PR #31, opened 2022-06-22
further simplifying the getting started experience

TyDunn:


QUICKSTART.md
```diff
@@ -6,194 +6,77 @@
...
-            c.name as child_name, c.id as child_id, c._pos as child_order_in_list
-        from `{schema_prefix}_example.json_doc` as p
-        left join `{schema_prefix}_example.json_doc__children`  as c
-            on p._record_hash = c._parent_hash
-    """
-    run_query(query)
-```
-
-b. See results like the following
+See results of querying the Google BigQuery table
```
rudolfix, line 79: Near `SQL result:` I'd explain that this is the result of joining two tables above on autogenerated keys: `p._record_hash = c._parent_hash`. otherwise it's hard to guess what is what. maybe leaving the SQL queries themselves makes sense here?

rudolfix (changes requested):
please consider explaining the SQL queries at the end more, see comment

QUICKSTART.md
TyDunn: Very much agree. I did that

PR #49, opened 2022-07-19
new pipeline api

rudolfix:


experiments/spotify/1_no_code_consume_source/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```
+pip install python-dlt
+pip install python-dlt-contrib
+
+dlt init
+
+vi .dlt/secrets
+
+dlt source info dlt.contrib.sources.spotify
```
TyDunn, line 12: I believe you mention this below, but can this just be aliased as `spotify` or at least `sources.spotify`

experiments/spotify/1_no_code_consume_source/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+> displays the source docstring and parameters to be passed - all from function signature!
+
+dlt load --to bigquery --dataset spotify dlt.contrib.sources.spotify --data-dir my_data --tables library,history
```
TyDunn, line 16: how does this look if you are using the Spotify API instead of a data directory with JSON files from Spotify?

TyDunn, line 16: @adrianbr is a CLI like this really that much better for folks than a short Python file? This command is quite unwieldy

experiments/spotify/1_no_code_consume_source/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+
+## How does it work
+
+1. `dlt init` created default pipeline folder structure with empty `secrets` and `config` where options are present but commented out. `dlt init` can take a parameter which is a name of the workflow ie. `consume` to consume source, `pipeline` to write a new pipeline, `source` to develop source
```
TyDunn, line 21: what does it mean to consum a source vs. write a new pipeline vs. develop a source? We need to better define these things and align on them. Based on the conversation this morning, I think we have different definitions here

experiments/spotify/1_no_code_consume_source/README.md
```diff
@@ -0,0 +1,36 @@
...
+## dlt load
+
+How does it work
+1. we tell the function to load to bigquery into dataset `spotify`
+2. we tell it where source module is `dlt.contrib.sources.spotify` - we could figure out for sure how to use an alias ie `sources.spotify` as well
+3. fire will figure out the CLI for the module above for the function signature (function must be decorated by @source)
+4. we pass the same parameters we would pass in python. in this case this is `data-dir` and `tables`
+5. as the source returns iterator or iterators of iterators we know what to do: we'll extract (in parallel if possible), unpack and load atomically and with retries
+
+We could use `.dlt/config`
```
TyDunn, line 36: for what?

experiments/spotify/2_consume_source/README.md
```diff
@@ -0,0 +1,20 @@
...
+3. with `dlt init pipeline [pipeline_name]` see also README for `consume_source` case
+
+In our case we do
+```
+dlt init consume spotify_pipeline
+```
+
+which will create folder structure and `spotify_pipeline.py` with necessary imports and maybe some code commented out...
+
+## Running the pipeline
```
TyDunn, line 15: what about credentials? in `.dlt/credentials`?

experiments/spotify/3_devlop_own_source/README.md
```diff
@@ -0,0 +1,80 @@
+# Developing a pipeline with a custom (own) source
+This is example of a pipeline that includes custom source. The pipeline will be developed interatively by letting the user to modify schemas and run it again to apply changes.
+At the end we'll consume a few custom sources to show how pipeline with many sources could work.
+
+## Pipeline init
+
+Please read READMEs for `no_code_consume_source` and `consume_source` on options to initialize pipeline. In this case we do
+```
+dlt init custom-source spotify_pipeline
```
TyDunn, line 9: We need to align on names and better delineate on custom sources for yourself vs. custom sources for others. I have been calling the first 'snippets' (though I kind of like 'scripts' that use the DLT library better) and the second 'sources'

experiments/spotify/4_develop_contrib_source_from_own_source/spotify/spotify_source.py
```diff
@@ -0,0 +1,64 @@
+import os
+from dlt import source as src
+
+# we start with the schema we developed in `own source` workflow.
+# we let people pass the directory with the spotify data
+@source(schema=src.load_schema("schema.yml"))
```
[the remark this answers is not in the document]
TyDunn: educating about two different things, when they are so similar adds a lot of overhead / complexity

experiments/spotify/5_develop_contrib_source_from_scratch/README.md
```diff
@@ -0,0 +1,44 @@
+# Developing a contrib source from scratch
+
+We develop new source with the intention to be importable. Probably we are a software developer or an agency and our source is something really popular ie. Hubspot. The workflow is not that different from `develop contrib source from own source` but:
+
+1. The source structure would be much cleaner and decorators will be used, not `as_table`
```
TyDunn, line 5: I don't understand why we don't just use decorators in 4. Can you help me understand?

experiments/spotify/1_no_code_consume_source/README.md
rudolfix: I forgot :D

experiments/spotify/4_develop_contrib_source_from_own_source/spotify/spotify_source.py
```diff
@@ -0,0 +1,64 @@
...
+                    item["searchTime"] = item["searchTime"][:-5]
+            return item
+
+        if file.startswith('Playlist'):
+            # we yield the data as table. we just provide the table name, the rest comes from the schema we loaded earlier
+            yield as_table(map(fix_timestamps, data['playlists']), table_name='Playlist')
+        elif file.startswith('SearchQueries'):
+            yield as_table(map(fix_timestamps, data), table_name='SearchQueries')
+        elif file.startswith('StreamingHistory'):
+            yield as_table(iter(data), table_name='StreamingHistory')
```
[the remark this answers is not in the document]
rudolfix: ok - this is very important. a difference between flow 3 and 4

our sources are iterators and I want to keep them like that. this is the 1. design principle. but here I have a problem: I want to pass not only data but also metadata with the iterator. this metadata tells the dlt how to load the iterator into database. this is for example the table name. but I can define additional properties of such table here,
`as_table` is attaching this metadata to data. this allows the flows 1 and 2 to be so simple. it also enables adrian to write those helper functions

from that point of view
1. in flow 3 we tell dlt what to do with the data explicitly in extract method. this is very natural in self-contained pipeline
2. in flow 4 we attach this metadata to the iterator so the user does not need to know for example that we need to replace `songs` table but `history` is loaded incrementally with specified merge key of given type...

experiments/spotify/4_develop_contrib_source_from_own_source/README.md
```diff
@@ -0,0 +1,51 @@
...
+* `spotify` folder with contrib source structure (see folder). The `__init__` and `spotify_source.py` will be created by `dlt init` and contain basic skeleton for the source
+* `tests` folder where you should place your tests and mock data
+* `source_requirements.txt` where you can put additional dependencies for the source
+
+## Contrib source code
+We re-use code from the demo with minimal modifications.
+
+1. Already developed `schema.yml` is used as source schema
+2. The code stays intact, except doing `extract` we yield the data (see code, it's simple)
+3. We provide minimal number of hints, instead relying on the already developed schema!
```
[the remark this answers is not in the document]
rudolfix: I'm trying to describe how I ported the code from `own source` to here. in other words
1. I kept the schema
2. I didn't change the code structure at all thanks to using `yield`
3. what was passed to extractor is now passed to `as_table`

experiments/spotify/4_develop_contrib_source_from_own_source/spotify/spotify_source.py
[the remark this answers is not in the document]
TyDunn: I think we should discuss further. If we clearly lay out the differences between 3 and 4, then I think we can decide if it is better or worse to educate people about one harder thing (4) or one easier thing (3) AND one harder thing (4)

experiments/spotify/4_develop_contrib_source_from_own_source/spotify/spotify_source.py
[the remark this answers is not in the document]
TyDunn: That makes sense, but why is it called `as_table`?

experiments/spotify/2_consume_source/README.md
rudolfix: yes, or any other name we pick

experiments/spotify/2_consume_source/spotify_pipeline.py
```diff
@@ -0,0 +1,18 @@
+from dlt import pipeline as p
+from dlt.sources import spotify
+
+# you can configure your pipeline explicitely
+# if not configured, sensible defaults will be used: the current file name will become the name of the pipeline
+# the working dir goes to tmp and is recreated each time pipeline is run
```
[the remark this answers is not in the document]
rudolfix: the consumer workflow does not change the schema. (workflows 1 and 2). that's the whole point. you do not work iteratively. you just load the data to warehouse (and do streamlit app or get slice of data as dataframe). the source has complete schema. this is the case Adrian is working on.

experiments/spotify/2_consume_source/spotify_pipeline.py
```diff
@@ -0,0 +1,18 @@
...
+# this also writes `.dlt/config` file
+# dlt.config("spotify_pipeline", work_dir="~/pipelines/spotify")
+
+# this get a list of iterators with associated schema
+data = spotify.get_data("/path/to/data", tables=["songs", "history"])
```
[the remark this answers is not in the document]
rudolfix: this is another good question. how do we name methods that return data. are this name always the same? can a source have several alternative functions that return data? in case of spotidy you could have

```
@source()
def spotify_file(data_dir, tables): ...

@source()
def spotify_api(user_id, credentials, tables): ...

```

experiments/spotify/3_devlop_own_source/README.md
```diff
@@ -0,0 +1,80 @@
...
+- schema - string or object: you can pass a ready schema or a schema name in whcih tables will be created
+
+## Working iteratively on schema
+
+1. you start with empty `schemas` folder
+2. you provide sensible and minimal setting for your tables into `extract` method
+3. you run your script
+4. in `schemas` folder you see inferred schemas: from your hints in `extract` and from the data itself
+5. modify the schemas and run your script again
+6. remove the `dry_run` option to actually load data
```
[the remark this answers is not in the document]
rudolfix: nothing else comes to my mind right now

experiments/spotify/3_devlop_own_source/spotify_pipeline.py
```diff
@@ -0,0 +1,49 @@
...
+    else:
+        continue
+
+# now we have a spotify source, let's say we want to add some related annotations from google sheets. this happens via
+# source you get from contrib
+# this you add to the pipeline once your custom source is fully developed - see README
+
+# put annotation in separate schema. annotations data will load into `spotifypipeline_annotations`
+# we also limit nesting of the schema
+ann_schema = Schema("annotations", max_nesting=2)
```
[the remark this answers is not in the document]
rudolfix: we convert nested json into relational form. a nested element is a separate table that is a child of the root document and so on (recursively). this parameter tells json normalizer to stop at certain depth and convert the nested json into STRING or JSON in the detabase.

metabase data is a good example why you want to do that @adrianbr 

max_nesting=0 sends intact documents into database as string (didn't try that but it should work)

experiments/spotify/4_develop_contrib_source_from_own_source/spotify/spotify_source.py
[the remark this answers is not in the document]
rudolfix: we can rename all decorators and helper functions... we should have a consistent naming.

rudolfix:
a few open questions mostly about default settings and behaviours

0. we need to have a glossary and consistent naming. I will propose something in https://www.notion.so/Glossary-of-DLT-usage-6a333efd4c894a2d9976e5e5a8e8738f

1. **sync schema** vs **in schema + out schema**. I have `sync_schema_folder` from where dlt will get schemas and export every change immediately.  in theory you should not have any of your changes overwritten as they are always imported. @adrianbr in your helper we have the `in` and `out` schemas which are separate files. so you must manually copy the `in` to `out` if dlt changed `in`. why it is like that? should we have separate folders for in and out or sync one will do?

2. what should dlt do when one of the tables does not load due to terminal (non recoverable) error? Such error should not be retried - typically your data is wrong. Currently: - **no exception is thrown**. - log message is created - failed data is preserved within the pipeline - there's a helper method that returns all the failed tables from the load. Should we do the above but throw the exception? when you develop a source that would be better setting. on production though it will just fail the worker node without any chance for it to recover.  

3. should the pipeline have a name? this is a deep question: please read my comment here why we need them - connected to the dataset name to which the data will load: https://github.com/scale-vector/dlt/issues/51

4. can two pipelines with different schemas load into the same dataset? (currently: there are no checks against that so if there are overlapping table names and columns are incompatible then loading will fail and schema version control will be messed up)

5. naming of the schema(s) problem. Imagine this: I have `spotify` source that I import from `contrib` and I also ad hoc import annotations from google sheets. how the data should be organized and loaded? (assuming pipeline is named `my_music`)
- Option 1: annotations (several tables corresponding to tabs in google sheets) go into `my_music`  because **default schema has no name and we assign all the ad hoc tables into default schema** and spotify data go into `my_music_spotify` (because source has a named schema `spotify`)
- Option 2: everything goes into `my_music` because all the tables are forced into default schema and names provided by `contrib` are ignored (if not requested otherwise). there are risks of conflicting table names though.
which one is better?

6. when we develop a source we often change the schema in a way that corresponding dataset (or tables within) must be dropped. @adrianbr @TyDunn what should be desired behavior
- user must take care herself, manually drop or load into new dataset everytime
- as above but data will not load at all if there's any such conflict
- DLT will drop and re-create all tables (it knows about - so your "manual" tables will be spared) and then load data anew
- DLT will drop the whole dataset with all tables and re-load it

7. The same as above but on production? What if someone changes the database schema directrly and our schema is incompatible? You have the same options as above plus
- DLT schema will be synced to database schema (but then we get into the huge problem of having incompatible data in JSON)

experiments/spotify/1_no_code_consume_source/README.md
rudolfix: @adrianbr are you sure this is our guy? someone that cannot save and run python script will not be able to run pipelines. he would be able only to experiment in notebook but then what's next?

experiments/spotify/1_no_code_consume_source/README.md
rudolfix: let's build a glossary more because you clearly think about pipelines differently. IMO pipeline can have many sources with different schemas (you call them manifests) and set of python transformations (ie to clean up the data in python). in general I see the pipeline not as blackbox but as an api where I have control over whats going on inside, not just input and output

experiments/spotify/1_no_code_consume_source/README.md
rudolfix: hmmm to me preference files is decade old pattern (probably from UNIX in 70s) to set the local defaults. every unix tool uses that and that includes the ML stack (streamlit, jupyter) and super advanced devops stuff (ie. Kubernetes, AWS tooling)

experiments/spotify/4_develop_contrib_source_from_own_source/spotify/spotify_source.py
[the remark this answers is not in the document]
rudolfix: adrian, good question on naming the sources. IDK. as long as decorators are properly used to mark the method returning data we could have any name we want and still be able to automatically discover and run the source (and yes in that case method is named spotify, this also names the schema)

experiments/spotify/5_develop_contrib_source_from_scratch/README.md
rudolfix: again I do not agree IMO decorators are a part of every on-framework - they let you keep your current code but us it in a new way. ie caching decorators are quite popular because
1. they hide a lot of complexity for the user
2. give the developer huge amount of control over the decorated code!

experiments/spotify/3_devlop_own_source/README.md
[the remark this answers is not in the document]
rudolfix: > point 5 is where stuff will break because it's a schema and IDK what my options are.
> 
> So we need to design a good way to recover - either we auto create a backup we can restore, or offer to overwrite the broken schema. If most of the iteration happens in the extract method, it's easier to test and run iteratively IMO.

see comment - I ask exactly for this behaviour in the last comment, see our dev channel

adrianbr:
> 
1. I was worried about manually editing the schema and breaking it, hence the "out" was a backup before editing it for "in" I will try it without this.
2. i would raise the error. There might be recovery built into the pipeline differently.
3. the identifier for a pipeline is 
- source (like a hash of the host/ credentials for instance)
- destination  (like a hash of the host/ credentials for instance)
I do not think a name in itself is useful.
4. Yes there are 2 common patterns: either load each thing to its own schema, or load all to "raw" with sourcename_tablename as table prefix. I think the former is most popular and most tools like 5tran do it that way too.

5. I would expect to have 2 pipelines each with their own config. One from contrib, and one i did myself. i would not plug my source into a perfectly good contrib pipeline that someone else maintains - i would run each separately. If the annotations are related to spotify, I would put it all in the spotify schema. I would not create a "my music" schema unless I have more music than spotify. If the annotations are a generic music annotation process, then perhaps I wouldhave a music and an annotation schema and load spotify to music and annotation to annotation

6. 1 or 2  with a message that says 'if you want to replace the data set during development, pass the flag "replace_dataset" = True. I would not default to anything that drops any data. Data drop is scary, people often do not have/use a test env.

7. same. as 6 as 6 will often happen directly on production anyway. we load to "raw" and the staging -> model layers prevent from further schema corruption. But we should never cause data loss.

You can also assume many schemas to be non-persistent, except perhaps the raw ingestion schemas. It's a pattern to deploy schema as needed so you can have it as code and not have to manage database migrations separately from your code deployment. Otherwise, synching schemas with code running on a schedule is a nightmare.

PR #85, opened 2022-11-17
first draft of glossary terms

TyDunn:


docs/website/docs/glossary.md
```diff
@@ -4,14 +4,37 @@ sidebar_position: 6
 
 # Glossary
 
-## Pipeline
+## [Resource](./walkthroughs/create-a-pipeline.md)
 
-## Run
+  If the source is an API, then the resource is an endpoint in that API. If the source is a
```
rudolfix, line 9: @matthauskrzykowski says that the best metaphor for source vs. resource is the spreadsheet (source) and a tab in it (resource). Each tab has a set of columns typically each with data of a given type. A spreadsheet groups many such tabs.

rudolfix (changes requested):
OK it's a first draft. we have a glossary. maybe you can copy from there?

TyDunn:
> OK it's a first draft. we have a glossary. maybe you can copy from there?

I did copy from there at first. However, many of the definitions were 1) inaccessible to folks outside the org and / or 2) outdated, so these are my proposed updates

PR #99, opened 2022-11-28
first draft of destinations

TyDunn:


docs/website/docs/destinations.md
```diff
@@ -0,0 +1,61 @@
...
+You will then need to install the necessary dependencies for Postgres by running
+```
+pip install -r requirements.txt
+```
+
+Finally, you will want to edit the `dlt` credentials file with your connection info
+```
+open ./.dlt/secrets.toml
+```
+
```
rudolfix, line 45: The easiest way to setup `dlt` is to have a dedicated database for the loading. In that case you do the following from the `psql` cli or query editor:

1. Create new database `CREATE DATABASE dlt_data`  -- put it into `secrets.toml`
2. Create new user, set password `CREATE USER loader WITH PASSWORD 'loader';`  -- change it to your password and then put it into `secrets.toml`
3. Set as database owner (we could set lower permission) `ALTER DATABASE dlt_data OWNER TO loader`

It is entirely possible to set up more restrictive credentials. For example you can create the database schema into which you plan to load the data and give the `loader` user access to do the DDL and DML queries only within that schema

adrianbr (approved):

Issue #126, opened 2023-02-12
Nullable column causes pipeline to fail in certain cases


steinitzu:
Hi hi

I'm just starting my contributions to dlt and ran into my first issue.

It happens when the column is `null` in first >=5000 rows (5000 matches buffer writer limit),
then a following row which has a non-null value for the column causes the pipeline to fail.

This pipeline reproduces the issue:

```python3
import dlt
import random

@dlt.source
def dummy():
    return [table_rows()]


@dlt.resource(write_disposition='append')
def table_rows():
    for _ in range(5000):
        yield dict(display_name=None, random_number=random.random())
    yield dict(display_name='Dummy name', random_number=random.random())


def main() -> None:
    p = dlt.pipeline(
        pipeline_name='dummy', destination='postgres', dataset_name='dummy_data', full_refresh=False
    )

    info = p.run(dummy())
    print(info)


if __name__ == '__main__':
    main()
```

Traceback:

```
Traceback (most recent call last):
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/common/runners/pool_runner.py", line 79, in run_pool
    run_metrics = run_f.run(cast(TPool, pool))
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/normalize/normalize.py", line 257, in run
    self.spool_schema_files(schema_name, list(files_in_schema))
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/normalize/normalize.py", line 235, in spool_schema_files
    self.spool_files(schema_name, load_id, map_parallel_f, files)
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/normalize/normalize.py", line 195, in spool_files
    total_items, schema_updates, chunk_files = map_f(schema, load_id, files)
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/normalize/normalize.py", line 171, in map_single
    processed_chunk = Normalize.w_normalize_files(
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/normalize/normalize.py", line 120, in w_normalize_files
    load_storage.close_writers(load_id)
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/common/storages/data_item_storage.py", line 33, in close_writers
    writer.close_writer()
  File "/home/steinthor/projects/dlt-adrian/dlt/dlt/common/data_writers/buffered.py", line 77, in close_writer
    self._flush_and_close_file()
[...]

PR #127, opened 2023-02-12
Fix current columns and new columns referencing the same object

steinitzu:
This solves https://github.com/dlt-hub/dlt/issues/126 for me

### DIFF
--- a/dlt/common/data_writers/buffered.py
+++ b/dlt/common/data_writers/buffered.py
@@ -1,5 +1,7 @@
 from typing import List, IO, Any
 
+from copy import deepcopy
+
 from dlt.common.utils import uniq_id
 from dlt.common.typing import TDataItem, TDataItems
 from dlt.common.data_writers import TLoaderFileFormat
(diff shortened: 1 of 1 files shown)


dlt/common/data_writers/buffered.py
```diff
@@ -54,7 +56,7 @@ def write_data_item(self, item: TDataItems, columns: TTableSchemaColumns) -> Non
         if self._writer and not self._writer.data_format().supports_schema_changes and len(columns) != len(self._current_columns):
             self._rotate_file()
         # until the first chunk is written we can change the columns schema freely
-        self._current_columns = columns
+        self._current_columns = deepcopy(columns)
```
rudolfix, line 59: your fix is correct as explained in the issue. I'm merging the PR and then will add another PR with tests. I will also change the `deepcopy` to shallow one, deepcopy is expensive and this function is often called

rudolfix (approved):

dlt/common/data_writers/buffered.py
steinitzu: Awesome! Thanks. Wasn't sure how the object gets mutated so `deepcopy` was the safer bet, better that it's not needed

PR #146, opened 2023-02-22
getting started: bigquery --> duckdb

TyDunn:
Changes getting started to use DuckDB rather than Google BigQuery

### DIFF
--- a/docs/website/docs/getting-started.md
+++ b/docs/website/docs/getting-started.md
@@ -10,35 +10,47 @@ Please make sure you have [installed `dlt`](./installation.mdx) before getting s
 
 ## 1. Initialize project
 
-Create a `dlt` project with a pipeline that loads data from the chess.com API to Google BigQuery by running:
+Create a `dlt` project with a pipeline that loads data from the chess.com API to DuckDB by running:
 
 ```
-dlt init chess bigquery
+dlt init chess duckdb
 ```
 
-Install the dependencies necessary for Google BigQuery:
+Install the dependencies necessary for DuckDB:
 ```
 pip install -r requirements.txt
 ```
 
-## 2. Set up Google BigQuery
+## 2. Run pipeline
 
-Follow [BigQuery setup](./destinations#google-bigquery) to create the service account credentials you'll need to add to `.dlt/secrets.toml`.
-
-## 3. Run pipeline
-
-Run the pipeline to load data from the chess.com API to Google BigQuery by running:
+Run the pipeline to load data from the chess.com API to DuckDB by running:
 ```
 python3 chess.py
 ```
 
-Go to the [Google BigQuery](https://console.cloud.google.com/bigquery) console and view the tables
-that have been loaded.
+## 3. Query DuckDB database
+
+Create a `query.py` file where you can write SQL queries:
+```
+import duckdb
+
+conn = duckdb.connect('quack.duckdb')
+conn.sql('SELECT COUNT(*) FROM games').show()
+```
+
+Run `query.py` to run the script and execute the SQL query to see the number of games:
+```
+python3 query.py
+```
+
+Learn more about how to query DuckDB [here](https://duckdb.org/docs/sql/introduction#querying-a-table).
+
+Read more about the default configuration of the DuckDB destination [here](destinations.md#destination-configuration).
 
 ## 4. Next steps
 
 Now that you have a working pipeline, you have options for what to learn next:
-- Try loading data to a [different destination](./destinations) like Amazon Redshift, Postgres or DuckDb
+- Try loading data to a [different destination](./destinations) like Google BigQuery, Amazon Redshift, or Postgres
 - [Deploy this pipeline](./walkthroughs/deploy-a-pipeline), so that the data is automatically
 loaded on a schedule
 - [Create a pipeline](./walkthroughs/create-a-pipeline) for an API that has data you want to load and use
--- a/docs/website/docs/getting-started.md
+++ b/docs/website/docs/getting-started.md
@@ -38,7 +38,7 @@ conn = duckdb.connect('quack.duckdb')
 conn.sql('SELECT COUNT(*) FROM games').show()
 ```
 
-Run `query.py` to run the script and execute the SQL query to see the number of games:
+Run `query.py` to execute the script and SQL query to see how many total games were loaded from the API:
 ```
 python3 query.py
 ```
--- a/docs/website/docs/getting-started.md
+++ b/docs/website/docs/getting-started.md
@@ -28,19 +28,13 @@ Run the pipeline to load data from the chess.com API to DuckDB by running:
 python3 chess.py
 ```
 
-## 3. Query DuckDB database
+## 3. Query the data in DuckDB
 
-Create a `query.py` file where you can write SQL queries:
-```
-import duckdb
+Your data has been loaded into `chess.duckdb`, which has been added to your project directory.
 
-conn = duckdb.connect('quack.duckdb')
-conn.sql('SELECT COUNT(*) FROM games').show()
+Run this command to see a quick overview of the data loaded:
 ```
-
-Run `query.py` to execute the script and SQL query to see how many total games were loaded from the API:
-```
-python3 query.py
+dlt pipeline chess show
 ```
 
 Learn more about how to query DuckDB [here](https://duckdb.org/docs/sql/introduction#querying-a-table).
--- a/dlt/helpers/streamlit.py
+++ b/dlt/helpers/streamlit.py
@@ -277,8 +277,8 @@ def display(pipeline_name: str) -> None:
     pipeline = dlt.attach(pipeline_name)
 
     pages: Dict[str, AnyFun] = {
-        "Pipeline info": write_load_status_page,
         "Explore data": write_data_explorer_page,
+        "Pipeline info": write_load_status_page,
     }
 
     st.title(f"Show {pipeline_name} pipeline")
--- a/docs/website/docs/command-line-interface.md
+++ b/docs/website/docs/command-line-interface.md
@@ -39,4 +39,4 @@ Follow the [Deploy a pipeline](./walkthroughs/deploy-a-pipeline.md) walkthrough 
 dlt pipeline <pipeline name> show
 ```
 
-Generates and launches a simple [Streamlit](https://streamlit.io/) app where you can inspect your pipeline state, loading status / stats and also explore generated schemas and data in the destination's dataset. Requires `streamlit` to be installed.
+Generates and launches a simple [Streamlit](https://streamlit.io/) app that you can use to inspect the schemas and data in the destination as well as your pipeline state and loading status / stats. Requires `streamlit` to be installed.
--- a/docs/website/docs/installation.mdx
+++ b/docs/website/docs/installation.mdx
@@ -8,7 +8,7 @@ import Tabs from '@theme/Tabs';
 
 ## Set up environment
 
-### Make sure you are using **Python 3.8, 3.9, or 3.10** and have `pip` installed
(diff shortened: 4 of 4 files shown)


docs/website/docs/getting-started.md
```diff
@@ -10,35 +10,47 @@ Please make sure you have [installed `dlt`](./installation.mdx) before getting s
...
-python3 chess.py
+import duckdb
+
+conn = duckdb.connect('quack.duckdb')
+conn.sql('SELECT COUNT(*) FROM games').show()
 ```
 
-Go to the [Google BigQuery](https://console.cloud.google.com/bigquery) console and view the tables
-that have been loaded.
+Run `query.py` to run the script and execute the SQL query to see the number of games:
```
TyDunn, line 41:
```suggestion
Run `query.py` to execute the script and SQL query to see how many total games were loaded from the API:
```

docs/website/docs/getting-started.md
```diff
@@ -10,35 +10,47 @@ Please make sure you have [installed `dlt`](./installation.mdx) before getting s
...
-## 3. Run pipeline
+## 3. Query DuckDB database
 
-Run the pipeline to load data from the chess.com API to Google BigQuery by running:
+Create a `query.py` file where you can write SQL queries:
 ```
-python3 chess.py
+import duckdb
+
+conn = duckdb.connect('quack.duckdb')
```
rudolfix, line 37: Yeah we'll move the `quack` to the working folder - so you can do the above. There's however a way that will work with any destination and does a lot of things behind the scenes for you. here's the script. take what you want from there
```python
import dlt

# this will use the same configuration as in `chess.py` and open the connection seamlessly. it is also destination independent way
# also:
# - it will let you query tables without schema prefix 
# - will apply the same settings to duckdb you had when loading
# - will inject s3 and other credentials so you can work with s3 bucket automatically
# - etc.

with dlt.attach(pipeline_name="chess").sql_client() as client:
    # this works on any destination
    rows = client.execute_sql('SELECT COUNT(*) FROM players_games')
    print(rows)
    
    # this works on any destination
    with client.execute_query('SELECT * FROM players_games LIMIT 10') as query:
        
        # print(query.fetchall())
        # this is duckdb specific
        print(query.fetch_df())

    # or call the duckdb directly
    client.native_connection.sql('SELECT COUNT(*) FROM players_games').show()
    # next dlt version will allow for this (no native_connection necessary)
    # client.sql('SELECT COUNT(*) FROM players_games').show()
```

rudolfix (changes requested):

docs/website/docs/getting-started.md
```diff
@@ -10,35 +10,47 @@ Please make sure you have [installed `dlt`](./installation.mdx) before getting s
...
-## 2. Set up Google BigQuery
+## 2. Run pipeline
 
-Follow [BigQuery setup](./destinations#google-bigquery) to create the service account credentials you'll need to add to `.dlt/secrets.toml`.
+Run the pipeline to load data from the chess.com API to DuckDB by running:
+```
+python3 chess.py
+```
 
-## 3. Run pipeline
```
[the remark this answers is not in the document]
TyDunn: This is pretty cool. It required me to `pip install pandas` and `pip install streamlit` though. Can we make those part of the things installed with dlt and / or requirements.txt?

docs/website/docs/getting-started.md
TyDunn: I am quite skeptical that we should be recommending folks to interact with `DuckDB` through `dlt`. The things you mentioned are nice, but I'm not sure they are worth the complexity / confusion they would introduce to our users (at least for `DuckDB`)

docs/website/docs/getting-started.md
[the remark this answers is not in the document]
rudolfix: both streamlit and pandas are huge packages that with many weird dependencies. I'd rather keep them out of the core. What I can do
1. make another extra so you can install streamlit  with `python-dlt[streamlit]`
2. the `dlt show` command can install streamlit + pandas for the user. if those are missing we could as if they should be added and install seamlessly.

what do you think?

docs/website/docs/getting-started.md
rudolfix: I'm totally for it if we move this into a place dedicated to `duckdb`. if this is getting started section, we should IMO show stuff that works with any destination.
Do we expect mostly duckdb crowd here? if this is so then keep your original script, just fix the table name you select from. we may put something more universal later

I'll be pushing the db location fix on monday

rudolfix (changes requested):

docs/website/docs/getting-started.md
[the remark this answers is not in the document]
TyDunn: Makes sense. Let's go with #2. We should also do #155 when we do that

docs/website/docs/getting-started.md
TyDunn: That would make sense to me if the way the user should interact with their destinations is through `dlt`. I don't think this should be the recommended way. I think `dlt` should be a simple, easy data loading library and not try to do too much. We can have this functionality since there are use cases where you need to query in order to load, but I don't think we should train folks to use `dlt` interfaces for their daily querying, their transformations, etc.

I would remove the script and just use the `dlt show` part for now

docs/website/docs/getting-started.md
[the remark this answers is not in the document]
rudolfix: https://github.com/dlt-hub/dlt/issues/158

docs/website/docs/getting-started.md
rudolfix: This is old UNIX principle of doing just one thing and doing this well :) `sql_client` is used mostly internally or to quickly test/fetch data in the same script that ran pipeline. people should use something else for transformations.

just note that `duckdb` is special because you see your database file and it is easy to connect. the user of `duckdb` also comes here for a database they love. but think about bigquery user which has no clue about the database and querying data in python. to actually use the data:
- you need to find which of hundreds objects is actual google sql client
- learn a special syntax that is different from python dbapi syntax (or dig out that there's dbapiconnection in google namespace)
- learn how to authenticate with service json, load the file and pass it somewhere
- that to query your dataset you need to use project_id.dataset_name.table_name enclosed in non standard quotation
- that there's a special `to_dataframe` helper that will execute query and convert it to a dataframe (if you install special deps)

this is all eliminated by dlt that is reusing the credentials you are supplied by loading and will return you an sql_client which is just bigquery client with a few standard methods attached

long rant heh but there are situations that sql_client makes a lot of sense

rudolfix (dismissed):
let's remove the script and merge. tickets with various `dlt` improvements are created

TyDunn:
@rudolfix I am getting this error when I try to follow the getting started and run `python3 chess.py`:
```
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/destinations/sql_client.py", line 138, in _wrap
    return f(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/destinations/duckdb/sql_client.py", line 27, in open_connection
    self._conn = self.credentials.borrow_conn(read_only=self.credentials.read_only)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/destinations/duckdb/configuration.py", line 37, in borrow_conn
    self._conn = duckdb.connect(database=self.database, read_only=read_only, config=config)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: connect(): incompatible function arguments. The following argument types are supported:
    1. (database: str = ':memory:', read_only: bool = False, config: dict = None) -> duckdb.DuckDBPyConnection

Invoked with: kwargs: database='/users/ty/.dlt/pipelines/chess/quack.duckdb', read_only=False, config=None

During handling of the above exception, another exception occurred:

  File "/Users/ty/Documents/dlthub/getting-started/chess.py", line 138, in <module>
    info = pipeline.run(
           ^^^^^^^^^^^^^
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 101, in _wrap
    step_info = f(self, *args, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 132, in _wrap
    return f(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 422, in run
    self.sync_destination(destination, dataset_name)
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 75, in _wrap
    rv = f(self, *args, **kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 464, in sync_destination
    remote_state = self._restore_state_from_destination()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 972, in _restore_state_from_destination
    job_client.sql_client.open_connection()
  File "/Users/ty/Documents/dlthub/getting-started/env/lib/python3.11/site-packages/dlt/destinations/sql_client.py", line 140, in _wrap
    raise DestinationConnectionError(type(self).__name__, self.dataset_name, str(ex), ex)
dlt.destinations.exceptions.DestinationConnectionError: Connection with DuckDbSqlClient to dataset name chess_data failed. Please check if you configured the credentials at all and provided the right credentials values. You can be also denied access or your internet connection may be down. The actual reason given is: connect(): incompatible function arguments. The following argument types are supported:
    1. (database: str = ':memory:', read_only: bool = False, config: dict = None) -> duckdb.DuckDBPyConnection

Invoked with: kwargs: database='/users/ty/.dlt/pipelines/chess/quack.duckdb', read_only=False, config=None
```

rudolfix:
@TyDunn you had another lucky strike to test just after another duckdb release. Last time I've limited the possible upgrade to minor version. Nevertheless they've changed their python interface in incompatible way with going from 0.7.0 to 0.7.1.

I'll try to bugfix it. I'm still against freezing the version completely but if the above happens again I will

docs/website/docs/getting-started.md
```diff
@@ -10,35 +10,41 @@ Please make sure you have [installed `dlt`](./installation.mdx) before getting s
...
+## 3. Query the data in DuckDB
 
-Run the pipeline to load data from the chess.com API to Google BigQuery by running:
+Your data has been loaded into `chess.duckdb`, which has been added to your project directory.
```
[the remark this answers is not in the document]
rudolfix: yes. this #159 changes it.

TyDunn:
@rudolfix Loading is now working. However, I still am unable to use the `dlt pipeline chess show` command: #158

TyDunn:
@rudolfix once we have `dlt pipeline chess show` working again, then we can merge this

rudolfix (approved):

PR #159, opened 2023-02-27
fixes duckdb database default location

### LINKED ISSUE #148: duckdb database location challenging to find and understand
Right now, after running a pipeline that loads to DuckDB, you get a message that says `the duckdb destination used duckdb:////users/<username>/.dlt/pipelines/chess/quack.duckdb location to store data`.

As a new dlt user, I would have expected the DuckDB database to be in the same directory as my pipeline script. If we were to do that, we would need to make sure that the DB is not committed in Git, so we should probably also add it to the `.gitignore` file?

rudolfix:
closes #148 

also fixes connect problem in `duckdb` 0.7.1

@TyDunn doc change is here https://github.com/dlt-hub/dlt/pull/159/files#diff-ee372fc4d5abe4107cca1a8ad146c3a2a87301168c13440cd2303618ff011bb0R158

when merged I can release a `duckdb` fix for yesterday's upgrade


docs/website/docs/destinations.md
```diff
@@ -155,9 +155,9 @@ python3 chess.py
 
 ### Destination Configuration
 
-By default, a DuckDB database will be created inside the pipeline working directory with a name `quack.duckdb`. It is available in `read/write` mode via `pipeline.sql_client`.
+By default, a DuckDB database will be created in the current working directory with a name `quack.duckdb`. After loading, it is available in `read/write` mode via `with pipeline.sql_client() as con:` which is a wrapper over `DuckDBPyConnection`. See [duckdb docs](https://duckdb.org/docs/api/python/overview#persistent-storage) for details.
```
TyDunn, line 158: @rudolfix is the name `quack.duckdb` or `<pipeline>.duckdb`?

TyDunn (approved):

docs/website/docs/destinations.md
rudolfix: it should be pipeline name. I will fix it and let you review the docs again

rudolfix: @TyDunn done

TyDunn (approved):

PR #162, opened 2023-02-28
add docs

adrianbr:


docs/website/docs/glossary.md
```diff
@@ -6,28 +6,114 @@ sidebar_position: 6
...
 
-  If endpoints in an API are the resources, then the API is the source. If tabs in a spreadsheet
-  are the resources, then the source is the spreadsheet. If tables in a database are the resources, 
+  > **💡** If endpoints in an API are the resources, then the API is the source. If tabs in a spreadsheet
+  are the resources, then the source is the spreadsheet. If tables in a database are the resources,
   then the source is the database. A source is organized into one or more resources.
 
+
+
+A dlt Source is the code that handles getting data from an api.
```
rudolfix, line 52: what about google sheet, sql, jsonl and many other sources that are not APIs?

docs/website/docs/glossary.md
```diff
@@ -6,28 +6,114 @@ sidebar_position: 6
...
+        yield {'id':1}
+
+pipeline = dlt.pipeline(destination='bigquery',dataset_name='sql_database_data')
+
+info = pipeline.run(gen(10))
+
+print(info)
+```
+
+A pipeline’s unique identifier is the name. If no name is given, then the pipeline takes the name of the source or generator you pass to it.
```
rudolfix, line 112: it takes the default name from the currently executing module. all the defaults are explained in the general usage documentation and in docstrings

rudolfix (changes requested):
1. this is better than what we have now :)
2. in glossary we should define terms and provide examples, not the instructions how to write those
3. it looks like a General Usage page - there's one already written in the technical docs. should we rather add a new page and work together to merge this with the other information there?

TyDunn (changes requested):
Nice work! I agree we need more explanations of these key parts of `dlt`. However, I think the glossary should just be for 1-2 sentence definitions. I recommend that we create a new sidebar item `Concepts` and then add these three deep explanations there. The specific details look pretty good, so assuming you get approval from Marcin and make the structural changes I am requesting, I think you can merge. There are small style things that we can clean up over time. Most important is to get this info shipped and iterate quickly on it

adrianbr:
> 1. this is better than what we have now :)
> 2. in glossary we should define terms and provide examples, not the instructions how to write those
> 3. it looks like a General Usage page - there's one already written in the technical docs. should we rather add a new page and work together to merge this with the other information there?

I am not sure either.

adrianbr:
> 1. this is better than what we have now :)
> 2. in glossary we should define terms and provide examples, not the instructions how to write those
> 3. it looks like a General Usage page - there's one already written in the technical docs. should we rather add a new page and work together to merge this with the other information there?

Going through the general usage, it is very hard to comprehend - why am i reading, what problem am i trying to solve? I will not remember that I can find information in this because it's just a dump of many things. Perhaps general usage will be recycled into in-detail pages about the concepts?

PR #164, opened 2023-03-02
dbt docs, could still use some docs about transformers maybe

adrianbr:


rudolfix (commented):
looks good from my side. there are really good docstrings in the code, you may mention that if you want.

thanks 👍

rudolfix (approved):

PR #174, opened 2023-03-11
adds segment telemetry and sentry tracing

### LINKED ISSUE #154: add basic telmetry
it would be helpful if we anonymously tracked when data is loaded to different destinations

it should be clear to users that we are sending this telemetry and easy to opt out

rudolfix:
closes #154 

@adrianbr @TyDunn please review the documentation change https://github.com/dlt-hub/dlt/pull/174/files#diff-c356300d02eea95c26251ef3b5a4f558d95835fc292b77f86a9790fbc50a6074


TyDunn (dismissed):
I refactored the docs pages. Everything looks good to me, assuming they remain accurate.

Issue #147, opened 2023-02-22
create a version of `requests` that retries on certain HTTP statuses by default


rudolfix:
**Motivation**
1. Most of the pipelines do hundreds of requests, if there's no retry built in the whole pipeline will fail.
2. We want a blueprint for "source helper" - something that is so often used that is a part of dlt core

**Tasks**
1. implement retry session using `HTTPAdapter`: https://majornetwork.net/2022/04/handling-retries-in-python-requests/
2. allow to configure timeout as well.
3. provide a default session with sane timeout and retry parameters
4. provide configurable session like: https://github.com/bustawin/retry-requests
5. look at thread safety ie. here someone explains how to make a shared connection pool which is thread safe https://stackoverflow.com/questions/18188044/is-the-session-object-from-pythons-requests-library-thread-safe

Place code in `dlt.sources.helpers.requests` module and expose the default session as `requests` and configurable session as `requests_with_retry(...)`

the point of the above is to provide drop-in replacement for people making requests in `pipelines` repository, They'd just import from some other module but the remaining code could stay the same.

PR #178, opened 2023-03-13
Requests retrying session

steinitzu:
https://github.com/dlt-hub/dlt/issues/147  
This is the requests implementation using tenacity to manage retries. + session with `timeout` and `raise_for_status` options.  
Added some basic customization options for retries.    
Defaults are retrying on 5xx and 429 status codes, and connection and timeout errors, using exponential backoff or `retry-after` header for delay.   

The default settings can be used like requests normally:

```python3
from dlt.sources.helpers import requests

requests.request(...)
requests.get(...)
requests.post(...)
...
```

Currently this manages a thread local session behind the scenes.   

Custom options e.g.

```python3
from dlt.sources.helpers.requests import requests_with_retry, Session

def should_retry(response, exception):
    return response is not None and response.text == 'this is an error'

session = requests_with_retry(Session(timeout=120), max_attempts=20, status_codes=[500, 503], condition=should_retry)
```

TODO: Thread safe pool for the default and custom session. E.g. using global adapter: https://stackoverflow.com/questions/30748200/reusing-connections-in-django-with-python-requests/73094972#73094972


rudolfix:
@steinitzu this looks good, I will dive in the code tomorrow. are you able to push this again as a branch here and not a fork? the credentials are not shared. if you want to work from forks then tell me. I'll give you access to bigquery and redshift

steinitzu:
Ah yeah, will push again here. Forgot to change my default remote

PR #179, opened 2023-03-13
Requests retry session

steinitzu:
Same as https://github.com/dlt-hub/dlt/pull/178 without the fork :)


dlt/common/utils.py
```diff
@@ -258,3 +260,20 @@ def derives_from_class_of_name(o: object, name: str) -> bool:
     """Checks if object o has class of name in its derivation tree"""
     mro = type.mro(type(o))
     return any(t.__name__ == name for t in mro)
+
+
+
+@overload
+def to_seconds(td: None) -> None:
```
rudolfix, line 267: time related utils are in `time.py`. I think it's a better for the utils below

dlt/sources/helpers/requests/retry.py
```diff
@@ -0,0 +1,160 @@
...
+    def __init__(self, predicate: RetryPredicate) -> None:
+        self.predicate = predicate
+
+    def __call__(self, retry_state: RetryCallState) -> bool:
+        response = _get_retry_response(retry_state)
+        exception = retry_state.outcome.exception()
+        return self.predicate(response, exception)
+
+
+class wait_exponential_retry_after(wait_exponential):
```
rudolfix, line 57: this code looks good. why did you decide to write it?
1. we could probably re-use logic from `tenacity`?
2. why not to use `HTTPAdapter` which does exactly what you do here? https://majornetwork.net/2022/04/handling-retries-in-python-requests/
```
adapter = TimeoutHTTPAdapter(timeout=(3, 60), max_retries=Retry(total=5, backoff_factor=1.5, allowed_methods=False, status_forcelist=[429, 500, 502, 503, 504]))
```

rudolfix: there may be good reason to write that from scratch, I'm curious why :)

pyproject.toml
```diff
@@ -51,7 +51,6 @@ gitpython = ">=3.1.29"
 pytz = ">=2022.6"
 tqdm = ">=4.64.1"
 giturlparse = ">=0.10.0"
-reretry = ">=0.11.8"
```
rudolfix, line 54: let's keep re-retry for some time. we need to port all our pipelines to new session. I'll write ticket for that

dlt/sources/helpers/requests/api.py
```diff
@@ -0,0 +1,61 @@
...
+from dlt.sources.helpers.requests.session import Session
+from dlt.sources.helpers.requests.retry import requests_with_retry
+
+
+_local = threading.local()
+
+
+def _get_session() -> Session:
+    # Manage a thread local session with default settings
+    session: Optional[Session] = getattr(_local, 'session', None)
```
rudolfix, line 15: ok this is cool. we just need to upgrade this code to use a shared connection pool that is thread safe.

dlt/sources/helpers/requests/api.py
```diff
@@ -0,0 +1,61 @@
...
+    get = requests.get
+    post = requests.post
+    options = requests.options
+    head = requests.head
+    put = requests.put
+    patch = requests.patch
+    delete = requests.delete
+
+
+def request(*args, **kwargs):  # type: ignore
```
rudolfix, line 32: hmmm this looks like an overkill :)

why not create the default session in `helpers/__init__,py`
`requests = _get_session()` and then users with deal with session object directly.

this will OFC require to change the name of this module from requests to maybe `session`

so IDK. I'm just giving you an idea so we can get rid of most the glue code here...

rudolfix (changes requested):

dlt/sources/helpers/requests/api.py
rudolfix: solution above also preserves typing of get/post etc. methods

dlt/sources/helpers/requests/retry.py
steinitzu: Yeah it's basically the same for the common case.    
But I'm thinking from the POV of having to deal with wonky APIs that are not very standards compliant, or just have a couple of edge cases where the status code is ambigious. I usually need to add at least a little bit of custom logic to decide whether to retry so being able to inject some callback to inspect the response is super handy.  
E.g. think cases like where `400` code is used both for rate limit and "invalid input" kind of error. Or the extreme case, I work with one internal API which returns 200 on everything and error code is in the body.   

There is a lot of flexibility in tenacity that could be exposed as well for more exotic use cases. E.g. dynamic delays (say one error takes longer to recover than others).

dlt/sources/helpers/requests/api.py
steinitzu: I only did this because having one global session with multi threading is not recommended.  

Maybe a session factory which could have a global instance with default settings.  

So you call e.g. 
```python3
from dlt... import requests
requests.session.get(...)  # session is property to manage thread local
```

Though I'm hoping I'm able to make an actual thread safe `Session` implementation (not just the inner adapter/pool). That would be ideal so none of these wrappers are needed.

steinitzu:
Hey  @rudolfix  So I'm not finding any solid information on thread safety, and a lot of other frameworks are using separate sessions per thread so gonna say best to do the same.  

Considering now a slightly different API  with the `Client` class (see last commit) which uses a global connection pool and thread-local sessions internally.  
Added the `get/post/request/...` glue there so you have the same interface for default and custom settings, and you can still drill into the session itself if needed.  So not quite a drop in requests replacement, but almost. :)

```
# Default client
from dlt.sources.helpers.requests import client

client.get('https://...')

# Custom
from dlt.sources.helpers.requests import Client

client = Client(timeout=..., status_codes=[....])
client.get('https://...')
```

rudolfix:
approved! thanks for all the tests

PR #202, opened 2023-03-24
Allow changing `write_disposition` in the resource without dropping dataset

### LINKED ISSUE #193: allow to change `write_disposition` in the resource without dropping dataset
**Background**
We do not allow the `write_disposition` argument to be changed after the dataset is created with a corresponding yaml schema. The check is artificial and actually prevents our from doing what they want

**Tasks**
1. * [ ] write a test that reproduces the problem (add to tests in `loader/pipeline`)
2. * [ ] remove the check (this happens when we check if two table schemas can be merged)
3. * [ ] fix the other tests

burnash:
1. Removes the check in `dlt.common.schema.utils.diff_table()`
2. Sets the new `write_disposition` value if present.
3. Adds and fixes tests

Resolves #193


rudolfix (changes requested):
this looks good! thanks for catching that detail with merging the schema, 
one thing: could you re-submit the PR directly from the repo. fork does not have access to CI credentials so IDK if some tests are passing or not. thanks!

Issue #194, opened 2023-03-21
wrap "last value" incremental loading into resource decorator


rudolfix:
**Background**
Our users struggle with a version of incremental loading where 

**Requirements**
* the mechanism should work per resource, when using dlt state, the last value and other state elements should be kept per resource
* user should be able to tell `dlt` which field in the data returned by given resource contains last value. this can be nested field (but not a part of list). we could use subset of json path to specify the location
* user should be able to tell `dlt` the "last value" function to that selects the value to be saved to state. the default should be max. we should also provide "min" (the minimum of all values) "last" (the last value) or custom function
* user should have access to last value in the resource ie. to pass it to API endpoint or db query. the access should be intuitive (argument? dlt.current.last_value?). to do it right is tricky!
* if there's a primary key specified in the resource decorator (#151  - not done yet) we can use it to deduplicate
* optional: a way to override the value in the code

**Use cases**
We'll apply it to existing pipelines:
1. `sql_database` - https://github.com/dlt-hub/pipelines/blob/master/pipelines/sql_database/sql_database.py#L20
2. `github` - https://github.com/dlt-hub/pipelines/blob/master/pipelines/github/__init__.py#L76
3. 'hubspot' - https://github.com/dlt-hub/pipelines/blob/master/pipelines/hubspot/__init__.py#L83

**Tasks**
1. [x] propose a way to implement the above before writing code. then we finalize the ticket

Issue #180, opened 2023-03-14
allow to store datetimes and other basic types in pipeline state 


rudolfix:
**Background**
When implementing incremental loading the last value is often datetime. People struggle with converting it to strings in order to store those is the state and then use to filter out or retrieve records incrementally

**Objective**
`dlt` has a typed JSON dumper that will serialize and deserialize several native python objects including datetime, time and dates. Use it when reading and saving state.

typed dumper is backward compatible with the standard dumper. we we do not need to migrate existing state stored in destination

**Tasks**
1. * [x] use the `typed_dump` to save state. state serialization and deserialization is done in many different places in `state.py` and `pipeline.py`. so we need to clean it up and do util files to handle this and then convert all loads and dumps
2. * [x] typed load requires additional decoding. see `custom_pua_decode` and see `test_coerce_type_complex_with_pua`
3. * [x] add tests: that store and restore state locally and in destination. there's `test_restore_state.py`

PR #204, opened 2023-03-24
Rudimentary implementation for incremental

steinitzu:
Hey @rudolfix  This is a quick and dirty implementation https://github.com/dlt-hub/dlt/issues/194 , wanted to get a starting point to experiment from. This is not close to final.

Usage:

```python
import dlt
from dlt.extract.incremental import incremental


@dlt.source
def incremental_test_source():
    return [some_data()]


@dlt.resource(incremental=incremental('my_column', initial_value=25))
def some_data(incremental: incremental[int, str]):
    for i in range(incremental.last_value, incremental.last_value + 100):
        yield {'my_column': i}


pipeline = dlt.pipeline(
    pipeline_name='incremental_test', destination='duckdb', dataset_name='my_data'
)
load_info = pipeline.run(incremental_test_source())
```

API wise I'm not that happy with having to write "incremental" so many times.

I think it would be nicer to just have:
```python
@dlt.resource
def some_data(incremental: Incremental)
```
And have it work like configspec with/without default value in code and allow different formats: `incremental='column'`, `incremental=Incremental('column', initial...)`, or dict, etc  
Have to experiment with that a bit more.


dlt/extract/source.py
```diff
@@ -135,24 +136,31 @@ class DltResource(Iterable[TDataItem], DltResourceSchema):
...
-    def __init__(self, pipe: Pipe, table_schema_template: TTableSchemaTemplate, selected: bool):
+    def __init__(self, pipe: Pipe, table_schema_template: TTableSchemaTemplate, selected: bool, incremental: Incremental[Any, Any] = None):
         # TODO: allow resource to take name independent from pipe name
         self.name = pipe.name
         self.selected = selected
         self._pipe = pipe
+        self.incremental = incremental
+        if self.incremental:
+            self.incremental.resource_name = self.name
+            # TODO: Incremental needs to go last, I think. What happens when other steps are added after this?
```
rudolfix, line 147: it probably should, but I'd ignore this problem for now. we could construct pipes dynamically when they are passed to `PipeIterator.from_pipes` or add some `factory` steps in the pipe itsef (like the `ForkPipe` I use to pump data across dependend resources)

steinitzu:
Yeah exactly that would work I think.

dlt/extract/source.py
```diff
@@ -341,11 +349,14 @@ def _wrap_gen_step(self, *args: Any, **kwargs: Any) -> Any:
         skip_items_arg = 1 if self.is_transformer else 0  # skip the data item argument for transformers
         sig = inspect.signature(head)
         no_item_sig = sig.replace(parameters=list(sig.parameters.values())[skip_items_arg:])
+        for key, value in no_item_sig.parameters.items():
+            # TODO: Use bind to support overriding with positional arg as well
```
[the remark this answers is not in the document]
steinitzu: Yeah would be fine I guess to require default (or `dlt.config.value`), that would be consistent with how config injection and such works already.  
This is pretty naive method I think how I did it , works for the basic case I tested, but I'm sure this needs a little more thought.

dlt/extract/incremental.py
```diff
@@ -0,0 +1,78 @@
...
+    def last_value(self) -> Optional[TCursorValue]:
+        return self.state['last_value']
+
+    @property
+    def unique_keys(self) -> List[TUniqueValue]:
+        return self.state['unique_keys']
+
+    def __call__(self, row: TDataItem) -> bool:
+        # TODO: A bit inefficient to run this for each row.
+        # Can we use a transform which accepts a list of rows when resource yields in batches?
```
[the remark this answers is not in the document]
steinitzu: >     * `_wrap_gen_step` may not be executed even once so we may need extract the procedure above to a separate function that will be always called

When is this the case? When resource func is not a generator?

steinitzu:
I'm playing around with different injection method in last commit, using config spec internally also so all of this can be done in config too.  

Idea is to be able to use the same `Incremental` class everywhere, for default values, args, annotations.

```python
@dlt.resource
def a_data(state = Incremental('my_cursor')):
    yield {'a': 1}


@dlt.resource
def b_data(state: Incremental = dlt.config.value):
    yield {'b': 1}
```

For me this kind of API makes sense.

Would also support only passing the cursor column as string, which could be parsed by the configspec.

(Not hooking up the transform/filter atm. Just figuring out injection for now).

dlt/extract/incremental.py
```diff
@@ -0,0 +1,78 @@
...
+        if row is None:
```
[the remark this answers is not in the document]
rudolfix: yes, this will get the data verbatim. my take is to fail (not skip) on the data that is not a dictionary... but maybe we should be more generous?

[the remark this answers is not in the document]
steinitzu: Yeah I think fine to stick with dicts to begin with at least.  
If it comes up we can add other types in a new ticket, could warrant some more general abstraction to pluck data out of different records/collections.

dlt/extract/incremental.py
```diff
@@ -0,0 +1,133 @@
...
+
+from dlt.common.json import json
+from dlt.common.typing import TDataItem, AnyFun, TFun
+from dlt.common.configuration import configspec, known_sections, with_config, resolve_configuration
+from dlt.common.configuration.specs import BaseConfiguration
+from dlt.common.configuration.specs.exceptions import NativeValueError
+from dlt.common.configuration.accessors import config
+
+
+TCursorValue = TypeVar("TCursorValue", bound=Any)
```
rudolfix, line 15: you probably want to bind as `str` here?

dlt/extract/incremental.py
```diff
@@ -0,0 +1,133 @@
...
+TUniqueValue = TypeVar("TUniqueValue", bound=Any)
+LastValueFunc = Callable[..., Any]
+
+
+class IncrementalColumnState(TypedDict):
+    last_value: Optional[Any]
+    unique_keys: List[Any]
+
+
+@configspec
```
rudolfix, line 25: why `configspec`? do you want to make this part configurable and specify it via `config.toml`? will it work separately for each resource?

dlt/extract/incremental.py
```diff
@@ -0,0 +1,133 @@
...
+            initial_value: Optional[TCursorValue]=None,
+            last_value_func: Optional[LastValueFunc]=None,
+            primary_key: Optional[str] = None
+    ) -> None:
+        self.resource_name = resource_name
+        self.cursor_column = cursor_column
+        self.initial_value = initial_value
+        self.last_value_func = last_value_func or max
+        self.primary_key = primary_key
+        from dlt.pipeline.current import state  # TODO: Resolve circular import
```
rudolfix, line 56: we'll need to split `state,py` and move several definitions out. doable. we can go back to it when we have a working version

dlt/extract/decorators.py
```diff
@@ -284,7 +285,11 @@ def decorator(f: Callable[TResourceFunParams, Any]) -> Callable[TResourceFunPara
             resource_sections = (known_sections.SOURCES, _get_source_section_name(func_module), resource_name)
             # standalone resource will prefer existing section context when resolving config values
             # this lets the source to override those values and provide common section for all config values for resources present in that source
-            conf_f = with_config(f, spec=spec, sections=resource_sections, sections_merge_style=ConfigSectionContext.resource_merge_style)
+            conf_f = with_config(
```
rudolfix, line 288: I like it

dlt/extract/incremental.py
```diff
@@ -0,0 +1,133 @@
...
+        new_kwargs = {}
+
+        for p in sig.parameters.values():
+            if p.name in kwargs:
+                value = kwargs[p.name]
+                value.resource_name = resource_name
+                new_kwargs[p.name] = value
+            elif issubclass(p.annotation, Incremental) or isinstance(p.default, Incremental):
+                cfg = resolve_configuration(IncrementalConfigSpec(), sections=resource_sections, explicit_value=p.default)
+                new_kwargs[p.name] = Incremental.from_config(cfg, resource_name)
```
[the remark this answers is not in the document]
steinitzu: Yeah just wanted to see if this would work first, and it does. Config resolution in dlt is powerful :D  
Next is to put all these ideas together.

Do you mean to put the `with_incremental` wrapping in `DltResource`?  
I was doing it in the decorator because the config sections were already defined there.  

I'm thinking `with_incremental` could resolve the `Incremental` instance outside the wrapper and return both the wrapped func and the instance:

```
def with_incremental(func: TFun, ...) -> Tuple[TFun, Incremental]:
    ...
```

And the @resource decorator simply then passes a completely configured `Incremental` object to `DltResource`:

```
conf_f, incremental = with_incremental(conf_f, ...)

...
# make_resource

DltResource.from_data(..., incremental=incremental)
```

`DltResource` then simply receives a fully configured incremental object and only has to add the pipe step.

dlt/extract/incremental.py
steinitzu: I would expect `int` also for now. Have https://github.com/dlt-hub/dlt/issues/180 in mind too.   
Technically anything that can serialize and supports "lessthan" comparison would work.

rudolfix: ahh this is `value` not `key` my bad

dlt/extract/incremental.py
```diff
@@ -0,0 +1,148 @@
...
+        from dlt.pipeline.current import state  # TODO: Resolve circular import
+        self._state = state
+
+    @classmethod
+    def from_config(cls, cfg: IncrementalConfigSpec, resource_name: str) -> "Incremental":
+        # TODO: last_value_func from name
+        return cls(cfg.cursor_column, initial_value=cfg.initial_value, resource_name=resource_name)
+
+    @property
+    def state(self) -> IncrementalColumnState:
```
rudolfix, line 67: I'm writing a ticket that will add resource scoped state. I'll link it to this PR. we should have unified lead for the state.

rudolfix:
this looks good to me

rudolfix:
state refactor and layout: #214 
we may do this ticket together here so you can use `state` from common modules + we have unified layout

steinitzu:
> state refactor and layout: #214 we may do this ticket together here so you can use `state` from common modules + we have unified layout

Yeah makes sense, I can do this here too. The "current pipe" global is a good idea.

dlt/extract/incremental.py
```diff
@@ -0,0 +1,174 @@
...
+    @property
+    def unique_keys(self) -> List[TUniqueValue]:
+        return self.state['unique_keys']
+
+    def __call__(self, row: TDataItem, primary_key: Optional[TTableHintTemplate[TColumnKey]]) -> bool:
+        if row is None:
+            return True
+        state = self.state
+        last_value = state['last_value']
+        row_value = json.loads(json.dumps(self.cursor_column_p.parse(row)[0]))  # For now the value needs to match deserialized presentation from state
```
rudolfix, line 101: I get the problem. What about making our `last_value_func` always working on string representations?
- all types of interest (int, date, timestamp, decimal, float(?)) are also string comparable
- then you can just use `serialize_value` from `configuration/utils`
- you can always write a custom last value func and use `deserialize_value` to go back to the native representation

dlt/extract/pipe.py
```diff
@@ -588,6 +591,8 @@ def _get_source_item(self) -> ResolvablePipeItem:
         # TODO: count items coming of the generator and stop the generator if reached. that allows for sampling the beginning of a stream
         # _counts(id(gen)).setdefault(0) + 1
         try:
+            global _CURRENT_PIPE
```
rudolfix, line 594: this will unfortunately not work with async and thread pool :/ my take for now: require that `resource_state` gets explicit resource name. this we can have in incremental. and we do not expose the `resource_state` via dlt.current for now until this is solved

dlt/extract/incremental.py
```diff
@@ -0,0 +1,174 @@
...
+        if primary_key:
+            unique_value = resolve_column_value(primary_key, row)
```
rudolfix, line 103: please hash it. use `digest256`

rudolfix (changes requested):
thanks for the tests!

rudolfix (commented):
in tests following cases are missing
- last value function is not a default ie `min` or user supplied
- we test a few last value types like date, datetime, int
- compound primary keys

again thx for all the tests. it is really appreciated

dlt/extract/incremental.py
steinitzu: String compare like this? :grimacing:  `max("11", "100")` 

last_value_func is not always called i.e. `if last_value == row_value:`  and on the first row with clean state. But now thinking it should always be used anyway. You may have some special parsing in a custom function.

steinitzu: I think best would be to use the typed json dumper for state. Least surprising when you get the same type back as you put in.

dlt/extract/pipe.py
steinitzu: Ah ok. Is it more complicated than just multiple of these generators running concurrently? A contextvar instead of global would not be enough?

dlt/extract/incremental.py
```diff
@@ -0,0 +1,174 @@
...
+        return self.__class__(self.cursor_column, initial_value=self.initial_value, last_value_func=self.last_value_func)
+
+    @classmethod
+    def from_config(cls, cfg: IncrementalConfigSpec) -> "Incremental[Any, Any]":
+        # TODO: last_value_func from name
+        return cls(cfg.cursor_column, initial_value=cfg.initial_value)
+
+    @property
+    def state(self) -> IncrementalColumnState:
+        return self._state().setdefault('incremental', {}).setdefault(self.cursor_column, {'last_value': json.loads(json.dumps(self.initial_value)), 'unique_keys': []}) # type: ignore
```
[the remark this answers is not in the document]
steinitzu: For sure. This state only or the whole pipeline/source state in general?  
Wanted to leave it like this for now to not make too many assumptions about when/how to invalidate.

dlt/extract/incremental.py
```diff
@@ -0,0 +1,166 @@
...
+    @property
+    def state(self) -> IncrementalColumnState:
+        return self._state().setdefault('incremental', {}).setdefault(self.cursor_column, {'last_value': self.initial_value, 'unique_keys': []}) # type: ignore
+
+    @property
+    def last_value(self) -> Optional[TCursorValue]:
+        return self.state['last_value']  # type: ignore
+
+    @property
+    def unique_keys(self) -> List[TUniqueValue]:
```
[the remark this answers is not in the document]
steinitzu: Removed this propery. Shouldn't be any need to access it externally.

dlt/extract/incremental.py
rudolfix: my idea with comparing integers as string was amazing 😁

rudolfix (approved):
let's merge this. next steps
1. #180 to have native types in state
2. to be released with merge write disposition as alpha
3. a few additional end to end tests will be included in followup ticket

thanks!

PR #206, opened 2023-03-28
add faqs

adrianbr:


TyDunn:
@adrianbr why did you close? is this content going somewhere?

PR #217, opened 2023-03-30
added airflow functionality and tests

mebusila:
Airflow functionality and tests.
One more thing left that we need to discus is how can i test how many records had been saved


rudolfix (changes requested):
@mebusila could you add the docket compose to run airflow into tests/helpers/airflow_tests/airflow?

mebusila:
> @mebusila could you add the docket compose to run airflow into tests/helpers/airflow_tests/airflow?

we will not need an airflow instance for running airflow, just airflow installed >= 2

rudolfix:
followed by #357

PR #231, opened 2023-04-05
improving seo of docs

TyDunn:
See https://docusaurus.io/docs/next/seo for SEO suggestions


rudolfix (approved):
thanks for doing this!

PR #234, opened 2023-04-06
Docs: update logo link to homepage and add redirect to intro page

burnash:


TyDunn:
@burnash This is almost good to go, except `/docs` errors when it should probably redirect to `docs/intro` for now. We could add netlify redirect for this, but I think there should be a way to do it with docusarus (so running it locally provides the same experience). Maybe @rudolfix has an idea?

burnash:
@TyDunn Thanks for bringing this up. I'm getting the same error too. 
I was thinking that this PR would be a temporary fix. If that's the case, adding a redirect as you suggested would be the easiest solution. We could also explore a client-side redirect if needed.

TyDunn:
@burnash Good point. The netlify redirect is good enough for now. Thanks for pulling this PR together!

TyDunn (approved):

PR #243, opened 2023-04-10
adds schema to docs

rudolfix:
please also review the incremental load update (merge, last value)
https://github.com/dlt-hub/dlt/blob/rfix/adds-schema-docs/docs/website/docs/customization/incremental-loading.md (@adrianbr )


TyDunn (approved):
Nice walkthrough! We should probably merge #245 first and then adjust this PR for the new structure :)

PR #244, opened 2023-04-11
Handle optional incremental arg with explicit spec

steinitzu:
@rudolfix this workaround due to (I think) https://github.com/dlt-hub/dlt/blob/ecd773057cd4c3de5d93b486839f2a66d8557a84/dlt/common/configuration/resolve.py#L238-L239  

SQL pipeline is working with it in this branch: https://github.com/dlt-hub/pipelines/compare/master...example-incremental#diff-52f3a49a9a78cdceaf9545d295067445ceb1b467436135d35227e84054b73841R14-R16  
Because of the embedded spec you get an empty `Incremental()` instance instead of `None` when there's nothing in config  

This fix is not that great cause you can explictly pass empty `incremental=Incremental()` and get no error, just silently disables incremental load.   
Do you know any better solution? If there is some support for this case already or maybe we could add an attribute to configspec to prevent partial for individual specs?

### DIFF
--- a/dlt/extract/incremental.py
+++ b/dlt/extract/incremental.py
@@ -180,13 +180,9 @@ class IncrementalResourceWrapper(FilterItem):
             elif isinstance(p.default, Incremental):
                 new_incremental = p.default.copy()
 
-            if new_incremental and not new_incremental.cursor_path:
-                # Workaround for optional arg in spec
-                new_incremental = None
-                bound_args.arguments[p.name] = None
-
-            if not new_incremental:
+            if not new_incremental or new_incremental.is_partial():
                 if is_optional_type(p.annotation):
+                    bound_args.arguments[p.name] = None  # Remove partial spec
                     return func(*bound_args.args, **bound_args.kwargs)
                 raise ValueError(f"{p.name} Incremental has no default")
             new_incremental.resource_name = self.resource_name
--- a/dlt/extract/incremental.py
+++ b/dlt/extract/incremental.py
@@ -180,9 +180,14 @@ class IncrementalResourceWrapper(FilterItem):
             elif isinstance(p.default, Incremental):
                 new_incremental = p.default.copy()
 
+            if new_incremental and not new_incremental.cursor_path:
+                # Workaround for optional arg in spec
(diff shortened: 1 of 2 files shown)


dlt/extract/incremental.py
```diff
@@ -180,9 +180,14 @@ def _wrap(*args: Any, **kwargs: Any) -> Any:
             elif isinstance(p.default, Incremental):
                 new_incremental = p.default.copy()
 
+            if new_incremental and not new_incremental.cursor_path:
+                # Workaround for optional arg in spec
+                new_incremental = None
+                bound_args.arguments[p.name] = None
+
             if not new_incremental:
```
rudolfix, line 188: `Incremental` derives from `BaseConfiguration` which has this method
```
def is_partial(self) -> bool:
        """Returns True when any required resolvable field has its value missing."""
        if self.__is_resolved__:
            return False
        # check if all resolvable fields have value
        return any(
            field for field, hint in self.get_resolvable_fields().items() if getattr(self, field) is None and not is_optional_type(hint)
        )
        ```

I think you can just add
```
if not new_incremental or new_incremental.is_partial():
```

no need to set it to none above

rudolfix:
If embed is optional it can be simply not present or be partially present. I'm not sure if I ever use the second option so maybe I'll remove that. See the review for the PR. there's `is_partial()` method that you can call on the embed to see if it was correctly evaluated

dlt/extract/incremental.py
steinitzu: Very good!  changed to  https://github.com/dlt-hub/dlt/pull/244/files#diff-759977be85aa9d332107726916584d2574a442a8119117d9010bd5fa26be8310R183-R186

steinitzu:
> If embed is optional it can be simply not present or be partially present. I'm not sure if I ever use the second option so maybe I'll remove that.   

Yeah maybe worth looking into if anything important depends on current behaviour. I found this quite confusing, at least for embedded spec like this I would expect a complete object or `None`.

rudolfix (approved):

rudolfix:
ok I changed that

PR #248, opened 2023-04-11
adds Agolia DocSearch to the dlt docs 🚀

TyDunn:
https://docsearch.algolia.com/


rudolfix (approved):
I see only magic number there :) approved

PR #253, opened 2023-04-12
Exclude file patterns from changes check in CI

burnash:


.github/workflows/get_docs_changes.yml
```diff
@@ -7,6 +7,10 @@ on:
         description: "Changes outside docs"
         value: ${{ jobs.get_docs_changes.outputs.changes_outside_docs }}
 
+env:
+  EXCLUDED_FILE_PATTERNS: '^docs/|^README.md|^LICENSE\.txt|\.editorconfig|\.gitignore'
```
burnash, line 11: @rudolfix what else is safe to exclude?

PR #255, opened 2023-04-12
moving `create a pipeline` to use weatherapi and duckdb

TyDunn:


TyDunn:
@rahuljo Please review! Ignore that there is no metadata at the top of the file. I added that in a different PR :)

rahuljo (approved):

PR #264, opened 2023-04-13
PyPI package name: migrate to `dlt`

burnash:
This PR updates PyPI package name from 'python-dlt' to 'dlt' for a more concise naming.


rudolfix (approved):
this looks good but from now on we are not able to release python-dlt. is there any way to do that automatically? for example having two `pyproject.toml` files or swapping the name during the build. we can easily probe for the right package name if `version.py` so that should not be a problem. the tests would OFC work on the `dlt` package

rudolfix (changes requested):
I've also found this: https://github.com/simonw/pypi-rename
another cool trick: https://stackoverflow.com/questions/61363534/whats-the-recommended-way-of-renaming-a-project-in-pypi

burnash:
@rudolfix Change version to stable, please reveiw

rudolfix (approved):

PR #272, opened 2023-04-14
Fix a typo in docstrings and an exception msg

burnash:


rudolfix (approved):
thanks for fixing this!

PR #276, opened 2023-04-17
clones and updates pipeline schemas after extract

rudolfix:
this avoids modification of pipeline stored schemas before extract step completes successfully


steinitzu (changes requested):
Could you add a test case that reproduces the problem you're fixing. Seems like an issue that's easy to re-introduce accidentally.  

Otherwise it looks like a good change. Cloning and ordering of extract -> save schema, makes sense from looking at it.

PR #284, opened 2023-04-18
adds anonymous id to telemetry

### LINKED ISSUE #216: Anonymous telemetry improvements
We should adjust anonymous telemetry to include
1. a randomly generated identifier that remains consistent across events from the same `dlt` install
2. tracking of source name for loads if it is using an existing pipeline or `custom` if it is not

rudolfix:
closes #216 without `tracking of source name for loads if it is using an existing pipeline or custom if it is not` I could not find an easy way to do it. I want to push anonymous id quickly


TyDunn (approved):
Looks good to me. Let's get this merged!

Issue #209, opened 2023-03-29
add `dlt pipeline drop` command to fully or partially "reset" the pipeline


rudolfix:
**Background**
Our users report a need to fully or partially reset the pipeline state. This is important when the pipeline has many resources and there's a need to reset (re-acquire data) for `append` resource or incremental resource without affecting the others. Such operation is possible right now but quite complicated so we want to add a pipeline helper and expose it via `dlt` cli

**How it should work**
1. The user passes a resource name, or a list of resources or a wildcard (see our `SimpleRegex`). The user also optionally passes additional keys in the `sources` pipeline state that should be removed. The user optionally passes schema name (for pipelines with many schemas)
2. `dlt` removes affected tables from the default schema (or from named schema). affected tables include: (1) tables matched by user input (2) all child tables of the matched tables
3. `dlt` removes all state keys belonging to given resource + all explicitly indicated keys. note that schema name is the same as source name in the state
4. `dlt` drops all the affected tables at the destination. if destination support ddl transactions, it is used. non existing tables are silently ignored (see destination capabilities)
5. only then `dlt` saves the schemas and state locally.
6. `dlt` does the schema sync
7. `dlt` loads the state to the destination

**Tasks**
1. * [ ] implement the process above in the `dlt.pipeline.helpers`. make sure the process is idempotent
2. * [ ] add `dlt pipeline drop` command which gets the user input + a few flags (`--drop-all` should just drop the dataset and sync for a full reset `--skip-state-wipe` will not reset default state keys)
3. * [ ] make sure state layout we wipe is compatible with the state layout of #194 
4. * [ ] ask user before any modification: display a list of dropped tables and a list of deleted state keys and ask for confirmation
5. * [ ] write test of the helper in `tests/load/pipelines` to test against all possible destinations
6. * [ ] write tests for a few cases of `dlt pipeline drop` usage
7. * [ ] document in docstrings and in docs

PR #285, opened 2023-04-18
Feat/pipeline drop command

steinitzu:
https://github.com/dlt-hub/dlt/issues/209   
TODO:
- [x] CLI command
- [x] Dry run implementation (`DropCommand.info`)
- [x] idempotency tests
- [x] Wildcard resource names (regex)
- [x] Test state-only
- [x] Test CLI command

### DIFF
--- a/dlt/cli/pipeline_command.py
+++ b/dlt/cli/pipeline_command.py
@@ -1,4 +1,5 @@
 import yaml
+from typing import Any
 import dlt
 from dlt.cli.exceptions import CliCommandException
 
@@ -14,7 +15,7 @@ from dlt.pipeline.helpers import DropCommand
 from dlt.cli import echo as fmt
 
 
-def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, verbosity: int, load_id: str = None, drop_args: DictStrAny = None) -> None:
+def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, verbosity: int, **command_kwargs: Any) -> None:
     if operation == "list":
         pipelines_dir = pipelines_dir or get_dlt_pipelines_dir()
         storage = FileStorage(pipelines_dir)
@@ -110,6 +111,7 @@ def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, ver
             p.sync_destination()
 
     if operation == "load-package":
+        load_id = command_kwargs.get('load_id')
         if not load_id:
             packages = sorted(p.list_normalized_load_packages())
             if not packages:
@@ -138,7 +140,7 @@ def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, ver
         fmt.echo(schema_str)
 
     if operation == "drop":
-        drop = DropCommand(p, **drop_args)
+        drop = DropCommand(p, **command_kwargs)
         fmt.echo("About to drop the following data for the pipeline:")
         for k, v in drop.info.items():
             fmt.echo("%s: %s" % (fmt.style(k, fg="green"), v))
--- a/dlt/cli/pipeline_command.py
+++ b/dlt/cli/pipeline_command.py
@@ -5,17 +5,19 @@ from dlt.cli.exceptions import CliCommandException
 
 from dlt.common import json
 from dlt.common.pipeline import get_dlt_pipelines_dir, TSourceState
+from dlt.common.destination.reference import TDestinationReferenceArg
 from dlt.common.runners import Venv
 from dlt.common.runners.stdout import iter_stdout
 from dlt.common.schema.utils import remove_defaults
 from dlt.common.storages.file_storage import FileStorage
 from dlt.common.typing import DictStrAny
 from dlt.pipeline.helpers import DropCommand
+from dlt.pipeline.exceptions import CannotRestorePipelineException
 
 from dlt.cli import echo as fmt
 
 
-def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, verbosity: int, **command_kwargs: Any) -> None:
+def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, verbosity: int, dataset_name: str = None, destination: TDestinationReferenceArg = None, **command_kwargs: Any) -> None:
     if operation == "list":
         pipelines_dir = pipelines_dir or get_dlt_pipelines_dir()
         storage = FileStorage(pipelines_dir)
@@ -28,7 +30,16 @@ def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, ver
             fmt.secho(_dir, fg="green")
         return
 
-    p = dlt.attach(pipeline_name=pipeline_name, pipelines_dir=pipelines_dir)
+    try:
+        p = dlt.attach(pipeline_name=pipeline_name, pipelines_dir=pipelines_dir)
+    except CannotRestorePipelineException as e:
+        fmt.echo(e)
+        if not fmt.confirm("Do you want to attempt to restore from destination?"):
+            return
+        destination = destination or fmt.text_input(f"Enter destination name for pipeline {pipeline_name}")
+        dataset_name = dataset_name or fmt.text_input(f"Enter dataset name for pipeline {pipeline_name}")
+        p = dlt.attach(pipeline_name=pipeline_name, pipelines_dir=pipelines_dir, from_destination=True, dataset_name=dataset_name, destination=destination)
+
     fmt.echo("Found pipeline %s in %s" % (fmt.bold(p.pipeline_name), fmt.bold(p.pipelines_dir)))
 
     if operation == "show":
--- a/dlt/cli/pipeline_command.py
+++ b/dlt/cli/pipeline_command.py
@@ -8,11 +8,13 @@ from dlt.common.runners import Venv
 from dlt.common.runners.stdout import iter_stdout
 from dlt.common.schema.utils import remove_defaults
 from dlt.common.storages.file_storage import FileStorage
+from dlt.common.typing import DictStrAny
+from dlt.pipeline.helpers import DropCommand
 
 from dlt.cli import echo as fmt
 
 
-def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, verbosity: int, load_id: str = None) -> None:
+def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, verbosity: int, load_id: str = None, drop_args: DictStrAny = None) -> None:
     if operation == "list":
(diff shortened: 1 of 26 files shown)


dlt/pipeline/helpers.py
```diff
@@ -32,4 +36,100 @@ def _retry_load(ex: Exception) -> bool:
...
+    def __call__(self) -> None:
+        if self.pipeline.has_pending_data:  # Raise when there are pending extracted/load files to prevent conflicts
+            raise PipelineHasPendingDataException(self.pipeline.pipeline_name, self.pipeline.pipelines_dir)
+        self.delete_pipeline_tables()
+        self.drop_destination_tables()
+        self.drop_state_keys()
+        self.pipeline.schemas.save_schema(self.schema)
+        # Send updated state to destination
+        self.pipeline.normalize()
+        self.pipeline._get_load_storage().delete_completed_package
```
rudolfix, line 120: that does nothing?

dlt/pipeline/helpers.py
```diff
@@ -32,4 +36,100 @@ def _retry_load(ex: Exception) -> bool:
...
+        try:
+            self.pipeline.load(raise_on_failed_jobs=True)
+        except Exception:
+            # Clear extracted state on failure so command can run again
+            self.pipeline._get_load_storage().wipe_normalized_packages()
```
rudolfix, line 125: cool!

dlt/pipeline/pipeline.py
```diff
@@ -1103,6 +1108,10 @@ def _managed_state(self, *, extract_state: bool = False) -> Iterator[TPipelineSt
             # check if any state element was changed
             merged_state = merge_state_if_changed(backup_state, state)
 
+            # Force exctract when state hasn't changed
+            if extract_unchanged:
```
rudolfix, line 1112: why we need that? if there is nothing to sync then we can skip that part. do you see any problem with that?

dlt/pipeline/pipeline.py
```diff
@@ -543,7 +543,12 @@ def has_data(self) -> bool:
         return not self.first_run or bool(self.schema_names) or len(self.list_extracted_resources()) > 0 or len(self.list_normalized_load_packages()) > 0
 
     @property
-    def schemas(self) -> Mapping[str, Schema]:
+    def has_pending_data(self) -> bool:
```
rudolfix, line 546: maybe you could use it in the prop above
```
@property
    def has_data(self) -> bool:
        """Tells if the pipeline contains any data: schemas, extracted files, load packages or loaded packages in the destination"""
        return not self.first_run or bool(self.schema_names) or len(self.list_extracted_resources()) > 0 or len(self.list_normalized_load_packages()) > 0
        ```

Comment (rudolfix) [reply dlt/pipeline/pipeline.py:546]: you should test it in the same place `has_data` is tested

File: tests/load/pipeline/test_drop.py [hunk 5]
```diff
@@ -0,0 +1,186 @@
...
+@pytest.mark.parametrize('destination_name', ALL_DESTINATIONS)
+def test_drop_destination_tables_fails(destination_name: str) -> None:
+    """Fail on drop tables. Command runs again."""
+    source = droppable_source()
+    pipeline = dlt.pipeline(pipeline_name='drop_test_' + uniq_id(), destination=destination_name, dataset_name='drop_data_'+uniq_id())
+    pipeline.run(source)
+
+    attached = _attach(pipeline)
+
+    with mock.patch.object(helpers._DropCommand, 'drop_destination_tables', side_effect=RuntimeError("Something went wrong")):
```
rudolfix, line 130: cool!

dlt/pipeline/helpers.py
```diff
@@ -32,4 +36,100 @@ def _retry_load(ex: Exception) -> bool:
...
+        state: TSourceState
+        with self.pipeline.managed_state(extract_state=True, extract_unchanged=True) as state:  # type: ignore[assignment]
+            source_states = state.get("sources")
+            if not source_states:
+                return
+            # TODO: Should only drop from source state matching the schema. Not possible atm
+            for source_state in source_states.values():
+                for p in self.drop_state_paths:  # Remove keys by JSON path
+                    p.filter(lambda _: True, source_state)
+                resources = source_state.get('resources')
```
rudolfix, line 100: this should be a part of `common/pipeline.py` there's a function that resets the state there. maybe you can use it. if not please keep all operations that rely on our state layout in there. I refer to you getting resources here and assuming it has a list of sources

rudolfix (commented):
looks good! I had just some minor comments. let's move on with the CLI

rudolfix:
@steinitzu I've got some feedback:
1. in case there's no local state of the pipeline we should try create a new one and sync it with destination. for that you'll need a dataset name, you request it from the user.
2. the resource name (s) to drop may be regexes (or file name patterns `fnmatch`) nothing super fancy required
3. we should have option `--state-only` that would just drop the state and keep the tables

dlt/pipeline/pipeline.py
steinitzu: It's a bit of an wonky state to get the pipeline in though.  
Might be ok to always run sync destination at the beginning instead (after checking for pending data)

dlt/pipeline/helpers.py
steinitzu: Yeah just clutter leftover. Was trying something.

dlt/common/destination/reference.py
```diff
@@ -139,6 +139,11 @@ def update_storage_schema(self, staging: bool = False, only_tables: Iterable[str
         self._verify_schema()
         return expected_update
 
+    @abstractmethod
+    def drop_tables(self, *tables: str, staging: bool = False) -> None:
```
rudolfix, line 143: this should go to `SqlJobClientBase`. `JobClientBase` may not support any "tables" (ie. be an event sink or a parquet file). you need to test it as wel

dlt/destinations/dummy/dummy.py
```diff
@@ -89,6 +89,9 @@ def update_storage_schema(self, staging: bool = False, only_tables: Iterable[str
             raise DestinationTransientException("Raise on schema update due to fail_schema_update config flag")
         return applied_update
 
+    def drop_tables(self, *tables: str, staging: bool = False) -> None:
```
rudolfix, line 92: see above

rudolfix (commented):
Thanks for the update! Besides the comments we need some tests:

- [ ] 1. test the `pipeline_command` like the other pipeline commands are tested. not all the cases but two are important: happy path when the local state exist and happy path when we get the state from the destination
- [ ] 2. tests for drop tables
- [ ] 3. some simple unit tests for deleting state keys
- [ ] 4. test the wipe of the normalized folder

dlt/pipeline/helpers.py
```diff
@@ -32,4 +40,140 @@ def _retry_load(ex: Exception) -> bool:
...
+        return state  # type: ignore[return-value]
+
+    def _drop_state_keys(self) -> None:
+        state: Dict[str, Any]
+        with self.pipeline.managed_state(extract_state=True, extract_unchanged=True) as state:  # type: ignore[assignment]
+            state.clear()
+            state.update(self._new_state)
+
+    def _drop_all(self) -> None:
+        with self.pipeline.sql_client(self.schema.name) as client:
```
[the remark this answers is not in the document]
steinitzu: Yes, doing this now. drop-all just puts the regex to `.*` so it matches all resources

dlt/common/destination/reference.py
steinitzu: Make sense. Done. Test added in https://github.com/dlt-hub/dlt/pull/285/commits/87be84bf8758f9aa75b9b44890795372cb38d5c5

dlt/cli/pipeline_command.py
```diff
@@ -134,3 +136,14 @@ def pipeline_command(operation: str, pipeline_name: str, pipelines_dir: str, ver
             fmt.echo("Found schema with name %s" % fmt.bold(p.default_schema_name))
         schema_str = p.default_schema.to_pretty_yaml(remove_defaults=True)
         fmt.echo(schema_str)
+
+    if operation == "drop":
+        drop = DropCommand(p, **drop_args)
+        if drop.drop_all:
+            fmt.echo("About to drop all destination and local data for the pipeline:")
```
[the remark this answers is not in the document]
steinitzu: Would need destination name too, right? Shouldn't we add this to the sync command also and have it as a general util? Maybe an option in `dlt.attach` or another `dlt.attach_from_destination`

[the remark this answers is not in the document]
steinitzu: @rudolfix I added this here for all the pipeline commands https://github.com/dlt-hub/dlt/pull/285/commits/942a52a954468e790e6886b6edfbdf6174375962  
When regular `attach` fails it will prompt and try to attach new pipeline and sync   

Let me know if you think it makes sense like this and I'll add tests, etc.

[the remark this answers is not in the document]
rudolfix: ohhh I missed this comment. but my last PR answers it

dlt/pipeline/helpers.py
```diff
@@ -1,11 +1,19 @@
+from collections import defaultdict
+from typing import Callable, Tuple, Iterable, Optional, Any, cast, List, Iterator, Dict, Union, TypedDict
+from itertools import chain
 
-
-from typing import Callable, Tuple
+# from jsonpath_ng import parse as jsonpath_parse, JSONPath
+from dlt.common.jsonpath import resolve_paths, TAnyJsonPath, compile_paths
```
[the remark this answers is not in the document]
steinitzu: Yes. Doing this now.  
May add tests for the helpers in another pr. For now they're indirectly tested.

dlt/pipeline/helpers.py
```diff
@@ -32,4 +40,140 @@ def _retry_load(ex: Exception) -> bool:
...
+            self.pipeline.schemas.save_schema(self.schema)
+        # Send updated state to destination
+        self.pipeline.normalize()
+        try:
+            self.pipeline.load(raise_on_failed_jobs=True)
+        except Exception:
+            # Clear extracted state on failure so command can run again
+            self.pipeline._get_load_storage().wipe_normalized_packages()
+            with self.pipeline.managed_state() as state:
+                state["_local"].pop("_last_extracted_at", None)
```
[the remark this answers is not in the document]
steinitzu: I guess on Pipeline where other local state is managed. Kinda blanking on what to name it :)  
Think I'd rather just leave it as is until there's a second use case for it or if/until this state layout changes.

rudolfix (approved):
thx for the tests :)

PR #287, opened 2023-04-20
Add Airflow poetry group

burnash:
This PR adds a poetry dependency group for Airflow required by #273 
Updated poetry.lock leads to mypy failing with:
```
dlt/common/storages/versioned_storage.py:51: error: Returning Any from function declared to return "Version"  [no-any-return]
```
This PR fixes that as well.


dlt/common/storages/versioned_storage.py
```diff
@@ -48,7 +48,8 @@ def version(self) -> semver.VersionInfo:
         return self._load_version()
 
     def _load_version(self) -> semver.VersionInfo:
-        return self.storage.load(VersionedStorage.VERSION_FILE)
+        version_str = self.storage.load(VersionedStorage.VERSION_FILE)
+        return semver.VersionInfo.parse(version_str)
```
burnash, line 52: @rudolfix I had to change this to it passes mypy check after I updated poetry lock. Let me know if this is the proper way to do this.

rudolfix: maybe semver got update and they have typings now. perfect!

rudolfix (approved):

PR #289, opened 2023-04-21
updates explore/transform the data in Python

rudolfix:


TyDunn (approved):
LGTM! You removed the transforming with Pandas section completely, so I brought some of the content from the other page about data frames over

rudolfix:
@TyDunn Yeah, because there were instructions to query the data, nothing about writing back panda frame back into destination! which would be a nice tutorial. WDT?

TyDunn:
@rudolfix This would be really cool!

PR #295, opened 2023-04-25
adding another google tag id

TyDunn:
see here for context: https://dlthubinc.slack.com/archives/C05409H5X0W/p1682429538342249


TyDunn:
@rudolfix can you review, so we can unblock Anna?

rudolfix (approved):

PR #298, opened 2023-04-27
Feat/new add_limit method for resources

z3z1ma:
This method is incredibly useful for end to end integration testing where a user requires a limited subset of the data. `add_filter` is not sufficient as it will still exhaust the generator in my experience and follows a very different code path than offering a legitimate cap to the generator before `return`ing. This is not to say you won't find a better way in the future @rudolfix but this solves the near term well enough.

### DIFF
--- a/dlt/extract/source.py
+++ b/dlt/extract/source.py
@@ -229,11 +229,15 @@ class DltResource(Iterable[TDataItem], DltResourceSchema):
             count = 0
             if inspect.isfunction(gen):
                 gen = gen()
-            for i in gen:  # type: ignore # TODO: help me fix this later
-                yield i
-                count += 1
-                if count > max_items:
-                    return
+            try:
+                for i in gen:  # type: ignore # TODO: help me fix this later
+                    yield i
+                    count += 1
+                    if count > max_items:
+                        return
+            finally:
+                if inspect.isgenerator(gen):
+                    gen.close()
             return
         # transformers should be limited by their input, so we only limit non-transformers
         if not self.is_transformer:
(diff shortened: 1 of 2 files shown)


z3z1ma:
This doesn't like transformers, currently. Any ideas?

dlt/extract/source.py
```diff
@@ -212,6 +212,34 @@ def add_filter(self, item_filter: ItemTransformFunc[bool], insert_at: int = None
...
+            """Wrap a generator to take the first 50 records"""
+            nonlocal max_items
+            count = 0
+            if inspect.isfunction(gen):
+                gen = gen()
+            for i in gen:  # type: ignore # TODO: help me fix this later
+                yield i
+                count += 1
+                if count > max_items:
+                    return
```
rudolfix, line 236: please close underlying gen before returning, otherwise the code will hang on last yield. we had problems with sources reading from databases that kept open connections until process terminated

dlt/extract/source.py
```diff
@@ -212,6 +212,34 @@ def add_filter(self, item_filter: ItemTransformFunc[bool], insert_at: int = None
...
+            "DltResource": returns self
+        """
+        def _gen_wrap(gen: TPipeStep) -> TPipeStep:
+            """Wrap a generator to take the first 50 records"""
+            nonlocal max_items
+            count = 0
+            if inspect.isfunction(gen):
+                gen = gen()
+            for i in gen:  # type: ignore # TODO: help me fix this later
+                yield i
```
rudolfix, line 233: dlt may close a pipe when another pipe fails. in that case `yield i` raises `GeneratorExit`. in that case you should also close the underlying `gen`

maybe a finally clause would do. also `GeneratorExit` does not derive from Exception but from its base

rudolfix (changes requested):
@z3z1ma amazing! I had the same idea when looking how `take_first` is used. I intended to raise an exception that would close the `gen` generator via StopIteration in the `Pipe` class. what you do is seems to be cleaner.

could you add a test in `tests/extract/test_sources`? we test such things ie. in `test_add_transform_steps`. extract tests run without setting up any credentials etc. so you should be able to run it yourself quickly.

z3z1ma:
I added the test and wrapped the function body consuming the generator in a try/finally which will guarantee the `close` fn is called @rudolfix 
Thanks!

PR #342, opened 2023-05-19
Feat/fsspec destination

steinitzu:
WIP


dlt/common/configuration/specs/aws_credentials.py
```diff
@@ -0,0 +1,35 @@
...
+from dlt.common.typing import TSecretStrValue
+from dlt.common.configuration.specs import CredentialsConfiguration, CredentialsWithDefault, configspec
+
+if TYPE_CHECKING:
+    from botocore.credentials import Credentials
+
+
+@configspec
+class AwsCredentials(CredentialsConfiguration, CredentialsWithDefault):
+    access_key_id: str = None
```
rudolfix, line 12: let's use the same names used in credentials file and env variables. this will also hopefully make gpt-4 hallucinate correct format when asked for aws credentials in dlt (works for connection strings and google credentials)
```
[default]
aws_access_key_id = YOUR_ACCESS_KEY
aws_secret_access_key = YOUR_SECRET_KEY
aws_session_token = YOUR_SESSION_TOKEN
```

rudolfix (commented):
this looks good! decorator to create dynamic type hints seems to work well.

we'll need a few tests for it for a few mock credentials cases, somewhere in tests/common/configuration

tests/load/filesystem/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+# env variables for URLs for all test buckets, e.g. "gcs://bucket_name", "s3://bucket_name", "file://bucket_name"
+bucket_env_vars = [
+    "tests.bucket_url_gcs", "tests.bucket_url_aws", "tests.bucket_url_file"
+]
+
+ALL_BUCKETS = [b for b in (dlt.config.get(var, str) for var in bucket_env_vars) if b]
+
+
+@pytest.fixture(scope='function', params=ALL_BUCKETS)
+def all_buckets_env(request) -> Iterator[str]:  # type: ignore[no-untyped-def]
```
rudolfix, line 19: please look how other client tests clean after themselves by deleting dataset. you should do the same. after each test the client should delete dataset (folder in bucket)

I added dataset delete to sql_client to handle that. you may do something similar

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -0,0 +1,106 @@
...
+
+    for job in jobs:
+        assert job.state() == 'completed'
+        dest_fn = LoadFilesystemJob.make_destination_filename(job.file_name(), client.schema.name, load_id)
+        assert client.fs_client.exists(root_path.joinpath(dest_fn))
+
+
+def test_replace_write_disposition(all_buckets_env: str) -> None:
+    dataset_name = 'test_' + uniq_id()
+    client, jobs1, root_path, load_id1 = perform_load(all_buckets_env, dataset_name, NORMALIZED_FILES, write_disposition='replace')
```
rudolfix, line 50: something wrong with "test_replace_write_disposition"
Always creates LoadEmptyJob. files. does not raise AccessDenied, is it working on a local file? `perform_load` is always passing

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -0,0 +1,106 @@
...
+    # assert that only these two files are in the destination folder
+    ls = set(client.fs_client.ls(root_path))
+    assert ls == {str(job_1_paths[0]), str(job_2_path)}
+
+
+
+def perform_load(
+    bucket_url: str, dataset_name: str, cases: Sequence[str], write_disposition: str='append'
+) -> Tuple[FilesystemClient, List[LoadJob], Path, str]:
+    load = setup_loader(dataset_name)
```
rudolfix, line 72: bucket url is not used.

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -0,0 +1,106 @@
...
+    files = load.load_storage.list_new_jobs(load_id)
+    jobs = []
+    for f in files:
+        job = Load.w_spool_job(load, f, load_id, schema)
+        jobs.append(job)
+
+    return client, jobs, root_path, load_id
+
+
+def prepare_load_package(load_storage: LoadStorage, cases: Sequence[str], write_disposition: str='append') -> Tuple[str, Schema]:
```
rudolfix, line 88: looks like it could go to common utils? dummy shares this

tests/common/configuration/test_credentials.py
```diff
@@ -286,3 +286,20 @@ def test_run_configuration_slack_credentials(environment: Any) -> None:
     environment["RUNTIME__SLACK_INCOMING_HOOK"] = "DBgAXQFPQVsAAEteXlFRWUoPG0BdHQ-EbAg=="
     c = resolve_configuration(RunConfiguration())
     assert c.slack_incoming_hook == "DBgAXQFPQVsAAEteXlFRWUoPG0BdHQ-EbAg=="
+
+
+def test_aws_credentials_resolved_from_default(environment: Dict[str, str]) -> None:
```
rudolfix, line 291: please test default credentials in tests/load - I do not want to install any extras for common tests

rudolfix (changes requested):
I did a few changes
- default buckets are defined in `tests/config.toml` ENV override will still work
- changes the scheme from gcs to gs
- some fixes when passing gcp credentials

my biggest issue is lack of test coverage, the client is tested in replace mode. you do not test if append is adding more files. not sure if you test if replace is testing if files are deleted.
you do
I do not see you testing if file layout is exactly as in ticket spec
I do not see tests for merge special case

Please add tests where filesystem is used as destination in pipelines. Maybe some of existing tests in load/pipelines could be converted (separating parts that require sql_client, you could look if files are present instead). it is some work but without that it is hard to say if this thing works

tests/load/filesystem/conftest.py
steinitzu: Doing this now in the filesystem_client fixture

dlt/destinations/filesystem/filesystem.py
```diff
@@ -0,0 +1,115 @@
+import random
+from pathlib import Path
+from copy import copy
+from types import TracebackType
+from typing import ClassVar, Dict, Optional, Sequence, Type, Iterable
+from fsspec.core import url_to_fs
```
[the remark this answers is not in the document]
steinitzu: Looked like this was handled in pipeline.py by https://github.com/dlt-hub/dlt/blob/feat%2Ffsspec-destination/dlt/pipeline/pipeline.py#L837-L849  
Bigquery and duckdb both import things without try/except. Am I missing something?

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -0,0 +1,106 @@
...
+NORMALIZED_FILES = [
+    "event_user.839c6e6b514e427687586ccc65bf133f.0.jsonl",
+    "event_loop_interrupted.839c6e6b514e427687586ccc65bf133f.0.jsonl"
+]
+
+
+def test_succesful_load(all_buckets_env: str) -> None:
+    dataset_name = 'test_' + uniq_id()
+    dataset_name = 'test_' + uniq_id()
+    client, jobs, root_path, load_id = perform_load(all_buckets_env, dataset_name, NORMALIZED_FILES, write_disposition='replace')
```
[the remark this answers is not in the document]
steinitzu: Added append test specifically.

tests/load/filesystem/test_filesystem_client.py
steinitzu: I'm not seeing this now. I'm getting fail with wrong aws and gc credentials. But I'll take a better look tomorrow if it's correct.

steinitzu:
I added the destination to some of the existing tests in `load/pipeline/test_pipelines.py`. Just the ones that check for table existance I do a filename check instead.   + few more fixes

I might add also a test to verify the file data tomorrow, could be good to have. Should just be whether the load package files and destination files are identical.  

And will add test for the merge fallback.

dlt/common/configuration/specs/base_configuration.py
```diff
@@ -169,7 +177,12 @@ def to_native_representation(self) -> Any:
     @classmethod
     def get_resolvable_fields(cls) -> Dict[str, type]:
         """Returns a mapping of fields to their type hints. Dunders should not be resolved and are not returned"""
-        return {f.name:f.type for f in cls.__dataclass_fields__.values() if not f.name.startswith("__")}
+        # Sort dynamic type hint fields last because they depend on other values
+        fields = sorted(
```
rudolfix, line 181: please make sure that original order of fields is kept for the fields that do not have dynamic type hint. does `sorted()` preserve the order? are you sorting on a boolean variable?

dlt/pipeline/pipeline.py
```diff
@@ -629,6 +629,16 @@ def sql_client(self, schema_name: str = None, credentials: Any = None) -> SqlCli
             schema = self.default_schema if self.default_schema_name else Schema(self.dataset_name)
         return self._sql_job_client(schema, credentials).sql_client
 
+    def destination_client(self, schema_name: str = None, credentials: Any = None) -> JobClientBase:
```
rudolfix, line 632: let's put `_` prefix - I'm not yet sure this should be public interface

rudolfix (changes requested):
looks good! thanks for the pipeline tests with all bucket types

dlt/common/configuration/specs/base_configuration.py
steinitzu: It is preserved. I'm sorting on bool  `key=lambda f: f.name in cls.__hint_resolvers__`  
So just the dynamic fields are resolved last. Otherwise original order is the same.   
It's mainly needed when dealing with a subclass like the fs destination config. Otherwise the hint for `credentials` gets resolved before `bucket_url` is known.

steinitzu: https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts  
> Sorts are guaranteed to be [stable](https://en.wikipedia.org/wiki/Sorting_algorithm#Stability). That means that when multiple records have the same key, their original order is preserved.

rudolfix: amazing 👍

dlt/destinations/filesystem/filesystem.py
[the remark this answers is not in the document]
rudolfix: it's part of main dependencies now. problem gone (if there was any :)

dlt/destinations/filesystem/filesystem.py
```diff
@@ -0,0 +1,119 @@
...
+            fs_path=self.fs_path
+        )
+
+    def restore_file_load(self, file_path: str) -> LoadJob:
+        return EmptyLoadJob.from_file_path(file_path, "completed")
+
+    def create_merge_job(self, table_chain: Sequence[TTableSchema]) -> NewLoadJob:
+        return None
+
+    def complete_load(self, load_id: str) -> None:
```
rudolfix, line 112: I was looking at our use cases and actually it would be useful to create a file here. the reason: when the file is present we now that load to the bucket was finalized and we can use the loaded data. You could probably execute LoadFileSystemJob. write_disposition: append, the file can be empty, and the table name is LOADS_TABLE_NAME (_dlt_loads)

rudolfix (changes requested):

dlt/common/configuration/resolve.py
```diff
@@ -117,6 +117,9 @@ def _resolve_config_fields(
     unresolved_fields: Dict[str, Sequence[LookupTrace]] = {}
 
     for key, hint in fields.items():
+        if key in config.__hint_resolvers__:
+            # Type hint for this field is created dynamically
+            hint = config.__hint_resolvers__[key](config)
```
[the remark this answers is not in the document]
steinitzu: I added an exception with message when the decorator is used without matching fields, e.g. the message:
```
The config spec ConfigWithInvalidDynamicType has dynamic type resolvers for fields: ['a', 'b', 'c'] but these fields are not defined in the spec.
When using @resolve_type() decorator, Add the fields with 'Any' or another common type hint, example:

>>> class ConfigWithInvalidDynamicType(BaseConfiguration)
>>>    a: Any
>>>    b: Any
>>>    c: Any
```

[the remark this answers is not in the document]
steinitzu: I'm not clear what you mean with the other cases. The standard config fields missing exception is raised as normal when e.g. aws credentials are missing.

dlt/destinations/filesystem/filesystem.py
steinitzu: https://github.com/dlt-hub/dlt/pull/342/commits/63a97ec6b0e19f25869f213fac6d9325a98e14d0#diff-524215919bce1cfd1cd9bae6d4ae82c9f753d4b3ffb1215a418dd06277113a5dR107-R110  
Will this do? Just creates empty file named `{schema_name}.{_dlt_loads}.{load_id}`  . Figure extension and file ID is not needed since it's just a marker per load ID?

dlt/common/configuration/resolve.py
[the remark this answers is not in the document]
rudolfix: it looks good now, thanks!

dlt/destinations/filesystem/filesystem.py
rudolfix: no, it is not needed. thx!

rudolfix (approved):

Issue #269, opened 2023-04-14
Update deprecated GitHub Actions 


burnash:
Some actions from dlt workflows are deprecated, and need to be updated to avoid any potential issues or disruptions to our workflows. Specifically in https://github.com/dlt-hub/dlt/blob/devel/.github/workflows/test_common.yml, following actions are deprecated:

- actions/setup-python@v1
- actions/cache@v2

Details: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

Screenshot:

<img width="871" alt="Screen Shot 2023-04-14 at 12 13 59" src="https://user-images.githubusercontent.com/264674/232017344-4894bf17-8aba-48df-9ba2-ef0e1543dc61.png">

The task is to check all the workflows within the project and update any actions that have been deprecated.

PR #345, opened 2023-05-20
Update deprecated GitHub action

tungbq:
Resolves: #269

Update workflow files to use:
- uses: actions/setup-python@v4
- uses: actions/cache@v3

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,7 +25,7 @@ jobs:
         uses: actions/checkout@master
 
       - name: Setup Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v4
         with:
           python-version: "3.10.x"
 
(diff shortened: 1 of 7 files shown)


tungbq:
There is no change in the test suites or the code, not sure why some CI checkers are failing with DB error:
```
Database Error
  connection to server at "dlttests.cwz0jfxu0m7m.eu-central-1.rds.amazonaws.com" (18.196.0.3), port 5432 failed: fe_sendauth: no password supplied
  
Error:   file_log:eventmgr.py:63 06:43:43.[120](https://github.com/dlt-hub/dlt/actions/runs/5030898708/jobs/9023625301?pr=345#step:7:121)150 [error] [MainThread]: Encountered an error:
```

rudolfix:
@tungbq thanks for fixing this. the reason is that PRs from fork do not have access to our repo secrets so destination tests will not run. I'll take a look and probably merge your work and test it again. we need to split the duckdb and postgres tests from other destinations so those can work from forks

PR #357, opened 2023-05-24
airflow dlt wrapper

### LINKED ISSUE #140: Airflow deployment
Create an airflow deployment similar to the current git actions deployment. The deployment should support

1. * [x] generation of a dag running the pipeline
2. * [x] ideally this dag should contain tasks representing resources, in the correct dependency. If not currently possible, start with just running the pipeline in a task before defining next steps.
3. * [ ] credentials should be passed into airflow or the appropriate user actions/guidance to take should be offered on CLI similar to the other deployment command.

### LINKED ISSUE #167: allow to log to `airflow.task` for Airflow deployments using KubernetesExecutor
To be able to check a task logs from the airflow UI in "realtime", one has to log to `airflow.task` (https://docs.astronomer.io/learn/logging?tab=traditional#add-custom-task-logs-from-a-dag). It would be good to see dlt logs there.

### LINKED ISSUE #190: airflow dlt helper (dlt operator)
**Background**
We want to make running `dlt` pipeliens with `airflow` as easy as possible in a manner similar to `dbt` helpers. Users should be able to add dlt pipelines to DAGs.
The idea is to create Airflow operator for dlt pipelines that will contain all required boilerplate.
The ideal operator would: (1) decompose the dlt pipeline into several DAG nodes (2) enable reading secrets from Airflow hooks (3) configure logging and notifications of dlt.

PoC of the operator is here: https://github.com/dlt-hub/airflow_operator/tree/draft

**Tasks**
1. * [x] implement a simple airflow operator that executes the whole pipeline as a single task
2. * [x] the helper function should accept pipeline instance, source instance, some additional settings (ie. if to redirect `dlt` logger into `airflow` logger #167 ) + additional parameters that airflow may need
3. * [x] it should return Airflow artifact (Operator/TaskGroup) that users can place in their DAGs
4. * [x] helper will become part of our Airflow deployment #140 
5. * [x] the airflow code should be placed in `dlt/helpers/airflow`
6. * [x] dlt will not install airflow by default so please use `MissingDependencyException` to inform user on missing dependencies
7. * [x] airflow does not ensure that you get a clean worker, the leftover from previous tasks may be present. pipelines must get a clean state. sure way to do that is to generate random `pipelines_dir` when creating pipeline or change the location in the operator
8. * [x] the progress log should be enabled and log period should be configured. we should log to the task logger by default
9. * [x] We should set the pipeline working dir to https://cloud.google.com/composer/docs/concepts/cloud-storage on Composer and it needs to have a random name (as specified before). The reason: there's plenty of space while on the worker itself there isn't
10. * [x] we should delete all the files by default https://dlthub.com/docs/running-in-production/running#data-left-behind
11. * [x] we should propose a good layout for imported dlt and pipeline modules. see below
12. * [x] we should add a tenacity retry with backoff option for extract and load stage with some sensible settings
13. * [x] we should add options to reduce the item buffer size 

**Python modules layout**
This is what we know about Airflow task runtime
- the CWD is always `AIRFLOW_HOME` which is a folder above the `dags` folder.
- Airflow does not import Python modules from AIRFLOW_HOME (but import it from the `dags` and `plugins`)
- (at least in this version) we want our users to create pipelines with `dlt init` and run them as usual locally. then be able to deploy with airflow similarly as they do deploy with github actions,

the solution

1. the whole project created with `dlt init` goes in to the `dags` folder of Airflow
2. on local Airflow I just do a symlink to the projects from dag folder
3. on composer I copy the root of the repo to dags in the cloud build

I make sure the `.dlt` folder in `dags` is used to get configs - the wrapper does it

The final version looks like that

```lua
$AIRFLOW_HOME/
|-- dags/
|   |-- my_dag.py
|   |-- .dlt/
|     |-- config.toml
|   |-- pipedrive/
|   |-- __init__.py
|   |-- pipedrive_pipeline.py
```

**Tests**
We need tests in `tests` for this. The test should assume working local airflow instance. The test will be executed on github CI in separate workflow
1. * [x] implement tests for a few DAG cases, preferably using duckdb to verify loaded data
5. * [x] add definiton for github actions workflow that spins up airflow and executes tests

rudolfix:
closes #190 
closes #167 
closes #140

### DIFF
--- a/dlt/helpers/airflow.py
+++ b/dlt/helpers/airflow.py
@@ -1,11 +1,255 @@
-from typing import Any
+import os
+from tempfile import gettempdir
+from typing import Any, Callable, List, Literal, Sequence
+from tenacity import retry_if_exception, wait_exponential, stop_after_attempt, Retrying, RetryCallState
 
 from dlt.common.exceptions import MissingDependencyException
 
 try:
+    from airflow.configuration import conf
     from airflow.utils.task_group import TaskGroup
+    #from airflow.decorators import task
+    from airflow.operators.python import PythonOperator
+    from airflow.operators.python import get_current_context
 except ImportError:
     raise MissingDependencyException("Airflow", ["airflow>=2.0.0"])
 
 
-# def run(data: Any) -> TaskGroup:
+import dlt
+from dlt.common import logger
+from dlt.common.schema.typing import TWriteDisposition
+from dlt.common.utils import uniq_id
+from dlt.common.configuration.container import Container
+from dlt.common.configuration.specs.config_providers_context import ConfigProvidersContext
+from dlt.common.runtime.collector import NULL_COLLECTOR
+
+from dlt.extract.source import DltSource
+from dlt.pipeline.helpers import retry_load
+from dlt.pipeline.pipeline import Pipeline
+from dlt.pipeline.progress import log
+from dlt.pipeline.typing import TPipelineStep
+
+
+DEFAULT_RETRY_NO_RETRY = Retrying(stop=stop_after_attempt(1), reraise=True)
+DEFAULT_RETRY_BACKOFF = Retrying(stop=stop_after_attempt(5), wait=wait_exponential(multiplier=1.5, min=4, max=10), reraise=True)
+
+
+class PipelineTasksGroup(TaskGroup):
+    """
+    Represents a DLT Airflow pipeline task group.
+    """
+
+    def __init__(
+        self,
+        pipeline_name: str,
+        use_data_folder: bool = False,
+        local_data_folder: str = None,
+        use_task_logger: bool = True,
+        log_progress_period: float = 30.0,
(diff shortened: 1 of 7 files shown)


dlt/helpers/airflow.py
```diff
@@ -1,11 +1,255 @@
...
+
+        with self:
+
+            # use factory function to make test, in order to parametrize it. passing arguments to task function (_run) is serializing them and
+            # running template engine on them
+            def make_task(pipeline: Pipeline, data: Any) -> PythonOperator:
+
+                def _run() -> None:
+                    # drop local data
+                    task_pipeline = pipeline.drop()
```
AstrakhantsevaAA, line 172: should we warn people that all local data are deleting?

AstrakhantsevaAA (approved):
this PR is a real pleasure for the eyes

PR #358, opened 2023-05-25
Feat/Add transaction management for filesystem operations using fsspec

z3z1ma:


z3z1ma:
Let me know the right way to test this against some live cloud storages in CI. You can point me to an example file where you are running integration tests against a live S3/GCS or w/e.

dlt/common/storages/transactions.py
```diff
@@ -0,0 +1,210 @@
...
+
+        self._original_contents = self.read()
+        self._is_locked = True
+        if self._heartbeat is not None:
+            self._heartbeat.cancel()
+        self._heartbeat = self._make_heartbeat_thread()
+        self._heartbeat.start()
+        return True
+
+    def release_lock(self) -> None:
```
rudolfix, line 172: should we add `release_lock` to the class destructor? then the lock will be released when garbage collected and heartbeat will stop. I'm not sure about that. IMO the only safe way to use locks is with a context manager or within other managed objects. not releasing lock is a programmer's mistake and maybe we should keep the current behavior

z3z1ma: I think it would be fair for the destructor to make an attempt at signalling the heartbeat to stop refreshing the mtime of the file but perhaps not to explicitly drop the lock. If a lock goes out of scope and hits the `gc`, it would be a non-obvious release of the lock before its TTL has expired.

z3z1ma:
> also, can we lock folders with it or create locks that do not correspond to files? looking at your code - yes. but this case is not tested

Yes if it is not a file, it should work all the same. In that case `rollback` and `read` are simply no-ops. So a directory lock would just be `TransactionalFile(path="path/to/a/dir", fs=gcs)`. Backup and restore of a directory is not in scope for this lock manager. That atomicity should be handled by higher level code which uses this lock to ensure safe access to the directory and can be more prescriptive.

I will add test for this too.

rudolfix (changes requested):
thanks for all the tests. the code is good now and can be merged. there are two things left
1. the linter is complaining (dlt/common/storages/transactions.py:115: error: Returning Any from function declared to return "Optional[bytes]"  [no-any-return]). are you able to run the linter locally? `make lint` after `make dev`. why it complain about pydantic IDK. let's fix this first
2. then it seems your lock does not pass tests on windows (hangs forever). 99% the reason is that the path to lock you provide to `__init__` is in diffrent format (slash/backslash etc.). and the locks is never acquired because paths will never be equal. you probably want to normalize path to posix if you work on local filesystem. if you do not have windows box to test I can also do that. let me know

dlt/common/storages/transactions.py
```diff
@@ -0,0 +1,223 @@
...
+        if self._heartbeat is not None:
+            self._heartbeat.cancel()
+            self._heartbeat = None
+
+    def _sync_locks(self) -> t.List[str]:
+        """Gets a map of lock paths to their last modified times and removes stale locks."""
+        output = []
+        now = datetime.now()
+
+        for lock in self._fs.glob(self.lock_prefix + ".*", refresh=True, detail=True).values():
```
rudolfix, line 101: @z3z1ma I've just went through fsspec glob implementation because it often returned no results when I was globbing in one process and deleting files from the same "folder" in another. the implementation is quite complicated and I still do not know why the problem happens. it surely tries to do recursive walk (which is normal for glob) which is not necessary here.
I ended up using `ls` and filtering by prefix myself. I problems gone. I think you should do the same here

z3z1ma:
Okay changes are all made. 🎉  This is sorely needing testing for Windows and at least 1 (ideally all 3) of the cloud storage providers as, now that I have removed `glob` and also added the `normpath` which hopefully helps on windows, its a bit different than what has been running with us for awhile (which was designed for *nix and GCS).

z3z1ma:
Tried another thing to help with windows path but is more trouble than its worth trying to mess with it vs you having a windows box to test on because I am just taking educated guesses. 

I am sure now that I normalize in the `init` or try to at least, that the `self._fs.ls` operation should do the same to its results during lock synchronization

rudolfix:
@z3z1ma got working this on windows and merged all your recent changes. the new PR is here: #358

PR #376, opened 2023-06-04
brings state info and cli up to date

rudolfix:


TyDunn (approved):
I don't have a lot of context on the content itself, so I focused on presentation, and it looks good overall!

PR #380, opened 2023-06-05
355 - add computed exhausted property

sh-rp:


dlt/extract/source.py
```diff
@@ -614,6 +614,15 @@ def root_key(self) -> bool:
         config = RelationalNormalizer.get_normalizer_config(self._schema).get("propagation")
         return config is not None and "root" in config and "_dlt_id" in config["root"] and config["root"]["_dlt_id"] == "_dlt_root_id"
 
+    @property
+    def exhausted(self) -> bool:
+        # check all selected pipes wether one of them is exhausted
```
rudolfix, line 619: better to use docstrings

dlt/extract/source.py
```diff
@@ -614,6 +614,15 @@ def root_key(self) -> bool:
...
+        for resource in self._resources.extracted.values():
+            for item in resource._pipe:
```
rudolfix, line 621: you do not need to enumerate items. `Pipe` has `gen` property which is the main generator of the pipe. all others are transforms. so check only `gen`

dlt/extract/source.py
```diff
@@ -614,6 +614,15 @@ def root_key(self) -> bool:
...
+                if isinstance(item, types.GeneratorType):
```
rudolfix, line 622: interesting! I'm using `inspect.isgenerator` and it's probably better to stick to that (or you double check if the above catches all cases of generators)

rudolfix (changes requested):
please link to issue in PR. maybe we should a template?

tests/extract/test_sources.py
```diff
@@ -819,3 +819,15 @@ def some_data():
...
     info = str(s)
     assert "Source is already iterated" in info
+
+
+def test_exhausted_after_first_iter_fetch() -> None:
+
+    def some_data():
+        yield from [1, 2, 3, 4]
+
+    s = DltSource("source", "module", Schema("source"), [dlt.resource(some_data())])
```
rudolfix, line 829: all good, but please add more resource types to source
- a regular list (iterable)
- an iterator (ie. on a list)
- join gen resource to transformer

I just want to make sure all cases are passing, you can do it on one test

rudolfix (changes requested):
also the linter is not passing

rudolfix (approved):
OK! finally we have all the tests and CI is passing!

PR #381, opened 2023-06-05
Feat: Bigquery load performance increase

z3z1ma:
The largest latency is by far network related during BigQuery loads. Especially when spooling many workers. They compete over network. It is much more efficient to spend the cpu cycles to 1-shot compress data and then send over the network. The advantages of BigQuery being able to read non-compressed data in parallel are moot given the network latency. I have found huge speedups with this locally. There can be a deferred consideration down the line for how dlt can handle compression more holistically as on another big load, I overloaded the disk memory of a Pod during an atomic move with a lot of data which would have likely taken much less room compressed if it were an option.

### DIFF
--- a/dlt/destinations/bigquery/bigquery.py
+++ b/dlt/destinations/bigquery/bigquery.py
@@ -242,7 +242,7 @@ class BigQueryClient(SqlJobClientBase):
                 max_bad_records=0)
             with open(file_path, "rb") as f:
                 return self.sql_client.native_connection.load_table_from_file(
-                        io.BytesIO(gzip.compress(f)),  # network latency is the bottleneck so compress the data here
+                        io.BytesIO(gzip.compress(f.read())),  # network latency is the bottleneck so compress the data here
                         self.sql_client.make_qualified_table_name(table_name, escape=False),
                         job_id=job_id,
                         job_config=job_config,
--- a/dlt/destinations/bigquery/bigquery.py
+++ b/dlt/destinations/bigquery/bigquery.py
@@ -1,3 +1,5 @@
+import gzip
+import io
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


dlt/destinations/bigquery/bigquery.py
```diff
@@ -240,7 +242,7 @@ def _create_load_job(self, table_name: str, write_disposition: TWriteDisposition
                 max_bad_records=0)
             with open(file_path, "rb") as f:
                 return self.sql_client.native_connection.load_table_from_file(
-                        f,
+                        io.BytesIO(gzip.compress(f.read())),  # network latency is the bottleneck so compress the data here
```
rudolfix, line 245: can't believe that they do not compress in transit! there are a few things that you'd need to improve in order to merge it:
1, should be optional. two reasons (https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json#limitations)
- gzip file max size is 4GB
- what you do above reads the file fully into memory
2. it compresses no matter what is the file type. we are adding parquet and avro so I doubt it is going to work

to fix that:
1. check if file type is eligible: get file format from ParsedLoadJobFileName.parse(file_name) and then DataWriter.data_format_from_file_format().is_binary_format must be false
2. add a config flag to BigQueryClientConfiguration
3. open a compressed gzip stream on the file, do not read full file into mem
4. if you have (3) working you may enable the compress option by default

rudolfix (changes requested):
this is really a good catch. I wonder if we should compress files that we store in the buckets... please see the review for improvements that we need

rudolfix:
@z3z1ma now when I think of it, it's better to add option to compress files to load storage. then it will work everywhere including bucket destination.

PR #385, opened 2023-06-07
update docusaurus and fix gtag

sh-rp:


rudolfix (commented):
Cant say this is right. Please confirm if we really use this tags maybe with @TyDunn or @adrianbr ?

TyDunn:
I don't believe we use any GTags at the moment

rudolfix (approved):

PR #392, opened 2023-06-08
adds duckdb->bigquery walkthrough

rudolfix:


docs/website/docs/walkthroughs/share-a-dataset.md
```diff
@@ -0,0 +1,108 @@
...
+---
+
+# Share a dataset: duckdb -> BigQuery
+In previous walkthroughs you used the local stack to create and run your pipeline. This saved you the headache of setting up cloud account, credentials and often also money. Our choice for local "warehouse" is `duckdb`, fast, feature rich and working everywhere. However at some point you want to move to production or share the results with your colleagues. The local `duckdb` file is not sufficient for that! Let's move the dataset to BigQuery now.
+
+## 1. Replace the "destination" argument with "bigquery"
+```python
+if __name__=='__main__':
+
+    # below we replaced "duckdb" in the "destination" argument with "bigquery"
```
TyDunn, line 14: @rudolfix Is `weatherapi` a verified pipeline or is this something folks would have to create themselves? If not, maybe we switch to using an example where the pipeline exists already?

rudolfix: OK! converted to chess api:
1. same pipeline used in run-a-pipeline
2. will be way easier to create a recipe!

Issue #388, opened 2023-06-07
support compression in data writers


rudolfix:
**Background**
`dlt` uses files to pass data from extract->normalize and normalize->load. those files are often text files and may be compressed. We can compress files on the fly by opening a gzip stream and passing it to a data writer. Compression should be optional with default value in `TFileFormatSpec` and a flag in `BufferedDataWriterConfiguration` to enable/disable.

The biggest benefit is the network i/o for staging operations. There may be also benefits if the storage is a mapped bucket.

**Limitations**
- some system ie BigQuery cannot deal with archives > 4GB
- as above: parallel loading may be disable

**Tasks**
1. * [x] all text files (see `TFileFormatSpec`) support compression. We add a new attribute to it `compress_default` that tells the default behavior if `compress` flag in config is not present.
2. * [x] all text writers should have `compress_default` flag enabled
3. * [x] the default config value is None
4. * [x] implement the gzip in `buffered.py` by using gzip stream instead of text stream if file support that and combination of defaults and flags sets it to on
5. * [x] implement a helper method that automatically opens a gzip stream and fallbacks to text if file is not gzip. You can add it to FileStorage
6. * [x] known places that will need that open method: normalize worker process (normalize.py) insert sql job (InsertValuesLoadJob)

**Tests**
1. * [x] tests for low level functions (buffered writers and opening gzips with fallbacks)
2. * [x] test enable/disable compression with configs
3. * [x] due to defaults all our pipelines will fallback to compression. please add a test in load/pipelines that will disable it and run on all destinations

PR #405, opened 2023-06-14
Feat: Use compression to maximize network and disk space efficiency 

z3z1ma:
Adds compression support

### DIFF
--- a/dlt/common/data_writers/writers.py
+++ b/dlt/common/data_writers/writers.py
@@ -16,6 +16,7 @@ class TFileFormatSpec:
     is_binary_format: bool
     supports_schema_changes: bool
     requires_destination_capabilities: bool = False
+    compress_default: bool = False
 
 
 class DataWriter(abc.ABC):
@@ -86,7 +87,13 @@ class JsonlWriter(DataWriter):
 
     @classmethod
     def data_format(cls) -> TFileFormatSpec:
-        return TFileFormatSpec("jsonl", "jsonl", True, True)
+        return TFileFormatSpec(
+            "jsonl",
+            file_extension="jsonl",
+            is_binary_format=True,
+            supports_schema_changes=True,
+            compress_default=True,
+        )
 
 
 class JsonlListPUAEncodeWriter(JsonlWriter):
@@ -101,7 +108,13 @@ class JsonlListPUAEncodeWriter(JsonlWriter):
 
     @classmethod
     def data_format(cls) -> TFileFormatSpec:
-        return TFileFormatSpec("puae-jsonl", "jsonl", True, True)
+        return TFileFormatSpec(
+            "puae-jsonl",
+            file_extension="jsonl",
+            is_binary_format=True,
+            supports_schema_changes=True,
+            compress_default=True,
+        )
 
 
 class InsertValuesWriter(DataWriter):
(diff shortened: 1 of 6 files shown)


z3z1ma:
Closes #388

dlt/common/data_writers/writers.py
```diff
@@ -152,4 +165,10 @@ def write_footer(self) -> None:
 
     @classmethod
     def data_format(cls) -> TFileFormatSpec:
-        return TFileFormatSpec("insert_values", "insert_values", False, False, requires_destination_capabilities=True)
+        return TFileFormatSpec(
+            "insert_values",
+            file_extension="insert_values",
+            is_binary_format=False,
+            supports_schema_changes=False,
+            requires_destination_capabilities=True,
```
rudolfix, line 173: compress this one as well!

dlt/common/storages/file_storage.py
```diff
@@ -251,3 +254,14 @@ def rmtree_del_ro(action: AnyFun, name: str, exc: Any) -> Any:
...
             else:
                 os.remove(name)
+
+    @staticmethod
+    def open_zipsafe(path: str, mode: str = "r", encoding: Optional[str] = None, **kwargs: Any) -> IO[Any]:
+        """Opens a file using gzip.open if it is a gzip file, otherwise uses open."""
+        try:
+            f = gzip.open(path, mode, encoding=encoding, **kwargs)
+            # Force gzip to read the first few bytes and check the magic number
+            f.read(2), f.seek(0)
```
rudolfix, line 264: cool that zipstream supports seek. i wonder how that is implemented (are they using internal buffer, reopen the file etc.)

dlt/common/storages/file_storage.py
```diff
@@ -251,3 +254,14 @@ def rmtree_del_ro(action: AnyFun, name: str, exc: Any) -> Any:
                 os.rmdir(name)
             else:
                 os.remove(name)
+
+    @staticmethod
+    def open_zipsafe(path: str, mode: str = "r", encoding: Optional[str] = None, **kwargs: Any) -> IO[Any]:
+        """Opens a file using gzip.open if it is a gzip file, otherwise uses open."""
+        try:
+            f = gzip.open(path, mode, encoding=encoding, **kwargs)
```
rudolfix, line 262: ohhh the encoding will probably not work here! see our CI and common tests. also mode may contain "b" or "t". this needs to be unit tested

dlt/common/storages/file_storage.py
```diff
@@ -251,3 +254,14 @@ def rmtree_del_ro(action: AnyFun, name: str, exc: Any) -> Any:
                 os.rmdir(name)
             else:
                 os.remove(name)
+
+    @staticmethod
+    def open_zipsafe(path: str, mode: str = "r", encoding: Optional[str] = None, **kwargs: Any) -> IO[Any]:
```
rudolfix, line 259: this opens only for read. two options here:
- make it work with write (and use in data writers :))
- make sure that mode contains "r" with assert and maybe rename the method

rudolfix (changes requested):
this is cool :) I'm surprised that it could be implemented in such a simple way
Looking at this implementation I see that the ticket skipped the important case of binary streams that also can be compressed.
- let's rename `compress_default` to ` supports_compression` - this indicates that data can be compressed, regardless if this is text stream or not
- the ticket didn't precise it but we should add an option to disable compression to `BufferedDataWriterConfiguration` (it just mentioned to configuration)
- we need the specified tests. even
- even if you do the PR from a fork the common and airflow tests must pass

dlt/common/storages/file_storage.py
z3z1ma: Internal buffer! :)

rudolfix:
resubmitted as #415 all tests are passing. amazing work! we'll be releasing this over the weekend

PR #406, opened 2023-06-14
fix link in doc

sh-rp:


rudolfix (changes requested):
1. there are many links like this in the docs. I probably didn't find all. everything that starts with `./` is suspicious
2. there's also README and the library code that links to documentation. those links are not checked at all by docosaurus and need to be fixed :/ 
3. I'm afraid we have similar situation in verified pipeline repo (building blocks)

rudolfix (dismissed):

PR #409, opened 2023-06-14
Added Notion Docs

dat-a-man:


docs/website/docs/dlt-ecosystem/verified-sources/notion.md
```diff
@@ -0,0 +1,124 @@
...
+    ```bash
+    dlt pipeline <pipeline_name> show
+    ```
+    
+    For example, the pipeline_name for the above pipeline example is `notion`, you may also use any custom name instead.
+    
+
+## Customizations
+
+To load data to the destination using `dlt`, you have the option to write your own methods. It is important to note that this method retrieves data from the databases.
```
AstrakhantsevaAA, line 104: > It is important to note that this method retrieves data from the databases.

i would change this sentence.  "this method retrieves data from all existed databases in your Notion account"

> It is important to note 

it duplicates line 122. rephrase it please.

Issue #22, opened 2022-06-15
[destination] implement loader for Snowflake with PUT stage


rudolfix:
**Background**
Implement Snowflake destination based on PUT stage. It will be quite similar to the `BigQuery` implementation (just simpler). 

**Tasks**
* [x] use `jsonl` as preferred type
* [x] use `PUT` to stage
* [x] allow option to remove files when COPY INTO succeeded.
* [x] our jsonl uses base64 to store binary so enable BINARY_FORMAT  in COPY INTO https://docs.snowflake.com/en/sql-reference/sql/copy-into-table
* [ ] fail on any error on COPY INTO, return up to 100 errors
* [x] i think you must use TRUNCATE for replace load, do TRUNCATE COPY in one transaction!
* [x] implement cluster hint with CLUSTER BY 

**Authentication**
Derived from ConnectionStringCredentials:
1. * [x] must understand sql alchemy syntax https://docs.snowflake.com/en/developer-guide/python-connector/sqlalchemy
2. * [x] should expose several query parameters as fields `warehouse` `role`
3. * [x] key pair auth should be supported https://docs.snowflake.com/en/developer-guide/python-connector/sqlalchemy#key-pair-authentication-support

**Tests**
1. * [ ] we should enable it as destination in tests. all tests using ALL_DESTINATIONS must pass
2. * [ ] it should pass the same common tests which BigQuery job client and sql client is passing
3. * [ ] note that to pass sql_client tests you'll need to map DBApi exceptions into 3 categories that dlt needs: relation not found, terminal exceptions, transient exceptions. see bigquery and postgres implementations. that's manual work

all in all writing the test suite is more work than the code here

rudolfix:
@adrianbr if you have any comments to the ticket...

PR #414, opened 2023-06-15
Feat/snowflake destination

steinitzu:
closes https://github.com/dlt-hub/dlt/issues/22


dlt/destinations/snowflake/snowflake.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+SCT_TO_BQT: Dict[TDataType, str] = {
+    "complex": "VARIANT",
+    "text": "STRING",
+    "double": "FLOAT",
+    "bool": "BOOLEAN",
+    "date": "DATE",
+    "timestamp": "TIMESTAMP_TZ",
+    "bigint": f"NUMBER({BIGINT_PRECISION}, 0)",  # Snowflake has no integer types
```
rudolfix, line 35: amazing 🤷

dlt/pipeline/pipeline.py
```diff
@@ -352,6 +352,7 @@ def load(
             self.first_run = False
             return info
         except Exception as l_ex:
+            raise
```
rudolfix, line 355: huh?

dlt/destinations/snowflake/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+    caps.supported_loader_file_formats = ["jsonl"]
+    caps.escape_identifier = escape_postgres_identifier
+    caps.max_identifier_length = 63
+    caps.max_column_identifier_length = 63
+    caps.max_query_length = 32 * 1024 * 1024
+    caps.is_max_query_length_in_bytes = True
+    caps.max_text_data_type_length = 1024 * 1024 * 1024
+    caps.is_max_text_data_type_length_in_bytes = True
+    caps.supports_ddl_transactions = True
+    caps.alter_add_multi_column = False
```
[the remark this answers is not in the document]
steinitzu: It says so in the docs, but I got errors with it. I think it might be different syntax with just one `ADD COLUMN` clause, I have to try this:
```ADD COLUMN col1 NUMBER, col2 NUMBER```

File: dlt/common/data_writers/escape.py [hunk 4]
```diff
@@ -68,3 +68,9 @@ def escape_redshift_identifier(v: str) -> str:
 def escape_bigquery_identifier(v: str) -> str:
     # https://cloud.google.com/bigquery/docs/reference/standard-sql/lexical
     return "`" + v.replace("\\", "\\\\").replace("`","\\`") + "`"
+
+
+def escape_snowflake_identifier(v: str) -> str:
+    # Snowcase uppercase all identifiers unless quoted. Match this here so queries on information schema work without issue
+    # See also https://docs.snowflake.com/en/sql-reference/identifiers-syntax#double-quoted-identifiers
+    return v.upper()
```
rudolfix, line 76: we asked our snowflake buddies what they use typically. let's see what they say

dlt/destinations/snowflake/__init__.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def capabilities() -> DestinationCapabilitiesContext:
+    caps = DestinationCapabilitiesContext()
+    caps.preferred_loader_file_format = "jsonl"
+    caps.supported_loader_file_formats = ["jsonl"]
+    caps.escape_identifier = escape_snowflake_identifier
+    caps.max_identifier_length = 255
+    caps.max_column_identifier_length = 255
+    caps.max_query_length = 32 * 1024 * 1024
```
rudolfix, line 26: https://docs.snowflake.com/en/user-guide/query-size-limits so I'd put maybe 2MB

dlt/destinations/snowflake/configuration.py
```diff
@@ -0,0 +1,39 @@
...
+from dlt.common.configuration import configspec
+from dlt.common.destination.reference import DestinationClientDwhConfiguration
+
+
+@configspec
+class SnowflakeCredentials(ConnectionStringCredentials):
+    drivername: Final[str] = "snowflake"  # type: ignore[misc]
+    password: TSecretStrValue = None
+    host: str = None
+    database: Optional[str] = None
```
rudolfix, line 16: database must be set

rudolfix (changes requested):

dlt/common/data_writers/escape.py
steinitzu: Forgot to take into account escaping sql keywords also, e.g. : `CREATE TABLE xxx (FROM VARCHAR)`  
Thinking now we should both quote and upcase.

rudolfix: it seems that people mostly ignore it: use the original casing without quotes. this of course translates to upper case internally. that for example makes dbt packages to look identical for different systems. I'm still waiting for more input. sql keywords? aren't they case insensitive as well?

steinitzu: I mean with keywords if you have a column named `from` or `update` or something. Those need to be quoted.

So I think if we uppercase and quote in the escape function: `some_column` -> `"SOME_COLUMN"` makes things work pretty seamlessly, least resistance to how snowflake works.

Then queries with unquoted cols like `SELECT some_column FROM ...` work since snowflake translates it to uppercase implicitly.

dlt/destinations/snowflake/__init__.py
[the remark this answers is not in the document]
steinitzu: This syntax works. I extracted a separate method for the `ADD COLUMN` generation and overrode it in the snowflake client: https://github.com/dlt-hub/dlt/pull/414/commits/69abda3bd0ad21ab94a71125207711ebffe57d08

dlt/destinations/snowflake/snowflake.py
```diff
@@ -0,0 +1,164 @@
...
+            stage_name = f'"{table_name}_{load_id}"'
+            # Create stage and file format
+            client.execute_sql(
+                """
+                CREATE FILE FORMAT IF NOT EXISTS dlt_jsonl_format
+                TYPE = 'JSON'
+                BINARY_FORMAT = 'BASE64'
+                """
+            )
+            client.execute_sql(f"CREATE STAGE IF NOT EXISTS {stage_name} FILE_FORMAT = dlt_jsonl_format")
```
rudolfix, line 67: please do not create a stage. use the table implicit stage: https://docs.snowflake.com/en/user-guide/data-load-local-file-system-create-stage

use the `load_id` as the folder in stage from which you will copy

let the user specify a stage name in the configuration. if the stage name is available, this stage will be used

dlt/destinations/snowflake/sql_client.py
```diff
@@ -0,0 +1,164 @@
...
+                return None
+            else:
+                f = curr.fetchall()
+                return f
+
+    @contextmanager
+    @raise_database_error
+    def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DBApiCursor]:
+        curr: DBApiCursor = None
+        db_args = args if args else kwargs if kwargs else None
```
rudolfix, line 93: I'd deal with the multiple statements here. Unfortunately snowflake does not really follow dbapi spec regarding multiple results sets in cursors. instead they just returns many cursors. so let's do the following:
1. if dbargs is None then use `execute_string`: https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-api#object-connection - it returns many cursor so let's return last cursor (the most typical use case)
2. else just execute assuming that all parametrized queries are single statement

this will IMO cover all DDL and merge jobs (they are not parametrized). all other parametrized statements are single. (if there's something in tests then rather fix the tests)

i prefer this than hacking the code to cover for snowflake not implementing dbapi

dlt/destinations/snowflake/sql_client.py
```diff
@@ -0,0 +1,164 @@
...
+            if curr.description is None:
+                return None
+            else:
+                f = curr.fetchall()
+                return f
+
+    @contextmanager
+    @raise_database_error
+    def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DBApiCursor]:
+        curr: DBApiCursor = None
```
rudolfix, line 92: snowflake has a nice API to get dataframes and chunked dataframes from cursor. could you implement something like
```
class BigQueryDBApiCursorImpl(DBApiCursorImpl):
    """Use native BigQuery data frame support if available"""
    native_cursor: BQDbApiCursor  # type: ignore

    def df(self, chunk_size: int = None, **kwargs: Any) -> DataFrame:
        query_job: bigquery.QueryJob = self.native_cursor._query_job

        if chunk_size is None:
            try:
                return query_job.to_dataframe(**kwargs)
            except ValueError:
                # no pyarrow/db-types, fallback to our implementation
                return super().df()
        else:
            return super().df(chunk_size=chunk_size)
```

btw. this could also be used to wrap mutliple cursors returned from snowflake into something dbapi-like but I think it is a waste of time

rudolfix (changes requested):
please also fix the ci jobs so they run test selectively. maybe let's run snowflake jobs in separate workflow.

dlt/destinations/snowflake/sql_client.py
steinitzu: Turns out multi statements are actually supported! Only documented in the rest API, https://docs.snowflake.com/en/developer-guide/sql-api/submitting-multiple-statements

But also works in python: `cur.execute("SELECT 1; SELECT 2;", num_statements=0)`  

dbapi cursor seems to work normally with this, `cursor.nextset()`, etc works.  

`execute_string` splits the string and sends one query at a time so this is better.

dlt/destinations/job_client_impl.py
```diff
@@ -42,8 +42,13 @@ def __init__(self, file_path: str, sql_client: SqlClientBase[Any]) -> None:
         # execute immediately if client present
         with FileStorage.open_zipsafe_ro(file_path, "r", encoding="utf-8") as f:
             sql = f.read()
+        if sql_client.capabilities.supports_multiple_statements:
+            statements = [sql]
+        else:
+            statements = sql.splitlines()
```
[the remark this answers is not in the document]
steinitzu: I changed the `SqlMergeJob` implementation to write one statement per line in the file.   
Found it kinda sketchy to split on `;` since it can appear anywhere in strings and comments too. Maybe overkill to consider that here...  

But will remove this anyway due to https://github.com/dlt-hub/dlt/pull/414#discussion_r1233439895

dlt/destinations/snowflake/snowflake.py
```diff
@@ -0,0 +1,164 @@
...
+class SnowflakeLoadJob(LoadJob, FollowupJob):
+    def __init__(self, file_path: str, table_name: str, write_disposition: TWriteDisposition, load_id: str, client: SnowflakeSqlClient) -> None:
+        file_name = FileStorage.get_file_name_from_file_path(file_path)
+        super().__init__(file_name)
+
+        with client.with_staging_dataset(write_disposition == "merge"):
+            qualified_table_name = client.make_qualified_table_name(table_name)
+            stage_name = f'"{table_name}_{load_id}"'
+            # Create stage and file format
+            client.execute_sql(
```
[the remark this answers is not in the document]
steinitzu: Done!

dlt/destinations/snowflake/snowflake.py
steinitzu: Updated default stage  https://github.com/dlt-hub/dlt/pull/414/commits/bf322ecd8a972290e7c28d652244c585ef30df17#diff-577441894b0e65b848310ea282c52bab00b6f649eb2c1b9402a5423d70e4d3e4

dlt/destinations/snowflake/snowflake.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, file_path: str, table_name: str, write_disposition: TWriteDis
             stage_name = client.make_qualified_table_name('%'+table_name)
             with client.begin_transaction():
                 # PUT and copy files in one transaction
-                client.execute_sql(f'PUT file://{file_path} @{stage_name}/"{load_id}" OVERWRITE = TRUE')
+                client.execute_sql(f'PUT file://{file_path} @{stage_name}/"{load_id}" OVERWRITE = TRUE, SOURCE_COMPRESSION = GZIP')
```
steinitzu, line 63: @rudolfix is there a good way to tell whether the file is gzipped? Is that something we need to worry about?  
Just saw this was added, `AUTO_DETECT` doesn't work here.

rudolfix: all the files are compressed by default including jsonl. this is already ON in this branch.
looking at this: https://docs.snowflake.com/en/sql-reference/sql/put#optional-parameters
AUTO_COMPRESS=FALSE and SOURCE_COMPRESSION = AUTO_DETECT
what does it mean it does not work? if it indeed does not then we have a problem... we could detect the compression ourselves but some files will be in the buckets and that makes them harder. why they even check that during staging...

dlt/common/destination/capabilities.py
```diff
@@ -30,6 +30,8 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     supports_ddl_transactions: bool
     naming_convention: str = "snake_case"
     alter_add_multi_column: bool = True
+    # Can send multiple SQL statements in one execute call
+    supports_multiple_statements: bool = True
```
rudolfix, line 34: we do not need that...

dlt/destinations/snowflake/configuration.py
```diff
@@ -0,0 +1,74 @@
...
+@configspec
+class SnowflakeCredentials(ConnectionStringCredentials):
+    drivername: Final[str] = "snowflake"  # type: ignore[misc]
+    password: Optional[TSecretStrValue] = None
+    host: str = None
+    database: str = None
+    warehouse: Optional[str] = None
+    role: Optional[str] = None
+    private_key: Optional[TSecretStrValue] = None
+    private_key_passphrase: Optional[TSecretStrValue] = None
```
rudolfix, line 38: please add `password`, `warehouse` and `role` like `location` for `GcpCredentials`:
```
__config_gen_annotations__: ClassVar[List[str]] = ["location"]
```
why we do that: so when the secrets are auto generated for the user s/he sees those even if they are optionsl

dlt/destinations/snowflake/configuration.py
```diff
@@ -0,0 +1,74 @@
...
+from dlt.common.typing import TSecretStrValue
+from dlt.common.configuration.specs import ConnectionStringCredentials
+from dlt.common.configuration import configspec
+from dlt.common.destination.reference import DestinationClientDwhConfiguration
+
+
+def _read_private_key(private_key: str, password: Optional[str] = None) -> bytes:
+    """Load an encrypted or unencrypted private key from string.
+    """
+    from cryptography.hazmat.backends import default_backend
```
rudolfix, line 14: OK I see there's no other way to decrypt and parse pkey. Let's throw the usual `MissingDependencyException` with some explanation what we need and why

rudolfix:
@steinitzu for some reason snowflake tests are not running

dlt/destinations/snowflake/configuration.py
steinitzu: Raising now with the "install dlt[snowflake]" message, this is a snowflake dependency.

dlt/destinations/snowflake/sql_client.py
steinitzu: https://github.com/dlt-hub/dlt/pull/414/files#diff-36ef69a1f6bf9feca5e587f6594e04c9750f29d36f5a75d00fcefef7c7226ac0R13-R20  
Opted to keep the df columns in their original uppercase, not sure if better to lowercase them. This way it matches what's in the db

dlt/destinations/snowflake/snowflake.py
steinitzu: Added the config option to use a different stage, it gets created if it doesn't exist. And optionally delete the stage file after copy: https://github.com/dlt-hub/dlt/pull/414/commits/6853c2747e5b5bac1c2a77233292ef81a9a541bc

dlt/destinations/snowflake/sql_client.py
rudolfix: we should mess with the resultsets that snowflake returns

rudolfix (dismissed):

rudolfix (approved):

Issue #379, opened 2023-06-05
support round robin in pipe iterator


rudolfix:
**Background**
Current iteration strategy exhaust generator steps in pipes in order they were passed to the iterator. This is to have deterministic behavior that you would get from a normal iterator. However there are the cases when we want to iterate with other strategies ie. round robin where we take next item from the next generator, not exhausting the first one first.

Currently the process works as follows:
0. all generator steps in pipes (resources) are added to iterator in reverse.
1. iterator always takes an item from the last generator in the list
2. the items is pushed through the pipeline. it may add more generators to a list
3. item is yielded
4. go to 1

**as a consequence: all the generators that were created by the item will be exhausted (we take last element from the list) before next item form any of the original gens is evaluated**. this guarantees that items are leaving pipeline and this property should be preserved.

**Tasks**
1. * [x] if `gen` yields `None` get next item immediately (default or round robin)
2. * [x] add new config ('next_item_mode`) option to `PipeIteratorConfiguration` that can be `always_last` (current) and `round_robin`
3. * [x]  add option to `extract` method of `Pipeline` to pass that explicitly
4. * [x] it think that `_get_source_item` method is quite good so we provide separate implementation for round robin (method or we subclass)
5. * [x] round robin should only apply to the gens created in step 0 (see above), not to gens that were added as consequence of processing items. in other words even in round robin we still evaluate new gens fully before coming back to old gens
6. * [x] in round robin mode we keep an index of last gen and subtract it modulo

The desired consequence is that if gens return deferred or awaitable items the futures queue fills up to `max_parallel_items` and will execute them in parallel. mind that for sync gens no parallelism will be achieved - just the resources will be evaluated round robin

PR #421, opened 2023-06-15
379 round robin pipe iterator

sh-rp:
implements https://github.com/dlt-hub/dlt/issues/379

### DIFF
--- a/dlt/extract/pipe.py
+++ b/dlt/extract/pipe.py
@@ -435,7 +435,8 @@ class PipeIterator(Iterator[PipeItem]):
         self.workers = workers
         self.futures_poll_interval = futures_poll_interval
 
-        self._round_robin_index: int = 0
+        self._round_robin_index: int = -1
+        self._initial_sources_count: int = 0
         self._async_pool: asyncio.AbstractEventLoop = None
         self._async_pool_thread: Thread = None
         self._thread_pool: ThreadPoolExecutor = None
@@ -458,6 +459,7 @@ class PipeIterator(Iterator[PipeItem]):
         extract = cls(max_parallel_items, workers, futures_poll_interval, next_item_mode)
         # add as first source
         extract._sources.append(SourcePipeItem(pipe.gen, 0, pipe, None))
+        cls._initial_sources_count = 1
         return extract
 
     @classmethod
@@ -499,6 +501,7 @@ class PipeIterator(Iterator[PipeItem]):
         for pipe in reversed(pipes):
             _fork_pipeline(pipe)
 
+        extract._initial_sources_count = len(extract._sources)
         return extract
 
     def __next__(self) -> PipeItem:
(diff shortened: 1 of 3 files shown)


dlt/extract/pipe.py
```diff
@@ -697,6 +708,39 @@ def _get_source_item(self) -> ResolvablePipeItem:
         except Exception as ex:
             raise ResourceExtractionError(pipe.name, gen, str(ex), "generator") from ex
 
+
+
+    def _get_source_item_round_robin(self) -> ResolvablePipeItem:
+        try:
+            # get items from last added iterator, this makes the overall Pipe as close to FIFO as possible
+            self._round_robin_index = self._round_robin_index % len(self._sources)
```
rudolfix, line 716: I'd start the index at -1 and increase it by 1 here then takin modulo in the next line. so we save on assignment. I know I'm pedantic but this is the most frequently called piece of code

dlt/extract/pipe.py
```diff
@@ -697,6 +708,39 @@ def _get_source_item(self) -> ResolvablePipeItem:
...
+                if isinstance(item, DataItemWithMeta):
+                    return ResolvablePipeItem(item.data, step, pipe, item.meta)
+                else:
+                    return ResolvablePipeItem(item, step, pipe, meta)
+        except StopIteration:
+            # we need to decrease the index to keep the round robin order
+            self._round_robin_index -= 1
+            # remove empty iterator and try another source
+            self._sources.pop(self._round_robin_index)
+            return self._get_source_item()
```
rudolfix, line 738: not the round robin again?

dlt/extract/pipe.py
```diff
@@ -670,13 +676,18 @@ def _get_source_item(self) -> ResolvablePipeItem:
         # no more sources to iterate
         if len(self._sources) == 0:
             return None
+        if self._next_item_mode == "current":
+            return self._get_source_item_current()
+        if self._next_item_mode == "round_robin":
```
rudolfix, line 681: I'd do elif because it is faster

tests/extract/test_extract_pipe.py
```diff
@@ -15,6 +15,35 @@
 # from tests.utils import preserve_environ
 
 
+def test_next_item_mode() -> None:
```
rudolfix, line 18: this is a good test!

rudolfix (changes requested):
there's one important things in the ticket https://github.com/dlt-hub/dlt/issues/379

that the newly added sources (gens) must be ALWAYS evaluated before those added by intialization.

execution of the step may create a generator. to preserve any deadlocks we evaluate the newly added generators without the round robin. see the ticket.

dlt/extract/pipe.py
sh-rp: I changed the code a bit, and it has to stay like this for round robin and current to work in conjunction

tests/extract/test_extract_pipe.py
sh-rp: it's even better now :D

rudolfix (changes requested):

tests/extract/test_extract_pipe.py
```diff
@@ -15,6 +15,43 @@
...
+        yield from [88, 89]
+
+    def nested_gen():
+        yield from [55, 56, nested_gen_level_2(), 77]
+
+    def source_gen1():
+        yield from [1, 2, nested_gen(), 3,4]
+
+    def source_gen2():
+        yield from range(11,16)
```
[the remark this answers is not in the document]
sh-rp: seems to work without a problem

tests/extract/test_extract_pipe.py
```diff
@@ -15,6 +15,43 @@
...
+            ]
+
+    # default mode is "current"
+    _l = list(PipeIterator.from_pipes(get_pipes(), next_item_mode="current"))
+    # items will be in order of the pipes, nested iterator items appear inline
+    assert [pi.item for pi in _l] ==  [1, 2, 55, 56, 88, 89, 77, 3, 4, 11, 12, 13, 14, 15, 20, 21]
+
+    # test mode "round robin"
+    reversed_pipes = get_pipes()
+    reversed_pipes.reverse() # reverse order of pipes for easier testing
```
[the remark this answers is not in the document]
sh-rp: done

rudolfix (approved):

PR #428, opened 2023-06-19
add colon in the end of sentence

TongHere:


rudolfix (changes requested):
hey! could you name the pull request so we know what info if modified and how? this let people to understand ie. after a week or a year what was really changed. this is why we use source control and git :)

TongHere:
> hey! could you name the pull request so we know what info if modified and how? this let people to understand ie. after a week or a year what was really changed. this is why we use source control and git :)

Thank you ! It had been changed.

rudolfix (approved):

PR #435, opened 2023-06-20
Feat: Spike sqlglot insert writer

z3z1ma:
There is probably many places in the codebase (anything SQL related) that there is better opportunity to leverage an AST instead of raw sring interpolation. With some customization, it may make the code more robust. 

This spike attempts to demonstrate an insert writer that can work across all sqlglot dialects via transpilation of the AST:

bigquery
clickhouse
databricks
drill
duckdb
hive
mysql
oracle
postgres
presto
redshift
snowflake
spark
sqlite
starrocks
teradata
trino
tsql

The AST also afford extremely unique manipulation opprtunities including finding and replacing nodes and so on.


sh-rp:
Closing due to inactivity

PR #445, opened 2023-06-21
enables parquet in snowflake and duckdb

rudolfix:


dlt/destinations/duckdb/duck.py
```diff
@@ -41,6 +47,24 @@
 }
 
 
+class DuckDbCopyJob(LoadJob):
+    def __init__(self, table_name: str, write_disposition: TWriteDisposition, file_path: str, sql_client: DuckDbSqlClient) -> None:
+        super().__init__(FileStorage.get_file_name_from_file_path(file_path))
+        qualified_table_name = sql_client.make_qualified_table_name(table_name)
```
steinitzu, line 53: This is always referencing the non-staging table, right? Should be done inside `with_staging_dataset` context.  
But the merge test runs so I'm not sure if this is right.

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -206,6 +206,23 @@ def test_merge_keys_non_existing_columns(destination_name: str) -> None:
         assert "m_a1" not in table_schema  # unbound columns were not created
 
 
+@pytest.mark.parametrize('destination_name', ["duckdb", "snowflake", "bigquery"])
+def test_pipeline_load_parquet(destination_name: str) -> None:
```
steinitzu, line 210: This only tests merge disposition, right?  
I would add replace test as well. Maybe add `loader_file_format` params to some existing replace tests, e.g. `[(dest, None) for dest in ALL_DESTINATIONS] + [('duckdb', 'parquet'), ('snowflake', 'parquet')]`

steinitzu (commented):
Added a couple of comments. But looks good overall!

dlt/destinations/duckdb/duck.py
rudolfix: I removed the merge with `replace` write disposition if you look closer at test.

rudolfix:
@steinitzu thx for finding the problem with merge. duckdb job was not generating followup jobs as well. this is all fixed. I also improved the test to have merge + replace

steinitzu (approved):

PR #447, opened 2023-06-22
documents snowflake destination

rudolfix:


AstrakhantsevaAA (dismissed):
there are problems with punctuation and code blocks, but I'm still doing refactoring, so I'll fix it all later so as not to delay this PR now

adrianbr (approved):

PR #448, opened 2023-06-22
dbt + snowflake

### LINKED ISSUE #434: support dbt-snowflake
**Background**
When #414 is merged, integrate `dbt-snowflake` with dlt's dbt wrapper. 

**Tasks**
1. * [ ] add dbt snowflake to `dbt` extras
2. * [ ] add snowflake profile (https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup) to `profiles.yml` in dlt/helper/dbt. try to add a single profile that can work with password or key auth

**Tests**
1. * [ ]  make sure that pipeline tests in `test_dbt_helper.py` run for snowflake
2. * [ ] we need to extend tests in `test_runner_dbt_versions.py`. currently the fixture tries all dbt core versions and runs jaffle shop on postgres. what we need to do is to create `test_runner_dbt_destinations` (which will take ie. dbt version 1.4.0) and run jaffle shop on all supported destinations. we may install all destinations in a single venv.

**Implementation Details**
dbt runner will try to create a dbt wrapper using destination name as dbt profile name. it then emits the content of destination's credentials as env variable and calls dbt commands in separate process
```
@with_custom_environ
    def _run_dbt_command(self, command: str, command_args: Sequence[str] = None, package_vars: StrAny = None) -> Sequence[DBTNodeResult]:
```

steinitzu:
closes #434


rudolfix (approved):
this is really good! thanks!

Issue #320, opened 2023-05-09
add `load_id` to `_dlt_versions` table


rudolfix:
**Background**
We want the users to be able to link the loads in `_dlt_loads` to `_dlt_versions` to identify which load updated the schema without enabling the trace (https://dlthub.com/docs/running-in-production/running#inspect-and-save-the-load-info-and-trace).

**Tasks**
1. * [x] insert `load_id` into the `_dlt_versions`, see `SqlJobClientBase`
2. * [x] you need to change the `_dlt_versions` schema definition (`schema/utils.py`) - make sure it is NULLABLE so old dlt installations can migrate
3. * [x] we need to increase schema engine version and write migration that will add the column to existing schema (`schema/utils.py`)

**Tests**
1. Extend existing tests for updating schema and make sure the `load_id` is stored (`test_job_client.py`)
2. Test the schema migration: mock the schema to not have this column, load data, then load the schema again and load the data again to make sure the schema was added

codingcyclist:
Hi @rudolfix thanks for opening this ticket 🙌 I recently ran into another use case wrt. schema version names: I wanted to identify which schema version was used for which load (i.e. the reverse of what you're describing above). Have you also considered adding the schema version to the `_dlt_loads` table?

rudolfix:
why not? we can do it in this ticket. there will be duplicated information (you can always join the tables) but I assume it is more user friendly
@codingcyclist btw. could you check if maybe this https://dlthub.com/docs/running-in-production/running#inspect-and-save-the-load-info-and-trace
solves your problem? `dlt` produces very rich trace of what got loaded including all the information that you need. you can load this trace into your destination.

rudolfix:
@tarunsamanta2k20 there was github outage and several copies of this ticket were created. glad to hear that you want to contribute! what kind of help do you need?

codingcyclist:
@rudolfix I just filed a small [PR](https://github.com/dlt-hub/dlt/pull/466) for this, let me know what you think. Instead of adding the `load_id` to the version table, I decided to add the version number to the loads table. The reason was that the `_update_schema_in_storage()` function, which updates the version table, is not aware of the load id, whereas the `complete_load()` function, which updates the load table, is aware of the current schema version.

rudolfix:
@codingcyclist thanks for the PR. I'll come back with the review soon

PR #466, opened 2023-07-01
320: add version to _dlt_loads table

codingcyclist:
Resolves [320](https://github.com/dlt-hub/dlt/issues/320)

### DIFF
--- a/dlt/common/schema/typing.py
+++ b/dlt/common/schema/typing.py
@@ -4,7 +4,7 @@ from dlt.common.data_types import TDataType
 from dlt.common.normalizers.typing import TNormalizersConfig
 
 # current version of schema engine
-SCHEMA_ENGINE_VERSION = 5
+SCHEMA_ENGINE_VERSION = 6
 
 # dlt tables
 VERSION_TABLE_NAME = "_dlt_version"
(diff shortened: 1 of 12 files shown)


rudolfix (changes requested):
code changes are solid and all is ok there! thanks!
we still need to improve some tests. I enabled CI for your fork and a few tests are not passing
1. small linter issue (if you have poetry and make you can just do `make dev && make lint`)
2. common test problems: https://github.com/dlt-hub/dlt/actions/runs/5430315589/jobs/9969881019?pr=466

why the common tests are failing: each schema has a content hash used to detect schema changes. increasing the version changed the hash (when I load fixtures in tests, your migration modifies the content and we get a different hash - then tests that check hashes fail)
to fix: we need a v6 version of this fixture: ethereum_schema_v5.yml. I generate it with a dlt command:
```
dlt schema tests/common/cases/schemas/eth/ethereum_schema_v5.yml --remove-defaults
```
and save the output to new name. from it you can get the new content hash and put it into IMPORTED_VERSION_HASH_ETH_V5 constant (used by the tests)

btw. we'll upgrade our CI to be more friendly for PRs from forks: https://github.com/dlt-hub/dlt/issues/476

if the above looks too complicated or unclear just ping us, someone will take over the final fixes

codingcyclist:
Hi @rudolfix Thanks a lot for your comments :) I took a stab at them and things look good locally. However, the nefly CI seems to be blocking the test from running on my fork. Is there anything I can do about that 🤔 ?

rudolfix:
@codingcyclist I just unblocked CI. again... maybe it blocks PRs after some inactivity. will check out

rudolfix (approved):
LGTM!

PR #469, opened 2023-07-03
271 replace with staging tables and fix clearing bug

sh-rp:
This PR is based on the work for loading bucket files, the pr is against development so the tests will run.

This PR uses the staging tables for doing the replace write disposition. This will make sure no data is lost in the event of multiple jobs in one LoadPackage. If a resource only ever yields None, the associated table and the child tables are also cleared.

Implementation notes:
* For achieving the clearing, I am creating empty jobs for the main resources that get forwarded through the normaliser to the load stage. 
* I changed the logic for updating tables when there is a disposition that requires staging tables, it depends on the existing tables in the db now, I quite like it but you'll have to judge
* I moved the logic for deciding which tables to use when into the job client, so merge and replace are not mentioned in the loader anymore, I would say that is nice too.

Question: One test seems to suggest that write dispositions are rather read from a previous database state than the current code config. I need to discuss this with @rudolfix.


rudolfix (changes requested):
right now it is too hard to review properly
1. let's merge staging and snowflake staging first
2. it is hard to say what we need to do staging via replace and what we need to fix the replace bug
3. that we need to get list of tables from database to work properly is worrying. we should be able to just use schema. it worked for merging somehow? (we may need to create tables for which we have jobs but also all child tables of tables for which we have jobs. then we know that all expected tables are there)
4. not all file types can be empty. how do you deal with empty INSERT VALUE files?
5. it seems that you do not handle the following case: we have a replace parent table which has an element. but that element does not have any child elements. so we do not truncate this child table and it keeps the previous values
6. using staging for replace should be optional. does the old way still work?

btw. there are several good ideas here. but I'd really try to fix the replace bug in a separate PR. it is really tricky

PR #472, opened 2023-07-05
snowflake staging with gcs and s3

sh-rp:
Fairly simple PR for snowflake staging. I could not get the aws s3 integration to work. I completed the setup process 3 times but I still get access denied..


dlt/destinations/snowflake/snowflake.py
```diff
@@ -48,44 +53,82 @@
...
+            stage_name: Optional[str] = None, storage_integration: Optional[str] = None, keep_staged_files: bool = True
     ) -> None:
         file_name = FileStorage.get_file_name_from_file_path(file_path)
         super().__init__(file_name)
 
-        with client.with_staging_dataset(write_disposition == "merge"):
+        with client.with_staging_dataset(write_disposition in ["merge", "replace"]):
             qualified_table_name = client.make_qualified_table_name(table_name)
 
-            if stage_name:
```
rudolfix, line 62: let's change how `stage_name` works:
1. if stage name is specified we expect that this stage is created and contains all expected integrations and credentials so we just pass it in copy command. we also need to use a pattern that will load only this specific file
2. if stage name is not provided then:
- for local files that use PUT we use table stage (like we do now)
- for bigquery we just simply require stage name. if someone can create integration then s/he can create stage as well
- for s3/azure (we'll add soon) we can pass credentials in copy command as you do, just see comment below

dlt/destinations/snowflake/snowflake.py
```diff
@@ -48,44 +53,82 @@
...
+            if bucket_path and storage_integration:
+                stage_name = client.make_qualified_table_name(stage_name) if stage_name else qualified_table_name
+                bucket = CopyFileLoadJob.get_bucket(file_path)
+                client.execute_sql(f"""
+                    CREATE STAGE IF NOT EXISTS {stage_name}
+                    URL = '{bucket}'
+                    STORAGE_INTEGRATION = {storage_integration};""")
+                from_clause = f"FROM @{stage_name}"
+                # select all files that match the file from our job
+                pattern_clause = f"PATTERN = '.*{file_name}'"
```
rudolfix, line 90: actually, we do we need pattern? we can just give a file name in FILES part of the COPY command

dlt/destinations/snowflake/snowflake.py
```diff
@@ -48,44 +53,82 @@
...
+            # s3 credentials case
+            elif bucket_path:
+                if bucket_path.startswith("s3://"):
+                    s3_config = _s3_config()
+                    credentials_clause = f"""CREDENTIALS=(AWS_KEY_ID='{s3_config.aws_access_key_id}' AWS_SECRET_KEY='{s3_config.aws_secret_access_key}')"""
```
rudolfix, line 95:
```
IAM user: IAM credentials are required. Temporary (aka “scoped”) credentials are generated by AWS Security Token Service (STS) and consist of three components:

AWS_KEY_ID

AWS_SECRET_KEY

AWS_TOKEN

All three are required to access a private/protected bucket. After a designated period of time, temporary credentials expire and can no longer be used. You must then generate a new set of valid temporary credentials.
```
you are not passing AWS_TOKEN. probably you need to get it from boto3. ask GPT-4 :)

rudolfix: maybe add it to `AwsCredentials` there's already a `aws_session_token` which we can populate from boto3

dlt/destinations/snowflake/snowflake.py
rudolfix: if the above works we do not need to create any stages and we do not need `storage_integration`

rudolfix (changes requested):
my take:
1. lets move it to #451  (staging PR)
2. see the review
3. we need proper tests for cases where stage name is specified and not
4. we need docs where we explain how to setup a stage for PUT, s3 and gcs and how to use without named stage

dlt/destinations/snowflake/snowflake.py
sh-rp: when does this happen? my tests pass here..

rudolfix:
@sh-rp should we close this?

PR #475, opened 2023-07-06
dbt 1.5 support

### LINKED ISSUE #458: enable dbt 1.5.0
**Background**
dbt helper is tightly integrated with `dbt`. this happens in `dbt_utils.py`. after upgrade of dbt-core to 1.5.0 the interface to CLI commands chaged. it now uses `click` and tbh. looks like it will be harder to invoke commands with arguments and get status and results (no single dispatch method)

**Tasks**
* [x] enable support for 1.5.0
* [x] add 1.5.0 to tested versions
* [ ] bump default dependencies to 1.5.x and update DEFAULT_DBT_VERSION
* [x] all tests need to pass
* [x] keep the current architecture.
* [ ] motherduck destination requires 1.5.x `dbt-duckdb` test it locally but not enable on CI

sh-rp:
closes #458


rudolfix (changes requested):
this looks really good!
pls look at the ticket: https://github.com/dlt-hub/dlt/issues/458
- we bump default dbt version
- please test if Motherduck works with dbt (you can mock tests to use motherduck instead of duck db and see if they work locally). then update motherduck destination docs

rudolfix (approved):
@sh-rp let's merge this because there's another PR growing (#478 ) that messes up with dbt runner a little

PR #479, opened 2023-07-08
Add a Weaviate destination

### LINKED ISSUE #532: Implement Weaviate destination in dlt
## Objective

Implement Weaviate vector database as a destination in the dlt library.

### Features:

1. Default Weaviate configuration via dlt resources adapter.
2. Ability to define an explicit schema for Weaviate, derived from the dlt's schema, facilitating proper data mapping and structure.
3. Vectorizer configuration:
- Provide an option to specify the vectorizer.
- By default, no data fields should be vectorized.
- Allow explicit specification of fields to be vectorized to enhance search and filtering capabilities.
4. Provide an option to configure tokenization for individual data fields following the Weaviate's [property tokenization guidelines](https://weaviate.io/developers/weaviate/config-refs/schema#property-tokenization).

#### Development Tasks:

- [x] Define the interface and configuration options
- [ ] Implement the core functionality
- [ ] Write unit and integration tests
- [ ] Update documentation and examples

#### Requested Community Feedback
Any feedback on the planned features or additional suggestions would be highly appreciated.

burnash:
This PR introduces the Weaviate database destination in the dlt Python library. 
Implements and closes #532


dlt/destinations/weaviate/configuration.py
```diff
@@ -0,0 +1,29 @@
...
+    api_key: str
+    additional_headers: Optional[Dict[str, str]] = None
+
+
+@configspec(init=True)
+class WeaviateClientConfiguration(DestinationClientConfiguration):
+    destination_name: str = "weaviate"
+    loader_file_format: TLoaderFileFormat = "jsonl"
+
+    fail_schema_update: bool = False
```
rudolfix, line 20: those were just used by `dummy` for testing

dlt/destinations/weaviate/weaviate.py
```diff
@@ -0,0 +1,279 @@
...
+DLT_TABLE_PREFIX = "_dlt"
+
+
+# TODO: move to common
+def is_dlt_table(table_name: str) -> bool:
+    return table_name.startswith(DLT_TABLE_PREFIX)
+
+
+def snake_to_camel(snake_str: str) -> str:
+    return "".join(x.capitalize() for x in snake_str.split("_"))
```
rudolfix, line 35: interesting. we could easily create a naming convention so you have camel case in schema (in normalize stage)

dlt/destinations/weaviate/weaviate.py
```diff
@@ -0,0 +1,279 @@
...
+            record = response["data"]["Get"][version_class_name][0]
+        except IndexError:
+            return None
+        return self._decode_schema(record)
+
+    def _decode_schema(self, record: dict) -> StorageSchemaInfo:
+        # XXX: Duplicate code from dlt/destinations/job_client_impl.py
+        schema_str = record["schema"]
+        try:
+            schema_bytes = base64.b64decode(schema_str, validate=True)
```
rudolfix, line 175: we probably could store uncompressed schemas

pyproject.toml
```diff
@@ -110,7 +110,8 @@ filesystem = ["s3fs", "boto3"]
 s3 = ["s3fs", "boto3"]
 gs = ["gcsfs"]
 snowflake = ["snowflake-connector-python"]
-motherduck= ["duckdb", "pyarrow"]
+motherduck = ["duckdb", "pyarrow"]
+weaviate = ["weaviate-client"]
```
rudolfix, line 114: I think the dependency itself is not added

dlt/destinations/weaviate/weaviate.py
```diff
@@ -0,0 +1,279 @@
...
+            return self._make_non_vectorized_class_schema(class_name)
+
+        return self._make_vectorized_class_schema(class_name)
+
+    def _make_vectorized_class_schema(self, class_name: str) -> dict:
+        return {
+            "class": class_name,
+            "vectorizer": "text2vec-openai",
+            "moduleConfig": {
+                "text2vec-openai": {
```
rudolfix, line 205: we should expose this as configuration

dlt/destinations/weaviate/weaviate.py
```diff
@@ -0,0 +1,279 @@
...
+    ) -> None:
+        file_name = FileStorage.get_file_name_from_file_path(local_path)
+        super().__init__(file_name)
+
+        class_name = table_name_to_class_name(table_name)
+
+        with db_client.batch(
+            batch_size=client_config.weaviate_batch_size,
+        ) as batch:
+            with FileStorage.open_zipsafe_ro(local_path) as f:
```
rudolfix, line 69: would it help if we register the types with the class in weaviate?

rudolfix (changes requested):
some of my questions may be just lack of knowledge:
1. shukri thinks that chunking is the most interesting things. should we add it? and where? to the loader or maybe as transform to extract?
2. any additional schema hints? ie. to mark content that should be vectorized, to set chunking style per "column" etc. any additional options in decorators? maybe new decorator even :)
3. maybe a custom normalizer? so we do not have any parent-child tables and work on single nested documents?
4. should we support langchain Document when yielding in dlt resources?

dlt/destinations/weaviate/weaviate.py
burnash: sure, it's still for WIP stage

hsm207:
@burnash it looks like this implementation relies on weaviate's autoschema feature.

for example, if the datatype is text, user can specify the [totkenization](https://weaviate.io/developers/weaviate/config-refs/schema#property-tokenization) method which will influence the search results. It does not look like it is possible to do this in the current implementation.

burnash:
@hsm207 great point! This initial implementation indeed leans on autoschema feature. We're currently figuring out how to best integrate additional functionality in dlt's API and use our own type hints.

dlt/destinations/weaviate/weaviate.py
```diff
@@ -0,0 +1,285 @@
...
+        client_config: WeaviateClientConfiguration,
+        load_id: str,
+    ) -> None:
+        file_name = FileStorage.get_file_name_from_file_path(local_path)
+        super().__init__(file_name)
+
+        class_name = table_name_to_class_name(table_name)
+
+        with db_client.batch(
+            batch_size=client_config.weaviate_batch_size,
```
rudolfix, line 67: I really hope that batch is all or nothing operation. otherwise we cannot guarantee that we do not duplicate something on retry... needs research

hsm207:
batch is not an all or nothing operation. Weaviate has no notion of
transactions
<https://weaviate.io/developers/weaviate/more-resources/faq#q-what-is-weaviates-consistency-model-in-a-distributed-setup>
.

If there is a duplicate, weaviate will output an error message. However,
the duplicate is done by uuid, which the user can specify when upload their
document. e.g. see this code snippet in this section
<https://weaviate.io/developers/weaviate/client-libraries/python#auto-batching>

Mohd Shukri Hasan

Community Solution Engineer

<https://weaviate.io/>

<https://www.linkedin.com/in/mohdshukrihasan>

<https://twitter.com/MohdShukriHasa1>

dlt/helpers/weaviate_helper.py
```diff
@@ -0,0 +1,53 @@
...
+                if prop in properties:
+                    properties[prop]["tokenization"] = method
+                else:
+                    properties[prop] = {
+                        "name": prop,
+                        "tokenization": method,
+                    }
+
+        column_hints = _weaviate_properties_to_hints(properties.values())
+        resource.apply_hints(columns=column_hints)
```
rudolfix, line 47: `apply_hints` is overriding the columns, in your case we must merge with existing columns: or add a new method to resource or merge in the helper. there's update_dict_nested in utils that merges two dicts recursively

dlt/helpers/weaviate_helper.py
```diff
@@ -0,0 +1,53 @@
...
+    else:
+        raise ValueError(
+            "Either 'class_schema' or 'vectorize' and/or 'tokenization' must be specified."
```
rudolfix, line 50: 👍

rudolfix (changes requested):

rudolfix (changes requested):
I have a list of things that still need work. Let's go through it and decide what to implement now and what to kick out.

dlt/destinations/weaviate/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+    additional_headers: Optional[Dict[str, str]] = None
+
+
+@configspec(init=True)
+class WeaviateClientConfiguration(DestinationClientConfiguration):
+    destination_name: str = "weaviate"
+    weaviate_batch_size: int = 100
+
+    credentials: WeaviateCredentials
+    vectorizer: str = "text2vec-openai"
```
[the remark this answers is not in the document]
burnash: Could you elaborate on why we need tokenizer setting at config level? If I'm not mistaken, its property setting, and there's a default already from Weaviate itself.

dlt/destinations/weaviate/weaviate.py
```diff
@@ -0,0 +1,370 @@
...
+        )
+
+    def make_weaviate_class_schema(self, table: TTableSchema) -> Dict[str, Any]:
+        """Creates a Weaviate class schema from a table schema."""
+        table_name = table["name"]
+
+        class_name = table_name_to_class_name(table_name)
+
+        if is_dlt_table(table_name):
+            return self._make_non_vectorized_class_schema(class_name, table)
```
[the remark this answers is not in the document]
burnash: Good point, the difference is that for `_make_non_vectorized_class_schema()` I skip vectorization on the class level versus property level in `_make_vectorized_class_schema()` case:

```
return {
    "class": class_name,
    "properties": properties,
    "vectorizer": "none",
    "vectorIndexConfig": {
        "skip": True,
    },
}
```

@hsm207 is there any difference in terms of internal implementation?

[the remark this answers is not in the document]
hsm207: @burnash if in the a class, the user sets:

```
"vectorIndexConfig": {
        "skip": True,
    }
```

then, it is recommended to also set `"vectorizer"` to `"none"` in the class too. Otherwise, the vectorizer will still run to create the embeddings but will not index it.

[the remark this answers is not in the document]
burnash: @hsm207 good point, that's what I'm currently doing for the classes that skip vectorization entirely (see the `return` above). My question is what's the difference between skipping vectorization for every property versus skipping it on the class level (both with `"vectorizer": "none"` and `"vectorIndexConfig": {"skip": True}`)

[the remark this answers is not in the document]
hsm207: @burnash skipping vectorization for every property is not equivalent to `"vectorizer": "none"` and `"vectorIndexConfig": {"skip": True}`. In the former, the text2vec module would still be executed to check whether to create an embedding for the object. In the latter, this check is skipped entirely.

rudolfix:
a recap of what to implement and what goes later:
**now**
@burnash 
* [ ]  weaviate adapter: does not implement hints as str, list of str or mapping column:vectorizer/tokenizer
* [ ]  must implement replace disposition (truncate tables in initialize storage). must implement append - by ignoring primary keys and using only uniq and using autogen uuid
* [ ]  we must support schema evolution by adding new fields (https://weaviate.io/developers/weaviate/api/rest/schema#method-and-url-5)
* [ ]  must implement `initialize_storage` and `is_initialize_storage` by creating sentinel class which is dataset name. must prefix all class names with dataset name.


@rudolfix 
* [ ]  we must recognize terminal and non terminal exceptions and wrap weaviate clients into something that does that. the same client should use tenacity to retry batch loading - will be trivial if we know the exception types
* [ ]  there's new JobClientBase and you can remove the merge but you must implement a few other base methods
* [ ]  create a naming convention for weaviate that will use camel case for classes and pass other identifiers intact (sans /n /r and some other control characters)


+ tests and docs

**Immediately after this PR**
- implement chunking
- add multitenancy to weaviate_helper
- implement sync with destination for weaviate
- allow the user to choose the nesting of the tables (0 prevents nesting)


**Later**
- default inverted index config in adapter and configuration
- default vector index config in adapter and configuration
- take weaviate client in credentials
- create a weaviate normalizer that will create weaviate style references for child tables (not sql style references)
- setting the vector indexes

dlt/helpers/weaviate_helper.py
```diff
@@ -0,0 +1,40 @@
...
+            for prop in vectorize:
+                column_hints[prop] = {
+                    "name": prop,
+                    "x-vectorize": True,
+                }
+        if tokenization:
+            for prop, method in tokenization.items():
+                if prop in column_hints:
+                    column_hints[prop]["x-tokenization"] = method
+                else:
```
hsm207, line 30: there are only a limited number of [possible values](https://weaviate.io/developers/weaviate/config-refs/schema#property-tokenization) for method: 

* word
* lowercase
* whitespace
* field

would it be possible for dlt to enforce this check in case user made a typo?

dlt/destinations/weaviate/configuration.py
```diff
@@ -0,0 +1,21 @@
...
+
+from dlt.common.configuration import configspec
+from dlt.common.configuration.specs.base_configuration import CredentialsConfiguration
+from dlt.common.destination import TLoaderFileFormat
+from dlt.common.destination.reference import DestinationClientConfiguration
+
+
+@configspec
+class WeaviateCredentials(CredentialsConfiguration):
+    url: str
```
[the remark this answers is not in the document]
hsm207: This will depend on the user. They can configure the weaviate container to disable public access and require authorization.

dlt/destinations/weaviate/configuration.py
```diff
@@ -0,0 +1,30 @@
...
+    api_key: str
+    additional_headers: Optional[Dict[str, str]] = None
+
+
+@configspec(init=True)
+class WeaviateClientConfiguration(DestinationClientConfiguration):
+    destination_name: str = "weaviate"
```
hsm207, line 19: how does this param affect the weaviate client config?

dlt/helpers/weaviate_helper.py
rudolfix: yes, we can add it to our schema definition

burnash: One alternative is to leave the validation to Weaviate itself. That way there's not need to sync validation code should the list of supported tokenization methods change. Weaviate's error message is very well written:

```
<class 'weaviate.exceptions.UnexpectedStatusCodeException'>
Create class! Unexpected status code: 422, with response body: 
{'code': 606, 'message': 'properties.0.tokenization in body should be one of [word lowercase whitespace field]'}.
```

hsm207: @burnash that's what I was worried about. Would an experiend dlt user know how to interpret which column properties.0 is referring to? Otherwise, if we follow this suggestion, then we should document how the mapping occurs so dlt user can figure it out themselves.

burnash: @hsm207 good point, let's handle the validation here

dlt/destinations/weaviate/weaviate_client.py
```diff
@@ -0,0 +1,540 @@
...
+            auth_client_secret=weaviate.AuthApiKey(api_key=config.credentials.api_key),
+            additional_headers=config.credentials.additional_headers,
+        )
+
+    def make_full_name(self, table_name: str) -> str:
+        """Make a full Weaviate class name from a table name by prepending
+        the dataset name if it exists.
+        """
+        return (
+            f"{self.config.dataset_name}_{table_name}"
```
rudolfix, line 250: here's the new thing:
use self.config.normalize_dataset_name(schema)
to get the right dataset name.

also I'm not sure we should use "_" as separator. I'd say not - naming convention does not allow for `_` in table names

dlt/destinations/weaviate/weaviate_client.py
```diff
@@ -0,0 +1,540 @@
...
+        except weaviate.exceptions.UnexpectedStatusCodeException as e:
+            if e.status_code == 404:
+                return False
+            raise
+        return True
+
+    def create_sentinel_class(self) -> None:
+        self.db_client.schema.create_class(
+            {
+                "class": self.config.dataset_name,
```
rudolfix, line 316: here's the same situation. my take: just like in the sql_client, add a method to get `dataset_name` that will normalize it etc.

rudolfix (changes requested):

rudolfix (changes requested):
also please add method `drop_dataset` that will drop all classes from dataset including the sentinel. empty dataset drops all classes.

please use it in tests

dlt/destinations/weaviate/weaviate_client.py
burnash: Thanks! That could be the source of the error in the tests. I'd love to use some kind of separator as it makes it easier to read the names if the prefixed classes

PR #481, opened 2023-07-09
Fix: use sets to prevent unecessary truncate calls

z3z1ma:
~~FYI, there seems to be an unnecessary extra call to initialize storage.~~

EDIT: I suppose its because the 2nd call does the truncate, but I think its an extra api call to check if dataset exists.


rudolfix (approved):
looks good! thx for the fix

PR #488, opened 2023-07-13
replace with staging tables

sh-rp:


rudolfix:
@sh-rp huh! IMO you are better off cherry-picking the replace changes than trying to merge this. unfortunately that was expected :/

dlt/destinations/sql_jobs.py
```diff
@@ -12,36 +12,71 @@
...
+
+
+class SqlStagingCopyJob(SqlBaseJob):
+    """Generates a list of sql statements that copy the data from staging dataset into destination dataset."""
+    failed_text: str = "Tried to generate a staging copy sql job for the following tables:"
+
+    @classmethod
+    def generate_sql(cls, table_chain: Sequence[TTableSchema], sql_client: SqlClientBase[Any]) -> List[str]:
+        sql: List[str] = []
+        for table in table_chain:
```
rudolfix, line 53: could we do this:
1. redshift, postgres and duckdb have transactional DDL so we could do the following
- DROP destination table
- ALTER TABLE SCHEMA staging -> destination
- CREATE TABLE staging LIKE destination (we must recreate the staging table after move)
2. bigquery
this is really really cool
https://cloud.google.com/bigquery/docs/table-clones-intro

basically
1. DROP destination TABLE if not exist
2. CLONE staging -> destination

bigquery does not have ddl transactions BUT the operation above is idempotent so `dlt` will retry it until it succeeds. IMO good enough

for snowflake we cannot do much -> no DDL transactions. we could rename or swap tables but if we retry the operation will fail. so we should stay with copy

we also need to upgrade `SqlLoadJob` so it detects if sql script contains DDL statements and does not begin transaction if DDL statements are not supported (ie. BigQuery and Motherduck case)

this is some work but IMO it is worthy esp. in case of Redshift and BigQuery we get a huge improvement in cost

tests/load/pipeline/test_replace_disposition.py
```diff
@@ -0,0 +1,165 @@
...
+@pytest.mark.parametrize(STAGING_COMBINATION_FIELDS, STAGING_AND_NON_STAGING_COMBINATIONS)
+def test_replace_disposition(destination: str, staging: str, file_format: str, bucket: str, settings: Dict[str, Any]) -> None:
+
+    # only allow 40 items per file
+    os.environ['DATA_WRITER__FILE_MAX_ITEMS'] = "40"
+    # use staging tables for replace
+    os.environ['DESTINATION__REPLACE_STRATEGY'] = "staging"
+
+    # snowflake requires gcs prefix instead of gs in bucket path
+    if destination == "snowflake":
```
rudolfix, line 18: maybe we should normalize paths in copy jobs in snowflake? it has a specialized one

rudolfix (changes requested):
this looks really good.
1. please take a look at `test_write_dispositions` and parametrize for strategies. the test should OFC pass
2. add new document to docs in general-usage: full-loading.md where we'll describe the mechanism, the strategies etc. 
3. if you implement those optimizations in warehouses, please mention those per warehouse docs
4. incremental-loading.md introduces full loading: refer the document (2) from it

dlt/load/load.py
```diff
@@ -268,15 +269,15 @@ def load_single_package(self, load_id: str, schema: Schema) -> None:
...
+                if staging_table_jobs:
                     logger.info(f"Client for {job_client.config.destination_name} will start initialize STAGING storage")
                     job_client.initialize_storage(staging=True)
                     logger.info(f"Client for {job_client.config.destination_name} will UPDATE STAGING SCHEMA to package schema")
-                    merge_tables = set(job.table_name for job in merge_jobs)
-                    job_client.update_storage_schema(staging=True, only_tables=merge_tables | {VERSION_TABLE_NAME}, expected_update=expected_update)
-                    logger.info(f"Client for {job_client.config.destination_name} will TRUNCATE STAGING TABLES: {merge_tables}")
-                    job_client.initialize_storage(staging=True, truncate_tables=merge_tables)
+                    staging_tables = set(job.table_name for job in staging_table_jobs)
+                    job_client.update_storage_schema(staging=True, only_tables=staging_tables | dlt_tables, expected_update=expected_update)
```
rudolfix, line 278: you solved the conflict incorrectly: only_tables=merge_tables | {VERSION_TABLE_NAME}

dlt/load/load.py
```diff
@@ -268,15 +269,15 @@ def load_single_package(self, load_id: str, schema: Schema) -> None:
                 # only update tables that are present in the load package
                 applied_update = job_client.update_storage_schema(only_tables=all_tables | dlt_tables, expected_update=expected_update)
                 # update the staging dataset
-                merge_jobs = self.get_new_jobs_info(load_id, schema, "merge")
-                if merge_jobs:
+                staging_table_jobs = self.get_new_jobs_info(load_id, schema, job_client.get_stage_dispositions())
+                if staging_table_jobs:
                     logger.info(f"Client for {job_client.config.destination_name} will start initialize STAGING storage")
                     job_client.initialize_storage(staging=True)
```
rudolfix, line 275: MAYBE we could remove the staging flag from `initialize_storage` and `update_storage_schema` and have `with_staging` context manager on SqlJobClientBase (the refactor that I proposed previously?). that would remove all traces of staging from JobClientBase (the top level ABC)

rudolfix (changes requested):

dlt/common/destination/reference.py
```diff
@@ -172,6 +176,20 @@ def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) ->
...
+    def get_stage_dispositions(self) -> List[TWriteDisposition]:
+        """Returns a list of dispositions that require staging tables to be populated"""
+        return []
+
+    def truncate_destination_table(self, disposition: TWriteDisposition) -> bool:
+        """Returns wether the destination table should be truncated before loading"""
+        return False
+
+    @abstractmethod
+    def create_table_chain_completed_followup_jobs(self, table_chain: Sequence[TTableSchema]) -> List[NewLoadJob]:
```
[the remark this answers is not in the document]
sh-rp: Well either we have _create_merge_job which can be overwritten by redshift and bigquery clients or we modify create_table_chain_completed_followup_jobs for those classes..

dlt/load/load.py
sh-rp: I have created a subclass that supports the staging interface, it could also be a mixin actually, not sure.

sh-rp: Also there is one method left called drop_tables, which has a staging parameter, maybe this should also be handled within the context manager?

sh-rp: Ok, I removed it there too..

dlt/destinations/sql_jobs.py
sh-rp: And the alter table append command in redshift can not run in a tx and can also not be run in the same string of statements as other append commands, so I don't think it makes sense to use it really.

tests/load/pipeline/test_replace_disposition.py
```diff
@@ -0,0 +1,170 @@
+import dlt, os, pytest
+from typing import Any, Dict
+
+from tests.pipeline.utils import  assert_load_info
+from tests.load.pipeline.utils import  load_table_counts
+from tests.utils import ALL_DESTINATIONS
+from tests.load.pipeline.utils import destinations_configs, DestinationTestConfiguration, set_destination_config_envs
+
+REPLACE_STRATEGIES = ["truncate-and-insert", "insert-from-staging", "optimized"]
```
rudolfix, line 9: I'd rename `optimized` -> `staging-optimized`

rudolfix (approved):

Issue #492, opened 2023-07-14
Clarify create-a-pipeline docs


willi-mueller:
## Observed

Following copy & pasting from the current doc page https://dlthub.com/docs/walkthroughs/create-a-pipeline 
creates the following file tree:
```
.
├── poetry.lock
├── pyproject.toml
├── weatherapi.py
└── weatherapi_duckdb
```

## Expected
However, I wonder if the convention is actually to have the `weatherapi.py` file inside the `weatherapi_duckdb` folder and thus the directory tree should look like as follows:

```
.
├── poetry.lock
├── pyproject.toml
└── weatherapi_duckdb
    ├── requirements.txt
    ├── weatherapi.duckdb
    └── weatherapi.py
```

Please let me know if I misunderstood your intention or if this proposed change could produce more clarity.

PR #493, opened 2023-07-14
Clarifies create-a-pipeline docs

willi-mueller:
This PR addresses https://github.com/dlt-hub/dlt/issues/492

## Alternatives

Instead of asking the user to `cd my_new_directory` we could put:

```bash
mkdir weatherapi_duckdb && cd "$_"
```

However, because it's less known and a bit cryptic I decided to be explicit with a simple `cd`.


adrianbr (approved):

willi-mueller:
@adrianbr Maybe the following is even better?

AstrakhantsevaAA (dismissed):

AstrakhantsevaAA (approved):

PR #499, opened 2023-07-17
replace child table handling

sh-rp:
One important note: This PR only handles propers child table clearing for the staging replace strategy. There are more changes needed if we also want to clear child tables on the classic strategy.

The one test that is always failing is: tests/load/pipeline/test_merge_disposition.py::test_merge_on_keys_in_schema. The problem here is, that the test makes sure that if there are tables defined in the schema but no data is coming for them they don't get created. But in this PR all the child tables of any table that has data are created on staging and cleared, so that the child table replacing works. The way to solve this is to look at the destination and see which tables are present already and prevent creation of tables that are not present, but that part I took out again.

### DIFF
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -181,7 +181,7 @@ class JobClientBase(ABC):
         pass
 
     def get_truncate_destination_table_dispositions(self) -> List[TWriteDisposition]:
-        return []
+        return ["replace"]
 
     @abstractmethod
     def create_table_chain_completed_followup_jobs(self, table_chain: Sequence[TTableSchema]) -> List[NewLoadJob]:
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -180,6 +180,9 @@ class JobClientBase(ABC):
     def restore_file_load(self, file_path: str) -> LoadJob:
         pass
 
+    def get_truncate_destination_table_dispositions(self) -> List[TWriteDisposition]:
+        return []
+
     @abstractmethod
     def create_table_chain_completed_followup_jobs(self, table_chain: Sequence[TTableSchema]) -> List[NewLoadJob]:
         """Creates a list of followup jobs that should be executed after a table chain is completed"""
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -181,6 +181,7 @@ class JobClientBase(ABC):
         pass
 
     def get_truncate_destination_table_dispositions(self) -> List[TWriteDisposition]:
+        # in the base job, all replace strategies are treated the same, see filesystem for example
         return ["replace"]
 
     @abstractmethod
--- a/dlt/destinations/filesystem/filesystem.py
+++ b/dlt/destinations/filesystem/filesystem.py
@@ -88,7 +88,9 @@ class FilesystemClient(JobClientBase):
             for table in truncate_tables:
                 search_prefix = posixpath.join(self.dataset_path, f"{self.schema.name}.{table}.")
                 for item in all_files:
+                    # NOTE: glob implementation in fsspec does not look thread safe, way better is to use ls and then filter
                     if item.startswith(search_prefix):
+                        # NOTE: deleting in chunks on s3 does not raise on access denied, file non existing and probably other errors
                         self.fs_client.rm_file(item)
 
         # create destination dir
--- a/dlt/destinations/filesystem/filesystem.py
+++ b/dlt/destinations/filesystem/filesystem.py
@@ -38,31 +38,7 @@ class LoadFilesystemJob(LoadJob):
 
         super().__init__(file_name)
         fs_client, _ = client_from_config(config)
-
-        # fallback to replace for merge without any merge keys
-        if write_disposition == 'merge':
-            write_disposition = 'append' if has_merge_keys else 'replace'
-
-        # replace existing files. also check if dir exists for bucket storages that cannot create dirs
-        if write_disposition == 'replace' and fs_client.isdir(dataset_path):
-            job_info = LoadStorage.parse_job_file_name(file_name)
-            # remove those files
-            search_prefix = posixpath.join(dataset_path, f"{schema_name}.{job_info.table_name}.")
-            # but leave actual load id - files may be loaded from other threads
-            ignore_prefix = posixpath.join(dataset_path, f"{schema_name}.{job_info.table_name}.{load_id}.")
-            # NOTE: glob implementation in fsspec does not look thread safe, way better is to use ls and then filter
-            all_files: List[str] = fs_client.ls(dataset_path, detail=False, refresh=True)
-            items = [item for item in all_files if item.startswith(search_prefix) and not item.startswith(ignore_prefix)]
-            # NOTE: deleting in chunks on s3 does not raise on access denied, file non existing and probably other errors
-            # if items:
-            #     fs_client.rm(items[0])
-            for item in items:
-                # ignore file not found as we can have races from other deleting threads
-                try:
-                    fs_client.rm_file(item)
-                except FileNotFoundError:
-                    pass
-
+        self.destination_file_name = LoadFilesystemJob.make_destination_filename(file_name, schema_name, load_id)
         fs_client.put_file(local_path, self.make_remote_path())
 
     @staticmethod
@@ -106,6 +82,16 @@ class FilesystemClient(JobClientBase):
         return posixpath.join(self.fs_path, self.config.dataset_name)
 
     def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
+        # clean up existing files for tables selected for truncating
+        if truncate_tables and self.fs_client.isdir(self.dataset_path):
+            all_files = self.fs_client.ls(self.dataset_path, detail=False, refresh=True)
+            for table in truncate_tables:
+                search_prefix = posixpath.join(self.dataset_path, f"{self.schema.name}.{table}.")
+                for item in all_files:
+                    if item.startswith(search_prefix):
+                        self.fs_client.rm_file(item)
+
+        # create destination dir
         self.fs_client.makedirs(self.dataset_path, exist_ok=True)
 
     def is_storage_initialized(self) -> bool:
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -157,7 +157,7 @@ class SqlJobClientBase(StagingJobClientBase):
         return dispositions
 
     def get_truncate_destination_table_dispositions(self) -> List[TWriteDisposition]:
-        if self.config.replace_strategy:
+        if self.config.replace_strategy == "truncate-and-insert":
             return ["replace"]
         return []
 
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -156,8 +156,10 @@ class SqlJobClientBase(StagingJobClientBase):
             dispositions.append("replace")
         return dispositions
 
-    def _should_truncate_destination_table(self, disposition: TWriteDisposition) -> bool:
-        return disposition == "replace" and self.config.replace_strategy == "truncate-and-insert"
+    def get_truncate_destination_table_dispositions(self) -> List[TWriteDisposition]:
+        if self.config.replace_strategy:
+            return ["replace"]
+        return []
 
     def _create_merge_job(self, table_chain: Sequence[TTableSchema]) -> NewLoadJob:
         return SqlMergeJob.from_table_chain(table_chain, self.sql_client)
--- a/dlt/extract/extract.py
+++ b/dlt/extract/extract.py
@@ -86,7 +86,8 @@ def extract(
             resources_with_items.add(resource_name)
             storage.write_data_item(extract_id, schema.name, table_name, item, None)
 
-        def _write_dynamic_table(table_name: str, resource: DltResource, item: TDataItem) -> None:
+        def _write_dynamic_table(resource: DltResource, item: TDataItem) -> None:
+            table_name = resource._table_name_hint_fun(item)
             existing_table = dynamic_tables.get(table_name)
             if existing_table is None:
                 dynamic_tables[table_name] = [resource.table_schema(item)]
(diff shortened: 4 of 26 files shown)


dlt/extract/extract.py
```diff
@@ -121,23 +126,42 @@ def _write_static_table(resource: DltResource, table_name: str) -> None:
...
                         # write item belonging to table with static name
                         table_name = resource.table_name
+                        populated_tables.add(table_name)
                         _write_static_table(resource, table_name)
                         _write_item(table_name, pipe_item.item)
+
+            # find defined resources that did not yield any pipeitems and create empty jobs for them
+            for resource in source.resources.selected.values():
+                if resource.write_disposition != "replace" or resource._table_name_hint_fun:
+                    continue
```
sh-rp, line 157: how does this _table_name_hint_fun work?

dlt/common/destination/reference.py
```diff
@@ -176,6 +176,9 @@ def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) ->
     def restore_file_load(self, file_path: str) -> LoadJob:
         pass
 
+    def get_truncate_destination_table_dispositions(self) -> List[TWriteDisposition]:
```
rudolfix, line 179: this is "replace" by default.

dlt/extract/extract.py
```diff
@@ -72,6 +72,11 @@ def extract(
 
     with collector(f"Extract {source.name}"):
 
+        def _write_empty_file(table_name: str) -> None:
+            table_name = schema.naming.normalize_identifier(table_name)
+            collector.update(table_name)
```
rudolfix, line 77: don't update collector with empty tables

dlt/extract/extract.py
```diff
@@ -80,8 +85,7 @@ def _write_item(table_name: str, item: TDataItems) -> None:
             collector.update(table_name)
```
rudolfix, line 85: 1. you should track the resource names that generated data, not table names 
2. this is the only place that you should update the list. it is called in every case
3. pass both table_name and resource_name here, use table name for collector, resource name to track your set of resources

rudolfix: see below why we track resources, not tables

rudolfix (changes requested):
review of extract.py

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -232,9 +232,9 @@ def test_pipeline_load_parquet(destination_name: str) -> None:
     github_data.max_table_nesting = 2
     info = p.run(github_data, loader_file_format="parquet", write_disposition="replace")
     assert_load_info(info)
-    # make sure it was parquet only
+    # make sure it was parquet or sql inserts
     files = p.get_load_package_info(p.list_completed_load_packages()[1]).jobs["completed_jobs"]
-    assert all(f.job_file_info.file_format == "parquet" for f in files)
+    assert all(f.job_file_info.file_format in ["parquet", "sql"] for f in files)
```
rudolfix, line 237: we should have no "sql" jobs because we replace with truncate and without staging. if we have sql jobs we doing something wrong

tests/load/pipeline/test_replace_disposition.py
```diff
@@ -57,15 +57,6 @@ def load_items():
     # we should have all items loaded
     table_counts = load_table_counts(pipeline, *[t["name"] for t in pipeline.default_schema.data_tables()])
 
-    # in the classic truncate and insert replace for now no one knows what will happen, just check that we have
-    # max the items that we expect
-    if replace_strategy == "truncate-and-insert":
```
rudolfix, line 62: amazing!

rudolfix (changes requested):
two bigger things:
1. please check the `filesystem` destination. it also has a replace "problem" that is brute force fixed in the copy job. what we should do is to adopt the truncate-insert strategy and truncate data in `initialize_storage`. 
this code can just go from copy job to the client
```
        # fallback to replace for merge without any merge keys
        if write_disposition == 'merge':
            write_disposition = 'append' if has_merge_keys else 'replace'

        # replace existing files. also check if dir exists for bucket storages that cannot create dirs
        if write_disposition == 'replace' and fs_client.isdir(dataset_path):
            job_info = LoadStorage.parse_job_file_name(file_name)
            # remove those files
            search_prefix = posixpath.join(dataset_path, f"{schema_name}.{job_info.table_name}.")
            # but leave actual load id - files may be loaded from other threads
            ignore_prefix = posixpath.join(dataset_path, f"{schema_name}.{job_info.table_name}.{load_id}.")
            # NOTE: glob implementation in fsspec does not look thread safe, way better is to use ls and then filter
            all_files: List[str] = fs_client.ls(dataset_path, detail=False, refresh=True)
            items = [item for item in all_files if item.startswith(search_prefix) and not item.startswith(ignore_prefix)]
            # NOTE: deleting in chunks on s3 does not raise on access denied, file non existing and probably other errors
            # if items:
            #     fs_client.rm(items[0])
            for item in items:
                # ignore file not found as we can have races from other deleting threads
                try:
                    fs_client.rm_file(item)
                except FileNotFoundError:
                    pass
```
and we do not need to filter by ignore_prefix  and catch `FileNotFoundError` because we do truncate BEFORE we start any loading. so there is just one writer. this is so good

2. we need more tests
- test the situation with many tables per resource (the resources for github will be a good test)
- you must test all the file types we have: jsonl, insert-values, parquet to make sure empty files are correctly processed. empty == without rows. in particular I'm not sure of insert-values

dlt/extract/extract.py
```diff
@@ -121,23 +126,42 @@ def _write_static_table(resource: DltResource, table_name: str) -> None:
...
+                            populated_tables.add(table_name)
+                            _write_dynamic_table(table_name, resource, pipe_item.item)
                     else:
                         # write item belonging to table with static name
                         table_name = resource.table_name
+                        populated_tables.add(table_name)
                         _write_static_table(resource, table_name)
                         _write_item(table_name, pipe_item.item)
+
+            # find defined resources that did not yield any pipeitems and create empty jobs for them
```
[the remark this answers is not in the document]
sh-rp: In the case where the main resource yields none, the table does not show up in the group_tables_by_resource, so I had to add an extra bit of code for this case. I hope this is alright.

tests/load/pipeline/test_replace_disposition.py
sh-rp: :)

sh-rp: I did like the part where it says "no one knows what will happen" it's a shame we remove this golden bit of commenting.

sh-rp:
I have also decided to have the file destination fallback to append for merge. I think If we do this replace, then there will be problems when loading from staging destinations to redshift etc with merge. Data will go missing.

dlt/extract/extract.py
```diff
@@ -121,23 +126,38 @@ def _write_static_table(resource: DltResource, table_name: str) -> None:
...
                     _write_static_table(resource, table_name)
-                    _write_item(table_name, pipe_item.item)
+                    _write_item(table_name, resource.name, pipe_item.item)
                 else:
                     # get partial table from table template
                     if resource._table_name_hint_fun:
                         if isinstance(pipe_item.item, List):
                             for item in pipe_item.item:
-                                _write_dynamic_table(resource, item)
+                                table_name = resource._table_name_hint_fun(item)
```
rudolfix, line 139: you can move it back into `_write_dynamic_table`

dlt/extract/extract.py
```diff
@@ -121,23 +126,38 @@ def _write_static_table(resource: DltResource, table_name: str) -> None:
...
+                            _write_dynamic_table(table_name, resource, pipe_item.item)
                     else:
                         # write item belonging to table with static name
                         table_name = resource.table_name
                         _write_static_table(resource, table_name)
-                        _write_item(table_name, pipe_item.item)
+                        _write_item(table_name, resource.name, pipe_item.item)
+
+            # find defined resources that did not yield any pipeitems and create empty jobs for them
+            tables_by_resources = utils.group_tables_by_resource(schema.tables)
```
rudolfix, line 151: I think we can safely use `data_tables`. this eliminates `_dlt` tables and incomplete tables that were never created in the destination

dlt/extract/extract.py
```diff
@@ -121,23 +126,38 @@ def _write_static_table(resource: DltResource, table_name: str) -> None:
...
+            for resource in source.resources.selected.values():
+                if resource.write_disposition != "replace" or resource.name in resources_with_items:
+                    continue
+                if resource.name not in tables_by_resources:
```
rudolfix, line 155: IMO we do not need empty file. if resource is not in present for any complete table this means that there's nothing to truncated on the destination

docs/website/docs/general-usage/full-loading.md
```diff
@@ -40,18 +40,16 @@ replace_strategy = "staging-optimized"
...
-that ensures that parent tables and child tables will be in a consistent state if your pipeline run fails at some point during the loading. You
-may end up with new data in the main table and stale data in the its child tables. Furthermore on large loads, there is a possibility that not
-all data present in the resource will actually end up in the destination table, as `dlt` chunks data into predefined (and configurable) filesizes
-before loading into the destination, and there is no mechnism to ensure that chunked data will be appended back together. If you have large loads
-or consistently failing pipeline runs, it is better to choose one of the other two strategies. This strategy behaves the same way across all
-destinations.
+destination tables will be truncated at the beginning of the load and the new data will be inserted consecutively but not within the same transaction. 
+The downside of this strategy is, that your tables will have no data for a while until the load is completed. You
+may end up with new data in some tables and no data in other tables if the load fails during the run.
```
rudolfix, line 45: add something like "Such incomplete load may be however detected by checking if the `_dlt_loads` table [contains load id from `_dlt_load_id` of the replaced tables](https://dlthub.com/docs/dlt-ecosystem/visualizations/understanding-the-tables#load-ids)

dlt/destinations/filesystem/filesystem.py
```diff
@@ -106,6 +82,16 @@ def dataset_path(self) -> str:
         return posixpath.join(self.fs_path, self.config.dataset_name)
 
     def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
+        # clean up existing files for tables selected for truncating
+        if truncate_tables and self.fs_client.isdir(self.dataset_path):
+            all_files = self.fs_client.ls(self.dataset_path, detail=False, refresh=True)
+            for table in truncate_tables:
+                search_prefix = posixpath.join(self.dataset_path, f"{self.schema.name}.{table}.")
+                for item in all_files:
```
rudolfix, line 90: please preserve the comment form the above. it is a nice warning that you cannot trust fsspec` too much:
```
# NOTE: glob implementation in fsspec does not look thread safe, way better is to use ls and then filter
```
and
```
 # NOTE: deleting in chunks on s3 does not raise on access denied, file non existing and probably other errors
```

tests/load/test_job_client.py
```diff
@@ -485,6 +499,7 @@ def test_write_dispositions(client: SqlJobClientBase, write_disposition: str, re
                 assert len(db_rows) == idx + 1
             elif write_disposition == "replace":
                 # we overwrite with the same row. merge fallbacks to replace when no keys specified
+                print(db_rows)
```
rudolfix, line 502: you can remove that

tests/load/pipeline/test_replace_disposition.py
```diff
@@ -114,19 +103,41 @@ def items_with_subitems():
...
+                "id": 101,
+                "name": "sub item 101"
+            },{
+                "id": 101,
+                "name": "sub item 102"
+            }]
+        }]
+        yield dlt.mark.with_table_name(data, "items")
+        yield dlt.mark.with_table_name(data, "other_items")
+        yield None
```
rudolfix, line 136: what a little bother me is that when we added the round robin strategy to pipe iterator, we also forced the iterator to get next value if None is yielded. to None should never get out of the pipe:
```
item = None
            while item is None:
                item = next(gen)
```
so how it is possible that you've got this INSERT VALUE with NULL that we were looking at?
```

**Review by rudolfix — CHANGES_REQUESTED** [review]

**Comment by rudolfix** [comment]
> I have also decided to have the file destination fallback to append for merge. I think If we do this replace, then there will be problems when loading from staging destinations to redshift etc with merge. Data will go missing.

hmmm why? redshift destination copies files from the load package. all the other files were already copied and may be deleted (or not). I do not mean you should restore this edge case (this is not so important) - but no data will be missing in that case!

when we start implementing the delta lake then yeah we will replace old files with new but that's according to write disposition so all is still good

File: tests/load/pipeline/test_replace_disposition.py (hunk 14 above)
Comment (sh-rp) [reply tests/load/pipeline/test_replace_disposition.py:136]: This happens if yield None is the only statement in the resource and no data is ever yielded. I did not investigate why this happens though.

File: dlt/extract/extract.py (hunk 10 above)
Comment (sh-rp) [reply dlt/extract/extract.py:155]: yes you are right, I had an error in the test that required this.

File: dlt/destinations/job_client_impl.py [hunk 15]
```diff
@@ -156,8 +156,10 @@ def get_stage_dispositions(self) -> List[TWriteDisposition]:
             dispositions.append("replace")
         return dispositions
 
-    def _should_truncate_destination_table(self, disposition: TWriteDisposition) -> bool:
-        return disposition == "replace" and self.config.replace_strategy == "truncate-and-insert"
+    def get_truncate_destination_table_dispositions(self) -> List[TWriteDisposition]:
+        if self.config.replace_strategy:
```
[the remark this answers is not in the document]
sh-rp: crap, this was an editing error, good catch!

tests/load/pipeline/test_replace_disposition.py
sh-rp: Should I? I think this would be a problem for a separate ticket though and we should merge the replace soonish

sh-rp:
> > I have also decided to have the file destination fallback to append for merge. I think If we do this replace, then there will be problems when loading from staging destinations to redshift etc with merge. Data will go missing.
> 
> hmmm why? redshift destination copies files from the load package. all the other files were already copied and may be deleted (or not). I do not mean you should restore this edge case (this is not so important) - but no data will be missing in that case!
> 
> when we start implementing the delta lake then yeah we will replace old files with new but that's according to write disposition so all is still good

Hm I think you are right, I had a thought error here, I was already thinking about s3 and Athena. But I think it would be either good to keep it like this or throw an error if merge is used with filesystem destination.

rudolfix (approved):

rudolfix:
> Hm I think you are right, I had a thought error here, I was already thinking about s3 and Athena. But I think it would be either good to keep it like this or throw an error if merge is used with filesystem destination.

FYI: all the hints are handled as best effort by destinations. if something is not supported it gets ignored... so the existing code is good

Issue #241, opened 2023-04-09
`dlt init` should check if the local `dlt` version support the pipeline


rudolfix:
**Background**
When pipeline code is upgraded, it often requires a more recent `dlt` version that the one that user has. We should at least warn the user before updating the code.

**Tasks**
1. * [x] all pipelines have a `requirements.txt`. use it to get the dlt dependency version
2. * [x] there are helper functions to get the local dlt version. use standard python means (distutils??) to check if local version is matching the version string in requirements
3. * [x] perform that check during dlt init. if version does not match ask user if to continue
4. * [x] when listing pipelines check for the version and show it in the listing. warn about incompatibilities.
5. * [x] when adding/updating pipeline we generate `requirements.txt`, now it assumes that `dlt` dependency is never there. it must be smarter and actually use the `dlt` provided one. just remember to add extra to it with destination name
6. * [x] keep the required dlt version in `TVerifiedSourceFileIndex` which is dumped to `.sources` folder after successful `dlt init`

You may prepare pipelines for this upgrade first here: https://github.com/dlt-hub/verified-sources/issues/97

PR #514, opened 2023-07-25
dlt init version check

steinitzu:
Implements: https://github.com/dlt-hub/dlt/issues/241


rudolfix (approved):
excellent! also I've checked manually a few mocked cases and all seems to work

PR #516, opened 2023-07-26
Add example values to data types docs

burnash:


docs/website/docs/general-usage/schema.md
```diff
@@ -126,9 +126,19 @@ coerced in existing column.
 
 ### Data types
 
+| dlt Data Type | Source Value Example |
+| --- | --- |
+| text | `"hello world"` |
+| double | `45.678` |
+| bool | `True` |
+| timestamp | `2023-07-26T14:45:00Z` |
```
rudolfix, line 134: datetimes may be given via python datetime
- pendulum.now() or datetime.now()
- as ISO strings which get recognized with default schema settings

docs/website/docs/general-usage/schema.md
```diff
@@ -126,9 +126,19 @@ coerced in existing column.
...
+| bigint | 9876543210 |
+| binary | b'\x00\x01\x02\x03' |
+| complex | (4+3j) |
+| decimal | Decimal('4.56') |
+| wei | - |
```
rudolfix: example: 2**56


note: `wei` represents an integer in range of 1 to 2^256-1. `dlt` will try to fit the wei type into the numeric type at the destination of largest precision. currently only Postgres and parquet files are able to represent `wei` fully

rudolfix (changes requested):

rudolfix (approved):

PR #522, opened 2023-07-30
Basic AWS Athena support

sh-rp:
Notes:
This poc has schema evolution, iceberg tables for the system information and generally seems to work very well. A few things we need to discuss:
* I am not sure how to handle complex types if arbitrary complex types are not supported. I don't think we can use the map column type present in Athena.
* There is quite a bit of sql which is the same or very similar to the default Postgres client. There are some differences for Athena though. Some examples:
  * The default timestamp format is different
  * pyathena does not support the same parameter format as the other destinations do. See section "query with parameter" on this page: https://pypi.org/project/pyathena/#basic-usage. I am wondering if we should abstract execute query to be able to handle timestamps and also the different types of parameter insertion.
  * Athena needs to have backticks as enclosing for the table names in case tables start with an underscore.
* I still have to understand the role of the custom cursor, for now I am just using the default one pyathena provides without a wrapper
* I had to raise the min py version from 8.0 to 8.1, otherwise pyathena would not install.


dlt/destinations/job_client_impl.py
```diff
@@ -404,18 +405,20 @@ def _update_schema_in_storage(self, schema: Schema) -> None:
...
         schema_bytes = schema_str.encode("utf-8")
         if len(schema_bytes) > self.capabilities.max_text_data_type_length:
             # compress and to base64
             schema_str = base64.b64encode(zlib.compress(schema_bytes, level=9)).decode("ascii")
+        self._commit_schema_update(schema, schema_str)
+
+    def _commit_schema_update(self, schema: Schema, schema_str: str) -> None:
         # insert
         name = self.sql_client.make_qualified_table_name(VERSION_TABLE_NAME)
+        now_ts = str(pendulum.now())
```
rudolfix, line 420: IMO it should not be required to convert to string here. dbapi should be able to convert from date time. let's drop it and see what happens to tests

dlt/destinations/job_client_impl.py
```diff
@@ -404,18 +405,20 @@ def _update_schema_in_storage(self, schema: Schema) -> None:
...
         # values =  schema.version_hash, schema.name, schema.version, schema.ENGINE_VERSION, str(now_ts), schema_str
         self.sql_client.execute_sql(
             f"INSERT INTO {name}({self.VERSION_TABLE_SCHEMA_COLUMNS}) VALUES (%s, %s, %s, %s, %s, %s);", schema.stored_version_hash, schema.name, schema.version, schema.ENGINE_VERSION, now_ts, schema_str
```
rudolfix, line 423: this must work, converting to dict style should be easy, see above

rudolfix (changes requested):
LGTM
- we need to pass formalize staging config and pass it to destination instead of staging credentials (which will ne included)
- AwsCredentials already have region. you can use it in pyathena
- that there's `s3_staging_dir` is a little worrying. this is administrative overhead. we could probably build a very simple catalogue destination where we just create catalogue over files in buckets. so IMO pure AWS Glue implementation still make sense (we could persist schemas in `filesystem`)
- there are no multistatement SQL and no multi statement transactions. still we can try to implement merging with iceberg but surely as followup

dlt/destinations/athena/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+from dlt.destinations.athena.configuration import AthenaClientConfiguration
+from dlt.common.destination.reference import JobClientBase, DestinationClientConfiguration
+
+@with_config(spec=AthenaClientConfiguration, sections=(known_sections.DESTINATION, "athena",))
+def _configure(config: AthenaClientConfiguration = config.value) -> AthenaClientConfiguration:
+    return config
+
+def capabilities() -> DestinationCapabilitiesContext:
+    caps = DestinationCapabilitiesContext()
+    caps.preferred_loader_file_format = "parquet"
```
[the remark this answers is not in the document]
sh-rp: I am also wondering wether we should have some unified way to check the bucket url and raise an exception if you try to load to Athena via gcs for example..

dlt/destinations/athena/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    caps.supported_loader_file_formats = ["parquet"]
+    caps.preferred_staging_file_format = "parquet"
+    caps.supported_staging_file_formats = ["parquet"]
+    caps.escape_identifier = escape_athena_identifier
+    caps.max_identifier_length = 255
+    caps.max_column_identifier_length = 255
+    caps.max_query_length = 1 # we can only have one statement per query
```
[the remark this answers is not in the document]
sh-rp: yes, but it seems to still work that it only does one sql statement. should I create a new config parameter for the amount of sql commands per execute request where we can set this to one without using this hack?

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+from dlt.common.schema.typing import LOADS_TABLE_NAME, VERSION_TABLE_NAME
+from dlt.common.schema.utils import add_missing_hints
+
+from dlt.destinations.job_client_impl import SqlJobClientBase, StorageSchemaInfo
+from dlt.destinations.athena.configuration import AthenaClientConfiguration
+from dlt.destinations.job_impl import NewReferenceJob
+
+
+SCT_TO_HIVET: Dict[TDataType, str] = {
+    # "complex": "JSON",
```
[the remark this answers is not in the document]
sh-rp: seems to work!

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def drop_dataset(self) -> None:
+        pass
+
+    def fully_qualified_dataset_name(self, escape: bool = True) -> str:
+        return self.capabilities.escape_identifier(self.dataset_name) if escape else self.dataset_name
+
+    @contextmanager
+    @raise_database_error
+    def begin_transaction(self) -> Iterator[Any]:
```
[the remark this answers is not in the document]
sh-rp: I think we might only need it once we actually start using iceberg for the loaded data, no? In all other cases (system tables) there is always just one line written or deleted.

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+        else:
+            sql.append(f"""ALTER TABLE `{table_name}` ADD COLUMNS ({columns});""")
+
+        return sql
+
+    def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
+        """Starts SqlLoadJob for files ending with .sql or returns None to let derived classes to handle their specific jobs"""
+        return DoNothingJob(file_path)
+
+    def complete_load(self, load_id: str) -> None:
```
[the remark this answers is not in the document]
sh-rp: works!

dlt/destinations/athena/configuration.py
```diff
@@ -0,0 +1,14 @@
...
+from dlt.common.typing import TSecretStrValue
+
+from dlt.common.configuration import configspec
+from dlt.common.destination.reference import DestinationClientDwhConfiguration
+
+
+@configspec(init=True)
+class AthenaClientConfiguration(DestinationClientDwhConfiguration):
+    destination_name: Final[str] = "athena"  # type: ignore[misc]
+    aws_access_key_id: str = None
```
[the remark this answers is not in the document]
sh-rp: done

dlt/destinations/job_client_impl.py
sh-rp: yeah this I overlooked, thanks for finding this

dlt/destinations/job_client_impl.py
sh-rp: it does, I had to add a custom formatter but this is quite nice now

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+                return f
+
+    @contextmanager
+    @raise_database_error
+    def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DBApiCursor]:
+        assert isinstance(query, str)
+        db_args = args if args else kwargs if kwargs else None
+        if db_args:
+            # TODO: must provide much better refactoring of params
+            query = query.replace("%s", "?")
```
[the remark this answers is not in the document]
sh-rp: done

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+    def _commit_schema_update(self, schema: Schema, schema_str: str) -> None:
+        # TODO: figure out timestamp insertion later, nees syntax TIMESTAMP 'proper format'
+        # we need a java compatible date timestamp here
+        now_ts = pendulum.now().to_datetime_string()
+        sql = f"""INSERT INTO "{VERSION_TABLE_NAME}" ({self.VERSION_TABLE_SCHEMA_COLUMNS}) VALUES ('{schema.stored_version_hash}', '{schema.name}', {schema.version}, {schema.ENGINE_VERSION}, TIMESTAMP '{now_ts}','{schema_str}');"""
+        self.sql_client.execute_sql(
+            sql
+        )
+
+    def get_storage_table(self, table_name: str) -> Tuple[bool, TTableSchemaColumns]:
```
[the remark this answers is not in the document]
sh-rp: with the conversion of the dbargs it works now

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+    def fully_qualified_dataset_name(self, escape: bool = True) -> str:
+        return self.capabilities.escape_identifier(self.dataset_name) if escape else self.dataset_name
+
+    @contextmanager
+    @raise_database_error
+    def begin_transaction(self) -> Iterator[Any]:
+        yield self
+
+    @staticmethod
+    def _make_database_exception(ex: Exception) -> Exception:
```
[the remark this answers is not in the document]
sh-rp: I'm not quite sure if this is done right now, but I think it should be more or less correct according to the docs

dlt/common/configuration/specs/aws_credentials.py
```diff
@@ -13,6 +13,7 @@ class AwsCredentialsWithoutDefaults(CredentialsConfiguration):
     aws_secret_access_key: TSecretStrValue = None
     aws_session_token: Optional[TSecretStrValue] = None
     aws_profile: Optional[str] = None
+    aws_region: Optional[str] = None
```
rudolfix, line 16: take it from the boto Session below!

dlt/destinations/athena/__init__.py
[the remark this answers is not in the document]
rudolfix: does it work really? heh I see I apply it to schema migration indeed. it splits on ";"
what you should do -> in sql_client split by ";" and execute the lines in a loop and put a real value here.

dlt/common/destination/capabilities.py
```diff
@@ -42,6 +42,7 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     naming_convention: str = "snake_case"
     alter_add_multi_column: bool = True
     supports_truncate_command: bool = True
+    supports_numeric_precision_in_schema: bool = True
```
rudolfix, line 45: add that to `add-a-destination.md` also I'm worried we inflate this capabilities so adding more destinations becomes harder

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+    "binary": "binary",
+    "decimal": "decimal(%i,%i)"
+}
+
+HIVET_TO_SCT: Dict[str, TDataType] = {
+    "varchar": "text",
+    "double": "double",
+    "boolean": "bool",
+    "date": "date",
+    "timestamp(6)": "timestamp", # TODO: what is going on here?
```
rudolfix, line 51: iceberg vs parquet?

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+    "decimal": "decimal"
+}
+
+# add a formatter for pendulum to be used by pyathen dbapi
+def _format_pendulum_datetime(formatter: Formatter, escaper: Callable[[str], str], val: Any) -> Any:
+    # copied from https://github.com/laughingman7743/PyAthena/blob/f4b21a0b0f501f5c3504698e25081f491a541d4e/pyathena/formatter.py#L114
+    val_string = val.strftime("%Y-%m-%d %H:%M:%S.%f")
+    return f"""TIMESTAMP '{val_string}'"""
+
+class DLTAthenaFormatter(DefaultParameterFormatter):
```
rudolfix, line 64: you are a hero

dlt/destinations/athena/athena.py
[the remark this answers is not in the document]
rudolfix: we should be able to push this destination through all standard tests. it will catch all edge cases. `test_sql_client` etc.

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+        assert isinstance(query, str)
+
+        db_args = kwargs
+        # convert sql and params to PyFormat, as athena does not support anything else
+        if args:
+            query, db_args = self._convert_to_old_pyformat(query, args)
+            if kwargs:
+                db_args.update(kwargs)
+        try:
+            cursor = self._conn.cursor(formatter=DLTAthenaFormatter())
```
rudolfix, line 169: also here you can split query by ";". that will OFC mess with parameters thing but in case of dict I expect you can just pass them all after splitting

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+            cursor.execute(query, db_args)
+            yield cursor  # type: ignore
+        # TODO: figure out how to handle this
+        except DatabaseError as outer:
+            self.close_connection()
```
rudolfix, line 174: I think this is REST api really. there's no connection

dlt/destinations/path_utils.py
```diff
@@ -0,0 +1,64 @@
+# this can probably go some other place, but it is shared by destinations, so for now it is here
+from typing import List, Tuple
+
+import pendulum
+import re
+
+from dlt.destinations.exceptions import InvalidFilesystemLayout, CantExtractTablePrefix
+
+# TODO: ensure lauyout only has supported placeholders
+supported_placeholders = {
```
rudolfix, line 10: maybe this should really go to separate PR? must be really unit tested IMO

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def _get_table_update_sql(self, table_name: str, new_columns: Sequence[TColumnSchema], generate_alter: bool) -> List[str]:
+        # TODO: get bucket var from outside
+        bucket = self.config.staging_config.bucket_url
+        dataset = self.sql_client.dataset_name
+        schema_name = self.schema.name
+        sql: List[str] = []
+
+        # for the system tables we need to create empty iceberg tables to be able to run, DELETE and UPDATE queries
+        iceberg = self.schema.tables[table_name]["write_disposition"] == "skip"
```
rudolfix, line 225: `is_iceberg`

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,253 @@
...
+        columns = ", ".join([self._get_column_def_sql(c) for c in new_columns])
+
+        # this will fail if the table prefix is not properly defined
+        table_prefix = path_utils.get_table_prefix(self.config.staging_config.layout, schema_name=schema_name, table_name=table_name)
+        location = f"{bucket}/{dataset}/{table_prefix}"
```
rudolfix, line 230: this will come from `layout` right?

rudolfix (changes requested):
we are really close regarding the code. the testing will be a big thing.
- it should pass most of the `test_sql_client` tests.
- it should pass `test_job_client` tests that do not require loading files (ie. schema update tests)
- if we do it right it should pass all the pipelines, replace and staging tests in load (maybe via mocking `filesystem` as staging when it is required by pipeline?)

dlt/destinations/athena/athena.py
sh-rp: it does, see function above that line

dlt/destinations/path_utils.py
sh-rp: There are some unit tests, not super many though. I think we can keep it here actually, since the code is not crazy complicated.

dlt/common/destination/capabilities.py
sh-rp: Yes, you are right, I can also add this as some class var or internal setting that is not part of capabilities, not sure.

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,262 @@
...
+        self._conn: Connection = None
+        self.config = config
+
+    @raise_open_connection_error
+    def open_connection(self) -> Connection:
+        self._conn = connect(
+            schema_name=self.dataset_name,
+            s3_staging_dir=self.config.query_result_bucket,
+            region_name=self.config.credentials.aws_region,
+            aws_access_key_id=self.config.credentials.aws_access_key_id,
```
lukas-gust, line 100: Doesn't this force the use of aws access keys in secrets.toml or the respective env vars? It won't be able to fall back onto other boto3 auth methods.

dlt/destinations/athena/athena.py
lukas-gust: Might be nice to trim trailing / from the bucket in case it's part of the string from the config?

dlt/destinations/athena/athena.py
lukas-gust: Something like this
```
@raise_open_connection_error
    def open_connection(self) -> Connection:
        self._conn = connect(
            schema_name=self.dataset_name,
            s3_staging_dir=self.config.query_result_bucket,
            **self.config.credentials.to_native_representation())
        return self._conn
```


but also requires we modify the aws_credentials spec function that is used to match the region name arg. From common.specs.aws_credentials.py

```
def to_native_representation(self) -> Dict[str, Optional[str]]:
        """Return a dict that can be passed as kwargs to boto3 session"""
        d = dict(self)
        d['profile_name'] = d.pop('aws_profile')  # boto3 argument doesn't match env var name
        d['region_name'] = d.pop('aws_region')  # boto3 argument doesn't match env var name
        return d
```


Or we can expose the session from the aws spec?

rudolfix: good point! look here https://github.com/dlt-hub/dlt/pull/534/files#diff-86471f800e38d78e3737c979d879eb02b8a0ca3e956068747989bba466503d68
I think it is doing exactly what you describe, you can also pass boto3 instead of credentials dict
@sh-rp `to_native_representation` should return dict with names compatible with Athena, let's try to pass it unstructured as @lukas-gust shows

docs/website/docs/dlt-ecosystem/destinations/athena.md
```diff
@@ -0,0 +1,19 @@
...
+description: AWS Athena `dlt` destination
+keywords: [aws, athena, glue catalog]
+---
+
+# AWS Athena
+
+### Setup:
+* Provide aws credentials for an identity that has the athena full access role
+* Provide an s3 bucket which serves as the output for Athena queries, the aws identity will need access to this bucket too.
+* Ensure that the aws identity also had read and write access for the buckets that will have the files for the athena tables
```
nicor88, line 12: maybe providing a basic set of iam permissions needed to make the setup working will be better.

docs/website/docs/dlt-ecosystem/destinations/athena.md
```diff
@@ -0,0 +1,19 @@
+---
+title: AWS Athena / Glue Catalog
+description: AWS Athena `dlt` destination
+keywords: [aws, athena, glue catalog]
+---
+
+# AWS Athena
+
+### Setup:
```
nicor88, line 9: As a minimum infra the user need to have the following:
* An S3 bucket to store target data written by dlt (via athena)
* An S3 bucket used by athena to store query results

Some optional configurations
* Glue target database - by default athena uses `default` database
* Athena workgroup - by default the primary workgroup can be used

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,261 @@
...
+
+        # this will fail if the table prefix is not properly defined
+        table_prefix = path_utils.get_table_prefix(self.config.staging_config.layout, schema_name=schema_name, table_name=table_name)
+        location = f"{bucket}/{dataset}/{table_prefix}"
+
+        if is_iceberg and not generate_alter:
+            sql.append(f"""CREATE TABLE `{table_name}`
+                    ({columns})
+                    LOCATION '{location}'
+                    TBLPROPERTIES ('table_type'='ICEBERG', 'format'='parquet');""")
```
nicor88, line 244: consider to add the possibility to write partitioned tables - full reference here https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-creating-tables.html

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,261 @@
...
+        elif not generate_alter:
+            sql.append(f"""CREATE EXTERNAL TABLE `{table_name}`
+                    ({columns})
+                    STORED AS PARQUET
+                    LOCATION '{location}';""")
```
nicor88, line 249: as stated above, you want to give the possibility to create partitioned tables. In this specific case, for external table there is something more to do as partition values need to be added after writing to s3 is done.

sh-rp:
Tests that I can't fix [WIP]
FAILED tests/load/pipeline/test_pipelines.py::test_default_pipeline_names[True-athena-staging] - AssertionError
FAILED tests/load/pipeline/test_pipelines.py::test_default_pipeline_names[False-athena-staging] - AssertionError

dlt/destinations/filesystem/filesystem.py
```diff
@@ -82,17 +88,31 @@ def dataset_path(self) -> str:
...
+                except CantExtractTablePrefix:
+                    # crazy deletions might happen if table files are not clearly separated, so do nothing
+                    # TODO: print a warning to console?
+                    pass
+
+        # create destination dirs for all tables
+        for tschema in self.schema.tables.values():
+            table_prefix = path_utils.get_table_prefix(self.config.layout, self.schema.name, tschema["name"])
+            destination_dir = posixpath.join(self.dataset_path, table_prefix)
+            self.fs_client.makedirs(destination_dir, exist_ok=True)
```
rudolfix, line 115: does it really work? so if I have {data_set}/{schema_name}.{table_name} it will just create {data_set}? this code seems to work only with `/` separator where each table has a separate folder. what you should to here:
1. collect all paths for table_prefixes. remove the file name component so we have only folders
2. find unique paths
3. create those paths

so in the above case, even if you have 100 tables but all are in {dataset_name} folder, the code will create just one path

this also needs to be tested with a few layouts

dlt/destinations/job_client_impl.py
```diff
@@ -232,9 +233,12 @@ def _null_to_bool(v: str) -> bool:
                 return True
             raise ValueError(v)
 
+        fields = ["column_name", "data_type", "is_nullable"]
```
rudolfix, line 236: OK! I just wonder if the INFORMATION_SCHEMA table is defined in itself and we could pull those columns. but indeed - your method is faster. let's keep it

dlt/helpers/dbt/__init__.py
```diff
@@ -13,6 +13,12 @@
 
 DEFAULT_DBT_VERSION = ">=1.1,<1.6"
 
+# a map of destination names to dbt package names in case they don't match the pure destination name
+DBT_DESTINATION_MAP = {
+    "athena": "athena-community",
```
rudolfix, line 18: amazing 👍

dlt/pipeline/pipeline.py
```diff
@@ -920,7 +922,7 @@ def _get_destination_client(self, schema: Schema, initial_config: DestinationCli
                 "Dependencies for specific destinations are available as extras of dlt"
             )
 
-    def _get_staging_client(self, schema: Schema, initial_config: DestinationClientConfiguration = None) -> JobClientBase:
+    def ync(self, schema: Schema, initial_config: DestinationClientConfiguration = None) -> JobClientBase:
```
rudolfix, line 925: do I see correctly? `ync1?

dlt/pipeline/pipeline.py
```diff
@@ -1026,6 +1028,8 @@ def _resolve_loader_file_format(
             possible_file_formats = [f for f in dest_caps.supported_staging_file_formats if f in stage_caps.supported_loader_file_formats]
         if not file_format:
             if not stage_caps:
+                if not dest_caps.preferred_loader_file_format:
```
rudolfix, line 1031: maybe we should really add staging automatically if destination is used that requires it? it could happen in `_set_destination` and we can issue warning that we did that. WDT? this simplifies the code and we can also fix dlt init (I can do that)

tests/load/bigquery/test_bigquery_client.py
```diff
@@ -17,12 +17,15 @@
...
 from dlt.destinations.exceptions import LoadJobNotExistsException, LoadJobTerminalException
 
-from tests.utils import TEST_STORAGE_ROOT, delete_test_storage, preserve_environ
+from tests.utils import TEST_STORAGE_ROOT, delete_test_storage, preserve_environ, ALL_DESTINATIONS
 from tests.common.utils import json_case_path as common_json_case_path
 from tests.common.configuration.utils import environment
 from tests.load.utils import expect_load_file, prepare_table, yield_client_with_storage, cm_yield_client_with_storage
 
 
+pytestmark = pytest.mark.skipif('bigquery' not in ALL_DESTINATIONS, reason="BigQuery not configured")
```
rudolfix, line 26: uh! didn't we merge this? is this done against `devel` branch?

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -54,7 +54,7 @@ def test_successful_load(write_disposition: str, all_buckets_env: str, filesyste
         job_info = LoadStorage.parse_job_file_name(job.file_name())
         destination_path = posixpath.join(
             dataset_path,
-            f"{client.schema.name}.{job_info.table_name}.{load_id}.{job_info.file_id}.{job_info.file_format}"
+            f"{client.schema.name}/{job_info.table_name}/{load_id}.{job_info.file_id}.{job_info.file_format}"
```
rudolfix, line 57: we must test against several layouts here. so another parametrization needed. other tests can use standard layout

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -263,8 +272,16 @@ def _get_shuffled_events(shuffle: bool = dlt.secrets.value):
...
 
 
-@pytest.mark.parametrize('destination_name,github_resource', itertools.product(ALL_DESTINATIONS, [github_repo_events, github_repo_events_table_meta]))
-def test_merge_with_dispatch_and_incremental(destination_name: str, github_resource: DltResource) -> None:
+
+@pytest.mark.parametrize("destination_config", destinations_configs(default_configs=True), ids=lambda x: x.name)
+@pytest.mark.parametrize("github_resource",[github_repo_events, github_repo_events_table_meta])
+def test_merge_with_dispatch_and_incremental(destination_config: DestinationTestConfiguration, github_resource: DltResource) -> None:
+
+    # for athena we need to skip this test, as athena does not want a query string longer than 262144 (bytes?) and this is the
```
rudolfix, line 280: what is failing here? getting table counts? btw. we could set correctly this:
caps.max_query_length = 16 * 1024 * 1024
caps.is_max_query_length_in_bytes = True

also this is merge resource. this means that for complicated datasets the merge resource will not work (or we'll be executing it line by line anyway, right)?

dlt/destinations/athena/athena.py
```diff
@@ -0,0 +1,289 @@
...
+        # convert sql and params to PyFormat, as athena does not support anything else
+        if args:
+            query, db_args = self._convert_to_old_pyformat(query, args)
+            if kwargs:
+                db_args.update(kwargs)
+        cursor = self._conn.cursor(formatter=DLTAthenaFormatter())
+        for query_line in query.split(";"):
+            if query_line.strip():
+                try: 
+                    cursor.execute(query_line, db_args)
```
rudolfix, line 197: should we wrap this cursor in one that provides `df()` method? like postgres?

rudolfix: it basically forwards most of the standard cursor methods to inner cursor so all shoudl work

tests/load/utils.py
```diff
@@ -267,6 +267,11 @@ def yield_client(
     # lookup for credentials in the section that is destination name
     with Container().injectable_context(ConfigSectionContext(sections=(destination_name,))):
         with destination.client(schema, dest_config) as client:
+            # athena requires staging config to be present, so stick this in there here
+            if destination_name == "athena":
```
rudolfix, line 271: true! or let's build this into pipeline itself?

tests/load/test_sql_client.py
```diff
@@ -295,7 +300,7 @@ def test_commit_transaction(client: SqlJobClientBase) -> None:
 
 @pytest.mark.parametrize('client', ALL_CLIENTS, indirect=True)
 def test_rollback_transaction(client: SqlJobClientBase) -> None:
```
rudolfix, line 302: this won't work on Athena

tests/load/test_sql_client.py
```diff
@@ -333,7 +338,7 @@ def test_rollback_transaction(client: SqlJobClientBase) -> None:
 
 @pytest.mark.parametrize('client', ALL_CLIENTS, indirect=True)
 def test_transaction_isolation(client: SqlJobClientBase) -> None:
```
rudolfix, line 340: this also will not work.

rudolfix (changes requested):
this is incredible work! thanks!
we should plan to merge this soon. there are parts that will overlap (slightly) with Weaviate
I do not have any deeper issues. Maybe we could select staging automatically if it is required by destination in the pipeline. this will also simplify tests and dlt init

rudolfix:
> Tests that I can't fix [WIP] tests/load/pipeline/test_pipelines.py::test_default_pipeline_names[True-athena-staging] - AssertionError tests/load/pipeline/test_pipelines.py::test_default_pipeline_names[False-athena-staging] - AssertionError tests/load/pipeline/test_dbt_helper.py::test_run_jaffle_package
> 
> tests/load/test_job_client.py::test_get_storage_table_with_all_types[athena_client]
> 
> * Fails because we don't store numeric precision in the Athena schema, so we can't return wei in _from_db_type in Athena
> 
> I am skipping a bunch of tests in tests/load/test_job_client.py which write datasets. I am not sure if I should somehow adopt these to destinations that have a staging destination, or just not test this part.

my take: let's make a release candidate and then I will review and fix failing tests. there's nothing worrying or serious here. yes - you can skip stuff that writes to storage. we could also use buckets to do that but that would be too far

lukas-gust:
@rudolfix I'd love to know when you've made a release candidate.

dlt/destinations/athena/configuration.py
```diff
@@ -0,0 +1,14 @@
+from typing import Final, Optional
+from dlt.common.typing import TSecretStrValue
+
+from dlt.common.configuration import configspec
+from dlt.common.destination.reference import DestinationClientDwhConfiguration
+from dlt.common.configuration.specs import  AwsCredentials
```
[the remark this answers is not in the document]
lukas-gust: I believe I'm mistaken here and just lack a full understanding.

After some testing however. If we create a boto3 session without specifying the profile name. The profile name comes out to be default. I.e.

```
>>> session = boto3.Session(region_name = 'us-west-2')
>>> session.profile_name
'default'
```

OR

```
>>> from pyathena import connect
>>> athena_client = connect(work_group="dev-dlt-athena-workgroup", region_name="us-west-2")
>>> athena_client.session
Session(region_name='us-west-2')
>>> athena_client.session.profile_name
'default'
```

AND from the a pod on our cluster the profile shows to be 'default' as well. The issue is that if there is no shared credentials file or config file but a profile name is provided then boto3 will not succeed in finding the profile. In our case there isn't one.

My suggestion is that we don't set the profile_name from the session and make it an intentional configuration.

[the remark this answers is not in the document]
lukas-gust: See here: 
https://github.com/lukas-gust/dlt/blob/7fd0b44170dc00394dfd6dc7301fbf83f3e4bac6/dlt/common/configuration/specs/aws_credentials.py#L53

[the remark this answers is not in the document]
rudolfix: I'm OK with doing that just let's make sure that all is clear. here's the part that is responsible for setting up the profile name (and other params) from boto3 session
```
def on_partial(self) -> None:
        # Try get default credentials
        session = self._to_session()
        if self._from_session(session) and not self.is_partial():
            self.resolve()
```
so first it passes the parameters provided by user to  instantiate new boto3 Session. if user sets the profile it will be passed to the session. If such profile is not found then boto3 raises:
```
botocore.exceptions.ProfileNotFound: The config profile (fake_profile) could not be found
```
as in the test below:
```
    environment["CREDENTIALS__PROFILE_NAME"] = "fake_profile"
    with pytest.raises(botocore.exceptions.ProfileNotFound):
        resolve_configuration(AwsCredentials())
```
so your problem is that boto3 returns incorrect profile name? so when we actually try to create new session later it will fail?

rudolfix:
> @rudolfix I'd love to know when you've made a release candidate.

we'll get back to you tomorrow. what is left is some tests not passing and possibly the AWS credentials. otherwise the scope is sufficient?

dlt/destinations/athena/configuration.py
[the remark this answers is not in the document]
lukas-gust: The problem seems to be if you don't give it a profile name it will set it to "default" and subsequently pass that argument to pyathena. That shouldn't happen in the cases where someone needs to go beyond config or credential files like I do. If I don't give a profile name then it should be set to `None` not `"default"`.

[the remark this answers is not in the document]
rudolfix: OK get it. @sh-rp I'll do a separate PR with it. it is just one line but will make several tests to fail which I'd fix separately

rudolfix:
> > @rudolfix I'd love to know when you've made a release candidate.
> 
> we'll get back to you tomorrow. what is left is some tests not passing and possibly the AWS credentials. otherwise the scope is sufficient?

@lukas-gust we'll do a pre-release on Wednesday

dlt/pipeline/pipeline.py
sh-rp: this one should be deleted, I have done that now

tests/load/bigquery/test_bigquery_client.py
sh-rp: somehow this one was missing, I am not sure why..

tests/load/test_sql_client.py
sh-rp: I have amended the caps with a setting wether transactions are supported

tests/load/pipeline/test_merge_disposition.py
sh-rp: 'INSERT INTO "github_3_202308141258537531"."_dlt_version"(version_hash, schema_name, version, engine_version, inserted_at, schema) is failing, because the serialised schema string is too long. So I would say this can not be solved, since this is a single command that is too long.

dlt/destinations/athena/athena.py
sh-rp: seems to work!

tests/load/pipeline/test_merge_disposition.py
rudolfix: SqlJobClient will compress the schema string if it goes beyond `max_text_data_type_length` so setting it to correct value will help :) anyway I'll try to do that and then we'll see

sh-rp:
More notes:
* the loads file of the filesystem destination does not respect the layout at the moment. I am not sure wether we need this or not..

lukas-gust:
> > > @rudolfix I'd love to know when you've made a release candidate.
> > 
> > 
> > we'll get back to you tomorrow. what is left is some tests not passing and possibly the AWS credentials. otherwise the scope is sufficient?
> 
> @lukas-gust we'll do a pre-release on Wednesday

Great to hear! The scope is absolutely sufficient! Using DLT as a library gives us a ton of flexibility to conform to the rest of our dev infrastructure. I greatly appreciate the timeliness of this implementation.

PR #525, opened 2023-07-31
Ignore hints prefixed with 'x-' in table_schema()

burnash:
This pull request adds a filtering feature to the validate_dict method in DltResourceSchema, ignoring fields starting with "x-".


rudolfix (changes requested):
Looks good but we have two more places where parts of the schema are validated (in schema/utils.py)
- `new_table`
- `new_column`

please also add tests for all the cases above so we know that we can apply such hints and they land in the schema at the end

dlt/extract/schema.py
```diff
@@ -67,7 +67,12 @@ def table_schema(self, item: TDataItem =  None) -> TPartialTableSchema:
         resolved_template.pop("incremental", None)
         table_schema = self._merge_keys(resolved_template)
         table_schema["resource"] = self._name
-        validate_dict(TPartialTableSchema, table_schema, f"new_table/{self._name}")
+        validate_dict(
+            spec=TPartialTableSchema,
+            doc=table_schema,
+            path=f"new_table/{self._name}",
+            filter_f=lambda k: not k.startswith("x-")
```
rudolfix, line 74: would be good to define this prefix ie. in schema/typings

rudolfix (changes requested):

dlt/extract/schema.py
burnash: Could you please elaborate on this?

burnash: @rudolfix how about

```
validate_dict_ignoring_xkeys = functools.partial(
    validate_dict,
    filter_f=lambda k: not k.startswith("x-")
)
```

rudolfix: OK I know what you mean. since we use "x-" prefix in several places I though that we can define it as constant in schema/typing.py
or better: the whole lambda could be reused

rudolfix:
@burnash 
regarding the tests: let's test `apply_hints` now. the other methods will test in #479 when we actually start adding `x-` hints to tables and column schemas:
1. add a test along `test_columns_argument` where you make sure you can add `x-` hints to column schema

we'll need to add some tests to `tests/pipeline/test_pipelines` where we use weaviate adapter but load to duck db to see if all the special hints are preserved OK - that we can do in #479 as well

burnash:
@rudolfix added tests and a partial to handle repeated lambda

rudolfix (changes requested):
please to not forget to use `validate_dict_ignoring_xkeys` in apply hints and new table/column methods!

burnash:
@rudolfix I believe I missed that only in `dlt/extract/schema.py` and `tests/common/test_validation.py`. Should be all fixed in db248717

tests/extract/test_decorators.py
```diff
@@ -167,20 +167,27 @@ def camelCase():
 
 def test_columns_argument() -> None:
 
-    @dlt.resource(name="user", columns={"tags": {"data_type": "complex"}})
+    @dlt.resource(name="user", columns={"tags": {"data_type": "complex", "x-extra": "x-annotation"}})
```
rudolfix, line 170: I like it - clever :)

rudolfix (approved):

PR #528, opened 2023-08-02
Separate credentials tests in CI

steinitzu:
WIP


.github/workflows/test_local_destinations.yml
```diff
@@ -0,0 +1,102 @@
...
+        uses: actions/cache@v3
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}-redshift
+
+      - name: Install dependencies
+        run: poetry install --no-interaction -E redshift -E duckdb -E gs -E s3 -E pyarrow
+
+      - run: |
+          poetry run pytest tests/load tests/cli -k '(not motherduck)'
```
rudolfix, line 80: let's move `tests/cli` here and remove it from test_destinations.

rudolfix: we use `postgres` for dlt deploy test

.github/workflows/test_destinations.yml
```diff
@@ -23,6 +23,8 @@ env:
   RUNTIME__SENTRY_DSN: https://6f6f7b6f8e0f458a89be4187603b55fe@o1061158.ingest.sentry.io/4504819859914752
   RUNTIME__LOG_LEVEL: ERROR
   RUNTIME__DLTHUB_TELEMETRY_SEGMENT_WRITE_KEY: TLJiyRkGVZGCi2TtjClamXpFcxAA1rSB
+  ALL_DESTINATIONS: "[\"duckdb\", \"postgres\"]"
```
rudolfix, line 26: the name is misleading :/
here also test redshift and filesystem (including s3 and gs buckets) so we should include it here, maybe also change the name to include redshift and filesystem

rudolfix: we ignore certain paths so tests do not fail on dependencies

tests/load/bigquery/test_bigquery_client.py
```diff
@@ -17,12 +17,15 @@
...
 from dlt.destinations.exceptions import LoadJobNotExistsException, LoadJobTerminalException
 
-from tests.utils import TEST_STORAGE_ROOT, delete_test_storage, preserve_environ
+from tests.utils import TEST_STORAGE_ROOT, delete_test_storage, preserve_environ, ALL_DESTINATIONS
 from tests.common.utils import json_case_path as common_json_case_path
 from tests.common.configuration.utils import environment
 from tests.load.utils import expect_load_file, prepare_table, yield_client_with_storage, cm_yield_client_with_storage
 
 
+pytestmark = pytest.mark.skipif('bigquery' not in ALL_DESTINATIONS, reason="BigQuery not configured")
```
rudolfix, line 26: what this line does?

rudolfix (changes requested):

PR #537, opened 2023-08-07
Add issue templates

burnash:
This PR adds a first draft of GitHub issues templates derived from GitHub generated templates.

Further improvement could be to implement [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms).


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,36 @@
...
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
```
rudolfix, line 11: please look at https://github.com/dlt-hub/verified-sources/blob/master/.github/ISSUE_TEMPLATE/bug-report.md

we should get more context on a bug (not always applicable) ie.
- data source type (verified source? API? just data)
- destination (bigquery? etc.)
- runtime env. so we know this is production or local laptop or Colab notebook

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
```
rudolfix, line 10: Before asking questions below I'd ask about context in which dlt is used: ie. we could ask
- I'm already dlt user
- I'd consider to use dlt but I'm lacking a feature
- I run dlt in production
- I run it for fun
etc.

rudolfix (changes requested):
check out the suggestions. let's add PR template that asks people to link issue

.github/ISSUE_TEMPLATE/bug_report.md
rudolfix: ohhh I'd also ask them to go to slack before and report there. I think we should have people check out tech help on slack before reporting bugs

burnash:
@rudolfix please review. You can see the rendered version in [my branch](https://github.com/dlt-hub/dlt/blob/enh/github_issues_template/.github/ISSUE_TEMPLATE/bug_report.yml).

burnash:
@rudolfix added a PR template as well, please take a look

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,126 @@
...
+description: Report a problem with dlt
+labels: ["kind:bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for reporting a bug for dlt! Please fill out the sections below.
+        If you are not sure if this is a bug or not, please join our [Slack](https://join.slack.com/t/dlthub-community/shared_invite/zt-1slox199h-HAE7EQoXmstkP_bTqal65g)
+        and ask in the #3-technical-help channel.
+  - type: dropdown
```
rudolfix, line 12: should be it a normal input? so people just paste the version

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,126 @@
...
+    attributes:
+      label: dlt version
+      description: What dlt version are you using?
```
rudolfix, line 15: we could maybe tell them to do `dlt --version` to get it

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,126 @@
...
+    attributes:
+      label: Python version
+      description: What Python version are you using?
+      multiple: false
+      options:
+        - "3.11"
+        - "3.10"
+        - "3.9"
+        - "3.8"
+        - "Below 3.8"
```
rudolfix, line 93: we do not support <3.8

rudolfix (changes requested):
just a few suggestions. then we can merge. I wonder what happens when issue is created using form. should still be in Markdown format I just wonder how it looks like :)

.github/ISSUE_TEMPLATE/bug_report.yml
burnash: Good point. Let's change that, also reduces the maintenance

.github/ISSUE_TEMPLATE/bug_report.yml
burnash: I've intentionally put this as a red flag, even though dlt won't install on Python<3.8. If you think we don't need it I can remove that option.

rudolfix (approved):

PR #538, opened 2023-08-07
aligns with base class, removes base64

rudolfix:


burnash:
Looks good

dlt/destinations/weaviate/weaviate.py
```diff
@@ -340,15 +395,15 @@ def _update_schema_in_storage(self, schema: Schema) -> None:
...
-            version_hash=record["version_hash"],
-            schema_name=record["schema_name"],
-            version=record["version"],
-            engine_version=record["engine_version"],
-            inserted_at=ensure_pendulum_datetime(record["inserted_at"]),
-            schema=schema_str,
-        )
+        # schema_str = record["schema"]
+        return StorageSchemaInfo(**record)
+        #     version_hash=record["version_hash"],
```
burnash, line 400: Should we remove the commented part?

tests/load/weaviate/test_pipeline.py
```diff
@@ -67,6 +72,7 @@ def some_data():
     info = pipeline.run(
         some_data(),
     )
+    assert_load_info(info)
 
     data.extend(next(generator_instance2))
     assert_class(info.pipeline, "SomeData", data)
```
burnash, line 78: This probably won't work now since `assert_class` relies on presence of `doc_id` key

burnash (changes requested):

tests/load/weaviate/test_pipeline.py
rudolfix: what `assert_load_info` does is what you do here:
```
package_info = pipeline.get_load_package_info(info.loads_ids[0])
    assert package_info.state == "loaded"
    assert len(package_info.jobs["failed_jobs"]) == 0
    assert len(package_info.jobs["completed_jobs"]) == 1
```

burnash: Yes, I understand, I was referring this line: 
`assert sorted(objects_without_dlt_keys, key=lambda d: d['doc_id']) == sorted(data, key=lambda d: d['doc_id'])`
It assumes that object will have `doc_id`, but it's only so in one test case.

Issue #378, opened 2023-06-05
Ensure python 3.11. compatibility


sh-rp:
There has been a change introduced in python 3.11 to prevent dataclasses from sharing mutable objects by accident. Our resources throw errors when we set mutable defaults such as Incremental. To reproduce install python 3.11 and simply run the common tests.`ERROR tests/extract/test_incremental.py - ValueError: mutable default <class 'dlt.extract.incremental.Incremental'> for field created_at is not allowed: use default_factory`. We could consider to run some tests on a machine with python 3.11 to ensure it works.

https://docs.python.org/3/library/dataclasses.html

PR #541, opened 2023-08-07
Fix config dataclasses on python 3.11

steinitzu:
https://github.com/dlt-hub/dlt/issues/378

Modified the `@configspec` decorator to automatically wrap any default values which are `BaseConfiguration` instances.

So that:

```python
@configspec
class MyConfig:
    created_at: dlt.sources.incremental = dlt.sources.incremental('created_at', '2022-01-01')
```
Becomes equivalent to:

```python
default_incremental = dlt.sources.incremental('created_at', '2022-01-01')

@configspec
class MyConfig:
    created_at: dlt.sources.incremental = dataclasses.field(
        default_factory=lambda: default_incremental.copy()
    )
```

Note:

1. `@configspec` `init` argument is removed.
   Default factories are called in the generated `__init__`, so now the
   dataclass is always created with `init=True` EXCEPT when the class (or base class) defines a custom init.
   
2. Wrapped default attributes are only set on the instance
    
    `MyConfig.created_at` is an `AttributeError`
    but `MyConfig().created_at` has the default value.  
    
    That's just how dataclass fields work but may lead to confusing bugs.

### DIFF
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10.x"
+          python-version: "3.11.x"
 
       - name: Install Poetry
         # https://github.com/snok/install-poetry#running-on-windows
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -23,6 +23,15 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        python-version: ["3.11.x"]
+        # Test all python versions on ubuntu only
+        include:
+          - os: "ubuntu-latest"
+            python-version: "3.8.x"
+          - os: "ubuntu-latest"
+            python-version: "3.9.x"
+          - os: "ubuntu-latest"
+            python-version: "3.10.x"
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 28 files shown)


.github/workflows/test_common.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.10.x"
+          python-version: "3.11.x"
```
rudolfix, line 38: is it possible to define a matrix of python versions only on ubuntu? and all other oses to run on 3.10 still? if not we can keep this and solve in some other place

rudolfix (changes requested):
looks like we could do a hack by making `Inceremental` hashable:
https://docs.python.org/3.11/whatsnew/3.11.html#dataclasses
> Change field default mutability check, allowing only defaults which are [hashable](https://docs.python.org/3.11/glossary.html#term-hashable) instead of any object which is not an instance of [dict](https://docs.python.org/3.11/library/stdtypes.html#dict), [list](https://docs.python.org/3.11/library/stdtypes.html#list) or [set](https://docs.python.org/3.11/library/stdtypes.html#set). (Contributed by Eric V. Smith in [bpo-44674](https://bugs.python.org/issue?@action=redirect&bpo=44674).)

but it is still a hack and passing mutable defaults is not a good idea in general and your solution looks good. my only worry is that configspec decorator is super hacky - but it was like that before

if you like hack with hashes above then you can go for it instead. IMO what we should really do is to convert
`dlt.sources.incremental` into a factory (maybe our Wrapper would do) so people use it when defining resources. what do you think? maybe simply wrapper should be always used which is a factory already?

rudolfix (approved):
this is good, thanx!

PR #544, opened 2023-08-08
weaviate naming convention

rudolfix:
there's a lot of small changes everywhere but right now just look at the tests and destination weaviate files. now all identifiers are coming in the correct namespace.


burnash (approved):
LGTM

burnash:
Awesome work!

PR #550, opened 2023-08-11
fixes the check for empty list of write dispositions to truncate

### LINKED ISSUE #549: dlt deletes pipeline state table
Hello! 👋 

Ran into quite a nasty bug,

Running a pipeline with `replace_strategy = "staging-optimized"` would consistently nuke the state table in BigQuery. Naturally I have months worth of state, luckily BigQuery snapshots everything so I was able to manually rollback. I confirmed the behavior across multiple invocations.

rudolfix:
fixes #549

### DIFF
--- a/tests/load/pipeline/utils.py
+++ b/tests/load/pipeline/utils.py
@@ -160,7 +160,7 @@ def drop_active_pipeline_data() -> None:
                 for schema_name in p.schema_names:
                     drop_func(schema_name)
 
-        p._wipe_working_folder()
+        # p._wipe_working_folder()
         # deactivate context
         Container()[PipelineContext].deactivate()
 
--- a/tests/load/test_dummy_client.py
+++ b/tests/load/test_dummy_client.py
@@ -20,6 +20,7 @@ from dlt.destinations import dummy
 from dlt.destinations.dummy import dummy as dummy_impl
 from dlt.destinations.dummy.configuration import DummyClientConfiguration
 from dlt.load.exceptions import LoadClientJobFailed, LoadClientJobRetry
+from dlt.common.schema.utils import get_top_level_table
 
 from tests.utils import clean_test_storage, init_test_logging, TEST_DICT_CONFIG_PROVIDER, preserve_environ
 from tests.load.utils import prepare_load_package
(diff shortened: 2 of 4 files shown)


tests/load/test_dummy_client.py
```diff
@@ -89,6 +90,46 @@ def test_unsupported_write_disposition() -> None:
...
+    assert load.get_completed_table_chain(load_id, schema, top_job_table) is None
+    # fake being completed
+    assert len(load.get_completed_table_chain(load_id, schema, top_job_table, "event_user.839c6e6b514e427687586ccc65bf133f.0.jsonl")) == 1
+    # actually complete
+    loop_top_job_table = get_top_level_table(schema.tables, "event_loop_interrupted")
+    load.load_storage.start_job(load_id, "event_loop_interrupted.839c6e6b514e427687586ccc65bf133f.0.jsonl")
+    assert load.get_completed_table_chain(load_id, schema, loop_top_job_table) is None
+    load.load_storage.complete_job(load_id, "event_loop_interrupted.839c6e6b514e427687586ccc65bf133f.0.jsonl")
+    assert load.get_completed_table_chain(load_id, schema, loop_top_job_table) == [schema.get_table("event_loop_interrupted")]
+    assert load.get_completed_table_chain(load_id, schema, loop_top_job_table, "event_user.839c6e6b514e427687586ccc65bf133f.0.jsonl") == [schema.get_table("event_loop_interrupted")]
```
z3z1ma, line 130: Why does the last pos arg say event users? What's that asserting by not being event loop?

tests/load/pipeline/utils.py
```diff
@@ -160,7 +160,7 @@ def _drop_dataset_sql(schema_name: str) -> None:
                 for schema_name in p.schema_names:
                     drop_func(schema_name)
 
-        p._wipe_working_folder()
+        # p._wipe_working_folder()
```
z3z1ma, line 163: Is this intentional?

z3z1ma (approved):
So what line made the difference. I see a function argument rename and a falsey check to an is None. Just for my own understanding :)

tests/load/pipeline/utils.py
rudolfix: no :) but also not harmful, I'll do separate commit to fix

rudolfix:
@z3z1ma the problem was None check but having a function that behaves differently for `None` and `[]` is not a good idea. at least now it is properly tested

PR #561, opened 2023-08-16
fixes sentinel class

rudolfix:


dlt/destinations/weaviate/weaviate_client.py
```diff
@@ -339,21 +344,23 @@ def drop_dataset(self) -> None:
...
         Deletes all classes in the dataset and all data associated with them.
         Deletes the sentinel class as well.
-        """
-        if not self.dataset_name:
-            raise ValueError(
-                "Cannot drop dataset. Dataset name is not set. "
-                "To delete all classes, use `delete_all_classes`."
-            )
 
+        If dataset name was not provided, it deletes all the tables in the current schema
```
burnash, line 348: I did not implement this intentionally: this could lead to unexpected data deletion if the user mistakenly forgets to specify dataset name (or it is somehow turns out to be empty. 
I suggest to be explicit about data deletion and split this functionality to two methods: 'drop_dataset' and 'delete_all'. So there's at least some safety net.

burnash (changes requested):

dlt/destinations/weaviate/weaviate_client.py
rudolfix: this does not delete all classes but only classes created by this client: the class names are taken from the current schema so only the data loaded by the client will be deleted. WDT?

burnash: In my opinion there's still a conceptual difference. I didn't notice `if class_name in class_name_list:` initially, it's not as destructive with this check I suppose.

burnash (approved):

PR #564, opened 2023-08-17
general state sync interface

sh-rp:


dlt/common/destination/reference.py
```diff
@@ -265,7 +276,23 @@ def _verify_schema(self) -> None:
...
+    @abstractmethod
+    def get_stored_schema_by_hash(self, version_hash: str) -> StorageSchemaInfo:
+        pass
+
+    @abstractmethod
+    def get_stored_state(self, state_table: str, pipeline_name: str) -> Optional[str]:
+        """Loads compressed state from destination storage"""
+        pass
+
+class WithStagingDataset(ABC):
```
sh-rp, line 295: I think ABC is needed as base class, otherwise the abstract method decorator will not fail when not implemented

dlt/destinations/weaviate/weaviate_client.py
```diff
@@ -454,37 +456,91 @@ def get_storage_table(self, table_name: str) -> Tuple[bool, TTableSchemaColumns]
...
+            for state in state_records:
+                load_id = state["_dlt_load_id"]
+                load_records = self.get_records(self.schema.loads_table_name,
+                     where={
+                        "path": ["load_id"],
+                        "operator": "Equal",
+                        "valueString": load_id,
+                     }, limit=1)
+                # if there is a load for this state which was successfull, return the state
+                if len(load_records) and load_records[0]["status"] == 0:
```
rudolfix, line 488: theoretically this will not work if there's transformation layer that adds more statuses ie. 1,2 etc. jus tthe presence of load is sufficient. don't check the status

dlt/pipeline/pipeline.py
```diff
@@ -903,10 +903,12 @@ def _get_destination_client_initial_config(self, destination: DestinationReferen
         return client_spec(credentials=credentials)
 
     def _get_destination_clients(self,
-        schema: Schema,
+        schema: Schema = None,
         initial_config: DestinationClientConfiguration = None,
         initial_staging_config: DestinationClientConfiguration = None
     ) -> Tuple[JobClientBase, JobClientBase]:
+        if not schema:
```
rudolfix, line 910: why you allow for that?

rudolfix (changes requested):
looks good. what we need to do next
1. weaviate should pass all the tests in tests/load/pipeline/test_restore_state.py. maybe we should add it to our configured pipelines with "supports_sql` as false which eliminates it from all tests
2. then we do need more tests for it.
3. all tests should pass before I'll review the `Pipeline`  class really :)

dlt/pipeline/pipeline.py
sh-rp: It made a few locations in the code easier to write, I can also not allow this and put the default schema everywhere manually where I need it. It's up to you.

dlt/common/destination/reference.py
```diff
@@ -185,6 +196,13 @@ def __init__(self, schema: Schema, config: DestinationClientConfiguration) -> No
         self.schema = schema
         self.config = config
 
+    @property
+    def dataset_name(self) -> str:
```
rudolfix, line 200: this belongs to `DestinationClientDwhConfiguration` at best and will also cover the property there. also check `normalize_dataset_name` in the config above if you need an effective dataset name as used by sql code

dlt/common/destination/reference.py
```diff
@@ -185,6 +196,13 @@ def __init__(self, schema: Schema, config: DestinationClientConfiguration) -> No
...
+        return None
+
+    def drop_dataset(self) -> None:
```
rudolfix, line 203: we should not need that... and if anything you can rename it to `drop_storage`. better let's discuss it

dlt/common/destination/reference.py
```diff
@@ -234,6 +252,11 @@ def complete_load(self, load_id: str) -> None:
         """Marks the load package with `load_id` as completed in the destination. Before such commit is done, the data with `load_id` is invalid."""
         pass
 
+    @abstractmethod
+    def clear_load(self, load_id: str) -> None:
```
rudolfix, line 256: what about `complete_load` above? what's wrong with it? looks like code duplication everywhere

dlt/pipeline/pipeline.py
```diff
@@ -556,6 +556,7 @@ def sync_destination(self, destination: TDestinationReferenceArg = None, staging
             self._props_to_state(state)
             self._save_state(state)
         except Exception as ex:
+            raise ex
```
rudolfix, line 559: debug code?

dlt/pipeline/pipeline.py
rudolfix: I'd require schema here. this methods is already doing too much inside

dlt/pipeline/pipeline.py
```diff
@@ -1114,38 +1117,29 @@ def _optional_sql_job_client(self, schema_name: str) -> Optional[SqlJobClientBas
...
             self.config.use_single_dataset = True
-            job_client = self._optional_sql_job_client(normalize_schema_name(self.pipeline_name))
-            if job_client:
-                # handle open connection exception silently
-                state = None
-                try:
-                    job_client.sql_client.open_connection()
-                except Exception:
-                    # pass on connection errors
-                    if raise_on_connection_error:
```
rudolfix, line 1132: why `raise_on_connection_error` was removed?

dlt/pipeline/pipeline.py
```diff
@@ -1114,38 +1117,29 @@ def _optional_sql_job_client(self, schema_name: str) -> Optional[SqlJobClientBas
...
-                    state = load_state_from_destination(self.pipeline_name, job_client.sql_client)
-                finally:
-                    job_client.sql_client.close_connection()
-
-                if state is None:
-                    logger.info(f"The state was not found in the destination {self.destination.__name__}:{dataset_name}")
+            schema_name = normalize_schema_name(self.pipeline_name)
+            with self._maybe_destination_capabilities():
+                schema = Schema(schema_name)
+                schema.update_normalizers()
```
rudolfix, line 1130: it is not a bad idea to do that but IMO all should work without it.  the schema should be created with naming convention coming from current destination caps. also see this method `_set_context` in Piepline. it is called in dlt.pipeline() when new instance is activated

rudolfix (changes requested):

dlt/pipeline/pipeline.py
sh-rp: it was never used anywhere..

dlt/pipeline/pipeline.py
```diff
@@ -1114,38 +1117,29 @@ def _optional_sql_job_client(self, schema_name: str) -> Optional[SqlJobClientBas
...
+                schema._schema_name = schema.naming.normalize_identifier(schema_name)
```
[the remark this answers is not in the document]
sh-rp: remove it and see, the tests will fail because the wrong not normalised schema name is used. maybe I am doing something wrong here, but the normalisers are not properly set in the schema if I keep it the way it was before.

dlt/pipeline/pipeline.py
```diff
@@ -1154,18 +1148,24 @@ def _get_schemas_from_destination(self, schema_names: Sequence[str], always_down
         # check which schemas are present in the pipeline and restore missing schemas
         restored_schemas: List[Schema] = []
         for schema_name in schema_names:
-            if not self._schema_storage.has_schema(schema_name) or always_download:
-                # assume client always exist
-                with self._optional_sql_job_client(schema_name) as job_client:
-                    schema_info = job_client.get_newest_schema_from_storage()
+            with self._maybe_destination_capabilities():
```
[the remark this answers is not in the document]
sh-rp: remove it and see, the tests will fail. maybe I am doing something wrong here, but the normalisers are not properly set in the schema if I keep it the way it was before.

dlt/pipeline/pipeline.py
sh-rp: removed

dlt/common/destination/reference.py
sh-rp: this is to clear the load with a certain id, I use that in the tests, because the tests where using sql before. it is a unified interface for sql and not sql clients. we can do something else though if you like.

dlt/common/destination/reference.py
sh-rp: up to you, we can rename it, I just needed a unified interface in the tests for sql and non sql clients.

dlt/pipeline/pipeline.py
sh-rp: this somehow does not work properly. If you create a pipeline and then sync with destination, the resulting schema does not have the right normalisers set.

dlt/pipeline/pipeline.py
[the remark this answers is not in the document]
rudolfix: I'd say tests are wrong ie. comparing dataset name without normalizing. OK I'll take a look

dlt/common/destination/reference.py
rudolfix: OK I get it now, the docstrings still say it is to add a load_id, not remove. this is needed by just one test. in that case I'd add it to the test itself. or even provoke the error by mocking `complete_load` wiht exception during the test so it is not executed

dlt/common/destination/reference.py
```diff
@@ -265,7 +296,28 @@ def _verify_schema(self) -> None:
...
+    def get_stored_schema_by_hash(self, version_hash: str) -> StorageSchemaInfo:
+        pass
+
+    @abstractmethod
+    def get_stored_state(self, state_table: str, pipeline_name: str) -> Optional[StateInfo]:
+        """Loads compressed state from destination storage"""
+        pass
+
+    @abstractmethod
+    def get_stored_states(self, state_table: str) -> List[StateInfo]:
```
rudolfix, line 316: we do not need that really. we just need `get_stored_state` above. it can be removed

rudolfix (changes requested):

dlt/common/destination/reference.py
sh-rp: def test_restore_state_parallel_changes(destination_config: DestinationTestConfiguration) -> None:

sh-rp:
@rudolfix about the clear_load, I can't comment on that inline for some reason. I have removed it and moved the clearing code back into the test. I understand that you don't want to mess up the production code with these special functions, but still we need to maintain them in the tests also, so intuitively to me it makes sense to keep them there. But for now I moved it out.

dlt/pipeline/state_sync.py
```diff
@@ -93,19 +94,16 @@ def state_resource(state: TPipelineState) -> DltResource:
...
 
-def load_state_from_destination(pipeline_name: str, sql_client: SqlClientBase[Any]) -> TPipelineState:
+def load_state_from_destination(pipeline_name: str, client: WithStateSync) -> TPipelineState:
     # NOTE: if dataset or table holding state does not exist, the sql_client will rise DestinationUndefinedEntity. caller must handle this
     # TODO: this must go into job client and STATE_TABLE_NAME + LOADS_TABLE_NAME must get normalized before using in the query
-    query = f"SELECT state FROM {STATE_TABLE_NAME} AS s JOIN {LOADS_TABLE_NAME} AS l ON l.load_id = s._dlt_load_id WHERE pipeline_name = %s AND l.status = 0 ORDER BY created_at DESC"
-    with sql_client.execute_query(query, pipeline_name) as cur:
-        row = cur.fetchone()
-    if not row:
+    table_name = cast(JobClientBase, client).schema.naming.normalize_table_identifier(STATE_TABLE_NAME)
```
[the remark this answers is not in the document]
sh-rp: Yes, maybe, but not in this PR I would say

dlt/pipeline/pipeline.py
sh-rp: I made it mandatory again

dlt/destinations/weaviate/weaviate_client.py
sh-rp: done

rudolfix (approved):

PR #572, opened 2023-08-21
tests refactor

sh-rp:


rudolfix (changes requested):
looks amazing! I also reviewed if the tests are run and we have a few skips:
- tests/extract tests/pipeline are not run in test_common.yml (not due to this PR). this is a big omission. let's add them. see also Makefile
- in test_local_destinations for some reason we are not running filesystem. why is it excluded? it should be able to run for selected buckets?

sh-rp:
> looks amazing! I also reviewed if the tests are run and we have a few skips:
> 
> * tests/extract tests/pipeline are not run in test_common.yml (not due to this PR). this is a big omission. let's add them. see also Makefile
> * in test_local_destinations for some reason we are not running filesystem. why is it excluded? it should be able to run for selected buckets?

The filesystem tests are all in test_destinations, I could split them up so s3 and gcs are in test_destinations and memory and file are in local_destinations? Intuitively I would somehow prefer to keep all the filesystem tests in one place to reduce the probability of incorrect configuration.

sh-rp:
Actually I have activated file system for local destinations now, it only runs in memory and locally there so it doesn't make much difference

rudolfix (approved):

PR #577, opened 2023-08-22
add article return of etl

adrianbr:


rudolfix (changes requested):
Adrian, without giving some working code samples and links to working code - like Rahul is doing - we are missing the target audience.

I see here 2-3 examples we could give:
- rename, reorder columns, anonymize/filter - those are 2-3 liners
- run regression model and score data on the fly
- convert PDF to text - this is 2-3 lines of code as well

I'm happy to write some of those. the others are done and you can include them.

The blogpost itself is very good - concise and short. I'd just stress the build part more!

adrianbr:
good point @rudolfix let me do that

Issue #579, opened 2023-08-23
Support s3 stage on Snowflake destination without requiring an access key & access key id


codingcyclist:
**Current behavior**

When using an external (s3) stage to back up my raw data while loading it into Snowflake, I need to provide an AWS access key and an AWS access key ID as credentials for the s3 filesystem destination. Otherwise, I get the following error:

Requiring an AWS access key and an AWS access key ID when using an s3 stage in DLT seems unnecessary because authentication between AWS and Snowflake is already configured when creating the storage integration. In fact, Snowflake allows loading from any external stage that references a storage integration without requiring additional credentials (see [here](https://community.snowflake.com/s/article/Cannot-access-S3-bucket-The-AWS-Access-Key-Id-you-provided-is-not-valid#:~:text=Cause%3A%20%5BThe%20AWS%20Access%20Key,user%20might%20have%20been%20deleted.)).

**Desired behavior**

DLT should reference the staged files through the name of the external stage instead of the bucket URL. AWS access key and an AWS access key ID are only required when referencing staged files via their bucket URL. When referencing them via the stage name, no additional credentials are required.

codingcyclist:
@rudolfix I'm curious about your opinion and your advice on how to properly test the underlying code change. So far, I just tested the PR "locally" with my own Snowflake / S3 instances

codingcyclist:
@rudolfix Happy to re-submit as a contributor. It's likely that this won't be my last PR as we're planning on using DLT quite a lot

rudolfix:
@codingcyclist invite sent. so please resubmit and then we can set up additional test. my take is that me or @sh-rp show you how to do it. it requires adding additional staging option test to existing config set here:
https://github.com/dlt-hub/dlt/blob/0.3.10/tests/load/pipeline/utils.py#L89

rudolfix:
oh actually we are testing this in here!
```
            DestinationTestConfiguration(destination="snowflake", staging="filesystem", file_format="jsonl", bucket_url=GCS_BUCKET, stage_name="PUBLIC.dlt_gcs_stage", extra_info="gcs-integration"),
            DestinationTestConfiguration(destination="snowflake", staging="filesystem", file_format="jsonl", bucket_url=AWS_BUCKET, stage_name="PUBLIC.dlt_s3_stage", extra_info="s3-integration")
```
it seems there was a bug in the code and we never allowed the stage without forwarding credentials. so we just need to make sure that existing tests are passing!

PR #580, opened 2023-08-23
579 - Don't require AWS credentials for Snowflake external stage

codingcyclist:
See #579


rudolfix:
merged in other branch

PR #582, opened 2023-08-24
NormalizerInfo: table counts trace

sh-rp:


dlt/pipeline/pipeline.py
```diff
@@ -617,6 +617,36 @@ def last_trace(self) -> PipelineTrace:
...
+        trace = self.last_trace
+        if not trace:
+            return None
+        for step in trace.steps:
+            if step.step == step_name:
+                return step
+        return None
+
+    @property
+    def last_extract_info(self) -> ExtractInfo:
```
rudolfix, line 630: 🙏

rudolfix (approved):
this looks perfect!

dlt/normalize/normalize.py
```diff
@@ -18,17 +18,29 @@
...
 TMapFuncType = Callable[[Schema, str, Sequence[str]], TMapFuncRV]  # input parameters: (schema name, load_id, list of files to process)
 # tuple returned by the worker
-TWorkerRV = Tuple[List[TSchemaUpdate], int, List[str]]
+TWorkerRV = Tuple[List[TSchemaUpdate], int, List[str], TRowCount]
+
+def increase_row_count(row_counts: TRowCount, table_name: str, count: int) -> None:
+    row_counts[table_name] = row_counts.get(table_name, 0) + count
 
+# merges row counts_2 into row_counts_1
+def merge_row_count(row_counts_1: TRowCount, row_counts_2: TRowCount) -> None:
```
rudolfix, line 40: actually you could move it common utils and give it 2-3 test cases?

rudolfix (changes requested):
we have normalize tests in `tests/normalize`. this is the only place where we test multi process normalization. so the only place that you will actually get counters across process boundary and merge them. please add tests there

sh-rp:
> we have normalize tests in `tests/normalize`. this is the only place where we test multi process normalization. so the only place that you will actually get counters across process boundary and merge them. please add tests there

I have added one that I think tests this, let me know if you think this is right.

dlt/pipeline/pipeline.py
```diff
@@ -617,6 +617,36 @@ def last_trace(self) -> PipelineTrace:
             return self._last_trace
         return load_trace(self.working_dir)
 
+    def last_pipeline_step_trace(self, step_name: TPipelineStep) -> PipelineStepTrace:
```
rudolfix, line 620: this is good. what I do not like it is that it should be a part of `PipelineTrace` class. we have plenty of methods and code in pipeline already and props below have nothing to do with pipeline itself...

rudolfix (changes requested):
heh sorry I still have one issue :)

PR #587, opened 2023-08-25
Feat: don't require AWS credentials for ext stage

codingcyclist:
Re-opening #580 as contributor

As special note on testing this "feature": I went down the pragmatic route by just having two s3-staged Snowflake destination configs in the tests, one with and one without a stage name.

For the config that contains a stage name, dlt will reference the staged data through the stage name and therefore, Snowflake will load the data through the underlying storage integration without requiring any explicit s3 credentials.

For the config that has NO stage name, dlt will reference the staged data through the s3 URL and therefore, Snowflake will rely on the access key and access key ID being provided explicitly in the loading statement.

A more "rigorous" way of testing it would have been the following, but I found it prohibitively complex to implement as pipelines don't have separate APIs for loading data into the staging location before loading it from there into the destination table.
1. Load the normalized data into the s3 stage
2. Remove the s3 credentials from the config
3. Load the data from the s3 stage into Snowflake

Please let me know how you feel about the testing approach @rudolfix @sh-rp

### DIFF
--- a/dlt/destinations/snowflake/snowflake.py
+++ b/dlt/destinations/snowflake/snowflake.py
@@ -68,8 +68,11 @@ class SnowflakeLoadJob(LoadJob, FollowupJob):
         stage_file_path = ""
 
         if bucket_path:
-            # s3 credentials case
-            if bucket_path.startswith("s3://") and staging_credentials and isinstance(staging_credentials, AwsCredentialsWithoutDefaults):
+            # referencing an external s3 stage does not require explicit AWS credentials
+            if bucket_path.startswith("s3://") and stage_name:
+                from_clause = f"FROM '@{stage_name}'"
+            # referencing an staged files via a bucket URL requires explicit AWS credentials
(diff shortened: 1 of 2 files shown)


dlt/destinations/snowflake/snowflake.py
```diff
@@ -68,8 +68,12 @@ def __init__(
         stage_file_path = ""
 
         if bucket_path:
-            # s3 credentials case
-            if bucket_path.startswith("s3://") and staging_credentials and isinstance(staging_credentials, AwsCredentialsWithoutDefaults):
+            # referencing an external s3 stage does not require explicit AWS credentials
+            if bucket_path.startswith("s3://") and stage_name:
+                from_clause = f"FROM '@{stage_name}'"
+                files_clause = f"FILES = ('{urlparse(bucket_path).path.lstrip('/')}')"
```
rudolfix, line 74: this one was missing - without FILES, we were loading all files in stage.
btw. we had support for external stages but it got removed in code refactor. good catch

rudolfix (approved):
approved. we'll release it this Monday. I've added a FILES spec, otherwise all good
(you can still see a comment in review)

PR #594, opened 2023-08-27
[wip] basic schema freezing

sh-rp:


dlt/normalize/normalize.py
```diff
@@ -139,31 +139,38 @@ def _w_normalize_chunk(load_storage: LoadStorage, schema: Schema, load_id: str,
...
-                    increase_row_count(row_counts, table_name, 1)
+                if not row:
+                    continue
+                # decode pua types
+                for k, v in row.items():
+                    row[k] = custom_pua_decode(v)  # type: ignore
+                # coerce row of values into schema table, generating partial table with new columns if any
+                row, partial_table = schema.coerce_row(table_name, parent_table, row)
+                # check update
+                row, partial_table = schema.check_schema_update(table_name, row, partial_table, config.schema_update_mode)
```
rudolfix, line 150: don't call it if partial table is None! this is 99.99% of cases and this code is time critical

dlt/normalize/normalize.py
```diff
@@ -196,7 +203,7 @@ def map_parallel(self, schema: Schema, load_id: str, files: Sequence[str]) -> TM
         workers = self.pool._processes  # type: ignore
         chunk_files = self.group_worker_files(files, workers)
         schema_dict: TStoredSchema = schema.to_dict()
-        config_tuple = (self.normalize_storage.config, self.load_storage.config, self.config.destination_capabilities, schema_dict)
+        config_tuple = (self.config, self.normalize_storage.config, self.load_storage.config, self.config.destination_capabilities, schema_dict)
```
rudolfix, line 206: just FYI: this goes through process boundary and will be pickled. pay attention

dlt/common/schema/schema.py
```diff
@@ -174,7 +177,32 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
                     updated_table_partial["columns"][new_col_name] = new_col_def
 
         return new_row, updated_table_partial
-
+    
+    def check_schema_update(self, table_name: str, row: DictStrAny, partial_table: TPartialTableSchema, schema_update_mode: TSchemaUpdateMode) -> Tuple[DictStrAny, TPartialTableSchema]:
```
rudolfix, line 181: this is a solid implementation. but I'd like to describe it in the ticket (we can resue it for the docs) and maybe modify the requirements.

rudolfix (changes requested):

dlt/common/validation.py
```diff
@@ -50,9 +50,15 @@ def validate_dict(spec: Type[_TypedDict], doc: StrAny, path: str, filter_f: TFil
...
     def verify_prop(pk: str, pv: Any, t: Any) -> None:
         if is_optional_type(t):
+            # pass if value actually is none
+            if pv is None:
+                return
             t = extract_optional_type(t)
 
-        if is_literal_type(t):
+        # TODO: support for union types?
+        if pk == "schema_evolution_settings":
```
sh-rp, line 59: we should have a look at this..

sh-rp:
Notes for updated implementation:
* We now have three locations for schema evolution settings which will take precedence in this order: source, resource and global override
* Schema evolution settings are stored in the schema, but will never be read from the schema. If a resource decorator does not have a schema evolution settings value set, it will also be cleared from the schema, as I think implicit behaviour is confusing. This is a bit different from write_disposition (which I think we could also discuss)
* There is a quite nice set of e2e tests for this, we could probably add some unit tests for the schema.

dlt/common/schema/utils.py
```diff
@@ -403,6 +412,9 @@ def merge_tables(table: TTableSchema, partial_table: TPartialTableSchema) -> TPa
     if table.get('parent') is None and (resource := partial_table.get('resource')):
         table['resource'] = resource
 
+    # always update evolution settings
```
sh-rp, line 415: note that here the schema always is updated.

dlt/extract/decorators.py
```diff
@@ -164,6 +168,10 @@ def _wrap(*args: Any, **kwargs: Any) -> DltSource:
                     if inspect.isgenerator(rv):
                         rv = list(rv)
 
+            # prepare schema
+            schema = schema.clone(update_normalizers=True)
+            schema.set_schema_evolution_settings(schema_evolution_settings)
```
sh-rp, line 173: note that here also the schema always is updated.

dlt/common/schema/schema.py
```diff
@@ -11,12 +11,19 @@
...
 from dlt.common.schema.typing import (COLUMN_HINTS, SCHEMA_ENGINE_VERSION, LOADS_TABLE_NAME, VERSION_TABLE_NAME, STATE_TABLE_NAME, TPartialTableSchema, TSchemaSettings, TSimpleRegex, TStoredSchema,
-                                      TSchemaTables, TTableSchema, TTableSchemaColumns, TColumnSchema, TColumnProp, TColumnHint, TTypeDetections)
+                                      TSchemaTables, TTableSchema, TTableSchemaColumns, TColumnSchema, TColumnProp, TColumnHint, TTypeDetections, TSchemaEvolutionModes, TSchemaEvolutionSettings)
 from dlt.common.schema.exceptions import (CannotCoerceColumnException, CannotCoerceNullException, InvalidSchemaName,
                                           ParentTableNotFoundException, SchemaCorruptedException)
 from dlt.common.validation import validate_dict
+from dlt.common.schema.exceptions import SchemaFrozenException
 
 
+DEFAULT_SCHEMA_EVOLUTION_MODES: TSchemaEvolutionModes = {
```
rudolfix, line 21: 1. do you have a preference for "Evolution" here? I like Schema Contract way more. ie DEFAULT_SCHEMA_CONTRACT_MODE
2. I'm a fan of simple settings that are strings (something like presets). can we have a few of them? they could be accepted in the same place where full TypedDict is accepted

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+        return settings
+
+
+    def check_schema_update(self, parent_table: str, table_name: str, row: DictStrAny, partial_table: TPartialTableSchema, schema_evolution_settings_override: TSchemaEvolutionSettings) -> Tuple[DictStrAny, TPartialTableSchema]:
+        """Checks if schema update mode allows for the requested changes, filter row or reject update, depending on the mode"""
+
+        assert partial_table
+
+        # for now we defined the schema as new if there are no data columns defined
+        has_columns = self.has_data_columns
```
rudolfix, line 215: I do not think we need this. if we have an empty schema that is completely frozen so be it- it will not change at all

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+        if not has_columns:
+            return row, partial_table
+
+        evolution_modes = self.resolve_evolution_settings_for_table(parent_table, table_name, schema_evolution_settings_override)
```
rudolfix, line 219: in my opinion `evolution_modes` should be passed from outside. it also allows to cache those per table externally. this code is very CPU critical

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+
+        # default settings allow all evolutions, skipp all else
+        if evolution_modes == DEFAULT_SCHEMA_EVOLUTION_MODES:
```
rudolfix, line 222: yeah that should happen outside

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+            return row, partial_table
+
+        table_exists = table_name in self.tables and len(self.tables[table_name].get("columns", {}))
```
rudolfix, line 225: use `get_table_columns` to make sure we allow only for complete columns. otherwise we can say that table is new

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+
+        # check case where we have a new table
+        if not table_exists:
+            if evolution_modes == "freeze-and-trim":
```
rudolfix, line 229: huh? `evolution_modes` is dict. how you can compare that. also you are missing tests

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+                return None, None
+            if evolution_modes["table"] in ["freeze-and-discard", "freeze-and-trim"]:
+                return None, None
+            if evolution_modes["table"] == "freeze-and-raise":
+                raise SchemaFrozenException(f"Trying to add table {table_name} but new tables are frozen.")
```
rudolfix, line 234: would be good to pass schema name here. also pass the partial table (not for display but for later inspection)

dlt/common/schema/schema.py
rudolfix: also this code assumes that when we update schema, all partial tables for a single tables are merged first. I do not think it is the case.
you need to cache which table is new in the normalizer. otherwise you can add only single column to a new table (when new table is allowed but adding columns to existing table is not)

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+
+        # check columns
+        for item in list(row.keys()):
+            for item in list(row.keys()):
+                # if this is a new column for an existing table...
+                if table_exists and item not in self.tables[table_name]["columns"]:
```
rudolfix, line 240: allow to update incomplete column here, treat incomplete column as new

dlt/common/schema/schema.py
```diff
@@ -177,6 +184,71 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+                    is_variant = item in partial_table["columns"] and partial_table["columns"][item].get("variant")
+                    if evolution_modes["column"] == "freeze-and-trim" or (is_variant and evolution_modes["column_variant"] == "freeze-and-trim"):
+                        row.pop(item)
+                        partial_table["columns"].pop(item)
+                    if evolution_modes["column"] == "freeze-and-discard" or (is_variant and evolution_modes["column_variant"] == "freeze-and-discard"):
+                        return None, None
+                    if evolution_modes["column"] == "freeze-and-raise" or (is_variant and evolution_modes["column_variant"] == "freeze-and-raise"):
+                        raise SchemaFrozenException(f"Trying to add column {item} to table {table_name}  but columns are frozen.")
```
rudolfix, line 248: we really need a good exception for variant type that explains that variant column tried to be created but failed

dlt/common/schema/typing.py
```diff
@@ -62,6 +62,15 @@ class TColumnSchema(TColumnSchemaBase, total=False):
 TColumnName = NewType("TColumnName", str)
 SIMPLE_REGEX_PREFIX = "re:"
 
+TSchemaEvolutionMode = Literal["evolve", "freeze-and-trim", "freeze-and-raise", "freeze-and-discard"]
+
+class TSchemaEvolutionModes(TypedDict, total=False):
+    """TypedDict defining the schema update settings"""
+    table: TSchemaEvolutionMode
+    column: TSchemaEvolutionMode
+    column_variant: TSchemaEvolutionMode
```
rudolfix, line 71: can we name it `data_type`? because here we allow for data type to evolve.

dlt/common/schema/typing.py
```diff
@@ -62,6 +62,15 @@ class TColumnSchema(TColumnSchemaBase, total=False):
 TColumnName = NewType("TColumnName", str)
 SIMPLE_REGEX_PREFIX = "re:"
 
+TSchemaEvolutionMode = Literal["evolve", "freeze-and-trim", "freeze-and-raise", "freeze-and-discard"]
```
rudolfix, line 65: I think those names do not explain what happens
* evolve - lets keep it for now
* freeze-and-trim -> discard-value
* freeze-and-raise -> freeze
* freeze-and-discard -> discard-row
WDYT?

dlt/common/schema/schema.py
```diff
@@ -351,6 +429,9 @@ def update_normalizers(self) -> None:
         normalizers["json"] = normalizers["json"] or self._normalizers_config["json"]
         self._configure_normalizers(normalizers)
 
+    def set_schema_evolution_settings(self, settings: TSchemaEvolutionSettings) -> None:
```
rudolfix, line 432: this should be called when new Schema instance is created or?

dlt/common/validation.py
rudolfix: yes - and that should be trivial

dlt/common/validation.py
```diff
@@ -50,9 +50,15 @@ def validate_dict(spec: Type[_TypedDict], doc: StrAny, path: str, filter_f: TFil
 
     def verify_prop(pk: str, pv: Any, t: Any) -> None:
         if is_optional_type(t):
+            # pass if value actually is none
```
rudolfix, line 53: huh I have plenty of places where optional types are not present and all works! how it could work before? IMO something is wrong here

dlt/extract/decorators.py
```diff
@@ -164,6 +168,10 @@ def _wrap(*args: Any, **kwargs: Any) -> DltSource:
                     if inspect.isgenerator(rv):
                         rv = list(rv)
 
+            # prepare schema
```
rudolfix, line 171: handle it in the same way as `root_key` and `max_table_nesting`. as a property of DltSource that does the magic inside

dlt/extract/decorators.py
```diff
@@ -164,6 +168,10 @@ def _wrap(*args: Any, **kwargs: Any) -> DltSource:
...
+            schema = schema.clone(update_normalizers=True)
+            schema.set_schema_evolution_settings(schema_evolution_settings)
+
             # convert to source
             s = DltSource.from_data(name, source_section, schema.clone(update_normalizers=True), rv)
             # apply hints
```
rudolfix, line 177: here!

dlt/normalize/normalize.py
```diff
@@ -26,22 +26,23 @@
...
-    def __init__(self, collector: Collector = NULL_COLLECTOR, schema_storage: SchemaStorage = None, config: NormalizeConfiguration = config.value) -> None:
+    def __init__(self, collector: Collector = NULL_COLLECTOR, schema_storage: SchemaStorage = None, config: NormalizeConfiguration = config.value, schema_evolution_settings: TSchemaEvolutionSettings = None) -> None:
         self.config = config
         self.collector = collector
         self.pool: ProcessPool = None
         self.normalize_storage: NormalizeStorage = None
         self.load_storage: LoadStorage = None
         self.schema_storage: SchemaStorage = None
         self._row_counts: TRowCount = {}
+        self.schema_evolution_settings = schema_evolution_settings
```
rudolfix, line 45: IMO there's no need to pass it here

dlt/normalize/normalize.py
```diff
@@ -98,7 +99,7 @@ def w_normalize_files(
                         items_count = 0
                         for line_no, line in enumerate(f):
                             items: List[TDataItem] = json.loads(line)
-                            partial_update, items_count, r_counts = Normalize._w_normalize_chunk(load_storage, schema, load_id, root_table_name, items)
+                            partial_update, items_count, r_counts = Normalize._w_normalize_chunk(load_storage, schema, load_id, root_table_name, items, schema_evolution_settings)
```
rudolfix, line 102: and here. it must be in the schema already

dlt/normalize/normalize.py
```diff
@@ -139,31 +140,38 @@ def _w_normalize_chunk(load_storage: LoadStorage, schema: Schema, load_id: str,
...
+                if not row:
+                    continue
+                # decode pua types
+                for k, v in row.items():
+                    row[k] = custom_pua_decode(v)  # type: ignore
+                # coerce row of values into schema table, generating partial table with new columns if any
+                row, partial_table = schema.coerce_row(table_name, parent_table, row)
+                # if we detect a migration, the check update
+                if partial_table:
+                    row, partial_table = schema.check_schema_update(parent_table, table_name, row, partial_table, schema_evolution_settings)
```
rudolfix, line 152: ok this will not happen often. so we do not need to cache. but evolving new tables with columns frozen will surely fail. we can add just one column this way. we need to let add columns until normalizer finishes. and we need to test it

dlt/pipeline/pipeline.py
```diff
@@ -458,7 +459,7 @@ def run(
 
         # normalize and load pending data
         if self.list_extracted_resources():
-            self.normalize(loader_file_format=loader_file_format)
+            self.normalize(loader_file_format=loader_file_format, schema_evolution_settings=schema_evolution_settings)
```
rudolfix, line 462: don't pass it to normalizer. just update schema here

dlt/pipeline/pipeline.py
```diff
@@ -878,6 +879,7 @@ def _extract_source(self, storage: ExtractorStorage, source: DltSource, max_para
         # note we are not merging props like max nesting or column propagation
         for table in source_schema.data_tables(include_incomplete=True):
             pipeline_schema.update_schema(pipeline_schema.normalize_table_identifiers(table))
+            pipeline_schema._settings["schema_evolution_settings"] = source_schema._settings.get("schema_evolution_settings")
```
rudolfix, line 882: then you do not need to do that here

dlt/pipeline/pipeline.py
```diff
@@ -878,6 +879,7 @@ def _extract_source(self, storage: ExtractorStorage, source: DltSource, max_para
         # note we are not merging props like max nesting or column propagation
         for table in source_schema.data_tables(include_incomplete=True):
             pipeline_schema.update_schema(pipeline_schema.normalize_table_identifiers(table))
```
rudolfix, line 881: ok this is very interesting place:
we actually create and change schemas in the extract stage.
currently we allow for that to happen freely. we do not apply data contract. to be precise. stuff changes when
- someone adds new resource
- someone is using apply_hints
- we create a new table based on data

should we apply data contract for the last case?

dlt/common/schema/schema.py
rudolfix: OK - I see (2) is implemented :)

rudolfix (changes requested):

rudolfix (changes requested):
this is really amazing. besides review comments (we are quite close in terms of functionality) I wonder if we should also apply contracts in the extract stage
specifically in `extract_with_schema` function. take a look. I think we should

I also think that we should mark tables that received data at the very end of normalization to be able to recognize new tables consistently starting from extract until the end of normalization

rudolfix:
@sh-rp besides review: I have a half done PR that fixes tons of undefined behaviour with incomplete coilumns and tables. also fixes several bugs and adds proper test coverage

my take: let's merge it first and then we rebase this branch. IMO it will help you a lot

dlt/common/validation.py
sh-rp: I have changed this whole section to support union types properly, I would say it is correct now, I really don't know why it worked before...

dlt/pipeline/pipeline.py
sh-rp: I have tests for adding new tables from a new resource and adding new tables as a subtable, and they both pass:
def test_freeze_new_tables(contract_setting: str, setting_location: str) -> None:
so it seems this case is covered already?

dlt/common/schema/schema.py
sh-rp: Are you sure? I thought all the schema updates are merged into the schema after the normalizer finished running, at least that is what the code looks like to me. I also have the case of adding a new table with multiple columns when new columns are disabled in the test and it passes.

dlt/normalize/normalize.py
sh-rp: Could you explain how a test would look exactly? As mentioned above, I am testing the creation of a table from a new resource and this works even if new columns are frozen.

sh-rp:
Generally I am really not so happy with the way this global override works now. It updates the schema and all main table settings in the schema if you set it. If you remove it, this stuff stays in the schema with (as I see it) very non-intuitive results. Similar to how it is with the write_dispositions as well.

dlt/pipeline/pipeline.py
sh-rp: I did this more or less, I do pass it here, but updating it in that function

sh-rp:
Plus I can't get this test to pass:
tests/extract/test_decorators.py::test_load_schema_for_callable, somehow the schema is changed on load and I don't know why.

dlt/common/schema/schema.py
```diff
@@ -351,6 +419,12 @@ def tables(self) -> TSchemaTables:
     def settings(self) -> TSchemaSettings:
         return self._settings
 
+    @property
+    def has_data_columns(self) -> bool:
```
rudolfix, line 423: please remove - I do not think it has use

dlt/common/schema/schema.py
rudolfix: you are right - but there are other places where schema is updated
1. schema is updated in extract: from hints but also from data (all hints may be lambdas - including columns)
2. each worker process in normalizer updates tables on the fly
```
                if partial_table:
                    # update schema and save the change
                    schema.update_schema(partial_table)
                    table_updates = schema_update.setdefault(table_name, [])
```
3. all partial tables are collected from all worker processes and "fresh" schema is updated again at the end of the normalization

how to test it
1. define a complete column in extract stage, it will get added to the table and passed to normalizer. in extract yield data which will create another column in normalizer (from data). that will IMO fail when you allow new tables but freeze columns creation
2. yield two rows. first row creates column1, second row creates column2 - that will also fail
3. make columns a lambda (dynamic hints) and yield two rows, for each rows lambda returns different column. that should already fail in extract phase

dlt/common/validation.py
rudolfix: just wow

dlt/normalize/normalize.py
rudolfix: I just explained how to test that above

rudolfix:
> > I also think that we should mark tables that received data at the very end of normalization to be able to recognize new tables consistently starting from extract until the end of normalization
> 
> What exactly do you mean with "mark tables". add some info into the schema that then gets removed again? or something else?

I "hope" that when you write tests I mentioned above it will be clear why we need some mark for new tables...

dlt/extract/decorators.py
```diff
@@ -77,6 +79,7 @@ def source(
     max_table_nesting: int = None,
     root_key: bool = False,
     schema: Schema = None,
+    schema_contract_settings: TSchemaContractSettings = None,
```
rudolfix, line 82: good but update docstrings

dlt/extract/decorators.py
```diff
@@ -245,6 +256,7 @@ def resource(
     columns: TTableHintTemplate[TAnySchemaColumns] = None,
     primary_key: TTableHintTemplate[TColumnNames] = None,
     merge_key: TTableHintTemplate[TColumnNames] = None,
+    schema_contract_settings: TTableHintTemplate[TSchemaContractSettings] = None,
```
rudolfix, line 259: update docstrings

dlt/extract/schema.py
```diff
@@ -206,7 +207,8 @@ def new_table_template(
         write_disposition: TTableHintTemplate[TWriteDisposition] = None,
         columns: TTableHintTemplate[TAnySchemaColumns] = None,
         primary_key: TTableHintTemplate[TColumnNames] = None,
-        merge_key: TTableHintTemplate[TColumnNames] = None
+        merge_key: TTableHintTemplate[TColumnNames] = None,
+        schema_contract_settings: TTableHintTemplate[TSchemaContractSettings] = None,
```
rudolfix, line 211: OK! but why it is not a part of `apply_hints`?

dlt/pipeline/pipeline.py
```diff
@@ -389,7 +393,8 @@ def run(
         columns: TAnySchemaColumns = None,
         primary_key: TColumnNames = None,
         schema: Schema = None,
-        loader_file_format: TLoaderFileFormat = None
+        loader_file_format: TLoaderFileFormat = None,
+        schema_contract_settings: TSchemaContractSettings = None
```
rudolfix, line 397: docstrings?

dlt/pipeline/pipeline.py
```diff
@@ -294,7 +294,7 @@ def extract(
     @with_runtime_trace
     @with_schemas_sync
     @with_config_section((known_sections.NORMALIZE,))
-    def normalize(self, workers: int = 1, loader_file_format: TLoaderFileFormat = None) -> NormalizeInfo:
+    def normalize(self, workers: int = 1, loader_file_format: TLoaderFileFormat = None, schema_contract_settings: TSchemaContractSettings = None) -> NormalizeInfo:
```
rudolfix, line 297: OK now I see that this setting does not make sense here OR we apply this to all the schemas in the pipeline. WDYT?

dlt/pipeline/pipeline.py
```diff
@@ -466,7 +471,7 @@ def run(
         # extract from the source
         if data is not None:
             self.extract(data, table_name=table_name, write_disposition=write_disposition, columns=columns, primary_key=primary_key, schema=schema)
-            self.normalize(loader_file_format=loader_file_format)
+            self.normalize(loader_file_format=loader_file_format, schema_contract_settings=schema_contract_settings)
```
rudolfix, line 474: it looks like we should pass it to extract stage, not normalize. and apply to all sources being extracted there.
do not send it to normalize - it will afffect all the schemas - not only those from sources actually extracted

dlt/pipeline/pipeline.py
```diff
@@ -877,6 +882,7 @@ def _extract_source(self, storage: ExtractorStorage, source: DltSource, max_para
             pipeline_schema.update_schema(
                 pipeline_schema.normalize_table_identifiers(table)
             )
+            pipeline_schema.set_schema_contract_settings(source_schema._settings.get("schema_contract_settings", {}))
```
rudolfix, line 885: why do you update this in a loop? once is enough. also it seems we need a method to merge two schemas - that would merge schema settings and properties uh.

docs/website/docs/general-usage/schema.md
```diff
@@ -152,6 +152,79 @@ or create a child table out of it.
 
 `time` data type is saved in destination without timezone info, if timezone is included it is stripped. E.g. `'14:01:02+02:00` -> `'14:01:02'`.
 
+## Data contracts and controlling schema evolution
```
rudolfix, line 155: I'd create a separate doc for schema contracts, I can help writing content here

dlt/pipeline/pipeline.py
rudolfix: I think you are not testing a case when 
1. table name is a lambda (dynamic) - see getting started with github_events. in that case you may get new tables when there's new event type. and I'm sure if you block new tables the test will fail
2. `columns` are lambdas and new columns are created from data
yes - there's a little overlap in (2) with what normalizer does and it will not happen often but we need to handle this

rudolfix:
> Generally I am really not so happy with the way this global override works now. It updates the schema and all main table settings in the schema if you set it. If you remove it, this stuff stays in the schema with (as I see it) very non-intuitive results. Similar to how it is with the write_dispositions as well.

OK I agree. let's keep existing behavior and in next step we can see what is more intuitive. IMO
1. `write_disposition` global override is not - you'd expect to back to previous one if you load again (ie. `replace` applies only once)
2. schema settings - quite the opposite. if I do freeze once I expect it to last
yeah - let's decide later

rudolfix (changes requested):

dlt/common/schema/schema.py
```diff
@@ -187,6 +195,66 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+        # check columns
+        for item in list(row.keys()):
+            for item in list(row.keys()):
+                # if this is a new column for an existing table...
+
+                if table_exists and (item not in self.tables[table_name]["columns"] or not utils.is_complete_column(self.tables[table_name]["columns"][item])):
+                    is_variant = (item in partial_table["columns"]) and partial_table["columns"][item].get("variant")
+                    if contract_modes["column"] == "discard_value" or (is_variant and contract_modes["data_type"] == "discard_value"):
+                        row.pop(item)
+                        partial_table = deepcopy(partial_table)
```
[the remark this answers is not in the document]
sh-rp: alright

dlt/common/schema/schema.py
```diff
@@ -187,6 +195,66 @@ def coerce_row(self, table_name: str, parent_table: str, row: StrAny) -> Tuple[D
...
+                        partial_table["columns"].pop(item)
+                    elif contract_modes["column"] == "discard_row" or (is_variant and contract_modes["data_type"] == "discard_row"):
+                        return None, None
+                    elif contract_modes["column"] == "freeze":
+                        raise SchemaFrozenException(self.name, table_name, f"Trying to add column {item} to table {table_name}  but columns are frozen.")
+                    elif is_variant and contract_modes["data_type"] == "freeze":
```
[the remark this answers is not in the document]
sh-rp: in my mind if columns are frozen, so are variants implicitly. But maybe you are right actually, I moved it.

dlt/pipeline/pipeline.py
sh-rp: I moved it, but am not quite sure if it is in the right location, I think so though.

dlt/pipeline/pipeline.py
sh-rp: moved it to extract

dlt/pipeline/pipeline.py
sh-rp: I added this method

sh-rp:
I have some questions about these additional tests and the behavior in the extract stage:
1. I have added the tests you recommended, could you have a look if they are correct? They're not all passing yet, but I want to know wether I am testing the right stuff
2. `test_dynamic_tables` passed without any code modifications, so I'm wondering if I made an error in the test there?
3. `test_dynamic_columns` I am not quite sure what I am supposed to test here. I made a lambda that returns different column definitions..

rudolfix:
>  I am not quite sure what I am supposed to test here. I made a lambda that returns different column definitions..
hmmm I wrote detailed specs for tests somewhere in the comments. but it seems gone... still looking

rudolfix:
@sh-rp hey!
1. I wrote missing tests. please see my commits and all the notes there
2. I changed how new_table behaves (see notes)
3. regarding the new_table mark yes I think it is a new table hint
- added when new table is added to the schema
- removed by the normalizer if data was seen
there are for sure some nuances but

dlt/normalize/normalize.py
```diff
@@ -254,20 +266,25 @@ def map_single(self, schema: Schema, load_id: str, files: Sequence[str]) -> TMap
...
     def spool_files(self, schema_name: str, load_id: str, map_f: TMapFuncType, files: Sequence[str]) -> None:
         schema = Normalize.load_or_create_schema(self.schema_storage, schema_name)
-
         # process files in parallel or in single thread, depending on map_f
         schema_updates, row_counts = map_f(schema, load_id, files)
+        # set all populated tables to populated
+        populated_updated = False
+        for table_name, count in row_counts.items():
+            if count > 0 and schema.tables[table_name]["populated"] is not True:
+                schema.tables[table_name]["populated"] = True
```
sh-rp, line 284: why does this not persist??

tests/load/test_freeze_and_data_contract.py
```diff
@@ -0,0 +1,591 @@
...
+        yield from [{
+            "id": 5,
+            "name": "dave",
+            "amount": 6,
+            "sub": [{"hello": "dave"}]
+        }]
+
+    # test variants
+    pipeline = get_pipeline()
+    pipeline.run([get_items_simple()])
```
sh-rp, line 422: @rudolfix for the tests to work and the populated attribute to be set we always have to do one load. should we maybe rename the populated hint to something else (but with the same functionality) and set it to true in the schema if a complete table schema is supplied via columns? I think that would be more intuitive. Of course we can also do apply_hints, but this seems a bit intuitive for the enduser.

dlt/extract/decorators.py
```diff
@@ -52,6 +52,7 @@ def source(
     max_table_nesting: int = None,
     root_key: bool = False,
     schema: Schema = None,
+    schema_contract_settings: TSchemaContractSettings = None,
```
rudolfix, line 55: rename everywhere

dlt/extract/source.py
```diff
@@ -625,6 +625,14 @@ def max_table_nesting(self) -> int:
     def max_table_nesting(self, value: int) -> None:
         RelationalNormalizer.update_normalizer_config(self._schema, {"max_nesting": value})
 
+    @property
+    def schema_contract_settings(self) -> TSchemaContractSettings:
```
rudolfix, line 629: rename....

dlt/pipeline/pipeline.py
```diff
@@ -872,11 +881,7 @@ def _extract_source(self, storage: ExtractorStorage, source: DltSource, max_para
         self._schema_storage.save_import_schema_if_not_exists(source_schema)
 
         # get the current schema and merge tables from source_schema
-        # note we are not merging props like max nesting or column propagation
-        for table in source_schema.data_tables(include_incomplete=True):
-            pipeline_schema.update_schema(
-                pipeline_schema.normalize_table_identifiers(table)
-            )
+        pipeline_schema.update_schema(source_schema)
```
rudolfix, line 884: on one hand - this is good idea. on the other hand the implementation must be similar to those in normalize py
- create partial (diff table)
- apply only if there are any changes
- apply the schema contract to check if allowed, see my big comment at the end

docs/website/docs/general-usage/data-contracts.md
```diff
@@ -0,0 +1,77 @@
+---
+title: Data Contracts
+description: Data contracts and controlling schema evolution
+keywords: [data contracts, schema, dlt schema, pydantic]
+---
+
+## Data contracts and controlling schema evolution
```
rudolfix, line 7: thx for doing it!

dlt/extract/extract.py
```diff
@@ -103,12 +109,43 @@ def _write_dynamic_table(resource: DltResource, item: TDataItem) -> None:
...
-                dynamic_tables[table_name] = [static_table]
+                if not _add_dynamic_table(resource, table_name=table_name):
+                    return
+            _write_item(table_name, resource.name, item)
+
+        def _add_dynamic_table(resource: DltResource, data_item: TDataItem = None, table_name: Optional[str] = None) -> bool:
+            """
+            Computes new table and does contract checks
+            """
+            # TODO: We have to normalize table identifiers here
```
sh-rp, line 123: TODO...

dlt/extract/extract.py
```diff
@@ -103,12 +109,43 @@ def _write_dynamic_table(resource: DltResource, item: TDataItem) -> None:
...
+            table = resource.compute_table_schema(data_item)
+            if table_name:
+                table["name"] = table_name
+
+            # fast exit if we already evaluated this
+            if table["name"] in disallowed_tables:
+                return False
+
+            # this is a new table so allow evolve once
+            # TODO: is this the correct check for a new table, should a table with only incomplete columns be new too?
```
sh-rp, line 133: question for @rudolfix

dlt/pipeline/pipeline.py
```diff
@@ -849,17 +856,18 @@ def append_data(data_item: Any) -> None:
...
         source_schema = source.schema
         source_schema.update_normalizers()
 
-        extract_id = extract_with_schema(storage, source, source_schema, self.collector, max_parallel_items, workers)
-
         # if source schema does not exist in the pipeline
         if source_schema.name not in self._schema_storage:
-            # save schema into the pipeline
-            self._schema_storage.save_schema(source_schema)
+            # TODO: here we should create a new schema but copy hints and possibly other settings
```
sh-rp, line 866: quetion @rudolfix

dlt/extract/schema.py
```diff
@@ -220,12 +224,19 @@ def new_table_template(
             columns = ensure_table_schema_columns_hint(columns)
             if not callable(columns):
                 columns = columns.values()  # type: ignore
+            is_table_complete = len([c for c in columns if c.get("name") and c.get("data_type")]) # type: ignore
         else:
             validator = None
+            is_table_complete = False
+
+        # freeze the columns if we have a fully defined table and no other explicit contract
+        if not schema_contract and is_table_complete:
```
sh-rp, line 233: this is how I enforce the pydantic models (and other schema column settings) automatically. I am not quite sure if this covers all cases really and if the "is_table_complete" does the right thing.

sh-rp: also i allow the table to be created if it adheres to the schema.. I think this makes more sense

dlt/extract/extract.py
```diff
@@ -193,9 +205,38 @@ def _write_dynamic_table(self, resource: DltResource, item: TDataItem) -> None:
...
         existing_table = self.dynamic_tables.get(table_name)
         if existing_table is None:
-            static_table = resource.compute_table_schema()
-            static_table["name"] = table_name
-            self.dynamic_tables[table_name] = [static_table]
+            if not self._add_dynamic_table(resource, table_name=table_name):
+                return
+        self._write_item(table_name, resource.name, items)
+
+    def _add_dynamic_table(self, resource: DltResource, data_item: TDataItem = None, table_name: Optional[str] = None) -> bool:
```
sh-rp, line 212: This will have to live here somewhere, as we need to be able to intercept the creation of new items if the table is not allowed. If you don't want the extraction code mixed up with the contracts stuff, we could subclass the base extractor and put this code there somewhere or alternatively inject some kind of function in extract_with_schema that deals with this. I am not quite sure myself how to do this more elegantly. Of course we could also have the full extraction run the way it was before and then clean up at the end, but then we would potentially write files that will be discarded afterwards.

rudolfix:
@sh-rp changes from previous version
- extract is working on just one schema that is the most recent schema present in the pipeline. it updates this schema as the changes come
- instead of filtering data in apply contracts in schema, I generate filters that get executed in extract and normalize phase. that's something we could do with Pydantic in the future
- Pydantic is now tightly integrated with contracts - models are synthesized according to the settings.
- I've improved arrow support. contracts now apply to it as well

PR #597, opened 2023-08-29
fix snipsync endless loop

sh-rp:
### Description
Fixes the endless loop of snip sync by implementing a custom watcher.


rudolfix (approved):
I can just trust that it works :)

PR #604, opened 2023-09-01
Feat: add time to pua serde

z3z1ma:
### Description

`datetime.time` types are not included in PUA, as noted by someone on slack -- this adds it. Its a bigger conversation around making json serialization more robust. In cases like Zendesk and Mongo, we have to clean the object before yielding it. Ideally, we could just extend the serializer.

### DIFF
--- a/dlt/common/json/__init__.py
+++ b/dlt/common/json/__init__.py
@@ -1,7 +1,7 @@
 import os
 import base64
 import dataclasses
-from datetime import date, datetime  # noqa: I251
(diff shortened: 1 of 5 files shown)


rudolfix:
@z3z1ma 
- good idea with customized serialization/deserialization. I'll add a followup ticket
- I changed your PR to keep time as ISO string. we do not have a TIME column implemented. another followup ticket

rudolfix (dismissed):

rudolfix (approved):

PR #606, opened 2023-09-02
Add support for TIME data type

### LINKED ISSUE #605: support TIME type
**Motivation**
we are getting requests to support TIME data type. this is the most helpful when syncing sql databases or mongo

**Tasks**
1. declare new schema type in dlt/common/schema/typing.py
2. map python type to schema type in `py_type_to_sc_type` function (there's TODO already)
3. map schema type to database type for each destination (ie. dlt/destinations/bigquery/bigquery.py)
4. update the json PUA encode/decode - this is half done (we serialize to string)
5. mind the parquet in `writers.py` - we'll need to put TIME type there as well

optional
6. add type coercions in `coerce_value` function. from iso string and from timestamp without days part. also add coercion from date to timestamp

**Tests**
1. add cases to `py_type_to_sc_type` tests
2. add cases to `coerce_value` tests
3. look at `test_load_with_all_types` and `TABLE_ROW` fixture, add datetime to tested columns
4.

steinitzu:
### Description

Implements support for "TIME" in all destinations

### Related Issues

Resolves #605 

### Additional Context

Not all destinations support "time with timezone" so I'm using db tz naive "time" types in all and the timezone gets stripped when value includes it.

### DIFF
--- a/dlt/common/json/__init__.py
+++ b/dlt/common/json/__init__.py
@@ -15,7 +15,6 @@ from dlt.common import pendulum
 from dlt.common.arithmetics import Decimal
 from dlt.common.wei import Wei
 from dlt.common.utils import map_nested_in_place
-from dlt.common.time import ensure_pendulum_date, ensure_pendulum_datetime, ensure_pendulum_time
 
 
 class SupportsJson(Protocol):
@@ -64,14 +63,7 @@ def custom_encode(obj: Any) -> str:
         return str(obj)
     # this works both for standard datetime and pendulum
     elif isinstance(obj, datetime):
-        # See "Date Time String Format" in the ECMA-262 specification.
-        r = obj.isoformat()
-        # leave microseconds alone
-        # if obj.microsecond:
-        #     r = r[:23] + r[26:]
-        if r.endswith('+00:00'):
-            r = r[:-6] + 'Z'
-        return r
+        return obj.isoformat()
     elif isinstance(obj, date):
         return obj.isoformat()
     elif isinstance(obj, time):
@@ -104,16 +96,23 @@ _WEI = '\uF02C'
 _TIME = '\uF02D'
 
 
+def _datetime_decoder(obj: str) -> datetime:
+    if obj.endswith('Z'):
+        # Backwards compatibility for data encoded with previous dlt version
+        # fromisoformat does not support Z suffix (until py3.11)
+        obj = obj[:-1] + '+00:00'
+    return pendulum.DateTime.fromisoformat(obj)  # type: ignore[attr-defined, no-any-return]
+
 # define decoder for each prefix
 DECODERS: List[Callable[[Any], Any]] = [
     Decimal,
-    ensure_pendulum_datetime,
-    ensure_pendulum_date,
+    _datetime_decoder,
+    pendulum.Date.fromisoformat,  # type: ignore[attr-defined]
     UUID,
     HexBytes,
     base64.b64decode,
     Wei,
-    ensure_pendulum_time,
+    pendulum.Time.fromisoformat,  # type: ignore[attr-defined]
 ]
 # how many decoders?
 PUA_CHARACTER_MAX = len(DECODERS)
@@ -127,10 +126,7 @@ def custom_pua_encode(obj: Any) -> str:
         return _DECIMAL + str(obj)
     # this works both for standard datetime and pendulum
     elif isinstance(obj, datetime):
-        r = obj.isoformat()
-        if r.endswith('+00:00'):
-            r = r[:-6] + 'Z'
-        return _DATETIME + r
+        return _DATETIME + obj.isoformat()
     elif isinstance(obj, date):
         return _DATE + obj.isoformat()
     elif isinstance(obj, time):
--- a/dlt/common/json/__init__.py
+++ b/dlt/common/json/__init__.py
@@ -11,10 +11,11 @@ try:
 except ImportError:
     PydanticBaseModel = None  # type: ignore[misc]
 
+from dlt.common import pendulum
 from dlt.common.arithmetics import Decimal
 from dlt.common.wei import Wei
 from dlt.common.utils import map_nested_in_place
-from dlt.common.time import parse_iso_like_datetime
+from dlt.common.time import ensure_pendulum_date, ensure_pendulum_datetime, ensure_pendulum_time
 
 
 class SupportsJson(Protocol):
@@ -106,13 +107,13 @@ _TIME = '\uF02D'
 # define decoder for each prefix
 DECODERS: List[Callable[[Any], Any]] = [
     Decimal,
-    parse_iso_like_datetime,
-    lambda s: parse_iso_like_datetime(s),
+    ensure_pendulum_datetime,
+    ensure_pendulum_date,
     UUID,
     HexBytes,
     base64.b64decode,
     Wei,
-    lambda s: s  # NOTE: we keep iso implementation until we have TIME, time.fromisoformat,
+    ensure_pendulum_time,
 ]
 # how many decoders?
 PUA_CHARACTER_MAX = len(DECODERS)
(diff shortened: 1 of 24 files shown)


dlt/common/json/__init__.py
```diff
@@ -106,13 +107,13 @@ def custom_encode(obj: Any) -> str:
 # define decoder for each prefix
 DECODERS: List[Callable[[Any], Any]] = [
     Decimal,
-    parse_iso_like_datetime,
-    lambda s: parse_iso_like_datetime(s),
+    ensure_pendulum_datetime,
```
rudolfix, line 110: we know the input is a string so here's no need to use `ensure_pendulum` this code is very time critical and if we can remove some code then better

dlt/common/time.py
```diff
@@ -114,13 +88,72 @@ def ensure_pendulum_datetime(value: TAnyDateTime) -> pendulum.DateTime:
...
+
+    Returns:
+        A pendulum.Time object
+    """
+
+    if isinstance(value, datetime.time):
+        if isinstance(value, pendulum.Time):
+            return value
+        return pendulum.time(value.hour, value.minute, value.second, value.microsecond)
+    elif isinstance(value, datetime.datetime):
```
rudolfix, line 115: is there any reason to extract time from date-time types? my take is to be more strict and accept only `str` which we can parse to a valid time and that's it.

dlt/destinations/athena/athena.py
```diff
@@ -41,7 +41,8 @@
     "timestamp": "timestamp",
     "bigint": "bigint",
     "binary": "binary",
-    "decimal": "decimal(%i,%i)"
+    "decimal": "decimal(%i,%i)",
+    "time": "time"
```
rudolfix, line 45: looks like athena does not support TIME for parquet based tables (only ICEBERG)
https://docs.aws.amazon.com/athena/latest/ug/querying-iceberg-supported-data-types.html

let's try to use timestamp or string and see if it is able to load. if it cannot load parquet time to str/timestamp then we'll need another destination capability to generate proper parquet files...

dlt/destinations/redshift/redshift.py
```diff
@@ -192,4 +194,3 @@ def _from_db_type(cls, pq_t: str, precision: Optional[int], scale: Optional[int]
             if (precision, scale) == cls.capabilities.wei_precision:
                 return "wei"
         return PGT_TO_SCT.get(pq_t, "text")
```
rudolfix, line 196: one of the tests (direct INSERT) fails with
```
>           assert c["data_type"] == expected_c["data_type"]
E           AssertionError: assert 'time' == 'text'
E             - text
E             + time
```
this means that the redshift type for time is not `time` but most probably "TIME WITHOUT TIME ZONE" (small caps I assume)

error when copy parquet/jsonl:
```
  -----------------------------------------------
  error:  Assert
  code:      1000
  context:   Unreachable - Type not supported.
  query:     61832098
  location:  dory_type_helper.hpp:427
  process:   padbmaster [pid=1073996391]
  -----------------------------------------------
```
maybe try time with time zone? if not you can block time handling in redshift copy job (bad luck for the users) but I really hope that you can load time from files :)

tests/cases.py
```diff
@@ -25,7 +25,7 @@
 }
 # TODO: a version after PUA decoder (time is not yet implemented end to end)
 JSON_TYPED_DICT_DECODED = dict(JSON_TYPED_DICT)
-JSON_TYPED_DICT_DECODED["time"] = JSON_TYPED_DICT["time"].isoformat()
+JSON_TYPED_DICT_DECODED["time"] = JSON_TYPED_DICT["time"]
```
rudolfix, line 28: you can remove this I asssume :)

tests/common/schema/test_coercion.py
```diff
@@ -200,6 +200,26 @@ def test_coerce_type_to_date() -> None:
     assert coerce_value("date", "text", "2022-04-26 10:36") == pendulum.parse("2022-04-26", exact=True)
 
 
+def test_coerce_type_to_time() -> None:
+    # from ISO time string
+    assert coerce_value("time", "text", "03:41:31.466000") == pendulum.parse("03:41:31.466000", exact=True)
+    # from datetime object
+    assert coerce_value("time", "timestamp", pendulum.datetime(1995, 5, 6, 00, 1, 1, tz='EST')) == pendulum.parse("00:01:01", exact=True)
```
rudolfix, line 207: as mentioned I'd be way more strict on coercion

tests/common/schema/test_coercion.py
```diff
@@ -200,6 +200,26 @@ def test_coerce_type_to_date() -> None:
     assert coerce_value("date", "text", "2022-04-26 10:36") == pendulum.parse("2022-04-26", exact=True)
```
rudolfix, line 200: also coerce time which should fail

rudolfix (changes requested):
we have tests on Athena and Redshift failing. see comments. ping me on slack if you need credentials to Athena.

btw. excellent job with simplifying date coercions

dlt/common/json/__init__.py
steinitzu: Yeah very true. Since we're only dealing with output from our own json encoder here I would just use the built in `fromisoformat`, that should be fastest.

dlt/destinations/athena/athena.py
steinitzu: Neither timestamp/string seems to work as-is:

`time: timestamp`:
```
pyathena.error.OperationalError: SERIALIZATION_ERROR: Could not serialize column 'col11' of type 'timestamp(3)' at position 1:11
```

`time: string`:
```
dlt.destinations.exceptions.DatabaseTerminalException: HIVE_CURSOR_ERROR: Failed to read Parquet file: s3://...
```
So by destination capability you mean some callback or mapping to convert types? There's nothing like this currently, right? Probably would need to convert the time value -> timestamp with dummy date also.

dlt/destinations/redshift/redshift.py
```diff
@@ -108,6 +108,8 @@ def execute(self, table: TTableSchema, bucket_path: str) -> None:
         file_type = ""
         dateformat = ""
         compression = ""
+        if table_schema_has_type(table, "time"):
```
rudolfix, line 111: heh. so let it be. please improve the message below. ie. 
"Redshift cannot load TIME columns from {ext} files. Switch to direct INSERT file format or convert datetime.time objects in your data to str or datetimes"

please also add a note in Redshift docs (we have other exceptions there)

dlt/destinations/athena/athena.py
rudolfix: let's just block it we do it in Redshift. so if there's a time column in the schema of the table to copy we reject the whole file. 
Please provide a nice explanation in exception and mention that in the docs.

I do not thing we need another mechanism to handle that. Maybe later....

rudolfix (changes requested):

dlt/destinations/redshift/redshift.py
steinitzu: Ok, blocked this in both athena and redshift and added to tests. Now we wait for CI :pray:

dlt/destinations/redshift/redshift.py
```diff
@@ -41,7 +41,8 @@
     "timestamp": "timestamp with time zone",
     "bigint": "bigint",
     "binary": "varbinary",
-    "decimal": "numeric(%i,%i)"
+    "decimal": "numeric(%i,%i)",
+    "time": "time with time zone"
```
rudolfix, line 45: drop time zone: redshift tests are not passing due to that

tests/load/pipeline/test_stage_loading.py
```diff
@@ -92,8 +93,12 @@ def test_all_data_types(destination_config: DestinationTestConfiguration) -> Non
 
     pipeline = destination_config.setup_pipeline('test_stage_loading', dataset_name="test_all_data_types" + uniq_id())
 
-    data_types = deepcopy(TABLE_ROW_ALL_DATA_TYPES)
     column_schemas = deepcopy(TABLE_UPDATE_COLUMNS_SCHEMA)
+    if destination_config.destination == "redshift" and destination_config.file_format in ('parquet', 'jsonl'):
```
rudolfix, line 97: you also need to exclude athena. test is not passing

rudolfix (changes requested):
PR is good! tests are not passing but reason is simple (relatively :)) see review

rudolfix (approved):

PR #612, opened 2023-09-06
fix apply hints

sh-rp:
Fixes a small bug in the apply_hints method, that would fail if column name in hints are not set.


rudolfix (approved):
fixed that in my branch. merged!

PR #613, opened 2023-09-07
remove snipsync

sh-rp:
This PR removes snip sync and adds a better custom implementation


rudolfix (changes requested):
I have a layout problem: putting all snippets in one big file where they are placed inside functions is putting our (developer) interest over the user interest (in my humble opinion). yes it is easier to test it (not sure if to manage it) but way harder to understand and run snippets separately by just getting into the right folder and doing `python snippet_1.py`

are we able to get just a fragment of a function? this will be handy in several other places. quickstart is quite special - all snippets are standalone examples. but in most cases we'll show only relevant parts of the code, not the whole

the other thing that I'm not sure about is that we remove snippets from the text. IMO `dhelp` (and possibly anything else that works with documents and embedding) will be less helpful if we remove all the code from docs. are we able to pre render those locally and then ingest in dhelp?

@burnash we ingest the md/mdx files directly without building anything right?

burnash:
@rudolfix yes, it would be much better if we keep the snippets along with the text. And yes, we ingest the files directly.

sh-rp:
@rudolfix and @burnash: Ok, then I will change it to update the code inline so dhelp does not break, I don't really like this code duplication though, and prefer the approach to assemble this at build time.

With regards to being able to run the scripts directly, I can also revert to having all the snippets in their own file. That said, we have a nice copy button in the docs that you can use to copy the whole code snippet out and paste it into the python file. So imho this should be fine. For the examples I am working on another approach.

@rudolfix yes we can have multiple snippets in one file, and we can also have overlapping snippets if we want.

rudolfix (approved):

PR #615, opened 2023-09-09
remove snipsync (#613)

rudolfix:
Here's snipsync remove branch resubmitted. I merged it to devel but I was not able to get it to work
1. no matter what I did, snippets were not inserted into code - even if I removed already existing snippets nothing happened
2. I tried to build docs with `npm run build && npm run serve` - still no snippets inserted
3. what are you using to get preview of mdx files in VSCode? could we use `md` extension instead? then the built in preview works fine. I do not see any difference in how docosaurus treats those files
4. `<!-- @@@DLT_SNIPPET_START start -->` is causing the file to render as badly formatted (all red) in VSCode. for some reason
`<!--@@@DLT_SNIPPET_START start-->` (without spaces after <-- and before -->) works well. again - maybe default markdown plugin is broken?
5. yarn.lock is back. it must be removed
6. please add mode nodules to ignored paths when running pytest in docs


docs/website/docs/getting-started.md
```diff
@@ -224,14 +227,16 @@ with engine.connect() as conn:
     pipeline = dlt.pipeline(
         pipeline_name='from_database',
         destination='duckdb',
-        dataset_name='genome_data',
+        dataset_name='genome_da ta',
```
rudolfix, line 230: what is this space doing here :)

docs/website/docs/reference/performance-snippets.py
```diff
@@ -0,0 +1,256 @@
+from conftest import string_toml_provider
+
+PARALLEL_CONFIG_TOML = """
```
rudolfix, line 3: will it create a toml snippet or python snippet
```
```toml
```

should be toml

docs/website/docs/reference/performance-snippets.py
```diff
@@ -0,0 +1,256 @@
...
+            yield item_slice
+    # @@@DLT_SNIPPET_END performance_chunking_chunk
+
+    assert len(list(database_cursor())) == 10000
+    assert len(list(database_cursor_chunked())) == 10000
+
+
+def test_toml_snippets() -> None:
+
+    string_toml_provider.update("""
```
rudolfix, line 131: toml snippets in performance were not actually used by the snippets themselves. those used by the snippets were in `config.toml` files

anyway what you do here is weird... you overwrite the values over and over. why? if you want to validate just parse toml

docs/website/docs/reference/performance.md
```diff
@@ -48,7 +49,8 @@ def database_cursor_chunked():
 * set extract buffers separately from normalize buffers
 * set extract buffers for particular source or resource
 
-```toml
+<!--@@@DLT_SNIPPET_START buffer_toml-->
+```py
```
rudolfix, line 53: yeah - should be toml

docs/website/package.json
```diff
@@ -4,10 +4,9 @@
   "private": true,
   "scripts": {
     "docusaurus": "docusaurus",
-    "start": "snipsync && node watch_snippets.js & docusaurus start",
-    "snipsync": "snipsync",
-    "build": "snipsync && docusaurus build",
-    "build:netlify": "snipsync && docusaurus build --out-dir build/docs",
+    "start": "node tools/update_snippets.js --watch & docusaurus start",
```
rudolfix, line 7: could you add update snippets as a command?

docs/website/docs/reference/performance-snippets.py
rudolfix: ok. ideally you would create those as toml files not parts of python code
- way easier to write for non-senior devs :)
- if we need them for snippets we can still load them from file via string provider
- you can test them in special test that will crawl the tomls and load them all

rudolfix (commented):
thx for handling this. pls see review

docs/website/docs/getting-started.md
sh-rp: whops :)

rudolfix (commented):
for completness:
```
OK this is very close now :slightly_smiling_face:
npm run start is updating snippets but after that the watcher does see toml changes just py. fixable?
will ctrl-C stop both watcher and http server? if not then we'll have two watchers when I run it next time. maybe let's have a separate command to watch snippets?
```

rudolfix (commented):
excellent! thanks!

PR #631, opened 2023-09-13
Use bare pip in the docs

burnash:
Python 3 is the default since long time ago, I don't think there's any reason to be explicit here.


burnash:
@rudolfix could you please double check if the code for Windows is correct?

rudolfix (approved):
approved. python3 and pip3 actually do not work on windows at all

Issue #636, opened 2023-09-14
Support base64 encoded private keys for snowflake destination


codingcyclist:
### Feature description

I feel it's very common to encode private RSA keys in base64 when reading their values from environment variables, particularly because you then no longer have to worry about escaping newline characters, etc. Hence, it would be fantastic if DLT could handle both plain text and base64-encoded private keys when doing key-pair authentication with Snowflake.

In fact, tools like DBT even require the private key to be in base64 format (see [here](https://docs.getdbt.com/docs/core/connect-data-platform/snowflake-setup#key-pair-authentication)):
> The private_key string should be in Base64-encoded DER format, representing the key bytes

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

We use GitHub actions to run our DLT pipelines that connect to Snowflake using key-pair authentication. It's much less a headache to expose the private key as a base64 encoded secret, rather than plain-text secret

### Proposed solution

When parsing Snowflake credentials, check whether the private key is provided in plain text or base64 encoded. If it's base64 encoded, turn it into plain text before proceeding as usual.

@rudolfix WDYT? I'm happy to take this one up :)

PR #637, opened 2023-09-14
Allow base64 encoded private keys for Snowflake destination

codingcyclist:
See https://github.com/dlt-hub/dlt/issues/636


tests/load/snowflake/test_snowflake_configuration.py
```diff
@@ -60,6 +60,31 @@ def test_to_connector_params() -> None:
         warehouse='warehouse1',
         role='role1',
     )
+    
+    pkey_str = Path('./tests/common/cases/secrets/encrypted-private-key-base64').read_text('utf8')
```
rudolfix, line 64: test is not passing because the RSA key string got decoded as base64 into a garbage that could not be decoded as utf-8 :) 

you need to catch decoding exceptions or just all exceptions

rudolfix (changes requested):
pls also look at linter tests. thanks for contributing this!

codingcyclist:
@rudolfix Looks like something is odd with the MS SQL connection? Otherwise, tests are passing now - thanks a lot for your helpful comments

rudolfix (approved):
LGTM!

rudolfix:
@codingcyclist MSSQL on azure is unstable. we'll move to testing with local version (most probably)

PR #638, opened 2023-09-15
Pydantic validation

### LINKED ISSUE #617: use full pydantic typing when defining columns and allow the pydantic model to validate data during extraction
**Background**
We have basic Pydantic support: we can define table schemas and load pydantic models. However we use just the basic Pydantic types and we do not use models for data validation.
Such validation was requested by community in #135 (data contracts). instead of extending our schema with all those enums, ranges and other validators let's just use Pydantic model in Python transform attached to a resource

**Requirements**
1. * [ ]  take look at types here https://docs.pydantic.dev/latest/usage/types/json/ and start recognize them when creating dlt schema from pydantic model. what for use we need is a JSON type - it was really missing. I also see Decimal there
2. * [ ] we must recognize Enums - ints and strings (and whatever else) when mapping to dlt schema
3. * [ ] let's use the model that got passed to validate data. it will be similar to our `incremental`. see below

**Validation Requirements**
1. * [ ] When pydantic model is passed to `apply_hints` we add a transform step (can be a add._map) where it is used to validate data
2. * [ ] we expose that step in DltResource - like we expose Incremental. we allow it to be set: if set to None we remove validation. otherwise it can be set to any other map function
3. * [ ] if validation fails we just raise. when #594 is merged we can recognize schema evolution settings and (1) drop the row or offending fields (2) allow for new/unknown fields to be present

@sh-rp @steinitzu WDYT? I think this is really easy to do and we'll get something way cooler than defining validators in JSON schema

steinitzu:
WIP, needs tests

resolves #617

### DIFF
--- a/dlt/common/libs/pydantic.py
+++ b/dlt/common/libs/pydantic.py
@@ -26,6 +26,10 @@ def pydantic_to_table_schema_columns(model: Union[BaseModel, Type[BaseModel]], s
     fields = model.__fields__
     for field_name, field in fields.items():
         annotation = field.annotation
+        if inner_annotation := getattr(annotation, 'inner_type', None):
+            # This applies to pydantic.Json fields, the inner type is the type after json parsing
+            # (In pydantic 2 the outer annotation is the final type)
+            annotation = inner_annotation
         nullable = is_optional_type(annotation)
 
         if is_union(annotation):
--- a/dlt/common/libs/pydantic.py
+++ b/dlt/common/libs/pydantic.py
@@ -1,4 +1,4 @@
-from typing import Type, Union, get_type_hints, get_args
+from typing import Type, Union, get_type_hints, get_args, Any
 
 from dlt.common.exceptions import MissingDependencyException
 from dlt.common.schema.typing import TTableSchemaColumns
@@ -6,7 +6,7 @@ from dlt.common.data_types import py_type_to_sc_type, TDataType
 from dlt.common.typing import is_optional_type, extract_inner_type, is_list_generic_type, is_dict_generic_type, is_union
 
 try:
-    from pydantic import BaseModel, Field
+    from pydantic import BaseModel, Field, Json
 except ImportError:
     raise MissingDependencyException("DLT pydantic Helpers", ["pydantic"], "DLT Helpers for for pydantic.")
 
@@ -36,6 +36,12 @@ def pydantic_to_table_schema_columns(model: Union[BaseModel, Type[BaseModel]], s
             inner_type = get_args(annotation)[0]
         else:
             inner_type = extract_inner_type(annotation)
+
+        if inner_type is Json:  # Same as `field: Json[Any]`
+            inner_type = Any
+
+        if inner_type is Any:  # Any fields will be inferred from data
+            continue
 
         if is_list_generic_type(inner_type):
             inner_type = list
--- a/dlt/common/schema/typing.py
+++ b/dlt/common/schema/typing.py
@@ -110,9 +110,3 @@ class TStoredSchema(TypedDict, total=False):
     settings: Optional[TSchemaSettings]
     tables: TSchemaTables
     normalizers: TNormalizersConfig
-
-
-class ColumnValidator(ABC):
-    @abstractmethod
-    def __call__(self, item: TDataItem, meta: Any = None) -> TDataItem:
-        ...
--- a/dlt/common/schema/typing.py
+++ b/dlt/common/schema/typing.py
@@ -1,10 +1,8 @@
-from abc import ABC, abstractmethod
-from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Set, Type, TypedDict, NewType, Union, get_args, Protocol
+from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Set, Type, TypedDict, NewType, Union, get_args
 from typing_extensions import Never
 
 from dlt.common.data_types import TDataType
 from dlt.common.normalizers.typing import TNormalizersConfig
-from dlt.common.typing import TDataItem
 
 try:
     from pydantic import BaseModel as _PydanticBaseModel
--- a/dlt/common/schema/typing.py
+++ b/dlt/common/schema/typing.py
@@ -114,5 +114,5 @@ class TStoredSchema(TypedDict, total=False):
 
 class ColumnValidator(ABC):
     @abstractmethod
-    def __call__(self, item: TDataItem) -> TDataItem:
+    def __call__(self, item: TDataItem, meta: Any = None) -> TDataItem:
         ...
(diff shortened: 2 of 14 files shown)


dlt/extract/exceptions.py
```diff
@@ -259,3 +259,9 @@ def __init__(self, source_name: str, schema_name: str) -> None:
 class IncrementalUnboundError(DltResourceException):
     def __init__(self, cursor_path: str) -> None:
         super().__init__("", f"The incremental definition with cursor path {cursor_path} is used without being bound to the resource. This most often happens when you create dynamic resource from a generator function that uses incremental. See https://dlthub.com/docs/general-usage/incremental-loading#incremental-loading-with-last-value for an example.")
+
+
+class ValidationError(ValueError, DltException):
+    def __init__(self, original_exception: Exception) ->None:
+        self.original_exception = original_exception
+        super().__init__(f"Schema validation failed: {original_exception}")
```
rudolfix, line 267: `Extracted data item could not be validated with %s` where %s is a validator class (ie`PydanticValidator`). Also please pass the model and the data item to this error so people can ew. catch it and print the item

dlt/common/schema/typing.py
```diff
@@ -108,3 +110,9 @@ class TStoredSchema(TypedDict, total=False):
     settings: Optional[TSchemaSettings]
     tables: TSchemaTables
     normalizers: TNormalizersConfig
+
+
+class ColumnValidator(ABC):
```
rudolfix, line 115: way better to add it to `extract/typing` where other transforms are defined,

dlt/extract/validation.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class PydanticValidator(ColumnValidator, Generic[_TPydanticModel]):
+    model: Type[_TPydanticModel]
+    def __init__(self, model: Type[_TPydanticModel]) -> None:
+        self.model = model
+
+    def __call__(self, item: TDataItem, meta: Any = None) -> _TPydanticModel:
+        """Validate a data item agains the pydantic model"""
+        if item is None:
```
rudolfix, line 24: note that `item` may be a list. if that is a list, validate elements of a list. (we could also synthesize a model that validates lists of objects automatically - could we way faster)

please also test it

tests/common/test_pydantic.py
```diff
@@ -31,6 +51,16 @@ class Model(BaseModel):
     blank_dict_field: dict  # type: ignore[type-arg]
     parametrized_dict_field: Dict[str, int]
 
+    str_enum_field: StrEnum
+    int_enum_field: IntEnum
+    # Both of these shouold coerce to str
+    mixed_enum_int_field: MixedEnum
+    mixed_enum_str_field: MixedEnum
+
+    json_field: Json[List[str]]
```
rudolfix, line 60: also test Any field and Json[Any]

tests/extract/test_validation.py
```diff
@@ -0,0 +1,126 @@
...
+    a: int
+    b: str
+
+
+def test_validator_model_in_decorator() -> None:
+    # model passed in decorator
+    @dlt.resource(columns=SimpleModel)
+    def some_data() -> t.Iterator[t.Dict[str, t.Any]]:
+        yield {"a": 1, "b": "2"}
+        yield {"a": 2, "b": "3"}
```
rudolfix, line 24: also yield a list of objects

dlt/extract/validation.py
```diff
@@ -0,0 +1,35 @@
...
+from dlt.common.schema.typing import TAnySchemaColumns, ColumnValidator
+from dlt.extract.typing import TTableHintTemplate
+
+
+_TPydanticModel = TypeVar("_TPydanticModel", bound=PydanticBaseModel)
+
+
+class PydanticValidator(ColumnValidator, Generic[_TPydanticModel]):
+    model: Type[_TPydanticModel]
+    def __init__(self, model: Type[_TPydanticModel]) -> None:
```
rudolfix, line 19: how does it work without implementing `bind`? something must be wrong because bind must be called on each transform step just before extract on the pipe starts. could you please check it out? if the reason is that I check some kind of base class somewhere then let's create a base class with `bind` for all transforms. this will be very useful

rudolfix (changes requested):
thx for nice tests! pls see review comments

dlt/common/libs/pydantic.py
```diff
@@ -26,6 +26,10 @@ def pydantic_to_table_schema_columns(model: Union[BaseModel, Type[BaseModel]], s
     fields = model.__fields__
     for field_name, field in fields.items():
         annotation = field.annotation
+        if inner_annotation := getattr(annotation, 'inner_type', None):
+            # This applies to pydantic.Json fields, the inner type is the type after json parsing
+            # (In pydantic 2 the outer annotation is the final type)
+            annotation = inner_annotation
```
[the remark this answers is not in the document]
steinitzu: Yeah, best to skip them. Currently I think it will just fail in py_type_to_sc_type

dlt/extract/pipe.py
```diff
@@ -143,7 +143,7 @@ def steps(self) -> List[TPipeStep]:
 
     def find(self, *step_type: AnyType) -> int:
         """Finds a step with object of type `step_type`"""
-        return next((i for i,v in enumerate(self._steps) if type(v) in step_type), -1)
+        return next((i for i,v in enumerate(self._steps) if isinstance(v, step_type)), -1)
```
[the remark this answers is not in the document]
steinitzu: It's better like this right? Or was it supposed to match only exact type for some reason?

dlt/extract/validation.py
steinitzu: Yep, it's easy to synthesize wrapper model. I assume it would be faster.   
Only tricky part is it would fail the whole list on one error, if you want to implement a "ignore invalid items" option. But think there's some ways to manage it.

rudolfix: we are adding such settings in #594 so we'll upgrade pydantic validation as well

dlt/extract/pipe.py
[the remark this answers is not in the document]
rudolfix: OK i didn't notice that `step_type` is a tuple. I think what you do is slightly better

dlt/extract/validation.py
rudolfix: OK - looks like will be synthesizing models depending on the settings above. so let's skip it for now. it will come back :)

dlt/extract/validation.py
steinitzu: Hmm yeah you mean this? https://github.com/dlt-hub/dlt/blob/sthor%2Fpydantic-validation/dlt/extract/pipe.py#L261-L264  
I guess it would be best to make this a subclass of `ItemTransform`?

dlt/extract/validation.py
steinitzu: So I just added synthesized model anyway: https://github.com/dlt-hub/dlt/pull/638/files#diff-ba6e7ddb3759fcdb3e599cf5ca145432d28d529845672325b5642e4d00aa4d3dR23-R26  
You could make this conditional depending how you  want to structure it. There are also some solutions to achieve list validation without raising in https://github.com/pydantic/pydantic/issues/2274 but could need separate implementations for pydantic 1 and 2

tests/extract/test_validation.py
steinitzu: All tests parametrized now to test both single item and list

steinitzu:
@rudolfix Please check now the updates :)

rudolfix (approved):
LGTM!

PR #642, opened 2023-09-19
Docs: auto api reference

AstrakhantsevaAA:
1. uses pydoc-markdown to render docosaurus reference
2. reference is rendered when building and committed to the repo, see `package.json`
3. placement in `Reference/API`
4. uses custom preprocessor which removes #### headers and converts REPL in code blocks
5. pydoc-markdown installed in docs group with poetry


rudolfix (approved):
good as a first version :)

Issue #39, opened 2022-06-27
add precision and scale to applicable types


rudolfix:
add precision and scale settings to following types
- text, decimal, binary

* [ ] represent precision and scale as table hints
* [ ] allow for database notation type(precision, scale) when loading schema
* [ ] save schema with that notation when `remove-defaults` are used

PR #646, opened 2023-09-20
Support for precision, scale in column schema

steinitzu:
### Description

WIP: Implement precision and scale hints for column schema. 

### Related Issues
- Resolves https://github.com/dlt-hub/dlt/issues/39


rudolfix (approved):
this is good! thanks for a clean implementation (type mapper) and tests. I just added precision info to docs

PR #649, opened 2023-09-21
[WIP] Synapse Destination

NeilRM:
### Description

Adds the Synapse destination to the dlt destinations list.

 This destination code is based off of the mssql destination, and makes the following changes:
 
 1. Updates "mssql" references to "synapse" in destination code
 2. Adds Synapse [Github Testing Workflow](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/.github/workflows/test_destination_synapse.yml)
 3. Adds synapse_escape_literal function in escape.py ([lines 91-115](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/common/data_writers/escape.py#L91))
 4. Updates max_text_data_type_length to 4000 (max character length for Synapse) ([init.py line 39](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/__init__.py#L39))
 5. Changes nvarchar(max) to nvarchar(4000) and varbinary(max) to varbinary(8000) in type mapping. (max length not supported in Synapse) (syanpse.py [lines 22-30](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/synapse.py#L22))
 6. Removes IF EXISTS statements from queries in `[generate_sql](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/synapse.py#L61)`, `[drop_dataset](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/sql_client.py#L94)` and `[drop_views](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/sql_client.py#L105)` functions (not supported in Synapse)
 7. Adds LongAsMax to connection string to update ntext data types to nvarchar (configuration.py [line 68](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/configuration.py#L68))

### Remaining Issues

1. Azure Blob/PolyBase Storage - We will need to implement method that lands data into Azure Blob Storage and utilizes PolyBase to load the data into SQL pools: https://learn.microsoft.com/en-us/azure/synapse-analytics/sql/load-data-overview
 
2. Add a handling to insert multiple values for several columns in separate INSERT INTO queries. Synapse does not support multi-row inserts using a single `INSERT INTO` statement (`Incorrect syntax near ','.` error in tests)

Ex. - The following query works in Synapse:

```
INSERT INTO [test_d4f1e1df62243254a890efb2c7ee82af].[event_weird_rows_4b1d585e2b863eea6d84b2c57961e845] ([idx], [str])
VALUES (1, N', NULL''); DROP SCHEMA Public --')

INSERT INTO [test_d4f1e1df62243254a890efb2c7ee82af].[event_weird_rows_4b1d585e2b863eea6d84b2c57961e845] ([idx], [str])
VALUES (2, N'イロハニホヘト チリヌルヲ ''ワカヨタレソ ツネナラム')
```

However, the below query does not work in Synapse:

```
INSERT INTO [test_d4f1e1df62243254a890efb2c7ee82af].[event_weird_rows_4b1d585e2b863eea6d84b2c57961e845] ([idx], [str])
VALUES
(1, N', NULL''); DROP SCHEMA Public --'),
(2, N'イロハニホヘト チリヌルヲ ''ワカヨタレソ ツネナラム')
```
Both perform the same action, but Synapse can only handle this as separate INSERT INTO statements

3. Resolve error in test_drop_tables: `Cannot drop the table '[TABLE]', because it does not exist or you do not have permission` (Table isn't generated with `CREATE TABLE` query)

4. Resolve error in tests : `Invalid object name '[TABLE]' (Table _is_ generated by `CREATE TABLE` query but isn't recognized)


dlt/destinations/synapse/synapse.py
```diff
@@ -0,0 +1,142 @@
...
+        for table in table_chain:
+            with sql_client.with_staging_dataset(staging=True):
+                staging_table_name = sql_client.make_qualified_table_name(table["name"])
+            table_name = sql_client.make_qualified_table_name(table["name"])
+            # drop destination table
+            sql.append(f"DROP TABLE {table_name};")
+            # moving staging table to destination schema
+            sql.append(f"ALTER SCHEMA {sql_client.fully_qualified_dataset_name()} TRANSFER {staging_table_name};")
+            # recreate staging table
+            sql.append(f"SELECT * INTO {staging_table_name} FROM {table_name} WHERE 1 = 0;")
```
sh-rp, line 65: does this really only create the staging table schema, or does it also copy data? It should only do the former and not copy any data over.

dlt/destinations/synapse/synapse.py
```diff
@@ -0,0 +1,142 @@
...
+            # MS SQL doesn't support alias in DELETE FROM
+            return [f"FROM {root_table_name} WHERE EXISTS (SELECT 1 FROM {staging_root_table_name} WHERE {' OR '.join([c.format(d=root_table_name,s=staging_root_table_name) for c in key_clauses])})"]
+        return SqlMergeJob.gen_key_table_clauses(root_table_name, staging_root_table_name, key_clauses, for_delete)
+
+    @classmethod
+    def _to_temp_table(cls, select_sql: str, temp_table_name: str) -> str:
+        return f"SELECT * INTO {temp_table_name} FROM ({select_sql}) as t;"
+
+    @classmethod
+    def _new_temp_table_name(cls, name_prefix: str) -> str:
```
sh-rp, line 84: shouldn't there be a uniq id included in the name to avoid collissions? See the base SqlMergeJob Class

dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,164 @@
...
+        curr = self._conn.cursor()
+        try:
+            print("Executing query:")
+            print(query)
+            # unpack because empty tuple gets interpreted as a single argument
+            # https://github.com/mkleehammer/pyodbc/wiki/Features-beyond-the-DB-API#passing-parameters
+            curr.execute(query, *args)
+            yield DBApiCursorImpl(curr)  # type: ignore[abstract]
+        except pyodbc.Error as outer:
+            print(f"Error in query:")
```
sh-rp, line 135: debug code

dlt/destinations/insert_job_client.py
```diff
@@ -72,6 +72,33 @@ def _insert(self, qualified_table_name: str, file_path: str) -> Iterator[List[st
                         else:
                             # Replace the , with ;
                             insert_sql.append("".join(chunk).strip()[:-1] + ";\n")
+                elif max_rows >= 1000:
+                    # This part breaks the multiple values in an insert statement into individual insert statements
```
sh-rp, line 76: I'm assuming you got this from here: https://stackoverflow.com/questions/36141006/how-to-insert-multiple-rows-into-sql-server-parallel-data-warehouse-table. If that is the case, it is always nice to leave a stackoverflow link in the comments to other devs can understand what is going on here.

sh-rp: Also, why is this only needed for max_rows >= 1000? Shouldn't this always fail if we don't change to the union all approach?

dlt/destinations/insert_job_client.py
```diff
@@ -72,6 +72,33 @@ def _insert(self, qualified_table_name: str, file_path: str) -> Iterator[List[st
                         else:
                             # Replace the , with ;
                             insert_sql.append("".join(chunk).strip()[:-1] + ";\n")
+                elif max_rows >= 1000:
```
sh-rp, line 75: i think you need to also test for max_rows is not None here, otherwise the gte operator will fail on none.

dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,164 @@
...
+    @raise_database_error
+    def rollback_transaction(self) -> None:
+        self._conn.rollback()
+        self._conn.autocommit = True
+
+    @property
+    def native_connection(self) -> pyodbc.Connection:
+        return self._conn
+
+    def drop_dataset(self) -> None:
```
sh-rp, line 79: doesn't the code from the baseclass work here?

sh-rp:
Thanks for the work so far. I have added a few comments, the biggest thing is, that we really need to support loading data from a blob storage (azure blob storage to start with). This is already implemented in a few other destinations, for example in redshift. The copy into command for this is specified here: https://learn.microsoft.com/en-us/sql/t-sql/statements/copy-into-transact-sql?view=azure-sqldw-latest

dlt/destinations/synapse/synapse.py
NeilRM: `_new_temp_table_name` is calling the base class function in the line `name = SqlMergeJob._new_temp_table_name(name_prefix)`, so unique_ids are being used. 

Should we instead just be calling the base function directly?

dlt/destinations/synapse/sql_client.py
NeilRM: This and all other debug code will be removed once the rest of the PR is set / tests are resolved

Issue #652, opened 2023-09-24
Bug when extracting to table named "state"


jkleinkauff:
### dlt version

0.3.17

### Describe the problem

When reading a csv file and trying to extract to a table named "state", dlt will thrown an error.

``` python
import dlt
import csv

def read_csv_file(file_path):
    with open(file_path, "r") as csvfile:
        reader = csv.reader(csvfile)
        header = next(reader)  # Read the header row

        for row in reader:
            yield dict(zip(header, row))

pipeline = dlt.pipeline(destination="filesystem", dataset_name="rawdata")

run_info = pipeline.run(
    read_csv_file("anycsv.csv"),
    table_name="state",
    write_disposition="replace",
)
```

> property 'state' of 'DltSource' object has no setter

If you change table_name to anything besides "state" the code will work.
The problem is not related to destination filesystem. Originally, I got when extracting to postgres as
well.

### Expected behavior

I expect the table to be extracted regardless it's name.

### Steps to reproduce

Run any pipeline using table_name = "state".

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

CSV

### dlt destination

Filesystem & buckets, Postgres

burnash:
@jkleinkauff thank you again for reporting this.

PR #657, opened 2023-09-27
Don't set source.<resource_name> attributes

steinitzu:
### Description

Fixes the bug when resource names conflict with other attributes on `DltSource`.  I just deleted the `setattr` line, `__getattr__` will already pull from the resources dict any attrs that don't exist on the object.

### Related Issues
Fixes https://github.com/dlt-hub/dlt/issues/652


rudolfix (approved):
OK. I can't remember why this line was there

PR #659, opened 2023-09-28
athena iceberg

sh-rp:
### Description
Implements iceberg tables on Athena.


dlt/destinations/job_client_impl.py
```diff
@@ -431,10 +433,17 @@ def _commit_schema_update(self, schema: Schema, schema_str: str) -> None:
...
 
 class SqlJobClientWithStaging(SqlJobClientBase, WithStagingDataset):
+
+    in_staging_mode: bool = False
+
     @contextlib.contextmanager
     def with_staging_dataset(self)-> Iterator["SqlJobClientBase"]:
-        with self.sql_client.with_staging_dataset(True):
-            yield self
+        try:
```
sh-rp, line 441: @rudolfix in athena client i need to know wether it is being run in staging mode. This breaks this nice whole idea of the client being (mostly) agnostic of the underlaying dataset, but I'm not sure how to do it otherwise.

dlt/destinations/sql_jobs.py
```diff
@@ -215,11 +223,17 @@ def gen_merge_sql(cls, table_chain: Sequence[TTableSchema], sql_client: SqlClien
...
-                    insert_sql += f" WHERE {uniq_column} IN (SELECT * FROM {insert_temp_table_sql});"
+                    insert_sql += f" WHERE {uniq_column} IN (SELECT * FROM {insert_temp_table_name});"
 
             if insert_sql.strip()[-1] != ";":
                 insert_sql += ";"
             sql.append(insert_sql)
             # -- DELETE FROM {staging_table_name} WHERE 1=1;
 
+        # clean up
+        if insert_temp_table_name:
```
sh-rp, line 234: iceberg does not support temp tables

dlt/destinations/sql_jobs.py
```diff
@@ -93,17 +100,17 @@ def gen_key_table_clauses(cls, root_table_name: str, staging_root_table_name: st
 
             A list of clauses may be returned for engines that do not support OR in subqueries. Like BigQuery
         """
-        return [f"FROM {root_table_name} as d WHERE EXISTS (SELECT 1 FROM {staging_root_table_name} as s WHERE {' OR '.join([c.format(d='d',s='s') for c in key_clauses])})"]
+        return [f"FROM {root_table_name} WHERE EXISTS (SELECT 1 FROM {staging_root_table_name} as s WHERE {' OR '.join([c.format(d=root_table_name,s='s') for c in key_clauses])})"]
```
sh-rp, line 103: icerberg does not support table aliases on the delete statement. I tried a few suggestions I found on stackoverflow (for other sql dialects), but none worked.

rudolfix (changes requested):
There's a lot of good stuff (ie. how we create followup jobs is generalized, some method renames connected to that 👍 )

regarding the merge: if we are getting into something really complicated let's drop it from this ticket (or altogether). I had no idea that there are no temp tables on Iceberg. we may need a fully specialized merge job for it

some other things look hacky to me ie. that we make Athena connector aware of working on staging dataset, cutting staging from the name etc.

maybe we should do something different. what about this:
- iceberg tables must be marked as such in table schema (ie. table_format hint) so all the components in dlt know and can react to that. btw. typically only 1-2 tables should be iceberg, not all (ie those that have GDPR data)
- having that we change how `filesystem` works. if it sees `iceberg` table, it always does replace on such table so we keep only the most recent load id (`filesystem` knows if it is used as staging so it can do that selectively)
- the athena destination creates all iceberg tables in staging dataset and generates copy/merge jobs to move data to iceberg tables. it has only the new data in the stage (due to replace on iceberg tables). we can also use `load_id` to selectively copy data but then the staging dataset will grow all the time

this will require a refactor but I think we'll get a simpler cleaner code

maybe if we make
```
def get_load_table(schema: Schema, file_name: str) -> TTableSchema:
```
a method of the JobClient then the job client will produce a transient table schema with write dispotion and table format properly set and possibly other settings

also if we replace
```
job_client.get_stage_dispositions():
```
with a more selective method - if it gets table schema it decides that it goes to staging dataset or not. then we may just move this logic out of load.py altogehter

dlt/destinations/athena/athena.py
```diff
@@ -309,15 +320,18 @@ def _get_table_update_sql(self, table_name: str, new_columns: Sequence[TColumnSc
 
         bucket = self.config.staging_config.bucket_url
         dataset = self.sql_client.dataset_name
+
         sql: List[str] = []
 
         # for the system tables we need to create empty iceberg tables to be able to run, DELETE and UPDATE queries
-        is_iceberg = self.schema.tables[table_name].get("write_disposition", None) == "skip"
+        # or if we are in iceberg mode, we create iceberg tables for all tables
+        is_iceberg = (self.schema.tables[table_name].get("write_disposition", None) == "skip") or (self._is_iceberg_table(self.schema.tables[table_name]) and not self.in_staging_mode)
```
sh-rp, line 328: this "in staging mode" is still here unfortunately.. i don't know how to do it any other way.

dlt/common/schema/utils.py
```diff
@@ -525,6 +536,18 @@ def get_top_level_table(tables: TSchemaTables, table_name: str) -> TTableSchema:
         return get_top_level_table(tables, parent)
     return table
 
+def get_load_table(tables: TSchemaTables, table_name: str) -> TTableSchema:
```
rudolfix, line 539: this should be IMO part of JobClient

dlt/destinations/athena/athena.py
```diff
@@ -69,13 +69,18 @@ class AthenaTypeMapper(TypeMapper):
...
+        super().__init__(capabilities)
+        self.iceberg_mode = iceberg_mode
+
     def to_db_integer_type(self, precision: Optional[int]) -> str:
         if precision is None:
             return "bigint"
+        # iceberg does not support smallint and tinyint
         if precision <= 8:
-            return "tinyint"
+            return "int" if self.iceberg_mode else "tinyint"
```
rudolfix, line 81: that's why JobClient should create/modify table schema. so you can modify precision there and do not hack the type mapper...

dlt/destinations/athena/athena.py
rudolfix: if you move the `get_load_table` to JobClient you can modify table format to keep iceberg (destination dataset) or not (staging dataset). OFC you need to pass is_staging flag to the method. I do not like it but probably it is a good compromise

dlt/destinations/filesystem/filesystem.py
```diff
@@ -82,16 +83,22 @@ def __init__(self, schema: Schema, config: FilesystemDestinationClientConfigurat
...
-        ds_path = posixpath.join(self.fs_path, self.config.normalize_dataset_name(self.schema))
-        return ds_path
+        self.dataset_path = posixpath.join(self.fs_path, self.config.normalize_dataset_name(self.schema))
 
     def drop_storage(self) -> None:
         if self.is_storage_initialized():
             self.fs_client.rm(self.dataset_path, recursive=True)
 
+    @contextmanager
+    def with_staging_dataset(self) -> Iterator["FilesystemClient"]:
```
rudolfix, line 93: why we need this in FileSystemClient?

dlt/destinations/filesystem/filesystem.py
```diff
@@ -169,3 +176,9 @@ def __enter__(self) -> "FilesystemClient":
 
     def __exit__(self, exc_type: Type[BaseException], exc_val: BaseException, exc_tb: TracebackType) -> None:
         pass
+
+    def table_needs_staging(self, table: TTableSchema) -> bool:
+        # not so nice, how to do it better, collect this info from the main destination as before?
+        if table["table_format"] == "iceberg":
```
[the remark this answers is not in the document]
rudolfix: I'd rather say that this table needs truncation!  We do not need to keep it in staging dataset, we just need to truncate it.
To me it makes sense... if you have iceberg for sure you need to copy/move data and keeping full history does not makes sense...

dlt/destinations/job_client_impl.py
```diff
@@ -431,15 +428,22 @@ def _commit_schema_update(self, schema: Schema, schema_str: str) -> None:
 
 
 class SqlJobClientWithStaging(SqlJobClientBase, WithStagingDataset):
+
+    in_staging_mode: bool = False
```
rudolfix, line 432: IMO we'll not need it

dlt/destinations/job_client_impl.py
```diff
@@ -431,15 +428,22 @@ def _commit_schema_update(self, schema: Schema, schema_str: str) -> None:
...
-        with self.sql_client.with_staging_dataset(True):
-            yield self
+        try:
+            with self.sql_client.with_staging_dataset(True):
+                self.in_staging_mode = True
+                yield self
+        finally:
+            self.in_staging_mode = False
+
+    def table_needs_staging(self, table: TTableSchema) -> bool:
```
rudolfix, line 443: 👍

dlt/common/schema/utils.py
```diff
@@ -525,6 +536,18 @@ def get_top_level_table(tables: TSchemaTables, table_name: str) -> TTableSchema:
...
+    try:
+        # make a copy of the schema so modifications do not affect the original document
+        table = copy(tables[table_name])
```
rudolfix, line 542: we'll need a deepcopy because we will modify columns

dlt/destinations/filesystem/filesystem.py
sh-rp: because we need a location where the "normal" athena table file live before they are copied into iceberg. The folder in the "final destination" for this table will be fully managed by iceberg, and we do not manually manipulated the files there.

dlt/destinations/job_client_impl.py
sh-rp: the problem is, that in the place where I need this information (building the tables), the table info is being taken from the schema and not passed into the function. I could pass down this information via update_stored_schema, but this will not be much nicer than this hook i'm afraid..

dlt/destinations/athena/athena.py
sh-rp: Wouldn't it be the cleanest to have a subclass for iceberg and then set that before the table sql is generated? I don't feel like changing the type mapper is hacking at all, that is what it is there for, changing the mapping of the types depending on database / table format you are storing into.

sh-rp: Actually we could extend the type mapper to have the info which table_format is currently being processed. that might be nice?

dlt/destinations/athena/athena.py
```diff
@@ -293,11 +303,12 @@ def __init__(self, schema: Schema, config: AthenaClientConfiguration) -> None:
         super().__init__(schema, config, sql_client)
         self.sql_client: AthenaSQLClient = sql_client  # type: ignore
         self.config: AthenaClientConfiguration = config
-        self.type_mapper = AthenaTypeMapper(self.capabilities)
+        self.type_mapper = AthenaTypeMapper(self.capabilities, True)
```
rudolfix, line 306: some tables are iceberg, some not. you can't set always True here, see below

dlt/destinations/athena/athena.py
rudolfix: as discussed: use get_load_table() to
- adjust the precision
- set the table format
and pass it to the code below

dlt/destinations/athena/athena.py
```diff
@@ -309,15 +320,18 @@ def _get_table_update_sql(self, table_name: str, new_columns: Sequence[TColumnSc
...
         # for the system tables we need to create empty iceberg tables to be able to run, DELETE and UPDATE queries
-        is_iceberg = self.schema.tables[table_name].get("write_disposition", None) == "skip"
+        # or if we are in iceberg mode, we create iceberg tables for all tables
+        is_iceberg = (self.schema.tables[table_name].get("write_disposition", None) == "skip") or (self._is_iceberg_table(self.schema.tables[table_name]) and not self.in_staging_mode)
         columns = ", ".join([self._get_column_def_sql(c) for c in new_columns])
 
         # this will fail if the table prefix is not properly defined
         table_prefix = self.table_prefix_layout.format(table_name=table_name)
         location = f"{bucket}/{dataset}/{table_prefix}"
+
```
rudolfix, line 334: one thing (maybe helpful) when we create parquet file we are setting precision of various fields ie datetime. make sure we do not have problems here (probably you should take into account both capabilities and table format but I think our current implementation is good - I hope it works)

dlt/destinations/athena/athena.py
```diff
@@ -345,9 +359,35 @@ def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) ->
...
+    
+    def table_needs_staging_dataset(self, table: TTableSchema) -> bool:
+        # all iceberg tables need staging
+        if self._is_iceberg_table(table):
+            return True
+        return super().table_needs_staging_dataset(table)
+    
+    def get_load_table(self, table_name: str, staging: bool = False) -> TTableSchema:
+        table = super().get_load_table(table_name, staging)
+        # if staging and table.get("table_format", None) == "iceberg":
```
rudolfix, line 398: heh I think you need to uncomment that and also modify the precision

dlt/destinations/athena/configuration.py
```diff
@@ -12,6 +12,8 @@ class AthenaClientConfiguration(DestinationClientDwhWithStagingConfiguration):
     credentials: AwsCredentials = None
     athena_work_group: Optional[str] = None
     aws_data_catalog: Optional[str] = "awsdatacatalog"
+    supports_truncate_command: bool = False
+    force_iceberg: Optional[bool] = False
```
rudolfix, line 16: is this used just for testing? I do not see other reason. you can just set all table formats to iceberg to force it

dlt/destinations/filesystem/configuration.py
```diff
@@ -10,6 +10,7 @@
 @configspec
 class FilesystemDestinationClientConfiguration(FilesystemConfiguration, DestinationClientStagingConfiguration): # type: ignore[misc]
     destination_name: Final[str] = "filesystem"  # type: ignore
+    force_iceberg: Optional[bool] = False
```
rudolfix, line 13: same

dlt/destinations/filesystem/filesystem.py
```diff
@@ -82,16 +83,22 @@ def __init__(self, schema: Schema, config: FilesystemDestinationClientConfigurat
...
 
     def drop_storage(self) -> None:
         if self.is_storage_initialized():
             self.fs_client.rm(self.dataset_path, recursive=True)
 
+    @contextmanager
+    def with_staging_dataset(self) -> Iterator["FilesystemClient"]:
+        current_dataset_path = self.dataset_path
+        try:
+            self.dataset_path = posixpath.join(self.fs_path, self.config.normalize_dataset_name(self.schema)) + "_staging"
```
rudolfix, line 96: `self.config.normalize_dataset_name(SqlClientBase.make_staging_dataset_name(self.schema))`

dlt/destinations/sql_jobs.py
```diff
@@ -48,14 +54,15 @@ class SqlStagingCopyJob(SqlBaseJob):
...
-    def generate_sql(cls, table_chain: Sequence[TTableSchema], sql_client: SqlClientBase[Any]) -> List[str]:
+    def generate_sql(cls, table_chain: Sequence[TTableSchema], sql_client: SqlClientBase[Any], params: Optional[SqlJobParams] = None) -> List[str]:
         sql: List[str] = []
         for table in table_chain:
             with sql_client.with_staging_dataset(staging=True):
                 staging_table_name = sql_client.make_qualified_table_name(table["name"])
             table_name = sql_client.make_qualified_table_name(table["name"])
             columns = ", ".join(map(sql_client.capabilities.escape_identifier, get_columns_names_with_prop(table, "name")))
-            sql.append(sql_client._truncate_table_sql(table_name))
+            if params["replace"]:
```
rudolfix, line 64: do we still use it?

rudolfix (changes requested):

dlt/destinations/sql_jobs.py
sh-rp: yeah

dlt/destinations/athena/configuration.py
sh-rp: how would I do this? i want to be able to set an env var and then have our standard tests all run against athena iceberg

dlt/destinations/filesystem/configuration.py
sh-rp: removed it here as it is no longer needed

dlt/destinations/athena/configuration.py
rudolfix: OK! then keep it. I think other methods would be too complicated (ie patching some functions)

dlt/destinations/filesystem/filesystem.py
sh-rp: are you sure you want to use a method from the sqlclientbase in the filesystem destination? this does not seem quite right..

rudolfix: so we need to move it to client base? or to `WithStagingDataset`?

dlt/common/destination/reference.py
```diff
@@ -309,15 +324,23 @@ class WithStagingDataset(ABC):
...
-        return []
+    def should_load_data_to_staging_dataset(self, table: TTableSchema) -> bool:
+        return False
 
     @abstractmethod
     def with_staging_dataset(self)-> ContextManager["JobClientBase"]:
         """Executes job client methods on staging dataset"""
         return self  # type: ignore
 
+class SupportsStagingDestination():
```
sh-rp, line 335: With this the main destination can control the behavior of the staging destination, this was kind of undefined before, so we were lucky that athena worked properly, but now it is well defined and we will also need this on other datalakes.

dlt/destinations/type_mapping.py
```diff
@@ -20,15 +20,15 @@ class TypeMapper:
     def __init__(self, capabilities: DestinationCapabilitiesContext) -> None:
         self.capabilities = capabilities
 
-    def to_db_integer_type(self, precision: Optional[int]) -> str:
+    def to_db_integer_type(self, precision: Optional[int], table_format: TTableFormat = None) -> str:
         # Override in subclass if db supports other integer types (e.g. smallint, integer, tinyint, etc.)
         return self.sct_to_unbound_dbt["bigint"]
 
-    def to_db_type(self, column: TColumnSchema) -> str:
+    def to_db_type(self, column: TColumnSchema, table_format: TTableFormat = None) -> str:
```
sh-rp, line 27: to me manipulating the load table to achieve the correct mapping feels hacky. I would say that the type mapper should be in charge of mapping schema types to db types and in the case of athena (and maybe others in the future) it has to take the chosen table format into account.

sh-rp:
Notes:
* i have disabled the dbt tests for iceberg athena for now. On the dbt athena community plugin page it says iceberg should be supported with some caveats, so this would need more work to be resolved.
* I think it should be fairly easy to get merge to work, but we can do that in the next step
* We should save the iceberg table_format into the schema when the env var is set, but I am not sure where in the code this should be done so that this data is persisted. Do we have any destination that changes the schema in our code and does this concept make sense?

rudolfix (changes requested):
this is very good now!
- see one comment regarding `"_staging" (we must normalize the whole name)
- dbt problems: it is clear from the error that we are copying timestamps from iceberg table (6) to regular table (precision 3). and this fails
```
10:05:28  NOT_SUPPORTED: Incorrect timestamp precision for timestamp(6); the configured precision is MILLISECONDS; column name: end_time. You may need to manually clean the data at location 's3://dlt-athena-output/tables/850461bf-aa78-4d8c-8be7-36c1630da099' before retrying. Athena will not delete data in your account.
```
my take: disable dbt chess tests on athena + iceberg and leave just jaffle shop - just mention the error in docs maybe
or you mess around with `table_type='iceberg',` in the materialization in chess

rudolfix (approved):
LGTM!

PR #661, opened 2023-09-28
Typing fixes and enable mypy in tests

steinitzu:
### Description

This does a few things to make mypy useful in test modules especially:

* Changes mypy config -> disable `ignore_missing_imports` globally and enable explicitly for untyped packages
* Runs mypy on `tests/` (with relaxed config) as part of lint stage
* Fixes/ignores all type errors in tests. Not perfectly, lot of ignores, but goes a long way


rudolfix (approved):
🤯 this is good!

Issue #634, opened 2023-09-13
Direct loading of structured data (arrow Tables) into destination


rudolfix:
**Background**
As loading of already structured data comes quite often let's build a PoC and required support for it into library core. In essence for formats/objects like
- arrow tables / parquet files / duckdb (via arrow)
- panda frames (also via pyarrow)

we would:
1. quickly push them via extract stage (with zero copy if possible - see below)
2. extract the schema
3. materialize as parquet (any better ideas for the storage) instead of jsonl
4. quickly move into load package in normalize stage

The PoC would be a demonstration of:
- a Connector-X source producing arrow tables that get almost instantly loaded into destination

**PoC Limitations**
- only destinations supported parquet will work
- dlt will not support all parquet features (structured fields, we should also skip partitions in PoC until we have them in the core) when generating schemas
- an instance of arrow tables will be passed through pipeline so several features will stop working or require decomposing into rows

1. map, filter and yield_map would need to feed with row by row version
2. incremental - is just a filter

that should be ok for PoC - we can easily implement `incremental` version that does `max` operation on the whole arrow table etc. tbd. later

**Requirements**
1. * [x] Write simple Connector-X source (already done)
2. * [x] Convert pyarrow schemas into dlt schemas (easy until we have structured fields - let's just mark them as complex and pass along)
3. * [x] implement a new writer ie. ArrowWriter that will write pyarrow Table instances to a open file
4. * [x] extend the `extract()` in `extract.py` so it takes `jsonl` and `arrow` storages which will be used depending on the type of item being written
5. * [x] before item is written schema needs to be extraced from table and updated (like we do for dynamic tables already)
6. * [x] we extend normalizer that whenever it sees `parquet` file extension it just moves it into load package, assuming that schema is inferred already
7. * [ ] optional: we decompose tables into rows in dlt map, filter and yield_map

**Release**
1. We'll release an alpha/preview from the branch for testing. 
2. The Connector-X may go intially into our new Examples category in `docs`

z3z1ma:
This will be so cool.

tomsej:
Thanks! I have already started to work on that!

PR #662, opened 2023-09-29
[WIP] POC - direct PyArrow load

tomsej:
### Description

See #634

It is just a very rough example, that kinda works. Currently it does not support precision and scale. I am also not sure about how to properly do a normalization. It just copy the parquet file to the new location. It also lacks tests.

Sorry had not much time lately :(

### DIFF
--- a/dlt/common/data_writers/writers.py
+++ b/dlt/common/data_writers/writers.py
@@ -1,14 +1,16 @@
 import abc
-
 from dataclasses import dataclass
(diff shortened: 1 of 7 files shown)


rudolfix:
@tomsej your PR moved to #679 and it looks good! we'll finalize implementation there and if you are interested in contribution - an example with connector_x would be nice - we have a new top level category for Code Examples in our docs. if you are interested ping me on slack

PR #665, opened 2023-10-04
Blog data product docs

adrianbr:


docs/website/blog/2023-10-03-data-product-docs.md
hibajamal, line 1: Reading this yesterday, it didn't strike to me how much this is needed in the last company I worked for, because when I read this yesterday, data product to me read: marketable data product. Even though you didn't explicitly say that here, it's where my mind went. But a data product can also be an internal tool, as you mentioned today in the demo, data coming from marketing can't be read by product/tech etc. And it usually is an internal tool in many companies (we just forget that it is), so maybe in the examples, that aspect could also be there.

adrianbr: thanks, will add

PR #672, opened 2023-10-05
Refactor documentation structure

burnash:
### Description

This pull request proposes a restructuring of our existing documentation.

### Rationale
After extensively reviewing the current documentation, we identified opportunities for enhancing clarity. By adopting the [Documentation System](https://documentation.divio.com/), we aim to provide a more navigable and coherent documentation.


burnash:
@rudolfix 
> 1. do we want to keep files in old folders that do not correspond to index locations? we probably preserve some links but make a debt in the future

Nope, we don't want to keep files in the old folders. As a followup PR I'll move the files to new, proper folders and add 301 redirects on the server side with netlify. I didn't want to move anything until there's consensus on the new structure.

> 2. as the links are old but placed differently in the index: I think all the reports in GA4 will be messed up (no category information really?). @adrianbr we really need your input here

With this PR there will be no change to reports in GA4 since I index (sidebar) does not affect how individual pages are displayed in GA (If I understand correctly what you mean). After we merge the follow-up PR with redirect, yes, the links will be different.

> 3. we have a few mentions of "dlt-ecosystem" in our codebase. you need to also fix that.
best if you look for https://dlthub.com in all files and see which are affected - especially if we change the file placements

This is going to be addressed with 301 redirects as well. After which we can update the hrefs to make the graceful transition.

rudolfix:
@burnash that's clear now! but I'd prefer that we merge with redirect. Let's try to get a go from the team today and do the last mile in this PR.

From my side I'm 100% a GO :)

rudolfix (approved):

PR #677, opened 2023-10-06
[WIP] Synapse Destination

NeilRM:
### Description

Adds the Synapse destination to the dlt destinations list.

 This destination code is based off of the mssql destination, and makes the following changes:
 
 1. Updates "mssql" references to "synapse" in destination code
 2. Adds Synapse [Github Testing Workflow](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/.github/workflows/test_destination_synapse.yml)
 3. Adds synapse_escape_literal function in escape.py ([lines 91-115](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/common/data_writers/escape.py#L91))
 4. Updates max_text_data_type_length to 4000 (max character length for Synapse) ([init.py line 39](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/__init__.py#L39))
 5. Changes nvarchar(max) to nvarchar(4000) and varbinary(max) to varbinary(8000) in type mapping. ("max" length not supported in Synapse) (synapse.py [lines 22-30](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/synapse.py#L22))
 6. Removes IF EXISTS statements from queries in [generate_sql](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/synapse.py#L61), [drop_dataset](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/sql_client.py#L94) and [drop_views](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/sql_client.py#L105) functions (not supported in Synapse)
 7. Adds LongAsMax to connection string to update ntext data types to nvarchar (configuration.py [line 68](https://github.com/ryan-miranda-partners/dlt/blob/9fff053a3dff2ac55795a4967700c4d895038cbd/dlt/destinations/synapse/configuration.py#L68))
 8. Adds handling to insert multiple values for several columns in separate INSERT INTO queries using UNION ALL. (Synapse does not support multi-row inserts using a single `INSERT INTO` statement) (Insert_job_client.py lines [62-104](https://github.com/ryan-miranda-partners/dlt/blob/4a9fbd631f327dfbd73635de11c66ca678a1bae2/dlt/destinations/insert_job_client.py#L62))

### Remaining Issues

1. Azure Blob Storage - We will need to implement method that lands data into Azure Blob Storage and utilizes COPY command to load the data into SQL pools: https://learn.microsoft.com/en-us/azure/synapse-analytics/sql/load-data-overview
 
2. Resolve error in tests: `String or binary data would be truncated` for INSERT INTO queries that are inserting values over the max length of 4000

3. Resolve error in tests : `Invalid object name '[TABLE]' (Table _is_ generated by `CREATE TABLE` query but isn't recognized)

4. Resolve Error in `test_data_writer_string_escape` test: `Operation cannot be performed within a transaction`

5. Resolve Error in `test_execute_df ` test: `Parse error at line: 1, column: 102: Incorrect syntax near ','`


dlt/destinations/insert_job_client.py
```diff
@@ -10,6 +10,7 @@
 from dlt.destinations.sql_client import SqlClientBase
 from dlt.destinations.job_impl import EmptyLoadJob
 from dlt.destinations.job_client_impl import SqlJobClientWithStaging
+from tests.utils import ACTIVE_DESTINATIONS
```
sh-rp, line 13: you can't imort test code into production code...

dlt/destinations/insert_job_client.py
```diff
@@ -58,20 +59,49 @@ def _insert(self, qualified_table_name: str, file_path: str) -> Iterator[List[st
...
-                    for chunk in chunks(values_rows, max_rows - 1):
-                        processed += len(chunk)
-                        insert_sql.extend([header.format(qualified_table_name), values_mark])
-                        if processed == len_rows:
-                            # On the last chunk we need to add the extra row read
-                            insert_sql.append("".join(chunk) + until_nl)
-                        else:
-                            # Replace the , with ;
-                            insert_sql.append("".join(chunk).strip()[:-1] + ";\n")
+                    if "synapse" in ACTIVE_DESTINATIONS:
```
sh-rp, line 62: if you need a different behavior here for synapse, you need to extend or replace this load job for synapse or somehow manage this with the capabilities object. this code is depended on some var that is only available in the test environment and will not work in production

dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,222 @@
...
+        WHERE TABLE_SCHEMA = ? AND TABLE_NAME = ?
+        """
+        result = self.execute_sql(query, self.dataset_name, table_name)
+        return result[0][0] > 0        
+
+    def drop_tables(self, *tables: str) -> None:
+        if not tables:
+            return
+        # If a transaction is active, commit it
+        if not self._conn.autocommit:
```
sh-rp, line 118: i don't think you need this kind of code here. In which case is a transaction avtive when you drop the dataset?

dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,222 @@
...
+                if self.table_exists(table):
+                    self.execute_sql(f"DROP TABLE {self.make_qualified_table_name(table)};")
+        finally:
+            # Reset autocommit to its previous state
+            self._conn.autocommit = prev_autocommit
+
+    def _drop_views(self, *tables: str) -> None:
+        # Save the current autocommit state
+        # If a transaction is active, commit it
+        if not self._conn.autocommit:
```
sh-rp, line 136: same as above

dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,222 @@
...
+    @raise_database_error
+    def rollback_transaction(self) -> None:
+        self._conn.rollback()
+        self._conn.autocommit = True
+
+    @property
+    def native_connection(self) -> pyodbc.Connection:
+        return self._conn
+
+    def drop_dataset(self) -> None:
```
sh-rp, line 78: why do you need to check if each table exists? can't you just drop all tables in the dataset like it is done in the mssql client?

PR #679, opened 2023-10-09
Pyarrow direct loading

steinitzu:
Continuation  of https://github.com/dlt-hub/dlt/pull/662


rudolfix:
@sh-rp please take a look
@steinitzu it looks good. code structure is OK. what we still need
- a pipeline test where we load arrow tables and panda frames to all destinations that support parquet
- a documentation: let's add it to "Verified Sources" under Arrow Table / Pandas
- I will still do the review

dlt/common/storages/normalize_storage.py
```diff
@@ -47,10 +47,17 @@ def build_extracted_file_stem(schema_name: str, table_name: str, file_id: str) -
...
 
-        parts = Path(file_name).stem.split(".")
+        parts = file_name_p.stem.split(".")
         if len(parts) != 3:
             raise ValueError(file_name)
         return TParsedNormalizeFileName(*parts)
+
+    def delete_extracted_files(self, files: Sequence[str]) -> None:
+        for file_name in files:
+            if Path(file_name).suffix == ".parquet":  # Parquet extract files are moved, not copied
```
rudolfix, line 61: we do not copy those files back into old location right? so we'll have a partially deleted original folder when something fails.

on next attempt files will not be processed again

what we should IMO do is not move but use FileStorage.link_hard to create hard links from the original into temp folder. 

in that case you just delete the original folder like any other.

the only problem I see is when normalize and load storages are on separate filesystem. but in that case rename also falls back to copy so you can use link_hard with fallback to copy. I think it is good compromise

dlt/extract/extract.py
```diff
@@ -48,12 +99,115 @@ def commit_extract_files(self, extract_id: str, with_delete: bool = True) -> Non
...
-    def _get_data_item_path_template(self, load_id: str, schema_name: str, table_name: str) -> str:
-        template = NormalizeStorage.build_extracted_file_stem(schema_name, table_name, "%s")
-        return self.storage.make_full_path(os.path.join(self._get_extract_path(load_id), template))
+    def write_data_item(self, file_format: TLoaderFileFormat, load_id: str, schema_name: str, table_name: str, item: TDataItems, columns: TTableSchemaColumns) -> None:
+        self.get_storage(file_format).write_data_item(load_id, schema_name, table_name, item, columns)
 
-    def _get_extract_path(self, extract_id: str) -> str:
-        return os.path.join(ExtractorStorage.EXTRACT_FOLDER, extract_id)
+
+def _item_format(item: TDataItems) -> TLoaderFileFormat:
```
rudolfix, line 106: maybe a static method of Extract?

rudolfix: also `item` may be a list here (`TDataItems`) so you need to check it for being a list first.

please also rename item to items. same for `_write_static_table`

dlt/extract/extract.py
```diff
@@ -125,26 +249,9 @@ def _write_static_table(resource: DltResource, table_name: str) -> None:
                 signals.raise_if_signalled()
 
                 # TODO: many resources may be returned. if that happens the item meta must be present with table name and this name must match one of resources
-                # if meta contains table name
                 resource = source.resources.find_by_pipe(pipe_item.pipe)
```
rudolfix, line 252: please merge with current `devel` this line was changed and you have a conflict

dlt/normalize/items_normalizers.py
```diff
@@ -0,0 +1,131 @@
...
+        load_id: str,
+        root_table_name: str,
+    ) -> Tuple[List[TSchemaUpdate], int, TRowCount]:
+        from dlt.common.libs import pyarrow
+        with normalize_storage.storage.open_file(extracted_items_file, "rb") as f:
+            items_count = pyarrow.get_row_count(f)
+        target_folder = load_storage.storage.make_full_path(os.path.join(load_id, LoadStorage.NEW_JOBS_FOLDER))
+        # normalize_storage.parse_normalize_file_name
+        parts = NormalizeStorage.parse_normalize_file_name(extracted_items_file)
+        new_file_name = load_storage.build_job_file_name(parts.table_name, parts.file_id, with_extension=False) + ".parquet"
```
rudolfix, line 129: heh interesting if that is going to work :) in theory we can have some files to be jsonl/INSERT and some parquet and the destination may handle that.

IMO you should create a valid `load_storage` with a proper extension and pass it here.

dlt/normalize/normalize.py
```diff
@@ -67,13 +69,13 @@ def load_or_create_schema(schema_storage: SchemaStorage, schema_name: str) -> Sc
...
-            load_id: str,
-            extracted_items_files: Sequence[str],
-        ) -> TWorkerRV:
+        normalize_storage_config: NormalizeStorageConfiguration,
+        loader_storage_config: LoadStorageConfiguration,
+        destination_caps: DestinationCapabilitiesContext,
+        stored_schema: TStoredSchema,
+        load_id: str,
+        extracted_items_files: Sequence[str],
+    ) -> TWorkerRV:
```
rudolfix, line 78: you must create load storage with a correct file format later
```
load_storage = LoadStorage(False, destination_caps.preferred_loader_file_format, LoadStorage.ALL_SUPPORTED_FILE_FORMATS, loader_storage_config)
```
goes for puae-json files

the parquet must have a parquet extension as preferred.

destination_caps have information if parquet is supported at all. if not you must raise a really descriptive exception why we can't do parquet for given destination.

we could OFC convert parequet to py_list and stream it into the desired writer like we stream jsonl files. but I think this is not the reason to use pyarrow directy. IMO a nice exception is enough.

this case also needs to be tested

dlt/normalize/normalize.py
```diff
@@ -93,22 +95,22 @@ def w_normalize_files(
...
-                        # make sure base tables are all covered
-                        increase_row_count(row_counts, root_table_name, 0)
+
+                    extension = Path(extracted_items_file).suffix
+                    normalizer: ItemsNormalizer
+                    if extension == ".parquet":
+                        normalizer = ParquetItemsNormalizer()
+                    else:
+                        normalizer = JsonLItemsNormalizer()
+                    partial_updates, items_count, r_counts = normalizer(extracted_items_file, load_storage, normalize_storage, schema, load_id, root_table_name)
```
rudolfix, line 105: create load storage here

rudolfix (changes requested):
this is good! 
- we need tests I mentioned in review
- please also test what happens if we add incremental or any other filter/map function
- I think it is trivial to modify `ItemTransform` to deconstruct Table or Panda into rows and make incremental working. this of course will destroy the table. maybe there's a way to "rescue" filter functions? maybe you can filter arrow table by a python lambda

sh-rp:
Nice PR! The one thing that came to my mind is, that with the way it is implemented now we are losing the ability to create variant columns. This means if we try to load an arrow table that has a different datatype on a column, it will now fail when updating the schema in the extract phase but not create a variant. It seems we could solve this with rename_columns in arrow, but this would mean having some kind of variant detection step in the extract phase, or alternatively push the schema updating downstream into the normalizer phase (which might be nice anyway to keep it similar to the way it is now). Also should we give the users the option to extract complex fields in parquet into subtables?

rudolfix:
@sh-rp good points! but in most cases people want Arrow tables to have strict schema (and typically the data is already well defined)

I like the idea to extract the structs into separate tables. We could do that in parquet normalizer but OFC the tables would need to be rewritten - most probably using **duckdb** as engine. btw. that would be FAST - compare to our standard normalizer

dlt/extract/incremental/transform.py
```diff
@@ -0,0 +1,251 @@
...
+            # Exclude rows from the table which have unique hashes already seen before
+
+            # Rows with same cursor as stored last value
+            eq_rows = tbl.filter(pa.compute.equal(tbl[cursor_path], last_value))
+            # compute index, unique hash mapping
+            unique_values = self.unique_values(eq_rows, primary_key, resource_name)
+            unique_values = [(i, uq_val) for i, uq_val in unique_values if uq_val not in incremental_state['unique_hashes']]
+            keep_idx = pa.array(i for i, _ in unique_values)
+            # Filter the table
+            tbl = tbl.filter(pa.compute.is_in(tbl["_dlt_index"], keep_idx))
```
rudolfix, line 218: 1. this will keep only records with unique values and remove all the records that are "newer" than unique right? because keep_idx is only on unique values. maybe you should use `is_not_in`?
2. also adding index could be deferred to the moment we need to remove something below. (and maybe primary key could be used for that if present)

rudolfix (changes requested):
added some comments on incremental

dlt/extract/incremental/__init__.py
```diff
@@ -371,13 +312,36 @@ def bind(self, pipe: SupportsPipe) -> "Incremental[TCursorValue]":
...
+                return self._transformers['arrow']
+            return self._transformers['json']
+        return self._transformers['json']
+
+    def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
+        if rows is None:
+            return rows
+
+        transformer = self._get_transformer(rows)
+        # TODO: Primary key is not applied at bind time?
```
rudolfix, line 336: yes - in case incremental is on transformer. then we call bind before _incremental in wrapper is created (transformer is called for a first time already when executing in the pipe)

dlt/normalize/normalize.py
```diff
@@ -51,7 +53,7 @@ def __init__(self, collector: Collector = NULL_COLLECTOR, schema_storage: Schema
     def create_storages(self) -> None:
         # pass initial normalize storage config embedded in normalize config
         self.normalize_storage = NormalizeStorage(True, config=self.config._normalize_storage_config)
-        # normalize saves in preferred format but can read all supported formats
+        # # normalize saves in preferred format but can read all supported formats
```
rudolfix, line 56: typo

rudolfix (commented):
looks good! I assume that is still WIP (some tests and docs mostly)
also you detect arrow tables in lists but IMO when you write to tables in extractor you assume that items are always actual objects not lists. if I'm not right you can ignore it.

also in arrow incremental IMO you delete too many records but that will come out in the tests

amazing work!

steinitzu:
> looks good! I assume that is still WIP (some tests and docs mostly) also you detect arrow tables in lists but IMO when you write to tables in extractor you assume that items are always actual objects not lists. if I'm not right you can ignore it.
> 
> also in arrow incremental IMO you delete too many records but that will come out in the tests
> 
> amazing work!

Yep tests and docs coming, needs more thorough testing. I'm hoping I can parametrize the current incremental tests easily to run with all formats, that would be best.

> also you detect arrow tables in lists but IMO when you write to tables in extractor you assume that items are always actual objects not lists

You might be right. Needs tests with lists as well.

dlt/extract/incremental/__init__.py
steinitzu: Hmm ok, I see. Then it's only the primary key that might be deferred, right? I could also instatiate the transformers lazily in `__call__` if that's safer. What do you think?

rudolfix: there's one test that uses incremental on transformer and it will fail (if there's any problem at all). also get the current devel. I improved this behavior (a little bit) recently

rudolfix (approved):
LGTM! thx for incremental tests!
this needs to be fixed though
```
 Redshift cannot load TIME columns from parquet files. Switch to direct INSERT file format or convert `datetime.time` objects in your data to `str` or `datetime.datetime`
 ```
Also Athena has the same problem

steinitzu:
> LGTM! thx for incremental tests! this needs to be fixed though
> 
> ```
>  Redshift cannot load TIME columns from parquet files. Switch to direct INSERT file format or convert `datetime.time` objects in your data to `str` or `datetime.datetime`
> ```

Should work now! Had a typo in this check https://github.com/dlt-hub/dlt/pull/679/commits/b9f7aaf2a2c48d4c9aba63424889c76c4612bf32

rudolfix:
> > LGTM! thx for incremental tests! this needs to be fixed though
> > ```
> >  Redshift cannot load TIME columns from parquet files. Switch to direct INSERT file format or convert `datetime.time` objects in your data to `str` or `datetime.datetime`
> > ```
> 
> Should work now! Had a typo in this check [b9f7aaf](https://github.com/dlt-hub/dlt/commit/b9f7aaf2a2c48d4c9aba63424889c76c4612bf32)

@steinitzu now there's some problems with types. ie. binary values are returned as hex strings on redshift. which is the case in other tests...

rudolfix (approved):
so good!

PR #680, opened 2023-10-09
Fix: read pkey as DER format, not PEM

codingcyclist:
### Description
I had a small glitch in my previous PR: instead of expecting the base64-encoded private key for a Snowflake connection to come in DER format, I implemented the key loading function to expect a PEM-formatted key. I took a second look at [how DBT implemented their loading of base64 encoded keys](https://github.com/dbt-labs/dbt-snowflake/pull/292/files) and adjusted it accordingly.

### Additional Context

Original PR for base64-encoded private keys: https://github.com/dlt-hub/dlt/pull/637


dlt/destinations/snowflake/configuration.py
```diff
@@ -23,21 +23,24 @@ def _read_private_key(private_key: str, password: Optional[str] = None) -> bytes
         from cryptography.hazmat.primitives.asymmetric import rsa
         from cryptography.hazmat.primitives.asymmetric import dsa
         from cryptography.hazmat.primitives import serialization
+        from cryptography.hazmat.primitives.asymmetric.types import PrivateKeyTypes
     except ModuleNotFoundError as e:
         raise MissingDependencyException("SnowflakeCredentials with private key", dependencies=[f"{version.DLT_PKG_NAME}[snowflake]"]) from e
 
+    load_key_func = serialization.load_pem_private_key
```
rudolfix, line 30: IMO code is more readable if those two lines are moved to exception handler

dlt/destinations/snowflake/configuration.py
```diff
@@ -23,21 +23,24 @@ def _read_private_key(private_key: str, password: Optional[str] = None) -> bytes
...
+    private_key_bytes_decoded = private_key.encode()
     try:
         # decode base64 encoded private key
-        private_key_decoded = base64.b64decode(private_key).decode('utf-8')
+        private_key_bytes_decoded = base64.b64decode(private_key)
+        load_key_func = serialization.load_der_private_key
```
rudolfix, line 35: again the PEM key also decodes as base64 so you will assign der key func to it. see CI - it fails
what I'd do here: just load the key as DER, if that fails - load as PEM in exception handler

rudolfix (changes requested):
please look at the CI - a test is not passing. please include both DER and PEM test cases

PR #686, opened 2023-10-11
change source schema handling and normalizer root_key propagation

sh-rp:
* This PR changes the way the initial schema of a source is constructed, so normalizer settings are retained but can be updated between loads.
* The normalizer now also adds a normalizer config entry for each table that ever had a merge disposition set, so subsequent loads with other write dispositions to not fail.


dlt/common/schema/schema.py
```diff
@@ -206,6 +207,16 @@ def update_schema(self, partial_table: TPartialTableSchema) -> TPartialTableSche
             partial_table = utils.merge_tables(table, partial_table)
         return partial_table
 
+
+    def update_schema(self, schema: "Schema") -> None:
+        """Updates this schema from an incoming schema"""
+        # update all tables
+        for table in schema.tables.values():
+            self.update_table(table)
+        # update other settings
```
sh-rp, line 216: it does not seem we have to update the _settings here, because they are not set within the source, but I am not 100% sure. Lmk what else we need to take care here..

dlt/pipeline/pipeline.py
```diff
@@ -856,29 +856,25 @@ def _extract_source(self, storage: ExtractorStorage, source: DltSource, max_para
...
-        self._schema_storage.save_import_schema_if_not_exists(source_schema)
-
-        # get the current schema and merge tables from source_schema
-        # note we are not merging props like max nesting or column propagation
-        for table in source_schema.data_tables(include_incomplete=True):
-            pipeline_schema.update_schema(
-                pipeline_schema.normalize_table_identifiers(table)
-            )
+        # save import with fully discovered schema
+        self._schema_storage.save_import_schema_if_not_exists(pipeline_schema)
```
sh-rp, line 877: this will now export the full schema up to this point, not only the source schema discovered in this run. I am not sure if this is good or not, but when I was playing with the yaml files, i was a bit confused that not the full schema was exported.

dlt/common/normalizers/json/relational.py
```diff
@@ -261,6 +258,16 @@ def extend_schema(self) -> None:
             }
         )
 
+        # for every table with the write disposition merge, we propagate the root_key
+        for table_name, table in self.schema.tables.items():
```
rudolfix, line 262: just one thing: this is called only once when schema is created. so if we add any tables that have write disposition "merge" it won't be reflected in the schema.

we could have a mechanism that whenever we add a new table to the schema we call `extend` or `extend_table` again but I think it is overkill?

rudolfix: or maybe not, any time we update or add table we should call `extend_table` even here we should call it for all existing tables

dlt/common/schema/schema.py
rudolfix: I'd just overwrite the settings from incoming schema.

dlt/pipeline/pipeline.py
rudolfix: import schema must have only stuff discovered in the source. no columns from data normalization. for that you have export schema.

dlt/pipeline/pipeline.py
```diff
@@ -856,29 +856,25 @@ def _extract_source(self, storage: ExtractorStorage, source: DltSource, max_para
...
         # if source schema does not exist in the pipeline
         if source_schema.name not in self._schema_storage:
-            # save schema into the pipeline
-            self._schema_storage.save_schema(source_schema)
+            # create a new schema with the right name
+            self._schema_storage.save_schema(Schema(source_schema.name))
+
+        # get the schema from pipeline storage and merge stuff from the new schema
+        pipeline_schema = self._schema_storage[source_schema.name]
+        pipeline_schema.update_schema(source_schema)
```
rudolfix, line 866: probably makes sense to update settings and configs but not tables. that will happen later in extract and will be controlled by contracts

dlt/pipeline/pipeline.py
```diff
@@ -856,29 +856,25 @@ def _extract_source(self, storage: ExtractorStorage, source: DltSource, max_para
...
         source_schema.update_normalizers()
 
-        extract_id = extract_with_schema(storage, source, source_schema, self.collector, max_parallel_items, workers)
-
         # if source schema does not exist in the pipeline
         if source_schema.name not in self._schema_storage:
-            # save schema into the pipeline
-            self._schema_storage.save_schema(source_schema)
+            # create a new schema with the right name
+            self._schema_storage.save_schema(Schema(source_schema.name))
```
rudolfix, line 862: we are we saving empty schema? I get that we update it later but I'd just save source schema and assign pipeline schema to source schema. and do the lines below in else

rudolfix (changes requested):
this is really good PR. we standardize even more behavior where schemas are updated (also some normalize stuff) - beside fixing the bug.

we need some tests
- schema update especially when normalizer config is merged
- various cases when "merge" is changed to "replace" and back, we didn't have good tests because the bug that Lukas had was not detected
- do we have a test where propagation of other columns is configured and we change write disposition to merge to trigger the update in "extend"? and config gets merged properly?

dlt/pipeline/pipeline.py
sh-rp: then we have to revert to the approach from before where we merge source and pipeline schema at the end, otherwise we will always have all the stuff in this export

dlt/pipeline/pipeline.py
sh-rp: I thought it would be nice to always start with an empty schema and then merge in incoming stuff, this will be easier for implementing the contracts too. There already is table stuff on the source schema before it runs through extract. But for now I have set it back to the way it was.

dlt/pipeline/pipeline.py
sh-rp: There are tables already on the source_schema when it comes from the source before we run extract in some cases, so we need to merge these in too.

dlt/common/normalizers/json/relational.py
sh-rp: done

dlt/pipeline/pipeline.py
rudolfix: you still save an empty schema but let's try that :)

dlt/pipeline/pipeline.py
rudolfix: OK - with this implementation pipeline_schema is a reference for the contract. so if there are changes done in the code (ie. schema is modified directly - they will be rejected if contract says so. hmmm I think it is OK.

dlt/common/schema/schema.py
```diff
@@ -204,8 +205,23 @@ def update_schema(self, partial_table: TPartialTableSchema) -> TPartialTableSche
...
 
+
+    def update_schema(self, schema: "Schema") -> None:
+        """Updates this schema from an incoming schema"""
+        # update all tables
+        for table in schema.tables.values():
+            self.update_table(table)
+        # update normalizer config nondestructively
+        self.data_item_normalizer.update_normalizer_config(self, self.data_item_normalizer.get_normalizer_config(schema))
+        # update and compile settings
```
rudolfix, line 220: should we update normalizers? in theory they could change

dlt/common/normalizers/json/relational.py
```diff
@@ -261,6 +258,17 @@ def extend_schema(self) -> None:
             }
         )
 
+    def extend_table(self, table_name: str) -> None:
```
rudolfix, line 261: should we call it in "extend_schema"? ie. someone is modifying files directly and adding tables

rudolfix (changes requested):
there are two questions in the review
also Athena does not have merge so test fails
if the failing test persist in common I'll take a look. I was changing that and it does not look related to what you do (but it may...)

dlt/pipeline/pipeline.py
sh-rp: Yes, I prefer that, because the update path will always be the same, it seems cleaner to me. HOWEVER: this one test in common "tests/extract/test_incremental.py::test_replace_resets_state"n fails if I create a new schema her, so I am doing it the old way again now. I had the same problem / question on the contracts PR where this happened and I did not track it down what was going on. It seems that there is a difference between saving the schema directly on the first load and merging it into an empty one.

rudolfix (commented):
@sh-rp we'll switch to saving empty schema and updating pipeline one. I'll merge this into data contracts PR

thanks for all the tests and fixing schema update behavior!

rudolfix (approved):

PR #701, opened 2023-10-19
Support aws config `endpoint_url` with fsspec

steinitzu:
Allows setting `endpoint_url` to use s3 compatible storages. I.e. https://s3fs.readthedocs.io/en/latest/#s3-compatible-storage

The tests only verify that endpoint is resolved from config and passed to fsspec. I don't know if we should really add any integration tests with the r2 bucket or have it part of all other fs/staging tests. Doesn't seem neccassary imo since it's like any other s3 bucket.


rudolfix (approved):
LGTM!

Issue #700, opened 2023-10-18
support all writer file types and dlt columns in `ParquetItemsNormalizer`


rudolfix:
**Background**
We want to make the data normalized via `ParquetItemsNormalizer` more similar to data normalized via `JsonLItemsNormalizer`. - but only optionally

**Tasks**
1. * [x] if the writer type is not parquet, read the rows from parquet files and feed the writer of required type - instead of just moving the file. we do not infer anything with the data and just pass data around. we can support destinations like postgres with that
2. * [x] optionally - add `_dlt_load_id` column to parquet file with the current load id. please research how effectively can that be done. it is the same value repeated possibly millions of times. can we add this in batches? maybe we can use some tricks for low cardinality columns. remember to add this column also to the schema
3. * [x] optionally - add `_dlt_id` same way we add it in `JsonLItemsNormalizer`
4. * [x] we need to rename the columns in parquet to match the naming convention of the schema. to me it looks that the best place is in extract. we can use `normalize_table_identifiers` of Schema and use https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.rename_columns before writing in ArrowWriter. so we do not need to rewrite any files

**Implementation**
extend `NormalizeConfiguration` so we can change those options above. by default they are switched off.

PR #704, opened 2023-10-23
Support all writer types in parquet normalizer

steinitzu:
Implements: https://github.com/dlt-hub/dlt/issues/700

### DIFF
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -139,6 +139,25 @@ def _get_column_type_from_py_arrow(dtype: pyarrow.DataType) -> TColumnType:
         raise ValueError(dtype)
 
 
+def remove_null_columns(item: TAnyArrowItem) -> TAnyArrowItem:
+    """Remove all columns of datatype pyarrow.null() from the table or record batch
+    """
+    if isinstance(item, pyarrow.Table):
+        return item.drop([field.name for field in item.schema if pyarrow.types.is_null(field.type)])
+    elif isinstance(item, pyarrow.RecordBatch):
+        null_idx = [i for i, col in enumerate(item.columns) if pyarrow.types.is_null(col.type)]
+        new_schema = item.schema
+        for i in reversed(null_idx):
+            new_schema = new_schema.remove(i)
+        return pyarrow.RecordBatch.from_arrays(
+            [col for i, col in enumerate(item.columns) if i not in null_idx],
+            schema=new_schema
+        )
+    else:
+        raise ValueError(item)
+
+
+
 def py_arrow_to_table_schema_columns(schema: pyarrow.Schema) -> TTableSchemaColumns:
     """Convert a PyArrow schema to a table schema columns dict.
 
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -1,10 +1,10 @@
-from typing import Any, Tuple, Optional, Union
+from typing import Any, Tuple, Optional, Union, Callable, Iterable, Iterator, Sequence, Tuple
 from dlt import version
 from dlt.common.exceptions import MissingDependencyException
 from dlt.common.schema.typing import TTableSchemaColumns
 
 from dlt.common.destination.capabilities import DestinationCapabilitiesContext
-from dlt.common.schema.typing import TColumnType
+from dlt.common.schema.typing import TColumnType, TColumnSchemaBase
 from dlt.common.data_types import TDataType
 from dlt.common.typing import TFileOrPath
 
@@ -173,3 +173,29 @@ def get_row_count(parquet_file: TFileOrPath) -> int:
 
 def is_arrow_item(item: Any) -> bool:
     return isinstance(item, (pyarrow.Table, pyarrow.RecordBatch))
+
+
+TNewColumns = Sequence[Tuple[pyarrow.Field, Callable[[pyarrow.RecordBatch], Iterable[Any]]]]
+
+def pq_stream_with_new_columns(
+    parquet_file: TFileOrPath, columns: TNewColumns, batch_size: int = 100
+) -> Iterator[pyarrow.RecordBatch]:
+    """Add a column to the parquet file.
+
+    Args:
+        parquet_file: path or file object to parquet file
+        columns: list of columns to add in the form of (`pyarrow.Field`, column_value_callback)
+            The callback should accept a `pyarrow.RecordBatch` and return an iterable of values for the column.
+        batch_size: batch size to use when reading the parquet file. Defaults to 100.
+
+    Yields:
+        `pyarrow.RecordBatch` objects with the new columns added.
+    """
+    with pyarrow.parquet.ParquetFile(parquet_file) as reader:
+        schema: pyarrow.Schema
+        schema = reader.schema_arrow
+        for col in columns:
+            schema = schema.append(col[0])
+        # append the column to the schema
+        for batch in reader.iter_batches(batch_size=batch_size):
+            yield pyarrow.record_batch(batch.columns + [col[1](batch) for col in columns], schema=schema)
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -175,27 +175,30 @@ def is_arrow_item(item: Any) -> bool:
     return isinstance(item, (pyarrow.Table, pyarrow.RecordBatch))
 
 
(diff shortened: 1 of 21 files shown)


dlt/normalize/items_normalizers.py
```diff
@@ -112,6 +122,45 @@ def __call__(
...
+        load_storage: LoadStorage, load_id: str, schema: Schema, root_table_name: str,
+        add_load_id: bool, add_dlt_id: bool
+    ) -> int:
+        import numpy as np
+        new_columns: List[Any] = []
+        if add_load_id:
+            schema.update_table({"name": root_table_name, "columns": {"_dlt_load_id": {"name": "_dlt_load_id", "data_type": "text"}}})
+            new_columns.append((
+                pyarrow.pyarrow.field("_dlt_load_id", pyarrow.pyarrow.string(), nullable=False),
+                lambda batch: np.full(batch.num_rows, load_id)
```
rudolfix, line 139: can we use categorical type here? https://github.com/huggingface/datasets/issues/5706 (I think Arrow does not have it but take a look). maybe we could at least do that in numpy?

dlt/normalize/items_normalizers.py
```diff
@@ -112,6 +122,45 @@ def __call__(
...
+            ))
+        if add_dlt_id:
+            schema.update_table({"name": root_table_name, "columns": {"_dlt_id": {"name": "_dlt_id", "data_type": "text"}}})
+            new_columns.append((
+                pyarrow.pyarrow.field("_dlt_id", pyarrow.pyarrow.string(), nullable=False),
+                lambda batch: (uniq_id_base64(10) for _ in range(batch.num_rows))
```
rudolfix, line 145: is this what we do in relational normalizer?

rudolfix (changes requested):
@steinitzu looking good, left a few comments

dlt/extract/extract.py
```diff
@@ -205,13 +206,31 @@ class JsonLExtractor(Extractor):
...
+    def _rename_columns(self, items: List[TDataItem], new_column_names: List[str]) -> List[TDataItem]:
+        """Rename arrow columns to normalized schema column names"""
+        if not items:
+            return items
+        # TODO: Could check whether columns need renaming before doing this
+        if isinstance(items[0], pyarrow.pyarrow.Table):
+            return [item.rename_columns(new_column_names) for item in items]
+        elif isinstance(items[0], pyarrow.pyarrow.RecordBatch):
+            # Convert the batches to table -> rename -> then back to batches
+            return pa.Table.from_batches(items).rename_columns(new_column_names).to_batches()  # type: ignore[no-any-return]
```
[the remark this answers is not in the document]
steinitzu: Hmm, I'm not sure, I assumed they underlying arrays are not copied. Could instead create new batches with the new column names. I'll check which performs better.

dlt/normalize/items_normalizers.py
steinitzu: Looked like it, I didn't look too deeply into this code: https://github.com/dlt-hub/dlt/blob/sthor%2Fnormalize-parquet/dlt/common/normalizers/json/relational.py#L145-L147  Am I missing something?   

Should make a common function to generate ID though.

dlt/normalize/items_normalizers.py
steinitzu: There's a dictionary type in arrow which should be the same thing, I'll look into it

dlt/extract/extract.py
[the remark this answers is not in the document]
steinitzu: Doesn't look like anything's copied. The arrays buffers still have the same memory address, just different views of the same data.  

I checked with a 5mil row table in ~1000 batches. This runs in ~0.5ms for me.  

The only other way I can find is to create new batches like this:

```python
    return [
        pa.RecordBatch.from_arrays(batch.columns, names=new_columns)
        for batch in items
    ]
```
But that takes ~7ms on the same table.

dlt/normalize/items_normalizers.py
steinitzu: Seems `pyarrow.dictionary` works: https://github.com/dlt-hub/dlt/pull/704/files#diff-91126b068ef86213d8e95fcc59fa504fc708a9a20bc74c36a165d080a7e26050R134-R138   
The array takes less memory, but we still have to create the list of strings initially. Not sure if any way around that...

dlt/normalize/items_normalizers.py
steinitzu: Optimized this too with ids created in bulk with the underlying function: https://github.com/dlt-hub/dlt/pull/704/files#diff-18351bb2bb37a436db3f126f90a273da557aa78cdc1e0c17164cc235a66c764dR43-R51    
The individual `token_bytes` really add up.

dlt/extract/extract.py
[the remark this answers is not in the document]
rudolfix: I just run a few tests and rewriting those parquets is slower than rewriting them as JSON. probably could be expected...

dlt/common/libs/pyarrow.py
```diff
@@ -173,3 +173,29 @@ def get_row_count(parquet_file: TFileOrPath) -> int:
...
+    Args:
+        parquet_file: path or file object to parquet file
+        columns: list of columns to add in the form of (`pyarrow.Field`, column_value_callback)
+            The callback should accept a `pyarrow.RecordBatch` and return an iterable of values for the column.
+        batch_size: batch size to use when reading the parquet file. Defaults to 100.
+
+    Yields:
+        `pyarrow.RecordBatch` objects with the new columns added.
+    """
+    with pyarrow.parquet.ParquetFile(parquet_file) as reader:
```
rudolfix, line 194: I just run a few tests and rewriting those parquets is slower than rewriting them as JSON. probably could be expected...

dlt/normalize/items_normalizers.py
```diff
@@ -112,23 +118,96 @@ def __call__(
 
 
 class ParquetItemsNormalizer(ItemsNormalizer):
+    RECORD_BATCH_SIZE = 100000
```
rudolfix, line 121: we need to do something with this. maybe we could use `https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read_row_groups` instead of reading batches? then we could configure the number of groups to read. it also could be faster because the method returns tables, not batches

rudolfix (changes requested):
we are almost there but we need more tests!
- test that Arrows containing Nones can be written as INSERT and JSONL files and loaded
- actually make a test that will normalize the names in the arrow table. use some caps in names or write to weaviate which will capitalize all the names

dlt/normalize/items_normalizers.py
steinitzu: Yeah that's probably better. It's using the chunk size the table is written with originally so should be more efficient.

steinitzu: Makes no difference in speed from my tests if we're using the same batch size/row group size. But still makes sense to use this I think so we're rewriting the file with the same chunk size and not changing it.

tests/cases.py
```diff
@@ -351,6 +351,7 @@ def arrow_table_all_data_types(object_format: TArrowFormat, include_json: bool =
         "binary": [random.choice(ascii_lowercase).encode() for _ in range(num_rows)],
         "decimal": [Decimal(str(round(random.uniform(0, 100), 4))) for _ in range(num_rows)],
         "bool": [random.choice([True, False]) for _ in range(num_rows)],
+        "string_null": [random.choice(ascii_lowercase) for _ in range(num_rows - 1)] + [None],
```
rudolfix, line 354: create a column with only None. and then in panda frame you'll get a column without a type and our schema conversion routine fails

please try to reproduce. if you cannot I'll do that (ConnectorX is doing that when you select a column that has only None)

rudolfix (changes requested):
we also need addon to our docs:
- we need to mention normalize configuration
- we must explain that we rewrite parquet and that takes time.
- add info on supporting json and insert file formats if native parquet is not supported

tests/cases.py
steinitzu: Ok, yeah I see, it comes up as `pyarrow.null()` type. How would be best to handle this? Could just remove such columns before writing, think that makes sense.

rudolfix: you can create column without data type as well. it is called "incomplete" and will not be created in the destination

steinitzu: Tried this, but duckdb complains when there are extra columns in the parquet. So think we have to delete them.

rudolfix (approved):
LGTM! fixed the mssql timestamp

PR #707, opened 2023-10-23
arrow example

rudolfix:
### Description
* adds arrow + connector x example
* fixes how items are counted in writers + arrow
* incremental optimizations
* drop unix timestamp autodetect


dlt/common/schema/utils.py
```diff
@@ -698,4 +698,4 @@ def standard_hints() -> Dict[TColumnHint, List[TSimpleRegex]]:
 
 
 def standard_type_detections() -> List[TTypeDetections]:
-    return ["timestamp", "iso_timestamp"]
+    return ["iso_timestamp"]
```
z3z1ma, line 701: Is this a concern / breaking change for any experienced users relying on dlt pipelines in production and its existing type detection?

dlt/common/data_writers/buffered.py
```diff
@@ -79,10 +81,20 @@ def write_data_item(self, item: TDataItems, columns: TTableSchemaColumns) -> Non
         if isinstance(item, List):
             # items coming in single list will be written together, not matter how many are there
             self._buffered_items.extend(item)
+            # update row count, if item supports "num_rows" it will be used to count items
+            if len(item) > 0 and hasattr(item[0], "num_rows"):
```
steinitzu, line 85: This is great, makes more sense than counting a million row table as 1 :smile: 

There's an unlikely edge case, you can write a pydantic or dataclass obj with a `num_rows` attribute or method.   
So I would do an `isinstance` check for arrow types or just `if file_format == "arrow" ...`,  since that means we're always dealing with arrow objects, right?

dlt/common/data_writers/buffered.py
```diff
@@ -79,10 +81,20 @@ def write_data_item(self, item: TDataItems, columns: TTableSchemaColumns) -> Non
...
+                self._buffered_items_count += reduce(lambda prev, tbl: prev + tbl.num_rows, item, 0)
```
steinitzu, line 86: I think list comprehension is always faster: `sum([tbl.num_rows for tbl in item])`

dlt/extract/incremental/transform.py
```diff
@@ -180,6 +187,25 @@ def __call__(
...
+            for pk in unique_columns:
+                if pk not in tbl.schema.names:
+                    raise IncrementalPrimaryKeyMissing(self.resource_name, pk, tbl)
+            # use primary key as unique index
+            if isinstance(primary_key, str):
+                self._dlt_index = primary_key
+        elif primary_key is None:
+            unique_columns = tbl.column_names
+        else:  # deduplicating is disabled
+            unique_columns = None
```
steinitzu, line 207: Nice! Good creating the index as needed only :+1:

dlt/common/configuration/exceptions.py
```diff
@@ -48,6 +50,14 @@ def __str__(self) -> str:
...
             for tr in field_traces:
                 msg += f'\t\tIn {tr.provider} key {tr.key} was not found.\n'
+        # check if entry point is run with path. this is common problem so warn the user
+        main_path = main_module_file_path()
+        main_dir = os.path.dirname(main_path)
+        abs_main_dir = os.path.abspath(main_dir)
+        if abs_main_dir != os.getcwd():
+            # directory was specified
+            msg += "WARNING: dlt looks for .dlt folder in your current working directory and your cwd (%s) is different from directory of your pipeline script (%s).\n" % (os.getcwd(), abs_main_dir)
+            msg += "If you keep your secret files in the same folder as your pipeline script but run your script from some other folder, secrets/configs will not be found\n"
```
steinitzu, line 60: :heart:

dlt/extract/incremental/transform.py
```diff
@@ -206,24 +232,19 @@ def __call__(
...
         # The new max/min value
         try:
-            row_value = compute(tbl[cursor_path]).as_py()
+            orig_row_value = compute(tbl[cursor_path])
+            row_value = orig_row_value.as_py()
+            # dates are not represented as datetimes but I see connector-x represents
+            # datetimes as dates and keeping the exact time inside. probably a bug
+            # but can be corrected this way
+            if isinstance(row_value, date) and not isinstance(row_value, datetime):
+                row_value = pendulum.from_timestamp(orig_row_value.cast(pa.int64()).as_py() / 1000)
```
steinitzu, line 241: Interesting. Should we have a test for this?  
Not common probably, but initial and end values should be converted the same way if they're dates so they're comparable.

steinitzu (approved):
Amazing! Looks like we're ready for a much faster sql-database source :)  
Added a few nitpicks but looks good.

dlt/common/schema/utils.py
rudolfix: existing pipelines will keep the already inferred type and all will work the same. only when dataset or table is deleted and data is reloaded, the type will be different. do you rely in this anywhere? such timestamps happen most often in logs

dlt/extract/incremental/transform.py
rudolfix: this is a bug in connector x. when I use arrow2 type with binary protocol, the types are correct. if the end and start values are dates, they will compare with datetimes just fine. it was working in my (manual) tests

dlt/common/data_writers/buffered.py
rudolfix: yes, it will be faster than using lambdas...

z3z1ma:
How does connector-x handle larger than memory datasets. In a regular SQL source, you chunk results from the cursor/server in proper generator fashion. this on the other hand is most like to one shot the entire result on the first call to `next()` right. I love the idea of pulling data directly to arrow in memory via rust, but am concerned about the lack of a generator-like interface (which I am not sure is possible via the FFI?)

dlt/common/schema/utils.py
z3z1ma: Ah okay cool, I was mostly wondering for the case of existing pipelines. I have some Mongodb/Timescaledb data pipelines which may inadvertently rely on this. Its nbd if this is no longer the default behavior but instead opt-in, just good to know! I can visualize many cases where this would not be desired but luckily, none of those ever happened to me :)

rudolfix:
> How does connector-x handle larger than memory datasets. In a regular SQL source, you chunk results from the cursor/server in proper generator fashion. this on the other hand is most like to one shot the entire result on the first call to `next()` right. I love the idea of pulling data directly to arrow in memory via rust, but am concerned about the lack of a generator-like interface (which I am not sure is possible via the FFI?)
> 
> Consider a warning for users if the above holds true. Regardless, I would be extremely keen on some benchmarking results against the existing sql source, I bet its blazingly fast by comparison 🆒

they load everything into memory first - pre-allocating the needed space. so you need to chunk the stream yourself. that's quite unfortunate. if you have any good index, it should be easy though. I do that in the example here.

https://github.com/sfu-db/connector-x#how-does-connectorx-download-the-data

PR #708, opened 2023-10-24
Synapse

eryanRM:
### Description
Initial PR bringing Synapse functions into subclass

Note: new errors (ex. 'str' mapping) popped up when running updated tests

ex 1. resolved: 'str' mapping errors in several test_job_client.py
ex 2. tests/load/test_job_client.py::test_load_with_all_types now failing due to column count not matching table

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### Additional Context
This PR is for David's review to confirm we are on correct path


dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,342 @@
...
+    dbapi: ClassVar[DBApi] = pyodbc
+    capabilities: ClassVar[DestinationCapabilitiesContext] = capabilities()
+
+    def __init__(self, dataset_name: str, credentials: SynapseCredentials) -> None:
+        super().__init__(credentials.database, dataset_name)
+        self._conn: pyodbc.Connection = None
+        self._transaction_in_progress = False  # Transaction state flag
+        self.credentials = credentials
+
+    def generate_insert_query(self, sql: str) -> tuple:
```
sh-rp, line 58: as previously discussed please leave a link here, for example: https://stackoverflow.com/questions/36141006/how-to-insert-multiple-rows-into-sql-server-parallel-data-warehouse-table. This method can be static and should get a bunch of unit tests to ensure it works properly.

dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,342 @@
...
+                if self._transaction_in_progress:
+                    self._conn.commit()
+                self._conn.close()
+        except pyodbc.Error as e:
+            # Log detailed error information here
+            logger.error(f"Failed to close connection due to pyodbc.Error: {str(e)}")
+
+
+    @contextmanager
+    def begin_transaction(self) -> Iterator[DBTransaction]:
```
sh-rp, line 135: what exactly happens when we use the same tx code we have in the mssql implementation? why can't we use that?

dlt/destinations/synapse/sql_client.py
```diff
@@ -0,0 +1,342 @@
...
+    def execute_sql(self, sql: AnyStr, *args: Any, **kwargs: Any) -> Optional[Sequence[Sequence[Any]]]:
+        #logging.info(f"Executing SQL: {sql} with arguments: {args} and keyword arguments: {kwargs}")
+        # Convert any JSON arguments to strings
+        args = tuple(json.dumps(arg) if isinstance(arg, dict) else arg for arg in args)
+
+        param_values = args  # This line remains unchanged as you're capturing all additional arguments into param_values
+        original_sql = sql
+
+        # Check if it's a multi-row insert
+        is_multi_row_insert = (
```
sh-rp, line 256: you should implement a custom InsertValuesLoadJob that does this and not hack the execute_sql function.

sh-rp (changes requested):
Hey @eryanRM, this PR is looking much better now. I have some requests:
* Please Make sure the branch is up to date with devel, there are some small changes in the way the loader resolved which tables are needed on staging etc. You can probably just use the settings from the mssql client for that
* Please make sure the tests are actually executed so I can see what passes and what does not. This includes removing debug/test code and making sure the poetry synapse extra works.
* Generally speaking it seems like there is a lot of code shared between mssql and synapse. I would be better to have common baseclasses for the two of them instead of having the same code in separate places.

dlt/destinations/synapse/sql_client.py
eryanRM: Thank you for the feedback, @sh-rp.  How about a subclass of InsertValuesLoadJob tailored for Microsoft Synapse that lives in synapse.py?

Given that DLT handles multiple destinations and insert_job_client.py provides a foundation for all of them, I can isolate Synapse-specific conversions and avoid execute_sql

SynapseInsertValuesLoadJob

dlt/destinations/synapse/sql_client.py
eryanRM: Shifted back to mssql implementation.  Current approach helped with error handling; however, we aren't facing transaction errors anymore and can switch back.

dlt/destinations/synapse/sql_client.py
eryanRM: @sh-rp thank you for feedback.  Googled and liked your link the most

Here is another: https://stackoverflow.com/questions/77014211/how-to-insert-values-in-a-table-in-azure-synapse-analytics

dlt/destinations/synapse/sql_client.py
eryanRM: @sh-rp circling back to this, please let me know if this is in right direction.  generate_insert_query() now lives in synapse.py 

Instead of taking a SQL string (vulnerable to injection) it now takes comparable params to InsertValuesLoadJob

eryanRM:
Hi @sh-rp, I appreciate the feedback.  This has been an illuminating process

- Updated with Devel.  Please let me know if it's still meaningfully behind
- I removed debug and print statements.  Can you see tests on your end?  Locally, I dropped to 40 passing tests after syncing with devel and migrating / modifying generate_insert_query to work in subclass.  I am confident I can get back to passing (while keeping as much common with mssql as possible)
- Will work on baseclass through weekend and consult with @steinitzu .  Priority has been getting back to passing vast majority of tests but with these improvements implemented (less hacky)

eryanRM:
Refreshing Synapse branch w/ 'devel'.  Continuing as new PR

PR #727, opened 2023-10-31
prototype platform connection

sh-rp:
Adds support for sending traces to the dlthub platform. Implements:

- Support for multiple tracking modules
- Sending full run trace after pipeline was run to platform on a separate thread
- Adds pipeline name to pipeline trace
- Adds Schemas to LoadInfo
- Adds execution context to pipeline trace


dlt/common/storages/load_storage.py
```diff
@@ -87,6 +87,7 @@ class LoadPackageInfo(NamedTuple):
     package_path: str
     state: TLoadPackageState
     schema_name: str
+    schema: Schema
```
sh-rp, line 90: fyi i am adding the full schema to the load info here

dlt/pipeline/platform.py
```diff
@@ -0,0 +1,31 @@
...
+def on_start_trace_step(trace: PipelineTrace, step: TPipelineStep, pipeline: SupportsPipeline) -> None:
+    # _send_to_beacon(trace, step, pipeline, None)
+    pass
+
+def on_end_trace_step(trace: PipelineTrace, step: PipelineStepTrace, pipeline: SupportsPipeline, step_info: Any) -> None:
+    # _send_to_beacon(trace, step, pipeline, step_info)
+    pass
+
+def on_end_trace(trace: PipelineTrace, pipeline: SupportsPipeline) -> None:
+    _send_to_beacon(trace, None, pipeline, None)
```
sh-rp, line 29: for now only send on end trace, but it would be nice to have the full progress available

sh-rp:
one thing that is a bit strange in the trace is, that there is this additional run step which duplicates the result of the loadinfo as stepinfo. It's not a problem really, but for one the run step is just a representation of the full trace really, and we are sending too much data around, especially given the fact that the loadinfo will always be the most verbose.

dlt/common/storages/load_storage.py
rudolfix: IMO way better if you add TypedDict with schema content, not the object itself. this is being pickled and dumped into trace so obviously dicts work better

dlt/pipeline/platform.py
```diff
@@ -0,0 +1,31 @@
...
+from dlt.pipeline.trace import PipelineTrace, PipelineStepTrace, TPipelineStep, SupportsPipeline
+from dlt.common import json
+
+count = 0
+
+def _send_to_beacon(trace: PipelineTrace, step: PipelineStepTrace, pipeline: SupportsPipeline, some):
+    if pipeline.runtime_config.beacon_token and pipeline.runtime_config.beacon_url:
+        trace_dump = json.dumps(trace.asdict())
+        url = f"{pipeline.runtime_config.beacon_url}/pipeline/{pipeline.runtime_config.beacon_token}/traces"
+        requests.put(url, json=trace_dump)
```
rudolfix, line 14: maybe you could reuse our telemetry thread executor to send the messages in the thread, without blocking the pipeline execution? the code is already there and I think could be repurposed.

rudolfix (changes requested):
two suggestions :) also do you want to merge it or keep a branch for testing?

sh-rp:
> two suggestions :) also do you want to merge it or keep a branch for testing?

For now this is only for testing i would say.

dlt/common/pipeline.py
```diff
@@ -54,7 +54,7 @@ def asdict(self) -> DictStrAny:
         """A dictionary representation of NormalizeInfo that can be loaded with `dlt`"""
         d = self._asdict()
         # list representation creates a nice table
-        d["row_counts"] = [(k, v) for k, v in self.row_counts.items()]
+        d["row_counts"] = [{"table_name": k, "count": v} for k, v in self.row_counts.items()]
```
sh-rp, line 57: this makes a much nicer table..

dlt/pipeline/trace.py
```diff
@@ -162,22 +166,24 @@ def on_end_trace(self, trace: PipelineTrace, pipeline: SupportsPipeline) -> None
...
+        with suppress_and_warn():
+            module.on_start_trace(trace, step, pipeline)
     return trace
 
 
 def start_trace_step(trace: PipelineTrace, step: TPipelineStep, pipeline: SupportsPipeline) -> PipelineStepTrace:
     trace_step = PipelineStepTrace(uniq_id(), step, pendulum.now())
-    with suppress_and_warn():
-        TRACKING_MODULE.on_start_trace_step(trace, step, pipeline)
+    trace.steps.append(trace_step)
```
sh-rp, line 183: is there any downside to attaching the trace step on trace start? it would be nicer to have it like this when sent to the platform

rudolfix:
> one thing that is a bit strange in the trace is, that there is this additional run step which duplicates the result of the loadinfo as stepinfo. It's not a problem really, but for one the run step is just a representation of the full trace really, and we are sending too much data around, especially given the fact that the loadinfo will always be the most verbose.

run step is used to correlate several pipeline steps into one transaction. it makes sense IMO to send some summary information. yeah I think we repeat the `LoadInfo` (or the last step info) in it? then let's change it to RunInfo with just basic information. we could also use it to send full state sync (but not everyone uses `run` method so probably bad idea

dlt/common/pipeline.py
rudolfix: absolutely! this will get a proper treatment in #757

dlt/common/runtime/typing.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Set, Type, TypedDict, NewType, Union, get_args
+
+
+TExecInfoNames = Literal["kubernetes", "docker", "codespaces", "github_actions", "airflow", "notebook", "colab","aws_lambda","gcp_cloud_function"]
```
rudolfix, line 4: could you format it? or maybe we should go back to blake formatter initially in the same mode we have in verified sources?

dlt/common/runtime/typing.py
```diff
@@ -0,0 +1,18 @@
...
+
+class TVersion(TypedDict):
+    """TypeDict representing a library version"""
+    name: str
+    version: str
+
+class TExecutionContext(TypedDict):
```
rudolfix, line 11: Looking good! for open telemetry we can collect way more information that is not anonymous. but in other PR

dlt/common/storages/load_storage.py
```diff
@@ -87,6 +88,7 @@ class LoadPackageInfo(NamedTuple):
     package_path: str
     state: TLoadPackageState
     schema_name: str
+    schema: TStoredSchema
```
rudolfix, line 91: schema hash. send a separate pipeine state message with all pipeline schemas

dlt/pipeline/platform.py
```diff
@@ -0,0 +1,47 @@
+"""Implements SupportsTracking"""
+from typing import Any
+import requests
+from concurrent.futures import ThreadPoolExecutor
+
+from dlt.pipeline.trace import PipelineTrace, PipelineStepTrace, TPipelineStep, SupportsPipeline
+from dlt.common import json
+from dlt.common.runtime import logger
+
+_THREAD_POOL: ThreadPoolExecutor = None
```
rudolfix, line 10: is this temporary code? we should reuse the segment pool or at least somehow add this pool to `at_exit` handler otherwise messages will not go out at the end. 

maybe extract this sending pool to common module that creates a pool and registers it in exit handler?

also implementation below does not have a method to stop the pool. won't that be a problem when testing?

dlt/pipeline/trace.py
```diff
@@ -92,6 +93,8 @@ def asdict(self) -> DictStrAny:
 class PipelineTrace:
     """Pipeline runtime trace containing data on "extract", "normalize" and "load" steps and resolved config and secret values."""
     transaction_id: str
+    pipeline_name: str
+    execution_context: TExecutionContext
```
rudolfix, line 97: cool!

dlt/common/configuration/specs/run_configuration.py
```diff
@@ -27,6 +27,8 @@ class RunConfiguration(BaseConfiguration):
     request_max_retry_delay: float = 300
     """Maximum delay between http request retries"""
     config_files_storage_path: str = "/run/config/"
+    """Platform connection"""
+    platform_dsn: Optional[str] = None
```
rudolfix, line 31: is workspace cookie part of platform_dsn? also this is a secret value so use TSecretStrValue here. maybe we could rename it to dlthub_dsn?

rudolfix (changes requested):
good! before we merge:
1. your thread pool must register in `at_exit` best if you could refactor segment pool to have many pools registered there
2. as usual some naming suggestions

dlt/common/runtime/typing.py
sh-rp: i am also for black formatter..

dlt/pipeline/platform.py
sh-rp: I have extracted a managedexecutor class. it does not seem a missing stop method is a problem when testing...

sh-rp:
I am not quite sure i have implemented reporting the current state data to the platform the way you had in mind, let me know.

dlt/common/runtime/typing.py
```diff
@@ -0,0 +1,28 @@
+from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Set, Type, TypedDict, NewType, Union, get_args
+
+
+TExecInfoNames = Literal[
+    "kubernetes",
```
rudolfix, line 5: look also at this. maybe we can copy code form there to have even more CI envs?
https://www.npmjs.com/package/ci-info

also there's `CI` env flag which says that code runs in CI so maybe we should add "generic_ci"

dlt/pipeline/__init__.py
```diff
@@ -245,8 +245,8 @@ def run(
     )
 
 # plug default tracking module
-from dlt.pipeline import trace, track
-trace.TRACKING_MODULE = track
+from dlt.pipeline import trace, track, platform
```
rudolfix, line 248: ok for platform but maybe we should just say opentelemetry?

dlt/pipeline/pipeline.py
```diff
@@ -90,48 +90,52 @@ def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
...
-                    assert self._trace == trace, f"Messed up trace reference {id(self._trace)} vs {id(trace)}"
-                    # if we end new trace that had only 1 step, add it to previous trace
-                    # this way we combine several separate calls to extract, normalize, load as single trace
-                    # the trace of "run" has many steps and will not be merged
-                    self._last_trace = merge_traces(self._last_trace, trace)
-                    self._trace = None
+                try:
+                    if trace_step:
+                        # if there was a step, finish it
+                        end_trace_step(self._trace, trace_step, self, step_info, send_state)
```
rudolfix, line 122: OK we can send it in `end_trace_step` or here... probably you are right - we can make it trace module specific

dlt/pipeline/pipeline.py
```diff
@@ -329,7 +333,7 @@ def normalize(self, workers: int = 1, loader_file_format: TLoaderFileFormat = No
             except Exception as n_ex:
                 raise PipelineStepFailed(self, "normalize", n_ex, normalize.get_normalize_info()) from n_ex
 
-    @with_runtime_trace
+    @with_runtime_trace()
```
rudolfix, line 336: I'd send the state here

dlt/pipeline/pipeline.py
```diff
@@ -382,7 +386,7 @@ def load(
         except Exception as l_ex:
             raise PipelineStepFailed(self, "load", l_ex, self._get_load_info(load)) from l_ex
 
-    @with_runtime_trace
+    @with_runtime_trace(send_state=True)
```
rudolfix, line 389: not here

dlt/pipeline/platform.py
```diff
@@ -0,0 +1,101 @@
...
+    # sync only if load step was processed
+    load_info: LoadInfo = None
+    for step in trace.steps:
+        if step.step == "load":
+            load_info = cast(LoadInfo, step.step_info)
+
+    if not load_info:
+            return
+
+    payload = TSchemaSyncPayload(
```
rudolfix, line 62: looks like PipelineStateSync. LGTM for now. we can add more data to it when we need it. we could also add the pipeline state. it will help with debugging. there's a method to retrieve it in the pipeline and methods to serialize it

rudolfix (changes requested):
> I am not quite sure i have implemented reporting the current state data to the platform the way you had in mind, let me know.

LGTM. I think we will modify it a lot when we actually use it for something :) also see my comments.


> I should probably check the load packages and not sync schemas for failed loadpackages? I am not 100 % sure on that atm.
heh good question and another nuance
1. you should not bump schema revision in the dataset to which we were loading if the package failed. soon load info will have an information if schema was upgraded in dataset - even if the package failed (#757 )
2. you should bump pipeline schema revision to the one that is in state.

pipeline may be ahead of the dataset in terms of schema revision. it may produce load packages that are loaded somewhere else. so we have revisions materialized in dataset and current revision in the pipeline. they may be different

> Maybe the whole state/schema synching should not be in the trace decorator at all. It should work the way I implemented it, but somehow it does not feel quite right.
depends if we think it has anything useful for open telemetry collector.

the traces are open telemetry friendly. the state sync MAY BE something more spcific. but for now - LGTM

and we can merge that branch soon

dlt/pipeline/platform.py
sh-rp: renamed it

dlt/pipeline/__init__.py
sh-rp: at the moment it is not opentelemetry format at all. i can rename it, but i would just say we switch to opentelemetry when the prototype is out and then also rename this file.

rudolfix (approved):
LGTM!

PR #732, opened 2023-11-02
Nit: Remove py.typed file that makes Pyright incessant

z3z1ma:
Pyright is incredibly incessant about symbols not being exported due to the presence of the py.typed file but not following the rules on imports denoted here

https://microsoft.github.io/pyright/#/typed-libraries?id=library-interface

Solutions

- either adhere to the rules, meaning employ redundant imports like `from dlt.common.configuration.accessors import config as config`

- use `__all__` in files where names are being exported, especially in `dlt/__init__.py`

- remove the `py.typed` file

I have done the latter. The static type checkers used in LSPs (like pyright, which majority use) are very incessant otherwise and I end up with this ugly bit all over my code `type: ignore[import]` just to silence errors.

If you prefer using `__all__` we can do that too

### DIFF
--- a/dlt/__init__.py
+++ b/dlt/__init__.py
@@ -41,3 +41,27 @@ TSecretValue = _TSecretValue
 
 TCredentials = _CredentialsConfiguration
 "When typing source/resource function arguments it indicates that a given argument represents credentials and should be taken from dlt.secrets. Credentials may be a string, dictionary or any other type."
+
+
+__all__ = [
+    "__version__",
+    "config",
+    "secrets",
+    "state",
+    "Schema",
+    "source",
+    "resource",
+    "transformer",
+    "defer",
+    "pipeline",
+    "run",
+    "attach",
+    "Pipeline",
+    "dbt",
+    "progress",
+    "current",
+    "mark",
+    "TSecretValue",
+    "TCredentials",
+    "sources",
+]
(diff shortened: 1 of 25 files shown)


steinitzu:
Mypy still needs py.typed, or has that changed?   
In that case dlt+mypy users will get missing import errors.   

So I'd say rather use __all__.   
Does this issue affect pyright users importing dlt in their own projects?

z3z1ma:
Yes any pyright user importing dlt gets LSP errors (not even warnings, but actual errors) when using common dlt components @steinitzu 
I can post screenshot too

> So I'd say rather use __all__.

Ok got it

steinitzu:
> Yes any pyright user importing dlt gets LSP errors (not even warnings, but actual errors) when using common dlt components @steinitzu I can post screenshot too

Got you, wasn't clear whether it's just when working in the dlt codebase.  Then definitely we should fix this imo.

Enabling [F401](https://www.flake8rules.com/rules/F401.html) in flake8 helps catch this too, currently ignored in our config.

rudolfix:
@z3z1ma is this a lot of work to add __all__? @steinitzu feel free to help here ie. by enabling linting and fixing imports. I'd rather keep `py.typed`.

steinitzu:
Started adding `__all__` here, good opportunity to remove unused imports at the same time. But it's a lot of busywork ...

rudolfix:
> Started adding `__all__` here, good opportunity to remove unused imports at the same time. But it's a lot of busywork ...

that can be done automatically. let's do __all__ whenever necessary and then merge this. at least this is my take :)

rudolfix (approved):
LGTM! @z3z1ma does pyright still complain?

z3z1ma:
LGTM @rudolfix 

Before
image


After
image

Thx @steinitzu

PR #734, opened 2023-11-03
Add authenticator to SnowflakeCredentials class

gjdevincentis:


rudolfix (approved):
thanks for contrib and this LGTM! one question:
what about the `token` as visible here: https://stackoverflow.com/questions/58715101/is-it-possible-to-connect-python-client-to-snowflake-using-oauth2?

do you need to pass it or your machine could generate the refresh token automtically?

gjdevincentis:
Oh interesting, not sure if a token is required when using the authenticator for 'oauth2' (or if other kwargs are needed for any other valid authenticators)

For my purposes I was referencing an external authenticator link ie `https://foo.okta.bar...` for which no token, or any other authenticator kwargs are needed

PR #735, opened 2023-11-06
add support for gdrive files

sehnem:
### Description

This PR makes some changes to the filesystem module in order to support gdrive.

### Additional Context

In gdrive when parsing a directory that is not the user root, eg. a directory that is shared with the user, you need to provide `root_file_id`, this PR adds support for it without changing `GcpCredentials`, it also removes the `netloc` that contains the `root_file_id` before requesting the data to fsspec.

For fetching a file using file_id the user will need to provide something like: `gdrive://<file_id>/path`
If the user wants to get the root from the user gdrive account the netloc should not be provided: `gdrive:///path`

### DIFF
--- a/dlt/common/storages/configuration.py
+++ b/dlt/common/storages/configuration.py
@@ -60,7 +60,7 @@ class FilesystemConfiguration(BaseConfiguration):
     PROTOCOL_CREDENTIALS: ClassVar[Dict[str, Any]] = {
         "gs": Union[GcpServiceAccountCredentials, GcpOAuthCredentials],
         "gcs": Union[GcpServiceAccountCredentials, GcpOAuthCredentials],
-        "gdrive": GcpOAuthCredentials,
+        "gdrive": Union[GcpServiceAccountCredentials, GcpOAuthCredentials],
         "s3": AwsCredentials,
         "az": Union[AzureCredentialsWithoutDefaults, AzureCredentials],
         "abfs": Union[AzureCredentialsWithoutDefaults, AzureCredentials],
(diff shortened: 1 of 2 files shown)


rudolfix (changes requested):
Hey! Now when I see what we need to modify in our code in order to get gdrive working I think we should push all of this into the gdrive implementation.

Actually `fsspec` provides a way to get additional arguments from query string. You need to implement `_get_kwargs_from_urls` class method in the filesystem. so we can get `root_file_id` from it. see what fsspec does
```
def _un_chain(path, kwargs):
    x = re.compile(".*[^a-z]+.*")  # test for non protocol-like single word
    bits = (
        [p if "://" in p or x.match(p) else p + "://" for p in path.split("::")]
        if "::" in path
        else [path]
    )
    # [[url, protocol, kwargs], ...]
    out = []
    previous_bit = None
    kwargs = kwargs.copy()
    for bit in reversed(bits):
        protocol = kwargs.pop("protocol", None) or split_protocol(bit)[0] or "file"
        cls = get_filesystem_class(protocol)
        extra_kwargs = cls._get_kwargs_from_urls(bit)
        kws = kwargs.pop(protocol, {})
        if bit is bits[0]:
            kws.update(kwargs)
        kw = dict(**extra_kwargs, **kws)
        bit = cls._strip_protocol(bit)
        if (
            protocol in {"blockcache", "filecache", "simplecache"}
            and "target_protocol" not in kw
        ):
            bit = previous_bit
        out.append((bit, protocol, kw))
        previous_bit = bit
    out = list(reversed(out))
    return out
```
then we can pass `root_file_id` in bucket url in query string. I think it is a fine compromise that can be accepted by `gdrive` maintainer. if not we'll go with out own implementation

TLDR:
1. please check out if the above works
2. resubmit the PR as contributor (you are added now)
3. add gdrive tests to all existing tests in `test_filesystem_common.py`

PR #740, opened 2023-11-07
synapse: updated _insert function, escape chars (reset w/ devel)

eryanRM:
### Description

Synapse destination integration

In addition to actioning earlier feedback, this PR condenses Synapse multi-row insert logic into '_insert' subclass

This replaces closed PR's: https://github.com/dlt-hub/dlt/pull/708 and https://github.com/dlt-hub/dlt/pull/725

### Related Issues

- Fixes #...
- Closes # https://github.com/dlt-hub/dlt/pull/708 and https://github.com/dlt-hub/dlt/pull/725 
- Resolves #...

### Additional Context
Currently stalling on test: tests/load/pipeline/test_merge_disposition.py::test_merge_no_child_tables 

VARBINARY(MAX) not working w/ Synapse column store tables
https://stackoverflow.com/questions/75517148/replacing-nvarcharmax-in-azure-synapse-table


dlt/destinations/synapse/synapse.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __init__(self, table_name: str, file_path: str, sql_client: SqlClientBase[Any]) -> None:
+        # First, set any attributes specific to this subclass
+        self._sql_client = sql_client
+        self._file_name = FileStorage.get_file_name_from_file_path(file_path)
+
+        # Then, call the parent class's __init__ method with the required arguments
+        super().__init__(table_name, file_path, sql_client)
+
+    def _insert(self, qualified_table_name: str, file_path: str) -> Iterator[List[str]]:
```
sh-rp, line 128: this is much better now, but my idea had been to hook into the _insert function of the superclass. the problem you have now is, that your implementation does not respect self._sql_client.capabilities.max_query_length and self._sql_client.capabilities.max_rows_per_insert. that's why I suggested to brancht out at "if max_rows is not None:" and not at the beginning of this method.

eryanRM: @sh-rp Thanks for feedback -- apologies for not implementing your idea correctly.  Is this better?  Still WIP ... am I branching out correctly?

Issue #142, opened 2023-02-20
empty rows get dropped when normalizing and loading


rudolfix:
If an empty row (ie `{}` or `{"value": None}` is passed to pipeline `run` method, it gets dropped and nothing is loaded
correct behavior: load a row with all columns set to NULL (except `dlt` added ones)

bug applies only to rows for existing tables with existing columns. In case of non existing tables and table stubs (without any columns) such empty rows should not generate records to load

this most probably happens in `normalize` stage where rows get dropped if all data is filtered out by `exclude` filters

PR #745, opened 2023-11-08
Empty rows fix

steinitzu:
Closes https://github.com/dlt-hub/dlt/issues/142


steinitzu:
@rudolfix I couldn't reproduce the bug, just added a test and it passes already. All the rows are loaded. So guessing it was fixed already?

rudolfix (approved):
@steinitzu that's possible. this code was refactored for a few times

Issue #508, opened 2023-07-21
allow to parametrize destinations by converting them into callables


rudolfix:
**Background**
destinations do not follow the general principle of exposing config and secret values as arguments (which for example sources and `Pipeline` do). Instead we partially allow for that by `credentials` argument to `dlt.pipeline` that then is used to instantiate destination configuration. This argument is not necessary and does not work with `staging`.

Currently we allow to pass the credentials as a name or module instance. the second one is rarely used:
```python
import dlt
from dlt.destinations import filesystem

p = dlt.pipeline(pipeline_name="bucket_pipeline", destination=filesystem)
```

We want to allow to pass the essential arguments explicitly
```python
import dlt
from dlt.destinations import filesystem

p = dlt.pipeline(pipeline_name="bucket_pipeline", destination=filesystem("s3://bucket_name/", credentials={...}))
```
but keep existing syntax

**Tasks**
1. * [ ] allow the existing destinations to be called in order to pass essential configuration arguments, how to implement that is tricky, a module import should still work, a call to that import should return same type but parametrized (so probably this is a class with a factory)
2. * [ ] convert all destinations to new structure. `credentials` must be present in all callables' signatures (following the destination config spec)
3. * [ ] deprecate the `credentials` argument of pipeline. warn when used and apply it over any other setting
4. * [ ] test such instantiation for all destinations (along existing tests that use names)
5. * [ ] new callables must have good docstrings (people will see them when adding to pipeline!)
6. * [ ] add information on instatiation to landing page of destinations

**Implementation details**
Mind that `DestinationReference` is always resolved in `reference.py` so probably your tricks should happen in `from_name` method. Also `DestinationReference` is a protocol that each destination module implements. We can do same trick we do for normalizers: they still are modules with protocol but protocol just says how to instantiate a class that does all the rest.

We want the destinations to still be modules as they are now. do not convert everything to one big class

PR #746, opened 2023-11-09
Parametrized destinations

steinitzu:
### Description

Draft implementation of parametrized destinations.   

I added a base `DestinationFactory` with same interface as the destination modules just to hold the config arguments.   

Usage is e.g. 

```python
import dlt
from dlt.destinations import snowflake, filesystem


pipeline = dlt.pipeline(
    "callable_dest_staging",
    destination=snowflake(),
    staging=filesystem("s3://my-bucket")
)
```

The biggest change is that I moved all destinations to `dlt.destinations.impl`. Just because I think it would be more confusing to have a module and a callable with the same name.   
I.e. `dlt.destinations.postgres(...)` is the factory, `dlt.destinations.impl.postgres` is the package with all the destination code.

### Related Issues

- Resolves https://github.com/dlt-hub/dlt/issues/508


steinitzu:
@rudolfix would be great to know what you think of the approach before I finalize it.

dlt/destinations/impl/snowflake/factory.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(
+        self,
+        credentials: SnowflakeCredentials = None,
+        stage_name: t.Optional[str] = None,
+        keep_staged_files: bool = True,
+        **kwargs: t.Any,
+    ) -> None:
+        cfg: SnowflakeClientConfiguration = kwargs['_dlt_config']
+        self.credentials = cfg.credentials
+        self.config_params = {
```
rudolfix, line 23: all configurations are also dictionaries. so you could get `dict`, assign credentials (part of the base) and pass rest of the dict as `config_params`

also possible at run time IMO - when actual client/capabilities are created

rudolfix (commented):
@steinitzu I really like it!. using partial to get initial config is smart. I think factory impls can be simplified even more.

btw. we'll have another ticket to improve destinations even further ie. to let people assign names to them so we have dynamic configs in case of having several environments ie.
`progress-dev`
will be taken from `[destination.progress-dev]` just a headsup

but we'll merge that first

rudolfix (commented):
@steinitzu on the second look: I think we should also instantiate destination factory in pipeline, when we pass the destination as string and completely get rid of ability to pass a module.

I bet we can get rid of most of the code in `__init__.py` of each of destination implementations: for sure the `_configure` and `client()` methods. maybe all of them?

steinitzu:
Not quite done, but some more progress.   
1. `DestinationReference` is removed entirely. `client/capabilities/spec` defined on the factory and always use it
2.  Also renamed `DestinationFactory` to `Destination`, think that makes sense as now it's "the" base destination
3. I create capabilities at import time ( https://github.com/dlt-hub/dlt/blob/sthor%2Fparametrized-destination/dlt/destinations/impl/postgres/factory.py#L15 ). I couldn't see a reason to have them behind a callable (except for dummy where they can be configured). Any problem with this?

dlt/common/destination/__init__.py
```diff
@@ -1,10 +1,9 @@
 from dlt.common.destination.capabilities import DestinationCapabilitiesContext, TLoaderFileFormat, ALL_SUPPORTED_FILE_FORMATS
-from dlt.common.destination.reference import DestinationReference, TDestinationReferenceArg
+from dlt.common.destination.reference import TDestinationReferenceArg, Destination
 
 __all__ = [
     "DestinationCapabilitiesContext",
     "TLoaderFileFormat",
     "ALL_SUPPORTED_FILE_FORMATS",
-    "DestinationReference",
```
rudolfix, line 8: `Destination`?

dlt/common/destination/reference.py
```diff
@@ -344,59 +345,140 @@ def should_truncate_table_before_load_on_staging_destination(self, table: TTable
...
+    @abstractmethod
+    def spec(self) -> Type[DestinationClientConfiguration]:
+        """Returns the destination configuration spec"""
+        ...
 
+    @property
+    @abstractmethod
     def capabilities(self) -> DestinationCapabilitiesContext:
-        """Destination capabilities ie. supported loader file formats, identifier name lengths, naming conventions, escape function etc."""
+        """Returns the destination capabilities"""
```
rudolfix, line 429: this should always return a new instance. there's a lot of code that takes capabilities and then modifies the instance ie. to restrict the file formats or change the preferred format. so I'd say make it explicit method makes it more obvious.

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -25,6 +25,7 @@ class DuckDbBaseCredentials(ConnectionStringCredentials):
     read_only: bool = False  # open database read/write
 
     def borrow_conn(self, read_only: bool) -> Any:
+        # TODO: Can this be done in sql client instead?
```
rudolfix, line 28: no - in loader each thread gets a private sql client and connection. however in duckdb there's just a single connection that can be borrowed. So I need a "singleton" which the instance of configuration is.

I agree that we could implement such singleton separately but this is close to perfect - each configuration describes a particular duckdb database so this mechanism fits here pretty well...

dlt/destinations/impl/duckdb/factory.py
```diff
@@ -0,0 +1,31 @@
...
+    def client_class(self) -> t.Type["DuckDbClient"]:
+        from dlt.destinations.impl.duckdb.duck import DuckDbClient
+
+        return DuckDbClient
+
+    @with_config(spec=DuckDbClientConfiguration, sections=(known_sections.DESTINATION, 'duckdb'), accept_partial=True)
+    def __init__(
+        self,
+        credentials: DuckDbCredentials = None,
+        create_indexes: bool = True,
```
rudolfix, line 28: the default for duckdb is False

dlt/destinations/impl/duckdb/factory.py
```diff
@@ -0,0 +1,31 @@
...
+    @property
+    def client_class(self) -> t.Type["DuckDbClient"]:
+        from dlt.destinations.impl.duckdb.duck import DuckDbClient
+
+        return DuckDbClient
+
+    @with_config(spec=DuckDbClientConfiguration, sections=(known_sections.DESTINATION, 'duckdb'), accept_partial=True)
+    def __init__(
+        self,
+        credentials: DuckDbCredentials = None,
```
rudolfix, line 27: would be better to get connection string or path and create DuckDbCredentials from it. it's constructor should accept

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -95,28 +96,25 @@ class DuckDbCredentials(DuckDbBaseCredentials):
 
     __config_gen_annotations__: ClassVar[List[str]] = []
 
-    def on_resolved(self) -> None:
```
rudolfix, line 98: What is the problem here? `on_resolved` is not called in new model because we allow for partial? but at some point we must make sure that configuration is complete. pls see my other comments

dlt/destinations/impl/duckdb/factory.py
```diff
@@ -0,0 +1,31 @@
...
+    capabilities = capabilities()
+    spec = DuckDbClientConfiguration
+
+    @property
+    def client_class(self) -> t.Type["DuckDbClient"]:
+        from dlt.destinations.impl.duckdb.duck import DuckDbClient
+
+        return DuckDbClient
+
+    @with_config(spec=DuckDbClientConfiguration, sections=(known_sections.DESTINATION, 'duckdb'), accept_partial=True)
```
rudolfix, line 24: OK, we can simplify this even more. 
1. I'd not use @with_config here at. I'd just let people pass initial parameters and do not pull config at all.
2. I'd use with_config not as decorator but within a function that would return a configuration. there's no partial.
so let's imagine we have this in `Destination`:
```
def configuration(initial: TConfig) -> TConfig:
   return with_config(lambda x: x, spec=DuckDbClientConfiguration, sections=(known_sections.DESTINATION, self.destination_name), initial_config=initial)
```
we'd need to extend `with_config` to accept inital config in the decorator BUT if you look in the code there's already mechanims to handle that:
```
# if one of arguments is spec the use it as initial value
                if spec_arg:
                    config = bound_args.arguments.get(spec_arg.name, None)                    
```
if we do that, we do not use partial, we always resolve configuration etc.
where `initial` is coming from? we form it in `__init__` from the aguments.

rudolfix (changes requested):
pls see how I'd simplify configuring the destination even more

dlt/destinations/impl/duckdb/factory.py
steinitzu: Hmm yeah I think that makes sense. So we'd just be leaving resolving config until the last moment when it's needed.

dlt/destinations/impl/duckdb/factory.py
steinitzu: Yeah probably all destinations we should accept either `TCredentials | str` or any native value types.

dlt/destinations/impl/duckdb/configuration.py
steinitzu: Yeah I get it. I ran into a problem trying to copy the config why I added the comment, but it's not a big deal right now I think.   
I agree it's pretty neat like this and nice to not need another global weakref mapping or something.

dlt/destinations/impl/duckdb/factory.py
steinitzu: Question, why use `with_config` and not `resolve_config` directly for this?

rudolfix: you can use `resolve_config`! looks even better

dlt/destinations/impl/duckdb/configuration.py
steinitzu: Ok my fix was a problem for dbt runner so I went back to the original `on_resolved` method, but added this https://github.com/dlt-hub/dlt/blob/sthor%2Fparametrized-destination/dlt/destinations/impl/duckdb/configuration.py#L99-L104 so it doesn't trigger right when the instance is created

dlt/destinations/impl/weaviate/factory.py
```diff
@@ -0,0 +1,30 @@
...
+    @property
+    def client_class(self) -> t.Type["WeaviateClient"]:
+        from dlt.destinations.impl.weaviate.weaviate_client import WeaviateClient
+
+        return WeaviateClient
+
+    def __init__(
+        self,
+        credentials: t.Optional[WeaviateCredentials] = None,
+        **kwargs: t.Any,
```
rudolfix, line 28: maybe we could add vectorizer and module_config?

tests/load/duckdb/test_duckdb_client.py
```diff
@@ -147,12 +148,12 @@ def test_keeps_initial_db_path() -> None:
...
         # new pipeline context took over
         # TODO: restore pipeline context on each call
-        assert conn.credentials.database.lower() != os.path.abspath(db_path).lower()
+        assert conn.credentials._conn_str().lower() != os.path.abspath(db_path).lower()
 
 
 def test_duckdb_database_delete() -> None:
     db_path = "_storage/path_test_quack.duckdb"
-    p = dlt.pipeline(pipeline_name="quack_pipeline", credentials=db_path, destination="duckdb")
+    p = dlt.pipeline(pipeline_name="quack_pipeline", destination=duckdb(credentials=DuckDbCredentials(db_path)))
```
rudolfix, line 156: this should still work. we just need to create the instance ourselves if stirng is passed. I explained that above

rudolfix (changes requested):
this is really close! please remember to deprecate `credentials` argument usage in `Pipeline` `run` and `load` method

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -95,6 +96,13 @@ class DuckDbCredentials(DuckDbBaseCredentials):
 
     __config_gen_annotations__: ClassVar[List[str]] = []
 
+    def is_partial(self) -> bool:
+        partial = super().is_partial()
```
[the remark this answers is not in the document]
steinitzu: Huh? How does that work? :/

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -173,6 +180,9 @@ def _path_from_pipeline(self, default_path: str) -> Tuple[str, bool]:
 
         return default_path, True
 
+    def _conn_str(self) -> str:
```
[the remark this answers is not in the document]
steinitzu: Seems just for motherduck compatiblity it's needed, uses the same method.

dlt/destinations/impl/athena/factory.py
```diff
@@ -0,0 +1,35 @@
...
+        self,
+        query_result_bucket: t.Optional[str] = None,
+        credentials: t.Optional[AwsCredentials] = None,
+        athena_work_group: t.Optional[str] = None,
+        aws_data_catalog: t.Optional[str] = "awsdatacatalog",
+        supports_truncate_command: bool = False,
+        force_iceberg: bool = False,
+        **kwargs: t.Any,
+    ) -> None:
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
steinitzu: Ah this one is a mistake

dlt/destinations/impl/athena/factory.py
```diff
@@ -0,0 +1,35 @@
...
+    @property
+    def client_class(self) -> t.Type["AthenaClient"]:
+        from dlt.destinations.impl.athena.athena import AthenaClient
+
+        return AthenaClient
+
+    def __init__(
+        self,
+        query_result_bucket: t.Optional[str] = None,
+        credentials: t.Optional[AwsCredentials] = None,
```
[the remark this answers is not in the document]
steinitzu: Yes should do this everywhere, don't even need to add anything to deal with different arguments since `resolve_configuration` handles it already.

rudolfix (dismissed):
LGTM! thanks for the docstrings!

PR #764, opened 2023-11-13
adds more name hashes to telemetry

rudolfix:
### Description
We want a few name hashes to telemetry (see docs edit in this PR for details). Motivation: to investigate if dataset, destinations and schema names change or stay constant for particular pipeline name


burnash (approved):
Looks good to me

Issue #766, opened 2023-11-15
ISO date strings don't get loaded with 'date' data type


codingcyclist:
### dlt version

0.3.22

### Describe the problem

We detect ISO-compliant date strings like `2023-11-15` as date objects in `parse_iso_like_datetime` (see [here](https://github.com/dlt-hub/dlt/blob/63c18fed82ad2618ae5f95274ceadf13a9fbe03c/dlt/common/time.py#L30C21-L30C21)), but neither does the `is_iso_timestamp` autodetection function pick them up as dates, nor is there a dedicated `is_iso_date` auto-detection function. 

As a result, sources that yield "native" date objects get loaded with data type `date`, whereas sources that yield ISO-compliant date strings get loaded with data type `text`

### Expected behavior

There should be an `iso_date` autodetection function.

### Steps to reproduce

Load data=[{'date':'2023-11-15'}] into any destination so see that it gets loaded as `text`

### Operating system

Linux

### Runtime environment

Other

### Python version

3.11

### dlt data source

Custom

### dlt destination

Snowflake

PR #767, opened 2023-11-15
Autodetector for ISO date strings

codingcyclist:
### Description
[ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) defines date strings as `YYYY-MM-DD`. We already have autodetection for ISO timestamp strings and detect ISO-compliant date strings like `2023-11-15` as date objects (see [here](https://github.com/dlt-hub/dlt/blob/63c18fed82ad2618ae5f95274ceadf13a9fbe03c/dlt/common/time.py#L30C21-L30C21)). However, neither does the `is_iso_timestamp` autodetection function pick date strings up as dates, nor is there a dedicated `is_iso_date` auto-detection function. 

As a result, sources that yield "native" date objects get loaded with data type `date`, whereas sources that yield ISO-compliant date strings get loaded with data type `text`

This PR adds an `iso_date` autodetection function to the list of default type detectors. This change is fully backward compatible with existing schemas, so that we don't need to bump the schema engine version.

### Related Issues

- Fixes [766](https://github.com/dlt-hub/dlt/issues/766)

### DIFF
--- a/dlt/common/schema/utils.py
+++ b/dlt/common/schema/utils.py
@@ -698,4 +698,4 @@ def standard_hints() -> Dict[TColumnHint, List[TSimpleRegex]]:
 
 
 def standard_type_detections() -> List[TTypeDetections]:
-    return ["iso_timestamp", "iso_date"]
+    return ["iso_timestamp"]
--- a/dlt/common/schema/utils.py
+++ b/dlt/common/schema/utils.py
@@ -698,4 +698,4 @@ def standard_hints() -> Dict[TColumnHint, List[TSimpleRegex]]:
 
 
 def standard_type_detections() -> List[TTypeDetections]:
-    return ["iso_timestamp"]
+    return ["iso_timestamp", "iso_date"]
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ help:
 	@echo "			runs flake and mypy"
 	@echo "		test"
 	@echo "			tests all the components including destinations"
-	@echo "		test-local"
+	@echo "		test-load-local"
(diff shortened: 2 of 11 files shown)


dlt/common/schema/utils.py
```diff
@@ -698,4 +698,4 @@ def standard_hints() -> Dict[TColumnHint, List[TSimpleRegex]]:
 
 
 def standard_type_detections() -> List[TTypeDetections]:
-    return ["iso_timestamp"]
+    return ["iso_timestamp", "iso_date"]
```
rudolfix, line 701: this is the only thing I'd change. I'm not sure that generating dates should be the default. date-like strings are simple and common - IMO automatic conversion to `date` database type is an unexpected behavior.

instead you can always use `add_type_detection` - from within a dlt.source like here https://dlthub.com/docs/general-usage/schema#schema-is-modified-in-the-source-function-body

or just before loading: `source().schema.add_type_detection("iso_date")`.

we'll be releasing dlt soon so if you are OK with it we can do this change here

rudolfix (changes requested):
looks great! I have a question to make `iso_date` default detector, see my comment

dlt/common/schema/utils.py
codingcyclist: I'm fine with that :)

codingcyclist: [91b7293](https://github.com/dlt-hub/dlt/pull/767/commits/91b729308b07c08b5b7578907a5ff387aa74782f)

rudolfix (approved):
LGTM!

PR #768, opened 2023-11-16
Website/docs/30 sql databases

### LINKED ISSUE #720: make clear in docs that we support any sql database as a source
### Feature description

Our docs:
https://dlthub.com/docs/dlt-ecosystem/verified-sources/sql_database
are not clear that we support all sql alchemy dialects and does not give any examples

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

ingesting data from production databases

### Proposed solution

Let's fix it as follows:
1. change title to +50 SQL Databases (or similar)
2. before we get into user guide add a section on supported databases. mention that we support any sql alchemy dialect.
3. take 15 most prominent engines from https://docs.sqlalchemy.org/en/20/dialects/ and list them, then link to this list
4. mention that there are many unofficial dialects and link to duckdb-engine as an example
5. provide more info on how to create a connection string for a given dialect. (just extend existing docs)

### Related issues

_No response_

anuunchin:
### Description
Made some final copy improvements on the SQL databases verified source. 

### Related Issues

- Closes #720

### DIFF
--- a/docs/website/docs/dlt-ecosystem/verified-sources/sql_database.md
+++ b/docs/website/docs/dlt-ecosystem/verified-sources/sql_database.md
@@ -1,4 +1,4 @@
-# SQL Database
+# +30 SQL Databases
 
 :::info Need help deploying these sources, or figuring out how to run them in your data stack?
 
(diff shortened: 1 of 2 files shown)


anuunchin:
Have to add some additional fixes.

Issue #758, opened 2023-11-11
enforce schema and source name to be the same


rudolfix:
**Background**
Behavior above is enforced by the decorators and warning is present for a long time. Since one source can hold only one schema there's no need to have different names - this makes tracing, state management and data lineage way simpler / possible at all

**Tasks**
1. * [ ] enforce the correct behavior in all cases by removing source name from `DltSource` initializer.

PR #769, opened 2023-11-16
source and schema changes

sh-rp:
* Keep ancestry information in schema as the 10 previous hashes
* Remove name from DLTSource initializer and properties.


rudolfix (changes requested):
we probably had some scope misunderstanding!
if you look at here https://github.com/dlt-hub/dlt/issues/758
we should drop the name in `DltSource` not in decorator. the point of the ticket is to use schema name as the source name. keeping the name in decorator which was enforcing the schema name made perfect sense... probably my fault with describing the scope?

you could also link the relevant issues to PR - the template has a section for that

thanks!

decorator should stay intact, you also do not need to change verified sources.

sh-rp:
Ah alright, I was going to ask anyway because it did not seem useful to remove the decorator arg. I have updated the PR and it should be ready for review again @rudolfix

dlt/common/schema/utils.py
```diff
@@ -240,12 +246,18 @@ def compile_simple_regexes(r: Iterable[TSimpleRegex]) -> REPattern:
 
 
 def validate_stored_schema(stored_schema: TStoredSchema) -> None:
+    # exclude validation of keys added later
+    ignored_keys = []
```
rudolfix, line 250: why this is required? first we migrate the dictionary and then we validate the schema. so the engine should be always 7

dlt/common/schema/schema.py
```diff
@@ -37,6 +37,7 @@ class Schema:
     _dlt_tables_prefix: str
     _stored_version: int  # version at load/creation time
     _stored_version_hash: str  # version hash at load/creation time
+    _stored_ancestors: Optional[List[str]] # list of ancestor hashes of the schema
```
rudolfix, line 40: this is not ancestor. this is previous version hash. maybe let's name it like that. we do not have any schema derivation scheme like we have in pydantic models (ie. base schema etc.) this is just a version (probably revision would be better - but it is too late to change it)

rudolfix: _previous_version_hashes or _previous_hashes would be better

dlt/common/schema/utils.py
```diff
@@ -143,8 +143,13 @@ def bump_version_if_modified(stored_schema: TStoredSchema) -> Tuple[int, str]:
         pass
     elif hash_ != previous_hash:
         stored_schema["version"] += 1
+        # unshift previous hash to ancestors and limit array to 10 entries
+        if previous_hash not in stored_schema["ancestors"]:
+            stored_schema["ancestors"].insert(0, previous_hash)
+            stored_schema["ancestors"] = stored_schema["ancestors"][:10]
```
rudolfix, line 149: maybe we could have it in the contants so it is easy to find and change?

dlt/common/validation.py
```diff
@@ -9,7 +9,7 @@
 TCustomValidator = Callable[[str, str, Any, Any], bool]
 
 
-def validate_dict(spec: Type[_TypedDict], doc: StrAny, path: str, filter_f: TFilterFunc = None, validator_f: TCustomValidator = None) -> None:
+def validate_dict(spec: Type[_TypedDict], doc: StrAny, path: str, filter_f: TFilterFunc = None, validator_f: TCustomValidator = None, filter_required: TFilterFunc = None) -> None:
```
rudolfix, line 12: filter_required - what it does? you added new argument without docstrings

dlt/extract/decorators.py
```diff
@@ -162,16 +159,16 @@ def _wrap(*args: Any, **kwargs: Any) -> TDltSourceImpl:
...
                     if rv is None:
-                        raise SourceDataIsNone(name)
+                        raise SourceDataIsNone(schema.name)
                     # if generator, consume it immediately
                     if inspect.isgenerator(rv):
                         rv = list(rv)
 
             # convert to source
-            s = _impl_cls.from_data(name, source_section, schema.clone(update_normalizers=True), rv)
+            s = _impl_cls.from_data(source_section, schema.clone(update_normalizers=True), rv)
```
rudolfix, line 171: 👍

dlt/extract/source.py
```diff
@@ -642,22 +642,17 @@ class DltSource(Iterable[TDataItem]):
     * You can use a `run` method to load the data with a default instance of dlt pipeline.
     * You can get source read only state for the currently active Pipeline instance
     """
-    def __init__(self, name: str, section: str, schema: Schema, resources: Sequence[DltResource] = None) -> None:
-        self.name = name
+    def __init__(self, section: str, schema: Schema, resources: Sequence[DltResource] = None) -> None:
```
rudolfix, line 645: could we swap schema with section? it feels more natural - schema is the most important arg

dlt/extract/source.py
```diff
@@ -642,22 +642,17 @@ class DltSource(Iterable[TDataItem]):
...
-        if self.name != schema.name:
-            # raise ValueError(f"Schema name {schema.name} differs from source name {name}! The explicit source name argument is deprecated and will be soon removed.")
-            warnings.warn(f"Schema name {schema.name} differs from source name {name}! The explicit source name argument is deprecated and will be soon removed.")
-
         if resources:
             self.resources.add(*resources)
 
     @classmethod
-    def from_data(cls, name: str, section: str, schema: Schema, data: Any) -> Self:
+    def from_data(cls, section: str, schema: Schema, data: Any) -> Self:
```
rudolfix, line 655: also here

rudolfix (changes requested):
thx for the tests! I have a few questions and I'd like to rename ancestors to something else... then we merge this because it LGTM

dlt/common/schema/utils.py
sh-rp: in test_upgrade_engine_v1_schema many different schema versions are validated. we could alternatively change that test.

dlt/common/schema/schema.py
sh-rp: done!

dlt/common/schema/utils.py
sh-rp: I'm not sure what you mean?

dlt/extract/source.py
sh-rp: done

dlt/extract/source.py
sh-rp: done

dlt/common/validation.py
rudolfix: maybe rename to `ignored_props_f`. And maybe it is time to consider Pydantic instead? honestly I wrote this code because pydantic handling of TypedDicts was weak (forgot the detaiuls already). 

on the other hand I do not want Pydantic in core deps

dlt/common/schema/utils.py
rudolfix: really I was sure we migrate and then validate. We should validate only the version after migration. So you can change the test.

but you can keep this `filter_required` providing that you add tests for it

rudolfix (changes requested):

rudolfix (approved):
totally LGTM! thanks!

PR #782, opened 2023-11-21
documents schema and data contract

rudolfix:
### Description
Improves data contracts documentation. We still need good code examples how to apply them in a few interesting cases ie. schema inference and then contracts, pydantic models + event processing, contracts on arrow tables

@adrianbr @sh-rp please review. I'm happy to rewrite stuff to make it clearer


sh-rp (approved):
To me this looks very good! One thing we could consider is to mark experimental features better, i am not sure that this test tube symbol is universally understood.

Issue #763, opened 2023-11-13
0.4.1 release notes and master merge


rudolfix:
## Master merge
We have a few bigger pull requests coming in that change internal object models and storage layouts. We want to release them as 0.4.0 version (which is a major release for us). Public APIs will not be impacted or changes will be backward compatible.
1. `devel` will be merged to `master` and 0.3.x released from there
2. docs will be updated from master branch
3. `devel` will collect the 0.4.0 features. docs will be released to devel site
4. we plan to merge everything to master end of November

## Post merge branches and releases
After the 0.4.x merge we will use the following branches
1. `devel` branch will be the default branch to which majority of PRs will be accepted
2. `master` branch will be used for hotfixes to code and documentation, also for publishing blogposts
3. on the release day, version of the library will be bumped on `devel` branch and then `devel` branch will be merged to `master`
4. all library releases happen from `master`. that includes regular releases and hotfix (post) releases
5. we may occasionally do pre-releases from the `devel` branch

Note: we support two version of documentation:
1. current `master` release on https://dlthub.com/docs/intro
2. incoming master / alpha release from `devel` on https://dlthub.com/devel/intro

## `0.4.1` release notes
- the destination adapters moved to separate namespace. **Weaviate** and **Qdrant** are affected:
`from dlt.destinations.weaviate import weaviate_adapter` -> `from dlt.destinations.adapters import weaviate_adapter`
- data and schema contracts (#135 and #594 ) - refactored extract and normalize processing, backward compatible API changes and no storage changes
- trace improvements - trace dumps with new layout and engine version
- extract → normalize data handover difference (normalize storage layout change and bump) (#756) if you have extracted packages that are not yet normalized left from 0.3.x, the 0.4.x will not migrate the data but rather stop.
- Extract, Load and Normalize Info → way more data collected, interface incompatibilities, loadable dictionary representation will change so the schema if such data is loaded with dlt (#757 #754)
- source and schema names must be identical → warning / deprecation converted to error. (#758 )
- schema engine version bump (happened already many times - should have no impact on working dlt instances) (#758)
- object model for destinations and thus the way they are implemented changes to more look as sources (#746 ) to enable implementation of sinks (#752 )
- telemetry changes:
`destination_type`: contains a full module of a destination (dlt.destinations.dummy)
`destination_name`: contains destination name OR fallback to the last part of destination type name (ie. dummy)
[...]

PR #783, opened 2023-11-22
destination config updates

sh-rp:
* Renames destionation_name to destionation_type on DestinationClientConfiguration
* Adds new variables name and environment to DestinationClientConfiguration


sh-rp:
General notes:
* I'm not sure that it makes sense to have the destination name in the config. I think it should be set in code and then you can select a config section based on that name. If you also have the name in the config it might be a bit confusing. Generally I am not quite sure how the setup for a user with multiple named destinations and environments would look like..

dlt/pipeline/configuration.py
```diff
@@ -11,8 +11,8 @@
 class PipelineConfiguration(BaseConfiguration):
     pipeline_name: Optional[str] = None
     pipelines_dir: Optional[str] = None
-    destination_name: Optional[str] = None
-    staging_name: Optional[str] = None
+    destination_type: Optional[str] = None
```
sh-rp, line 14: maybe we should just call this destination?

dlt/destinations/impl/athena/configuration.py
```diff
@@ -23,3 +23,19 @@ def __str__(self) -> str:
             return str(self.staging_config.credentials)
         else:
             return "[no staging set]"
+
+    if TYPE_CHECKING:
```
sh-rp, line 27: 🥔🥔🥔

dlt/common/destination/reference.py
```diff
@@ -449,12 +469,14 @@ def client_class(self) -> Type[TDestinationClient]:
...
     def configuration(self, initial_config: TDestinationConfig) -> TDestinationConfig:
         """Get a fully resolved destination config from the initial config"""
-        return resolve_configuration(
+        config = resolve_configuration(
             initial_config,
             sections=(known_sections.DESTINATION, self.name),
             # Already populated values will supersede resolved env config
             explicit_value=self.config_params,
         )
+        config.name = self.name
```
rudolfix, line 478: why? we pass the `name` to kwargs in init. are you protecting against someone overwriting the name from config? then place `name` as `Final[str]`

dlt/common/pipeline.py
```diff
@@ -92,8 +92,11 @@ class LoadInfo(NamedTuple):
     """A tuple holding the information on recently loaded packages. Returned by pipeline `run` and `load` methods"""
 
     pipeline: "SupportsPipeline"
-    destination_name: str
+    destination_type: str
     destination_displayable_credentials: str
+    destination_name: str
```
rudolfix, line 97: ok for now, but we'll create destination info in next ticket

dlt/destinations/impl/athena/factory.py
```diff
@@ -45,6 +47,8 @@ def __init__(
         super().__init__(
```
rudolfix, line 47: name and environment missing. Also everywhere else

dlt/destinations/impl/athena/factory.py
```diff
@@ -25,6 +25,8 @@ def __init__(
         self,
         query_result_bucket: t.Optional[str] = None,
         credentials: t.Union[AwsCredentials, t.Dict[str, t.Any], t.Any] = None,
+        name: t.Optional[str] = None,
```
rudolfix, line 28: now when I see this, I think we should use `destination_name` everywhere

rudolfix: what about placing the name and environment always at the end of argument list?

dlt/destinations/impl/bigquery/configuration.py
```diff
@@ -10,7 +10,7 @@
 
 @configspec
 class BigQueryClientConfiguration(DestinationClientDwhWithStagingConfiguration):
-    destination_name: str = "bigquery"
+    destination_type: Final[str] = "bigquery"  # type: ignore
```
rudolfix, line 13: interesting option would be to have `destination_name = "bigquery"` as well. so this is more explicit. but up to you! maybe better to mark it as Final in base class and never redefine

dlt/pipeline/configuration.py
rudolfix: I'd keep `_type` to be more explicit. also please link to this PR and add info that we changed the names in this config in here https://github.com/dlt-hub/dlt/issues/763

also please add `on_resolved` and please keep `destination_name` and `staging_name` in the config but display deprecation warnings. Steini added a method to warn on deprecation somewhere in the Pipeline class

rudolfix: there are people that configure destinations via config.

btw. we may keep destination name in the longer run. so in fact we'll create a destination with given type and given name. so let's do that:
- if only name is present or only type is present, we create a destination with name == type
- if they are different we pass both as destination reference.
i think destination reference supports it

rudolfix (changes requested):
thanks for all missing mypy init delcarations...

> I'm not sure that it makes sense to have the destination name in the config. I think it should be set in code and then you can select a config section based on that name. If you also have the name in the config it might be a bit confusing. Generally I am not quite sure how the setup for a user with multiple named destinations and environments would look like..

yes - that may be good idea. I even proposed to make "name" final in the config. so you can completely take it out.
User setup: if destination is entity independent from pipeline we can have many devel or production BigQuery project and users may want to name them and refer to them in many pipelines. 

> Should we also store the destination name in the pipeline state? I am not quite sure why the destination info is stored there in the first place, this is one of these (for me) confusing mechanisms where stuff gets stored somewhere and then dlt does thing that are not really reflected in the code. Anyway if we do not save the name but have config segments that are reliant on the name, then the second run might produce other results.

We should store both name and type of the destination in the pipeline state. You'll also need to bump state version and write a migration. The goal is to be able to retrieve exactly the same configuration for the pipeline (now we need both type and name - because the type is not stored in the config - I think we can keep it that way). 

Right now it is used by streamlit app and CLI to do `dlt.attach` and obtain fully functional pipeline object outside of the original pipeline script.
We do not talk much about this in the docs - right. But this is quite powerful way for doing automations.

> I have implemented the dynamic configuration resolution in the same way it was done before in the from_reference constructor on the destination. This seems to work good, but it is different than was specified in the ticket. Let me know if this also works. the one things that does not work now though, is to have common credentials for all duckdb destinations if those duckdb destinations have different names for example (if we also have other destinations types)

Heh you are right. But putting the config under `[destination]` will work

> I have also made the dataset configurable via the config, I am not sure why this was not possible before, so maybe I am missing something there and will have to revert that..

I think it makes sense.

dlt/common/destination/reference.py
sh-rp: this was the place to set name or type if name was not set, but I put that in an onresolved method now.

dlt/destinations/impl/athena/factory.py
sh-rp: I think it was missing only here, i added it.

dlt/destinations/impl/bigquery/configuration.py
sh-rp: i think i prefer to have it the way it is now

dlt/pipeline/configuration.py
sh-rp: Ok, so we do not deprecated destination_name and staging_name, right?

sh-rp:
Questions:
* Did I understand correctly that destination_name is not to be marked as deprecated?
* The pipeline steps currently support a destination and stage argument which can be a reference or a destination_type, should it be possible to passe type and name there too? Or should we just assume people will pass a destination instance with this info if they really want to use it?

tests/pipeline/test_pipeline_state.py
```diff
@@ -503,3 +513,15 @@ def test_migrate_state(test_storage: FileStorage) -> None:
...
     assert p.dataset_name == "debug_pipeline_data"
     assert p.default_schema_name == "example_source"
+
+    # specifically check destination v2 to v3 migration
+    state_v2 = {"destination": "redshift", "staging": "filesystem", "_state_engine_version": 2}
+    migrate_state(
+        "test_pipeline", state_v2, state_v2["_state_engine_version"], STATE_ENGINE_VERSION  # type: ignore
+    )
+    assert state_v2["destination_name"] == "redshift"
+    assert state_v2["destination_type"] == "redshift"
```
rudolfix, line 523: this looks wrong. it must always be a full path to a module that can be imported.

tests/pipeline/test_pipeline_state.py
```diff
@@ -503,3 +513,15 @@ def test_migrate_state(test_storage: FileStorage) -> None:
...
+    assert "destination" not in state_v2
+    assert state_v2["staging_name"] == "filesystem"
+    assert state_v2["staging_type"] == "filesystem"
```
rudolfix, line 526: same!

tests/pipeline/test_pipeline_state.py
```diff
@@ -485,6 +490,11 @@ def test_migrate_state(test_storage: FileStorage) -> None:
     assert state["_state_engine_version"] == STATE_ENGINE_VERSION
     assert "_local" in state
 
+    # check destination migration
+    assert state["destination_name"] == "dlt.destinations.postgres"
```
rudolfix, line 494: name should be postgres

dlt/common/destination/reference.py
```diff
@@ -453,9 +450,9 @@ def capabilities(self) -> DestinationCapabilitiesContext:
...
 
     @property
-    def name(self) -> str:
+    def destination_name(self) -> str:
         """The destination name will either be explicitely set while creating the destination or will be taken from the type"""
-        return self.config_params.get("name") or self.destination_type
+        return self.config_params.get("destination_name") or self.destination_type
 
     @property
     def destination_type(self) -> str:
```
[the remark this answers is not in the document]
rudolfix: the line below returns wrong type

dlt/common/destination/reference.py
```diff
@@ -453,9 +450,9 @@ def capabilities(self) -> DestinationCapabilitiesContext:
         ...
 
     @property
-    def name(self) -> str:
+    def destination_name(self) -> str:
         """The destination name will either be explicitely set while creating the destination or will be taken from the type"""
-        return self.config_params.get("name") or self.destination_type
+        return self.config_params.get("destination_name") or self.destination_type
```
rudolfix, line 455: use ` def to_name(ref: TDestinationReferenceArg) -> str:` to get name from type. use it everywhere you do that

dlt/common/destination/reference.py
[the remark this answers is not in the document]
rudolfix: maybe let's make it to return Type? then you'll catch all the bugs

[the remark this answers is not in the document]
sh-rp: I am not sure what you mean. We already have the type, no? I have changed it that it will return the full path to the destination module and added / updated tests for this.

sh-rp:
> let's not change the `TDestinationReferenceArg ` but what you can (should do) is to define a NewType for `destination_type` that is a special kind of string that can be a full module name or destination from dlt default module. and then use it `TDestinationReferenceArg ` . you'll catch a lot of problems with that!

This i don't get. A module name can just be a string, no? In that case the linter will not be able to differentiate between some wrong string and a valid module name. But I probably am missing sth here.

dlt/pipeline/pipeline.py
```diff
@@ -1500,26 +1509,36 @@ def managed_state(self, *, extract_state: bool = False) -> Iterator[TPipelineSta
...
+                continue
             if prop in state and not prop.startswith("_"):
                 setattr(self, prop, state[prop])  # type: ignore
         for prop in Pipeline.LOCAL_STATE_PROPS:
             if prop in state["_local"] and not prop.startswith("_"):
                 setattr(self, prop, state["_local"][prop])  # type: ignore
-        if "destination" in state:
-            self._set_destinations(self.destination, self.staging if "staging" in state else None)
+        self._set_destinations(
+            destination=state.get("destination_type"),
```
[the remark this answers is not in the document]
sh-rp: sure, default return value for get is None. Or what do you mean?

tests/destinations/test_destination_name_and_config.py
```diff
@@ -20,7 +20,7 @@ def test_set_name_and_environment() -> None:
     p = dlt.pipeline(pipeline_name="quack_pipeline", destination=duck)
     load_info = p.run([1, 2, 3], table_name="table", dataset_name="dataset")
 
-    assert p.destination.name == "duck1"
+    assert p.destination.destination_name == "duck1"
```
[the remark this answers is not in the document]
sh-rp: There is some stuff i don't understand going on here again. I will put a comment to the test.

sh-rp:
Changing the detination_type property to the full module path changes the output of some other places to this full path as well. I am not sure if we want that everywhere, or if we want to shorten it to only the class name for dlt destinations, you can let me know if the way it is is alright.
I have two questions in the other comments.

tests/destinations/test_destination_name_and_config.py
```diff
@@ -0,0 +1,77 @@
...
+    duck = duckdb(credentials="quack.duckdb", name="duck2")
+    p = dlt.pipeline(pipeline_name="quack_pipeline", destination=duck)
+    load_info = p.run([1, 2, 3], table_name="table", dataset_name="dataset")
+    with p.destination_client() as client:
+        assert client.config.environment == "production"
+        assert client.config.dataset_name == "production_dataset"  # type: ignore
+    assert load_info.environment == "production"
+
+
+def test_pipeline_config() -> None:
```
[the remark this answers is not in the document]
rudolfix: OK probably you are right. We inject the `PipelineConfiguration` into arguments in `dlt.pipeline()` function but we don't do it with `destination_name`:
```
destination = Destination.from_reference(destination or kwargs["destination_name"])
    staging = (
        Destination.from_reference(staging or kwargs.get("staging_name", None))
        if staging is not None
        else None
    )
```
so we just need to make sure that all combinations with `destination_name` and `destination_type` work as expected with `from_reference`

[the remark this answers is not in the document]
sh-rp: so i don't need to change anything here? or should it work with destination_name in the pipelineconfig? this is not clear to me..

sh-rp:
I can't get tests/cli/test_pipeline_command.py::test_pipeline_command_operations to work. If I use the dlt commands regularly from the console, i can create a pipeline, run it and then drop a table. but somehow in the test it fails and I am not sure why..

tests/destinations/test_destination_name_and_config.py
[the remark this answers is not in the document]
sh-rp: i put it in now, i think it should work

tests/destinations/test_destination_name_and_config.py
```diff
@@ -0,0 +1,97 @@
...
+    load_info = p.run([1, 2, 3], table_name="table", dataset_name="dataset")
+
+    assert p.destination.destination_name == "duck1"
+    assert p.destination.destination_type == "duckdb"
+    assert load_info.destination_name == "duck1"
+    assert load_info.destination_type == "duckdb"
+    assert load_info.environment == "production"
+
+
+def test_config_respects_name() -> None:
```
sh-rp, line 31: @rudolfix here is the test for config with name

rudolfix (approved):
this is good now! thanks! a few simple tests that make sure that pipeline props to state works right is needed but I'll do this with new ticket

PR #785, opened 2023-11-23
Fix: ensure accessor typing does not make static type checker error

z3z1ma:
Another quick patch for one of the more common public APIs.

Anywhere you use `dlt.config.value` or `dlt.secrets.value` will throw an error with Pyright (possibly other checkers depending on settings but pyright is the defacto standard in vscode for example which a large number of users use) unless the value is marked optional. This is antithetical to the point of the sentinel value as its presence indicates that a value will in fact contain something. 

So we should use `Any` here. Using Any fixes the below errors.

image

image


rudolfix (approved):
huh! it should be Any form the very start. thx!

Issue #502, opened 2023-07-18
expose dlt commands to notebooks as magics + improve colab experience


rudolfix:
**Background**
See how sql-mesh is doing that. amazing: https://sqlmesh.readthedocs.io/en/stable/reference/notebook/
if we do that we allow our users to easily add verified sources in notebooks and inspect working pipelines. we may also think about some notebook-specific magic ie. showing current config and secrets

**Tasks**
1. * [ ] add magics for `init`, `pipeline` and other commands
2. * [ ] there's a known limitation on Colab - their file editor cannot show files starting with `.` the result is that users do not see the `.dlt` folder with secrets and configs. we should solve this somehow. ie. by `dlt init` that detects running on colab (there's a helper) and then creates a link ie. `_dlt` that is visible
3. * [ ] add a demo colab notebook and documentation to `reference` section of the docs

zilto:
The latest ticket #1860 was closed because feature is not planned.

PR #793, opened 2023-11-29
Add dlt commands as Jupyter Notebook magic commands

Vasilije1990:
### Description
Non-interactive ipython magics that work in the notebooks or in the Ipython shell

- [ x] add magics for init, pipeline and other commands

- [ x]  there's a known limitation on Colab - their file editor cannot show files starting with . the result is that users do not see the .dlt folder with secrets and configs. we should solve this somehow. ie. by dlt init that detects running on colab (there's a helper) and then creates a link ie. _dlt that is visible

- [ ]  add a demo colab notebook and documentation to reference section of the docs -> Note, documentation was added, but there is a question of the colab notebook. 
- [ ] run a notebook in Databrics

Implements #502


dlt/cli/magics.py
```diff
@@ -0,0 +1,288 @@
...
+    @argument('--pipeline_script_path', type=str, help="Path to a pipeline script")
+    @argument('--deployment_method', type=str, help="Deployment method")
+    @argument('--repo_location', type=str, help="Repository location")
+    @argument('--branch', type=str, help="Git branch")
+    @argument('--schedule', type=str, default=None, help="Schedule for the deployment (optional)")
+    @argument('--run_on_push', type=bool, default=False, help="Run on push to Git (optional)")
+    @argument('--run_manually', type=bool, default=False, help="Run manually on Git (optional)")
+    @register_line_magic
+    @line_magic
+    def deploy(self, line:str) ->t.Any :
```
rudolfix, line 164: I think we can comment out deploy part for now. why? I think it will be rarely used from Notebook and if we leave it here we'll need to run a lot of tests

dlt/cli/magics.py
```diff
@@ -0,0 +1,288 @@
...
+            # Assume Databricks' 'display' is a callable with an unknown signature
+            databricks_display = self._shell.user_ns.get("display")
+            if callable(databricks_display):
+                return databricks_display
+            else:
+                raise RuntimeError("Expected a callable for Databricks' display, got something else.")
+        return display  # Assuming 'display' is a predefined callable
+
+    def success_message(self, messages: t.Dict[str, str]) -> HTML:
+        unstyled = messages.get("unstyled")
```
rudolfix, line 55: hmmm interesting. does our `fmt.bold` etc. still works or we should patch it with magic-specific verions?

dlt/common/runtime/exec_info.py
```diff
@@ -62,11 +62,23 @@ def is_github_actions() -> bool:
 
 def is_notebook() -> bool:
     try:
-        return bool(str(get_ipython()))  # type: ignore
+        # Import get_ipython from IPython
+        from IPython import get_ipython
+        # Get the class name of the current IPython instance
+        shell = get_ipython().__class__.__name__
```
rudolfix, line 68: did you have a false positive? so notebook detected while not being in the environment? what was wrong with previous function?

dlt/common/runtime/exec_info.py
```diff
@@ -62,11 +62,23 @@ def is_github_actions() -> bool:
...
+        # Check if the shell is a Jupyter notebook (ZMQInteractiveShell) and return as boolean
+        return shell == 'ZMQInteractiveShell'
     except NameError:
+        # Return False if get_ipython() is not available, indicating a non-IPython environment
         return False
 
 
+def is_ipython() -> bool:
```
rudolfix, line 76: are you able to detect DATABRICKS and add it? then we'll add it to detected environments in telemetry

pyproject.toml
```diff
@@ -77,6 +77,10 @@ pydantic = {version = ">=1.10,<2.0", optional = true}
 adlfs = {version = ">=2022.4.0", optional = true}
 pyodbc = {version = "^4.0.39", optional = true}
 qdrant-client = {version = "^1.6.4", optional = true, extras = ["fastembed"]}
+ipython = "7.34.0"
```
rudolfix, line 80: please do not add those to main dependencies. add them as separate group ie. name it ipython and make this group optional. people will install this group only if they need to work on the magics. see other groups that are done like that.

the other thing - to we really need all those dependencies? maybe just ipython is enough? this we'll probably solve if we can write tests for those magics

dlt/cli/magics.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        args = parse_argstring(self.settings, line)
+        global DEBUG_FLAG
+        try:
+            from dlt.cli._dlt import telemetry_change_status_command_wrapper
+            if args.enable_telemetry:
+                telemetry_change_status_command_wrapper(True)
+            if args.disable_telemetry:
+                telemetry_change_status_command_wrapper(False)
+            if args.non_interactive:
```
rudolfix, line 114: you can always switch non interactive with magics. and this is another way to get rid of all those special cases where notebook is non interactive

rudolfix (changes requested):
thanks for this PR. it looks good! adding magics is really like writing cli with click library. maybe we'll switch from argparse to click.

what is most important in review:
1. please investigate if we really cannot take user input. I was able to take it when launching dlt cli via !dlt ... 
maybe regular Python `input` works. ie like in here. just placing `input` in a cell definitely works
```
from IPython.core.magic import register_line_magic

@register_line_magic
def confirm_magic(line):
    # Splitting the arguments if you have any
    args = line.split()

    # Your custom operation logic
    print("Starting the operation...")

    # Interactive part
    confirmation = input("Do you want to continue? [y/n]: ")
    if confirmation.lower() == 'y':
        print("Operation confirmed.")
        # Place the rest of your operation here
    else:
        print("Operation cancelled.")

# Use the line magic with '%confirm_magic'
```
2. if that does not work - I have an idea how to use non interactive mode and some slight `echo.confirm` modifications instead of all ifs you added. please see my review
3. please see my remarks in `pyproject.toml`

what else we need
1. docs are started but not yet finished
2. we need to figure out if there's any decent way to write a test for a magic. this may be complicated. I can surely help with that

-

dlt/cli/magics.py
```diff
@@ -0,0 +1,288 @@
...
+        shell.register_magics(DltMagics)
+
+        # line_magics = shell.magics_manager.magics['line']
+        # fmt.echo("Registered Line Magics:" + str(line_magics))
+    except NameError:
+        pass
+
+
+def check_notebook_runtime() -> None:
+    fmt.echo("Checking if ipython")
```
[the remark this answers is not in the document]
Vasilije1990: Done

dlt/cli/magics.py
Vasilije1990: Sure, easier for me.

dlt/cli/magics.py
Vasilije1990: It does. This is more pretty, can go back to defaults or move this there. wdyt?

dlt/cli/echo.py
```diff
@@ -46,6 +46,7 @@ def note(msg: str) -> None:
     click.secho("NOTE: " + msg, fg="green")
 
 
+
```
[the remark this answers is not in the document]
Vasilije1990: cleaned up

dlt/__init__.py
```diff
@@ -42,6 +45,12 @@
 TCredentials = _CredentialsConfiguration
 "When typing source/resource function arguments it indicates that a given argument represents credentials and should be taken from dlt.secrets. Credentials may be a string, dictionary or any other type."
 
+try:
+    from dlt.cli.magics import check_notebook_runtime
+    check_notebook_runtime()
```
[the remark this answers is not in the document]
Vasilije1990: fixed, added docstrings

dlt/__init__.py
```diff
@@ -31,6 +31,9 @@
 from dlt.extract.decorators import source, resource, transformer, defer
 from dlt.pipeline import pipeline as _pipeline, run, attach, Pipeline, dbt, current as _current, mark as _mark
 from dlt.pipeline import progress
+from dlt.cli import echo as fmt
```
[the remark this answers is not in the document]
Vasilije1990: understood. That makes sense, felt bad adding stuff there, but removed it now. removed init too, so it doesn't impact it

docs/website/docs/reference/notebook-magic.md
```diff
@@ -0,0 +1,58 @@
...
+  3. `show`
+  4. `info`
+  5. `trace`
+  6. `failed-jobs`
+  7. `drop-pending-packages`
+  8. `load-package`
+  9. `drop`
+  10. `schema`
+
+For more information on these commands check our [CLI documentation](https://docs.dataloadingtool.com/cli/).
```
AstrakhantsevaAA, line 41: this website is not real :D
use just:  [CLI documentation](command-line-interface)

docs/website/docs/reference/notebook-magic.md
```diff
@@ -0,0 +1,58 @@
+---
+title: Command Line Interface for DLT in IPython Notebooks
+description:
```
AstrakhantsevaAA, line 3: add description, it cannot be empty

docs/website/docs/reference/notebook-magic.md
```diff
@@ -0,0 +1,58 @@
...
+
+## `%dlt_version`
+
+Check the DLT version.
+
+```ipython
+%dlt_version
+```
+
+To check an example, please see [notebooks/dlt_version.ipynb](notebooks/dlt_version.ipynb).
```
AstrakhantsevaAA, line 58: create Google Colab, we are not going to store jupyter notebook in repo.
I can create colab and share the edit access with you

dlt/__init__.py
```diff
@@ -32,6 +32,8 @@
 from dlt.pipeline import pipeline as _pipeline, run, attach, Pipeline, dbt, current as _current, mark as _mark
 from dlt.pipeline import progress
 
+
```
rudolfix, line 35: too many empty lines. we actually have black in `devel` branch so those problems will be fixed automatically...

dlt/cli/deploy_command_helpers.py
```diff
@@ -217,8 +218,12 @@ def parse_pipeline_info(visitor: PipelineScriptVisitor) -> List[Tuple[str, Optio
                 if f_r_value is None:
                     fmt.warning(f"The value of `full_refresh` in call to `dlt.pipeline` cannot be determined from {unparse(f_r_node).strip()}. We assume that you know what you are doing :)")
                 if f_r_value is True:
-                    if fmt.confirm("The value of 'full_refresh' is set to True. Do you want to abort to set it to False?", default=True):
-                        raise CliCommandException("deploy", "Please set the full_refresh to False")
+                    if is_notebook():
```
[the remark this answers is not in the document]
rudolfix: do we still need this if?

dlt/cli/init_command.py
```diff
@@ -360,3 +375,6 @@ def init_command(source_name: str, destination_name: str, use_generic_template:
     if dependency_system is None:
         requirements_txt = "\n".join(source_files.requirements.compiled())
         dest_storage.save(utils.REQUIREMENTS_TXT, requirements_txt)
+
+    if is_notebook():
```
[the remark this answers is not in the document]
rudolfix: did you take a look at the above? do you see the `.dlt` for example in Databricks notebook?

dlt/cli/magics.py
rudolfix: keep it here if it is prettier :)

dlt/cli/magics.py
```diff
@@ -0,0 +1,286 @@
...
+from IPython.core.magic import ( Magics, cell_magic,line_cell_magic,line_magic, magics_class, register_line_magic,)
+from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+
+
+@magics_class
+class DltMagics(Magics):
+
+    @property
+    def display(self) ->  t.Any:
+        if os.getenv("DATABRICKS_RUNTIME_VERSION"):
```
rudolfix, line 49: best if we can add databricks detection to exec_info so this can be reused

dlt/cli/magics.py
```diff
@@ -0,0 +1,286 @@
...
+    @magic_arguments()
+    @argument('--enable-telemetry', action='store_true', help="Enables telemetry before command is executed")
+    @argument('--disable-telemetry', action='store_true', help="Disables telemetry before command is executed")
+    @argument('--non-interactive', action='store_true', help="Non interactive mode. Default choices are automatically made for confirmations and prompts.")
+    @argument('--debug', action='store_true', help="Displays full stack traces on exceptions.")
+    @line_magic
+    @register_line_magic
+    def settings(self, line:str)->str:
+        """
+        A DLT line magic command to set global settings like telemetry, debug mode, etc.
```
rudolfix, line 108: we do not capitalize `dlt`. I know you can find wrong capitalization in the code. let's fix them when we see them

dlt/cli/magics.py
```diff
@@ -0,0 +1,286 @@
+
+
+from dlt.common.runtime.exec_info import is_notebook, is_ipython
+from dlt.cli import echo as fmt, echo
+
+import typing as t  # type: ignore
```
rudolfix, line 6: I know this is PITA but if you could place the standard imports over the dlt imports it would be awesome (we'll add isort soon)

dlt/cli/magics.py
```diff
@@ -0,0 +1,286 @@
...
+from hyperscript import h  # type: ignore
```
rudolfix, line 7: does hyperscript comes with iPython or people need to install this dependency to use magics? if the second one better you rewrite the html generating code

also this lib looks rather obscure and was not recently updated...

dlt/cli/magics.py
rudolfix: please remove on interactive option - we force it anyway

dlt/common/configuration/paths.py
```diff
@@ -51,3 +51,32 @@ def get_dlt_data_dir() -> str:
 
 def _get_user_home_dir() -> str:
     return os.path.expanduser("~")
+
```
rudolfix, line 54: two lines is standard spacing

dlt/common/runtime/exec_info.py
rudolfix: bump

dlt/common/runtime/exec_info.py
rudolfix: so you can have `get_ipython()` returning true without actually being a notebook?

pyproject.toml
```diff
@@ -100,6 +100,7 @@ weaviate = ["weaviate-client"]
 pydantic = ["pydantic"]
 mssql = ["pyodbc"]
 qdrant = ["qdrant-client"]
+notebook = ["ipython", "ipywidgets", "jupyter"]
```
rudolfix, line 103: please do not create extra for notebook. people will install that themselves. what we need is a poetry group (optional one) that will be used in order to run tests. so please revert changes to pyproject, add optional poetry group `notebook` instead

rudolfix (changes requested):
looks really decent now! thanks!
we are missing tests. if there is easy way to test magics ie via some virtual ipython interface when
1. place the tests in tests/cli
you can implement tests any way you see fit. here are some hints
1. you need to test if the command wrappers are called with right arguments so you can patch them magic mock and then test the arguments
2. one end to end test (ie. dlt init) that also needs non-interactive mode enabled will suffice to make sure that the whole magic works. you do not need to test each possible sub command of `dlt pipleine .. ` for example

dlt/common/runtime/exec_info.py
```diff
@@ -91,7 +115,7 @@ def airflow_info() -> StrAny:
 def is_airflow_installed() -> bool:
     try:
         with contextlib.redirect_stdout(io.StringIO()), contextlib.redirect_stderr(io.StringIO()):
-            import airflow
+            pass
```
sh-rp, line 118: why was this removed???

dlt/common/runtime/exec_info.py
```diff
@@ -153,7 +177,7 @@ def is_docker() -> bool:
 
     # if that didn't work, try to use proc information
     try:
-        with open("/proc/self/cgroup", mode="r", encoding="utf-8") as f:
+        with open("/proc/self/cgroup", encoding="utf-8") as f:
```
sh-rp, line 180: why is this change required?

rudolfix (changes requested):
my quick manual testing:
- detecting notebook presence was changed and now colab is not detected as notebook
- so the `_dlt` alias is not created
- documentation to magics does not mention that you need import dlt first
- magics are decorated in a way that positional arguments are optional that should not be the case
- would be cool to tell people how to get help ie `%init?` gives you argument for init
@sh-rp @AstrakhantsevaAA still it would be good to merge it somehow... on a notebook it will be really helpful

PR #795, opened 2023-11-29
Fix/test utils

AstrakhantsevaAA:
### Description

Fixes the `make test-common` failure due to a missing environment variable.

### DIFF
--- a/docs/website/docs/conftest.py
+++ b/docs/website/docs/conftest.py
@@ -6,7 +6,7 @@ from dlt.common.utils import set_working_dir
 from dlt.common.configuration.container import Container
 
 # patch which providers to enable
-from dlt.common.configuration.providers import  EnvironProvider, SecretsTomlProvider, ConfigTomlProvider
+from dlt.common.configuration.providers import EnvironProvider, SecretsTomlProvider, ConfigTomlProvider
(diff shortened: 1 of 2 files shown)


docs/website/docs/conftest.py
```diff
@@ -1,16 +1,13 @@
...
 
 from dlt.common.utils import set_working_dir
 from dlt.common.configuration.container import Container
 
 # patch which providers to enable
-from dlt.common.configuration.providers import StringTomlProvider, EnvironProvider, SecretsTomlProvider, ConfigTomlProvider
-from dlt.common.configuration.specs.config_providers_context import ConfigProvidersContext, ConfigProvidersConfiguration
-
-from tests.utils import patch_home_dir, autouse_test_storage, preserve_environ, duckdb_pipeline_location, wipe_pipeline
+from dlt.common.configuration.providers import  EnvironProvider, SecretsTomlProvider, ConfigTomlProvider
```
burnash, line 9:
```suggestion
from dlt.common.configuration.providers import EnvironProvider, SecretsTomlProvider, ConfigTomlProvider
```

burnash (changes requested):
Thanks for the fix! There's an extra space in the import, I've commented on it with a suggested change.

PR #796, opened 2023-11-29
send some traces to the platform from ci

sh-rp:


rudolfix:
@sh-rp do you still need it?

PR #805, opened 2023-12-05
Fix links to source code

AstrakhantsevaAA:
Fixed the problem with source code links in [API reference.](https://dlthub.com/docs/api_reference/common/time).


burnash (dismissed):
Thanks, Alena!

burnash (approved):
LGTM

PR #809, opened 2023-12-08
Clarify docs dev process

deanja:
### Description

- Add docs development guidance for less-technical docs authors.
- Enforce Node engine version from `package.json` , which also helps devs/authors get the right node version.

### Related Issues

Slack chat with Anton

### Additional Context
Hopefully it will save new authors some time in their first contribution attempts.

Some authors will benefit from hints about the nvm/npm/node universe and explicit info about the website structure.  Enforcing minimum Node version helps keep docs authors safe as some OS have a very old Node version installed as system default, which generate loads of errors on `npm install`.


docs/website/README.md
```diff
@@ -1,38 +1,63 @@
 # Website
 
-This website is built using [Docusaurus 2](https://docusaurus.io/), a modern static website generator. The actual content resides in the `docs` folder.
+The website is a Node.js application.
 
+The documentation is generated using [Docusaurus 2](https://docusaurus.io/), a modern static website generator. 
+Docusaurus consumes content from the `./docs` folder (at `./docs/website/docs` in this repo). The content includes:
```
AstrakhantsevaAA, line 6: Also Docusaurus consumes content from the `./blog` (`./docs/website/blog`), please mention this in the doc.
here is a generated blog: https://dlthub.com/docs/blog

docs/website/.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
AstrakhantsevaAA, line 1: do we really need this? we don't fix npm version in package.json

AstrakhantsevaAA (commented):
Thank you for your contribution, great work!

docs/website/README.md
deanja: Thanks for adding that.   Appended a line about blogs.  It would be a bigger rewrite to generalise docs and blogs as 'website content`, so for now a shovel remains a shovel, not a digging implement.

docs/website/.npmrc
deanja: Node version range is already specified in `package,json` in the `engines` section.   

Reason for my change was that without the `engine-strict=true` it just produces a version warning, which for me as a new Node user was lost in a lot of other frightening output.  Even a recent edition of Ubuntu (22.04) only has Node 12.22, which is older than the package's required `>=16.14".  

With `engine-strict=true`  it simply refuses to install and succinctly tells the user that they need a later version of Node.  The website README helps them get the latest stable version.  

I have not considered how making it strict could affect broader CI/CD processes. ie, are there other parts of dlt that depend on Node.  

Background info:
- https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines 
- upgrading node also upgrades its bundled npm, so fixing node version indirectly fixes npm version too.

AstrakhantsevaAA: let's try =)

AstrakhantsevaAA (approved):

PR #807, opened 2023-12-06
[WIP] adds data writer metrics

rudolfix:
### Description
1. Adds metrics for closed files in data writers and collects them when creating StepInfo [WIP]
2. Fixes edge cases with multiple load storages in normalize

PR #813, opened 2023-12-09
allows to run parallel pipelines in separate threads

rudolfix:
### Description
1. Allows to run pipelines in parallel, providing that they run in separate threads.
2. Makes injection containers thread-affine.
3. See `test_parallel_threads_pipeline` for an example.
4. Also updates docs in Performance to show how to use `asyncio` to run pipelines in parallel

Contains content of #807 
1. Adds metrics for closed files in data writers and collects them when creating StepInfo [WIP]
2. Fixes edge cases with multiple load storages in normalize


dlt/common/configuration/container.py
```diff
@@ -40,48 +55,112 @@ def __getitem__(self, spec: Type[TConfiguration]) -> TConfiguration:
...
-        del self.contexts[spec]
+        context = self._thread_context(spec)
+        self._thread_delitem(context, spec)
 
     def __contains__(self, spec: Type[TConfiguration]) -> bool:
-        return spec in self.contexts
+        context = self._thread_context(spec)
+        return spec in context
+
+    def _thread_context(
```
sh-rp, line 83: can't you use pythons thread local context to do all this? https://docs.python.org/3/library/threading.html#thread-local-data

dlt/common/data_writers/writers.py
```diff
@@ -23,6 +23,12 @@ class TFileFormatSpec:
     supports_compression: bool = False
 
 
+class DataWriterMetrics(NamedTuple):
+    file_path: str
+    items_count: int
+    file_size: int
```
sh-rp, line 29: maybe column count? but that is not really important tbh.

docs/website/docs/reference/performance.md
```diff
@@ -249,6 +250,17 @@ The default is to not parallelize normalization and to perform it in the main pr
 Normalization is CPU bound and can easily saturate all your cores. Never allow `dlt` to use all cores on your local machine.
 :::
 
+:::caution
+The default method of spawning a process pool on Linux is **fork**. If you are using threads in your code (or libraries that use threads),
```
sh-rp, line 254: should we add a link to some further explanation of this in the python docs maybe?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dlt"
-version = "0.4.1a0"
+version = "0.4.1a1"
```
sh-rp, line 3: should we add a test somewhere to check that the version number is in sync everywhere?

sh-rp (changes requested):
I think my main question is why not use the python built in thread locals? or do you need to be able to access another threads locals?

dlt/common/data_writers/writers.py
rudolfix: I plan to add elapsed time (start stop). Column count is not known at this moment

dlt/common/configuration/container.py
rudolfix: yeah I know it but when you look at the code, there are exceptions to that behavior. 
1. some type of context are available globally (I use main thread id)
2. there's a special treatment of the executor thread pool. I use a context of a thread that started a pool, not the current thread

so yeah I could use `local()` but there are exceptions so I'd need to keep more dictionaries. or you can force the thread id for local()?

dlt/common/data_writers/writers.py
rudolfix: not during extract. but it is known during normalize. you can however get the column count from the relevant schema...

docs/website/docs/reference/performance.md
rudolfix: good point! could you propose a link? then I'll add it

pyproject.toml
rudolfix: what do you mean. there's just one source of truth for the version and this is the toml file. or you want to have a tests that compares the toml file with the installed package? or maybe that should be a lint step where we force people to `make dev` when versions are not in sync?

dlt/common/data_writers/writers.py
sh-rp: yes elapsed would be cool too!

sh-rp (approved):

PR #817, opened 2023-12-11
dependency changed for dlt installation in local 

rchangdar:
dependency changed for local

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...


rudolfix:
@rchangdar why we should keep wheel files in the root of the repo? maybe that was intended for your personal fork?

rchangdar:
It was not allowing to proceed without installing those wheel files. No personal as i installed everything under special virtual environment.

PR #820, opened 2023-12-12
fix documentation typos

IlyaFaer:


IlyaFaer:
@rudolfix

burnash (approved):
Thanks for spotting this, @IlyaFaer

rudolfix (changes requested):
thx for the fixes and for the review!
the tests for docs examples are not passing. the reason is that @IlyaFaer submitted PR from his fork of dlt and in that case, github actions do not have access to the secrets

I think I already asked to mark mongo tests to not run on fork

@IlyaFaer please do the above :) you can find a @skipifgithubfork decorator and mark the failing tests

in case they are already marked we need to fix the decorator. I know there were some problems with it

thanks :)

.github/workflows/test_doc_snippets.yml
```diff
@@ -24,6 +24,8 @@ env:
   RUNTIME__SLACK_INCOMING_HOOK: ${{ secrets.RUNTIME__SLACK_INCOMING_HOOK }}
   # Mongodb url for nested data example
   MONGODB_PIPELINE__SOURCES__CONNECTION_URL: ${{ secrets.MONGODB_PIPELINE__SOURCES__CONNECTION_URL }}
+
+  IS_FORK: ${{ github.event.pull_request.head.repo.fork }}
```
IlyaFaer, line 28: This var is intended for forks detection, so it's functionality is guaranteed

tests/utils.py
```diff
@@ -247,9 +247,8 @@ def skip_if_not_active(destination: str) -> None:
 
 def is_running_in_github_fork() -> bool:
     is_github_actions = os.environ.get("GITHUB_ACTIONS") == "true"
-    head_ref = os.environ.get("GITHUB_HEAD_REF", "")
-    repo = os.environ.get("GITHUB_REPOSITORY", "")
-    return is_github_actions and ":" in head_ref and not head_ref.startswith(repo.split("/")[0])
```
IlyaFaer, line 252: These variables are not very reliable (like any paths on GitHub):

image

rudolfix (changes requested):
I like it!
Please add comments both to the decorator and to the workflow yaml that explain how they work together - that you need to explicitly define the variable in order for it to work

then I'm good with it and we can merge

IlyaFaer:
@rudolfix, okay, pushed

rudolfix (approved):
LGTM!

PR #827, opened 2023-12-15
Fix Windows lint issue and implement CI lint matrix strategy

### LINKED ISSUE #826: make lint fails on black command on Windows
### dlt version

0.4.1a2

### Describe the problem

When running `make lint` on the `devel` branch on a Windows machine, `black` throws this error:
image

The issue seems to be caused by the boolean ORs (`|`) in the regex for `--exclude`.

The error doesn't show on WSL.

### Expected behavior

_No response_

### Steps to reproduce

On Windows, Python 3.8:

1. Clone `dlt` repo
2. Switch to `devel` branch
3. Run `make dev`
4. Run `poetry shell`
5. Run `make lint`

### Operating system

Windows

### Runtime environment

Local

### Python version

3.8

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

jorritsandbrink:
### Description
1. Fix for issue with `black` command when running `make lint` on Windows: #826 
2. Implementation of matrix strategy for linting CI: #830 

### Related Issues

- Fixes #826
- Implements #830 

### Additional Context
The issue seems to be caused by the boolean ORs (`|`) in the regex for `--exclude`. These ORs can be circumvented when using `--extend-exclude` instead of `--extend`, because `.venv` and `_storage` are included in `.gitignore`, and everything in `.gitignore` is excluded by default ([link](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#exclude)).

> Note: `--exclude` overrides all default exclusions (such as `.gitignore`), while `--extend-exclude` adds on top of the default values. So this new solution potentially excludes more.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -29,6 +29,13 @@ jobs:
         #     os: "ubuntu-latest"
         #   - python-version: "3.10.x"
         #     os: "ubuntu-latest"
+        include:
+          - os: "windows-latest"
+            python-version: "3.8.x"
+          - os: "windows-latest"
+            python-version: "3.9.x"
+          - os: "windows-latest"            
+            python-version: "3.10.x"
 
     defaults:
       run:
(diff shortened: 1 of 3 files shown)


rudolfix (changes requested):
good! let's make sure that lining works on all platforms please extend our `lint` workflow in the same way we run `test_common`. btw. there's additional step at the end that gathers all the sub-workflows from the matrix and fails the whole workflow (matrix_job_required_check) - it needs to be included

rudolfix (approved):
LGTM! thanks for CI improvements!

PR #833, opened 2023-12-17
sync master to devel

rudolfix:
### Description
Take all changes from master to devel before 0.4.x release


rudolfix:
@sh-rp 
conflicts:
.github/workflows/test_doc_snippets.yml - we need both
docs/website/docs/examples/connector_x_arrow/code/load_arrow-snippets.py - we take master
tests/utils.py - we take devel

PR #837, opened 2023-12-18
plugins experiment

sh-rp:
### Description
A prototype of a possible plugin / callback interface to be discussed


dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,55 @@
...
+    #
+    # normalizer callbacks
+    #
+    def filter_row(self, table_name: str, item: TDataItem, **kwargs):
+        return item
+    
+    #
+    # contracts callbacks
+    #
+    def on_schema_contract_violation(self, schema_contract: TSchemaContract, table_name: str, violating_item: TDataItem, **kwargs):
```
sh-rp, line 37: We could repurpose the DataValidationError class which currently only is used for raising in the freeze case to send the specific violation info to this callback.

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,55 @@
...
+        pass
+
+
+_PLUGINS: list[Type[Plugin]] = []
+def register_plugin(plugin: Plugin):
+    global _PLUGINS
+    _PLUGINS.append(plugin)
+
+@configspec
+class PluginsContext(ContainerInjectableContext):
```
sh-rp, line 47: I'm not quite sure wether this is supposed to be used this way, but it works ;)

dlt/extract/extractors.py
```diff
@@ -63,6 +66,7 @@ def __init__(
         self._filtered_columns: Dict[str, Dict[str, TSchemaEvolutionMode]] = {}
         self._storage = storage
         self._caps = _caps or DestinationCapabilitiesContext.generic_capabilities()
+        self._plugins = Container()[PluginsContext].plugins
```
sh-rp, line 69: it would be nicer to have a decorator similar to with_config to inject the plugins here.

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,55 @@
+from typing import Type
+from dlt.common.typing import TDataItem
+from dlt.common.pipeline import SupportsPipeline
+from dlt.common.configuration.specs import ContainerInjectableContext
+from dlt.common.schema.typing import TSchemaContract
+from dlt.common.configuration.specs import configspec
+
+
+class Plugin:
```
sh-rp, line 9: we could consider having multiple plugin base classes which each implement a set of related hooks. This way we can group them in the pluginsprovider and not have to iterate all of the for filtering rows, only the relevant ones. tbd.

rudolfix:
@sh-rp I like it but IMO it is not going far enough :)
We have a pattern to pass such things to the dlt.pipeline. examples: `destination`, `progress` I'd see that we could add this as `callbacks` or `hooks`.

You can provide them as we provide destination and progress (but we accept a list as well, btw. we we'll probably accept a list of destinations and sinks so this will be unified)
- a name or a list of names
- a module or list of modules 
- an instance or instance of the class

Registration: we dump it for now or we reuse `register_plugin` to register modules in order to use a short name. ie we could register
- destinations
- progress
- hooks
etc. and then use them ie destination="my_custom_destination" without a need to do module. could be useful in bigger projects :)
the registry must be part of the Container! (with a global scope - like config providers)

now to the implementation:
- why this is scoped to the step? I'd probably instantiate those with the pipeline and let them manage state via step start and step end events
- there will be issues with keeping the state. there's cross process boundary in normalizer so you'd need to picke and pass all plugins to the normalizer and them move them back and merge the state :/ so for example if someone will use the plugin class to store the data items that violate the contract, bad things can happen
- on the other hand using many normalize processes is not that often so we could ignore it
- let's specify the exact interface we need for data contracts. that will be our initial use case. we must somehow fit it. the problem is that current implementation will filter out the bad data pro-actively... we must do a code review

closing remarks:
1. let's implement tracking in the same way! so additional track modules are passed to pipeline same way as the above. or maybe a tracking module should be a case of a plugin? WDYT?
2. let's give ourselves some time to thing that through

sh-rp:
I think it would be cool to have different plugin types like written above. But I can also implement this "hooks" argument for now if you like. Probably those hooks plugins should also have config section support similar to the destinations? But we can always provide this at a later point.

Your question about the implemenation:
* I think scoping it to the steps is good, because if you keep local state in the plugin an try to carry it over from extract to normalize it will fail if only one the normalize step is run. if we want the callback plugins to be able to have proper state, we have to add it to the load package. But maybe this is overthinking it...
* With regards to the cross boundary problem, I would say we can have two types of hooks. One that is marked as a hook that may not access local state but must be a self contained function. For example the custom row filter could be one such hook. We can make it clear in the docs that using the row filter with local state will only work if the workers are set to 
* Then in the case of the on_contract_violation hook we could create some kind of queue that will send all the violating items to the main process and call the hook there, and the user can consume the items there. 
* To have a custom data contract that does something more than what we have implemented, the easiest approach would be to have the user not use the built in contracts and implement some custom contract in the filter_row method of a plugin. Alternatively, our implementation of the contracts could be somehow implemented as plugin and made extendable, but I don't really think it makes sense if we do not develop a plugininterface that can also modify state and schema.

dlt/extract/extractors.py
```diff
@@ -43,6 +46,7 @@ class ExtractorConfiguration(BaseConfiguration):
         _caps: Optional[DestinationCapabilitiesContext] = None
 
     @with_config(spec=ExtractorConfiguration)
+    @with_plugins()
```
rudolfix, line 49: you do not need this decorator, `with_config` will do that for you. there's a special config provider `ContextProvider` that will inject the `PluginContext` into `_plugins`. exactly like `_caps`

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,99 @@
...
+        elif isinstance(plugin, Plugin):
+            pass
+        else:
+            raise TypeError(
+                f"Plugin {plugin} is not a subclass of Plugin, nor a Plugin instance, nor a plugin"
+                " name string"
+            )
+        return plugin
+
+    def setup_plugins(self, plugins: TPluginArg, pipeline: SupportsPipeline) -> None:
```
rudolfix, line 35: why we need `pipeliene` here? I'd add it only when we need it for a first time

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,99 @@
...
+            return
+        if not isinstance(plugins, Iterable):
+            plugins = [plugins]
+        for p in plugins:
+            resolved_plugin = self._resolve_plugin(p, pipeline)
+            self._plugins.append(resolved_plugin)
+            if isinstance(resolved_plugin, CallbackPlugin):
+                self._callback_plugins.append(resolved_plugin)
+
+    def on_step_start(self, step: str) -> None:
```
rudolfix, line 46: we should pass `SuppportsPipeline` here

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,99 @@
...
+            p.step = self.steps[-1] if self.steps else None
+
+    #
+    # callback interfaces
+    #
+
+    #
+    # extraction callbacks
+    #
+    def on_extractor_item_written(self, item: TDataItem, **kwargs: Any) -> None:
```
rudolfix, line 65: hmmmm we have plenty of ways to put any kind of python function at the end of pipeline. so I do not see use for this one.

if you want something really useful: add a callback when we have a state change on the load package in `load.py` (main use case: control retry and fail messages ie by stopping the loading)

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,99 @@
...
+    def on_schema_contract_violation(
+        self,
+        error: DataValidationError,
+        **kwargs: Any,
+    ) -> None:
+        for p in self._callback_plugins:
+            p.on_schema_contract_violation(error, **kwargs)
+
+
+def with_plugins() -> Callable[[TFun], TFun]:
```
rudolfix, line 89: this wont' be needed. `with_config` will suffice

dlt/common/plugins/reference.py
```diff
@@ -0,0 +1,60 @@
+from typing import Type, Union, List, Any
+from dlt.common.typing import TDataItem
+from dlt.common.configuration import resolve_configuration
+from dlt.common.configuration.specs.base_configuration import BaseConfiguration, configspec
+from dlt.common.schema.exceptions import DataValidationError
+
+
+class SupportsCallbackPlugin:
```
rudolfix, line 8: maybe it should be a Protocol?

dlt/normalize/items_normalizers.py
```diff
@@ -49,20 +51,24 @@ def __call__(
 
 
 class JsonLItemsNormalizer(ItemsNormalizer):
+    @with_plugins()
```
rudolfix, line 54: you need to pass plugins to normalizer in the same way we pass `DestinationCapabilities`. I just wonder if they are pickable

rudolfix: this may run in a process that will be spawned. You cannot expect that, for example, pipeline instance is still present. list of plugins must be self contained. and best if it is stateless (probably impossible)

dlt/pipeline/pipeline.py
```diff
@@ -170,6 +171,24 @@ def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
     return _wrap  # type: ignore
 
 
+def with_plugins_step() -> Callable[[TFun], TFun]:
+    def decorator(f: TFun) -> TFun:
+        @wraps(f)
+        def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
+            plugins_context = self._container[PluginsContext]
```
rudolfix, line 178: this must be implement the same way `with_state` is implemented. you need to inject `PluginContext` in the container (so you create instance here, pass a list of plugins, do the setup, I hope it is not too heavy.)

btw. I still do not understand if `Plugin` instance is supposed to survive steps? so for example if someone is sending contract violations to a kafka topic, the Kafka client should be created for each step?

dlt/pipeline/pipeline.py
```diff
@@ -314,6 +335,11 @@ def __init__(
         # initialize pipeline working dir
         self._init_working_dir(pipeline_name, pipelines_dir)
 
+        # setup plugins context
+        plugins_ctx = PluginsContext()
```
rudolfix, line 339: this should be in `with_plugin_step` decorator

pipeline.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+@configspec
+class MyPluginConfig(PluginConfig):
+    second_name: str
+
+class MyPlugin(CallbackPlugin):
+
+    NAME: str = "my_plugin"
```
rudolfix, line 16: those should be optional.
NAME should come from class name (by default)
SPEC should be None or `BaseConfiguration` by defaiult. you could also put `with_config` on an __init__ functions so you do not need any formalized way to pass SPEC. user does that or not.

rudolfix (changes requested):
I think we are moving in a good direction! instantiation of the plugins may be simplified though. Also the way you pass it to a normalize worker is wrong. see my review

I'll verify and push all relevant places for a contract violation callback in a commit

rudolfix (changes requested):
problem with pydantic validator: when we ignore extra properties, there's no way (right now) to know that it happened. it is possible to detect those by doing some kind of trickery (Pydantic 2.0 only).

1. we can ignore the problem
2. we can always make extra fields `forbid` and remove them ourselves in the validator exception (we do it with `discard_row` already)
3. we can implement some trickery ie. here https://stackoverflow.com/questions/69617489/can-i-get-incoming-extra-fields-from-pydantic

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,99 @@
...
+    # extraction callbacks
+    #
+    def on_extractor_item_written(self, item: TDataItem, **kwargs: Any) -> None:
+        for p in self._callback_plugins:
+            p.on_extractor_item_written(item, **kwargs)
+
+    #
+    # normalizer callbacks
+    #
+    def filter_row(self, table_name: str, item: TDataItem, **kwargs: Any) -> TDataItem:
```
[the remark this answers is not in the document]
sh-rp: We don't need this now, but my general hope is to come up with some plugin interface that hooks into as many places as possible so users can contribute really useful plugins to the community and this not only is a feature for individual pipeline development.

dlt/common/plugins/plugins.py
sh-rp: I removed it for now

pipeline.py
sh-rp: the way it is built now, the plugin name is part of the sections tuple in the config resolution. we ofc can also leave it up to the user, but as I wrote above, my hope is that these plugins can somehow be share among our users and then configured like destinations without the user having to even open the plugin code (maybe only to register the plugin)

dlt/pipeline/pipeline.py
sh-rp: I built in a way now that is survives the steps. I also don't really know which approach is right here, but we can do it like this for now and see where we get when implementing new stuff.

dlt/extract/extractors.py
```diff
@@ -249,7 +282,11 @@ def _apply_contract_filters(
                 if filtered_columns.get(rename_mapping[name]) is not None
             ]
             if removed_columns:
+                name = rename_mapping[removed_columns[0]]
+                mode = filtered_columns.get(name)
                 item = pyarrow.remove_columns(item, removed_columns)
+                # TODO: should we maybe emit dictionaries here? or are arrow tables better? not sure...
```
sh-rp, line 288: should we send the arrow table here or emit dicts?

sh-rp:
I have added some basic pydantic tests, the plugins callback for pydantic validation is not there yet though. My thoughts on this:
- [ ] We could send some kind of context object when traversing through the steps in the pipe (similar to this meta object) which includes the plugin context. I am not sure how many parts of the code this will touch though
- [ ] I have a fairly strong feeling we should not be doing any schema contracts stuff in pydantic at all. Either we do a full proper implemention (if at all possible) or we pydantic for validating structure and field level validators including a specialized exception for failures on that, but do the rest the same way we do for the other data. This would make documentation and usage much easier than having special behavior in this case.

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@configspec
+class PluginsContext(ContainerInjectableContext, SupportsCallbackPlugin):
+    def __init__(self, main: bool = True) -> None:
+        self._plugins: List[Plugin[BaseConfiguration]] = []
+        self._callback_plugins: List[CallbackPlugin[BaseConfiguration]] = []
+        self._initial_plugins: TPluginArg = []
+        self._main = main
+        self._queue: Optional[queue.Queue[Any]] = None
```
sultaniman, line 43: I think here you might potentially have linting issues because `queue.Queue[Any]` is not providing generic way of doing thus mypy might complaint that `Queue is not subscriptable....`
So I did the following workaround
```py
# Please read more at https://mypy.readthedocs.io/en/stable/runtime_troubles.html#not-generic-runtime
T = TypeVar("T")
if TYPE_CHECKING:

    class _Queue(Queue[T]):
        pass

else:

    class _Queue(Generic[T], Queue):
        pass


class BaseQueue(_Queue[T]):
    pass
```

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,143 @@
...
+    def __getstate__(self) -> Dict[str, Any]:
+        return {"plugins": self._initial_plugins, "queue": self._queue}
+
+    def __setstate__(self, d: Dict[str, Any]) -> None:
+        self.__init__(False)  # type: ignore[misc]
+        self.setup_plugins(d["plugins"])
+        self._queue = d["queue"]
+
+    def process_queue(self) -> None:
+        assert self._main
```
sultaniman, line 73: Maybe adding a message to assert will be helpful for debugging, something like `Expected to run in the main thread`

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,143 @@
...
+            if isinstance(resolved_plugin, CallbackPlugin):
+                self._callback_plugins.append(resolved_plugin)
+
+        if self._plugins and self._main:
+            manager = mp.Manager()
+            self._queue = manager.Queue()
+
+    def get_plugin(self, plugin_name: str) -> Optional[Plugin[BaseConfiguration]]:
+        for p in self._plugins:
+            if p.NAME == plugin_name:
```
sultaniman, line 102: are we case sensitive here or should we not?

dlt/common/plugins/plugins.py
```diff
@@ -0,0 +1,143 @@
...
+                return p
+        return None
+
+    #
+    # Main Pipeline Callbacks
```
sultaniman, line 107: Is it interesting to us to collect statistics about popular plugins in the future?

dlt/common/plugins/reference.py
```diff
@@ -0,0 +1,52 @@
...
+from dlt.common.configuration import resolve_configuration
+from dlt.common.configuration.specs.base_configuration import BaseConfiguration, configspec
+from dlt.common.schema.exceptions import DataValidationError
+from dlt.common.pipeline import SupportsPipeline
+
+
+TPluginConfig = TypeVar("TPluginConfig", bound=BaseConfiguration, covariant=True)
+
+
+class SupportsCallbackPlugin:
```
sultaniman, line 12: Is this an interface, if so should be `Protocol` or `ABC`?

dlt/common/runners/pool_runner.py
```diff
@@ -56,13 +57,15 @@ def create_pool(config: PoolRunnerConfiguration) -> Executor:
...
-    return NullExecutor()
+    return CurrentThreadExecutor()
 
 
 def run_pool(
     config: PoolRunnerConfiguration,
     run_f: Union[Runnable[TExecutor], Callable[[TExecutor], TRunMetrics]],
+    perdiodocal_callables: List[Callable[[], None]] = None,
 ) -> int:
+    perdiodocal_callables = perdiodocal_callables or []
```
sultaniman, line 68: Is it a typo here `perdiod>>o<<cal_callables` or it is intended?

dlt/extract/extractors.py
sultaniman: Is it possible to freeze them before sending?

dlt/pipeline/pipeline.py
```diff
@@ -172,6 +173,40 @@ def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
...
 
+def with_plugins() -> Callable[[TFun], TFun]:
+    def decorator(f: TFun) -> TFun:
+        @wraps(f)
+        def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
+            is_new_context = self._plugin_ctx is None
+            if self._plugin_ctx is None:
+                self._plugin_ctx = PluginsContext()
+                self._plugin_ctx.setup_plugins(self.plugins)
+                self._plugin_ctx.on_start(self)
```
sultaniman, line 184: what is step and how `on_step_start` differs from `on_start` should it be something like `on_init`?

dlt/pipeline/pipeline.py
```diff
@@ -172,6 +173,40 @@ def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
     return _wrap  # type: ignore
 
 
+def with_plugins() -> Callable[[TFun], TFun]:
+    def decorator(f: TFun) -> TFun:
+        @wraps(f)
+        def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
+            is_new_context = self._plugin_ctx is None
+            if self._plugin_ctx is None:
+                self._plugin_ctx = PluginsContext()
```
sultaniman, line 182: Also can `PluginsContext` be a context manager which calls some of callbacks?

dlt/pipeline/pipeline.py
```diff
@@ -796,6 +847,12 @@ def last_trace(self) -> PipelineTrace:
             return self._last_trace
         return load_trace(self.working_dir)
 
+    def get_plugin(self, plugin_name: str) -> Any:
```
sultaniman, line 850: I think this should something like `Optional[TPlugin]`?

dlt/common/plugins/reference.py
```diff
@@ -0,0 +1,52 @@
...
+    def on_schema_contract_violation_safe(self, error: DataValidationError, **kwargs: Any) -> None:
+        pass
+
+
+class Plugin(Generic[TPluginConfig]):
+    NAME: Optional[str] = None
+    SPEC: Type[TPluginConfig] = BaseConfiguration  # type: ignore[assignment]
+
+    def __init__(self) -> None:
+        self.step: Union[str, None] = None
```
sultaniman, line 40: Maybe `Optional[str]`?

sh-rp:
If we keep this multiprocessing queue approach, we will have to disable sending messages to the mp queue if there is no plugin active that will actually consume the message. I have some ideas for this so just keeping this here as a reminder.

Issue #757, opened 2023-11-11
refactor `NormalizeInfo` and `LoadInfo`


rudolfix:
**Background**
`NormalizeInfo` and `LoadInfo` do not hold all relevant tracking information from normalize and load step. The biggest issue being not aware of many load packages that can be processed in a single step. In this ticket we apply similar structure to the step infos.

Note: each step info:
- must have a human readable representation with several verbosity levels.
- must generate an object that is dlt friendly to load
- must be pickable without any weird and complicated behaviors

`ExtractInfo` must behave exactly the same (see #754 )

**Tasks**
1. * [x] all step infos must contain `pipeline`, `started_at`, `load_ids` fields which works exactly like in `LoadInfo`
2. * [x] we add `completed_at` DateTime to have a real moment processing was completed (LoadInfo cheats by using now)
3. * [x] `load_packages` holds the information on load packages - like in `LoadInfo`, each step deals with load packages (even if they are represented by different tuple types) - see #756 
4. * [x] when step info is collected after an exception it should contain **only** information that actually happen and got committed - see `LoadInfo`

`LoadInfo`
1. * [ ] use #751 destination info to hold destination and staging infos, convert current fields to properties to keep the interface but not duplicate data
2. * [ ] dataset name may be different for each `load_id` so extend load info to keep this information
3. * [x] add schema name and hash per `load_id`
5. * [ ] remove `elapsed` from `LoadJobInfo`. Instead implement proper metrics for jobs generated in `load.py`. We should get start, stop times of the jobs initially. **mind that you need to persist the metrics as this step is not atomic and may be restarted**
6. * [x] make sure that when load step raises exception we collect partial `LoadInfo` (I think it is implemented).

when collecting partial `LoadInfo`
1. * [x] include schema update ONLY if it got committed to db
2. * [ ] ~include a flag that schemas were actually updated~
- we'll need to (possibly) store some information (per load id) to survive restarts...

`NormalizeInfo`
1. * [x] add list of normalized packages like in LoadInfo
2. * [x] collect row counts per load_id and per file, keep current row count as property that you aggregate from data per file
4. * [x] collect processing time per load id and file
5. * [x] make sure that when norm step raises exception we collect partial info
- we may collect partial information via files (normalize is multiprocess)

Issue #754, opened 2023-11-10
Implement ExtractInfo


rudolfix:
**Background**
extract step does not generate any execution / tracing info, let's fix that. Should not be started before #757 is implemented and common interface of all step infos is clear

**Tasks**
1. * [x] implement `ExtractInfo` with `asdict` and `asstr` methods like the other infos. `asdict` must generate `dlt` friendly object that can be loaded into a relational structure, `asstr` must be friendly to our `pipeline info` command that displays traces
2. * [x] extract step may extract many sources. each extracted source is a separate operation and `ExtractInfo` must collect all of them.
3. * [x] implement all common fields as per #757 

**extract info should contain the following information**
in extract info per source
- * [x] source name and schema name (they must be identical - we'll enforce that see #758 ), load_id (see #756 ) 
- * [x] a DAG of selected resources (by name - they are unique). mind that source has a method that generates DAGs
- * [x] for each source and resource a list of applied hints
- * [x] for each resource collect metrics: the count of the items, number of written bytes, the counts of extracted items per table + bytes, user defined metrics (ticket icoming)
- * [x] for each table: the list of extracted files and their sizes
- * [x] elapsed time for each resource
- * [ ] resource state, source state (this should be optional, maybe extractor config)
- * [ ] resource arguments (if created with decorator) - we have that partially implemented!
- * [ ] source arguments (if created with decorator) - to be implemented

**Implementation**
You must be able to get all the data above after execution is finished. You must be able to collect partial info also when we have an exception during extract (info is part of exception)

users can add custom metrics to resources that will be part of extract info. please read #755 for some implementation hints. we can collect metrics in organized way even before custom metrics are added

PR #839, opened 2023-12-18
adds extract and normalize traces

rudolfix:
### Description
Implements #754 
Implements #757


dlt/common/pipeline.py
```diff
@@ -41,12 +41,22 @@
...
 from dlt.common.utils import RowCounts, merge_row_counts
 
 
+class _StepInfo(NamedTuple):
+    pipeline: "SupportsPipeline"
+    loads_ids: List[str]
+    """ids of the loaded packages"""
+    load_packages: List[LoadPackageInfo]
+    """Information on loaded packages"""
+    started_at: datetime.datetime
```
sh-rp, line 56: This should have a finished_at entry, no? currently I am getting a value here for the load step I think, but not for the others.

dlt/extract/extract.py
```diff
@@ -170,6 +175,64 @@ def __init__(
...
+                        )
+                    continue
+                hints[name] = hint
+
+        return {
+            "schema_name": source.schema.name,
+            "job_metrics": {job.job_id(): metrics for job, metrics in job_metrics.items()},
+            "table_metrics": table_metrics,
+            "resource_metrics": resource_metrics,
+            "dag": source.resources.selected_dag,
```
sh-rp, line 232: so from this we can see which resources where selected during pipeline run, right? But the info which table stems from which resource is not in here, is it?

tests/load/pipeline/test_replace_disposition.py
```diff
@@ -313,6 +315,16 @@ def static_items():
     def yield_none():
         yield
 
+    @dlt.resource(name="main_resource", write_disposition="replace", primary_key="id")
+    def no_yield():
+        # this will not yield even once
+        if False:
```
sh-rp, line 321: good catch that I missed this one

dlt/extract/extract.py
rudolfix: you can get this information from the schema (there's `resource` tag for each table). also it won't be too hard to add this info.
I'd merge the stuff as it is. or make a clear request that you want it :) @sh-rp

dlt/common/pipeline.py
rudolfix: OK I will add it! for extract and normalize it is easy because those steps are atomic so I can just use now() as finished at

rudolfix:
@sh-rp I'll be adding more tests mostly for the shapes of the data to this PR

PR #854, opened 2023-12-23
moves destination adapters to `dlt.destination.adapters`

rudolfix:
### Description
This is breaking change :/ that got slipped. Some modules were fixed, some not (weaviate example does not have tests)

### DIFF
--- a/dlt/destinations/adapters.py
+++ b/dlt/destinations/adapters.py
@@ -1,0 +1,6 @@
+"""This module collects all destination adapters present in `impl` namespace"""
+
+from dlt.destinations.impl.weaviate import weaviate_adapter
+from dlt.destinations.impl.qdrant import qdrant_adapter
+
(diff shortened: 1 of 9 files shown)


burnash:
@rudolfix what is the rationale for moving adapters to dlt.destination.adapters? Could you elaborate on the background?

rudolfix:
@burnash we could not keep backward compatibility because the destinations are not modules but classes now (so we can parametrize them easily). 
this makes it impossible to import ie from `dlt.destinations.weaviate`. callable modules are pure hacks and I'm still not 100% sure it is possible to import from them.

I think putting adapters in a separate namespace is a good compromise now. I expect all destinations to have adapters (ie. #855 ). the alternative is to import form `impl` (which someone did to Qdrant)

This slipped through. In the future we must be more aware of non compatible changes...

PR #863, opened 2023-12-28
corrected data type

### LINKED ISSUE #862: `from_db_type` method in `MsSqlTypeMapper` class references non-existing data type
### dlt version

0.4.1

### Describe the problem

The `from_db_type` method in class `MsSqlTypeMapper`  refers to a `numeric` data type that isn't used for `mssql`:

```python
def from_db_type(
    self, db_type: str, precision: Optional[int], scale: Optional[int]
) -> TColumnType:
    if db_type == "numeric":
        if (precision, scale) == self.capabilities.wei_precision:
            return dict(data_type="wei")
    return super().from_db_type(db_type, precision, scale)
```

`MsSqlTypeMapper` uses `decimal` in `sct_to_dbt` and `dbt_to_sct`, not `numeric`. Because of this, the `wei` data type is never returned.

Perhaps this method has been copied from `from_db_type` in `PostgresTypeMapper`, for which the method is correct, because `postgres` does actually use the `numeric` type.

### Expected behavior

_No response_

### Steps to reproduce

N/A

### Operating system

Windows

### Runtime environment

Local

### Python version

3.8

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

jorritsandbrink:
### Description
This PR corrects the data type used in the `from_db_type` method in class `MsSqlTypeMapper` which referred to a `numeric` data type that isn't used for `mssql`.

### Related Issues

- Fixes #862 

### Additional Context
I also removed the `wei` exclusion for `mssql` in `test_get_storage_table_with_all_types`. With the corrected data type, this test passes without the exclusion.

### DIFF
--- a/dlt/destinations/impl/mssql/mssql.py
+++ b/dlt/destinations/impl/mssql/mssql.py
@@ -76,7 +76,7 @@ class MsSqlTypeMapper(TypeMapper):
     def from_db_type(
         self, db_type: str, precision: Optional[int], scale: Optional[int]
     ) -> TColumnType:
-        if db_type == "numeric":
+        if db_type == "decimal":
             if (precision, scale) == self.capabilities.wei_precision:
                 return dict(data_type="wei")
         return super().from_db_type(db_type, precision, scale)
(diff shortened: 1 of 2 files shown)


rudolfix (approved):
LGTM!

PR #869, opened 2024-01-01
Enhancements in Filesystem Configuration

Pipboyguy:
Re-submission of [Enhancements in Filesystem Configuration](https://github.com/dlt-hub/dlt/pull/856) after changing default branch (see https://github.com/dlt-hub/dlt/pull/856#pullrequestreview-1795937970)


Pipboyguy:
Documentation on usage added to filesystem destination

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -52,39 +52,66 @@ class FileItem(TypedDict, total=False):
...
+        fs_kwargs.update(config.kwargs)
+
+    if config.client_kwargs is not None:
+        fs_kwargs["client_kwargs"] = config.client_kwargs
+
     if proto == "s3":
-        fs_kwargs.update(cast(AwsCredentials, config.credentials).to_s3fs_credentials())
+        # we are careful not to override the client_kwargs key when updating fs_kwargs
+        s3_credentials = cast(AwsCredentials, config.credentials).to_s3fs_credentials()
+        if "client_kwargs" in fs_kwargs and "client_kwargs" in s3_credentials:
```
sh-rp, line 105: would it not be possible to do the client_kwargs update after the if-clause thus having this code only once? you can extract the apropriate credentials dictionary for gcs also and use that. not super important, but might save some headache in future fs implementations

tests/load/filesystem/conftest.py
```diff
@@ -0,0 +1,49 @@
+import os
```
sh-rp, line 1: i think @rudolfix wanted to have this in a utils.py file

tests/load/filesystem/test_filesystem_common.py
```diff
@@ -30,7 +30,12 @@ def test_filesystem_configuration() -> None:
         == Union[AzureCredentialsWithoutDefaults, AzureCredentials]
     )
     # make sure that only bucket_url and credentials are there
```
sh-rp, line 32: outdated comment

sh-rp (changes requested):

tests/load/filesystem/test_filesystem_common.py
Pipboyguy: Fixed in 7ecae7f7

tests/load/filesystem/conftest.py
Pipboyguy: Sorry about that! Fixing now

Pipboyguy: Fixed in f137adfff6e693cabfe16dbaeead46de7b313927

dlt/common/storages/fsspec_filesystem.py
Pipboyguy: I believe I simplified `fsspec_from_config` by moving the credential logic to `GcpCredentials`. The function became to unwieldy. Hopefully it is more readable now

sh-rp (approved):

Issue #881, opened 2024-01-09
Add an example of variant columns to the docs


burnash:
### Feature description

To illustrate how [variant columns](https://dlthub.com/docs/general-usage/schema#variant-columns) work, let's add an example to this section. 

The example should show two cases:
- the data that couldn't be coerced, so the variant column is created
- list some example of successful coercion.

sultaniman:
Relevant updates have been merged.

PR #883, opened 2024-01-09
Update field type coercion with more specific examples

sultaniman:
In this PR I update column variance related type coercion and factors contributing related to introduction of a new column for the same field but with a different type.
Related to https://github.com/dlt-hub/dlt/issues/881


docs/website/docs/general-usage/schema.md
```diff
@@ -127,7 +127,46 @@ Postgres ignore it when creating tables.
...
+Let's consider our [getting started](../getting-started#quick-start) example with slightly different approach,
+where `id` is an integer type at the beginning
+
+```py
+data = [
+  {"id": 1, "name": "Alice"}
+]
+```
+
+once pipeline runs we will have the following schema
```
burnash, line 141:
```suggestion
once pipeline runs we will have the following schema:
```

docs/website/docs/general-usage/schema.md
```diff
@@ -127,7 +127,46 @@ Postgres ignore it when creating tables.
...
+
+| name          | data_type     | nullable |
```
burnash, line 143: Let's make "name" here more explicit so it doesn't clash with "name" the field below.

burnash (changes requested):
This is so good, @sultaniman 
The only thing imo we can improve is to use a more descriptive name for the "name" column so make it super clear what's the table represents.

burnash (approved):
LGTM

PR #888, opened 2024-01-12
Update airflow helper

AstrakhantsevaAA:
### Description

Community member Hawkar Mahmod noticed we don't have some args in `add_run` function:

> I'm using the airflow helper, and would like to set the loader_file_format argument, which is normally passed to pipeline.run. However with Airflow we use the .add_run of the PipelineTaskGroup. Additional kwargs passsed into this method are provded to the airflow task rather than the pipeline run method. I've tried the GPT-4 helper but to no avail. Anyone have any idea where I can change the file format. I've looked at what add_run does and it does not pass any additional args to the pipeline.run call.

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -25,7 +25,8 @@ except ModuleNotFoundError:
 
 import dlt
 from dlt.common import logger
-from dlt.common.schema.typing import TWriteDisposition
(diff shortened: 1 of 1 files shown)


burnash (approved):
Looks good, thanks for fixing this @AstrakhantsevaAA

PR #890, opened 2024-01-12
Blog/mode

hibajamal:
### Description
Co-marketing article for with Mode - about creating the modern data stack with dlt and Mode. Including demo of methods of how to use them both together.


AstrakhantsevaAA (commented):
I have all the same comments:
- Great article! Love your writing style
- Highlight `dlt` everywhere or nowhere
- Use black style for your python code-snippets
- Some of the pictures are too smalL, I can't see anything, for example this one
image

burnash (changes requested):

Issue #752, opened 2023-11-10
implement destination sinks


rudolfix:
**Background**
Sinks are the reverse of the source. A Python function decorated with `@dlt.sink` should generate destination factory (#746 ) that may be passed to pipeline in `destination` argument. Sink will consume data that is sent to it during the `load` stage.
Sink will be fed data in configurable chunks and will keep state of what got consumed already so it will be "transactional" when size of the chunk == 1.
Sink state is separate from source state but initially will be handled together with it - and preserved in file. For "transactional" sinks there will be tons of state updates so general mechanism we use now (store state at destination) does not make sense. We may support bucket storage for `load.py` and load storage for fully resumable, transactional sinks.

**Tasks**
This is rather a big ticket, tasks below are just outline:
1. * [ ] implement sink destination (like any other) with a proper destination factory
2. * [ ] implement `@dlt.sink` decorator working similarly to `@dlt.transformer`. It returns configured destination factory
3. * [ ] sink decorator should accept a sink function, where first argument is `TDataItems`, second a table schema (and maybe something else), and optional parameters (same as parametrized transformer)
4. * [ ] sink decorator should allow to specify batch size, file format (json, parquet). batch_size=1 sends single rows, >1 sends lists of objects (or arrow tables if parquet), batch_size=0 sends the FileDict as data item so user can handle file processing as s/he wants
5. * [ ] (optional) sink decorator should work well with source decorator and sinks should be part of `DltSource` objects (let's think about it)

**Implementation**
Basic implementation may be quite simple. Sink destination implements a copy job that opens a job file and then streams it into a sink function in chunks that were configured. After each chunk it emits sink state.

On restart the state must be retrieved and processed rows must be skipped.

State handling could be part of `LoadStorage` and we may make the `LoadStorage` comparible with fsspec (maybe in separate tickets - it depends a lot on "file move" to be fast and atomic)

rudolfix:
optionally sinks could be chained after main destination consumes data and not just replace it. there's for sure a clever way of doing that without complicating user interface

sh-rp:
Questions:
* Should we support awaitable sink functions?
* Should we add a flag to guarantee sink function execution on the main thread for each item, probably using multiprocessing queues?

PR #891, opened 2024-01-12
data sink destination

sh-rp:
### Description
Implementation of https://github.com/dlt-hub/dlt/issues/752

Notes:
* We also introduce the concept of a load package state here


dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,220 @@
...
+        job_execution_storage: Dict[str, int],
+    ) -> None:
+        super().__init__(FileStorage.get_file_name_from_file_path(file_path))
+        self._file_path = file_path
+        self._config = config
+        self._table = table
+        self._schema = schema
+        self._job_execution_storage = job_execution_storage
+
+        # TODO: is this the correct way to tell dlt to retry this job in the next attempt?
```
rudolfix, line 44: read the docstrings for LoadJob. particular instance will not be retried. "retry" is a terminal state for the `LoadJob` class. loader.py will create a new instance of SinkLoadJob will be created

dlt/destinations/impl/sink/sink.py
```diff
@@ -193,23 +144,21 @@ def update_stored_schema(
         return super().update_stored_schema(only_tables, expected_update)
 
     def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
+        load_state = destination_state().setdefault(load_id, {})
```
rudolfix, line 147: set default is not thread safe. but there won't be anyone else accessing this key so OK

dlt/common/pipeline.py
```diff
@@ -679,6 +684,50 @@ def source_state() -> DictStrAny:
...
+    if not destination_state_key:
+        raise SourceSectionNotAvailable()
+
+    state, _ = pipeline_state(Container())
+
+    state.setdefault("destinations", {}).pop(destination_state_key)
+
+
+def commit_pipeline_state() -> None:
+    container = Container()
```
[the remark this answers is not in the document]
sh-rp: i did this in the new code in the place where the context is created, do you think we should create the lock here?

dlt/common/storages/load_package.py
```diff
@@ -200,6 +201,9 @@ class PackageStorage:
     PACKAGE_COMPLETED_FILE_NAME = (  # completed package marker file, currently only to store data with os.stat
         "package_completed.json"
     )
+    LOAD_PACKAGE_STATE_FILE_NAME = (  # internal state of the load package, will not be synced to the destination
+        "load_package_state.json"
+    )
```
sh-rp, line 206: should i have one new file for the loadpackage state and keep the package_completed.json as is, or should we have a properly typed state dict that has both infos?

dlt/destinations/decorators.py
```diff
@@ -0,0 +1,12 @@
+from typing import Any, Callable
+from dlt.destinations.impl.sink.factory import sink as _sink
+from dlt.destinations.impl.sink.configuration import SinkClientConfiguration, TSinkCallable
+from dlt.common.destination import TDestinationReferenceArg
+from dlt.common.destination import TLoaderFileFormat
+
+
+def sink(loader_file_format: TLoaderFileFormat = None, batch_size: int = 10) -> Any:
+    def decorator(f: TSinkCallable) -> TDestinationReferenceArg:
+        return _sink(credentials=f, loader_file_format=loader_file_format, batch_size=batch_size)
```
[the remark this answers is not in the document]
sh-rp: I'm not sure what you mean with cleaning up the name? Do you mean extract it from the callable name or something like that?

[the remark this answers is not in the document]
sh-rp: Ok, i added it so it will take a name param or use the callable name if possible

dlt/destinations/impl/sink/configuration.py
```diff
@@ -0,0 +1,76 @@
...
+        # a path to a callable was passed in
+        if isinstance(native_value, str):
+            self.callable = native_value
+
+    def to_native_representation(self) -> Any:
+        return self.resolved_callable
+
+    def on_resolved(self) -> None:
+        if self.callable:
+            try:
```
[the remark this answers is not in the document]
sh-rp: I copied this from the code that is used for destinations to be resolved. Do you want me to change it?

dlt/destinations/impl/sink/factory.py
```diff
@@ -0,0 +1,45 @@
...
+
+if t.TYPE_CHECKING:
+    from dlt.destinations.impl.sink.sink import SinkClient
+
+
+class sink(Destination[SinkClientConfiguration, "SinkClient"]):
+    spec = SinkClientConfiguration
+
+    def capabilities(self) -> DestinationCapabilitiesContext:
+        return capabilities(self.config_params.get("loader_file_format", "parquet"))
```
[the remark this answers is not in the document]
sh-rp: i have set the transaction related ones and kept the rest to the default settings.

dlt/load/load.py
```diff
@@ -556,12 +566,29 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
         schema = self.load_storage.normalized_packages.load_schema(load_id)
         logger.info(f"Loaded schema name {schema.name} and version {schema.stored_version}")
 
+        # prepare load package state context
```
sh-rp, line 569: i am injecting the load package state context here, i think it makes the most sense but it also clutters the load class a bit, wdyt? For now we only use the load package state for the destinations, so there is no need to implement this any other stage than load.

sh-rp:
@rudolfix you can have another look at this PR. The main questions are:
* Is the load package state done right? Should we add a schema there and combine it with this file that for now only has one line of text (loaded etc) in it?
* Should the sink function get a bit more info than just dataitems and table schema? Maybe sentinel values that indicate start and finish of a job or something like that? A context object could also be nice, but maybe no-one needs it.

dlt/common/storages/load_package.py
```diff
@@ -380,6 +386,30 @@ def save_schema_updates(self, load_id: str, schema_update: TSchemaTables) -> Non
...
+        package_path = self.get_package_path(load_id)
+        if not self.storage.has_folder(package_path):
+            raise LoadPackageNotFound(load_id)
+        try:
+            state = self.storage.load(
+                os.path.join(package_path, PackageStorage.LOAD_PACKAGE_STATE_FILE_NAME)
+            )
+            return cast(DictStrAny, json.loads(state))
+        except FileNotFoundError:
+            return {}
```
sultaniman, line 402: Is logging details is useful for debugging?

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def drop_storage(self) -> None:
+        pass
+
+    def update_stored_schema(
+        self, only_tables: Iterable[str] = None, expected_update: TSchemaTables = None
+    ) -> Optional[TSchemaTables]:
+        return super().update_stored_schema(only_tables, expected_update)
+
+    def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
```
sultaniman, line 146: Since `None` is also returned the annotation should be `Optional[LoadJob]`

dlt/common/storages/load_package.py
sh-rp: no, this just means that there is no statefile yet

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,170 @@
...
+        except Exception as e:
+            self._state = "retry"
+            raise e
+        finally:
+            # save progress
+            commit_load_package_state()
+
+    @abstractmethod
+    def run(self, start_index: int) -> Iterable[TDataItems]:
+        pass
```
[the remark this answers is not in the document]
sh-rp: i think you can instantiate this a subclass of this class without implementing a method marked as abstractmethod, so this is not needed.

dlt/destinations/impl/sink/configuration.py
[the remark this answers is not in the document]
rudolfix: heh I think I misunderstood the code. `callable` here is misleading. it is a full class name no?

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def call_callable_with_items(self, items: TDataItems) -> None:
+        if not items:
+            return
+
+        # coerce items into correct format specified by schema
+        coerced_items: TDataItems = []
+        for item in items:
+            coerced_item, table_update = self._schema.coerce_row(self._table["name"], None, item)
```
rudolfix, line 67: why we need to do it. this is expensive operation. all rows are coerced in the normalizer?

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,170 @@
...
+        pass
+
+    def update_stored_schema(
+        self, only_tables: Iterable[str] = None, expected_update: TSchemaTables = None
+    ) -> Optional[TSchemaTables]:
+        return super().update_stored_schema(only_tables, expected_update)
+
+    def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
+        # save our state in destination name scope
+        load_state = load_package_state().setdefault(self.config.destination_name, {})
```
rudolfix, line 148: now when I think of it we do not need injection at all. you have load id so create PackageStorage here and pass it to sink load job. then use the method directly. (don't forget to lock on commit)

dlt/load/load.py
rudolfix: Now when I look at this I think we do not need to inject state at all, what we could inject is load id and instance of the package storage as a context available to a function decorated with sink. we can do that later.

dlt/load/load.py
```diff
@@ -556,12 +566,29 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
...
+        load_package_state = self.load_storage.normalized_packages.get_load_package_state(load_id)
```
rudolfix, line 570: IMO code below should be a part of the `LoadPackageStateInjectableContext` and also incliude
1. load id
2. package storage instance (normalized packages)

rudolfix (changes requested):
heh probably this review is a PITA but IMO we do not need the injection at all now when we use storage state.

we can create instance of PackageStorage in the sink job and do everything directly

we could use the injection for the benefit of decorated sink function ie via
dlt.current.sink that could get the state, load id, pacjage name etc.

your choice :)

dlt/destinations/impl/sink/sink.py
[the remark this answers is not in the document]
sh-rp: yes, sorry, I misspelled. i meant to say it's ok like this :)

dlt/destinations/impl/sink/sink.py
sh-rp: hm, are you sure? if we get the current loadpackagestate from the state file in every thread from the storage, we have multiple copies in memory and it will not be thread safe anymore even if we lock on commit. we'd have to lock, then load update dict and commit and then release the lock. Or am i misunderstanding what you are suggesting?

dlt/load/load.py
sh-rp: Alright, I changed it like this, it looks much nicer this way :)

dlt/destinations/impl/sink/sink.py
sh-rp: Alright, if I just keep a copy of the state in memory it should not be a problem.

dlt/destinations/impl/sink/sink.py
sh-rp: @rudolfix I'd still need an answer on this one.

sh-rp:
> heh probably this review is a PITA but IMO we do not need the injection at all now when we use storage state.
> 
> we can create instance of PackageStorage in the sink job and do everything directly
> 
> we could use the injection for the benefit of decorated sink function ie via dlt.current.sink that could get the state, load id, pacjage name etc.
> 
> your choice :)

I'm not quite sure what you mean by some of this, but I have functions in there now that restrict the access of the destination that is manipulating the loadpackage state to only their own chunk in there, similar to how the resources have a scoped access to the pipeline state.

sh-rp:
I have sort of extracted the the common parts between the pipeline state and the loadpackage state with the versioning etc. We could probably also have the schema make use of this, but the properties for hash and current version etc are not prefixed with an underline. I think it would be nice to unify this at some point, maybe in an upcoming PR, not sure.

dlt/common/storages/load_package.py
```diff
@@ -26,13 +28,65 @@
...
+    generate_state_version_hash,
+    bump_state_version_if_modified,
+    TVersionedState,
+    default_versioned_state,
+)
+from typing_extensions import NotRequired
+
+
+class TLoadPackageState(TVersionedState, total=False):
+    created: int
```
sh-rp, line 43: should I put a datetime string here or is utc timestamp ok?

docs/website/docs/dlt-ecosystem/destinations/sink.md
```diff
@@ -0,0 +1,109 @@
...
+2. There are a few other ways for declaring sink functions for your pipeline described below.
+:::
+
+## Sink decorator function and signature
+
+The full signature of the sink decorator and a function is
+
+```python
+@dlt.sink(batch_size=10, loader_file_format="jsonl", name="my_sink")
+def sink(items: TDataItems, table: TTableSchema) -> None:
```
sultaniman, line 56: Should we add a comment clarifying where users can import our internal types from?

tests/common/test_versioned_state.py
```diff
@@ -0,0 +1,43 @@
...
+)
+
+
+def test_versioned_state() -> None:
+    state = default_versioned_state()
+    assert state["_state_version"] == 0
+    assert state["_state_engine_version"] == 1
+
+    # first hash_ generation does not change version, attrs are not modified
+    version, hash_, previous_hash = bump_state_version_if_modified(state)
```
sultaniman, line 14: should `_hash` be `new_hash`?

tests/cases.py
```diff
@@ -226,8 +231,16 @@ def assert_all_data_types_row(
     if "col11" in db_mapping:
         db_mapping["col11"] = db_mapping["col11"].isoformat()
 
-    for expected, actual in zip(expected_rows.values(), db_mapping.values()):
-        assert expected == actual
+    if expect_filtered_null_columns:
+        for key, expected in expected_rows.items():
+            if expected is None:
+                assert db_mapping.get(key, None) is None
```
sultaniman, line 237: maybe better to check that key is not in `key not in db_mapping`?

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,172 @@
...
+            for line in f:
+                # find correct start position
+                if start_index > 0:
+                    start_index -= 1
+                    continue
+                current_batch.append(json.loads(line))
+                if len(current_batch) == self._config.batch_size:
+                    yield current_batch
+                    current_batch = []
+            yield current_batch
```
sultaniman, line 124: should we check if batch actually has data?

dlt/destinations/impl/sink/sink.py
rudolfix: use jsonl-pue file format and prevent using jsonl. `pue` has additional encoding and and a set of functions that recursively restore python objects.

dlt/common/data_types/type_helpers.py
```diff
@@ -105,6 +105,14 @@ def coerce_value(to_type: TDataType, from_type: TDataType, value: Any) -> Any:
                 return int(value.value)
         return value
 
+    if to_type == "complex":
+        # try to coerce from text
+        if from_type == "text":
+            try:
+                return json.loads(value)
+            except Exception:
+                raise ValueError("Cannot load text as json for complex type")
```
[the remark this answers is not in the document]
rudolfix: the pattern is to raise ValueError(value). you'll get the error message from exception.__context__

dlt/common/pipeline.py
```diff
@@ -599,6 +602,28 @@ class StateInjectableContext(ContainerInjectableContext):
         def __init__(self, state: TPipelineState = None) -> None: ...
 
 
+@configspec
+class LoadPackageStateInjectableContext(ContainerInjectableContext):
```
rudolfix, line 606: cool but why here? it is a part of package storage. and it is a big 👍 that it is decoupled from pipeline (it will help me to handle bad data for example)

dlt/common/storages/load_package.py
rudolfix: hmmm we use ISO timestamps everywhere

dlt/common/storages/load_package.py
```diff
@@ -26,13 +28,65 @@
...
+    """Timestamp when the loadpackage was created"""
+
+    """A section of state that does not participate in change merging and version control"""
+    destinations: NotRequired[Dict[str, Dict[str, Any]]]
```
rudolfix, line 47: a load package will be processed by just one destination. the processing is destructive (we move files around, create followup jobs etc.) so I assume we just need `destination_state which is a dict?

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,172 @@
...
+            start_index % self._config.batch_size
+        ) == 0, "Batch size was changed during processing of one load package"
+
+        start_batch = start_index / self._config.batch_size
+        with pyarrow.parquet.ParquetFile(self._file_path) as reader:
+            for record_batch in reader.iter_batches(batch_size=self._config.batch_size):
+                if start_batch > 0:
+                    start_batch -= 1
+                    continue
+                batch = record_batch.to_pylist()
```
rudolfix, line 105: hmmmmm. IMO just yield arrow here. Alex for example is pushing panda frames to salesforce

dlt/pipeline/pipeline.py
```diff
@@ -724,7 +734,7 @@ def sync_destination(
 
             # write the state back
             self._props_to_state(state)
-            bump_version_if_modified(state)
+            bump_pipeline_state_version_if_modified(state)
```
rudolfix, line 737: heh we are close to  implement a this as a class :)

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,172 @@
...
+        if file_path.endswith("jsonl"):
+            return SinkJsonlLoadJob(table, file_path, self.config, self.schema, load_state)
+        return None
+
+    def restore_file_load(self, file_path: str) -> LoadJob:
+        return EmptyLoadJob.from_file_path(file_path, "completed")
+
+    def complete_load(self, load_id: str) -> None:
+        # pop all state for this load on success
+        clear_loadpackage_destination_state()
```
rudolfix, line 164: IMO it is better if we do not delete it. You can use it ie. to inspect how many rows were processed.

rudolfix (changes requested):
logic LGTM! I think we have basics covered. 
1. Should we rename sink to something else? Alex is using "publisher", we could also use "dlt.destination" but maybe that would be too big for a name?
2. could we add `dlt.current.destination_state()` and dlt.current.load_state()` that could be used in the decorated function. It may make sense to keep some state or be able to inspect full schema etc.
3. I will probably try wrap a sink function so additional configuration (ie. credentials) can be passed to it. IMO we can merge into destination factory better than making credentials a callable

dlt/pipeline/pipeline.py
sh-rp: yeah, i think so too..

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class SinkLoadJob(LoadJob, ABC):
+    def __init__(
+        self,
+        table: TTableSchema,
+        file_path: str,
+        config: SinkClientConfiguration,
+        schema: Schema,
+        load_package_state: Dict[str, int],
```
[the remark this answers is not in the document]
sh-rp: no, but i renamed it to make it more clear what is is :)

dlt/destinations/impl/sink/factory.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def client_class(self) -> t.Type["SinkClient"]:
+        from dlt.destinations.impl.sink.sink import SinkClient
+
+        return SinkClient
+
+    def __init__(
```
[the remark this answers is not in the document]
sh-rp: yes, the important configs can be listed here for better autocompletion

dlt/common/storages/load_package.py
sh-rp: changed

docs/website/docs/dlt-ecosystem/destinations/sink.md
sh-rp: you can see that a couple of lines above this example

tests/cases.py
sh-rp: no, the key could be there and set to none and that would also be ok i would say.

dlt/common/storages/load_package.py
```diff
@@ -26,13 +28,65 @@
...
+    """private space for destinations to store state relevant only to the load package"""
```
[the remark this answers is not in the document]
sh-rp: i have no idea, i took this from the way it is done in other places @rudolfix is this some convention?

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,172 @@
...
+    def state(self) -> TLoadJobState:
+        return self._state
+
+    def exception(self) -> str:
+        raise NotImplementedError()
+
+
+class SinkParquetLoadJob(SinkLoadJob):
+    def run(self, start_index: int) -> Iterable[TDataItems]:
+        # stream items
```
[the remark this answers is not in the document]
sh-rp: no, this is a developer comment

dlt/destinations/impl/sink/sink.py
sh-rp: this is already done in the call_callable_with_items function above

dlt/common/pipeline.py
sh-rp: i moved it, but i am not 100% that it is in the right location now..

dlt/destinations/impl/sink/sink.py
sh-rp: done

dlt/destinations/impl/sink/sink.py
sh-rp: alright, changed!

sh-rp:
@rudolfix i implemented all the changes, since puae-jsonl files are also stored with the .jsonl extension, i had to fix a place in the loader which does not seem super nice. If we wanted to do this properly I think we would have puae-json (the current format but renamed) and puae-jsonl (new format with individually encoded lines) and save them with a different exension then jsonl, and it would also be nice to have an abstracted reader class (similar to the write) where you just stick a file path and it emits rows or batches of rows, so that this decoding code is not in the sink destination. But we could do a later ticket for that, not sure.

dlt/destinations/impl/sink/configuration.py
```diff
@@ -0,0 +1,71 @@
...
+from dlt.common.typing import TDataItems
+from dlt.common.schema import TTableSchema
+from dlt.common.configuration.exceptions import ConfigurationValueError
+
+
+TSinkCallable = Callable[[Union[TDataItems, str], TTableSchema], None]
+
+
+@configspec
+class SinkClientCredentials(CredentialsConfiguration):
```
rudolfix, line 19: Those are not credentials. User will provide credentials (eventually) in the decorated function. if you need to keep the reference to decorated function, use some other field name

dlt/destinations/impl/sink/configuration.py
```diff
@@ -0,0 +1,71 @@
...
+        if not hasattr(self, "resolved_callable"):
+            raise ConfigurationValueError("Please specify callable for sink destination.")
+
+        if not callable(self.resolved_callable):
+            raise ConfigurationValueError("Resolved Sink destination callable is not a callable.")
+
+
+@configspec
+class SinkClientConfiguration(DestinationClientConfiguration):
+    destination_type: Final[str] = "sink"  # type: ignore
```
rudolfix, line 58: mhhhh I think we'll abuse this config and use the name of the destination as type?

dlt/destinations/impl/sink/configuration.py
```diff
@@ -0,0 +1,71 @@
...
+    loader_file_format: TLoaderFileFormat = "puae-jsonl"
+    batch_size: int = 10
+
+    if TYPE_CHECKING:
+
+        def __init__(
+            self,
+            *,
+            credentials: Union[SinkClientCredentials, TSinkCallable, str] = None,
+            loader_file_format: TLoaderFileFormat = "puae-jsonl",
```
rudolfix, line 69: this signature must be generated from the decorated function signature by skipping two first mandatory parameters. I can show you how to do it in mypy ie.
```
@overload
def transformer(
    f: None = ...,
    /,
    data_from: TUnboundDltResource = DltResource.Empty,
    name: TTableHintTemplate[str] = None,
    table_name: TTableHintTemplate[str] = None,
    write_disposition: TTableHintTemplate[TWriteDisposition] = None,
    selected: bool = True,
    spec: Type[BaseConfiguration] = None,
    parallelized: bool = False,
    standalone: Literal[True] = True,
) -> Callable[
    [Callable[Concatenate[TDataItem, TResourceFunParams], Any]],
    Callable[TResourceFunParams, DltResource],
]: ...
```
```
 Callable[
    [Callable[Concatenate[TDataItem, TResourceFunParams], Any]],
    Callable[TResourceFunParams, DltResource],
]: ...
```
cuts first parameter from the signature, and you can have more

dlt/destinations/impl/sink/configuration.py
```diff
@@ -0,0 +1,71 @@
...
+            batch_size: int = 10,
```
rudolfix, line 70: file_format and batch_size are part of the decorator, not destination signature, remove

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,175 @@
...
+        self._table = table
+        self._schema = schema
+        self._resolved_callable = resolved_callable
+
+        self._state: TLoadJobState = "running"
+        self._storage_id = f"{self._parsed_file_name.table_name}.{self._parsed_file_name.file_id}"
+        try:
+            if self._config.batch_size == 0:
+                # on batch size zero we only call the callable with the filename
+                self.call_callable_with_items(self._file_path)
```
rudolfix, line 50: this is super good

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,175 @@
...
+        # stream items
+        from dlt.common.libs.pyarrow import pyarrow
+
+        # guard against changed batch size after restart of loadjob
+        assert (
+            start_index % self._config.batch_size
+        ) == 0, "Batch size was changed during processing of one load package"
+
+        start_batch = start_index / self._config.batch_size
+        with pyarrow.parquet.ParquetFile(self._file_path) as reader:
```
rudolfix, line 94: mhhhh this looks to me like `file_readers` - opposite to our `file_writers`. so you can read parquet and jsonl (with pua) / json with them.

makes sense to extract it now?

dlt/destinations/impl/sink/sink.py
```diff
@@ -0,0 +1,175 @@
...
+        try:
+            if self._config.batch_size == 0:
+                # on batch size zero we only call the callable with the filename
+                self.call_callable_with_items(self._file_path)
+            else:
+                current_index = destination_state.get(self._storage_id, 0)
+                for batch in self.run(current_index):
+                    self.call_callable_with_items(batch)
+                    current_index += len(batch)
+                    destination_state[self._storage_id] = current_index
```
rudolfix, line 56: are you able to store batch number per table in state? (you'll need to lock the state). then user will be able to figure out that a batch is a first batch for a given table. you already store stuff in state

dlt/__init__.py
```diff
@@ -62,6 +64,7 @@
     "resource",
     "transformer",
     "defer",
+    "destination",
```
rudolfix, line 67: please also import `load_package()` into `current` module so it can be requested from within decorated function

rudolfix (changes requested):
this is getting so good!
we'll have another round of comments. now I was focused on generating proper config and destination signature

regarding tests:
please implement a BigQuery insert api sink, use #1037 for a reference. you can add this in bigquery specific tests and in examples. what it should do

- accept GcpCredentials and credentials + some default parameters
- insert json in batches
- bigquery should be able to infer the schema

dlt/destinations/impl/sink/configuration.py
sh-rp: yes, this is the way it works already (see the config/secrets tests), unless I misunderstand what you are saying.

dlt/destinations/decorators.py
```diff
@@ -0,0 +1,27 @@
+from typing import Any, Callable
+from dlt.destinations.impl.sink.factory import sink as _sink
+from dlt.destinations.impl.sink.configuration import SinkClientConfiguration, TSinkCallable
+from dlt.common.destination import TDestinationReferenceArg
+from dlt.common.destination import TLoaderFileFormat
+from dlt.common.utils import get_callable_name
+
+
+def sink(
```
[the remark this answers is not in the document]
sh-rp: yes, this is what the decorator is called when importing it from dlt (see the tests). I can also rename this function though.

dlt/destinations/impl/sink/sink.py
sh-rp: Let's do a separate PR for this where filetypes are fully pluggable..

dlt/destinations/impl/sink/configuration.py
sh-rp: Alright, moved the callable into the regular config

dlt/__init__.py
sh-rp: It is there, no?

dlt/destinations/impl/sink/configuration.py
sh-rp: I'm not quite sure what you mean. When I create a destination in the decorator, I need a way to hand over these values to the destination instance, and in all destinations this kind of stuff is done via the config, no?

dlt/common/configuration/container.py
```diff
@@ -87,10 +87,14 @@ def _thread_context(
             context = self.main_context
         else:
             # thread pool names used in dlt contain originating thread id. use this id over pool id
-            if m := re.match(r"dlt-pool-(\d+)-", threading.currentThread().getName()):
-                thread_id = int(m.group(1))
-            else:
-                thread_id = threading.get_ident()
+            # print(threading.currentThread().getName())
```
sh-rp, line 90: I'm not quite sure how this is supposed to work, in any case it will not work to inject multiple configcontexts of the same type in different threads if I keep this in. lmk.

PR #892, opened 2024-01-12
Databricks destination

### LINKED ISSUE #762: [destinations] Databricks
### Feature description

Implement Databricks destination. It will be quite similar to the `Snowflake` implementation.

- use [Databricks Python Connector](https://docs.databricks.com/en/dev-tools/python-sql-connector.html)
- decide what to use [COPY INTO](https://docs.databricks.com/en/ingestion/copy-into/index.html) or [AutoLoader](https://docs.databricks.com/en/ingestion/auto-loader/index.html)

[Authorization](https://docs.databricks.com/en/dev-tools/auth.html):

- Databricks personal access token
- OAuth with refresh token

Support data sources ([staging storage](https://dlthub.com/docs/dlt-ecosystem/staging)):
- [Managed tables](https://docs.databricks.com/en/lakehouse/data-objects.html#what-is-a-managed-table)
- Amazon S3 (External storage)
- Azure Blob Storage (External storage)

Tests

- [ ] we should enable it as destination in tests. all tests using ALL_DESTINATIONS must pass
- [ ] it should pass the same common tests which BigQuery job client and sql client is passing
- [ ] note that to pass sql_client tests you'll need to map DBApi exceptions into 3 categories that dlt needs: relation not found, terminal exceptions, transient exceptions. see bigquery and postgres implementations. that's manual work

steinitzu:
### Description

Draft of databricks destination #762 

### Related Issues

- Resolves #762

### Additional Context

At the moment `COPY INTO` is always used and requires `s3`/`azure` staging bucket


dlt/destinations/impl/databricks/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+def _configure(
+    config: DatabricksClientConfiguration = config.value,
+) -> DatabricksClientConfiguration:
+    return config
+
+
+def capabilities() -> DestinationCapabilitiesContext:
+    caps = DestinationCapabilitiesContext()
+    caps.preferred_loader_file_format = "parquet"
+    caps.supported_loader_file_formats = ["parquet"]
```
steinitzu, line 30: I could not get `jsonl` to work with `COPY INTO`. Suspect it's because of gzip, i.e. either databricks doesn't support reading gzipped json or more likely requires file extension `json.gz`

dlt/destinations/impl/databricks/configuration.py
```diff
@@ -0,0 +1,112 @@
...
+    _credentials_provider: Optional[Dict[str, Any]] = None
+
+    __config_gen_annotations__: ClassVar[List[str]] = [
+        "server_hostname",
+        "http_path",
+        "catalog",
+        "schema",
+    ]
+
+    def __post_init__(self) -> None:
```
sh-rp, line 37: we also have the on_resolved function that gets called after the configuration is resolved, maybe you can use that, or do you need this to run before?

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -0,0 +1,336 @@
...
+                    )
+                )
+                from_clause = f"FROM '{bucket_path}'"
+            else:
+                raise LoadJobTerminalException(
+                    file_path,
+                    f"Databricks cannot load data from staging bucket {bucket_path}. Only s3 and azure buckets are supported",
+                )
+        else:
+            raise LoadJobTerminalException(
```
sh-rp, line 177: somehow i think it would be nice to be able to define in the capabilities which staging filesystem types are supported by each destination. not sure wether to add this now or later, but it would save some of this code in a few places and we could also catch config errors earlier.

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -0,0 +1,336 @@
...
+        )
+        file_name = (
+            FileStorage.get_file_name_from_file_path(bucket_path) if bucket_path else file_name
+        )
+        from_clause = ""
+        credentials_clause = ""
+        files_clause = ""
+        # stage_file_path = ""
+        format_options = ""
+        copy_options = "COPY_OPTIONS ('mergeSchema'='true')"
```
sh-rp, line 131: is mergeschema needed? normally the loader should have updated the destination schema before parquet files start getting copied into there.

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -0,0 +1,336 @@
...
+    def state(self) -> TLoadJobState:
+        return "completed"
+
+    def exception(self) -> str:
+        raise NotImplementedError()
+
+
+class DatabricksStagingCopyJob(SqlStagingCopyJob):
+    @classmethod
+    def generate_sql(
```
sh-rp, line 206: i think some of the other destinations have very similar syntax for this, check bq and snowflake. maybe we can have a common stagingcopyjob shared between them, or a capability "supports_clone_command" so that we can just enable this on any sql destination and have a different syntax being emmitted from the SqlStagingCopyJob.

sh-rp: syntax = sql_syntax

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -0,0 +1,336 @@
...
+    # def _create_staging_copy_job(self, table_chain: Sequence[TTableSchema]) -> NewLoadJob:
+    #     return DatabricksStagingCopyJob.from_table_chain(table_chain, self.sql_client)
+
+    def _make_add_column_sql(
+        self, new_columns: Sequence[TColumnSchema], table_format: TTableFormat = None
+    ) -> List[str]:
+        # Override because databricks requires multiple columns in a single ADD COLUMN clause
+        return ["ADD COLUMN\n" + ",\n".join(self._get_column_def_sql(c) for c in new_columns)]
+
+    # def _create_optimized_replace_job(self, table_chain: Sequence[TTableSchema]) -> NewLoadJob:
```
sh-rp, line 284: dev leftover?

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -0,0 +1,336 @@
...
+    def restore_file_load(self, file_path: str) -> LoadJob:
+        return EmptyLoadJob.from_file_path(file_path, "completed")
+
+    def _create_merge_job(self, table_chain: Sequence[TTableSchema]) -> NewLoadJob:
+        return DatabricksMergeJob.from_table_chain(table_chain, self.sql_client)
+
+    def _create_merge_followup_jobs(self, table_chain: Sequence[TTableSchema]) -> List[NewLoadJob]:
+        return [DatabricksMergeJob.from_table_chain(table_chain, self.sql_client)]
+
+    # def _create_staging_copy_job(self, table_chain: Sequence[TTableSchema]) -> NewLoadJob:
```
sh-rp, line 275: dev leftover?

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -0,0 +1,336 @@
...
+        cluster_list = [
+            self.capabilities.escape_identifier(c["name"]) for c in new_columns if c.get("cluster")
+        ]
+
+        if cluster_list:
+            sql[0] = sql[0] + "\nCLUSTER BY (" + ",".join(cluster_list) + ")"
+
+        return sql
+
+    def _execute_schema_update_sql(self, only_tables: Iterable[str]) -> TSchemaTables:
```
sh-rp, line 312: why do you need to override this method? won't the one in the baseclass work?

dlt/destinations/impl/databricks/sql_client.py
```diff
@@ -0,0 +1,166 @@
...
+        return self._conn
+
+    @raise_open_connection_error
+    def close_connection(self) -> None:
+        if self._conn:
+            self._conn.close()
+            self._conn = None
+
+    @contextmanager
+    def begin_transaction(self) -> Iterator[DBTransaction]:
```
sh-rp, line 52: we have a (seemingly currently unused) supports_transactions flag in the capabilities, why not use that in the superclass to determin wether transations are started here?

sh-rp: and the following methods.

dlt/destinations/impl/databricks/sql_client.py
```diff
@@ -0,0 +1,166 @@
...
+        self, sql: AnyStr, *args: Any, **kwargs: Any
+    ) -> Optional[Sequence[Sequence[Any]]]:
+        with self.execute_query(sql, *args, **kwargs) as curr:
+            if curr.description is None:
+                return None
+            else:
+                f = curr.fetchall()
+                return f
+
+    def execute_many(
```
sh-rp, line 89: ok, now I see why you had to override further up. maybe this also is more a case for a capabilty for sql destinations, I am not 100% sure though.

sh-rp: Actually you have it with supports_many... why not use that in the super class?

tests/load/test_job_client.py
```diff
@@ -387,7 +387,9 @@ def test_get_storage_table_with_all_types(client: SqlJobClientBase) -> None:
             "time",
         ):
             continue
-        if client.config.destination_type == "mssql" and c["data_type"] in ("complex"):
+        if client.config.destination_type in "mssql" and c["data_type"] in ("complex"):
```
sh-rp, line 390: mssql should be in parentheses or brackets here.

sh-rp (changes requested):
very nice work, I have a few comments

dlt/destinations/impl/databricks/__init__.py
sh-rp: can you confirm if using a different extension works?

sh-rp: https://docs.databricks.com/en/sql/language-manual/delta-copy-into.html

steinitzu: The multiline mode applies to records that span multiple lines, jsonl works otherwise by default but only with gz compression disabled.    
I have to try some script with `jsonl.gz` filename, looks like a lot of work to actually change extensions in dlt.

dlt/destinations/impl/databricks/configuration.py
steinitzu: Yeah on_resolved should work. Didn't touch these classes much from the previous PR since it seemed to work, needs some tweaks.

dlt/destinations/impl/databricks/databricks.py
steinitzu: Now it does. Updated the baseclass to use `execute_many`. The base sqlclient implements it as:
https://github.com/dlt-hub/dlt/blob/sthor%2Fdatabricks/dlt/destinations/sql_client.py#L122-L139  depending on capabilities

dlt/destinations/impl/databricks/sql_client.py
steinitzu: Hmm yeah `supports_transactions` is only used in tests currently. Every db driver needs slightly different transaction implementation, so if we implement this in base class I think it makes sense to just do this:

```python
    def begin_transaction(self) -> ContextManager[DBTransaction]:
        if not self.capabilities.supports_transactions:
            yield self
        else:
            raise NotImplementedError(f"{type(self).__name__} does not implement begin_transaction")

    def commit_transaction(self) -> None:
        if not self.capabilities.supports_transactions:
            return
        raise NotImplementedError(f"{type(self).__name__} does not implement commit_transaction")

    def rollback_transaction(self) -> None:
        if not self.capabilities.supports_transactions:
            return
        raise NotImplementedError(f"{type(self).__name__} does not implement rollback_transaction")
```

dlt/destinations/impl/databricks/databricks.py
steinitzu: Implemented it like this in https://github.com/dlt-hub/dlt/commit/64b7e2e5  . The `CLONE` syntax is exactly the same for bq/snowflake.  
https://github.com/dlt-hub/dlt/blob/64b7e2e583209146c0e13bbd532f22c66b0adea6/dlt/destinations/sql_jobs.py#L71-L127

dlt/destinations/impl/databricks/__init__.py
steinitzu: Confirmed that gzip works with `jsonl.gz` extension.   
Might be good to change this so dlt creates files with `gz` ext, that's pretty universally supported, but would leave it for another pr.  
I just have parquet enabled for now.

dlt/destinations/impl/databricks/configuration.py
steinitzu: For now I removed a lot of this and am only supporting the "personal access token" auth.   
I don't have a way to test oauth atm, will need to be done somehow so we're not prompting on every connection.

dlt/destinations/impl/databricks/databricks.py
steinitzu: Good idea to have this in capabilities. Can do that and refactor in another PR.

dlt/destinations/impl/databricks/__init__.py
rudolfix: @steinitzu my 2 cents:
there's really no way to set the compression explictly ie. in copy or format options?
if not then please do not disable JSONL, just detect that compression is enabled and fail the job with an exception (ie. ask to disable compression)

rudolfix: another ask @steinitzu could you write a ticket to add `gz` extensions to zipped files in dlt load packages?

steinitzu: > @steinitzu my 2 cents: there's really no way to set the compression explictly ie. in copy or format options?

Hey @rudolfix no there's no option for it that's documented.

>  if not then please do not disable JSONL, just detect that compression is enabled and fail the job with an exception (ie. ask to disable compression)

Hmm yeah that makes sense for now. Until we change the extensions.

steinitzu: It fails on empty jsonl files too. Produced by this test: https://github.com/dlt-hub/dlt/blob/sthor%2Fdatabricks/tests/load/pipeline/test_replace_disposition.py#L154-L163  
Is this supposed to produce a stage file? Not sure if I'm doing something wrong.

rudolfix: yes, empty jsonl files are correct. they mean that content of the table should be dropped as there are no rows in current load. we generate them ie. to delete data from child tables on replace if there were no data at all for given table. your workaround is good and will do what we expected! thanks

rudolfix (dismissed):
LGTM! thanks everyone for the quality work! if @sh-rp agrees we should merge this.

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -0,0 +1,296 @@
...
+                load_id,
+                self.sql_client,
+                staging_config=cast(FilesystemConfiguration, self.config.staging_config),
+            )
+        return job
+
+    def restore_file_load(self, file_path: str) -> LoadJob:
+        return EmptyLoadJob.from_file_path(file_path, "completed")
+
+    def _create_merge_job(self, table_chain: Sequence[TTableSchema]) -> NewLoadJob:
```
sh-rp, line 250: i don't think this is used anywhere, is it?

dlt/destinations/impl/databricks/factory.py
```diff
@@ -0,0 +1,54 @@
...
+            credentials: Credentials to connect to the databricks database. Can be an instance of `DatabricksCredentials` or
+                a connection string in the format `databricks://user:password@host:port/database`
+            stage_name: Name of the stage to use for staging files. If not provided, the default stage will be used.
+            keep_staged_files: Should staged files be kept after loading. If False, staged files will be deleted after loading.
+            **kwargs: Additional arguments passed to the destination config
+        """
+        super().__init__(
+            credentials=credentials,
+            stage_name=stage_name,
+            keep_staged_files=keep_staged_files,
```
sh-rp, line 50: is this used anywhere?

dlt/destinations/impl/databricks/sql_client.py
sh-rp: yes, that would be cool, maybe in a follow-up PR?

dlt/destinations/impl/snowflake/__init__.py
```diff
@@ -21,4 +21,5 @@ def capabilities() -> DestinationCapabilitiesContext:
     caps.is_max_text_data_type_length_in_bytes = True
     caps.supports_ddl_transactions = True
     caps.alter_add_multi_column = True
+    caps.supports_clone_table = True
```
sh-rp, line 24: thanks for this :)

docs/website/docs/dlt-ecosystem/destinations/databricks.md
```diff
@@ -0,0 +1,103 @@
...
+http_path = "/sql/1.0/warehouses/12345"
+access_token "MY_ACCESS_TOKEN"
+catalog = "my_catalog"
+```
+
+## Write disposition
+All write dispositions are supported
+
+## Data loading
+Data is loaded using `INSERT VALUES` statements by default.
```
sh-rp, line 48: is this still WIP? we only have a copy job for now, no?

sh-rp (changes requested):
a few small comments again :)

docs/website/docs/dlt-ecosystem/destinations/databricks.md
```diff
@@ -52,20 +52,25 @@ For more information on staging, see the [staging support](#staging-support) sec
...
-* [jsonl](../file-formats/jsonl.md) supported when staging is enabled. **Note**: Currently loading compressed jsonl files is not supported. `data_writer.disable_compression` should be set to `true` in dlt config
+* [jsonl](../file-formats/jsonl.md) supported when staging is enabled (see limitations below)
 * [parquet](../file-formats/parquet.md) supported when staging is enabled
 
+The `jsonl` format has some limitations when used with Databricks:
+
+1. Compression must be disabled to load jsonl files in databricks. Set `data_writer.disable_compression` to `true` in dlt config when using this format.
+2. The following data types are not supported when using `jsonl` format with `databricks`: `decimal`, `complex`, `date`, `binary`. Use `parquet` if your data contains these types.
+3. `bigint` data type with precision is not supported with `jsonl` format
+
```
steinitzu, line 63: @sh-rp @rudolfix  
Json support is even more limited. All these came up in `tests/load/pipeline/test_stage_loading.py::test_all_data_types`   
* `binary` in base64 does not work, not sure if it works at all in some other encoding.  
* Databricks parses `date` strings to timestamp and fails  
* There is no general `JSON` or equivelent type (I think), you need to create a struct column with a defined schema. And json objects/arrays don't convert automatically to string  
* Somehow a 16 bit int in JSON is detected as LONG

sh-rp: I definately think we should contact databricks about this and see what they say. @rudolfix who do we know there?

PR #893, opened 2024-01-14
Add git to filesystem source 301 devel

deanja:
Support git filesystem Resource per https://github.com/dlt-hub/verified-sources/issues/301

May also include general improvements to dlt fsspec handling.


deanja:
Supporting parameters in fsspec URLs requires some changes in `dlt.common.storages` .   Some comments on the bigger changes.

```
class FileItemDict(DictStrAny):
    """A FileItem dictionary with additional methods to get fsspec filesystem, open and read files."""

    def __init__(
        self,
        mapping: FileItem,
        credentials: Optional[Union[FileSystemCredentials, AbstractFileSystem]] = None,
    ):
```
[show code](https://github.com/dlt-hub/dlt/blob/3e93685a670476a99467fc4e28963ecf8c923d97/dlt/common/storages/fsspec_filesystem.py#L122C1-L129C7)

Comments re supporting fsspec url arguments:

It has complex responsibilities and not enough scope to fulfull them:
  1. store file metadata. ok, that's easy.
  2. lazy fsspec factory, which also exists as functions above, albeit different signature.
  3. 'file_url' is not enough to instantiate an fs because url params are stripped out in `file_url`. The  params would need to have been copied into `credentials` object or turned into kwargs.   Regardless of urls it's fragile to use somewhat "formatted" output of a Resource as input to further program flow.

Possible way to tidy up is to allow only AbstractFileSystem instance in constructor. Change:
`credentials: Optional[Union[FileSystemCredentials, AbstractFileSystem]] = None,`
to:
`fs_client:  AbstractFileSystem`

- Mostly, this class is instantiated by filesystem Resource, which has an fs instance it can pass in. Stop this class needing to act as a factory.
- Edge case is sources/inbox/__init__.py which instantiates it directly. But only accesses simple dict items and doesn't call fsspec.
- Are dlt users likely to have instantiated this class directly in their own code?

General Commnents:
- Confirm it is a stateful class/object because it's an output of dlt Resource? If yes, could that be made clearer by it inheriting/implementing a dlt Python type?
- Regardless of fsspec urls it's fragile to use somewhat "formatted" output of a Resource as input to further program flow. Is it a common pattern in dlt Resources?   Better to inject dependencies explicitly.

deanja:
Fsspec urls can contain some strange things that urlparse can't understand.  For example, a path in the `username:` element.

Workaround here is to call fs_client._strip_protocol, which is [extended by quite a few fsspec implementations](https://github.com/search?q=repo%3Afsspec%2Ffilesystem_spec+%22def+_strip_protocol%22&type=code).

```
def glob_files(
    fs_client: AbstractFileSystem, bucket_url: str, file_glob: str = "**"
) -> Iterator[FileItem]:
    """Get the files from the filesystem client.

    Args:
        fs_client (AbstractFileSystem): The filesystem client.
        bucket_url (str): The url to the bucket.
        file_glob (str): A glob for the filename filter.

    Returns:
        Iterable[FileItem]: The list of files.
    """
    import os

    protocol = urlparse(bucket_url).scheme
    fsspec_path = fs_client._strip_protocol(bucket_url)
    bucket_url = f"{protocol}://{fsspec_path}"

    bucket_url_parsed = urlparse(bucket_url)
```
[show code](https://github.com/dlt-hub/dlt/blob/3e93685a670476a99467fc4e28963ecf8c923d97/dlt/common/storages/fsspec_filesystem.py#L203-L222)

But _strip_protocol is an _internal method.    

The only public way to get fsspec to clean the url is via `url_to_fs()`, which returns a cleaned path (as well as the fs instance).  Using `url_to_fs()` could be elegant and simplify this `glob_files()` method, but may require some breaking changes to filesystem Resource.   [WIP on a branch](https://github.com/deanja/dlt/commit/5b22d0e6d7d4cc59e68463c15d143dbe8f944ab8)

There's similar url parsing issue in `FileSystemConfiguration` class. 

```
        url = urlparse(self.bucket_url)
```
[view code](https://github.com/dlt-hub/dlt/blob/3e93685a670476a99467fc4e28963ecf8c923d97/dlt/common/storages/configuration.py#L114-L115)

That `url` is used for validation rules.    It happens to  work now, but it could be a problem for and any new rules that might be added to `on_resolved()`.       This one is harder to address with `_strip_protocol()` because there is  no fs client available.    We could do something like a class factory for fsspec implementations, to get at the `_strip_protocol()` @classmethod, without having to instantiate.

deanja:
Final little issue with fsspec url parameters.   How to redact sensitive info.  

```
    def __str__(self) -> str:
        """Return displayable destination location"""
        url = urlparse(self.bucket_url)
        # do not show passwords
        if url.password:
            new_netloc = f"{url.username}:****@{url.hostname}"
            if url.port:
```
[show code](https://github.com/dlt-hub/dlt/blob/3e93685a670476a99467fc4e28963ecf8c923d97/dlt/common/storages/configuration.py#L136-L142)

An fssepec implementation might:
1. Put sensitive info in a generally non-sensitive url element.  eg a password in the `port`.    Recall we are considering hostname:port as a shorthand to clone/fetch a repo from github, bitbucket etc.
2.  Put non-sensitive info in a generally sensitive element, then user can't see it!  eg, git ref in the `password` element, as currently happens with gitpythonfs.

Issue #787, opened 2023-11-23
implement missing arrow and data contracts tests


rudolfix:
**Tasks**

- [ ] - test basic pyarrow lib
- [ ] - test pyarrow normalize
- [ ] - define columns and then load arrow table with incorrect order
- [ ] - fix _dlt_load_id added only if needed
- [ ] - test discard value and discard row in nested json: discard row in child table that has another child table should discard data from descendants
- [ ] - test removal of x-normalizer tags (also when {})
- [ ] - pydantic nested model test
- [x] - ~allow Annotated when synthesizing models~
- [ ] - test freeze exception with Pydantic

sh-rp:
Questions:
* What should we test with nested models on pydantic? Right now it is not implemented, so maybe this is the task?

PR #896, opened 2024-01-16
Add some missing tests

sh-rp:
**Tasks**

- [ ] - test basic pyarrow lib
- [x] - test pyarrow normalize
- [x] - define columns and then load arrow table with incorrect order
- [ ] - fix _dlt_load_id added only if needed
- [x] - test discard value and discard row in nested json: discard row in child table that has another child table should discard data from descendants
- [x] - test removal of x-normalizer tags (also when {})
- [ ] - pydantic nested model test
- [x] - test freeze exception with Pydantic

original ticket: https://github.com/dlt-hub/dlt/issues/787


sh-rp:
@rudolfix what exactly needs to be tested on the pydantic nested models?

tests/libs/pyarrow/test_pyarrow_normalizer.py
```diff
@@ -0,0 +1,105 @@
...
+    assert _row_at_index(result, 0) == ["hello", 1]
+
+
+def test_dlt_columns_not_added() -> None:
+    table = pa.Table.from_pylist(
+        [
+            {"col1": 1},
+        ]
+    )
+    columns = [new_column("_dlt_something", "bigint"), new_column("col2", "text")]
```
rudolfix, line 89: because of SCD2 only _dlt_versions and loads are removed. other columns may be added. check the normalize and pls fix the test

tests/normalize/test_normalize.py
```diff
@@ -708,3 +709,32 @@ def assert_timestamp_data_type(load_storage: LoadStorage, data_type: TDataType)
     event_schema = load_storage.normalized_packages.load_schema(loads[0])
     # in raw normalize timestamp column must not be coerced to timestamp
     assert event_schema.get_table_columns("event")["timestamp"]["data_type"] == data_type
+
+
+def test_removal_of_normalizer_schema_section(raw_normalize: Normalize) -> None:
```
rudolfix, line 714: so this test must be a little extended.
1. we remove evolove-columns-once
2. but we add `seen-data` to all tables that had data items (see shcema utils: has_table_seen_data)
this is a good test, needs to be upgraded

rudolfix (changes requested):
LGTM! 2 tests need upgrade after few months of work

rudolfix (approved):
LGTM!

Issue #832, opened 2023-12-17
implement synapse destination


rudolfix:
**Background**
We'd like to add destination for Synapse. There's previous work done here:
https://github.com/dlt-hub/dlt/pull/740

Our goals are:
1. support most common use case of loading data into synapse (is it copying PARQUET/JSON/CSV from azure storage into synapse)?
2. supporting insert-values would be nice (but not necessary AFAIK synapse can insert 1 record per query)
3. support all write dispositions (append, replace, merge)
4. reuse as much code from MSSQL destination by subclassing, extracting common base class etc. - whatever necessary
5. possibly supporting Azure specific authentication modes (again: following the most common use cases)

Issue #855, opened 2023-12-23
create BigQuery adapter for dlt resources


rudolfix:
**Background**
Following the idea of adapter as used in Weaviate and Qdrant () we implement one for BigQuery. The idea of an adapter is to add destination specific table and column hinst using `x-<destination>-` table and column hints. In case of BigQuery we want to fully specify
- partitions
- clusters
- data retention
+ 2 more
Part of this ticket is to figure out the most popular settings of this kind and implement them. Ask our community for help :)

**Tasks**
1. * [x]  reuse Qdrant / Weaviate concept. Define and adapter that adds `x-` elements as columns definitions
2. * [x] design adapter function signature, write nice docstrings, place it in `dlt.destinations.adapters`
3. * [x] extend SQL builder in `bigquery.py` to interpret new hints.
4. * [x] add unit tests for to existing sql builder tests
5. * [x] add specialized adapter test that will create table with those hints
6. * [x] update the destination documentation

PR #900, opened 2024-01-18
Synapse destination

jorritsandbrink:
### Description
This PR adds initial support for the Synapse destination. ~~It only implements the `insert_values` file format. I'll add the `staging_file_format` `parquet` (using `COPY INTO` to load from Azure Blob Storage into Synapse) in an upcoming separate PR. That next PR will also contain user docs.~~ It implements the `insert_values` file format and the `staging_file_format` `parquet` (using `COPY INTO` to load from Azure Blob Storage into Synapse).

### Related Issues

- Implements #832 

### Additional Context

- I aimed to reuse the code for the `mssql` destination as much as possible.
- Since Synapse cannot handle multi row inserts, I introduced a new `insert_values_writer_type` called `select_union` that is based on `SELECT` and `UNION_ALL`. Relevant SO [here](https://stackoverflow.com/questions/73579485/azure-synapse-unable-to-insert-multiple-rows-into-table).
- ~~All tables are created as `HEAP` tables. This is the most robust choice as it always works (as opposed to `CLUSTERED COLUMNSTORE INDEX` tables, which do not support `varchar(max)`, `nvarchar(max)`, and `varbinary(max)` data types). See Synapse docs about [indexes](https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-tables-index) for context.~~
- The user can configure which table [index type](https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-tables-index) is used: `heap` or `clustered_columnstore_index`. The default is `heap`, because that's the most robust choice that always works with all data types. The default can be overridden by setting `destination.synapse.default_table_index_type` in a config TOML or by exporting the `DESTINATION__SYNAPSE__DEFAULT_TABLE_INDEX_TYPE` env var. All data tables will be created with the table index type specified as the default, unless another table index type is specified at the resource level. ~~The `table_index_type` argument can be provided to the `@dlt.resource` decorator to configure indexes at a more granular level.~~ The `x-table-index-type` argument can be provided to the `synapse_adapter` to configure indexes at a more granular level. Dlt system tables (the ones prefixed with `_dlt`) always have `heap` as index type, regardless of any configuration, because the recommendation is that "For small lookup tables, less than 60 million rows, consider using HEAP or clustered index for faster query performance.". ~~The setup for handling table index type is inspired by the way `table_format` is handled.~~
- I included support for `primary_key` and `unique` column hints, which are implemented with `PRIMARY KEY NONCLUSTERED NOT ENFORCED` and `UNIQUE NOT ENFORCED`, respectively. Because these [constraints](https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-table-constraints) are a little tricky (they are not enforced and can lead to innacurate results if the user does not ensure all column values are unique), they are **not** applied by default. The user would need to set `create_indexes` to `True` explicitly to apply them.
- Since long types (`text`, `ntext`, `image`) are not supported on Synapse, the `LongAsMax` keyword is set to `yes` so long types get converted to `MAX` types as described [here](https://stackoverflow.com/a/57926224). `LongAsMax` got introduced in `ODBC Driver 18 for SQL Server`. Hence, as opposed to for the `mssql` destination, `ODBC Driver 17 for SQL Server` is **not** supported for Synapse.
- The logic for byte literals in `escape_mssql_literal` has been adapted to make it work for Synapse. The old logic only worked for `mssql`, the new logic works for both `mssql` and `synapse`.
- **OUTDATED, NO LONGER APPLIES AFTER CODE CHANGES:** Synapse seems to struggle with concurrency in specific scenario's (completely hangs and need to execute `KILL` commands to release the stuck process). ~~Because of this, I had to set the `LOAD__WORKERS` env var to 1 for the test `tests/load/pipeline/test_replace_disposition.py::test_replace_table_clearing[staging-optimized-synapse-no-staging]`.~~ This is now handled automatically. When `auto_disable_concurrency` is set to `true` (the default value), the number of load workers automatically gets set to 1 at runtime in cases that are known to have issues with concurrency (for now, that's the case when there are tables with `write_disposition == replace` and `replace_strategy == staging-optimized`. A warning message is shown to the user in such cases, to make them aware of this intervention and to inform them of their configuration options.
- dbt with Synapse has multiple complications. Since I could not make it pass all the tests in `tests/load/pipeline/test_dbt_helper.py`, I set `supports_dbt` to False in the `DestinationTestConfiguration` for Synapse.

Points regarding the support for `staging_file_format` `parquet`:

- The implementation mostly follows `redshift`'s implementation for staged file loading.
- The `COPY INTO` [command](https://learn.microsoft.com/en-us/sql/t-sql/statements/copy-into-transact-sql?view=azure-sqldw-latest) is used to load the Parquet files into Synapse tables.
- `COPY INTO` does not support JSON, that's why only Parquet has been implemented.
- Authorizing access from the Synapse workspace to the Azure Storage Account can be done in two ways:
1) based on `AzureCredentialsWithoutDefaults`, using `azure_storage_sas_token`. The same credentials are used for both the write into the Storage Account, and the read from the Storage Account into Synapse. This is the default option.
2) using the Synapse workspace managed identity. This requires that the managed identity has been assigned the `Storage Blob Data Reader` role (or a role with higher permissions) on the Blob container (or at the Storage Account level). The user can opt in for managed identity authentication by setting `staging_use_msi` to `True`.
- Parquet `TIME` columns are not supported. Synapse reads those columns as `bigint`, and raises an incompatibility error as such. This might actually be caused by a [bug](https://github.com/apache/arrow/issues/32766) in the `pyarrow` Parquet writer, but I'm not sure about that.
- The Parquet staging files are **not** deleted after the load into the Synapse table has been completed. I have looked into this though—file deletion can relatively easily be implemented as a step after the `COPY INTO` (potentially optional similar to how `snowflake` uses `keep_staged_files`), but this would leave empty folders dangling behind. I think also deleting the folders would require orchestration at a higher level, which is less straightforward to implement.

Other changes in this PR:

- Added a proper way to skip `dbt`-specific tests for destinations that don't support `dbt`.


dlt/destinations/impl/synapse/synapse.py
```diff
@@ -0,0 +1,99 @@
...
+    "unique": "UNIQUE NOT ENFORCED",
+}
+
+
+class SynapseClient(MsSqlClient, SupportsStagingDestination):
+    capabilities: ClassVar[DestinationCapabilitiesContext] = capabilities()
+
+    def __init__(self, schema: Schema, config: SynapseClientConfiguration) -> None:
+        sql_client = SynapseSqlClient(config.normalize_dataset_name(schema), config.credentials)
+        InsertValuesJobClient.__init__(self, schema, config, sql_client)
```
sh-rp, line 33: what is this line about?

jorritsandbrink: The code mirrored `MsSqlClient.__init__`. While `MsSqlClient` can call `InsertValuesJobClient.__init__` through `super()`, `SynapseClient.__init__` needs to call it explicitly (because it inherits from `MsSqlClient`, and not from `InsertValuesJobClient` like `MsSqlClient` does).

That being said, I have added a new commit to simplify `SynapseClient.__init__`. Would you agree the new code is better?

dlt/common/schema/typing.py
```diff
@@ -165,6 +167,7 @@ class TTableSchema(TypedDict, total=False):
     columns: TTableSchemaColumns
     resource: Optional[str]
     table_format: Optional[TTableFormat]
+    table_index_type: Optional[TTableIndexType]
```
sh-rp, line 170: @rudolfix we can switch between 2 index types (https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-tables-index) and we have a new table hint here. I am wondering if we can re-use table_format or wether this needs to be a separate hint. I am not quite sure.

dlt/destinations/impl/synapse/configuration.py
```diff
@@ -0,0 +1,106 @@
...
+    auto_disable_concurrency: Optional[bool] = True
+    """Whether concurrency is automatically disabled in cases where it might cause issues."""
+
+    __config_gen_annotations__: ClassVar[List[str]] = [
+        "default_table_index_type",
+        "create_indexes",
+        "auto_disable_concurrency",
+    ]
+
+    def get_load_workers(self, tables: TSchemaTables, workers: int) -> int:
```
sh-rp, line 64: My take is, that this needs to be an interface for all destinations. So:
* Implement a method on JobClientBase "def enable_parallel_loading" which returns true as a default value
* Evaluate this in the Load class for each destination, and set workers to 1 if it returns false
* Implement this method for synapse, all the other fall back to true.

dlt/destinations/impl/synapse/synapse.py
```diff
@@ -0,0 +1,175 @@
...
+
+from dlt.common.schema import TTableSchema, TColumnSchema, Schema, TColumnHint
+from dlt.common.schema.typing import TTableSchemaColumns, TTableIndexType
+
+from dlt.destinations.sql_jobs import SqlStagingCopyJob, SqlJobParams
+from dlt.destinations.sql_client import SqlClientBase
+from dlt.destinations.insert_job_client import InsertValuesJobClient
+from dlt.destinations.job_client_impl import SqlJobClientBase
+
+from dlt.destinations.impl.mssql.mssql import (
```
sh-rp, line 16: @rudolfix is cross-importing between destinations ok for "destination families"?

dlt/destinations/impl/synapse/synapse.py
```diff
@@ -0,0 +1,175 @@
...
+    def _get_table_update_sql(
+        self, table_name: str, new_columns: Sequence[TColumnSchema], generate_alter: bool
+    ) -> List[str]:
+        table = self.get_load_table(table_name)
+        if table is None:
+            table_index_type = self.config.default_table_index_type
+        else:
+            table_index_type = table.get("table_index_type")
+            if table_index_type == "clustered_columnstore_index":
+                new_columns = self._get_columstore_valid_columns(new_columns)
```
sh-rp, line 62: I think instead of doing this here, you should probably implement _get_column_def_sql (extend it to also accept index type) and react to it in the type_mapper. see how we do it with athena tables.

dlt/destinations/impl/synapse/synapse.py
```diff
@@ -0,0 +1,175 @@
...
+            # dlt tables should always be heap tables, regardless of the user
+            # configuration. Why? "For small lookup tables, less than 60 million rows,
+            # consider using HEAP or clustered index for faster query performance."
+            # https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-tables-index#heap-tables
+            table["table_index_type"] = "heap"
+        if table["table_index_type"] is None:
+            table["table_index_type"] = self.config.default_table_index_type
+        return table
+
+    def get_storage_table_index_type(self, table_name: str) -> TTableIndexType:
```
sh-rp, line 125: I think you can move this to the tests if it is not used in production code

dlt/destinations/impl/synapse/synapse.py
```diff
@@ -0,0 +1,175 @@
...
+                f"ALTER SCHEMA {sql_client.fully_qualified_dataset_name()} TRANSFER"
+                f" {staging_table_name};"
+            )
+            # recreate staging table
+            # In some cases, when multiple instances of this CTAS query are
+            # executed concurrently, Synapse suspends the queries and hangs.
+            # This can be prevented by setting the env var LOAD__WORKERS = "1".
+            sql.append(
+                f"CREATE TABLE {staging_table_name}"
+                " WITH ( DISTRIBUTION = ROUND_ROBIN, HEAP )"  # distribution must be explicitly specified with CTAS
```
sh-rp, line 170: could be configurable, but maybe let's jsut stick with this until someone requests otherwise. you can add a note in the docs about it

sh-rp (changes requested):
some more small changes

dlt/common/schema/typing.py
rudolfix: this looks like a case for destination adapter. it sets per-destination options on a resource using x- hints. please look into:
https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/adapters.py
and bigquery adapter that @Pipboyguy is doing: https://github.com/dlt-hub/dlt/issues/855

dlt/destinations/impl/synapse/configuration.py
rudolfix: yes, this is good idea. we may need in a few other places (ie. sink destination if user wants to serialize processing).
I'd add it as an option to base configuration of each destination. (should we do it in this PR or a followup?)

dlt/destinations/impl/synapse/synapse.py
rudolfix: I think we already import duckdb in motherduck. so I say yes

rudolfix (commented):
@jorritsandbrink this is really good! also thanks for diving in into synapse (ie various types of indexes and union inserts)

dlt/common/schema/typing.py
jorritsandbrink: I rewrote the code to make use of a destination adapter. Looked at `qdrant_adapter` and `weaviate_adapter` for inspiration. The `synapse_adapter` is slightly different because it works with a table hint, not column hints like `qdrant` and `weaviate`. Since `x-table-index-type` is not defined on `TResourceHints`, I have to ignore mypy's `typeddict-unknown-key` errors. Is this what you had in mind @rudolfix?

dlt/destinations/impl/synapse/synapse.py
jorritsandbrink: Now that `table_index_type` (generic for all destinations) has been "downgraded" to `x-table-index-type` (Synapse-specific), would you still recommend this?

I ask because taking this approach would require fetching `x-table-index-type` from the table schema in `SqlJobClientBase._get_table_update_sql`, and then pass this value to `_get_column_def_sql(...)`, thus having to add `table_index_type` as an argument to all implementations for that function (similar to how `table_format` is handled). Destination-specific logic would end up in generic and other-destination-specific code (where it is unused).

dlt/destinations/impl/synapse/synapse.py
jorritsandbrink: Done.

dlt/destinations/impl/synapse/configuration.py
jorritsandbrink: This is no longer needed, at least not for what I used it for. I replaced the CTAS that caused the concurrency issues with regular `CREATE TABLE` statements. This makes it simpler (no need to manage the number of workers) and also faster because loads can be done concurrently.

dlt/destinations/impl/synapse/synapse.py
jorritsandbrink: I included a note in the docs.

jorritsandbrink:
@sh-rp and @rudolfix I addressed all the points you raised and requested a re-review from both of you.

dlt/destinations/impl/synapse/synapse.py
sh-rp: alright, let's keep it like this!

dlt/destinations/impl/synapse/synapse_adapter.py
```diff
@@ -0,0 +1,50 @@
...
+    resource = ensure_resource(data)
+
+    if table_index_type is not None:
+        if table_index_type not in TABLE_INDEX_TYPES:
+            allowed_types = ", ".join(TABLE_INDEX_TYPES)
+            raise ValueError(
+                f"Table index type {table_index_type} is invalid. Allowed table index"
+                f" types are: {allowed_types}."
+            )
+        resource._hints[TABLE_INDEX_TYPE_HINT] = table_index_type  # type: ignore[typeddict-unknown-key]
```
sh-rp, line 49: my take here is to do this with resource.apply_hints() but extend apply_hints to accept a new arg which is a dict which then get's merged into _hints. one of the other freelancers was asking about this (or was it you) so there may be some conflicts. @rudolfix wdyt.

sh-rp: so this new arg would be something like additional_hints: DictStrAny.

docs/website/docs/general-usage/full-loading.md
```diff
@@ -67,6 +67,4 @@ opportunities, you should use this strategy. The `staging-optimized` strategy be
   recreated with a [clone command](https://docs.snowflake.com/en/sql-reference/sql/create-clone) from the staging tables. This is a low cost and fast way to create a second independent table from the data of another. Learn
   more about [table cloning on snowflake](https://docs.snowflake.com/en/user-guide/object-clone).
 
-For all other destinations the `staging-optimized` will fall back to the behavior of the `insert-from-staging` strategy.
-
-
+For all other [destinations](../dlt-ecosystem/destinations/index.md), please look at their respective documentation pages to see if and how the `staging-optimized` strategy is implemented. If it is not implemented, `dlt` will fall back to the `insert-from-staging` strategy.
```
sh-rp, line 70: thanks for cleaning up these random parts all over the place, really appreciated :)

docs/website/docs/dlt-ecosystem/destinations/synapse.md
```diff
@@ -0,0 +1,209 @@
...
+## Supported column hints
+
+Synapse supports the following [column hints](https://dlthub.com/docs/general-usage/schema#tables-and-columns):
+
+* `primary_key` - creates a `PRIMARY KEY NONCLUSTERED NOT ENFORCED` constraint on the column
+* `unique` - creates a `UNIQUE NOT ENFORCED` constraint on the column
+
+> ❗ These hints are **disabled by default**. This is because the `PRIMARY KEY` and `UNIQUE` [constraints](https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/sql-data-warehouse-table-constraints) are tricky in Synapse: they are **not enforced** and can lead to innacurate results if the user does not ensure all column values are unique. For the column hints to take effect, the `create_indexes` configuration needs to be set to `True`, see [additional destination options](#additional-destination-options).
+
+## Load concurrency issue
```
sh-rp, line 153: this is outdated, no?

sh-rp: and you need to link this page in the sidebars.js. by the way you can see a preview of this here: https://deploy-preview-900--dlt-hub-docs.netlify.app/docs/dlt-ecosystem/destinations/synapse/

dlt/destinations/impl/synapse/synapse.py
```diff
@@ -0,0 +1,297 @@
...
+        if self.staging_use_msi:
+            credential = "IDENTITY = 'Managed Identity'"
+        else:
+            sas_token = staging_credentials.azure_storage_sas_token
+            credential = f"IDENTITY = 'Shared Access Signature', SECRET = '{sas_token}'"
+
+        # Copy data from staging file into Synapse table.
+        with self._sql_client.begin_transaction():
+            dataset_name = self._sql_client.dataset_name
+            sql = dedent(f"""
```
sh-rp, line 266: out of curiosity, is dedent needed here?

sh-rp (changes requested):
almost there, only a few small improvements left.

dlt/destinations/impl/synapse/synapse.py
jorritsandbrink: No. Functionally it doesn't make a difference, but I like the idea the query is properly formatted without unnecessary indentation.

dlt/destinations/impl/synapse/synapse_adapter.py
jorritsandbrink: I extended `apply_hints()` to handle table hints as you suggested. Also aligned with @Pipboyguy about this on Slack.

dlt/extract/hints.py
```diff
@@ -208,6 +209,14 @@ def apply_hints(
                 t["incremental"] = None
             else:
                 t["incremental"] = incremental
+        if additional_table_hints is not None:
+            # loop through provided hints and add, overwrite, or remove them
+            for k, v in additional_table_hints.items():
+                if v:
```
sh-rp, line 215: This has to be "if v is not None" otherwise a value of false or [] will also be removed although it needs to stay in.

jorritsandbrink: I adjusted it.

I had it like that to maintain consistency with how the standard table hints are handles (remove when "empty value" is provided), but I agree that isn't desired behavior for the additional table hints.

Issue #895, opened 2024-01-15
Loading behavior of optional fields is inconsistent when using nested pydantic models


codingcyclist:
### dlt version

0.4.2

### Describe the problem

```python
class Child(BaseModel):
    any_string: str
    optional_child_attribute: t.Optional[str] = None

class Parent(BaseModel):
    child: Child
    optional_parent_attribute: t.Optional[str] = None

    dlt_config: t.ClassVar[DltConfig] = {"skip_complex_types": True}
```
**Context:**
- I'm using `Parent` as a **nested** pydantic model to enforce a schema on a rescue. 
- I'm setting `skip_complex_types` on the parent model to `True` and `max_table_nesting` on the resource to `1` so that my child model does not get coerced into a single `complex` column or into several child tables, but flattened into several columns (i.e. `child__{child-column-name}`)
- Both the parent- and the child model have optional fields with well-defined, primitive data types

**Current behavior:**
The optional child model field won’t appear as a `child__optional_child_attribute` column in the destination until the first data comes through where this filed is not None. The optional parent attribute, however, appears as and `optional_parent_attribute` column from the first time when the resource gets loaded.

The problem here seems to be that the table schema has no knowledge about the data type of child model attributes as the schema is being generated before the flattening logic gets applied. 

### Expected behavior

Optional child model fields should always get loaded into the destination if they have a primitive data type

```python
# example source data
{
    "child": {"any_string": "hello_world"},
}

# desired format when loaded into destination
{
    "optional_parent_attribute": None,
    "child__any_string": "hello_world",
    "child__optional_child_attribute": None
}
```

### Steps to reproduce

See above

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

Custom (i.e. AWS EventBridge)

### dlt destination

Snowflake

rudolfix:
@sultaniman looking #901 we can't really fix how we map Pydantic models into our schema fully. It may require some conceptual change. It is a good outcome of the research.

Still we should fix this ticket and we can be (a little) hacky. We'll ignore recurring into lists for now (those generate dlt tables) but we can support nested pydantic models (that are not part of the list). We we see a Pydantic model like in @codingcyclist example below we should recursively create a dlt schema from it and then flatten this model into parent model. 

we use `make_path` of `NamingConvention` protocol to do that. You can take implementation from snake_case.py or create global instance to make paths. this is OFC a hack but will work for all our naming conventions as of now.

[...]

PR #901, opened 2024-01-19
Pydantic improvements

sultaniman:
This PR is to address the issue https://github.com/dlt-hub/dlt/issues/895 which is when pydantic column schema is provided.
Adjustment recursively calls `pydantic_to_table_schema_columns` to expand for sub model.


tests/load/pipeline/test_pydantic_models.py
```diff
@@ -0,0 +1,41 @@
...
+        child: Child
+        optional_parent_attribute: t.Optional[str] = None
+        dlt_config: t.ClassVar[DltConfig] = {"skip_complex_types": True}
+
+    example_data = {
+        "optional_parent_attribute": None,
+        "child": {"child_attribute": "any string", "optional_child_attribute": None},
+    }
+
+    @dlt.resource
```
codingcyclist, line 23: Re: my comment above, I think we should test for both cases (i.e. `skip_complex_types = {True/False}`)
```python
# Test Case 1: "skip_complex_types" = False -> both pydantic models and dicts should be coerced into complex column
class Child(BaseModel):
    child_attribute: str
    optional_child_attribute: t.Optional[str] = None


class Parent(BaseModel):
    optional_parent_attribute: t.Optional[str] = None
    child: Child
    dct: t.Dict[str, t.Any]
    dlt_config: t.ClassVar[DltConfig] = {"skip_complex_types": False}


example_data = {
    "optional_parent_attribute": None,
    "dct": {"child_attribute": "any string", "optional_child_attribute": None},
}

expected_loaded_data = {
    "child__child_attribute": "any string",
    "dct": '{"child_attribute": "any string", "optional_child_attribute": None}',
}


# Test Case 2 "skip_complex_types" = False -> both pydantic models and dicts should be flattened. Optional fields should be kept only for pydantic models
class Child(BaseModel):
    child_attribute: str
    optional_child_attribute: t.Optional[str] = None


class Parent(BaseModel):
    optional_parent_attribute: t.Optional[str] = None
    child: Child
    dct: t.Dict[str, t.Any]
    dlt_config: t.ClassVar[DltConfig] = {"skip_complex_types": True}


example_data = {
    "optional_parent_attribute": None,
    "dct__child_attribute": "any string",
    "child__child_attribute": "any string", 
    "child__optional_child_attribute": None,
}

expected_loaded_data = {"optional_parent_attribute": None, "dct__child_attribute": "any string"}

```

codingcyclist (changes requested):
Great stuff - I think the behavior between pydantic and other complex constructs is not 100% consistent yet in cases where `skip_complex_types = False`. I left some comments

dlt/common/libs/pydantic.py
```diff
@@ -121,14 +128,27 @@ def pydantic_to_table_schema_columns(
             # try to coerce unknown type to text
             data_type = "text"
 
-        if data_type == "complex" and skip_complex_types:
+        if not is_inner_type_pydantic_model and data_type == "complex" and skip_complex_types:
```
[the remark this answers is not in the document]
sultaniman: @codingcyclist of the following two the first is flattening a single field with Pydantic model the second one I think should have `data_type == "complex"`

```py
if is_inner_type_pydantic_model and skip_complex_types:
    ...
elif is_inner_type_pydantic_model and not skip_complex_types:
    ...
```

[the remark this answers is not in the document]
codingcyclist: Have you tired it? I tried it locally and the second line seems to have `data_type == "text"`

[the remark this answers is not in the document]
sultaniman: @codingcyclist I just got back to PR, I will adjust and incorporate your proposed changes and check it out as well 🤗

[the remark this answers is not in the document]
sultaniman: @sh-rp @codingcyclist  so the logic here is the following right?
1. If `skip_complex_types=True` and the field is a Pydantic model then we need to convert it to flattened representation,
2. If it is a Pydantic model but `skip_complex_types=False` then we mark it as complex,
3. And finally default logic if the `data_type=complex` and `skip_complex_types=True` then we skip this field altogether

tests/load/pipeline/test_pydantic_models.py
```diff
@@ -0,0 +1,41 @@
...
+    def res():
+        yield [example_data]
+
+    @dlt.source(max_table_nesting=1)
+    def src():
+        yield res()
+
+    p = dlt.pipeline("example", full_refresh=True, destination="duckdb")
+    p.run(src(), table_name="items", columns=Parent)
+
```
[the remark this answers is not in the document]
sultaniman: Added.

codingcyclist (dismissed):
Hooray, great stuff 🙌

dlt/common/libs/pydantic.py
```diff
@@ -121,14 +128,32 @@ def pydantic_to_table_schema_columns(
             # try to coerce unknown type to text
             data_type = "text"
```
rudolfix, line 129: please detect pydantic model in exception handler. if detected: data_type is complex and set the flag is_inner_type_pydantic_model

dlt/common/libs/pydantic.py
```diff
@@ -121,14 +128,32 @@ def pydantic_to_table_schema_columns(
...
 
-        if data_type == "complex" and skip_complex_types:
```
rudolfix, line 124: then you can leave this skip as it is!

dlt/common/libs/pydantic.py
```diff
@@ -121,14 +128,32 @@ def pydantic_to_table_schema_columns(
...
+        # This case is for a single field schema/model
+        # we need to generate snake_case field names
+        # and return flattened field schemas
+        if is_inner_type_pydantic_model and skip_complex_types:
```
rudolfix, line 134: hmmm I assume that it should be `is_inner_type_pydantic_model and not skip_complex_types`? if we have `skip_complex_types` we leave the task of inferring the complex type (no matter if there's a model or not) to the normalizer.
@codingcyclist what would you expect? if we have a pydantic model defined, should we ignore `skip_complex_types` or not?

rudolfix (changes requested):
IMO we have an edge case when there's pydantic model defined and we have `skip_complex_types` set to true. please see my review

dlt/common/libs/pydantic.py
sultaniman: then we set `data_type=complex` if it is pydantic model right?

dlt/common/libs/pydantic.py
sultaniman: @rudolfix I updated it this way. Will wait for @codingcyclist's feedback.

codingcyclist: @rudolfix you're exactly right, thank's for catching this one. It should be `is_inner_type_pydantic_model and not skip_complex_types`, exactly as I mentioned it in [my previous comment](https://github.com/dlt-hub/dlt/pull/901#discussion_r1470269082). This should also be caught by one of the two test cases that we have

rudolfix (changes requested):
@sultaniman this looks good now. and I know where the problems with duckdb come from. we have a special area in tests suite for local pipeline tests

tests/pipeline/test_pipeline_extra

you can move your tests there, our workflow will make sure that duckdb and pydantic are installed

tests/pipeline/test_pipeline_extra.py
```diff
@@ -200,3 +200,157 @@ def generic(start=8):
...
+            "child_attribute": "any string",
+            "optional_child_attribute": None,
+        },
+    }
+
+    @dlt.resource
+    def res():
+        yield [example_data]
+
+    @dlt.source(max_table_nesting=1)
```
rudolfix, line 314: please remove nesting level. they should have no impact on the tests.

tests/pipeline/test_pipeline_extra.py
```diff
@@ -200,3 +200,157 @@ def generic(start=8):
...
+
+    with p.sql_client() as client:
+        with client.execute_query("SELECT * FROM items") as cursor:
+            loaded_values = {
+                col[0]: val
+                for val, col in zip(cursor.fetchall()[0], cursor.description)
+                if col[0] not in ("_dlt_id", "_dlt_load_id")
+            }
+
+            assert loaded_values == {
```
rudolfix, line 329: please explain what the test does here. you get the unflattened dict and child (as complex types) due to the flag. it is easy to understand to you but to the reviewer it is not easy

rudolfix (changes requested):
logic LGTM!
@sultaniman this code asks for unit tests. we have unit tests for the mapper in libs. why you didn't add the tests there? please test all combinations explictly

sultaniman:
> logic LGTM! @sultaniman this code asks for unit tests. we have unit tests for the mapper in libs. why you didn't add the tests there? please test all combinations explictly

@rudolfix do you mean to add tests for [TypeMapper](https://github.com/dlt-hub/dlt/blob/a58d9739de10853eaab0ad840901e81c7e00a4c0/dlt/destinations/type_mapping.py#L8) in destinations?

rudolfix (approved):
LGTM! thanks for all the tests!

PR #902, opened 2024-01-21
fix: use truthy value in ternary since 0 cause div by zero

z3z1ma:
### Related Issues

- Fixes #...
- Closes #...
- Resolves #...


rudolfix (approved):
thanks!

PR #905, opened 2024-01-22
enable async generators as resources

sh-rp:
A simple implementation to support async generators in our resources and transformers. This implementation basically converts an async generator to an awaitable list, so it can be run in the futures pool. 

Implementation details: The way it is implemented now, the async generator will block one futures slot in the extraction until it is completed. The question is, should we aim at implementing something that will finish a future regularly after a certain batch size and yield control back to the main thread, but have some kind of info in the FuturePipeItem indicating that there still is an open async generator that it can be rescheduled? Somehow a follow-up future of sorts?


dlt/extract/pipe.py
```diff
@@ -791,6 +806,27 @@ def _get_source_item(self) -> ResolvablePipeItem:
         elif self._next_item_mode == "round_robin":
             return self._get_source_item_round_robin()
 
+    def _get_next_item_from_generator(
```
sh-rp, line 809: here there only is refactoring to get rid of non-dry code.

dlt/extract/utils.py
```diff
@@ -119,6 +119,25 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
...
 
 
+def wrap_async_generator(f: Any) -> Any:
+    """Wraps an async generator into a list with one awaitable"""
+
+    async def run() -> List[TDataItem]:
+        result: List[TDataItem] = []
+        try:
+            item: TDataItems = None
+            while item := await f.__anext__():
```
rudolfix, line 129: the problem with this thing is that it will consume the whole generator in memory and then return results once. so ie. when there's a transformer it will receive a full dataset. right?

maybe we need to send each f.__next__() to a future pool and wrap this in a function that returns both a next result (after await) and the async iterator itself? so you put it into futures pool again and return the current value

rudolfix (changes requested):
wrapping is very clever! but we need to yield item by item. we cannot collect all the items and yield at once...

sh-rp:
@rudolfix yes you are right, i was already thinking about that after submitting the PR. I have changed it now that the wrapped async generator returns an iterator of awaitables. There is some magic inside which essentially emulates this "yield none" behavior we implemented last summer to make round_robin extract work (I can point you to the commit I think if you don't remember). My remaining questions are:
* Should we maybe change the default item mode to round_robin? I don't think the newbie user will find out about this if they just start yielding stuff
* Should we make the chunk-size of this generator configurable, or are we happy with yielding single items? I suppose we can keep it like this and it will replicate the behavior of the regular generator.
* I have modified the function that evaluates the result of futures a bit to not have stopiteration exceptions be an error, we could also have a custom exception there if you think we should still treat this as an error in other cases. wdyt?

sh-rp:
If we follow the approach of adding a follow up future, which i think is doable, then parallelism I think will also work for the fifo mode, I somehow like my approach better though, because it does not modify the pipe code much.

dlt/extract/utils.py
```diff
@@ -119,6 +119,29 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
...
+        try:
+            return await gen.__anext__()
+        except StopAsyncIteration:
+            exhausted = True
+            raise
+        finally:
+            is_running = False
+
+    # it is best to use the round robin strategy here if multiple async generators are used in resources
+    while not exhausted:
```
rudolfix, line 138: I think we should use the same strategy that user wants? anyway. IMO if you have two async generators they will be executed in parallel in futures, right?

dlt/extract/utils.py
```diff
@@ -119,6 +119,29 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
...
+        while is_running:
+            yield None
```
rudolfix, line 140: 100% cpu usage here? this will be executed in tight loop right? if there's None we go for the next resource and we get None again so we go again etc.

btw. I think we are close and the idea with yielding None is good.

dlt/extract/utils.py
```diff
@@ -119,6 +119,29 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
...
+            return await gen.__anext__()
+        except StopAsyncIteration:
+            exhausted = True
+            raise
+        finally:
+            is_running = False
+
+    # it is best to use the round robin strategy here if multiple async generators are used in resources
+    while not exhausted:
+        while is_running:
```
rudolfix, line 139: hmmm maybe asyncio critical sections are better? https://docs.python.org/3.8/library/asyncio-sync.html (if they work here because this code is sync)

rudolfix (changes requested):

rudolfix:
@sh-rp 

> * Should we maybe change the default item mode to round_robin? I don't think the newbie user will find out about this if they just start yielding stuff
I'd say no. the default behavior is to evaluate resources one by another fully. I think it is good.
we can switch to round robin if there's an async generator in any of the pipes though. otherwise your current implementation will starve the async generators is there are normal ones. btw. there are a lot of nuances due to this hack with yielding None in a loop

> * Should we make the chunk-size of this generator configurable, or are we happy with yielding single items? I suppose we can keep it like this and it will replicate the behavior of the regular generator.
single item can be a list/batch. regular generators work the same way, no?

> * I have modified the function that evaluates the result of futures a bit to not have stopiteration exceptions be an error, we could also have a custom exception there if you think we should still treat this as an error in other cases. wdyt?
I do not see this modification. do you mean this?:
```
if isinstance(ex, StopAsyncIteration):
                return None
```
this is handling this new async iterator and LGTM

but yeah we should test if limiting async iterator still works. could you add a test with does `add_limit` on a resource that is async generator to make sure we close it properly before it is exhausted.

also please add a test where we have async and sync generator to see if we are not starving the async.

dlt/extract/utils.py
sh-rp: not if we emit a list of awaitables, then one list will be processed first, then the other (you can see that in the test)

sh-rp:
@rudolfix do you remember what the reasoning of these lines was?
```
        # if there are currently more sources than added initially, we need to process the new ones first
        if sources_count > self._initial_sources_count:
            return self._get_source_item_fifo()
```
This makes the round robin fall back to fifo if more pipes were added during extraction. I remember that I put this in, but i have no recollection why. I will slow down parallel execution as soon as dynamically creates pipes are present in async generator scenarios and I am wondering if we can remove it.

dlt/extract/utils.py
```diff
@@ -142,8 +178,12 @@ def wrap_resource_gen(
...
-        # always wrap generators and generator functions. evaluate only at runtime!
+
+    if (
+        inspect.isgeneratorfunction(inspect.unwrap(f))
+        or inspect.isgenerator(f)
+        or inspect.isasyncgenfunction(f)
+    ):
 
         def _partial() -> Any:
             # print(f"_PARTIAL: {args} {kwargs}")
```
sultaniman, line 189: Is this supposed to be here?

dlt/extract/resource.py
```diff
@@ -310,10 +310,16 @@ def _gen_wrap(gen: TPipeStep) -> TPipeStep:
...
+
+            # wrap async gen already here
+            if inspect.isasyncgen(gen):
+                gen = wrap_async_generator(gen)
+
             try:
                 for i in gen:  # type: ignore # TODO: help me fix this later
                     yield i
-                    count += 1
+                    if i is not None:
```
sh-rp, line 321: this line is needed for the async generator to work properly, it changes the behavior of the limit, but probably that is ok, i am not sure.

dlt/extract/pipe.py
```diff
@@ -773,91 +792,63 @@ def _resolve_futures(self) -> ResolvablePipeItem:
...
-
-    def _get_source_item_round_robin(self) -> ResolvablePipeItem:
         sources_count = len(self._sources)
         # no more sources to iterate
         if sources_count == 0:
             return None
-        # if there are currently more sources than added initially, we need to process the new ones first
-        if sources_count > self._initial_sources_count:
-            return self._get_source_item_current()
         try:
```
sh-rp, line 818: this function now nicely combines fifo and round_robin. in fifo mode it says on the first source and only ventures into the next ones if that returns none. It would be quite easy to switch it back to the old behavior though. I removed this part that switches from round robin to fifo in some cases as it does not really make sense anymore imho if fifo also can switch the source index.

dlt/extract/utils.py
```diff
@@ -119,6 +131,35 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
...
+                raise
+
+    # this generator yields None while the async generator is not exhausted
+    try:
+        while not exhausted:
+            while lock.locked():
+                yield None
+            yield run()
+    except GeneratorExit:
+        # clean up async generator
```
sh-rp, line 159: i don't think I need to clean up here, this can happen at the end of the loop in the pipe

tests/pipeline/test_resources_evaluation.py
```diff
@@ -0,0 +1,280 @@
...
+    pipeline_1.run(async_resource1().add_limit(13))
+
+    with pipeline_1.sql_client() as c:
+        with c.execute_query("SELECT * FROM table1") as cur:
+            rows = list(cur.fetchall())
+            assert len(rows) == 13
+
+
+@pytest.mark.parametrize("parallelized", [True, False])
+def test_async_decorator_experiment(parallelized) -> None:
```
sh-rp, line 192: this for now is just an experiment, but this could be the extension for the defer decorator. if you put it on a resource function, it will make every iteration run as a future, so quite cool i think.

dlt/extract/pipe.py
```diff
@@ -703,6 +716,12 @@ def stop_background_loop(loop: asyncio.AbstractEventLoop) -> None:
 
         # print("stopping loop")
         if self._async_pool:
+            # wait for all async generators to be closed
+            future = asyncio.run_coroutine_threadsafe(
+                self._async_pool.shutdown_asyncgens(), self._ensure_async_pool()
+            )
+            while not future.done():
+                sleep(self.futures_poll_interval)
```
sh-rp, line 724: not 100% certain i implemented this right, but i think i did. we need to shut down all the open gens in this pool.

tests/pipeline/test_resources_evaluation.py
sh-rp: PS: i know that i was not supposed to do this yet ;)

dlt/extract/utils.py
```diff
@@ -119,6 +131,38 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
     return meta_arg
 
 
+def wrap_async_generator(
+    gen: AsyncGenerator[TDataItems, None]
+) -> Generator[Awaitable[TDataItems], None, None]:
+    """Wraps an async generator into a list of awaitables"""
+    exhausted = False
+    busy = False
```
sh-rp, line 139: using an async.lock here will fail on python 3.8, and 3.9 because there is no current loop on the main thread. i am not sure why it is allowed in python 3.10. threading locks should not be used with coroutines as far as I understand.

tests/extract/test_extract_pipe.py
```diff
@@ -39,13 +39,13 @@ def get_pipes():
...
-    assert [pi.item for pi in _l] == [1, 2, 55, 56, 77, 88, 89, 3, 4, 11, 12, 13, 14, 15, 20, 21]
+    # items will be in order of the pipes, nested iterator items appear inline, None triggers rotation
+    assert [pi.item for pi in _l] == [1, 2, 55, 56, 3, 77, 88, 89, 4, 11, 12, 13, 14, 15, 20, 21]
 
     # round robin mode
     _l = list(PipeIterator.from_pipes(get_pipes(), next_item_mode="round_robin"))
-    # items will be round robin, nested iterators are fully iterated and appear inline as soon as they are encountered
-    assert [pi.item for pi in _l] == [1, 11, 20, 2, 12, 21, 55, 56, 77, 88, 89, 13, 3, 14, 4, 15]
+    # items will be round robin, nested iterators are integrated into the round robin
+    assert [pi.item for pi in _l] == [1, 11, 20, 2, 12, 21, 3, 13, 55, 4, 14, 56, 15, 77, 88, 89]
```
rudolfix, line 48: please test a case when we have too many nested iterators

rudolfix (changes requested):
thx for this! I think we fixed many flaws with parallelism, not only async gens
1. we need a test where exception happens during async generation. I have a set of such tests somewhere, please add this one
2. we need a test where large amount of nested iterators is created, see review
3. we should move defer decorator (we can alias it!) to separate PR. nevertheless I reviewed it :)

dlt/extract/pipe.py
```diff
@@ -613,8 +614,19 @@ def __next__(self) -> PipeItem:
...
-                self._sources.append(
-                    SourcePipeItem(item, pipe_item.step, pipe_item.pipe, pipe_item.meta)
+                self._sources.insert(
+                    0, SourcePipeItem(item, pipe_item.step, pipe_item.pipe, pipe_item.meta)
+                )
+                pipe_item = None
+                continue
+
+            # handle async iterator items as new source
+            if inspect.isasyncgen(item):
```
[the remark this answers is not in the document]
sh-rp: yeah you are totally right, this is a leftover from something else, I updated it in the other places too.

dlt/extract/utils.py
sh-rp: the problem is, that the "parent" generator gets closed before the last item is evaulated that we want to evaluate. so it has to stay this way. do you know what I mean?

sh-rp: But i just had a nice idea of how to close the generator in the right moment this way, which is a good improvement :)

dlt/extract/pipe.py
sh-rp: @rudolfix I am not sure what you mean with "async". there is a test called `def test_limit_async_resource() -> None:` which tests limiting async resources (which get turned into generators by the wrapper). Can you maybe explain?

sh-rp: Do you mean async iterators?

tests/pipeline/test_resources_evaluation.py
```diff
@@ -0,0 +1,285 @@
...
+
+        @wraps(f)
+        def _wrap(*args: Any, **kwargs: Any) -> Any:
+            exhausted = False
+            busy = False
+
+            gen = f(*args, **kwargs)
+            # unpack generator
+            if inspect.isfunction(gen):
+                gen = gen()
```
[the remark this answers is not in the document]
sh-rp: next pr!

tests/pipeline/test_resources_evaluation.py
```diff
@@ -0,0 +1,285 @@
...
+            # if we have an async gen, no further action is needed
+            if inspect.isasyncgen(gen):
```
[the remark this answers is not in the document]
sh-rp: next pr!

tests/pipeline/test_resources_evaluation.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+@pytest.mark.parametrize("parallelized", [True, False])
+def test_async_decorator_experiment(parallelized) -> None:
+    os.environ["EXTRACT__NEXT_ITEM_MODE"] = "fifo"
+    execution_order = []
+    threads = set()
+
+    def parallelize(f) -> Any:
+        """converts regular itarable to generator of functions that can be run in parallel in the pipe"""
```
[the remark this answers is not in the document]
sh-rp: let's move this stuff to a new PR as you suggested, I have marked the test as skipped and we can discuss there.

dlt/extract/pipe.py
```diff
@@ -773,91 +792,63 @@ def _resolve_futures(self) -> ResolvablePipeItem:
...
+                if first_evaluated_index == -1:
+                    first_evaluated_index = self._current_source_index
         except StopIteration:
             # remove empty iterator and try another source
-            self._sources.pop(self._round_robin_index)
+            self._sources.pop(self._current_source_index)
             # we need to decrease the index to keep the round robin order
-            self._round_robin_index -= 1
-            # since in this case we have popped an initial source, we need to decrease the initial sources count
-            self._initial_sources_count -= 1
```
sh-rp, line 859: this is actually a bug i would say..

dlt/extract/pipe.py
sh-rp: I implemented it this way (and corrected the decreasing of initial sources counter along the way). Now there is a diferentiation between fifo and strict_fifo, I think this is necessary to prevent a scenario as you describe it.

tests/extract/test_extract_pipe.py
```diff
@@ -603,6 +603,10 @@ def pass_gen(item, meta):
 
 
 def test_close_on_async_exception() -> None:
+    global close_pipe_got_exit, close_pipe_yielding
+    close_pipe_got_exit = False
+    close_pipe_yielding = False
```
sh-rp, line 608: globals need to be reset in each test, they will not recover between tests in one file!

tests/extract/test_extract_pipe.py
```diff
@@ -674,6 +699,33 @@ def raise_gen(item: int):
...
+
+        # will be closed by PipeIterator
+        try:
+            close_pipe_yielding = True
+            for i in range(0, 10000):
+                await asyncio.sleep(0.01)
+                yield i
+            close_pipe_yielding = False
+        # we have a different exception here
+        except asyncio.CancelledError:
```
sh-rp, line 718: i expected to also get a generatorexit here, but it seems to not be the case.

tests/extract/test_extract_pipe.py
sh-rp: maybe a couple more tests would be not so bad here, especially since it is a bit hard to read this test (ie to verify that it is correct) :)

tests/extract/test_extract_pipe.py
```diff
@@ -715,13 +717,14 @@ async def long_gen():
                 yield i
             close_pipe_yielding = False
         # we have a different exception here
-        except asyncio.CancelledError:
+        except GeneratorExit:
```
sh-rp, line 720: I don't think this works. the test fails now. The error on cancelling async gens is a cancellederror, not a generatorexit.

rudolfix: huh this was working for me. I was not cancelling async gen directly but the wrapped gen and propagating the exception. I'll check it out

dlt/extract/pipe.py
```diff
@@ -498,20 +504,20 @@ def __init__(
...
-        self._round_robin_index: int = -1
-        self._initial_sources_count: int = 0
         self._async_pool: asyncio.AbstractEventLoop = None
         self._async_pool_thread: Thread = None
         self._thread_pool: ThreadPoolExecutor = None
-        self._sources: List[SourcePipeItem] = []
+        self._sources = sources
         self._futures: List[FuturePipeItem] = []
-        self._next_item_mode = next_item_mode
+        self._next_item_mode: TPipeNextItemMode = next_item_mode
```
sultaniman, line 518: since it is annotated in the argument should we also remove the annotation here?

rudolfix: good catch! but somehow mypy sees this as string here so I forced the type (or maybe it is VSCode language server)

dlt/extract/pipe.py
```diff
@@ -773,91 +793,70 @@ def _resolve_futures(self) -> ResolvablePipeItem:
...
-                    return ResolvablePipeItem(item, step, pipe, meta)
+                self._current_source_index = (self._current_source_index - 1) % sources_count
+            while True:
+                # if we have checked all sources once and all returned None, then we can sleep a bit
+                if self._current_source_index == first_evaluated_index:
+                    sleep(self.futures_poll_interval)
+                # get next item from the current source
+                gen, step, pipe, meta = self._sources[self._current_source_index]
+                set_current_pipe_name(pipe.name)
+                if (item := next(gen)) is not None:
```
sultaniman, line 837: imo this already will check for not None as `item := next(gen)`

dlt/extract/pipe.py
```diff
@@ -773,91 +793,70 @@ def _resolve_futures(self) -> ResolvablePipeItem:
...
+                        # keep the item assigned step and pipe when creating resolvable item
+                        if isinstance(item, DataItemWithMeta):
+                            return ResolvablePipeItem(item.data, step, pipe, item.meta)
+                        else:
+                            return ResolvablePipeItem(item, step, pipe, meta)
+                # remember the first evaluated index
+                if first_evaluated_index is None:
+                    first_evaluated_index = self._current_source_index
+                # always go round robin if None was returned
+                self._current_source_index = (self._current_source_index - 1) % sources_count
```
sultaniman, line 852: should we also check if `self._current_source_index > 0`?

dlt/extract/utils.py
```diff
@@ -119,6 +132,45 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
     return meta_arg
 
 
+def wrap_async_iterator(
+    gen: AsyncIterator[TDataItems],
+) -> Generator[Awaitable[TDataItems], None, None]:
+    """Wraps an async generator into a list of awaitables"""
+    exhausted = False
+    busy = False
```
sultaniman, line 140: is it better to pass some object instance to `run` function which will contain the state the variables?
```py
    exhausted = False
    busy = False
```

tests/extract/test_extract_pipe.py
rudolfix: @z3z1ma right! but we wrap async generator in regular generator so I expect GeneratorExit here. if we get both it means that we are leaking some edge cases ie. when pool is closed and we cancel async generators using asyncio module. I'll investigate in separate branch

dlt/extract/pipe.py
rudolfix: no! check how the modulo operator works for negative numbers :)

dlt/extract/utils.py
```diff
@@ -119,6 +132,45 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
...
+        # on stop iteration mark as exhausted
+        # also called when futures are cancelled
+        except StopAsyncIteration:
+            exhausted = True
+            raise
+        finally:
+            nonlocal busy
+            busy = False
+
+    # this generator yields None while the async generator is not exhausted
```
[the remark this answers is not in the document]
rudolfix: in a sense. most of this PR was to handle exactly this case. it is done upstream, if we see that all generators want to idle and we have no more data we sleep and yield control to other threads

rudolfix (approved):
LGTM! thanks @sh-rp and everyone else that participated and gave us ideas to implement that!

PR #906, opened 2024-01-23
feat(csv_reader): separate a method for csv_reader

IlyaFaer:
A PR additional to the https://github.com/dlt-hub/verified-sources/pull/319

Splits a method into two to allow auth of a DuckDB CSV reader.


rudolfix (approved):
LGTM!

PR #912, opened 2024-01-26
feat(filesystem): add compression flag if the read file is GZ

IlyaFaer:
This changes is needed for the csv_duckdb_reader:

https://github.com/dlt-hub/verified-sources/pull/319


dlt/common/storages/fsspec_filesystem.py
```diff
@@ -185,7 +185,12 @@ def open(self, mode: str = "rb", **kwargs: Any) -> IO[Any]:  # noqa: A003
                 **text_kwargs,
             )
         else:
-            opened_file = self.fsspec.open(self["file_url"], mode=mode, **kwargs)
+            opened_file = self.fsspec.open(
+                self["file_url"],
+                mode=mode,
+                compression="gzip" if self["file_url"].endswith(".gz") else None,
```
IlyaFaer, line 191: This line is the only actual change - set a compression flag for the `gz` files. Tried it locally with the csv_duckdb_reader - works fine

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -185,7 +185,12 @@ def open(self, mode: str = "rb", **kwargs: Any) -> IO[Any]:  # noqa: A003
                 **text_kwargs,
             )
         else:
-            opened_file = self.fsspec.open(self["file_url"], mode=mode, **kwargs)
+            opened_file = self.fsspec.open(
```
rudolfix, line 188: yeah this is close to what we want. but please allow people to switch this behavior on or off:
- add an argument `compression` with `auto` (default - behavior exactly as below), "enable", "disable" to hardcore enable / diable

dlt/common/storages/fsspec_filesystem.py
rudolfix: ok this is good (for auto mode, but please use mime type (it is a part of FileDict) to detect compression type (does fsspec support anything else?)

rudolfix (changes requested):

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -157,23 +172,48 @@ def fsspec(self) -> AbstractFileSystem:
...
         """
+        if compression == "auto":
+            compression_arg = "gzip" if self["encoding"] == "gzip" else None
+        elif compression == "enable":
+            compression_arg = "gzip"
+        elif compression == "disable":
+            compression_arg = None
+        else:
+            raise ValueError("""The argument `compression` must have one of the following values:
+                "auto", "enable", "disable".""")
```
IlyaFaer, line 205: With these tabs in in-line format, the message will look like:
```
The argument `compression` must have one of the following values:                "auto", "enable", "disable".
```

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -237,22 +281,29 @@ def glob_files(
...
             " 2023.9.0 or later"
         )
 
     for file, md in glob_result.items():
         if md["type"] != "file":
             continue
+        # filter out weird files that look like folders
+        size = int(md["size"])
+        if file.endswith("/") and size == 0:
+            continue
```
IlyaFaer, line 294: I think this part may be confusing. There are no folders in buckets actually. And even if there is a strange object in a bucket, I think it still should be shown, if it fits the user's glob. Hiding them seems like unexpected behavior, because of which the user may stuck for some time.

IlyaFaer: It's required to update the test files on `gs`. See this failure:

image

It's only on `gs` bucket. But previously I saw the same failures on S3 and Azure, and I fixed it. There are two things needed:

1. In the file `tests/filesystem/samples/csv/freshman_lbs.csv` the headline must be like this:
```
"Sex","Weight(lbs,Sep)","Weight(lbs,Apr)","BMI(Sep)","BMI(Apr)"
```
2. Just delete this strange folder-like file that ends with `/`. On S3 I found it after you wrote, and deleted it, keeping only the `tests/filesystem/samples/gzip/taxi.csv.gz`, and it worked fine.

I did both of these things on S3 and Azure, and I would do on GS, but those credentials I have don't work, so I just don't have access to the GS bucket. However, this is not a functionality problem, there is just some mess in the test files on GS bucket.

PR #916, opened 2024-01-29
D#/fix pipeline schema command

sh-rp:
Fixes implementation of "dlt pipeline <my_pl> schema" cli command to respect schema format and remove_default flags.


sultaniman (dismissed):
🚀 LGTM

PR #918, opened 2024-01-30
fix bucket tests

sh-rp:
There are some new files in our test buckets that mess up the existing tests, so now we just make sure all the known files exist and can be accessed.

### DIFF
--- a/tests/common/storages/utils.py
+++ b/tests/common/storages/utils.py
@@ -30,7 +30,26 @@ def assert_sample_files(
     config: FilesystemConfiguration,
     load_content: bool,
 ) -> None:
+    minimally_expected_file_items = {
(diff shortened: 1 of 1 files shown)


sultaniman (approved):
💃 LGTM

tests/common/storages/utils.py
```diff
@@ -49,6 +68,7 @@ def assert_sample_files(
         with file_dict.open() as f:
             assert content == f.read()
         # read via various readers
+        print(item)
```
rudolfix, line 71: @sh-rp @sultaniman please do not let debug prints into the `devel`. if this goes into release we'll need a quick post release to fix it. and this is PITA :)

PR #922, opened 2024-01-31
test(filesystem): drop folders ignore

IlyaFaer:


rudolfix (approved):
@IlyaFaer @sh-rp 
I was able to reproduce this problem: if you use "create folder" from google cloud UI to create folder in a bucket and then place file into it you'll get this weird size 0 file. if you upload a folder with files you won't so IMO there's some kind of a problem with fsspec.
for now LGTM

PR #923, opened 2024-01-31
SCD2 Experiment

sh-rp:
### Description
Proof of concept for scd2 implementation. This is an implementation of scd2 as described here: https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row with a valid_from and valid_until column. In this prototype we treat scd2 as a merge_strategy rather than it's own write_disposition, given the similarities between scd2 and merge this might make sense, we could also introduce a new write disposition.

Loading with this strategy always assumes a full dataset sync from the source, this way we detect missing columns and mark them as not valid anymore. 

Child tables get handled the same way as their parents. All tables are retained but gain validity columns. This part still needs to be hammered out.

### Open questions and ideas
* Where is the right place to extend the schema if additional columns are needed as in this case?
* Is it safe to convert the _dlt_id into a hash of the row in this scenario? Since we merge on this key in the merge job, there will never be two of the same _dlt_ids.
* How do we integrate soft delete columns into this?
* We could also have a scd2 mode that does not need a full sync but will not mark tables as deleted if there is no new value as a compromise.

### DIFF
--- a/dlt/common/normalizers/json/relational.py
+++ b/dlt/common/normalizers/json/relational.py
@@ -149,6 +149,14 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
     def _add_row_id(
         self, table: str, row: TDataItemRow, parent_row_id: str, pos: int, _r_lvl: int
     ) -> str:
+        # sometimes row id needs to be hash for now hardcode here
+        cleaned_row = {k: v for k, v in row.items() if not k.startswith("_dlt")}
+        row_hash = digest128(json.dumps(cleaned_row, sort_keys=True))
+        row["_dlt_id"] = row_hash
+        if _r_lvl > 0:
+            DataItemNormalizer._link_row(cast(TDataItemRowChild, row), parent_row_id, pos)
+        return row_hash
+
         # row_id is always random, no matter if primary_key is present or not
         row_id = generate_dlt_id()
         if _r_lvl > 0:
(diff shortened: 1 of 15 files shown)


dlt/destinations/job_client_impl.py
```diff
@@ -224,7 +224,14 @@ def _create_append_followup_jobs(self, table_chain: Sequence[TTableSchema]) -> L
         return []
 
     def _create_merge_followup_jobs(self, table_chain: Sequence[TTableSchema]) -> List[NewLoadJob]:
-        return [SqlMergeJob.from_table_chain(table_chain, self.sql_client)]
+        now = pendulum.now()
```
sh-rp, line 227: I would say we need a cutoff date that is the same for the whole load. Probably we should take the timestamp when the loadpackage was created or something like this, i am just not sure if this is saved somewhere atm.

dlt/common/schema/schema.py
```diff
@@ -240,6 +240,15 @@ def coerce_row(
                         updated_table_partial["columns"] = {}
                     updated_table_partial["columns"][new_col_name] = new_col_def
 
+        # insert columns defs for scd2 (TODO: where to do this properly, maybe in a step after the normalization?)
```
sh-rp, line 243: where is the right place to update the schema based on destination settings? maybe we should do this at the beginning of the load step? I am not sure.

dlt/common/normalizers/json/relational.py
```diff
@@ -148,6 +149,14 @@ def _extend_row(extend: DictStrAny, row: TDataItemRow) -> None:
     def _add_row_id(
         self, table: str, row: TDataItemRow, parent_row_id: str, pos: int, _r_lvl: int
     ) -> str:
+        # sometimes row id needs to be hash for now hardcode here
```
sh-rp, line 152: for scd2 it makes the most sense to have a row_hash in the row_id..

PR #932, opened 2024-02-05
feat(filesystem): implement Google Drive source

IlyaFaer:
Towards: https://github.com/dlt-hub/verified-sources/issues/308

I took the original implementation, moved it into the `dlt` repo and fixed several small problems, which were preventing the driver from working with Google Drive. Most of the original driver code is left untouched - even though I think some parts of it are not needed (file uploading, deleting, etc.)


dlt/common/storages/fsspec_filesystem.py
```diff
@@ -276,6 +281,7 @@ def glob_files(
 
     bucket_path = bucket_url_parsed._replace(scheme="").geturl()
     bucket_path = bucket_path[2:] if bucket_path.startswith("//") else bucket_path
+    bucket_path = bucket_path.split("?")[0]
```
IlyaFaer, line 284: Query params (`RootId=xxx`) should be dropped before using the `bucket_path` further

rudolfix (changes requested):
the biggest missing piece are tests:
the minimum option: we need to test glob functionality
the tests are in tests/load/filesystem
1. add gdrive bucket to `test_filesystem_dict` 
2. buckets are defined in tests/load/utils
3. bucket url is defined in tests/.dlt/config.toml, you can use https://drive.google.com/drive/u/0/folders/15eC3e5MNew2XAIefWNlG8VlEa0ISnnaG as the bucket url (contains `standard_source` inside)
NOTE: if you modify `DEFAULT_BUCKETS` to include gdrive, all other tests (that test write files) will pick it up.

in theory they should all pass so you may try it out, to pass them may be our maximum plan (and maybe that's the easiest). otherwise you need to create another fixture...

tests/common/storages/utils.py
```diff
@@ -49,7 +49,8 @@ def assert_sample_files(
 
     for item in all_file_items:
         # only accept file items we know
-        assert item["file_name"] in minimally_expected_file_items
+        if not item["file_name"] in minimally_expected_file_items:
+            continue
```
IlyaFaer, line 53: That seems like a very brutal check. In the Google Drive case it doesn't work, because there is another folder in the bucket (in addition to the `samples`).

dlt/common/configuration/specs/gdrive_credentials.py
```diff
@@ -0,0 +1,569 @@
+import re
+import os
+from urllib.parse import urlparse, parse_qs
+from typing import Any, Dict, List, Literal, Optional
+
+from fsspec import register_implementation
+from fsspec.spec import AbstractFileSystem, AbstractBufferedFile
+from googleapiclient.discovery import build
```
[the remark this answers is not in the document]
IlyaFaer: Done

pyproject.toml
```diff
@@ -76,6 +76,8 @@ weaviate-client = {version = ">=3.22", optional = true}
 adlfs = {version = ">=2022.4.0", optional = true}
 pyodbc = {version = "^4.0.39", optional = true}
 qdrant-client = {version = "^1.6.4", optional = true, extras = ["fastembed"]}
+pydata-google-auth = "^1.8.2"
```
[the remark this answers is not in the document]
IlyaFaer: Done

dlt/common/configuration/specs/__init__.py
```diff
@@ -21,6 +21,7 @@
 from .api_credentials import OAuth2Credentials
 from .aws_credentials import AwsCredentials, AwsCredentialsWithoutDefaults
 from .azure_credentials import AzureCredentials, AzureCredentialsWithoutDefaults
+from .gdrive_credentials import GoogleDriveFileSystem
```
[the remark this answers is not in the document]
IlyaFaer: You mean to move the whole file `gdrive_credentials`? Okay, done

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -50,6 +50,9 @@ def test_filesystem_destination_configuration() -> None:
 @pytest.mark.parametrize("layout", ALL_LAYOUTS)
 def test_successful_load(write_disposition: str, layout: str, default_buckets_env: str) -> None:
     """Test load is successful with an empty destination dataset"""
+    if default_buckets_env.startswith("gdrive://"):
+        pytest.skip("GDrive is not supported for this test")
```
IlyaFaer, line 54: Temporary skip until the GDrive starts to work as a destination as well

rudolfix: I think it would be better to create another fixture along `default_buckets_env` that contains gdrive and use it in dict test. ie `source_buckets_env` and then remove gdrive from default. otherwise you'll need to add this condition to all destination tests

IlyaFaer: Hm, sounds cunning. Done

IlyaFaer:
There is a bunch of failures, but at the first glance they do not seem related. Checking...

image

dlt/common/storages/fsspecs/google_drive.py
```diff
@@ -367,17 +162,30 @@ def makedirs(self, path: str, exist_ok: Optional[bool] = True) -> None:
...
             else:
                 raise FileExistsError(path)
-        if self._parent(path):
+        if len(path.split("/")) > 2:
             self.makedirs(self._parent(path), exist_ok=True)
         self.mkdir(path, create_parents=False)
 
+    def _is_path_root_id(self, path: str) -> bool:
+        """Checks if path is root id"""
+        return len(path.split("/")) == 0
```
IlyaFaer, line 154: I don't think that's a correct statement. `split()` returns a list with at least one element - with no separator, it'll still have `len() == 1`.

image

IlyaFaer (commented):
Well, at the first gaze, there is one suspicious item about `split()`.
Also, there is a bunch of changes to Athena, BigQuery and Snowflake.

dlt/common/storages/fsspecs/google_drive.py
rudolfix: you are so right. that needs to be fixed

rudolfix (approved):
LGTM!

PR #938, opened 2024-02-06
add mark function to emit resource hints from decorated function

rudolfix:
### Description
1. adds `dlt.mark.with_hints` which allow to create/update resource hints from within decorated function. allows lazy creation of table schema during extract process
2. allows several current functions: to get source instance and resource name from within decorated function


dlt/extract/hints.py
```diff
@@ -40,10 +38,66 @@ class TResourceHints(TypedDict, total=False):
     merge_key: TTableHintTemplate[TColumnNames]
     incremental: Incremental[Any]
     schema_contract: TTableHintTemplate[TSchemaContract]
+    table_format: TTableHintTemplate[TTableFormat]
     validator: ValidateItem
     original_columns: TTableHintTemplate[TAnySchemaColumns]
 
 
+class HintsMeta:
+    __slots__ = "hints"
```
sh-rp, line 47: i never heard of this before, this is purely here for performance reasons, right?

dlt/extract/hints.py
```diff
@@ -176,7 +232,12 @@ def apply_hints(
                     # normalize columns
                     columns = ensure_table_schema_columns(columns)
                     # this updates all columns with defaults
+                    print("PREV")
```
sh-rp, line 235: dev code

dlt/extract/hints.py
```diff
@@ -202,8 +263,13 @@ def apply_hints(
                 )
                 if schema_contract is not None:
                     t["schema_contract"] = schema_contract
+            if table_format is not None:
```
sh-rp, line 266: this is kind of unrelated, but what about a utility function that does all of the table_format updating (and all other props that behave the same way) in one line? this makes it all more readable and you can also use it in make_hints (although the popping part is not needed there really)

sh-rp (changes requested):
maybe add a small note in the docs somewhere?

sh-rp (approved):

Issue #753, opened 2023-11-10
allow custom `DltResource` impl in `dlt.resource` and `dlt.transformer` decorators


rudolfix:
`dlt.source` decorator allows for custom `DltSource` implementation by setting `_impl_cls` argument. We want the same feature for resource and transformer decorators

rudolfix:
done

PR #942, opened 2024-02-07
annotations experiment

sh-rp:
### Description

This PR has an example of how we could leverage python annotations to have a more universal and more readable format of our schema.

Check out test_pipeline.py on what the interface for the user would be in this example. It actually already works. Basically all you need to to is to put our annotations on class vars, and you can use any new or pre-existing class as a basis for our schema. Typehints that are unknown to us will be ignored (as defined per "Annotated" PEP)

#### Notes and Considerations:
* Not all hints and data_types are support here, this is a prototype, but one that is easy to extend
* There is a meta data attribute "__table__" on the class, we can use this to set table level hints, currently the dlt core is not set up to support this via the columns attribute, but I think it should be quite easy to do.

### DIFF
--- a/dlt/common/schema/annotations.py
+++ b/dlt/common/schema/annotations.py
@@ -1,0 +1,162 @@
+import inspect
+
+from typing import Type, List, Any, Tuple, Union, Dict
+from typing_extensions import Annotated, get_origin, get_args
+from decimal import Decimal
+
+from dataclasses import dataclass
+from dlt.common.schema.typing import TTableSchema, TColumnSchema, TWriteDisposition, TDataType
+from datetime import datetime, date
+
+
+#
+# base dataclasses used for hints
+#
+@dataclass
+class BoolValue:
+    value: bool
+
+
+@dataclass
+class StringValue:
+    value: str
+
+
+@dataclass
+class IntValue:
+    value: int
+
+
+@dataclass
+class StrListValue:
+    value: List[str]
+
+
+#
+# Column Hints
+#
+class PrimaryKey(BoolValue): ...
+
+
+class Unique(BoolValue): ...
+
+
+#
+# Table Hints
+#
+class TableName(StringValue): ...
+
+
+class Description(StringValue): ...
+
+
+class Classifiers(StringValue): ...
+
+
+@dataclass
+class WriteDisposition:
+    value: TWriteDisposition
+
+
+#
+# Converters
+#
+
+TypeMap: Dict[Any, TDataType] = {
+    str: "text",
+    int: "bigint",
+    float: "double",
+    date: "date",
+    datetime: "time",
+    bool: "bool",
+    Decimal: "decimal",
+}
+
+
+def unwrap(t: Type[Any]) -> Tuple[Any, List[Any]]:
+    """Returns python type info and wrapped types if this was annotated type"""
+    if get_origin(t) is Annotated:
+        args = get_args(t)
+        return args[0], list(args[1:])
+    return t, []
+
+
+def to_full_type(t: Type[Any]) -> TColumnSchema:
+    result: TColumnSchema = {}
+    if get_origin(t) is Union:
+        for arg in get_args(t):
+            if arg is type(None):
+                result["nullable"] = True
+            else:
+                result["data_type"] = TypeMap.get(arg, None)
+    else:
+        result["data_type"] = TypeMap.get(t, None)
+    return result
+
+
+def to_table_hints(h: List[Type[Any]]) -> TTableSchema:
+    result: TTableSchema = {}
+    for hint in h:
+        if isinstance(hint, TableName):
+            result["name"] = hint.value
+        elif isinstance(hint, WriteDisposition):
+            result["write_disposition"] = hint.value
+    return result
+
+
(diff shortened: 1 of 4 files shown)


test_pipeline.py
```diff
@@ -0,0 +1,37 @@
...
+    p = dlt.pipeline("my_pipe", destination="duckdb", full_refresh=True)
+
+    data = [{
+        "id": "my_id"
+    }]
+
+    # run simple pipeline and see wether schema was used
+    load_info = p.run(data, columns=Items, table_name="blah")
+    print(load_info)
+    print(p.default_schema.to_pretty_yaml())
```
sh-rp, line 35: The items example will produce this segment in the final schema:
```
  blah:
    columns:
      id:
        data_type: text
        primary_key: true
        unique: true
      name:
        data_type: text
        nullable: true
        x-classifiers:
        - pii.name
      email:
        data_type: text
        nullable: true
        unique: true
        x-classifiers:
        - pii.email
      likes_herring:
        data_type: bool
        x-classifiers:
        - pii.food_preference
      _dlt_load_id:
        data_type: text
        nullable: false
      _dlt_id:
        data_type: text
        nullable: false
        unique: true
    write_disposition: append
```

dlt/common/schema/annotations.py
```diff
@@ -0,0 +1,162 @@
...
+class WriteDisposition:
+    value: TWriteDisposition
+
+
+#
+# Converters
+#
+
+TypeMap: Dict[Any, TDataType] = {
+    str: "text",
```
rudolfix, line 66: we have py_type_to_sc_type. look at the type_helpers.py. tons of edge cases are handled there when converting types

dlt/common/schema/annotations.py
```diff
@@ -0,0 +1,162 @@
...
+    float: "double",
+    date: "date",
+    datetime: "time",
+    bool: "bool",
+    Decimal: "decimal",
+}
+
+
+def unwrap(t: Type[Any]) -> Tuple[Any, List[Any]]:
+    """Returns python type info and wrapped types if this was annotated type"""
```
rudolfix, line 77: look at typing.py in common, I think I have similar function

test_pipeline.py
```diff
@@ -0,0 +1,37 @@
...
+
+from dlt import annotations as a
+from dlt.common import json
+from typing_extensions import Annotated, Never, Optional
+
+
+class Items:
+
+    # metadata for the table, currently not picked up by the pipeline
+    __table__: Annotated[Never, a.TableName("my_items"), a.WriteDisposition("merge")]
```
rudolfix, line 11: hmmm there must be a better way! ie.
load_info = p.run(data, columns=Annotated[Items, a.TableName("my_items"), a.WriteDisposition("merge")] , table_name="blah")

if items are not annotated then we go to default: Items as name, append and write dispositions

rudolfix: also using `columns` for the above is a kind of abuse... let's sync on table level hints. they may require to change our core library. ie. to make resource a generic that takes model as T. also see: https://github.com/dlt-hub/dlt/issues/753

rudolfix: or we add "model" to resource definition. but then we are going into a big overhaul of our schema system where relational and python schemas are different.

test_pipeline.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # primary keys
+    id: Annotated[str, a.PrimaryKey, a.Unique]
+
+    # additional columns
+    name: Annotated[Optional[str], a.Classifiers(["pii.name"])]
```
rudolfix, line 17: I'd probably generate literals for those

rudolfix (changes requested):
this is soo cool

dlt/common/schema/annotations.py
```diff
@@ -0,0 +1,162 @@
...
+    if get_origin(t) is Annotated:
+        args = get_args(t)
+        return args[0], list(args[1:])
+    return t, []
+
+
+def to_full_type(t: Type[Any]) -> TColumnSchema:
+    result: TColumnSchema = {}
+    if get_origin(t) is Union:
+        for arg in get_args(t):
```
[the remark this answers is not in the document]
sh-rp: we can only have one type in the schema, so either we have a default way of resolving if there are multiple types or we throw an error.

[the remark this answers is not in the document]
sh-rp: Or types of int and string will produce a string, but I think that is taking it to far for now.

test_pipeline.py
sh-rp: you mean for the classifiers?

dlt/common/schema/annotations.py
sh-rp: ah yes, great :) I was wondering if we had something like this somewhere.

dlt/common/schema/annotations.py
sh-rp: yes, i extracted the essential part.

test_pipeline.py
sh-rp: i changed it to accept annotated tables now, so this would work.

PR #951, opened 2024-02-08
Handle UnionType when checking is_union_type and is_optional_type

sultaniman:
We have `typing.Union` and `types.UnionType` and a new union and an optional syntax which came in Python 3.10.
So using for example `str | None` will return `UnionType` for which we don't check during schema hint extraction/generation

```py
>>> type(str | int)
<class 'types.UnionType'>
>>> type(str | None)
<class 'types.UnionType'>
>>> type(Union[int, str])
<class 'typing._GenericAlias'>
```

## TODO
* [x] Tests

### DIFF
--- a/dlt/common/libs/pydantic.py
+++ b/dlt/common/libs/pydantic.py
@@ -140,7 +140,7 @@ def pydantic_to_table_schema_columns(
             # This case is for a single field schema/model
             # we need to generate snake_case field names
             # and return flattened field schemas
-            schema_hints = pydantic_to_table_schema_columns(field.annotation)
+            schema_hints = pydantic_to_table_schema_columns(inner_type)
 
             for field_name, hints in schema_hints.items():
                 schema_key = snake_case_naming_convention.make_path(name, field_name)
--- a/dlt/common/typing.py
+++ b/dlt/common/typing.py
@@ -2,6 +2,7 @@ from collections.abc import Mapping as C_Mapping, Sequence as C_Sequence
 from datetime import datetime, date  # noqa: I251
 import inspect
 import os
+import sys
 from re import Pattern as _REPattern
 from typing import (
     ForwardRef,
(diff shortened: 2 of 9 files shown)


dlt/common/typing.py
```diff
@@ -25,6 +26,13 @@
     runtime_checkable,
     IO,
 )
+
+if sys.version_info >= (3, 10):
+    from types import UnionType
+else:
+    # it is defined as type(int | str)
```
codingcyclist, line 33: NP: Mabe make the comment a bit more verbose to make it clear what we're doing here: UnionType has only been introduced in Python 3.10, so we need to override it for older python version for backward compatibility

codingcyclist (commented):
Nice one 😎 ! I'm not sure what's the best way to test this? Maybe it works if we patch `sys.version_info` with `return_value = (3,9)`?

dlt/common/typing.py
sultaniman: Sure, thanks for pointing this out. It is an important thing imo to avoid confusion in the future.

sultaniman: @codingcyclist please check again :)

dlt/common/libs/pydantic.py
```diff
@@ -140,7 +140,7 @@ def pydantic_to_table_schema_columns(
             # This case is for a single field schema/model
             # we need to generate snake_case field names
             # and return flattened field schemas
-            schema_hints = pydantic_to_table_schema_columns(field.annotation)
+            schema_hints = pydantic_to_table_schema_columns(inner_type)
```
sultaniman, line 143: Here I am now passing `inner_type` @rudolfix

rudolfix (approved):
@sultaniman LGTM! thanks for detailed descriptions of the problem and hard work on the tests

Issue #828, opened 2023-12-15
SCD2 write disposition


adrianbr:
### Feature description

Just as we have the "merge" write disposition, we should have a SCD2 write disposition

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Multiple users have requested slowly changing dimension support

### Proposed solution

We should implement SCD2 as it's the most common, versatile and sufficient (users can add their own flags or modify it to other types via a view)
https://en.wikipedia.org/wiki/Slowly_changing_dimension#Type_2:_add_new_row

### Implementation notes / ideas (by @sh-rp)
- Compute hash per row and base updates on the hash
- have configurable names for validity columns (from-until)
- Figure out how subtables behave, also one hash per row?
- Respect soft-delete columns (maybe as a second step?)

euanjohnston-dev:
- So DBT handles this with snapshots (implements type-2 Slowly Changing over mutable source tables). 
- DBT runner seems to currently accept seed and models but not sources.

Can currently navigate by running an incremental model in dbt as a workaround but its messy: 

e.g. 

`

{{
    config({
        "materialized": "incremental",
        "unique_key": "__etl_row_unique_id",
        "on_schema_change":"append_new_columns"
    })
}}


WITH get_base_incremental as (

-- dynamically establish base columns from input.

SELECT
  t.*,
  (
  select string_agg(trim(val, '"'), '-' order by offset)
  from unnest(regexp_extract_all(to_json_string(t), r'"[^"]+":("[^"]+"|null)')) val with offset
) concatenated_columns
FROM 
(

-- exclude fields not necessary to establish increments.

SELECT * Except(LIST_COLUMNS_TO_DROP)
FROM SOURCE_TABLE
WHERE id is not null
) as t
)

-- id creation including __etl_row_unique_id required for increments.
-- new columns always appended to end of table despite explicit ordering below.

SELECT * Except(concatenated_columns),
    '{{ var("job_id") }}' AS __etl_job_id,
    concatenated_columns AS __etl_row_unique_id,
    CURRENT_DATETIME() AS __etl_last_seen_at_utc
FROM get_base_incremental`

rudolfix:
@euanjohnston-dev what do you mean by
> DBT runner seems to currently accept seed and models but not sources.
?

you can define sources in your dbt package, we just run it with `dbt` command... probably I'm missing something here and possibly we can fix your issue quickly

rudolfix:
[...]

PR #960, opened 2024-02-12
Introduce `hard_delete` and `dedup_sort` columns hint for `merge`

### LINKED ISSUE #947: Core extensions to support database replication
### Feature description

Extend the core to enable database replication. This needs to be done before #933. 

Context:

1. https://dlthub-community.slack.com/archives/C06EYJ2Q37C/p1707135498114549?thread_ts=1706550675.731069&cid=C06EYJ2Q37C
image

2. https://github.com/dlt-hub/dlt/issues/933#issuecomment-1931740029
image

### Are you a dlt user?

None

### Use case

Database replication, CDC, efficient incremental processing

### Proposed solution

_No response_

### Related issues

#933

jorritsandbrink:
### Description
~~This PR introduces a new write disposition `replicate` that can be used to propagate captured change data (INSERTs, UPDATEs, and DELETEs) from the source to the target destination.~~

- ~~requires a `primary_key` hint—raises `SchemaException` if not provided~~
- ~~requires a `cdc_config` hint—raises `SchemaException` if not provided~~
  - ~~`cdc_config` holds information on how the change data is organized, such as which column holds the operation type and which values in that column corresponds with which DML operation~~
- ~~extended `SqlMergeJob` to implement `replicate`—`SqlMergeJob` now handles both `merge` and `replicate` write dispositions~~
- ~~mechanism: first load change data to staging table, then propagate changes in final table using simple "delete-and-insert" logic (similar to how `merge` works, but here we have to filter out `delete` records before we insert from the staging table)~~
  - ~~all primary key values present in the staging table (corresponding with `insert`, `update`, and `delete` operations) are deleted from the final table~~
  - ~~records in the staging table corresponding with `insert` and `update` operations are inserted in the final table~~
- ~~also works with child tables~~

The above no longer applies—see https://github.com/dlt-hub/dlt/pull/960#issuecomment-1947529052

Potentially useful functionality not yet implemented:
- **Partial updates**. We now expect all columns of an updated record to be available (also the ones that didn't change). Most source systems (Postgres, MySQL, SQL Server, Delta tables...) seem to provide this info so perhaps this isn't so important.
- **Conflict detection and resolution**. The current logic implicitly implements a _source wins_ resolution strategy, but it would be nice to make this configurable. I.e. to add support for _target wins_ or the option to raise an error in case of a conflict. This [page](https://www.ibm.com/docs/en/idr/11.4.0?topic=console-setting-conflict-detection-resolution) provides a useful overview.
- **Soft deletes**. Add a configuration option that lets users choose between _hard deletes_ (current behavior) and _soft deletes_ (extra column is added to final table that marks a record as deleted or not).

### Related Issues

- Closes #947


jorritsandbrink:
@rudolfix can you have a look at this and see if it aligns with your ideas?

dlt/common/schema/utils.py
```diff
@@ -317,6 +318,19 @@ def validate_stored_schema(stored_schema: TStoredSchema) -> None:
             if parent_table_name not in stored_schema["tables"]:
                 raise ParentTableNotFoundException(table_name, parent_table_name)
 
+        # check for "replicate" tables that miss a primary key or "cdc_config"
```
rudolfix, line 321: this makes sense but we should move it to
1. end of normalize stage OR
2. beginning of load stage
at this moment schema may be still partial. not all columns may be present (100% after extract stage)

also we should check merge disposition.

also take a look at `_verify_schema` in `JobClientBase` looks like our place :)

dlt/common/schema/typing.py
```diff
@@ -166,6 +198,7 @@ class TTableSchema(TypedDict, total=False):
     columns: TTableSchemaColumns
     resource: Optional[str]
     table_format: Optional[TTableFormat]
+    cdc_config: Optional[TCdcConfig]
```
rudolfix, line 201: this is one way to go. but IMO a better way would be to define a column level hint.
`cdc_op` which could be integer or single char (u/d/i)

do we really need a sequence? if so we could reuse `sort` or add a new hint ie. `cdc_seq`. There are helper methods to find column(s) with hints

it looks simpler to me.

dlt/destinations/sql_jobs.py
```diff
@@ -253,20 +292,36 @@ def gen_merge_sql(
         sql: List[str] = []
         root_table = table_chain[0]
 
+        escape_identifier = sql_client.capabilities.escape_identifier
+        escape_literal = sql_client.capabilities.escape_literal
+
+        insert_condition = "1 = 1"
+        write_disposition = root_table["write_disposition"]
+        if write_disposition == "replicate":
```
rudolfix, line 300: IMO SqlMergeJob should not be aware of write disposition... Is this possible to create a base class and two subclasses for merge and replicate.

this is very hacky

rudolfix (changes requested):
we are going into a good direction. my biggest issue is with the way SqlMergeJob is hacked. IMO we can radically simplify this if:
- we want to modify existing SqlMergeJob
- we assume that updates are not partial but full

in that case the only thing we need is to define a hard delete column hint and give it a special treatment.

the distinction of insert/update is superfluous if updates are full. you delete records anyway so everything is "i" or "d"

and frankly this is what I'd do. we may even drop a separate write disposition and start interpeting hard delete flag.

partial deletes:
if we have partial deletes we'll need new write disposition and a completely separate merge job which will be based on sql MERGE statement.
if we do not need this now - let's do it later.

rudolfix:
> Partial updates. We now expect all columns of an updated record to be available (also the ones that didn't change). Most source systems (Postgres, MySQL, SQL Server, Delta tables...) seem to provide this info so perhaps this isn't so important.

see my comment above. we skip it for now. it is way more work.

> Conflict detection and resolution. The current logic implicitly implements a source wins resolution strategy, but it would be nice to make this configurable. I.e. to add support for target wins or the option to raise an error in case of a conflict. This [page](https://www.ibm.com/docs/en/idr/11.4.0?topic=console-setting-conflict-detection-resolution) provides a useful overview.

OK so here we'd need a "i" and "u" distinction. IMO part of advanced replication above

> Soft deletes. Add a configuration option that lets users choose between hard deletes (current behavior) and soft deletes (extra column is added to final table that marks a record as deleted or not).

yeah! look at this maybe: #923 and #828 definitely next thing we do

jorritsandbrink:
@rudolfix I undid a lot of the changes based on your feedback.

What's new:

- a `hard_delete` column hint that gets interpreted in `merge`
- interpretation of the `sort` column hint in `merge` so changes can be processed in order (e.g. if a record gets inserted then deleted in the same load, it should not be inserted in the destination—if it gets deleted then inserted in the same load, it should be inserted)
- `append` and `replace` ignore the `hard_delete` hint
- `merge` now has three "steps" if a `hard_delete` and/or `sort` column hint is provided: 1) delete from destination dataset 2) delete from staging dataset 3) copy staging to destination

1. Is deleting from the staging dataset a good approach or is it better to add filters in the copy-staging-to-destination step?
2. If it's a good approach, can the existing primary key deduplication also be done in that way rather than using temp tables?

dlt/destinations/sql_jobs.py
```diff
@@ -333,6 +337,44 @@ def gen_merge_sql(
...
+                        SELECT {sort_column} FROM (
+                            SELECT {sort_column}, ROW_NUMBER() OVER (partition BY {", ".join(primary_keys)} ORDER BY {sort_column} DESC) AS _rn
+                            FROM {staging_root_table_name}
+                        ) AS a
+                        WHERE a._rn > 1
+                    );
+                """)
+
+        # remove deleted records from staging tables if a hard_delete column is provided
+        if has_column_with_prop(root_table, "hard_delete"):
```
rudolfix, line 356: I think (I hope) there's a simpler way to handle `hard_deletes`. The code below does not need any modifications. It will delete all rows from destination dataset (using primary and merge keys) that are present in the staging dataset. it does not matter if hard delete flag is set or not. we must delete those rows anyway.

we only must change how we insert, from here:
```
# insert from staging to dataset, truncate staging table
        for table in table_chain:
```
the only think you need to do is to filter out rows that have deleted flag set so this is another clause in where

rudolfix: overall it should be way less code, we do not interfere with any edge cases by deleting and deduplicating the staging dataset + it looks like less row reads

dlt/destinations/sql_jobs.py
```diff
@@ -333,6 +337,44 @@ def gen_merge_sql(
...
+            hard_delete_column = escape_identifier(
+                get_columns_names_with_prop(root_table, "hard_delete")[0]
+            )
+            # first delete from root staging table
+            sql.append(f"""
+                DELETE FROM {staging_root_table_name}
+                WHERE {hard_delete_column} IS NOT DISTINCT FROM {escape_literal(True)};
```
rudolfix, line 363: ok so you assume that hard deleted column is boolean. probably makes the most sense. but then you must check the type somewhere. my take:
delete if value is IS NOT NULL OR (only in case of boolean) when true as above. maybe someone wants to have deleted flag as timestamp?

dlt/destinations/job_client_impl.py
```diff
@@ -588,3 +589,15 @@ def should_load_data_to_staging_dataset(self, table: TTableSchema) -> bool:
         ):
             return True
         return False
+
+    def _create_table_update(
+        self, table_name: str, storage_columns: TTableSchemaColumns
+    ) -> Sequence[TColumnSchema]:
+        updates = super()._create_table_update(table_name, storage_columns)
+        table = self.schema.get_table(table_name)
+        if has_column_with_prop(table, "hard_delete"):
```
rudolfix, line 598: this is really good point. but my take would be to have identical schemas in staging and destination datasets. also: what about append and replace? this data won't be dropped from parquet/json files so just dropping from schema wont't help.

I'd say let's remove it. also all the code in merge job that skips deleted columns

dlt/load/load.py
```diff
@@ -246,8 +247,14 @@ def get_completed_table_chain(
                 for job in table_jobs
             ):
                 return None
-            # if there are no jobs for the table, skip it, unless the write disposition is replace, as we need to create and clear the child tables
-            if not table_jobs and top_merged_table["write_disposition"] != "replace":
+            # if there are no jobs for the table, skip it, unless child tables need to be replaced
+            needs_replacement = False
+            if top_merged_table["write_disposition"] == "replace" or (
```
rudolfix, line 252: why it is changed?

rudolfix (changes requested):
on top of the review:
you must test a data where there are child tables. looks at other merge tests. we need a test with one and two nesting levels (can be same dataset)

dlt/load/load.py
jorritsandbrink: This is needed to propagate deletes to child tables. If we provide only a primary key and the `hard_delete` column for a nested table, such as happens on lines 584 and 599 of `test_merge_disposition.py`, the child tables wouldn't get included in the table chain, and those deletes would only be executed on the root table.

dlt/destinations/job_client_impl.py
jorritsandbrink: Removed. Staging and destination datasets now how identical schemas.

dlt/destinations/sql_jobs.py
jorritsandbrink: Good point, I implemented your suggestion.

dlt/destinations/sql_jobs.py
jorritsandbrink: I have changed the approach to extending the where clause in the insert stage, rather than deleting from the staging dataset. It didn't turn out to be less code but it makes more sense nonetheless.

dlt/destinations/sql_jobs.py
```diff
@@ -333,6 +337,44 @@ def gen_merge_sql(
                 )
             )
 
+        # remove "non-latest" records from staging table (deduplicate) if a sort column is provided
+        if len(primary_keys) > 0:
+            if has_column_with_prop(root_table, "sort"):
```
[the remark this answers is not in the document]
jorritsandbrink: Done.

jorritsandbrink:
@rudolfix I addressed your feedback and the PR is ready for another review!

dlt/destinations/sql_jobs.py
```diff
@@ -333,47 +388,59 @@ def gen_merge_sql(
...
+            if root_table["columns"][hard_delete_col]["data_type"] == "bool":
+                # only True values indicate a delete for boolean columns
+                not_deleted_cond += f" OR {escape_id(hard_delete_col)} = {escape_lit(False)}"
+
+        # get name of column with dedup_sort hint, if specified
+        dedup_sort_col = get_first_column_name_with_prop(root_table, "dedup_sort")
+
+        insert_temp_table_name: str = None
+        if len(table_chain) > 1:
+            if len(primary_keys) > 0 or (len(primary_keys) == 0 and hard_delete_col is not None):
```
rudolfix, line 406: `if len(primary_keys) > 0 or hard_delete_col is not None:` should be sufficient.

dlt/destinations/sql_jobs.py
```diff
@@ -200,19 +205,63 @@ def gen_delete_temp_table_sql(
...
+        Expects column names provided in arguments to be escaped identifiers.
+        """
+        if sort_column is None:
+            order_by = "(SELECT NULL)"
+        else:
+            order_by = f"{sort_column} DESC"
+        if condition is None:
+            condition = "1 = 1"
+        col_str = ", ".join(columns)
+        inner_col_str = copy(col_str)
```
rudolfix, line 229: do not need to copy! strings are immutable

dlt/destinations/sql_jobs.py
```diff
@@ -333,47 +388,59 @@ def gen_merge_sql(
...
+                # only True values indicate a delete for boolean columns
+                not_deleted_cond += f" OR {escape_id(hard_delete_col)} = {escape_lit(False)}"
+
+        # get name of column with dedup_sort hint, if specified
+        dedup_sort_col = get_first_column_name_with_prop(root_table, "dedup_sort")
+
+        insert_temp_table_name: str = None
+        if len(table_chain) > 1:
+            if len(primary_keys) > 0 or (len(primary_keys) == 0 and hard_delete_col is not None):
+                condition_colummns = [hard_delete_col] if not_deleted_cond is not None else None
```
rudolfix, line 407: `condition_colummns` typo

dlt/load/load.py
rudolfix: I still do not get it. We have jobs for this table because in both those lines we declare some data. The exception for replace is only for the case that there is no data at all. does not happen here. IMO you should try to remove it and find the problem elsewhere or ping me on slack to discuss it

dlt/destinations/sql_jobs.py
```diff
@@ -200,19 +205,63 @@ def gen_delete_temp_table_sql(
...
+        table_name: str,
+        primary_keys: Sequence[str],
+        columns: Sequence[str],
+        sort_column: str = None,
+        condition: str = None,
+        condition_columns: Sequence[str] = None,
+    ) -> str:
+        """
+        Returns SELECT FROM statement where the FROM clause represents a deduplicated version of the `table_name` table.
+        Expects column names provided in arguments to be escaped identifiers.
```
rudolfix, line 220: please explain what is the condition and condition columns and how they are related, you use this function in two ways:
 to get temp table and to gen full select. this is also worthy of mention.

rudolfix (changes requested):
OK! we are almost there. our sql jobs is almost unreadable now (if it ever was)

we are missing a test where you have a merge key on non-unique column and you have a hard delete flag (you should be able to delete whole day of data with just one flag)

also question: does hard deleted flag make sense if there's no primary key? if the answer is no and primary key is required we can simplify code even more

also maybe a test case when we have a dedup sort and two rows one with deleted flag and one without (could run on duckdb only to make it faster)

dlt/destinations/sql_jobs.py
```diff
@@ -253,28 +302,34 @@ def gen_merge_sql(
         sql: List[str] = []
         root_table = table_chain[0]
 
+        escape_id = sql_client.capabilities.escape_identifier
+        escape_lit = sql_client.capabilities.escape_literal
+        if escape_id is None:
```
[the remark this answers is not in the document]
jorritsandbrink: `sql_client.capabilities.escape_literal` is None for `snowflake`. `sql_client.capabilities.escape_identifier` always has a value (at least with the current set of destinations), but I included the `if escape_id is None:`  for consistency.

dlt/destinations/sql_jobs.py
```diff
@@ -200,19 +205,63 @@ def gen_delete_temp_table_sql(
...
+        condition_columns: Sequence[str] = None,
+    ) -> str:
+        """
+        Returns SELECT FROM statement where the FROM clause represents a deduplicated version of the `table_name` table.
+        Expects column names provided in arguments to be escaped identifiers.
+        """
+        if sort_column is None:
+            order_by = "(SELECT NULL)"
+        else:
+            order_by = f"{sort_column} DESC"
```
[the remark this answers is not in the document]
jorritsandbrink: Higher values typically indicate more recent (think timestamps, or the LSN in a Postgres WAL). So if we sort in descending order, we get the most recent value, which makes sense for most typical use cases.

I could also change the `dedup_sort` column hint from boolean to string and have it accept `"asc"` or `"desc"` values to make it configurable for the user. What do you think?

jorritsandbrink:
@rudolfix I addressed all your points. I added the test cases you mentioned. Only remaining point is the one about child table skipping we are discussing on Slack.

> also question: does hard deleted flag make sense if there's no primary key? if the answer is no and primary key is required we can simplify code even more

I think it makes as much sense as doing inserts/updates without a primary key. I would say deleting multiple records sharing the same key is a valid use case we should support.

rudolfix (changes requested):
looks like asc/desc on update is left.

I've found and fixed a lot of heresy in the code. you can see my two commits. there were edge cases that tables were not created even if they should or vice versa.

now normalizer is marking tables which seen data so loader knows more which tables to create.

rudolfix (approved):
bugs fixed & some tests added. LGTM!

Issue #934, opened 2024-02-06
Simplify Parallelization for generator resources


sh-rp:
### Feature description
We want to provide a simple interface that enables our users to annotate a resource create from a generator in some way which will make dlt execute each iterator of the executor on a thread from the threadpool in pipe.py. An example implementation of this in a commented out test can be found in `def test_async_decorator_experiment` on the devel branch.

### Reasoning
Even with this feature, a generator can of course only be evaluated one item after another, so with a single generator resource this approach will not be faster. But with multiple parallelized decorators active at the same time, dlt can evaluate them in parallel across many workers with the user only needing to supply one flag or annotation.

### Implementation Notes
- We need to decide on the interface, either a decorator as shown in the example, or maybe just a parameter on the resource decorator called "parallelize".
- We need to update the docs, specially the "performance" page.
- We need to make sure all generators are closed properly in the event of an exception of an early exit of the extract phase. See the according tests for regular and async generator resources.
- We could consider adding a global flag to make all resources behave this way automatically if they are a regular generator.
- The example demonstrates how to wrap an existing generator into a threadsafe generator that yields callables. Callables will be evaluated in the futures pool automatically by dlt extract. This is probably the best way to implement this, but maybe there is a better way.

PR #965, opened 2024-02-13
Sthor/parallelize decorator

steinitzu:
### Description
Implements #934 

Draft for now, may need to test more.

Added a `parallelized` argument to resource. For me that seems like a good interface.  
Under the hood it's basically the same decorator from the test implementation (maybe it makes sense to use it directly in some cases?) or we could integrate this more deeply in resource.

There was a lot of overhead coming from all the polling in `PipeIterator` when constantly yielding `None`, so I tried to implement more intelligent waiting for futures. Basically just block until some future completes when the pool is full instead of blindly polling and looping over the resources and that seems to make a big difference.   
+ pretty big refactor of `PipeIterator` -> the futures worker pool is a separate class   

Benchmark before and after optimizing: https://gist.github.com/steinitzu/671b47cdb7cbf61b2fab46cf1faefd86  
The original iterator had a lot of overhead, especially when number of resources is <= number of workers.

### DIFF
--- a/dlt/extract/concurrency.py
+++ b/dlt/extract/concurrency.py
@@ -1,0 +1,219 @@
+import asyncio
+from uuid import uuid4
+from asyncio import Future
+from concurrent.futures import (
+    ThreadPoolExecutor,
+    as_completed,
+    wait as wait_for_futures,
+    FIRST_COMPLETED,
+)
+from threading import Thread
+from typing import List, Awaitable, Callable, Any, Dict, Set, Optional
+
+from dlt.common.exceptions import PipelineException
+from dlt.common.configuration.container import Container
+from dlt.common.runtime.signals import sleep
+from dlt.extract.typing import DataItemWithMeta, TItemFuture
+from dlt.extract.items import ResolvablePipeItem, FuturePipeItem
+
+from dlt.extract.exceptions import (
+    DltSourceException,
+    ExtractorException,
+    PipeException,
+    ResourceExtractionError,
+)
+
+
+class WorkerPool:
+    """Worker pool for pipe items that can be resolved asynchronously.
+
+    Items can be either asyncio coroutines or regular callables which will be executed in a thread pool.
+    """
+
+    def __init__(
+        self, workers: int = 5, poll_interval: float = 0.01, max_parallel_items: int = 20
+    ) -> None:
+        self.futures: Dict[TItemFuture, FuturePipeItem] = {}
+        self._thread_pool: ThreadPoolExecutor = None
+        self._async_pool: asyncio.AbstractEventLoop = None
+        self._async_pool_thread: Thread = None
+        self.workers = workers
+        self.poll_interval = poll_interval
+        self.max_parallel_items = max_parallel_items
+        self.used_slots: int = 0
+
+    def __len__(self) -> int:
+        return len(self.futures)
+
+    @property
+    def free_slots(self) -> int:
+        # Done futures don't count as slots, so we can still add futures
+        return self.max_parallel_items - self.used_slots
+
+    @property
+    def empty(self) -> bool:
+        return len(self.futures) == 0
+
+    def _ensure_thread_pool(self) -> ThreadPoolExecutor:
+        # lazily start or return thread pool
+        if self._thread_pool:
+            return self._thread_pool
+
+        self._thread_pool = ThreadPoolExecutor(
+            self.workers, thread_name_prefix=Container.thread_pool_prefix() + "threads"
+        )
+        return self._thread_pool
+
+    def _ensure_async_pool(self) -> asyncio.AbstractEventLoop:
+        # lazily create async pool is separate thread
+        if self._async_pool:
+            return self._async_pool
+
+        def start_background_loop(loop: asyncio.AbstractEventLoop) -> None:
+            asyncio.set_event_loop(loop)
+            loop.run_forever()
+
+        self._async_pool = asyncio.new_event_loop()
+        self._async_pool_thread = Thread(
+            target=start_background_loop,
+            args=(self._async_pool,),
+            daemon=True,
+            name=Container.thread_pool_prefix() + "futures",
+        )
+        self._async_pool_thread.start()
+
+        # start or return async pool
+        return self._async_pool
+
+    def _vacate_slot(self, _: TItemFuture) -> None:
+        # Used as callback to free up slot when future is done
+        self.used_slots -= 1
+
+    def submit(self, pipe_item: ResolvablePipeItem) -> Optional[TItemFuture]:
+        """Submit an item to the pool.
+
+        Args:
+            pipe_item: The item to submit.
+
+        Returns:
+            The future if the item was successfully submitted, otherwise None.
+        """
+        if self.free_slots == 0:
+            return None
+
+        if self.free_slots < 0:  # TODO: Sanity check during dev, should never happen
+            raise RuntimeError("Used more slots than max_parallel_items")
+
+        # submit to thread pool or async pool
+        item = pipe_item.item
+        if isinstance(item, Awaitable):
+            future = asyncio.run_coroutine_threadsafe(item, self._ensure_async_pool())
+        elif callable(item):
+            future = self._ensure_thread_pool().submit(item)
+        else:
+            raise ValueError(f"Unsupported item type: {type(item)}")
+
+        # Future is not removed from self.futures until it's been consumed by the
+        # pipe iterator. But we always want to vacate a slot so new jobs can be submitted
+        future.add_done_callback(self._vacate_slot)
+        self.used_slots += 1
+
+        self.futures[future] = FuturePipeItem(
+            future, pipe_item.step, pipe_item.pipe, pipe_item.meta
+        )
+        return future
+
+    def poll(self) -> None:
+        sleep(self.poll_interval)
+
+    def _resolve_future(self, future: TItemFuture) -> Optional[ResolvablePipeItem]:
+        future, step, pipe, meta = self.futures.pop(future)
+
+        if ex := future.exception():
+            if isinstance(ex, StopAsyncIteration):
+                return None
+            # Raise if any future fails
+            if isinstance(
+                ex, (PipelineException, ExtractorException, DltSourceException, PipeException)
+            ):
+                raise ex
+            raise ResourceExtractionError(pipe.name, future, str(ex), "future") from ex
+
+        item = future.result()
+
+        if item is None:
+            return None
+        elif isinstance(item, DataItemWithMeta):
+            return ResolvablePipeItem(item.data, step, pipe, item.meta)
+        else:
+            return ResolvablePipeItem(item, step, pipe, meta)
+
+    def resolve_next_future(self) -> Optional[ResolvablePipeItem]:
+        """Wait for the next future to be done. Returns None if no futures done."""
+        if not self.futures:
+            return None
+
+        if (item := self.resolve_next_future_no_wait()) is not None:
+            return item
+        for future in as_completed(self.futures):
+            if future.cancelled():
+                # Get the next not-cancelled future
+                continue
+
+            return self._resolve_future(future)
+
+        return None
+
+    def resolve_next_future_no_wait(self) -> Optional[ResolvablePipeItem]:
+        """Resolve the first done future in the pool.
+        This does not block and returns None if no future is done.
+        """
+        # Get next done future
+        future = next((fut for fut in self.futures if fut.done() and not fut.cancelled()), None)
+        if not future:
+            return None
+
+        return self._resolve_future(future)
+
+    def wait_for_free_slot(self) -> None:
+        """Wait until any future in the pool is completed to ensure there's a free slot."""
+        if self.free_slots >= 1:
+            self.poll()  # Just sleep a bit to give other threads a chance
+            return
+
+        for future in as_completed(self.futures):
+            if future.cancelled():
+                # Get the next not-cancelled future
+                continue
+            if self.free_slots == 0:
+                # Future was already completed so slot was not freed
+                continue
+            return  # Return when first future completes
+
+    def close(self) -> None:
+        # Cancel all futures
+        for f in self.futures:
+            if not f.done():
+                f.cancel()
+
+        def stop_background_loop(loop: asyncio.AbstractEventLoop) -> None:
+            loop.stop()
+
+        if self._async_pool:
+            # wait for all async generators to be closed
+            future = asyncio.run_coroutine_threadsafe(
+                self._async_pool.shutdown_asyncgens(), self._ensure_async_pool()
+            )
+
+            wait_for_futures([future])
+            self._async_pool.call_soon_threadsafe(stop_background_loop, self._async_pool)
+
+            self._async_pool_thread.join()
+            self._async_pool = None
+            self._async_pool_thread = None
+
+        if self._thread_pool:
+            self._thread_pool.shutdown(wait=True)
+            self._thread_pool = None
+
+        self.futures.clear()
--- a/dlt/extract/decorators.py
+++ b/dlt/extract/decorators.py
@@ -456,7 +456,7 @@ def resource(
             raise DynamicNameNotStandaloneResource(get_callable_name(f))
 
         if parallelized:
-            f = parallel(f)
+            f = parallelize(f)
 
         # resource_section = name if name and not callable(name) else get_callable_name(f)
         resource_name = name if name and not callable(name) else get_callable_name(f)
@@ -758,9 +758,9 @@ def defer(
     return _wrap
 
 
-def parallel(f: Callable[TResourceFunParams, Any]) -> Callable[TResourceFunParams, Any]:
+def parallelize(f: Callable[TResourceFunParams, Any]) -> Callable[TResourceFunParams, Any]:
     @wraps(f)
-    def _wrap(*args, **kwargs):
+    def _wrap(*args: Any, **kwargs: Any) -> Any:  # TODO: Type correctly
         gen = f(*args, **kwargs)
         if inspect.isfunction(gen):
             gen = gen()
@@ -770,7 +770,7 @@ def parallel(f: Callable[TResourceFunParams, Any]) -> Callable[TResourceFunParam
         exhausted = False
         busy = False
 
-        def _parallel_gen():
+        def _parallel_gen() -> Any:  # TODO: Type correctly
             try:
                 return next(gen)
             except StopIteration:
--- a/dlt/extract/decorators.py
+++ b/dlt/extract/decorators.py
@@ -277,6 +277,7 @@ def resource(
     table_format: TTableHintTemplate[TTableFormat] = None,
     selected: bool = True,
     spec: Type[BaseConfiguration] = None,
+    parallelized: bool = False,
 ) -> DltResource:
     ...
 
@@ -295,6 +296,7 @@ def resource(
     table_format: TTableHintTemplate[TTableFormat] = None,
     selected: bool = True,
     spec: Type[BaseConfiguration] = None,
+    parallelized: bool = False,
 ) -> Callable[[Callable[TResourceFunParams, Any]], DltResource]:
     ...
 
@@ -314,6 +316,7 @@ def resource(
     selected: bool = True,
     spec: Type[BaseConfiguration] = None,
     standalone: Literal[True] = True,
+    parallelized: bool = False,
 ) -> Callable[[Callable[TResourceFunParams, Any]], Callable[TResourceFunParams, DltResource]]:
     ...
 
@@ -332,6 +335,7 @@ def resource(
     table_format: TTableHintTemplate[TTableFormat] = None,
     selected: bool = True,
     spec: Type[BaseConfiguration] = None,
+    parallelized: bool = False,
 ) -> DltResource:
     ...
 
@@ -351,7 +355,7 @@ def resource(
     spec: Type[BaseConfiguration] = None,
     standalone: bool = False,
     data_from: TUnboundDltResource = None,
-    parallelize: bool = False,
+    parallelized: bool = False,
 ) -> Any:
     """When used as a decorator, transforms any generator (yielding) function into a `dlt resource`. When used as a function, it transforms data in `data` argument into a `dlt resource`.
 
@@ -451,7 +455,7 @@ def resource(
         if not standalone and callable(name):
             raise DynamicNameNotStandaloneResource(get_callable_name(f))
 
-        if parallelize:
+        if parallelized:
             f = parallel(f)
 
         # resource_section = name if name and not callable(name) else get_callable_name(f)
@@ -785,5 +789,6 @@ def parallel(f: Callable[TResourceFunParams, Any]) -> Callable[TResourceFunParam
                 yield _parallel_gen
             except GeneratorExit:
                 gen.close()
+                raise
 
     return _wrap
--- a/dlt/extract/decorators.py
+++ b/dlt/extract/decorators.py
@@ -77,7 +77,8 @@ class SourceSchemaInjectableContext(ContainerInjectableContext):
 
     if TYPE_CHECKING:
 
-        def __init__(self, schema: Schema = None) -> None: ...
+        def __init__(self, schema: Schema = None) -> None:
+            ...
 
 
 @configspec
@@ -90,7 +91,8 @@ class SourceInjectableContext(ContainerInjectableContext):
 
     if TYPE_CHECKING:
 
-        def __init__(self, source: DltSource = None) -> None: ...
+        def __init__(self, source: DltSource = None) -> None:
+            ...
 
 
 TSourceFunParams = ParamSpec("TSourceFunParams")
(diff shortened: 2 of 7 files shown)


dlt/extract/concurrency.py
```diff
@@ -0,0 +1,219 @@
...
+        Args:
+            pipe_item: The item to submit.
+
+        Returns:
+            The future if the item was successfully submitted, otherwise None.
+        """
+        if self.free_slots == 0:
+            return None
+
+        if self.free_slots < 0:  # TODO: Sanity check during dev, should never happen
```
rudolfix, line 104: user assert for that!

dlt/extract/concurrency.py
```diff
@@ -0,0 +1,219 @@
...
+        if isinstance(item, Awaitable):
+            future = asyncio.run_coroutine_threadsafe(item, self._ensure_async_pool())
+        elif callable(item):
+            future = self._ensure_thread_pool().submit(item)
+        else:
+            raise ValueError(f"Unsupported item type: {type(item)}")
+
+        # Future is not removed from self.futures until it's been consumed by the
+        # pipe iterator. But we always want to vacate a slot so new jobs can be submitted
+        future.add_done_callback(self._vacate_slot)
```
rudolfix, line 118: why you call it then? it just decreases the counter

dlt/extract/pipe.py
```diff
@@ -822,34 +682,58 @@ def _get_source_item(self) -> ResolvablePipeItem:
             # if too many new sources is added we switch to fifo not to exhaust them
             if (
                 self._next_item_mode == "fifo"
-                or (sources_count - self._initial_sources_count) >= self.max_parallel_items
+                or sources_count - self._initial_sources_count
```
rudolfix, line 685: I know that it works but lack of parens here is scary :)

dlt/extract/pipe.py
```diff
@@ -822,34 +682,58 @@ def _get_source_item(self) -> ResolvablePipeItem:
...
-                            return ResolvablePipeItem(item.data, step, pipe, item.meta)
+                        if isinstance(pipe_item, DataItemWithMeta):
+                            pipe_item = ResolvablePipeItem(
+                                pipe_item.data, step, pipe, pipe_item.meta
+                            )
                         else:
-                            return ResolvablePipeItem(item, step, pipe, meta)
+                            pipe_item = ResolvablePipeItem(pipe_item, step, pipe, meta)
+
+                    if isinstance(pipe_item.item, Awaitable) or callable(pipe_item.item):
```
rudolfix, line 714: this duplication looks really suspicious. AFAIK we were returning Awaitable/Callable item to the __next__ function and resolved the item there. I do not see a reason for this code to be there. if this is a result of the code refactor then IMO something is still not right there

dlt/extract/decorators.py
```diff
@@ -736,3 +756,39 @@ def _curry() -> TBoundItems:
         return _curry
 
     return _wrap
+
+
+def parallelize(f: Callable[TResourceFunParams, Any]) -> Callable[TResourceFunParams, Any]:
+    @wraps(f)
+    def _wrap(*args: Any, **kwargs: Any) -> Any:  # TODO: Type correctly
+        gen = f(*args, **kwargs)
+        if inspect.isfunction(gen):
```
rudolfix, line 765: if `gen` is a function, we'll call it when pipe iterator is constructed. do not start evaluating gen here! context:
- pipes are cloned before evaluation. so if gen is generator function you'll be able to evaluate it multiple times
- maybe someone will pass the source to a different thread for evaluation. we support parallel pipelines now (and this decorator is made to support that)
instead check if this is generator function and if not raise a nice exception. please look into exceptions.py and use proper base class (where you can pass function name and a few other so user knows what is going on)

dlt/extract/decorators.py
```diff
@@ -736,3 +756,39 @@ def _curry() -> TBoundItems:
...
+            gen = gen()
+        if inspect.isasyncgen(gen):
```
rudolfix, line 767: see above.

rudolfix (changes requested):
I'm OK with the code refactor. Code duplication you mentioned worries me. We didnt have it before!

Could you also add a `parallelize` method to the `DltResource` that will wrap existing gen so regular resources (ie. from verified sources) can be converted to parallel ones?

also you must test parallel transformers!

rudolfix:
one more comment to refactor: please split `Pipe` from `PipeIterator`. Also I think you can move `items.py` to `typing.py` but that's up to you

dlt/extract/pipe.py
steinitzu: It's an optimization, but I'm not totally happy with how this looks either. Want to give it a little more love.  
The basic idea behind this is that we start multiple futures on each `__next__`  iteration, instead of just adding one future and doing the whole dance "poll -> iterate all sources -> repeat" each time. It seems to make a big difference.

steinitzu: My benchmark with and without this block:

  ```
  Done in 10.724883079528809 seconds
  Extracted 5000 items total from 5 resources

  Done in 20.511144161224365 seconds
  Extracted 5000 items total from 5 resources
  ```

dlt/extract/pipe.py
steinitzu: I like the parens too :) I didn't meant to edit this line

dlt/extract/concurrency.py
steinitzu: It's the same idea (I think) as the way it checked for free slots before:
```python
                if len(self._futures) < self.max_parallel_items or self._next_future() >= 0:
```
But the difference now there can be more than `max_parallel_items + 1` futures in the pool if some of them are done, so this keeps track of how many slots are being blocked by pending futures only.  
The callback fires soon as the future is done which can be way before pipe iterator actually consumes the result and this seems faster than counting slots every time.

sh-rp:
I did not review the PR, but want to point out that returning None from a resource to skip to the next in round robin mode should still work for backwards compatibility at least alex is using this.

rudolfix:
> I did not review the PR, but want to point out that returning None from a resource to skip to the next in round robin mode should still work for backwards compatibility at least alex is using this.

💯  we have a  test that makes sure that round robin really works `test_rotation_on_none` and all other tests must pass

dlt/extract/resource.py
```diff
@@ -331,7 +331,7 @@ def _gen_wrap(gen: TPipeStep) -> TPipeStep:
                         # async gen yields awaitable so we must count one awaitable more
                         # so the previous one is evaluated and yielded.
                         # new awaitable will be cancelled
-                        if count == max_items + int(is_async_gen):
```
rudolfix, line 334: this is cool. but why can we get rid of it now?
```

```

dlt/extract/resource.py
```diff
@@ -343,6 +343,19 @@ def _gen_wrap(gen: TPipeStep) -> TPipeStep:
             self._pipe.replace_gen(_gen_wrap(self._pipe.gen))
         return self
 
+    def parallelize(self) -> "DltResource":
+        """Wraps the resource to execute each item in a threadpool to allow multiple resources to extract in parallel."""
+
+        if (
+            not inspect.isgenerator(self._pipe.gen)
```
rudolfix, line 350: does that really work? IMO you cannot parallelize executing generator. you need to write a test for that. start generator in the main thread and then use parallelize and evaluate it. it must fail. we should detect and raise a proper exception.

dlt/extract/source.py
```diff
@@ -333,6 +334,11 @@ def add_limit(self, max_items: int) -> "DltSource":  # noqa: A003
             resource.add_limit(max_items)
         return self
 
+    def parallelize(self) -> "DltSource":
+        for resource in self.resources.selected.values():
```
rudolfix, line 338: cool but IMO we should skip resources that cannot be parallelized. so those that will raise exception from resource.parallelize. also docstrings

dlt/extract/utils.py
```diff
@@ -171,6 +178,57 @@ async def run() -> TDataItems:
...
+    """Wraps a generator for parallel extraction"""
+
+    def _wrapper(*args: Any, **kwargs: Any) -> Generator[TDataItems, None, None]:
+        is_generator = True
+        gen: Union[Generator[TDataItems, Optional[Any], Optional[Any]], AnyFun]
+        if callable(f):
+            if inspect.isgeneratorfunction(f):
+                gen = f(*args, **kwargs)
+            else:
+                is_generator = False
```
rudolfix, line 193: why do we allow regular functions here? this will also execute them in a pool right?

tests/extract/test_decorators.py
```diff
@@ -853,3 +854,51 @@ def __call__(self, more: int = 1):
...
+def test_parallelized_resource_decorator() -> None:
+    """Test paralellized resources are wrapped correctly.
+    Note: tests for parallel execution are in test_resource_evaluation
+    """
+
+    def some_gen():
+        yield from [1, 2, 3]
+
+    # Create resource with decorated function
+    resource = dlt.resource(some_gen, parallelized=True)
```
rudolfix, line 868: also try `resource = dlt.resource(some_gen(), parallelized=True)` which will pass generator, not generator function.

we need separate test that will actually run such resource. you'll see that it will fail

tests/pipeline/test_resources_evaluation.py
```diff
@@ -212,99 +215,174 @@ async def async_resource1():
...
+            assert {r[0] for r in rows} == {"a", "b", "c"}
+
+        with c.execute_query("SELECT * FROM resource2") as cur:
+            rows = list(cur.fetchall())
+            assert len(rows) == 3
+            assert {r[0] for r in rows} == {"e", "f", "g"}
+
+    if parallelized:
+        assert len(threads) > 1
+        assert execution_order == ["one", "two", "one", "two", "one", "two"]
```
rudolfix, line 264: I hope it sleeps long enough so test is really deteministic :)

tests/pipeline/test_resources_evaluation.py
```diff
@@ -212,99 +215,174 @@ async def async_resource1():
...
-
-#     if parallelized:
-#         assert len(threads) > 1
-#         assert execution_order == ["one", "two", "one", "two", "one", "two"]
-#     else:
-#         assert execution_order == ["one", "one", "one", "two", "two", "two"]
-#         assert len(threads) == 1
+@pytest.mark.parametrize("parallelized", [True, False])
+def test_parallelized_resource(parallelized: bool) -> None:
+    os.environ["EXTRACT__NEXT_ITEM_MODE"] = "fifo"
```
rudolfix, line 220: what for round robin? test it for both. I want to see what is the difference!

rudolfix (changes requested):
besides all the comments:
- modification to `add_limit` worries my. why it works right now? we are still yielding awaitables from the resource itslef no?
- we need a test for add limit for parallelize (both resource and transformer)
- also we need docs, but let's make this work first

dlt/extract/concurrency.py
```diff
@@ -0,0 +1,239 @@
...
+            return ResolvablePipeItem(item.data, step, pipe, item.meta)
+        else:
+            return ResolvablePipeItem(item, step, pipe, meta)
+
+    def resolve_next_future(self) -> Optional[ResolvablePipeItem]:
+        """Wait for the next future to be done. Returns None if no futures done."""
+        if not self.futures:
+            return None
+
+        if (item := self.resolve_next_future_no_wait()) is not None:
```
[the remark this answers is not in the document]
steinitzu: It's to guarantee ordered results from each resource. I.e. if there are 2 or more futures from the same pipe already done when this is called, we get the results in FIFO order, since the `self._futures` dict is in insertion order but `as_completed` yields in undefined order.

dlt/extract/concurrency.py
```diff
@@ -0,0 +1,239 @@
...
+        return self._resolve_future(future)
+
+    def wait_for_free_slot(self) -> None:
+        """Wait until any future in the pool is completed to ensure there's a free slot."""
+        if self.free_slots >= 1:
+            self.poll()  # Just sleep a bit to give other threads a chance
+            return
+
+        for future in as_completed(self.futures):
+            if future.cancelled():
```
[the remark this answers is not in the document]
steinitzu: Yeah only when closing the pool I think. The old code did this same thing. Probably prevents some random errors when you kill the program early.

dlt/extract/pipe_iterator.py
```diff
@@ -0,0 +1,395 @@
...
+                                pipe_item.data, step, pipe, pipe_item.meta
+                            )
+                        else:
+                            pipe_item = ResolvablePipeItem(pipe_item, step, pipe, meta)
+
+                    if isinstance(pipe_item.item, Awaitable) or callable(pipe_item.item):
+                        # Send callables to the worker pool right away, collect futures from multiple sources in one iteration
+                        self._worker_pool.submit(pipe_item, block=True)
+                        pipe_item = None
+                        if len(self._worker_pool) >= sources_count:
```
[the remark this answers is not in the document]
steinitzu: Yeah `sources_count` seemed like a good number to stop collecting so we've roughly iterated all sources ( but could be tuneable).   
We're just going round-robin while adding futures to the pool (same as when source yields `None`), but with this break condition so we don't spawn 5,000,000 futures before yielding anything.

dlt/extract/pipe_iterator.py
```diff
@@ -0,0 +1,395 @@
...
+                    if not isinstance(pipe_item, ResolvablePipeItem):
+                        # keep the item assigned step and pipe when creating resolvable item
+                        if isinstance(pipe_item, DataItemWithMeta):
+                            pipe_item = ResolvablePipeItem(
+                                pipe_item.data, step, pipe, pipe_item.meta
+                            )
+                        else:
+                            pipe_item = ResolvablePipeItem(pipe_item, step, pipe, meta)
+
+                    if isinstance(pipe_item.item, Awaitable) or callable(pipe_item.item):
```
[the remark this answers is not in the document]
steinitzu: 1. Hmm, think maybe the `resolve_next_future_no_wait()` part can be removed, and we could do blocking `resolve_next_future()` there instead -> if pool is empty we go back to `get_source_item()`. That would reduce some polling, and there should always be something in the pool to wait for after the first iteration.  

2. This block `if isinstance(item, Awaitable) or callable(item): ...` in `__next__` is duplicate code, but it's only triggered when followup pipe steps return callables (`next_item`). `_get_source_item` never returns callable/awaitable, it sends them to the future pool itself and returns `None`

tests/pipeline/test_resources_evaluation.py
steinitzu: Yeah can parametrize it. Should be not much difference. It's going round-robin implicitly but resetting at some points depending on timings.

rudolfix:
@steinitzu one more ask - when this is implemented could you check if our `sql_database` can be executed in parallel? looks like `engine` is thread safe and we open connection when already (possibly) in a thread

steinitzu:
> @steinitzu one more ask - when this is implemented could you check if our `sql_database` can be executed in parallel? looks like `engine` is thread safe and we open connection when already (possibly) in a thread

Good call, wonder if that will work with streaming results. The conn pool is thread safe, but we're streaming chunks on one connection to load a table.

dlt/extract/concurrency.py
```diff
@@ -0,0 +1,239 @@
...
+        # Get next done future
+        future = next((fut for fut in self.futures if fut.done() and not fut.cancelled()), None)
+        if not future:
+            return None
+
+        return self._resolve_future(future)
+
+    def wait_for_free_slot(self) -> None:
+        """Wait until any future in the pool is completed to ensure there's a free slot."""
+        if self.free_slots >= 1:
```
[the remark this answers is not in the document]
steinitzu: Yeah removed this, seems not useful to have it

dlt/extract/resource.py
steinitzu: This works fine, or am I missing something? See: https://github.com/dlt-hub/dlt/blob/sthor%2Fparallelize-decorator/tests/pipeline/test_resources_evaluation.py#L391-L414

It's still executing the generator in one thread at a time, the `while busy: yield None` loop acts as a lock

dlt/extract/utils.py
steinitzu: Transformes that are not generators: 
https://github.com/dlt-hub/dlt/blob/sthor%2Fparallelize-decorator/tests/pipeline/test_resources_evaluation.py#L365-L371  
So they execute in pool. Any reason not to support it?

tests/pipeline/test_resources_evaluation.py
steinitzu: Probably beter to not count on it. Don't think we need to check exact order.

dlt/extract/pipe_iterator.py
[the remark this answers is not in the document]
steinitzu: Cleaned this part up a bit: https://github.com/dlt-hub/dlt/blob/sthor%2Fparallelize-decorator/dlt/extract/pipe_iterator.py#L151-L164  
Less polling going on.

dlt/extract/pipe_iterator.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def __next__(self) -> PipeItem:
+        pipe_item: Union[ResolvablePipeItem, SourcePipeItem] = None
+        # __next__ should call itself to remove the `while` loop and continue clauses but that may lead to stack overflows: there's no tail recursion opt in python
+        # https://stackoverflow.com/questions/13591970/does-python-optimize-tail-recursion (see Y combinator on how it could be emulated)
+        while True:
+            # do we need new item?
+            if pipe_item is None:
+                # if none then take element from the newest source
+                pipe_item = self._get_source_item()
```
rudolfix, line 153: W8,
1. we were taking items from futures pool without blocking before going for next item in the sources.
2. if None we tried to get items from the source. 
3. if None we checked if iterator is maybe exhausted and possibly exit
4. if not we were sleeping a little bit and go to 1

this pattern must remain. we cannot have any blocking operations anywhere because we'll start starving our pools.

we added additional complications by adding generators that return None. we prevent sources from starving futures by exiting when all sources returned None and taking items from future pool

there is OFC a problem because we sleep when there's no data anywhere. and this sleep is unconditional. we can improve it a little by waiting on a futures pool instead of sleeping (if any elements in it) and make our parallel and async generators to trigger a semaphore on which we sleep (but must be in thread storage let's not do it)

you can decrease CPU usage by moving done to special done dicts and improving `_next_done_future` method

rudolfix (changes requested):
I have a feeling that this PR got more chaotic. but now I think I understand various edge cases.  Please read my comments. There's not so much fixing.

btw. I realized that generators can be read from multiple threads not sure where I got the idea that it is not the case :)

tests/pipeline/test_resources_evaluation.py
```diff
@@ -212,99 +215,200 @@ async def async_resource1():
...
+    def some_source():
+        return [
+            # Resources created from generators directly (not generator functions) can be parallelized
+            dlt.resource(pos_data(), parallelized=True, name="pos_data"),
+            dlt.resource(neg_data(), parallelized=True, name="neg_data"),
+        ]
+
+    result = list(some_source())
+
+    assert set(result) == {1, 2, 3, 4, 5, -1, -2, -3, -4, -5}
```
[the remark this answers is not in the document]
steinitzu: I'm not checking the order here since it's not that guaranteed, just that all items are here. But I updated all the tests to check thread idents ( >1 thread and not main thread)  
Would be nice to check execution order is roughly interleaved but tricky without making flaky tests

dlt/extract/pipe_iterator.py
```diff
@@ -0,0 +1,388 @@
...
+            # always reset to end of list for fifo mode, also take into account that new sources can be added
+            # if too many new sources is added we switch to fifo not to exhaust them
+            if self._next_item_mode == "fifo" or (
+                sources_count - self._initial_sources_count >= self._futures_pool.max_parallel_items
+            ):
+                self._current_source_index = sources_count - 1
+            else:
+                self._current_source_index = (self._current_source_index - 1) % sources_count
+            while True:
+                # get next item from the current source
```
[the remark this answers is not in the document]
steinitzu: Figured the future submitting would return eventually. But yes also need to handle the returning `None` case. I'm returning `None` from this now and `__next__` does polling

dlt/extract/concurrency.py
```diff
@@ -0,0 +1,242 @@
...
+        assert self.free_slots >= 0, "Worker pool has negative free slots, this should never happen"
+
+        if self.free_slots == 0:
+            if block:
+                # Wait until some future is completed to ensure there's a free slot
+                # Note: This is probably not thread safe. If ever multiple threads will be submitting
+                # jobs to the pool, we ned to change this whole method to be inside a `threading.Lock`
+                self._wait_for_free_slot()
+            else:
+                return None
```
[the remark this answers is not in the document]
steinitzu: Done this, submit always blocks

dlt/extract/pipe_iterator.py
```diff
@@ -0,0 +1,388 @@
...
+        # https://stackoverflow.com/questions/13591970/does-python-optimize-tail-recursion (see Y combinator on how it could be emulated)
+        while True:
+            # do we need new item?
+            if pipe_item is None:
+                # if none then take element from the newest source
+                pipe_item = self._get_source_item()
+
+                if pipe_item is None:
+                    # Block until a future is resolved
+                    pipe_item = self._futures_pool.resolve_next_future()
```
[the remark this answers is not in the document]
steinitzu: This makes sense and I don't see any performance impact. That is using the `poll_interval` as timeout for done future, doing that now

steinitzu:
@rudolfix  Few more adjustements and tests.

The duplicate "submit futures" block in `_get_source_item` is gone. I was getting huge overhead (2x runtime in some cases) without it originally, but now after all the changes I see no real difference so there's no need for it.

dlt/extract/pipe_iterator.py
steinitzu: So now this is pretty much it. "get source item" -> "wait for a resolved future (with timeout)" -> repeat. Just the order is opposite, which I don't think matters other than we don't waste 10ms polling at the beginning of extract, right?   
https://github.com/dlt-hub/dlt/blob/sthor%2Fparallelize-decorator/dlt/extract/pipe_iterator.py#L151-L169

steinitzu: I considered doing something with locks or async "Event" yielded from resource initially, to skip pending sources and only wait exactly as needed. But got complicated. But don't think we have too much overhead anyway.

steinitzu: > you can decrease CPU usage by moving done to special done dicts and improving _next_done_future method

This is good, will do that.

rudolfix (changes requested):
Now LGTM!
Check this commit: https://github.com/dlt-hub/dlt/pull/965/commits/6af5c1a931503db96fe37fe29acd5c05ec5a8f05 WDYT?

Please amend docs:
1. In resource.md please add a chapter
```
### Declare parallel and async resources
```
Where you just mention that async generators or parallel flag can be used to allow many resources to be executed at once. Then link to performance.md
In the performance kick our @dlt.defer example and use parallelize instead. Explain how it works with generators and regular functions.
provide code snippets, mind that code snippets are tested.

There are two examples in examples section where defer on transformers should be replaced with parallelize
Remember to change the source of example in the relevant test!
in chess_production:
```
# this resource takes data from players and returns profiles
    # it uses `defer` decorator to enable parallel run in thread pool.
    # defer requires return at the end so we convert yield into return (we return one item anyway)
    # you can still have yielding transformers, look for the test named `test_evolve_schema`
    @dlt.transformer(data_from=players, write_disposition="replace")
    @dlt.defer
```

and transformers
```
# a special case where just one item is retrieved in transformer
    # a whole transformer may be marked for parallel execution
    @dlt.transformer
    @dlt.defer
    def species(pokemon_details):
```

steinitzu:
Updated the docs. Couple of concerns:

1. I left one instance of `@defer` in the pokemon example since it's a different use case. We're not planning on removing it, right?
2. The "parallel pipelines asyncio" example I'm not sure if it works. For me it hangs sometimes and the generators don't look like they're even executing (print debugging) though the test passes https://github.com/dlt-hub/dlt/blob/sthor%2Fdestination-docs/docs/website/docs/reference/performance_snippets/performance-snippets.py#L123 . It's the same on the devel branch, was this working at some point?

rudolfix:
> Updated the docs. Couple of concerns:
> 
> 1. I left one instance of `@defer` in the pokemon example since it's a different use case. We're not planning on removing it, right?

No, it still works well when you want parallelize a loop. 


> 2. The "parallel pipelines asyncio" example I'm not sure if it works. For me it hangs sometimes and the generators don't look like they're even executing (print debugging) though the test passes https://github.com/dlt-hub/dlt/blob/sthor%2Fdestination-docs/docs/website/docs/reference/performance_snippets/performance-snippets.py#L123 . It's the same on the devel branch, was this working at some point?

It runs. But console output gets disabled when two pipelines are gathered at the same time. Something evil must happen there... It does not matter if we run the pools in extract step or if we run extract step at all. if I serialize all the calls by locking in `_run_pipeline` then the problem is gone. any ideas what could be happening?
also it indeed locks from time to time. probably connected to the above

Issue #931, opened 2024-02-02
Airflow helper: expose internal runner and introduce parallel execution of resources


rubenhelsloot:
## Requirements
1. allow airflow helper users to construct their own Dags but still to use internal runner with all the options (see example code at the very bottom)
2. introduce two new "decompose" types
- **parallel** will execute scc components in parallel, except the first component that must complete first (just in case)
- **parallel-isolated** will execute scc components in parallel. before execution pipeline will be forked under new name so state will be isolated
3. parallel options must be documented in docstring and docs

### additional requirement for `parallel`
We need to warn users if they modify the state of the resource in any of the tasks that are run in parallel. Please do the following:
when creating DAG enumerate all the resources in given component. if any of the resources has `incremental` property set, please log a warning like this
`The resource {resource.name} in task {task.name} is using incremental loading and may modify the state. Resources that modify the state should not run in parallel within the single pipeline as the state will not be correctly merged. Please use "serialize" or "parallel-isolated" modes instead`

First component may modify the state.

### **parallel-isolated** requirements
The goal of this decompose type is to run parallel several resources that modify the state. We achieve that by allocating a separate pipeline name (which will isolate the state) to each component. All such pipelines will write to the same dataset and work on the same schema,
1. * [ ] the overall DAG should be exactly  as in `parallel` mode, with a single starting task
2. * [ ] before creating task, the existing pipeline must be dropped `pipeline.drop()` - but we must be allowed to give it a separate name
3. * [ ] you must add a new argument to drop that will override the default pipeline name and pass it internally
4. * [ ] please add a test for the above alongside the regular drop test
5. * [ ] `drop` returns a clone of the original pipeline object. pass it to the task run

### dlt version

0.4.1

### Describe the problem

**Situation**
I have several Airflow dags that each select a subset of tables from SAP Hana to transfer using dlt. To do this, I'm using the generic SQL loader with a few minor tweaks. I define each subset as a dag using the `PipelineTasksGroup` class and split each subset into one task per table using `decompose`.

My problem is that on each heartbeat, Airflow evaluates the structure of all dags, which means it constructs the `dlt.source` object in every dag, and that means that it calls the SQLAlchemy constructor `Table()` on every named table. This actually connects to the Hana instance to fetch table metadata.

[...]

PR #966, opened 2024-02-14
feat(airflow): add the "parallel" mode

IlyaFaer:
Towards: https://github.com/dlt-hub/dlt/issues/931

Implements the `parallel` decomposition mode in the Airflow `dlt` helper.

### DIFF
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -155,11 +155,14 @@ class PipelineTasksGroup(TaskGroup):
             pipeline (Pipeline): An instance of pipeline used to run the source
             data (Any): Any data supported by `run` method of the pipeline
             decompose (Literal["none", "serialize", "parallel"], optional):
-                A source decomposition strategy into Airflow tasks. Defaults to "none".
-                serialize - decompose the source into a sequence of Airflow tasks.
-                parallel - decompose the source into a parallel Airflow task group (NOTE:
-                    in case the SequentialExecutor is used by Airflow, the tasks will
-                    remain sequential - use another executor, e.g. CeleryExecutor).
+                A source decomposition strategy into Airflow tasks:
+                    none - no decomposition, default value.
+                    serialize - decompose the source into a sequence of Airflow tasks.
+                    parallel - decompose the source into a parallel Airflow task group.
+                        NOTE: In case the SequentialExecutor is used by Airflow, the tasks
+                              will remain sequential. Use another executor, e.g. CeleryExecutor)!
+                        NOTE: The first component of the source is done first, after that
+                              the rest are executed in parallel to each other.
             table_name: (str): The name of the table to which the data should be loaded within the `dataset`
             write_disposition (TWriteDisposition, optional): Same as in `run` command. Defaults to None.
             loader_file_format (Literal["jsonl", "insert_values", "parquet"], optional): The file format the loader will use to create the load package.
@@ -315,8 +318,12 @@ class PipelineTasksGroup(TaskGroup):
 
                 end = DummyOperator(task_id=f"{task_name(pipeline, data)}_end")
 
-                start >> tasks >> end
-                return [start] + tasks + [end]
+                if tasks:
+                    start >> tasks >> end
+                    return [start] + tasks + [end]
+
+                start >> end
+                return [start, end]
             else:
                 raise ValueError("decompose value must be one of ['none', 'serialize', 'parallel']")
 
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -158,13 +158,15 @@ class PipelineTasksGroup(TaskGroup):
                 A source decomposition strategy into Airflow tasks. Defaults to "none".
                 serialize - decompose the source into a sequence of Airflow tasks.
                 parallel - decompose the source into a parallel Airflow task group (NOTE:
-                requires LocalExecutor to be used in Airflow instead of SequentialExecutor).
+                    in case the SequentialExecutor is used by Airflow, the tasks will
+                    remain sequential - use another executor, e.g. CeleryExecutor).
             table_name: (str): The name of the table to which the data should be loaded within the `dataset`
             write_disposition (TWriteDisposition, optional): Same as in `run` command. Defaults to None.
             loader_file_format (Literal["jsonl", "insert_values", "parquet"], optional): The file format the loader will use to create the load package.
                 Not all file_formats are compatible with all destinations. Defaults to the preferred file format of the selected destination.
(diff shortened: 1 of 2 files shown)


dlt/helpers/airflow_helper.py
```diff
@@ -135,7 +135,7 @@ def add_run(
         pipeline: Pipeline,
         data: Any,
         *,
-        decompose: Literal["none", "serialize"] = "none",
+        decompose: Literal["none", "serialize", "parallel"] = "none",
```
IlyaFaer, line 138: The method `add_run()` looks very big. Maybe we should split it into several smaller methods?

dlt/helpers/airflow_helper.py
```diff
@@ -291,8 +296,28 @@ def log_after_attempt(retry_state: RetryCallState) -> None:
...
+                    for source in data.decompose("scc"):
+                        tasks.append(make_task(pipeline, source))
+
+                end = DummyOperator(task_id=f"{group_name}_end")
+
+                start >> tasks >> end
+                return [start] + tasks + [end]
             else:
-                raise ValueError(decompose)
+                raise ValueError("decompose value must be one of ['none', 'serialize', 'parallel']")
```
IlyaFaer, line 320: I've added a `TaskGroup()`, as it seems to me parallel tasks should be grouped - it'll make analyzes easier, both in the runtime and in the Airflow UI.

tests/helpers/airflow_tests/test_airflow_wrapper.py
```diff
@@ -212,6 +212,64 @@ def dag_decomposed():
...
+    )
+    assert pipeline_dag_decomposed_counts == pipeline_standalone_counts
+
+    s = []
+    for task in dag_def.tasks:
+        s.append((task.task_id, task.downstream_task_ids, task.upstream_task_ids))
+
+    for task in dag_def.tasks[1:4]:
+        assert task.downstream_task_ids == set([dag_def.tasks[-1].task_id])
+        assert task.upstream_task_ids == set([dag_def.tasks[0].task_id])
```
IlyaFaer, line 266: The first and the last tasks are dummies only needed for grouping. So, we're checking that the real three tasks are up/down-streamed to them.

dlt/helpers/airflow_helper.py
```diff
@@ -291,8 +296,28 @@ def log_after_attempt(retry_state: RetryCallState) -> None:
...
+                start = DummyOperator(task_id=f"{group_name}_start")
+
+                with TaskGroup(group_id=group_name):
+                    for source in data.decompose("scc"):
+                        tasks.append(make_task(pipeline, source))
+
+                end = DummyOperator(task_id=f"{group_name}_end")
+
+                start >> tasks >> end
+                return [start] + tasks + [end]
```
IlyaFaer, line 318: Not 100% sure that we should return `start` and `end` too. On the other hand - they are marking the start and the end of the parallel group. What do you think?

dlt/helpers/airflow_helper.py
rudolfix: yes - I want to extract internal runner `def _run() -> None:` an let people use it to run any task. this is the second part of this ticket.

this will require refactoring of this function

dlt/helpers/airflow_helper.py
```diff
@@ -154,22 +154,26 @@ def add_run(
         Args:
             pipeline (Pipeline): An instance of pipeline used to run the source
             data (Any): Any data supported by `run` method of the pipeline
-            decompose (Literal["none", "serialize"], optional): A source decomposition strategy into Airflow tasks. Defaults to "none".
+            decompose (Literal["none", "serialize", "parallel"], optional):
+                A source decomposition strategy into Airflow tasks. Defaults to "none".
+                serialize - decompose the source into a sequence of Airflow tasks.
+                parallel - decompose the source into a parallel Airflow task group (NOTE:
```
rudolfix, line 160: right - but this is relevant for people running dags locally. so I'd just warn that in case of SequentialExecutor everything is serialized anyway. There are many many other executors that local - it is not the only alternative

dlt/helpers/airflow_helper.py
rudolfix: do we have a task group within task group? mind `with self:` up there. I do not think we need another one. also it may make sense for you to run Airflow with the UI and simply see the dags you are creating in it.

dlt/helpers/airflow_helper.py
rudolfix: hmmm so here is one trick I asked to implement:
the `start` should be the first source component. when you run the pipeline, it will do a lot of setup work (ie create schemas and initial state) that rather should not happen in parallel.

full parallelism happens in `parallel-isolated` mode

the end task makes sense. I think it is used with a condition that all predecessors succeeded. look at examples in the discussion in the original tickets

rudolfix (changes requested):
this is really good! I have one request for a `start` task.

dlt/helpers/airflow_helper.py
IlyaFaer: A-ah, I think I got it about `start`. So, no dummy is needed for a start, just the first element of the decomposition, and when it's finished, other tasks are ran in parallel.

About the `end`: by default, `all_success` trigger is used, so it should already work like that:
image

dlt/helpers/airflow_helper.py
IlyaFaer: Well, yes, to indicate the group of those tasks executed in parallel to each other (as some tasks under `self` are sequential). Okay, dropped the group.

dlt/helpers/airflow_helper.py
rudolfix: hmmm cool! but what if `tasks` is an empty list? maybe you should test a pipeline with just one component (resource).

had no idea that `>>` works with lists!

dlt/helpers/airflow_helper.py
```diff
@@ -154,22 +154,28 @@ def add_run(
...
```
rudolfix, line 160: Please explain that we run first source compinent (ie. resource) and then all the rest in parallel

rudolfix (changes requested):
just small details

dlt/helpers/airflow_helper.py
IlyaFaer: It appears it'll not work fine. The `start` and the `end` will not be up/down streamed to each other if the middle list is empty - a bit unexpected for me. Fixed.

dlt/helpers/airflow_helper.py
IlyaFaer: Done

rudolfix (approved):
LGTM!

PR #973, opened 2024-02-16
Docs update on how to set query limits.

dat-a-man:
### Related Issues

- Fixes #...
- Closes #...
- Resolves #...


docs/website/docs/dlt-ecosystem/verified-sources/salesforce.md
```diff
@@ -274,3 +270,11 @@ steps:
    > overwriting existing data. Conversely, the "task" endpoint supports "merge" mode for
    > incremental loads, updating or adding data based on the 'last_timestamp' value without erasing
    > previously loaded data.
+
+1. To limit the number of Salesforce API data requests, you can set `IS_PRODUCTION` variable
+   to `False` in "[salesforce/settings.py](https://github.com/dlt-hub/verified-sources/blob/master/sources/salesforce/settings.py)",
+   which limits API call requests to 100. To modify this limit, you can update the query limit in
```
VioletM, line 276: I would also add an explanation that by default data requests are not limited. Otherwise, user might think, that limit exists, but it's different.

dat-a-man: I added a few lines as per sf docs.

PR #979, opened 2024-02-19
feat(airflow): implement parallel-isolated mode

IlyaFaer:
Towards: https://github.com/dlt-hub/dlt/issues/931

Implements `parallel-isolated` decomposition mode.
Made in about the same way as the `parallel` mode, but the first source component is executed in parallel with others.

### DIFF
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -135,7 +135,7 @@ class PipelineTasksGroup(TaskGroup):
         pipeline: Pipeline,
         data: Any,
         *,
-        decompose: Literal["none", "serialize", "parallel"] = "none",
+        decompose: Literal["none", "serialize", "parallel", "parallel-isolated"] = "none",
         table_name: str = None,
         write_disposition: TWriteDisposition = None,
         loader_file_format: TLoaderFileFormat = None,
@@ -163,6 +163,9 @@ class PipelineTasksGroup(TaskGroup):
                               will remain sequential. Use another executor, e.g. CeleryExecutor)!
                         NOTE: The first component of the source is done first, after that
                               the rest are executed in parallel to each other.
+                    parallel-isolated - decompose the source into a parallel Airflow task group.
+                        NOTE: In case the SequentialExecutor is used by Airflow, the tasks
+                              will remain sequential. Use another executor, e.g. CeleryExecutor)!
             table_name: (str): The name of the table to which the data should be loaded within the `dataset`
             write_disposition (TWriteDisposition, optional): Same as in `run` command. Defaults to None.
             loader_file_format (Literal["jsonl", "insert_values", "parquet"], optional): The file format the loader will use to create the load package.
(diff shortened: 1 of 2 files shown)


IlyaFaer:
@rudolfix, I'm not 100% sure I understood what you mean by `before execution pipeline will be forked under new name so state will be isolated`. You mean that it'll be cloned with the `drop()` method here?

https://github.com/dlt-hub/dlt/blob/cf2e8864e1e29453b2daea69b2474e67e04cb8cb/dlt/helpers/airflow_helper.py#L202

It'll clone the pipeline, so seems to me it's what you mean by forking. Or maybe you mean that `ForkPipe()`?

rudolfix:
> @rudolfix, I'm not 100% sure I understood what you mean by `before execution pipeline will be forked under new name so state will be isolated`. You mean that it'll be cloned with the `drop()` method here?
> 
> https://github.com/dlt-hub/dlt/blob/cf2e8864e1e29453b2daea69b2474e67e04cb8cb/dlt/helpers/airflow_helper.py#L202
> 
> It'll clone the pipeline, so seems to me it's what you mean by forking. Or maybe you mean that `ForkPipe()`?

you are right! ticket was not clear. please take a look at updated requirements. also I added a requirement to add users if state is written in parallel mode

rudolfix (changes requested):
looks good! but spec was not clear so please see my comment...

dlt/helpers/airflow_helper.py
```diff
@@ -194,12 +202,12 @@ def task_name(pipeline: Pipeline, data: Any) -> str:
...
             # passing arguments to task function (_run) is serializing
             # them and running template engine on them
-            def make_task(pipeline: Pipeline, data: Any) -> PythonOperator:
+            def make_task(pipeline: Pipeline, data: Any, name: str = None) -> PythonOperator:
                 def _run() -> None:
                     # activate pipeline
                     pipeline.activate()
                     # drop local data
-                    task_pipeline = pipeline.drop()
+                    task_pipeline = pipeline.drop(pipeline_name=name)
```
rudolfix, line 210: clever!

dlt/helpers/airflow_helper.py
```diff
@@ -308,15 +316,53 @@ def log_after_attempt(retry_state: RetryCallState) -> None:
...
+                if pipeline.full_refresh:
+                    raise ValueError("Cannot decompose pipelines with full_refresh set")
+
+                # parallel tasks
+                tasks = []
+                t_name = task_name(pipeline, data)
+
+                start = DummyOperator(task_id=f"{t_name}_start")
+
+                for task_num, source in enumerate(data.decompose("scc"), start=1):
```
rudolfix, line 362: let's do the same trick and in the mode above. task1 goes first and then all the rest in parallel. first task possibly will create the dataset and initial schema. sorry if that was not clear in the ticket

dlt/helpers/airflow_helper.py
```diff
@@ -163,6 +168,9 @@ def add_run(
                               will remain sequential. Use another executor, e.g. CeleryExecutor)!
                         NOTE: The first component of the source is done first, after that
                               the rest are executed in parallel to each other.
+                    parallel-isolated - decompose the source into a parallel Airflow task group.
+                        NOTE: In case the SequentialExecutor is used by Airflow, the tasks
```
rudolfix, line 172: please explain that we run each task in a pipeline with different (but derived) name to separate the state

rudolfix (changes requested):
I still have one issue (probably tickets were not clear). tests are really top!

dlt/helpers/airflow_helper.py
IlyaFaer: In fact, first I did it, but reverted before committing

rudolfix: @IlyaFaer I think your revert is not pushed. otherwise this PR is top

PR #980, opened 2024-02-19
Issue 951

sultaniman:
This PR fixes multiples issues related to Pydantic schema inference and type detection logic.

### DIFF
--- a/dlt/common/data_types/type_helpers.py
+++ b/dlt/common/data_types/type_helpers.py
@@ -1,5 +1,6 @@
 import binascii
 import base64
+import dataclasses
 import datetime  # noqa: I251
 from collections.abc import Mapping as C_Mapping, Sequence as C_Sequence
 from typing import Any, Type, Literal, Union, cast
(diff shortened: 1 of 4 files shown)


tests/libs/test_typing_extended.py
```diff
@@ -0,0 +1,39 @@
+from typing import (
```
sultaniman, line 1: @rudolfix @sh-rp where should I put this test file because it doesn't fit with minimum test suite run profile I created this here?

rudolfix: why can't it fit in test_typing.py? just declare pydantic UUID4 like type yourself to drop the dependency!

rudolfix (changes requested):
nice improvement of the tests!
- did you really needed to update poetry.lock? library deps didn't change so maybe you could revert this change and push again (let's see what fails, if anything)
- you have consistent error in all common test on is Union so I think it is a problem with a test

PR #984, opened 2024-02-20
Use BIGNUMERIC for large decimals in bigquery

### LINKED ISSUE #983: BigQuery destination does not pick right decimal type
**Background**
After we've added a precision and scale to decimal types, we are not really checking if those are compatible with destinations. We'll add those checks in normalizer (other ticket). Here we fix BigQuery behavior which does not switch to BIGNUMERIC if precision/scale does not fit

**Requirements**
1. * [ ] assume that precision and scale is properly normalized, assume that precison and scale will fit one of decimal types
2. * [ ] pick the right decimal type using precision and scale of the column. refer to bigquery documentation when to switch (it looks quite complicated IMO)
3. * [ ] unit test (query builder) and end to end test (load some data containing big decimals that would not fit normally)

steinitzu:
### Description
* Adjust type mapper to use BIGNUMERIC for decimals with `precision > 38 or scale > 9`
* Tests -> added big decimal to table builder unit test and one e2e test with pipeline

### Related Issues
- Fixes #983

### DIFF
--- a/dlt/destinations/impl/bigquery/bigquery.py
+++ b/dlt/destinations/impl/bigquery/bigquery.py
@@ -54,7 +54,6 @@ class BigQueryTypeMapper(TypeMapper):
     sct_to_dbt = {
         "text": "STRING(%i)",
         "binary": "BYTES(%i)",
-        "decimal": "NUMERIC(%i,%i)",
     }
 
     dbt_to_sct = {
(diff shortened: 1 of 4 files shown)


rudolfix (changes requested):
heh this looks good but somehow vector database tests are not passing (see CI)

steinitzu:
> heh this looks good but somehow vector database tests are not passing (see CI)

Should be fixed in last commit. Was assuming all dbs had a decimal type/default precision set, weaviate doesn't.

PR #1012, opened 2024-02-27
clear normalized package in case it already existed

sh-rp:
…eady

### Description
In the current version of dlt, if the run failed with an exception during normalization, re-running the pipeline will yield an error such as 
FileExistsError: [Errno 17] File exists: '/Users/dave/.dlt/pipelines/my_pipe2/load/new/1709031489.142404'
Because at the beginning of a normalize step the folders for the load package are created. Since the normalization was run before but failed, this attempt will also fail. This PR solves this problem.


dlt/common/storages/load_package.py
```diff
@@ -328,12 +328,20 @@ def complete_job(self, load_id: str, file_name: str) -> str:
     #
 
     def create_package(self, load_id: str) -> None:
-        self.storage.create_folder(load_id)
+        self.storage.create_folder(load_id, exists_ok=True)
```
sh-rp, line 331: I am not sure wether by allowing this we will run into problems down the line. I can also try to make this more sensitive to only be allowed for normalize load packages. Or maybe I am also overlooking something else here..

sultaniman (approved):

dlt/common/storages/load_package.py
sultaniman: Can this affect incremental loads?

dlt/normalize/normalize.py
```diff
@@ -341,7 +341,7 @@ def spool_files(
 
     def spool_schema_files(self, load_id: str, schema: Schema, files: Sequence[str]) -> str:
         # normalized files will go here before being atomically renamed
-        self.load_storage.new_packages.create_package(load_id)
+        self.load_storage.new_packages.create_package(load_id, True)
```
rudolfix, line 344: if package exists you must delete the previous package, there's a partial content there possibly so can easily do the double load. btw. a good catch! try to improve test so we have a double load. not sure how to provoke a normalizer error. maybe with mocking?

dlt/common/storages/load_package.py
rudolfix: @sultaniman not more than a regular loads. in the normalization stage `dlt` deals only with files, format etc. the extraction logic does not matter

@sh-rp when I look at the code, it looks like that when new package is created it is always initially "private" for a given step. (in new state). so we do not need owner!

also we do not need exits_ok -> it is better we catch those double create errors. please see my other comment

sh-rp:
Deleting the whole half-normalized package seems to be safe to me, but maybe we want to move that into the pipeline code? I am not sure, your call. The test is a bit hacky, bc I modify the schema in the package but it works well.

sh-rp:
One other thought that I occured to me: imho the schema contracts override should override the schema stored in the loadpackage, so for this case that i have in the test you can normalize an extracted package with different rules. But I'm not 100% certain that this has to be the case. I somehow would expect it thought as a user.

rudolfix:
> One other thought that I occured to me: imho the schema contracts override should override the schema stored in the loadpackage, so for this case that i have in the test you can normalize an extracted package with different rules. But I'm not 100% certain that this has to be the case. I somehow would expect it thought as a user.

this is really a good point! we should overwrite package schema from the storage schema before running normalizer. why? the reason we have this step is to create schema in schema storage. so IMO we should always start the process with the newest schema!
```
# schema is updated, save it to schema volume
self.schema_storage.save_schema(schema)
```
so actually we had 2 subtle bugs after introducing end to end load package

I fixed this and also moved your tests into the normalize tests (there are lower level normalizer tests)

sh-rp (commented):
approved!

rudolfix (approved):
OK LGTM!

PR #1013, opened 2024-02-27
fix import schema yaml

sh-rp:
### Description
Schemas references by the import_schema_path are completely ignore by the current dlt version which is a bug. This PR makes sure these schemas get loaded if no other schema was discovered locally or in the destination.


tests/load/pipeline/test_import_export_schema.py
```diff
@@ -0,0 +1,109 @@
...
+    p = dlt.pipeline(
+        pipeline_name=name,
+        destination="duckdb",
+        import_schema_path=IMPORT_SCHEMA_PATH,
+        export_schema_path=EXPORT_SCHEMA_PATH,
+    )
+
+    p.run(EXAMPLE_DATA, table_name="person", columns={"id": {"data_type": "text"}})
+
+    # discovered columns are in export
```
sh-rp, line 60: @rudolfix is this the correct behavior, that columns provided in the columns are not going to be in the exported schema?

tests/load/pipeline/test_import_export_schema.py
```diff
@@ -0,0 +1,114 @@
...
+        export_schema_path=EXPORT_SCHEMA_PATH,
+        full_refresh=True,
+    )
+    p.run(EXAMPLE_DATA, table_name="person")
+    assert p.default_schema.tables["person"]["columns"]["id"]["data_type"] == "text"
+
+    # import schema is not overwritten
+    assert _get_import_schema(name).tables["person"]["columns"]["id"]["data_type"] == "text"
+
+    # export now includes the modified column type
```
sh-rp, line 111: related to the above, if importing the schema, the updated column will be in the export

sh-rp:
@rudolfix i think this got lost during the contracts merge.

rudolfix:
@sh-rp I'm almost 100% I've seen this code? is this restoring the old code somehow? this part is tricky but was working correctly before. we need to review this thing together

sh-rp:
@rudolfix the import schema behavior broke with your changes in Rfix/load package extract (#790). I am looking to find the exact line or part of code that changed it.

sh-rp:
Ok, what I think is going on is this:

* In the code before the above pr, you are running the load_schema function of the SchemaStorage in def spool_files of the normalizer, which will use the import schema if no other schema is found
* In the code after the above pr, you are running the self.normalize_storage.extracted_packages.load_schema(load_id) function to aquire the schema in the normalization step, but this will load the schema generated during the extraction which does not respect the import schema.

sh-rp:
@rudolfix one additional comment, now the import schema only get used if there is no other schema present. I can change it to always be used which somehow makes sense to me, we should be getting an error when trying to merge it in and has a conflict with an existing schema, so that should be safe.

rudolfix (changes requested):
@sh-rp import schema is implemented on the level of storage. there's a built in mechanism that
- when loading schema from storage, checks if import schema exists
- loads import schema if it were modified (remembers original imported schema hash)
- let the schema in storage evolve otherwise

I think we lost the import schema long time ago. You are right that normalize lost the ability to import schemas, but it looks to me extract lost that really long time ago.

I restored that (hopefully) by using load_schema, I also do that in a way that changes introduced during extract stay in memory until extact ends and they are comitted in the decorator

tests are amazing. there's not much to add. what we must test above all:
- only changes introduced in extract (so coming from source and resource hints) are present in import schema. normalize changes are not present there!
- second run of the pipeline with other resource hints will not overwrite import schema
- but pipeline schema will evolve freely
- until someone changes the import schema then all gets reset

dlt/pipeline/pipeline.py
```diff
@@ -161,13 +161,28 @@ def _wrap(self: "Pipeline", *args: Any, **kwargs: Any) -> Any:
...
-        self.schema_names = self._list_schemas_sorted()
-        return rv
+        try:
+            rv = f(self, *args, **kwargs)
+        except Exception:
+            # because we committed live schema before calling f, we may safely
+            # drop all changes in live schemas
+            for name in list(self._schema_storage.live_schemas.keys()):
+                try:
+                    schema = self._schema_storage.load_schema(name)
```
sultaniman, line 171: I am wondering if we can load schemas in parallel if there are more than N schemas?

sh-rp: @sultaniman what is the question exactly?

dlt/extract/extract.py
```diff
@@ -75,8 +75,9 @@ def choose_schema() -> Schema:
         """Except of explicitly passed schema, use a clone that will get discarded if extraction fails"""
         if schema:
             schema_ = schema
+        # if we have a default schema name, use it but do not collect any info from the existing schema
         elif pipeline.default_schema_name:
-            schema_ = pipeline.schemas[pipeline.default_schema_name].clone()
+            schema_ = Schema(pipeline.default_schema_name)
```
sh-rp, line 80: let's see if this passes

dlt/pipeline/pipeline.py
sultaniman: I mean the cases when there will be dozens or >100 of schemas when calling `self._schema_storage.live_schemas.keys()` and was wondering if we should do it in parallel.

dlt/common/storages/live_schema_storage.py
```diff
@@ -55,6 +56,19 @@ def commit_live_schema(self, name: str) -> Schema:
...
         return live_schema
 
+    def is_live_schema_committed(self, name: str) -> bool:
+        """Checks if live schema is present in storage and have same hash"""
+        live_schema = self.live_schemas.get(name)
+        if live_schema is None:
+            raise SchemaNotFoundError(name, f"live-schema://{name}")
+        try:
+            stored_schema_json = self._load_schema_json(name)
+            return live_schema.stored_version_hash == cast(
```
rudolfix, line 66: we must use `version_hash` here.
stored_version_hash - is a hash of last commit to storage
version_hash - incorporates all the changes
I wonder how the tests are passing. we must somehow bump schema version (which should only happen when saving)

dlt/extract/extract.py
```diff
@@ -75,6 +75,7 @@ def choose_schema() -> Schema:
         """Except of explicitly passed schema, use a clone that will get discarded if extraction fails"""
         if schema:
             schema_ = schema
+        # if we have a default schema name, use it but do not collect any info from the existing schema
```
rudolfix, line 78: I think is is a wrong comment? we still clone pipeline schema

rudolfix (changes requested):
there's issue with detecting changes in committed schemas. but somehow tests are passing?

and we still clone the pipeline schema, but comment days otherwise

dlt/extract/extract.py
sh-rp: ah yes, i had to revert this but forgot to remove this comment

dlt/common/storages/live_schema_storage.py
```diff
@@ -55,6 +56,17 @@ def commit_live_schema(self, name: str) -> Schema:
...
         return live_schema
 
+    def is_live_schema_committed(self, name: str) -> bool:
+        """Checks if live schema is present in storage and have same hash"""
+        live_schema = self.live_schemas.get(name)
+        if live_schema is None:
+            raise SchemaNotFoundError(name, f"live-schema://{name}")
+        try:
+            stored_schema_json = self._load_schema_json(name)
+            return live_schema.version_hash == cast(str, stored_schema_json.get("version_hash"))
```
sultaniman, line 66: is using `cast` relevant here, I mean we are not using it for any type hints and detections?

PR #1026, opened 2024-02-28
Add Dremio as a destination

maxfirman:
### Description
This PR adds Dremio as a destination to DLT

### Additional Context
- Tested locally against OSS Dremio running in docker and against Dremio Enterprise cluster. 
- dockercompose.yaml provided for creating local Dremio environment with minio as an object store for staging.

Some details for future PRs:
- implement `merge` write disposition using `MERGE INTO`
- implement table partition evolution


rudolfix:
@maxfirman this is amazing! we'll assign a reviewer tomorrow and add tests to our CI jobs after merging

sh-rp (commented):
Hey @maxfirman, thanks for this PR and Apologies that I took my time to have a look at this. I went over the code and the structure looks very good so far, thanks for all the work. The next steps would be to verify that all our standard tests for destination pass, and set up our CI to run your dremio destination for every PR, you can see the destination workflows in the github folder. I will set up dremio locally next week and try to run all the tests and get back to you then. I have a few questions now though you could answer if you have the time:

* Where is this pydremio taken from? Is that available somewhere and do we need to give credit?
* Did you run all the destination tests against your dremio implementation? These are all tests in tests/load with ACTIVE_DESTINATIONS set to {'dremio'}? You will also need to register variants of dremio for s3 and gcs in def destination_configs. 
* Should all the bucket providers also work for the local dremio instance? Or do we need the cloud for that?

If you need help running those tests or with fixing individual tests, let me know we're here to help, but for starters the answers to the three questions above will be enough for me to know how to proceed on monday.

sh-rp:
@maxfirman PS: You're welcome to join our community at dlthub.com/community and text me directly if you'd like to discuss this PR.

maxfirman:
Hi @sh-rp, thanks for taking the time to review this. To answer your questions:

- `pydremio.py` contains some code originally taken from dremio arrow flight examples repo [here](https://github.com/dremio-hub/arrow-flight-client-examples/tree/main/python/dremio-flight), with some modifications and additional code. The code in that repo is licensed under Apache 2.0.   I believe this means I need to copy the original license header text into the file and state that the file has been modified from the original.
-  I hadn't run the all tests/load tests, thanks for pointing out how this is done. I've just tried to run this now and there were some failures that I will need to address.
- I've only tested against s3 so far. Dremio supports Azure and Google storage, so there is no reason why this couldn't work. s3 is straightforward to test locally as I can spin up minio in docker container, I assume there are equivalents for google and azure but I'm unfamiliar. Would you be happy to only include s3 support as part of this PR? Google and Azure could always be added as a followup if demand arises.

I should be make the changes tomorrow otherwise it will be mid next week.

maxfirman:
@sh-rp I've got this mostly passing against the tests in tests/load except for the following tests:

- `tests.load.pipeline.test_replace_disposition.test_replace_disposition`
- `tests.load.pipeline.test_restore_state.test_restore_state_utils`

I've run out of time today to get these working, but will have another look next week.

sh-rp:
@maxfirman could you provide me with your secrets.toml that you are using to connect to the local dremio and s3-mocker?

maxfirman:
That's right. Here is the secrets.toml:
```
[destination.filesystem]
bucket_url = "s3://dlt-ci-test-bucket"

aws_access_key_id = "minioadmin" # copy the access key here
aws_secret_access_key = "minioadmin" # copy the secret access key here
endpoint_url = "http://127.0.0.1:9010"

[destination.dremio]
staging_data_source = "minio" # the name of the "Object Storage" data source in Dremio containing the s3 bucket

[destination.dremio.credentials]
database = "dlt" # the highest level of hierarchy under "data_source". e.g for AWS Glue catalog this is the Glue Database name
host = "localhost" # the Dremio hostname
port = 32010 # the Dremio Arrow Flight grpc port
data_source = "nas" # the name of the destination data source in Dremio
flatten=false # if true tables will be written as <dataset>__<table_name>, otherwise <dataset>/<table_name>
drivername="grpc" # either 'grpc' or 'grpc+tls'
username="dremio"
password="dremio123"

```

sh-rp:
@maxfirman thanks! When running tests/load/dremio/test_dremio.py::test_dremio with your creds and the docker containers running, I am getting:
```
E           dlt.pipeline.exceptions.PipelineStepFailed: Pipeline execution failed at stage load when processing package 1710345885.017133 with exception:
E
E           <class 'OSError'>
E           [Errno 5] An error occurred (426) when calling the CreateBucket operation: Upgrade Required
```

Is this something you have encountered? All the other tests also fail, but I wanted to get this one to run first.

maxfirman:
Unfortunately I haven't seen that one before. 

Perhaps the bucket creation failed during the docker compose startup and dlt is trying to create the bucket and failing for some reason.

Its worth confirming that the bucket in minio is being created by the `createbuckets` service in the docker compose. Once the docker-compose is running you should be able to login to the minio ui at `http://localhost:9011` and confirm that the bucket has been created. I would then check the Dremio UI at `http://localhost:9047` to confirm that minio is configured as a data source in Dremio and that you can see the `dlt-ci-test-bucket` within it.

tests/load/utils.py
```diff
@@ -302,6 +302,12 @@ def destinations_configs(
                 bucket_url=AZ_BUCKET,
                 extra_info="az-authorization",
             ),
+            DestinationTestConfiguration(
```
sh-rp, line 305: This needs to be removed here and appended to the array at line 211, similar to how databricks is appended there.

sh-rp (changes requested):
Hey @maxfirman, 

I got it to work now, I just restarted the whole thing, so you probably where right about a setup task failing :)

I have one important change in your test setup which escaped my initial look at your PR. I have marked that in the code. You added dremio to the special list of tests that are executed for the destinations that support loading from buckets. Since dremio does not have non-bucket-loading capabilities, this means that all the standard tests are not executed if you set it up this way. You'll have to move the test setup to the list of standard tests and run the tests again (which will be many more this time around, so expect some more red ones) 

Let me know how it goes and if you get stuck with any of the tests, just send me a message or contact me on slack and I can help you out.

dlt/destinations/impl/dremio/dremio.py
```diff
@@ -0,0 +1,251 @@
...
+"""
+        rows = self.sql_client.execute_sql(query, *db_params)
+
+        # if no rows we assume that table does not exist
+        schema_table: TTableSchemaColumns = {}
+        if len(rows) == 0:
+            return False, schema_table
+        for c in rows:
+            numeric_precision = (
+                c[3] if self.capabilities.schema_supports_numeric_precision else None
```
sh-rp, line 220: this will always be true for dremio, no? so you don't need to fallback to None

dlt/destinations/impl/dremio/dremio.py
```diff
@@ -0,0 +1,251 @@
...
+            )
+            numeric_scale = c[4] if self.capabilities.schema_supports_numeric_precision else None
```
sh-rp, line 222: same as above

dlt/destinations/impl/dremio/dremio.py
```diff
@@ -0,0 +1,251 @@
...
+                **self._from_db_type(c[1], numeric_precision, numeric_scale),
+            }
+            schema_table[c[0]] = schema_c  # type: ignore
+        return True, schema_table
+
+    def _create_merge_followup_jobs(self, table_chain: Sequence[TTableSchema]) -> List[NewLoadJob]:
+        def _new_temp_table_name(name_prefix: str) -> str:
+            return self.sql_client.make_qualified_table_name(f"_temp_{name_prefix}_{uniq_id()}")
+
+        class DremioMergeJob(SqlMergeJob):
```
sh-rp, line 235: don't define this as an inner class, you can put it at the top.

sh-rp: are you sure there won't be any name colission if the temp tables are all named the same? why could you not use the uniq_id of the standard merge job?

sh-rp: ah, ok, you need to make a qualified table name right? why not ammend the main sqlmergejob to do this? i hope it would work for all destination.. :)

dlt/destinations/impl/dremio/sql_client.py
```diff
@@ -0,0 +1,154 @@
...
+                return DatabaseTerminalException(ex)
+        else:
+            return ex
+
+    @staticmethod
+    def is_dbapi_exception(ex: Exception) -> bool:
+        return isinstance(ex, (pyarrow.lib.ArrowInvalid, pydremio.MalformedQueryError))
+
+    def create_dataset(self) -> None:
+        table_name = self.make_qualified_table_name("_dlt_dataset_exists")
```
sh-rp, line 133: why exactly do we need this table to be created? You probably mentioned it already.. in any case it needs to go into the docs that this happens.

sh-rp (changes requested):
Really nice work max! I have a couple of comments and questions, but it's really clean and clear.

dlt/destinations/impl/dremio/configuration.py
```diff
@@ -0,0 +1,55 @@
...
+@configspec
+class DremioCredentials(ConnectionStringCredentials):
+    drivername: str = "grpc"
+    username: str = None
+    password: Optional[TSecretStrValue] = None
+    host: str = None
+    port: Optional[int] = 32010
+    database: str = None
+
+    __config_gen_annotations__: ClassVar[List[str]] = ["password", "warehouse", "role"]
```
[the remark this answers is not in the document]
maxfirman: Not 100% sure what this does, but have removed the non-existent settings

dlt/destinations/sql_jobs.py
```diff
@@ -254,7 +254,7 @@ def gen_select_from_dedup_sql(
             1) To select the values for an INSERT INTO statement.
             2) To select the values for a temporary table used for inserts.
         """
-        order_by = "(SELECT NULL)"
+        order_by = "NULL"
```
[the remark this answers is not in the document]
maxfirman: Fingers crossed!

tests/load/pipeline/test_pipelines.py
```diff
@@ -76,11 +80,11 @@ def data_fun() -> Iterator[Any]:
     # this will create default schema
     p.extract(data_fun)
     # _pipeline suffix removed when creating default schema name
-    assert p.default_schema_name in ["dlt_pytest", "dlt"]
+    assert p.default_schema_name in ["dlt_pytest", "dlt", "dlt_jb_pytest_runner"]
```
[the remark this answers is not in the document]
maxfirman: This was needed to allow me to run tests through PyCharm IDE

dlt/destinations/impl/dremio/dremio.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class DremioLoadJob(LoadJob, FollowupJob):
+    def __init__(
+        self,
+        file_path: str,
+        table_name: str,
+        client: DremioSqlClient,
+        stage_name: Optional[str] = None,
+        staging_credentials: Optional[CredentialsConfiguration] = None,
```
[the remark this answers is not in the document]
maxfirman: Yes these were redundant. I've removed.

dlt/destinations/impl/dremio/dremio.py
```diff
@@ -0,0 +1,251 @@
...
+    def _get_table_update_sql(
+        self,
+        table_name: str,
+        new_columns: Sequence[TColumnSchema],
+        generate_alter: bool,
+        separate_alters: bool = False,
+    ) -> List[str]:
+        sql = super()._get_table_update_sql(table_name, new_columns, generate_alter)
+
+        if not generate_alter:
```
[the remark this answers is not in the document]
maxfirman: The way it is written it won't alter the partitions or sort order. From what I could tell it didn't appear that any other adapters had implemented ALTER PARTIONS/LOCALSORT?

dlt/destinations/impl/dremio/dremio.py
```diff
@@ -0,0 +1,251 @@
...
+            partition_list = [
+                self.capabilities.escape_identifier(c["name"])
```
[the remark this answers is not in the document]
maxfirman: I had a look at this, but the signature of `get_columns_names_with_prop` takes a `TTableSchema` rather than ` Sequence[TColumnSchema]`, so not sure it would work here.

dlt/destinations/impl/dremio/dremio.py
maxfirman: Yeah this is a bit wacky. Dremio doesn't have `CREATE TEMP TABLE`, so I needed to create a normal table in the same location. I'm not sure I can easily move this logic into `SqlMergeJob` the way everything is currently written using `@classmethod`s.

dlt/destinations/impl/dremio/sql_client.py
maxfirman: I've removed this after some further consideration. This was necessary to get the `test_default_schema_name_init_storage` test to pass. It relates back to the lack of `CREATE SCHEMA`. The tests are being run using a "NAS" data source which allows us to circumvent the `CREATE SCHEMA` issue, as it can create tables at arbitrary paths in the file system. I was using this dummy table as a marker that dataset had been created. I think this is probably unnecessary, and I've simply disabled the `test_default_schema_name_init_storage` test for Dremio.

dlt/destinations/impl/dremio/configuration.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def fingerprint(self) -> str:
+        """Returns a fingerprint of host part of a connection string"""
+        if self.credentials and self.credentials.host:
+            return digest128(self.credentials.host)
+        return ""
+
+    if TYPE_CHECKING:
+
+        def __init__(
```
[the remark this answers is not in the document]
sh-rp: fyi this is now no longer needed with changes merged it from level

tests/utils.py
```diff
@@ -60,7 +61,7 @@
 
 # filter out active destinations for current tests
 ACTIVE_DESTINATIONS = set(dlt.config.get("ACTIVE_DESTINATIONS", list) or IMPLEMENTED_DESTINATIONS)
-
+ACTIVE_DESTINATIONS = {"dremio"}
```
sh-rp, line 64: note to self: remove!

dlt/destinations/impl/dremio/dremio.py
sh-rp: Ok, I added the sql_client as an arg for the classmethod and now it looks a bit nicer :)

rudolfix (dismissed):
@maxfirman @sh-rp this is so good! this destination was quite different from the other but you were able to fit it perfectly into existing model and test structure (thx for making tests working)

I opened a PR that will be used that the newest code works #1194 

We upgraded how we deal with configurations, parquet files and redid our CI structure so when all tests in the branch above are passing I'll push the newest version here and then we merge it.

IMO this destination is really good even now (maybe I'm missing something that dremio does and we do not support :))
- we already have ticket to start using MERGE SQL statements
- adding gcp and azure blob storages are mostly matter of config and we can do that later
- using a sentinel table for destinations that have no schemas/datasets actually makes sense. we do that for vector databases. we also support empty dataset names so tables are not prefixed (not sure it applies here after all)
- we can implement dremio adapter to expose all the partition/clustering/retention features (some are implemented but adapters let users to add the hints conveniently to the resources)

I also found following comment in tests
```
    # schemaless destinations allow adding of root key without the pipeline failing
    # for now this is only the case for dremio
    # doing this will result in somewhat useless behavior
    destination_allows_adding_root_key = destination_config.destination == "dremio"
```
tbh. to me it looks like dremio does not enforce nulls and allows to add `root_id` column even if it is NOT NULL to an existing table with the data. @maxfirman could that be the case?

anyway thanks again for your amazing work!

maxfirman:
@rudolfix  thanks for time and effort reviewing this PR! 

Indeed Dremio does not support `NOT NULL` constraints on columns. I think you are right that is the cause of the test failure. 

I agree with all of your points. I have a few other thoughts as well for follow up PRs:

- Dremio recently fixed a bug that enables the use of the ADBC client https://github.com/apache/arrow-adbc/issues/1559. We could therefore remove the `pydremio.py` file and cut over to ADBC. I don't think there is any rush on this. 
- I'd like to test a wider selection of destination data source types in Dremio. We are somewhat hamstrung by the lack of `CREATE SCHEMA` support, which is why the CI tests run against a "NAS" (really local file system). I know that Dremio does support `CREATE FOLDER` [sql statements](https://docs.dremio.com/current/reference/sql/commands/nessie/create-folder/) for Nessie catalogs, so Nessie might be next logical thing to add to the testing pipelines as it should be possible to get all the tests passing against it. It might also be worth running a subset of the tests that don't require `CREATE SCHEMA` against a Hive Metastore data source.
- I can foresee that we may want to expose a "path_prefix" config option to build tables at arbitrarily nested levels in a folder structure.
- `MERGE INTO` sql support is vital given that Dremio doesn't support transactions, so I'm really glad to hear that is on your backlog!

rudolfix:
@maxfirman one question to:

> I'd like to test a wider selection of destination data source types in Dremio. We are somewhat hamstrung by the lack of CREATE SCHEMA support, which is why the CI tests run against a "NAS" (really local file system). I know that Dremio does support CREATE FOLDER [sql statements](https://docs.dremio.com/current/reference/sql/commands/nessie/create-folder/) for Nessie catalogs, so Nessie might be next logical thing to add to the testing pipelines as it should be possible to get all the tests passing against it. It might also be worth running a subset of the tests that don't require CREATE SCHEMA against a Hive Metastore data source.

we will be able to test against dremio cloud right? this "NAS" is also available there - the "local" filesystem is in the cloud then, right?

also: https://github.com/dlt-hub/dlt/issues/1199

maxfirman:
>this "NAS" is also available there - the "local" filesystem is in the cloud then, right?

@rudolfix I'm not 100% sure as I'm only familiar with their on-prem offering. 

If we are running tests against Dremio Cloud it may make sense to use the built in ["Arctic" catalog](https://docs.dremio.com/cloud/arctic/) (which is basically managed Nessie) instead. 

Nessie can be run in a docker container, so we could also extend the docker-compose.yaml to stand up a Nessie server if we wanted to test this out against a Dockerised deployment.

rudolfix (approved):

Issue #459, opened 2023-06-27
upgrade built-in streamlit app


rudolfix:
**Background**
We generate a Streamlit app with `dlt pipeline ... show`. It needs updates to catch up with existing features.

**Tasks**
1. * [ ] display the staging destination information
2. * [ ] display the staging dataset information (if there are any merge resources)
3. * [x] **allow to select schemas if there's more than 1**
4. * [x] display more information on a resource: primary keys, merge keys and other hints

look at `dlt pipeline` commands and port some of the functionalities. there are nice data structures below
1. * [ ] display source state for current schema
2. * [ ] display resource state for particular resource
3. * [ ] list extracted, normalized and loaded packages and allow to see the job, their statuses and errors inside
4. * [ ] display last run trace

Bonus task:
Use decompose method of the `DltSource` to show default execution DAG. you can use networkx to vis (make it optional dependency though)

sultaniman:
@rudolfix should we keep this open to work on the remaining points?

PR #1036, opened 2024-02-29
Upgrades to streamlit app

sultaniman:
This PR upgrades our streamlit app #459 which is launched when `dlt pipeline PIPELINE_NAME show` command is called.
It also includes better prompt to detect and install streamlit and additional packages if it is missing.

image

image


dlt/cli/helpers/dependencies.py
```diff
@@ -0,0 +1,71 @@
...
+            install_missing_package("streamlit")
+        else:
+            raise MissingDependencyException(
+                "DLT Streamlit Helpers",
+                ["streamlit"],
+                "DLT Helpers for Streamlit should be run within a streamlit app.",
+            )
+
+
+maybe_install_streamlit()
```
sultaniman, line 71: I will remove this and will call it in `streamlit_app/init.py`

rudolfix (changes requested):
Let's review UI together! First take:
- the sidebar should contain the most important information like pipeline name, destination etc. maybe last load if trace is present
- I do not want to distribute images with the library
- OR we should distribute the streamlit app as dlt extra: we extract this app as seprate dependency

otherwise I'm so happy we are pushing that forward

dlt/cli/helpers/dependencies.py
```diff
@@ -0,0 +1,71 @@
...
+        try:
+            check_call(["pip", "install", package])
+        except CalledProcessError as e:
+            print(f"Could not install {package}")
+            sys.exit(e.returncode)
+    finally:
+        globals()[package] = importlib.import_module(package)
+
+
+def query_yes_no(question, default="yes"):
```
rudolfix, line 20: what's wrong with current `echo.py` and `clikck` library we use. this should IMO work

rudolfix (changes requested):

sultaniman:
Closing this in favor of https://github.com/dlt-hub/dlt/pull/1060 because I decided to split automatic dependency installation utilities in the separate PR.

PR #1041, opened 2024-03-02
adds row_order to incremental

rudolfix:
### Description
1. users can declare if rows are ordered in incremental
2. this allows to early close the generator when cursor field is out of range
3. small fix in `add_limit`: does not evaluate generators
4. updates docs on incremental load and sql_database


dlt/extract/incremental/__init__.py
```diff
@@ -305,6 +315,16 @@ def _transform_item(
         row, start_out_of_range, end_out_of_range = transformer(row)
         self.start_out_of_range = start_out_of_range
         self.end_out_of_range = end_out_of_range
+        # if we know that rows are ordered we can close the generator automatically
+        # mind that closing pipe will not immediately close processing. it only closes the
+        # generator so this page will be fully processed
+        if self.row_order:
+            # ordered ascending, check if we cross upper bound
+            if self.row_order == "asc" and end_out_of_range:
+                self._pipe.close()
```
steinitzu, line 324: There is an issue with parallelized, which is more of a problem with the paralellized wrapper.
Because of: https://github.com/dlt-hub/dlt/blob/rfix%2Fadds-row-order-incremental/dlt/extract/utils.py#L189

`incremental.bind()` is called with an empty pipe and this close call fails. I'm not sure how to fix it, any ideas?

steinitzu (changes requested):
This looks great and the docs are good.  Just one issue I found when combined with `parallelized=True`  

Think some other optimizations are possible leading from this in the future, i.e. we can process chunks of rows more efficiently instead of checking each item.

dlt/extract/incremental/__init__.py
rudolfix: thanks for pointing out. we should not evaluate gens in the wrapper, only at runtime. I fixed `add_limit` to exactly do that... I need a test to reproduce it. there are other things that are passed in bind so it may be a bigger problem

rudolfix: @steinitzu pls review commits below, they fix the late bind issue

Issue #1043, opened 2024-03-04
Change LoadInfo structure to not create new tables every time it is loaded into the db


sh-rp:
### dlt version

0.4.5

### Describe the problem

load_info.metrics hast the load_ids as keys, which ends up creating a new table for every load when the max_nesting_level is the default. We should change it so all the load_info metrics stay in the same table but can be filtered by the load id.

Reported here: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1709412446163619

This change will require an update to the platform to ingest a different datastrcture.

### Steps to reproduce

Load any load_info object and see the table structure.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.10

### dlt destination

DuckDB

PR #1046, opened 2024-03-04
Refactor LoadInfo metrics layout schema

sultaniman:
This issue was reported by a community member in [slack](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1709412446163619) and related issue https://github.com/dlt-hub/dlt/issues/1043

When we capture load_info data in the destination database the following occurs:
* Pipeline execution slows down significantly after x number of incremental pipeline runs (it is back to the original speed when load_info is not captured in the database). My specific job (sourcing data from rest API capturing it in MotherDuck) slows down from 1.5 minutes to over 10 minutes (after a few hundred runs) and seems to keep getting slower.
* Each pipeline run creates a new table with one record (at least in my simple pipeline) with names like _load_info__metrics___1709356777_4431682.

## TODO
* [ ] Adjust `_ExtractInfo.metrics` from `Dict[str, List[ExtractMetrics]]` to just `List[ExtractMetrics]`,
* [ ] Add `load_id` field to `StepMetrics`,
* [ ] Adjust dependent code to use and extract information collection to lookup load ids etc.

### DIFF
--- a/dlt/common/pipeline.py
+++ b/dlt/common/pipeline.py
@@ -61,8 +61,12 @@ class _StepInfo(NamedTuple):
 class StepMetrics(TypedDict):
     """Metrics for particular package processed in particular pipeline step"""
 
+    load_id: str
+    """Related load id"""
+
     started_at: datetime.datetime
     """Start of package processing"""
+
     finished_at: datetime.datetime
     """End of package processing"""
 
@@ -72,7 +76,7 @@ TStepMetricsCo = TypeVar("TStepMetricsCo", bound=StepMetrics, covariant=True)
 
 class StepInfo(SupportsHumanize, Generic[TStepMetricsCo]):
     pipeline: "SupportsPipeline"
-    metrics: Dict[str, List[TStepMetricsCo]]
+    metrics: List[TStepMetricsCo]
     """Metrics per load id. If many sources with the same name were extracted, there will be more than 1 element in the list"""
     loads_ids: List[str]
     """ids of the loaded packages"""
@@ -99,6 +103,10 @@ class StepInfo(SupportsHumanize, Generic[TStepMetricsCo]):
             return max(m["finished_at"] for l_m in self.metrics.values() for m in l_m)
         except ValueError:
             return None
+
+    @property
+    def metrics_as_dict(self) -> Dict[str, List[TStepMetricsCo]]:
+        return {load_metric["load_id"]: load_metric for load_metric in self.metrics}
 
     def asdict(self) -> DictStrAny:
         # to be mixed with NamedTuple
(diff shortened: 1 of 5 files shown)


tests/pipeline/test_pipeline_load_info.py
```diff
@@ -0,0 +1,26 @@
...
+def test_pipeline_load_info_metrics_schema_is_not_chaning() -> None:
+    pipeline = dlt.pipeline(
+        pipeline_name="quick_start",
+        destination="duckdb",
+        dataset_name="mydata",
+    )
+
+    load_info = pipeline.run(data, table_name="users")
+
+    pipeline.run([load_info], table_name="_load_info")
```
rudolfix, line 19: 1. please add load_info, normalize_info and extract_info.
2. please load to a separate schema (you have schema argument to run()
3. use this second schema to compare hashes (mind that it wont be a default)

dlt/common/pipeline.py
```diff
@@ -61,8 +61,12 @@ class _StepInfo(NamedTuple):
 class StepMetrics(TypedDict):
     """Metrics for particular package processed in particular pipeline step"""
 
+    load_id: str
```
rudolfix, line 64: this is OK and overall even better than a dictionary. But in general the shape of the data is changed in `asdict()` like here:
```
def asdict(self) -> DictStrAny:
        # to be mixed with NamedTuple
        d: DictStrAny = self._asdict()  # type: ignore
        d["pipeline"] = {"pipeline_name": self.pipeline.pipeline_name}
        d["load_packages"] = [package.asdict() for package in self.load_packages]
        if self.metrics:
            d["started_at"] = self.started_at
            d["finished_at"] = self.finished_at
        return d
```
and the problem was that we didn't reformat `metric` to convert form dict to list.

tests/pipeline/test_pipeline_load_info.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def test_pipeline_load_info_metrics_schema_is_not_chaning() -> None:
+    pipeline = dlt.pipeline(
+        pipeline_name="quick_start",
+        destination="duckdb",
+        dataset_name="mydata",
+    )
+
+    load_info = pipeline.run(data, table_name="users")
```
rudolfix, line 17: please load something more complicated than this. ie. source with a resource that have several hints

tests/pipeline/test_pipeline_load_info.py
```diff
@@ -0,0 +1,26 @@
...
+
+    pipeline.run([load_info], table_name="_load_info")
+    first_version_hash = pipeline.default_schema.version_hash
+
+    load_info = pipeline.run(data, table_name="users")
```
rudolfix, line 22: here let's load again but we should add another source with different resource and some schema hints

tests/pipeline/test_pipeline_load_info.py
```diff
@@ -0,0 +1,26 @@
...
+    pipeline.run([load_info], table_name="_load_info")
```
rudolfix, line 23: you may have a schema difference when loading extract_info and the new resource has a new hint type. then indeed we may add column dynamically.

rudolfix (changes requested):

sultaniman:
closing in favor of https://github.com/dlt-hub/dlt/pull/1051

PR #1047, opened 2024-03-04
Make merge write_disposition fall back to append behavior when no merge or primary keys are set

sh-rp:
### Description
Currently the merge jobs will delete all data in the tables if no merge or primary keys are present, this behavior is not intuitive and can also lead to unexpected data loss.


dlt/destinations/sql_jobs.py
```diff
@@ -362,66 +362,70 @@ def gen_merge_sql(
...
-            # use unique hint to create temp table with all identifiers to delete
-            unique_columns = get_columns_names_with_prop(root_table, "unique")
-            if not unique_columns:
-                raise MergeDispositionException(
-                    sql_client.fully_qualified_dataset_name(),
-                    staging_root_table_name,
-                    [t["name"] for t in table_chain],
-                    f"There is no unique column (ie _dlt_id) in top table {root_table['name']} so"
-                    " it is not possible to link child tables to it.",
+        if not fallback_to_append:
```
rudolfix, line 371: maybe we could extract delete code block to separate method?

rudolfix (changes requested):
it is a good start :)
1. there's already test for it (`test_merge_no_merge_keys`)
2. we should also test what happens if we have hard delete column
3. filesystem also fallbacks to replace when there are no keys (and I think there's a test for it)
4. what about vector databases? they support limited merge (on primary key). I think they do not fallback to replace but make sure

rudolfix:
@sh-rp also you must update docs where I think we say we replace

PR #1051, opened 2024-03-05
Quick fix to serialize load metrics as list instead of a dictionary

sultaniman:
This issue was reported by a community member in [slack](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1709412446163619) and related issue https://github.com/dlt-hub/dlt/issues/1043

When we capture load_info data in the destination database the following occurs:
* Pipeline execution slows down significantly after x number of incremental pipeline runs (it is back to the original speed when load_info is not captured in the database). My specific job (sourcing data from rest API capturing it in MotherDuck) slows down from 1.5 minutes to over 10 minutes (after a few hundred runs) and seems to keep getting slower.
* Each pipeline run creates a new table with one record (at least in my simple pipeline) with names like _load_info__metrics___1709356777_4431682.

## TODO
* Create 2 separate sources to load
  1. One has resource w/ primary key
  2. Another with two resources
  3. load both sources
* load the first source
* load extract, normalize and load info
* run pipeline with schema parameter


dlt/common/configuration/providers/airflow.py
```diff
@@ -14,7 +14,6 @@ def name(self) -> str:
 
     def _look_vault(self, full_key: str, hint: type) -> str:
         """Get Airflow Variable with given `full_key`, return None if not found"""
-
```
rudolfix, line 17: what is your black version? why this line is removed? did we have unformatted code (maybe)

tests/pipeline/test_pipeline_load_info.py
```diff
@@ -0,0 +1,130 @@
...
+import random
+
+from tests.common.utils import json_case_path
+
+data = [
+    {"id": 1, "name": "Alice"},
+    {"id": 2, "name": "Bob"},
+]
+
+faker.Faker.seed(random.randint(0, 10000))
```
rudolfix, line 12: please remove it, we are not using it yet

tests/pipeline/test_pipeline_load_info.py
```diff
@@ -0,0 +1,130 @@
...
+        [pipeline.last_trace.last_normalize_info], table_name="_normalize_info", schema=schema
+    )
+
+    pipeline.run([pipeline.last_trace.last_extract_info], table_name="_extract_info", schema=schema)
+    schema_hashset.add(pipeline.schemas["nice_load_info_schema"].version_hash)
+
+    taxi_load_info = pipeline.run(
+        taxi_demand_source(),
+        table_name="taxi_demands",
+        primary_key="id",
```
rudolfix, line 86: also drop it here

tests/pipeline/test_pipeline_load_info.py
```diff
@@ -0,0 +1,130 @@
...
+    )
+    schema_hashset.add(pipeline.schemas["nice_load_info_schema"].version_hash)
+
+    pipeline.run([pipeline.last_trace.last_extract_info], table_name="_extract_info", schema=schema)
+    schema_hashset.add(pipeline.schemas["nice_load_info_schema"].version_hash)
+
+    assert len(schema_hashset) == 1
+
+    users_load_info = pipeline.run(
+        users_source(),
```
rudolfix, line 104: pass a list of sources to run, drop primary key and write disposition

rudolfix (changes requested):
1. We do not need another test file. Please add it to pipelines tests.
2. we do not use faker (yet). please revert lock and pyproject for now
3. why those primary keys are added to run? you have them in the resources

tests/pipeline/test_pipeline_load_info.py
sultaniman: Do you mean `[users_source(), taxi_demand_source()]` or just `[users_source()]`?

rudolfix (dismissed):
LGTM now,

@sultaniman 
1. there were bugs in test (loading tables into trace schema)
2. there was an inconsistent way to replace schema content (bug got fixed)

PR #1052, opened 2024-03-05
fix add_limit behavior in edge cases

sh-rp:
### Description
Limiting resources with add_limit was not working correctly or predictably for some values, with this update we:
* Emit 0 values if the limit is 0
* Disable the limiting for values None or below zero.


rudolfix (dismissed):
this is lovely! pls merge when test pass

rudolfix (approved):

PR #1053, opened 2024-03-05
adds test case where payload data contains PUA unicode characters

willi-mueller:
### Description
Adds end-to-end pipeline test case where payload data contains PUA unicode characters.

### Related Issues

- Complements #974

### Additional Context
Currently, we have only unit tests which verify that PUA are parsed correctly. This PR adds a full end-to-end pipeline test.

### Verification
`pytest tests/pipeline/test_pipeline.py::test_run_with_pua_payload` outputs:

Screenshot 2024-03-05 at 18 30 14


rudolfix (approved):
LGTM!

PR #1056, opened 2024-03-06
Docs: Add note about google secret name normalization

sultaniman:
Added a quick info section about expected secret name format for google secrets.
The exact page url: https://deploy-preview-1056--dlt-hub-docs.netlify.app/docs/general-usage/credentials/config_providers/#provider-key-formats

image


AstrakhantsevaAA (approved):
I would also add some info about google secrets (how to enable it and how to use with dlt). 
now this page and section doesn't look like a right place for this info.
Approve it as a hot fix to not lose this info

sultaniman:
@AstrakhantsevaAA should it be another sub section about enabling google secrets?

PR #1057, opened 2024-03-06
feat(core): drop default value for write disposition

### LINKED ISSUE #1031: `make_hints` returns default `write_disposition` if no disposition is provided as input
### dlt version

0.4.4

### Describe the problem

The `make_hints` function returns a `TResourceHints` object containing a value for `write_disposition` when `make_hints` was called without the `write_disposition` argument.

This is for example problematic when yielding hints in a resource:

image
image
image

### Expected behavior

`make_hints` only returns hints that were explicitly provided as input in the function's arguments, not using any default values.

### Steps to reproduce

```python
from dlt.extract.hints import make_hints

print(make_hints())
```

output:
`{'columns': {}, 'write_disposition': 'append'}`

### Operating system

Windows

### Runtime environment

Local

### Python version

3.8

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

IlyaFaer:
Fixes #1031


dlt/common/schema/utils.py
```diff
@@ -684,7 +684,8 @@ def new_table(
         assert schema_contract is None
     else:
         # set write disposition only for root tables
-        table["write_disposition"] = write_disposition or DEFAULT_WRITE_DISPOSITION
+        if write_disposition is not None:
```
rudolfix, line 687: OK this looks like good place to fix it. but way better would be to use `DEFAULT_WRITE_DISPOSITION` as a default argument value. then it will be backward compatible but also fix the bug

dlt/common/schema/utils.py
```diff
@@ -542,7 +542,7 @@ def get_write_disposition(tables: TSchemaTables, table_name: str) -> TWriteDispo
     """Returns table hint of a table if present. If not, looks up into parent table"""
     return cast(
         TWriteDisposition,
-        get_inherited_table_hint(tables, table_name, "write_disposition", allow_none=False),
+        get_inherited_table_hint(tables, table_name, "write_disposition", allow_none=True),
```
rudolfix, line 545: this should never allow for none. write disposition of a table chain must resolve. please revert

tests/common/schema/test_schema.py
```diff
@@ -817,13 +817,11 @@ def test_group_tables_by_resource(schema: Schema) -> None:
                 "columns": {},
                 "name": "c_products",
                 "resource": "products",
-                "write_disposition": "append",
```
rudolfix, line 820: please remove changes in all the tests below. they should pass. the only problem I see is in `def apply_hints(` in `hints.py where we use `make_hints` to create new table, there you must pass `DEFAULT_WRITE_DISPOSITION` if `write_disposition` is None

rudolfix (changes requested):
@IlyaFaer 
please add test that will reproduce and check the case described by Jorrit

dlt/common/schema/utils.py
IlyaFaer: Done

dlt/common/schema/utils.py
IlyaFaer: Not sure I understand. The task is to drop the default value, but if I put it into the default argument value, it'll be the same as it's now, no?

IlyaFaer: Or you mean the default value for disposition in `apply_hints()`?

rudolfix: > Not sure I understand. The task is to drop the default value, but if I put it into the default argument value, it'll be the same as it's now, no?
no, because when argument is skipped (like in 90% of cases) we'll get default values. when we pass None, None will be used. previously when you passed None, default value was used

rudolfix:
@IlyaFaer changing the default param is still a bad advice. Better we do not touch the `new_table` and its default behavior. Instead fix `make_hints` by popping `write_disposition` from `new_template` if the argument was None.

please see what happens when you do that
- write test to reproduce the bug
- there's IMO one usage in apply_hints that expects the old behavior let's see if any test catches that
- why are you pushing so many `fix` commits? I think you can amend and force push, I'd prefer to have meaningful commit names at the end

IlyaFaer:
@rudolfix 

> why are you pushing so many fix commits? I think you can amend and force push, I'd prefer to have meaningful commit names at the end

Well, tests take a lot of time. Usually, locally I ran only obviously related tests, and then push the code. While all the tests are passing, I'm working on something else.

I think, you always do "squash and merge" anyway, so in `master` it's just one commit.

rudolfix (dismissed):
LGTM! I fixed one edge case with empty apply hints on empty resource

PR #1060, opened 2024-03-06
Streamlit improvements

sultaniman:
This PR upgrades our streamlit app https://github.com/dlt-hub/dlt/issues/459 which is launched when dlt pipeline PIPELINE_NAME show command is called.
It also includes better prompt to detect and install streamlit and additional packages if it is missing.
Screenshot 2024-03-06 at 13 16 10
Screenshot 2024-03-06 at 13 15 57


dlt/cli/pipeline_command.py
```diff
@@ -101,12 +101,21 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
...
-                venv, "streamlit", "run", streamlit_helper.__file__, pipeline_name
+                venv,
+                "streamlit",
+                "run",
+                dashboard.__file__,
+                pipeline_name,
+                "--client.showSidebarNavigation",
+                "false",
+                # TODO: once done remove the option below
+                "--server.runOnSave",
```
sultaniman, line 117: I will remove this before merging once approved.

rudolfix (changes requested):
A few improvements in the UI before merging:

image
this is non-essential information. please move it to Load Info to the bottom. Also it would be cool to show the full pipeline state there but collapsed (hidden) by default

image
this is actually important. can we show it extended?

image
please do not show state view at all if state does not exist

image
order of the columns is for some reason random. it should be:
name | data_type | nullable

I'm happy we are able to test this

dlt/helpers/streamlit_app/blocks/table_hints.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+        )
+
+        if merge_keys:
+            table_hints.append("merge key(s): **%s**" % ", ".join(merge_keys))
+
+        st.subheader(f"Table: {table['name']}", divider=True)
+        st.markdown(" | ".join(table_hints))
+        if "resource" in table:
+            resource_state_info(
```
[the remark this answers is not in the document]
sultaniman: I will use it.

sultaniman:
Added sorting of table by column names.

dlt/cli/pipeline_command.py
```diff
@@ -101,13 +103,26 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
 
     if operation == "show":
         from dlt.common.runtime import signals
-        from dlt.helpers import streamlit_helper
+        from dlt.helpers.streamlit_app import dashboard
 
         with signals.delayed_signals():
+            # exit when there are not schemas
+            reload_flag = os.getenv("DLT_STREAMLIT_HOT_RELOAD")
```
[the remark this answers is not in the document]
sultaniman: Done!

dlt/cli/pipeline_command.py
```diff
@@ -101,13 +103,27 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
 
     if operation == "show":
         from dlt.common.runtime import signals
-        from dlt.helpers import streamlit_helper
+        from dlt.helpers.streamlit_app import dashboard
 
         with signals.delayed_signals():
+            # exit when there are not schemas
+            if pipelines_dir:
```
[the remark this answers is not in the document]
rudolfix: another option is to use:
```
def get_dlt_data_dir() -> str:
    """Gets default directory where pipelines' data will be stored
    1. in user home directory: ~/.dlt/
    2. if current user is root: in /var/dlt/
    3. if current user does not have a home directory: in /tmp/dlt/
    4. if DLT_DATA_DIR is set in env then it is used
    """
    if "DLT_DATA_DIR" in os.environ:
        return os.environ["DLT_DATA_DIR"]
```
then you do not need to modify anything in `attach`.
if you go this way use a context manager so environ is restored after you exit `custom_environ` in utils

dlt/helpers/streamlit_app/blocks/table_hints.py
```diff
@@ -0,0 +1,72 @@
...
+                pipeline,
+                current_schema.name,
+                table["resource"],
+            )
+
+        # table schema contains various hints (like clustering or partition options)
+        # that we do not want to show in basic view
+        def essentials_f(c: Any) -> Dict[str, Any]:
+            return {k: v for k, v in sorted(c.items()) if k in ["name", "data_type", "nullable"]}
```
rudolfix, line 68: please do not sort! the order of the columns corresponds to order in database/parquet etc!!

dlt/helpers/streamlit_app/dashboard.py
```diff
@@ -0,0 +1,66 @@
...
+    list_table_hints(pipeline, tables)
+    maybe_run_query(
+        pipeline,
+        show_charts=show_charts,
+        example_query=example_query,
+    )
+
+
+def display(pipeline_name: str) -> None:
+    pipeline = dlt.attach(pipeline_name)
```
rudolfix, line 49: `attach` has `pipelines_dir`, why not use it?

dlt/helpers/streamlit_app/dashboard.py
```diff
@@ -0,0 +1,66 @@
...
+import streamlit as st
+from dlt.cli import echo as fmt
+from dlt.helpers.streamlit_app.blocks.query import maybe_run_query
+from dlt.helpers.streamlit_app.blocks.table_hints import list_table_hints
+from dlt.helpers.streamlit_app.menu import menu
+from dlt.helpers.streamlit_app.widgets import schema_picker
+from dlt.pipeline import Pipeline
+
+
+def write_data_explorer_page(
```
rudolfix, line 14: why is this not a page?

dlt/helpers/streamlit_app/dashboard.py
```diff
@@ -0,0 +1,66 @@
...
+        example_query=example_query,
+    )
+
+
+def display(pipeline_name: str) -> None:
+    pipeline = dlt.attach(pipeline_name)
+    if pipelines_dir := os.getenv("DLT_PIPELINES_DIR"):
+        pipeline.pipelines_dir = pipelines_dir
+
+    st.session_state["pipeline_name"] = pipeline_name
```
rudolfix, line 53: why we use session to keep it? previous version had just one `display` method and could write a selected page. and each page had the same interface that was accepting the pipeline instance.

dlt/helpers/streamlit_app/menu.py
```diff
@@ -0,0 +1,42 @@
...
+from dlt.helpers.streamlit_app.utils import HERE
+from dlt.helpers.streamlit_app.widgets import logo, stat, tag, mode_selector
+from dlt.helpers.streamlit_app.blocks.load_info import last_load_info
+from dlt.pipeline.state_sync import load_pipeline_state_from_destination
+
+
+def menu(pipeline: dlt.Pipeline) -> None:
+    mode_selector()
+    logo()
+    st.page_link(f"{HERE}/dashboard.py", label="Explore data", icon="🕹️")
```
rudolfix, line 14: does that mean that each page is a separate script with separate main function? I think streamlit is a single page app by nature. I'm not against changing the structure but your change added code duplication and will result in more bugs later

dlt/helpers/streamlit_app/menu.py
```diff
@@ -0,0 +1,42 @@
...
+    st.page_link(f"{HERE}/pages/load_info.py", label="Load info", icon="💾")
+    pipeline_state_info(pipeline)
+    last_load_info(pipeline)
+
+
+def pipeline_state_info(pipeline: dlt.Pipeline) -> None:
```
rudolfix, line 20: why this is here and not in widgets or blocks?

dlt/helpers/streamlit_app/pages/load_info.py
```diff
@@ -0,0 +1,146 @@
...
+            display="block",
+            border_left_width=4,
+        )
+
+
+def show() -> None:
+    if not st.session_state.get("pipeline_name"):
+        st.switch_page("dashboard.py")
+
+    pipeline = dlt.attach(st.session_state["pipeline_name"])
```
rudolfix, line 133: this is clearly a code duplication and reproduces the whole app init logic just to run a different page than dashboard

dlt/helpers/streamlit_app/pages/load_info.py
```diff
@@ -0,0 +1,146 @@
...
+        stat(
+            label="Remote version",
+            value=remote_state_version,
+            display="block",
+            border_left_width=4,
+        )
+
+
+def show() -> None:
+    if not st.session_state.get("pipeline_name"):
```
rudolfix, line 130: you are clearly fighting with not having a single page app

rudolfix (changes requested):
we need to refactor this app a little because it got messy.
1. old app code is still there, why?
2. old app code was a single page app with single entry and init logic also it forced an interface for displaying pages which was good.
3. new app duplicates init code, pages are not self contained etc.
4. why dashboard is not a page?

I like that pages are broken into blocks etc. but top level structure is gone in this PR

dlt/helpers/streamlit_app/pages/load_info.py
sultaniman: I will extract this part

rudolfix:
@sultaniman now I see
> NOTE: I will remove old streamlit_helper.py once we merge this PR.

forget my (1) bullet point, didn't see that

dlt/helpers/streamlit_app/menu.py
sultaniman: yes each page has it's own main function in our case but we can also remove them this is however will not prevernt streamlit from complete re-renderings.

dlt/helpers/streamlit_app/pages/load_info.py
sultaniman: I extracted this logic now and take pipeline name from cli arguments.

dlt/helpers/streamlit_app/menu.py
sultaniman: Moved it to blocks.

dlt/helpers/streamlit_app/dashboard.py
sultaniman: the reason for this is that streamlit doesn't have an option to select a default page in multipage apps thus I had to put it in the main entrypoint file 😞

dlt/cli/pipeline_command.py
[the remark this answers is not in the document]
sultaniman: @rudolfix I re-used `get_dlt_data_dir` from configuration module

dlt/helpers/streamlit_app/menu.py
sultaniman: So now I created an `index.py` which redirects to dashboard when first loaded.

dlt/cli/pipeline_command.py
```diff
@@ -255,7 +278,12 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
                 tables = remove_defaults({"tables": package_info.schema_update})  # type: ignore
                 fmt.echo(fmt.bold("Schema update:"))
                 fmt.echo(
-                    yaml.dump(tables, allow_unicode=True, default_flow_style=False, sort_keys=False)
+                    yaml.dump(
```
sultaniman, line 281: black formatted this.

dlt/cli/pipeline_command.py
```diff
@@ -101,13 +114,23 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
 
     if operation == "show":
         from dlt.common.runtime import signals
-        from dlt.helpers import streamlit_helper
+        from dlt.helpers.streamlit_app import index
+
+        with signals.delayed_signals(), custom_environ({"DLT_PIPELINES_DIR": pipelines_dir or ""}):
```
rudolfix, line 119: OK so if we use custom env: why not to pass this as another argument? and use a env in the tests? it is way more clean

rudolfix (changes requested):
just two small thing in review. and one bigger:

when displaying schemas the order is not fixed! should be
name | data type | nullability
 and is reverse

dlt/cli/pipeline_command.py
```diff
@@ -25,8 +27,15 @@ def pipeline_command(
     verbosity: int,
     dataset_name: str = None,
     destination: TDestinationReferenceArg = None,
+    hot_reload: Optional[bool] = False,
     **command_kwargs: Any,
 ) -> None:
+    if pipelines_dir:
+        fmt.secho(
+            "When passing pipelines directory please make sure it contains relevant pipelines",
```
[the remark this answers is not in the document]
sultaniman: Fair I will remove it, I just thought it would be helpful for users if they get confused.

dlt/cli/pipeline_command.py
sultaniman: Done

dlt/cli/pipeline_command.py
```diff
@@ -101,13 +107,29 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
...
+            if hot_reload:
+                streamlit_cmd.append("--server.runOnSave")
+                streamlit_cmd.append("true")
+
+            streamlit_cmd.append("--")
+            streamlit_cmd.append(pipeline_name)
+            if pipelines_dir:
+                streamlit_cmd.append("--pipelines-dir")
+                streamlit_cmd.append(pipelines_dir)
+
```
sultaniman, line 130: @rudolfix now we pass the streamlit app specific params after terminating them `--`

dlt/helpers/streamlit_app/utils.py
```diff
@@ -0,0 +1,77 @@
...
+def parse_args() -> Tuple[str, str]:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("pipeline_name", nargs=1)
+    parser.add_argument(
+        "--pipelines-dir",
+        help="Pipelines working directory",
+        default=None,
+    )
+    known_options, _ = parser.parse_known_args()
+    return known_options.pipeline_name[0], known_options.pipelines_dir
```
sultaniman, line 26: @rudolfix and here is our simple parser which only parses known params

sultaniman:
Also we're now manually enforcing the order of columns.

rudolfix (approved):
LGTM! we should think how to eject this app from the lib so people can build their own stuff on top

tests/helpers/streamlit_tests/test_streamlit_show_resources.py
```diff
@@ -120,48 +124,41 @@ def test_multiple_resources_pipeline_with_dummy_destination():
...
     )
-    streamlit_app.run()
+    pipeline.run([{"n": 1}, {"n": 2}], table_name="numbers")
+    os.environ["DLT_TEST_PIPELINE_NAME"] = "test_resources_pipeline_dummy_destination"
+    base_args = ["dlt-show", "pipeline_name", "--pipelines-dir"]
+
+    def dummy_render(pipeline: dlt.Pipeline) -> None:
+        pass
+
+    old_args = sys.argv[:]
```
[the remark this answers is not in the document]
sultaniman: Fixed it in the cli runner PR 🙂

Issue #588, opened 2023-08-25
Full refresh mode adjustment


adrianbr:
**Background**
Please read **rationale** below on why we do it. In essence we want to rename current `full_refresh` mode into `dev_mode` (creation of new datasets with each run) and have certain functionalities of `dlt pipeline ... drop` command available from code.

**Tasks**
1. * [ ] we keep `full_refresh` flag but issue deprecation warning if it is used. there's too much code using it. it is ok to remove it from docs, configuration and docstrings though
2. * [ ] we add `dev_mode=True|False` that does the same as `full_refresh`. both should work the same. please use `dev_mode` in the code though
3. * [ ] we add a `refresh` argument to `run` method of `Pipeline` and `refresh` method to the Pipeline. they take 3 refresh modes
- `dataset` - drops the whole storage and completely wipes pipeline state, schemas and local storage
- 'full' the same as `dataset`
- `tables` - drops the tables belonging to all selected resources in all loaded sources, resets the corresponding resource schemas and states.
- 'data' - truncates all tables belonging to all selected resources in all loaded sources. keeps the schemas but resets the resource state, source level state stays intact
4. * [ ] we need to update the documentation to use `dev_mode` and `refresh`
5. * [ ] all methods should work with all destinations that we have.

**Implementation**
dev mode:
when deployments are generated we always check if `full_refresh` is present. we must start checking `dev_mode` as well

refresh:
Ideal implementation would reuse the code from drop command, but would make it in two steps:
1. modify / wipe local pipeline state and schemas to "simulate" dropped tables and state until we actually load data. we do very similar thing when resources are set to "replace" (maybe it can be re-used? but without changing the actual schema)
2. modify destination, this ideally should happen in `load.py`

details:

1. dataset - each destination client supports now `drop_storage` that will drop all data in the dataset. after that the local state can be wiped (by calling `sync_destination` or `drop`). there's no need to load anything into the dataset.
2. tables - we should automatically drop only the selected resources across all sources (or all tables across all schemas if nothing is selected). similar to setting write disposition to "replace" but instead of truncating we drop the tables before load and also delete tables from schema (before extract).  note: drop command depends on `sql_client` being present. but if it is not available, the method in `data` can be used
3. data - destination client, `initialize_storage` now accepts a list of tables to truncate. resource state is reset so it must be normalized and loaded into destination before proceeding. this is the same as setting write disposition to replace (but without changing the schema)

**Rationale**
------
[...]

PR #1063, opened 2024-03-07
New "refresh" mode and "dev_mode"

steinitzu:
### Description

Still needs some cleanup and testing.

* `full_refresh arg` replaced with `dev_mode`
*  Deprecation warning when `full_refresh=True`
* Add `refresh: "full" | "replace"` arg to pipeline
* `full` drops all tables/resource states/source state keys
* `replace` drops tables/state from the resources selected to extract

The new refresh mode introduces a new file to the load package: `dropped_tables.json`, this gets created in extract     
That was the simplest way I could think of to drop tables in appropriate stages (local schema and state is updated in extract and destination tables get dropped in load step).   
Otherwise we'd have to start poking at destination storage before extract which feels wrong.

Refresh works as follows:

1. Before extract -> uses drop command to wipe all/selected resource state and tables (local state only, destination tables are not deleted and normalize/load doesn't run as part of drop)
2. Add `dropped_tables.json` to load package with the tables found by drop command
3. In load stage -> read `dropped_tables.json` and drop the tables from destination schema (non-sql destination fall back to truncate on dataset init)

### Related Issues

- Resolves # #588


rudolfix (changes requested):
two biggest issues:
1. drop command must join extract transaction, should not persist anything
2. IMO you can merge `init_client` and `_refresh` and have way less code. we may need a drop_schema.json which contains tables that should be dropped

dlt/common/pipeline.py
```diff
@@ -47,6 +48,9 @@
 from dlt.common.utils import RowCounts, merge_row_counts
 
 
+TRefreshMode = Literal["full", "replace"]
```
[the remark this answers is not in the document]
steinitzu: 1. To be clear by `drop_dataset`, you don't mean `DROP SCHEMA ...`, but just all tables in the dlt schema, right? I.e. we don't want to wipe anything unrelated to the pipeline.  
2. What's the difference between `drop_data` and `write_disposition` replace? Is it just case of getting the same behaviour without changing `write_disposition` in schema?

rudolfix:
@steinitzu this looks pretty cool now. in the meantime we refactored extract code so please merge the current changes. sorry for that!

dlt/extract/extract.py
```diff
@@ -367,14 +377,44 @@ def extract(
...
+                    _state, _ = pipeline_state(Container())
+                    new_schema, new_state, drop_info = drop_resources(
+                        source.schema,
+                        _state,
+                        resources=_resources_to_drop,
+                        drop_all=self.refresh == "drop_dataset",
+                        state_only=self.refresh == "drop_data",
+                        state_paths="*" if self.refresh == "drop_dataset" else [],
+                    )
+                    _state.update(new_state)
```
rudolfix, line 407: we should replace the _state with a `new_state`? you just update it here. hmmm it may work because drop command just changes `sources` in the state so they will be overwritten

dlt/extract/extract.py
```diff
@@ -367,14 +377,44 @@ def extract(
...
+                    if drop_info["tables"]:
```
rudolfix, line 408: yeah this is good. but IMO we should also save tables to truncate if `drop_data` is used. so we have `dropped_tables` and `truncated_tables` in package state. no?

dlt/extract/extract.py
```diff
@@ -367,14 +377,44 @@ def extract(
...
+                        drop_tables = [
+                            table
+                            for table in source.schema.tables.values()
+                            if table["name"] in drop_info["tables"]
+                        ]
+                        load_package.state["dropped_tables"] = drop_tables
+                    source.schema.tables.clear()
+                    source.schema.tables.update(new_schema.tables)
```
rudolfix, line 416: I think we'll have a problem here. because the source schema is used to update the pipeline schema. so if you delete a table, it will be kept anyway.  not sure what we could do here. maybe this piece of code should be removed from `extract` and put into a separate function that `pipeline` will call just before extract? then you can return a modified schema

dlt/load/load.py
```diff
@@ -68,6 +68,7 @@ def __init__(
         config: LoaderConfiguration = config.value,
         initial_client_config: DestinationClientConfiguration = config.value,
         initial_staging_client_config: DestinationClientConfiguration = config.value,
+        refresh: Optional[TRefreshMode] = None,
```
rudolfix, line 71: why we need this? IMO it is enough if we keep dropped and truncated tables in package state and just execute the changes

dlt/load/utils.py
```diff
@@ -128,12 +140,26 @@ def _init_dataset_and_update_schema(
...
     staging_info: bool = False,
+    drop_tables: Optional[List[TTableSchema]] = None,
 ) -> TSchemaTables:
     staging_text = "for staging dataset" if staging_info else ""
     logger.info(
         f"Client for {job_client.config.destination_type} will start initialize storage"
         f" {staging_text}"
     )
+    if drop_tables:
+        old_schema = job_client.schema
```
rudolfix, line 151: we already dropped the tables from the `old_schema` right? so if we drop them again this means they were added from the data. (or we have a bug that we do not drop this tables effectively).

maybe we do not need to drop anything here? we should have here a schema in a final state

dlt/extract/extract.py
```diff
@@ -367,14 +377,44 @@ def extract(
...
+                        source.schema,
+                        _state,
+                        resources=_resources_to_drop,
+                        drop_all=self.refresh == "drop_dataset",
+                        state_only=self.refresh == "drop_data",
+                        state_paths="*" if self.refresh == "drop_dataset" else [],
+                    )
+                    _state.update(new_state)
+                    if drop_info["tables"]:
+                        drop_tables = [
```
rudolfix, line 409: we need to drop tables from a full table chain. look what happens in load. also we have a marker (seen-data) that tells which tables were really created.

same thing for truncated tables. we should truncate table chains.

maybe when we have table chains we do not really need to modify schema in load.py

dlt/normalize/normalize.py
```diff
@@ -354,6 +354,12 @@ def spool_files(
         self.load_storage.new_packages.save_schema_updates(
             load_id, merge_schema_updates(schema_updates)
         )
+        # self.load_storage.new_packages.save_dropped_tables(
+        #     load_id, self.normalize_storage.extracted_packages.load_dropped_tables(load_id)
+        # )
+        self.load_storage.new_packages.save_load_package_state(
```
rudolfix, line 360: OK! so we do not move the state from extract package to normalize package? maybe that should be a part of the normalize_storage

rudolfix (changes requested):
@steinitzu I think we can still simplify and restructure the code. see my comments

dlt/extract/extract.py
steinitzu: I was updating it so it would mutate the state instance in context, wasn't sure how to replace the state dict completely or if there's a better to extract new state?  

> hmmm it may work because drop command just changes sources in the state so they will be overwritten
Yeah exactly cause of that it works

dlt/extract/extract.py
steinitzu: True, that would be cleaner

dlt/load/load.py
steinitzu: Yes, keeping truncated tables would be cleaner. This doesn't need to know about refresh

dlt/load/utils.py
steinitzu: I think at this point the dropped tables are back in `old_schema` if the same resources are being loaded again.  
So I had to do this and replace stored schema in `_dlt_version` so that during schema update it correctly detects these tables don't exist and they get re-created.   
It's a little convoluted though for sure, maybe there's a better way.

dlt/extract/extract.py
steinitzu: Where does the pipeline schema get updated? Seems to work like this, after extract `pipeline.default_schema` doesn't contain any of the dropped tables

dlt/extract/extract.py
rudolfix: maybe just replace `sources`? as nothing else is changed in drop command. or maybe I'm wrong?

dlt/normalize/normalize.py
rudolfix: this one got fixed! If you take the newest devel!

dlt/pipeline/helpers.py
```diff
@@ -152,13 +125,14 @@ def is_empty(self) -> bool:
...
         )
 
-    def _drop_destination_tables(self) -> None:
-        table_names = [tbl["name"] for tbl in self.tables_to_drop]
-        for table_name in table_names:
-            assert table_name not in self.schema._schema_tables, (
-                f"You are dropping table {table_name} in {self.schema.name} but it is still present"
-                " in the schema"
-            )
+    def _drop_destination_tables(self, allow_schema_tables: bool = False) -> None:
```
rudolfix, line 128: Do we still need it? we load state package anyway so all this modifications happen in the load right?

dlt/extract/extract.py
steinitzu: @rudolfix  I'm not 100% what you mean. (aside from checking seen-data) the table chain is already generated in drop far as I can tell, the `dropped_tables` list is the result of:

```python
def group_tables_by_resource(
    tables: TSchemaTables, pattern: Optional[REPattern] = None
) -> Dict[str, List[TTableSchema]]:
    """Create a dict of resources and their associated tables and descendant tables
    If `pattern` is supplied, the result is filtered to only resource names matching the pattern.
    """
    result: Dict[str, List[TTableSchema]] = {}
    for table in tables.values():
        resource = table.get("resource")
        if resource and (pattern is None or pattern.match(resource)):
            resource_tables = result.setdefault(resource, [])
            resource_tables.extend(get_child_tables(tables, table["name"]))
    return result

...
data_tables = {t["name"]: t for t in schema.data_tables()}  # could probably include seen_data_only here?
resource_tables = group_tables_by_resource(data_tables, pattern=resource_pattern)
```

dlt/common/storages/load_package.py
```diff
@@ -480,6 +486,20 @@ def save_schema_updates(self, load_id: str, schema_update: TSchemaTables) -> Non
         ) as f:
             json.dump(schema_update, f)
 
+    def save_dropped_tables(self, load_id: str, dropped_tables: Sequence[str]) -> None:
```
rudolfix, line 489: we do not need those anymore (package state is used, right?)

dlt/extract/extract.py
rudolfix: @steinitzu I overlooked this. we are interested in root tables that seen data so
```
def _write_empty_files(
        self, source: DltSource, extractors: Dict[TDataItemFormat, Extractor]
    ) -> None:
        schema = source.schema
        json_extractor = extractors["object"]
        resources_with_items = set().union(*[e.resources_with_items for e in extractors.values()])
        # find REPLACE resources that did not yield any pipe items and create empty jobs for them
        # NOTE: do not include tables that have never seen data
        data_tables = {t["name"]: t for t in schema.data_tables(seen_data_only=True)}
        tables_by_resources = utils.group_tables_by_resource(data_tables)
```
looks like this should also go through truncated table state. and other things in `_write_empty_files`... better we fix it after this PR is merged

dlt/load/utils.py
rudolfix: radical solution would be to:
1. add a flag to `update_stored_schema` that always forces to re-check the schema and create/update existing tables
2. remove `replace_schema` from `drop_tables`
3. remove `_drop_destination_tables` from drop command and always create empty package (just modified state, schema and tables that needs to be dropped and truncated and do this here)

rudolfix (changes requested):

dlt/load/utils.py
steinitzu: > add a flag to update_stored_schema that always forces to re-check the schema and create/update existing tables
This would have to run every time, right? We don't know if a table was dropped in some previous load.

steinitzu: Potentially solved it with this: https://github.com/dlt-hub/dlt/pull/1063/commits/8990f114e9664b16fad55d6d2eccc73c217af5d2  
`replace_schema` is not needed now, but I check both `version_hash` and `version` when getting stored destination schema, so we are not comparing with the "pre-drop" schema. Maybe version number or previous hashes or something could be part of the hash so it's simpler?   
Are there any downsides to this?

dlt/pipeline/helpers.py
steinitzu: Nope, now drop command creates a load package: https://github.com/dlt-hub/dlt/blob/sthor%2Ffull-refresh-adjustment/dlt/pipeline/helpers.py#L150-L154

dlt/pipeline/configuration.py
```diff
@@ -25,9 +26,15 @@ class PipelineConfiguration(BaseConfiguration):
     use_single_dataset: bool = True
     """Stores all schemas in single dataset. When False, each schema will get a separate dataset with `{dataset_name}_{schema_name}"""
     full_refresh: bool = False
+    """Deprecated. Use `dev_mode` instead. When set to True, each instance of the pipeline with the `pipeline_name` starts from scratch when run and loads the data to a separate dataset."""
+    dev_mode: bool = False
     """When set to True, each instance of the pipeline with the `pipeline_name` starts from scratch when run and loads the data to a separate dataset."""
     progress: Optional[str] = None
     runtime: RunConfiguration
+    refresh: Optional[TRefreshMode] = None
```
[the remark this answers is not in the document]
steinitzu: https://github.com/dlt-hub/dlt/pull/1063/commits/7120835610f283c210209d4af821f0819cff6152

dlt/pipeline/pipeline.py
```diff
@@ -1063,8 +1077,33 @@ def _wipe_working_folder(self) -> None:
...
 
+    def _refresh_source(self, source: DltSource) -> Tuple[Schema, TPipelineState, Dict[str, Any]]:
+        if self.refresh is None or self.first_run:
+            return source.schema, self.state, {}
+        _resources_to_drop = (
+            list(source.resources.extracted) if self.refresh != "drop_dataset" else []
+        )
+        drop_result = drop_resources(
+            source.schema,
+            self.state,
```
rudolfix, line 1088: state from context manager should be used, so current_pipeline_dyate

dlt/pipeline/pipeline.py
```diff
@@ -1063,8 +1077,33 @@ def _wipe_working_folder(self) -> None:
     def _attach_pipeline(self) -> None:
         pass
 
+    def _refresh_source(self, source: DltSource) -> Tuple[Schema, TPipelineState, Dict[str, Any]]:
```
rudolfix, line 1080: maybe we could move this to helpers, pipeline.py is large enough

dlt/pipeline/pipeline.py
```diff
@@ -1063,8 +1077,33 @@ def _wipe_working_folder(self) -> None:
...
+
     def _extract_source(
-        self, extract: Extract, source: DltSource, max_parallel_items: int, workers: int
+        self,
+        extract: Extract,
+        source: DltSource,
+        max_parallel_items: int,
+        workers: int,
+        with_refresh: bool = False,
+        load_package_state_update: Optional[Dict[str, Any]] = None,
```
rudolfix, line 1106: @steinitzu @sh-rp IMO we should unify how we  generate package state in the pipeline. right now there's a hack we use to save pipeline state into package state at the end of extract
```
for source in data_to_sources(
                    data,
                    self,
                    schema,
                    table_name,
                    parent_table_name,
                    write_disposition,
                    columns,
                    primary_key,
                    schema_contract,
                ):
                    if source.exhausted:
                        raise SourceExhausted(source.name)

                    self._extract_source(
                        extract_step, source, max_parallel_items, workers, with_refresh=True
                    )
                # extract state
                state: TPipelineStateDoc = None
                if self.config.restore_from_destination:
                    # this will update state version hash so it will not be extracted again by with_state_sync
                    state = self._bump_version_and_extract_state(
                        self._container[StateInjectableContext].state, True, extract_step
                    )
                # commit load packages with state
                extract_step.commit_packages(state)
```
here:                 extract_step.commit_packages(state)

and there's a mechanism from @steinitzu to inject state during extract:
```
def extract(
....
      if load_package_state_update:
                    load_package.state.update(load_package_state_update)  # type: ignore[typeddict-item]
```
IMO we do not need both. and what steini did seems more generic so if we just add pipeline state to `load_package_state_update`
```
if should_extract and extract_state:
            data, doc = state_resource(state)
            extract_ = extract or Extract(
                self._schema_storage, self._normalize_storage_config(), original_data=data
            )
            self._extract_source(
                extract_,
                data_to_sources(data, self, schema or self.default_schema)[0],
                1,
                1,
                load_package_state_update=load_package_state_update,
```
under the right key, everything well work (and we fix one inconsistency - if we have many sources we add pipeline state to all packages, not only to the one that contains state document)
@sh-rp WDYT?

rudolfix, line 1106: btw. this parameter is never passed from outside so we can just remove it IMO

rudolfix (changes requested):
two important things:
1. investigate why we need to SELECT by version number as well (not only hash) when finding schema in db
2. unify how package state is written by the pipeline

dlt/pipeline/pipeline.py
sh-rp: yes, that sounds good to me. I was thinking we should put the pipeline state into the load package state always and let the destination figure out wether it needs to be stored based on the hash (right now it's only the filesystem that is using this anyway)

sh-rp (changes requested):

dlt/destinations/job_client_impl.py
```diff
@@ -375,10 +379,17 @@ def get_stored_state(self, pipeline_name: str) -> StateInfo:
             return None
         return StateInfo(row[0], row[1], row[2], row[3], pendulum.instance(row[4]))
 
-    def get_stored_schema_by_hash(self, version_hash: str) -> StorageSchemaInfo:
+    def get_stored_schema_by_hash(
+        self, version_hash: str, version: Optional[int] = None
```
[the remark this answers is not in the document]
steinitzu: It's a problem when trying to reload a previous schema version after some tables are dropped.  

1. Load some source -> stored schema has tables a, b, c
2. Drop table "b" -> new stored schema has tables a, c -> new hash
3. Load resource "b" again -> new schema is identical to 1. and has same hash -> update_stored_schema incorrectly decides that current schema is 1. (Where table b alredy exists) -> no changes are detected, dropped table is not recreated and load fails

The version number makes sense to me, cause it's bumped only when there are changes. So even if we have an identical schema to some previous version we need to know there have been changes in between

dlt/pipeline/pipeline.py
steinitzu: Makes sense. Could we then not extract this `state_resource`? I.e. for drop command just create an empty load package with state and schema and no sources.

rudolfix: @steinitzu sometimes we must update the state. right now AFAIK you somehow force the state hash to change so state is always extracted. you can try empty package (when there's not state change). I expect it will be ignored by the normalizer. but actually we should push it to load. you can try to fix it

steinitzu: So I tried this, most of the changs in https://github.com/dlt-hub/dlt/pull/1063/commits/94020900aafc70b22712a91fc6b1e70d0c3fd15d  
State table is by default created like other `_dlt_` tables and state is inserted in destination in `complete_load()`  

Loading "empty" load package works (only load package state and new schema), but I had to remove this in normalize:
```python
            # if len(schema_files) == 0:
            #     # delete empty package
            #     self.normalize_storage.extracted_packages.delete_package(load_id)
            #     logger.info(f"Empty package {load_id} processed")
            #     continue
```
Is this just optimization or some other reason to skip empty packages? Would it be better to load it conditionally, e.g. with some marker file that marks it as "state only" package?

rudolfix:
@steinitzu do you need a review here? PR is still draft

dlt/destinations/job_client_impl.py
```diff
@@ -186,7 +187,10 @@ def update_stored_schema(
     ) -> Optional[TSchemaTables]:
         super().update_stored_schema(only_tables, expected_update)
         applied_update: TSchemaTables = {}
-        schema_info = self.get_stored_schema_by_hash(self.schema.stored_version_hash)
+        self.schema.version
```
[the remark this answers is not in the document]
rudolfix: @steinitzu please remove this :)

dlt/destinations/job_client_impl.py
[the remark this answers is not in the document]
rudolfix: OK but I still think it is too hacky. We already were dropping tables in the drop command that was not needed. IMO when you drop any tables and modify the schema we should remove the schema with a hash that become invalid. I think we were doing that in the drop command, no?

you can pass the schema hash to delete in the package state

dlt/pipeline/pipeline.py
rudolfix: @steinitzu I do not see any changes in `normalize` so let's keep it like that. we'll have a bigger push to reorganize how we write state

dlt/pipeline/__init__.py
```diff
@@ -30,6 +30,8 @@ def pipeline(
     import_schema_path: str = None,
     export_schema_path: str = None,
     full_refresh: bool = False,
+    dev_mode: bool = False,
+    refresh: Optional[TRefreshMode] = None,
```
rudolfix, line 34: docstring missing

tests/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,266 @@
...
+        "refresh_full_test",
+        destination="duckdb",
+        refresh="drop_dataset",
+        dataset_name="refresh_full_test",
+    )
+
+    info = pipeline.run(my_source())
+    assert_load_info(info)
+
+    # Second run of pipeline with only selected resources
```
rudolfix, line 70: would be good to mutate schema of one of the resources. this mode should allow to replace ie nullable with not nullable column or change the data type (but really - in the database)

this should be tested

tests/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,266 @@
...
+    assert result == [(1,), (2,)]
+
+    # Loaded state contains only keys created in second run
+    with pipeline.destination_client() as dest_client:
+        destination_state = load_pipeline_state_from_destination(
+            pipeline.pipeline_name, dest_client  # type: ignore[arg-type]
+        )
+    assert destination_state["sources"]["my_source"]["resources"] == {
+        "some_data_1": {"resource_key_3": "resource_value_3"},
+    }
```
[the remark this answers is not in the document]
rudolfix: it should be a separate test

tests/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def test_refresh_drop_tables():
+    first_run = True
+
+    @dlt.source
```
rudolfix, line 106: is this code by any chance fully duplicated?

tests/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,266 @@
...
+    )
+
+    info = pipeline.run(my_source())
+    assert_load_info(info)
+
+    # Second run of pipeline with only selected resources
+    first_run = False
+    info = pipeline.run(my_source().with_resources("some_data_1", "some_data_2"))
+
+    # Confirm resource tables not selected on second run got wiped
```
rudolfix, line 163: w8 - in this mode tables not selected should not be touched. I think that just the comment is outdated

tests/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,266 @@
...
+    with pipeline.sql_client() as client:
+        result = client.execute_sql("SELECT id FROM some_data_3 ORDER BY id")
+    assert result == [(5,), (6,)]
+
+    with pipeline.sql_client() as client:
```
rudolfix, line 168: again: if you do not change the schema in non-backward-compatible way you have no idea if table was truncated or dropped so please mutate the resource schema in some radical way :)

tests/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,266 @@
...
+from dlt.common.pipeline import resource_state
+from dlt.destinations.exceptions import DatabaseUndefinedRelation
+from dlt.destinations.impl.duckdb.sql_client import DuckDbSqlClient
+from dlt.pipeline.state_sync import load_pipeline_state_from_destination
+
+from tests.utils import clean_test_storage, preserve_environ
+from tests.pipeline.utils import assert_load_info
+
+
+def test_refresh_drop_dataset():
```
rudolfix, line 14: I think it is possible to test also filesystem? we have now a client that can read table data. it would be cool to test local filesystem here

rudolfix (changes requested):
it seems that code is complete. there are still a few things to do:
1. there are comments and non executable code left please take look at this PR and clean it
2. we should improve tests, I've left you a few instructions
3. docs are missing

also @sh-rp could you look at pipeline.py and review how we update package state and pass the pipeline state to it. you were the implementer... and we'll have a followup here to fully move state processing to package

dlt/pipeline/helpers.py
```diff
@@ -152,58 +133,6 @@ def is_empty(self) -> bool:
...
-                f"You are dropping table {table_name} in {self.schema.name} but it is still present"
-                " in the schema"
-            )
-        with self.pipeline._sql_job_client(self.schema) as client:
-            client.drop_tables(*table_names, replace_schema=True)
-            # also delete staging but ignore if staging does not exist
-            if isinstance(client, WithStagingDataset):
-                with contextlib.suppress(DatabaseUndefinedRelation):
-                    with client.with_staging_dataset():
-                        client.drop_tables(*table_names, replace_schema=True)
```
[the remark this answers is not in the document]
steinitzu: @rudolfix 
We were deleting all schema versions with `replace_schema` flag before, and I think we need to do that, or at least delete all schemas that contain the tables being dropped, not just the "current version".   
I might be wrong there but think otherwise we can on later run accidentally re-create a schema that's identical to some previous version it will think those table already exist.   

I thought it was kinda drastic to erase history completely like that, but maybe it's a better workaround? Should we work on properly supporting "subtractive" schema migrations? It is a purely additive versioning system basically so it's all a bit hacky imo.

tests/load/pipeline/test_drop.py
```diff
@@ -200,48 +200,48 @@ def test_drop_command_only_tables(destination_config: DestinationTestConfigurati
...
     attached = _attach(pipeline)
 
-    with mock.patch.object(
-        helpers.DropCommand,
-        "_drop_destination_tables",
-        side_effect=RuntimeError("Something went wrong"),
-    ):
-        with pytest.raises(RuntimeError):
-            helpers.drop(attached, resources=("droppable_a", "droppable_b"))
+    # with mock.patch.object(
```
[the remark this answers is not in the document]
steinitzu: Was supposed to be temporary. Should now mock destination client `drop_tables` to fail instead.

tests/load/pipeline/test_drop.py
```diff
@@ -200,48 +200,48 @@ def test_drop_command_only_tables(destination_config: DestinationTestConfigurati
...
 
-@pytest.mark.parametrize(
-    "destination_config", destinations_configs(default_sql_configs=True), ids=lambda x: x.name
-)
-def test_fail_after_drop_tables(destination_config: DestinationTestConfiguration) -> None:
-    """Fail directly after drop tables. Command runs again ignoring destination tables missing."""
-    source = droppable_source()
-    pipeline = destination_config.setup_pipeline("drop_test_" + uniq_id(), full_refresh=True)
-    pipeline.run(source)
+# @pytest.mark.parametrize(
```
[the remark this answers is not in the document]
steinitzu: Same here, yeah cause it runs in load. Letting `update_stored_schema` fail should be equivalent in this case

tests/pipeline/test_refresh_modes.py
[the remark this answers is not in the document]
steinitzu: Added test for this

tests/pipeline/test_refresh_modes.py
steinitzu: Pretty much. Cleaned this up a bit, the source is defined only once now

tests/pipeline/test_refresh_modes.py
steinitzu: I updated the tests and check this by removing a column in the second run data. So when truncated it has `null` values, and when dropped it's not in the schema, i.e.:

```python
        if first_run:
            yield {"id": 5, "name": "Joe"}
            yield {"id": 6, "name": "Jill"}
        else:
            yield {"id": 7}
            yield {"id": 8}
```

tests/load/pipeline/test_drop.py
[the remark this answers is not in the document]
steinitzu: Updated both drop cmd tests

dlt/pipeline/helpers.py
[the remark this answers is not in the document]
rudolfix: OK you are right. still I prefer to drop all schemas (with given name!). we just need to document this and maybe use :::caution::: for that

dlt/pipeline/__init__.py
```diff
@@ -72,6 +72,8 @@ def pipeline(
         dev_mode (bool, optional): When set to True, each instance of the pipeline with the `pipeline_name` starts from scratch when run and loads the data to a separate dataset.
         The datasets are identified by `dataset_name_` + datetime suffix. Use this setting whenever you experiment with your data to be sure you start fresh on each run. Defaults to False.
 
+        refresh (str | TRefreshMode): One of `drop_dataset`, `drop_tables` or `drop_data`. Set this to fully or partially delete and reset the schema, state and destination dataset when running the pipeline.
```
rudolfix, line 75: please be more specific :)
- drop dataset: - drops all tables for any source that will processed in `run` or `extract` methods. also drops the source state. also warn that schema history is erased
- drop_tables - in similar manner. . also warn that schema history is erased
user should know what will happen when given options is specified

docs/website/docs/general-usage/pipeline.md
```diff
@@ -92,6 +92,28 @@ experimenting a lot. If you want that each time the pipeline resets its state an
...
+You can use the `refresh` argument of `dlt.pipeline` to fully or partially reset the destination
+dataset and pipeline state when running the pipeline. This is useful if you need to reload all
+data from the beginning or remove/re-create tables in the destination.
+
+This involves dropping or wiping data from tables and fully or partially wiping the pipeline state.
+
+There are three possible refresh modes that can be used. The `refresh` argument should have one of these string values:
+
+* `drop_dataset`
+  All tables listed in the pipeline's schema will be dropped and all source state is wiped.
```
rudolfix, line 106: we drop only the tables for schemas actually processed right? so if we have two schemas (two different sources) in the pipeline and we use drop and load only source_2, only that source will be wiped? if so we should add a test for that

heh maybe we should rename it to `drop_schemas` or `drop_sources`?

rudolfix (changes requested):
OK! now tests are almost in order! still to do:
1. I have a strong opinion to rename our modes  to `drop_sources`, `drop_resources` and `drop_data`. WDYT? this corresponds more to what happens
2. is it much work to add `refresh` to `run` and `extract` methods of `Pipeline`?
3. we need to improve docstrings and docs
4. let's remove all schemas for a given name from a destination. Old flag was not that bad IMO - it leaves all details to the implementer
5. a simple test for local filesystem - it really makes sense when I look how many users use that destination

docs/website/docs/general-usage/pipeline.md
steinitzu: This was in fact not being done. It was wiping al source state.
Fixed and tested now. `drop_sources` is good naming imo

steinitzu:
> OK! now tests are almost in order! still to do:
> 
>     1. I have a strong opinion to rename our modes  to `drop_sources`, `drop_resources` and `drop_data`. WDYT? this corresponds more to what happens
> 
>     2. is it much work to add `refresh` to `run` and `extract` methods of `Pipeline`?
> 
>     3. we need to improve docstrings and docs
> 
>     4. let's remove all schemas for a given name from a destination. Old flag was not that bad IMO - it leaves all details to the implementer
> 
>     5. a simple test for local filesystem - it really makes sense when I look how many users use that destination

1. Yes. I think that naming is much better. I changed it.
2. Added this and tests for each. Nice to have this.
3. I updated both a bit. Let me know if anything missing.
4. Doing this again now, but I renamed the `drop_tables` `replace_schema` argument to `delete_schema` and changed the behaviour.  
    This is simpler for initializing client. No need to do the dance with swapping out client schema, we just  drop tables and delete schema history in one transaction -> then "update stored schema" runs as normal after.
5. We don't support `drop_tables` in filesystem. Should we add it? I guess it would be same as truncate basically (maybe delete the folders also?). But I made a test for `refresh=drop_data`

dlt/pipeline/helpers.py
```diff
@@ -1,5 +1,17 @@
...
+    Iterable,
+    Optional,
+    Any,
+    List,
+    Dict,
+    Union,
+    TypedDict,
+    TYPE_CHECKING,
+)
 from itertools import chain
```
rudolfix, line 15: 80% of the dependencies are not needed. you could clean them up

rudolfix (commented):
LGTM! just a few small things are left
regarding the filesystem: it has `truncate_tables` method which is equivalent to `drop_tables`. IMO you could just add `drop_tables` to it, call `truncate_tables`. dropping schemas is easy: just remove the directory:
```
self.get_table_dir(self.schema.version_table_name)
```

@sh-rp please review https://github.com/dlt-hub/dlt/pull/1063#discussion_r1572451165

this is big PR... we'll make sure to split those into smaller ones

docs/website/docs/dlt-ecosystem/destinations/duckdb.md
```diff
@@ -184,3 +184,12 @@ This destination fully supports [dlt state sync](../../general-usage/state#synci
 
 <!--@@@DLT_TUBA duckdb-->
 
+=======
```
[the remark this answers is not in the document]
sh-rp: how did this even get here??

dlt/pipeline/__init__.py
```diff
@@ -99,6 +107,8 @@ def pipeline(
     import_schema_path: str = None,
     export_schema_path: str = None,
     full_refresh: bool = False,
```
sh-rp, line 109: Imho we should set full_refresh to "None" and issue a warning if it set to True or False explicitely. So:
* If full_refresh is None, take the value of dev_mode
* If full_refresh is not None, issue warning and set dev_mode to full_refresh

sh-rp: @rudolfix wdyt

tests/load/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,484 @@
...
+    pipeline.extract(refresh_source(first_run=False).with_resources("some_data_2"))
+
+    tables = set(t["name"] for t in pipeline.default_schema.data_tables(include_incomplete=True))
+    assert tables == {"some_data_2", "some_data_3", "some_data_4"}
+
+
+@pytest.mark.parametrize(
+    "destination_config", destinations_configs(local_filesystem_configs=True), ids=lambda x: x.name
+)
+def test_refresh_drop_sources_local_filesystem(destination_config: DestinationTestConfiguration):
```
sh-rp, line 454: imho this does not need to be tested in a special test for local filesystem. we have a bunch of utils where you can query the items for each table regardless of wether the destination is a filesystem or a sql db. or is there something here that will not work this way?

docs/website/docs/dlt-ecosystem/destinations/duckdb.md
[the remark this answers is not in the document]
steinitzu: Ok wow I don't know how I did that :confused:  
How is this generated? Trying to figure out if I ran something by accident.

tests/load/pipeline/test_refresh_modes.py
sh-rp: Generally speaking it would be cool to have some kind of unified interface to query all items from a table that needs to be implemented for all destinations so we do not need these extra tests for filesystem that are still around in a couple of places. But that is for another PR.

tests/load/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,484 @@
...
+
+    yield some_data_1
+    yield some_data_2
+    yield some_data_3
+    yield some_data_4
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["duckdb"]),
```
sh-rp, line 99: why only duckdb?

sh-rp: as mentioned below, you should be able to test the filesystems with the same test

tests/load/pipeline/test_refresh_modes.py
```diff
@@ -0,0 +1,484 @@
...
+    with pipeline.destination_client() as dest_client:
+        destination_state = load_pipeline_state_from_destination(
+            pipeline.pipeline_name, dest_client  # type: ignore[arg-type]
+        )
+    assert_source_state_is_wiped(destination_state["sources"]["refresh_source"])
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["duckdb"]),
```
sh-rp, line 147: same questions as above

sh-rp (changes requested):
I will have another closer look tomorrow as this is really a lot of stuff going on here. For now I would suggest merging filesystem and duckdb tests, I think it is possible with our helpers and it will reduce complexity. Also @rudolfix is testing on duckdb enough?

dlt/pipeline/__init__.py
steinitzu: That's a good approach. Distinguish between `False` and "not used"

tests/load/pipeline/test_refresh_modes.py
steinitzu: I managed to do this mostly with the `load_tables_to_dicts` utils. All tests run on filesystem now!

tests/load/pipeline/test_refresh_modes.py
steinitzu: Only variance could be the implementation of "drop/truncate tables" methods, but they're already tested elsewhere. So didn't think it would be necessary to run other destinations here. Just going by the assumption that all sql destinations have their interface tested.    
But maybe doesn't hurt to be thorough if something changes with the implementation in the future.

rudolfix:
@sh-rp thanks for review. the part where I asked you to look deeper is how we pass and write the package state and if the pipeline state is handled correctly. I do not see any updates on that. so I assume this is mostly right?

@steinitzu plenty of tests are failing. my take is that you should try to run common test locally so we have less turnover on CI (if you do not do that already)

steinitzu:
> @steinitzu plenty of tests are failing. my take is that you should try to run common test locally so we have less turnover on CI (if you do not do that already)

@rudolfix got it. I don't always run them but should for sure before pushing. Usually I try to run the tests I think are relevant, in this case lot more than I thought. 

Common + load postgres/duck/local filesystem tests are passing locally now

rudolfix (approved):
LGTM!

sh-rp:
The state stuff looks good to me too. The way it is loaded to the destination has not been changed here yet, this is for an upcoming PR (correct me if I'm wrong there @steinitzu), only those new fields have been added and the state is committed after extraction, so I don't see a problem, or any migration stuff we need to take care of.

PR #1065, opened 2024-03-07
Generic destination / sink decorator

sh-rp:
### Description
Implementation of https://github.com/dlt-hub/dlt/issues/752

This is a re-submit of PR https://github.com/dlt-hub/dlt/pull/891 which I cannot reopen after an accidental merge 

Notes:
* We also introduce the concept of a load package state here

ToDos Alpha release:
- [x] Tests for locked injection context
- [x] Tests for resolved partial
- [ ] Tests for migration from no load package state to having load package state

Follow up tasks:
- [ ] Nice example of loading to bigquery or kafka with the generic destination
- [ ] Update CLI scaffolding scripts to create a sink function if "destination" is selected as Destination


sh-rp:
@rudolfix current questions:
* I have create a method to create a partial from a function decorated with "with_config", can you let me know if this looks good? Then I can add some tests
* Can you show me how an example of a sink function with google cloud credentials should look like? Do you want to send the spec as one arg or what is the interface for the user you have in mind? Then I will better understand what you have in mind

Generally you can have a look at my changes, I did most of the stuff of your last review.

dlt/common/configuration/inject.py
```diff
@@ -167,12 +172,48 @@ def _wrap(*args: Any, **kwargs: Any) -> Any:
...
+
+                # call the function with the pre-resolved config
+                bound_args = sig.bind(*args, **kwargs)
+                update_bound_args(bound_args, config, args, kwargs)
+                return f(*bound_args.args, **bound_args.kwargs)
+
+            return creator
+
+        @wraps(f)
+        def _wrap(*args: Any, **kwargs: Any) -> Any:
```
rudolfix, line 198: 1. we must sure that config is resolved at run time (when the inner function is actually called)
2. we do not want to cache here in interactive mode

dlt/common/reflection/spec.py
```diff
@@ -26,7 +26,10 @@ def _first_up(s: str) -> str:
 
 
 def spec_from_signature(
-    f: AnyFun, sig: Signature, include_defaults: bool = True
+    f: AnyFun,
+    sig: Signature,
+    include_defaults: bool = True,
+    base: Type[BaseConfiguration] = BaseConfiguration,
```
rudolfix, line 32: this is so good

dlt/common/storages/load_package.py
```diff
@@ -17,23 +19,92 @@
...
+
+def generate_loadpackage_state_version_hash(state: TLoadPackageState) -> str:
+    return generate_state_version_hash(state)
+
+
+def bump_loadpackage_state_version_if_modified(state: TLoadPackageState) -> Tuple[int, str, str]:
+    return bump_state_version_if_modified(state)
+
+
+def migrate_loadpackage_state(
```
rudolfix, line 81: load_package? we do snake case

dlt/common/storages/load_package.py
```diff
@@ -334,8 +408,13 @@ def create_package(self, load_id: str) -> None:
         self.storage.create_folder(os.path.join(load_id, PackageStorage.COMPLETED_JOBS_FOLDER))
         self.storage.create_folder(os.path.join(load_id, PackageStorage.FAILED_JOBS_FOLDER))
         self.storage.create_folder(os.path.join(load_id, PackageStorage.STARTED_JOBS_FOLDER))
+        # ensure created timestamp is set in state when load package is created
+        state = self.get_load_package_state(load_id)
```
rudolfix, line 412: we must make this change backward compatible or explicitly upgrade all storages that have load packages internally,

see this test `test_pipeline_with_dlt_update` and similar: they actually upgrade dlt from 0.3 to 0.4 and test packages in process

my worry: people will have old completed load packages or packages in process, they upgrade dlt and it stops working

so you need to test it.

dlt/common/storages/load_package.py
```diff
@@ -514,3 +621,59 @@ def filter_jobs_for_table(
...
     ) -> Sequence[LoadJobInfo]:
         return [job for job in all_jobs if job.job_file_info.table_name == table_name]
+
+
+@configspec
+class LoadPackageStateInjectableContext(ContainerInjectableContext):
+    storage: PackageStorage
+    load_id: str
+    can_create_default: ClassVar[bool] = False
+    global_affinity: ClassVar[bool] = True
```
rudolfix, line 631: it should be false, no? you can have many pipelines running at once and each of them will maintain separate context. the only context that has global affitity are config providers

dlt/common/storages/load_package.py
```diff
@@ -514,3 +621,59 @@ def filter_jobs_for_table(
...
+
+def load_package() -> TLoadPackage:
+    """Get full load package state present in current context. Across all threads this will be the same in memory dict."""
+    container = Container()
+    # get injected state if present. injected load package state is typically "managed" so changes will be persisted
+    # if you need to save the load package state during a load, you need to call commit_load_package_state
+    try:
+        state_ctx = container[LoadPackageStateInjectableContext]
+    except ContextDefaultCannotBeCreated:
+        raise Exception("Load package state not available")
```
rudolfix, line 654: Please derive exception properly. see other similar exceptions.
provide explanation why this is happening. ie when you access this outside of `dlt.destination` decorated function (we inject package only in loader, btw. we can start injecting it in normalize and extract as well)

dlt/destinations/decorators.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def destination(
+    loader_file_format: TLoaderFileFormat = None,
+    batch_size: int = 10,
+    name: str = None,
+    naming_convention: str = "direct",
+    spec: Type[SinkClientConfiguration] = SinkClientConfiguration,
+) -> Any:
+    def decorator(destination_callable: TSinkCallable) -> TDestinationReferenceArg:
```
rudolfix, line 15: Decorator returns `Destination` but the user should see the signature of the decorated function as return type, We''ll add `__call__` to Destination to simulate it being a function. this is same trick we do with DltResource

I can do the above if you want

dlt/destinations/impl/destination/factory.py
```diff
@@ -0,0 +1,114 @@
...
+
+    @property
+    def client_class(self) -> t.Type["SinkClient"]:
+        from dlt.destinations.impl.destination.sink import SinkClient
+
+        return SinkClient
+
+    def __init__(
+        self,
+        destination_callable: t.Union[TSinkCallable, str] = None,  # noqa: A003
```
rudolfix, line 56: we can still configure this right?

dlt/destinations/impl/destination/factory.py
```diff
@@ -0,0 +1,114 @@
...
+        naming_convention: str = "direct",
+        spec: t.Type[SinkClientConfiguration] = SinkClientConfiguration,
+        **kwargs: t.Any,
+    ) -> None:
+        # resolve callable
+        if callable(destination_callable):
+            pass
+        elif destination_callable:
+            try:
+                module_path, attr_name = destination_callable.rsplit(".", 1)
```
rudolfix, line 70: do you allow to configure this via module.function_name?

dlt/load/load.py
```diff
@@ -452,12 +463,19 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
...
 
+        container = Container()
         # get top load id and mark as being processed
         with self.collector(f"Load {schema.name} in {load_id}"):
-            # the same load id may be processed across multiple runs
-            if not self.current_load_id:
-                self._step_info_start_load_id(load_id)
-            self.load_single_package(load_id, schema)
+            with container.injectable_context(
+                LoadPackageStateInjectableContext(
```
rudolfix, line 470: I have no idea how this is supposed to work in global context...

rudolfix (changes requested):
pls see review. I'll have more comments but let's fix those first. overall this is so good

dlt/destinations/impl/destination/factory.py
sh-rp: yes, see tests

dlt/destinations/impl/destination/factory.py
sh-rp: as in give a function name / module path in a config bar? then yes, see the tests.

dlt/destinations/decorators.py
sh-rp: Ah, now I think I get what you want here. Yes why not, maybe you can add that one small change

dlt/common/storages/load_package.py
sh-rp: - This is a left over of yesterdays battle i think :D

dlt/common/configuration/inject.py
sh-rp: 1. This is the case
2. I'm not sure what interactive mode is exactly?

dlt/destinations/decorators.py
sh-rp: Actually I think to me it is not quite clear what i should do here, i changed the decorator typings a bit and made it so a wrapped function is returned with the args in place (i think at least) but maybe you need to make some changes here.

dlt/common/configuration/inject.py
sh-rp: In any case, in the tests you can see that you can create a pipeline, then define env vars and run the pipeline without failure.

dlt/common/configuration/inject.py
```diff
@@ -88,15 +97,16 @@ def decorator(f: TFun) -> TFun:
...
         if spec is None:
-            SPEC = spec_from_signature(f, sig, include_defaults)
+            SPEC = spec_from_signature(f, sig, include_defaults, base=base)
         else:
             SPEC = spec
 
         if SPEC is None:
             return f
 
+        func_id = id(f)
```
sh-rp, line 108: we use the id of the wrapped function as part of the key to created the threadlock, if the same function is wrapped multiple times, then the key will collide, so we could also consider a uniqid

dlt/common/configuration/inject.py
```diff
@@ -58,6 +65,7 @@ def with_config(
     include_defaults: bool = True,
     accept_partial: bool = False,
     initial_config: Optional[BaseConfiguration] = None,
```
sh-rp, line 67: should we have a "thread_safe" parameter so we only lock on decorated functions where we know that there might be issues?

rudolfix (changes requested):
OK we are really close here! changes seem to be easy to make. test coverage is good. my take:
implement two tests that you mention
- we should add tests for the inject context with lock (when you fix the code) where we makes sure two threads are getting into critical section at the same time
- that we actually resolve config once in `create_resolved_partial` (and possibly can pass external one - see my review)
- test for __callable__ destination factory
after that we can merge and do alpha release. a dlt migration test for load packages is a must before stable release

dlt/common/configuration/inject.py
```diff
@@ -167,12 +180,55 @@ def _wrap(*args: Any, **kwargs: Any) -> Any:
...
+        def create_resolved_partial() -> Any:
+            # creates a pre-resolved partial of the decorated function
+            empty_bound_args = sig.bind_partial()
+            config = resolve_config(empty_bound_args)
+
+            def wrapped(*args: Any, **kwargs: Any) -> Any:
+                nonlocal config
+
+                # Do we need an exception here?
+                if spec_arg and spec_arg.name in kwargs:
```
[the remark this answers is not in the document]
sh-rp: Yeah, but in the case of partially resolving, the resolve_config is called before even the func args are available, do you know what i mean? It's a bit complicated but also quite clear. We can also remove it if you like, it's just a warning that made sense to put in given the mechanics of the code.

dlt/destinations/impl/destination/sink.py
```diff
@@ -0,0 +1,185 @@
...
+class SinkClient(JobClientBase):
+    """Sink Client"""
+
+    capabilities: ClassVar[DestinationCapabilitiesContext] = capabilities()
+
+    def __init__(self, schema: Schema, config: SinkClientConfiguration) -> None:
+        super().__init__(schema, config)
+        self.config: SinkClientConfiguration = config
+        # create pre-resolved callable to avoid multiple config resolutions during execution of the jobs
+        self.destination_callable = create_resolved_partial(self.config.destination_callable)
```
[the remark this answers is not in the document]
sh-rp: this works, nice!

dlt/destinations/impl/destination/factory.py
```diff
@@ -0,0 +1,114 @@
...
+            self.config_params.get("naming_convention", "direct"),
+        )
+
+    @property
+    def spec(self) -> t.Type[SinkClientConfiguration]:
+        """A spec of destination configuration resolved from the sink function signature"""
+        return self._spec
+
+    @property
+    def client_class(self) -> t.Type["SinkClient"]:
```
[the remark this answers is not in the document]
sh-rp: I did not even need to implement the call function, just forward the kwargs to the destination and then use the pre-resolved config in the create resolved partial function (see the code, I have a test for this too). Very nice!

tests/load/sink/test_sink.py
```diff
@@ -0,0 +1,428 @@
...
+        yield items
+
+    p = dlt.pipeline("sink_test", destination=test_sink, full_refresh=True)
+    p.run([items_resource()])
+
+    return calls
+
+
+@pytest.mark.parametrize("loader_file_format", SUPPORTED_LOADER_FORMATS)
+def test_all_datatypes(loader_file_format: TLoaderFileFormat) -> None:
```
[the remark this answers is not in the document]
sh-rp: I removed the dependency on google stuff and moved them into the commons test area

tests/destinations/test_generic_destination.py
```diff
@@ -0,0 +1,435 @@
...
+    print(my_sink)
+
+    # if no value is present, it should raise
+    with pytest.raises(ConfigFieldMissingException):
+        dlt.pipeline("sink_test", destination=my_sink, full_refresh=True).run(
+            [1, 2, 3], table_name="items"
+        )
+
+    # we may give the value via __callable__ function
+    dlt.pipeline("sink_test", destination=my_sink(my_val="something"), full_refresh=True).run(
```
sh-rp, line 375: fyi: callable my_sink is tested here

rudolfix (approved):
LGTM!

I'm not entirely sure that calling former sinks GenericDestination is what we want. maybe DecoratedDestination? anyway! let's merge this and write a followup ticket

PR #1072, opened 2024-03-11
docs(airflow): add description of new decompose methods

### LINKED ISSUE #931: Airflow helper: expose internal runner and introduce parallel execution of resources

IlyaFaer:
Closes #931

### DIFF
--- a/docs/website/docs/walkthroughs/deploy-a-pipeline/deploy-with-airflow-composer.md
+++ b/docs/website/docs/walkthroughs/deploy-a-pipeline/deploy-with-airflow-composer.md
@@ -155,8 +155,41 @@ def load_data():
     )
     # Create the source, the "serialize" decompose option
     # will convert dlt resources into Airflow tasks.
-    # Use "none" to disable it
-    tasks.add_run(pipeline, source(), decompose="serialize", trigger_rule="all_done", retries=0, provide_context=True)
+    # Use "none" to disable it.
+    tasks.add_run(
+        pipeline,
+        source(),
+        decompose="serialize",
+        trigger_rule="all_done",
+        retries=0,
+        provide_context=True
+    )
+
+    # The "parallel" decompose option will convert dlt
+    # resources into parallel Airflow tasks, except the
+    # first one, which will be executed before any other tasks.
+    # All the tasks will be executed in the same pipeline state.
+    # tasks.add_run(
+    #   pipeline,
+    #   source(),
+    #   decompose="parallel",
+    #   trigger_rule="all_done",
+    #   retries=0,
+    #   provide_context=True
+    # )
+
+    # The "parallel-isolated" decompose option will convert dlt
+    # resources into parallel Airflow tasks, except the
+    # first one, which will be executed before any other tasks.
+    # In this mode, all the tasks will use separate pipeline states.
+    # tasks.add_run(
+    #   pipeline,
+    #   source(),
+    #   decompose="parallel-isolated",
+    #   trigger_rule="all_done",
+    #   retries=0,
+    #   provide_context=True
+    # )
 
 load_data()
 ```


IlyaFaer:
There are already deployment docs for Airflow, so I don't think it makes much sense to repeat it. The task was to describe the new parallel decomposition methods, which can be added into the existing docs, to the related section.

rudolfix:
@AstrakhantsevaAA @burnash should we add a general information on Airflow helper somewhere or we are fine with this. I'm keen on having a place for helpers documentation

AstrakhantsevaAA:
@rudolfix don't forget about API reference. We have not bad docs for helpers. IMO we should extend our docstrings and clean up our reference API list. 

image

AstrakhantsevaAA (approved):

PR #1073, opened 2024-03-11
Adding the proposed segment migration blog

zem360:
Adding the segment migration blog.


AstrakhantsevaAA (approved):
Love it! Amazing work Zaeem!
one small moment (probably for future): usually we lowercase dlt and write it as code `dlt`.

Issue #407, opened 2023-06-14
Airflow: DuplicateTaskIdFound


AstrakhantsevaAA:
In case of `pipedrive` pipeline we tried to load data from “custom_fields_mapping” twice.

```python
# First source configure to load everything except activities from the beginning
source = pipedrive_source()
source.resources["activities"].selected = False

# Another source configured to activities starting at the given date (custom_fields_mapping is included to translate custom field hashes to names)
activities_source = pipedrive_source(
    since_timestamp="2023-03-01 00:00:00Z"
).with_resources("activities", "custom_fields_mapping")
```

Because of this we got the following error:

```python
airflow.exceptions.DuplicateTaskIdFound: 
Task id ‘pipedrive.pipedrive_custom_fields_mapping’ has already been added to the DAG
```

Task ids in the task group should be still unique globally

dat-a-man:
Hey there! Just wanted to drop a quick comment to see if the Github pipeline is working properly. I made some modifications and I want to make sure that the reactions on comments are loading correctly. Please don't mind this comment - it's just a test. Thanks!

IlyaFaer:
Can't reproduce it. Is there kinda Airflow configurations that must be set to repeat the error?

rudolfix:
@IlyaFaer I think the error happens when you pass two instances of a source that have identical resources. it is a valid use case ie. you can have a source configured to backfil a table and the other one to just do incremental load. try to reproduce this error by creating a simple source with one resource and pass two instances of it to the runner

IlyaFaer:
@rudolfix, okay, I reproduced it 👌

PR #1080, opened 2024-03-12
feat(airflow): allow re-using sources in airflow wrapper

IlyaFaer:
Fixes https://github.com/dlt-hub/dlt/issues/407

### DIFF
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -1,7 +1,7 @@
 import functools
 import os
 from tempfile import gettempdir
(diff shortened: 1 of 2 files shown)


IlyaFaer:
@rudolfix, as I understand this names generation method has a particular purpose, and you don't want to change it.

So, I did this: if a task name is already used in this group, we simply add a number of the task to the id. Thus, the first task will have the original name, and the next ones with the same name will be numbered.

rudolfix (approved):
LGTM! and quite elegant :)

PR #1082, opened 2024-03-12
validates class instances in typed dict

rudolfix:
### Description
Allows classes in typed dict and validates class instances, also respect unions


burnash (approved):
LGTM

PR #1084, opened 2024-03-12
[do not merge] Make snippets great again

sh-rp:
### Description
This is the first attempt at making our snippet system more accessible, further description to follow.

### DIFF
--- a/docs/snippets/performance_chunking.py
+++ b/docs/snippets/performance_chunking.py
@@ -1,0 +1,18 @@
+"""
+description: Chunking example
+tags: performance
+"""
+
+import dlt
+
+if __name__ == "__main__":
+
+    def get_rows(limit):
+        yield from map(lambda n: {"row": n}, range(limit))
+
+    @dlt.resource
+    def database_cursor():
+        # here we yield each row returned from database separately
+        yield from get_rows(10000)
+
+    assert len(list(database_cursor())) == 10000
--- a/docs/website/check_snippets.py
+++ b/docs/website/check_snippets.py
@@ -1,0 +1,75 @@
+"""
+Load all snippets in snippet folder, check wether they parse with ast and run them
+"""
+import typing as t
+import os
+import ast
+import sys
+
+SNIPPET_DIR = "../snippets"
+
+def get_snippet_list() -> t.List[str]:
+    """Get list of available snippets in the snippet folder."""
+    return [s.replace(".py", "") for s in os.listdir(SNIPPET_DIR) if s.endswith(".py") and s != "__init__.py"]
+
+def get_snippet(snippet_name: str) -> str:
+    """Get the content of a snippet."""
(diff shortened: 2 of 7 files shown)


docs/snippets/performance_chunking.py
```diff
@@ -0,0 +1,18 @@
+"""
+description: Chunking example
+tags: performance
+"""
+
+import dlt
+
+if __name__ == "__main__":
```
sh-rp, line 8: when using this snippet in the docs, only the code inside the main clause is used and any asserts are stripped (we can add an additional marker to have certain lines removed, but possibly this is not necessary). Also you can decide which imports show up in the snippet by either placing them in the head or in the main clause.

VioletM:
Maybe we can have the same structure for snippets as for docs? To be able to find them easily
image

Or we can put snippets folder to each docs directory and gather them from all docs section. 
* It's easier when you place a new snippet on the doc page, because you will have your code right next to the docs.
* But it's harder to collect snippets for "snippets gallery"

docs/website/check_snippets.py
```diff
@@ -0,0 +1,75 @@
...
+            failed_parsing.append(snippet)
+            continue
+
+        # snippet needs to be run in main function for some reason
+        if run_snippet(snippet) is False:
+            failed_running.append(snippet)
+
+    print()
+    if failed_parsing or failed_running:
+        print(f"\033[91m{len(failed_parsing)} snippets failed to parse, {len(failed_running)} snippets failed to run")
```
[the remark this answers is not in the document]
sh-rp: Yeah, for each PR all snippets are run, and if they fail the PR is red. So don't need to check on docs build, as only PRs that the normal snippets tests will be merged.

sh-rp:
@VioletM we can use folders as categories, but we would need to decide how to categorize the snippets. I think the snippets categories might be different from the docs ones.

docs/website/check_snippets.py
```diff
@@ -0,0 +1,75 @@
...
+Load all snippets in snippet folder, check wether they parse with ast and run them
+"""
+import typing as t
+import os
+import ast
+import sys
+
+SNIPPET_DIR = "../snippets"
+
+def get_snippet_list() -> t.List[str]:
```
rudolfix, line 11: this must be run from `pytest`. all fixtures to drop pipelines, change the segment key, set the path to credentials should be preserved

rudolfix: generate a test for each script. it is I think possible to register tests at run time

rudolfix (changes requested):

docs/website/check_snippets.py
sh-rp: hm, good idea, i will see if i can do it

sh-rp:
closing for now

PR #1093, opened 2024-03-14
check embedded code blocks

sh-rp:
### Description
This PR will add a new short python script that will find all embedded code blocks in our docs and parse them to see if there are any very obvious errors. We may even be able to lint the python ones, but this is unclear as of now. 

What it does now:

- Forces you to mark a code block with the type of language, if it is not a language, setting it to text is allowed
- Will parse python, toml, yaml and json blocks and fail if they are not parseable


rudolfix (approved):
LGTM!
@sh-rp but `npm run start` still runs some python right? didn't we plan to decouple this?

sh-rp:
> LGTM! @sh-rp but `npm run start` still runs some python right? didn't we plan to decouple this?
@rudolfix yes, I want to do these things step by step. In the next pr i plan to remove python stuff from the npm commands (except for the main build command) and remove duplicated snippet code, but for the latter I still have to do some tests.

PR #1097, opened 2024-03-16
Clickhouse 

### LINKED ISSUE #1055: Implement Clickhouse Destination
### Objective

Implement Clickhouse as a destination in the dlt library.

### Features:

1. **Default Clickhouse configuration via dlt resources adapter:**

   - Provide a way to configure the Clickhouse connection details using dlt's configuration system using the [clickhouse-connect](https://clickhouse.com/docs/en/integrations/python) driver. (`clickhouse_connect.driver`)
   - Allow users to specify additional connection-specific settings if needed. 
   
2. **Schema mapping and data type compatibility:**

   - Map dlt's schema to the appropriate Clickhouse data types.
   - The sqlalchemy dialect contained in (`clickhouse_connect.cc_sqlalchemy`) could be used to define these mappings. The docs mention that it is limited though.
   - Handle necessary data type conversions to ensure compatibility between dlt's data and Clickhouse's supported types.
   - Handle Clickhouse's definition of primary keys.
   
3. **Efficient data loading:**
   - Utilize Clickhouse's bulk loading capabilities to optimize data ingestion performance.
   - Explore options for parallel data loading to leverage Clickhouse's distributed architecture.
   
4. **Support for Clickhouse-specific features:**

   - Provide support for Clickhouse-specific features that can enhance data loading and querying, such as:
     - Partitioning and sharding strategies
     - Materialized views
     - Aggregation and compression settings
     
5. **SQL client integration:**
   - Ensure integration with dlt's SQL client to enable users to query and interact with the loaded data in Clickhouse.

### Development Tasks:

- [ ] Define the interface and configuration options for the Clickhouse destination.
- [ ] Implement the core functionality for schema mapping, data loading, and destination-specific features.
- [ ] Write unit and integration tests to ensure the correctness and reliability of the Clickhouse destination in `test/load/pipelines`.
- [ ] Update documentation and provide examples showcasing the usage of the Clickhouse destination in dlt pipelines.

### Requested Community Feedback

We would appreciate any feedback or suggestions from the community regarding:

- Clickhouse features or optimizations that would be valuable to support in the dlt destination.
- Use cases or scenarios where a Clickhouse destination would be beneficial.
- Any potential challenges or considerations to keep in mind when implementing the Clickhouse destination.

Your input will help us prioritize the development efforts and ensure that the Clickhouse destination meets the needs of the dlt community. Thank you for your feedback.

Pipboyguy:
## Description

Implements a basic ClickHouse destination target, along with an adapter to specify table engine type.

## Related Issues

- Resolves #1055 #

## Additional Context

This PR only implements the `MergeTree` and `ReplicatedMergeTree` engines from the `MergeTree` family.

### Features NOT Implemented in this PR

- Any table engine not specified in the mentioned table engines.
-  **Custom** sorting keys (`ORDER BY` when creating tables) - Only useful when using the [SummingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/summingmergetree) and [AggregatingMergeTree](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/aggregatingmergetree) table engines. See [MergeTree family docs](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree#choosing-a-primary-key-that-differs-from-the-sorting-key). Note that sorting key, by default, is implicitly equal to the primary anyway.
- Clustering (future work for adapter).
- Partitioning (future work for adapter).
- `SETTINGS` clause (future work for adapter).
- Taking advantage of multiple replicas: table functions like [s3Cluster](https://clickhouse.com/docs/en/sql-reference/table-functions/s3Cluster) take advantage of multiple replicas with parallel inserts. Managed services like Clickhouse Cloud will most benefit from this. 

### Things to keep in mind for future work
- Not all table engines support a [primary index](
https://clickhouse.com/docs/en/sql-reference/statements/create/table#with-explicit-schema).
- Clickhouse has no unique constraint support.

## TODO

- [ ] Tests
- [x] Documentation
- [ ] `Settings` Options
- [ ] Compression codecs


Pipboyguy:
```sql
ALTER TABLE `default`.`event_test_table`
ADD COLUMN `col` JSON;
```
throws:

```console
 Code: 344. DB::Exception: Adding a new column of a type which has dynamic subcolumns to an existing table is not allowed. It has known bugs. (SUPPORT_IS_DISABLED) (version 24.3.1.1699 (official build))
 ```
 
 This is unfortunate because the `JSON` type lends some nice operations on dict types. We'll have to stick to `String` type it seems.

**Comment by Pipboyguy** [comment]
@rudolfix I'm getting somewhat ambiguous excpetions for `tests/load/pipeline/test_clickhouse.py`

```
dlt/pipeline/pipeline.py:541: PipelineStepFailed
exceptions must derive from BaseException
exceptions must derive from BaseException

Process finished with exit code 1
```

Not sure whether the sqlclient's Error definitions has somethign to do with this

File: dlt/destinations/impl/clickhouse/clickhouse.py [hunk 1]
```diff
@@ -0,0 +1,277 @@
...
+        bucket_url = urlparse(bucket_path)
+        bucket_scheme = bucket_url.scheme
+
+        table_function: str
+
+        if bucket_scheme in ("s3", "gs", "gcs"):
+            bucket_http_url = convert_storage_to_http_scheme(bucket_url)
+
+            table_function = (
+                render_s3_table_function(
```
sh-rp, line 139: so this s3 rendering also works for gcs? maybe then rename the function, but from the code in there I'm not sure it actually works this way?

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,277 @@
...
+                # Authenticated access.
+                account_name = staging_credentials.azure_storage_account_name
+                storage_account_url = (
+                    f"{staging_credentials.azure_storage_account_name}.blob.core.windows.net"
+                )
+                account_key = staging_credentials.azure_storage_sas_token
+                container_name = bucket_url.netloc
+                blobpath = bucket_url.path
+
+                format_mapping = {"jsonl": "JSONEachRow", "parquet": "Parquet"}
```
sh-rp, line 162: this mapping is done in two different places, maybe make it more DRY

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,277 @@
...
+
+class ClickhouseClient(SqlJobClientWithStaging, SupportsStagingDestination):
+    capabilities: ClassVar[DestinationCapabilitiesContext] = capabilities()
+
+    def __init__(
+        self,
+        schema: Schema,
+        config: ClickhouseClientConfiguration,
+    ) -> None:
+        # TODO: There are no schemas in Clickhouse. No point in having schemas, only dataset names and table names for example "dataset1_mytable".
```
sh-rp, line 195: I'm not sure what the todo is here, but we have the schema anyway and I don't think it is harmful to have it a as a property on the client. At least this is my take.

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,277 @@
...
+        return super().start_file_load(table, file_path, load_id) or ClickhouseLoadJob(
+            file_path,
+            table["name"],
+            self.sql_client,
+            staging_credentials=(
+                self.config.staging_config.credentials if self.config.staging_config else None
+            ),
+        )
+
+    def _get_table_update_sql(
```
sh-rp, line 214: just for me to understand, clickhouse does not have any schemas at all, you just define the table and an engine and then the schema is whatever is found in the files loaded to that table?

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -0,0 +1,174 @@
...
+
+    @raise_open_connection_error
+    def close_connection(self) -> None:
+        if self._conn:
+            self._conn.close()
+            self._conn = None
+
+    @contextmanager
+    @raise_database_error
+    def begin_transaction(self) -> Iterator[DBTransaction]:
```
sh-rp, line 68: I think we may at some point want to a capabiliy wether a destination supports transactions and then deal with this automatically, but maybe not now.

dlt/destinations/impl/clickhouse/utils.py
```diff
@@ -0,0 +1,64 @@
...
+def render_s3_table_function(
+    url: str,
+    access_key_id: Optional[str] = None,
+    secret_access_key: Optional[str] = None,
+    file_format: Optional[S3_TABLE_FUNCTION_FILE_FORMATS] = "jsonl",
+) -> str:
+    if file_format not in ["parquet", "jsonl"]:
+        raise ValueError("Clickhouse s3/gcs staging only supports 'parquet' and 'jsonl'.")
+
+    format_mapping = {"jsonl": "JSONEachRow", "parquet": "Parquet"}
```
sh-rp, line 52: as mentioned abovel, this mapping exists twice

pyproject.toml
```diff
@@ -145,6 +145,7 @@ google-api-python-client = ">=1.7.11"
 pytest-asyncio = "^0.23.5"
 types-sqlalchemy = "^1.4.53.38"
 ruff = "^0.3.2"
+clickhouse-driver = "^0.2.7"
```
sh-rp, line 148: please create an extra for clickhouse, see how it is done in other destinations e.g. bigquery

sh-rp (changes requested):
I have a few small questions and comments, now we should get this running on ci in the tests

sh-rp:
@Pipboyguy we need to set up testing, for this you need to:
* Create a github workflow for cllickhouse, see the other destinations how to do this, basically you need to set active destinations to clickhouse and install dlt with the clickhouse extra (which you still have to create)
* We need at least one test where we also test loading from parquet
* We need tests for gcs and azure integration

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: From https://clickhouse.com/docs/en/sql-reference/table-functions/gcs:

> This is an alias of the [s3 table function](https://clickhouse.com/docs/en/sql-reference/table-functions/s3).

The s3 table function should work for both **s3** and **gcs**, but not for **az**.  I've renamed the function to `render_object_storage_table_function` for clarity. This function should ideally contain azure logic as well.

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: Fixed in 86685fd684bff29c8c016d9f20bfb4b5d9055e8d

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: You are 100% right. I've opted for using underscore and the separator: `database`.`{schema/dataset}_tablename`. 

https://github.com/dlt-hub/dlt/blob/86685fd684bff29c8c016d9f20bfb4b5d9055e8d/dlt/destinations/impl/clickhouse/sql_client.py#L154

Thoughts? Is this robust enough for a first draft?

dlt/destinations/impl/clickhouse/sql_client.py
Pipboyguy: Good idea. Please advise if you see any way I can make this PR deal better with CH's lack of transactions, if so.

(As a side note, [Experimental support](https://clickhouse.com/docs/en/guides/developer/transactional#enable-experimental-transaction-suppor) is available for single node deployments, mergetree family)

dlt/destinations/impl/clickhouse/utils.py
Pipboyguy: Fixed

pyproject.toml
Pipboyguy: Fixed in 83b38121b6081afd98d96264a2091a612df559c6

Pipboyguy:
@sh-rp 

- Added gcs and azure integration
- The destination only supports parquet for now, all tests pass
- Working on merge functionality and CI

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: Correct, the schema is just the prefix

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,334 @@
...
+            else ""
+        )
+        file_name = (
+            FileStorage.get_file_name_from_file_path(bucket_path) if bucket_path else file_name
+        )
+        file_extension = os.path.splitext(file_name)[1][
+            1:
+        ].lower()  # Remove dot (.) from file extension.
+        if file_extension not in ["parquet", "jsonl"]:
+            raise ValueError("Clickhouse staging only supports 'parquet' and 'jsonl' file formats.")
```
sh-rp, line 138: For now only parquet, no?

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,334 @@
...
+        table_function: str
+
+        if bucket_scheme in ("s3", "gs", "gcs"):
+            bucket_http_url = convert_storage_to_http_scheme(bucket_url)
+
+            if isinstance(staging_credentials, AwsCredentialsWithoutDefaults):
+                access_key_id = staging_credentials.aws_access_key_id
+                secret_access_key = staging_credentials.aws_secret_access_key
+            elif isinstance(staging_credentials, GcpCredentials):
+                # TODO: HMAC keys aren't implemented in `GcpCredentials`.
```
sh-rp, line 157: will you create a follow up ticket for this kind of stuff?

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,334 @@
...
+        if primary_key_list := [
+            self.capabilities.escape_identifier(c["name"])
+            for c in new_columns
+            if c.get("primary_key")
+        ]:
+            sql[0] += "\nPRIMARY KEY (" + ", ".join(primary_key_list) + ")"
+        else:
+            sql[0] += "\nPRIMARY KEY tuple()"
+
+        # TODO: Apply sort order and cluster key hints.
```
sh-rp, line 253: also for follow up ticket

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def open_connection(self) -> clickhouse_driver.dbapi.connection.Connection:
+        self._conn = clickhouse_driver.dbapi.connect(
+            dsn=self.credentials.to_native_representation()
+        )
+        with self._conn.cursor() as cur:
+            # Set session settings. There doesn't seem to be a way to set these
+            # without using the library's top-level, non-dbapi2 client.
+            cur.execute("set allow_experimental_object_type = 1")
+            cur.execute("set allow_experimental_lightweight_delete = 1")
```
sh-rp, line 63: please mention in the docs that certain features are experimental

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -0,0 +1,209 @@
...
+        return f"{database_name}.{dataset_name}"
+
+    def make_qualified_table_name(self, table_name: str, escape: bool = True) -> str:
+        database_name = (
+            self.capabilities.escape_identifier(self.database_name)
+            if escape
+            else self.database_name
+        )
+        dataset_table_name = (
+            self.capabilities.escape_identifier(f"{self.dataset_name}_{table_name}")
```
sh-rp, line 165: we should make the separator between dataset name and table name configurable and set it to something special, so we don't end up dropping tables by mistake because you are deleting a dataset that is also contained in the name of another dataset that exists, do you know what I mean? maybe the default dataset default separator should be three underscores or something like this.

dlt/destinations/utils.py
```diff
@@ -1,16 +1,48 @@
...
-    else:
-        resource = data
-    return resource
+    if isinstance(data, DltResource):
+        return data
+    resource_name = None if hasattr(data, "__name__") else "content"
+    return cast(DltResource, make_resource(data, name=resource_name))
+
+
+def _convert_to_old_pyformat(
```
sh-rp, line 16: great :)

dlt/sources/helpers/rest_client/auth.py
```diff
@@ -30,9 +30,7 @@
 else:
```
sh-rp, line 30: please revert the changes on the rest helpers, this is under active development by @burnash and will just confuse i think :)

tests/sources/helpers/rest_client/conftest.py
```diff
@@ -28,9 +28,7 @@ def __init__(self, base_url: str):
         self.routes: List[Route] = []
```
sh-rp, line 28: same as for the rest client

tests/load/utils.py
```diff
@@ -302,6 +299,27 @@ def destinations_configs(
                 bucket_url=AZ_BUCKET,
                 extra_info="az-authorization",
             ),
+            DestinationTestConfiguration(
```
sh-rp, line 302: one of these tests, it does not matter which one, need to be moved into the default_sql_configs sections, below clickouse, so we run the default sql tests against it.

sh-rp (changes requested):
Good work, I still have a few points.

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: I added `jsonl` back, its much faster. It was decryption that didn't work, should be easy fix in future so CI decrypted for now

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: https://github.com/dlt-hub/dlt/issues/1181

dlt/sources/helpers/rest_client/auth.py
Pipboyguy: Reverted!

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: https://github.com/dlt-hub/dlt/issues/1183

dlt/destinations/impl/clickhouse/sql_client.py
Pipboyguy: Added configuration option `[destination.clickhouse.credentials.dataset_table_separator]`

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,416 @@
...
+                secret_access_key = dlt.config[
+                    "destination.filesystem.credentials.gcp_secret_access_key"
+                ]
+            else:
+                access_key_id = None
+                secret_access_key = None
+
+            structure = "auto"
+
+            template = Template("""
```
rudolfix, line 203: why do you use jinja template to render this? do we have it in dependencies? is it maybe coming with clickhouse client? if not then most probably this will stop working on clean install

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,416 @@
...
+            schema_table[c[0]] = schema_c  # type: ignore
+        return True, schema_table
+
+    def get_stored_schema(self) -> StorageSchemaInfo:
+        return super().get_stored_schema()
+
+    @staticmethod
+    def _gen_not_null(v: bool) -> str:
+        # ClickHouse fields are not nullable by default.
+        # We use the `Nullable` modifier instead of NULL / NOT NULL modifiers to cater for ALTER statement.
```
rudolfix, line 407: OK. but I have a test that adds NON NULL column to existing column with the data and it passes. so we are not adding this constraint or it is impossible?
`test_many_schemas_single_dataset`

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -0,0 +1,218 @@
...
+    capabilities: ClassVar[DestinationCapabilitiesContext] = capabilities()
+
+    def __init__(self, dataset_name: str, credentials: ClickHouseCredentials) -> None:
+        super().__init__(credentials.database, dataset_name)
+        self._conn: clickhouse_driver.dbapi.connection = None
+        self.credentials = credentials
+        self.database_name = credentials.database
+
+    def has_dataset(self) -> bool:
+        return len(self._list_tables()) > 0
```
rudolfix, line 58: way better would be to create "sentinel" table to indicate that dataset exists. please check waviate that does exactly that

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -0,0 +1,218 @@
...
+        return self._conn
+
+    def execute_sql(
+        self, sql: AnyStr, *args: Any, **kwargs: Any
+    ) -> Optional[Sequence[Sequence[Any]]]:
+        with self.execute_query(sql, *args, **kwargs) as curr:
+            return None if curr.description is None else curr.fetchall()
+
+    def create_dataset(self) -> None:
+        # ClickHouse doesn't have schemas.
```
rudolfix, line 96: yeah here

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -0,0 +1,218 @@
...
+        db_args = kwargs.copy()
+
+        if args:
+            query, db_args = _convert_to_old_pyformat(query, args, OperationalError)
+            db_args.update(kwargs)
+
+        # Prefix each query transaction with experimental settings.
+        # These are necessary for nested datatypes to be available and other operations to work.
+        query = (
+            "set allow_experimental_lightweight_delete = 1;"
```
rudolfix, line 142: do we need to set it for each query? below you are sending those in separate calls so we have 4 roundtrips to the server for a single statement. maybe `settings` parameter would do the trick? or you can send those when opening connection (or just after)

dlt/destinations/impl/clickhouse/configuration.py
```diff
@@ -0,0 +1,90 @@
...
+            ]
+        )
+        return url
+
+
+@configspec
+class ClickHouseClientConfiguration(DestinationClientDwhWithStagingConfiguration):
+    destination_type: Final[str] = dataclasses.field(default="clickhouse", init=False, repr=False, compare=False)  # type: ignore[misc]
+    credentials: ClickHouseCredentials = None
+    dataset_name: Final[str] = dataclasses.field(default="", init=False, repr=False, compare=False)  # type: ignore[misc]
```
rudolfix, line 78: why do we have this line? why the default is ""? I agree that it would be cool to allow "empty" dataset so tables are created without prefixes. we have this in waviate. but that will require more code and testing

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,416 @@
...
+    def state(self) -> TLoadJobState:
+        return "completed"
+
+    def exception(self) -> str:
+        raise NotImplementedError()
+
+
+class ClickHouseMergeJob(SqlMergeJob):
+    @classmethod
+    def _to_temp_table(cls, select_sql: str, temp_table_name: str) -> str:
```
rudolfix, line 290: this is still failing. some problems with table aliasing
```
DB::Exception: Missing columns: 'd.node_id' 'd.url' 'd.id' while processing query: 'SELECT 1 FROM dlt.github_3_202404091123570088_staging___issues AS s WHERE (d.id = id) OR ((d.url = url) AND (d.node_id = node_id))', required columns: 'id' 'node_id' 'url' 'd.id' 'd.url' 'd.node_id', maybe you meant: 'id', 'node_id' or 'url': While processing 1 IN ((SELECT 1 FROM (SELECT 1 FROM dlt.github_3_202404091123570088_staging___issues AS s WHERE (d.id = s.id) OR ((d.url = s.url) AND (d.node_id = s.node_id))) LIMIT 1) AS _subquery5).
```

I expect this will be the biggest problem to fix

rudolfix (changes requested):
looks good! I fixed some tests and issues (see commit log)

the biggest thing is to make `merge` work. temp tables are not created right now. run any of the merge tests you'll see

some other tests are not working. you should fix sql_client and job_client tests first.

dlt/destinations/impl/clickhouse/configuration.py
Pipboyguy: Removed! I agree we don't need the added complexity.

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: There was no real reason to use jinja templating in this simple case, agreed! It makes the code less readable now that I had another look.

Removed it in favor of built in string formatting.

dlt/destinations/impl/clickhouse/sql_client.py
Pipboyguy: I initially had trouble with `clickhouse-driver` not recognizing the desired settings in `settings` parameter, but after moving the settings to the DSN, it worked!

Updated `to_url` to reflect this and removed the unnecessary round trips.

dlt/common/time.py
```diff
@@ -142,6 +141,8 @@ def ensure_pendulum_time(value: Union[str, datetime.time]) -> pendulum.Time:
         result = parse_iso_like_datetime(value)
         if isinstance(result, pendulum.Time):
             return result
+        elif isinstance(result, pendulum.DateTime):
```
sh-rp, line 144: note: not sure what is going on here, I will have to investigate and see if clickhouse just saves times as datetime or wether there is something else going on here.

Pipboyguy:
https://github.com/ClickHouse/ClickHouse/issues/6697

@rudolfix  @sh-rp  Correlated sub-queries aren't supported in CH as yet, so the merge support is complicated to implement

Pipboyguy:
Clickhouse does not support correlated subqueries, see https://github.com/ClickHouse/ClickHouse/issues/6697. We'll have to using an inner join instead of `EXISTS`. Making progress by following this strategy in test results.

sh-rp:
Open questions (wip):
* dlt support
* binary insert
* experimental json support
*

dlt/common/destination/capabilities.py
```diff
@@ -53,6 +54,9 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     insert_values_writer_type: str = "default"
     supports_multiple_statements: bool = True
     supports_clone_table: bool = False
+    scd2_high_timestamp: pendulum.DateTime = HIGH_TS
```
sh-rp, line 57: clickhouse max timestamp is 2099, so I needed to make this configurable for each destinations, I'm not satisfied that the capabilities are the right place, if you have a better idea for it, lmk

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,402 @@
...
+        return [
+            f"FROM {root_table_name} AS d JOIN {staging_root_table_name} AS s ON {join_conditions}"
+        ]
+
+    @classmethod
+    def gen_update_table_prefix(cls, table_name: str) -> str:
+        return f"ALTER TABLE {table_name} UPDATE"
+
+    @classmethod
+    def requires_temp_table_for_delete(cls) -> bool:
```
sh-rp, line 279: this will make the merge job use the merge strategy for table with child tables which does the delete also on the root table with a temp table. I could not get clickhouse to do a delete in one sql call when selecting multiple keys, so this was the easiest way. The only drawback is, that you will need to have a unique column defined.

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,402 @@
...
+            self.active_hints.get(hint)
+            for hint in self.active_hints.keys()
+            if c.get(hint, False) is True
+            and hint not in ("primary_key", "sort")
+            and hint in self.active_hints
+        )
+
+        # Alter table statements only accept `Nullable` modifiers.
+        # JSON type isn't nullable in ClickHouse.
+        type_with_nullability_modifier = (
```
sh-rp, line 315: Clickhouse will not fail when updating a table to add a non-nullable column to a table. I think we should probably prevent update statements that contain non-nullable columns in _make_add_column_sql for all destinations so make the behavior uniform. There is one test that is not passing because of this (adding a dlt_root_it to an existing table does not raise an exception on clickhouse at this point)

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,402 @@
...
+                    password=client.credentials.password,
+                    secure=bool(client.credentials.secure),
+                ) as clickhouse_connect_client:
+                    insert_file(
+                        clickhouse_connect_client,
+                        qualified_table_name,
+                        file_path,
+                        fmt=clickhouse_format,
+                        settings={
+                            "allow_experimental_lightweight_delete": 1,
```
sh-rp, line 184: I use the lightweight delete in the merge job, it seems to be working well.

dlt/destinations/impl/clickhouse/clickhouse.py
sh-rp: see above

dlt/destinations/impl/clickhouse/clickhouse_adapter.py
```diff
@@ -0,0 +1,60 @@
...
+- Whether multithread request execution is possible.
+- Data replication parameters.
+
+See https://clickhouse.com/docs/en/engines/table-engines.
+"""
+TABLE_ENGINE_TYPES: Set[TTableEngineType] = set(get_args(TTableEngineType))
+TABLE_ENGINE_TYPE_HINT: Literal["x-table-engine-type"] = "x-table-engine-type"
+
+
+def clickhouse_adapter(data: Any, table_engine_type: TTableEngineType = None) -> DltResource:
```
sh-rp, line 26: note: this still needs a test

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -0,0 +1,210 @@
...
+            else:
+                return DatabaseTransientException(ex)
+        elif isinstance(
+            ex,
+            (
+                clickhouse_driver.dbapi.errors.OperationalError,
+                clickhouse_driver.dbapi.errors.InternalError,
+            ),
+        ):
+            if term := cls._maybe_make_terminal_exception_from_data_error():
```
sh-rp, line 186: this is in a couple of destinations but does not do anything for most. I'll keep it for now, but some destination overall cleanup is due sometime i think.

dlt/destinations/sql_jobs.py
```diff
@@ -582,8 +579,22 @@ def gen_scd2_sql(
                     staging_table_name = sql_client.make_qualified_table_name(table["name"])
                 sql.append(f"""
                     INSERT INTO {table_name}
-                    SELECT *
+                    SELECT s.*
                     FROM {staging_table_name} AS s
-                    WHERE NOT EXISTS (SELECT 1 FROM {table_name} AS f WHERE f.{unique_column} = s.{unique_column});
+                    LEFT JOIN {table_name} as t
+                        on s.{unique_column} = t.{unique_column}
+                    WHERE t.{unique_column} IS NULL or t.{unique_column} = '';
```
sh-rp, line 586: unique column ends up being empty string in clickhouse...

dlt/helpers/dbt/profiles.yml
```diff
@@ -171,3 +171,16 @@ databricks:
       http_path: "{{ env_var('DLT__CREDENTIALS__HTTP_PATH') }}"
       token: "{{ env_var('DLT__CREDENTIALS__ACCESS_TOKEN') }}"
       threads: 4
+
+
+clickhouse:
```
sh-rp, line 176: I have disabled the dbt tests for clickhouse, the connection set up here works, but since we have the table prefixes after running the seed on jaffle for example the tests will fail because the tables are not found. generally I think it will work though.

tests/cases.py
```diff
@@ -197,6 +197,7 @@ def assert_all_data_types_row(
     timestamp_precision: int = 6,
     schema: TTableSchemaColumns = None,
     expect_filtered_null_columns=False,
+    allow_string_binary: bool = False,
```
sh-rp, line 200: on clickhouse,  loading a binary row from parquet end's up being a base64 string and loading from jsonl ends up being a binary just converted to a string (or vice versa, forgot, you can see it in the tests). I will mention this in the docs, I supposed we'll have to live with it.

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -139,7 +139,7 @@ def test_merge_on_ad_hoc_primary_key(destination_config: DestinationTestConfigur
 
 
 @dlt.source(root_key=True)
-def github():
+def github(remove_lists: bool = False):
```
sh-rp, line 142: see above, for clickhouse i need to remove the lists in the dataset, otherwise it will look for a List[json] field and not find it. needs to go into the docs

tests/load/pipeline/test_stage_loading.py
```diff
@@ -203,9 +204,13 @@ def test_all_data_types(destination_config: DestinationTestConfiguration) -> Non
         exclude_types=exclude_types, exclude_columns=exclude_columns
     )
 
-    # bigquery cannot load into JSON fields from parquet
+    # clickhouse json is experimental, will not work for parquet and makes strange changes for jsonl
```
sh-rp, line 207: there is another strange issue in the json object in clickhouse, where a nested list that has [1,2,3,"4"] as a value end's up being a list of strings in the destination. another reason to maybe disabled json by default for now.

tests/load/test_job_client.py
```diff
@@ -772,18 +790,21 @@ def _load_something(_client: SqlJobClientBase, expected_rows: int) -> None:
...
         _load_something(client, 3)
 
-        # adding new non null column will generate sync error
+        # adding new non null column will generate sync error, except for clickhouse, there it will work
         event_3_schema.tables["event_user"]["columns"]["mandatory_column"] = new_column(
             "mandatory_column", "text", nullable=False
         )
         client.schema._bump_version()
-        with pytest.raises(DatabaseException) as py_ex:
+        if destination_config.destination == "clickhouse":
```
sh-rp, line 798: with the suggestion above, we should be able to have this run like all other destinations

dlt/destinations/sql_jobs.py
```diff
@@ -582,8 +579,22 @@ def gen_scd2_sql(
                     staging_table_name = sql_client.make_qualified_table_name(table["name"])
                 sql.append(f"""
```
jorritsandbrink, line 580: Could this SQL statement be the following, for consistency with the logic for the root table?

```sql
  INSERT INTO {table_name}
  SELECT *
  FROM {staging_table_name}
  WHERE {unique_column} NOT IN (SELECT {unique_column} FROM {table_name});
```

dlt/common/destination/capabilities.py
rudolfix: we need the same timestamp in each destination. my question to @jorritsandbrink why not to allow nulls? so the active record is marked as having no end date

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,402 @@
...
+
+TABLE_ENGINE_TYPE_TO_CLICKHOUSE_ATTR: Dict[TTableEngineType, str] = {
+    "merge_tree": "MergeTree",
+    "replicated_merge_tree": "ReplicatedMergeTree",
+}
+
+
+class ClickHouseTypeMapper(TypeMapper):
+    sct_to_unbound_dbt = {
+        "complex": "JSON",
```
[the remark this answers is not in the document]
rudolfix: do we have the same problem when `jsonl` and `parquet` are used? the difference
- jsonl stores complex as json fragments
- parquet stores comples as string
please check it out. maybe if we have a string in jsonl it will start to work

I'd disable JSON support by default and make it configurable. still jsonl keeping complex as dict/list (not string) may be a problem

dlt/destinations/impl/clickhouse/clickhouse.py
rudolfix: I think this is the only way to go :)

dlt/destinations/impl/clickhouse/clickhouse.py
rudolfix: I can't find code generating MERGE sql. but if you do it then it is completely fine

dlt/destinations/impl/clickhouse/clickhouse.py
rudolfix: dremio has the same behavior and I think we are skipping this check in the tests. my take would be to generate log warning that such thing happened (ALTER with NOT NULL) and let destination handle it. if the table is empty adding NOT NULL will work. also some weird things like automatic default value and whatever may be the case. So I'd let destination handle this

rudolfix: we may add another capability

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -0,0 +1,408 @@
...
+        sql = SqlJobClientBase._get_table_update_sql(self, table_name, new_columns, generate_alter)
+
+        if generate_alter:
+            return sql
+
+        # Default to 'ReplicatedMergeTree' if user didn't explicitly set a table engine hint.
+        table_type = cast(
+            TTableEngineType, table.get(TABLE_ENGINE_TYPE_HINT, "replicated_merge_tree")
+        )
+        sql[0] = f"{sql[0]}\nENGINE = {TABLE_ENGINE_TYPE_TO_CLICKHOUSE_ATTR.get(table_type)}"
```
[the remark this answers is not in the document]
rudolfix: probably it is. I'm just wonder how the table without primary key or order defined is created. we have such tables in tests. clickhouse do not allow for that. you can ignore this comment :)

tests/load/pipeline/test_stage_loading.py
rudolfix: let's disable by default!

tests/load/pipeline/test_clickhouse.py
```diff
@@ -0,0 +1,83 @@
+from typing import Iterator
```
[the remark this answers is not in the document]
sh-rp: done

dlt/common/destination/capabilities.py
sh-rp: this is how I did it in my initial implementation, that would make it a bit easier

dlt/destinations/impl/clickhouse/clickhouse_adapter.py
sh-rp: done

dlt/destinations/impl/clickhouse/clickhouse.py
sh-rp: no, what I meant to say, it will fallback to the code that is used to to a merge for a tablechain with more than one table. there are no MERGE statements.

dlt/destinations/sql_jobs.py
sh-rp: works! thanks!

dlt/destinations/impl/clickhouse/clickhouse.py
sh-rp: the logger already prints a warning in the superclass if you do this.

dlt/destinations/impl/clickhouse/clickhouse.py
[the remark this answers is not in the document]
sh-rp: I have disabled it completely for now, and updated the tests. it will be possible to put it in once someone wants to have it. I'm not sure wether it is worth the trouble at this point tbh.

dlt/destinations/impl/clickhouse/sql_client.py
sh-rp: ok, I kicked it out, let's see if all the tests still pass

tests/load/pipeline/test_merge_disposition.py
sh-rp: removed this, no longer needed for disabled complex types

sh-rp:
Follow up ticket for filesystem gcs s3 interoperability mode: https://github.com/dlt-hub/dlt/issues/1272

Pipboyguy:
> @Pipboyguy could you update the docs a bit to explain better
> 
>     * why there are two different ports used?
> 
>     * How to use gcs? A link to set up the ineroperability mode and instruction on how to provide the credentials for gcs are needed. This has changed now by the way and these credentials are part of the clickhouse config, so will be set like this:
>       [destination.clickhouse.credentials]
>       gcp_access_key_id="..."
>       gcp_secret_access_key="..."

Added!

sh-rp:
Note to self: loading certain values (such as 12.7001) via jsonl will produce rounding errors predictably, tests need an update and we need a note in the docs.

sh-rp (dismissed):

sh-rp (approved):

PR #1103, opened 2024-03-18
add a script to check wether our poetry lockfile is in order

sh-rp:
### Description
Add a script that checks the poetry lockfile for version hashes to ci


rudolfix (approved):

rudolfix (changes requested):
this does not work on windows :/ better to parse this in Python

sh-rp:
@rudolfix alright, done. I have also moved all the  tools to a tools folder (which I like, but you let me know if this is any good). I made sure that check-package still works as intended for the root dir and all above.

rudolfix (approved):
LGTM!

PR #1107, opened 2024-03-18
Extend custom destination

sh-rp:
### Description
This PR introduces two new settings for the custom destination decorator:
* max_nesting_level to control the normalizer
* skip_dlt_columns_and_tables to skip internal columns and tables


dlt/pipeline/pipeline.py
```diff
@@ -456,7 +456,18 @@ def normalize(
             return None
 
         # make sure destination capabilities are available
-        self._get_destination_capabilities()
+        caps = self._get_destination_capabilities()
```
sh-rp, line 459: @rudolfix i need some guidance where to inject / overwrite the max_nesting_level coming from a destination. I realize this place is very likely not the right one, but I am not sure where and how to do it. Should I get the capabilities context in the relationalnormalizer and not persist this setting to the schema at all, or what is the best way?

rudolfix: the only thing you need to do is to fix `NormalizersConfiguration`
```
        # get naming from capabilities if not present
        if self.naming is None:
            if self.destination_capabilities:
                self.naming = self.destination_capabilities.naming_convention
```
detect the type of the json normalizer and apply the settings to it like the below. you can override existing settings. I think capabilities (if not None) should have precedence over the source settings.

what happens later:
when new schema is created this setting will be used
when schema is loaded - it will not but when we call update_normalize - it will. and we do that in normalizer (in schema.clone). so it should work!

rudolfix (changes requested):
this is good! see the trick to apply max nesting to schema

dlt/destinations/impl/destination/destination.py
```diff
@@ -93,9 +100,14 @@ def run(self, start_index: int) -> Iterable[TDataItems]:
...
         ) == 0, "Batch size was changed during processing of one load package"
 
+        # on record batches we cannot drop columns, we need to
+        # select the ones we want to keep
+        keep_columns = list(self._table["columns"].keys())
         start_batch = start_index / self._config.batch_size
         with pyarrow.parquet.ParquetFile(self._file_path) as reader:
-            for record_batch in reader.iter_batches(batch_size=self._config.batch_size):
+            for record_batch in reader.iter_batches(
+                batch_size=self._config.batch_size, columns=keep_columns
```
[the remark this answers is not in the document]
sh-rp: yes, the docs say it does: 
https://arrow.apache.org/docs/12.0/_modules/pyarrow/parquet/core.html#ParquetFile.iter_batches

dlt/pipeline/pipeline.py
sh-rp: since this is set on the nested json normalizer settings i had to change a bit more but not much, I hope it is ok to change the type from mapping to dict in there.

rudolfix (dismissed):
I have a few additions to docs and cross references. Let's merge this first. LGTM!

PR #1109, opened 2024-03-18
Cancel previous CI runs for PRs when new commits pushed

sultaniman:
This PR adjusts workflows on PRs groups to cancel all previously running jobs according to docs https://docs.github.com/en/actions/using-jobs/using-concurrency

I added the following rule for each workflow
```yml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,6 +8,10 @@ on:
       - devel
   workflow_dispatch:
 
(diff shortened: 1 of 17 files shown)


sh-rp (approved):
very cool!

PR #1114, opened 2024-03-19
Docs: Add preprocessor

sh-rp:
### Description
We want to run all markdown pages through a preprocessor which injects snippets and other links. This will make snippets inserting easier and will remove code duplication for the most part. Re-organizing the snippets will come in a next step.

### DIFF
--- a/docs/website/.gitignore
+++ b/docs/website/.gitignore
@@ -4,6 +4,9 @@
 # Production
 /build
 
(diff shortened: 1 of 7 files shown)


AstrakhantsevaAA (approved):
Sehr gut!
I love that we don't need to re-generate docs/examples locally before commit

PR #1119, opened 2024-03-20
custom destination fixes

rudolfix:
### Description
see commit list

@sh-rp you can take over this:
- sink decorator must work without parens
- add good docstrings to the decorators
- write tests for:
1. add test for _DESTINATIONS
2. passing SPEC in the decorator
3. do we have test for instantiated factory like pipeline.run(destination=my_gcp_sink(credentials=...)

### DIFF
--- a/tests/destinations/test_custom_destination.py
+++ b/tests/destinations/test_custom_destination.py
@@ -7,13 +7,14 @@ import os
 import inspect
 
 from copy import deepcopy
+from dlt.common.configuration.specs.base_configuration import configspec
 from dlt.common.typing import TDataItems
 from dlt.common.schema import TTableSchema
 from dlt.common.data_writers.writers import TLoaderFileFormat
 from dlt.common.destination.reference import Destination
 from dlt.pipeline.exceptions import PipelineStepFailed
 from dlt.common.utils import uniq_id
-from dlt.common.exceptions import InvalidDestinationReference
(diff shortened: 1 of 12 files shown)


sh-rp:
I think 3. is already done.

tests/extract/test_decorators.py
```diff
@@ -631,14 +631,14 @@ def test_sources_no_arguments() -> None:
     def no_args():
         return dlt.resource([1, 2], name="data")
 
-    # there is no spec if no arguments
+    # there is a spec even if no arguments
     SPEC = _SOURCES[no_args.__qualname__].SPEC
-    assert SPEC is None
```
sh-rp, line 636: since the spec is now created even if there are no args, i assume updating this test is in order

tests/destinations/test_custom_destination.py
```diff
@@ -437,6 +451,54 @@ def my_gcp_sink(
     )
 
 
+def test_destination_with_spec() -> None:
```
sh-rp, line 454: lmk if this is enough wrt to testing the custom base

rudolfix: improved a little bit. spec is not a custom base. it is THE spec that will be used

rudolfix (commented):
LGTM! you can merge @sh-rp

PR #1122, opened 2024-03-21
materializes table schemas for empty tables

### LINKED ISSUE #1116: Generate data with just table structure in the destination if there was nothing in the extract phase.
Slack thread: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1710518926508509

User ran into the following issue:
When using the Stripe example to push data to S3 in Parquet format, an issue occurs where no Parquet file is created in S3 if the Stripe endpoint returns an empty array (e.g., for a new Stripe account with no transactions). This behavior seems to be an unintended consequence, as discussed in our team's communication. The expected behavior is to have a Parquet file created in S3 with a defined schema, even if the Stripe data array is empty.

It was suggested that this is an unintended behavior and there's a  possibility of allowing the yield of an empty list to create tables/files.

rudolfix:
### Description
fixes #1116 

Here we allow to create empty tables / files at the destination. The standard `dlt` behavior is to delay creation of tables until first data arrives. This is beneficial when schema is determined from the data (fully or partially). Specifically yielding empty list will not create empty tables / files. We do not want to change this behavior.

In some cases ie. when full schema is known upfront or generated at runtime, empty tables and files should be created without waiting for data.

This PR introduces a new marker in `dlt.mark` that, when yielded, will materialize empty tables. Example (dynamic schema)
```
    @dlt.resource
    def users():
        yield dlt.mark.with_hints(
            # this is a special empty item which will materialize table schema
            # emit table schema with the item
            dlt.mark.make_hints(
                columns=[
                    {"name": "id", "data_type": "bigint", "precision": 4, "nullable": False},
                    {"name": "name", "data_type": "text", "nullable": False},
                ]
            ),
        )
```
example (defined schema)
```
    @dlt.resource(columns=UsersPydanticModel)
    def users():
        yield dlt.mark.materialize_table_schema()
```

### DIFF
--- a/dlt/normalize/items_normalizers.py
+++ b/dlt/normalize/items_normalizers.py
@@ -201,10 +201,16 @@ class JsonLItemsNormalizer(ItemsNormalizer):
                 schema_updates.append(partial_update)
                 logger.debug(f"Processed {line_no} lines from file {extracted_items_file}")
             if line is None and root_table_name in self.schema.tables:
-                partial_update = self._normalize_chunk(
-                    root_table_name, [{}], False, skip_write=True
-                )
-                schema_updates.append(partial_update)
+                # TODO: we should push the truncate jobs via package state
+                # not as empty jobs. empty jobs should be reserved for
+                # materializing schemas and other edge cases ie. empty parquet files
+                root_table = self.schema.tables[root_table_name]
+                if not has_table_seen_data(root_table):
+                    # if this is a new table, add normalizer columns
+                    partial_update = self._normalize_chunk(
+                        root_table_name, [{}], False, skip_write=True
+                    )
+                    schema_updates.append(partial_update)
                 self.load_storage.write_empty_items_file(
                     self.load_id,
                     self.schema.name,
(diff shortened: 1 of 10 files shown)


dlt/normalize/items_normalizers.py
```diff
@@ -193,22 +195,31 @@ def __call__(
...
                 items: List[TDataItem] = json.loadb(line)
-                partial_update = self._normalize_chunk(root_table_name, items, may_have_pua(line))
+                partial_update = self._normalize_chunk(
+                    root_table_name, items, may_have_pua(line), skip_write=False
+                )
                 schema_updates.append(partial_update)
                 logger.debug(f"Processed {line_no} lines from file {extracted_items_file}")
             if line is None and root_table_name in self.schema.tables:
-                # write only if table seen data before
+                # TODO: we should push the truncate jobs via package state
```
sh-rp, line 204: yes, this is probably a good idea

sh-rp (approved):
Looks good, I'm assuming there are existing tests in the code that test that no table will be created in the destination for empty resources if this marker is not emitted, right?

sh-rp:
PS: What about docs?

rudolfix:
@sh-rp where in the docs we should add this trick? I want to rewrite the Pipeline doc. it is outdated. and there I can mention all the possible markers for items

rudolfix:
> Looks good, I'm assuming there are existing tests in the code that test that no table will be created in the destination for empty resources if this marker is not emitted, right?

you wrote this tests when we were adding empty files to replace resources

Issue #1037, opened 2024-02-29
add fast loading methods for bigquery destination


rudolfix:
**Background**
Current method using load jobs is optimized for large loads. However most of the load packages is quite small. We can improve loading speeds by implementing streaming insert copy jobs. There's a singer target https://github.com/z3z1ma/target-bigquery by @z3z1ma from where we can take code.

we'll implement unconstrained (schema-less) version of the above when #891 in merged. this is about extending existing destination

**Tasks**
1. * [ ] allow user to select the loading API via destination configuration and per resource/table via bigquery adapter
2. * [ ] implement insert api and optionally storage write api
3. * [ ] allow both parquet and jsonl to be loaded this way. consider adding standard file readers for those types (port them from #891 )
4. * [ ] tests and documentation

PR #998, opened 2024-02-25
allows naming conventions to be changed

rudolfix:
### Description
This PR attempts to allow any naming convention to be used (ie UPPER CASE or CamelCase). Until now any naming convention that changed casing of columns was failing. Also destinations have problems when reflecting such columns  from information schema.

Fixes following issues
#1470 
#964 
#1074 (by offering alternative naming convention)
#1085 
#860  (most probably, fixture order was messed up)
#1471 (partially, without internal source)

read the commit log for details!

naming conventions:
1. Make sure dlt engine is agnostic to the naming convention used on schema, table and column level (done, not well tested)
2. Make sure that schema settings (preferred types, normalizer settings, default hints) are also updated in (1)
3. Expose and document methods to change default hints, preferred settings and other compiled properties, update documentation
4. Make sure that all destinations work with forced lower and upper case naming conventions (not done)
5. destination capabilities are extended to define the following
- naming convention (the default snake case is dropped, convention set in Schema is adopted by default - it is snake case)
- a flag if destination is or is not case sensitive
- case folding function used by destination to get insensitive identifiers (`str`, `lower`, `upper`)
- all destinations are by default configured in **case insensitive mode**
6. we use INFORMATION SCHEMA to read all columns for all tables at once. with many tables the speed increase should be significant
7. `sql_cs` and `sql_ci` naming conventions to create case sensitive and insensitive names compatible with SQL
8. `duck_case` and `direct` will be unified and will allow any characters in the identifiers
9. quadrant now works with in memory and local engine

todo:
- `sql_cs` and `sql_ci` not yet implemented (but that's trivial)
- Also fix bug in Qdrant - state and recent schemas are incorrectly retrieved (somehow sort order was ignored during implementation) (90% fixed)
- also fix bugs in Weaviate tests
- documentation is still missing

reference:
(4) an overview of case sensitivity and preferred naming convention per destination: https://www.linkedin.com/posts/toby-mao_sql-activity-7170104665412423680-17gv/

PR #1123, opened 2024-03-21
feat(bigquery): add streaming inserts support

IlyaFaer:
Towards #1037


dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -217,33 +221,44 @@ def restore_file_load(self, file_path: str) -> LoadJob:
...
+            job = DestinationJsonlLoadJob(
+                table,
+                file_path,
+                self.config,
+                self.schema,
+                destination_state(),
+                functools.partial(streaming_load, self.sql_client),
+                [],
+            )
+        else:
```
IlyaFaer, line 234: If `streaming` is set in config, we use another type of job with a specific callable.

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -425,3 +438,10 @@ def _from_db_type(
         self, bq_t: str, precision: Optional[int], scale: Optional[int]
     ) -> TColumnType:
         return self.type_mapper.from_db_type(bq_t, precision, scale)
+
+
+def streaming_load(sql_client, items, table):
+    full_name = sql_client.make_qualified_table_name(table["name"], escape=False)
+
+    bq_client = sql_client._client
+    bq_client.insert_rows_json(full_name, items)
```
IlyaFaer, line 447: Callable for streaming inserts

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -217,33 +221,44 @@ def restore_file_load(self, file_path: str) -> LoadJob:
...
-                    raise DestinationTransientException(gace) from gace
-        return job
+        if self.config.loading_api == "streaming":
+            job = DestinationJsonlLoadJob(
+                table,
+                file_path,
+                self.config,
+                self.schema,
+                destination_state(),
+                functools.partial(streaming_load, self.sql_client),
```
IlyaFaer, line 231: Here I do a trick with `partial` to add the third argument into the callable

IlyaFaer:
@rudolfix, if you're wondering if it's working. Yes, it is, here is a scratchy proof:

Untitled2

So, I inserted the data and then read it from the BigQuery backend. I don't see yet why it would be a bad solution, so I'm tidying it up 👌

dlt/destinations/impl/bigquery/bigquery.py
rudolfix: this is cool!

rudolfix: can we set the batch size here?

dlt/destinations/impl/bigquery/configuration.py
```diff
@@ -19,6 +19,8 @@ class BigQueryClientConfiguration(DestinationClientDwhWithStagingConfiguration):
     retry_deadline: float = (
         60.0  # how long to retry the operation in case of error, the backoff 60 s.
     )
+    loading_api: str = "default"
+    batch_size: int = 0
```
rudolfix, line 23: does that work? batch_size==0 makes job to pass just a file name? or maybe I'm wrong

rudolfix (changes requested):
good direction! but it can't be so hacky:
1. do not import from other destination. move jobs that you need to `job_impl.py`. we need a small refactor here
2. also support parquet via the other job
3. config option is cool but we need per table definition. look at `bigquery_adapter` there you can add table hint with the loader type and interpret it during loading. should be very easy

dlt/destinations/impl/bigquery/configuration.py
IlyaFaer: Actually, I have a config set to 1000 locally, so I didn't yet use the default value.

I'm thinking, do we actually need a batch size? BigQuery streaming API streams rows one by one, not in batches. Seems not useful to split the data into batches.

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -217,32 +222,53 @@ def restore_file_load(self, file_path: str) -> LoadJob:
...
-        job = super().start_file_load(table, file_path, load_id)
-
-        if not job:
-            try:
-                job = BigQueryLoadJob(
-                    FileStorage.get_file_name_from_file_path(file_path),
-                    self._create_load_job(table, file_path),
-                    self.config.http_timeout,
-                    self.config.retry_deadline,
+        insert_api = table.get("insert_api", self.config.loading_api)
```
IlyaFaer, line 225: Parameter given through adapter has higher priority that config

dlt/common/schema/typing.py
```diff
@@ -171,6 +171,7 @@ class TTableSchema(TypedDict, total=False):
     columns: TTableSchemaColumns
     resource: Optional[str]
     table_format: Optional[TTableFormat]
+    insert_api: Optional[Literal["streaming", "default"]]
```
IlyaFaer, line 174: I guess it's okay that it's a global param. BigQuery is not the only service, which allows different insert approaches.

rudolfix: I'd prefer not to specify it until we have more cases... please use `x-` prefix to have custom hints.

rudolfix (changes requested):
OK we are almost there. I have one comment regarding the hint.

there's one more thing:
```
google-cloud-bigquery = {version = ">=2.26.0", optional = true}
google-cloud-bigquery = {version = ">=3.14", optional = true}
```
which is good, because we should work on the newest version. but we have two problems:
1. I use some kind of hack to access query job from the dbapi cursor to load data frame:
https://github.com/dlt-hub/dlt/actions/runs/8431727786/job/23089624139?pr=1123#step:8:921
are you able to make it work for both versions? I have another PR where I do it without any hacks
https://github.com/dlt-hub/dlt/pull/998/files#diff-94cc93a6e16c7508322344071cf19788557e561db8b55b6755261155bce91cb8R48
please take code from there

2. something is going on here:
https://github.com/dlt-hub/dlt/actions/runs/8431727786/job/23089624139?pr=1123#step:8:932

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -217,32 +223,53 @@ def restore_file_load(self, file_path: str) -> LoadJob:
         return job
 
     def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
-        job = super().start_file_load(table, file_path, load_id)
```
rudolfix, line 220: you should keep this first. it will launch various sql jobs ie merge operations. tables that do insert also participate in this

tests/load/bigquery/test_bigquery_streaming_insert.py
```diff
@@ -0,0 +1,33 @@
+import dlt
+from dlt.destinations.impl.bigquery.bigquery_adapter import bigquery_adapter
+from tests.pipeline.utils import assert_load_info
+
+
+def test_bigquery_streaming_insert():
```
rudolfix, line 6: please write a test where you use merge write disposition and use nested data. we need to check if merge sql jobs run

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -217,32 +223,53 @@ def restore_file_load(self, file_path: str) -> LoadJob:
...
-                    return self.restore_file_load(file_path)
-                elif reason in BQ_TERMINAL_REASONS:
-                    # google.api_core.exceptions.BadRequest - will not be processed ie bad job name
-                    raise LoadJobTerminalException(
-                        file_path, f"The server reason was: {reason}"
-                    ) from gace
-                else:
-                    raise DestinationTransientException(gace) from gace
+
+            job = job_cls(
```
rudolfix, line 237: why this is not put into the same exception handler as the `LoadJob` below?
```
try:
...
except api_core_exceptions.GoogleAPICallError as gace:
```
what happens there is to tell terminal exceptions from regular exceptions. are the error codes from insert api completely different?

rudolfix: what singer-target from Alex Butler does on error?

rudolfix (changes requested):
OK I've found a few inconsistences

tests/load/bigquery/test_bigquery_streaming_insert.py
IlyaFaer: Seems it's not so easy with merge disposition. Streaming means buffers are used in BigQuery. Merge is trying to update data, to rearrange, but it's not allowed for 90 mins after streaming insert. Only reading.

IlyaFaer: @rudolfix, so, yes, merge is not that easy with streaming. It may take from several minutes to 1.5 h for the inserted data to become fully available. Thus, merge will work, but its speed is limited (because it does `TRUNCATE`). Streaming insert is just not intended for that kind of operations.

A warning in the docs will be needed, I guess. And something smart for the tests as well... Thinking about it...

rudolfix: @IlyaFaer this will be super confusing for our users! yeah I knew that stream insert is doing internal batching but if timeouts are that long it is really not suitable for what we do. my proposition is to ditch this and replace with https://github.com/z3z1ma/target-bigquery/blob/main/target_bigquery/storage_write.py which let's you flush the cache.

for append-only streaming insert we just did this: https://dlthub.com/docs/examples/custom_destination_bigquery/

tests/load/bigquery/test_bigquery_streaming_insert.py
```diff
@@ -0,0 +1,41 @@
...
+    with pipe.sql_client() as client:
+        with client.execute_query("SELECT * FROM test_streaming_items;") as cursor:
+            res = cursor.fetchall()
+            assert tuple(res[0])[:2] == (1, 2)
+
+
+def test_bigquery_streaming_nested_data():
+    @dlt.resource
+    def test_resource():
+        yield {"field1": {"nested_field": 1}, "field2": {"nested_field": 2}}
```
rudolfix, line 28: please put `{"nested_field": 2}` in list. that will generate a nested table that you should also test

tests/load/bigquery/test_bigquery_streaming_insert.py
```diff
@@ -0,0 +1,41 @@
...
+
+    bigquery_adapter(test_resource, insert_api="streaming")
+
+    pipe = dlt.pipeline(pipeline_name="insert_test", destination="bigquery")
```
rudolfix, line 32: please use `full_refresh` or random `dataset_name` to prevent race condition if somehow we run in parallel on CI

tests/load/bigquery/test_bigquery_streaming_insert.py
```diff
@@ -0,0 +1,41 @@
...
+    pack = pipe.run(test_resource, table_name="test_streaming_items")
```
rudolfix, line 33: here's a big chance your two tests are interfering. they create the same table and write to it. see above to fix it

rudolfix (changes requested):
1. pls read my comment on MERGE problem
2. see my comments on failing tests

rudolfix:
> @rudolfix, there is a failure in `tests/load/bigquery/test_bigquery_streaming_insert.py::test_bigquery_streaming_nested_data FAILED`. It works locally fine.
> I suspect there is something in the Ci/Cl test table. Is it possible to clear it?

test tables and datasets are dropped in test fixture. what happens on CI is weird because there's a record but one of the values is None... my take:
your both tests are writing to the same dataset and table. so because stream insert goes via buffer you can still get a row written to a table from previous test. and the table was in the meantime dropped and recreated with different but compatible schema.

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -425,3 +451,41 @@ def _from_db_type(
...
+
+        Args:
+            exc (google.api_core.exceptions.GoogleAPICallError):
+                Exception raised by the client.
+
+        Returns:
+            bool: True if the exception is retryable, False otherwise.
+        """
+        reason = exc.errors[0]["reason"]
+        return reason in (list(_RETRYABLE_REASONS) + ["notFound"])
```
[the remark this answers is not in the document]
IlyaFaer: I've got a couple of temporary `NotFound` failures during development. I know, BigQuery tables are supposed to be NOT eventually consistent, and still it happened. I erased it, let's see how it'll go...

dlt/destinations/impl/bigquery/bigquery_adapter.py
```diff
@@ -30,6 +30,7 @@ def bigquery_adapter(
     round_half_even: TColumnNames = None,
     table_description: Optional[str] = None,
     table_expiration_datetime: Optional[str] = None,
+    insert_api: Optional[Literal["streaming", "default"]] = None,
```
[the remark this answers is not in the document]
IlyaFaer: Done

dlt/destinations/impl/bigquery/configuration.py
```diff
@@ -20,6 +20,7 @@ class BigQueryClientConfiguration(DestinationClientDwhWithStagingConfiguration):
     retry_deadline: float = (
         60.0  # how long to retry the operation in case of error, the backoff 60 s.
     )
+    batch_size: int = 500
```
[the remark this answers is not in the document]
IlyaFaer: This config goes straight to `DestinationJsonlLoadJob` and `DestinationParquetLoadJob` (I don't use the value anywhere by myself), both of which expect `batch_size` to be in config. I can make kinda reassignment, but... does it make much sense?

tests/load/bigquery/test_bigquery_streaming_insert.py
IlyaFaer: Done

tests/load/bigquery/test_bigquery_streaming_insert.py
IlyaFaer: Done

dlt/destinations/impl/bigquery/configuration.py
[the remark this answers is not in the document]
rudolfix: OK you are right!

rudolfix (changes requested):
code is good! please update bigquery docs.
1. the BigQuery adapter part
2. rewrite
```
Data Loading
```
chapter to add information on streaming writes. please mention that only append mode works and that there is a buffer etc.

rudolfix (dismissed):
LGTM!

PR #1127, opened 2024-03-21
fixes column order and add hints table variants

### LINKED ISSUE #1105: allow dynamic table names for items yielded with `with_hints`
### Feature description

`dlt` allows to pass resource hints with the item to extract pipeline. When item is processed, the hints are merged into the table schema. Currently we do not allow to dispatch schema changes to different tables and they will accumulate in a resource hints.
This feature will use actual item and dynamic table name to create a variant of the hints that will compute tables later.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

I have a data source that processes a stream and want to dynamically create model for columns depending on event type. I want to use `def with_hints(item: TDataItems, hints: TResourceHints) -> DataItemWithMeta:` marker to pass columns bound to a data item.

* I define the table name to be based on item type
* I pass item on that type with column hints
* I want those hints to apply only to event of that type

### Proposed solution

When applying hints at run time and if table name is dynamic, create a copy if hints for that table that will be then returned in `compute_table_schema`

### Related issues

_No response_

### LINKED ISSUE #1111: Column hints break column order
### dlt version

0.4.5

### Describe the problem

Providing a column hint leads to that column being placed first in the schema. The column order as provided by the source is not respected.

Example: column `c3` is placed first in the schema, while it should come after `c1` and `c2` to respect the source column order.

image

### Expected behavior

The source column order is respected, also when hints are provided.

### Steps to reproduce

```python
import dlt

@dlt.resource(columns={"c3": {"precision": 32}})
def items():
    yield {"c1": 1, "c2": 1, "c3": 1}

    p = dlt.pipeline(pipeline_name="column_order", destination="duckdb", full_refresh=True)
    p.run(items)
    print(p.default_schema.get_table_columns("items"))
```

### Operating system

Windows

### Runtime environment

Local

### Python version

3.8

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
1. fixes #1111 
2. fixes #1105 

docs tbd.

creation hints variant per table happens as follows:
```
def test_mark_hints_with_variant() -> None:
    @dlt.resource
    def with_table_hints():
        # create variant for table a,  item dispatch to table a
        yield dlt.mark.with_hints(
            {"id": 1, "pk": "A"},
            dlt.mark.make_hints(
                table_name="table_a", columns=[{"name": "id", "data_type": "bigint"}]
            ),
            create_table_variant=True,
        )
     # then you can dispatch subsequent items like
     yield dlt.mark.with_table_name({"id": 4, "pk": "D"}, "table_a")
```

### DIFF
--- a/dlt/extract/hints.py
+++ b/dlt/extract/hints.py
@@ -44,10 +44,7 @@ class TResourceHints(TypedDict, total=False):
 
 
 class HintsMeta:
-    __slots__ = ["hints", "create_table_variant"]
-
-    hints: TResourceHints
-    create_table_variant: bool
+    __slots__ = ("hints", "create_table_variant")
 
     def __init__(self, hints: TResourceHints, create_table_variant: bool) -> None:
         self.hints = hints
@@ -169,7 +166,6 @@ class DltResourceHints:
         table_template = self._clone_hints(table_template)
         if "name" not in table_template:
             table_template["name"] = self.name
-        # table_template["columns"] = copy(self._hints["columns"])
 
         # if table template present and has dynamic hints, the data item must be provided.
         if self._table_name_hint_fun and item is None:
--- a/dlt/extract/hints.py
+++ b/dlt/extract/hints.py
@@ -12,7 +12,8 @@ from dlt.common.schema.typing import (
     TTableFormat,
     TSchemaContract,
 )
-from dlt.common.schema.utils import DEFAULT_WRITE_DISPOSITION, merge_columns, new_column, new_table
+from dlt.common import logger
+from dlt.common.schema.utils import DEFAULT_WRITE_DISPOSITION, merge_column, new_column, new_table
 from dlt.common.typing import TDataItem, DictStrAny, DictStrStr
 from dlt.common.utils import update_dict_nested
 from dlt.common.validation import validate_dict_ignoring_xkeys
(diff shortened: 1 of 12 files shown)


dlt/extract/hints.py
```diff
@@ -143,16 +152,24 @@ def columns(self) -> TTableHintTemplate[TTableSchemaColumns]:
...
+        if not table_template:
             return new_table(self.name, resource=self.name)
 
         # resolve a copy of a held template
-        table_template = copy(self._hints)
+        table_template = self._clone_hints(table_template)
         if "name" not in table_template:
             table_template["name"] = self.name
-        table_template["columns"] = copy(self._hints["columns"])
+        # table_template["columns"] = copy(self._hints["columns"])
```
jorritsandbrink, line 172: I think this commented line can be deleted.

dlt/extract/items.py
```diff
@@ -81,20 +81,20 @@ class SourcePipeItem(NamedTuple):
 
 
 class DataItemWithMeta:
-    __slots__ = "meta", "data"
+    __slots__ = ["meta", "data"]
 
-    meta: Any
-    data: TDataItems
+    # meta: Any
```
jorritsandbrink, line 86: Why are these commented? Same question for `table_name` on `TableNameMeta`.

jorritsandbrink (commented):
Besides two minor comments, LGTM!

dlt/extract/items.py
rudolfix: those are slot classes: https://wiki.python.org/moin/UsingSlots

I was using them incorrectly. this still needs a cleanup. ie. use tuples instead of lists

PR #1130, opened 2024-03-22
docs: add typechecking to embedded snippets

sh-rp:
### Description
This PR adds typechecking to snippets embedded in docs. Some small fixes to the existing snippets were required.

We are ignoring a bunch of errors, and we are also not checking out the verified sources for typechecking which would improve this a lot. All unknown vars are ignore, I have tried to find the most common ones and have added them to template.py


AstrakhantsevaAA (approved):
Amazing!

PR #1132, opened 2024-03-22
Fix typo with switched column names in schema evolution docs page

b-per:
### Description
Fixes a small typo in the docs on [the schema evolution page](https://dlthub.com/docs/general-usage/schema-evolution).
The new field is `main_block` and the old one is `building`


burnash:
@b-per thanks for spotting this and for the fix!

burnash (approved):

PR #1139, opened 2024-03-23
Add docs component with onboarding call

VioletM:
Update the onboarding link

### DIFF
--- a/docs/website/docs/_book-onboarding-call.md
+++ b/docs/website/docs/_book-onboarding-call.md
@@ -1,0 +1,1 @@
+<a href="https://calendar.app.google/EMZRS6YhM11zTGQw7">book a call</a> with our support engineer Violetta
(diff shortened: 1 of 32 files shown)


sh-rp:
@VioletM we have this docs preprocessor now which enables us to inject stuff into doc pages on compile time. If there are certain patterns for when and where these headers get added, we can also do it in this step to keep marketing stuff out of the GitHub docs. I'm not sure if we need to do this, but maybe you can keep that in mind.

VioletM:
@sh-rp 

For me, it is more or less the same because any change will go through PR anyway (or am I wrong?), I mostly care, that it's one change instead of 20. If the dev team thinks separation is needed, let's put it to the docs preprocessing.

Currently, the pattern where headers get added -- at the beginning of each source and at the central destination page.

Issue #930, opened 2024-02-02
Extended filesystem layout placeholders for destinations


sultaniman:
## Problem statement

There are already few requests to extend the supported variable placeholders for `layout` option, which has the following default value

```toml
[destination.filesystem]
layout="{table_name}/{load_id}.{file_id}.{ext}"
```

There are already couple of requests on GitHub [1](https://github.com/dlt-hub/dlt/issues/759), [2](https://github.com/dlt-hub/dlt/issues/555) and [slack](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1706533824819229). This means users would love to have more fine grained control for layout formatting options using more placeholders.

### Request [#1](https://github.com/dlt-hub/dlt/issues/759)

> `curr_year`, `curr_month`, `curr_day`, `curr_hour` and `curr_minute`
> 

### Request [#2](https://github.com/dlt-hub/dlt/issues/555)

> Partition config should support resource parameters, for example, if `date` (or `year, month, date`) is a parameter to a resource, that should be available in the partition layout.
> 

> Partition config should support resource metadata, such as *object path* or *creation date,* for example, in order to replicate the partitioning from the source to destination.
> 

### Request [#3](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1706533824819229)

> For us we would also like to be able to partition based on date for which the job is running. Very similar to airflow's [ds field](https://airflow.apache.org/docs/apache-airflow/stable/templates-ref.html)
> 

> So we can have hive style partitions `s3:///..../year={job_run.year}/month={job_run.month/day={job_run.day}/table/schema/...`
> 

> It would also be nice to be able to pass in these kind of fields as parameters to the dlt operator on airflow.
> 

## TODO

We need to augment [LoadFilesystemJob.make_destination_filename](https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/filesystem/filesystem.py#L54) and expand the the available parameters with datetime fields for `year, month, day, hour, minute, second, microsecond` 

Extending `[SUPPORTED_PLACEHOLDERS](https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/path_utils.py#L10)` with known set of variables should suffice the minimum use case needs currently it has the following values

`SUPPORTED_PLACEHOLDERS = {"schema_name", "table_name", "load_id", "file_id", "ext", "curr_date"}`

After the extension this will look like

[...]

PR #1149, opened 2024-03-26
Extend layout placeholder params for filesystem destinations

sultaniman:
This PR addresses https://github.com/dlt-hub/dlt/issues/930 

Requirements:

* [x] Make sure you can place something after the extension,
* Should be able to pass custom placeholders dictionary to fs and merge w/ existing placeholders,
	* [ ] if placeholder is not known then show a warning and skip,
	* [x] add validation for placeholders if they are defined with user friendly messages,
	* [x] support custom current date formatting placeholders (interpolation options on top of pendulum or ISO standard)
* [x] Placeholders in dictionary can be values or callbacks,
* [x] Make configurable via config toml,
* [x] We would like to have overridable current date to filesystem arguments

@sh-rp recommends to test with staging destination
`you’ll have to test it, write a test where you run the destination settings that use staging and add a map function for the filename`

## TODO
* [ ] Tests,
* [ ] Docs update


dlt/destinations/impl/filesystem/factory.py
```diff
@@ -46,12 +52,27 @@ def __init__(
...
         """
+        destinatination_kwargs: t.Dict[str, t.Any] = {}
+        if current_datetime:
+            destinatination_kwargs["current_datetime"] = current_datetime
+
+        if datetime_format:
+            destinatination_kwargs["datetime_format"] = datetime_format
+
+        if extra_params:
+            destinatination_kwargs["extra_params"] = extra_params
```
sultaniman, line 69: I am passing these three arguments via `kwargs` so [here](https://github.com/dlt-hub/dlt/pull/1149/files#diff-650cdec3d48ccec34c894eb426306aeabc9dccae1d3872850666abe396d53024R64) on resolve I reconcile and can override the values from `config.toml`.
It doesn't feel right however is it fine @sh-rp @rudolfix ?

dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -1,19 +1,71 @@
...
 
     @resolve_type("credentials")
     def resolve_credentials_type(self) -> Type[CredentialsConfiguration]:
         # use known credentials or empty credentials for unknown protocol
-        return self.PROTOCOL_CREDENTIALS.get(self.protocol) or Optional[CredentialsConfiguration]  # type: ignore[return-value]
+        return (
+            self.PROTOCOL_CREDENTIALS.get(self.protocol) or Optional[CredentialsConfiguration]  # type: ignore[return-value]
+        )
+
+    def on_resolved(self) -> None:
```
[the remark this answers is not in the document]
rudolfix: OK you need it to validate the layout. at this moment you know all the config values. some come from providers some from factory,

so here you need to validate the layout. see if all placeholders are known. if not we need a really good explanation which are missing etc.

dlt/destinations/impl/filesystem/layout.py
```diff
@@ -0,0 +1,250 @@
...
+    "timestamp",
+}
+
+SUPPORTED_TABLE_NAME_PREFIX_PLACEHOLDERS = ("schema_name",)
+
+
+class ExtraParams:
+    def __init__(
+        self,
+        config: FilesystemDestinationClientConfiguration,
```
rudolfix, line 37: it would be good if we have no dependencies on config in this module. just pass layout and extra_placeholders directly

dlt/destinations/impl/filesystem/layout.py
```diff
@@ -0,0 +1,250 @@
...
+
+        # If current_datetime is not given
+        # Then take current moment in time
+        # If current_datetime is callable
+        # Then call it and check it's instance
+        # If the result id DateTime
+        # Then take it
+        # Else exit.
+        if not self.config.current_datetime:
+            self.config.current_datetime = pendulum.now()
```
rudolfix, line 87: ok this is tricky! we should set a current_datetime only once per instance of factory. so imo you should override `configuration` method and set the `current_datetime` if it is none from a timestamp that is created in factory `__init__`

dlt/destinations/impl/filesystem/layout.py
```diff
@@ -0,0 +1,250 @@
...
+
+        # if extension is not defined, we append it at the end
+        if "ext" not in placeholders:
+            path += f".{job_info.file_format}"
+
+        return path
+
+
+def get_table_prefix_layout(
+    config: FilesystemDestinationClientConfiguration,
```
rudolfix, line 208: hopefully this will merge into one nice path_utils or layout.py :)

dlt/destinations/impl/filesystem/layout.py
```diff
@@ -0,0 +1,250 @@
...
+class layout_helper:
+    def __init__(
+        self,
+        path_layout: str,
+        params: Dict[str, str],
+        allowed_placeholders: Optional[Set[str]] = SUPPORTED_PLACEHOLDERS,
+    ) -> None:
+        self.params = params
+        self.allowed_placeholders = allowed_placeholders.copy()
+        self.layout_placeholders = re.findall(r"\{(.*?)\}", path_layout)
```
rudolfix, line 148: what's wrong with using `path_utils` method?

dlt/destinations/impl/filesystem/layout.py
```diff
@@ -0,0 +1,250 @@
...
+            self._params["curr_date"] = now.format(self.config.datetime_format)
+        else:
+            self._params["curr_date"] = str(now.date())
+
+        # For each callable extra parameter
+        # otherwise take it's value
+        if self.config.extra_params:
+            for key, value in self.config.extra_params.items():
+                if callable(value):
+                    self._params[key] = value(
```
rudolfix, line 111: here you must catch TypeError which indicates a wrong signature and explain what the correct signature is. otherwise people will be confused

ideally you could inspect all the signatures when validating placeholders

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -36,6 +34,13 @@ def logger_autouse() -> None:
     "{schema_name}/{table_name}/{load_id}.{file_id}.{ext}",  # new default layout with schema
     "{schema_name}.{table_name}.{load_id}.{file_id}.{ext}",  # classic layout
     "{table_name}88{load_id}-u-{file_id}.{ext}",  # default layout with strange separators
+    # Extra layout options
+    "{table_name}/{curr_date}/{load_id}.{file_id}.{ext}{timestamp}",
+    "{table_name}/{year}/{month}/{day}/{load_id}.{file_id}.{ext}",
+    "{table_name}/{day}/{hour}/{minute}/{load_id}.{file_id}.{ext}",
+    "{table_name}/{timestamp}/{load_id}.{file_id}.{ext}",
+    "{table_name}/{dow}/{load_id}.{file_id}.{ext}",
```
rudolfix, line 42: also add
`{YY}/{MM}/{DD}/{table_name}/...` and see where it fails

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -94,13 +93,17 @@ class FilesystemClient(JobClientBase, WithStagingDataset):
...
+        schema: Schema,
+        config: FilesystemDestinationClientConfiguration,
+    ) -> None:
         super().__init__(schema, config)
         self.fs_client, self.fs_path = fsspec_from_config(config)
         self.config: FilesystemDestinationClientConfiguration = config
         # verify files layout. we need {table_name} and only allow {schema_name} before it, otherwise tables
         # cannot be replaced and we cannot initialize folders consistently
-        self.table_prefix_layout = path_utils.get_table_prefix_layout(config.layout)
+        self.table_prefix_layout = get_table_prefix_layout(config)
```
rudolfix, line 106: I'll comment on that in next review. this is a bigger chunk. we want to support all paths in append-only and require old layout only when we replace data (we can't guarantee replacement otherwise)

rudolfix (changes requested):

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -38,36 +43,28 @@ def __init__(
...
 
         super().__init__(file_name)
         fs_client, _ = fsspec_from_config(config)
-        self.destination_file_name = LoadFilesystemJob.make_destination_filename(
-            config.layout, file_name, schema_name, load_id
-        )
+        self.destination_file_name = make_filename(config, file_name, schema_name, load_id)
+
+        # Make sure directory exists before moving file
+        dir_path = self.make_remote_path(only_dir=True)
```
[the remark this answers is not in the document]
sultaniman: @rudolfix the strange thing was that ffspec didn't create directory that's why I added it, I will remove this and see if test pipeline will work out.

dlt/destinations/impl/filesystem/layout.py
sultaniman: Got it

dlt/destinations/impl/filesystem/factory.py
sultaniman: Done.

sultaniman:
Closing this in favor of https://github.com/dlt-hub/dlt/pull/1182

PR #1141, opened 2024-03-24
Add RESTClient: 

burnash:
### Description

This pull request introduces a `RESTClient` class along with supporting components to interact with RESTful APIs. The RESTClient provides:
- an interface for making HTTP requests, 
- handling pagination (via a set of paginator classes: e.g. `HeaderLinkPaginator`, `JSONResponsePaginator`.
- managing authentication (via `BearerTokenAuth`, `HttpBasicAuth`, etc.)

Users are able to customize request handling, response parsing, and data extraction according to the specifics of the target API. `RESTClient` also features automatic paginator detection with the `PaginatorFactory`

Example usage:

```python
import dlt
from dlt.sources.helpers.rest_client import RESTClient

github_client = RESTClient(base_url="https://api.github.com")

@dlt.resource(
    table_name="issues",
    write_disposition="merge",
    primary_key="id",
)
def get_issues(
    updated_at=dlt.sources.incremental(
        "updated_at", initial_value="1970-01-01T00:00:00Z"
    )
):
    for page in github_client.paginate(
        "/repos/dlt-hub/dlt/issues",
        params={
            "since": updated_at.last_value,
            "per_page": 100,
            "sort": "updated",
            "direction": "desc",
            "state": "open",
        },
    ):
        yield page


# The rest of the pipeline remains the same
pipeline = dlt.pipeline(
    pipeline_name="github_issues_merge",
    destination="duckdb",
    dataset_name="github_data_merge",
)
load_info = pipeline.run(get_issues)
print(load_info)
```

### Related Issues

- This is an extract from the `rest_api` source from https://github.com/dlt-hub/verified-sources/pull/313

PR #1151, opened 2024-03-26
Add a topic guide for the `RESTClient` class

burnash:
### Description

This PR adds documentation for the `RESTClient` class, including a brief example and instructions on how to handle paginated endpoints in APIs.

### Related Issues

- #1141


docs/website/docs/general-usage/rest-client.md
```diff
@@ -0,0 +1,259 @@
+# Reading data from RESTful APIs
+
+RESTful APIs are a common way to interact with web services. They are based on the HTTP protocol and are used to read and write data from and to a web server. dlt provides a simple way to read data from RESTful APIs using two helper methods: [a wrapper around the `requests`](../reference/performance#using-the-built-in-requests-client) library and a `RESTClient` class.
```
VioletM, line 3: I would suggest splitting this in two pages: 
1) RESTClient -- takes care of pagination, authentification, customization
2) requests module (migrate stuff from performance to separate page) -- takes care of retries

Maybe even abstract the requests helper and make retry customization available through RESTClient so users have one entry point.

docs/website/docs/general-usage/rest-client.md
```diff
@@ -0,0 +1,259 @@
...
+The `RESTClient` class offers a powerful interface for interacting with RESTful APIs, supporting features like
+- automatic pagination,
+- various authentication mechanisms,
+- customizable request/response handling.
+
+This guide demonstrates how to use the `RESTClient` class to read data APIs focusing on its `paginate()` method to fetch data from paginated API responses.
+
+## Quick example
+
+Here's a simple pipeline that reads issues from the [dlt GitHub repository](https://github.com/dlt-hub/dlt/issues). The API endpoint is `https://api.github.com/repos/dlt-hub/dlt/issues`. The result is "paginated", meaning that the API returns a limited number of issues per page. The `paginate()` iterates over all pages and yields the results which are then processed by the pipeline.
```
VioletM, line 19: I would add that in this example authentification and pagination are guessed automatically. And after this example put another one that shows how you can specify the paginator and auth if they are special/not guessed correctly.

Honestly, these two examples would be enough to give users an understanding of how to use RESTClient.

So overall I suggest:
* Quick example simple
* Quick example with specific paginator/auth
* Available Paginators
* Available Auths
* How to write my own if nothing fits

In this case user WOW with this doc would be: 
1) Copy-past code which guesses everything automatically 
2) If it doesn't work, copy-past code with a specification of pagination and auth.
3) Search in docs which auth and paginator is right for you
4) If nothing fits -> How-to write custom auth/paginator.

rudolfix (approved):
LGTM!

@burnash if you want to top up this then add add an example with incremental loading. this will be the most common "advanced" task people have

it maybe a separate ticket - this is really good

burnash:
> one topic is really missing

@rudolfix which one?

rudolfix (changes requested):
We do not explain `data_selector`. It requires a separate chapter because it does a lot of detection magic. this must be explained - at least briefly

IMO we should do a followup ticket on this:
- I agree with @VioletM - we should explain the pagination detection better and say when it works (links paginator, next field at known places) and when it does not (offset pagination using non standard names) and what happens if we cannot detect.
- more in depth on `data_selector` response detections
- I'd add an example with incremental loading. this will be the most common "advanced" task people have

burnash:
@VioletM @rudolfix please take a look at the reworked version. I've covered missing sections and moved the docs for requests wrapper to this guide. I agree with @VioletM we may consider going even further and streamlining the code so there's one entry point. Open for discussion.

docs/website/docs/general-usage/http/rest-client.md
```diff
@@ -0,0 +1,413 @@
...
+## Paginating API responses
+
+The `RESTClient.paginate()` method is specifically designed to handle paginated responses, yielding `PageData` instances for each page:
+
+```py
+for page in client.paginate("/posts"):
+    print(page)
+```
+
+If `paginator` is not specified, the `paginate()` method will attempt to automatically detect the pagination mechanism used by the API. If the API uses a standard pagination mechanism like having a `next` link in the response's headers or JSON body, the `paginate()` method will handle this automatically. Otherwise, you can specify a paginator object explicitly or implement a custom paginator.
```
VioletM, line 56:
```suggestion
:::tip
If `paginator` is not specified, the `paginate()` method will attempt to automatically detect the pagination mechanism used by the API. If the API uses a standard pagination mechanism like having a `next` link in the response's headers or JSON body, the `paginate()` method will handle this automatically. Otherwise, you can specify a paginator object explicitly or implement a custom paginator.
:::
```

docs/website/docs/general-usage/http/rest-client.md
```diff
@@ -0,0 +1,413 @@
...
+- [OffsetPaginator](#offsetpaginator) - pagination based on offset and limit query parameters.
+- [JSONResponseCursorPaginator](#jsonresponsecursorpaginator) - pagination based on a cursor in the JSON response.
+
+If the API uses a non-standard pagination, you can [implement a custom paginator](#implementing-a-custom-paginator) by subclassing the `BasePaginator` class.
+
+#### JSONResponsePaginator
+
+`JSONResponsePaginator` is designed for APIs where the next page URL is included in the response's JSON body. This paginator uses a JSON path to locate the next page URL within the JSON response.
+
+##### Parameters:
```
VioletM, line 116: NIT:
The 5th and 6th headings looks very strange, smaller than a regular text :D 
Maybe just use bold?
image

VioletM (dismissed):
Looks great! Everything is explained with examples, this is a very good quality docs :)

docs/website/docs/general-usage/http/rest-client.md
burnash: Thanks for suggestion, @VioletM

docs/website/docs/general-usage/http/rest-client.md
burnash: Good point

Issue #1002, opened 2024-02-26
tuple: pivot as a built in transformations


VioletM:
### Feature description

 I want to be able to declare a pivot transformation in a resource if a column contains tuples

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

A response from API contains tabular data in form of list of lists. The nested list is always of the same length (a tuple) and in fact represents a table.
Another good example could be panda frame as list of lists:
```
df = pd.DataFrame({
    'A': [1, 2],
    'B': [3, 4],
    'C': [5, 6]
})

list_of_lists = df.values.tolist()
print(list_of_lists)
```

dlt will represent such lists as child tables with single `value` column. What we want to achieve is to convert list of lists to list of dicts where keys are col1, col2 .. col5 and values are elements of a list
ie. `[[1, 3, 5], [2, 4, 6]]` -> `[{"col1": 1, "col2": 3, "col3": 5}, ...]

### Proposed solution

1. we want to implement this as a transform function added with `add_map`. we keep standard transform functions in `sources.helpers.transforms`
2. you need to implement a function that returns a configured map function. the config arguments are: a single or a list of strings representing jsonpath and a column prefix
3. when the map function receives it should find all the matching elements, make sure it is a list of lists and then pivot internal lists creating columns with a given prefix
4. please write tests and docstrings

PR #1152, opened 2024-03-27
feat(transform): implement columns pivot map function

IlyaFaer:
Closes https://github.com/dlt-hub/dlt/issues/1002


dlt/sources/helpers/transform.py
```diff
@@ -24,3 +26,31 @@ def _filter(_: TDataItem) -> bool:
...
+    def _transformer(item: TDataItem) -> TDataItem:
+        """Pivot columns into a dictionary.
+
+        Args:
+            item (TDataItem): a data item.
+
+        Returns:
+            TDataItem: a data item with pivoted columns.
+        """
+        return {prefix + col: item[ind] for ind, col in enumerate(columns)}
```
rudolfix, line 54: when `columns` are string (you allow that in the Union) it will enumerate characters. also you didn't write a test for that

dlt/sources/helpers/transform.py
```diff
@@ -24,3 +26,31 @@ def _filter(_: TDataItem) -> bool:
         return count > max_items
 
     return _filter
+
+
+def pivot(columns: Union[str, List[str]], prefix: str) -> ItemTransformFunctionNoMeta[TDataItem]:
+    """Pivot a list of columns into a dictionary.
+
+    Args:
+        columns (Union[str, List[str]]): list of column names
```
rudolfix, line 35: pls check out https://github.com/dlt-hub/dlt/issues/1002 the proposed solution. the arguments are
- json path to the element(s) containing lists of lists, let's put default "$" to select the root. also we have jsonpath_ng in our deps so please use it (jsonpath.py in common)
- a prefix for column prefix. so columns start with prefix and have an index as suffix. we can also accept a list of columns like the above. but then you should make sure that len of the list is exactly the number of columns

rudolfix (changes requested):

dlt/sources/helpers/transform.py
IlyaFaer: On lines 42-43 I cast `str` to `list`, so it'll work. Added a test however.

rudolfix (changes requested):
OK now the implementation is good. tests look OK as well. just two things:
1. please create a new test module for transform tests. extract/test_transform.py and move tests there
2. please also write a test for @VioletM case where we have a panda frame producing a list of lists and we feed it into the pivot function without any arguments (it should transform the root element)

ahhhh and
3. please set the default prefix to "col" :)

IlyaFaer:
@rudolfix, okay, done. One thing that concerns me is `$`. In most of the databases it has special meaning. We'll not get any problems by returning keys with it?

tests/extract/test_transform.py
```diff
@@ -0,0 +1,134 @@
...
+    import pandas
+
+    df = pandas.DataFrame({"A": [1, 2], "B": [3, 4], "C": [5, 6]})
+
+    yield df.values.tolist()
+
+
+def test_pandas_dataframe_pivot() -> None:
+    res = pivot_test_pandas_resource()
+    res.add_map(pivot())
```
rudolfix, line 128: 0to pivot the example above you need to `pivot("*")` that will select all children of $ (A,B,C)

we need one more test where you pivot "$" where item is just list of lists

dlt/sources/helpers/transform.py
```diff
@@ -24,3 +29,86 @@ def _filter(_: TDataItem) -> bool:
...
+            item (TDataItem): The data item to transform.
+
+        Returns:
+            TDataItem: the data item with pivoted columns.
+        """
+        trans_item: Dict[str, List[Any]] = {}
+
+        for path in paths:
+            expr = jsonpath_ng.parse(path)
+            for match in expr.find([item] if path == "$" else item):
```
rudolfix, line 103: this should not be necessary, for "$" to work, the data item must be a list of lists

rudolfix (changes requested):

dlt/sources/helpers/transform.py
```diff
@@ -24,3 +29,86 @@ def _filter(_: TDataItem) -> bool:
...
+                _raise_if_not_sequence(match)
+
+                f_path = str(match.full_path)
```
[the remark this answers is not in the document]
IlyaFaer: Done

tests/extract/test_transform.py
IlyaFaer: When we do `tolist()`, it becomes `[[1, 3, 5], [2, 4, 6]]` - a list of lists

rudolfix (approved):
LGTM!

PR #1155, opened 2024-03-27
Check for default schema and schema name in streamlit session

sultaniman:
This PR adds additional checks in case if pipeline ran but no data was loaded so we can show better error messages instead of showing a stack trace.


rudolfix (dismissed):

rudolfix:
@sultaniman is this fixing also this problem?
image

sultaniman:
Updated the streamlit app so it doesn't show annoying error message if there is no resource state in pipeline

rudolfix:
@sultaniman pls post a pic with the current UI :)

sultaniman:
Yessir!

PR #1159, opened 2024-03-28
ci improvements

sh-rp:
### Description
This PR will include improvements to our ci.

NOTE: I can't see the athena iceberg tests being executed, I don't get what is going on there..

## Details
- Removes unneeded matrix job collectors for destination tests
- Cleans up naming of workflows and job to produce more legible output
- Only runs tests marked as "essential" for most destination on PRs
- Add the possibility to run all tests with a "ci-full" tag on the PR
- Add the possibility to run tests on a fork if tag "ci from fork" is defined
- there is a nightly schedule at 2am where the full test suite is run

## To Do
- Update the required tests on the repo to not point to the matrix collection tests

## Other fixes
- Snippets tests also need secrets, so they are now also prevented to run on forks by default

## To discuss
- Local tests require the secrets, although they shouldn't really. there are some filesystem tests that use gdrive. For now they are also not disabled on forks. we should talk about it.

## Ideas
- Run a simple load for all destinations on all OSes to see that the drivers work everywhere?


.github/workflows/test_common.yml
```diff
@@ -1,4 +1,4 @@
-name: test common
+name: common | common
```
sultaniman, line 1: maybe just name it common?

.github/workflows/test_destination_snowflake.yml
```diff
@@ -71,22 +69,14 @@ jobs:
         run: pwd && echo "$DLT_SECRETS_TOML" > tests/.dlt/secrets.toml
 
       - run: |
-          poetry run pytest tests/load
-        if: runner.os != 'Windows'
-        name: Run tests Linux/MAC
+          poetry run pytest tests/load -m "essential"
+        name: Run essential tests Linux
+        if: ${{ ! (contains(github.event.pull_request.labels.*.name, 'ci full') || github.event_name == 'schedule')}}
```
sultaniman, line 74: will this negate the final result of if condition?

sh-rp: yes

.github/workflows/test_common.yml
sh-rp: I'm trying to keep the same format across all files

.github/workflows/test_destination_qdrant.yml
```diff
@@ -22,22 +24,18 @@ env:
...
 jobs:
   get_docs_changes:
+    name: docs changes
     uses: ./.github/workflows/get_docs_changes.yml
-    if: ${{ !github.event.pull_request.head.repo.fork }}
+    if: ${{ !github.event.pull_request.head.repo.fork || contains(github.event.pull_request.labels.*.name, 'ci from fork')}}
 
   run_loader:
-    name: Tests Qdrant loader
+    name: test ß
```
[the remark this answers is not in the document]
sh-rp: ah thanks!

rudolfix (changes requested):
1. you can mark whole modules: https://docs.pytest.org/en/7.1.x/example/markers.html#marking-whole-classes-or-modules
2. I'd also test
- test_job_client
- test_insert_job_client
3. load/pipelines/test_pipeline.py
4. load/pipelines/test_arrow_loading

up to you but I would mark all destination specific tests as essential
- all folders in tests/load/... testing specific destinations
- same for tests/load/pipelines

sh-rp:
note: make mssql run fully each time

rudolfix (approved):
this is so good! @sh-rp problems with athena and synapse were found and fixed in the other branches

Issue #1129, opened 2024-03-22
Add `upsert` merge strategy


rudolfix:
**Background**
`dlt` uses DELETE + INSERT to merge staging dataset into destination. It will also create temporary tables with _dlt_id of rows to be deleted and inserted. This may all not be necessary if MERGE sql is used. On supported database backend, MERGE looks even more standardized that various forms of DELETE that we had to hack.

**Tasks**
1. * [ ] look at replace strategies and add merge strategies to config
2. * [ ] pick a right merge job depending on merge strategy
3. * [ ] make sure the deduplication of the staging datasets by primary key also takes place (hope MERGE will do it automatically)
4. * [ ] which destinations to convert: duckdb (testing), bigquery, snowflake. postgres should work if duckdb works
5. * [ ] tests and docs!

jorritsandbrink:
My take, based on previous work in https://github.com/dlt-hub/dlt/pull/1294:

Add a new key-based merge strategy called `upsert`:
- if key is not present in destination table ➜ insert record
- if key is present in destination table ➜ update record

Source data can be both full or incremental extracts.

Constraints:
- needs a `primary_key`
- does not support `merge_key` (we need a one-to-one relationship between records in source and destination)

Furthermore, as with `delete-insert`:
- do primary key-based deduplication
- support `dedup_sort` and `hard_delete` column hints

Implementation:
- do real updates as much as possible, i.e. no delete+insert mechanism
  - use the `MERGE` command to implement it for SQL destinations that support it
  - use `INSERT INTO ... ON CONFLICT DO UPDATE` on DuckDB ([reference](https://duckdb.org/docs/sql/statements/insert.html#do-update-clause-upsert))
  - use the [TableMerger API](https://delta-io.github.io/delta-rs/api/delta_table/delta_table_merger/) for `delta` table format on `filesystem` destination (implement in separate PR)
- child tables still need deletes, to delete records for elements which have been removed from an updated list
- hash primary key and store as `_dlt_id` in root table
  - use this as upsert key

rudolfix:
@jorritsandbrink my biggest question to the above: what you can do with it that you cannot do with the current merge? can you update from partial records? (on first look we'd need to add something on top to do that)

I had something more modest in mind:
1. replace INSERT / DELETE with MERGE wherever we can and make it somehow optional - but this is just internal change without impact on the UX
2. implement existing merge strategy for delta (using their TableMerger API whenever possible, INSERT DELETE otherwise)

[...]

PR #1168, opened 2024-04-01
SCD2 support

### LINKED ISSUE #828: SCD2 write disposition

jorritsandbrink:
### Description
This PR adds basic support for SCD2. Scope is defined [here](https://github.com/dlt-hub/dlt/issues/828#issuecomment-2016791397).

- Treats `scd2` as a special case of the `merge` write disposition.
- Support configurable names for the validity columns ("valid from" / "valid to").
- Validity column names default to `_dlt_valid_from` and `_dlt_valid_to` if the user does not specify names.
- Uses timestamps (not dates) to support sub-daily changes.
- Uses high timestamp (`9999-12-31T00:00:00+00:00`) in "valid to" column to indicate active records.
- Includes validity columns only in the root table. Validity dates for records in child tables can be linked through the root id.
- Does **not** use "standard" [configuration](https://dlthub.com/docs/general-usage/credentials/configuration) because it doesn't support resource-scoped configuration (correct me if I'm wrong). We want resource-scoped configuration because a user might want to load one table with the `scd2` strategy and another one with "regular" merge behavior, or use different validity column names for different tables.
- ~~Introduces `merge_config` argument on the `resource` decorator~~ Extends `write_disposition` argument such that it also accepts configuration dictionaries to let the user specify the merge strategy and validity column names..
- ~~Adds `merge_config` key to `TResourceHints`, but not to `TTableSchema`.~~
- Uses x-hints to store user configuration in the schema:
  - `x-merge-strategy` at the table level
  - `x-valid-from` and `x-valid-to` at the column level

### Related Issues

- Closes #828


tests/load/pipeline/test_scd2.py
```diff
@@ -0,0 +1,370 @@
...
+from tests.pipeline.utils import assert_load_info
+from tests.load.pipeline.utils import (
+    destinations_configs,
+    DestinationTestConfiguration,
+    load_tables_to_dicts,
+)
+
+
+ACTIVE_TS = datetime.fromisoformat(HIGH_TS.isoformat()).replace(tzinfo=None)
+h = DataItemNormalizer.get_row_hash
```
sultaniman, line 23: maybe alias this to `get_row_hash`?

dlt/common/schema/typing.py
```diff
@@ -155,6 +155,14 @@ class NormalizerInfo(TypedDict, total=True):
     new_table: bool
 
 
+class TMergeConfig(TypedDict, total=False):
+    strategy: Optional[TLoaderMergeStrategy]
+    validity_column_names: Optional[List[str]]
```
sultaniman, line 160: Should it allow duplicates column names?

dlt/destinations/sql_jobs.py
```diff
@@ -478,3 +491,61 @@ def gen_merge_sql(
...
+            UPDATE {root_table_name} SET {to} = '{boundary_ts}'
+            WHERE NOT EXISTS (
+                SELECT s._dlt_id FROM {staging_root_table_name} AS s
+                WHERE {root_table_name}._dlt_id = s._dlt_id
+            ) AND {to} = '{active_record_ts}';
+        """)
+
+        # insert new active records in root table
+        columns = map(escape_id, list(root_table["columns"].keys()))
+        col_str = ", ".join([c for c in columns if c not in (from_, to)])
```
sultaniman, line 532: afaik there are requests to preserve column names and in the future we might want to support it, should we put column names inside quotes?

tests/load/pipeline/test_scd2.py
```diff
@@ -0,0 +1,370 @@
...
+
+from tests.pipeline.utils import assert_load_info
+from tests.load.pipeline.utils import (
+    destinations_configs,
+    DestinationTestConfiguration,
+    load_tables_to_dicts,
+)
+
+
+ACTIVE_TS = datetime.fromisoformat(HIGH_TS.isoformat()).replace(tzinfo=None)
```
sultaniman, line 22: Can't we just use `HIGHT_TS.replace(tzinfo=None)`?

dlt/destinations/sql_jobs.py
```diff
@@ -478,3 +491,61 @@ def gen_merge_sql(
...
+        boundary_ts = current_load_package()["state"]["created_at"]
+        active_record_ts = HIGH_TS.isoformat()
+
+        # retire updated and deleted records
+        sql.append(f"""
+            UPDATE {root_table_name} SET {to} = '{boundary_ts}'
+            WHERE NOT EXISTS (
+                SELECT s._dlt_id FROM {staging_root_table_name} AS s
+                WHERE {root_table_name}._dlt_id = s._dlt_id
+            ) AND {to} = '{active_record_ts}';
```
sultaniman, line 527: with this condition aren't we making a useless update?

dlt/destinations/sql_jobs.py
jorritsandbrink: I think we need to escape column names. Without escaping, for example, the column name `from` (a logical candidate name for the "valid from" column) would cause an error because it's a SQL keyword.

tests/load/pipeline/test_scd2.py
jorritsandbrink: `HIGH_TS` is a `pendulum` `DateTime` because the `datetime` module is a banned import. However, when fetching data from the destination tables to assert its content, timestamps are returned as `datetime` objects. Hence the conversion from `DateTime` to `datetime.`

tests/load/pipeline/test_scd2.py
jorritsandbrink: Changed in https://github.com/dlt-hub/dlt/pull/1168/commits/7726d98a5a412da45b6a6a1f9bc3ffb736b781cd

dlt/common/schema/typing.py
jorritsandbrink: It shouldn't, didn't think of that. Added validity column name checking in https://github.com/dlt-hub/dlt/pull/1168/commits/30bb2e0e9fff508a76c9cea56aed997c849cf377. An exception is raised if a configured validity column name appears in the data.

dlt/destinations/sql_jobs.py
sh-rp: yes this looks totally fine for me too

tests/load/pipeline/test_scd2.py
```diff
@@ -0,0 +1,399 @@
...
+    ]
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["duckdb"]),
+    ids=lambda x: x.name,
+)
+@pytest.mark.parametrize("simple", [True, False])
+def test_child_table(destination_config: DestinationTestConfiguration, simple: bool) -> None:
```
sh-rp, line 170: i think we should test the child table functionality on all destinations, so either enable all dests here or add a child table to the basic test. the reason is, that we should run all sql code that we have against all destinations, just to make sure it works everywhere.

sh-rp (changes requested):
Very nice work, thanks! Only a few small changes and we still need the docs. Also I want to discuss two points with @rudolfix before we can merge.

jorritsandbrink:
> @rudolfix two questions on this PR:
> 
> * Do you agree with the interface? We now have additional fields on the resource to configure this as opposed to a settings on the destination how we do it with the replace strategy. I think this makes sense, but also looks a bit weird provided the replace strategy is set in a different way.
> * Right now the "cutoff date" is the created date of the load package. I am thinking we might want to make this overridable with a setting for certain cases, but maybe it is good like this now and we can add this feature when people request it.

Note that this is why I chose a different interface than the one used to configure replace strategy:
image

dlt/destinations/sql_jobs.py
rudolfix: there's another PR that will handle that @sultaniman there will be proper escaping and normalization of column names

dlt/destinations/sql_jobs.py
```diff
@@ -478,3 +492,60 @@ def gen_merge_sql(
...
+        root_table_name = sql_client.make_qualified_table_name(root_table["name"])
+        with sql_client.with_staging_dataset(staging=True):
+            staging_root_table_name = sql_client.make_qualified_table_name(root_table["name"])
+
+        # get validity column names
+        escape_id = sql_client.capabilities.escape_identifier
+        from_, to = list(map(escape_id, get_validity_column_names(root_table)))
+
+        # define values for validity columns
+        boundary_ts = current_load_package()["state"]["created_at"]
```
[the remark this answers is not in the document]
rudolfix: looks good for now!

dlt/common/schema/typing.py
```diff
@@ -65,6 +64,7 @@
 ]
 """Known hints of a column used to declare hint regexes."""
 TWriteDisposition = Literal["skip", "append", "replace", "merge"]
+TLoaderMergeStrategy = Literal["scd2"]
```
rudolfix, line 67: we should have a default strategy which is "delete-insert". and we'll add one more `merge` to support MERGE sql statements #1129

dlt/extract/decorators.py
```diff
@@ -297,6 +297,7 @@ def resource(
     columns: TTableHintTemplate[TAnySchemaColumns] = None,
     primary_key: TTableHintTemplate[TColumnNames] = None,
     merge_key: TTableHintTemplate[TColumnNames] = None,
+    merge_config: TMergeConfig = None,
```
rudolfix, line 300: @jorritsandbrink @sh-rp maybe instead of doing this, we could extend `write_disposition` to
``
write_disposition: TTableHintTemplate[Union[TWriteDisposition, TMergeConfig, TReplaceConfig]] = None
```
exactly what we do with `schema_contract` below? then we can support short hand strings and full definitions. and still use the same parameter
in that case I'd rename TMergeConfig to TMergeDispositionConfig

**Review by rudolfix — CHANGES_REQUESTED** [review]
this is really good! @jorritsandbrink congrats on plugging the work into our normalizer and hints perfectly.

I think we can do the user interface better by extending `write_disposition` instead of adding a new argument. @sh-rp exactly what we did for data contracts with shorthand and full notation - see my review

also: I trust you that SQL transformation is good and properly tested :)

File: dlt/extract/decorators.py (hunk 9 above)
Comment (sh-rp) [reply dlt/extract/decorators.py:300]: and we also handle the replace strategies in there? (not in the pr but later)? i think it's a good idea

File: tests/load/pipeline/test_scd2.py (hunk 6 above)
Comment (jorritsandbrink) [reply tests/load/pipeline/test_scd2.py:173]: Done in https://github.com/dlt-hub/dlt/pull/1168/commits/11748a63ffee50d2f01eb80a6138444a56743980

File: dlt/common/schema/typing.py (hunk 8 above)
Comment (jorritsandbrink) [reply dlt/common/schema/typing.py]: Added in https://github.com/dlt-hub/dlt/pull/1168/commits/396ec59393f691573d68585befd99abcbb4fc5c8

File: dlt/extract/decorators.py (hunk 9 above)
Comment (jorritsandbrink) [reply dlt/extract/decorators.py]: Extended `write_disposition` in https://github.com/dlt-hub/dlt/pull/1168/commits/8f4d4cee6e59e37962963670bf2d23a1d1454e99 and https://github.com/dlt-hub/dlt/pull/1168/commits/c8d84d88a8fc8e221cb3b60c8930be77be67009f

**Comment by jorritsandbrink** [comment]
@rudolfix I adjusted the user interface according to your suggestion and introduced a default `delete-insert` merge strategy. WDYT?

I'll add docs after we've settled on the user interface.

**Review by rudolfix — CHANGES_REQUESTED** [review]
this is good! I have two optimizations that IMO are OK to add.
thx for adding apply_hints tests!

File: dlt/destinations/sql_jobs.py [hunk 10]
```diff
@@ -486,3 +499,60 @@ def gen_merge_sql(
...
+        active_record_ts = HIGH_TS.isoformat()
+
+        # retire updated and deleted records
+        sql.append(f"""
+            UPDATE {root_table_name} SET {to} = '{boundary_ts}'
+            WHERE NOT EXISTS (
+                SELECT s._dlt_id FROM {staging_root_table_name} AS s
+                WHERE {root_table_name}._dlt_id = s._dlt_id
```
rudolfix, line 533: let's add x-row-hash hint that you can attach to `_dlt_id` if it not present in the schema (same thing we do with `root_key`). why:
1. no dependency on particular name when generating sql but on hints
2. users can create their own hash during extraction and ie. just take into account 2-3 columns not all of them. also our default hash (via json dump) will be quite slow

rudolfix (changes requested):
@jorritsandbrink what should we do with `merge_key` here? should we warn/raise when it is specified or use it for something ie. to apply the scd2 to a subset of data defined by a merge_key?
in the second case it is still possible to do incremental ie. day by updated

dlt/destinations/sql_jobs.py
jorritsandbrink: We could handle three user input cases:
1. `merge_key` = `None` (no merge key) ➜ assign `x-row-hash` to `_dlt_id` and generate row hash including all resource columns (the ones not generated by `dlt`)
2. `merge_key` = `col_x` (simple merge key) ➜ assign `x-row-hash` to `col_x` and use random `_dlt_id`—this can be used as "Bring Your Own Hash / Surrogate Key"
3. `merge_key` = [`col_x`, `col_y`, `col_z`] (composite merge key) ➜ assign `x-row-hash` to `_dlt_id` and generate row hash including only `col_x`, `col_y`,  and `col_z`

I suppose we would need to restrict `merge_key` changes. If we don't, we would identify unchanged records as being updated because the way their hash was calculated changed.

dlt/common/normalizers/json/relational.py
```diff
@@ -296,10 +319,18 @@ def normalize_data_item(
         row = cast(TDataItemRowRoot, item)
         # identify load id if loaded data must be processed after loading incrementally
         row["_dlt_load_id"] = load_id
+        # determine if row hash should be used as dlt id
```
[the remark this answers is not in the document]
jorritsandbrink: Solved with caching as you suggested on Slack: https://github.com/dlt-hub/dlt/pull/1168/commits/6b243782f3f53dc8fb6e8632795073c4dc110dcb

dlt/extract/hints.py
```diff
@@ -391,17 +399,56 @@ def _merge_key(hint: TColumnProp, keys: TColumnNames, partial: TPartialTableSche
...
+                    "data_type": "timestamp",
+                    "x-valid-from": True,
+                }
+                dict_["columns"][to] = {"name": to, "data_type": "timestamp", "x-valid-to": True}
+
+    @staticmethod
+    def _create_table_schema(resource_hints: TResourceHints, resource_name: str) -> TTableSchema:
+        """Creates table schema from resource hints and resource name."""
+
+        dict_ = cast(Dict[str, Any], deepcopy(resource_hints))
```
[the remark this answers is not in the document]
jorritsandbrink: `deepcopy` was indeed obsolete—removed it in https://github.com/dlt-hub/dlt/pull/1168/commits/1f399bcf5c306c036e643ce82d291755974737b3

dlt/destinations/sql_jobs.py
jorritsandbrink: Added `row_hash_column_name` configuration option in https://github.com/dlt-hub/dlt/pull/1168/commits/4124d61574d87c42230c33913d1373603af527f5. Translates to `x-row-hash` hint behind the scenes. With this, users can bring their own hash. As discussed on Slack, let's move other extensions involving `merge_key` to follow up PRs.

rudolfix (dismissed):
LGTM!

oh `dremio` does not like ZULU timestamps in sql merge job... do you generate them for SCD2? I think Z timestamps are not really part of ISO

we have a dremio container in tests if you want to fix it

PR #1176, opened 2024-04-03
Feat: parameterize pipeline class in the primary factory method

z3z1ma:
### Description

This is an  easy addition to parameterize and will simplify some hackery in `cdf`. Furthermore it makes the factory method significantly more flexible in possibilities since we can pass subclasses of Pipeline. 

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### Additional Context

In the future, it would be even simpler if this sort of factory method was a `classmethod` on the `Pipeline` class and simply aliased/exported like `pipeline = Pipeline.create` where `Pipeline.create` has a sig like `def create(cls, ...)` which also lets subclasses benefit from the "standard" setup procedure. 

For now, we make the simpler change though...

### DIFF
--- a/dlt/pipeline/__init__.py
+++ b/dlt/pipeline/__init__.py
@@ -1,4 +1,4 @@
-from typing import Sequence, Type, cast, overload
+from typing import Sequence, Type, TypeVar, cast, overload
 
 from dlt.common.schema import Schema
 from dlt.common.schema.typing import TColumnSchema, TWriteDisposition, TSchemaContract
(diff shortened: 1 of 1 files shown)


rudolfix (commented):
for some reason linter is not able to assign types to instances of pipeline created via `dlt.pipeline`. looks weird. maybe try to temporarily remove the pipeline()` overload (it is not parametrized)

somehow this worked with DltSource

rudolfix:
@z3z1ma OK I know what the problem is. there's a special TypeVar in typing extensions that let's you to set a default. I spend several hours looking for this when upgrading `DltSoure`. See https://github.com/dlt-hub/dlt/pull/1176/commits/48cdfa1eb7d7b0fb37411d61cc0eb747fba814a7

is this PR doing what you need in CDF now?

rudolfix (approved):

z3z1ma:
Ah I definitely would not have found that for some time 😅 

Thanks for the update @rudolfix 🙌 

I just tried it out and indeed it works perfectly. This is very powerful.

PR #1184, opened 2024-04-03
filesystem state sync

sh-rp:
### Description
Implements support for synching state and schema with a filesystem destination. Also saves the info to the loads table the way it is saved in other destinations. The saving of the state is done a bit different, mostly to have the same fileformat always and making things easier. I'm not sure if exracting the state at such a late time creates any problems?

### Notes:
- This approach does not respect a custom layout for the dlt tables for now and always uses jsonl (i think this is fine). Maybe we could provide a configurable path to the dlt internal tables that the user can change, and by default it is _dlt/ (in which all the dlt table folders then lie)
- I touch an "init" file in each dlt table folder to differentiate between "table does not exist" and "table exists but no state found"
- Should we gzip the schema and state files?
- Don't upload files if they exist already (if the hash is the same, the contents should be the same)?

### ToDo:
- Write a couple of filesystem specific state tests, may not be necessary, but I had one instance where the schema was created too often, so maybe a simple test for this would be good.


dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +232,111 @@ def __exit__(
...
+        """gets tuple of dir and fullpath for schema file for a given hash"""
+        safe_hash = "".join(
+            [c for c in hash if re.match(r"\w", c)]
+        )  # remove all special chars from hash
+        return (
+            f"{self.dataset_path}/{self.schema.state_table_name}/{pipeline_name}__{safe_hash}.jsonl"
+        )
+
+    def store_current_state(self) -> None:
+        # get state doc from current pipeline
```
sh-rp, line 278: This is a bit of a hack to get the state from the current pipeline instance and store it as json. You can see that I send an empty load job for the state table. I am not sure we should do it this way but it saves us dealing with different output fileformats and paths.

sh-rp:
Screenshot 2024-04-03 at 18 49 08
This is what you end up with when running the example pipeline currently included. You can always find the current state and current schema with a direct key as well as all the previous versions via hash. I think this is the easiest and fastest way of doing this.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
+    #
+    # Schema read/write
+    #
+
+    def _get_schema_file_name(self, version_hash: str) -> str:
+        """gets full path for schema file for a given hash"""
+        safe_hash = "".join(
+            [c for c in version_hash if re.match(r"\w", c)]
+        )  # remove all special chars from hash
+        return f"{self.dataset_path}/{self.schema.version_table_name}/{self.schema.name}__{safe_hash}.jsonl"
```
rudolfix, line 329: IMO hash is enough. also it would be good to have load_id

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
+
+    def get_stored_schema(self) -> Optional[StorageSchemaInfo]:
+        """Retrieves newest schema from destination storage"""
+        return self.get_stored_schema_by_hash("current")
```
rudolfix, line 333: same thing like in state: find the oldest load id

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
+
+        return None
+
+    def store_current_schema(self) -> None:
+        # get paths
+        current_path = self._get_schema_file_name("current")
+        hash_path = self._get_schema_file_name(self.schema.stored_version_hash)
+
+        # TODO: duplicate of weaviate implementation, should be abstracted out
+        version_info = {
```
rudolfix, line 353: heh I'll be never able to merge my naming convention PR :)

fs_testing_pipe.py
```diff
@@ -0,0 +1,20 @@
+import dlt
```
rudolfix, line 1: keep those in some ignored folder ;)

rudolfix (changes requested):
also make sure that it works when state sync is disabled. otherwise it is really good!

fs_testing_pipe.py
sh-rp: yeah, i have that also, but since i am working on two different machines i need to do this sometimes ;)

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
+        from dlt.common.configuration.container import Container
+        from dlt.common.pipeline import PipelineContext
+        from dlt.pipeline.state_sync import state_doc
+
+        pipeline = Container()[PipelineContext].pipeline()
+        state = pipeline.state
+        doc = state_doc(state)
+
+        # get paths
+        current_path = self._get_state_file_name(pipeline.pipeline_name, "current")
```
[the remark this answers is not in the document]
sh-rp: I am not sure what you mean here, maybe we should discuss it briefly. Imho this setup replicated the behavior of the other destinations, with the same lookups/where clauses.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
+            pipeline.pipeline_name, self.schema.stored_version_hash
+        )
+
+        # write
+        self._write_to_json_file(current_path, doc)
+        self._write_to_json_file(hash_path, doc)
+
+    def get_stored_state(self, pipeline_name: str) -> Optional[StateInfo]:
+        # raise if dir not initialized
+        filepath = self._get_state_file_name(pipeline_name, "current")
```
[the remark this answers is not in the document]
sh-rp: This is what is does, no? I mean it starts with the pipeline name, so you can look up the state with the pipeline, and instead of looking for the highest load_id (which we shouldn't do anyway, since we should not rely on load ids being timestamps) it has this current marker. I have a screenshot above of the file layout this pr produces. This way we can avoid iterating through a list of files to find the newest one which will eventually slow down against destinations with many loads, at least that would be my expectation.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -171,27 +179,44 @@ def update_stored_schema(
...
+        # write schema to destination
+        self.store_current_schema()
+
         return expected_update
 
-    def _get_table_dirs(self, table_names: Iterable[str]) -> Set[str]:
+    def _get_table_dirs(self, table_names: Iterable[str]) -> List[str]:
         """Gets unique directories where table data is stored."""
-        table_dirs: Set[str] = set()
+        table_dirs: List[str] = []
```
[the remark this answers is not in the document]
sh-rp: so that the def update_stored_schema for the filesystem works, for that I need an ordered list of dirs. I can also change it so we keep the set here and implement another one for the the schema update.

sh-rp:
> Screenshot 2024-04-03 at 18 49 08 This is what you end up with when running the example pipeline currently included. You can always find the current state and current schema with a direct key as well as all the previous versions via hash. I think this is the easiest and fastest way of doing this.

this image is not quite up to date, there will now also be these init files in each folder.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
     def should_load_data_to_staging_dataset(self, table: TTableSchema) -> bool:
         return False
+
+    #
+    # state stuff
+    #
+
+    def _write_to_json_file(self, filepath: str, data: DictStrAny) -> None:
+        dirname = os.path.dirname(filepath)
+        if not self.fs_client.isdir(dirname):
```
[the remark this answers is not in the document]
sh-rp: there is a test that fails because _write_to_json_file will be called without an initialised dataset. I can't quite remember which one, but i will have a look.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
+        # TODO: this is also duplicate across all destinations. DRY this.
+        load_data = {
+            "load_id": load_id,
+            "schema_name": self.schema.name,
+            "status": 0,
+            "inserted_at": pendulum.now().isoformat(),
+            "schema_version_hash": self.schema.version_hash,
+        }
+        filepath = (
+            f"{self.dataset_path}/{self.schema.loads_table_name}/{self.schema.name}.{load_id}.jsonl"
```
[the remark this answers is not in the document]
sh-rp: i wanted to align it with the way all the other dlt tables are stored. i somehow like it more, we could have a setting though for backwards compatibility or something? your call.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: I need the name in the filepath so I can find the right schema when looking for the newest version of a schema, so I will keep it.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -220,3 +239,126 @@ def __exit__(
...
+    # state read/write
+    #
+
+    def _get_state_file_name(self, pipeline_name: str, version_hash: str) -> str:
+        """gets full path for schema file for a given hash"""
+        safe_hash = "".join(
+            [c for c in version_hash if re.match(r"\w", c)]
+        )  # remove all special chars from hash
+        return (
+            f"{self.dataset_path}/{self.schema.state_table_name}/{pipeline_name}__{safe_hash}.jsonl"
```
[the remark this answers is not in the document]
sh-rp: i put load id there now, not sure what you mean with the "you will not get the same hash twice" comment?

dlt/destinations/impl/filesystem/filesystem.py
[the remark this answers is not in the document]
sh-rp: I have changed it to iterate over the files in the dir and select the correct one

dlt/destinations/impl/filesystem/filesystem.py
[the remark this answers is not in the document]
sh-rp: changed this now.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: done (assuming you mean the newest load id :) )

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -168,30 +179,49 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
...
+        table_names = only_tables or self.schema.tables.keys()
+        dirs_to_create = self._get_table_dirs(table_names)
+        for tables_name, directory in zip(table_names, dirs_to_create):
             self.fs_client.makedirs(directory, exist_ok=True)
+            # we need to mark the folders of the data tables as initialized
+            if tables_name in self.schema.dlt_table_names():
+                self.fs_client.touch(posixpath.join(directory, INIT_FILE_NAME))
+
+        # write schema to destination
+        self.store_current_schema(load_id or "1")
```
sh-rp, line 197: When "sync_destination" is called, we are not inside the context of a load. I am not quite sure how to handle this case. I first just did not store the schema, but there is a test that verifies that there is a schema in the destination after "sync_destination" is called on a pipeline with nothing in the versions folder. Either we change the the tests or think of some default value. I am not sure..

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: I have done this, but somehow this feels like a bit strange solution too. The place where this state resource is extracted is right before the load package gets committed for normalization, so why do we not just put the current pipeline state in the load package state right before committing in this place?

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -167,31 +178,53 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
                         " should be created previously!"
                     )
 
+        # we mark the storage folder as initialized
+        self.fs_client.touch(posixpath.join(self.dataset_path, INIT_FILE_NAME))
```
sh-rp, line 182: unfortunately when dropping the azure dataset, the folders still remain, only the files will be gone, so we need a marker to see wether the dataset is initialized..

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -166,31 +177,57 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
...
 
     def is_storage_initialized(self) -> bool:
-        return self.fs_client.isdir(self.dataset_path)  # type: ignore[no-any-return]
+        return self.fs_client.exists(posixpath.join(self.dataset_path, INIT_FILE_NAME))  # type: ignore[no-any-return]
 
     def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
+        # skip the state table, we create a jsonl file in the complete_load step
+        # this does not apply to scenarios where we are using filesystem as staging
+        # where we want to load the state the regular way
+        if table["name"] == self.schema.state_table_name and not self.config.as_staging:
```
sh-rp, line 228: we can probably remove this special case as soon as we tackle a fully unified and pluggable state management interface, right now it is too deeply mixed up with the regular loading mechanism for the other destinations.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: We can also find the newest load_id for this schema present and increase it by one, but that also does not feel right..

sh-rp: or we use a current timestamp, then it should be in line with the other destinations.

sh-rp:
@rudolfix I have added some more tests for the statesync. Imho the best approach would be to standardize the state loading and retrieving further and run all standard sql tests for the filesystem also, right now many standard sql tests rely on executing sql, so this would be a bigger refactor including probably an extension of the "WithStateSync" interface. Additionally the drop command will currently not work on the filesystem, since that also relies on the sql client, but I don't think it needs to if the destinations would have an improved interface that could be implemented by both the sql and filesystem destinations. Maybe you could have a look if the coverage seems enough at this point and we tackle the harmonization of (possibly all?) destinations when we also add the ability to route statesync to a separate place.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: i have changed it to be like this though. For lineage purposes it would be interesting to also have the load_id (if available) in the file/table, but for now it is inline with the other destinations

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -219,3 +250,160 @@ def __exit__(
...
+    #
+
+    def _write_to_json_file(self, filepath: str, data: DictStrAny) -> None:
+        dirname = posixpath.dirname(filepath)
+        if not self.fs_client.isdir(dirname):
+            return
+        self.fs_client.write_text(filepath, json.dumps(data), "utf-8")
+
+    def _to_path_safe_string(self, s: str) -> str:
+        return "".join([c for c in s if re.match(r"\w", c)]) if s else None
```
sultaniman, line 264: should we instead use `c in strings.ascii_letters`?
```py
In [2]: import string
In [3]: string.ascii_letters
Out[3]: 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
```

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -219,3 +249,161 @@ def __exit__(
...
+
+    def _list_dlt_dir(self, dirname: str) -> Iterator[Tuple[str, List[str]]]:
+        if not self.fs_client.exists(posixpath.join(dirname, INIT_FILE_NAME)):
+            raise DestinationUndefinedEntity({"dir": dirname})
+        for filepath in self.fs_client.listdir(dirname, detail=False):
```
sultaniman, line 269: `listir` is just an alias to `fs_client.ls` or is it implementing some additional behavior?

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -219,3 +249,161 @@ def __exit__(
...
+        # write entry to load "table"
+        # TODO: this is also duplicate across all destinations. DRY this.
+        load_data = {
+            "load_id": load_id,
+            "schema_name": self.schema.name,
+            "status": 0,
+            "inserted_at": pendulum.now().isoformat(),
+            "schema_version_hash": self.schema.version_hash,
+        }
+        filepath = f"{self.dataset_path}/{self.schema.loads_table_name}/{self.schema.name}__{load_id}.jsonl"
```
sultaniman, line 286: Should we use `path.join` instead?

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -219,3 +249,161 @@ def __exit__(
...
+        self._store_current_state(load_id)
+        self._store_load(load_id)
+
+    #
+    # state read/write
+    #
+
+    def _get_state_file_name(self, pipeline_name: str, version_hash: str, load_id: str) -> str:
+        """gets full path for schema file for a given hash"""
+        return f"{self.dataset_path}/{self.schema.state_table_name}/{pipeline_name}__{load_id}__{self._to_path_safe_string(version_hash)}.jsonl"
```
sultaniman, line 300: Should we use `path.join` here as well?

dlt/destinations/impl/filesystem/filesystem.py
sultaniman: is it beneficial to use zip for state files at all?

sultaniman (approved):

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -155,7 +155,12 @@ def test_replace_write_disposition(layout: str, default_buckets_env: str) -> Non
             for basedir, _dirs, files in client.fs_client.walk(
                 client.dataset_path, detail=False, refresh=True
             ):
+                # remove internal paths
+                if "_dlt" in basedir or "init" in basedir:
```
sultaniman, line 159: "init" maybe should be a reference to constant

docs/website/docs/dlt-ecosystem/destinations/filesystem.md
```diff
@@ -249,7 +248,13 @@ You can choose the following file formats:
...
-You can however easily [backup and restore the pipeline working folder](https://gist.github.com/rudolfix/ee6e16d8671f26ac4b9ffc915ad24b6e) - reusing the bucket and credentials used to store files.
+This destination fully supports [dlt state sync](../../general-usage/state#syncing-state-with-destination). To this end, special folders and files that will be created at your destination which hold information about your pipeline state, schemas and completed loads. These folders DO NOT respect your
+settings in the layout section. When using filesystem as a staging destination, not all of these folders are created, as the state and schemas are
+managed in the regular way by the final destination you have configured.
+
+You will also notice `init` files being present in the root folder and the special `dlt` folders. In the absence of the concepts of schemas and tables
+in blob storages and directories, `dlt` uses these special files to harmonize the behavior of the `filesystem` destination with the other implemented destinations.
```
sultaniman, line 256: is this much `gpt` 🙂 ?

tests/load/pipeline/test_filesystem_pipeline.py
```diff
@@ -19,6 +19,8 @@
 from tests.common.utils import load_json_case
 from tests.utils import ALL_TEST_DATA_ITEM_FORMATS, TestDataItemFormat, skip_if_not_active
 from dlt.destinations.path_utils import create_path
+from tests.load.pipeline.utils import destinations_configs, DestinationTestConfiguration
+from tests.load.pipeline.utils import load_table_counts
```
sultaniman, line 23: these two can be merged

rudolfix (changes requested):
LGTM! we can remove the mark. if we are running filesystem through standard state tests, it should be good

docs/website/docs/dlt-ecosystem/destinations/filesystem.md
sh-rp: I'm not sure what you mean.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: maybe we could have our own fsclient wrapper that only exposes stuff we think is reliable..

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: changed it to a hex string now

dlt/destinations/impl/filesystem/filesystem.py
sultaniman: +1 on having our own abstraction with things we really need

dlt/pipeline/pipeline.py
```diff
@@ -427,13 +429,14 @@ def extract(
...
                 if self.config.restore_from_destination:
                     # this will update state version hash so it will not be extracted again by with_state_sync
-                    self._bump_version_and_extract_state(
+                    state = self._bump_version_and_extract_state(
                         self._container[StateInjectableContext].state, True, extract_step
                     )
-                # commit load packages
-                extract_step.commit_packages()
+                # commit load packages with state
+                extract_step.commit_packages(state)
```
sh-rp, line 439: the state only gets committed to the load package when it has changed now, this is how it is done for other destinations too, but I think we should probably always add the pipeline state to the load package and then check in the destination wether a state entry with this hash exists, right? eventually (I think) all other destinations will also use this mechanism to store the state, this would make sense to me at least.

sh-rp:
Follow up ticket for state synching to separate place: https://github.com/dlt-hub/dlt/issues/1233

dlt/pipeline/pipeline.py
rudolfix: here we could add info if version was bumped. easy to check the pre and post bump and extract state. but for now it is OK

rudolfix (dismissed):
LGTM!

PR #1188, opened 2024-04-04
Update example connection string

MiConnell:
### Description

This is a quick fix to the docs to update the example connection string, `{port}{database}` -> `{port}/{database}` and removes extra variable `connection_string`


sh-rp (approved):
thanks for this :)

PR #1207, opened 2024-04-10
RESTClient: Extend docstrings of paginator classes

burnash:
### Description
This PR extends class and method docstrings in the paginator classes of the RESTClient module and adds usage examples.

### Related Issues

- #1141


AstrakhantsevaAA (commented):
Amazing! The most useful part is a response examples like:

> with a 'next' key like this:
        `{"items": [...], "pagination": {"next": "https://api.example.com/items?page=2"}}`.
        
I would add them for every type of paginators (e.g. HeaderLinkPaginator)

dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -7,45 +7,71 @@
...
-        """Update the paginator state based on the response.
+        """Updates the paginator's state based on the response from the API.
+
+        This method should extract necessary pagination details (like next page
+        references) from the response and update the paginator's state
+        accordingly.
 
         Args:
-            response (Response): The response object from the API.
+            response (Response): The response object from the API request.
```
rudolfix, line 60: This method must call `next_reference` setter with a value of next reference right? otherwise `has_next_page` is not going to work correctly. Maybe we should mention this here?

I also wonder if we really need to force `next_referece` in base class and just leave `has_next_page` abstract? This property is kind of useful in  `BaseNextUrlPaginator` (and I wonder why it is public?)

rudolfix (changes requested):
I have one comment regarding `next_reference`. I'm not sure if it belongs to base class but if it is we should tell the user when to set it

dlt/sources/helpers/rest_client/paginators.py
rudolfix: otherwise this is really high quality!

burnash: > This method must call `next_reference` setter with a value of next reference right? otherwise `has_next_page` is not going to work correctly. Maybe we should mention this here?
> 
> I also wonder if we really need to force `next_referece` in base class and just leave `has_next_page` abstract? This property is kind of useful in `BaseNextUrlPaginator` (and I wonder why it is public?)

Yes, both are valid points: 
1. `next_reference` is useless in offset/limit pagination (or page-based), so having it in the base class is not a good design. 
Let's address it a follow-up PR. 
2. `update_state` needs to modify `_has_next_page` indeed. I'll add this to the docstring.

rudolfix (dismissed):
LGTM!

however you need to solve conflict then use your admin power to merge or ask me for another review @burnash

Issue #1175, opened 2024-04-02
fix how `filesystem` deals with windows UNC paths


rudolfix:
### dlt version

0.4.7

### Describe the problem

Windows UNC paths ie `\\localhost\c$\my_dir` are mangled both by `filesystem` source and a destination.

### Expected behavior

They can be used like any other path

### Steps to reproduce

in the code below it is clear that UNC path got completely mangled
```python
import os
import posixpath
from typing import Iterator
import dlt
from dlt.sources import TDataItems
try:
    from .filesystem import FileItemDict, filesystem, readers, read_csv  # type: ignore
except ImportError:
    from filesystem import (
        FileItemDict,
        filesystem,
        readers,
        read_csv,
    )
TESTS_BUCKET_URL = "file://////cifs-fin/oracle_erp_extracts/FNPD/Avetta/In"
def load_csv() -> None:
    """Demonstrates how to scan folder with csv files, load them in chunk and merge on date column with the previous load"""
    pipeline = dlt.pipeline(
        pipeline_name="standard_filesystem_csv",
        destination='snowflake',
        dataset_name="file_data",
    )
    data_file = readers(
        bucket_url=TESTS_BUCKET_URL, file_glob="VendorProvince.csv"
    ).read_csv()
    load_info = pipeline.run(data_file.with_name("data_file"),write_disposition="replace")
    print(load_info)
    print(pipeline.last_trace.last_normalize_info)
```
```
Error:
  File "c:\bgisrepos\dlt_pipeline_flows\.venv\Lib\site-packages\dlt\extract\pipe_iterator.py", line
270, in _get_source_item
    pipe_item = next(gen)
                ^^^^^^^^^
  File "C:\bgisrepos\dlt_pipeline_flows\filesystem\readers.py", line 30, in _read_csv
    with file_obj.open() as file:
         ^^^^^^^^^^^^^^^
  File "c:\bgisrepos\dlt_pipeline_flows\.venv\Lib\site-packages\dlt\common\storages\fsspec_filesystem.py", line 237, in open
    opened_file = self.fsspec.open(
                  ^^^^^^^^^^^^^^^^^
  File "c:\bgisrepos\dlt_pipeline_flows\.venv\Lib\site-packages\fsspec\spec.py", line 1293, in open
    f = self._open(
        ^^^^^^^^^^^
  File "c:\bgisrepos\dlt_pipeline_flows\.venv\Lib\site-packages\fsspec\implementations\local.py", line 197, in _open
[...]

PR #1209, opened 2024-04-11
fix(filesystem): UNC paths are not supported

IlyaFaer:
Towards #1175


IlyaFaer:
@rudolfix, that's a couple of cunning fixes. I'd say I didn't find ideological errors - it's just about a couple of symbols (normal ones), which make `fsspec` (not our code) to turn the path into something weird. Can't say I like it, but it worked. Pushed a test in `verified-sources`: https://github.com/dlt-hub/verified-sources/pull/423

rudolfix (commented):
@IlyaFaer it is hard to say if this works or not without comments and tests or maybe I'm reviewing too early?

rudolfix (changes requested):
OK good!

but you can add tests also here. we are testing `glob` in `test_local_fileststem`. please unit test this
- UNC path
- windows path starting with drive letter
- windows path using backslashes

rudolfix (changes requested):
more specific on the test, following https://en.wikipedia.org/wiki/File_URI_scheme
please test following windows paths

* C:\a\b\b
* \\localhost\c$\a
* file://server/folder/data.xml
* file:////server/folder/data.xml (if the above works, you can skip this one)
 with the `glob` function

IlyaFaer:
@rudolfix, hm, I don't think it actually works. Surprisingly, when you try to read one file, it goes fine. But if you give a glob, `fsspec` does path translation of some sort, and always adds a disk name into the start, creating a pattern like:

`(?s:[\\\\/][\\\\/]localhost[\\\\/][\\\\/]C\\$[\\\\/]git_reps[\\\\/]dlt[\\\\/]tests[\\\\/]common[\\\\/]storages[\\\\/]samples[\\\\/].*)\\Z`

Here:
https://github.com/fsspec/filesystem_spec/blob/37c1bc63b9c5a5b2b9a0d5161e89b4233f888b29/fsspec/utils.py#L742

So, basically, it can see files by UNC path (I logged, it really does), but it filters them out, not returning anything. Not sure if it can be overridden...

However, if I give a file path (not a glob), it reads the file correctly, not going to this part of code, which does the path translation.

If I give a normal full path, like `C:/a/b/c`, it works fine. But UNC - no.

IlyaFaer:
@rudolfix, the only way I see how to fix it, is to write our own filesystem, inheriting the `fsspec.LocalFileSystem` and override the `glob` method to fix this translation. As there are no conditions in the translation, there is no way to skip it for globs (it's skipped only for precise paths). What do you think?

rudolfix:
@IlyaFaer 

> @rudolfix, the only way I see how to fix it, is to write our own filesystem, inheriting the fsspec.LocalFileSystem and override the glob method to fix this translation. As there are no conditions in the translation, there is no way to skip it for globs (it's skipped only for precise paths). What do you think?

heh this is so bad :) could you check first if regular Python glob works correctly? then my take would be to use it for `file://` and just translate the results to something fsspec understand. but let's check it out.

IlyaFaer:
@rudolfix, yes, Python works fine, I tried it even before started to look at `fsspec`

Untitled

IlyaFaer:
> translate the results to something fsspec understand

I don't think it's possible. Because, if we read `localhost`, it's okay, we can translate the path. But what if we read a network entity? It'll start with an ip address - it can't be translated into a path, which starts with a disk name. So, `fsspec` will add it into the pattern by itself.

tests/load/filesystem/test_filesystem_common.py
```diff
@@ -256,3 +258,42 @@ def test_filesystem_destination_passed_parameters_override_config_values() -> No
...
+                "freshman_lbs.csv",
+                "mlb_players.csv",
+                "mlb_teams_2012.csv",
+                "mlb_players.jsonl",
+                "A881_20230920.csv",
+                "A803_20230919.csv",
+                "A803_20230920.csv",
+                "mlb_players.parquet",
+                "taxi.csv.gz",
+                "sample.txt",
```
IlyaFaer, line 293: When working with UNC, paths are always full and absolute (to include the IP address of the host), so it requires a bit different test case.

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -282,13 +283,22 @@ def glob_files(
         # this is a file so create a proper file url
         bucket_url = pathlib.Path(bucket_url).absolute().as_uri()
         bucket_url_parsed = urlparse(bucket_url)
+
     bucket_url_no_schema = bucket_url_parsed._replace(scheme="", query="").geturl()
     bucket_url_no_schema = (
         bucket_url_no_schema[2:] if bucket_url_no_schema.startswith("//") else bucket_url_no_schema
     )
     filter_url = posixpath.join(bucket_url_no_schema, file_glob)
+    is_unc_path = fs_client.protocol == "file" and "$" in filter_url
```
rudolfix, line 292: I started to experiment with this place but it still does not work. all paths (bucket_url) and filter_url are normalized and $ is never there - at least for your test cases

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -302,10 +312,16 @@ def glob_files(
         # make that absolute path on a file://
         if bucket_url_parsed.scheme == "file" and not file.startswith("/"):
             file = f"/{file}"
-        file_name = posixpath.relpath(file, bucket_url_no_schema)
-        file_url = bucket_url_parsed._replace(
-            path=posixpath.join(bucket_url_parsed.path, file_name)
-        ).geturl()
+
+        if is_unc_path:
+            file_name = file
```
rudolfix, line 317: name must contain only name (final path component)

tests/common/storages/test_local_filesystem.py
```diff
@@ -54,3 +55,41 @@ def test_filesystem_decompress() -> None:
...
         with file_dict.open(compression="enable") as f:
             assert f.read().startswith(b'"1200864931","2015-07-01 00:00:13"')
+
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="Test it only on Windows")
+def test_windows_unc_path() -> None:
+    config = FilesystemConfiguration(bucket_url=TEST_SAMPLE_FILES)
+    config.read_only = True
+
+    unc_path = r"\\localhost\\" + os.path.abspath("tests/common/storages/samples").replace(":", "$")
```
rudolfix, line 65: this returns no elements in the glob

tests/common/storages/test_local_filesystem.py
```diff
@@ -54,3 +55,41 @@ def test_filesystem_decompress() -> None:
...
+                "mlb_teams_2012.csv",
+                "mlb_players.jsonl",
+                "A881_20230920.csv",
+                "A803_20230919.csv",
+                "A803_20230920.csv",
+                "mlb_players.parquet",
+                "taxi.csv.gz",
+                "sample.txt",
+            ]
+            for file in all_file_items:
```
rudolfix, line 91: if there are no elements this tests still passes

tests/common/storages/test_local_filesystem.py
```diff
@@ -54,3 +55,41 @@ def test_filesystem_decompress() -> None:
...
+                file_name = file["file_name"].split("\\")[-1].split("/")[-1]
+                assert file_name in expected_files
+        except NotImplementedError as ex:
+            pytest.skip(f"Skipping due to {str(ex)}")
```
rudolfix, line 95: what does it mean?

rudolfix (changes requested):
there are things that still does not work

tests/common/storages/test_local_filesystem.py
IlyaFaer: Oh, yeah, it is. There were a couple of small changes, which I reverted, because tests were till passing without them 😵‍💫Fixed

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -276,19 +279,30 @@ def glob_files(
     """
     import os
 
+    is_unc_path = "$" in bucket_url
     bucket_url_parsed = urlparse(bucket_url)
     # if this is a file path without a scheme
-    if not bucket_url_parsed.scheme or (os.path.isabs(bucket_url) and "\\" in bucket_url):
+    if (
+        not bucket_url_parsed.scheme or (os.path.isabs(bucket_url) and "\\" in bucket_url)
+    ) and not is_unc_path:
```
IlyaFaer, line 287: This part must not be done for UNC, as it breaks them completely

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -299,20 +313,28 @@ def glob_files(
...
         if bucket_url_parsed.scheme == "file" and not file.startswith("/"):
             file = f"/{file}"
-        file_name = posixpath.relpath(file, bucket_url_no_schema)
-        file_url = bucket_url_parsed._replace(
-            path=posixpath.join(bucket_url_parsed.path, file_name)
-        ).geturl()
+
+        if is_unc_path:
+            file_url = "file:///" + file
+            file_name = file.replace(bucket_url_no_schema, "").replace("\\", "/").lstrip("/")
```
IlyaFaer, line 323: It's the most straightforward and stupid solution, but I think it'll work. For UNC we're getting `file` from `glob`, which is supposed to always return the same result, so `str` edits like this should work. This way it produces the same result as `relpath` for normal paths.

dlt/common/storages/fsspec_filesystem.py
rudolfix: good point! although our docs mention that file_name can be a relative path, it is very misleading. so let's fix it:
1. rename `file_name` to `relative_path` and add `file_name` with just the name part
2. update all the tests
3. update documentation: https://dlthub.com/docs/dlt-ecosystem/verified-sources/filesystem#fileitem-representation

luckily we do not use `file_name` in verified sources

tests/common/storages/test_local_filesystem.py
```diff
@@ -54,3 +55,39 @@ def test_filesystem_decompress() -> None:
...
+            "met_csv/A803/A803_20230919.csv",
+            "met_csv/A803/A803_20230920.csv",
+            "parquet/mlb_players.parquet",
+            "gzip/taxi.csv.gz",
+            "sample.txt",
+        ]
+        assert len(all_file_items) == len(expected_files)
+
+        for file in all_file_items:
+            assert file["file_name"] in expected_files
```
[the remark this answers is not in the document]
rudolfix: this will also try to load the content of the file so it will verify if fsspec `open` works. I think it is crucial here

dlt/common/storages/fsspec_filesystem.py
rudolfix: what is `file` here? file_url must contain UNC path so we are able to use it to open later. make sure it is like that

rudolfix (changes requested):

tests/common/storages/test_local_filesystem.py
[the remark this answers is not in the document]
IlyaFaer: @rudolfix, opening doesn't work. Hm-m, I didn't notice it, but our code uses `fsspec.open()`, which (just like any other `fsspec` method) break UNC paths. O-o-okay, I added like a separate branch for UNC. Locally, the tests all passed, pushing (there are still several small things I didn't do, but I wanna see if I didn't break anything).

IlyaFaer:
@rudolfix, uf-f, that's quite a big re-writing. But I think I understood it

rudolfix (approved):

rudolfix:
@IlyaFaer it was actually impossible to correctly implement all file paths styles on all platforms using fsspec. what this PR does is to use Python `glob` for all `file://` uris. it also converts all `file://` uris to native paths before opening. somehow fsspec is able to open UNC paths, but it is not able to do that when they are in `file://` form.

as usual tests are way more complicated than code...

dlt/common/storages/configuration.py
```diff
@@ -117,11 +116,57 @@ def fingerprint(self) -> str:
...
+                # filesystem root
+                if local_path == "/":
+                    return str(pathlib.Path("/").resolve())
+                # this prevents /C:/ or ///share/ where both POSIX and Windows root are present
+                if os.path.isabs(local_path[1:]):
+                    local_path = local_path[1:]
+        return str(pathlib.Path(local_path))
+
+    @staticmethod
+    def make_file_uri(local_path: str) -> str:
```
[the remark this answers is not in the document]
rudolfix: you are right that this needs a better place. we have file utils in `FileStorage` but they probably need to go somewhere else, then we'll move those as well

PR #1212, opened 2024-04-11
Add paginate shortcut to `dlt.sources.helpers.requests`

burnash:
### Description
This PR adds a new function `paginate` to `dlt.sources.helpers.requests` avoiding the circular import issue from #1154 
`paginate` is a shortcut which creates a RESTClient under the hood and runs `paginate()` method.

### Related PRs

- #1141 
- #1154

### DIFF
--- a/dlt/sources/helpers/requests/__init__.py
+++ b/dlt/sources/helpers/requests/__init__.py
@@ -15,6 +15,7 @@ from requests import (
 from requests.exceptions import ChunkedEncodingError
 from dlt.sources.helpers.requests.retry import Client
 from dlt.sources.helpers.requests.session import Session
+from dlt.sources.helpers.rest_client import paginate # noqa: F401
 from dlt.common.configuration.specs import RunConfiguration
 
 client = Client()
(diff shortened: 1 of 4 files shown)


burnash:
@rudolfix, it works now; however, I'm not fully convinced that `dlt.sources.helpers.requests` is an appropriate place for the `paginate()` shortcut function. I believe it still logically belongs in the `rest_client` namespace. A real-world example is when someone needs to explicitly use an authenticator or a paginator; they would need to import that from the `dlt.sources.helpers.rest_client` module:

```py
from dlt.sources.helpers import requests
from dlt.sources.helpers.rest_client import paginators

for page in requests.paginate(
    "https://api.example.com/endpoint",
    paginator=paginators.JSONResponsePaginator(next_url_path="metadata.nextPage"),
    auth=auth.APIKeyAuth(name="apikey", location="query", api_key="..."),
    params={"size": 100},
):
    print(page)
```

In my opinion, it would be logically simpler if the related parts lived under one package, e.g.:

```py
from dlt.sources.helpers.rest_client import paginate, paginators

for page in paginate(
    "https://api.example.com/endpoint",
    paginator=paginators.JSONResponsePaginator(next_url_path="metadata.nextPage"),
    auth=auth.APIKeyAuth(name="apikey", location="query", api_key="..."),
    params={"size": 100},
):
    print(page)
```

rudolfix:
@burnash following on the above: let's close it. probably my fault to push the idea but it looked good at that point :)

PR #1219, opened 2024-04-15
Pass options to parse iso like strings

VioletM:
Now if datetime cannot be parsed by `parse_iso8601` it will be parsed by `_parse_common` with default pendulum options.

### DIFF
--- a/dlt/common/time.py
+++ b/dlt/common/time.py
@@ -1,12 +1,16 @@
 import contextlib
-from typing import Any, Optional, Union, overload, TypeVar, Callable  # noqa
 import datetime  # noqa: I251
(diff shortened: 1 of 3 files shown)


rudolfix (approved):
LGTM! it seems this will also work on pendulum 2.0 so no need to upgrade deps

rudolfix (changes requested):
just make sure all tests are passing:
https://github.com/dlt-hub/dlt/actions/runs/8686738863/job/23818843284?pr=1219

this function is used to detect dates from strings so run the affected tests

sh-rp:
@VioletM with this change the function will now detect more datetime objects and you have to update the tests to reflect that. let me know if something is unclear there.

sh-rp (approved):
looks good :)

PR #1225, opened 2024-04-16
Make merge write-disposition fall back to append if no primary or merge keys are specified

sh-rp:
### Description
Fall back to append if no merge keys are available


sh-rp:
@rudolfix comments from the last pr (which somehow got automatically closed when I reset this branch)

===
it is a good start :)

    there's already test for it (test_merge_no_merge_keys)
    we should also test what happens if we have hard delete column
    filesystem also fallbacks to replace when there are no keys (and I think there's a test for it)
    what about vector databases? they support limited merge (on primary key). I think they do not fallback to replace but make sure
===

tests/pipeline/utils.py
```diff
@@ -3,6 +3,7 @@
 import pytest
 import random
 from os import environ
+import io
```
sh-rp, line 6: i pulled this over from my filesystem state branch because i needed those changes here.

tests/load/utils.py
```diff
@@ -226,6 +226,10 @@ def destinations_configs(
             DestinationTestConfiguration(destination="synapse", supports_dbt=False),
         ]
 
+        # sanity check that when selecting default destinations, one of each sql destination is actually
+        # provided
+        assert set(SQL_DESTINATIONS) == {d.destination for d in destination_configs}
```
sh-rp, line 231: this has happened a few times already... not really related to this pr

rudolfix (approved):
LGTM!

PR #1236, opened 2024-04-17
Segment Migration: Making Endpoint Configurable.

zem360:
### Description
Introducing env variable `RUNTIME__TELEMETRY_ENDPOINT` to overwrite default telemetry endpoint.

### Changes
-  Adding default endpoint to `RunConfiguration`
-  Overwriting default endpoint with `RUNTIME__TELEMETRY_ENDPOINT`.


dlt/common/runtime/segment.py
```diff
@@ -23,24 +23,30 @@
...
+_SEGMENT_ENDPOINT: str = None
 _SEGMENT_CONTEXT: TExecutionContext = None
 
 
 def init_segment(config: RunConfiguration) -> None:
-    assert (
-        config.dlthub_telemetry_segment_write_key
-    ), "dlthub_telemetry_segment_write_key not present in RunConfiguration"
+    global _WRITE_KEY, _SESSION, _SEGMENT_ENDPOINT
+    if "RUNTIME__TELEMETRY_ENDPOINT" not in os.environ:
```
sh-rp, line 32: you do not need to access os.environ, accessing the values in the config is enough. what you need to do here:
* if config.dlthub_telemetry_endpoint is "https://api.segment.io/v1/track", config.dlthub_telemetry_segment_write_key should be provided, if not fail
* if config.dlthub_telemetry_endpoint is None, fail
* if config.dlthub_telemetry_endpoint is something else, it's ok

sh-rp: fyi, your environ var does not match the config, so maybe that is the reason for your confusion.

dlt/common/runtime/segment.py
```diff
@@ -179,13 +190,15 @@ def _send_event(event_name: str, properties: StrAny, context: StrAny) -> None:
...
         resp = _SESSION.post(
             _SEGMENT_ENDPOINT, headers=headers, json=payload, timeout=_SEGMENT_REQUEST_TIMEOUT
         )
-        # print(f"SENDING TO Segment done {resp.status_code} {time.time() - start_ts} {base64.b64decode(_WRITE_KEY)}")
+        # end_ts = time.time_ns()
+        # elapsed_time = (end_ts - start_ts) / 10e6
+        # print(f"SENDING TO Segment done: {elapsed_time}ms Status: {resp.status_code}")
         # handle different failure cases
-        if resp.status_code != 200:
+        if resp.status_code != 200 or resp.status_code != 204:
```
sh-rp, line 201: `if resp.status_code not in [200,204]:`

dlt/common/runtime/segment.py
```diff
@@ -23,24 +23,30 @@
...
+            config.dlthub_telemetry_segment_write_key
+        ), "dlthub_telemetry_segment_write_key not present in RunConfiguration"
+        # store write key
+        key_bytes = (config.dlthub_telemetry_segment_write_key + ":").encode("ascii")
+        _WRITE_KEY = base64.b64encode(key_bytes).decode("utf-8")
+        # store endpoint
+        _SEGMENT_ENDPOINT = config.dlthub_telemetry_endpoint
+    else:
+        _SEGMENT_ENDPOINT = os.environ["RUNTIME__TELEMETRY_ENDPOINT"]
+        _WRITE_KEY = None
```
sh-rp, line 43: if the checks mentioned above passed, you can just set _SEGMENT_ENDPOINT and _WRITE_KEY to the values form the config at the end

dlt/common/runtime/segment.py
```diff
@@ -170,7 +181,7 @@ def _send_event(event_name: str, properties: StrAny, context: StrAny) -> None:
         logger.debug("Skipping request to external service: payload was filtered out.")
         return
 
-    if not _WRITE_KEY:
+    if not _WRITE_KEY and _SEGMENT_ENDPOINT != RunConfiguration.dlthub_telemetry_endpoint:
```
sh-rp, line 184: here you can simply check if _SEGMENT_ENDPOINT is None, if so, return. all the other checks where done in the init.

sh-rp (changes requested):

zem360:
@sh-rp I messed up the rebase. Need git support

sh-rp (approved):
this is looking good now @zem360 please create a follow up ticket so we can add some basic tests!

sh-rp (approved):

PR #1238, opened 2024-04-17
Revert tzdata update and update lock

sultaniman:
This PR reverts tzdata update in pyproject.toml and updates lockfile.


rudolfix (changes requested):
@sultaniman please add
1. new label "changes dependencies" that we will add to PRs that change deps
2. explanation to CONTRIBUTING: 
```
### Add / update / remove core library dependencies
```
where you'll explain the process and why we do it!

CONTRIBUTING.md
```diff
@@ -62,6 +63,34 @@ only the `duckdb` and `postgres` are available to forks.
...
+2. **Using the '>=' operator:**
+   When specifying dependencies, please make sure to use the `>=` operator while also maintaining version minima. This approach ensures our project remains compatible with older systems and setups, mitigating potential unsolvable dependency conflicts.
+
+For example, if our project currently uses a package `example-package==1.2.3`, and a security update is
+released as `1.2.4`, instead of updating to `example-package==1.2.4`, we can set it to `example-package>=1.2.3,<2.0.0`. This permits the necessary security update and at the same time
+prevents the automatic jump to a potentially incompatible major version update in the future.
+The other important note on using possible version minimas is to prevent potential cases where package
+versions will not be resolvable.
+
+### Hypothetical example:
```
sh-rp, line 86: I think we can scrap this section, the above section is clear enough. Also we are not as much concerned about breaking changes (I think) than about other libs our users are using which will collide with our requirements, so in this case we would not enforce a higher minimal version I think. @rudolfix wdyt.

rudolfix (changes requested):
yeah now it makes sense! @sultaniman @sh-rp should we introduce a label to warns that dependencies are updated? or maybe we add this to the PR template? so you must say if you update core deps or not

@sultaniman you can drop the example

sultaniman:
@rudolfix @sh-rp I added dependency update label

rudolfix (approved):
OK! let's merge this and change our PR template - ask about deps changes there

PR #1239, opened 2024-04-17
adds options to write headers, change delimiter

rudolfix:
### Description
adds config to csv writer, like we have for parquet


sh-rp (approved):
lgtm

PR #1241, opened 2024-04-18
fixes bug, where configs where not injected for async functions

sh-rp:
### Description
Fixes signature discovery for async functions.


rudolfix (approved):
good catch! LGTM!

Issue #945, opened 2024-02-07
allow to define maximum nesting level per resource


rudolfix:
**Background**
We want our users to specify maximum nesting level on a resource. Nesting level is a property of relational json normalizer so I'm reluctant to add it to dlt.resource, on the other hand it would be very useful to do so.
Implementation may require a change in schema representation. currently relational normalizer has its own configuration fragment with it's own hints for tables. we may want to keep those together using `x-normalizer-` hint.

nauxliu:
Is that possible to support specify a subset of fields to un-nesting?

rudolfix:
@nauxliu yes, you can tell dlt which fields to unnest and which do not:
https://dlthub.com/docs/general-usage/resource#define-schema

here `tags` field will be loaded as JSON and not as child table

rudolfix:
my take would be to use `x-normalize: - nesting-level` or similar hint on the table in the normalizer area and additionally interpret it in the `relational`. since we already have it in `dlt.source` I think I can tolerate it in dlt.resource :)

PR #1242, opened 2024-04-18
Add max_table_nesting to resource decorator

sultaniman:
This PR addresses https://github.com/dlt-hub/dlt/issues/945 and implements `max_table_nesting` for resources


dlt/common/normalizers/json/relational.py
```diff
@@ -79,6 +81,10 @@ def _is_complex_type(self, table_name: str, field_name: str, _r_lvl: int) -> boo
         # turn everything at the recursion level into complex type
         max_nesting = self.max_nesting
         schema = self.schema
+        table = schema.get_table(table_name)
+        max_table_nesting = table.get("x-normalizer", {}).get("max_nesting")
+        if max_table_nesting:
+            max_nesting = max_table_nesting
```
sultaniman, line 87: Here I am overriding `max_nesting`

tests/common/normalizers/test_max_nesting.py
```diff
@@ -0,0 +1,79 @@
...
+            1000,
+            8,
+            ALL_TABLES_FOR_RASA_EVENT,
+        ),
+    ),
+)
+def test_resource_max_nesting(
+    nesting_level: int, expected_num_tables: int, expected_table_names: List[str]
+):
+    @dlt.resource(max_table_nesting=nesting_level)
```
sh-rp, line 45: you should have multiple resources with different nesting levels at the same time here, also add a source with a nesting level and check that it is overwritten by the resource setting but kept when there is no resource settingl.

dlt/common/normalizers/json/relational.py
```diff
@@ -79,21 +80,27 @@ def _is_complex_type(self, table_name: str, field_name: str, _r_lvl: int) -> boo
         # turn everything at the recursion level into complex type
         max_nesting = self.max_nesting
         schema = self.schema
+        table = schema.tables.get(table_name)
+        if table:
```
rudolfix, line 84: are we able to cache nesting per table in `extend_table`? and here to only access a dict with a cache? why: because this is called for each dict/list in the results so is time critical

dlt/extract/decorators.py
```diff
@@ -449,6 +455,15 @@ def make_resource(
             schema_contract=schema_contract,
             table_format=table_format,
         )
+
+        # If custom nesting level was specified then
+        # we need to add it to table hints so that
+        # later in normalizer dlt/common/normalizers/json/relational.py
+        # we can override max_nesting level for the given table
+        if max_table_nesting is not None:
```
rudolfix, line 463: do we have `max_table_nesting` available as property of dlt source? I think so. so let's do the same for resource. will be handy for users

rudolfix (changes requested):
LGTM! just two small improvements and a chapter in the docs :)

dlt/extract/decorators.py
sultaniman: Done

dlt/common/normalizers/json/relational.py
sultaniman: Also I am not sure if I understand `extend_table` ’s purpose and how to use it, grepping in source code doesn’t give definitive examples I could use, should I just define some dictionary for this? Now that we have getter and setter in a resource we will have to invalidate this as well.

sultaniman: @rudolfix can you please take a look again?

docs/website/docs/general-usage/resource.md
```diff
@@ -343,6 +343,61 @@ for user in users().add_filter(lambda user: user["user_id"] != "me").add_map(ano
     print(user)
 ```
 
+### Reduce the nesting level of generated tables
+
+You can limit how deep `dlt` goes when generating child tables. By default, the library will descend
+and generate child tables for all nested lists, without limit.
+
+:::note
+You can completely `max_table_nesting`, in this case dlt will use it from the source if it is specified
```
rudolfix, line 352: what do you mean here?

docs/website/docs/general-usage/resource.md
```diff
@@ -343,6 +343,61 @@ for user in users().add_filter(lambda user: user["user_id"] != "me").add_map(ano
...
+                        "text": "string",
+                        "author": "string",
+                    }
+                ]
+            }
+        ]
+    }
+```
+
+In the example above we want only 1 level of child tables to be generates (so there are no child
```
rudolfix, line 380: typo, read again :)

rudolfix (changes requested):
1. please improve the docs and maybe ask @VioletM for proof reading
2. you do not test getter / setter for `max_table_nesting`

otherwise LGTM!

tests/normalize/test_max_nesting.py
```diff
@@ -0,0 +1,290 @@
...
+    pipeline_schema = pipeline.schemas[pipeline.default_schema_name]
+    all_table_names = pipeline_schema.data_table_names()
+    count_all_tables_first_run = len(all_table_names)
+    tables = pipeline_schema.data_table_names()
+    assert count_all_tables_first_run == 1
+    assert tables == ["rasa_bot_events_with_nesting_lvl_one"]
+
+    # now adjust the max_table_nesting for resource and check
+    pipeline.drop()
+    rasa_bot_events_resource.max_table_nesting = 2
```
sultaniman, line 274: @rudolfix here I am manually setting/overriding the `max_table_nesting` for the resource.

tests/common/normalizers/test_max_nesting.py
sultaniman: @sh-rp I updated the test with additional check 
1. setting the `max_table_nesting`
2. then running the pipeline
3. unsetting the `max_table_nesting`
4. check if the tables are still there.

rudolfix (approved):
LGTM!

PR #1262, opened 2024-04-22
adds quoting style option to csv writer config

### LINKED ISSUE #1254: configure quoting for csv writers
Allow for full and minimal quoting to be configurable for both (arrow and Python) csv writers.
1. Add new config parameter
2. Keep the current as default
3. Confgure writers with it
4. Update tests and documentation

rudolfix:
### Description
Closes #1254

### DIFF
--- a/dlt/common/data_writers/writers.py
+++ b/dlt/common/data_writers/writers.py
@@ -381,10 +381,14 @@ class ParquetDataWriter(DataWriter):
         )
 
 
+CsvQuoting = Literal["quote_all", "quote_needed"]
+
+
 @configspec
 class CsvDataWriterConfiguration(BaseConfiguration):
     delimiter: str = ","
     include_header: bool = True
+    quoting: CsvQuoting = "quote_needed"
 
     __section__: ClassVar[str] = known_sections.DATA_WRITER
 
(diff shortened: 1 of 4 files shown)


dlt/common/data_writers/writers.py
```diff
@@ -398,23 +402,32 @@ def __init__(
...
         self.include_header = include_header
         self.delimiter = delimiter
+        self.quoting: CsvQuoting = quoting
         self.writer: csv.DictWriter[str] = None
         self.bytes_encoding = bytes_encoding
 
     def write_header(self, columns_schema: TTableSchemaColumns) -> None:
         self._columns_schema = columns_schema
+        if self.quoting == "quote_needed":
+            quoting: Literal[1, 2] = csv.QUOTE_NONNUMERIC
```
sultaniman, line 418: `csv.QUOTE_NONNUMERIC` has the value of `2` and `1` corresponds to `csv.QUOTE_ALL` should we change this to int from `Literal`?

dlt/destinations/impl/postgres/postgres.py
```diff
@@ -112,6 +112,10 @@ def __init__(self, table_name: str, file_path: str, sql_client: Psycopg2SqlClien
         with FileStorage.open_zipsafe_ro(file_path, "rb") as f:
             # all headers in first line
             headers = f.readline().decode("utf-8").strip()
+            # quote headers if not quoted - all special keyword like "binary" must be quoted
+            headers = ",".join(
+                h if h.startswith('"') else '"' + h + '"' for h in headers.split(",")
```
sultaniman, line 117: I think we need to use `f`-strings for readbility

sh-rp (dismissed):

dlt/common/data_writers/writers.py
sultaniman: so I did a read up I thought, type information would be erased and become any in the case of `else` branch but according to mypy it reads the first type definition for a variable and later uses it to validate the assigned type.

dlt/destinations/impl/postgres/postgres.py
rudolfix: OK I agree, will change it

PR #1263, opened 2024-04-23
Enable auto_mkdir for local filesystem and run on all supported stora…

sultaniman:
This PR adjusts filesystem pipeline tests for local filesystem and enables `auto_mkdir` in tests and runs through all storage backends.

### DIFF
--- a/tests/load/pipeline/test_filesystem_pipeline.py
+++ b/tests/load/pipeline/test_filesystem_pipeline.py
@@ -305,14 +305,9 @@ def test_filesystem_destination_extended_layout_placeholders(
         "hiphip": counter("Hurraaaa"),
     }
     now = pendulum.now()
-    kwargs = {}
-    if default_buckets_env.startswith("file://"):
-        kwargs = {"auto_mkdir": True}
-
     fs_destination = filesystem(
         layout=layout,
         extra_placeholders=extra_placeholders,
-        kwargs=kwargs,
         current_datetime=counter(now),
     )
     pipeline = dlt.pipeline(
--- a/tests/load/pipeline/test_filesystem_pipeline.py
+++ b/tests/load/pipeline/test_filesystem_pipeline.py
@@ -4,6 +4,9 @@ import posixpath
 from pathlib import Path
 from typing import Any, Callable, List, Dict, cast
 
+from fsspec import register_implementation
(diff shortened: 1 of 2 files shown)


tests/load/pipeline/test_filesystem_pipeline.py
```diff
@@ -299,30 +301,36 @@ def count(*args, **kwargs) -> Any:
         "hiphip": counter("Hurraaaa"),
     }
     now = pendulum.now()
-    os.environ["DESTINATION__FILESYSTEM__BUCKET_URL"] = "file://_storage"
+    kwargs = {}
+    if default_buckets_env.startswith("file://"):
```
rudolfix, line 305: maybe I'm missing something but we decided to make path just before `put_file` in copy job and only for local filesystem.

rudolfix (changes requested):

tests/load/pipeline/test_filesystem_pipeline.py
sultaniman: @rudolfix sorry I've been busy with other two prs, I will do it soon :)

sultaniman: @rudolfix I created a separate PR for this https://github.com/dlt-hub/dlt/pull/1267

sultaniman: Merged.

tests/load/pipeline/test_filesystem_pipeline.py
```diff
@@ -355,6 +363,38 @@ def count(*args, **kwargs) -> Any:
...
+
+    register_implementation("file", MockLocalFS, clobber=True)
+
+    yield
+
+    register_implementation("file", LocalFileSystem, clobber=True)
+
+
+@pytest.mark.parametrize("layout", TEST_LAYOUTS)
+def test_filesystem_local_filesystem_put_file_fails_if_we_do_not_create_destination_folder_regression(
```
sultaniman, line 380: @rudolfix @sh-rp here is the regression test, in which I replaced local filesystem implementation with my own mocked version and then later checking the exception chain.

sh-rp (approved):

rudolfix (approved):
LGTM! and thx for all the fixes

PR #1265, opened 2024-04-23
mark scd2 child table test essential

jorritsandbrink:
### Description
Mark scd2 child table test `essential` to ensure all SQL logic is tested against all engines.


sh-rp (approved):
thanks :)

PR #1266, opened 2024-04-23
Add snowflake to application parameter to configuration

sultaniman:
To allow snowflake to track users using dlt we need to pass application parameter as query parameter.

~~If the value is set to `skip` we do not include it (some users might want to disable). @rudolfix wdyt?~~
Will use `""` as a way to skip the addition of `application` parameter.

**Config example**
```toml
[destination.snowflake.credentials]
application="some identifier"
```

### DIFF
--- a/dlt/destinations/impl/snowflake/configuration.py
+++ b/dlt/destinations/impl/snowflake/configuration.py
@@ -93,7 +93,7 @@ class SnowflakeCredentials(ConnectionStringCredentials):
         if "application" not in query:
             if self.application is None:
                 query["application"] = snowflake_application_id
-            elif self.application and self.application != "skip":
+            elif self.application and self.application != "":
                 query["application"] = self.application
 
         return URL.create(
@@ -127,7 +127,7 @@ class SnowflakeCredentials(ConnectionStringCredentials):
 
         if self.application is None:
             conn_params["application"] = snowflake_application_id
-        elif self.application and self.application != "skip":
+        elif self.application and self.application != "":
             conn_params["application"] = self.application
 
         return conn_params
--- a/dlt/destinations/impl/snowflake/configuration.py
+++ b/dlt/destinations/impl/snowflake/configuration.py
@@ -110,6 +110,7 @@ class SnowflakeCredentials(ConnectionStringCredentials):
         private_key: Optional[bytes] = None
         if self.private_key:
             private_key = _read_private_key(self.private_key, self.private_key_passphrase)
+
         conn_params = dict(
             self.query or {},
             user=self.username,
--- a/dlt/destinations/impl/snowflake/configuration.py
+++ b/dlt/destinations/impl/snowflake/configuration.py
@@ -63,7 +63,7 @@ class SnowflakeCredentials(ConnectionStringCredentials):
     authenticator: Optional[str] = None
     private_key: Optional[TSecretStrValue] = None
     private_key_passphrase: Optional[TSecretStrValue] = None
-    application: Optional[str] = None
+    application: Optional[str] = snowflake_application_id
 
     __config_gen_annotations__: ClassVar[List[str]] = ["password", "warehouse", "role"]
 
@@ -90,11 +90,8 @@ class SnowflakeCredentials(ConnectionStringCredentials):
         if self.role and "role" not in query:
             query["role"] = self.role
 
-        if "application" not in query:
-            if self.application is None:
-                query["application"] = snowflake_application_id
-            elif self.application and self.application != "":
-                query["application"] = self.application
+        if self.application != "" and "application" not in query:
+            query["application"] = self.application
 
         return URL.create(
             self.drivername,
@@ -125,9 +122,7 @@ class SnowflakeCredentials(ConnectionStringCredentials):
         if self.authenticator:
             conn_params["authenticator"] = self.authenticator
 
-        if self.application is None:
-            conn_params["application"] = snowflake_application_id
-        elif self.application and self.application != "":
+        if self.application != "" and self.application != "":
             conn_params["application"] = self.application
 
         return conn_params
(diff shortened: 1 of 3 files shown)


dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -49,6 +49,9 @@ def _read_private_key(private_key: str, password: Optional[str] = None) -> bytes
     )
 
 
+snowflake_application_id = "dltHub_dlt"
```
rudolfix, line 52: is this the application id we picked for `dlt`? @VioletM ?

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -60,6 +63,7 @@ class SnowflakeCredentials(ConnectionStringCredentials):
     authenticator: Optional[str] = None
     private_key: Optional[TSecretStrValue] = None
     private_key_passphrase: Optional[TSecretStrValue] = None
+    application: Optional[str] = None
```
rudolfix, line 66: 1. I think you can just set the default here
2. `application_id` looks like a better name? or we want to be compatible with snowflake conn param name? if the latter you can keep `application`

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -85,6 +89,13 @@ def to_url(self) -> URL:
             query["warehouse"] = self.warehouse
         if self.role and "role" not in query:
             query["role"] = self.role
+
+        if "application" not in query:
+            if self.application is None:
+                query["application"] = snowflake_application_id
+            elif self.application and self.application != "skip":
```
rudolfix, line 96: hmmmm you could just allow empty `""` application name and if detected then skip it

rudolfix (changes requested):
1. did we pick a name for our application?
2. I think we should mention that we send it in our docs for snowflake...

dlt/destinations/impl/snowflake/configuration.py
sultaniman: I think it is best to be compatible with standard snowflake parameters because `application_id` will create more confusion for end users.

dlt/destinations/impl/snowflake/configuration.py
sultaniman: Got it I will then just use `""` to skip it

dlt/destinations/impl/snowflake/configuration.py
sh-rp: I would make the var uppercase and import it into the test so we don't have any magic strings there

sultaniman: Done.

docs/website/docs/dlt-ecosystem/destinations/snowflake.md
```diff
@@ -39,11 +39,16 @@ username = "loader"
...
+application = "dltHub_dlt"
 ```
 In the case of Snowflake, the **host** is your [Account Identifier](https://docs.snowflake.com/en/user-guide/admin-account-identifier). You can get it in **Admin**/**Accounts** by copying the account URL: https://kgiotue-wn98412.snowflakecomputing.com and extracting the host name (**kgiotue-wn98412**).
 
 The **warehouse** and **role** are optional if you assign defaults to your user. In the example below, we do not do that, so we set them explicitly.
 
+:::note
+We use `application = "dltHub_dlt"` to let Snowflake to know about dlt users if you would like to disable this behavior
+please set it to empty string `application = ""` this will disable it.
+:::
```
sultaniman, line 51: @VioletM @rudolfix is it fine to word instructions to disable it this way?

rudolfix: @sultaniman @VioletM I agree. And btw. this is not how application id is even used :) it is used to get help from Snowflake SE on customer request. IMO we should just describe what it does: that this field sends application id to snowflake, the default value is ... and that you can disable it by setting "" and point to Snowflake documentation.

@VioletM you are free to add commit to this PR!

rudolfix (changes requested):

docs/website/docs/dlt-ecosystem/destinations/snowflake.md
sultaniman: I see I thought it is mostly used for tracking.

docs/website/docs/dlt-ecosystem/destinations/snowflake.md
```diff
@@ -39,11 +39,18 @@ username = "loader"
 host = "kgiotue-wn98412"
 warehouse = "COMPUTE_WH"
 role = "DLT_LOADER_ROLE"
+application = "dltHub_dlt"
```
VioletM, line 42: I don't think we should put it here, because for most users it's not relevant and it just will be a default value.

sultaniman: Got it I will remove this :)

tests/load/snowflake/test_snowflake_configuration.py
```diff
@@ -103,12 +120,14 @@ def test_snowflake_credentials_native_value(environment) -> None:
...
     # set password via env
     os.environ["CREDENTIALS__PASSWORD"] = "pass"
+    os.environ["CREDENTIALS__APPLICATION"] = "dlt"
     c = resolve_configuration(
         SnowflakeCredentials(),
         explicit_value="snowflake://user1@host1/db1?warehouse=warehouse1&role=role1",
     )
     assert c.is_resolved()
     assert c.password == "pass"
+    assert "application=dlt" in str(c.to_url())
```
VioletM, line 130: Could you put a test for the exact `dltHub_dlt` string? Because we'll be submitting it and we need to be sure it's the same

VioletM (approved):
Otherwise a small comment LGTM!

tests/load/snowflake/test_snowflake_configuration.py
sultaniman: We are checking the default at https://github.com/dlt-hub/dlt/pull/1266/files#diff-7d1eaf8541220ca58e41bdbb425431e2930d50076f8ac88f5282800ae1be696fR82

rudolfix (approved):
LGTM! but now tests must pass

PR #1267, opened 2024-04-23
Automatically create destination folder for local filesystem

sultaniman:
This PR adjusts `LoadFilesystemJob` for local filesystem to automatically create the destination folder.


sh-rp (changes requested):
if this is a bug that was discovered, adding a test to reproduce this would be good

sultaniman:
@sh-rp I think we can add some regression test for this, @rudolfix asked to incorporate this PR into the main PR https://github.com/dlt-hub/dlt/pull/1263

PR #1274, opened 2024-04-24
filesystem improvement and fixes

sh-rp:
### Description
* Add a fs base client that can be accessed from the pipeline with a few useful methods to inspect the loaded files and in the process refactors the filesystem destination a bit for more clarity and easier testing.
* Cleans up some tests utils a bit, still ongoing
* Adds better tests for state and truncating tables with different path layouts and fixes 2 tests along the way

Fixes two bugs:
* State restore not working for certain layout settings
* Truncating tables in subfolders for certain layout settings did not work


dlt/pipeline/pipeline.py
```diff
@@ -956,6 +958,21 @@ def sql_client(self, schema_name: str = None, credentials: Any = None) -> SqlCli
         schema = self._get_schema_or_create(schema_name)
         return self._sql_job_client(schema, credentials).sql_client
 
+    def fs_client(self, schema_name: str = None, credentials: Any = None) -> FSClientBase:
```
sh-rp, line 961: more or less analog to sql_client

rudolfix: I'd rename it to `_fs_client` until we discuss the exact interface.

PR #1277, opened 2024-04-24
Experiment: Guard add_limit counter using lock

sultaniman:


dlt/extract/resource.py
```diff
@@ -343,7 +344,9 @@ def _gen_wrap(gen: TPipeStep) -> TPipeStep:
                 for i in gen:  # type: ignore # TODO: help me fix this later
                     yield i
                     if i is not None:
-                        count += 1
+                        with counter_lock:
```
sh-rp, line 347: @sultaniman I think all this code is run on the main pipeline thread anyway, so the lock will not help here, but I'm not 100% sure

sultaniman: Tbh I also didn't dig deeper, my theory was that we submit the work to `ThreadPoolExecutor` which can potentially case data races.

rudolfix: operations on integers in CPython are atomic so we are not locking anything. IMO race condition goes deep here and we are better off by making our tests to accept 10th element 🤷

rudolfix:
@sultaniman @sh-rp this kind of things should not go into PR, let's discuss them on Slack or f2f. we just waste CPU cycles on running CI

PR #1278, opened 2024-04-24
master merge for 0.4.9 release

rudolfix:
### Description
master merge for 0.4.9 release


sh-rp:
@rudolfix you need to add the ci-full label when creating the pr, if you add it later, it will only be applied if there is another push unfortunately

PR #1280, opened 2024-04-25
RESTClient: refactor BasePaginator

burnash:
### Description
This PR refactors `BasePaginator` to:
- adjust class hierarchy by introducing `BaseReferencePaginator` as an intermediate class that handles the next reference state management.
- encapsulate the `next_reference` attribute (which was publicly accessible) into a protected `_next_reference` attribute of `BaseReferencePaginator`.

### Related PRs

- #1141 
- #1207


rudolfix (approved):
LGTM! I think we should port class tests into a module test at some point

Issue #1257, opened 2024-04-22
Configurable partner identifiers for snowflake and motherduck


sultaniman:
### Feature description

To let Snowflake and Motherduck to identify dlt users we need to pass certain parameters during or via the connection string

1. **For snowflake**: use `<connection-string>?application=dltHub_dlt`, because they require this format,
2. **For motherduck**: for now use the same parameter name until it is clarified

### Are you a dlt user?

Yes, I'm already a dlt user.

PR #1284, opened 2024-04-25
Extend motherduck client configuration to pass custom user agent

sultaniman:
This PR implements passing of `custom_user_agent` property via [connection configuration](https://motherduck.com/docs/integrations/how-to-integrate#language--framework-examples) for motherduck and resolves https://github.com/dlt-hub/dlt/issues/1257 

MotherDuck's guidelines on naming the integration https://motherduck.com/docs/integrations/how-to-integrate#user-agent-guidelines

**User-agent guidelines**

* The format is integration/version(custom-metadata1;custom-metadata2) where the version and metadata sections are optional.
* Avoid using spaces in integration and version sections.
* Multiple custom metadata sections should be separated by semicolons.

**Some examples:**

- my-integration
- my-integration/2.9.0
- my-integration/2.9.0(linux_amd64)
- my-integration/2.9.0(linux_amd64;us-east-1)

**Suggestion**

I suggest to use `dlt/VERSION(OS_NAME)`

**Example from motherduck's docs**
```py
con = duckdb.connect("md:my_database", config={
    "motherduck_token": token,
    "custom_user_agent": "INTEGRATION_NAME"
})
```

**Configuration example**

```toml
[destination.motherduck.credentials]
custom_user_agent="some motherduck identifier"
```

### DIFF
--- a/dlt/destinations/impl/duckdb/configuration.py
+++ b/dlt/destinations/impl/duckdb/configuration.py
@@ -2,7 +2,7 @@ import os
 import dataclasses
 import threading
 from pathvalidate import is_valid_filepath
-from typing import Any, ClassVar, Final, List, Optional, Tuple, Type, Union
+from typing import Any, ClassVar, Dict, Final, List, Optional, Tuple, Type, Union
 
 from dlt.common import logger
 from dlt.common.configuration import configspec
@@ -37,10 +37,16 @@ class DuckDbBaseCredentials(ConnectionStringCredentials):
         if not hasattr(self, "_conn_lock"):
             self._conn_lock = threading.Lock()
 
+        config = {}
+        if hasattr(self, "_conn_config"):
+            config = self._conn_config
+
         # obtain a lock because duck releases the GIL and we have refcount concurrency
         with self._conn_lock:
             if not hasattr(self, "_conn"):
-                self._conn = duckdb.connect(database=self._conn_str(), read_only=read_only)
+                self._conn = duckdb.connect(
+                    database=self._conn_str(), read_only=read_only, config=config
+                )
                 self._conn_owner = True
                 self._conn_borrows = 0
 
@@ -82,6 +88,14 @@ class DuckDbBaseCredentials(ConnectionStringCredentials):
                 self.database = native_value
             else:
                 raise
+
+    @property
+    def conn_config(self) -> Dict[str, Any]:
+        return self._conn_config or {}
+
+    @conn_config.setter
+    def conn_config(self, new_config: Dict[str, Any]) -> None:
+        self._conn_config = new_config
 
     def _conn_str(self) -> str:
         return self.database
--- a/dlt/destinations/impl/duckdb/configuration.py
+++ b/dlt/destinations/impl/duckdb/configuration.py
@@ -2,16 +2,13 @@ import os
 import dataclasses
 import threading
 from pathvalidate import is_valid_filepath
-from typing import Any, ClassVar, Final, List, Optional, Tuple, TYPE_CHECKING, Type, Union
+from typing import Any, ClassVar, Final, List, Optional, Tuple, Type, Union
 
 from dlt.common import logger
 from dlt.common.configuration import configspec
 from dlt.common.configuration.specs import ConnectionStringCredentials
 from dlt.common.configuration.specs.exceptions import InvalidConnectionString
-from dlt.common.destination.reference import (
-    DestinationClientDwhWithStagingConfiguration,
-    DestinationClientStagingConfiguration,
-)
+from dlt.common.destination.reference import DestinationClientDwhWithStagingConfiguration
 from dlt.common.typing import TSecretValue
 
 try:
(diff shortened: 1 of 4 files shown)


dlt/destinations/impl/motherduck/configuration.py
```diff
@@ -32,11 +35,38 @@ def _token_to_password(self) -> None:
         if self.query and "token" in self.query:
             self.password = TSecretValue(self.query.pop("token"))
 
+    def _connect(self, read_only: bool) -> Any:
```
sultaniman, line 38: This is mostly copy-paste from the base client of our `duckdb` destination because the whole logic was hard to compose I decided to make it separate for motherduck

dlt/destinations/impl/motherduck/configuration.py
```diff
@@ -32,11 +35,38 @@ def _token_to_password(self) -> None:
...
+        # TODO: Can this be done in sql client instead?
+        import duckdb
+
+        if not hasattr(self, "_conn_lock"):
+            self._conn_lock = threading.Lock()
+
+        # obtain a lock because duck releases the GIL and we have refcount concurrency
+        with self._conn_lock:
+            config = {}
+            if self.custom_user_agent and self.custom_user_agent != "":
```
sh-rp, line 48: can this not be all in one class and solved with proper subclassing?

tests/load/duckdb/test_motherduck_client.py
```diff
@@ -47,15 +52,49 @@ def test_motherduck_configuration() -> None:
...
+    con.sql("SHOW DATABASES")
+    config.credentials.return_conn(con)
+
+    # check for the default user agent value
+    connect_spy.assert_called()
+    assert "config" in connect_spy.call_args.kwargs
+    assert connect_spy.call_args.kwargs["config"]["custom_user_agent"] == MOTHERDUCK_USER_AGENT
+
+
+def test_motherduck_connect_disable_user_agent_passing(mocker: MockerFixture) -> None:
```
sh-rp, line 75: this test and the above one are almost the same, it would be better to parametrize them with the setting, it has two upsides
1. less code
2. you can really check that the the custom_user_agent is the only thing that changes between the test runs.

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -86,6 +85,14 @@ def parse_native_representation(self, native_value: Any) -> None:
             else:
                 raise
 
+    @property
```
sh-rp, line 88: much better, thanks :) you don't need to change this anymore it is good, but one feedback from me would be that you don't really need getter and setter here, it adds more coded that is not needed since this is not really a public interface used by anyone. adding the conn_config function alone and bulding the connection string there would be enough. But we can keep it like this.

dlt/destinations/impl/motherduck/configuration.py
```diff
@@ -1,24 +1,29 @@
...
 from dlt.common.destination.reference import DestinationClientDwhWithStagingConfiguration
 from dlt.common.destination.exceptions import DestinationTerminalException
 from dlt.common.typing import TSecretValue
 from dlt.common.utils import digest128
-from dlt.common.configuration.exceptions import ConfigurationValueError
 
 from dlt.destinations.impl.duckdb.configuration import DuckDbBaseCredentials
 
 MOTHERDUCK_DRIVERNAME = "md"
+MOTHERDUCK_USER_AGENT = f"dlt/{__version__}({sys.platform})"
```
sh-rp, line 15: this you need to clarify with @rudolfix, I'm not sure what we want here, but from what I saw I think this is according to motherduck docs, right?

sh-rp (changes requested):

dlt/destinations/impl/motherduck/configuration.py
sultaniman: So we have agreed on this in slack.

dlt/destinations/impl/duckdb/configuration.py
sh-rp: @sultaniman: sorry for another change request, maybe I was not clear in my comment. Using hasattr and getattr is not really preferred as we lose typing and it also really is not needed here. They way to leverage inheritance here is the following:
1. create a private function def _connection_config() in the superclass that returns the default value (empty dict or none, whatever you need there)
2. Leave the default behavior for duckdb (do not change anything there)
3. Overwrite this function in the motherduck implementation to return the correct dict (you do not need to prepare this dict in the init or anywhere, just build it there. less lines of code the better)
4. call connection_config() in the superclass when creating the connection.

sh-rp (changes requested):
see comment

dlt/destinations/impl/duckdb/configuration.py
sultaniman: Thanks I think yeah it was a minor misunderstanding, I will adjust it.

sh-rp (approved):

PR #1286, opened 2024-04-25
Fix filesystem layout timestamps with milliseconds

sultaniman:
This PR uses `datetime.timestamp()` for timestamps with milliseconds instead of pendulum.format().
I thought it was fixed but apparently got mixed up.


rudolfix (commented):
Please tell me what was the bug really? "SSS" is a milisecond part of a datetime, not a full timestamp? hard to review without knowing what got fixed
otherwise LGTM

sultaniman:
@rudolfix yes it `SSS` is just a millisecond part of it and pendulum has `x` but it is a stringified timestamp with seconds so switching between the tasks I overlooked this. Sorry.

dlt/destinations/path_utils.py
```diff
@@ -92,7 +92,7 @@ def prepare_datetime_params(
     if load_package_timestamp:
         current_timestamp = ensure_pendulum_datetime(load_package_timestamp)
         params["load_package_timestamp"] = str(int(current_timestamp.timestamp()))
-        params["load_package_timestamp_ms"] = current_timestamp.format("SSS")
+        params["load_package_timestamp_ms"] = str(int(current_timestamp.timestamp() * 1000))
```
sh-rp, line 95: imho we have a place for these kind of function (for example timestamp to seconds or timestamp to ms) which is dlt.common.time. it should go there as util function (the line above too) and there should be some basic tests that only test this function.

sultaniman: I like it, I think this will be helpful for us.

sultaniman: @sh-rp I extracted two helper functions.

tests/common/test_time.py
```diff
@@ -100,3 +102,13 @@ def test_ensure_pendulum_date_utc() -> None:
...
     ) == pendulum.date(2020, 12, 31)
+
+
+@pytest.mark.parametrize("date_value,datetime_obj", test_params)
+def test_datetime_to_timestamp_helpers(
+    date_value: TAnyDateTime, datetime_obj: pendulum.DateTime
+) -> None:
+    dt = ensure_pendulum_datetime(date_value)
+    assert isinstance(dt, pendulum.DateTime)
+    assert datetime_to_timestamp(datetime_obj) == int(datetime_obj.timestamp())
```
sh-rp, line 113: what you want to do here is to create a date time object with a value that you select and then compare it to an actual integer that it should produce, this way you are basically just testing the same code again:
datetime(2021, 1, 1, 5, 6, 55) == 123456789

sultaniman: Nice, I will adjust this with parametrized fixture

sh-rp:
looks good, maybe one additional request: put an example in the docs. Right now it says "timestamp_ms - the current timestamp in Unix Timestamp format rounded to milliseconds" which is not quite right, because it is not a unix timestamp rounded to ms (this would look like this: 12334455.233) but it looks like this: 12334455233 actually.

sh-rp (approved):

Issue #1270, opened 2024-04-24
Docs: update tutorial to use `RESTClient.paginate()`


burnash:
Change the tutorial with an additional step that replaces pagination with ‘while True’ with `RESTClient.paginate()`

burnash:
Implemented in #1287

PR #1287, opened 2024-04-25
Update the tutorial to use `rest_client.paginate` for pagination

burnash:
Implements #1270

### DIFF
--- a/docs/website/docs/tutorial/grouping-resources.md
+++ b/docs/website/docs/tutorial/grouping-resources.md
@@ -141,13 +141,14 @@ For the next step we'd want to get the [number of repository clones](https://doc
 Let's handle this by changing our `fetch_github_data()` first:
 
 ```py
+from dlt.sources.helpers.rest_client.auth import BearerTokenAuth
+
 def fetch_github_data(endpoint, params={}, access_token=None):
-    headers = {"Authorization": f"Bearer {access_token}"} if access_token else {}
     url = f"{BASE_GITHUB_URL}/{endpoint}"
     return paginate(
         url,
         params=params,
-        headers=headers,
+        auth=BearerTokenAuth(token=access_token) if access_token else None,
     )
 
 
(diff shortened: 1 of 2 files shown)


docs/website/docs/tutorial/load-data-from-an-api.md
```diff
@@ -148,6 +148,49 @@ and `updated_at.last_value` to tell GitHub to return issues updated only **after
...
+    dataset_name="github_data_merge",
+)
+load_info = pipeline.run(get_issues)
+row_counts = pipeline.last_trace.last_normalize_info
+
+print(row_counts)
+print("------")
+print(load_info)
+```
+
```
rudolfix, line 193: We should provide 2-3 references to `REST Client` documentation, ie.
- linl to `paginate` docs
- [link to rest client docs]
- `paginate` function above will automatically detect link based pagination that Github uses, refer to ... on how to configure pagination

rudolfix (changes requested):

docs/website/docs/tutorial/grouping-resources.md
```diff
@@ -22,17 +25,11 @@ In the previous tutorial, we loaded issues from the GitHub API. Now we'll prepar
...
-
-        # get next page
-        if "next" not in response.links:
-            break
-        url = response.links["next"]["url"]
+    for page in paginate(
+        "https://api.github.com/repos/dlt-hub/dlt/comments"
+        params={"per_page": 100}
+    ):
+        yield page
```
AstrakhantsevaAA, line 32: I would keep one "pure python" example, so people could understand that they also can use just python if they want.
I would mention below that we have a rest api helper, we recommend to use it to save your time and build more reliable api source

burnash: You probably mean "Requests", right? We have those examples in the first part of tutorial. I was just thinking it would be strange to return back to the hands-on Requests pagination after we switch to `paginate()'

burnash: So the pure-requests examples are here https://deploy-preview-1287--dlt-hub-docs.netlify.app/docs/tutorial/load-data-from-an-api#load-only-new-data-incremental-loading repeating multiple times until "Using pagination helper" section.

AstrakhantsevaAA: right!

docs/website/docs/tutorial/load-data-from-an-api.md
burnash: @rudolfix fixed that in 3fd2c5ecf812c1b2a02ab084136cb3a76c1edc35

PR #1289, opened 2024-04-25
fix athena edge case and adds layout tests for athena

sh-rp:
### Description
* Adds layout tests for athena and athena iceberg
* Athena needs each table in its own subfolder, a folder with multiple tables and a prefix does not work, we raise an error now when this happens


rudolfix (approved):
LGTM!

PR #1292, opened 2024-04-26
feat(pipeline): add an ability to auto truncate

### LINKED ISSUE #1104: truncate the staging dataset tables after load
### Feature description

I want to free space in the staging dataset and truncate the tables after load

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

_No response_

### Proposed solution

We can reuse the storage init method in the loader and run the truncation again after the load is fully completed. this is done on best effort basis and should not fail the overall load

### Related issues

_No response_

IlyaFaer:
Closes #1104


dlt/pipeline/pipeline.py
```diff
@@ -519,6 +519,7 @@ def load(
         *,
         workers: int = 20,
         raise_on_failed_jobs: bool = False,
+        truncate_staging: bool = True,
```
IlyaFaer, line 522: Motivation to use an argument in `run()` and `load()` is that every request to cloud databases costs money. User may be willing to avoid frequent truncation.

tests/pipeline/test_pipeline.py
```diff
@@ -163,6 +163,51 @@ def test_file_format_resolution() -> None:
...
+        full_refresh=True,
+    )
+    info = pipeline.run(
+        [{"field": 1}, {"field": 2}, {"field": 3}],
+        table_name="staging_cleared_resource",
+    )
+    assert_load_info(info)
+
+    _, staging = pipeline._get_destination_clients(pipeline.default_schema)
+    assert staging.list_table_files("staging_cleared_resource") == []
```
IlyaFaer, line 180: Checking that there are no files

tests/pipeline/test_pipeline.py
```diff
@@ -163,6 +163,51 @@ def test_file_format_resolution() -> None:
...
+    info = pipeline.run(
+        [{"field": 1}, {"field": 2}, {"field": 3}],
+        table_name="staging_cleared_resource",
+        truncate_staging=False,
+    )
+    assert_load_info(info)
+
+    _, staging_client = pipeline._get_destination_clients(pipeline.default_schema)
+    assert (
+        "staging_cleared_resource" in staging_client.list_table_files("staging_cleared_resource")[0]
```
IlyaFaer, line 203: Checking that a file remained

tests/pipeline/test_pipeline.py
sh-rp: @IlyaFaer i think it would be better to have a parametrized test that runs the same code once with the prop set to true and once to false and so the asserts accordingly at the end

sh-rp:
@rudolfix since we have two different concepts of "staging", a staging destination as well as a staging dataset, maybe we should use those terms in config vars more clearly and not just "staging" which could mean both really

sh-rp:
@IlyaFaer in the ticket it says we want to truncate the staging dataset, not the staging destination. I also think that the setting should probably be part of the LoaderConfiguration and not be passed down from the pipeline. @rudolfix should we always truncate or only if all loadjobs where successfull? If a mergejob fails because of a connection error or something like that I don't think we should truncate the staging tables, because you can retry the load.

rudolfix:
@sh-rp we should truncate only when we complete the load id successfully. so after the complete method is executed.

IlyaFaer:
One thing that I didn't manage to solve is how to patch the `LoaderConfiguration`, to run tests for both cases. I tried mocking, direct replacing the class in the module, but none worked 🤔

dlt/load/load.py
```diff
@@ -490,6 +490,28 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
...
+
+        Args:
+            schema (Schema): Schema to use for the staging dataset.
+        """
+        try:
+            if self.config.truncate_staging_dataset:
+                with self.get_destination_client(schema) as client:
+                    if isinstance(client, WithStagingDataset):
+                        with client.with_staging_dataset():
+                            client.initialize_storage(truncate_tables=schema.data_table_names())
```
rudolfix, line 506: here we need to pick only tables that are eligible for staging
- use `_extend_tables_with_table_chain` and pass schema.data_tables()
- select only tables that are eligible for staging (job_client.should_load_data_to_staging_dataset) - as table filter

and truncate returned tables

dlt/pipeline/pipeline.py
```diff
@@ -555,7 +555,9 @@ def load(
             with signals.delayed_signals():
                 runner.run_pool(load_step.config, load_step)
             info: LoadInfo = self._get_step_info(load_step)
+
             self.first_run = False
+            load_step.maybe_trancate_staging_dataset(self.default_schema)
```
rudolfix, line 560: makes sense to put it into a separate method, but I'd prefer to have `load.py` self-contained so please add it at the end of `complete_package` method

tests/pipeline/test_pipeline.py
```diff
@@ -2161,3 +2161,27 @@ def test_yielding_empty_list_creates_table() -> None:
             rows = list(cur.fetchall())
             assert len(rows) == 1
             assert rows[0] == (1, None)
+
+
+def test_staging_dataset_truncate() -> None:
+    @dlt.resource(write_disposition="merge", merge_key="id")
```
rudolfix, line 2167: cool! but remember to enable deletion in config

rudolfix (changes requested):
OK almost there, see review and

1. add info on new config ie to `## Data left behind` of `running.md`
2. we need to enable this behavior for one of the merge tests ie. `test_pipeline_upfront_tables_two_loads` and make sure that all tables in staging dataset are truncated - this is to test it on all destinations

dlt/pipeline/pipeline.py
IlyaFaer: There were failures, if I do truncation somewhere inside the `Load`. Let's see...

dlt/load/configuration.py
```diff
@@ -15,6 +15,9 @@ class LoaderConfiguration(PoolRunnerConfiguration):
     raise_on_max_retries: int = 5
     """When gt 0 will raise when job reaches raise_on_max_retries"""
     _load_storage_config: LoadStorageConfiguration = None
+    # if set to `True`, the staging dataset will be
+    # truncated after loading the data
+    truncate_staging_dataset: bool = True
```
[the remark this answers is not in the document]
IlyaFaer: Done

tests/pipeline/test_pipeline.py
IlyaFaer: Done

dlt/load/load.py
IlyaFaer: Done

IlyaFaer:
@rudolfix, okay, what do we do, if the client doesn't have `should_load_data_to_staging_dataset()`? Qdrant supports both `merge` and `replace`, but doesn't have the method.

tests/load/pipeline/test_pipelines.py
```diff
@@ -881,6 +881,8 @@ def some_source():
 def test_pipeline_upfront_tables_two_loads(
     destination_config: DestinationTestConfiguration, replace_strategy: str
 ) -> None:
+    dlt.config["truncate_staging_dataset"] = True
```
rudolfix, line 884: this is good but we have auto cleanup fixture only for os.environ. so please use it instead

dlt/load/load.py
```diff
@@ -490,6 +492,38 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
...
+        """
+        Truncate the staging dataset if one used,
+        and configuration requests truncation.
+
+        Args:
+            schema (Schema): Schema to use for the staging dataset.
+            job_client (JobClientBase):
+                Job client to use for the staging dataset.
+        """
+        data_tables = schema.data_tables()
```
rudolfix, line 505: you have `data_table_names`

dlt/load/load.py
```diff
@@ -490,6 +492,38 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
...
+        table_names = [tab["name"] for tab in data_tables]
+
+        tables = _extend_tables_with_table_chain(
```
rudolfix, line 508: this is available only in `WithStagingDataset` so move it below the check!

rudolfix (changes requested):

rudolfix:
> @rudolfix, okay, what do we do, if the client doesn't have `should_load_data_to_staging_dataset()`? Qdrant supports both `merge` and `replace`, but doesn't have the method.

yeah because qdrand does not use staging dataset for merge!

look at my comments: client must implement `WithStagingDataset` trait which has `should_load_data_to_staging_dataset` so just move stuff below instance type check!

IlyaFaer:
@rudolfix, hm, some tests fail for Dremio. It seems to me, there staging is always used. Does it mean that staging dataset is not used? I don't see credentials for Dremio in the Secret Manager, so I can't test it locally what's going on there.

dlt/load/load.py
```diff
@@ -490,6 +492,39 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
...
+        """
+        Truncate the staging dataset if one used,
+        and configuration requests truncation.
+
+        Args:
+            schema (Schema): Schema to use for the staging dataset.
+            job_client (JobClientBase):
+                Job client to use for the staging dataset.
+        """
+        if not isinstance(job_client, WithStagingDataset):
```
rudolfix, line 505: OK also return if config option is disabled

dlt/load/load.py
```diff
@@ -490,6 +492,39 @@ def run(self, pool: Optional[Executor]) -> TRunMetrics:
...
+
+        data_tables = schema.data_table_names()
+        tables = _extend_tables_with_table_chain(
+            schema, data_tables, data_tables, job_client.should_load_data_to_staging_dataset
+        )
+
+        try:
+            if self.config.truncate_staging_dataset:
+                with self.get_destination_client(schema) as client:
+                    if isinstance(client, WithStagingDataset):
```
rudolfix, line 516: you just returned when this is not true

tests/load/pipeline/test_pipelines.py
```diff
@@ -994,6 +996,19 @@ def table_3(make_data=False):
         is True
     )
 
+    job_client, _ = pipeline._get_destination_clients(schema)
+
+    if isinstance(job_client, WithStagingDataset):
+        with pipeline.sql_client() as client:
```
rudolfix, line 1002: this is OK but you are hardcoding the staging dataset and qualified table name. this will sometimes work sometimes not depending on the destination. look here:
```
with client.sql_client.with_staging_dataset(
        client.should_load_data_to_staging_dataset(client.schema.tables[table_name])  # type: ignore[attr-defined]
    ):
        canonical_name = client.sql_client.make_qualified_table_name(table_name)
    # write row
    with io.BytesIO() as f:
        write_dataset(client, f, [data_types], column_schemas)
        query = f.getvalue().decode()
    expect_load_file(client, file_storage, query, table_name)
    db_row = list(client.sql_client.execute_sql(f"SELECT * FROM {canonical_name}")[0])
```
it switches to staging dataset to create qualified table name. and then does the select. IMO dremio will start working

rudolfix (changes requested):

rudolfix:
> @rudolfix, hm, some tests fail for Dremio. It seems to me, there staging is always used. Does it mean that staging dataset is not used? I don't see credentials for Dremio in the Secret Manager, so I can't test locally what's going on there.

we only have local container and there are instructions. take a look in dremio folder in destinations. but I think you won't need it if you fix the pr. see review

dlt/load/load.py
IlyaFaer: Done

dlt/load/load.py
IlyaFaer: Pf-f-f, yes! I'm just a bit distracted the last few days 😨

tests/load/pipeline/test_pipelines.py
IlyaFaer: Okay, let's see

rudolfix (approved):
LGTM!

rudolfix (approved):
LGTM!

PR #1294, opened 2024-04-27
Add `upsert` merge strategy

### LINKED ISSUE #633: enable merge on ICEBERG via Athena
**Background**
See **rationale** below for why iceberg/open table. We can support it easily via existing destinations: Athena and Snowflake.
- snowflake ICEBERG https://www.snowflake.com/blog/iceberg-tables-powering-open-standards-with-snowflake-innovations/ - still in private preview so we skip that
- athena - supports ICEBERG and actually our destination is already using it for dlt tables

Why not spark:
- too much work for this ticket.

 **Requirements**
1. * [x] add a new table hint **table_format** that will tell destination which table format to use
2. * [x] for Athena: add global destination setting to set all tables to ICEBERG
3. * [ ] enable merge support for iceberg tables on athena

**Implementation Notes**
* Bulk loading into iceberg on athena can only be done from one table to another. So we need to retain our code for defining regular athena tables and additionally add iceberg tables to copy from the first tables. 
* Iceberg has nice "merge into ... on" sql directives that we can hopefully use for the merge write disposition
* Do we need to run maintenance commands that clean up the underlying parquet files on iceberg or should we leave this to the user?

**Rationale**
Just few points on why iceberg is a game changer on data lakes:
* it allows evolving the partitions
e.g. often data engineers design partitioning based on few use cases that they change over time, and Iceberg allows to just change those partition definitions, thanks to hidden partition (without rewriting the all dataset, that's the case for pure parquet)
* it has build in procedures to optimize the dataset
are you aware of the small file problem in data-lakes? if so iceberg has a build in function to compact "objects" and improve performance and cost. Just a note on that: I wrote quite some compaction procedures on my life, and I believe that Iceberg make that way of compacting data standardised (same as DeltaLake)
* using disposition like Merge > e.g. you can upsert by id
* being able to delete data from a dataset - that's what I mentioned already, and that's why I choose Iceberg to ingest sensitive data from Kakfa for example.
this decision came after leading a GDPR project that allow to remove sensitive data on a data lake - again game changer that make data engineers life much easier.

last but not least: Iceberg is becoming one of the facto open table format in the data landscape:
* Redshift now supports it (read only in preview - I net to test it)
* Snowflake supports it in preview (read/write)
* BiqQuery should support it (read only? - not sure)
* Trino supports it (read/write)
* The Athena supports it too - as it's an AWS fork of Trino (on engine v3)
* Spark supports it - the most complete engine that support iceberg
* Flink supports it - (read/write)
* Dremio Support it - read for sure, not sure about writes
* Clickhouse support it - only read
* DuckDB - support  it - only read
* Java client - read and write
* python client - read only for now
* rust client - coming soon
* go client - coming soon

jorritsandbrink:
### Description
This started as building `merge` support for Athena Iceberg, but the SQL logic runs on some other engines out-of-the-box too (e.g. `postgres`, `mssql`, `snowflake`, `databricks`, `bigquery`). It primarily uses the `MERGE INTO` statement to implement upsert (update + insert) logic. Since both `delete-insert` and `scd2` don't apply, I introduced a new merge strategy called `upsert`.

Characteristics:
- needs a `primary_key`
- uses hash of `primary_key` as deterministic row identifier (`_dlt_id`) in root table
- implements "true" upsert for root table (only updates and inserts), but still needs deletes for child tables (to delete records for elements which have been removed from an updated list)

### Related Issues

- Closes #633

### DIFF
--- a/dlt/destinations/sql_jobs.py
+++ b/dlt/destinations/sql_jobs.py
@@ -159,6 +159,8 @@ class SqlMergeJob(SqlBaseJob):
         merge_strategy = table_chain[0].get("x-merge-strategy", DEFAULT_MERGE_STRATEGY)
         if merge_strategy == "delete-insert":
             return cls.gen_merge_sql(table_chain, sql_client)
+        elif merge_strategy == "upsert":
+            return cls.gen_upsert_sql(table_chain, sql_client)
         elif merge_strategy == "scd2":
             return cls.gen_scd2_sql(table_chain, sql_client)
 
@@ -339,6 +341,67 @@ class SqlMergeJob(SqlBaseJob):
         return f"CREATE TEMP TABLE {temp_table_name} AS {select_sql};"
 
     @classmethod
+    def gen_upsert_sql(
+        cls, table_chain: Sequence[TTableSchema], sql_client: SqlClientBase[Any]
+    ) -> List[str]:
+        sql: List[str] = []
+        root_table = table_chain[0]
+        root_table_name, staging_root_table_name = sql_client.get_qualified_table_names(
+            root_table["name"]
+        )
+        caps = sql_client.capabilities
+        escape_id = caps.escape_identifier
+        root_unique_key = escape_id(get_first_column_name_with_prop(root_table, "unique"))
+
+        # generate merge statement for root table
+        col_str = ", ".join(["{alias}" + escape_id(c) for c in root_table["columns"]])
+        update_str = ", ".join(
+            [escape_id(c) + " = " + "s." + escape_id(c) for c in root_table["columns"]]
+        )
+        sql.append(f"""
+            MERGE INTO {root_table_name} d USING {staging_root_table_name} s
+            ON d.{root_unique_key} = s.{root_unique_key}
+            WHEN MATCHED
+                THEN UPDATE SET {update_str}
+            WHEN NOT MATCHED
+                THEN INSERT ({col_str.format(alias="")}) VALUES ({col_str.format(alias="s.")});
+        """)
+
+        # generate statements for child tables if they exist
+        child_tables = table_chain[1:]
+        if child_tables:
+            if len(child_tables) > 1 and caps.supports_temp_table:
+                # store unique keys in temp table for efficiency
+                temp_table_name = cls._new_temp_table_name("delete", sql_client)
+                select_statement = f"SELECT {root_unique_key} FROM {staging_root_table_name}"
+                sql.append(cls._to_temp_table(select_statement, temp_table_name))
+
+            for table in child_tables:
+                unique_key = escape_id(get_first_column_name_with_prop(table, "unique"))
+                root_key = escape_id(get_first_column_name_with_prop(table, "root_key"))
+                table_name, staging_table_name = sql_client.get_qualified_table_names(table["name"])
+
+                # delete records for elements no longer in the list
+                sql.append(f"""
+                    DELETE FROM {table_name}
+                    WHERE {root_key} IN (SELECT {root_unique_key} FROM {temp_table_name if caps.supports_temp_table else staging_root_table_name})
+                    AND {unique_key} NOT IN (SELECT {unique_key} FROM {staging_table_name});
+                """)
+
+                # insert records for new elements in the list
+                col_str = ", ".join(
+                    ["{alias}" + escape_id(c) for c in list(table["columns"].keys())]
+                )
+                sql.append(f"""
+                    MERGE INTO {table_name} d USING {staging_table_name} s
+                    ON d.{unique_key} = s.{unique_key}
+                    WHEN NOT MATCHED
+                        THEN INSERT ({col_str.format(alias="")}) VALUES ({col_str.format(alias="s.")});
+                """)
+
+        return sql
+
+    @classmethod
     def gen_merge_sql(
         cls, table_chain: Sequence[TTableSchema], sql_client: SqlClientBase[Any]
     ) -> List[str]:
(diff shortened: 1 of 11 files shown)


jorritsandbrink:
@rudolfix @sh-rp Before continuing development on this branch, does the direction I've taken here make sense?

rudolfix (commented):
OK for a new merge strategy! My overall comments
- why do we need to modify the normalizer? `upsert` should be compatible with `insert-delete`
- when generating SQL, the only difference I see is that we can replace INSERT / DELETE into root table with MERGE statement and we do not need a temp table. the child tables however still need temp table to know which rows to delete. maybe our original merge sql should be restructured to separate the child tables processing so it can be reused?
- we'll need to support merge keys and hard deletes as well. 
- we still need temp tables. was the problem with ATHENA a lack of temp tables?

jorritsandbrink:
@rudolfix 

**First addressing a main consideration here:**
I don't think `upsert` can (and should) be compatible with `delete-insert`. `upsert` needs a `primary_key` to ensure a one-to-one relationship between records in the source and target, while `delete-insert` can go without `primary_key`.

**Then addressing your points in order:**
1. The normalizer is changed because the `upsert` logic relies on a deterministic `_dlt_id` (hash of `primary_key`). I think these hashes could also be calculated on the fly in the SQL engine, but that would be inefficient.
2. Biggest benefit will be for the root table indeed. Note though that the delete-insert logic I have here for `upsert` behaves different than the delete-insert logic for `delete-insert`. Difference is that the former is more specific—it only deletes/inserts records for removed/added elements in the list while the latter deletes/inserts all records for all elements, even if they didn't change.
3. 
    - `merge_key` might not really apply here. The composite of `primary_key` and `merge_key` would still need to be unique to ensure the one-to-one relationship, in which case it would make more sense to just include all columns in `primary_key` rather than splitting them across `primary_key` and `merge_key`.
    - `hard_delete` should be possible.
4. Two problems with Athena for `delete-insert`: doesn't support temp tables and doesn't support transactions. 
    - We could probably still do `delete-insert` and use CTEs instead of temp tables, but that would be inefficient (but not more inefficient than the Databricks implementation that uses temporary views). Using an actual table is another approach, but would not have the same performance as a temp table.

**Two options:**
1. Just make `delete-insert` work for Athena Iceberg and accept the drawbacks (no transactions, no temp tables).
2. Continue `upsert` and make it complete/solid. This goes beyond just Athena Iceberg and can be beneficial for more destinations.

sh-rp:
@jorritsandbrink Athena has some kind of concept of temp tables when you use the WITH statement. I'm not sure wether that is helpful in your case, but that is what I found when looking at your pr this morning and doing some research.

sh-rp:
Also why do we need a primary key? The merge into has this ON clause which should support and AND condition for merging on composite keys. Or am I mistaken here?

jorritsandbrink:
@sh-rp We already discussed this on Slack, but I'll add it here too for transparency and completeness.

> Also why do we need a primary key? The merge into has this ON clause which should support AND and OR conditions for merging on composite keys. Or am I mistaken here?

We need `primary_key` to ensure a unique key for the `ON` clause in `MERGE INTO`. SQL engines throw a syntax error when trying to update a target table with duplicate keys. For example, `mssql` says:

_"The MERGE statement attempted to UPDATE or DELETE the same row more than once. This happens when a target row matches more than one source row. A MERGE statement cannot UPDATE/DELETE the same row of the target table multiple times. Refine the ON clause to ensure a target row matches at most one source row, or use the GROUP BY clause to group the source rows."_

`postgres` throws a similar error.

Consider this case to understand why these errors make sense:

staging table:
| id | att |
| -- | -- |
| 1 | bar |
| 1 | baz |

destination table:
| id | att |
| -- | -- |
| 1 | foo |

What should be the result of the destination table after upserting the staging table into it? It's not properly defined.

> Athena has some kind of concept of temp tables when you use the WITH statement. I'm not sure wether that is helpful in your case, but that is what I found when looking at your pr this morning and doing some research.

Yes, these are the CTEs I mention in my previous comment under (4). We can definitely use that to make it work, but it would imply recalculation for each table, rather than calculating once and reusing results as is the case with a temp table. As such, it won't be efficient.

rudolfix:
@jorritsandbrink thx for the explanation. I reviewed the code and now I understand your concept better.
What I think we should do (@sh-rp do you agree):
1. implement athena using insert-delete. I'd just create a table with a random name and then drop it (best effort). we can also create a view (but I doubt those are cached in any way). on databricks we use temporary views and MERGE statements to delete data. take a look
2. stash the existing upsert code, we have a ticket for it #1129 where we can write a good spec

now regarding the upsert itself:
- OK I get why merge key is not possible
- you still use temp table in your code to optimize INSERT/MERGE into child tables. so it will not work for Athena
- I get why we have primary key based hash! we had it in very early version of dlt! I think we could let people use such hash not only for upsert ie, by adding another hint that would provoke such hash
- for the parent table MERGE statement I'd use primary key as it is. why: because arrow tables are not processed in the relational normalizer and will not work with `upsert` (we had the same problem with scd2 and it is temporarily solved which is not convenient).

jorritsandbrink:
@rudolfix Okay, I will leave this branch and PR open as input for #1129 and create a new PR that simply implements `delete-insert` for Athena Iceberg.

dlt/destinations/sql_jobs.py
```diff
@@ -338,6 +340,67 @@ def _to_temp_table(cls, select_sql: str, temp_table_name: str) -> str:
...
+                select_statement = f"SELECT {root_unique_key} FROM {staging_root_table_name}"
+                sql.append(cls._to_temp_table(select_statement, temp_table_name))
+
+            for table in child_tables:
+                unique_key = escape_id(get_first_column_name_with_prop(table, "unique"))
+                root_key = escape_id(get_first_column_name_with_prop(table, "root_key"))
+                table_name, staging_table_name = sql_client.get_qualified_table_names(table["name"])
+
+                # delete records for elements no longer in the list
+                sql.append(f"""
```
jorritsandbrink, line 385: Dummy comment to create link.

jorritsandbrink:
Closed PR. This has moved to https://github.com/dlt-hub/dlt/pull/1466.

Issue #1264, opened 2024-04-23
Streamlit app cache is invalid & load info crashes


willi-mueller:
### dlt version

0.4.8

### Describe the problem

When the schema changes by either
a) dropping a table manually
b) unselecting a resource while having the `write_disposition='replace'`

the streamlit app does not show the changed schema.
![Screenshot 2024-04-23 at 13 06 40](https://github.com/dlt-hub/dlt/assets/217980/ccc8dd66-ab39-4ba7-9ffc-81393b8e18bd)

Also, when I delete a row it is still shown when I click "Show Data".

Further, the "Number of loaded rows" in the load info tab only shows a run-time error and stack trace but no row count.

<img width="1211" alt="Screenshot 2024-04-23 at 12 57 05" src="https://github.com/dlt-hub/dlt/assets/217980/5faff41a-cebb-4664-ad7b-ea4c2a8a15e5">

### Expected behavior

1. When I drop a table the streamlit app should not show it
2. When I deselect a resource and dlt removes the table from the destination the streamlit app should not show it
3. The streamlit app should show the row count in the load info tab
4. Different topic: The tables created by the rest_api should not be filed in dlt under a non-existing schema.

### Steps to reproduce

See screencast: https://www.loom.com/share/700e9f4a1cbe48a5988f55f27c022588?sid=b937409a-329a-4770-85b6-b65afe05aa51

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

rest_api

### dlt destination

DuckDB

### Additional information
My test code:

```python
import dlt

from rest_api import rest_api_source

pokemon_config = {
    "client": {
        "base_url": "https://pokeapi.co/api/v2/",
    },
    "resource_defaults": {
        "write_disposition": "replace",
        "endpoint": {
            "params": {
                "limit": 1000,
            },
        },
    },
    "resources": [
        {
          "name": "berries",
          "endpoint": {
            "path": "berry"
          },
          # "selected": False
        },
        "pokemon",
    ],
}

pokemon_source = rest_api_source(pokemon_config)

pipeline = dlt.pipeline(
    pipeline_name="pokemon_pipeline",
    destination="duckdb",
    dataset_name="pokemon",
    progress="log",
)

load_info = pipeline.run(pokemon_source)
print(load_info)

```

sultaniman:
Tried to reproduce this by running pipeline with the example code multiple times and then trying out 

a) dropping a table manually
[...]

PR #1300, opened 2024-04-29
Do not show a notice if there is no resource state for schema

sultaniman:
This PR adjusts the UI in streamlit app to not show notice message if there is no resource state for the given schema.
Related to: https://github.com/dlt-hub/dlt/issues/1264

### DIFF
--- a/dlt/helpers/streamlit_app/blocks/resource_state.py
+++ b/dlt/helpers/streamlit_app/blocks/resource_state.py
@@ -24,9 +24,7 @@ def resource_state_info(
     sources_state = pipeline.state.get("sources") or {}
     schema = sources_state.get(schema_name, {})
     resource = schema.get("resources", {}).get(resource_name)
-    with st.expander("Resource state", expanded=(resource is None)):
-        if not resource:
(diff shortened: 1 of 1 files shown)


sh-rp:
@sultaniman Is there some way of writing some tests for this?

sultaniman:
@sh-rp unfortunately current testing support is not very pytest friendly https://github.com/streamlit/streamlit/issues/5683 and I would say is incomplete for example if you would like to go beyond simply testing clicks and rendering it will be impossible to do because `AppTest` which streamlit has hides many things and overcoming them using workarounds turned bad when I tried, because it made testing things unreliable.

dlt/helpers/streamlit_app/blocks/resource_state.py
```diff
@@ -24,9 +24,9 @@ def resource_state_info(
     sources_state = pipeline.state.get("sources") or {}
```
rudolfix, line 24: we need to start using helper methods to get source and resource state but OK for now

rudolfix (approved):
LGTM

Issue #1213, opened 2024-04-11
Intermittent CI failure for `test_limit_edge_cases` in async and sync resource limits


burnash:
## Description

CI tests intermittently fail on `test_limit_edge_cases` with a limit of 10. `async_list` unexpectedly contains an extra item compared to the `sync_list`. Race condition?

Test file: tests/extract/test_sources.py
Test case: test_limit_edge_cases[10]
CI link: https://github.com/dlt-hub/dlt/actions/runs/8646067265/job/23704568405?pr=1211#step:8:1018

rudolfix:
I know that does not sound good but let's just allow this additional element. so both lists are valid (including one causing the problem). the whole `add_limit` should be rewritten to limit async generators properly (and we wrap them in threads) and I'm not sure it makes sense now

PR #1310, opened 2024-05-02
Fix add_limit wrapper for async generators

sultaniman:
This PR is an attempt to address https://github.com/dlt-hub/dlt/issues/1213

### DIFF
--- a/dlt/extract/resource.py
+++ b/dlt/extract/resource.py
@@ -341,30 +341,28 @@ class DltResource(Iterable[TDataItem], DltResourceHints):
 
         def _gen_wrap(gen: TPipeStep) -> TPipeStep:
             """Wrap a generator to take the first `max_items` records"""
-
+            nonlocal max_items
             # zero items should produce empty generator
             if max_items == 0:
                 return
 
             count = 0
-            is_async_gen = False
             if callable(gen):
                 gen = gen()  # type: ignore
 
             # wrap async gen already here
             if isinstance(gen, AsyncIterator):
                 gen = wrap_async_iterator(gen)
-                is_async_gen = True
 
             try:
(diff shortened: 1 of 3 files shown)


dlt/extract/resource.py
```diff
@@ -352,14 +352,14 @@ def _gen_wrap(gen: TPipeStep) -> TPipeStep:
...
                 for i in gen:  # type: ignore # TODO: help me fix this later
-                    yield i
                     if i is not None:
                         count += 1
                         # async gen yields awaitable so we must count one awaitable more
                         # so the previous one is evaluated and yielded.
                         # new awaitable will be cancelled
                         if count == max_items + int(is_async_gen):
                             return
+                    yield i
```
sultaniman, line 362: we need to yield only after `if` section

dlt/extract/utils.py
```diff
@@ -180,39 +180,20 @@ def wrap_async_iterator(
...
-            if exhausted:
-                raise StopAsyncIteration()
-            item = await gen.__anext__()
-            return item
-        # on stop iteration mark as exhausted
-        # also called when futures are cancelled
+            while True:
+                if should_stop:
+                    break
+                yield loop.run_until_complete(gen.__anext__())  # type: ignore[arg-type]
```
sultaniman, line 190: I used eventloop because with the old implementation it has been complaining about not awaited generator
```py
tests/extract/test_sources.py::test_add_limit_async
  /[...]/dlt/dlt/extract/pipe_iterator.py:275: RuntimeWarning: coroutine 'wrap_async_iterator.<locals>.run' was never awaited
    pipe_item = next(gen)
  Enable tracemalloc to get traceback where the object was allocated.
  See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
```

Issue #1295, opened 2024-04-29
REST Client: extend base paginator class to prepare initial request


rudolfix:
The request used to get the first page of results does not go through `update_request` so for example we are not able to set the initial page in `OffestPaginator`. 

Add a new method `init_request` that will receive the initial request and will be able to modify it. The base class implementation should leave original request unmodified.

burnash:
Implemented in #1311

PR #1311, opened 2024-05-02
RESTClient: add paginator initialization

burnash:
This PR adds a new method `BasePaginator.init_request()` which is called in `RESTClient.paginate()` before sending it for the first time.

Implements #1295


sultaniman (dismissed):
LGTM. I think we could use [vcrpy](https://vcrpy.readthedocs.io/en/latest/usage.html) to prepare, record and replay requests.

PR #1315, opened 2024-05-03
Enable `merge` write disposition for `athena` Iceberg

### LINKED ISSUE #633: enable merge on ICEBERG via Athena

jorritsandbrink:
### Description
This PR enables the `merge` write disposition (both `delete-insert` and `scd2` strategies) for the `athena` destination with `iceberg` table format.

for `delete-insert`:
- uses regular tables instead of temporary tables because Athena does not support those
    - these tables are created in the staging schema
    - each run, the previous' run table(s) get dropped and new table(s) for that run are created—hence, a version of these tables will always be visible in the staging schema (not super elegant, but not a big problem either I'd say)

### Related Issues

- Closes #633 

### Additional Context
Discussion on how we end up using this approach on this PR: https://github.com/dlt-hub/dlt/pull/1294

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -35,14 +35,20 @@ from dlt.common.exceptions import TerminalValueError
 from dlt.common.utils import without_none
 from dlt.common.data_types import TDataType
 from dlt.common.schema import TColumnSchema, Schema, TSchemaTables, TTableSchema
-from dlt.common.schema.typing import TTableSchema, TColumnType, TWriteDisposition, TTableFormat
+from dlt.common.schema.typing import (
+    TTableSchema,
+    TColumnType,
+    TWriteDisposition,
+    TTableFormat,
+    TSortOrder,
+)
 from dlt.common.schema.utils import table_schema_has_type, get_table_format
 from dlt.common.destination import DestinationCapabilitiesContext
 from dlt.common.destination.reference import LoadJob, DoNothingFollowupJob, DoNothingJob
 from dlt.common.destination.reference import TLoadJobState, NewLoadJob, SupportsStagingDestination
 from dlt.common.storages import FileStorage
 from dlt.common.data_writers.escape import escape_bigquery_identifier
-from dlt.destinations.sql_jobs import SqlStagingCopyJob
+from dlt.destinations.sql_jobs import SqlStagingCopyJob, SqlMergeJob
 
 from dlt.destinations.typing import DBApi, DBTransaction
 from dlt.destinations.exceptions import (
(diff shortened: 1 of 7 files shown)


dlt/destinations/impl/athena/athena.py
```diff
@@ -154,6 +160,64 @@ def __init__(self) -> None:
...
+            staging_root_table_name,
+            sql_client,
+            primary_keys,
+            unique_column,
+            dedup_sort,
+            condition,
+            condition_columns,
+        )
+        # DROP needs backtick as escape identifier
+        sql.insert(0, f"""DROP TABLE IF EXISTS {temp_table_name.replace('"', '`')};""")
```
rudolfix, line 198: yes, and there's PR (#998 ) that allows to escape for DDL and DML separately

rudolfix (approved):
LGTM!

docs/website/docs/dlt-ecosystem/destinations/athena.md
```diff
@@ -137,6 +137,13 @@ force_iceberg = "True"
 
 For every table created as an iceberg table, the Athena destination will create a regular Athena table in the staging dataset of both the filesystem and the Athena glue catalog, and then copy all data into the final iceberg table that lives with the non-iceberg tables in the same dataset on both the filesystem and the glue catalog. Switching from iceberg to regular table or vice versa is not supported.
 
+#### `merge` support
+The `merge` write disposition is supported for Athena when using iceberg tables.
+
+> Note that:
+> 1. there is a risk of tables ending up in inconsistent state in case a pipeline run fails mid flight, because Athena doesn't support transactions, and `dlt` uses multiple DELETE/UPDATE/INSERT statements to implement `merge`,
```
toddy86, line 144: Why not use the `merge` syntax that is available in Athena?

jorritsandbrink: Full discussion is here: https://github.com/dlt-hub/dlt/pull/1294.

_TLDR_: we're planning to add a new merge strategy based on the MERGE statement, but this isn't properly specced yet. We want consistent behavior accross different destinations, hence the implementation of `delete-insert` for `athena` now. MERGE-based `merge` for `athena` will come together with the other destinations.

Issue #1255, opened 2024-04-21
allow lazily evaluated sources/resources and additional setup to be passed to Airflow helper


rudolfix:
**Background**
We want to allow our users
1. to perform additional setup of the worker (ie. injecting env variables) just before pipeline is run
2. to use Airflow-unfriendly sources that ie. read data in the `dlt.source` function instead of resource. (good example is google sheets which reads all data when `dlt.source` is called and just returns sheet data from generated resources)

To make that happen we:
1. allow the `data` argument to be callable and evaluate it just before  `pipeline.run`
2. add new argument `on_before_run` to `run` and `add_run` methods which gets called in `_run` method just before pipeline is run

both callbacks should not accept any parameters. NOTE that **airflow task context will be available in them**

**Tasks**
1. * [x] allow `data` argument to be callable and evaluate it just before `pipeline.run` (only once! - see retry)
2. * [x] accept new `on_before_run` callback and evaluate it at the beginning of `_run` method
3. * [x] document in docstrings
4. * [x] write proper tests for both: make sure that task context is available in both callbacks,

PR #1318, opened 2024-05-03
feat(airflow): allow data to be a callable

IlyaFaer:
Towards #1255 

This PR is for the `data` argument item

### DIFF
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -179,7 +179,10 @@ class PipelineTasksGroup(TaskGroup):
 
         Args:
             pipeline (Pipeline): The pipeline to run
-            data (Any): The data to run the pipeline with
+            data (Any):
+                The data to run the pipeline with. If a non-resource
+                callable given, it's called before the load to get
+                the data.
             table_name (str, optional): The name of the table to
                 which the data should be loaded within the `dataset`.
             write_disposition (TWriteDispositionConfig, optional): Same as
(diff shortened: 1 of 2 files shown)


dlt/helpers/airflow_helper.py
```diff
@@ -271,6 +277,9 @@ def log_after_attempt(retry_state: RetryCallState) -> None:
                 )
 
         try:
+            if callable(data):
+                data = data()
```
IlyaFaer, line 281: Called outside retries

tests/helpers/airflow_tests/test_airflow_wrapper.py
```diff
@@ -917,3 +917,48 @@ def dag_parallel():
...
+
+
+def callable_source():
+    @dlt.resource
+    def test_res():
+        context = get_current_context()
+        yield [
+            {"id": 1, "tomorrow": context["tomorrow_ds"]},
+            {"id": 2, "tomorrow": context["tomorrow_ds"]},
+            {"id": 3, "tomorrow": context["tomorrow_ds"]},
```
IlyaFaer, line 929: Add a context field just to be able to check it (as this thing is actually executed outside of the main loop)

dlt/helpers/airflow_helper.py
```diff
@@ -179,7 +179,10 @@ def run(
 
         Args:
             pipeline (Pipeline): The pipeline to run
-            data (Any): The data to run the pipeline with
+            data (Any):
+                The data to run the pipeline with. If a non-resource
+                callable given, it's called before the load to get
```
rudolfix, line 184: `before the load` is not helpful. what user should now is that it is called when the DAG is run, and not when the DAG is loaded - so just before actual pipeline run happens,.

rudolfix (commented):
LGTM, pls continue with the rest of the ticket

rudolfix (approved):
LGTM!

PR #1321, opened 2024-05-05
adds std pipe iterator for stdin and stderr

rudolfix:
### Description
An iterator that reads both stdout and stderr line by line from a process run via `Venv`

### DIFF
--- a/dlt/common/logger.py
+++ b/dlt/common/logger.py
@@ -1,4 +1,3 @@
-import sys
 import contextlib
 import logging
 import traceback
--- a/dlt/common/logger.py
+++ b/dlt/common/logger.py
@@ -1,3 +1,4 @@
+import sys
 import contextlib
 import logging
 import traceback
@@ -81,7 +82,7 @@ def _init_logging(
         logger = logging.getLogger(logger_name)
         logger.propagate = False
         logger.setLevel(level)
-        # get or create logging handler
+        # get or create logging handler, we log to stderr by default
         handler = next(iter(logger.handlers), logging.StreamHandler())
         logger.addHandler(handler)
 
--- a/dlt/common/runners/stdout.py
+++ b/dlt/common/runners/stdout.py
@@ -1,8 +1,9 @@
 import sys
+import queue
 from contextlib import contextmanager
 from subprocess import PIPE, CalledProcessError
 from threading import Thread
-from typing import Any, Generator, Iterator, List
+from typing import Any, Generator, Iterator, List, Tuple
 
 from dlt.common.runners.venv import Venv
 from dlt.common.runners.synth_pickle import decode_obj, decode_last_obj, encode_obj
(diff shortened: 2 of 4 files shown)


dlt/common/logger.py
```diff
@@ -1,3 +1,4 @@
+import sys
```
sultaniman, line 1: Was this import added by accident?

dlt/common/runners/stdout.py
```diff
@@ -24,6 +25,45 @@ def exec_to_stdout(f: AnyFun) -> Iterator[Any]:
             print(encode_obj(rv), flush=True)
 
 
+def iter_std(venv: Venv, command: str, *script_args: Any) -> Iterator[Tuple[int, str]]:
+    """Starts a process `command` with `script_args` in environment `venv` and returns iterator
+    of (filno, line) tuples where `fileno` is 1 for stdout and 2 for stderr. `line` is
```
sultaniman, line 30: Maybe we should create a simple enum to denote 1 for stdout and 2 as sdterr.

dlt/common/runners/stdout.py
```diff
@@ -24,6 +25,45 @@ def exec_to_stdout(f: AnyFun) -> Iterator[Any]:
...
+    with venv.start_command(
+        command, *script_args, stdout=PIPE, stderr=PIPE, bufsize=1, text=True
+    ) as process:
+        exit_code: int = None
+        q_: queue.Queue[Tuple[int, str]] = queue.Queue()
+
+        def _r_q(std_: int) -> None:
+            stream_ = process.stderr if std_ == 2 else process.stdout
+            for line in iter(stream_.readline, ""):
+                q_.put((std_, line[:-1] if line[-1] == "\n" else line))
```
sultaniman, line 44: I think `line.strip()` is simpler.

sultaniman: Also can line in theory be `None`?

dlt/common/runners/stdout.py
```diff
@@ -24,6 +25,45 @@ def exec_to_stdout(f: AnyFun) -> Iterator[Any]:
...
+            # close queue
+            q_.put(None)
+
+        # read stderr with a thread, selectors do not work on windows
+        t_out = Thread(target=_r_q, args=(1,), daemon=True)
+        t_out.start()
+        t_err = Thread(target=_r_q, args=(2,), daemon=True)
+        t_err.start()
+        while line := q_.get():
```
sultaniman, line 53: Should we have some `hard` timeout after it that we shut down or do we just let users use `Ctrl+C`?

dlt/common/runners/stdout.py
```diff
@@ -24,6 +25,45 @@ def exec_to_stdout(f: AnyFun) -> Iterator[Any]:
             print(encode_obj(rv), flush=True)
 
 
+def iter_std(venv: Venv, command: str, *script_args: Any) -> Iterator[Tuple[int, str]]:
```
sultaniman, line 28: I know that `shlex` is not windows friendly, I am just wondering if we should sanitize script arguments, wdyt?

dlt/common/runners/stdout.py
rudolfix: I will create a Literal with those codes, they correspond to unix file numbers for std so IMO they are fine

dlt/common/runners/stdout.py
rudolfix: I think None stops the iterator so no. I could not find it in my tests

dlt/common/runners/stdout.py
rudolfix: I'm not sure what I should do here? @sultaniman

sultaniman: This was just a security related concern that something might get injected via arguments which will lead RCE or LCE

dlt/common/runners/stdout.py
rudolfix: not now I think. adding timeout is trivial if we need it

PR #1326, opened 2024-05-06
Streamlit caching and session state store fixes

sultaniman:
This PR does the following things

1. Lowers the TTL for query result cache to 1 minute (old was 5 minutes),
2. Avoids storing schema in streamlit's session state store.

Issue: https://github.com/dlt-hub/dlt/issues/1264

### DIFF
--- a/dlt/helpers/streamlit_app/blocks/show_data.py
+++ b/dlt/helpers/streamlit_app/blocks/show_data.py
@@ -1,5 +1,3 @@
-from typing import List
-
 import dlt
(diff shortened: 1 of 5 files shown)


rudolfix (approved):
LGTM!

and to be 100% honest I'd go back to single page application where we do not have any sessions state and use cache only for queries. way less side effects

sultaniman:
@rudolfix yeah, I also agree with this, it rather better to have a single file instead of later trying to tackle the side effects.

PR #1327, opened 2024-05-06
improves dict union validation

### LINKED ISSUE #1316: better error message for unions failing `validate_dict`.
**Background**
`validate_dict` will descend into unions and fail the node if none of union types can validate a value. the error raised does not collect errors for each checked type but provides just generic error. Such message is hard to understand for a user of ie. our REST API source

**Tasks**
Allow `DictValidationException` to contain a list of `DictValidationException` for union types and display their error messages.

rudolfix:
### Description
1. fixes #1316 
2. allows callables to be present in `TypedDict` definitions

Improved messages for unions looks as follow (see. `test_typeddict_friendly_exceptions`):
```
In path .: field 'write_disposition' expects the following types: typing.Literal, TWriteDispositionDict, TMergeDispositionDict, Callable. Provided value {'strategy': 'scd2'} with type 'dict' is invalid with the following errors:
For TWriteDispositionDict: In path ./write_disposition: following required fields are missing {'disposition'}
For TMergeDispositionDict: In path ./write_disposition: following required fields are missing {'disposition'}
For typing.Literal: In path .: field 'write_disposition' with value {'strategy': 'scd2'} is not one of: ('skip', 'append', 'replace', 'merge')
For Callable: In path .: field 'write_disposition' expects callable (function or class instance) but got  '{'strategy': 'scd2'}'. Mind that signatures are not validated

In path .: field 'write_disposition' expects the following types: typing.Literal, TWriteDispositionDict, TMergeDispositionDict, Callable. Provided value unknown with type 'str' is invalid with the following errors:
For typing.Literal: In path .: field 'write_disposition' with value unknown is not one of: ('skip', 'append', 'replace', 'merge')
For TWriteDispositionDict: In path .: field 'write_disposition' with value unknown has invalid type 'str' while 'TWriteDispositionDict' is expected
For TMergeDispositionDict: In path .: field 'write_disposition' with value unknown has invalid type 'str' while 'TMergeDispositionDict' is expected
For Callable: In path .: field 'write_disposition' expects callable (function or class instance) but got  'unknown'. Mind that signatures are not validated

In path .: field 'endpoint' expects the following types: str, Endpoint. Provided value {'path': '/path', 'data_selector': Root(), 'incremental': {'cursor_path': '$', 'start_param': 1}} with type 'dict' is invalid with the following errors:
For Endpoint: In path ./endpoint/incremental: field 'start_param' with value 1 has invalid type 'int' while 'str' is expected
For str: In path .: field 'endpoint' with value {'path': '/path', 'data_selector': Root(), 'incremental': {'cursor_path': '$', 'start_param': 1}} has invalid type 'dict' while 'str' is expected
```

Note that we show the longest path at the top so the "most" fitting case is always show first. it works pretty well in the last test case above


dlt/common/schema/utils.py
```diff
@@ -243,16 +243,18 @@ def simple_regex_validator(path: str, pk: str, pv: Any, t: Any) -> bool:
...
                     path,
+                    t,
                     pk,
                     pv,
                 )
         else:
             if RE_NON_ALPHANUMERIC_UNDERSCORE.match(pv):
                 raise DictValidationException(
-                    f"In {path}: field {pk} value {pv} looks like a regex, please prefix with re:",
+                    f"field {pk} value {pv} looks like a regex, please prefix with re:",
```
sultaniman, line 255: is `:` supposed to be at the end of the message?

dlt/common/typing.py
```diff
@@ -119,7 +120,34 @@ def asstr(self, verbosity: int = 0) -> str:
         ...
 
 
-def extract_type_if_modifier(t: Type[Any]) -> Type[Any]:
+def get_type_name(t: Type[Any]) -> str:
+    """Returns a human-friendly name of type `t`"""
+    if name := getattr(t, "__name__", None):
```
sultaniman, line 125: Should we instead use `type(t)` and combine it with `get_origin`?

dlt/common/typing.py
```diff
@@ -119,7 +120,34 @@ def asstr(self, verbosity: int = 0) -> str:
...
+        return name  # type: ignore[no-any-return]
+    t = get_origin(t) or t
+    if name := getattr(t, "__name__", None):
+        return name  # type: ignore[no-any-return]
+    return str(t)
+
+
+def is_callable_type(hint: Type[Any]) -> bool:
```
sultaniman, line 133: why can't we use builtin `callable` function I checked it out and it works

```ipython
In [1]: class B:
   ...:     def __call__(self):
   ...:         return "B"
   ...: 

In [2]: b = B()

In [3]: callable(b)
Out[3]: True

In [4]: def h():
   ...:     print("aa")
   ...: 

In [5]: callable(h)
Out[5]: True

In [6]: callable(dict)
Out[6]: True

In [7]: callable(None)
Out[7]: False

In [8]: import pendulum

In [9]: callable(pendulum)
Out[9]: False
```

tests/common/test_typing.py
```diff
@@ -65,6 +68,50 @@ class MyDataclass:
...
+    async def async_func():
+        pass
+
+    def non_async_func():
+        pass
+
+    CallableType = Callable[[Any], Any]
+
+    # new type is a callable and will be ignored
+    ANewType = NewType("ANewType", str)
```
sultaniman, line 95: Maybe also add some generic type aliases like `AnyDict = Dict[str, Any]` or `SomeString = Optional[str]`

dlt/common/typing.py
```diff
@@ -119,7 +120,34 @@ def asstr(self, verbosity: int = 0) -> str:
...
+    # this skips NewType etc.
+    if getattr(hint, "__module__", None) == "typing":
+        return False
+    if isinstance(hint, FunctionType):
+        return True
+    # this is how we check if __call__ is implemented in the mro
+    if inspect.isclass(hint) and any("__call__" in t_.__dict__ for t_ in inspect.getmro(hint)):
+        return True
+
+    return False
```
sultaniman, line 147: I think as a last resort we can return the result of `callable(hint)`

dlt/common/typing.py
```diff
@@ -119,7 +120,34 @@ def asstr(self, verbosity: int = 0) -> str:
...
+    t = get_origin(t) or t
+    if name := getattr(t, "__name__", None):
+        return name  # type: ignore[no-any-return]
+    return str(t)
+
+
+def is_callable_type(hint: Type[Any]) -> bool:
+    """Checks if `hint` is callable: a function or callable class. This function does not descend
+    into type arguments ie. if Union, Literal or NewType contain callables, those are ignored"""
+    if get_origin(hint) is get_origin(Callable):
```
sultaniman, line 136: Also can't we directly compare it to `Callable` like `get_origin(hint) is Callable:`

dlt/common/typing.py
sultaniman: Also I came up with less convoluted option which works and tests pass
```py
if get_origin(hint) is Callable:
        return True
    # Avoid unnecessary checks for built-in typing types by checking for __call__ attribute
    if callable(hint):
        return True

    # If hint is a class, check if it or any of its parents have an overridden __call__ method
    if inspect.isclass(hint) and any("__call__" in t.__dict__ for t in inspect.getmro(hint)):
        return True

    return False
```

dlt/common/typing.py
rudolfix: `t` is a type, not instance. what do you mean?

tests/common/test_typing.py
rudolfix: we have them in common typing. and honestly we should probably ditch them...

dlt/common/typing.py
rudolfix: do you retract part of your code above?

dlt/common/typing.py
rudolfix: I tried many things. and this
```
 if callable(hint):
        return True
```
is true for any class type, at least in python 3.8 because class types are callable
```
Python 3.8.11 (default, Jul  3 2021, 17:53:42)
[GCC 7.5.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class C:
...     pass
...
>>> callable(C)
True
>>>
```
so IMO your code will not work

dlt/common/typing.py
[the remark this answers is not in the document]
rudolfix: hmmm then we need to test it. but IMO behavior should be backwards compatible:
```
Python 3.12.1 (main, Dec 10 2023, 15:16:41) [GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> d = dict[str, str]
>>> d
dict[str, str]
>>> import inspect
>>> inspect.isclass(d)
False
>>> from typing import get_origin
>>> get_origin(d)
<class 'dict'>
>>> from typing import Callable
>>> get_origin(Callable[[str], str])
<class 'collections.abc.Callable'>
>>>
```
I do not see any differences. what was your worry?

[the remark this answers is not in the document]
sultaniman: I was just worrying that people will start using builting types for type hints and annotations with newer versions of Python.

dlt/common/typing.py
sultaniman: Ah I got confused because of `getattr`, it's fine.

sultaniman (approved):

dlt/common/typing.py
rudolfix: @sultaniman any update how the tests were passing with the code above?

PR #1335, opened 2024-05-07
fixes filesystem dest on windows

rudolfix:
### Description
`filesystem` destination was not working well with local windows path. we have more and more windows users so this needs ot be corrected.
1. uses native local paths and native pathlib
2. normalizes all paths to pathlib separator
3. adds end to end filesystem tests on windows and mac + local filesystems
4. tbd. handling of windows extended paths \\?\ (to support more than 256 chars in path)

### DIFF
--- a/dlt/common/storages/configuration.py
+++ b/dlt/common/storages/configuration.py
@@ -168,5 +168,4 @@ class FilesystemConfiguration(BaseConfiguration):
         """
         p_ = pathlib.Path(local_path)
         p_ = p_.expanduser().resolve()
-        # return "file:///" + p_.as_posix().lstrip("/")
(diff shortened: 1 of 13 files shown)


sh-rp:
LGTM, the new test is failing on mac and ubuntu now, works for windows though :D

sh-rp:
One note, not directly related: maybe we should organize the tests slightly differently. In the test-common workflow we now are listing tests and folders behind pytest commands. This is prone to us adding tests that might not be run on CI in the end or might be run on the wrong platform. Maybe we should reorganize them just a bit at some point to remove complexity.

sh-rp (approved):

PR #1339, opened 2024-05-08
disable qdrant example test and fix snippet errors

sh-rp:
### Description
Disables failing qdrant example test and fixes some linter errors on snippets

### DIFF
--- a/docs/website/docs/general-usage/http/overview.md
+++ b/docs/website/docs/general-usage/http/overview.md
@@ -58,7 +58,7 @@ from dlt.sources.helpers.rest_client.paginators import JSONResponsePaginator
 
 github_client = RESTClient(
     base_url="https://pokeapi.co/api/v2",
-    paginator=JSONResponsePaginator(next_url_path="next")    # (1)
+    paginator=JSONResponsePaginator(next_url_path="next"),   # (1)
     data_selector="results",                                 # (2)
 )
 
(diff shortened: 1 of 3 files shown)


docs/website/docs/general-usage/http/overview.md
```diff
@@ -58,7 +58,7 @@ from dlt.sources.helpers.rest_client.paginators import JSONResponsePaginator
 
 github_client = RESTClient(
     base_url="https://pokeapi.co/api/v2",
-    paginator=JSONResponsePaginator(next_url_path="next")    # (1)
+    paginator=JSONResponsePaginator(next_url_path="next"),   # (1)
```
burnash, line 61: Nice catch :)

docs/website/docs/general-usage/http/rest-client.md
```diff
@@ -191,7 +191,7 @@ E.g. `https://api.example.com/items?offset=0&limit=100`, `https://api.example.co
 
 ```json
 {
-  "items": [...],
+  "items": ["one", "two", "three"],
```
burnash, line 194: Is there a way to leave ellipsis?

sh-rp: it does not pass json parsing that way.. we can mark the block as text (which will disabled syntax highlighting) or write something like [{}, {}, {}].

burnash: Let's use the strings then

PR #1340, opened 2024-05-08
improves trackers thread pool

rudolfix:
### Description
Tracker pools were lazily instantiated. In case of many pipelines running at the same time this could cause problems.
1. all tracker pools are instantiated when telemetry starts
2. tracker pools have mandatory prefixes
3. they unregister at_exit when stopped

### DIFF
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -104,7 +104,7 @@ jobs:
         if: runner.os != 'Windows'
         name: Run pipeline smoke tests with minimum deps Linux/MAC
       - run: |
-          poetry run pytest tests/pipeline/test_pipeline.py tests/pipeline/test_import_export_schema.py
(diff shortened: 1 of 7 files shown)


sh-rp (dismissed):
looks alright. what where the exact problems that where happening before?

.github/workflows/test_common.yml
```diff
@@ -130,7 +130,7 @@ jobs:
         if: runner.os != 'Windows'
         name: Run extract and pipeline tests Linux/MAC
       - run: |
-          poetry run pytest tests/extract tests/pipeline tests/libs tests/cli/common tests/destinations
+          poetry run pytest tests/extract tests/pipeline tests/libs tests/cli/common tests/destinations  -m "not forked"
```
sh-rp, line 133: why do we need to skip these tests on windows?

sh-rp (approved):

PR #1349, opened 2024-05-13
ability to add partitions in Athena to resources

Vitalii0-o:
dlt version
0.4.10

Describe the problem
[Athena code](https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/athena/athena.py#L408) there is no way to add partitions for Athena tables

Expected behavior
Added the ability to partition Athena tables in a schema

Steps to reproduce
try to create a partitioned table in Athena

Operating system
Linux

Runtime environment
Airflow

Python version
3.11

dlt data source
No response

dlt destination
AWS Athena / Glue Catalog

Other deployment details
No response

Additional information
No response

### DIFF
--- a/dlt/common/schema/typing.py
+++ b/dlt/common/schema/typing.py
@@ -91,6 +91,7 @@ class TColumnType(TypedDict, total=False):
     data_type: Optional[TDataType]
     precision: Optional[int]
     scale: Optional[int]
+    partition: Optional[bool]
 
 
 class TColumnSchemaBase(TColumnType, total=False):
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -418,7 +418,8 @@ class AthenaClient(SqlJobClientWithStaging, SupportsStagingDestination):
         table = self.prepare_load_table(table_name, self.in_staging_mode)
         table_format = table.get("table_format")
         is_iceberg = self._is_iceberg_table(table) or table.get("write_disposition", None) == "skip"
-        columns = ", ".join([self._get_column_def_sql(c, table_format) for c in new_columns])
+        columns = ", ".join([self._get_column_def_sql(c, table_format) for c in new_columns if not c.get("partition")])
+        partition_by = [c for c in new_columns if c.get("partition")]
 
         # this will fail if the table prefix is not properly defined
         table_prefix = self.table_prefix_layout.format(table_name=table_name)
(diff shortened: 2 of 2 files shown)


dlt/common/schema/typing.py
```diff
@@ -91,6 +91,7 @@ class TColumnType(TypedDict, total=False):
     data_type: Optional[TDataType]
     precision: Optional[int]
     scale: Optional[int]
+    partition: Optional[bool]
```
sh-rp, line 94: Please have a look at how we add partitions in bigquery with the bigquery adapter, it's very easy to do and you can copy most of the code from there probably.

sh-rp (changes requested):
Thanks for the contribution :)
We need:
* Move to adapter (see comment)
* A test for creating a table with partitions
* A test for adding a partition to an existing table
* A small update in the athena docs about the partitions (see how it is done bigquery)

nicor88:
Few comments from my side, because there is something missing:
* Adding partitions for external tables (hive) without having dlt writing to the right partitioning schema in s3 is not enough. The data in S3 must match the partitions schema defined - you should add some example on how we you can leverage the s3 layout: `"{table_name}/{test-params}/{YYYY}-{MM}-{DD}/{load_id}.{file_id}.{ext}"` in this case the partition can be called for example `extaction_date` . Also, after dlt writes to s3 you should add the partition to the table via a metadata call, for example `ALTER TABLE elb_logs_raw_native_part ADD PARTITION (dt='2015-01-01') location 's3://athena-examples-us-west-1/elb/plaintext/2015/01/01/'.` - on the first run you can leverage `MSCK REPAIR TABLE impressions`, but for incremental runs this operation is too expensive.
* You should add also partitioning for iceberg tables, in that case the writer is athena itself, not dlt, so what you proposed for hive/external tables should be enough.


Not sure if dlt has integration tests available to run, but you should add integrations tests to verify that the result of adding data to a partitioned table is correct - @sh-rp can for sure help you out here.

rudolfix:
@nicor88 yes we have end to end test and we can test the layout. ideally we could somehow test this via some SQL command. we'll put you as a reviewer here. OK?

rudolfix:
@Vitalii0-o we are moving to #1403

nicor88:
@rudolfix 
>also when looking at the documentation: if I use ADD PARTITION and LOCATION I can have any file layout I want. I just need to provide a file name to LOCATION and that's it, right?
In theory yes, pratically it's a bad practises, because we introduce inconsistency between the S3 layout and the partition definition, and on scale could slow the engine down.

Also, feel free to add me anywhere or ping me when necessary :) happy to help.

PR #1354, opened 2024-05-14
Fast Postgres to Postgres with Connector-X and DuckDB

sspaeti:
### Description
Load data fast from Postgres to Postgres with ConnectorX and DuckDB for bigger Postgres tables (GBs)

This examples shows you how to export and import data from postgres to postgres in a fast way with ConnectorX and DuckDB as the default export will [generate Insert_statement](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1711818064548909?thread_ts=1711033737.320439&cid=C04DQA7JJN6) in the normalization phase which is super slow for large tables.

As it's an initial load, we create a separate schema with a timestamp initially and then replace the existing schema with the new one.

### Related Issues

- Installing DuckDB wasn't as straight forward. There was a strange problem on our servers only (locally wasn't any problem). More details on https://github.com/duckdb/duckdb/issues/8035#issuecomment-2020803032.
- Problem with Data Type TIME with nanoseconds: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1711579390028279?thread_ts=1711477727.553279&cid=C04DQA7JJN60
	- Also we had to convert Timestamp columns with nanosecond when they had `9999-12-31 23:59:59.999999`. We replaced them with NULL instead.
	

#### initial load only
This approach is tested and works well for initial load (`--replace`), the incremental load (`--merge`) might need some adjustments (loading of load-tables of dlt, setting up first run after initial.
load, etc.).

#### run it:
Start with the following command but be aware that you need to define the environment variables in a `.env` and adjust the table names (table1 and table2):

```sh
postgres_to_postgres_with_x_and_duckdb.py --replace
```

#### Default behavior of generating insert-statements

I'm unsure if the normalization phase only generates insert statements when connector-x and parquet are used. We also used different source connectors like Oracle/MSSQL, and these were again bottlenecked in the normalization state with neverending creating insert-statement with a large table. Might there be a better way, as I would think this problem would appear to everyone using dlt (except if you only have small tables)? I'm sure I also didn't know all the right ways of using it, but I couldn't figure it out; that's why I created this workaround with DuckDB, which works very well for the initial load, but now we need to load the initial load every day, which was not the idea.

#### Working example

Right now, you'd need to define env-variables with existing Postgres DBs. It would be nice to update the example with an existing public Postgres DB so that people could easily test it.

I'm happy for feedback, but most importantly, I wanted to share the code here so people who have the same problem can have an example of how to improve speed with DuckDB.

### DIFF
--- a/docs/examples/connector_x_arrow/postgres_to_postgres_with_x_and_duckdb.py
+++ b/docs/examples/connector_x_arrow/postgres_to_postgres_with_x_and_duckdb.py
@@ -1,0 +1,248 @@
+
+"""
+---
+title: Load from Postgres to Postgres with ConnectorX & Arrow export as Parquet, normalizing and exporting as DuckDB, and attaching it to Postgres
+description: Load data fast from Postgres to Postgres with ConnectorX and DuckDB for bigger posgres tables (GBs)
+keywords: [connector x, pyarrow, zero copy, duckdb, postgres, initial load]
+---
+
+This examples shows you how to export and import data from postgres to postgres in a fast way with ConnectorX and DuckDB as the default export will
+generate Insert_statement in the normalization phase which is super slow for large tables.
+
+As it's an initial load, we create a seperate schema with timestamp initially and then replace the existing schema with the new one.
+
+:::note
+This approach is tested and works well for initial load (`--replace`), the incremental load (`--merge`) might need some adjustments (loading of load-tables of dlt, setting up first run after intial
+load, etc.).
+:::
+
+We'll learn:
+
+- How to get arrow tables from [connector X](https://github.com/sfu-db/connector-x) and yield them in chunks.
+- That merge and incremental loads work with arrow tables.
+- How to use DuckDB for a speedy normalization.
+
+
+Start with the following command but be aware that you need to define the environment variables in a `.env` and addjust the tables names (table1 and table2):
+
+```sh
+postgres_to_postgres_with_x_and_duckdb.py --replace
+```
+
+
+:::warn
+Attention: There were problems with data type TIME that includes nano seconds. More details in
+[Slack](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1711579390028279?thread_ts=1711477727.553279&cid=C04DQA7JJN60)
+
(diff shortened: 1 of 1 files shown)


rudolfix:
@sspaeti this is SO good! @AstrakhantsevaAA please take a look. let's merge this

as a side note: after a discussion with you I realized how inefficient we are when dealing with parquet and databases. so we started implementing some shortcuts. in case of postgres I discovered that you can actually stream parquet as csv via psycopg2 and it is at least 10x faster than insert. Surely not as fast as what you have here but on the other hand is able to support merge loads.
https://dlthub.com/docs/dlt-ecosystem/destinations/postgres#fast-loading-with-arrow-tables-and-csv

the next one is mssql and bcp. but we'd need to run external tool for that which is quite obnoxious

ogierpaulmck:
Hi, +1 on how to load mssql very fast!

sspaeti:
> Hi, +1 on how to load mssql very fast!

This example easily works for MSSQL, Oracle and other SQLAlchemy supported connections too. We use it now with exactly these two too. Only thing you need to change is:

1. [connection string](https://github.com/dlt-hub/dlt/pull/1354/files#diff-cae048d3def0718d887bf20f7eb2408a0ea87d3739f186bf5bd786ffd053500aR101) to MSSQL or anything else.
    1. e.g. MSSQL we use `conn = f"mssql://{MSSQL_USERNAME}:{quote_plus(MSSQL_PASSWORD)}@{MSSQL_HOST}: {MSSQL_PORT}/{table_definition['source_db']}?TrustServerCertificate=yes&Encrypt=yes"` 
    2. Oracle: `oracle+oracledb://{POSTGRES_USERNAME}:{POSTGRES_PASSWORD}@{POSTGRES_HOST}:{POSTGRES_PORT}/?service_name={POSTGRES_DATABASE}`
2. `LIMIT {CHUNKSIZE} OFFSET {offset}` in the `read_sql_x_chunked` need to be adjusted to each database. 
    1. E.g. Oracle uses: `OFFSET {CHUNKSIZE * i} ROWS FETCH NEXT {CHUNKSIZE} ROWS ONLY`

PR #1362, opened 2024-05-14
rest_api: add response_actions documentation

burnash:


AstrakhantsevaAA (approved):
looks good!

sultaniman (dismissed):

rudolfix (approved):
@burnash LGTM! just one thought: maybe we should say this is an experimental feature? Or you think we are going to keep the interface

burnash:
> @burnash LGTM! just one thought: maybe we should say this is an experimental feature? Or you think we are going to keep the interface

good point, I'll add a note

Issue #1237, opened 2024-04-17
Data Contracts block `_dlt` tables creation


DrRaider:
### dlt version

0.4.8

### Describe the problem

I tried to use `table: freeze` on an existing database, to write to an existing table, but since the `_dlt` tables were missing, I got the following error: `Contract on tables with mode freeze is violated. Trying to add table xxxxxx but new tables are frozen.`.

### Expected behavior

`_dlt` tables should be create even with `freeze` enabled, or it should be documented that we need to create them first. 
And maybe differentiate the error so that the user knows that it's the `_dlt` tables that are missing.

### Steps to reproduce

On a clean project (first run) if I run my pipeline with table: freeze and I have created the table I want to write to, then I get an error `Contract on tables with mode freeze is violated. Trying to add table xxxxxx but new tables are frozen.`.
If I then remove the `table: freeze` run it once, and finally add it back, it will run without issue.

The main differences between the two runs is that on the second one I have the `_dlt` tables already created.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

Cloud Pub/Sub -> JSON

### dlt destination

Postgres

DrRaider:
I'm adding the slack thread link: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1712848047582969?thread_ts=1712847020.454509&cid=C04DQA7JJN6

rudolfix:
@sh-rp let's investigate! this was one of our test cases and I'm pretty sure we have exception for it...

sh-rp:
@DrRaider this is not a problem with the dlt tables, these will always be created, it is a problem that our internal schema does not know about the table that dlt has never seen but exists on the destination. To make this work allowing table evolution once is the way to go as you suggest. I will add a small note in the docs.

PR #1366, opened 2024-05-15
Create _dlt_* tables even when schema contract is frozen

sultaniman:
This PR let's dlt to create `_dlt_*` tables even when `schema_contract=freeze`

Resolves: https://github.com/dlt-hub/dlt/issues/1237

### DIFF
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -2314,3 +2314,4 @@ def test_pipeline_with_frozen_schema_contract() -> None:
         "_dlt_version",
         "_dlt_pipeline_state",
     }
+    assert table_counts["test_items"] == 2
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -1,14 +1,15 @@
 import asyncio
-import pathlib
-from concurrent.futures import ThreadPoolExecutor
 import itertools
 import logging
 import os
+import pathlib
 import random
+import threading
+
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_pipeline.py
```diff
@@ -2230,3 +2270,48 @@ def stateful_resource():
...
+    pipeline.run(
+        data,
+        table_name="test_items",
+    )
+
+    with pipeline.sql_client() as c:
+        c.execute_sql("DROP TABLE _dlt_loads")
+        c.execute_sql("DROP TABLE _dlt_version")
+        c.execute_sql("DROP TABLE _dlt_pipeline_state")
+        c.execute_sql("TRUNCATE TABLE test_items")
```
sultaniman, line 2297: I am not sure if it is a valid way to do it.

tests/pipeline/test_pipeline.py
```diff
@@ -2230,3 +2270,48 @@ def stateful_resource():
     assert len(fs_client.list_table_files("_dlt_loads")) == 2
     assert len(fs_client.list_table_files("_dlt_version")) == 1
     assert len(fs_client.list_table_files("_dlt_pipeline_state")) == 1
+
+
+def test_pipeline_with_frozen_schema_contract() -> None:
```
rudolfix, line 2275: Please reproduce a test case that user described. All starts with a database that already has a table and a pipeline in total freeze mode! then we can look for a fix in the code

rudolfix (changes requested):

PR #1367, opened 2024-05-15
RESTClient: docs: add the troubleshooting section

burnash:


docs/website/docs/general-usage/http/rest-client.md
```diff
@@ -527,4 +527,70 @@ from dlt.sources.helpers.rest_client import paginate
...
+```
+
+### `RESTClient.paginate()`
+
+Debugging `paginate()` is trickier because it's a generator function that yields [`PageData`](#pagedata) objects. Here's several ways to debug the `paginate()` method:
+
+1. Enable [logging](../../running-in-production/running.md#set-the-log-level-and-format) to see detailed information about the HTTP requests:
+
+```bash
+RUNTIME__LOG_LEVEL=INFO python my_script.py
```
sultaniman, line 559: is `INFO` sufficient here?

sultaniman (approved):
LGTM 🚀 imo we need to improve exceptions and logging as @rudolfix already did maybe we can do extended logging and rich formatted exception logs if `level=DEBUG`?

docs/website/docs/general-usage/http/rest-client.md
burnash: Yes, this is enough to see the requests info

burnash:
Good point @sultaniman. Do you have good examples of rich formatting the exceptions? I think one thing that would be great is to see the actual response content in case of HTTP exceptions, as sometimes there's required context that can clarify the error.

sultaniman:
@burnash I only know about rich traceback printer https://rich.readthedocs.io/en/stable/traceback.html

PR #1371, opened 2024-05-16
rest_api: add troubleshooting section

burnash:


rudolfix (changes requested):
What I'm lacking here: you do not tell the user when do certain action. I left some comments. Also I think we should mention how to troubleshot autodetections
- tell the user that is paginator is not specified we try to autodetect one. if autodetection fails or we are not sure, s/he will get a warning in the log. we recommend to put explicit paginator for production code
- same thing for data_selector. we'll try to detect a valid one but if you are getting weird data in the database you should provide a right one (we log what we guess at the INFO level)
-

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
```diff
@@ -576,3 +576,18 @@ See the [incremental loading](../../general-usage/incremental-loading.md#increme
 - `root_key` (bool): Enables merging on all resources by propagating root foreign key to child tables. This option is most useful if you plan to change write disposition of a resource to disable/enable merge. Defaults to False.
 - `schema_contract`: Schema contract settings that will be applied to this resource.
 - `spec`: A specification of configuration and secret values required by the source.
+
+## Troubleshooting
+
+If you encounter any issues while setting up or running the pipeline, check the following:
+
+- Make sure you have the correct access token and other credentials in the `secrets.toml` file.
```
rudolfix, line 584: if you are getting 401 your authorization data is wrong or you use wrong `auth` type. If you get https://dlthub.com/docs/general-usage/credentials/configuration#understanding-the-exceptions you didnt' provide credentials or dlt cannot find them

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
```diff
@@ -576,3 +576,18 @@ See the [incremental loading](../../general-usage/incremental-loading.md#increme
...
+- Check the API documentation for the correct endpoint paths and query parameters.
```
rudolfix, line 585: When does that happen? What is the symptom? I assume 404? Also ask the user to check if base url is ok and if ask them to verify what they get when they combine it with path. is anything missing?
also if the path starts with `/path` - the path in base url is ignored. it is a common mistake worth mentioning

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
```diff
@@ -576,3 +576,18 @@ See the [incremental loading](../../general-usage/incremental-loading.md#increme
...
+- Verify the configuration settings in the `rest_api_pipeline.py` file.
```
rudolfix, line 586: This does not help. You can explain what DictValidationError is and how to troubleshoot it

rudolfix (changes requested):

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
burnash: > also if the path starts with /path - the path in base url is ignored. it is a common mistake worth mentioning

@rudolfix could you elaborate? I don't think it the base url would be ignored. That's why we have a custom [join_url](https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/helpers/rest_client/utils.py#L1)

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
burnash: done

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
burnash: this should be done as well

rudolfix (dismissed):
LGTM!
really clear

PR #1375, opened 2024-05-16
LanceDB Destination

### LINKED ISSUE #1370: LanceDB Destination
### Feature description

LanceDB's serverless architecture and use of the Lance columnar format, compared to [qdrant](https://dlthub.com/devel/dlt-ecosystem/destinations/qdrant)'s client-server model and custom vector format makes it simpler to deploy, perhaps as part of a self-contained RAG application or serverless pipeline.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

Enabling AI use cases.

### Proposed solution

_No response_

### Related issues

https://github.com/dlt-hub/dlt/issues/1322

Pipboyguy:
### Description

This PR adds support for using LanceDB as a destination.

- [x] Updated documentation
- [x] Unit tests

### Related Issues

- closes #1370 

### Additional Context

With this change, dlt users can easily load their transformed data into LanceDB for efficient vector similarity search, full-text search, and SQL querying. LanceDB's ability to store raw data alongside vector embeddings makes it a powerful destination for AI applications.


dlt/destinations/impl/lancedb/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+]
+
+
+@configspec
+class LanceDBClientConfiguration(DestinationClientDwhConfiguration):
+    destination_type: Final[str] = dataclasses.field(  # type: ignore
+        default="LanceDB", init=False, repr=False, compare=False
+    )
+    credentials: LanceDBCredentials = None
+    dataset_separator: str = "_"
```
sh-rp, line 71: imho we should have a triple underscore here. if datasets have an underscore in their name, we might start dropping tables that are in the wrong "dataset" this way.

dlt/destinations/impl/lancedb/configuration.py
```diff
@@ -0,0 +1,102 @@
...
+    CredentialsConfiguration,
+)
+from dlt.common.destination.reference import DestinationClientDwhConfiguration
+from dlt.common.typing import TSecretStrValue
+from dlt.common.utils import digest128
+
+
+@configspec
+class LanceDBCredentials(CredentialsConfiguration):
+    uri: str = "./.lancedb"
```
sh-rp, line 16: is this a file or a folder? why do we need the "." prefix?

dlt/destinations/impl/lancedb/errors.py
```diff
@@ -0,0 +1,29 @@
...
+
+from dlt.common.destination.exceptions import (
+    DestinationUndefinedEntity,
+    DestinationTerminalException,
+)
+from dlt.common.destination.reference import JobClientBase
+from dlt.common.typing import TFun
+
+
+def lancedb_error(f: TFun) -> TFun:
```
sh-rp, line 16: why can't we use raise_database_error for this like in the other destinations? you'll need to adapt _make_database_exception() for the lancedb needs. Or am I missing something here?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,567 @@
...
+from dlt.destinations.impl.lancedb.utils import (
+    list_unique_identifiers,
+    generate_uuid,
+    set_non_standard_providers_environment_variables,
+)
+from dlt.destinations.job_impl import EmptyLoadJob
+from dlt.destinations.type_mapping import TypeMapper
+
+
+class LanceDBTypeMapper(TypeMapper):
```
sh-rp, line 63: so it is not possible to use precision for decimals? Or is this still to be implemented? also it seems like there are different precision settings for timestamp from what I can tell from the docs.

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,567 @@
...
+        pa.date32(): "date",
+        pa.time64("us"): "time",
+    }
+
+
+class NullSchema(LanceModel):
+    pass
+
+
+class VersionSchema(LanceModel):
```
sh-rp, line 97: do we need these models or can we not just load to those tables the same way the loadpackages are loaded?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,567 @@
...
+
+    def delete_table(self, table_name: str) -> None:
+        """Delete a LanceDB table.
+
+        Args:
+            table_name: The name of the table to delete.
+        """
+        self.db_client.drop_table(table_name)
+
+    def delete_all_tables(self) -> None:
```
sh-rp, line 187: is this ever used?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,567 @@
...
+                for table_name in self.db_client.table_names()
+                if table_name.startswith(prefix)
+            ]
+        else:
+            table_names = self.db_client.table_names()
+
+        for table_name in table_names:
+            self.db_client.drop_table(table_name)
+
+        self._delete_sentinel_table()
```
sh-rp, line 265: Please make sure that when creating and deleting the sentinel table, this table is prefixed with the datasetname. Every "virtual" dataset in this database needs its own sentinel table, otherwise stuff will stop working!

sh-rp: Actually I think you are doing this, but please do it the same way we do it in the other destinations that do not have datasets as a concept. The sentinel table should always be {dataset_name}{separator}{sentinel_table_name}

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,567 @@
...
+            )
+        return applied_update
+
+    def _execute_schema_update(self, only_tables: Iterable[str]) -> None:
+        for table_name in only_tables or self.schema.tables:
+            exists = self.table_exists(self.make_qualified_table_name(table_name))
+            if not exists:
+                self.create_table(
+                    self.make_qualified_table_name(table_name),
+                    schema=cast(LanceModel, NullSchema),
```
sh-rp, line 328: will the actual schema then be inferred from the incoming data or how does this work?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,567 @@
...
+            raise
+
+        tbl.add(lancedb_model(**record))
+
+    def get_stored_state(self, pipeline_name: str) -> Optional[StateInfo]:
+        """Loads compressed state from destination storage by finding a load ID that was completed."""
+        while True:
+            try:
+                state_table_name = self.make_qualified_table_name(self.schema.state_table_name)
+                state_records = (
```
sh-rp, line 368: you should order this by load_id descending, is that possible? Check the exact similar thing in weaviate

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,567 @@
...
+            record = response.to_list()[0]
+            return StorageSchemaInfo(**record)
+        except Exception:
+            raise
+
+    def get_stored_schema(self) -> Optional[StorageSchemaInfo]:
+        """Retrieves newest schema from destination storage."""
+        try:
+            version_table_name = self.make_qualified_table_name(self.schema.version_table_name)
+            response = (
```
sh-rp, line 410: same as above, you need to sort to make the newest one be the first, otherwise this will not work properly

dlt/destinations/impl/lancedb/configuration.py
Pipboyguy: Fixed!

dlt/destinations/impl/lancedb/configuration.py
Pipboyguy: I remove the prefix so now its just `.lancedb` . This good?

dlt/destinations/impl/lancedb/errors.py
Pipboyguy: LanceDB is similar to qdrant and weaviate and doesn't follow sql  semantics or offer a DB API 2.0 specification. Looking at qdrant and weaviate they also only raise `DestinationException`.

Let me know your thoughts?

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: It is indeed possible!

Kindly review latest changes that addresses decimal and timestamp precision conversion.

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Changed it to 

```python
    @property
    def sentinel_table(self) -> str:
        return self.make_qualified_table_name("dltSentinelTable")
```

Please advise if this is good.

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Ideally yes, but the loadpackages loading utilities have a bunch of extra stuff like "special" fields and rather intricate workings to convert to pydantic models.

Can I request for this to be put as a separate ticket? Its not immediately obvious to me how to create a single interface here

dlt/destinations/impl/lancedb/lancedb_client.py
sh-rp: yes, thanks :)

dlt/destinations/impl/lancedb/exceptions.py
```diff
@@ -0,0 +1,62 @@
...
+)
+from dlt.common.destination.reference import JobClientBase
+from dlt.common.typing import TFun
+
+
+class LanceDBBatchError(DestinationException):
+    pass
+
+
+def lancedb_error(f: TFun) -> TFun:
```
sh-rp, line 23: I thought I left a comment about this somewhere: Why can we not use the same error handling mechanism we use on all the other destinations: @raise_database_error?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+                data_type="time",
+                precision=UNIT_TO_TIMESTAMP_PRECISION[db_type.unit],
+                scale=scale,
+            )
+        if isinstance(db_type, pa.Decimal128Type):
+            precision, scale = db_type.precision, db_type.scale
+            if (precision, scale) == self.capabilities.wei_precision:
+                return cast(TColumnType, dict(data_type="wei"))
+            return dict(data_type="decimal", precision=precision, scale=scale)
+        return cast(
```
sh-rp, line 144: why not call super here? this code is implemented there isn't it?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+
+    Raises:
+        ValueError: If the write disposition is unsupported, or `id_field_name` is not
+            provided for update/merge operations.
+    """
+
+    try:
+        tbl = db_client.open_table(table_name)
+    except FileNotFoundError as e:
+        raise LanceDBBatchError(e) from e
```
sh-rp, line 183: I'm assuming this is done like this because the filenotfounderror means different things in different lancedb contexts right? You could directly raise the correct dlt error here.

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+        self.vector_field_name = self.config.vector_field_name
+        self.id_field_name = self.config.id_field_name
+
+    @property
+    def dataset_name(self) -> str:
+        return self.config.normalize_dataset_name(self.schema)
+
+    @property
+    def sentinel_table(self) -> str:
+        return self.make_qualified_table_name("dltSentinelTable")
```
sh-rp, line 245: i think the name of the sentinel table should be configurable, I think it is done like this in qdrant and weavieate, no?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+            arrow_schema = self.db_client.open_table(fq_table_name).schema
+        except FileNotFoundError:
+            return False, table_schema
+        except Exception:
+            raise
+
+        for field_name, field_type in arrow_schema_to_dict(arrow_schema).items():
+            schema_c: TColumnSchema = {
+                "name": self.schema.naming.normalize_identifier(field_name),
+                **self._from_db_type(field_type, None, None),
```
sh-rp, line 397: Do you not need to send precision data from the arrow schema to the typemapper here? You will loose this info if you don't, I think tests should fail on this too

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+
+        Since arrow tables are immutable, this is done via a staging mechanism.
+        The data is stored in-memory in a staging arrow table, evolved then stored
+        written over the old table.
+
+        Args:
+            table_name: The name of the table to create the field on.
+            field_schema: The field to create.
+        """
+        arrow_table = self.db_client.open_table(table_name).to_arrow()
```
sh-rp, line 414: does this mean all data needs to be loaded from the destination to create a field? we should definately add that to the docs page in a performance section or something like that.

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+            )
+            logger.info(f"Found {len(new_columns)} updates for {table_name} in {self.schema.name}")
+            if len(new_columns) > 0:
+                if exists:
+                    for column in new_columns:
+                        field_schema: TArrowField = make_arrow_field_schema(
+                            column["name"], column, self.type_mapper, embedding_fields
+                        )
+                        fq_table_name = self.make_qualified_table_name(table_name)
+                        self.add_table_field(fq_table_name, field_schema)
```
sh-rp, line 445: when adding multiple new fields this code will load the full table, add a new column, save it and repeat for every column. I think .add_table_field should take an array of new columns and add them at once. Or is there some other reason you solved it this way?

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+        )
+        if len(state_records) == 0:
+            return None
+        for state in state_records:
+            load_id = state["_dlt_load_id"]
+            # If there is a load for this state which was successful, return the state.
+            if (
+                self.db_client.open_table(fq_loads_table_name)
+                .search()
+                .where(f'load_id = "{load_id}"')
```
sh-rp, line 520: you need to filter by status = 0 also, other entries are not considered successful loads

sh-rp: You can check out get_stored_state from the job_client_impl. You need to somehow replicate the join that is done there with the tables that you have here. The code of the other vector DBs will help here.

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+        self.write_disposition: TWriteDisposition = cast(
+            TWriteDisposition, self.table_schema.get("write_disposition", "append")
+        )
+
+        records: List[DictStrAny]
+        with FileStorage.open_zipsafe_ro(local_path) as f:
+            try:
+                records = json.load(f)
+            except JSONDecodeError:
+                # If parsing as a single object fails, try a line at a time.
```
sh-rp, line 654: you don't need the try except block here, just always go line by line

dlt/destinations/impl/lancedb/schema.py
```diff
@@ -0,0 +1,84 @@
+"""Utilities for creating Pydantic model schemas from table schemas."""
```
sh-rp, line 1: you mean pyarrow schema right?

dlt/destinations/impl/lancedb/utils.py
```diff
@@ -0,0 +1,53 @@
...
+def list_unique_identifiers(table_schema: TTableSchema) -> Sequence[str]:
+    """Returns a list of unique identifiers for a table.
+
+    Args:
+        table_schema (TTableSchema): a dlt table schema.
+
+    Returns:
+        Sequence[str]: A list of unique column identifiers.
+    """
+    if table_schema.get("write_disposition") == "merge":
```
sh-rp, line 43: i think if merge keys are defined this should be computed on the merge keys. This whole code is meant for merging right?

dlt/destinations/impl/lancedb/utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+PROVIDER_ENVIRONMENT_VARIABLES_MAP: Dict[TEmbeddingProvider, str] = {
+    "cohere": "COHERE_API_KEY",
+    "gemini-text": "GOOGLE_API_KEY",
+    "openai": "OPENAI_API_KEY",
+    "huggingface": "HUGGINGFACE_API_KEY",
+}
+
+
+def generate_uuid(data: DictStrAny, unique_identifiers: Sequence[str], table_name: str) -> str:
```
sh-rp, line 19: this code seems to be a copy from the other vector DBs, generally speaking if this happens, we should extract it into a utils class that is used by all destinations, for now let's keep it, I think some refactoring is due at some point in any case.

tests/load/lancedb/test_pipeline.py
```diff
@@ -0,0 +1,371 @@
...
+        some_data(),
+        write_disposition="replace",
+    )
+    assert_load_info(info)
+
+    data = next(generator_instance2)
+    assert_table(pipeline, "some_data", items=data)
+
+
+def test_pipeline_merge() -> None:
```
sh-rp, line 197: test with merge keys also

sh-rp (changes requested):
i have a bunch of requests

dlt/destinations/impl/lancedb/exceptions.py
Pipboyguy: @sh-rp I don't think it will work because lancedb doesn't have a DBAPI driver we can use, i.e. doesn't inheret from SqlClientBase

Pipboyguy: Let me know if I'm missing something?

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: I overrode `from_dbt_type` as associated methods because we have to return a `pa.DataType` instance for `LanceDBTypeMapper` to work correctly.

In addition to this, the timestamp precision in pyarrow has to be accounted for, the constructors don't just take a scalar but a string. Please see the unit parameter at https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html

Pipboyguy:
```python
...
        return cast(
            TColumnType,
            without_none(
                dict(
                    data_type=self.dbt_to_sct.get(db_type, "text"),
                    precision=precision,
                    scale=scale,
                )
            ),
        )
```

I did replace the above with `return super().from_db_type(db_type, precision, scale)`  though to reduce redundancy. Thanks for the spot!

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -0,0 +1,698 @@
...
+
+    def get_storage_table(self, table_name: str) -> Tuple[bool, TTableSchemaColumns]:
+        table_schema: TTableSchemaColumns = {}
+
+        try:
+            fq_table_name = self.make_qualified_table_name(table_name)
+            arrow_schema = self.db_client.open_table(fq_table_name).schema
+        except FileNotFoundError:
+            return False, table_schema
+        except Exception:
```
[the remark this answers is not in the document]
Pipboyguy: You're right. I removed all of these bare except clauses.

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: I removed `LanceDBBatchError` class and directly raised the appropriate dlt destination errors as requested

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Doesn't seem like they are:

https://github.com/dlt-hub/dlt/blob/23c6365310d60ed86e16b25196ae4e41e71e8b3b/dlt/destinations/impl/weaviate/weaviate_client.py#L264

But I've made it configurable now: https://github.com/dlt-hub/dlt/blob/bbc00a5237f08d18a5d81c04caf6cd25bf84fe94/dlt/destinations/impl/lancedb/configuration.py#L88

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: The precision and scale, if present, is obtained from the arrow schema by way of `from_db_type`:

https://github.com/dlt-hub/dlt/blob/9322477461288b77ee317232ee62fcc0897ab969/dlt/destinations/impl/lancedb/lancedb_client.py#L122-L145

To make this clearer to the reader, I've removed the None arguments:

https://github.com/dlt-hub/dlt/blob/9322477461288b77ee317232ee62fcc0897ab969/dlt/destinations/impl/lancedb/lancedb_client.py#L393-L397

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Will do.

I couldn't find a 'lancedb' way to do cloning. I'm sure there is a better way to do it. Shall we add an issue for this?

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Thanks for the spot!

There was no reason to do it this way, I've add supported for multiple fields in a single operation now with `add_table_fields`:

https://github.com/dlt-hub/dlt/blob/9b4c5196436f54c73a7fa61a8aa33bf5fe37f3b0/dlt/destinations/impl/lancedb/lancedb_client.py#L393-L422

Now the arrow table is mutated in memory, instead of read from the database n times over.

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Oh yes its already wrapped in a `JSONDecodeError` 🤦

Thanks! Fixed

Pipboyguy:
LanceDB's ORDER BY clause not working as expected:

```python
import lancedb
db = lancedb.connect("/tmp/.lancedbx")

data = [
    {"id": 1, "value": 10},
    {"id": 2, "value": 5},
    {"id": 3, "value": 15},
    {"id": 4, "value": 7}
]
db.create_table("test_table", data, mode="overwrite")

print("Ascending order:")
result = db["test_table"].search().select(["id", "value"]).where("1=1 ORDER BY value ASC").to_list()
print(result)

print("\nDescending order:")
result = db["test_table"].search().select(["id", "value"]).where("1=1 ORDER BY value DESC").to_list()
print(result)
```

```
Ascending order:
[{'id': 1, 'value': 10}, {'id': 2, 'value': 5}, {'id': 3, 'value': 15}, {'id': 4, 'value': 7}]

Descending order:
[{'id': 1, 'value': 10}, {'id': 2, 'value': 5}, {'id': 3, 'value': 15}, {'id': 4, 'value': 7}]
```

I'm refactoring all state and schema getters to use either in-memory arrow or python data structures

dlt/destinations/impl/lancedb/exceptions.py
sh-rp: Ah yes, you are right, we should make this decorator more universal, forget about it for now ;)

dlt/destinations/impl/lancedb/lancedb_client.py
sh-rp: yes, that's whay I meant I think, thanks

Pipboyguy:
When using the OpenAI API for embeddings, passing empty strings in the input can cause a BadRequestError 
with the message "'$.input' is invalid". This is because the API does not accept empty strings.

```python
from lancedb.embeddings import OpenAIEmbeddings

os.environ["OPENAI_API_KEY"] = "..."

embedding_func = OpenAIEmbeddings()
texts = ["hello", "", "world"]
embeddings = embedding_func.generate_embeddings(texts)
```

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "'$.input' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference.", 'type': 'invalid_request_error', 'param': None, 'code': None}}
```

I patched this in:

https://github.com/dlt-hub/dlt/blob/e08bb8aa013c2ac910b757e39b318080c749262d/dlt/destinations/impl/lancedb/models.py#L10

sh-rp:
@Pipboyguy we still need to do the following:
* Fix state retrieval (you are not getting the newest load id in get_stored_state, so the results are unpredictable)
* Create a lancedb extra (now it's in the main dependencies). The cohere dependency can completely go, you should add a section in the lancedb docs that for certain embedding providers additional dependencies are needed. Maybe there is a page in the lance docs that shows this that you can link to
* We need openai dependency on CI, just add it as a pip install in the lancedb workflow.

tests/load/lancedb/utils.py
```diff
@@ -0,0 +1,96 @@
...
+        dlt.config.get("destination.lancedb.credentials.vector_field_name", str) or "vector__",
+    ]
+    objects_without_dlt_or_special_keys = [
+        {k: v for k, v in record.items() if k not in drop_keys} for record in records
+    ]
+
+    assert_unordered_dicts_equal(objects_without_dlt_or_special_keys, items)
+
+
+def drop_active_pipeline_data() -> None:
```
sh-rp, line 68: quick question: why can't you use the one that is used in all other places?

sh-rp: ok, I see that this is done like this in the other vector destinations, another thing I'll have to look at when extracting common stuff from the vectordbs

akelad:
Hey @Pipboyguy @sh-rp I'm just testing this out a bit to see how this works. Two questions so far:

1. I can't omit the definition of the api_key and embedding_api_key, even though those aren't strictly required. So to go around it this is what I put in the toml file, would be great if those could be marked as optional:
```
[destination.lancedb.credentials]
uri = ".lancedb"
api_key = ""
embedding_model_provider_api_key = ""
```

2. I actually can't get the lancedb_adapter to work properly so far. I did a `dlt init rest_api lancedb`, which doesn't produce the code to use the lancedb_adapter, it just treats lancedb as a normal db. Not sure how `dlt init` works and whether it's even feasible to adjust that. But either way I started trying to work with the pokemon declarative API source as follows:
```
def load_pokemon() -> None:
    pipeline = dlt.pipeline(
        pipeline_name="rest_api_pokemon",
        destination='lancedb',
        dataset_name="poke_data",
    )

    pokemon_source = rest_api_source(
        {
            "client": {
                "base_url": "https://pokeapi.co/api/v2/",
                # If you leave out the paginator, it will be inferred from the API:
                # paginator: "json_response",
            },
            "resource_defaults": {
                "endpoint": {
                    "params": {
                        "limit": 1000,
                    },
                },
            },
            "resources": [
                "pokemon",
                "berry",
                "location",
            ],
        }
    )
    load_info = pipeline.run(lancedb_adapter(pokemon_source.resources['pokemon'], embed="name"))
```
and I keep getting an error: `TypeError: 'module' object is not callable`. What am I doing wrong here? Side note - I understood you can only pass a resource if you also want to embed, and not a full source - is that correct?

akelad:
Actually if I follow the exact example you have in the docs I can't get it to work either 🤔

akelad:
Update: if I change the lancedb_adapter to `from dlt.destinations.impl.lancedb.lancedb_adapter import lancedb_adapter` the script runs. The docs say `from dlt.destinations.adapters import lancedb_adapter`, and that gave me the TypeError

Pipboyguy:
@akelad the `api_key` and `embedding_model_provider_api_key` have been marked as optional now. Please see associated tests as well. Mind checking whether it works for you?

akelad:
@Pipboyguy I'll try it out, thanks. Can you also take a look at my [other comments](https://github.com/dlt-hub/dlt/pull/1375#issuecomment-2178955609)? Both Anuun and I get the `TypeError: 'module' object is not callable` when using `from dlt.destinations.adapters import lancedb_adapter` - that should be fixed.

akelad:
Can confirm it works without defining the api/embedding api key

Pipboyguy:
@akelad 

> I actually can't get the lancedb_adapter to work properly so far. I did a dlt init rest_api lancedb, which doesn't produce the code to use the lancedb_adapter, it just treats lancedb as a normal db. Not sure how dlt init works and whether it's even feasible to adjust that.

Not all fields contain search-relevant information, and fields may require distinct embedding functions, especially for multi-modal data. Users must explicitly specify source fields for embeddings in their schema to ensure optimal performance and accuracy in vector search operations.

This is why the adapter has to be written by the user that understands what they are trying to achieve. If we want the adapter auto generated for the example we'd have to add this to the `init` command manually for each example as you pointed out.

I'm not sure how to do this though. @rudolfix @sh-rp ?

Pipboyguy:
> I understood you can only pass a resource if you also want to embed, and not a full source - is that correct?

💯

Embeddings are computed on a field by field basis. The user needs to specify which fields they want to embed so that when they perform semantic search they understand what results they are getting back. (i.e. don't want some query to match on the dataset's ID field, and don't want to embed this field either)

I suppose we could just embed all of them by default, but this will definitely hit the provider API hard (5 times as much if there are 5 fields). Some providers do somehow seem to combine the embeddings up together into one vector but I don't know whether this is useful.

So to answer your question, I think we should keep as is and require the user to specify the embedding source fields of each resource, just like lancedb does it.

Pipboyguy:
> Update: if I change the lancedb_adapter to from dlt.destinations.impl.lancedb.lancedb_adapter import lancedb_adapter the script runs. The docs say from dlt.destinations.adapters import lancedb_adapter, and that gave me the TypeError

Thanks for this catch @akelad !

This should be fixed now. Kindly test again?

akelad:
The fixed import works now! About the dlt init command - that's fine with me, maybe we should add some sort of note either in the generated code or as a warning that you have to add the lancedb_adapter for embeddings? Because out of the box it will just act as a normal DB. @sh-rp what do you think?

> Actually it seems that you can. Like you demonstrated above ( also ran it on my end), the adapter seems to work just fine with the pokemon source.

Actually the example I showed there only runs on a resource: `pokemon_source.resources['pokemon']`. I don't think it works properly with a full source, at least when I tried it last week it did run, but there was some weird behaviour.  I think we need to make it extra clear in the docs that this is intended to be used with resources and not on a source level - I can try and make those docs updates.

Pipboyguy:
@akelad thank you so much for the detailed investigation.

Apologies, you are right the following does indeed work:

```python
    lancedb_adapter(pokemon_source.resources['pokemon'], embed=["name"])
    load_info = pipeline.run(pokemon_source)
```

whereas the following doesn't raise a warning or exception, but nothing is embedded:

```python
    lancedb_adapter(pokemon_source, embed=["name"])
    load_info = pipeline.run(pokemon_source)
```

PR #1377, opened 2024-05-16
Azure service principal credentials

steinitzu:
### Description
Adds support for authenticating Azure with service principal (client ID/secret/tenant ID):

```toml
[destination.filesystem]
bucket_url = "az://[your_container name]" # replace with your container name

azure_client_id = "client_id" # please set me up!
azure_client_secret = "client_secret
azure_tenant_id = "tenant_id" # please set me up!
```

### Related Issues
 - Resolves https://github.com/dlt-hub/dlt/issues/1378


rudolfix (changes requested):
@steini LGTM

just one thing: ping me on Slack and let's add test credentials to the vault and please write a test that just connects to out accounts and does a few ops. you can use our fsspec abstraction to do that

rudolfix (approved):

PR #1392, opened 2024-05-21
Fix snippet linting errors

sh-rp:
### Description
Some snippets needed fixing


burnash (dismissed):
Awesome, thank you @sh-rp!

PR #1398, opened 2024-05-22
RESTClient: implement AuthConfigBase.__bool__ + update docs

burnash:
This PR implements __bool__ on AuthConfigBase so instances of this class evaluate to True in requests.sessions.Session.prepare_request() when auth argument is evaluated otherwise instances of AuthConfigBase have no effect when passed as arguments to RESTClient's methods.

The PR also updates the documentation to suggest explicit inheritance from `requests.auth.AuthBase` for custom authentication.


sultaniman (approved):
LGTM

rudolfix (approved):
so good!

PR #1399, opened 2024-05-22
RESTClient: Added "values" to the data pattern of the rest_api helper

francescomucio:
### Description
Quick PR to add the value `values` to the path recognized automatically as data in an API response. This is actually used by the JIRA api, but it looks like a quite generic name

### Related Issues

None


burnash (approved):
Awesome, thanks @francescomucio!

Issue #555, opened 2023-08-14
Partitions in AWS Athena 


sh-rp:
We want to enable the user to make use of AWS Athena partitions. For this we will most likely need to change the schema a bit and add the option for the user to define partitions on the resources. 

Tasks:
- [ ] Alter schema to have an order of partitions (possibly replace the "bool" hint with an "int" hint which is null or 0 if this column should not be a partition, and a number if it should which then defines the order?)
- [ ] Expose partition config on the resource as array of strings which are the column names?
- [ ] When creating parquet files, respect the partitions and create a file for every partition key combination and upload that to the right path.
- [ ] Make the partitions known to Athena, either we can run explicit ddl commands, but it should also be possible to infer the partitions with a repair command, the latter is more costly though, so doing this with our schema makes sense. This means we will have schema m migration operations that not only add columns but also partition definitions.

To think about:
- [ ] How to guard against the user uploading to the same bucket with different partition settings?
- [ ] How to migrate existing schemas if we change the type of the partition settings?
- [ ] How to behave if there are partitions defined, but the yielded resource item does not have a value for this column? Will there always be a "None" path for those cases? Will this be queryable by Athena?

gamgi:
Documenting a potential use-case. I'm loading files from S3 + SFTP sources to S3 destination.

I'd like to use properties from the source (not data itself) to design the partition layout in the destination. The properties may be resource *arguments*, or indirectly source metadata like  *object path* and/or *file name*.

Examples:
* Single daily object from SFTP
  * `sftp://example.com/foo/bar/report_20231114000000.csv` --> `s3://destination-bucket/baz/quz/yyyy=2023/mm=11/dd=14/report.csv`
* Multiple daily objects from S3
  * `s3://example-bucket/foo/bar/report_20231114000000.csv` --> `s3://destination-bucket/baz/quz/yyyy=2023/mm=11/dd=14/report_20231114000000.csv`
  * `s3://example-bucket/foo/bar/report_20231114005000.csv` --> `s3://destination-bucket/baz/quz/yyyy=2023/mm=11/dd=14/report_20231114005000.csv`
* Single daily API call to S3
  * `https://example.com/api/report?date=2023-11-14` --> `s3://destination-bucket/baz/quz/yyyy=2023/mm=11/dd=14/report.csv`


Implications for this ticket:
* Partition config should support resource parameters
    * For example, if `date` (or `year, month, date`) is a parameter to a resource, that should be available in the partition layout.
[...]

PR #1403, opened 2024-05-23
Support partitioning hints for athena iceberg

steinitzu:
### Description
Adds `athena_adapter` to add partitioning hints in athena. Generates `PARTITIONED BY` clause on iceberg tables.

The adapter generates a table hint: `x-athena-partition`.
This is a dict in form of `{<column_name>: <template_string>}`  
E.g. `{"department": "{column_name}", "date_hired": "year({column_name})"}`  

Using format strings so we can generate transform (`bucket(...)`, `year(...)`, etc) strings and inject the escaped column name later in the loader.

### Related Issues

Implements partially #555 -> Iceberg support

### DIFF
--- a/tests/load/athena_iceberg/test_athena_adapter.py
+++ b/tests/load/athena_iceberg/test_athena_adapter.py
@@ -57,9 +57,7 @@ def test_iceberg_partition_hints():
         )[0]
 
     # Partition clause is generated with original order
-    expected_clause = (
-        "PARTITIONED BY (category, month(created_at), bucket(10, product_id), truncate(2, name))"
-    )
+    expected_clause = "PARTITIONED BY (`category`, month(`created_at`), bucket(10, `product_id`), truncate(2, `name`))"
     assert expected_clause in sql_partitioned
 
     # No partition clause otherwise
--- a/tests/load/athena_iceberg/test_athena_adapter.py
+++ b/tests/load/athena_iceberg/test_athena_adapter.py
@@ -1,0 +1,66 @@
+import pytest
+
+import dlt
+from dlt.destinations import filesystem
+from dlt.destinations.impl.athena.athena_adapter import athena_adapter, athena_partition
+from tests.load.utils import destinations_configs, DestinationTestConfiguration
+
+
+def test_iceberg_partition_hints():
+    """Create a table with athena partition hints and check that the SQL is generated correctly."""
+
+    @dlt.resource(table_format="iceberg")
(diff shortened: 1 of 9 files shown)


tests/load/athena_iceberg/test_athena_adapter.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+
+import dlt
+from dlt.destinations import filesystem
+from dlt.destinations.impl.athena.athena_adapter import athena_adapter, athena_partition
+from tests.load.utils import destinations_configs, DestinationTestConfiguration
+
+
+def test_iceberg_partition_hints():
```
rudolfix, line 9: should we use `destination_config` to filter for `athena` so we run it only when athena tests run on CI?

tests/load/pipeline/test_athena.py
```diff
@@ -231,3 +233,68 @@ def test_athena_file_layouts(destination_config: DestinationTestConfiguration, l
...
     )
     assert table_counts == {"items1": 3, "items2": 7}
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["athena"], force_iceberg=True),
+    ids=lambda x: x.name,
+)
+def test_athena_partitioned_iceberg_table(destination_config: DestinationTestConfiguration):
```
rudolfix, line 243: please mark this test with `essentials` so we run it on a each CI run

rudolfix (changes requested):
code and docs: LGTM!
I think we can reconfigure tests, see my review

tests/load/athena_iceberg/test_athena_adapter.py
steinitzu: The whole package is skipped by: https://github.com/dlt-hub/dlt/blob/athena-partitions/tests/load/athena_iceberg/__init__.py#L1-L4

tests/load/pipeline/test_athena.py
steinitzu: The module is marked https://github.com/dlt-hub/dlt/blob/athena-partitions/tests/load/pipeline/test_athena.py#L23-L24

steinitzu: Or was this meant for the `athena_adapter` test?

rudolfix: heh I meant here... so it seems that all is good

PR #1407, opened 2024-05-24
Feat/1406 bumps duckdb 0.10

rudolfix:
### Description
1. bumps duckdb to 0.10
2. enables dbt 1.8

### DIFF
--- a/dlt/helpers/dbt/__init__.py
+++ b/dlt/helpers/dbt/__init__.py
@@ -11,7 +11,7 @@ from dlt.version import get_installed_requirement_string
 
 from dlt.helpers.dbt.runner import create_runner, DBTPackageRunner
 
-DEFAULT_DBT_VERSION = ">=1.1,<1.6"
+DEFAULT_DBT_VERSION = ">=1.5,<1.9"
 
 # a map of destination names to dbt package names in case they don't match the pure destination name
 DBT_DESTINATION_MAP = {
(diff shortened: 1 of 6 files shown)


dlt/sources/helpers/rest_client/client.py
```diff
@@ -57,7 +59,7 @@ class RESTClient:
     Args:
         base_url (str): The base URL of the API to make requests to.
         headers (Optional[Dict[str, str]]): Default headers to include in all requests.
-        auth (Optional[AuthConfigBase]): Authentication configuration for all requests.
+        auth (Optional[AuthBase]): Authentication configuration for all requests.
```
burnash, line 62: Will it work with out existing authenticators from `dlt.sources.helpers.rest_client.auth` which derive from `AuthConfigBase`? E.g., from our example code:

```py
from dlt.sources.helpers.rest_client import RESTClient
from dlt.sources.helpers.rest_client.auth import BearerTokenAuth

client = RESTClient(
    base_url="https://api.example.com",
    auth=BearerTokenAuth(token="your_access_token_here")  # type: ignore
)

for page in client.paginate("/protected/resource"):
    print(page)
```

Shouldn't it be a union of `AuthConfigBase` and `AuthBase` instead?

burnash (changes requested):

dlt/sources/helpers/rest_client/client.py
rudolfix: this is due to linting problems that I merged from #1398 so I reverted it and we continue this review in #1413

PR #1413, opened 2024-05-25
RESTClient: implement AuthConfigBase.__bool__ + update docs

rudolfix:
### Description
This PR is resubmitted because I merged the last one with typing problems that made linting to fail. Let's try to fix our qdrant problem so we see CI green again...

linting fixes included
- using correct types in Auths defined in tests
- changing the types for auth in `RESTClient` to `AuthBase` which is a common base for `AuthConfigBase` and the one used by requests (`AuthBase` itself)

all tests including those in rest api source are passing after this

----
This PR implements __bool__ on AuthConfigBase so instances of this class evaluate to True in requests.sessions.Session.prepare_request() when auth argument is evaluated otherwise instances of AuthConfigBase have no effect when passed as arguments to RESTClient's methods.

The PR also updates the documentation to suggest explicit inheritance from requests.auth.AuthBase for custom authentication.

### DIFF
--- a/dlt/sources/helpers/rest_client/client.py
+++ b/dlt/sources/helpers/rest_client/client.py
@@ -6,12 +6,14 @@ from typing import (
     Any,
     TypeVar,
     Iterable,
+    Union,
     cast,
 )
 import copy
 from urllib.parse import urlparse
 from requests import Session as BaseSession  # noqa: I251
 from requests import Response, Request
+from requests.auth import AuthBase
 
 from dlt.common import jsonpath, logger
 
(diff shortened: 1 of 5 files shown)


dlt/sources/helpers/rest_client/client.py
```diff
@@ -57,7 +59,7 @@ class RESTClient:
     Args:
         base_url (str): The base URL of the API to make requests to.
         headers (Optional[Dict[str, str]]): Default headers to include in all requests.
-        auth (Optional[AuthConfigBase]): Authentication configuration for all requests.
+        auth (Optional[AuthBase]): Authentication configuration for all requests.
```
burnash, line 62: @rudolfix should't this be a Union type?

burnash: resolved

PR #1423, opened 2024-05-28
1272 Support ClickHouse GCS S3 compatibility mode in filesystem destination

### LINKED ISSUE #1181: HMAC keys aren't implemented in `GcpCredentials`
### dlt version

0.4.7

### Describe the problem

Clickhouse can only use HMAC keys to ingest from GCS. `GcpCredentials` Needs to allow for HMAC keys.

See: 

- https://clickhouse.com/docs/en/sql-reference/table-functions/gcs
- https://cloud.google.com/storage/docs/authentication/managing-hmackeys#create

### Expected behavior

HMAC keys needs to be read from

```toml
gcp_access_key_id = "..."
gcp_secret_access_key = "..."
```

This is the equivalent of aws access key

### Steps to reproduce

A temporary workaround is to read from config file `dlt.config["destination.filesystem.credentials.gcp_access_key_id"]`. Ideally this should be part of the official google credential class.

### Operating system

Linux, macOS, Windows

### Runtime environment

Local

### Python version

3.11

### dlt data source

All

### dlt destination

_No response_

### Other deployment details

Clickhouse

### Additional information

_No response_

### LINKED ISSUE #1272: Make filesystem destination work with gcs in s3 compatibility mode
GCS has an s3 interoperability mode which emulates s3. This way users could use the s3 bucket filesystem implementation with gcs. This is useful when loading to a final destination which supports connecting to s3 buckets but not gcs, such as clickhouse. Clickhouse now has a custom implementation to handle this, but it would be nice if our filesystem could be set up to use the s3 implementation with gcs buckets, right now based on the bucket url, gcs will always be selected automatically.

If this is done, we can update clickhouse and remove the specialized implementation.

Pipboyguy:
### Description

Updates the filesystem destination to support using the S3 implementation with GCS buckets that have S3 interoperability mode (HMAC keys) enabled. 

Key changes:
- Update `convert_storage_to_http_scheme` utility function to properly handle GCS endpoints
- Modify ClickHouse load job to raise an exception if GCS bucket is used without providing S3 compatible credentials (access key ID and secret access key)
- Add test case to verify ClickHouse loading from GCS bucket using S3 compatibility mode
- Remove old gcs tests. These will be added back in when named destinations are supported in templated tests.

### Related Issues

- Resolves #1181
- Fixes #1272


rudolfix (changes requested):
code: LGTM!
@Pipboyguy do we have the right credentials in our CI so this test is really connecting to clickhouse?

@sh-rp you should review this

docs/website/docs/dlt-ecosystem/destinations/clickhouse.md
```diff
@@ -173,51 +175,42 @@ pipeline = dlt.pipeline(
 )
 ```
 
-### Using Google Cloud Storage as a Staging Area
+### Using S3-Compatible Storage as a Staging Area
```
sh-rp, line 178: maybe mention google cloud storage in the title here too

tests/load/clickhouse/test_clickhouse_gcs_s3_compatibility.py
```diff
@@ -0,0 +1,28 @@
...
+import pytest
+
+import dlt
+from dlt.destinations import filesystem
+from tests.load.utils import GCS_BUCKET
+from tests.pipeline.utils import assert_load_info
+
+
+@pytest.mark.essential
+def test_clickhouse_gcs_s3_compatibility() -> None:
```
sh-rp, line 12: looks good!

sh-rp:
> code: LGTM! @Pipboyguy do we have the right credentials in our CI so this test is really connecting to clickhouse?
> 
Yeah I put those in yesterday. The question you were asking was wether it is really connecting to GCS, right?

Pipboyguy:
> code: LGTM! @Pipboyguy do we have the right credentials in our CI so this test is really connecting to clickhouse?
> 
> @sh-rp you should review this

@rudolfix  We are indeed connecting to GCS with HMAC credentials and [test passes](https://github.com/dlt-hub/dlt/actions/runs/9285866234/job/25551337483?pr=1423#step:8:43)

PR #1425, opened 2024-05-30
Fix streamlit bug on chess example

sh-rp:
### Description
Fixes a bug in the streamlit app which would produce an error if the nullable or datatype hints on a column where missing.


dlt/helpers/streamlit_app/blocks/table_hints.py
```diff
@@ -64,17 +64,21 @@ def list_table_hints(pipeline: dlt.Pipeline, tables: List[TTableSchema]) -> None
 
         # table schema contains various hints (like clustering or partition options)
         # that we do not want to show in basic view
-        def essentials_f(c: Any) -> Dict[str, Any]:
-            essentials: Dict[str, Any] = {}
-            for k, v in c.items():
-                if k in ["name", "data_type", "nullable"]:
-                    essentials[k] = v
-
+        def essentials_f(c: TColumnSchema) -> Dict[str, Any]:
```
rudolfix, line 67: looks like this function is not used anymore?

rudolfix (approved):

PR #1428, opened 2024-05-31
Example: fast postgres to postgres

AstrakhantsevaAA:
tested example , fixed bugs, typos and reformatted

- Created schema `example_data_1` in tests Postgres database (which we are using as the destination for ci-cd)
- Loaded tests data `table_1` and `table_2`
- Formatted example accordingly CONTRIBUTION.md
- Tested


Makefile
```diff
@@ -70,6 +70,10 @@ lint-and-test-examples:
 	poetry run mypy --config-file mypy.ini docs/examples
 	poetry run flake8 --max-line-length=200 docs/examples
 	cd docs/tools && poetry run python prepare_examples_tests.py
+	# install duckdb extension for postgres_to_postgres example
```
rudolfix, line 73: since postgres is now official extension IMO all those scripts can be removed
https://duckdb.org/docs/extensions/postgres

you just execute
```sql
INSTALL postgres;
```

and that we can do in the example script. this will be a huge relief because the code below binds the extension to particular duckdb version and will prevent many people from running the script (needs wget, extension works only on linux)

docs/examples/postgres_to_postgres/postgres_to_postgres.py
```diff
@@ -0,0 +1,292 @@
+"""
+---
+title: Load from Postgres to Postgres faster
+description: Load data fast from Postgres to Postgres with ConnectorX & Arrow export as Parquet, normalizing and exporting as DuckDB, and attaching it to Postgres for bigger Postgres tables (GBs)
+keywords: [connector x, pyarrow, zero copy, duckdb, postgres, initial load]
+---
+
+This examples shows you how to export and import data from Postgres to Postgres in a fast way with ConnectorX and DuckDB
```
rudolfix, line 8: We should credit the user that submitted that :)

rudolfix (changes requested):
@AstrakhantsevaAA this is so good! I think we can radically simplify extension installation. pls see my comments

rudolfix (approved):
LGTM!

PR #1429, opened 2024-05-31
Ensure arrow field's nullable flag matches the schema column

steinitzu:
### Description

Updates nullability of fields in arrow schema as part of `normalize_py_arrow_item`.   
Fixes issue that affects at least sql database source when pandas backend and bigquery destination is used. Any resource yielding dataframes when pk/non-nullable columns are defined has the same problem, e.g.

```python
import dlt
import pandas as pd


@dlt.resource(primary_key="col1")
def some_data():
    yield pd.DataFrame({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})

# Fails due to schema mismatch
info = dlt.pipeline(full_refresh=True, destination="bigquery").run(some_data)
print(info)
```

Fails with:

```
Provided Schema does not match Table serene-anagram-189621:dlt_ipython_dataset_20240531072948.some_data. 
Field col1 has changed mode from REQUIRED to NULLABLE
```

### Related Issues
Somewhat related to https://github.com/dlt-hub/verified-sources/issues/430 , part of the reason bigquery fails

### DIFF
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -223,9 +223,18 @@ def should_normalize_arrow_schema(
     schema: pyarrow.Schema,
     columns: TTableSchemaColumns,
     naming: NamingConvention,
-) -> Tuple[bool, Mapping[str, str], Dict[str, str], TTableSchemaColumns]:
+) -> Tuple[bool, Mapping[str, str], Dict[str, str], Dict[str, bool], TTableSchemaColumns]:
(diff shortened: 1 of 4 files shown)


rudolfix (approved):
LGTM! is this PR still draft?

steinitzu:
> LGTM! is this PR still draft?

@rudolfix no just didn't have time to wait for all tests to run.  
You can merge if it looks good to you 🙂

PR #1436, opened 2024-06-03
update dependencies to fix bug on databricks has_dataset

oonyoontong:
### Description
Fix error in parsing results back from the underlying databricks sql client by updating dependencies.

> Traceback (most recent call last):
  File "dlt\destinations\sql_client.py", line 243, in _wrap_gen
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "dlt\destinations\impl\databricks\sql_client.py", line 115, in execute_query
    yield DBApiCursorImpl(curr)  # type: ignore[abstract]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "dlt\destinations\impl\databricks\sql_client.py", line 84, in execute_sql   
    f = curr.fetchall()
        ^^^^^^^^^^^^^^^
  File "databricks\sql\client.py", line 670, in fetchall
    return self.active_result_set.fetchall()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "databricks\sql\client.py", line 944, in fetchall
    return self._convert_arrow_table(self.fetchall_arrow())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "databricks\sql\client.py", line 884, in _convert_arrow_table
    res = df.to_numpy(na_value=None)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas\core\frame.py", line 1993, in to_numpy
    result = self._mgr.as_array(dtype=dtype, copy=copy, na_value=na_value)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas\core\internals\managers.py", line 1703, in as_array
    arr[isna(arr)] = na_value
    ~~~^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

<!--
Provide any additional context about the PR here.
-->
### Additional Context
https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1716857509703029
<!--
Please ensure that
    - you have read the [Contributing to dlt](../CONTRIBUTING.md) guide.
    - you have run the tests locally and they have passed before submitting your PR.
-->


dlt/destinations/impl/databricks/sql_client.py
```diff
@@ -65,6 +66,18 @@ def rollback_transaction(self) -> None:
     def native_connection(self) -> "DatabricksSqlConnection":
         return self._conn
 
+    def has_dataset(self) -> bool:
+        query = """
+    SELECT CATALOG_NAME
```
steinitzu, line 71: What's the context on this? Would be great to have a comment describing why this override is necessary. Looks like an odd case that `SELECT 1` doesn't work.  
I also cannot reproduce the false-positive issue with the original `has_dataset` in our databricks. Is this something that happens now with new clusters only?

dlt/destinations/impl/databricks/sql_client.py
```diff
@@ -65,6 +66,18 @@ def rollback_transaction(self) -> None:
...
+    FROM INFORMATION_SCHEMA.SCHEMATA
+    WHERE """
+        db_params = self.fully_qualified_dataset_name(escape=False).split(".", 2)
+        if len(db_params) == 2:
+            query += " catalog_name = %s AND "
+        query += "schema_name = %s"
```
steinitzu, line 77: Since databricks always uses `catalog_name` I would skip this `if` and just hardcode the whole query, it's easier to read.

dlt/destinations/impl/databricks/sql_client.py
```diff
@@ -89,27 +102,21 @@ def execute_sql(
     def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DBApiCursor]:
         curr: DBApiCursor = None
         # TODO: databricks connector 3.0.0 will use :named paramstyle only
```
steinitzu, line 104: We can remove the "TODO:" from this comment now :)

pyproject.toml
```diff
@@ -81,8 +81,8 @@ weaviate-client = {version = ">=3.22", optional = true}
 adlfs = {version = ">=2022.4.0", optional = true}
 pyodbc = {version = "^4.0.39", optional = true}
 qdrant-client = {version = "^1.6.4", optional = true, extras = ["fastembed"]}
-databricks-sql-connector = {version = ">=2.9.3,<3.0.0", optional = true}
-dbt-databricks = {version = ">=1.7.3", optional = true}
+databricks-sql-connector = "3.1.2"
```
steinitzu, line 84: Any rason to pin this specific versions? Otherwise we should just set lower bound on the minimum version that works. E.g. `^3` or `^3.1`  
Also make sure to keep the `optional` flag

pyproject.toml
```diff
@@ -81,8 +81,8 @@ weaviate-client = {version = ">=3.22", optional = true}
...
+dbt-databricks = "1.8.1"
```
steinitzu, line 85: Same with the `dbt` dependency

steinitzu (changes requested):
Hi @oonyoontong , thanks for the contribution. I have a few comments/questions, so please take a look :)

dlt/destinations/impl/databricks/sql_client.py
oonyoontong: hey @steinitzu thanks for the comments! I've tested it again with the new dependencies, it works without changing has_dataset. the underlying `databricks_sql_connector` had some changes in the way they were parsing the results of the query, causing it to fail. This change was initially an attempt to get the lib to work properly (but only solved the symptom not the root cause)

>   File "databricks\sql\client.py", line 670, in fetchall
    return self.active_result_set.fetchall()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "databricks\sql\client.py", line 944, in fetchall
    return self._convert_arrow_table(self.fetchall_arrow())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "databricks\sql\client.py", line 884, in _convert_arrow_table
    res = df.to_numpy(na_value=None)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas\core\frame.py", line 1993, in to_numpy
    result = self._mgr.as_array(dtype=dtype, copy=copy, na_value=na_value)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas\core\internals\managers.py", line 1703, in as_array
    arr[isna(arr)] = na_value
    ~~~^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'

oonyoontong:
sorry for the messy PR @steinitzu, was planning clean it up today. The PR was done early to let the others facing this issue that it was being actively worked on

steinitzu (changes requested):
Ok! So this is due to an incompatibility with latest pandas https://github.com/databricks/databricks-sql-python/issues/326   
Latest connector pins pandas version, so updating is good imo and we should do it anyway.   

The only problem is that the named paramstyle fails with a confusing error on cluster runtime version `13.x` for some reason. I had to upgrade the test cluster to `14.3`.    
If there's no way around that I think we should try to detect which version is running at the beginning of `has_dataset`. I.e. https://docs.databricks.com/en/sql/language-manual/functions/current_version.html     
And raise a `DatabaseTerminalException` with a helpful message.  
What do you think?

oonyoontong:
> Ok! So this is due to an incompatibility with latest pandas [databricks/databricks-sql-python#326](https://github.com/databricks/databricks-sql-python/issues/326) Latest connector pins pandas version, so updating is good imo and we should do it anyway.
> 
> The only problem is that the named paramstyle fails with a confusing error on cluster runtime version `13.x` for some reason. I had to upgrade the test cluster to `14.3`. If there's no way around that I think we should try to detect which version is running at the beginning of `has_dataset`. I.e. https://docs.databricks.com/en/sql/language-manual/functions/current_version.html And raise a `DatabaseTerminalException` with a helpful message. What do you think?

from my investigation of  `current_version()` and digging through the docs of databricks-sql-python, i dont see any clear way of getting 
1. the cluster runtime version from databricks sql
2. the versions of the databricks runtime which this implementation breaks for, where we can clearly say `This is incompatible with databricks version x.x.x, please upgrade to >= y.y.y for this connector to work`

I am more inclined to catch the `name paramstyle error` and provide a error message like `Known error seen with databricks cluster <14.x, try again after upgrading the databricks cluster to >= 14.x` 

I dont have the ability to change my databricks cluster to 13.x to test named paramstyle error returned, if you think the above solution works, could you help me add a commit to except the `Error`?

PR #1441, opened 2024-06-04
docs for loading with contracts to existing tables

sh-rp:
### Description
Update the docs on how to work with schema contracts and existing tables and columns

### DIFF
--- a/docs/website/docs/general-usage/schema-contracts.md
+++ b/docs/website/docs/general-usage/schema-contracts.md
@@ -151,18 +151,24 @@ Following tables are considered new:
 
 For example such table is considered new because column **number** is incomplete (define primary key and NOT null but no data type)
 ```yaml
-  blocks:
-    description: Ethereum blocks
-    write_disposition: append
-    columns:
-      number:
-        nullable: false
-        primary_key: true
-        name: number
+blocks:
+  description: Ethereum blocks
+  write_disposition: append
+  columns:
+    number:
+      nullable: false
+      primary_key: true
+      name: number
 ```
 
 What tables are not considered new:
 1. Those with columns defined by Pydantic modes
+
+### Working with datasets that have manually added tables and columns on the first load
+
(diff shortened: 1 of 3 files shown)


docs/website/docs/general-usage/schema-contracts.md
```diff
@@ -151,19 +151,25 @@ Following tables are considered new:
...
 ```
 
 What tables are not considered new:
 1. Those with columns defined by Pydantic modes
 
+### Working with datasets that have manually added tables and columns on the first load
+
+In some cases you might be working with datasets that have tables or columns created outside of dlt. If you are loading to a table not created by `dlt` for the first time, `dlt` will not know about this table while enforcing schema contracts. This means that if you do a load where the `tables` are set to `evolve`, all will work as planned. If you have `tables` set to `freeze`, dlt will raise an exception because it thinks you are creating a new table (which you are from dlts perspective). You can allow `evolve` for one load and then switch back to `freeze`. 
+
+The same thing will happen if `dlt` knows your table, but you have manually added a column to your destination and you have `columns` set to `freeze`.
```
rudolfix, line 171: I had just one thought. We have `pipeline.sync_schema` method. What if instead of letting to evolve once we use it after adding a column? should work and will make stuff easier

rudolfix (changes requested):
This issue escalated a little bit and my intention would be to allow people to load data into existing tables. If you are up to this:
In https://dlthub.com/docs/general-usage/destination-tables

I'd add a chapter (top level)
"Loading data into existing tables"
- you can load data into existing tables
- if the tables are empty (no data) dlt may success with adding `dlt` columns to it
- since columns are not NULL some database engines (like duckdb) will not be able to do it
- if the table contains data you always need to add such columns yourself
(show how to do it)

sh-rp:
@rudolfix I'm not quite sure, are you suggestion to only update the docs or also to change some code now?

rudolfix:
@sh-rp just to update docs - just as I described in the comment above about https://dlthub.com/docs/general-usage/destination-tables

so again: it is a short guide how to load data into existing table. you describe what your test does

docs/website/docs/general-usage/schema-contracts.md
rudolfix: maybe that was misleading. this is an idea to check in the test you write. but if it looks confusing let's just close it

rudolfix (approved):
LGTM!

PR #1444, opened 2024-06-05
exp/logging-messages-style

cacrespo:
### Description
Hi all!

I noticed that the logging output levels are currently left-aligned with a width of 21 characters. Is there any specific reason for this?

Here's an example of the current formatting:
image

While it's not a major issue, I believe removing the white spaces could improve the visualization. 

In this PR, I've adjusted the formatting accordingly and included a new test to validate the output format, specifically for error and warning messages:
image

Another option could be centering the text and reducing the total width slightly...

Perhaps in the future, we could explore using colors to differentiate between different log levels.

### Related Issues

Nope

Thanks!

### DIFF
--- a/dlt/common/configuration/specs/run_configuration.py
+++ b/dlt/common/configuration/specs/run_configuration.py
@@ -17,7 +17,7 @@ class RunConfiguration(BaseConfiguration):
     dlthub_telemetry: bool = True  # enable or disable dlthub telemetry
     dlthub_telemetry_endpoint: Optional[str] = "https://telemetry.scalevector.ai"
     dlthub_telemetry_segment_write_key: Optional[str] = None
-    log_format: str = "{asctime}|[{levelname:<21}]|{process}|{thread}|{name}|{filename}|{funcName}:{lineno}|{message}"
+    log_format: str = "{asctime}|[{levelname}]|{process}|{thread}|{name}|{filename}|{funcName}:{lineno}|{message}"
     log_level: str = "WARNING"
     request_timeout: float = 60
     """Timeout for http requests"""
--- a/tests/common/configuration/test_configuration.py
+++ b/tests/common/configuration/test_configuration.py
@@ -549,7 +549,7 @@ def test_configuration_is_mutable_mapping(environment: Any, env_provider: Config
         "dlthub_telemetry": True,
         "dlthub_telemetry_endpoint": "https://telemetry-tracker.services4758.workers.dev",
         "dlthub_telemetry_segment_write_key": None,
-        "log_format": "{asctime}|[{levelname:<21}]|{process}|{thread}|{name}|{filename}|{funcName}:{lineno}|{message}",
+        "log_format": "{asctime}|[{levelname}]|{process}|{thread}|{name}|{filename}|{funcName}:{lineno}|{message}",
         "log_level": "WARNING",
         "request_timeout": 60,
         "request_max_attempts": 5,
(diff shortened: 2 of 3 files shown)


rudolfix:
@cacrespo heh good point :) I took this formatter from my older projects when we started writing dlt. the intention was to align all log lines where file name and line number where... but since that time we added more things and that does not make sense.

btw. log format is configurable via toml files and env variables (`RUNTIME__LOG_FORMAT`). so if you want colors you can add them (but no variable colors without changing code I'm afraid)

I'm ok with merging this. @burnash @sh-rp WDYT?

cacrespo:
Hi there!

I've just reviewed the tests and found something odd. **If you run test_logging.py before test/common/runners, everything works fine**. However, if you execute the **logging tests after the runners, we encounter problems** (as is happening now).

The conflicts are particularly with these tests:

```python
@pytest.fixture(scope="module", autouse=True)
def logger_autouse() -> None:
    init_test_logging()

@pytest.mark.parametrize("method", ALL_METHODS)
def test_pool_runner_process_methods_forced(method) -> None:
    multiprocessing.set_start_method(method, force=True)
    r = _TestRunnableWorker(4)
    # make sure signals and logging is initialized
    C = resolve_configuration(RunConfiguration())
    initialize_runtime(C)

    runs_count = runner.run_pool(configure(ProcessPoolConfiguration), r)
    assert runs_count == 1
    assert [v[0] for v in r.rv] == list(range(4))


@pytest.mark.parametrize("method", ALL_METHODS)
def test_pool_runner_process_methods_configured(method) -> None:
    r = _TestRunnableWorker(4)
    # make sure signals and logging is initialized
    C = resolve_configuration(RunConfiguration())
    initialize_runtime(C)

    runs_count = runner.run_pool(ProcessPoolConfiguration(start_method=method), r)
    assert runs_count == 1
    assert [v[0] for v in r.rv] == list(range(4))

```
My conclusion is that we have **some problem happening when the process instantiates the configuration** through the different tests.

Honestly, I don't know enough about pytest, but I made some changes (see the last commit), and now it seems that the problems are resolved (at least in my local environment).

Whatever you need, I'm here to help! 😃

rudolfix:
@cacrespo OK! let me check it. runner will initialize logging on their own and we do not have a proper way to "uninitialize" them. So some settings may not be applied. I'm running CI again and the see which tests are failing. thx!

sh-rp:
@cacrespo You can change the logging format via your config.toml if you like by the way.

cacrespo:
> @cacrespo You can change the logging format via your config.toml if you like by the way.

Yes, you're right. I think it's the best option instead of trying to change the application logger. 😜

For your information, I used [pytest-randomly](https://pypi.org/project/pytest-randomly/) and I believe there are some flaky tests (apart from my problem). Here are my steps and an example:

Add the package:
```
poetry add pytest-randomly
```

Run the tests without randomness:
```
poetry run pytest -p no:randomly tests/common/runners tests/common/runtime
```
Everything goes fine.

Run the tests with randomness:
```
poetry run pytest --randomly-seed=1330321047 tests/common/runners tests/common/runtime
```
We encounter some errors.

Hope this helps improve this great tool. Thanks!

sh-rp:
@cacrespo I have created a new branch (since this one is yours) with only the change you suggested: https://github.com/dlt-hub/dlt/pull/1517. I'm happy to attribute it to you, I don't think we need any tests for this at the moment.

cacrespo:
Closed by #1517

PR #1445, opened 2024-06-06
bump node version used in docs

sh-rp:
### Description
Netlify is complaining about us using a deprecated node js version (namely 16), so this PR bumps this to 20.


burnash (approved):
Looks good!

Issue #1317, opened 2024-05-03
pyarrow/pandas: add load id and dlt id in the extract phase and unify the behavior


rudolfix:
**Background**
By default we do not add `load_id` and `dlt_id` to arrow tables. This must be configured explicitly and happens in the normalizer.
As a consequence, we need to decompress and rewrite parquet files which takes a lot of resources.
In this ticket we move this behavior to the extract phase. This is against general architecture but I do not see any other way to do that without rewriting files. 
We also unify the behavior making `relational` normalizer to follow `ItemsNormalizerConfiguration`

**Tasks**
1. * [ ] move the logic that adds `_dlt_id` and `load_id` from `ArrowItemsNormalizer` to the extract phase. looks like the best place to add them is `_write_item` (just before normalization)
2. * [ ] we (probably) do not need the logic that adds the columns when writing a file. we can just add them to existing table
3. * [ ] when adding _dlt_id we must follow table settings and generate `_dlt_id` according to hints (ie. SCD2 look add_row_hash_to_table)
4. * [ ] `ItemsNormalizerConfiguration` must be taken into account. this is probably a breaking change because we need to move it from `normalize` to `extract` so old settings will stop working
5. * [ ] in `_compute_table` when we add new columns we should also infer hints like for any new columns. currently schema settings will be ignored

unify behavior
6. * [ ] in `relational.py` do not generate _load_id and _dlt_id if not switched on. (both are ON by default)

PR #1449, opened 2024-06-07
Add load_id to arrow tables in extract step instead of normalize

steinitzu:
### Description

- Add `_dlt_load_id` to arrow tables in extract step before writing file to disk.
- Remove the load_id adding logic from normalize

Should be backwards compatible, aside from the edge case where you upgrade dlt in between extract and normalize

Will add/check tests

### Related Issues

Step 1 of https://github.com/dlt-hub/dlt/issues/1317

### DIFF
--- a/dlt/extract/extractors.py
+++ b/dlt/extract/extractors.py
@@ -342,15 +342,23 @@ class ArrowExtractor(Extractor):
             # normalize arrow table before merging
             arrow_table = self.schema.normalize_table_identifiers(arrow_table)
             # Add load_id column
+            dlt_load_id_col = self.naming.normalize_table_identifier("_dlt_load_id")
             if (
                 self._normalize_config.add_dlt_load_id
-                and "_dlt_load_id" not in arrow_table["columns"]
+                and dlt_load_id_col not in arrow_table["columns"]
             ):
-                arrow_table["columns"]["_dlt_load_id"] = {
-                    "name": "_dlt_load_id",
-                    "data_type": "text",
-                    "nullable": False,
-                }
+                self.schema.update_table(
+                    {
+                        "name": arrow_table["name"],
+                        "columns": {
+                            dlt_load_id_col: {
+                                "name": dlt_load_id_col,
+                                "data_type": "text",
+                                "nullable": False,
+                            }
+                        },
+                    }
+                )
             # issue warnings when overriding computed with arrow
             override_warn: bool = False
             for col_name, column in arrow_table["columns"].items():
--- a/dlt/extract/extractors.py
+++ b/dlt/extract/extractors.py
@@ -1,9 +1,8 @@
 from copy import copy
 from typing import Set, Dict, Any, Optional, List
 
-from dlt.common.configuration import known_sections
+from dlt.common.configuration import known_sections, resolve_configuration, with_config
 from dlt.common import logger
-from dlt.common.configuration.inject import with_config
 from dlt.common.configuration.specs import BaseConfiguration, configspec
 from dlt.common.destination.capabilities import DestinationCapabilitiesContext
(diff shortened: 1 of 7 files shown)


dlt/extract/extractors.py
```diff
@@ -215,13 +217,26 @@ class ObjectExtractor(Extractor):
...
+    - `pyarrow.Table`
+    - `pyarrow.RecordBatch`
+    - `pandas.DataFrame` (is converted to arrow `Table` before processing)
     """
 
+    # Inject the parts of normalize configuration that are used here
+    @with_config(
+        spec=ItemsNormalizerConfiguration, sections=(known_sections.NORMALIZE, "parquet_normalizer")
+    )
+    def __init__(self, *args: Any, add_dlt_load_id: bool = False, **kwargs: Any) -> None:
```
steinitzu, line 236: This was my solution to support the old normalize config.  
I think it makes sense to keep it rather than move to `extract` section and have consistent with object normalizer.

rudolfix (changes requested):
LGTM!

I think we miss some tests. There's for sure test that checks if load_id is added. However now we need test two things:

1. adding load_id in extract + sensitivity to your config "hack" - just pipeline.extract() arrow table and see if extracted parquet has load_id
2. load some json data and request parquet in nromalize stage - this will create parquet file in normalizer and we test if load id is added there

dlt/extract/extractors.py
rudolfix: hmmm idea is very good.. but maybe you could decorate a method in this class. not __init__? so you call it just to retrieve config,

dlt/extract/extractors.py
```diff
@@ -318,6 +341,16 @@ def _compute_table(
             arrow_table["columns"] = pyarrow.py_arrow_to_table_schema_columns(item.schema)
             # normalize arrow table before merging
             arrow_table = self.schema.normalize_table_identifiers(arrow_table)
+            # Add load_id column
+            if (
+                self._normalize_config.add_dlt_load_id
+                and "_dlt_load_id" not in arrow_table["columns"]
```
rudolfix, line 347: 1. normalize _dlt_load_id before checking if in table columns
2. add column to table using schema.update_table so we are sure identifiers are properly normalized

rudolfix (changes requested):
@steinitzu thx for the tests!
still a few issues with normalizing column names. pls check

steinitzu:
Updated with normalized identifiers :)

rudolfix (changes requested):
@steinitzu looks good! but new test is not passing. IMO some column mismatch?

steinitzu:
> @steinitzu looks good! but new test is not passing. IMO some column mismatch?

@rudolfix this is fixed by https://github.com/dlt-hub/dlt/pull/1449/commits/b1be9c93f1d656f8da5adcb51ab62fd993076abe  
The issue was that `schema.update_table` was adding the column at the front. I'm not sure if it's right to use that method here since it creates the table in schema and that should be left for after all columns are computed?  
Added check for `_dlt_load_id` being last in common tests now too.

rudolfix:
> > @steinitzu looks good! but new test is not passing. IMO some column mismatch?
> 
> @rudolfix this is fixed by [b1be9c9](https://github.com/dlt-hub/dlt/commit/b1be9c93f1d656f8da5adcb51ab62fd993076abe) The issue was that `schema.update_table` was adding the column at the front. I'm not sure if it's right to use that method here since it creates the table in schema and that should be left for after all columns are computed? Added check for `_dlt_load_id` being last in common tests now too.

this is really weird. I added one more tests and the columns are added at the end. current version is btw. better, update_table is done at the end to modify actual schema

PR #1457, opened 2024-06-12
Loader parallelism strategies

sh-rp:
### Description
* Add parallelism strategies to loader
* Add max_load_jobs and parallelism strategies override to destination capabilities
* Add settings for both on the custom destination

### DIFF
--- a/dlt/common/destination/capabilities.py
+++ b/dlt/common/destination/capabilities.py
@@ -30,6 +30,8 @@ from dlt.common.wei import EVM_DECIMAL_PRECISION
 # insert_values - insert SQL statements
 # sql - any sql statement
 TLoaderFileFormat = Literal["jsonl", "typed-jsonl", "insert_values", "parquet", "csv"]
+TLoaderParallelismStrategy = Literal["parallel", "table_sequential", "sequential"]
+
 ALL_SUPPORTED_FILE_FORMATS: Set[TLoaderFileFormat] = set(get_args(TLoaderFileFormat))
 
 
@@ -81,12 +83,18 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     insert_values_writer_type: str = "default"
     supports_multiple_statements: bool = True
     supports_clone_table: bool = False
-
     """Destination supports CREATE TABLE ... CLONE ... statements"""
-    max_table_nesting: Optional[int] = None  # destination can overwrite max table nesting
+
+    max_table_nesting: Optional[int] = None
+    """Allows a destination to overwrite max_table_nesting from source"""
 
     # do not allow to create default value, destination caps must be always explicitly inserted into container
     can_create_default: ClassVar[bool] = False
+
+    max_parallel_load_jobs: Optional[int] = None
+    """The destination can set the maxium amount of parallel load jobs being executed"""
+    loader_parallelism_strategy: Optional[TLoaderParallelismStrategy] = None
+    """The destination can override the parallelism strategy"""
 
     @staticmethod
     def generic_capabilities(
--- a/dlt/common/destination/capabilities.py
+++ b/dlt/common/destination/capabilities.py
@@ -30,7 +30,7 @@ from dlt.common.wei import EVM_DECIMAL_PRECISION
 # insert_values - insert SQL statements
 # sql - any sql statement
 TLoaderFileFormat = Literal["jsonl", "typed-jsonl", "insert_values", "parquet", "csv"]
-TLoaderParallelismStrategy = Literal["parallel", "table_sequential", "sequential"]
+TLoaderParallelismStrategy = Literal["parallel", "table-sequential", "sequential"]
 
 ALL_SUPPORTED_FILE_FORMATS: Set[TLoaderFileFormat] = set(get_args(TLoaderFileFormat))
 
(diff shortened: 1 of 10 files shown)


dlt/load/utils.py
```diff
@@ -216,3 +218,45 @@ def _extend_tables_with_table_chain(
                 continue
             result.add(chain_table_name)
     return result
+
+
+def filter_new_jobs(
+    file_names: Sequence[str],
+    capabilities: DestinationCapabilitiesContext,
+    config: LoaderConfiguration,
+) -> Sequence[str]:
```
sh-rp, line 227: This will now use the capabilities of the main destination for all jobs, regardless of the settings on the staging destination and regardless of which jobs are for staging or main destination. I can add this differentiation too, but I'm not sure we really need it..

dlt/destinations/impl/destination/__init__.py
```diff
@@ -1,16 +1,21 @@
 from typing import Optional
 from dlt.common.destination import DestinationCapabilitiesContext, TLoaderFileFormat
+from dlt.common.destination.capabilities import TLoaderParallelismStrategy
 
 
 def capabilities(
     preferred_loader_file_format: TLoaderFileFormat = "typed-jsonl",
     naming_convention: str = "direct",
     max_table_nesting: Optional[int] = 0,
+    max_parallel_load_jobs: Optional[int] = 0,
```
sh-rp, line 10: I'm not sure we really need this here, this can be configured via config toml, same for the strategy. But maybe it is nice to have.

sh-rp:
@rudolfix which one is the destination that requires sequentials jobs?

dlt/common/destination/capabilities.py
```diff
@@ -30,6 +30,8 @@
 # insert_values - insert SQL statements
 # sql - any sql statement
 TLoaderFileFormat = Literal["jsonl", "typed-jsonl", "insert_values", "parquet", "csv"]
+TLoaderParallelismStrategy = Literal["parallel", "table_sequential", "sequential"]
```
rudolfix, line 33: hmm I think we are using dashes not underscores in other places? surely the above both are present :)

dlt/destinations/impl/destination/__init__.py
rudolfix: it is really helpful. use case: qdrant in local mode must be sequential. (single thread). in #998 capabilities are adjusted per config values (and naming convention) so I'll be able to request a strategy based on it.

dlt/load/configuration.py
```diff
@@ -20,4 +24,6 @@ class LoaderConfiguration(PoolRunnerConfiguration):
     truncate_staging_dataset: bool = False
 
     def on_resolved(self) -> None:
-        self.pool_type = "none" if self.workers == 1 else "thread"
+        self.pool_type = (
```
rudolfix, line 27: good. must be None (no threads spawned)

dlt/load/utils.py
rudolfix: we don't need this now. eventually the more strict settings could be used. We already have function that combines capabilities in Pipeline. we could extend it and use it here. but definitely not in scope of this

dlt/load/utils.py
```diff
@@ -216,3 +218,44 @@ def _extend_tables_with_table_chain(
...
+    """Filters the list of new jobs to adhere to max_workers and parallellism strategy"""
+
+    # nothing to do
+    if not file_names:
+        return file_names
+
+    # destination can overwrite ps
+    parallelism_strategy = capabilities.loader_parallelism_strategy or config.parallelism_strategy
```
rudolfix, line 235: config should always has last say. by default config should be None - no strategy should be forced

dlt/load/utils.py
```diff
@@ -216,3 +218,44 @@ def _extend_tables_with_table_chain(
...
+    if mp := capabilities.max_parallel_load_jobs:
+        max_workers = min(max_workers, mp)
+
+    # regular sequential will take max worker amount no matter the table
+    if parallelism_strategy == "parallel":
+        return file_names[:max_workers]
+
+    # we must ensure there only is one job per table
+    if parallelism_strategy == "table_sequential":
+        filtered_jobs: List[str] = []
```
rudolfix, line 252: this looks good. but you could also use `group_by` by table name, maybe the code will be simpler?

rudolfix (changes requested):
a few small issues

tests are amazing! good work!

dlt/load/configuration.py
sh-rp: I'm not sure what you mean? Should I change something here?

dlt/load/utils.py
sh-rp: fixed

dlt/load/utils.py
sh-rp: i made a one-liner out of it, not sure if it is easier to read now, but it is more pythonic i guess :)

sh-rp:
@rudolfix I have added an explanation of the new config values to the custom destination page. I briefly thought about adding the parallelization strategy setting to the performance docs page, but I am actually not sure it is useful there, for now I believe we only need this in the custom destination.

dlt/load/utils.py
sh-rp: Turns out you need to sort by the key you group by for the groupby to work as expected..

sh-rp:
Also I don't know what is going on with this failing duckdb test, this does not happen locally for me and I also don't see any way my changes could influence this..

rudolfix (approved):
LGTM! no idea where duckdb problem is coming from. I bet some previous test is changing the default database location that's why file is not there

PR #1466, opened 2024-06-14
Add `upsert` merge strategy

### LINKED ISSUE #1129: Add `upsert` merge strategy

jorritsandbrink:
### Description
- adds a new merge strategy called `upsert`
- requires `primary_key`, should be unique (see note 1)
  - stores primary-key hash as `_dlt_id`
  - does not support primary key-based deduplication like the `delete-insert` strategy
- does not support `merge_key`
- only tested for `postgres` and `snowflake` destinations (other destinations come in follow up PR)

**Note 1:**

A primary key-based `_dlt_id` requires that the primary key is unique before going into the `normalize` step. Primary key-based deduplication in the `load` step, as we do in the `delete-insert` merge strategy, is not possible.

Problems when primary key is not unique (i.e. it has duplicates):
- database error because `UNIQUE` constraint on `_dlt_id` column is violated (can be solved by not imposing the constraint)
- foreign keys in child staging tables link to multiple records in root staging table ➜ can't deduplicate child tables

### Related Issues

- Closes #1129


dlt/common/destination/reference.py
```diff
@@ -342,59 +343,72 @@ def _verify_schema(self) -> None:
...
+                        table, "merge_key"
+                    ):
+                        logger.warning(
+                            f"Table {table['name']} has `write_disposition` set to `merge`"
+                            " and `merge_strategy` set to `delete-insert`, but no primary or"
+                            " merge keys defined."
+                            " dlt will fall back to `append` for this table."
+                        )
+                elif table.get("x-merge-strategy") == "upsert":
+                    if self.config.destination_name not in ("postgres", "snowflake"):
```
jorritsandbrink, line 364: Should we add a `supported_merge_strategies` destination capability?

dlt/destinations/sql_jobs.py
```diff
@@ -515,6 +550,84 @@ def gen_merge_sql(
...
+            WHEN MATCHED
+                THEN UPDATE SET {update_str}
+            WHEN NOT MATCHED
+                THEN INSERT ({col_str.format(alias="")}) VALUES ({col_str.format(alias="s.")});
+        """)
+
+        # generate statements for child tables if they exist
+        child_tables = table_chain[1:]
+        if child_tables:
+            root_row_key = escape_id("_dlt_id")
```
jorritsandbrink, line 597: Should the row id be marked with a new hint (other than `unique`) to prevent hard-coding `_dlt_id`?

dlt/common/destination/reference.py
rudolfix: yes - if you at some point (as early as possible ie. when passing data to normalizer when we must have destination capabilities) are able to issue a warning and say which strategy will be used instead.

same thing with replace strategies (we have 3 afaik)

also if you want to do it: we do not even store which write dispositions are supported. so maybe this is a separate ticket :)

dlt/common/normalizers/json/relational.py
```diff
@@ -252,14 +259,15 @@ def _normalize_row(
...
-        row_hash: bool = False,
+        row_id_type: TRowIdType = None,
     ) -> TNormalizedRowIterator:
         schema = self.schema
         table = schema.naming.shorten_fragments(*parent_path, *ident_path)
-        # compute row hash and set as row id
-        if row_hash:
-            row_id = self.get_row_hash(dict_row)  # type: ignore[arg-type]
-            dict_row["_dlt_id"] = row_id
+        if row_id_type in ("key_hash", "row_hash"):
```
rudolfix, line 266: this logic must be moved to `_add_row_id`. I think the way we handle `_dlt_id` needs a little bit more work
- we have super simple bring your own `_dlt_id`
```
row_id = flattened_row.get("_dlt_id", None)
        if not row_id:
            row_id = self._add_row_id(table, flattened_row, parent_row_id, pos, _r_lvl)
```
we should replace it with hint based method - if there's any unique column we use it for `_dlt_id`. that may be a separate ticket. but current "bring your own" must work and now you ignore it here
- we have many methods to generate `_dlt_id` for parent table: random, deterministic (primary key based) and content based (used by scd2). they way you do it now is good enough. in the future I'd like people to be able to pick the way dlt id is generated (ie via `ItemsNormalizerConfiguration`)
- for child tables we always have deterministic `_dlt_id`

dlt/common/normalizers/json/relational.py
```diff
@@ -410,12 +420,18 @@ def _get_validity_column_names(schema: Schema, table_name: str) -> List[Optional
 
     @staticmethod
     @lru_cache(maxsize=None)
-    def _dlt_id_is_row_hash(schema: Schema, table_name: str) -> bool:
-        return (
-            schema.get_table(table_name)["columns"]  # type: ignore[return-value]
-            .get("_dlt_id", {})
-            .get("x-row-version", False)
-        )
+    def _get_row_id_type(schema: Schema, table_name: str) -> TRowIdType:
```
rudolfix, line 423: lru cache will hash schema by id (instance reference)? right?

dlt/common/normalizers/typing.py
```diff
@@ -1,8 +1,11 @@
-from typing import List, Optional, TypedDict
+from typing import List, Optional, TypedDict, Literal
 
 from dlt.common.typing import StrAny
 
 
+TRowIdType = Literal["random", "row_hash", "key_hash"]
```
rudolfix, line 6: yes this is good.

dlt/destinations/sql_jobs.py
rudolfix: why not `unique`? technically you just need a unique column to delete / merge child tables correctly. `delete-insert` does that, right?

rudolfix (changes requested):
this looks good, is short and clean. important info
- I must merge #998 - there all column names are abstracted away etc. your PR will generate a lot of conflict and it will be way easier to merge it after that
- for later: it would be cool if users can request dlt id type for append/replace as well - ie to just use content or primary key (deterministic) hash
- what happens if input data is not deduplicated? so we have several values with the same  primary key? `delete-insert` is deduplicating input data. see what happens in your case

dlt/common/normalizers/json/relational.py
jorritsandbrink: Right, because `Schema` is a user-defined class. 

_"Objects which are instances of user-defined classes are hashable by default. They all compare unequal (except with themselves), and their hash value is derived from their [id()](https://docs.python.org/3/library/functions.html#id)."_

https://docs.python.org/3/glossary.html#term-hashable

dlt/common/normalizers/json/relational.py
jorritsandbrink: I have moved the logic to `_add_row_id`. Current "bring your own" (if I understand it correctly) should work now.

dlt/common/destination/reference.py
jorritsandbrink: - added `supported_merge_strategies` capability and configured it for all SQL destinations
- added `_verify_destination_capabilities` that checks configured merge strategy against supported merge strategies
  - raises error if not supported
    - (I know you suggested falling back to a supported strategy and issueing a warning instead, but I'm not a fan of that approach. Will of course change it if that's how we do things in `dlt`, but wanted to challenge it first.)
  - called right before `normalize` step — is that the right place?

I can create a separate ticket for supported replace strategies and supported write dispositions capabilities.

jorritsandbrink:
@rudolfix I addressed your comments. Can you review?

> what happens if input data is not deduplicated? so we have several values with the same primary key? delete-insert is deduplicating input data. see what happens in your case

See "Note 1" in the PR description.

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -984,15 +1033,57 @@ def r():
...
+    p.drop()
+    r.apply_hints(
+        write_disposition={"disposition": "merge", "strategy": "upsert"},
+        primary_key="foo",
+        merge_key="foo",
+    )
+    assert "primary_key" in r._hints
+    assert "merge_key" in r._hints
+    p.run(r())
+    assert (
```
jorritsandbrink, line 1086: This assertion fails on CI because `capsys.readouterr().err` is an empty string there. It does work on my local machine. Any idea on how to best test warning logs?

dlt/destinations/utils.py
```diff
@@ -77,17 +77,32 @@ def verify_sql_job_client_schema(schema: Schema, warnings: bool = True) -> List[
...
+                if not has_column_with_prop(table, "primary_key"):
+                    exception_log.append(
+                        SchemaCorruptedException(
+                            schema.name,
+                            f"No primary key defined for table `{table['name']}`."
+                            " `primary_key` needs to be set when using the `upsert`"
+                            " merge strategy.",
+                        )
+                    )
+                if has_column_with_prop(table, "merge_key"):
```
jorritsandbrink, line 100: I noticed these warnings are logged more than once because `verify_sql_job_client_schema` is called multiple times in a single pipeline run. Do we have something to prevent that?

dlt/extract/hints.py
```diff
@@ -440,9 +441,14 @@ def _merge_merge_disposition_dict(dict_: Dict[str, Any]) -> None:
 
         mddict: TMergeDispositionDict = deepcopy(dict_["write_disposition"])
         if mddict is not None:
-            dict_["x-merge-strategy"] = (
-                mddict["strategy"] if "strategy" in mddict else DEFAULT_MERGE_STRATEGY
-            )
+            dict_["x-merge-strategy"] = DEFAULT_MERGE_STRATEGY
+            if "strategy" in mddict:
+                if mddict["strategy"] not in MERGE_STRATEGIES:
+                    raise ValueError(
```
jorritsandbrink, line 447: Is this the right way/place to do user input validation?

dlt/destinations/sql_jobs.py
rudolfix: no, we do not have such test case. what I want to achieve here is that we rely on annotations, not on column names.
- `_dlt_id` is annotated as `unique` by a standard schema
- `root_key` is a also an annotation, not a hardcoded field
`relational.py` adds config for each table with `merge` write disposition:
```
"propagation": {
                        "tables": {
                            table_name: {
                                TColumnName(self.c_dlt_id): TColumnName(self.c_dlt_root_id)
                            }
                        }
                    }
```
which will propagate `_dlt_id` from root to each child table as `_dlt_root_id` and
```
self.schema._merge_hints(
            {
                "not_null": [
                    TSimpleRegex(self.c_dlt_id),
                    TSimpleRegex(self.c_dlt_root_id),
                    TSimpleRegex(self.c_dlt_parent_id),
                    TSimpleRegex(self.c_dlt_list_idx),
                    TSimpleRegex(self.c_dlt_load_id),
                ],
                "foreign_key": [TSimpleRegex(self.c_dlt_parent_id)],
                "root_key": [TSimpleRegex(self.c_dlt_root_id)],
                "unique": [TSimpleRegex(self.c_dlt_id)],
            },
            normalize_identifiers=False,  # already normalized
        )
```
makes sure that hints are applied. this is quite old version of `dlt` and we'll reimplement it. but high level mechanism (we rely on annotations, not names) is good

tests/load/pipeline/test_merge_disposition.py
rudolfix: OK I have no idea. maybe the log is routed to stdout on CI? but I doubt that, or the test is really not passing

dlt/destinations/impl/bigquery/factory.py
```diff
@@ -42,6 +42,7 @@ def _raw_capabilities(self) -> DestinationCapabilitiesContext:
         caps.supports_ddl_transactions = False
         caps.supports_clone_table = True
         caps.schema_supports_numeric_precision = False  # no precision information in BigQuery
+        caps.supported_merge_strategies = ["delete-insert", "scd2"]
```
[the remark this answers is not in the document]
rudolfix: OK I think you intend to merge some kind of simple version and then add support one by one. Good!

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -30,13 +33,30 @@
 # ACTIVE_DESTINATIONS += ["motherduck"]
 
 
+def skip_if_not_supported(
+    merge_strategy: TLoaderMergeStrategy,
+    destination: TDestination,
+) -> None:
+    if merge_strategy not in destination.capabilities().supported_merge_strategies:
```
rudolfix, line 40: this is good. but should be a part of `destinations_configs` filters. definitely a task for the future: if we add supported write disposiitons to the capabilities we'll be able to generate a lot of filters automatically

rudolfix (changes requested):
This is really good and test coverage is sufficient. I had a few remarks regarding using hardcoded column names vs annotation but it looks mostly fixed now.

other things are minor

we also lack some docs but can go to the next PR. up to you

again - sorry for the big merge and conflicts - you'll have a few more places to fix when you merge current devel - but way less than last time

dlt/destinations/impl/bigquery/factory.py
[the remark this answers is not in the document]
jorritsandbrink: Correct, the plan was `snowflake` first. We get `postgres` "for free" because that's what I was using to develop the logic. Some of the other destinations might already work, but have not been tested. They will be added in a followup PR.

dlt/common/normalizers/json/relational.py
```diff
@@ -188,18 +195,37 @@ def _extend_row(extend: DictStrAny, row: DictStrAny) -> None:
...
+
+        if row_id_type == "random":
+            row_id = generate_dlt_id()
+        else:
+            if _r_lvl == 0:  # root table
+                if row_id_type in ("key_hash", "row_hash"):
+                    subset = None
+                    if row_id_type == "key_hash":
+                        subset = self._get_primary_key(self.schema, table)
+                    row_id = self.get_row_hash(dict_row, subset=subset)
```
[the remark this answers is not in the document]
jorritsandbrink: Not sure if it's what you had in mind but I added a comment in this commit: https://github.com/dlt-hub/dlt/pull/1466/commits/245cd0f3b67ffbe49af38b28fe9bc21ec0546226

dlt/common/normalizers/json/relational.py
```diff
@@ -463,12 +486,24 @@ def _get_validity_column_names(schema: Schema, table_name: str) -> List[Optional
...
+        schema: Schema, table_name: str, primary_key: bool, _r_lvl: int
+    ) -> TRowIdType:
+        if _r_lvl == 0:  # root table
+            merge_strategy = DataItemNormalizer._get_merge_strategy(schema, table_name)
+            if merge_strategy == "upsert":
+                return "key_hash"
+            elif merge_strategy == "scd2":
+                if (
+                    schema.get_table(table_name)["columns"]
+                    .get("_dlt_id", {})
```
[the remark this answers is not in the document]
jorritsandbrink: I adjusted in https://github.com/dlt-hub/dlt/pull/1466/commits/0d84e6dcda244c95525c2613ccc095a4e8f5a1aa

dlt/destinations/utils.py
jorritsandbrink: I made a commit to implement this but [reverted](https://github.com/dlt-hub/dlt/pull/1466/commits/cec667170817860bd4c1b050710314d471ba1d79) it because some tests were failing and we have a bit of urgency to merge this PR. Let's do it in another PR.

dlt/extract/hints.py
jorritsandbrink: Moved it to `_set_hints` in https://github.com/dlt-hub/dlt/pull/1466/commits/479e52adacd7b504123bf8b2b7bb74f802c63c68

tests/load/pipeline/test_merge_disposition.py
jorritsandbrink: Commented out the test for now so this doesn't block progress: https://github.com/dlt-hub/dlt/pull/1466/commits/0f641e44c2ae79c147947c44173635bbc443c7a6

rudolfix (approved):
LGTM!

PR #1469, opened 2024-06-16
Bump orjson WIP

Vasilije1990:
### Description
Bumped version of orjson in order to be able to use dlt in projects that also require langchain due to version conflicts

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ astunparse = ">=1.6.3"
 gitpython = ">=3.1.29"
 pytz = ">=2022.6"
 giturlparse = ">=0.10.0"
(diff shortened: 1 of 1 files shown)


rudolfix:
@Vasilije1990 I spent few hours last week researching if orjson is any good. they had tons of problems with crashes in the past that were fixing slowly. we can try the version that langchain allows. could you tell us the version spec they use?

rudolfix:
@steinitzu maybe you could take this over. we should add working dependencies for orjson, enabling a few ranges we know actually work
- keep our existing range
- check langchain range and add it

rudolfix:
#1501 implements this

PR #1472, opened 2024-06-16
fix: service principal auth support for synapse copy job

### LINKED ISSUE #1460: Assertion in SynapseCopyFileLoadJob funtion blocks use of service principal credentials for staging load
### dlt version

0.4.12

### Describe the problem

The problem arises when using service principal credentials to perform a file load in the SynapseCopyFileLoadJob function. Specifically, an assertion check is failing, which blocks the use case where different credentials are used for different parts of the loading process.

Key Points:
1. Assertion Failure:

The code contains an assertion assert isinstance(staging_credentials, AzureCredentialsWithoutDefault) at [line 207 in synapse.py](https://github.com/dlt-hub/dlt/blob/23c6365310d60ed86e16b25196ae4e41e71e8b3b/dlt/destinations/impl/synapse/synapse.py#L207)

This assertion checks if staging_credentials is an instance of `AzureCredentialsWithoutDefault`.

2. Service Principal Credentials:

The failure occurs because when using service principal credentials, staging_credentials is an instance of 
`AzureServicePrincipalCredentialsWithoutDefaults` instead of `AzureCredentialsWithoutDefault`.

3.  Use Case:

The intended scenario involves using service principal credentials (`AzureServicePrincipalCredentialsWithoutDefaults`) to 
perform the staging load to Azure Blob Storage.
Simultaneously, a managed identity is used to perform the COPY INTO operation.

4. Impact:

The assertion blocks this scenario by enforcing that staging_credentials must be of type `AzureCredentialsWithoutDefault`, which is not necessary.
As a result, the assertion is unnecessarily restrictive and prevents the intended use case.

5.  Proposed Fix:

Removing the assertion assert isinstance(staging_credentials, `AzureCredentialsWithoutDefault`) resolves the issue.
By removing this assertion, the code allows the use of `AzureServicePrincipalCredentialsWithoutDefaults` for staging load and a managed identity for the COPY INTO operation.

### Expected behavior

Allow AzureServicePrincipalCredentialsWithoutDefaults for staging load to Azure Blob Storage.
Permit the use of managed identity for COPY INTO operation.
Avoid unnecessary assertions restricting credential types.

### Steps to reproduce

Reported in community: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1716949523567929

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

destination synapse

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

jorritsandbrink:
### Description
Adds service principal authentication support for Synapse `COPY INTO` job.

### Related Issues

- Fixes #1460

### DIFF
--- a/dlt/destinations/impl/databricks/databricks.py
+++ b/dlt/destinations/impl/databricks/databricks.py
@@ -264,7 +264,7 @@ class DatabricksClient(InsertValuesJobClient, SupportsStagingDestination):
         sql_client = DatabricksSqlClient(config.normalize_dataset_name(schema), config.credentials)
         super().__init__(schema, config, sql_client)
         self.config: DatabricksClientConfiguration = config
-        self.sql_client: DatabricksSqlClient = sql_client  # type: ignore[assignment]
+        self.sql_client: DatabricksSqlClient = sql_client
         self.type_mapper = DatabricksTypeMapper(self.capabilities)
 
     def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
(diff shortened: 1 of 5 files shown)


rudolfix (approved):
LGTM!

PR #1480, opened 2024-06-18
fix(incremental): don't filter Arrow tables with empty filters

IlyaFaer:
Towards https://github.com/dlt-hub/verified-sources/pull/486


rudolfix (changes requested):
We need a test that reproduces this error. Looks like that happens when you send empty arrow as a second one and use incremental in your resource.

if this is correct please reproduce that

PR #1482, opened 2024-06-18
set default next item mode to round robin

sh-rp:
### Description
Sets the default next item mode in the pipe iterator to round robin.

### DIFF
--- a/tests/extract/test_decorators.py
+++ b/tests/extract/test_decorators.py
@@ -812,7 +812,10 @@ def standalone_transformer_returns(item: TDataItem, init: int = dlt.config.value
     return "A" * item * init
 
 
-def test_standalone_transformer() -> None:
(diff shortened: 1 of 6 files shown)


tests/extract/test_decorators.py
```diff
@@ -824,7 +824,7 @@ def test_standalone_transformer() -> None:
         bound_tx(1)
     assert isinstance(bound_tx, DltResource)
     # the resource sets the start of the range of transformer + transformer init
-    assert list(standalone_signature(1, 3) | bound_tx) == [6, 7, 8, 9, 7, 8, 9]
+    assert list(standalone_signature(1, 3) | bound_tx) == [6, 7, 7, 8, 8, 9, 9]
```
rudolfix, line 827: maybe also test both modes here? @sh-rp

PR #1483, opened 2024-06-18
fix: allow loggeradapter in addition to logger in logcollector

matsmhans1:
the _log method in LogCollector now checks for either logging.Logger or logging.LoggerAdapter unlike previously were it only allowed the former.

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...


rudolfix (approved):
LGTM!

PR #1485, opened 2024-06-18
[experiment] Add resource time limit and rate limiting

sh-rp:
### Description
This PR extends the add_limit function to add time limits and rate limits to resources. This approach is to be discussed,  but very straightforward, easy to test and works for both sync and async resources.

TODOs (if we go this route):
* More tests that test pipe iterators with multiple resources (we want to allow a global rate limit for APIs for example)
* Docs
* Extend source level add_limit to have the same functionality as the resource level one
* Improve the logger warning if add_limit is declared on non-incremental resources.
* Investigate fifo extractor strategy, maybe do not go to round robin if none is yielded...

Other thoughts:
* We might want to apply the rate limit wait also once before the original generator is used plus allow rate limiting on the transformers, otherwise global rate limiting for APIs will not work.

### DIFF
--- a/dlt/extract/resource.py
+++ b/dlt/extract/resource.py
@@ -1,4 +1,5 @@
 import inspect
+import time
 from functools import partial
 from typing import (
     AsyncIterable,
@@ -11,6 +12,7 @@ from typing import (
     Union,
     Any,
     Optional,
+    Literal,
 )
 from typing_extensions import TypeVar, Self
 
@@ -345,8 +347,14 @@ class DltResource(Iterable[TDataItem], DltResourceHints):
             self._pipe.insert_step(FilterItem(item_filter), insert_at)
         return self
 
-    def add_limit(self: TDltResourceImpl, max_items: int) -> TDltResourceImpl:  # noqa: A003
-        """Adds a limit `max_items` to the resource pipe.
+    def add_limit(
+        self: TDltResourceImpl,
+        max_items: Optional[int] = None,
+        max_time: Optional[float] = None,
+        min_wait: Optional[float] = None,
+    ) -> TDltResourceImpl:  # noqa: A003
+        """Adds a limit `max_items` to the resource pipe
+
 
         This mutates the encapsulated generator to stop after `max_items` items are yielded. This is useful for testing and debugging.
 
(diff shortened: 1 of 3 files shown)


dlt/extract/resource.py
```diff
@@ -372,14 +393,32 @@ def _gen_wrap(gen: TPipeStep) -> TPipeStep:
...
+                        # if we crossed the max time, we will stop
+                        if max_time and time.time() - start_time > max_time:
+                            return
+
+                    # apply rate limiting
+                    if min_wait:
+                        while (last_iteration + min_wait) - time.time() > 0:
+                            # we give control back to the pipe iterator
+                            yield None
+                            time.sleep(0.1)
```
sh-rp, line 416: we could make this configurable, I am not sure wether it is needed though.

tests/extract/test_sources.py
```diff
@@ -788,52 +788,6 @@ def test_add_transformer_right_pipe() -> None:
         iter([1, 2, 3]) | dlt.resource(lambda i: i * 3, name="lambda")
 
 
-def test_limit_infinite_counter() -> None:
```
sh-rp, line 791: these three tests were moved to the new location where a couple more tests will be added specifically for the limits

dlt/extract/resource.py
```diff
@@ -356,10 +365,22 @@ def add_limit(self: TDltResourceImpl, max_items: int) -> TDltResourceImpl:  # no
         def _gen_wrap(gen: TPipeStep) -> TPipeStep:
             """Wrap a generator to take the first `max_items` records"""
 
+            if max_items >= 0 or max_time and not self.incremental:
+                from dlt.common import logger
+
+                logger.warning(
```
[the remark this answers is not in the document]
sh-rp: also: generally speaking it would be cool to have some wrapper around log messages to be able to test them, maybe mocking would be enough, not sure

sh-rp:
Closing this PR in favor of:
* https://github.com/dlt-hub/dlt/pull/2149
* https://github.com/dlt-hub/dlt/pull/2131

@joscha the api rate limiting things you have suggested would be a layer above in the rest_api implementation. There might already be a ticket for this or you could open a new one.

joscha:
> would be a layer above in the rest_api implementation

Okay. How so, if it relies on a 429 answer or an explicit rest API resource returning the limits? Would you assume each resource to have some sort of hook to report back any time limits?

PR #1491, opened 2024-06-19
[experiment] ibis integration with connection sharing

sh-rp:
### Description
This PR has a connection borrowing approach which theoretically would make the integration quite easy with the wrapping contextmanager suggested here. Ibis does not have support for sharing an open dbapi connection so there is some hackery here and this does not work for most destinations unfortunately. Also about half of our sql destinations are not really supported by ibis.

The current test in here only passes for duckdb, postgres and snowflake. There are various problems on other destinations, that may or may not stem from some init code in the ibis backend not being run which would ordinarily run when executing the do_connect method.

Some notes:
* Redshift does work with the postgres backend

### DIFF
--- a/dlt/helpers/ibis_helper.py
+++ b/dlt/helpers/ibis_helper.py
@@ -23,5 +23,4 @@ def ibis_helper(p: Pipeline) -> Iterator[BaseBackend]:
     ibis_backend = cast(BaseBackend, ibis_module.Backend())
 
     with p.sql_client() as c:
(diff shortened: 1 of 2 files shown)


sh-rp:
For reference, issue for connection sharing in ibis: https://github.com/ibis-project/ibis/issues/8877

sh-rp:
https://github.com/ibis-project/ibis/pull/9603

PR #1494, opened 2024-06-19
Add support for continuously starting load jobs as slots free up in the loader

sh-rp:
### Description
In the current implementation we more or less always start n (=max workers) load jobs, let them complete and then rerun the whole loader to schedule the next n jobs. In this PR we submit new load jobs as slots free up.

What is happening here:
* The loader now periodically checks all jobs wether they are done and schedules new ones as needed.
* Runnable Jobs now manage their own internal state (file moving still needs to be done by the loader on the main thread) and have a dedicated "run" method which is the one called on a thread.
* Renaming of the Job base classes to make clearer what is going on:
  * RunnableLoadJob: Jobs that actually do something and should be executed on a thread
  * FollowupJob: Class that creates a new job persisted to the disc
  * HasFollowupJobs (ex-FollowupJob): Trait that tells loader to look for followup Jobs
  * FinalizedJob: Not runnable because it already has an actionable state (completed, failed, retry). Used for indicating failed restored jobs, completed restored jobs and cases where nothing needs to be done
* FollowupJobs always go to the "new_jobs" folder and need to be picked up by the loader as any other jobs do, there were some just directly executed in the main thread: not good! :) We now assume that creation of FollowupJobs does not yield exceptions, I don't think this is needed and it makes the code simpler.
 * (Hopefully) simplification of the load class
 * Restoring jobs is now handled the same way as creating jobs is, this simplifies the code in a few places. Jobs that are found in the running folder now are simply started again, the bigquery job can figure out wether it needs to be resumed on its own.

Possible FollowupWork:
* Performance Tuning in loader, a lot of time is spent on parsing filenames (I think).


dlt/load/load.py
```diff
@@ -371,6 +397,18 @@ def complete_package(self, load_id: str, schema: Schema, aborted: bool = False)
             f"All jobs completed, archiving package {load_id} with aborted set to {aborted}"
         )
 
+    def update_loadpackage_info(self, load_id: str) -> None:
```
sh-rp, line 400: I am not quite sure when this needs to be run and how often, so any help there would be apreciated

dlt/load/load.py
```diff
@@ -414,72 +452,49 @@ def load_single_package(self, load_id: str, schema: Schema) -> None:
...
                 raise
 
+        # always update load package info
+        self.update_loadpackage_info(load_id)
+
+        # complete the package if no new or started jobs present after loop exit
+        if (
+            len(self.load_storage.list_new_jobs(load_id)) == 0
+            and len(self.load_storage.normalized_packages.list_started_jobs(load_id)) == 0
+        ):
```
sh-rp, line 495: is this the correct "package completion" condition? I think so, but am not 100% sure.

tests/load/test_dummy_client.py
```diff
@@ -96,15 +96,15 @@ def test_unsupported_write_disposition() -> None:
     load.load_storage.normalized_packages.save_schema(load_id, schema)
```
sh-rp, line 96: I needed to change a bunch of tests, since we do not rely on multiple executions of the run method anymore. All the changes make sense, it might be good to add a few more tests cases specific to the new implementation.

dlt/load/load.py
```diff
@@ -414,72 +452,49 @@ def load_single_package(self, load_id: str, schema: Schema) -> None:
...
-                                raise LoadClientJobRetry(
-                                    load_id,
-                                    new_job.job_file_info.job_id(),
-                                    r_c,
-                                    self.config.raise_on_max_retries,
-                                )
+                # we continously spool new jobs and complete finished ones
+                running_jobs, pending_exception = self.complete_jobs(load_id, running_jobs, schema)
+                # do not spool new jobs if there was a signal
+                if not signals.signal_received() and not pending_exception:
```
sh-rp, line 470: signal support still here :)

dlt/load/load.py
```diff
@@ -274,14 +278,19 @@ def create_followup_jobs(
...
         """Run periodically in the main thread to collect job execution statuses.
 
         After detecting change of status, it commits the job state by moving it to the right folder
         May create one or more followup jobs that get scheduled as new jobs. New jobs are created
         only in terminal states (completed / failed)
         """
+        # list of jobs still running
         remaining_jobs: List[LoadJob] = []
+        # if an exception condition was met, return it to the main runner
+        pending_exception: Exception = None
```
sh-rp, line 293: I need to collect the exception to raise in the main loop  here now, we could alternatively collect all problems we find and print them out, not just raise on exception.

sh-rp:
Note: PR on hold until naming conventions PR is merged.

tests/pipeline/test_pipeline_trace.py
```diff
@@ -368,7 +368,8 @@ def test_trace_telemetry() -> None:
                 # dummy has empty fingerprint
                 assert event["properties"]["destination_fingerprint"] == ""
         # we have two failed files (state and data) that should be logged by sentry
-        assert len(SENTRY_SENT_ITEMS) == 2
+        # TODO: make this work
```
sh-rp, line 371: I am not sure why this is not working anymore, can you point me to the place in the code where these sentry events should be raised?

tests/pipeline/test_pipeline.py
```diff
@@ -1657,9 +1658,16 @@ def test_remove_pending_packages() -> None:
     os.environ["EXCEPTION_PROB"] = "1.0"
     os.environ["FAIL_IN_INIT"] = "False"
     os.environ["TIMEOUT"] = "1.0"
-    # should produce partial loads
+    # will make job go into retry state
     with pytest.raises(PipelineStepFailed):
         pipeline.run(airtable_emojis())
+    # move job into running folder manually
```
sh-rp, line 1664: in the new loader implementation, jobs that retry are moved into the new jobs folder again, I'm wondering wether "has_pending_data" should also return true if there are jobs that are marked as "to be retried"

tests/load/test_job_client.py
```diff
@@ -719,15 +719,12 @@ def test_retrieve_job(client: SqlJobClientBase, file_storage: FileStorage) -> No
...
+        # dataset = f.getvalue().decode()
+    # job = expect_load_file(client, file_storage, dataset, user_table_name)
     # now try to retrieve the job
     # TODO: we should re-create client instance as this call is intended to be run after some disruption ie. stopped loader process
-    r_job = client.restore_file_load(file_storage.make_full_path(job.file_name()))
-    assert r_job.state() == "completed"
-    # use just file name to restore
-    r_job = client.restore_file_load(job.file_name())
-    assert r_job.state() == "completed"
+    # r_job = client.restore_file_load(file_storage.make_full_path(job.file_name()))
```
sh-rp, line 726: NOTE: todo

tests/cli/test_pipeline_command.py
```diff
@@ -207,9 +207,19 @@ def test_pipeline_command_drop_partial_loads(repo_dir: str, project_files: FileS
     os.environ["EXCEPTION_PROB"] = "1.0"
     os.environ["FAIL_IN_INIT"] = "False"
     os.environ["TIMEOUT"] = "1.0"
+
     venv = Venv.restore_current()
     with pytest.raises(CalledProcessError) as cpe:
         print(venv.run_script("chess_pipeline.py"))
+
+    # move job into running folder manually
```
sh-rp, line 215: see comment about this below

tests/load/test_job_client.py
sh-rp: for bigquery restoring is tested now, the other just re-run the job when found in the running folder (which as discussed is the right way to do it)

dlt/load/utils.py
```diff
@@ -251,19 +249,30 @@ def filter_new_jobs(
...
+    if len(running_jobs) >= max_workers:
+        return []
+    max_jobs = max_workers - len(running_jobs)
+
     # regular sequential works on all jobs
     eligible_jobs = file_names
 
     # we must ensure there only is one job per table
     if parallelism_strategy == "table-sequential":
+        # TODO: this whole code block may be quite inefficient for long lists of jobs
```
sh-rp, line 262: Possible todo

dlt/load/load.py
```diff
@@ -381,6 +436,18 @@ def complete_package(self, load_id: str, schema: Schema, aborted: bool = False)
             f"All jobs completed, archiving package {load_id} with aborted set to {aborted}"
         )
 
+    def update_loadpackage_info(self, load_id: str) -> None:
+        # update counter we only care about the jobs that are scheduled to be loaded
+        package_jobs = self.load_storage.normalized_packages.get_load_package_jobs(load_id)
```
rudolfix, line 441: this is really costly operation... make sure we update counters only when job changed state

dlt/load/load.py
```diff
@@ -428,74 +495,51 @@ def load_single_package(self, load_id: str, schema: Schema) -> None:
...
+                    # if a pending exception was discovered during completion of jobs
+                    # we can raise it now
+                    if pending_exception:
+                        raise pending_exception
                     break
-                # process remaining jobs again
-                jobs = remaining_jobs
                 # this will raise on signal
-                sleep(1)
+                sleep(
```
rudolfix, line 524: I think you should really create a mock package with 50k jobs. this currently will call list new jobs 10x / second

dlt/load/load.py
rudolfix: this was the previous condition:
```
        if file_count == 0:
            logger.info(f"No new jobs found in {load_id}")
            return 0, []
```
so checking just new jobs. I do not fully get why you do it here again? the loop above should exit only when all jobs are completed, no?

dlt/load/utils.py
```diff
@@ -251,19 +249,30 @@ def filter_new_jobs(
     if mp := capabilities.max_parallel_load_jobs:
         max_workers = min(max_workers, mp)
 
+    # if all slots are full, do not create new jobs
+    if len(running_jobs) >= max_workers:
```
rudolfix, line 253: this must be checked before you call this function. do not pass `running_jobs` here. you are mixing two separate things

rudolfix (changes requested):
amazing job! and I think code got greatly simplified. also I think changes are quite comprehensible.

I reviewed load.py and I think you can take a look already

dlt/load/load.py
```diff
@@ -129,118 +131,147 @@ def maybe_with_staging_dataset(
...
+            else:
+                use_staging_dataset = isinstance(
+                    job_client, WithStagingDataset
+                ) and job_client.should_load_data_to_staging_dataset(load_table)
+
+            # set job vars
+            job.set_run_vars(load_id=load_id, schema=schema, load_table=load_table)
+
+            # submit to pool
+            self.pool.submit(Load.w_run_job, *(id(self), job, active_job_client, use_staging_dataset))  # type: ignore
```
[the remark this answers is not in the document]
sh-rp: Yep, done!

dlt/destinations/sql_jobs.py
```diff
@@ -54,22 +52,16 @@ def from_table_chain(
...
-            # to support clients that need to execute one statement at a time (i.e. snowflake)
-            sql = [
-                " ".join(stmt.splitlines())
-                for stmt in cls.generate_sql(table_chain, sql_client, params)
-            ]
-            job = cls(file_info.file_name(), "running")
-            job._save_text_file("\n".join(sql))
-        except Exception:
-            # return failed job
-            tables_str = yaml.dump(
```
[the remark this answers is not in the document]
sh-rp: In my view, failed jobs should be the result of something not right in the destination (=async errors). Creating a FollowUpJob should never fail, and if it does it is a programming error on our side and should create a proper exception with stacktrace and not a failed job. Followupjobs, at least for now, have no destination dependent or async component. The one exception is a full disk that might occur. If you want me to keep it though, I will put it back in.

dlt/load/load.py
```diff
@@ -269,60 +304,61 @@ def create_followup_jobs(
...
             state: TLoadJobState = job.state()
-            if state == "running":
+            if state in ("ready", "running"):
                 # ask again
                 logger.debug(f"job {job.job_id()} still running")
                 remaining_jobs.append(job)
             elif state == "failed":
                 # create followup jobs
-                _schedule_followup_jobs(self.create_followup_jobs(load_id, state, job, schema))
+                self.create_followup_jobs(load_id, state, job, schema)
```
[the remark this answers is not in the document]
sh-rp: I would say the old variant was not easier to test, the importing of a job was a nested function within the code that creates them. How would you split this? The part that imports the jobs only calls a function on the normalized_packages of the load_storage. If you have a good idea I'll change it.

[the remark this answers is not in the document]
sh-rp: That said, I will check the existing tests for the followupjobs and maybe this will result in a refactor here, let's see.

dlt/load/load.py
sh-rp: this should be done correctly now, I am also updating it once after the loop exits, I hope this is good

dlt/load/utils.py
sh-rp: done

dlt/load/load.py
```diff
@@ -129,118 +131,147 @@ def maybe_with_staging_dataset(
...
+
+    def start_new_jobs(
+        self, load_id: str, schema: Schema, running_jobs: Sequence[LoadJob]
+    ) -> Sequence[LoadJob]:
+        """
+        will retrieve jobs from the new_jobs folder and start as many as there are slots available
+        """
+        # get a list of jobs elligble to be started
         load_files = filter_new_jobs(
             self.load_storage.list_new_jobs(load_id),
```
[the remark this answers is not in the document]
sh-rp: I have changed it this way. We could cache the list of files, let me see what a test with 30k files does.

tests/pipeline/test_pipeline.py
sh-rp: @rudolfix I need your opinion here.

tests/cli/test_pipeline_command.py
sh-rp: above actually

dlt/common/destination/reference.py
```diff
@@ -272,75 +309,96 @@ class LoadJob:
...
         File name is also a job id (or job id is deterministically derived) so it must be globally unique
         """
         # ensure file name
-        assert file_name == FileStorage.get_file_name_from_file_path(file_name)
-        self._file_name = file_name
-        self._parsed_file_name = ParsedLoadJobFileName.parse(file_name)
+        super().__init__(file_path)
+        self._state: TLoadJobState = "ready"
+        self._exception: Exception = None
+        self._job_client = job_client
```
sh-rp, line 320: client here can possibly be removed.

tests/load/test_dummy_client.py
```diff
@@ -107,15 +107,24 @@ def test_unsupported_write_disposition() -> None:
...
+    start_time = time()
     with ThreadPoolExecutor(max_workers=20) as pool:
         load.run(pool)
-    print("done" + str(time.time() - start_time))
+    duration = float(time() - start_time)
+
+    # sanity check
+    assert duration > 5
+
+    # we want 1000 empty processed jobs to need less than 15 seconds total (locally it runs in 10)
```
sh-rp, line 124: I did the same test on devel and it is way slower there, I'd need to dig further into it why that is the case, maybe the long sleep and the restarting of the load, or maybe something else idk.

dlt/common/destination/reference.py
```diff
@@ -393,23 +451,20 @@ def update_stored_schema(
         return expected_update
 
     @abstractmethod
-    def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) -> LoadJob:
+    def get_load_job(
```
rudolfix, line 454: this is very misleading name. it does not get any jobs. it creates new jobs. please rename

dlt/common/destination/reference.py
```diff
@@ -393,23 +451,20 @@ def update_stored_schema(
...
+        self, table: TTableSchema, file_path: str, load_id: str, restore: bool = False
+    ) -> LoadJob:
         """Creates and starts a load job for a particular `table` with content in `file_path`"""
```
rudolfix, line 457: it does not start anything

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -254,30 +247,24 @@ def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) ->
...
                 else:
                     job = BigQueryLoadJob(
-                        FileStorage.get_file_name_from_file_path(file_path),
-                        self._create_load_job(table, file_path),
+                        file_path,
                         self.config.http_timeout,
                         self.config.retry_deadline,
                     )
+            # TODO: this section may not be needed, BigQueryLoadJob will not through errors here and the streaming insert i don't know
             except api_core_exceptions.GoogleAPICallError as gace:
```
rudolfix, line 263: this is 100% not needed and no calls to big query should be made from here. AFAIK you create jobs before sending them to threads and without calling `with` client so the connection in `sql_client` is closed

tests/load/bigquery/test_bigquery_client.py
```diff
@@ -300,14 +277,23 @@ def test_bigquery_job_errors(client: BigQueryClient, file_storage: FileStorage)
...
+    r_job = cast(
+        RunnableLoadJob,
+        client.get_load_job(
+            client.schema.get_table(user_table_name),
+            file_storage.make_full_path(job.file_name()),
+            uniq_id(),
+        ),
     )
+
+    # job will be automatically found and resumed
```
rudolfix, line 292: this test is not testing any resumes because `get_load_job` does not create anything in bigquery. you'll need to create a job that in big query, make sure it exists and then complete it. how to do that exactly IDK.

it would be cool to test it inside `load.py`. so you load a job that fails once you create another load instance run it again and make sure it is completed with retry count == 1 in file name

rudolfix (changes requested):
OK there are a lot of comments. we'll need another review... overall looking good!

are you able to add a test (ie for INSERT VALUES client) that simulates killing loader when inserts are done and then resuming? can run on duckdb

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -104,60 +106,82 @@ def from_db_type(
...
+                # google.api_core.exceptions.BadRequest - will not be processed ie bad job name
+                raise LoadJobTerminalException(
+                    self._file_path, f"The server reason was: {reason}"
+                ) from gace
+            else:
+                raise DatabaseTransientException(gace) from gace
+
+        # we loop on the job thread until we detect a status change
+        while True:
+            time.sleep(1)
```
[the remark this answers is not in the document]
sh-rp: done

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -104,60 +106,82 @@ def from_db_type(
...
+                continue
+            # done, break loop and go to completed state
+            if self._bq_load_job.output_rows is not None and self._bq_load_job.error_result is None:
+                break
+            reason = self._bq_load_job.error_result.get("reason")
+            if reason in BQ_TERMINAL_REASONS:
+                # the job permanently failed for the reason above
+                raise DatabaseTerminalException(Exception("Bigquery Load Job failed"))
+            elif reason in ["internalError"]:
+                continue
```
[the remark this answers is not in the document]
sh-rp: I have added the warning and also the reason to the exception string. Imho it was right except for the missing log message.

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -104,60 +106,82 @@ def from_db_type(
...
+            self._created_job = True
+        except api_core_exceptions.GoogleAPICallError as gace:
+            reason = BigQuerySqlClient._get_reason_from_errors(gace)
+            if reason == "notFound":
+                # google.api_core.exceptions.NotFound: 404 – table not found
+                raise DatabaseUndefinedRelation(gace) from gace
+            elif (
+                reason == "duplicate"
+            ):  # google.api_core.exceptions.Conflict: 409 PUT – already exists
+                self._bq_load_job = self._job_client._retrieve_load_job(self._file_path)
```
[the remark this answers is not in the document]
sh-rp: done

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -104,60 +106,82 @@ def from_db_type(
...
+                raise DatabaseTerminalException(Exception("Bigquery Load Job failed"))
+            elif reason in ["internalError"]:
+                continue
+            else:
+                raise DatabaseTransientException(Exception("Bigquery Job needs to be retried"))
 
     def exception(self) -> str:
-        exception: str = json.dumps(
+        if not self._bq_load_job:
+            return ""
```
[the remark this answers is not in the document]
sh-rp: I have removed it, i don't remember why this is here

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -77,79 +73,53 @@ def __init__(
...
 
-        file_paths = [job.file_path for job in self.table_jobs]
-
+        files = ReferenceFollowupJob.resolve_references(self._file_path)
         write_delta_table(
-            path=self.client.make_remote_uri(self.make_remote_path()),
-            data=pa.dataset.dataset(file_paths),
-            write_disposition=self.table["write_disposition"],
-            storage_options=_deltalake_storage_options(self.client.config),
+            path=self._job_client.make_remote_uri(
```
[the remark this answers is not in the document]
sh-rp: The behavior of the delta load has not changed here, i think delta manages the folders for the tables.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -326,22 +296,25 @@ def list_files_with_prefixes(self, table_dir: str, prefixes: List[str]) -> List[
...
+            return FinalizedLoadJob(file_path)
         if table.get("table_format") == "delta":
             import dlt.common.libs.deltalake  # assert dependencies are installed
 
-            return DoNothingFollowupJob(file_path)
-
-        cls = FollowupFilesystemJob if self.config.as_staging else LoadFilesystemJob
-        return cls(self, file_path, load_id, table)
+            # a reference job for a delta table indicates a table chain followup job
+            if ReferenceFollowupJob.is_reference_job(file_path, "delta"):
```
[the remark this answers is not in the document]
sh-rp: I also don't like this. The problem is the behavior of "is_staging_destination_job" in load.py. If I add "reference" to the supported file formats for the filesystem, then reference files will also be sent to the filesystem when it is configured as staging destination in combination with another destination, although it should actually be consumed by the main destination.

dlt/load/load.py
```diff
@@ -125,118 +128,164 @@ def maybe_with_staging_dataset(
...
+                ) and job_client.should_load_data_to_staging_dataset_on_staging_destination(
+                    load_table
+                )
+            else:
+                use_staging_dataset = isinstance(
+                    job_client, WithStagingDataset
+                ) and job_client.should_load_data_to_staging_dataset(load_table)
+
+            # set job vars
+            job.set_run_vars(load_id=load_id, schema=schema, load_table=load_table)
```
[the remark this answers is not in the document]
sh-rp: We can put it in the init also. There will just be tons of repetitive code where each destination has to provide these vars to every load job and every load job has to pass these to "super", so I chose this way. Your call.

dlt/common/destination/reference.py
```diff
@@ -257,11 +260,45 @@ class DestinationClientDwhWithStagingConfiguration(DestinationClientDwhConfigura
...
+    def __init__(self, file_path: str) -> None:
+        self._file_path = file_path
+        self._file_name = FileStorage.get_file_name_from_file_path(file_path)
+        # NOTE: we only accept a full filepath in the constructor
+        assert self._file_name != self._file_path
+        self._parsed_file_name = ParsedLoadJobFileName.parse(self._file_name)
+
+    def job_id(self) -> str:
+        """The job id that is derived from the file name and does not changes during job lifecycle"""
+        return self._parsed_file_name.job_id()
```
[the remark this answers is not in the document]
sh-rp: yes, as you saw below I think this should be good.

dlt/load/load.py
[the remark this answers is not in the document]
sh-rp: I looked at the query tagging stuff, I am not really sure how that is related to these run_vars? Do you want the query tags to be set this way on the job too? Or what do you have in mind?

dlt/load/load.py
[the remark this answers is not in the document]
sh-rp: In my view they are all there. I will add one new test to see what happens if creating a followupjob throws an exception and wether the package will load once the problem is resolved.

dlt/load/load.py
```diff
@@ -424,74 +513,46 @@ def load_single_package(self, load_id: str, schema: Schema) -> None:
...
+                    load_id, running_jobs, schema
+                )
+                # update load package info if any jobs where finalized
+                if finalized_jobs:
+                    self.update_loadpackage_info(load_id)
+
+                pending_exception = pending_exception or new_pending_exception
+                # do not spool new jobs if there was a signal
+                if not signals.signal_received() and not pending_exception:
+                    running_jobs += self.start_new_jobs(load_id, schema, running_jobs)
```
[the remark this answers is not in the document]
sh-rp: done

tests/cli/test_pipeline_command.py
sh-rp: the comment right above. `def is_package_partially_loaded` returns False if all jobs in a normalized package are in the started folder. So if I have a test that needs `is_package_partially_loaded` to return true, I have to move one job into the running or completed folder. Before my changes, the job that needed retrying remained in the started_jobs folder so this was not needed. There are two places where I test this. My question above is wether we should change is_package_partially_loaded to return true also if there was at least one failed retry.

tests/load/bigquery/test_bigquery_client.py
sh-rp: I call "r_job.run_managed(client)" here which actually connects to bigquery and resumes the job (as indicated by the _resumed_job boolean var). The test is as good as the tests that where there for bigquery before. I am happy to improve them, but I don't really want to bloat the scope of these changes with things are are not directly related, at least in my view.

dlt/destinations/impl/bigquery/bigquery.py
sh-rp: yes that is correct, now that you mention it I have removed it. I think the test coverage for the streaming insert is not super good. we can improve that in the future maybe too, but generally speaking it is protected by the custom destination state mechanism, so I think it should be good enough for now.

tests/pipeline/test_pipeline_trace.py
rudolfix: what this test was doing: failing two jobs. those generated two log messages on error level. those messages were sent to sentry.

you changed the code so we do not log anything when messages are failed or dummy was changed in a way to not fail the packages when `FAIL_PROB` is set.

this test should work again

rudolfix (changes requested):
> What exactly is the desired behavior from a conceptual standpoint when creating a followup job fails? In the old version, loading just continues and those jobs are marked as failed. I don't think this makes sense, because the load will be useless if for example a mergejob can't be created and executed. So either we decide
this is a transient problem (e.g. the user can manually fix the schema and restart he load) , in that case we just raise an exception on the main thread, so that the job that triggered the scheduling of a followupjob remains in "started_jobs" and will be rerun on the pipeline execution including the scheduling of the followupjob (it is implemented and tested like this now)
or this is a terminal problem, in which case we should also stop the load but mark the loadpackage as failed

my take:
make it a transient error

and we need how we deal with failed packages. right now we continue load and do not raise exception at the end. IMO we should change that. we should continue load but raise exception at the end automatically.

I'll write a ticket for that - it ia a breaking change

tests/load/test_dummy_client.py
sh-rp: I can go back to 1 second, no problem (and maybe modify it for the speed test). If you look at "start_new_jobs", I made a change in the last round so that it checks how many slots are available before actually querying any files on the disc, so if all slots are full, it exits early and nothing is done. I'd say you can tell that it is not starving in my speed test where many files are present on the disc.

sh-rp: The number for running empty jobs in this test are:
1000 jobs = 4s
2000 jobs = 12s
10k jobs = 372s
So it does go up quite a bit, but definately not from listing all files on every loop (as it does not do this). What I would do: Merge this like it is now, it already is faster than the old implementation, and in a followup ticket with plenty tests we cache the parsed jobs in some in memory structure that gets refreshed periodically or when the new jobs in this structure fall below the count of open slots.

tests/pipeline/test_pipeline_trace.py
sh-rp: Alright, now I get it. I fixed the dummy config and I was logging 2 exceptions per failed job, which I have changed to only log the one from the loader, now it is good again.

dlt/load/load.py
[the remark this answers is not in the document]
sh-rp: I have added a generic method for a client to prepare for a job to be executed on the thread and am using this to set the querytags, this seemed to me to be the cleanest solution

dlt/common/storages/load_package.py
```diff
@@ -723,19 +723,12 @@ def build_job_file_name(
 
     @staticmethod
     def is_package_partially_loaded(package_info: LoadPackageInfo) -> bool:
```
sh-rp, line 725: the behavior is unified now between the different package states, I'd say this is correct.

rudolfix (approved):
LGTM!

PR #1496, opened 2024-06-19
Support Open Source ClickHouse Deployments

### LINKED ISSUE #1387: ClickHouse MergeTree Support
# Support for MergeTree Engine in ClickHouse Destination

## Problem
The ClickHouse destination in dlt 0.4.11 currently only supports the `ReplicatedMergeTree` table engine. However, many users have requested support for the standard non-replicated `MergeTree` engine to enable local development and testing deployments.

## Desired Behavior 
The ClickHouse destination should correctly handle `ENGINE=MergeTree` and create the appropriate table structure in ClickHouse.

## Proposed Solution
Enhance the ClickHouse destination adapter with the following:

1. Automatically detect whether the connected ClickHouse instance is a Cloud instance (replicated) or a local self-managed instance. 
   - For ClickHouse Cloud, default to `ReplicatedMergeTree` engine
   - For non-cloud instances, default to standard `MergeTree`
   
```sql
SELECT value
FROM system.settings
WHERE name = 'cloud_mode';
```

2. Allow users to explicitly specify the desired engine (`MergeTree` or `ReplicatedMergeTree`) in the destination configuration. This will override the automatic detection.

3. Properly interpolate the provided engine into the `CREATE TABLE` statement executed against ClickHouse.

Note: To limit scope, this enhancement will not include support for specifying replication, zookeeper or shard details for the `ReplicatedMergeTree` engine. Users requiring those customizations can continue to specify the full engine definition in their configuration.

## Considerations
- Thoroughly test the solution against both ClickHouse Cloud and various self-managed ClickHouse versions
- Update documentation to explain the new behavior and configuration options
- Consider any impacts to downstream systems or processes relying on the current hard-coded `ReplicatedMergeTree` behavior

### LINKED ISSUE #1503: Open Source ClickHouse fails on Timestamp with Timezone
### dlt version

0.4.13a0

### Describe the problem

```shell
poetry run pytest tests/load/test_job_client.py::test_load_with_all_types -k "clickhouse"
```

```
        # do not use the DBAPI driver for local files.
        if not bucket_path:
            # Local filesystem.
            if ext == "jsonl":
                compression = "gz" if FileStorage.is_gzipped(file_path) else "none"
            try:
                with clickhouse_connect.create_client(
                    host=client.credentials.host,
                    port=client.credentials.http_port,
                    database=client.credentials.database,
                    user_name=client.credentials.username,
                    password=client.credentials.password,
                    secure=bool(client.credentials.secure),
                    settings={
                        "allow_experimental_lightweight_delete": 1,
                        "enable_http_compression": 1,
                        "input_format_json_named_tuples_as_objects": 1,
                        "input_format_json_read_numbers_as_strings": 0,
                    },
                ) as clickhouse_connect_client:
                    insert_file(
                        clickhouse_connect_client,
                        qualified_table_name,
                        file_path,
                        fmt=clickhouse_format,
                        settings={
                            "allow_experimental_lightweight_delete": 1,
                            "enable_http_compression": 1,
    
                    },
                        compression=compression,
                    )
            except clickhouse_connect.driver.exceptions.Error as e:
>               raise LoadJobTerminalException(
                    file_path,
                    f"ClickHouse connection failed due to {e}.",
                ) from e
E               dlt.destinations.exceptions.LoadJobTerminalException: Job with id/file name /home/marcel/Work/dltHub/dlt/_storage/event_test_table647d6e37d7652dc525157f9c058a030a.f45bf72149.0.jsonl encountered unrecoverable problem: ClickHouse connection failed due to :HTTPDriver for http://localhost:8123 returned response code 400)
E                Code: 27. DB::Exception: Cannot parse input: expected '"' before: '+00:00","col5":"string data \\n \\r  🦆","col6":"2323.34","col7":"YmluYXJ5IGRhdGEgCiANIA==","col8":72057686131818776,"col9":{"complex":[1,2,3,"a"],"link":"?comm': (while reading the value of key col4): (at row 1)
E               : While executing ParallelParsingBlockInputFormat. (CANNOT_PARSE_INPUT_ASSERTION_FAILED) (version 24.6.1.4322 (official build))
E               .
```

### Expected behavior

Passes all tests same as Clickhouse cloud.

### Steps to reproduce

Start a clickhouse server as described in https://clickhouse.com/docs/en/getting-started/quick-start

Set the username and password etc on server, then amend config file:

```toml
[destination.clickhouse.credentials]
database = "dlt"
username = "dlt_ci"
password = "..."
host = "localhost"
port = 9000
http_port = 8123
secure = 0
```

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

**ClickHouse local version 24.6.1.4322 (official build).**

### Additional information

_No response_

Pipboyguy:
### Description

This will **improves support for self-managed ClickHouse open source deployments**, while maintaining compatibility with ClickHouse Cloud deployments.

- Allow explicitly specifying the desired engine via `table_engine_type` in `clickhouse_adapter`. Valid types are `merge_tree`, `replicated_merge_tree`, `shared_merge_tree`, `stripe_log`, `tiny_log`
- Default to `MergeTree` if no engine is specified, which now works for both Cloud and self-managed deployments
- Update tests to check for the appropriate engine based on annotation
- CH cloud has "date_time_input_format" set to "best_effort", while OS deployments don't. This caused some tests to fail for OS deployments. We override setting for `clickhouse_connect` sessions.
- Clarify some networking details for making dlt work with OS deployments in docs

### Related Issues

- Closes #1387
- Closes #1503 

### Additional Context

Note that this change does not include support for specifying replication, ZooKeeper or shard details for the `ReplicatedMergeTree` engine. Users requiring those customizations can continue to specify the full engine definition in their configuration.


docs/website/docs/dlt-ecosystem/destinations/clickhouse.md
```diff
@@ -104,7 +111,8 @@ Data is loaded into ClickHouse using the most efficient method depending on the
 
 `Clickhouse` does not support multiple datasets in one database, dlt relies on datasets to exist for multiple reasons.
 To make `clickhouse` work with `dlt`, tables generated by `dlt` in your `clickhouse` database will have their name prefixed with the dataset name separated by
-the configurable `dataset_table_separator`. Additionally, a special sentinel table that does not contain any data will be created, so dlt knows which virtual datasets already exist in a
+the configurable `dataset_table_separator`.
+Additionally, a special sentinel table that doesn't contain any data will be created, so dlt knows which virtual datasets already exist in a
```
jorritsandbrink, line 115: Perhaps out of scope for this PR, but have you considered using pipeline [state](https://dlthub.com/docs/general-usage/state) instead of a sentinel table?

tests/load/clickhouse/test_clickhouse_adapter.py
```diff
@@ -25,37 +36,50 @@ def not_annotated_resource():
...
+                ):
+                    assert tuple(res[0])[2] in (
+                        "MergeTree",
+                        "SharedMergeTree",
+                        "ReplicatedMergeTree",
+                    )
+                else:
+                    # Non annotated resource needs to default to detected installation
+                    # type, i.e. cloud or self-managed.
+                    # CI runs on CH cloud, so will be `SharedMergeTree`.
```
jorritsandbrink, line 71: Should we run CI both for ClickHouse cloud and ClickHouse local, like Postgres?

Seems like we don't have proper test coverage without local (e.g. for the `Log` table engines).

jorritsandbrink (commented):
Code itself looks good to me. Just two comments regarding test coverage and the use of sentinel tables.

tests/load/clickhouse/test_clickhouse_adapter.py
Pipboyguy: - Added ClickHouse OSS as a local destination test in `.github/workflows/test_local_destinations.yml`, so CI also runs local tests.
- Removed the log engine family for now. I'm not even sure why I brought it into this ticket, but it will require more work than what can be done here.

docs/website/docs/dlt-ecosystem/destinations/clickhouse.md
Pipboyguy: That is a really good suggestion!

I've dropped the sentinel table workaround for more canonical usage of dlt:

 https://github.com/dlt-hub/dlt/blob/409487ce0669a06941769c0a5414c8d71289d2b7/dlt/destinations/impl/clickhouse/clickhouse.py#L310-L345

.github/clickhouse-compose.yml
jorritsandbrink, line 1: I think this file should be in `tests/load/clickhouse`, similar to where `dremio` keeps it `docker-compose.yml`. I know `weaviate` also keeps it under `.github`, but that makes less sense to me, because you can also use the file when testing locally, not only on CI.

.github/workflows/test_destination_clickhouse.yml
```diff
@@ -67,12 +66,49 @@ jobs:
...
+      # OSS ClickHouse
+      - run: |
+          docker-compose -f ".github/clickhouse-compose.yml" up -d
+          echo "Waiting for ClickHouse to be healthy..."
+          timeout 30s bash -c 'until docker-compose -f ".github/clickhouse-compose.yml" ps | grep -q "healthy"; do sleep 1; done'
+          echo "ClickHouse is up and running"
+        name: Start ClickHouse OSS
+
+
+      - run: poetry run pytest tests/load -m "essential"
```
jorritsandbrink, line 78: This is missing `if` conditions. Look at https://github.com/dlt-hub/dlt/blob/af7752725b34ff9fa6f081f1c2483ac767944284/.github/workflows/test_destinations.yml#L86.

The essential tests run twice without such conditions.

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -302,13 +307,50 @@ def __init__(
         self.active_hints = deepcopy(HINT_TO_CLICKHOUSE_ATTR)
         self.type_mapper = ClickHouseTypeMapper(self.capabilities)
 
+    def has_dataset(self) -> bool:
```
jorritsandbrink, line 310: Not sure if this is the correct way to interact with state. I think we need @rudolfix to review this block.

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -217,3 +217,9 @@ def _make_database_exception(cls, ex: Exception) -> Exception:
     @staticmethod
     def is_dbapi_exception(ex: Exception) -> bool:
         return isinstance(ex, clickhouse_driver.dbapi.Error)
+
+    def _get_deployment_type(self) -> TDeployment:
```
jorritsandbrink, line 221: Seems this is only used in the tests. Probably better to turn it into a test utility.

(I did the same in a PR once, and that was the feedback I received then.)

jorritsandbrink (commented):
Cool stuff :) Just some minor comments.

Note that `tests/load/pipeline/test_restore_state.py::test_reset_pipeline_on_deleted_dataset[clickhouse-jsonl-no-staging]` is failing. Perhaps a regression after moving away from sentinal tables.

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: Yeah I've a bit stumped here. Thanks @rudolfix

dlt/destinations/impl/clickhouse/sql_client.py
Pipboyguy: Agreed - refactored into test utils

.github/workflows/test_destination_clickhouse.yml
Pipboyguy: Thank you Jorrit! I missed this one.

.github/clickhouse-compose.yml
Pipboyguy: Agreed! It felt rather clumsy to put it in .github directory.

Moved to `tests/load/clickhouse`

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: This test seems to be passing intermittently.

jorritsandbrink (commented):
Thanks for addressing the comments, LGTM!

Just this comment regarding state interaction is still open: https://github.com/dlt-hub/dlt/pull/1496#discussion_r1659801825

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -340,18 +336,21 @@ def start_file_load(self, table: TTableSchema, file_path: str, load_id: str) ->
...
         sql = SqlJobClientBase._get_table_update_sql(self, table_name, new_columns, generate_alter)
 
         if generate_alter:
             return sql
 
-        # Default to 'ReplicatedMergeTree' if user didn't explicitly set a table engine hint.
-        table_type = cast(
-            TTableEngineType, table.get(TABLE_ENGINE_TYPE_HINT, "replicated_merge_tree")
-        )
+        # Default to 'MergeTree' if the user didn't explicitly set a table engine hint.
```
rudolfix, line 350: doing this in adapter is really cool. just one thought? should we add this to clickhouse configuration anyway? so users can change the engine for all the resources without the need for adapter for each one? it will be similar to other `dlt` settings ie. where you can `iceberg` table format in athena for all resources (and also per resource with the decorator)

docs/website/docs/dlt-ecosystem/destinations/clickhouse.md
rudolfix: @Pipboyguy where is this change? I can't find it in the PR. Overall idea is good but in this very case it is impossible to execture. `dlt` stores state in the very dataset you are creating here. so it cannot be used to check if the dataset was created (it must be in order to use state).

in this specific code you are using package state. the scope of this state is a single load. so you will not be able to refer to it in the next load. some simple tests will pass. but not much else

sentinel table is quite a good idea and let's keep it. I plan to standardize it maybe in some kind of mixin class. We use sentinels in vector databases and dremio and the pattern is always the same.

rudolfix (changes requested):
@Pipboyguy @jorritsandbrink this is really good. thanks for working on it and all reviews. from my side
- pls keep sentinel table. I elaborated quite a bit why
- consider adding the table engine in the clickhouse configuration to have easy global swithc

docs/website/docs/dlt-ecosystem/destinations/clickhouse.md
Pipboyguy: @rudolfix I already reverted back to sentinel table method in latest commits. It is indeed a much simpler a more robust solution imho

dlt/destinations/impl/clickhouse/clickhouse.py
Pipboyguy: @rudolfix Added!

https://github.com/dlt-hub/dlt/blob/f852d44d41ff77327a4735772ccd8f7affb03bb3/dlt/destinations/impl/clickhouse/configuration.py#L43-L44

dlt/destinations/impl/clickhouse/sql_client.py
```diff
@@ -98,13 +103,16 @@ def execute_sql(
...
             self.credentials.dataset_sentinel_table_name
         )
-        self.execute_sql(
-            f"""CREATE TABLE {sentinel_table_name} (_dlt_id String NOT NULL PRIMARY KEY) ENGINE=ReplicatedMergeTree COMMENT 'internal dlt sentinel table'"""
-        )
+        # `MergeTree` is guaranteed to work in both self-managed and cloud setups.
+        self.execute_sql(f"""
+            CREATE TABLE {sentinel_table_name}
+            (_dlt_id String NOT NULL PRIMARY KEY)
+            ENGINE=MergeTree
```
rudolfix, line 114: I'm probably nitpicking but should it use config.table_engine_type here?

rudolfix (dismissed):
LGTM! I have just one small suggestion which may not make sense. let's wait for @jorritsandbrink review and we can merge it

jorritsandbrink (commented):
Overall it looks great, just a comment regarding organization of config attributes.

dlt/destinations/impl/clickhouse/configuration.py
```diff
@@ -36,6 +40,8 @@ class ClickHouseCredentials(ConnectionStringCredentials):
     """Timeout for sending and receiving data. Defaults to 300 seconds."""
     dataset_table_separator: str = "___"
     """Separator for dataset table names, defaults to '___', i.e. 'database.dataset___table'."""
+    table_engine_type: Optional[TTableEngineType] = "merge_tree"
```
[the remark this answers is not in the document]
rudolfix: @Pipboyguy yes I agree with @jorritsandbrink 
this is probably due to sql_client not seeing config but credentials - but you can pass additional arguments to it
so my take is that we move it and if it is hard then we'll need to change how `sql_client` is instantiated. I can help in that case

dlt/destinations/impl/clickhouse/sql_client.py
Pipboyguy: @rudolfix Thanks for the spot!

Changed to config table engine:

https://github.com/dlt-hub/dlt/blob/9c67c65cc1e50fb4ab746bebf0437485b600fff0/dlt/destinations/impl/clickhouse/sql_client.py#L116-L121

dlt/destinations/impl/clickhouse/configuration.py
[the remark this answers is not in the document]
Pipboyguy: Totally agree. These don't belong in credentials.

Please find latest changes in both code base and docs.

jorritsandbrink (approved):
LGTM!

I like the `typing` module.

Issue #1009, opened 2024-02-27
CVE in orjson < 3.9.15 and orjson pinned to <=3.9.10 in dlt


b-per:
### dlt version

0.4.4

### Describe the problem

"orjson.loads in orjson before 3.9.15 does not limit recursion for deeply nested JSON documents" ([dependabot link](https://github.com/advisories/GHSA-pwr2-4v36-6qpr))

But dlt 0.4.4 pins `orjson` to "<=3.9.10" so we can't upgrade `orjson` to the fix

### Expected behavior

`dlt` doesn't pin`orjson` to "<=3.9.10" and we can update `orjson` to the new version with the fix

### Steps to reproduce

N/A

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

rudolfix:
@b-per we limit the version because there's a buffer overrun bug introduced to orjson that makes bigger load crash. and it is probably exploitable. 
the CVE looks like more DDOS, stack overflow should not be easily exploitable so we probably keep it until they really fix the above.
The PR log does not look promising (already 3 PRs that failed to fix the bug: 

we can go back to simplejson which is still available:
https://dlthub.com/docs/reference/performance#use-built-in-json-parser
just uninstall orjson for that to happen

PR #1501, opened 2024-06-20
Extend orjson dependency allowed range with excluded versions

steinitzu:
### Description

Extends the orjson version range to allow `>=3.6.7,<4` with some known problematic versions excluded

Segafault bug introduced in 3.9.11 seems to be fixed in 3.9.15 https://github.com/ijl/orjson/pull/459#issuecomment-1961733799  

Another crashing bug mentioned in changelog introduced in 3.10.1 (and fixed in 3.10.2) so I'm excluding that version too https://github.com/ijl/orjson/blob/7b081e10768a44699f6f7231521b1803c3dfdd87/CHANGELOG.md#fixed

This is within lanchain's range as well:

```
$ pipdeptree -p langchain|grep orjson
│   │   ├── orjson [required: >=3.9.14,<4.0.0, installed: 3.10.5]
```

### Related Issues

- Fixes https://github.com/dlt-hub/dlt/issues/1009

### Additional Context

https://github.com/dlt-hub/dlt/pull/1469


rudolfix:
@steinitzu that's right. there were more bugs introduced later and fixed. if we see any problems with `orjson` in the future I'll remove it from `dlt`

rudolfix (dismissed):
LGTM!

PR #1502, opened 2024-06-20
rest_api: add an example to the incremental load section

burnash:
Adds an example and elaborates on incremental configuration following up the Slack community question:
https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1718889819509209


docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
```diff
@@ -560,49 +560,81 @@ This will include the `id`, `title`, and `created_at` fields from the `issues` r
...
 
-    This configuration is more flexible and allows you to specify the start and end conditions for the incremental loading.
+The alternative method is to use the `incremental` field in the [endpoint configuration](#endpoint-configuration). This method is more flexible and allows you to specify the start and end conditions for the incremental loading:
+
+```py
+{
+    "incremental": {
+        "start_param": "<parameter_name>",
+        "end_param": "<parameter_name>",
+        "cursor_path": "<path_to_cursor_field>",
```
AstrakhantsevaAA, line 632: could you note in comments what types and formats these parameters have, e.g.

```
    "incremental": {
        "cursor_path": "<path_to_cursor_field>", # "created_at"
```

for me it's not clear what I should put in `start_param` and `end_param`

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
```diff
@@ -560,49 +560,81 @@ This will include the `id`, `title`, and `created_at` fields from the `issues` r
...
+    "<parameter_name>": {
+        "type": "incremental",
+        "cursor_path": "<path_to_cursor_field>",
+        "initial_value": "<initial_value>",
+    },
+}
+```
+
+- `type`: The type of the incremental parameter. Set to `incremental`.
+- `cursor_path`: The JSONPath to the field within each item in the list that will be used as the cursor value. In this case, it's `created_at`. Note that the path starts from the root of the item (dict) and not from the root of the response.
```
omri-antman-imagen, line 620:
```suggestion
- `cursor_path`: The JSONPath to the field within each item in the list that will be used as the cursor value. In this case, it's `created_at`. Note that the path starts from the root of the item (dict) and not from the root of the response.

`{
    "responses": [
        {
            "submissionId": "ggg",
            "submissionTime": "2024-04-17T20:09:38.716Z",
            "lastUpdatedAt": "2024-04-17T20:09:38.716Z",
        },
        {
            "submissionId": "yyy",
            "submissionTime": "2024-04-17T20:11:38.716Z",
            "lastUpdatedAt": "2024-04-17T20:11:38.716Z",
        }
    ]
}`
For example, in the response above, and the cursor path is inside the array.
Thus, we should modify the `data_selector` parameter to "responses", and the `cursor_path` to "submissionTime"
`                "endpoint": {
                    "data_selector": "responses",
                    "params": {
                        "since": {
                            "type": "incremental",
                            "cursor_path": "submissionTime",
                            "initial_value": "2022-01-01T11:21:28Z",
                        },
                    },
                },`
```

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
burnash: Thanks for the comment, added more details in 70e43afc94719ec428730961827205e3e547ed31

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
burnash: Hey @omri-antman-imagen, thanks for the suggestion. I believe introducing a second JSON example would be confusing for the reader, so I suggest using only one example (e.g. "posts"). Nevertheless, I think I get your point, so I adjusted the text to really stress that the `cursor_path` is within the data item and added an example to illustrate that in commit 70e43afc94719ec428730961827205e3e547ed31. Please take a look.

AstrakhantsevaAA (approved):
Very good Anton!

Issue #1463, opened 2024-06-14
Wrong Merge Key Not Throwing Error


zem360:
### dlt version

0.4.12

### Describe the problem

While working on a community support request regarding snake_case, camel_case @dat-a-man discovered this bug. 

A typo in the `merge_key` or the wrong `merge_key` doesn't throw an error.

### Expected behavior

If there is a typo in the merge_key or the provided key is not in the data an error should be thrown, but the code is running normally.

In the code snippet provided in **Steps to Reproduce** the `merge_key = 'mana'` should throw an error as it is not present in the data, but that is not the case.

### Steps to reproduce

Run the following code:
```
@dlt.resource(name='table_name', merge_key = "mana", write_disposition={"disposition": "merge"})
def func():
    data = [{'id':1, 'NAme':'abcaaaa', 'status':'bronze'},
            {'id':2,'NAme':'deaf','status':'bronze'}]
    yield data
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.10

rudolfix:
the reason is that we use `merge_key` only during the loading. so before that no checks are done. same thing will happen to any other hint including `primary_key` (if not part of incremental which is actually checking the data).

to really fix this issue (before loading starts) we'd need to track which columns received data, currently we track only table level. maybe we can do that in a separate ticket.

what we can do now:
- there's a warning in the code somewhere that warns the user if there are any columns that do not have a data type (so we know they've never seen data). maybe it is gone? maybe we should warn the user at the end of normalization stage?
- let's verify why the code is working normally. merge keys are NOT NULL. probably columns are skipped when syncing to database. how to fix: escalate the warning to exception for all incomplete columns that are NOT NULL so we fail at normalize

PR #1504, opened 2024-06-21
Raise/warn on incomplete columns in normalize

steinitzu:
### Description
Turns the "unbound column" warning into an exception for not-null columns and move it to normalize

### Related Issues

- Fixes https://github.com/dlt-hub/dlt/issues/1463

### DIFF
--- a/dlt/common/schema/exceptions.py
+++ b/dlt/common/schema/exceptions.py
@@ -8,6 +8,7 @@ from dlt.common.schema.typing import (
     TSchemaEvolutionMode,
 )
 from dlt.common.normalizers.naming import NamingConvention
+from dlt.common.schema.typing import TColumnSchema, TColumnSchemaBase
 
 
 class SchemaException(DltException):
@@ -231,3 +232,27 @@ class TableIdentifiersFrozen(SchemaException):
 
 class ColumnNameConflictException(SchemaException):
     pass
+
+
+class UnboundColumnException(SchemaException):
+    def __init__(self, schema_name: str, table_name: str, column: TColumnSchemaBase) -> None:
+        self.column = column
+        self.schema_name = schema_name
+        self.table_name = table_name
+        nullable = column.get("nullable", True)
+        merge_key = column.get("merge_key", False)
+        primary_key = column.get("primary_key", False)
+        if merge_key:
+            col_type_msg = "merge key"
+        elif primary_key:
+            col_type_msg = "primary key"
+        elif nullable:
+            col_type_msg = "nullable"
+        else:
+            col_type_msg = "non-nullable"
+        super().__init__(
+            f"A {col_type_msg} column {column['name']} in table {table_name} in schema"
+            f" {schema_name} is incomplete. It was not bound to the data during normalizations"
+            " stage and its data type is unknown. Did you add this column manually in code? Make"
+            " sure spelling correctly matches the data."
+        )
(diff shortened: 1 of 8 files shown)


sh-rp:
I'm wondering if we should not do this in the extraction step already. All columns that are non-nullable (and merge and primary keys should be that) should raise if not populated. Extraction spends time on I/O mostly and not on python code as the normalizer, so the check would not make a big difference in performance in my opinion.

steinitzu:
> I'm wondering if we should not do this in the extraction step already. All columns that are non-nullable (and merge and primary keys should be that) should raise if not populated. Extraction spends time on I/O mostly and not on python code as the normalizer, so the check would not make a big difference in performance in my opinion.

I agree it would be much better to fail early if possible. Ideally we could tell right after the first data item is extracted.  
But I wasn't sure if we can always tell whether the column is populated in extract. The "seen data" marker for the table is only set in normalize so I was going by that. But I'll give it a try.

dlt/common/schema/exceptions.py
```diff
@@ -156,3 +157,27 @@ def __init__(self, table_name: str) -> None:
...
+        if merge_key:
+            col_type_msg = "merge key"
+        elif primary_key:
+            col_type_msg = "primary key"
+        elif nullable:
+            col_type_msg = "nullable"
+        else:
+            col_type_msg = "non-nullable"
+        super().__init__(
+            f"A {col_type_msg} column {column['name']} in table {table_name} in schema"
```
sh-rp, line 179: I would make this error message better. Think about the perspective of the user that does not know the inner workings of dlt. It should be something like "The column xxx on table yyy is marked as non-nullable but did not receive any data during this load. This can happen if you manually specify non-nullable columns (for example primary or merge keys) in your code or import schema but this column does not match any incoming data."

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -989,3 +989,24 @@ def r():
     with pytest.raises(PipelineStepFailed) as pip_ex:
         p.run(r())
     assert isinstance(pip_ex.value.__context__, SchemaException)
+
+
+@pytest.mark.parametrize(
```
sh-rp, line 994: Can you write a test (or check if one exists) to see what happens when we do a merge on merge keys but some rows have null in the merge key? It's not super important right now, but if it would be interesting to know what happens :)

sh-rp (changes requested):
Looks good, small requests

tests/load/pipeline/test_merge_disposition.py
steinitzu: I couldn't find a test so I added one. This was raising an exception already through `schema.coerce_row` in normalize

steinitzu:
> I'm wondering if we should not do this in the extraction step already. All columns that are non-nullable (and merge and primary keys should be that) should raise if not populated. Extraction spends time on I/O mostly and not on python code as the normalizer, so the check would not make a big difference in performance in my opinion.

Was looking into if this was possible also, but I don't think so without moving a lot of normalize logic into extract. I wasn't sure how much schema inferrence is done in extract, seems there is none.

rudolfix:
@steinitzu this needs merge from devel. we did a lot of updates. @sh-rp otherwise this PR is good to go?

steinitzu:
> @steinitzu this needs merge from devel. we did a lot of updates. @sh-rp otherwise this PR is good to go?

branch is up to date now and tests passing

rudolfix (approved):
LGTM!

PR #1506, opened 2024-06-21
fix(airflow): clear dag folders after run

### LINKED ISSUE #1464: airlfow helper does not clean pipeline working folder when DAG is created
### dlt version

0.4.12

### Describe the problem

`dlt` creates a new working folder in a `/tmp` for each (1) dag creation (2) each actual dag execution. in case of (2) we clean the working directory after run (see `_run`) method of the task group
in case of (1) we do not do it, ie. in this snippet
```python
@dag(
    schedule_interval='@daily',
    start_date=pendulum.datetime(2023, 7, 1),
    catchup=False,
    max_active_runs=1,
    default_args=default_task_args
)
def load_data():
    # set `use_data_folder` to True to store temporary data on the `data` bucket. Use only when it does not fit on the local storage
    tasks = PipelineTasksGroup("pipeline_decomposed", use_data_folder=False, wipe_local_data=True)

    # import your source from pipeline script
    from github import github_repo_events

    source = github_repo_events("apache", "airflow")

    # modify the pipeline parameters 
    pipeline = dlt.pipeline(pipeline_name='pipeline_name',
                     dataset_name='dataset_name',
                     destination='duckdb',
                     full_refresh=False # must be false if we decompose
                     )
    # create the source, the "serialize" decompose option will converts dlt resources into Airflow tasks. use "none" to disable it
    tasks.add_run(pipeline, source, decompose="serialize", trigger_rule="all_done", retries=0, provide_context=True)
```
`pipeline` creates a temp working dir (just to create a DAG), the code ends and the working dir is left.

### Expected behavior

`dlt` will clean the working folder from the DAG creation, observing the `wipe_local_data=True` of `PipelineTasksGroup`. 
In order to fix it we could convert our task group into a context manager:
```python
with PipelineTasksGroup("pipeline_decomposed", use_data_folder=False, wipe_local_data=True) as tasks:
  ...
  tasks.add_run(pipeline, source, decompose="serialize", trigger_rule="all_done", retries=0, provide_context=True)
```
what happens here:
1. `tasks.add_run` keeps the pipeline instance(s) so it can use them to wipe the folder
2. on `__exit__` all working folders are wiped for all pipelines (flag permitting)

### Steps to reproduce

Please write a test that reproduces this behavior

### Operating system

Linux

### Runtime environment

Airflow

### Python version

3.11

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

IlyaFaer:
Fixes #1464

### DIFF
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -1,5 +1,6 @@
 import functools
 import os
+import shutil
 from tempfile import gettempdir
 from typing import Any, Callable, Dict, List, Literal, Optional, Sequence, Tuple
 
@@ -337,6 +338,7 @@ class PipelineTasksGroup(TaskGroup):
             if self.wipe_local_data:
                 logger.info(f"Removing folder {pipeline.working_dir}")
                 task_pipeline._wipe_working_folder()
+                shutil.rmtree(os.environ["DLT_DATA_DIR"])
 
     @with_telemetry("helper", "airflow_add_run", False, "decompose")
     def add_run(
--- a/dlt/common/libs/deltalake.py
+++ b/dlt/common/libs/deltalake.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict, Union
+from typing import Optional, Dict, Union, List
 from pathlib import Path
 
 from dlt import version, Pipeline
@@ -71,9 +71,13 @@ def write_delta_table(
     table_or_uri: Union[str, Path, DeltaTable],
     data: Union[pa.Table, pa.RecordBatchReader],
     write_disposition: TWriteDisposition,
+    partition_by: Optional[Union[List[str], str]] = None,
     storage_options: Optional[Dict[str, str]] = None,
 ) -> None:
-    """Writes in-memory Arrow table to on-disk Delta table."""
+    """Writes in-memory Arrow table to on-disk Delta table.
+
+    Thin wrapper around `deltalake.write_deltalake`.
+    """
 
     # throws warning for `s3` protocol: https://github.com/delta-io/delta-rs/issues/2460
     # TODO: upgrade `deltalake` lib after https://github.com/delta-io/delta-rs/pull/2500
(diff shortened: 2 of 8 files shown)


dlt/helpers/airflow_helper.py
```diff
@@ -337,6 +338,7 @@ def log_after_attempt(retry_state: RetryCallState) -> None:
             if self.wipe_local_data:
                 logger.info(f"Removing folder {pipeline.working_dir}")
                 task_pipeline._wipe_working_folder()
+                shutil.rmtree(os.environ["DLT_DATA_DIR"])
```
rudolfix, line 341: here you fully wipe a folder ie `/tmp`. it is not a private pipeline folder. it is a folder where all the pipelines dirs are stored.
you can erase tons of unrelated data like that. also does not solve the problem in the issue

rudolfix (changes requested):
@IlyaFaer please read the issue again. Airflow creates `Pipeline` instance twice
1. to infer, visualise and gather task to be executed. This happens every time user uses UI and every 10 seconds during execution. typically on the main node, not the executor
2. To actually run the pipeline

the problem we create is in (1) - every 10 seconds we create another pipeline dir in /tmp and we are not wiping it because we wipe it only in `_run`

dlt/helpers/airflow_helper.py
IlyaFaer: Hm. At least locally it doesn't work like this. Every dag creates (and leaves) a folder in a form of `/tmp/dlt_******`. And when I remove `DLT_DATA_DIR`, it actually points to this folder, not to the entire `/tmp`. So, it only wipes the folder related to this particular run, not touching anything else.

image

So... I don't know, for now I don't see why this change is incorrect.

rudolfix: it is a "minor" problem (user can configure a persistent `DLT_DATA_DIR` for his/her pipelines, this is what this setting is for) - your code will wipe it all.

still the main issue is not addressed, did you read point (1)? is there something wrong with this observation?

Issue #1095, opened 2024-03-15
access data after load load as dataframes with ibis


rudolfix:
**Background**
ibis https://github.com/ibis-project/ibis is a library that translates dataframe expressions into SQL statement and then executes them in the destination. they do nice work of compiling final SQL statement with sqlglot (so probably resultant SQL is quite optimized)

We have large overlap in destinations and we were looking for decent dataframe -> sql thing from the very start. it seems that's it: we can easily build a helper that exposes any dlt dataset as dataframe, share credentials etc.

**Implementation**
We can integrate deeply or via a helper scheme. In case of helper, we allow users to get `ibis` connection from `dlt` destination and/or pipeline. The UX will be similar to `dbt` helper.

Deep integration means that we expose the loaded data from the `Pipeline`, `DltSource` and `DltResource` instances. ie.
```
pipeline.run(source, destination="bigquery", dataset_name="data")

# duckdb-like interface

# get rows dbapi style
for row in pipeline.dataset.sql("SELECT * FROM table"):
    ...

# get materialized dataframes
for batch in pipeline.dataset.sql("SELECT * FROM table").arrow(batch_size=10000):
    ...

# get lazy dataframe via ibis
ibis_table = pipeline.dataset.df(table_name)
ibis_connection = ibis_table = pipeline.dataset.df()

# we can expose the `dataset` property of the pipeline via source (and internal resources as well), in that case we automatically bind 
to right schema

ibis_table = resource.dataset.df()
ibis_connection = source.dataset.df()
```

Implementation is straightforward for sql-like destinations. We won't support vector databases.
It would be really interesting to support filesystem destination as above. ie. by registering the json and parquet files in temporary duckdb database and then exposing the database for `ibis` and `sql` access methods

** Ibis Connection sharing**
We are discussing a connection sharing approach with ibis here: https://github.com/ibis-project/ibis/issues/8877. As mentioned in the comments there, we could build it in a way that we manage the connection and ibis provides backends that accept an open connection and DO NOT need any addtionally dependencies.

lostmygithubaccount:
hi @rudolfix, I'm working on Ibis and we were just discussing `dlt` and a potential integration! I'm curious how you found Ibis per:

> we were looking for decent dataframe -> sql thing from the very start

we'd be happy to help move this along, particularly if there are any questions we can answer. in my cursory look at `dlt` and its destinations last week, basically all except the dedicated vector databases are covered as Ibis backends. let us know if we can help move this forward!

[...]

PR #1507, opened 2024-06-21
[experiment] expose readable datasets as dataframes and arrow tables

sh-rp:
### Description
As an alternative to the ibis integration, we are testing out wether we can create our own data reader with not too much effort that works across all destinations.


composable_pipeline_1.py
```diff
@@ -0,0 +1,109 @@
...
+                        "id": order_no,
+                        "customer_id": customer["id"],
+                        "product": product["name"],
+                        "price": product["price"],
+                    }
+
+    # run and print result
+    print("RUNNING WAREHOUSE INGESTION")
+    print(duck_pipeline.run([c(), o()]))
+    print(duck_pipeline.dataset.df(table="customers"))
```
rudolfix, line 57: This is super close. I'd modify this a little bit (following the ticket #1095)
access data in a table
`dataset.customers` or `dataset["customers"]` or `dataset.sql("SELECT * FROM customers")` creates a relation. relation is leazily evaluated. that happens when those methods are called:
* df()
* iter_df(batch_size) - iterator
* arrow()
* iter_arrow(batch_size)
* fetch() -> db api fetch all
* iter_fetch(batch_size) -> fetchmany
* ibis() - converts relation to IBIS relation via https://ibis-project.org/how-to/extending/sql

plugging in IBIS:
(not sure about that)
* dataset.df() - IBIS client on which we can get Table relation
* dataset.df("customer") - IBIS table relation on customers
* datasert.customers.ibis() - same as above

rudolfix (commented):
@sh-rp pls see my comments

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -565,3 +577,83 @@ def get_table_jobs(
...
+            sql = f"SELECT * FROM {table}"
+        elif not prepare_tables or not sql:
+            raise Exception(
+                "You must either provide a table argument or sql and prepare table arguments to"
+                " access this dataset"
+            )
+
+        # create in memory table, for now we read all available files
+        db = duckdb.connect(":memory:")
+        db.register_filesystem(self.fs_client)
```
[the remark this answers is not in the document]
sh-rp: yes it is!

dlt/common/destination/reference.py
```diff
@@ -498,6 +503,30 @@ def should_truncate_table_before_load_on_staging_destination(self, table: TTable
         return True
 
 
+class SupportsDataAccess(ABC):
+    """Add support for accessing data as arrow tables or pandas dataframes"""
+
+    @abstractmethod
+    def iter_df(
```
jorritsandbrink, line 510: I wouldn't explicitly expose Pandas dataframes. I would only expose Arrow data structures , because the user can call `to_pandas` on those.

dlt/common/destination/reference.py
```diff
@@ -498,6 +503,30 @@ def should_truncate_table_before_load_on_staging_destination(self, table: TTable
...
+        self,
+        *,
+        table: str = None,
+        batch_size: int = 1000,
+        sql: str = None,
+        prepare_tables: List[str] = None,
+    ) -> Generator[DataFrame, None, None]: ...
+
+    @abstractmethod
+    def iter_arrow(
```
jorritsandbrink, line 520: It's probably a good idea to support `pyarrow.RecordBatchReader` alongside `pyarrow.Table` for larger-than-memory data.

Or just expose a `pyarrow.Dataset`, from which the user can create either a `pyarrow.RecordBatchReader` or `pyarrow.Table`.

dlt/destinations/sql_client.py
```diff
@@ -274,6 +286,39 @@ def _truncate_table_sql(self, qualified_table_name: str) -> str:
...
+        prepare_tables: List[str] = None,
+    ) -> Generator[DataFrame, None, None]:
+        if not sql:
+            table = self.make_qualified_table_name(table)
+            sql = f"SELECT * FROM {table}"
+
+        # iterate over results in batch size chunks
+        with self.execute_query(sql) as cursor:
+            while True:
+                if not (result := cursor.fetchmany(batch_size)):
```
jorritsandbrink, line 304: Ideally we immediately receive Arrow data from the SQL database. Columnar-to-row-to-columnar conversions will slow things down a lot.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -565,3 +577,83 @@ def get_table_jobs(
...
+
+        # create all tables in duck instance
+        for ptable in prepare_tables:
```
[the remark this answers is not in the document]
sh-rp: it seems to be working, i have added a test for this.

dlt/destinations/sql_client.py
```diff
@@ -299,18 +321,42 @@ def df(self, chunk_size: int = None, **kwargs: Any) -> Optional[DataFrame]:
...
-            else:
-                return df
+
+        # if no chunk size, fetch all
+        if not chunk_size:
+            yield _wrap_result(self.fetchall(), columns)
+            return
+
+        # otherwise iterate over results in batch size chunks
+        for result in self.iter_fetchmany(chunk_size=chunk_size):
```
sh-rp, line 351: I'm not quite sure if this part is right. The tests pass but I am creating a dataframe with a pyarrow backend so the function below is more efficient, I am not sure how it is done in the rest of the codebase though. are we using numpy backends or pyarrow backends there?

rudolfix:
IMO the interface is OK. here is feedback for the code. it is still too early for a regular review:

General
1. move all dataset-related code to `dlt.destinations`. try to remove as many inner imports as possible
2. do not pass `Pipeline` anywhere. We can implement everything with simpler object

Type hierarchy
I think what you did can be really simplified

Dataset:
1. we do not need to take `Pipeline` here. What we need is to pass destination reference, dataset name and Schema(s) that are in the pipeline so we can create right destination client. some code that creates initial configuration may be moved from pipeline to dlt.destinations namespace but that is even better
2. We should not need any list of prepared tables. AFAIK only `filesystem` needs that right? see below.
3. I like that `Dataset` is fully lazy. we must keep it like that.
4. Create an instance of destination client right away. It will keep all the context we need and also cache this in-memory database for filesystem

Relation:
We do not need this class at all. OR drop any references to clients there and make it implement SupportsDataAccess, merge it with DBAPICursor and return it from `execute_query` in SqlClient. You do not need `SupportsRelationshipAccess` which does exactly the same.
Relation of the form above should be very lightweight and do the same thing that cursor does

Filesystem:
implement SqlClient for filesystem:
- derive from duckdb sql_client
- re-implement `execute_query` so we create views as needed by parsing SQL. here's oneliner with sqlglot. https://stackoverflow.com/questions/69684115/python-library-for-extracting-table-names-from-from-clause-in-sql-statetments OFC create only tables that you know: so those that are in the Schema associated with destination client (there's always one)
- let's start with read only mode (views). we can add write mode ie. as a flag in the Dataset later
- we can IMO drop jsonl support but if not, use the `columns` hint on the `read_json` to pass the exact schema that we have in the Schema for given table. in your current implementation autodetect will be used and schema will be different.

in the future we can implement second SqlClient for data fusion so the engine can be replaced (not only duckdb).
WDYT?

composable_pipeline_1.py
jorritsandbrink: I suggest these adjustments:

- `df()` -> `pandas()`
- `iter_df` -> `iter_pandas()`

Because:
- there are other dataframe flavors besides Pandas, e.g. Polars and Ibis
- it's consistent with `arrow()` and `ibis()` (they are also project names)

dlt/destinations/sql_client.py
```diff
@@ -299,18 +321,48 @@ def df(self, chunk_size: int = None, **kwargs: Any) -> Optional[DataFrame]:
...
+            # TODO: ensure that this is arrow backed
+            yield _wrap_result(result, columns, dtype_backend="pyarrow")
+
+    def iter_arrow(self, chunk_size: int) -> Generator[ArrowTable, None, None]:
+        """Default implementation converts df to arrow"""
+        for df in self.iter_df(chunk_size=chunk_size):
+            # TODO: is this efficient?
```
jorritsandbrink, line 364: We're piping everything through Pandas here. Pandas is known for being slow. Perhaps these are faster:

- https://docs.pola.rs/api/python/stable/reference/api/polars.read_database.html
- https://docs.pola.rs/api/python/stable/reference/api/polars.read_database_uri.html#polars.read_database_uri

They return a Polars dataframe from a database connection. Polars is Arrow-backed. The Polars dataframe can be turned into an Arrow table with https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_arrow.html.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -590,3 +607,74 @@ def create_table_chain_completed_followup_jobs(
...
+                db.sql(create_table_sql_gzipped)
+
+        return db
+
+    @contextmanager
+    def get_readable_relation(
+        self, *, table: str = None, sql: str = None
+    ) -> Generator[DBApiCursor, Any, Any]:
+        from dlt.destinations.impl.duckdb.sql_client import DuckDBDBApiCursorImpl
+        import sqlglot
```
sh-rp, line 665: I assume this should go in some extra, or do we just ask the user to install sqlglot if not present?

sh-rp:
@rudolfix you can have another look at the file and class structure.

- I would say I have worked out the placement and naming of the interfaces/protocols now. 
- There are no inner imports left (except for sqlglot), I also put all the stuff in the destination folder
- the instantiation of the dataset is now done in the destination client, i think this makes sense, but let me know if you think otherwise.
- I have not implemented the sql_client for the filesystem, the SQLClientBase does a lot of stuff next to plainly executing queries. I think I'd leave it the way it is for now, the read access is completely abstracted with these new Protocols anyway.
- Maybe there is a smart way to forward method calls from ReadableRelation to the cursor, I have not found it yet ;)
- How should we handle sqlglot? As an extra or just ask the user to install sqlglot when he accesses the filesystem dataset?

Also one other question, what do you think about using polars for creating arrow tables or exposing polars?

sh-rp:
Per our discussion: Maybe we can also add a `source` function on the ReadableDataset and a `resource` function on the ReadableRelation which will create a DltSource and a DltResource respectively which also contains the schema if used.

dlt/destinations/sql_client.py
rudolfix: I agree. We already code doing the above efficiently in `sql_database` and we'll use it here. There's once "complication" though:
1. if the relation is a table, we have a correct schema and we can create properly typed arrow table. that also applies to pandas above
2. if not we must infer types from the data
`sql_database` code handles the above. Again that should be a next ticket. Somehow we should be able to attach Relation schema

dlt/destinations/dataset.py
```diff
@@ -0,0 +1,104 @@
...
+
+from dlt.common.typing import DataFrame, ArrowTable
+
+
+class ReadableRelation(SupportsReadableRelation):
+    def __init__(
+        self,
+        *,
+        client: WithReadableRelations,
+        table: str = None,
```
rudolfix, line 18: we also should know the relation schema here. you'll need it to form correctly typed frames and arrow tables

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -590,3 +607,74 @@ def create_table_chain_completed_followup_jobs(
...
+        import sqlglot.expressions as exp
+
+        if table:
+            sql = f"SELECT * FROM {table}"
+
+        # find all tables to preload
+        expression = sqlglot.parse_one(sql, read="oracle")
```
rudolfix, line 672: there's duckdb dialect. you can parse it here

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -590,3 +607,74 @@ def create_table_chain_completed_followup_jobs(
...
+        load_tables = [t.name for t in expression.find_all(exp.Table)]
```
rudolfix, line 673: only tables that are found in the schema associated with the client should be mapped

dlt/destinations/impl/filesystem/filesystem.py
rudolfix: I still do not understand why you'll not derive from `DuckdbSqlClient` and put all this code into `execute_query`? why this is good:
1. we do not refer to duckdb here. we'll be able to swap engines by swapping sql clients
2. you can move sqlglot there
3. you can rewrite `get_duckdb` to just create required relations
4. then you just call super() on `execute_query` to get the right cursor

rudolfix: added bonus: filesystem now implements `sql_client` that works on every bucket and our streamlit app magically works

dlt/common/destination/reference.py
```diff
@@ -399,6 +408,50 @@ def create_followup_jobs(self, final_state: TLoadJobState) -> List[FollowupJob]:
...
+
+    def iter_fetchmany(self, chunk_size: int) -> Generator[List[Tuple[Any, ...]], Any, Any]: ...
+
+    def fetchone(self) -> Optional[Tuple[Any, ...]]: ...
+
+
+class SupportsReadableDataset(Protocol):
+    """A readable dataset retrieved from a destination, has support for creating readable relations for a query or table"""
+
+    def sql(self, sql: str) -> SupportsReadableRelation: ...
```
rudolfix, line 448: you shoudl accept dbapi params. WDYT?

dlt/common/destination/reference.py
```diff
@@ -561,6 +617,18 @@ def should_truncate_table_before_load_on_staging_destination(self, table: TTable
         return True
 
 
+class WithReadableRelations(ABC):
+    """Add support for getting readable reletions form a destination"""
+
+    @abstractmethod
+    def get_readable_relation(
```
rudolfix, line 624: I think you should split into `table_relation` and `query_relation`

dlt/common/destination/reference.py
rudolfix: maybe `query` is a better name? hmmm.. practically this will be sql

dlt/destinations/sql_client.py
rudolfix:
```
@staticmethod
    def _query_iterator(
        result,
        exit_stack: ExitStack,
        chunksize: int,
        columns,
        index_col=None,
        coerce_float: bool = True,
        parse_dates=None,
        dtype: DtypeArg | None = None,
        dtype_backend: DtypeBackend | Literal["numpy"] = "numpy",
    ):
        """Return generator through chunked result set"""
        has_read_data = False
        with exit_stack:
            while True:
                data = result.fetchmany(chunksize)
                if not data:
                    if not has_read_data:
                        yield _wrap_result(
                            [],
                            columns,
                            index_col=index_col,
                            coerce_float=coerce_float,
                            parse_dates=parse_dates,
                            dtype=dtype,
                            dtype_backend=dtype_backend,
                        )
                    break

                has_read_data = True
                yield _wrap_result(
                    data,
                    columns,
                    index_col=index_col,
                    coerce_float=coerce_float,
                    parse_dates=parse_dates,
                    dtype=dtype,
                    dtype_backend=dtype_backend,
                )
```

our biggest challenge is to provide `dtype` in case of tables

rudolfix (commented):
on top of the review: lets implement as many native arrow/df processing as possible
1. add native arrow and df to snowflake. they have really good interface now! ask @rahuljo for details
2. clickhouse: https://clickhouse.com/docs/en/integrations/python#consuming-query-results-with-numpy-pandas-or-arrow

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: yes, this works now :)

composable_pipeline_1.py
```diff
@@ -0,0 +1,109 @@
+"""Example of a composable pipeline"""
+
+import dlt
+import os
+import random
+from dlt.destinations import filesystem, duckdb
+
+# fixtures
+customers = [
```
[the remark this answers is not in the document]
sh-rp: this is an example file so adrian and others can review how to use tihs interface, it will go when I finish this

sh-rp:
@rudolfix I have an architectural question: As you can see from my latest commit, I am not creating an arrow table schema from our dlt schema if there is no native arrow support on the cursor. For this the table columns are pushed right down into the cursor where we will eventually also need the destination capabilities of another destination if we pipe from one dataset into the next pipeline. So I'm thinking we should probably implement this stuff in the ReadableRelations class and only forward the desired arrow schema to the cursor, or even contruct the arrow table in the ReadableRelations if no native support is present on the cursor in question. Wdyt?

dlt/common/destination/reference.py
```diff
@@ -399,6 +408,50 @@ def create_followup_jobs(self, final_state: TLoadJobState) -> List[FollowupJob]:
...
+        """
+        ...
+
+    def arrow(self, chunk_size: int = None) -> Optional[ArrowTable]: ...
+
+    def iter_df(self, chunk_size: int) -> Generator[DataFrame, None, None]: ...
+
+    def iter_arrow(self, chunk_size: int) -> Generator[ArrowTable, None, None]: ...
+
+    def fetchall(self) -> List[Tuple[Any, ...]]: ...
```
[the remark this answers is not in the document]
sh-rp: You are talking only about the fetch methods? or the arrow and df methods too?

dlt/common/destination/reference.py
sh-rp: this is done

dlt/destinations/dataset.py
sh-rp: this is now implemented to querying tables

dlt/common/destination/reference.py
```diff
@@ -55,6 +69,16 @@
 
 DEFAULT_FILE_LAYOUT = "{table_name}/{load_id}.{file_id}.{ext}"
 
+try:
+    from dlt.common.libs.pandas import DataFrame
```
sh-rp, line 73: I need these in the reference for the DataAccess classes. I could also type those as any, but It would be nicer if the user had fully typed results here.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -235,6 +256,21 @@ def __init__(
         # cannot be replaced and we cannot initialize folders consistently
         self.table_prefix_layout = path_utils.get_table_prefix_layout(config.layout)
         self.dataset_name = self.config.normalize_dataset_name(self.schema)
+        self._sql_client: SqlClientBase[Any] = None
+
+    @property
+    def sql_client(self) -> SqlClientBase[Any]:
+        # we use an inner import here, since the sql client depends on duckdb and will
+        # only be used for read access on data, some users will not need the dependency
+        from dlt.destinations.impl.filesystem.sql_client import FilesystemSqlClient
```
sh-rp, line 265: this is "plus" area..

dlt/destinations/sql_client.py
sh-rp: I am not really sure what you mean with the code snippet above. I can't find it in the code.

tests/load/test_read_interfaces.py
```diff
@@ -0,0 +1,218 @@
...
+    # check iterfetchmany
+    chunks = list(relationship.iter_fetchmany(chunk_size=chunk_size))
+    assert [len(chunk) for chunk in chunks] == expected_chunk_counts
+    ids = reduce(lambda a, b: a + b, [[item[0] for item in chunk] for chunk in chunks])
+    assert set(ids) == set(range(total_records))
+
+    # simple check that query also works
+    relationship = pipeline.dataset().query("select * from items where id < 20")
+
+    # we selected the first 20
```
sh-rp, line 136: will need better tests

tests/load/test_sql_client.py
```diff
@@ -309,57 +309,6 @@ def test_execute_query(client: SqlJobClientBase) -> None:
             assert len(rows) == 0
 
 
-@pytest.mark.parametrize(
-    "client", destinations_configs(default_sql_configs=True), indirect=True, ids=lambda x: x.name
-)
-def test_execute_df(client: SqlJobClientBase) -> None:
```
sh-rp, line 315: this is now tested with the data accessing

dlt/common/destination/reference.py
sh-rp: @jorritsandbrink there are some destinations that have native pandas support and I think it would be cool to be able to expose those directly for the user

composable_pipeline_1.py
sh-rp: makes sense to me, @rudolfix wdyt?

dlt/destinations/sql_client.py
sh-rp: This is implemented now, we have the same code that the sql database uses to create arrow tables from cursor results, in the case of querying full tables we can use the known dlt schema for proper typing, for all results from calling "query" the types are autodiscovered. We could take this further by:
* Somehow using sql glot to figure out which column came from which table
* Allowing filtering, sorting and selecting a subste of columns of a given table where we can then easily use the known schema. Maybe this could be a next step.

dlt/common/destination/reference.py
```diff
@@ -561,6 +570,53 @@ def should_truncate_table_before_load_on_staging_destination(self, table: TTable
...
+        """
+        ...
+
+    def arrow(self, *, chunk_size: int = None) -> Optional[ArrowTable]: ...
+
+    def iter_df(self, chunk_size: int) -> Generator[DataFrame, None, None]: ...
+
+    def iter_arrow(self, chunk_size: int) -> Generator[ArrowTable, None, None]: ...
+
+    def fetchall(self) -> List[Tuple[Any, ...]]: ...
```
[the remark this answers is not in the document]
sh-rp: I think I would keep it this way. The user may use the cursor too if they like, but this possiblity to iterate without having to open your own context manager is quite nice from the user perspective I think.

dlt/dataset.py
```diff
@@ -0,0 +1,128 @@
+from typing import cast, Any, TYPE_CHECKING, Generator, List, Tuple, Optional
```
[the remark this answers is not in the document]
sh-rp: this is moved

dlt/dataset.py
```diff
@@ -0,0 +1,128 @@
...
+
+from dlt.common.typing import DataFrame, ArrowTable
+
+
+class Relation:
+    def __init__(
+        self, *, pipeline: Any, table: str = None, sql: str = None, prepare_tables: List[str] = None
+    ) -> None:
+        """Create a lazy evaluated relation to for the dataset of a pipeline"""
+        from dlt.pipeline import Pipeline
```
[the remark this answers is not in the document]
sh-rp: I changed the architector so that this is not needed anymore

dlt/dataset.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.pipeline: Pipeline = cast(Pipeline, pipeline)
+        self.prepare_tables = prepare_tables
+        self.sql = sql
+        self.table = table
+
+    @contextmanager
+    def _client(self) -> Generator[SupportsRelationshipAccess, Any, Any]:
+        from dlt.destinations.job_client_impl import SqlJobClientBase
```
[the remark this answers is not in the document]
sh-rp: done

dlt/common/typing.py
```diff
@@ -77,6 +77,22 @@
     REPattern = _REPattern
     PathLike = os.PathLike
 
+
+try:
```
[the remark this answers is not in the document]
sh-rp: I moved it around a bit. We have the dataaccess stuff in the destination, and imho it would be nice to have the correct typings there.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: done

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: done

dlt/common/destination/reference.py
sh-rp: done

dlt/destinations/sql_client.py
```diff
@@ -316,18 +341,54 @@ def df(self, chunk_size: int = None, **kwargs: Any) -> Optional[DataFrame]:
...
+            # NOTE: we go via arrow table, types are created for arrow is columns are known
+            # https://github.com/apache/arrow/issues/38644 for reference on types_mapper
+            yield table.to_pandas(types_mapper=pd.ArrowDtype)
+
+    def iter_arrow(self, chunk_size: int) -> Generator[ArrowTable, None, None]:
+        """Default implementation converts query result to arrow table"""
+        from dlt.common.libs.pyarrow import row_tuples_to_arrow
+
+        # if loading to a specific pipeline, it would be nice to have the correct caps here
+        caps = DestinationCapabilitiesContext.generic_capabilities()
```
rudolfix, line 380: all `sql_client` and `clients` have caps. this must be present at this point if we create dataset out of the `pipeline`. so assume you have caps in `init` of the cursor. but if it is None, you use generic caps.

also note that caps are often dynamic ie. filesystem + delta generates caps

dlt/destinations/sql_client.py
```diff
@@ -292,11 +302,26 @@ def _truncate_table_sql(self, qualified_table_name: str) -> str:
...
+    ) -> None:
+        pass
+
+
 class DBApiCursorImpl(DBApiCursor):
     """A DBApi Cursor wrapper with dataframes reading functionality"""
 
     def __init__(self, curr: DBApiCursor) -> None:
         self.native_cursor = curr
+        self.columns = {}
```
rudolfix, line 324: where columns are coming from? they look they are always empty.
1. ideally we know the columns in a advance (from table or via sqlglot - which I doubt will work)
2. or we set the columns after first row/chunk is available by reflecting back from data: we have code for python object (in Schema - infer column, in arrow obviously, and we lack similar things for pandas but that's trivial to do)

dlt/destinations/sql_client.py
```diff
@@ -316,18 +341,54 @@ def df(self, chunk_size: int = None, **kwargs: Any) -> Optional[DataFrame]:
...
+        while True:
+            if not (result := self.fetchmany(chunk_size)):
+                return
+            yield result
+
+    def iter_df(self, chunk_size: int) -> Generator[DataFrame, None, None]:
+        """Default implementation converts arrow to df"""
+        from dlt.common.libs.pandas import pandas as pd
+
+        for table in self.iter_arrow(chunk_size=chunk_size):
```
rudolfix, line 370: we need a followup PR to implement `row_tuples_to_pandas` that will behave similarly to the arrow one but without using arrow.

dlt/destinations/sql_client.py
rudolfix: this is how pandas converts row tuples into panda frames. we do it OK in sql_database code so you can close it

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,94 @@
...
+from dlt.destinations.impl.duckdb.sql_client import DuckDbSqlClient
+from dlt.destinations.impl.duckdb.factory import duckdb as duckdb_factory
+
+
+class FilesystemSqlClient(DuckDbSqlClient):
+    def __init__(self, fs_client: FSClientBase, protocol: str) -> None:
+        """For now we do all operations in the memory dataset"""
+        """TODO: is this ok?"""
+        super().__init__(
+            dataset_name="memory",
```
rudolfix, line 26: we need exactly the same dataset name that was used in the pipeline and was passed to JobClient

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,94 @@
...
+                self._conn.execute(create_table_sql_base)
+            except (duckdb.InvalidInputException, duckdb.IOException):
+                # try to load non gzipped files
+                self._conn.execute(create_table_sql_gzipped)
+
+    @contextmanager
+    @raise_database_error
+    def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DBApiCursor]:
+        # find all tables to preload
+        expression = sqlglot.parse_one(query, read="duckdb")  # type: ignore
```
rudolfix, line 80: we can add sqlglot to filesystem deps

dlt/common/destination/reference.py
```diff
@@ -441,6 +465,65 @@ def create_followup_jobs(self, final_state: TLoadJobState) -> List[FollowupJobRe
...
+
+class DBApiCursor(SupportsReadableRelation):
+    """Protocol for DBAPI cursor"""
+
+    description: Tuple[Any, ...]
+
+    native_cursor: "DBApiCursor"
+    """Cursor implementation native to current destination"""
+
+    columns: TTableSchemaColumns
```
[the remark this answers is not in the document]
sh-rp: I moved it to the readablerelation interface for now

dlt/destinations/sql_client.py
sh-rp: As far as I understand these capabilities are needed to create an arrow table that can be properly loaded into the destination that it will go. Since this is a read interface, it does not make sense to use the capabilities of the source database. If we know where the data is going (from one pipeline into the next one) the capabilities of the destination pipeline should be known here and used for creating the arrow tables, this is what I wrote in the comment above. Or is there some benefit of using the ones of the datasource?

dlt/destinations/sql_client.py
sh-rp: In case where we do not know the schema (sql queries) they are set to this:
`{c: {"name": c, "nullable": True} for c in self._get_columns()}`

In case where we know the schema they are set to the tablecolumns of the table we are querying.

dlt/destinations/impl/filesystem/sql_client.py
sh-rp: this is done now!

dlt/destinations/impl/filesystem/sql_client.py
sh-rp: done

tests/load/test_read_interfaces.py
sh-rp: actually probably not, this simple test should be enough to check that sql queries are passed on.

dlt/common/destination/reference.py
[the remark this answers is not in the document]
sh-rp: fetchone and fetchmany are default dbapi cursors functions though, so I would keep those, no?

dlt/destinations/dataset.py
```diff
@@ -0,0 +1,96 @@
...
+        client: SqlClientBase[Any],
+        query: str,
+        schema_columns: TTableSchemaColumns = None,
+    ) -> None:
+        """Create a lazy evaluated relation to for the dataset of a destination"""
+        self.client = client
+        self.schema_columns = schema_columns
+        self.query = query
+
+        # wire protocol functions
```
sh-rp, line 27: I couldn't do much better than this, if you have a nicer implementation of this that still keeps the same interface from the outside, let me know.

dlt/destinations/sql_client.py
sh-rp: I added this to the folloup ticket

tests/load/test_read_interfaces.py
```diff
@@ -0,0 +1,247 @@
...
+
+    # simple check that query also works
+    tname = pipeline.sql_client().make_qualified_table_name("items")
+    query_relationship = pipeline.dataset().query(f"select * from {tname} where id < 20")
+
+    # we selected the first 20
+    table = query_relationship.arrow()
+    assert table.num_rows == 20
+
+    # check that hints are carried over to arrow table
```
sh-rp, line 153: I think testing one column should be enough to test wether hints are carried over, we are testing all datatypes for all destinations in a few other locations

tests/load/test_read_interfaces.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+@pytest.mark.essential
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(
+        local_filesystem_configs=True,
+        all_buckets_filesystem_configs=True,
+        bucket_exclude=[AZ_BUCKET, ABFS_BUCKET, SFTP_BUCKET],
+    ),  # TODO: make AZ work
```
sh-rp, line 190: az and sftp support are moved to the followup ticket

tests/load/pipeline/test_restore_state.py
```diff
@@ -674,6 +674,10 @@ def some_data(param: str) -> Any:
     # nevertheless this is potentially dangerous situation 🤷
     assert ra_production_p.state == prod_state
 
+    # for now skip sql client tests for filesystem
```
sh-rp, line 677: this can be commented out when sftp and az sql_client support is there

sh-rp:
@rudolfix if you like you can have another look. I think all your comment are adressed. Open questions for me still are:
* I don't think we should be using the capabilities of the source destination when reading, rather it should be possible to pass in the capabilities of given destination pipeline if we know where the data is going
* I was not quite sure what you mean with this dataframe document. Do you want to change anything in the place where we emit dataframes (which is done in the standard arrow way now and can be improved in followup work)
* There were still some questions around the methods we expose on the readable dataset, I put comments there too.

tests/load/test_read_interfaces.py
```diff
@@ -0,0 +1,247 @@
...
+    assert len(many) == chunk_size
+
+    # check iterfetchmany
+    chunks = list(table_relationship.iter_fetchmany(chunk_size=chunk_size))
+    assert [len(chunk) for chunk in chunks] == expected_chunk_counts
+    ids = reduce(lambda a, b: a + b, [[item[0] for item in chunk] for chunk in chunks])
+    assert set(ids) == set(range(total_records))
+
+    # simple check that query also works
+    tname = pipeline.sql_client().make_qualified_table_name("items")
```
rudolfix, line 146: both qualified and unqualified table name should work here

rudolfix (changes requested):
it looks like only cosmetic changes are left. please see all open comments (I closed solved).

on top of that:
1. what exactly does not work with AZ? seems duckdb supports it very well
2. a few fundamental test cases are missing
- make a query with a join of known tables
- write join to a new table and then select from it
- try to update a mapped table (view) to see what happens and maybe provide nice exception message

followup ticket:
- support views on delta and iceberg tables (ie. if we open staging destination from athena or dremio)

dlt/common/destination/reference.py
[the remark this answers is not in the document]
sh-rp: ah now i get it, you wanted me to rename iter_fetchmany to iter_fetch, right? yeah I did that now :)

dlt/common/destination/reference.py
```diff
@@ -442,6 +465,65 @@ def create_followup_jobs(self, final_state: TLoadJobState) -> List[FollowupJobRe
         return []
 
 
+class SupportsReadableRelation(Protocol):
+    """A readable relation retrieved from a destination that supports it"""
+
+    schema_columns: TTableSchemaColumns
+    """Known dlt table columns for this relation"""
+
+    def pandas(self, chunk_size: int = None) -> Optional[DataFrame]:
```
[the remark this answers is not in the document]
sh-rp: done

dlt/destinations/dataset.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.iter_pandas = self._wrap_iter("iter_pandas")  # type: ignore
+        self.iter_arrow = self._wrap_iter("iter_arrow")  # type: ignore
+        self.iter_fetchmany = self._wrap_iter("iter_fetchmany")  # type: ignore
+
+    @contextmanager
+    def cursor(self) -> Generator[SupportsReadableRelation, Any, Any]:
+        """Gets a DBApiCursor for the current relation"""
+        with self.client as client:
+            # this hacky code is needed for mssql to disable autocommit, read iterators
```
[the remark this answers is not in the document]
sh-rp: the problem arises when keeping the connection open and iterating over the result, yes. I don't think we have any tests that do this right now on devel.

dlt/destinations/sql_client.py
sh-rp: I added this plus a fallback to the defaults

dlt/common/destination/reference.py
sh-rp: Ok. I'm assuming i still need to catch the MissingDependencyException exceptions in case a user uses mypy on a project and does not need these libs, right?

dlt/common/libs/pyarrow.py
```diff
@@ -31,6 +33,7 @@
     import pyarrow.compute
     import pyarrow.dataset
     from pyarrow.parquet import ParquetFile
+    from pyarrow import Table as ArrowTable
```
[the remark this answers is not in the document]
sh-rp: done

dlt/pipeline/pipeline.py
```diff
@@ -1707,3 +1716,9 @@ def _save_state(self, state: TPipelineState) -> None:
     def __getstate__(self) -> Any:
         # pickle only the SupportsPipeline protocol fields
         return {"pipeline_name": self.pipeline_name}
+
+    def dataset(self, dataset_type: Literal["dbapi", "ibis"] = "dbapi") -> SupportsReadableDataset:
```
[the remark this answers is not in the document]
sh-rp: done

tests/load/test_read_interfaces.py
sh-rp: I added a test for this

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,123 @@
...
+        self.is_local_filesystem = protocol == "file"
+
+        if protocol not in SUPPORTED_PROTOCOLS:
+            raise NotImplementedError(
+                f"Protocol {protocol} currently not supported for FilesystemSqlClient. Supported"
+                f" protocols are {SUPPORTED_PROTOCOLS}."
+            )
+
+        # set up duckdb instance
+        self._conn = duckdb.connect(":memory:")
```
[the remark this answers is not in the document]
sh-rp: I did this now, it will retain the in memory version of the database and fall back to the mechanism in the parent class when using an externally passed in database connection

dlt/common/destination/reference.py
```diff
@@ -442,6 +465,65 @@ def create_followup_jobs(self, final_state: TLoadJobState) -> List[FollowupJobRe
...
+    """Cursor implementation native to current destination"""
+
+    def execute(self, query: AnyStr, *args: Any, **kwargs: Any) -> None: ...
+    def close(self) -> None: ...
+
+
+class SupportsReadableDataset(Protocol):
+    """A readable dataset retrieved from a destination, has support for creating readable relations for a query or table"""
+
+    def query(self, query: str) -> SupportsReadableRelation: ...
```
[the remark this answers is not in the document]
sh-rp: I did both

pyproject.toml
```diff
@@ -86,6 +86,8 @@ deltalake = { version = ">=0.19.0", optional = true }
 sqlalchemy = { version = ">=1.4", optional = true }
 alembic = {version = "^1.13.2", optional = true}
 paramiko = {version = ">=3.3.0", optional = true}
+sqlglot = "^25.9.0"
+db-dtypes = { version = ">=1.2.0", optional = true }
```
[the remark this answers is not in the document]
sh-rp: I think this came from changing arrowtables to dataframes, I removed it and will have to see wether it works now without

pyproject.toml
```diff
@@ -86,6 +86,8 @@ deltalake = { version = ">=0.19.0", optional = true }
 sqlalchemy = { version = ">=1.4", optional = true }
 alembic = {version = "^1.13.2", optional = true}
 paramiko = {version = ">=3.3.0", optional = true}
+sqlglot = "^25.9.0"
```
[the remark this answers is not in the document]
sh-rp: done

tests/load/test_read_interfaces.py
sh-rp: az is now working, sftp needs a deeper investigation

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,123 @@
...
+        self._conn.sql(f"CREATE SCHEMA {self.dataset_name}")
+        self._conn.sql(f"USE {self.dataset_name}")
+        self._conn.register_filesystem(self.fs_client.fs_client)
```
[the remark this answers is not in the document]
sh-rp: I am using the native implementation where possible now

tests/load/test_read_interfaces.py
```diff
@@ -314,7 +324,44 @@ def test_read_interfaces_filesystem(destination_config: DestinationTestConfigura
...
+)
+def test_delta_tables(destination_config: DestinationTestConfiguration) -> None:
+    os.environ["DATA_WRITER__FILE_MAX_ITEMS"] = "700"
+
+    pipeline = destination_config.setup_pipeline(
+        "read_pipeline",
+        dataset_name="read_test",
+    )
+
+    # in case of gcs we use the s3 compat layer for reading
```
sh-rp, line 347: this is a bit bad tbh. writing delta tables only works with gcs native, reading them in duckdb requrires the s3 api. this should be investigated at some point by @jorritsandbrink I think.

tests/load/test_read_interfaces.py
```diff
@@ -218,30 +221,48 @@ def double_items():
...
+        # we create a duckdb with a table an see wether we can add more views
+        duck_db_location = "_storage/" + uniq_id()
+        external_db = duckdb.connect(duck_db_location)
+        external_db.execute("CREATE SCHEMA first;")
+        external_db.execute("CREATE SCHEMA second;")
+        external_db.execute("CREATE TABLE first.items AS SELECT i FROM range(0, 3) t(i)")
+        assert len(external_db.sql("SELECT * FROM first.items").fetchall()) == 3
 
         # now we can use the filesystemsql client to create the needed views
+        fs_client: Any = pipeline.destination_client()
```
sh-rp, line 243: this interface is a bit clunky at the moment. we should probably have a unified interface that enables us to mount views into duckdb from any destination that supports it, if you have an idea of where it could live and what the signature should be you can let me know. for the first little while it should be ok, since we are using inner classes here. the secrets are actually also not stored in the database file but in a global directory for duckdb, so I've built it in a way that whoever controls the duckdb connection from the outside can also create secrets with a given key and is then able to delete them again.

dlt/common/destination/reference.py
rudolfix: Note: using Dataset possible allows us to "mount" any external table as a cursor that reads data so we do not need to download full table locally. we may be able to ie. mount snowflake or bigquery tables are duckdb tables

tests/load/test_read_interfaces.py
```diff
@@ -0,0 +1,405 @@
...
+
+    # special filesystem sql checks
+    if destination_config.destination_type == "filesystem":
+        import duckdb
+        from dlt.destinations.impl.filesystem.sql_client import FilesystemSqlClient
+
+        # check we can create new tables from the views
+        with pipeline.sql_client() as c:
+            c.create_views_for_tables({"items": "items", "double_items": "double_items"})
+            c.execute_sql(
```
rudolfix, line 219: why do you need to register views explicitly? there's a subquery referencing those tables so you should be able to parse them out with sqlglot no? (I just checked and you get list of all tables from this expression so this line is not needed)

tests/load/test_read_interfaces.py
```diff
@@ -0,0 +1,405 @@
...
+                assert list(joined_table[10]) == [10, 20]
+
+            # inserting values into a view should fail gracefully
+            try:
+                c.execute_sql("INSERT INTO double_items VALUES (1, 2)")
+            except Exception as exc:
+                assert "double_items is not an table" in str(exc)
+
+        # we create a duckdb with a table an see wether we can add more views
+        duck_db_location = "_storage/" + uniq_id()
```
rudolfix, line 237: use TEST_STORAGE constant....

tests/load/test_read_interfaces.py
rudolfix: the interface is already there. it is called `update_stored_schema` on job client. in this case `filesystem` is a staging destination for `duckdb` and you create all tables as views, pass the credentials etc. maybe we need some convenience method that creates `dataset` instance out of such strucutre (so we take not just destination but also staging as input to dataset).

the nice thing is that all this duckdb related code that creates views and does permission handover could go to duckdb client.

https://github.com/dlt-hub/dlt/issues/1692

this is a good followup ticket

dlt/pipeline/pipeline.py
```diff
@@ -1707,3 +1716,9 @@ def _save_state(self, state: TPipelineState) -> None:
     def __getstate__(self) -> Any:
         # pickle only the SupportsPipeline protocol fields
         return {"pipeline_name": self.pipeline_name}
+
+    def _dataset(self, dataset_type: Literal["dbapi", "ibis"] = "dbapi") -> SupportsReadableDataset:
```
rudolfix, line 1720: my idea to pass external connection to dataset would be to accept `SqlClientBase` as `dataset_type` in that case we'd create `ReadableDBAPIDataset` with it,

however we can do it directly by getting sql_client form the pipeline already.

rudolfix (changes requested):

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def open_connection(self) -> duckdb.DuckDBPyConnection:
+        # we keep the in memory instance around, so if this prop is set, return it
+        if not self._conn:
+            super().open_connection()
+
+            # set up connection and dataset
+            self._existing_views: List[str] = []  # remember which views already where created
+
+            self.autocreate_required_views = False
```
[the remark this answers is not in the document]
sh-rp: there is some bug in sqlglot that throws an internal exception when parsing the information schema query, and I think for now this method is better than catching all exceptions and ignore them in the sqlglot parsing.

tests/load/test_read_interfaces.py
sh-rp: done

tests/load/test_read_interfaces.py
sh-rp: I remove it; i think this was a leftover from another query.

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+else:
+    FilesystemClient = Any
+
+
+class FilesystemSqlClient(DuckDbSqlClient):
+    def __init__(
+        self,
+        fs_client: FilesystemClient,
+        dataset_name: str = None,
+        duckdb_connection: duckdb.DuckDBPyConnection = None,
```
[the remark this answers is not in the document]
sh-rp: I changed it so you can now hand over a credentials object to the filesystem duckdb client. I find it less convenient because you need to wrap the connection in a credentials object, but I suppose it is better like that.

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+    ) -> None:
+        super().__init__(
+            dataset_name=dataset_name or fs_client.dataset_name,
+            staging_dataset_name=None,
+            credentials=DuckDbCredentials(duckdb_connection or ":memory:"),
+            capabilities=duckdb_factory()._raw_capabilities(),
+        )
+        self.fs_client = fs_client
+        self.using_external_database = duckdb_connection is not None
```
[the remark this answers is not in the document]
sh-rp: I removed it

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+        if persistent:
+            persistent_stmt = " PERSISTENT "
+
+        # abfss buckets have an @ compontent
+        scope = self.fs_client.config.bucket_url
+        if "@" in scope:
+            scope = scope.split("@")[0]
+
+        # add secrets required for creating views
+        if self.fs_client.config.protocol == "s3":
```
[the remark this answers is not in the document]
sh-rp: done, although I'm not quite sure what it means

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+            # create authentication to data provider
+            self.create_authentication()
+
+        # the line below solves problems with certificate path lookup on linux
+        # see duckdb docs
+        self._conn.sql("SET azure_transport_option_type = 'curl';")
+
+        return self._conn
+
+    def close_connection(self) -> None:
```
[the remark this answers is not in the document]
sh-rp: I removed it

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+            view_name = self.make_qualified_table_name(view_name)
+            create_table_sql_base = f"CREATE VIEW {view_name} AS SELECT * FROM {from_statement}"
+            self._conn.execute(create_table_sql_base)
+
+    @contextmanager
+    @raise_database_error
+    def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DBApiCursor]:
+        # find all tables to preload
+        if self.autocreate_required_views:  # skip this step when operating on the schema..
+            expression = sqlglot.parse_one(query, read="duckdb")  # type: ignore
```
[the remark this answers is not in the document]
sh-rp: there is a bug in sqlglot, that's why I need to disbale it for certain queries, see above

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+                self.create_dataset()
+            self.autocreate_required_views = True
+            self._conn.sql(f"USE {self.dataset_name}")
+
+            # create authentication to data provider
+            self.create_authentication()
+
+        # the line below solves problems with certificate path lookup on linux
+        # see duckdb docs
+        self._conn.sql("SET azure_transport_option_type = 'curl';")
```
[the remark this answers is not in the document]
sh-rp: I moved it up

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+            )
+            self._conn.register_filesystem(self.fs_client.fs_client)
+
+        # for memory we also need to register filesystem
+        elif self.fs_client.config.protocol == "memory":
+            self._conn.register_filesystem(self.fs_client.fs_client)
+
+    def open_connection(self) -> duckdb.DuckDBPyConnection:
+        # we keep the in memory instance around, so if this prop is set, return it
+        if not self._conn:
```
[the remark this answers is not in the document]
sh-rp: I've changed it a bit to keep the in memory connection in this class and it can be treated as an external database in the credentials, this way the borrowing will work just the same way it always does and the instance should go away when this class gets garbarge collected.

dlt/pipeline/pipeline.py
sh-rp: ok, let's do this in the next iteration, then you can explain what you want to do exactly

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -0,0 +1,250 @@
...
+            return super().close_connection()
+
+    @raise_database_error
+    def create_views_for_tables(self, tables: Dict[str, str]) -> None:
+        """Add the required tables as views to the duckdb in memory instance"""
+
+        # create all tables in duck instance
+        for table_name in tables.keys():
+            view_name = tables[table_name]
+            if view_name in self._existing_views:
```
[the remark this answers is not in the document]
sh-rp: yes, good point, I changed it to only create views if there are no tables or views of the same name present yet

dlt/destinations/impl/filesystem/sql_client.py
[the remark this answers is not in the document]
rudolfix: we should tell sqlglot to fix that IMO :) maybe other queries are also affected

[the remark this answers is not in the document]
rudolfix: sqlglot was complaining that it cannot parse parametrized queries, which is fair... this was also a good opportunity to fix things so I'm skipping parsing in this case now

rudolfix (changes requested):
code is good now but we miss a few tests specific to sql_client for filesystem:

1. make sure you can "open_connection"  several times on sql_client
2. pass external connection ie. to duckdb that is persisted and try to create a few views then use it from existing connection (or after reopen of the persistent db)
3. test if we skip creating view for tables from "foreign" schemas (not dataset) ie. by querying a table with known table but with schema mismatch

ideal we'd move checks in test_read_interfaces.py done only for filesystem to the specific tests. we can do it later as well

sh-rp:
@rudolfix I think all is addressed in the additional commit I just made. I'm not quite sure what you mean with testing open_connection several times. I am testing this by opening the sql_client context on the filesystem a few times in a row now, but there is no test for parallel access, if that is what you need lmk.

rudolfix (dismissed):
LGTM!

rudolfix (dismissed):

PR #1508, opened 2024-06-23
Updated credentials docs

VioletM:


burnash:
@VioletM looks like some links are broken, so the build fails https://app.netlify.com/sites/dlt-hub-docs/deploys/66787c07e3f41b0008026c3f:

```
9:51:04 PM: - On source page path = /docs/examples/custom_destination_bigquery:
9:51:04 PM:    -> linking to ../general-usage/credentials/config_specs#gcp-credentials (resolved as: /docs/general-usage/credentials/config_specs)
9:51:04 PM: 
9:51:04 PM: - On source page path = /docs/examples/google_sheets:
9:51:04 PM:    -> linking to ../general-usage/credentials/config_specs#gcp-credentials (resolved as: /docs/general-usage/credentials/config_specs)
9:51:04 PM:    -> linking to ../general-usage/credentials/config_specs#working-with-alternatives-of-credentials-union-types (resolved as: /docs/general-usage/credentials/config_specs)
```

akelad (commented):
I started reviewing content before remembering you wanted a structure review first, so I've submitted those comments anyways. 

About the structure:


**How to set up credentials**
* I think it's weird to have only `ENV variables` and `secrets.toml and config.toml` as sections if we mention vaults + default argument values in the beginning too. Should we add a section around vaults at least? The default argument values we can just refer people to the other docs page
* I'm wondering whether the naming convention should go first in this docs page? And I think we should start with the example in that section instead of explaining theoretical stuff. If we start with the naming convention in the docs, we can then keep the naming convention for ENV variables in the ENV variable section so we're not linking out to a different section from there.

**Using configs and secrets in code**
* I think this page should go last out of the three pages we have because it's advanced, right? 
* I don't understand why you would "read configs and secrets yourself" - perhaps we add an explanation to that there.

**Prebuilt credential types**
* How does this section relate to the first page? I think we need to link to it somehow because it's unclear to me atm 
* I'm actually also wondering if these should even be on a separate page? Aren't they all connected to a specific source or destination? Should they instead be described in those pages? 
* I mentioned this somewhere else in my comments I think, but this page needs a better name if we're gonna keep it. One that's more generally understandable.

VioletM:
> * I think it's weird to have only `ENV variables` and `secrets.toml and config.toml` as sections if we mention vaults + default argument values in the beginning too. Should we add a section around vaults at least? The default argument values we can just refer people to the other docs page

Added a couple sentences, but there is not much I can add.

> * I'm wondering whether the naming convention should go first in this docs page? And I think we should start with the example in that section instead of explaining theoretical stuff. If we start with the naming convention in the docs, we can then keep the naming convention for ENV variables in the ENV variable section so we're not linking out to a different section from there.
@akelad 

Moved naming convention closer to the top. I don't think we should put an example first there. Each source and destination has their own example. Also, if people want examples, they can just go to the examples section right away

> **Using configs and secrets in code**
> 
> * I think this page should go last out of the three pages we have because it's advanced, right?

No, it's not the most advanced, it's right in the middle. Because it explains how the injection mechanism works. So if you use a verified source you need `How to setup credentials` section. If you're creating a custom source, you'd need this section.

> * I don't understand why you would "read configs and secrets yourself" - perhaps we add an explanation to that there.

Added

> **Prebuilt credential types**

> * How does this section relate to the first page? I think we need to link to it somehow because it's unclear to me atm

Added a section and a link to the first page

> * I'm actually also wondering if these should even be on a separate page? Aren't they all connected to a specific source or destination? Should they instead be described in those pages?

I think it should be a separate page because there is some generic information as well on how to write custom credential type.

> * I mentioned this somewhere else in my comments I think, but this page needs a better name if we're gonna keep it. One that's more generally understandable.

Hm, what is not understandable about this name? I'm having trouble coming up with a different one :) Maybe just Credential types?

rahuljo:
I like the structure, but it took me a while to understand it. Maybe we could try to be more explicit with the purpose of each section by using titles and section headers. That would make it easier to navigate. Example:

**Configuration and secrets**  
|\-\-\- **How to: Set-up**    
|\-\-\- **How to: Pass it to the pipeline script**  
|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|\-\-\-  Automatically (Injection Mechanism)  
|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|\-\-\-  Manually  
|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|\-\-\-  Advanced: write it from the pipeline script  
|\-\-\- **Special Credentials Types**

akelad (approved):
yay amazing work!!!

docs/website/sidebars.js
```diff
@@ -136,6 +136,19 @@ const sidebars = {
...
         'general-usage/source',
+        {
+          type: 'category',
+          label: 'Configuration and secrets',
+           link: {
+            type: 'doc',
+            id: 'general-usage/credentials/index',
+          },
+          items: [
+            'general-usage/credentials/setup',
```
burnash, line 147: Since we're renaming the section, before merging to master, we'd need to set up redirect from the older pages.

docs/website/sidebars.js
```diff
@@ -136,6 +136,19 @@ const sidebars = {
...
+            'general-usage/credentials/advanced',
```
burnash, line 148: Again, since we're renaming the sections anyway, maybe the `general-usage/configuration` path would be more consistent with the title of the page?

docs/website/docs/general-usage/credentials/index.md
```diff
@@ -0,0 +1,18 @@
+---
+title: Configuration and Secrets
+description: How to configure dlt pipelines and set up credentials
+keywords: [credentials, secrets.toml, secrets, config, configuration, environment variables]
+---
+import DocCardList from '@theme/DocCardList';
+
+`dlt` pipelines usually require configurations and credentials. These can be set up in [various ways](setup):
```
burnash, line 8:
```suggestion
`dlt` pipelines usually require configurations and credentials (secrets):
- **Configuration** refers to non-sensitive settings that define a data pipeline's behavior. These include file paths, database hosts, timeouts, API URLs, and performance settings.
- **Secrets** are sensitive data like passwords, API keys, and private keys. They should never be hard-coded to avoid security risks. 
These can be set up in [various ways](setup):
```

I would add a short definition of configuration and secrets:

docs/website/docs/general-usage/credentials/setup.md
```diff
@@ -0,0 +1,607 @@
...
+password="password"
+database = "database"
+host = "service-account"
+warehouse = "warehouse_name"
+role = "role"
+```
+
+`dlt` can work with both ways and convert one to another. To learn more about which credential types are supported, visit the [prebuilt credential types](prebuilt_types) page.
+
+## ENV variables
```
burnash, line 105:
```suggestion
## Environment variables
```
I'd use the full name for consistency

burnash (dismissed):
Awesome work @VioletM. I left a couple of comments for improvements. We can also add them after the merge.
The important point is that we need to setup redirects for old config docs urls after before we merge this to master.

docs/website/docs/general-usage/credentials/index.md
VioletM: Hm I wanted to keep this section short otherwise the other sections wouldn't be visible
image

burnash: I don't think it's a big problem. The "Details" will still be visible.

akelad:
Reposting my comment from DMs just so it doesn't get lost: Maybe we can just rename the "Prebuilt credential types" to "Complex credential types" and the description could be "Instructions for credentials like xyz" so it's clear what it's about

docs/website/sidebars.js
VioletM: good spotting!

docs/website/sidebars.js
VioletM: hm, I don't think it that important, also there are far more info about credentials then configurations

docs/website/docs/general-usage/credentials/index.md
VioletM: Added this info to `How to set up credentials` because I really want the index page to be short

docs/website/docs/general-usage/credentials/setup.md
VioletM: Fixed!

akelad (approved):

Issue #1325, opened 2024-05-06
RESTClient: add integration tests for all paginators


burnash:
Every paginator in the repo should have an offline (mock) integration test validating that the paginator actually works with the `RESTClient.paginate()`

burnash:
Implemented in #1509

PR #1509, opened 2024-06-23
RESTClient: add integrations tests for paginators

burnash:
This PR:
- extends mock_api_server in `tests/sources/helpers/rest_client/conftest.py` to enable it to serialize test records using different pagination methods. 
- uses new pagination methods to extend paginators tests to actually test if records fetched by `paginate()` paginated correctly.

Implements #1325

TODO:
- [x] HeaderLinkPaginator  
- [x] JSONResponsePaginator
- [x] SinglePagePaginator
- [x] OffsetPaginator
- [x] PageNumberPaginator
- [x] JSONResponseCursorPaginator
- [x] Update docs


tests/sources/helpers/rest_client/test_paginators.py
```diff
@@ -317,3 +371,60 @@ def test_maximum_page(self):
...
+        paginator.update_state(response)
+        assert paginator._next_reference == "cursor-2"
+        assert paginator.has_next_page is True
+
+    def test_update_request(self):
+        paginator = JSONResponseCursorPaginator(cursor_path="next_cursor")
+        paginator._next_reference = "cursor-2"
+        request = Request(method="GET", url="http://example.com/api/resource")
+        paginator.update_request(request)
+        assert request.params["cursor"] == "cursor-2"
```
willi-mueller, line 420: I was surprised that this method modifies the argument by reference without returning it. Could we make it return the modified object?

```suggestion
        updated_request = paginator.update_request(request)
        assert updated_request.params["cursor"] == "cursor-2"
```

tests/sources/helpers/rest_client/test_mock_api_server.py
willi-mueller, line 1: It took me a moment to understand the intention of this file. 
The purpose of the test suite in this module is to test the functionality of our mock API server. This does not test the functionality of the paginators defined in dlt/sources/helpers/rest_client/paginators.py.

docs/website/docs/general-usage/http/rest-client.md
willi-mueller, line 1: I noticed that we did not comment on the `SinglePagePaginator` yet. I think that would be useful because maybe there are single-page API responses that our auto-detection does not recognize. Specifying a paginator manually might fail. 

I'd have to think more deeply about in which cases only the SinglePagePaginator would work. I remember that I encountered that in February while this was still under active development. 
On the other hand, if we have that SinglePagePaginator code available, I suggest documenting it. Otherwise, people might not know about it and thus not use it.

tests/sources/helpers/rest_client/test_paginators.py
willi-mueller, line 1: I wonder if it would make sense to also add tests that ensure that:
- the first request is without the pagination parameter
- only starting from the second page do we find a `page=foo` etc. in the request params

willi-mueller (approved):
Wonderful! I just have some minor comments that could be addressed later.

rudolfix:
@willi-mueller @burnash I'm merging it so we can do another alpha release now. please write followup ticket (for actionable items) and Anton you can take a look at the comments in the closed PR

Issue #751, opened 2023-11-10
improve destination fingerprinting and info


rudolfix:
**Background**
Building on #746 we want a robust fingerprinting and destination info to include in `LoadInfo` and pipeline traces

**Tasks**
1. * [ ] add method to destination factory that will return a destination info as named tuple/dict
2. * [ ] improve fingerprinting of destinations that run locally (duckdb/filesystem/postgres)
3. * [ ] improve `filesystem` fingerprint by including only the schema and netloc. currently we also include path in the bucket which is too much

**Implementation**
destination info contains:
- destination type, name, environment and fingerprint
- destination str() representation (which is display safe)
- local/remote flag

local destination:
for destinations that may run locally (duckdb, postgres, waeviate, quadrant, filesystem etc.) we should start generating fingerprints
- if destination runs locally (always duckdb, `file://` on filesystem, localhost and 127.0.0.1 + ip6 localhost if connection string) a local flag must be set to true in info
- fingerprint must include `.anonymous_id` as used by telemetry and path to file/database when applicable
- `.anonymous_id` should be detached from telemetry code and become independent (maybe part of `paths.py` module)

PR #1516, opened 2024-06-26
feat(filesystem): use only netloc and scheme for fingerprint

IlyaFaer:
Towards #751 

Implements item 1


IlyaFaer:
A couple of errors are happening, but they seem to be not related temporary problems with cloud databases.

sh-rp:
@IlyaFaer could you update your branch to devel? I think the failing tests should go away then.

IlyaFaer:
@sh-rp, @rudolfix, yes, it's passing now

rudolfix (approved):
LGTM!

PR #1523, opened 2024-06-27
Feat/simplifies naming convention writing

rudolfix:
### Description
A followup on #998 
- adds examples for custom naming conventions
- simplifies how naming conventions are implemented and configured
- adds some tests


docs/examples/custom_naming/sql_ci_no_collision.py
```diff
@@ -0,0 +1,34 @@
+from typing import ClassVar
```
sh-rp, line 1: these extra files will not show up in the docs, maybe this is ok for now, but we might need to adapt this somehow?

docs/examples/custom_naming/custom_naming.py
```diff
@@ -0,0 +1,91 @@
...
+With this example you will learn to:
+* Create a naming convention module with a recommended layout
+* Use naming convention by explicitly passing it to `duckdb` destination factory
+* Use naming convention by configuring it config.toml
+* Changing the declared case sensitivity by overriding `is_case_sensitive` property
+* Providing custom normalization logic by overriding `normalize_identifer` method
+
+"""
+
+import dlt
```
sh-rp, line 28: really cool example!

docs/tools/prepare_examples_tests.py
```diff
@@ -54,8 +54,12 @@
             os.unlink(test_example_file)
             continue
 
-        with open(example_file, "r", encoding="utf-8") as f:
-            lines = f.read().split("\n")
+        try:
+            with open(example_file, "r", encoding="utf-8") as f:
+                lines = f.read().split("\n")
+        except FileNotFoundError:
```
sh-rp, line 60: what is this for? when does this happen?

docs/website/docs/general-usage/naming-convention.md
```diff
@@ -66,16 +58,20 @@ In the snippet below, we define a resource with various "illegal" unicode charac
...
-* `dlt` applies final naming convention in `normalize` step. Naming convention comes from (1) explicit configuration (2) from destination capabilities. Naming convention
-in schema will be ignored.
-* You can change the naming convention in the capabilities: (name, case-folding, case sensitivity)
+* Each destination may define a preferred naming convention (ie. Weaviate), otherwise **snake case** will be used
+* This naming convention is used when new schemas are created. This happens when pipeline is run for a first time.
+* Schemas preserve naming convention when saved. Your running pipelines will maintain existing naming conventions if not requested otherwise
+* `dlt` applies final naming convention in `normalize` step. Naming convention comes from (1) explicit configuration (2) from destination capabilities.
+* Naming convention will be used to put destination is case sensitive/insensitive mode and apply the right case folding function.
+
+:::caution
```
sh-rp, line 67: Maybe we should add here that we do not recommend to change the naming convention for an existing dataset and if need to do it, a fresh load to a new dataset would be the least errorprone.

dlt/common/normalizers/naming/naming.py
```diff
@@ -31,15 +39,13 @@ def normalize_table_identifier(self, identifier: str) -> str:
         """Normalizes and shortens identifier that will function as a dataset, table or schema name, defaults to `normalize_identifier`"""
         return self.normalize_identifier(identifier)
 
-    @abstractmethod
     def make_path(self, *identifiers: str) -> str:
```
sh-rp, line 42: imho these should be called split_path and join_path (same as string operations). not really important but it makes it much clearer what happens here.

docs/examples/custom_naming/sql_ci_no_collision.py
rudolfix: heh true but people may click in the link and they will go to github. LGTM, for now

docs/tools/prepare_examples_tests.py
rudolfix: something (I think Airflow) creates `logs` folder in the examples folder when I run any example. this prevents me from refreshing examples and I must remove the `logs` folder each time

docs/website/docs/general-usage/naming-convention.md
rudolfix: this documentation needs at least a day of work...

Issue #1337, opened 2024-05-07
Fix Qdrant tests on CI


burnash:
## Background
#998 makes qdrant working again but leaves a few things to be fixed. Qdrant is querying state in a wrong way (relies on order which is not there). Some tests are incorrect. Also we should test embedded qdrant.

Note we use following hack to go around ordering:
```python
    def _create_point_no_vector(self, obj: Dict[str, Any], collection_name: str) -> None:
        """Inserts a point into a Qdrant collection without a vector.

        Args:
            obj (Dict[str, Any]): The arbitrary data to be inserted as payload.
            collection_name (str): The name of the collection to insert the point into.
        """
        # we want decreased ids because the point scroll functions orders by id ASC
        # so we want newest first
        id_ = 2**64 - int(precise_time() * 10**6)
        self.db_client.upsert(
            collection_name,
            points=[
                models.PointStruct(
                    id=id_,
                    payload=obj,
                    vector={},
                )
            ],
        )
```

## Task
1. * [x] decide if the keep the hack above or to enable ordering of records by created_at. ordering needs additional indexes, and indexes work only on server version of qdrant. so not sure it is worth the effort.
2. * [x] add qdrant to all the tests where weaviate is added
3. * [x] write additional tests that make sure we get the newest schema and right state when we have state that is not committed (no corresponding entry in the _dlt_loads collection)
4. * [ ] fix `test_pipeline_replace` which is loading same data twice 
5. * [ ] enable local qdrant tests on CI (we have workflow with local tests)
6. * [x] enable server qdrant tests

rudolfix:
@steinitzu #1545 is fixing some of issues above. IMO two of them still remain:
- run tests on embedded/file version of qdrant (I think you spin up server on local tests)
- `test_pipeline_replace` looks like loading same data twice so is not testing `replace` disposition
pls do another PR for those two

PR #1526, opened 2024-06-27
Fix/qdrant tests in CI

steinitzu:
### Description

* Fixes issues with qdrant state restore
* Support local qdrant
* Test both local and server qdrant (same tests as weaviate)
*  server qdrant runs in docker in CI instead of cloud

### Related Issues

- Resolves https://github.com/dlt-hub/dlt/issues/1337

### DIFF
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -63,6 +63,28 @@ class StorageSchemaInfo(NamedTuple):
     inserted_at: datetime.datetime
     schema: str
 
+    @classmethod
+    def from_normalized_mapping(
+        cls, normalized_doc: Dict[str, Any], naming_convention: NamingConvention
+    ) -> "StorageSchemaInfo":
+        """Instantiate this class from mapping where keys are normalized according to given naming convention
+
+        Args:
+            normalized_doc: Mapping with normalized keys (e.g. {Version: ..., SchemaName: ...})
+            naming_convention: Naming convention that was used to normalize keys
+
+        Returns:
+            StorageSchemaInfo: Instance of this class
+        """
+        return cls(
+            version_hash=normalized_doc[naming_convention.normalize_identifier("version_hash")],
+            schema_name=normalized_doc[naming_convention.normalize_identifier("schema_name")],
+            version=normalized_doc[naming_convention.normalize_identifier("version")],
+            engine_version=normalized_doc[naming_convention.normalize_identifier("engine_version")],
+            inserted_at=normalized_doc[naming_convention.normalize_identifier("inserted_at")],
+            schema=normalized_doc[naming_convention.normalize_identifier("schema")],
+        )
+
 
 @dataclasses.dataclass
 class StateInfo:
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -82,6 +82,29 @@ class StateInfo:
             doc.pop("version_hash")
         return doc
 
+    @classmethod
+    def from_normalized_mapping(
+        cls, normalized_doc: Dict[str, Any], naming_convention: NamingConvention
+    ) -> "StateInfo":
+        """Instantiate this class from mapping where keys are normalized according to given naming convention
+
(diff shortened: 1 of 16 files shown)


dlt/destinations/impl/qdrant/factory.py
```diff
@@ -26,6 +28,20 @@ def _raw_capabilities(self) -> DestinationCapabilitiesContext:
...
+        cls,
+        caps: DestinationCapabilitiesContext,
+        config: QdrantClientConfiguration,
+        naming: t.Optional[NamingConvention],
+    ) -> DestinationCapabilitiesContext:
+        caps = super(qdrant, cls).adjust_capabilities(caps, config, naming)
+        if config.credentials.is_local():
+            # Local qdrant can not load in parallel
+            caps.loader_parallelism_strategy = "sequential"
+            caps.max_parallel_load_jobs = 1
```
steinitzu, line 42: workaround for local qdrant, it is not thread safe

dlt/load/load.py
```diff
@@ -196,7 +196,9 @@ def w_spool_job(
     def spool_new_jobs(self, load_id: str, schema: Schema) -> Tuple[int, List[LoadJob]]:
         # use thread based pool as jobs processing is mostly I/O and we do not want to pickle jobs
         load_files = filter_new_jobs(
-            self.load_storage.list_new_jobs(load_id), self.destination.capabilities(), self.config
+            self.load_storage.list_new_jobs(load_id),
+            self.destination.capabilities(self.initial_client_config),
```
steinitzu, line 200: I had to pass config here to get the correct adjusted capabilities (see `qdrant/factory.py` changes). Please double check.

tests/load/pipeline/test_restore_state.py
```diff
@@ -65,9 +65,10 @@ def test_restore_state_utils(destination_config: DestinationTestConfiguration) -
     with p.destination_client(p.default_schema.name) as job_client:  # type: ignore[assignment]
         with pytest.raises(DestinationUndefinedEntity):
             load_pipeline_state_from_destination(p.pipeline_name, job_client)
-        # sync the schema
-        p.sync_schema()
-        # check if schema exists
+    # sync the schema
+    p.sync_schema()
+    # check if schema exists
```
steinitzu, line 70: Local qdrant uses flock and can't handle more than one instance of client, why I moved all the pipeline operations out of `with destination_client()` scope

dlt/destinations/impl/qdrant/factory.py
rudolfix: very true! this is why this capability got added :)

dlt/common/destination/reference.py
```diff
@@ -63,6 +63,28 @@ class StorageSchemaInfo(NamedTuple):
     inserted_at: datetime.datetime
     schema: str
 
+    @classmethod
+    def from_normalized_mapping(
```
rudolfix, line 67: this is good!

rudolfix (dismissed):
LGTM! I can't find where you test local (not server) quadrant. the test suite seems to be run only once

dlt/load/load.py
rudolfix: I even improved it and requested full (not just initial) config here

steinitzu:
> LGTM! I can't find where you test local (not server) quadrant. the test suite seems to be run only once

I just added two destination configs so they both run together. (see `local-file`/`server` suffix)

```
tests/load/pipeline/test_arrow_loading.py::test_load_arrow_with_not_null_columns[arrow-batch-qdrant-no-staging-local-file] PASSED [ 33%]
tests/load/pipeline/test_arrow_loading.py::test_load_arrow_with_not_null_columns[arrow-batch-qdrant-no-staging-server] PASSED [ 33%]
```

https://github.com/dlt-hub/dlt/pull/1526/files#diff-136cf9e252c2cfe3f09641909b446ac65b5d00eb9286caf484bc5f1032437988R288-R293

PR #1531, opened 2024-07-01
rest_api: add a quick example to rest_api docs

burnash:
This PR adds a short example to the top of the rest_api docs illustrating the configuration format.

### DIFF
--- a/docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
+++ b/docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
@@ -9,6 +9,54 @@ import Header from './_source-info-header.md';
 
 This is a generic dlt source you can use to extract data from any REST API. It uses [declarative configuration](#source-configuration) to define the API endpoints, their [relationships](#define-resource-relationships), how to handle [pagination](#pagination), and [authentication](#authentication).
 
+### Quick example
+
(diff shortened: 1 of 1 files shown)


docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
```diff
@@ -9,6 +9,54 @@ import Header from './_source-info-header.md';
...
+        "auth": {
+            "token": dlt.secrets["your_api_token"],
+        },
+        "paginator": {
+            "type": "json_response",
+            "next_url_path": "paging.next",
+        },
+    },
+    "resources": [
+        "posts",
```
sh-rp, line 32: one suggestion: maybe add a python comment here that explains that a simple entry of "posts" will hit https://api.example.com/posts and no further configuration for this endpoint is needed.

docs/website/docs/dlt-ecosystem/verified-sources/rest_api.md
```diff
@@ -9,6 +9,54 @@ import Header from './_source-info-header.md';
...
+        {
+            "name": "comments",
```
sh-rp, line 34: and here you could add a very short comment to explain that this is a more complex endpoints, hence a dict is needed.

sh-rp (dismissed):
I have a small suggestion, but it's not a must imho, you can decide

PR #1534, opened 2024-07-02
feat(destinations): implement destination_info property

IlyaFaer:
Towards #751 

Implements item 2

### DIFF
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -577,7 +577,8 @@ TDestinationReferenceArg = Union[
 
 
 class Destination(ABC, Generic[TDestinationConfig, TDestinationClient]):
-    """A destination factory that can be partially pre-configured
+    """
+    A destination factory that can be partially pre-configured
     with credentials and other config params.
     """
 
@@ -652,6 +653,22 @@ class Destination(ABC, Generic[TDestinationConfig, TDestinationClient]):
     def _raw_capabilities(self) -> DestinationCapabilitiesContext:
         """Returns raw capabilities, before being adjusted with naming convention and config"""
         ...
+
+    @property
+    def destination_info(self) -> Dict[str, Any]:
+        """Return the destination info as a dict.
+
+        Returns:
+            Dict[str, Any]: Destination info.
+        """
+        conf = self.configuration(self.spec(), accept_partial=True)
+        return {
+            "destination_name": self.destination_name,
+            "destination_type": self.destination_type,
+            "environment": conf.get("environment"),
+            "fingerprint": conf.fingerprint(),
+            "repr": self.destination_description,
+        }
 
     @property
     def destination_name(self) -> str:
--- a/dlt/destinations/impl/destination/configuration.py
+++ b/dlt/destinations/impl/destination/configuration.py
@@ -9,6 +9,7 @@ from dlt.common.destination.reference import (
 )
 from dlt.common.typing import TDataItems
 from dlt.common.schema import TTableSchema
+from dlt.common.utils import digest128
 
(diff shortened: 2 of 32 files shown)


dlt/destinations/impl/destination/configuration.py
```diff
@@ -27,6 +28,14 @@ class CustomDestinationClientConfiguration(DestinationClientConfiguration):
     skip_dlt_columns_and_tables: bool = True
     max_table_nesting: Optional[int] = 0
 
+    def fingerprint(self) -> str:
+        """Returns fingerprint of the custom destination.
+
+        Returns:
+            str: The custom destination fingerprint.
+        """
+        return digest128(self.destination_callable.__name__)
```
IlyaFaer, line 37: I guess fingerprint for custom destinations is also needed. Otherwise, it's always `''`

IlyaFaer:
@rudolfix, made a draft for the item 2 (you said we have to talk about it first) to see if it's close to what was requested.

dlt/common/destination/reference.py
```diff
@@ -518,6 +519,22 @@ def _raw_capabilities(self) -> DestinationCapabilitiesContext:
...
         ...
 
+    @property
+    def destination_info(self) -> Dict[str, Any]:
+        """Return the destination info as a dict.
+
+        Returns:
+            Dict[str, Any]: Destination info.
+        """
+        conf = self.configuration(self.spec())
```
rudolfix, line 529: pls pass `accept_partial=True` here so we can get basic config info without ie. credentials

dlt/destinations/impl/destination/configuration.py
```diff
@@ -27,6 +28,19 @@ class CustomDestinationClientConfiguration(DestinationClientConfiguration):
...
+    def fingerprint(self) -> str:
+        """Returns fingerprint of the custom destination.
+
+        Returns:
+            str: The custom destination fingerprint.
+        """
+        name = (
+            self.destination_callable
+            if isinstance(self.destination_callable, str)
+            else self.destination_callable.__name__
```
rudolfix, line 40: cool! but IMO a fully qualified function name __module__ + __name__ would be better

tests/destinations/test_custom_destination.py
```diff
@@ -58,10 +61,55 @@ def items_resource() -> TDataItems:
...
+        "destination_name": "test_sink",
+        "destination_type": "dlt.destinations.destination",
+        "environment": None,
+        "fingerprint": digest128("test_sink"),
+        "repr": "test_sink(dlt.destinations.destination)",
+    }
+
+
+@pytest.mark.parametrize("destination_name", IMPLEMENTED_DESTINATIONS)
+def test_destination_info(destination_name):
```
rudolfix, line 83: this is a good test. it should go to `tests/common/test_destinations.py`. I'm not sure it will pass - you create configs and sometimes this requires destination specific deps.

pls. fix the PR and let's see if tests are passing

rudolfix (changes requested):

dlt/common/destination/reference.py
IlyaFaer: Done

dlt/destinations/impl/destination/configuration.py
IlyaFaer: Done

tests/destinations/test_custom_destination.py
IlyaFaer: Done

IlyaFaer:
@rudolfix, oh, so in `dlt` repo there are no secrets for destinations? Hm-m. This test is supposed to be moved to `verified-sources` then?!

PR #1536, opened 2024-07-02
improves collision detection when naming convention changes 

rudolfix:
### Description
1. we prevent changing naming convention when there are any tables that are already created in the destination
2. we allow to change naming convention when all such tables are blocked and make sure the right names (pre change) are used when dropping tables
3. bugfix: tables were not dropped for staging dataset so `refresh` was not working with `merge`
4. no `print` linter rule added and some prints removed


dlt/common/known_env.py
```diff
@@ -0,0 +1,25 @@
+"""Defines env variables that `dlt` uses independently of its configuration system"""
+
+DLT_PROJECT_DIR = "DLT_PROJECT_DIR"
```
sh-rp, line 3: are these in the docs somewhere? I think people had been asking about how change the datadir etc.

dlt/load/utils.py
```diff
@@ -158,12 +162,17 @@ def _init_dataset_and_update_schema(
...
-        drop_table_names = [table["name"] for table in drop_tables]
         if hasattr(job_client, "drop_tables"):
             logger.info(
-                f"Client for {job_client.config.destination_type} will drop tables {staging_text}"
+                f"Client for {job_client.config.destination_type} will drop tables"
+                f" {drop_tables} {staging_text}"
+            )
+            job_client.drop_tables(*drop_tables, delete_schema=True)
+        else:
+            logger.warning(
```
sh-rp, line 172: which client does not support this? I thought this was a universal interface, it's implemented for filesystem too.

sh-rp: ah ok, the custom destination..

dlt/common/schema/schema.py
```diff
@@ -967,42 +982,91 @@ def _verify_update_normalizers(
...
+                    table_seen_data = utils.has_table_seen_data(norm_table)
+                    if schema_seen_data:
+                        _verify_identifiers(table, norm_table)
+                    schema_tables[norm_table["name"]] = norm_table
+
+            schema_seen_data = False
+            for table in self.data_tables(include_incomplete=True):
+                # TODO: when lineage is fully implemented we should use source identifiers
+                # not `table` which was already normalized
                 norm_table = utils.normalize_table_identifiers(table, to_naming)
```
sh-rp, line 1052: Does this make sense here? Maybe I have not thought it through properly, but we don't have the original identifiers in the schema for the data tables do we? Imho there is no way to really discover wether a new naming scheme is compatible with an old one except for on the internal tables. My intuitive solution for this would have been to allow naming changes and solve the problem by explaining in the docs that you'll need to manually update your destination table names and columns if you add a new convention that ends up loading to a different table or column.

dlt/load/utils.py
rudolfix: vector databases as well

dlt/common/schema/schema.py
rudolfix: you are right. next step in this naming convention thing is to store source identifiers in the schema (ie. attached to every table and column names). in that case whenever naming convention changes we are able to restore source identifiers and normalize them again. I do exactly this with `dlt` tables where we know the source identifiers (hardcoded in utils).

if we have source identifiers we are able to ensure that tables that have data didn't change any names when naming convention changes. currently I block changing naming conventions if there any tables with data. this is to prevent a catastrophic situation when someone changes naming by accident and tables are broken.

if you do not care of above, there's a config setting to enable that. I'm still writing the doc here:
https://dlthub.com/devel/general-usage/naming-convention#avoid-identifier-collisions

dlt/common/known_env.py
rudolfix: let's just link this
https://deploy-preview-1536--dlt-hub-docs.netlify.app/docs/api_reference/common/known_env

from our docs

rudolfix: https://github.com/dlt-hub/dlt/issues/1546

PR #1553, opened 2024-07-05
Fix: handle column order gracefully when loading into ddb from pqt

codingcyclist:
### Description
Loads into duck-db can fail after a schema change if the added columns are anything but varchar. The reason is that the `COPY ... FROM` command, which reads from the parquet file, does not re-order the columns to match the order of columns in the destination table (see https://github.com/duckdb/duckdb/issues/3506). Here's a scenario that I ran into by adding an additional `timestamp` column (`detail__started_at`) after I had already run the pipeline a few times (I'm using pydantic schema contracts):

#### Data in the load package (`detail__started_at` is at column index -3):
```
(Pdb) pd.read_parquet("/private/var/folders/55/d2bjkrjd693_3qcwmz41sjhc0000gn/T/pytest-of-sbumm/pytest-142/test_schema_migration0/.dlt/pipeline/consumer_api/load/loaded/1720192417.667019/failed_jobs/decision_outcome.eb28698587.0.parquet")
  version                                    id       detail_type  ...               detail__started_at       _dlt_load_id         _dlt_id
0       0  ad0d6a8b-50c4-4860-91ab-0538ab96fde9  decision_outcome  ... 2024-07-05 15:13:37.657248+00:00  1720192417.667019  puWD37sjTI/iUA
1       0  4c74b2fd-73b4-46e5-94d3-c405d2a3c04b  decision_outcome  ... 2024-07-05 15:13:37.657261+00:00  1720192417.667019  f1tqhvAFs3D7+A
```
#### Table schema after updates have already been applied (`detail__started_at` is at column index -1):
```
(Pdb) conn.sql("select data_type, is_nullable, column_name from information_schema.columns where table_name = 'decision_outcome'")
┌──────────────────────────┬─────────────┬─────────────────────┐
│        data_type         │ is_nullable │     column_name     │
│         varchar          │   varchar   │       varchar       │
├──────────────────────────┼─────────────┼─────────────────────┤
│ VARCHAR                  │ NO          │ version             │
│ VARCHAR                  │ NO          │ id                  │
│ VARCHAR                  │ NO          │ detail_type         │
│ VARCHAR                  │ NO          │ source              │
│ VARCHAR                  │ NO          │ account             │
│ TIMESTAMP WITH TIME ZONE │ NO          │ time                │
│ VARCHAR                  │ NO          │ region              │
│ VARCHAR                  │ NO          │ resources           │
│ VARCHAR                  │ NO          │ detail__decision_id │
│ BOOLEAN                  │ NO          │ detail__is_sandbox  │
│ JSON                     │ NO          │ detail__meta        │
│ VARCHAR                  │ NO          │ _dlt_load_id        │
│ VARCHAR                  │ NO          │ _dlt_id             │
│ TIMESTAMP WITH TIME ZONE │ YES         │ detail__started_at  │
├──────────────────────────┴─────────────┴─────────────────────┤
│ 14 rows                                            3 columns │
└──────────────────────────────────────────────────────────────┘
```
Hence, what happens is the pipeline fails with the following error:
````
In file "/private/var/folders/55/d2bjkrjd693_3qcwmz41sjhc0000gn/T/pytest-of-sbumm/pytest-142/test_schema_migration0/.dlt/pipeline/consumer_api/load/normalized/1720192417.667019/new_jobs/decision_outcome.eb28698587.0.parquet" the column "_dlt_id" has type VARCHAR, but we are trying to read it as type TIMESTAMP WITH TIME ZONE.
This can happen when reading multiple Parquet files. The schema information is taken from the first Parquet file by default. Possible solutions:
* Enable the union_by_name=True option to combine the schema of all Parquet files (duckdb.org/docs/data/multiple_files/combining_schemas)
* Use a COPY statement to automatically derive types from an existing table.


#### Fix
By using `INSERT INTO` in lieu of `COPY .... FROM` we can use the `BY NAME` option to be agnostic of the column ordering.

### DIFF
--- a/dlt/destinations/impl/duckdb/duck.py
+++ b/dlt/destinations/impl/duckdb/duck.py
@@ -125,24 +125,21 @@ class DuckDbCopyJob(LoadJob, FollowupJob):
                 lock: threading.Lock = TABLES_LOCKS.setdefault(
                     qualified_table_name, threading.Lock()
                 )
-                
+
                 with lock:
                     with sql_client.begin_transaction():
-                        sql_client.execute_sql(
-                            f"""
+                        sql_client.execute_sql(f"""
                             INSERT INTO {qualified_table_name}
                                 BY NAME
                                 SELECT * FROM read_parquet('{file_path}', union_by_name = true);
-                            """
-                        )
-                        
+                            """)
+
         elif file_path.endswith("jsonl"):
             # NOTE: loading JSON does not work in practice on duckdb: the missing keys fail the load instead of being interpreted as NULL
             options = ", COMPRESSION GZIP" if FileStorage.is_gzipped(file_path) else ""
             with sql_client.begin_transaction():
                 sql_client.execute_sql(
-                    f"COPY {qualified_table_name} FROM '{file_path}' ( FORMAT"
-                    f" JSON {options});"
+                    f"COPY {qualified_table_name} FROM '{file_path}' ( FORMAT JSON {options});"
                 )
 
     def state(self) -> TLoadJobState:
--- a/dlt/destinations/impl/duckdb/duck.py
+++ b/dlt/destinations/impl/duckdb/duck.py
@@ -119,27 +119,30 @@ class DuckDbCopyJob(LoadJob, FollowupJob):
 
         qualified_table_name = sql_client.make_qualified_table_name(table_name)
         if file_path.endswith("parquet"):
-            source_format = "PARQUET"
-            options = ""
             # lock when creating a new lock
             with PARQUET_TABLE_LOCK:
                 # create or get lock per table name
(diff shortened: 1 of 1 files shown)


rudolfix:
@codingcyclist thanks for this report! we'll need a deeper look because all parquet files that we generate must correspond to the order of columns in the schema. and columns in the schema must correspond to the columns in the database. so there's a deeper problem

fix on duck db level will be helpful as well. we now (from 0.5.x) allow to import external files and there constraints are more lax

codingcyclist:
I see. I wasn't able to figure out all the underlying workings, but I thought filing a PR would still be the easiest way to raise the bug and describe the specifics of the failure mode that I stumbled into

rudolfix:
finally we have a proper fix: #1674

Issue #919, opened 2024-01-30
Ability to define /configure target schema for staging tables and dlt internal tables.


sohail-ali:
## Requirements:
Allow to specify the name of the staging dataset in destination configuration.
1. * [x] The default setting should be the current name `%s_staging` where %s is schema name
2. * [x] also allow a string without placeholders (ie. `_dlt_staging` as discussed below)
3. * [x] currently both filesystem.py and sql_client.py create staging dataset name. this should happen in one place, preferably in the destination configuration itself.
4. * [x] write a test (duckdb is sufficient) where a string without placeholders is used and load data to a few different schemas, make sure that staging dataset for particular schema is
5. * [x] document in using dlt / destination

----------
### Feature description

Currently dlt internal tables are created in same schema where target table was created. If you have multiple target schemas than internal tables are created in all those schemas. I would be better to have an option to specify dedicated schema for internal tables and also for staging tables.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Currently dlt internal tables are created in same schema where target table was created. If you have multiple target schemas than internal tables are created in all those schemas. I would be better to have an option to specify dedicated schema for internal tables and also for staging tables.

### Proposed solution

Option to specify target schema for dlt internal and staging tables.  

Destination: snowflake

rudolfix:
@sohail-ali a few questions:
1. staging dataset is kept in a separate schema (`<schema_name>_staging`) or are you referring to something else here
> I would be better to have an option to specify dedicated schema for internal tables **and also for staging tables.**

2. if there are multiple target schemas we indeed create internal schemas in all of them. assuming that we create them in a single schema: is it ok to have a internal table name prefixed with schema name so we can keep data separated or we need to add schema name to the table as a column?
3. are you by any chance using `use_single_dataset` set to False in the pipeline configuration?
4. I do not understand how this change would help you at the end. is this more aesthetic? or you have some data replication/backup problem?
sorry if I missed any details in our slack

sohail-ali:
1. For staging dataset instead of having <schema_name>_staging. Can we configure single staging dataset for all the resources like _dlt_staging. Current approach will create multiple staging schemas in the database. This is more of an an aesthetic requirement rather than functional need. We like our target database / schema to look more like source database / schema. So  any internal tables or staging schemas created by dlt like them to be under one schema.

rudolfix:
[...]

PR #1555, opened 2024-07-07
adds missing docs / tests and configurable staging dataset name

rudolfix:
### Description
Adds naming convention docs, missing tests and is implementing #919 (to get normalized staging dataset name without hacks)

### DIFF
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -184,6 +184,8 @@ class DestinationClientDwhConfiguration(DestinationClientConfiguration):
     """name of default schema to be used to name effective dataset to load data to"""
     replace_strategy: TLoaderReplaceStrategy = "truncate-and-insert"
     """How to handle replace disposition for this destination, can be classic or staging"""
+    staging_dataset_name_layout: str = "%s_staging"
+    """Layout for staging dataset, where %s is replaced with dataset name. placeholder is optional"""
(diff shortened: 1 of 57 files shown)


dlt/common/destination/reference.py
```diff
@@ -201,21 +203,37 @@ def normalize_dataset_name(self, schema: Schema) -> str:
...
+            if not dataset_name
+            else schema.naming.normalize_table_identifier(dataset_name)
+        )
+
+    def normalize_staging_dataset_name(self, schema: Schema) -> str:
+        """Builds staging dataset name out of dataset_name and staging_dataset_name_layout."""
+        if "%s" in self.staging_dataset_name_layout:
+            # if dataset name is empty, staging dataset name is also empty
+            dataset_name = self._make_dataset_name(schema.name)
+            if not dataset_name:
```
sh-rp, line 218: does this ever happen? I'm reading the code and trying to understand the mechanism, but in which case would the result of _make_dataset_name be None? Should't self.dataset_name alway be set?

rudolfix: yes - in vector databases you can work on collections without dataset prefix. they override configuration making that field optional. btw. we should support that as well in clickhouse or dremio where we do not have schemas as well

sh-rp (dismissed):

Issue #1495, opened 2024-06-19
RESTClient: Rename JSONResponsePaginator to JSONLinkPaginator


burnash:
### Background 

Both `JSONResponsePaginator` and `HeaderLinkPaginator` derive from `BaseNextUrlPaginator` providing a pagination based on the extracted next url value. 

### Proposed change

To make naming of paginators consistent and better reflect their functionality, rename `JSONResponsePaginator` to `JSONLinkPaginator`.

### Task

- [x] Rename `JSONResponsePaginator` to `JSONLinkPaginator` in `dlt/sources/helpers/rest_client`
- [x] Update tests
- [x] Update the documentation

PR #1558, opened 2024-07-08
feat/1495 rest_client: renames JSONResponsePaginator to JSONLinkPaginator

willi-mueller:
### Description

Implements: https://github.com/dlt-hub/dlt/issues/1495

### Related Issues
- once this is merged and released, also merge the depending PR in verified-sources: https://github.com/dlt-hub/verified-sources/pull/523

### DIFF
--- a/docs/website/docs/general-usage/http/overview.md
+++ b/docs/website/docs/general-usage/http/overview.md
@@ -58,11 +58,11 @@ Note that we do not explicitly specify the pagination parameters in the example.
 ```py
 import dlt
 from dlt.sources.helpers.rest_client import RESTClient
-from dlt.sources.helpers.rest_client.paginators import JSONResponsePaginator
+from dlt.sources.helpers.rest_client.paginators import JSONLinkPaginator
 
 github_client = RESTClient(
     base_url="https://pokeapi.co/api/v2",
-    paginator=JSONResponsePaginator(next_url_path="next"),   # (1)
+    paginator=JSONLinkPaginator(next_url_path="next"),   # (1)
     data_selector="results",                                 # (2)
 )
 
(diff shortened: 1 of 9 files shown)


rudolfix (changes requested):
Looks good but this is a breaking change so what we need is
- JSONResponsePaginator as an alias or derive it directly from `JSONLinkPaginator`
- deprecate it, I think you can deprecate classes

what are you planning to do with `rest_api`? will it continue to work with old `dlt` or you will bump the required version?

burnash:
@willi-mueller Looks good, I agree with @rudolfix we need to deprecate JSONResponsePaginator,

@rudolfix will something like this work?

```python
import warnings

class JSONLinkPaginator(BaseNextUrlPaginator):
    ...


class JSONResponsePaginator(JSONLinkPaginator):
    def __init__(self):
        warnings.warn("JSONResponsePaginator is deprecated and will be removed in version 1.0.0. Use JSONLinkPaginator instead.", DeprecationWarning, stacklevel=2)
        super().__init__()
```

burnash (changes requested):
Great work, let's also deprecate the old class. See the comment above.

willi-mueller:
> Great work, let's also deprecate the old class. See the comment above.

Thanks! I marked the old class as deprecated and rebased onto devel.

docs/website/docs/general-usage/http/overview.md
```diff
@@ -58,11 +58,11 @@ Note that we do not explicitly specify the pagination parameters in the example.
...
 import dlt
 from dlt.sources.helpers.rest_client import RESTClient
-from dlt.sources.helpers.rest_client.paginators import JSONResponsePaginator
+from dlt.sources.helpers.rest_client.paginators import JSONLinkPaginator
 
 github_client = RESTClient(
     base_url="https://pokeapi.co/api/v2",
-    paginator=JSONResponsePaginator(next_url_path="next"),   # (1)
+    paginator=JSONLinkPaginator(next_url_path="next"),   # (1)
     data_selector="results",                                 # (2)
```
burnash, line 66:
```suggestion
    data_selector="results",                             # (2)
```

burnash (approved):
Looks good! Thank you @willi-mueller

PR #1563, opened 2024-07-08
feat: add config dataset_name_prefix

### LINKED ISSUE #1520: Create profile-based destination db schema names
### Feature description

When working in a team in a data setup the data layer in a shared database (be it development environment or Prod environment) should have an isolated data space for each member of the team. When I test my code on a shared development database, I want to make sure, that the data that I push into somewhere is located somewhere, where it doesn't interfere with the work of other. For example, if I decide to name a destination in my dlt pipeline `customers` and another colleague decides to use this name at the same time, we would have a conflict while developing our dlt pipeline. In dbt, there is the [custom schema](https://docs.getdbt.com/docs/build/custom-schemas) setup, which allows to use the same naming convention in the code but under the hood, the resulting schema gets a prefix based on the individual profile setup. This feature would make it easier to work with dlt as well and in my opinion should be added. 

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Currently, two colleagues wouldn't be able to create a code with the same destination's dataset name when developing a pipeline as they both would overwrite each other's table content while developing.

### Proposed solution

Add a `profile` attribute to the `secrets.toml` which is then used to prefix destination dataset names (like database schemas). 

### Related issues

_No response_

donotpush:
### Description
Adds `dataset_name_prefix` in `config.toml`. This variable will add a prefix to all destinations that require a dataset name.

### Related Issues
- Resolves #1520 

### Steps WIP:

1. * [x] add a new field to `PipelineConfiguration` ie. `dataset_name_prefix` which will be used to prefix dataset names generated by the `Pipeline`
2. * [x] amend the `Pipeline` class to generate correct datasets using the configuration. we do not need to add a new parameter to `dlt.pipeline`, just looking into the `config` is enough
3. * [x] a bunch of tests (in `tests/pipeline`) that will make sure a correct dataset name is created (also in `dev_mode`!). use the ENV representation for tests
4. * [ ] I just found out that we do not document pipeline configuration at all. Please add a chapter `Configure the pipeline` with a description how to use `config.toml` to specify the prefix. We'll most probably add a few more details there after a first review

the end result is that you'll be able to use `config.toml`
```toml
dataset_name_prefix="rudolfix"
```
to get all datasets created by affected pipelines prefixed with `rudolfix_`

### DIFF
--- a/dlt/pipeline/configuration.py
+++ b/dlt/pipeline/configuration.py
@@ -1,5 +1,6 @@
 from typing import Any, Optional
 
+import dlt
 from dlt.common.configuration import configspec
 from dlt.common.configuration.specs import RunConfiguration, BaseConfiguration
 from dlt.common.typing import AnyFun, TSecretValue
@@ -18,7 +19,7 @@ class PipelineConfiguration(BaseConfiguration):
     staging_name: Optional[str] = None
     loader_file_format: Optional[TLoaderFileFormat] = None
     dataset_name: Optional[str] = None
-    dataset_name_prefix: Optional[str] = None
+    dataset_name_prefix: Optional[str] = dlt.config.value
     pipeline_salt: Optional[TSecretValue] = None
     restore_from_destination: bool = True
     """Enables the `run` method of the `Pipeline` object to restore the pipeline state and schemas from the destination"""
@@ -42,10 +43,6 @@ class PipelineConfiguration(BaseConfiguration):
             self.runtime.pipeline_name = self.pipeline_name
         if not self.pipeline_salt:
             self.pipeline_salt = TSecretValue(digest256(self.pipeline_name))
-        if not self.dataset_name_prefix:
-            self.dataset_name_prefix = self.runtime.dataset_name_prefix
-        else:
-            self.runtime.dataset_name_prefix = self.dataset_name_prefix
 
 
 def ensure_correct_pipeline_kwargs(f: AnyFun, **kwargs: Any) -> None:
--- a/dlt/pipeline/configuration.py
+++ b/dlt/pipeline/configuration.py
@@ -19,7 +19,8 @@ class PipelineConfiguration(BaseConfiguration):
     staging_name: Optional[str] = None
     loader_file_format: Optional[TLoaderFileFormat] = None
     dataset_name: Optional[str] = None
-    dataset_name_prefix: Optional[str] = dlt.config.value
+    dataset_name_layout: Optional[str] = dlt.config.value
+    """Layout for dataset_name, where %s is replaced with dataset_name."""
     pipeline_salt: Optional[TSecretValue] = None
     restore_from_destination: bool = True
     """Enables the `run` method of the `Pipeline` object to restore the pipeline state and schemas from the destination"""
--- a/dlt/pipeline/configuration.py
+++ b/dlt/pipeline/configuration.py
@@ -7,6 +7,7 @@ from dlt.common.typing import AnyFun, TSecretValue
 from dlt.common.utils import digest256
 from dlt.common.destination import TLoaderFileFormat
 from dlt.common.pipeline import TRefreshMode
+from dlt.common.configuration.exceptions import ConfigurationValueError
 
 
 @configspec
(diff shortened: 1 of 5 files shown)


dlt/pipeline/configuration.py
```diff
@@ -41,6 +42,10 @@ def on_resolved(self) -> None:
             self.runtime.pipeline_name = self.pipeline_name
         if not self.pipeline_salt:
             self.pipeline_salt = TSecretValue(digest256(self.pipeline_name))
+        if not self.dataset_name_prefix:
```
rudolfix, line 45: do we ever use the prefix from `run_configuration`?

rudolfix (changes requested):
This looks good! Thanks for the proper test coverage. I have a few things to consider:
1. why do you need `dataset_name_prefix` in `run_configuration`?
`run_configuration` is embedded in pipeline configuration - it is one of the mechanisms that `dlt` offers. `run_configuration` is used to start the logs, initialize tracing etc. and is required by several component in `dlt`. we express this dependency via embedding
(I'm trying to explain why we have it here)
IMO you do not need `dataset_name_prefix` there

2. dataset prefix feature is for power users. so I had another though and maybe we should let people to define a layout with a placeholder (like we do for staging dataset in https://github.com/dlt-hub/dlt/issues/919
so `dataset_name_layout` could be `rudolfix_%s` and we substitute dataset name for %s?
then it is even more flexible. WDYT?

dlt/pipeline/pipeline.py
```diff
@@ -1372,6 +1372,13 @@ def _maybe_destination_capabilities(
             if injected_caps:
                 injected_caps.__exit__(None, None, None)
 
+    def _normalize_dataset_name(self, dataset_name: str) -> str:
+        """Normalizes the dataset name using the dataset_name_layout."""
+        if "%s" not in self.config.dataset_name_layout:
```
rudolfix, line 1377: please move this check to `on_resolved` method of `PipelineConfiguration`. this method is called after all fields were successfully resolved by config providers and is used to do additional processing and checks.

Raise `ConfigurationValueError` (specialized ValueError), you can also inspect other `on_resolved` impls to see what we typically do there

dlt/pipeline/pipeline.py
```diff
@@ -1372,6 +1372,13 @@ def _maybe_destination_capabilities(
...
+            raise ValueError("The dataset_name_layout must contain a '%s' placeholder.")
```
rudolfix, line 1378: You could tell the user why we need `%s` (to substitute `dataset_name` from the pipeline). Sometimes we even give examples so users can self-correct without reading more docs or coming to our slack

dlt/pipeline/configuration.py
```diff
@@ -18,6 +19,8 @@ class PipelineConfiguration(BaseConfiguration):
     staging_name: Optional[str] = None
     loader_file_format: Optional[TLoaderFileFormat] = None
     dataset_name: Optional[str] = None
+    dataset_name_layout: Optional[str] = dlt.config.value
```
rudolfix, line 22: it is OK to use it here but not necessary for correctly typed fields.
`dlt.config.value` and `dlt.secrets.value` are mostly used on `dlt.source` and `dlt.resource` function signatures to mark arguments that are required to be present and (when user is not adding types to arguments) to indicate which arguments contain secrets.

internally we convert function signatures into `configspec` (dataclasses and use those markers to correctly type them, mark as optional etc.)

in other words: as long as you create SPECs (not signatures) you do not need to use the marker above

rudolfix (changes requested):
We need some cosmetic changes and we are good to go! Docs can be updated in a separate PR since we have tests already passing

dlt/pipeline/pipeline.py
rudolfix: > return self.config.dataset_name_layout % dataset_name
this line is certainly correct

but the validation line
> if "%s" not in self.config.dataset_name_layout:
could be moved to `on_resolved` because it depends only on config instance content.

ping me on Slack if this is not clear... maybe I'm missing some nuances :)

rudolfix (approved):
LGTM! thx for the tests

PR #1569, opened 2024-07-09
Surface original err msg from pydantic as extended_info on DataValidationError

codingcyclist:
### Description
After I found myself spending the better half of a day trying to understand why my schema contract was failing with `In Table: decision_outcome Column: ('detail', 'meta', 'node_level_timings') . Contract on data_type with mode freeze is violated. `, I figured I might as well just file a PR to add the original pydantic error message (i.e. `Field required`) to dlt's `DataValidationError`. 

I think this is an important piece of information because right now, one cannot tell from the `DataValidationError` whether it failed for nullability or for actual dtype mismatch. Here are two examples

```
# Before:
In Table: decision_outcome Column: ('detail', 'complex_list', 1, 'str') . Contract on data_type with mode freeze is violated.

# After
In Table: decision_outcome Column: ('detail', 'complex_list', 1, 'str') . Contract on data_type with mode freeze is violated. Input should be a valid string

# Before
In Table: decision_outcome Column: ('detail', 'meta', 'node_level_timings') . Contract on data_type with mode freeze is violated. 

# After
In Table: decision_outcome Column: ('detail', 'meta', 'node_level_timings') . Contract on data_type with mode freeze is violated. Field required
```

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/dlt/common/libs/pydantic.py
+++ b/dlt/common/libs/pydantic.py
@@ -332,6 +332,7 @@ def validate_items(
                     list_model,
                     {"columns": "freeze"},
                     items,
+                    err["msg"]
                 ) from e
             # raise on freeze
             if err["type"] == "extra_forbidden":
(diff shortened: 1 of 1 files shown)


rudolfix (approved):
LGTM!

Issue #1571, opened 2024-07-09
Pipeline crashes when single record has None value at incremental cursor path


willi-mueller:
### dlt version

0.5.1

### Describe the problem

The following data crashes when trying to load it incrementally with the `cursor_path="created_at"`

```py
    data = [
        {"id": 1, "created_at": 1},
        {"id": 2, "created_at": None},
        {"id": 3, "created_at": 2},
    ]
```

I could not isolate it yet, but in version 0.4.12 the second row where the `created_at is None` was under some unknown conditions skipped over and not loaded into the destination.

### Expected behavior
- [x] Allow the user to specify whether the incremental transformer raises an error or accepts the row:
```
inc_0 = dlt.sources.incremental(cursor_path="updated_at", on_cursor_value_none="include")
inc_1 = dlt.sources.incremental(cursor_path="updated_at", on_cursor_value_none="raise")
```
- [x] Documentation how to set a default cursor_path in case the value is `None`.
- [x] Clarify with the exception whether the cursor path is missing or the value at the cursor path is `None`.

### Steps to reproduce

See test suite in https://github.com/dlt-hub/dlt/pull/1576

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

rudolfix:
@willi-mueller how `dlt` should behave in this case? We cannot really say if this record is in range or not if cursor column is not provided.

IMO we need nice exception message here explaining what happen, now I bet `max` function is simply failing on some Python internals

rudolfix:
@willi-mueller OK so it seems we need another setting in the incremental:
on_none_cursor_value: raise, include, exclude to govern such behavior

also should work on object and arrow tables
WDYT?

rudolfix:
@willi-mueller idea of the callable is also good but I'm not sure if `add_map` is not enough?
https://dlthub.com/docs/general-usage/resource#filter-transform-and-pivot-data

it does exactly what the user wants

willi-mueller:
> @willi-mueller OK so it seems we need another setting in the incremental: on_none_cursor_value: raise, include, exclude
> 
> so it is user's conscious decision what to do with such values. btw. this is possible even now by defining own `max` (last_value_func) that accepts nones.
> 
> also should work on object and arrow tables WDYT?

from what I observed, it crashes in `find_cursor_value()` which is executed even before the `last_value_func()`, such as `max()`.

This, a custom last_value_func does not seem to solve this problem.

[...]

PR #1576, opened 2024-07-10
Fix/1571 Incremental: Optionally load or ignore records with cursor_path missing or None value

willi-mueller:
### Description

**This PR:**
Allows users to specify what happens when the value at the incremental cursor path is `None` or the field is missing in a row. It also unifies the handling of null values in pandas/arrow with python objects.

Consider the following example data where `created_at` is the incremental cursor path:
```py
data_1 = [
  {"a": 1, "created_at": 1},
  {"a": 2, "created_at": None},
]

data_2 = [
  {"a": 1, "created_at": 1},
  {"a": 2},
]
```

The options are:
1. `incremental(..., on_cursor_value_missing="raise")`. This will raise `IncrementalCursorPathHasValueNone` for `data_1` and `IncrementalCursorPathMissing` for `data_2`.
2. `incremental(..., on_cursor_value_missing="include")`. This will load all rows for both `data_1` and `data_2` respectively.
3. `incremental(..., on_cursor_value_missing="exclude")`. This will load only the first row for both `data_1` and `data_2` respectively.

This PR also adds documentation on how to load data with `None` at the cursor path incrementally

All outlined features are implemented and tested for all 4 data formats (object, pandas, arrow-table, arrow batch). However, JSON path cursors are still only supported for JSON objects but not for arrow and pandas. 

Done in collaboration with @francescomucio 

### TODO
- [x] docs: explain new parameter
- [x] docs: explain how we can leverage `add_map()` to add default values

### Related Issues

- Resolves https://github.com/dlt-hub/dlt/issues/1571


VioletM:
@willi-mueller this is an important feature, thank you!
I have a question: records with None value at the incremental cursor will be re-loaded each time, right? Could maybe some fall-back cursor be specified in this case?

dlt/extract/incremental/typing.py
```diff
@@ -1,8 +1,9 @@
-from typing import TypedDict, Optional, Any, List, TypeVar, Callable, Sequence
+from typing import TypedDict, Optional, Any, List, Literal, TypeVar, Callable, Sequence
 
 
 TCursorValue = TypeVar("TCursorValue", bound=Any)
 LastValueFunc = Callable[[Sequence[TCursorValue]], Any]
+OnCursorValueNone = Literal["raise", "include"]
```
rudolfix, line 6: why no "exclude"? this OFC can be done with add_map.

tests/extract/test_incremental.py
```diff
@@ -633,6 +637,260 @@ def some_data(last_timestamp=dlt.sources.incremental("item.timestamp")):
...
+    ]
+    source_items = data_to_item_format(item_type, data)
+
+    @dlt.resource
+    def some_data(created_at=dlt.sources.incremental("created_at", on_cursor_value_none="include")):
+        yield source_items
+
+    p = dlt.pipeline(pipeline_name=uniq_id())
+    p.extract(some_data())
+    s = p.state["sources"][p.default_schema_name]["resources"]["some_data"]["incremental"][
```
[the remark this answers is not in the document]
rudolfix: another thing: better if you use duckdb and run the whole resource and then count rows in database. and make sure you have NULL value for the expected id

tests/extract/test_incremental.py
```diff
@@ -633,6 +637,260 @@ def some_data(last_timestamp=dlt.sources.incremental("item.timestamp")):
...
+        "created_at"
+    ]
+    assert s["last_value"] == 2
+
+
+def test_cursor_path_none_can_raise_on_none() -> None:
+    # No nested json path support for pandas and arrow. See test_nested_cursor_path_arrow_fails
```
rudolfix, line 704: cursor is not nested. arrow is ignoring nulls by default... which is different from json. we need to unify this

dlt/extract/incremental/transform.py
```diff
@@ -298,13 +319,13 @@ def __call__(
             tbl = tbl.filter(end_compare(tbl[cursor_path], end_value_scalar))
```
rudolfix, line 319: currently IMO arrow is ignoring nulls when doing max function: https://arrow.apache.org/docs/python/generated/pyarrow.compute.ScalarAggregateOptions.html#pyarrow-compute-scalaraggregateoptions

and I expect they are sometimes left in the table (initial load when no filters are applied) and sometimes removed (if end value is set or initial value is set please test it)

this is indeed a problem - even in the current code. what I'd do
if the cursor column is nullable: I'd filter out all the rows with cursor null. and raise if we do not allow for that. if we include nulls I'd add them back to table at the end of processing. otherwise we do not need to do anything

dlt/extract/incremental/__init__.py
```diff
@@ -132,14 +140,15 @@ def __init__(
...
         self.row_order = row_order
         self.allow_external_schedulers = allow_external_schedulers
+        self.on_cursor_value_none = on_cursor_value_none
 
         self._cached_state: IncrementalColumnState = None
         """State dictionary cached on first access"""
         super().__init__(lambda x: x)  # TODO:
 
-        self.end_out_of_range: bool = False
+        self.is_above_end_value: bool = False
```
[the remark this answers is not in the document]
willi-mueller: Got it, then I would keep it as it is

dlt/extract/incremental/typing.py
willi-mueller: Yes, I didn't want to duplicate the functionality of add_filter(). But, since we already have one parameter here I could add that too without much trouble.

dlt/extract/incremental/__init__.py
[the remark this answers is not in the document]
willi-mueller: I reverted this renaming.

tests/extract/test_incremental.py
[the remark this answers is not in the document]
willi-mueller: interesting: I adjusted the test and it turns out that all 3 rows are loaded for all 4 data item formats. I tested with both table_metrics, i.e. only extraction (see dc1696def35ab3195134937352c827477abee416) and full pipeline run (see dfd9ba6056fb3ecf492b6beec512ac90fe9989d8)

tests/extract/test_incremental.py
willi-mueller: Indeed. I corrected the comment

dlt/extract/incremental/transform.py
```diff
@@ -164,14 +177,22 @@ def __call__(
         ):
             return None, False, True
 
-        check_values = (row_value,) + ((last_value,) if last_value is not None else ())
+        if (row_value is None and last_value is None) or (
```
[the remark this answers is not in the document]
willi-mueller: Yes, indeed! 
I just needed to make one adjustment to support the exclusion of `row_value is None`

dlt/extract/incremental/__init__.py
```diff
@@ -95,6 +100,7 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
             specified range of data. Currently Airflow scheduler is detected: "data_interval_start" and "data_interval_end" are taken from the context and passed Incremental class.
             The values passed explicitly to Incremental will be ignored.
             Note that if logical "end date" is present then also "end_value" will be set which means that resource state is not used and exactly this range of date will be loaded
+        on_cursor_value_none: Specify what happens when a record has `None` at the cursor_path: raise, include
```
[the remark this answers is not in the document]
willi-mueller: Good point, I made it now so that value at cursor_path and cursor_path not in records are treated equally

dlt/extract/incremental/transform.py
```diff
@@ -117,18 +120,28 @@ def find_cursor_value(self, row: TDataItem) -> Any:
...
-            row_values = find_values(self._compiled_cursor_path, row)
-            if row_values:
-                row_value = row_values[0]
-        else:
-            try:
-                row_value = row[self.cursor_path]
-            except Exception:
-                pass
-        if row_value is None:
+        if self.cursor_path not in row.keys() and not self._compiled_cursor_path:
```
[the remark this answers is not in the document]
willi-mueller: Thank you! I reworked that. Please let me know if this is what you had in mind.

Alternatively, we could structure it:
```py
if self._compiled_cursor_path:
    if self.on_cursor_path_missing == "raise":
        ...
    if self.on_cursor_path_missing == "include":
        ...
    if self.on_cursor_path_missing == "exclude":
        ...
else: 
    # same structure as above
```

Please let me know which you prefer.

dlt/extract/incremental/transform.py
```diff
@@ -117,18 +120,28 @@ def find_cursor_value(self, row: TDataItem) -> Any:
...
+
+        row_value = self._value_at_cursor_path(row)
+
+        if self.on_cursor_value_none == "raise" and row_value is None:
+            raise IncrementalCursorPathHasValueNone(self.resource_name, self.cursor_path, row)
+
+        return row_value
+
+    def _value_at_cursor_path(self, row: TDataItem) -> Any:
+        row_value = row.get(self.cursor_path, None)
```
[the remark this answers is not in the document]
willi-mueller: Thanks. It's reworked now.

dlt/extract/incremental/transform.py
willi-mueller: That's a pragmatic idea, I implemented that approach. Please let me know what you think.

willi-mueller:
@rudolfix The test suite fails due to missing access rights to git@github.com:scale-vector/test_private_repo.git. 

The error message asks us to approve the key here: https://github.com/dlt-hub/test_private_repo/settings/keys/62219229. However, both I and @burnash could not do that approval. Could you please help?

willi-mueller:
@rudolfix Currently, the following test is failing:
`tests/pipeline/test_pipeline_extra.py::test_dump_trace_freeze_exception`

The reason is that before this PR, the following exception was ignored: `AttributeError: 'TestRow' object has no attribute 'get'` when trying to access `TestRow.get("id")`. Instead, the `IncrementalCursorPathMissing` was raised.

in this PR, we do not ignore the exception when the record object does not implement `__get__()`.

As a developer, I would appreciate the most precise exception and thus I would change the test case. Also, the test case seems to be concerned with how dlt handles traces and it seems to use the incremental only to produce a certain trace and it is not testing the incremental implementation.

What is your thought?

dlt/extract/incremental/transform.py
```diff
@@ -117,18 +120,49 @@ def find_cursor_value(self, row: TDataItem) -> Any:
...
+            # ignores the other found values, e.g. when the path is $data.items[*].created_at
+            row_value = cursor_values[0]
         else:
+            row_value = None
             try:
-                row_value = row[self.cursor_path]
-            except Exception:
+                row_value = row.get(self.cursor_path)
+            except AttributeError:
+                # just to pass tests/pipeline/test_pipeline_extra.py::test_dump_trace_freeze_exception
```
rudolfix, line 145: OK from my side to try getting with `getattr` on top of `get`. Then we support Pydantic models as well. You can leave failing trace test to me. This is testing some kind of super edge case when using pickle on local object. I can mess around to recreate it

dlt/extract/incremental/transform.py
```diff
@@ -117,18 +120,49 @@ def find_cursor_value(self, row: TDataItem) -> Any:
...
+            elif None in cursor_values:
+                if self.on_cursor_value_missing == "raise":
+                    raise IncrementalCursorPathHasValueNone(
+                        self.resource_name, self.cursor_path, row
+                    )
+                elif self.on_cursor_value_missing == "exclude":
+                    return None
+
+            # ignores the other found values, e.g. when the path is $data.items[*].created_at
+            row_value = cursor_values[0]
```
rudolfix, line 139: IMO you should take the first element and then check for None. Also let's add this behavior to docs (if not yet added). Currently you check None in all possible locations

dlt/extract/incremental/transform.py
```diff
@@ -117,18 +120,49 @@ def find_cursor_value(self, row: TDataItem) -> Any:
...
         Will use compiled JSONPath if present, otherwise it reverts to column search if row is dict
         """
-        row_value: Any = None
         if self._compiled_cursor_path:
-            row_values = find_values(self._compiled_cursor_path, row)
-            if row_values:
-                row_value = row_values[0]
+            cursor_values = find_values(self._compiled_cursor_path, row)
+            if cursor_values == []:
+                if self.on_cursor_value_missing == "raise":
```
rudolfix, line 126: I'm sure you can refactor this function that this code:
```
if self.on_cursor_value_missing == "raise":
                    raise IncrementalCursorPathMissing(self.resource_name, self.cursor_path, row)
                elif self.on_cursor_value_missing == "exclude":
                    return None
```
will be in it only once. right now it is duplicated 3 times

tests/extract/test_incremental.py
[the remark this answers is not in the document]
rudolfix: Interesting. If I remember correctly previous code in arrow transform sometimes was passing null values and sometimes not. For sure when loaded for a first times you'll get null values (3 records) but on next load data were filtered and old values were removed. This removed also null values (this is how filters work in arrow). So maybe this is not yet tested?

1. load 3 rows
2. load 3 rows where 1 row overlap with previous load and is filtered out, next row contains null and next a regular value. 2 rows shoudl stay

rudolfix (changes requested):
I kind of like it now! Please see the comments. thanks for all tests and docs!

dlt/extract/incremental/transform.py
willi-mueller: Thank you for the suggestion. I'm not sure how to achieve this because:
1. the exception in the case of "raise" is different. We have `IncrementalCursorPathMissing` and `IncrementalCursorPathHasValueNone`
2. we want to return None from this function in case the user wants to exclude such rows. I don't know how to return from the `find_cursor_value()` function if I extract the check for "exclude" and then `return None` into a separate function.

dlt/extract/incremental/transform.py
willi-mueller: Do you recommend taking the first element even before checking if the result is an empty list? In that case, I think we'd need to manage the control flow using try/except which is much slower than if/else so I'd avoid that.

The other alternative I see is flipping the order of checking for `None` and taking the first element. As follows:
```py
# ignores the other found values, e.g. when the path is $data.items[*].created_at
row_value = cursor_values[0]
if None in cursor_values:
    if self.on_cursor_value_missing == "raise":
        raise IncrementalCursorPathHasValueNone(
            self.resource_name, self.cursor_path, row
        )
    elif self.on_cursor_value_missing == "exclude":
        return None
```

tests/extract/test_incremental.py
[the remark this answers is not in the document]
willi-mueller: I added this test and as you predicted, it fails for arrow & pandas.

dlt/extract/incremental/transform.py
willi-mueller: Thanks for stepping in and doing the refactoring!

rudolfix (approved):
LGTM!

PR #1559, opened 2024-07-08
fix: add airflow db reset for all tests

donotpush:
### Description
Forces Airflow local database reset when running tests.

### Related Issues

- Fixes #1550 
- PRs #1554

PR #1578, opened 2024-07-10
fix: suppress alembic logs for tests

donotpush:
### Description
Suppresses alembic logs when running airflow db reset in tests.

### Related Issues

- Fixes #1559


rudolfix (approved):
so good!

PR #1579, opened 2024-07-10
Uses `python -m pip install` syntax for docs

kushaldas:
### Additional Context

https://snarky.ca/why-you-should-use-python-m-pip/ explains why we should use `python -m pip install dlt` this syntax.


sh-rp:
@kushaldas thanks for the PR. We have installation instructions on how to set up a venv, they were just not linked on the first page, I am closing this PR without merging in favor of adding this note here: https://github.com/dlt-hub/dlt/pull/1625

Issue #1583, opened 2024-07-11
RESTClient does not use configured request timeout


TyrantKingBen:
### dlt version

0.5.1

### Describe the problem

`RESTClient` does not use the configured `request_timeout` nor does it have a way to make it use a timeout.

### Expected behavior

The `RESTClient` should use the configured `request_timeout`.

### Steps to reproduce

Use the following configuration.
```toml
[runtime]
log_level = "DEBUG"
request_timeout = 0.001
request_max_attempts = 1
```

Create a resource that makes an API request using `RESTClient`. Run a pipeline with the resource. The pipeline will complete, but it should fail due to request timing out and max retries being reached.

### Operating system

Linux

### Runtime environment

Virtual Machine

### Python version

3.10

### dlt data source

Custom API data source using `RESTClient`.

### dlt destination

DuckDB

### Other deployment details

WSL2

### Additional information

By default `RESTClient` uses the `Session` requests helper. `Session` currently overrides the `request` method to set the default timeout. `Session` could also override the `send` method to set the default timeout.

```python
    def send(self, request, **kwargs):
        kwargs.setdefault("timeout", self.timeout)

        return super().send(request, **kwargs)
```

rudolfix:
@willi-mueller please take a look. we re-configure our built in session when config changes. look for `test_init_default_client` that is checking that

TyrantKingBen:
A related issue is that you cannot provide any of the standard request parameters (`proxies`, `stream`, `verify`, `cert`, `allow_redirects`) through the `RESTClient` methods. Specifically, `allow_redirects` cannot be used at all (unlike the others which you can workaround by configuring session defaults).

```python
RESTClient().get("https://example.com", allow_redirects=False) # errors
```

Something along the lines of the following code should allow passing these parameters while preserving the standard behavior of request/session settings merging.

```python
    def _send_request(self, request: Request, **kwargs: Any) -> Response:
        logger.info(
            f"Making {request.method.upper()} request to {request.url}"
            f" with params={request.params}, json={request.json}"
        )

        prepared_request = self.session.prepare_request(request)

        send_kwargs = self.session.merge_environment_settings(
            prepared_request.url,
            kwargs.pop("proxies", {}),
            kwargs.pop("stream", None),
            kwargs.pop("verify", None),
            kwargs.pop("cert", None),
        )

        send_kwargs.update(**kwargs)

        return self.session.send(prepared_request, **send_kwargs)

[...]

PR #1590, opened 2024-07-13
Feat/1583 rest client session timeout configuration

willi-mueller:
### Related Issues

- Implements #1583

### DIFF
--- a/docs/website/docs/general-usage/http/rest-client.md
+++ b/docs/website/docs/general-usage/http/rest-client.md
@@ -590,6 +590,22 @@ for page in paginate("https://api.example.com/posts"):
     print(page)
 ```
 
+
+## Retry
+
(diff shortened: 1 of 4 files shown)


docs/website/docs/general-usage/http/rest-client.md
```diff
@@ -590,6 +590,22 @@ for page in paginate("https://api.example.com/posts"):
     print(page)
 ```
 
+
+## Retry
```
willi-mueller, line 594: I think it would be beneficial to repeat the docs here, wdyt?
Or, should we only place a link to the request wrapper docs?

willi-mueller: we agreed upon that in meeting

PR #1591, opened 2024-07-14
allows nested types in BigQuery via autodetect_schema

rudolfix:
### Description
This PR allows users to offload schema inference and evolution (type inference from data, create and alter tables) to BigQuery.
As a consequence nested data types may be created (if present in arrow tables/parquet or inferred from json). 
Besides being useful in itself it is a PoC of how nested types support may look in `dlt`.
1. in case of arrow tables / parquet files containing nested fields, such fields are created and migrated/evolved in the destination
2. in case or json documents, nested fields are inferred and stored as nested types. `max_table_nesting` still works - user has a choice how far `dlt` should decompose json into tables and when to switch to nested types


docs/website/docs/dlt-ecosystem/verified-sources/arrow-pandas.md
```diff
@@ -120,6 +123,21 @@ pipeline.run(orders)
...
+* **duckdb** can do the same with `read_json`
+
+```py
+import duckdb
+
+conn = duckdb.connect()
+table = conn.execute(f"SELECT * FROM read_json_auto('{json_file_path}')").fetch_arrow_table()
+```
+
+Note that **duckdb** and **pyarrow** methods will generate [nested types](#loading-nested-types) for nested data, which are only partially by `dlt`.
```
sh-rp, line 139: there is a word missing here, probably "supported"

sh-rp:
@rudolfix the PR looks good, but I have a question. Are you sure that this schema auto detection really creates the schema the exact same way in bigquery as our explicit code would (with the exception of the complex fields of course)? At least this is what I understand it says in the docs. I think I would at least add a test that checks all data types or something like that.

sh-rp:
Or maybe it would be possible to only have the complex fields be autodetected and run the other migrations the way we always do, but I am actually not sure this makes sense.

rudolfix:
@sh-rp the schema is identical in case of parquet (because parquet schema is source of truth both for bigquery and dlt) - with the exception of `complex` fields which here are structs and in our case JSON

JSON documents are inferred differently and our schema does not match 100% regarding the types and column order. What is unfortunate - I can't create some columns with dlt and some (complex) with BigQuery because it complains that the schema is different. I do not think the problem is too great. The biggest difference so far is `_dlt_load_id` that is inferred as float

anyway this is a trick for power user until we do #1592 is done

Deninc:
@rudolfix can we do this with Redshift SUPER type? Loading from jsonl.

PR #1594, opened 2024-07-15
Add clarification for add_limit

VioletM:
Docs were confusing about whether the number of records is limited.

Closes https://github.com/orgs/dlt-hub/projects/11?pane=issue&itemId=70933002


rudolfix (changes requested):
@VioletM are you able to review docstring as well? we have `add_limit` method on `DltSource` and `DltResource`

docs/website/docs/general-usage/source.md
```diff
@@ -111,6 +111,10 @@ load_info = pipeline.run(pipedrive_source().add_limit(10))
 print(load_info)
 ```
 
+:::note
+Note that `add_limit` **does not limit the number of records** but rather the "number of yields". dlt will close the iterator/generator that produces data after the limit is reached.
```
AstrakhantsevaAA, line 115:
```suggestion
Note that `add_limit` **does not limit the number of records** but rather the "number of yields". `dlt` will close the iterator/generator that produces data after the limit is reached.
```

AstrakhantsevaAA (commented):
Very good!

AstrakhantsevaAA (dismissed):

AstrakhantsevaAA (approved):

rudolfix (approved):

PR #1599, opened 2024-07-16
Fix/1549 propagate max table nesting

### LINKED ISSUE #1549: max_table_nesting parameter at source level does not get inherited when selecting a resouce from the source.
### dlt version

dlt[snowflake]==0.4.12

### Describe the problem

There is a bug regarding the "max_table_nesting" parameter when applied at the source level.
It does not propagate to the resources when we select them before passing them to the pipeline.

Out of the three ways to select a resource from a source tested :
- source.resources["resource_1"]
- source.selected_resources.get("resource_1")
- source.with_resources("resource_1")

only this one ( **_source.with_resources("resource_1")_** ) is correctly inheriting the "max_table_nesting" parameter from the source.
The other two selection methods fail to do so.

### Expected behavior

_No response_

### Steps to reproduce

```
@dlt.source(max_table_nesting=0)
def the_ultimate_source():
    def table_1():
        yield [
            {"the_date": {"one": [1, 2, 3], "two": [5, 6, 7]}, "something_else": i}
            for i in range(5)
        ]

    def table_2():
        yield [{"the_date": i} for i in range(5)]

    yield dlt.resource(table_1, name="table_1", table_name="table_1")
    yield dlt.resource(table_2, name="table_2", table_name="table_2")


resource = the_ultimate_source().resources["table_1"] # does not propagate nesting parameter
# resource = the_ultimate_source().selected_resources.get("table_1") # does not propagate nesting parameter
# resource = the_ultimate_source().with_resources("table_1")  # propagates correctly nesting parameter
pipeline = dlt.pipeline(
    pipeline_name="testing_pipeline",
    destination="snowflake",
    pipelines_dir=".dlt/pipelines/",
)

info = pipeline.run(resource)
    print(step.step_info)
```

### Operating system

Linux

### Runtime environment

GitHub Codespaces

### Python version

3.10

### dlt data source

_No response_

### dlt destination

Snowflake

### Other deployment details

_No response_

### Additional information

Here is a the slack thread where a DLT support member in the #3-technical-help channel confirmed the issue :

https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1719437121433839

donotpush:
### Description
Considering the following:
```
 @dlt.source(max_table_nesting=nesting_level)
    def nesting_source():
        # Data with 4 nesting levels
        # Table names: table_1, table_1_children, table_1_children_children, table_children_children_children
        data = [
            {
                "level": 1,
                "children": [{"level": 2, "children": [{"level": 3, "children": [{"level": 4}]}]}],
            }
        ]
        yield dlt.resource(data, name="nest_source", table_name="table_1")

...
p = dlt.pipeline("sink_test_max_table_nesting", destination=nesting_sink, dev_mode=True)
p.run(nesting_source())
```

When the function `nesting_source()` executes, a sequence of `DltResource` objects is created with `max_table_nesting=None`, since this parameter isn't specified in the `dlt.resource` function signature. Later, these `DltResource` objects are included in the `DltSource` object.

The `DltSource` object does contain the expected value for `max_table_nesting`, but the `DltResource` objects never obtain this value from the source.

This PR impl sets max_table_nesting in the resource when source is created.

### Related Issues

- Fixes #1549

### DIFF
--- a/dlt/extract/decorators.py
+++ b/dlt/extract/decorators.py
@@ -215,6 +215,9 @@ def source(
             # apply hints
             if max_table_nesting is not None:
                 s.max_table_nesting = max_table_nesting
+                for resource in s._resources.values():
+                    if resource.max_table_nesting is None:
+                        resource.max_table_nesting = max_table_nesting
             s.schema_contract = schema_contract
             # enable root propagation
             s.root_key = root_key
--- a/dlt/extract/source.py
+++ b/dlt/extract/source.py
@@ -2,7 +2,7 @@ import contextlib
 from copy import copy
 import makefun
 import inspect
-from typing import Dict, Iterable, Iterator, List, Sequence, Tuple, Any
+from typing import Dict, Iterable, Iterator, List, Sequence, Tuple, Any, Optional
 from typing_extensions import Self
 
 from dlt.common.configuration.resolve import inject_section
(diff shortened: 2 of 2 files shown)


sh-rp:
@donotpush as discussed on slack, this is not a bug and this PR can be closed without merging.

Issue #1453, opened 2024-06-10
Support `merge` write disposition for `delta` table format on `fileystem` destination


jorritsandbrink:
### Feature description

I would like to use the `merge` write disposition when using the `delta` table format on the `fileystem` destination.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Proposed solution

This can be implemented using `delta-rs` `TableMerger` API: https://delta-io.github.io/delta-rs/api/delta_table/delta_table_merger/

### Related issues

https://github.com/dlt-hub/dlt/issues/1129

PR #1600, opened 2024-07-16
Add basic `upsert` support for `delta` table format in `filesystem` destination

jorritsandbrink:
### Description
This PR adds basic `merge` support for the `fileystem` destination when using `delta` as table format.

- only supports `upsert` merge strategy (not `delete-insert` and `scd2`)
- does not support deleting records from child tables (so updates to complex columns that involve element removals are not propagated)

### Related Issues

#1453 

### Additional context

- First 35 commits are from earlier PR: https://github.com/dlt-hub/dlt/pull/1466.
- New commits start from https://github.com/dlt-hub/dlt/pull/1600/commits/04437e72eed778e53bb2a89be1008b8d429ef277.


dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -123,16 +124,51 @@ def write(self) -> None:
...
 
         file_paths = [job.file_path for job in self.table_jobs]
 
-        write_delta_table(
-            path=self.client.make_remote_uri(self.make_remote_path()),
-            data=pa.dataset.dataset(file_paths),
-            write_disposition=self.table["write_disposition"],
-            storage_options=_deltalake_storage_options(self.client.config),
-        )
+        if (
```
jorritsandbrink, line 132: This should probably be refactored to match how we do it for SQL merge jobs:
- use separate class for merge job
- process entire table chain in job

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -123,16 +124,51 @@ def write(self) -> None:
...
+                    storage_options=_deltalake_storage_options(self.client.config),
+                )
+            )
+            is not None
+        ):
+            assert self.table["x-merge-strategy"] in self.client.capabilities.supported_merge_strategies  # type: ignore[typeddict-item]
+
+            if self.table["x-merge-strategy"] == "upsert":  # type: ignore[typeddict-item]
+                if "parent" in self.table:
+                    unique_column = get_first_column_name_with_prop(self.table, "unique")
```
jorritsandbrink, line 146: Perhaps these column identifiers need to be escaped.

dlt/destinations/sql_jobs.py
```diff
@@ -660,10 +660,14 @@ def gen_upsert_sql(
...
 
                 # insert records for new elements in the list
-                col_str = ", ".join(["{alias}" + escape_column_id(c) for c in table["columns"]])
+                table_column_names = list(map(escape_column_id, table["columns"]))
+                update_str = ", ".join([c + " = " + "s." + c for c in table_column_names])
+                col_str = ", ".join(["{alias}" + c for c in table_column_names])
                 sql.append(f"""
                     MERGE INTO {table_name} d USING {staging_table_name} s
                     ON d.{unique_column} = s.{unique_column}
+                    WHEN MATCHED
```
jorritsandbrink, line 669: Not related to `delta` tables.

This fixes a bug in our SQL destination `upsert` implementation. It had slipped trough because we didn't have a test that properly checked record updates. That test is added in this PR.

dlt/pipeline/pipeline.py
```diff
@@ -474,11 +474,13 @@ def _verify_destination_capabilities(
 
         # verify merge strategy
         for table in self.default_schema.data_tables(include_incomplete=True):
-            # temp solution to prevent raising exceptions for destinations such as
-            # `fileystem` and `weaviate`, which do handle the `merge` write
-            # disposition, but don't implement any of the defined merge strategies
-            if caps.supported_merge_strategies is not None:
-                if "x-merge-strategy" in table and table["x-merge-strategy"] not in caps.supported_merge_strategies:  # type: ignore[typeddict-item]
+            if "x-merge-strategy" in table and table["x-merge-strategy"] not in caps.supported_merge_strategies:  # type: ignore[typeddict-item]
+                if self.destination.destination_name in ("filesystem", "weaviate") and table["x-merge-strategy"] == "delete-insert":  # type: ignore[typeddict-item]
```
jorritsandbrink, line 478: Current behaviour of `merge` with `filesystem` destination:
1. `write_disposition="merge"` (which implicitly defaults to `delete-insert` merge strategy) ➜ fall back to `append`
2. `write_disposition={"disposition="merge", strategy="upsert"}` ➜ `upsert` merge strategy

I'm not a fan of this.

I think (1) should raise an error, because `delete-insert` is not implemented for `filesystem`.

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -116,8 +132,125 @@ def data(slice_: slice = None):
...
+    "destination_config",
+    destinations_configs(
+        default_sql_configs=True,
+        local_filesystem_configs=True,
+        supports_merge=True,
+    ),
+    ids=lambda x: x.name,
+)
+@pytest.mark.parametrize("merge_strategy", ("delete-insert", "upsert"))
+def test_merge_record_updates(
```
jorritsandbrink, line 146: Added this test to explicitly test record updates, which we didn't do so far in any of the existing tests (at least not beyond row counts).

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -597,39 +731,53 @@ def duplicates_no_child():
...
     ids=lambda x: x.name,
 )
 @pytest.mark.parametrize("merge_strategy", ("delete-insert", "upsert"))
 def test_complex_column_missing(
     destination_config: DestinationTestConfiguration,
     merge_strategy: TLoaderMergeStrategy,
 ) -> None:
+    if destination_config.table_format == "delta":
+        pytest.skip(
+            "Record updates that involve removing elements from a complex"
```
jorritsandbrink, line 746: In the SQL implementation for `upsert` we use subqueries for this.

`delta-rs` doesn't currently support subqueries in predicates (see this [ticket](https://github.com/delta-io/delta-rs/issues/2668) that I created on their repo).

dlt/common/normalizers/json/relational.py
```diff
@@ -505,6 +504,10 @@ def _get_row_id_type(
                 if x_row_version_col == DataItemNormalizer.C_DLT_ID:
                     return "row_hash"
         elif _r_lvl > 0:  # child table
+            merge_strategy = DataItemNormalizer._get_merge_strategy(schema, table_name)
+            if merge_strategy in ("upsert", "scd2"):
```
sh-rp, line 508: do I understand correctly, that one has to set merge strategy to "upsert" explicitely when using merge with delta tables on the filesystem destination? If that is the case, it would be nice to have a mechanism that does it automatically.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: I don't think it is 100% necessary to have a separate merge job, but processing all table chain jobs at once maybe yes. This is kind of there already in my loader changes, so I think you can have a look at the new structure once it is done and figure it out.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -123,16 +124,51 @@ def write(self) -> None:
...
+            self.table["write_disposition"] == "merge"
+            and (
+                dt := try_get_deltatable(
+                    self.client.make_remote_uri(self.make_remote_path()),
+                    storage_options=_deltalake_storage_options(self.client.config),
+                )
+            )
+            is not None
+        ):
+            assert self.table["x-merge-strategy"] in self.client.capabilities.supported_merge_strategies  # type: ignore[typeddict-item]
```
sh-rp, line 142: is this just for dev purposes or do you want to leave this here? we might want to extend the mechanism in the loader to check for non-loadable jobs (similar to the way we check for incorrect fileformats); but maybe not for now.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -123,16 +124,51 @@ def write(self) -> None:
...
+
+            if self.table["x-merge-strategy"] == "upsert":  # type: ignore[typeddict-item]
+                if "parent" in self.table:
+                    unique_column = get_first_column_name_with_prop(self.table, "unique")
+                    predicate = f"target.{unique_column} = source.{unique_column}"
+                else:
+                    primary_keys = get_columns_names_with_prop(self.table, "primary_key")
```
sh-rp, line 149: we don't use the merge keys? or did you overlook this? we should have failing tests if you don't and the tests are set up correctly.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: just for my understanding: this first unique key will be the articifial row_hash created further above, right?

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -123,16 +124,51 @@ def write(self) -> None:
...
+                    .when_not_matched_insert_all()
+                )
+
+                qry.execute()
+
+        else:
+            write_delta_table(
+                path=self.client.make_remote_uri(self.make_remote_path()),
+                data=pa.dataset.dataset(file_paths),
+                write_disposition=self.table["write_disposition"],
```
sh-rp, line 169: yeah, here we just fallback to append for different merge strategies. so i think upsert should be autoselected. not 100% sure how to do that in a nice way though.

tests/load/pipeline/test_merge_disposition.py
sh-rp: oh wow, thanks for catching this!!

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -1059,34 +1202,32 @@ def r():
...
+        default_sql_configs=True,
+        local_filesystem_configs=True,
+        supports_merge=True,
+        subset=["postgres", "filesystem"],  # test one SQL and one non-SQL destination
+    ),
+    ids=lambda x: x.name,
+)
+def test_upsert_merge_strategy_config(destination_config: DestinationTestConfiguration) -> None:
+    if destination_config.destination == "filesystem":
+        # TODO: implement validation and remove this test exception
```
sh-rp, line 1218: what do you mean with "validation"?

sh-rp:
@jorritsandbrink this is already looking quite good. I have a question about the tests though. Shouldn't we be able to make most tests in "/Volumes/dlthub/dlt/tests/load/pipeline/test_merge_disposition.py" work with delta tables? I can see that you are not using the merge keys (at least it seems to be this way) and those tests would've caught that.

dlt/destinations/impl/filesystem/filesystem.py
jorritsandbrink: The `upsert` merge strategy does not support merge keys. It requires a primary key.

This [discussion](https://github.com/dlt-hub/dlt/pull/1294#issuecomment-2082372128) explains why.

tests/load/pipeline/test_merge_disposition.py
jorritsandbrink: Validation of user-provided configuration.

For example, we want to raise if `upsert` is chosen as merge strategy, but no `primary_key` has been defined.

For SQL destinations this happens in `verify_sql_job_client_schema`. We need similar checks for the `delta` route. Ideally we use the same code for both.

dlt/destinations/impl/filesystem/filesystem.py
jorritsandbrink: Ultimately this should be removed.

This check happens in `_verify_destination_capabilities` for the other destinations, but `filesystem` is exempted there because it has [this](https://github.com/dlt-hub/dlt/pull/1600#discussion_r1680069255) (IMO undesired) behavior.

dlt/destinations/impl/filesystem/filesystem.py
jorritsandbrink: I'd not autoselect `upsert`.

`delete-insert` is our default merge strategy. I think it's good to have the same default for all destinations.

docs/website/docs/general-usage/incremental-loading.md
```diff
@@ -391,6 +392,62 @@ must be unique for a root table. We are working to allow `updated_at` style trac
 column in the root table to stamp changes in nested data.
 * `merge_key(s)` are (for now) ignored.
 
+### `upsert` strategy
+
+:::caution
+The `upsert` merge strategy is currently only supported for these destinations:
+- `postgres`
```
sh-rp, line 399: you could consider mentioning this on the three destination pages.

jorritsandbrink: `upsert` will soon be supported for all destinations. I think it's better to leave this here for now and remove the `caution` block when generic support has been implemented.

dlt/destinations/impl/filesystem/filesystem.py
jorritsandbrink: The first `unique` key resolves to the `C_DLT_ID` column, which contains a key created by `_get_child_row_hash`.

https://github.com/dlt-hub/dlt/blob/e4d912eac07648cf713d6eabc5900a0a77ebb267/dlt/common/normalizers/json/relational.py#L506

Issue #1593, opened 2024-07-15
Rest Client: allow setting of request kwargs (stream, verify, proxies, cert, timeout, allow_redirects) in the `paginate()` method


willi-mueller:
### Feature description

The REST client should allow setting all request kwargs also in the methods `paginate()`, `get()`, and `post()` methods.

1. stream responses:
```py
RESTClient(base_url="https://httpbin.org/").get(path="stream/2)", stream=True)
```

2. Verify the SSL certificate of the server with a trusted certificate from the local file system
```py
RESTClient(base_url="https://github.com").get(path="", verify='/path/to/certfile')
```
Currently, it is possible to set the `REQUESTS_CA_BUNDLE` environment variable.

3. Allow to pass requests through a proxy (might be useful for enterprise deployments where https traffic from the private network is passed through a proxy).
```py
RESTClient(base_url="https://httpbin.org").get(path="json", proxies={'http': 'http://10.10.1.10:3128', 'https': 'http://10.10.1.10:1080'})
```

4. Use client-side certificates
```py
RESTClient(base_url="https://httpbin.org").get(path="json",  cert=('/path/client.cert', '/path/client.key'))
```

5. Disable allowing to redirect
```py
RESTClient(base_url="https://httpbin.org/").get(path="redirect-to", params={"url": "https://httpbin.org/json"}, allow_redirects=False)"
```

### Are you a dlt user?

Yes, I'm already a dlt user.

### Proposed solution

Copying here, what @TyrantKingBen proposed because it seems to work well!

1. add `**kwargs` to `RESTClient.paginate()`
2. modify`_send_request()`:
```py
 def _send_request(self, request: Request, **kwargs: Any) -> Response:
        logger.info(
            f"Making {request.method.upper()} request to {request.url}"
            f" with params={request.params}, json={request.json}"
        )

        prepared_request = self.session.prepare_request(request)

        send_kwargs = self.session.merge_environment_settings(
            prepared_request.url,
            kwargs.pop("proxies", {}),
            kwargs.pop("stream", None),
            kwargs.pop("verify", None),
            kwargs.pop("cert", None),
        )

        send_kwargs.update(**kwargs)

        return self.session.send(prepared_request, **send_kwargs)
```

  4. modify `request()`:
  ```py
 def request(self, path: str = "", method: HTTPMethod = "GET", **kwargs: Any) -> Response:
        prepared_request = self._create_request(
            path=path,
            method=method,
            params=kwargs.pop("params", None),
[...]

PR #1609, opened 2024-07-18
Feat/1593 rest_client: allow setting of request kwargs

willi-mueller:
### Related Issues

- Resolves https://github.com/dlt-hub/dlt/issues/1593

### DIFF
--- a/dlt/sources/helpers/rest_client/client.py
+++ b/dlt/sources/helpers/rest_client/client.py
@@ -116,7 +116,7 @@ class RESTClient:
             hooks=hooks,
         )
 
-    def _send_request(self, request: Request) -> Response:
+    def _send_request(self, request: Request, **kwargs: Any) -> Response:
         logger.info(
             f"Making {request.method.upper()} request to {request.url}"
             f" with params={request.params}, json={request.json}"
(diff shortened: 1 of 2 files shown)


tests/sources/helpers/rest_client/test_client.py
```diff
@@ -430,3 +430,33 @@ def test_post_json_body_without_params(self, rest_client) -> None:
...
+    def test_request_kwargs(self, mocker) -> None:
+        rest_client = RESTClient(
+            base_url="https://api.example.com",
+            session=Client().session,
+        )
+
+        mocked_send = mocker.Mock()
+        rest_client.session.send = mocked_send  # type: ignore[method-assign]
+
+        rest_client.get(
```
willi-mueller, line 443: @burnash when I call `rest_client.paginate()` here then the test fails because it returns immediately from `paginate()` with `None`. Can you understand why?

dlt/sources/helpers/rest_client/client.py
```diff
@@ -170,6 +179,9 @@ def paginate(
             hooks (Optional[Hooks]): Hooks to modify request/response objects. Note that
                 when hooks are not provided, the default behavior is to raise an exception
                 on error status codes.
+            **kwargs (Any): Optional arguments to that the request library accepts, such as
```
burnash, line 182:
```suggestion
            **kwargs (Any): Optional arguments to that the Request library accepts, such as
```

burnash (dismissed):

PR #1617, opened 2024-07-19
Add empty source handling for `delta` table format on `filesystem` destination

### LINKED ISSUE #1613: Can't load empty table when using `delta` table format on `filesystem` destination
### dlt version

0.5.2a0

### Describe the problem

An error is thrown when loading an empty Arrow table when using the `delta` table format on the `filesystem` destination:

```sh
PipelineStepFailed: Pipeline execution failed at stage load when processing package 1721321020.398795 with exception:

<class '_internal.DeltaError'>
Generic error: No data source supplied to write command.
```

Using `dlt.mark.materialize_table_schema()` leads to the same error. Related Slack discussion: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1720776803670029?thread_ts=1720526414.546919&cid=C04DQA7JJN6

### Expected behavior

- no error
- if Delta table does not exist yet (i.e. it's the first run), an empty Delta table with schema from Arrow table or column hints (in case of  `dlt.mark.materialize_table_schema()`) is created
- if Delta table already exists, nothing happens

### Steps to reproduce

```py
import dlt
import pyarrow as pa

@dlt.resource(table_format="delta")
def my_resource():
    empty_arrow_table = pa.schema([
        pa.field('foo', pa.int64()),
        pa.field('bar', pa.string())
    ]).empty_table()

    yield empty_arrow_table

pipeline = dlt.pipeline(
    destination=filesystem("file://_storage")
)

pipeline.run(my_resource())
```

### Operating system

Windows

### Runtime environment

Local

### Python version

3.8

### dlt data source

_No response_

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

jorritsandbrink:
### Description
This PR adds support for the "empty source" case for `delta` table format on `filesystem` destination.

Specifically, it enables:
- use of resources that yield empty Arrow tables
- use of  `dlt.mark.materialize_table_schema()`

These cases need explicit handling because `delta-rs` throws errors in case of empty tables/datasets: https://github.com/delta-io/delta-rs/issues/2686

### Related Issues

- Closes #1613

### DIFF
--- a/dlt/destinations/impl/filesystem/filesystem.py
+++ b/dlt/destinations/impl/filesystem/filesystem.py
@@ -124,33 +124,33 @@ class DeltaLoadFilesystemJob(NewReferenceJob):
         from dlt.common.libs.deltalake import (
             DeltaTable,
             write_delta_table,
-            ensure_delta_compatible_arrow_table,
+            ensure_delta_compatible_arrow_schema,
             _deltalake_storage_options,
             try_get_deltatable,
         )
 
-        # create Arrow table from Parquet files
-        # this eagerly materializes table in memory
-        # ideally, we'd use an Arrow dataset everywhere, but we need to
-        # explicitly check if there is data
-        # (https://github.com/delta-io/delta-rs/issues/2686)
+        # create Arrow dataset from Parquet files
         file_paths = [job.file_path for job in self.table_jobs]
-        arrow_table = pa.dataset.dataset(file_paths).to_table()
+        arrow_ds = pa.dataset.dataset(file_paths)
 
         # create Delta table object
         dt_path = self.client.make_remote_uri(self.make_remote_path())
         storage_options = _deltalake_storage_options(self.client.config)
         dt = try_get_deltatable(dt_path, storage_options=storage_options)
 
-        if arrow_table.num_rows == 0:
+        # explicitly check if there is data
+        # (https://github.com/delta-io/delta-rs/issues/2686)
+        if arrow_ds.head(1).num_rows == 0:
             if dt is None:
                 # create new empty Delta table with schema from Arrow table
                 DeltaTable.create(
                     table_uri=dt_path,
-                    schema=ensure_delta_compatible_arrow_table(arrow_table).schema,
+                    schema=ensure_delta_compatible_arrow_schema(arrow_ds.schema),
                     mode="overwrite",
                 )
             return
+
+        arrow_rbr = arrow_ds.scanner().to_reader()  # RecordBatchReader
 
         if self.table["write_disposition"] == "merge" and dt is not None:
             assert self.table["x-merge-strategy"] in self.client.capabilities.supported_merge_strategies  # type: ignore[typeddict-item]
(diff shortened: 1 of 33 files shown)


dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -122,23 +122,37 @@ def __init__(
...
+        # (https://github.com/delta-io/delta-rs/issues/2686)
         file_paths = [job.file_path for job in self.table_jobs]
-
-        if (
-            self.table["write_disposition"] == "merge"
-            and (
-                dt := try_get_deltatable(
-                    self.client.make_remote_uri(self.make_remote_path()),
-                    storage_options=_deltalake_storage_options(self.client.config),
+        arrow_table = pa.dataset.dataset(file_paths).to_table()
```
sh-rp, line 138: on large loads this will probably fail or be super slow, no? there must be ways to cheat here, for example just load the first file and if there is something in there we know that we have rows and write the table the same way as before or something like that.

sh-rp (changes requested):
Thanks for the work on this! I think we should have another think about the mechanism which we use to determine wether a table is empty. If I understood you correctly then delta-rs uses some kind of multithreaded loading under the hood, so we kill this speed benefit when loading the full table eagerly into memory. I'm sure we can cheat here somehow. A few ideas:

* Only load the first file for a table and if there is something there we can use the old way of loading
* Work with the filesizes of the jobs. I'm not sure how predictable they are, but a 1mb parquet file very likely has data, right?
* Arrow might have some system to only load the first row or a parquet file or list of parquet files, that should be enough for peaking into the files to see if there is anything.

The main point is to avoid a situation where we have 10gb of files for one table and load them all into memory.

jorritsandbrink:
I've refactored the code to use `pyarrow`'s `Dataset` and `RecordBatchReader` to prevent loading all data into memory at once.

I had to upgrade `pyarrow` to `17.0.0`, which led to a dependency conflict with `pylance`:
image

I don't know how to elegantly work around this conflict. Poetry does not support dependency overriding: https://github.com/python-poetry/poetry/issues/697.

This problem will soon solve itself, because `pylance` has already loosened the `pyarrow` dependency recently: https://github.com/lancedb/lance/commit/ac3f75a17bc4ba0cd9d90729925b94f4158b8e7f.

This change hasn't been released yet. `pylance` releases pretty often, so I expect it will be released soon.

I disabled the `lancedb` dependency in `pyproject.toml` to be able to progress. Of course, the `lancedb`-dependent pipelines are now failing on CI, so this can't be merged like this.

@sh-rp can you:
- review
- suggest an approach: wait for `pylance` release or do something else

jorritsandbrink:
The part about `pyarrow` versioning in my comment above no longer applies. As discussed with @rudolfix on Slack, we don't upgrade `pyarrow` in `pyproject.toml`. Instead, we expect the user to run dlt in an environment with `pyarrow>=17.0.0.` when they're using the `delta` table format on `filesystem`. This requirement is asserted at runtime, and a `DependencyVersionException` is raised if not satisfied.

I had to extend the test setup a bit:
- `delta` tests are now marked as `needspyarrow17`
- `needspyarrow17` tests are automatically skipped if `pyarrow<17.0.0`
- a new github workflow `test_pyarrow17` explicitly installs `pyarrow==17.0.0` to make sure `needspyarrow17` tests run on CI

sh-rp (approved):

Issue #1615, opened 2024-07-19
User-Provided Text Splitters for Document Chunking


Pipboyguy:
### Feature description
Implement support for user-provided text splitters in dlt for flexible document chunking.

### Are you a dlt user?
Yes

### Use case
Users need custom control over document chunking for various document formats.

### Proposed solution
Add an option to specify a custom text splitter callback when configuring the resource.

### Related issues
https://github.com/dlt-hub/dlt/issues/1587

zilto:
Adding text splitters is not currently planned. We will add reference architectures in docs. 

I like the lightweight library [chonkie](https://docs.chonkie.ai/python-sdk/getting-started/introduction)

Issue #1533, opened 2024-07-01
Document Chunking Management


Pipboyguy:
### Are you a dlt user?

None

rudolfix:
@Pipboyguy the biggest challenge is to update the document in the database effectively (`merge` write disposition). if we chunk a document, should we yield many objects (documents) per chunk or we yield a single compound object with all the chunks inside?

for the latter we already solve this problem for sql databases (child table merge) so the same mechanism could be reused if vector database supports SQL (ie. pgvector).

for other vector databases we'd need to drop all the previous chunks and then re-insert all new chunks (or something more optimized) but I have some doubts if that is scalable?

any thoughts on that?

Pipboyguy:
@rudolfix Thank you for providing some specificity and color to the issue.

Here's a 

```python
import hashlib
import random
from typing import List, Generator

import dlt
from dlt.common.typing import DictStrAny


def get_md5_hash(data: str) -> str:
    return hashlib.md5(data.encode("utf-8")).hexdigest()


def mock_embed(dim: int = 10) -> str:
    # We return string instead of compound (list) type here to prevent
    # further normalization. this will be a vector type in target db.
    return str([random.uniform(0, 1) for _ in range(dim)])


def chunk_document(doc: str, chunk_size: int = 10) -> List[str]:
    return [doc[i : i + chunk_size] for i in range(0, len(doc), chunk_size)]


@dlt.resource(
    standalone=True,
    write_disposition="merge",
    merge_key=["doc_id", "chunk_hash"],
    table_name="document",
)
def documents(docs: List[DictStrAny]) -> Generator[DictStrAny, None, None]:
    for doc in docs:
        doc_id = doc["doc_id"]
        chunks = chunk_document(doc["text"])
        embeddings = [
            {
                "chunk_hash": get_md5_hash(chunk),
                "chunk_text": chunk,
                "embedding": mock_embed(),
            }
            for chunk in chunks
        ]
        yield {"doc_id": doc_id, "doc_text": doc["text"], "embeddings": embeddings}


pipeline = dlt.pipeline(
    pipeline_name="chunked_docs",
    destination="postgres",
    dataset_name="chunked_documents",  # dev_mode=True,
)

initial_docs = [
    {
        "text": "This is the first document. It contains some text that will be chunked and embedded. (I don't want "
        "to be seen in updated run's embedding chunk texts btw)",
        "doc_id": 1,
    },
    {
[...]

PR #1620, opened 2024-07-21
LanceDB - Remove Orphaned Chunks

### LINKED ISSUE #1587: Chunk and Embedding Management in LanceDB
### Feature description

Prevent orphaned records during updates and deletions, especially with parent-child table relationships to maintain referential integrity. This will enable chunking mechanisms to work seamlessly once documents are dropped or updated.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

When working with hierarchical data structures in LanceDB, it's crucial to maintain referential integrity between parent and child records. This is particularly important for scenarios like document chunking, where a large document (parent) is split into smaller chunks (children) for embedding and similarity searches. The main challenge is to update or delete parent records without orphaning their associated child records.

### Proposed solution

- Enhance merge operations to handle parent-child relationships effectively.
- Implement automatic removal of orphaned child records when parent records are updated or deleted.

### Related issues

- https://github.com/dlt-hub/dlt/issues/1533
- https://github.com/dlt-hub/dlt/issues/1615

Pipboyguy:
### Description

This PR lays the groundwork for managing chunked documents and their embeddings efficiently in LanceDB, focusing **merge with referential integrity**. This PR does not implement document chunking/splitting, which will be addressed in #1615.

- Automatically remove orphaned chunks when the parent document is updated or deleted.

### Related Issues

- Closes #1587
- Relates to #1533
- Relates to #1615


dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -190,6 +191,29 @@ def upload_batch(
             tbl.merge_insert(
                 id_field_name
             ).when_matched_update_all().when_not_matched_insert_all().execute(records)
+
+            # Remove orphaned parent IDs.
+            if parent_table_name:
+                try:
+                    parent_tbl = db_client.open_table(parent_table_name)
```
sh-rp, line 198: this will not work, there is no guarantee that the parent table is fully loaded (with all possible load jobs) at the point in time of the execution of the child table job. If we do it this way, we will need table chain followup jobs, similar to the way merge jobs are done. You can have a look at the default sql client for that.

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -190,6 +191,29 @@ def upload_batch(
...
+                    parent_tbl.checkout_latest()
+                except FileNotFoundError as e:
+                    raise DestinationTransientException(
+                        "Couldn't open lancedb database. Batch WILL BE RETRIED"
+                    ) from e
+
+                parent_ids = set(pc.unique(parent_tbl.to_arrow()["_dlt_id"]).to_pylist())
```
sh-rp, line 205: I wonder if this approach will work well for large dataset..

sh-rp:
Thanks for starting work on this. I am wondering if it would be possible to get this to work without relying on the parent tables so that adding a parent table is optional in this scenario. Theoretically it should be possible to use the dlt_load_id to discover outdated embeddings, but I have not thought it through 100%. Do you think the following is somehow possible assuming we are only using 1 table:

* Every "original" document has a unique ID which will be inherited as a column by all embeddings (it will just be sent as column from the resource but we need a hint for the schema to identify this column)
* We have this compound key created from the original document id and the row hash for doing the merge update the way you already do it (I think)
* After the full table is loaded, we run a follow up job, which will identify each "original" document id that has new data loaded during this run (we can find those with the dlt_load_id) and then will do a delete on all records that match these original document IDs not do NOT match the current load id.

Pipboyguy:
@sh-rp I think that can work! Just one thing:

> We have this compound key created from the original document id and the row hash

I've simply added the document id to the current merge keys:

https://github.com/dlt-hub/dlt/blob/a60737aeaa1dc84f6735c7e62afe4ca6d6ac5168/dlt/destinations/impl/lancedb/utils.py#L45-L48

I feel that if we maintain a separate merge key, this might lead to confusing behaviour, especially if the user also specifies a primary_key and merge key as well.

Thoughts?

Pipboyguy:
I'm having some trouble with scheduling the orphan removal jobs:

https://github.com/dlt-hub/dlt/blob/7f6f1cd0e0b3eb9a3cd1b6fcc3d46f9a63e75c4d/dlt/destinations/impl/lancedb/lancedb_client.py#L759

Any idea what's happening here? Need someone a bit more familiar with the Job Orchestrator

Pipboyguy:
#TODO: Use parquet instead of jsonl

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Fixed in `LanceDBRemoveOrphansJob`

Pipboyguy:
Made a performance improvement in schema updating: We don't materialize empty NULL tables in memory any more but use datafusion api. I think this engine will prove useful going forward, one perhaps we defer to this engine as intermediary between dlt and lancedb in future

https://github.com/dlt-hub/dlt/blob/7f245e2534c03e23e08db359be45125880bb881a/dlt/destinations/impl/lancedb/lancedb_client.py#L431

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Subqueries don't work so we can't use the datafusion engine here. It will have to be batched

Pipboyguy:
Parquet file format now only supported format. much faster to ingest and load

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -690,67 +691,177 @@ def complete_load(self, load_id: str) -> None:
...
+
+                    if orphaned_ids := child_ids - parent_ids:
+                        if len(orphaned_ids) > 1:
+                            child_table.delete(f"_dlt_parent_id IN {tuple(orphaned_ids)}")
+                        elif len(orphaned_ids) == 1:
+                            child_table.delete(f"_dlt_parent_id = '{orphaned_ids.pop()}'")
+
+                else:
+                    # Chunks and embeddings in the root table.
+                    document_id_field = get_columns_names_with_prop(table, DOCUMENT_ID_HINT)
```
rudolfix, line 835: in adapter you mention that DOCUMENT ID HINT will be appended to primary key but here I see something else...

IMO what we want here
primary_key = (DOCUMENT ID, CHUNK ID)

so what adapter must make sure is: DOCUMENT ID is set on a subset of PRIMARY KEY

for example primary_key = ("doc_id", "chunk_id"), document_id=("doc_id",) leaving "chunk_id" as chunk_id

you can force both chunk and doc ids to be single columns if that simplifies your code

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -690,67 +691,177 @@ def complete_load(self, load_id: str) -> None:
...
+                            " resource when using merge disposition."
+                        )
+
+                    # If document ID is defined, we use this as the sole grouping key to identify stale chunks,
+                    # else fallback to the compound `id_field_name`.
+                    grouping_key = document_id_field or id_field_name
+                    grouping_key = (
+                        grouping_key if isinstance(grouping_key, list) else [grouping_key]
+                    )
+                    child_table_arrow: pa.Table = child_table.to_lance().to_table(
```
rudolfix, line 848: please never use constant names for columns. you have normalized names in "schema" instance
schema.dlt_id_name (AFAIK) look for it

rudolfix (commented):
@Pipboyguy 
code structure is good now, however we should really use the staging dataset to do the merging. IMO the code will be way simpler and faster.

currently we load huge datasets into memory. I hope data fusion will let us join and delete from tables properly (?). duckdb definitely could do it!

my take: merge your staging dataset PR into this one and let's review that together.

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Changed as requested, now reads from normalizer

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Turns out that LanceDB currently does not have built-in support for primary keys: https://github.com/lancedb/lancedb/issues/1120
 
 The Lance merge methods also [claims to support compound merge keys with an Iterable[str]](https://lancedb.github.io/lance/api/python/lance.html#lance.dataset.LanceDataset.merge_insert)

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -696,67 +694,149 @@ def complete_load(self, load_id: str) -> None:
...
+        ]
+
+        for job in table_lineage:
+            target_is_root_table = "parent" not in job.table_schema
+            fq_table_name = self._job_client.make_qualified_table_name(job.table_name)
+            file_path = job.file_path
+            with FileStorage.open_zipsafe_ro(file_path, mode="rb") as f:
+                payload_arrow_table: pa.Table = pq.read_table(f)
+
+            if target_is_root_table:
```
rudolfix, line 818: OK this is really good but IMO there's a mistake when deleting nested tables:
- to delete orphan from root table we use merge key of the root table (good)
- to delete from child table we use all `dlt_ids` that you can find in root table - not in child table why? because if you remove (1) all nested elements in root table you will not delete all orphans (2) or you remove one row from root table you also won't find the proper root key

please write tests for that.

also could you take taxi dataset and try to merge all rows (where IN clause will be insane long). I wonder what happens: will that be o(n^2) operation or they will do a hash table on IN clause

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -696,67 +694,149 @@ def complete_load(self, load_id: str) -> None:
...
+    def __init__(
+        self,
+        file_path: str,
+    ) -> None:
+        super().__init__(file_path)
+        self._job_client: "LanceDBClient" = None
+        self.references = ReferenceFollowupJobRequest.resolve_references(file_path)
+
+    def run(self) -> None:
+        dlt_load_id = self._schema.data_item_normalizer.C_DLT_LOAD_ID  # type: ignore[attr-defined]
```
rudolfix, line 795: those are not normalized! please use `c_dlt_load_id` (small caps). apply that to everywhere!!

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -192,11 +206,14 @@ def upload_batch(
         elif write_disposition == "replace":
             tbl.add(records, mode="overwrite")
```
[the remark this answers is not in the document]
Pipboyguy: Fixed

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Fixed

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Setting a scalar index does seem to offers constant time complexity. Clearly this is worth investigating for future PRs.

rudolfix (approved):
LGTM!
reviewed the code and added a few small fixes. overall procedure is good and thanks for the tests!!

PR #1628, opened 2024-07-23
Enable `upsert` merge strategy for more SQL destinations

### LINKED ISSUE #1621: Extend `upsert` support to other SQL destinations
### Feature description

The `upsert` merge strategy is currently supported for:
- `postgres` and `snowflake` SQL destinations 
- `delta` table format on `filesystem` destination

It would be good to make `upsert` available on other SQL destinations (`bigquery`, `databricks`, `mssql`, ...).

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

_No response_

### Proposed solution

_No response_

### Related issues

_No response_

jorritsandbrink:
### Description
Currently, only `postgres` and `snowflake` support the `upsert` merge strategy.

This PR enables `upsert` for:
- `athena`
- `bigquery`
- `databricks`
- `mssql`

### Related Issues

Closes #1621 

### Additional context

The other SQL destinations are not enabled because they either:
- don't support `MERGE INTO`
- support `MERGE INTO` with limitations/complications

We could enable some more destinations if we're okay with partial support: basic upsert works, but `hard_delete` hint does not:
- `redshift` ➜ works with current SQL logic
- `dremio` ➜ perhaps works with current SQL logic (haven't been able to test it locally)
- `duckdb` /  `motherduck` ➜ requires custom SQL logic 

Details:

- `clickhouse`
  - does not support `MERGE INTO` or similar functionality
- `duckdb` / `motherduck`
  - does not support `MERGE INTO`
  - It supports the [upsert](https://duckdb.org/docs/sql/statements/insert.html#do-update-clause-upsert) pattern with `INSERT INTO ... ON CONFLICT DO UPDATE`, but this can't handle the `hard_delete` hint.
- `dremio`
  - does not support `WHEN MATCHED THEN DELETE` ➜ `hard_delete` hint can't be handled 
- `synapse`
  - [requires](https://learn.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=azure-sqldw-latest&preserve-view=true#azure-synapse-analytics-considerations) the target table to have `HASH` distribution when using a `WHEN NOT MATCHED` clause
  - I tried making it work by configuring tables with `HASH` distribution when `upsert` is chosen, but this comes with its own set of challenges related to setting/choosing the distribution column(s).
- `redshift`
  - only supports having one `WHEN MATCHED` clause in the `MERGE` statement (makes supporting `hard_delete` hint difficult)
  - I tried working around this limitation but couldn't find an elegant solution.
  - (`redshift` also doesn't support using an alias for the target table in a `MERGE` statement. I managed to work around this with some conditional logic, so this isn't a blocking limitation.)

### DIFF
--- a/dlt/destinations/impl/athena/factory.py
+++ b/dlt/destinations/impl/athena/factory.py
@@ -46,7 +46,7 @@ class athena(Destination[AthenaClientConfiguration, "AthenaClient"]):
         caps.schema_supports_numeric_precision = False
         caps.timestamp_precision = 3
         caps.supports_truncate_command = False
-        caps.supported_merge_strategies = ["delete-insert", "scd2"]
+        caps.supported_merge_strategies = ["delete-insert", "upsert", "scd2"]
         return caps
 
     @property
--- a/dlt/destinations/impl/bigquery/factory.py
+++ b/dlt/destinations/impl/bigquery/factory.py
@@ -42,7 +42,7 @@ class bigquery(Destination[BigQueryClientConfiguration, "BigQueryClient"]):
         caps.supports_ddl_transactions = False
         caps.supports_clone_table = True
         caps.schema_supports_numeric_precision = False  # no precision information in BigQuery
-        caps.supported_merge_strategies = ["delete-insert", "scd2"]
+        caps.supported_merge_strategies = ["delete-insert", "upsert", "scd2"]
 
         return caps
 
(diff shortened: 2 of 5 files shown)


rudolfix (approved):
LGTM! we need to investigate this logbook problem. I bet it is coming from dbt in dev deps which uses it

rudolfix:
@jorritsandbrink looks like the logbook problem is gone

Issue #1567, opened 2024-07-09
improve `progress` in normalize and load steps


rudolfix:
**Background**
Progress reporting in `normalize` and `load` steps are far from perfect.
1. in `normalize` we report progress on file level but that only is updated when a worker process is finished
2. in `load` the reported metrics do not survive restarts (see #853 )

**Tasks**
Step1. fix `normalize`:
- use metrics collected in `extract` (per job and resource) to correctly report processed row per resource (where we have total number of records as well)
- right now there's no communication between worker and main process. but we need to start reporting metrics back. so we need to update 

Step 2. Fix `load`:
- see #853 use package state to track the elapsed times (task created, start, stop of job)
- we are interested in following metrics to be displayed: jobs processed, average elapsed time, average lag (from job created to job started)

**Implementation**
1. you'll need to use package state to store extract metrics (ExtractInfo) and normalize metrics
2. if those elements are not present in the state you must fallback gracefully ie. reporting only the progress of the files. the job processing must be plain: if there are files they will be processed even if state is not present

IlyaFaer:
It seems to me, to make progress tracking work for cases, when we normalize data, extracted during the previous run, we need to use the pipeline state somehow. Trying to do it...

rudolfix:
implemented

PR #1639, opened 2024-07-26
feat(pipeline): implement normalize progress tracking for JSON normalizer

IlyaFaer:
Towards #1567

Implements item 1 for `JsonLItemsNormalizer`

### DIFF
--- a/dlt/normalize/items_normalizers.py
+++ b/dlt/normalize/items_normalizers.py
@@ -8,6 +8,7 @@ from dlt.common.data_writers.writers import ArrowToObjectAdapter
 from dlt.common.json import custom_pua_decode, may_have_pua
 from dlt.common.normalizers.json.relational import DataItemNormalizer as RelationalNormalizer
 from dlt.common.runtime import signals
+from dlt.common.runtime.collector import Collector, NULL_COLLECTOR
 from dlt.common.schema.typing import TSchemaEvolutionMode, TTableSchemaColumns, TSchemaContractDict
 from dlt.common.schema.utils import has_table_seen_data
 from dlt.common.storages import NormalizeStorage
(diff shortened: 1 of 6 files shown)


dlt/normalize/items_normalizers.py
```diff
@@ -83,9 +93,15 @@ def _normalize_chunk(
         schema_name = schema.name
         normalize_data_fun = self.schema.normalize_data_item
 
+        total = self.extract_info.total_rows_count if self.extract_info else None
+
+        import time
+
         for item in items:
+            time.sleep(0.7)
```
IlyaFaer, line 101: I left his testing stuff for now, so you'll be able to see that it works

dlt/normalize/items_normalizers.py
```diff
@@ -83,9 +93,15 @@ def _normalize_chunk(
...
             items_gen = normalize_data_fun(item, self.load_id, root_table_name)
             try:
+                self.collector.update("Items", total=total)
```
IlyaFaer, line 104: For every row we update the collector

IlyaFaer:
For the case when we're first extracting data and then normalizing, it works fine

image

What I don't like is that we have to through collector and extract data through a bunch of methods and classes, but for now I don't see any solution, as the code is chained.

Thinking about how to test it... Seems to me something cunning is needed.

dlt/normalize/items_normalizers.py
```diff
@@ -83,9 +87,13 @@ def _normalize_chunk(
         schema_name = schema.name
         normalize_data_fun = self.schema.normalize_data_item
 
+        # import time
+
         for item in items:
+            # time.sleep(0.7)
```
IlyaFaer, line 93: Use it for testing

IlyaFaer:
@rudolfix, hm-m, looks like pipeline state is a better solution - allows to cover both cases, when the data was just extracted, and when it's been extracted on the previous run.

dlt/pipeline/pipeline.py
```diff
@@ -445,7 +445,12 @@ def extract(
                 )
                 # commit load packages with state
                 extract_step.commit_packages()
-                return self._get_step_info(extract_step)
+
+                info = self._get_step_info(extract_step)
+                state = self._container[StateInjectableContext].state
+                state["_local"]["_last_extracted_count"] = info.total_rows_count
```
IlyaFaer, line 451: Figuring out if it's possible to do this in a less hacky way without breaking anything

rudolfix:
collected metrics will not go through process boundary, we'll need different approach here

Issue #1629, opened 2024-07-23
allow to bind config fields to env variables for toml/yaml config providers 


rudolfix:
**Background**
`dlt` often works in an environment where some secret/config values are already defined as env variables. Those variables will not have keys supported by `dlt` env config provider. This requires the user to map them (in code) via

or read them and pass them explicitly

Here we want to support a simple mapping mechanism:
```toml
[sources.zendesk.credentials]
subdomain = "{{ ZENDESK_SUBDOMAIN }}"
```

or

```yaml
snowflake:
    stage_name: {{SNOWFLAKE_STAGE}}
    connection_timeout: {{CONNECTION_TIMEOUT}}
```
using basic jinja template syntax, we'll replace the `{{ ENV_VAR_NAME }}` with value of env variable `ENV_VAR_NAME`

**Requirements**
1. * [ ] after loading the file (as text) or obtaining it any other way (ie from the vault provider) replace all placeholders with env variables of specified name. Then parse the file as usual
2. * [ ] for missing env variables, provide descriptive exception message (name of missing variable, number of the line in file, text of the line and explanation what dlt was doing and why)
3. * [ ] document the behavior somewhere here https://dlthub.com/docs/general-usage/credentials/configuration mind that this PR will change it: https://github.com/dlt-hub/dlt/pull/1508

PR #1642, opened 2024-07-27
Feat/1596 adds custom docs config provider

### LINKED ISSUE #1596: allow `yaml` as alternative to `toml` in config/secrets
**Background**
`dlt` stores "project" (`.dlt`) configuration in `toml` files. Here we want to support `yaml` as well. `dlt` should probe for possible files and use proper parser to read them. From the user perspective it is enough if ie. `secrets.yaml` or `secrets.yml` be present in `.dlt` folder to use it. No other configuration should be required.

Example: `.dlt/secrets.toml` and it's alternative `.dlt/secrets.yaml`:
```toml

destination.postgres.credentials="postgresql://loader:loader@localhost:5432/dlt_data"

[sources.zendesk.credentials]
subdomain = "subdomain" # Zendesk subdomain
email = "set me up" # Email used to login to Zendesk
password = "set me up" # Password for Zendesk account
token = "set me up" # For API token auth
oauth_token = "set me up" # Use Zendesk support OAuth token or Zendesk chat OAuth token

[destination.athena]
query_result_bucket="s3://dlt-ci-test-bucket"
[destination.athena.credentials]
aws_access_key_id="AK"
aws_secret_access_key="b+secret"

[destination.snowflake]
stage_name="PUBLIC.my_s3_stage"
connection_timeout=60.4
[destination.snowflake.csv_format]
delimiter="|"
include_header=false
on_error_continue=true
[destination.snowflake.credentials]
query_tag='{{"source":"{source}", "resource":"{resource}", "table": "{table}", "load_id":"{load_id}", "pipeline_name":"{pipeline_name}"}}'


[data_types]
datetime=1979-05-27T07:32:00-08:00
```

```yaml
destination:
  postgres:
    credentials: postgresql://loader:loader@localhost:5432/dlt_data
  athena:
    query_result_bucket: s3://dlt-ci-test-bucket
    credentials:
      aws_access_key_id: AK
      aws_secret_access_key: b+secret
  snowflake:
    stage_name: PUBLIC.my_s3_stage
    connection_timeout: 60.4
    csv_format:
      delimiter: '|'
      include_header: false
      on_error_continue: true
    credentials:
      query_tag: '{{"source":"{source}", "resource":"{resource}", "table": "{table}",
        "load_id":"{load_id}", "pipeline_name":"{pipeline_name}"}}'
sources:
  zendesk:
    credentials:
      subdomain: subdomain
      email: set me up
      password: set me up
      token: set me up
      oauth_token: set me up
data_types:
  datetime: 1979-05-27 07:32:00-08:00
```
We expect many users will prefer `yaml` which is shorter more opinionated regarding the layout

**Requirements**
1. We have many config providers based on `toml` format. The basic file provider, string providers, vaults and airflow providers that work with toml fragments. Luckily the `toml` part is (almost) abstracted away. We want to change the base in a way so all our downstream config providers also support yaml loading.
2. For toml we use `tomlkit` which preserves comments. This is done thanks to wrapping all Python types in tomlkit object model. If config was loaded with toml, we want to still preserve the comments to be able to write it back.
3. Both `get_value` and `set_value` must be implemented for toml and yaml
4. File provider should probe `toml`, `yaml` and `yml` files and load just one of them. If there are more files with same name but different extension, a warning message (use warn module, not logger which may not yet be configured) must explain the situation and which file got loaded
5. It would be nice if we can write both toml and yaml files back but yaml is not required. `toml` must work like it should before. We do not care for comments in yaml at this moment.

**Documentation**
1. We must document the choice in config docs
2. As a separate ticket: all `toml` snippets should become tabs with both `toml` and `yaml` representation. This is optional

rudolfix:
### Description
fixes #1596  and indirectly #1629 

allows to instantiate custom config provider with user supplied loading function which may process any file format from any location

### DIFF
--- a/dlt/common/configuration/providers/toml.py
+++ b/dlt/common/configuration/providers/toml.py
@@ -37,10 +37,6 @@ class BaseDocProvider(ConfigProvider):
                 if not isinstance(node, dict):
                     raise KeyError(k)
                 node = node[k]
-            assert not isinstance(
-                node, (TOMLContainer, TOMLItem)
-            ), f"toml injected in {full_path}: {node.as_string()}"
-            # rv = node.unwrap() if isinstance(node, (TOMLContainer, TOMLItem)) else node
             return node, full_key
         except KeyError:
             return None, full_key
--- a/dlt/common/configuration/providers/toml.py
+++ b/dlt/common/configuration/providers/toml.py
@@ -1,9 +1,10 @@
 import os
 import tomlkit
 import yaml
+import functools
 from tomlkit.items import Item as TOMLItem
 from tomlkit.container import Container as TOMLContainer
-from typing import Any, Dict, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, Optional, Tuple, Type
 
 from dlt.common.configuration.paths import get_dlt_settings_dir, get_dlt_data_dir
 from dlt.common.configuration.utils import auto_cast, auto_config_fragment
@@ -135,9 +136,49 @@ class StringTomlProvider(BaseDocProvider):
         return "memory"
 
 
-class FileDocProvider(BaseDocProvider):
+class CustomLoaderDocProvider(BaseDocProvider):
     def __init__(
-        self, file_name: str, project_dir: str = None, add_global_config: bool = False
+        self, name: str, loader: Callable[[], Dict[str, Any]], supports_secrets: bool = True
+    ) -> None:
+        """Provider that calls `loader` function to get a Python dict with config/secret values to be queried.
+        The `loader` function typically loads a string (ie. from file), parses it (ie. as toml or yaml), does additional
+        processing and returns a Python dict to be queried.
+
+        Instance of CustomLoaderDocProvider must be registered for the returned dict to be used to resolve config values.
+        >>> import dlt
+        >>> dlt.config.register_provider(provider)
+
+        Args:
+            name(str): name of the provider that will be visible ie. in exceptions
+            loader(Callable[[], Dict[str, Any]]): user-supplied function that will load the document with config/secret values
+            supports_secrets(bool): allows to store secret values in this provider
+
+        """
+        self._name = name
+        self._supports_secrets = supports_secrets
+        super().__init__(loader())
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def supports_secrets(self) -> bool:
+        return self._supports_secrets
+
+    @property
+    def is_writable(self) -> bool:
+        return True
+
+
+class ProjectDocProvider(CustomLoaderDocProvider):
+    def __init__(
+        self,
+        name: str,
+        supports_secrets: bool,
+        file_name: str,
+        project_dir: str = None,
+        add_global_config: bool = False,
     ) -> None:
         """Creates config provider from a `toml` file
 
@@ -147,6 +188,8 @@ class FileDocProvider(BaseDocProvider):
         If none of the files exist, an empty provider is created.
 
         Args:
+            name(str): name of the provider when registering in context
+            supports_secrets(bool): allows to store secret values in this provider
             file_name (str): The name of `toml` file to load
             project_dir (str, optional): The location of `file_name`. If not specified, defaults to $cwd/.dlt
             add_global_config (bool, optional): Looks for `file_name` in `dlt` home directory which in most cases is $HOME/.dlt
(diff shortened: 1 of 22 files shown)


rudolfix:
@VioletM here's preview of the example https://deploy-preview-1642--dlt-hub-docs.netlify.app/docs/examples/custom_config_provider

docs/website/docs/general-usage/credentials/config_providers.md
```diff
@@ -33,6 +33,11 @@ providers.
 4. **Default Argument Values**: These are the values specified in the function's signature.
    They have the lowest priority in the provider hierarchy.
 
+:::tip
+Implement and use your own config provider by just providing loader function. See [example](../../examples/custom_config_provider) for `yaml` based config that
+supports switchable profiles.
+:::
+
```
sh-rp, line 40: I realize that the latter is already covered by the example, I'd still add two super simple examples of the BaseDocProvider and the CustomLoaderDocProvider here, but this could be done in a followup PR.

docs/examples/custom_config_provider/custom_config_provider.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # dlt standard providers work at this point (we have profile name in config.toml)
+    profile_name = dlt.config["dlt_config_profile_name"]
+
+    # instantiate custom provider using `prod` profile
+    # NOTE: all placeholders (ie. GITHUB_API_KEY) will be evaluated in next line!
+    provider = CustomLoaderDocProvider("profiles", functools.partial(loader, profile_name))
+    # register provider, it will be added as the last one in chain
+    dlt.config.register_provider(provider)
+
```
sh-rp, line 81: I'd add a comment here that this is the place to run the pipeline (imho examples should always be thought of in the context of a pipeline execution)

dlt/common/configuration/providers/toml.py
```diff
@@ -113,124 +132,49 @@ def name(self) -> str:
...
-        self._update_from_vault(SECRETS_TOML_KEY, None, AnyType, None, ())
-
-    def get_value(
-        self, key: str, hint: type, pipeline_name: str, *sections: str
-    ) -> Tuple[Optional[Any], str]:
-        full_key = self.get_key_name(key, pipeline_name, *sections)
+        """
+        self._name = name
+        self._supports_secrets = supports_secrets
+        super().__init__(loader())
```
sh-rp, line 155: What is the use of this CustomLoaderProvider if it is not lazily evaluated on first access of the dict?

docs/examples/custom_config_provider/custom_config_provider.py
```diff
@@ -0,0 +1,90 @@
...
+if __name__ == "__main__":
+    # mock env variables to fill placeholders in profiles.yaml
+    os.environ["GITHUB_API_KEY"] = "secret_key"  # mock expected var
+
+    # dlt standard providers work at this point (we have profile name in config.toml)
+    profile_name = dlt.config["dlt_config_profile_name"]
+
+    # instantiate custom provider using `prod` profile
+    # NOTE: all placeholders (ie. GITHUB_API_KEY) will be evaluated in next line!
+    provider = CustomLoaderDocProvider("profiles", functools.partial(loader, profile_name))
```
sh-rp, line 78: I would not use stuff like partials in examples if it is not absolutely necessary. While I agree the code looks nicer this way, my intuition says this makes it harder to understand what is going on for less experienced devs. It's up to you to change it :)

sh-rp (changes requested):
Super useful PR imho. I have a few comments, none of which are super important for this release, but the docs could be better / more helpful.

dlt/common/configuration/providers/toml.py
rudolfix: heh you are right. in this implementation it does not make sense. BUT maybe we want to change it to lazy evaluation ie. on first use. may be necessary ie. on Airflow. so let's keep it like that!

docs/examples/custom_config_provider/custom_config_provider.py
```diff
@@ -0,0 +1,90 @@
...
+    # register provider, it will be added as the last one in chain
```
VioletM, line 79: Last one in the chain means, that if they have `secrets.toml` or env variable with the same key it will override the explicitly registered `CustomLoaderDocProvider` provider, right?

VioletM: It just seems more intuitive for me that the custom provider I wrote to be the first in chain. But I guess it's OK, as long as it's stated in the docs

VioletM (approved):
This looks good! 

I'd say we should add more docs and examples for `CustomLoaderDocProvider`. But it can be done in the separate PR later after we'll get the first feedback :)

Issue #1474, opened 2024-06-03
Incorrect RESTAPIConfig can print secrets to stdout


b-per:
### dlt version

0.4.12

### Source name

rest_api

### Describe the problem

When setting up `auth` in `RESTAPIConfig`, some incorrect config can write secrets to stdout

```
...
    raise DictValidationException(
dlt.common.exceptions.DictValidationException: In path ./client: field 'auth' expects the following types: AuthConfigBase, Literal, BearerTokenAuthConfig, ApiKeyAuthConfig, HttpBasicAuthConfig, BearerTokenAuth, APIKeyAuth, HttpBasicAuth. Provided value {'type': 'bearer', 'token': '<mycleartexttokenhere>', 'name': 'Authorization', 'location': 'header'} with type 'dict' is invalid with the following errors:
...
```

### Expected behavior

The sensitive fields like `api_key` or `token` should either be removed from the logs or scrubbed. 

### Steps to reproduce

Create a `RESTAPIConfig` with the following auth config:

```
            "auth": {
                "type": "bearer",
                "token": "this-will-show-in-clear-text",
				"location": "Header",
            },
```

### How you are using the source?

I run this source for fun.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

PR #1643, opened 2024-07-29
implements secret masking for auth helpers

willi-mueller:
### Related Issues

Follow-up to https://github.com/dlt-hub/dlt/issues/1474

### DIFF
--- a/dlt/sources/helpers/rest_client/auth.py
+++ b/dlt/sources/helpers/rest_client/auth.py
@@ -1,6 +1,5 @@
 import math
 import dataclasses
-from abc import abstractmethod
 from base64 import b64encode
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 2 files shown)


sh-rp:
I know this is not ready for review yet, but I think a better approach would be to use introspection to find the parts of a config that are secrets (they will be annotated as secrets) and then mask those. Another thing we could potentially do is to save all secrets that were retrieved from any of the providers in a list somewhere and just clean all log messages with regexes, this might be a bit slow though..

willi-mueller:
> I know this is not ready for review yet, but I think a better approach would be to use introspection to find the parts of a config that are secrets (they will be annotated as secrets) and then mask those. Another thing we could potentially do is to save all secrets that were retrieved from any of the providers in a list somewhere and just clean all log messages with regexes, this might be a bit slow though..
> 
> So in short: Some implementation where we do not need to take care of every individual secret in any config would be good.

Thank you, that's helpful!

We have this PR for review where we mask the sensitive fields of known classes and config values: https://github.com/dlt-hub/verified-sources/pull/542. Is this better?

sh-rp:
@sh-rp we need to make sure no secrets leak on validation and no secrets leak when calling __str__ on a ConfigurationObject.

sh-rp:
closing in favor of this implementation: https://github.com/dlt-hub/dlt/pull/2027/

PR #1645, opened 2024-07-29
prevent accidental wrapping of sources in resources when using adapters

sh-rp:
### Description
A couple of times our users used a destination adapter on a source which creates unexpected results. The small change in this PR changes this.

### DIFF
--- a/dlt/destinations/utils.py
+++ b/dlt/destinations/utils.py
@@ -14,7 +14,7 @@ from dlt.common.schema.utils import (
 from typing import Any, cast, Tuple, Dict, Type
 
 from dlt.destinations.exceptions import DatabaseTransientException
-from dlt.extract import DltResource, resource as make_resource
+from dlt.extract import DltResource, resource as make_resource, DltSource
 
 RE_DATA_TYPE = re.compile(r"([A-Z]+)\((\d+)(?:,\s?(\d+))?\)")
 
@@ -23,6 +23,12 @@ def ensure_resource(data: Any) -> DltResource:
     """Wraps `data` in a DltResource if it's not a DltResource already."""
     if isinstance(data, DltResource):
         return data
+    # prevent accidental wrapping sources with adapters
+    if isinstance(data, DltSource):
+        raise Exception(
+            "You are trying to use an adapter on a dlt source. You can only use adapters on pure"
+            " data or dlt resources."
+        )
     resource_name = None if hasattr(data, "__name__") else "content"
     return cast(DltResource, make_resource(data, name=resource_name))
 


dlt/destinations/utils.py
```diff
@@ -23,6 +23,12 @@ def ensure_resource(data: Any) -> DltResource:
     """Wraps `data` in a DltResource if it's not a DltResource already."""
     if isinstance(data, DltResource):
         return data
+    # prevent accidentally wrapping sources with adapters
+    if isinstance(data, DltSource):
+        raise Exception(
+            "You are trying to use an adapter on a dlt source. You can only use adapters on pure"
```
VioletM, line 29: Is this function only used with adapters? The error message seems to be out of context for `ensure_resource` function. Also, `TypeError` maybe?

dlt/destinations/utils.py
```diff
@@ -23,6 +23,12 @@ def ensure_resource(data: Any) -> DltResource:
...
+            " data or dlt resources."
+        )
     resource_name = None if hasattr(data, "__name__") else "content"
```
VioletM, line 32: I think this else close should throw a warning. Otherwise, you get some unrecognizable table in the destination with the name `Content` out of nowhere :scream:

VioletM (changes requested):

dlt/destinations/utils.py
sh-rp: Yes, it only is used in adapters, I checked that.

rudolfix (changes requested):
LGTM! but let's adapt sources with single resource as well

rudolfix:
@VioletM we should document how to adapt sources. the example below IMO is not the best pattern:
```
from sql_database import sql_database
from dlt.destinations.adapters import weaviate_adapter
import dlt

products_table = sql_database().with_resources("products")

pipeline = dlt.pipeline(
        pipeline_name="postgres_to_weaviate_pipeline",
        destination="weaviate",
    )

products_table.table_name = "Products"

info = pipeline.run(
    weaviate_adapter(products_table.products, vectorize="description"),
)

print(info)
```
beacuse it takes the resource out of source and adapts it and loads it. what user should do:

```
products_tables = sql_database().with_resources("products", "customers")

pipeline = dlt.pipeline(
        pipeline_name="postgres_to_weaviate_pipeline",
        destination="weaviate",
    )

# adapt the resource within the source
weaviate_adapter(products_tables.products, vectorize="description")
weaviate_adapter(products_tables.customers, vectorize="bio")

info = pipeline.run(products_tables)
```

sh-rp:
I have added support for sources with one resource as well as some tests. I'm not sure we should support setting of this default "content" resource name. I have a feeling we should raise similar to when loading data without table name..

dlt/destinations/utils.py
```diff
@@ -14,16 +16,35 @@
...
+    Helper function for adapters. Wraps `data` in a DltResource if it's not a DltResource already.
+    Alternatively if `data` is a DltSource, throws an error if there are multiple resource in the source
+    or returns the single resource if available.
+    """
     if isinstance(data, DltResource):
         return data
-    resource_name = None if hasattr(data, "__name__") else "content"
+    # prevent accidentally wrapping sources with adapters
+    if isinstance(data, DltSource):
+        if len(data.resources.keys()) == 1:
```
rudolfix, line 34: please use "selected_resources". only those are evaluated

dlt/destinations/utils.py
```diff
@@ -14,16 +16,35 @@
...
+        else:
+            raise Exception(
+                "You are trying to use an adapter on a DltSource with multiple resources. You can"
+                " only use adapters on pure data, direclty on a DltResouce or a DltSource"
+                " containing a single DltResource."
+            )
+
+    resource_name = None
+    if not hasattr(data, "__name__"):
+        logger.warning("Setting default resource name to `content` for adapted resource.")
```
rudolfix, line 45: this is not warning. IMO passing "dumb" data like a list of dicts is pretty normal. so maybe info level?

rudolfix (changes requested):

dlt/destinations/utils.py
sh-rp: good point!

VioletM (dismissed):

rudolfix (approved):

PR #1656, opened 2024-08-02
switch to docker compose subcommand

sh-rp:
### Description
Test wether we can switch from docker-compose to docker compose without other changes in the tests.


rudolfix (approved):
LGTM!

PR #1665, opened 2024-08-05
Update salesforce.md

makies:
### Description

I have modified the config sample code used to connect to Salesforce
The salesforce user name must be specified as user_name, not username.

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...


burnash (approved):
Thanks for catching that and for the fix, @makies

PR #1669, opened 2024-08-07
Feat/1492 extend timestamp config

### LINKED ISSUE #1492: Option to keep default date columns as TIMESTAMP_NTZ in oracle to snowflake migration
### Feature description

Community user : When migrating data from Oracle to Snowflake, I would like the ability to default date columns to TIMESTAMP_NTZ instead of TIMESTAMP_TZ.

### Are you a dlt user?

None

### Use case

When migrating data from Oracle to Snowflake, the default behavior sets date columns to TIMESTAMP_TZ. In many cases, it is more desirable to use TIMESTAMP_NTZ to avoid timezone-related complications and align with existing data standards in Snowflake.

### Proposed solution

- Provide an option to set the default timestamp type to TIMESTAMP_NTZ during the migration process.
- Alternatively, allow users to specify the desired timestamp type in the configuration.

### Related issues

_No response_

donotpush:
### Description
`dlt` does not support timestamps without timezones and to be able to pass such timestamps form source to destination we need to extend the core library.

The scope of this PR is to implement a proof of concept (POC) for DuckDB and parquet. This includes updating the type mapper and adding tests where the timezone is passed via columns.

Additionally, a complementary notebook with various experiments is included in this PR.

https://github.com/donotpush/dlt-notebooks/blob/main/notebooks/duckdb-exploration.ipynb

@rudolfix requirements to solve #1492 :
1. We'd need a new hint in `TColumnType` that will say if data_type contains timezone or not (or define time zone as string, with default to "UTC"). See `precision` and `scale` hints already there. The idea is the same.
2. Each of our destinations has a type mapper which needs to be be able to interpret this new thing.
3. `pyarrow.py` contains functions that convert dlt schema into arrow schema and vice versa. those functions needs to be upgraded
4. we'll need plenty of tests: loading non-tz aware datetimes via json and parquet into all possible destinations
5. let's make it work for duckdb+postgres+snowflake first

### Related Issues
- Resolves #1492


dlt/destinations/impl/duckdb/duck.py
```diff
@@ -87,20 +87,28 @@ def to_db_integer_type(
...
 
     def to_db_datetime_type(
-        self, precision: Optional[int], table_format: TTableFormat = None
+        self,
+        timezone: Optional[bool],
+        precision: Optional[int],
+        table_format: TTableFormat = None,
     ) -> str:
+        # TIMESTAMP and TIMESTAMPTZ(TIMESTAMP WITH TIME ZONE) supports microsecond precision
+        if timezone:
```
rudolfix, line 96: here I wound not ignore wrong combinations of precision and timezone for duckdb.

if we see that user wants timezone but precision does not support it we should raise exception where we explain exactly what is going on (must include table name, column name) and how to fix it (change the timezone to False)

rudolfix (commented):
LGTM! implementation idea and code structure is good. duckdb was a hard case because we had technical debt where we silently created columns without timezones if precision required it. now we can do it right

my take: try to implement postgres and snowflake next. if you need any credentials ping us but I think you have access to CI vault

thx for the tests

donotpush:
> Collaborator



> this is really good! I think we are really close to close it.
> 
> 1. could we support BigQuery and Athena? those are popular destinations
> 2. for other destinations log a warning that we do not support ts without timezones
> 
> then we are good to go. I like your tests

I have tried to implement timezone flag for BigQuery using TIMESTAMP, and DATETIME but I have failed. BigQuery does not automatically casts the incoming data for DATETIME - which causes errors. That isn't the case in other databases.

I have documented my findins here: https://github.com/donotpush/dlt-notebooks/blob/main/notebooks/bigquery-exploration.ipynb

docs/website/docs/dlt-ecosystem/destinations/duckdb.md
```diff
@@ -77,7 +77,45 @@ to disable tz adjustments.
 :::
 
 ## Supported column hints
-`duckdb` may create unique indexes for all columns with `unique` hints, but this behavior **is disabled by default** because it slows the loading down significantly.
+
+`duckdb` can create unique indexes for columns with `unique` hints. However, **this feature is disabled by default** as it can significantly slow down data loading.
+
+## Supported column flags
```
rudolfix, line 83: those are still called hints. I'd move this section under
## Data loading

and replace this header with 
### Data types

(it is used in several other destinations)

please apply this to other docs

rudolfix (changes requested):
LGTM! I have suggestion to move docs to different sections. the code is good. thanks for investigating and cleaning up timestamp mess :)

docs/website/docs/dlt-ecosystem/destinations/duckdb.md
donotpush: Done!

**Question**:  What about this section? 
```
## Supported column hints

`duckdb` can create unique indexes for columns with `unique` hints. However, **this feature is disabled by default** as it can significantly slow down data loading.
```
It was there before this PR. Should we keep it as it is, or move it under **Data Loading** as well?

rudolfix (approved):
LGTM!

PR #1675, opened 2024-08-08
[Experiment] dlt cli plugin

sh-rp:
### Description

An PoC on how to install plugins from other packages in dlt by using entry points and decorators for plugin registering (similar to pytest). How to test this out:

1. Run `dlt` from the command line and check the available commands
2. Run `make install-example-plugin` from `tests/plugins/example_plugin', the example plugin will be built and installed in the venv
3. Run `dlt`again, now we have one extra command which also can be executed.


tests/plugins/example_plugin/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.1.0"
+description = ""
+authors = ["dave <shrps@posteo.net>"]
+readme = "README.md"
+packages = [
+    { include = "example_plugin" },
+]
+
+[tool.poetry.plugins."dlt.plugin"]
+example_plugin = "example_plugin"
```
sh-rp, line 12: here we define the entry point which is then discoverable in dlt core.

rudolfix:
@sh-rp I think overall structure is ok

1. how pytest keeps those plugin registries? Also global dicts?
2. do not add code to `dlt.plugins` let’s use it as namespace. Maybe we move code to dlt.common.plugins?
3. please never expose those global dicts to external modules. Could `_dlt.py` just call one method in plugins module that would add collected parsers?
4. this
def cli_plugin(subcommand: str, cli_help: str) -> Callable[[TCliPlugin], None]:
    """decorator for creating a dlt cli subcommand"""

    def register(cli_plugin: TCliPlugin) -> None:
        # print("Registering cli command")
        assert _CURRENT_PLUGIN
        assert (
            subcommand not in _DLT_CLI_PLUGINS
        ), f"CLI Plugin with name {subcommand} already exists"
        _DLT_CLI_PLUGINS[subcommand] = {"func": cli_plugin, "cli_help": cli_help}

    return register

assumes that it will be called during `discover_plugins` where first import happens. (assert _CURRENT_PLUGIN) why? IMO some plugins will contain directly importable code. So maybe it will be imported before dlt call “import_module”

rudolfix:
top level view/summary:
1. all plugins in a module register themselves when module containing them is imported
2. dlt sources are plugins from this POV. they register themselves via a decorator (SourceInfo). Same for standalone resources
3. dlt destinations should also be plugins. they should register themselves via a decorator or explicitly (build in or external destinations) to a destination registry from which we'll instantiate them
4. dlt pipelines should also register themselves (ie via container which tracks them already)

PR #1676, opened 2024-08-08
Add enable_dataset_name_normalization option

VioletM:
Now there is an option to control the dataset name normalization separately. `enable_dataset_name_normalization=true`

### DIFF
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -190,7 +190,7 @@ class DestinationClientDwhConfiguration(DestinationClientConfiguration):
     """How to handle replace disposition for this destination, can be classic or staging"""
     staging_dataset_name_layout: str = "%s_staging"
     """Layout for staging dataset, where %s is replaced with dataset name. placeholder is optional"""
-    dataset_name_normalization: bool = True
+    enable_dataset_name_normalization: bool = True
     """Whether to normalize the dataset name. Affects staging dataset as well."""
 
     def _bind_dataset_name(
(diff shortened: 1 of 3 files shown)


dlt/common/destination/reference.py
```diff
@@ -190,6 +190,8 @@ class DestinationClientDwhConfiguration(DestinationClientConfiguration):
     """How to handle replace disposition for this destination, can be classic or staging"""
     staging_dataset_name_layout: str = "%s_staging"
     """Layout for staging dataset, where %s is replaced with dataset name. placeholder is optional"""
+    dataset_name_normalization: bool = True
```
sh-rp, line 193: I think an imperative makes more sense her "normalize_dataset_name"

rudolfix (commented):
thanks for the tests and improved docs! code LGTM! I agree with Dave's comment to rename the config field

dlt/common/destination/reference.py
VioletM: yes, but we already have this field, and I don't want to change it in all places
```
def normalize_dataset_name(self, schema: Schema) -> str:
```

I was thinking of `to_normalize_dataset_name`, but we don't use such a naming in `dlt`. Do you have any ideas for naming?

rudolfix: ok then `enable_dataset_name_normalizaton`?

rudolfix (approved):
LGTM! thanks for docs and tests

PR #1677, opened 2024-08-09
RESTClient: stops pagination after empty page (Feat/1637)

### LINKED ISSUE #1637: Update RESTClient range paginators to support stopping after empty pages
### Feature description

Update RESTClient's [PageNumberPaginator](https://github.com/dlt-hub/dlt/blob/0d955dee409498824de140490462480625e51b77/dlt/sources/helpers/rest_client/paginators.py#L177) to optionally stop paging after receiving empty pages.

I'm unsure if other paginators need this same functionality. Perhaps it needs to be added to the base `RangePaginator` class.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

I am pulling data from a few REST APIs that do not include the information necessary to determine the total number of records or pages. Setting the max number of pages isn't sufficient because the REST helper will keep sending requests up to the max page limit, even if everything past the first page is empty.

### Proposed solution

Add a boolean keyword argument to the `PageNumberPaginator` (or its base class `RangePaginator`) called `stop_after_empty` that is `False` by default.

When it is true, consider the paging complete when the first empty page is received.

### Related issues

_No response_

willi-mueller:
### Description
- Adds a new option to `RangePaginator` and its subclasses `OffsetPaginator` and `PageNumberPaginator` to stop pagination upon receiving a page without data items
- By default, stops pagination after an empty page
- Every `RangePaginator` now requires either `total_path` or `maximum_offset` or `stop_after_empty_page`
- If `total_path` is not provided, pagination is controlled by `maximum_page` and `stop_after_empty_page` or `maximum_offset` and `stop_after_empty_page` respectively.

### Related Issues

- Resolves #1637

### DIFF
--- a/dlt/sources/helpers/rest_client/paginators.py
+++ b/dlt/sources/helpers/rest_client/paginators.py
@@ -96,7 +96,7 @@ class RangePaginator(BasePaginator):
         maximum_value: Optional[int] = None,
         total_path: Optional[jsonpath.TJsonPath] = None,
         error_message_items: str = "items",
-        stop_after_empty_page: bool = True,
+        stop_after_empty_page: Optional[bool] = True,
     ):
         """
         Args:
@@ -117,6 +117,8 @@ class RangePaginator(BasePaginator):
                 If not provided, `maximum_value` must be specified.
             error_message_items (str): The name of the items in the error message.
                 Defaults to 'items'.
+            stop_after_empty_page (bool): Whether pagination should stop when
+              a page contains no result items. Defaults to `True`.
         """
         super().__init__()
         if total_path is None and maximum_value is None:
@@ -237,7 +239,7 @@ class PageNumberPaginator(RangePaginator):
         page_param: str = "page",
         total_path: jsonpath.TJsonPath = "total",
         maximum_page: Optional[int] = None,
-        stop_after_empty_page: bool = True,
+        stop_after_empty_page: Optional[bool] = True,
     ):
         """
         Args:
@@ -255,6 +257,8 @@ class PageNumberPaginator(RangePaginator):
                 will stop once this page is reached or exceeded, even if more
                 data is available. This allows you to limit the maximum number
                 of pages for pagination. Defaults to None.
+            stop_after_empty_page (bool): Whether pagination should stop when
+              a page contains no result items. Defaults to `True`.
         """
         if total_path is None and maximum_page is None:
             raise ValueError("Either `total_path` or `maximum_page` must be provided.")
@@ -340,7 +344,7 @@ class OffsetPaginator(RangePaginator):
         limit_param: str = "limit",
         total_path: jsonpath.TJsonPath = "total",
         maximum_offset: Optional[int] = None,
-        stop_after_empty_page: bool = True,
+        stop_after_empty_page: Optional[bool] = True,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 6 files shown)


dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -116,44 +117,55 @@ def __init__(
...
                 Defaults to 'items'.
+            stop_after_empty_page (bool): Whether pagination should stop when
+              a page contains no result items. Defaults to `True`.
         """
         super().__init__()
-        if total_path is None and maximum_value is None:
-            raise ValueError("Either `total_path` or `maximum_value` must be provided.")
+        if total_path is None and maximum_value is None and not stop_after_empty_page:
+            raise ValueError(
+                "Either `total_path` or `maximum_value` or stop_after_empty_page must be provided."
```
burnash, line 126:
```suggestion
                "Either `total_path` or `maximum_value` or `stop_after_empty_page` must be provided."
```

dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -116,44 +117,55 @@ def __init__(
...
+
+            self.current_value += self.value_step
+
+            if (total is not None and self.current_value >= total + self.base_index) or (
+                self.maximum_value is not None and self.current_value >= self.maximum_value
+            ):
+                self._has_next_page = False
+
+    def _stop_after_this_page(self, data: Optional[List[Any]]) -> bool:
+        return self.stop_after_empty_page and data == []
```
burnash, line 168: I'm wondering if `data == []` is too restrictive. Maybe `not data`? Also the `data` is defined as an `Optional` in `_stop_after_this_page()`, should it have `None` as a default value?

burnash (dismissed):
Great work, @willi-mueller. I've left a couple of comments

dlt/sources/helpers/rest_client/paginators.py
willi-mueller: Good point! The argument `data` comes from `RESTClient.extract_response()` which always returns a `List[Any]` because its final line is `return cast(List[Any], data)`. 

Thus, I think that this is always a list with 0 or more elements.

But, I prefer providing `data=None` to make it consistent with the other functions, make it more versatile, and explicit.

willi-mueller: Regarding `data == []` vs `data is None`

The existing test suite calls `update_state()` without the `data` argument and thus `data is None` and the tests break which assert that calling `update_state(a_response)` leaves it in a state which has a next page.  I wonder if there are some custom paginator usages that call `update_state()`. In the dlt code, it is only `RESTClient.paginate()`.

I see the following options:

1. `[]` means that there's no data, `None` means that it should not stop after an empty page even if `stop_after_empty_page is True`. Does not break existing test suite and maybe usage.
2. Both `[]` and `None` mean that there's no data. Breaks existing test suite and maybe usage. Necessary adjustment: 
```py
paginator.update_state(response)
```
becomes 
```py
paginator.update_state(response, data=[{}])
```

I favor option 2 because I find it most clear. But, this requires that we not that in the v1.0 release notes.

What do you think @burnash? Do you see more options?

willi-mueller: Done in 83bab151

burnash (approved):
Awesome work, thank you @willi-mueller

PR #1680, opened 2024-08-11
Allow different from credentials project_id

VioletM:


burnash (approved):
Looks great, thank you @VioletM

PR #1690, opened 2024-08-14
rest_api: documents pluggable custom auth

willi-mueller:
### Description

Documents new possibility to register custom auth classes and then use the config dict syntax for them.

### Related Issues

- Implements https://github.com/dlt-hub/verified-sources/issues/524


burnash (approved):
Looks good!

PR #1702, opened 2024-08-15
fixes load job counter

rudolfix:
### Description
Load job counters were incorrectly increased (absolute value was used). This also will skip loading full package details on each completed job decreasing cpu usage on large packages

### DIFF
--- a/dlt/common/runtime/collector.py
+++ b/dlt/common/runtime/collector.py
@@ -170,6 +170,7 @@ class LogCollector(Collector):
                 total=total,
             )
             self.messages[counter_key] = None
+            self.last_log_time = None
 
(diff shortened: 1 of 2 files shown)


dlt/common/runtime/collector.py
```diff
@@ -170,6 +170,7 @@ def update(
                 total=total,
             )
             self.messages[counter_key] = None
+            self.last_log_time = None
```
jorritsandbrink, line 173: > Load job counters were incorrectly increased (absolute value was used).

How does adding this line solve the issue of incorrectly increasing the counter?

jorritsandbrink (approved):

PR #1707, opened 2024-08-19
Enable `scd2` record reinsert

### LINKED ISSUE #1683: `scd2` merge strategy does not reinsert records
### dlt version

0.5.2

### Describe the problem

When using the `scd2` merge strategy, records do not get reinserted after retirement:

- **load 1:** source snapshot contains record x ➜ record x is inserted in destination
- **load 2:** source snapshot does not contain record x ➜ record x is retired in destination ("valid to" field is filled with timestamp)
- **load 3:** source snapshot contains record x ➜ nothing happens (this is the bug)

Issue only occurs if record x is exactly the same. If the natural key is the same, but one or more attributes is different, the record gets inserted as expected.

### Expected behavior

Record x gets reinserted in load 3. Resultant destination table has two records for record x with different validity windows.

### Steps to reproduce

```py
import dlt

@dlt.resource(
    table_name="my_table", write_disposition={"disposition": "merge", "strategy": "scd2"}
)
def r(data):
    yield data

pipe = dlt.pipeline(destination="postgres", dataset_name="scd2_reinsert_bug")

# load 1 — initial load
dim_snap = [
    {"nk": 1, "c1": "foo", "c2": "foo"},
    {"nk": 2, "c1": "bar", "c2": "bar"},
]
pipe.run(r(dim_snap))
# result: inserted 2 records (table size: 2 records)

# load 2 — delete natural key 1
dim_snap = [
    {"nk": 2, "c1": "bar", "c2": "bar"},
]
pipe.run(r(dim_snap))
# result: retired record for natural key 1 (table size: 2 records)

# load 3 — reinsert natural key 1
dim_snap = [
    {"nk": 1, "c1": "foo", "c2": "foo"},
    {"nk": 2, "c1": "bar", "c2": "bar"},
]
pipe.run(r(dim_snap))
# result: nothing happened (table size: 2 records)
# desired result: reinsert record for natural key 1 (table size: 3 records)
```

### Operating system

Windows

### Runtime environment

Local

### Python version

3.8

### dlt data source

_No response_

### dlt destination

Postgres

### Other deployment details

_No response_

### Additional information

_No response_

jorritsandbrink:
### Description
- enables the _insert-delete-reinsert_ pattern (see https://github.com/dlt-hub/dlt/pull/1707/commits/55850de5500b23b9c40e794947308d4a6540ff44) for `scd2`
- drops `unique` constraint on `_dlt_id` in root tables (except when users bring their own hash)

### Related Issues

Fixes #1683

### DIFF
--- a/dlt/destinations/sql_jobs.py
+++ b/dlt/destinations/sql_jobs.py
@@ -749,20 +749,20 @@ class SqlMergeFollowupJob(SqlFollowupJob):
             INSERT INTO {root_table_name} ({col_str}, {from_}, {to})
             SELECT {col_str}, {boundary_ts} AS {from_}, {active_record_literal} AS {to}
             FROM {staging_root_table_name} AS s
-            WHERE {hash_} NOT IN (SELECT {hash_} FROM {root_table_name});
+            WHERE {hash_} NOT IN (SELECT {hash_} FROM {root_table_name} WHERE {is_active_clause});
         """)
 
         # insert list elements for new active records in child tables
         child_tables = table_chain[1:]
         if child_tables:
-            unique_column = escape_column_id(
-                cls._get_unique_col(table_chain, sql_client, root_table)
-            )
             # TODO: - based on deterministic child hashes (OK)
             # - if row hash changes all is right
             # - if it does not we only capture new records, while we should replace existing with those in stage
             # - this write disposition is way more similar to regular merge (how root tables are handled is different, other tables handled same)
             for table in child_tables:
+                unique_column = escape_column_id(
+                    cls._get_unique_col(table_chain, sql_client, table)
+                )
                 table_name, staging_table_name = sql_client.get_qualified_table_names(table["name"])
                 sql.append(f"""
                     INSERT INTO {table_name}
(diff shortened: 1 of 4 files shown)


dlt/destinations/sql_jobs.py
```diff
@@ -749,20 +749,20 @@ def gen_scd2_sql(
             INSERT INTO {root_table_name} ({col_str}, {from_}, {to})
             SELECT {col_str}, {boundary_ts} AS {from_}, {active_record_literal} AS {to}
             FROM {staging_root_table_name} AS s
-            WHERE {hash_} NOT IN (SELECT {hash_} FROM {root_table_name});
+            WHERE {hash_} NOT IN (SELECT {hash_} FROM {root_table_name} WHERE {is_active_clause});
```
jorritsandbrink, line 752: This change enables reinsertion of a record if there is no active version of that record in the destination table.

One possible issue: users might have a UNIQUE constraint on the `_dlt_id` column in their existing `scd2` tables. This will cause a uniqueness error. Those users would have to drop those constraints manually when upgrading to a newer version of `dlt`.

PR #1708, opened 2024-08-19
feat/1681 collects load job metrics and adds remote uri

### LINKED ISSUE #1681: Add actual file paths to traces in case of filesystem destination
### Feature description

When the destination is a filesystem it would be really nice to have a file paths for each table visible in traces.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Using traces for monitoring it's important to know exact path to files loaded in destination.

### Proposed solution

This information could be added to `TRACE__STEPS__STEP_INFO__LOAD_PACKAGES__JOBS` table.

### Daves Notes:
We will probably want to implement this the same way as the traces of the normalizer stage. Every load job (including the follow up jobs) produce a metrics entry with at least the local name and execution time of the job, the destination (staging or final) it was executed against and a string or json structure that the job itself may provide as result of the execution. In the case of the filesystem jobs, this would be the remote path for example.

rudolfix:
### Description
fixes #1681 

also changes shape of the trace and puts contract on it (in tests)
* job_id added to `trace__steps__step_info__load_packages__jobs`
* `run` step removed (is a copy of last step in the pipeline) + corresponding columns in `trace`
* added `table_name` in `job_metrics`
* added `trace__steps__load_info__job_metrics`

also refactors code

### DIFF
--- a/dlt/common/metrics.py
+++ b/dlt/common/metrics.py
@@ -1,4 +1,4 @@
-import datetime
+import datetime  # noqa: I251
 from typing import Any, Dict, List, NamedTuple, Optional, Tuple, TypedDict  # noqa: 251
 
 
--- a/dlt/common/metrics.py
+++ b/dlt/common/metrics.py
@@ -1,0 +1,71 @@
+import datetime
+from typing import Any, Dict, List, NamedTuple, Optional, Tuple, TypedDict  # noqa: 251
+
+
+class DataWriterMetrics(NamedTuple):
+    file_path: str
+    items_count: int
+    file_size: int
+    created: float
+    last_modified: float
+
+    def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:
+        if isinstance(other, DataWriterMetrics):
+            return DataWriterMetrics(
+                self.file_path if self.file_path == other.file_path else "",
+                # self.table_name if self.table_name == other.table_name else "",
+                self.items_count + other.items_count,
+                self.file_size + other.file_size,
+                min(self.created, other.created),
+                max(self.last_modified, other.last_modified),
+            )
+        return NotImplemented
+
+
+class StepMetrics(TypedDict):
+    """Metrics for particular package processed in particular pipeline step"""
+
+    started_at: datetime.datetime
+    """Start of package processing"""
+    finished_at: datetime.datetime
+    """End of package processing"""
+
+
+class ExtractDataInfo(TypedDict):
+    name: str
+    data_type: str
+
+
+class ExtractMetrics(StepMetrics):
+    schema_name: str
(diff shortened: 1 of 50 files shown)


dlt/load/load.py
```diff
@@ -392,6 +398,11 @@ def complete_jobs(
                 # create followup jobs
                 self.create_followup_jobs(load_id, state, job, schema)
 
+                # preserve metrics
+                metrics = job.metrics()
+                if metrics:
```
sh-rp, line 403: when does it happen that a job does not have metrics?

dlt/common/metrics.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class NormalizeMetrics(StepMetrics):
+    job_metrics: Dict[str, DataWriterMetrics]
+    """Metrics collected per job id during writing of job file"""
+    table_metrics: Dict[str, DataWriterMetrics]
+    """Job metrics aggregated by table"""
+
+
+class LoadJobMetrics(NamedTuple):
```
sh-rp, line 60: I think it might be of interested to also track whether the data from this job was loaded to the staging destination or the staging dataset

dlt/common/metrics.py
```diff
@@ -0,0 +1,71 @@
...
+    job_id: str
+    file_path: str
+    table_name: str
+    started_at: datetime.datetime
+    finished_at: datetime.datetime
+    state: Optional[str]
+    remote_uri: Optional[str]
```
sh-rp, line 67: I was thinking that LoadJobs could have a space for arbitrary data (defined as Dict[str, str]) for information such as the remote_uri. What comes to mind is, that for example BigQuery Jobs have an internal BigQuery Job Id which we could store here. Many SQL Databases also have transaction_ids which could be saved in the traces.

dlt/common/metrics.py
sh-rp: It would also be cool to track if this was a table chain followup job. With these three flags you could make the/any platform UI much more legible.

sh-rp (approved):
Cool PR! I think it is good, but I have some more ideas of what to track in the load metrics when I think about how to display trace data in a useful way to the user in some tool.

dlt/common/metrics.py
sh-rp: Thinking a bit more about it, maybe a job_type would be interesting too in the context of displaying data to the user. Types could be:

single table jobs
* copy (all jobs that just move files around)
* load (all jobs that move data into a database)

table chain jobs
* merge (merge jobs)
* upsert (upsert merge jobs)
* scd2 (scd2 merge)
* replace (replace via insert from staging)
* whatever we call this chunking update

rudolfix:
@sh-rp we followup here #1743 and here #1744

PR #1709, opened 2024-08-21
`scd2` custom "valid from" / "valid to" value feature 

### LINKED ISSUE #1601: Allow Custom Override of `_dlt_valid_from` and `_dlt_valid_to` Timestamp in SCD2 Incremental Loading
### Feature description

Currently, when using the SCD2 strategy for incremental loading in dlt, the `_dlt_valid_from` and `_dlt_valid_to`  timestamp is automatically set to the time the pipeline is executed. This can cause issues when records need to reflect a specific historical period rather than the pipeline execution time.

Request:

I would like to request a feature that allows users to override the _dlt_valid_from/to timestamp with an arbitrary value. This would provide more flexibility and control over the data's validity periods, making it possible to accurately represent historical changes.

### Are you a dlt user?

### Use case

When processing historical data snapshots, it is often necessary to set specific `_dlt_valid_from` and `_dlt_valid_to` timestamps to reflect accurate validity periods. The ability to manually set this value would help maintain the integrity of historical data and ensure that the SCD2 strategy works as intended.

### Proposed solution

Allow the `_dlt_valid_from` and `_dlt_valid_to` timestamp to be overridden by specifying it as a parameter when executing the `dlt` command.

### Related issues

_No response_

jorritsandbrink:
### Description
Adds `boundary_timestamp` argument that's interpreted for `scd2` merge strategy. Lets users configure the timestamp value used in record validity windows (i.e. the value used as "valid from" for new records and as "valid to" for retired records).

### Related Issues

Closes #1601

### DIFF
--- a/tests/load/pipeline/test_scd2.py
+++ b/tests/load/pipeline/test_scd2.py
@@ -3,6 +3,7 @@
 import pytest
 from typing import List, Dict, Any, Optional
 from datetime import date, datetime, timezone  # noqa: I251
+from contextlib import nullcontext as does_not_raise
 
 import dlt
 from dlt.common.typing import TAnyDateTime
@@ -583,6 +584,7 @@ def test_validity_column_name_conflict(destination_config: DestinationTestConfig
         "9999-12-31T00:00:00",
         "9999-12-31T00:00:00+00:00",
         "9999-12-31T00:00:00+01:00",
+        "i_am_not_a_timestamp",
     ],
 )
 def test_active_record_timestamp(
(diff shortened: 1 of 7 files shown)


tests/load/pipeline/test_scd2.py
```diff
@@ -69,20 +71,6 @@ def strip_timezone(ts: datetime) -> datetime:
         return table
     return sorted(table, key=lambda d: d[sort_column])
 
-    return sorted(
```
jorritsandbrink, line 72: This removed block is not related to the PR. It's dead code I forgot to remove in another recent PR.

rudolfix (approved):
LGTM!

PR #1716, opened 2024-08-22
Make `make lint` fail on `black` format diff

jorritsandbrink:
### Description
- makes `make lint` exit with error code 1 (and thus makes the `lint.yml` workflow fail) when `black` detects a formatting diff—this should lead to less unformatted code in the `devel` branch
- adds coloring to diff output

see it in action here: https://github.com/dlt-hub/dlt/actions/runs/10502842793/job/29095135609?pr=1716#step:7:100

### Related Issues
Slack chat: https://dlthub-community.slack.com/archives/C06EYJ2Q37C/p1724234001980989


rudolfix (approved):
this is good!

PR #1717, opened 2024-08-22
Expose staging tables truncation to config

VioletM:


dlt/common/destination/reference.py
```diff
@@ -268,6 +268,8 @@ class DestinationClientDwhWithStagingConfiguration(DestinationClientDwhConfigura
 
     staging_config: Optional[DestinationClientStagingConfiguration] = None
     """configuration of the staging, if present, injected at runtime"""
+    truncate_table_before_load_on_staging_destination: bool = True
```
sh-rp, line 271: I think I'd call this "truncate_tables_on_staging_destination_before_load"

dlt/common/destination/reference.py
```diff
@@ -560,14 +562,19 @@ def with_staging_dataset(self) -> ContextManager["JobClientBase"]:
 class SupportsStagingDestination:
     """Adds capability to support a staging destination for the load"""
 
+    def __init__(self, truncate_table_before_load_on_staging_destination: bool = True) -> None:
```
sh-rp, line 565: I think you can take the full DestinationClientDwhWithStagingConfiguration as argument here.

sh-rp: And it should be required, no default value. That way we can ensure that it is implemented correctly in each destination.

dlt/destinations/impl/athena/athena.py
```diff
@@ -389,6 +389,9 @@ def __init__(
             config,
             capabilities,
         )
+        SupportsStagingDestination.__init__(
```
sh-rp, line 392: In athena this method actually is implemented because we need special behavior for athena to work properly. I think here we can just ignore this setting and mention it in the docs.

sh-rp (changes requested):
Very good already!

dlt/common/destination/reference.py
```diff
@@ -577,14 +579,19 @@ def with_staging_dataset(self) -> ContextManager["JobClientBase"]:
 class SupportsStagingDestination:
     """Adds capability to support a staging destination for the load"""
 
+    def __init__(self, config: DestinationClientDwhWithStagingConfiguration) -> None:
```
rudolfix, line 582: this should be removed. `SupportsStagingDestination` is a mixing class, should not hold any state.

dlt/common/destination/reference.py
```diff
@@ -577,14 +579,19 @@ def with_staging_dataset(self) -> ContextManager["JobClientBase"]:
...
+
     def should_load_data_to_staging_dataset_on_staging_destination(
         self, table: TTableSchema
     ) -> bool:
         return False
 
     def should_truncate_table_before_load_on_staging_destination(self, table: TTableSchema) -> bool:
         # the default is to truncate the tables on the staging destination...
-        return True
+        return self.truncate_tables_on_staging_destination_before_load
```
rudolfix, line 594: make this method abstract to force everyone to implement it since we can't use this nice default anymore

dlt/destinations/impl/athena/athena.py
```diff
@@ -389,6 +389,7 @@ def __init__(
             config,
             capabilities,
         )
+        SupportsStagingDestination.__init__(self, config)
```
rudolfix, line 392: not needed

dlt/destinations/impl/athena/athena.py
```diff
@@ -531,7 +532,7 @@ def should_truncate_table_before_load_on_staging_destination(self, table: TTable
         if table["write_disposition"] == "replace" and not self._is_iceberg_table(
             self.prepare_load_table(table["name"])
         ):
-            return True
+            return self.truncate_tables_on_staging_destination_before_load
```
rudolfix, line 535: use `self.config.truncate_tables_on_staging_destination_before_load`

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -229,6 +229,7 @@ def __init__(
             config.http_timeout,
             config.retry_deadline,
         )
+        SupportsStagingDestination.__init__(self, config)
```
rudolfix, line 232: all of those are not needed!

dlt/destinations/impl/dummy/configuration.py
```diff
@@ -34,6 +34,8 @@ class DummyClientConfiguration(DestinationClientConfiguration):
     """raise terminal exception in job init"""
     fail_transiently_in_init: bool = False
     """raise transient exception in job init"""
+    truncate_tables_on_staging_destination_before_load: bool = True
```
rudolfix, line 37: my take is to not change dummy at all but write a single test that works on all destinations that support staging destination. see review comment below

rudolfix (changes requested):
we need a proper test (look at test_stage_loading.py):
1. make sure that staging destination is not truncated (flag at default) and truncated (flag to false). at the same time make sure that destination tables are NOT truncated
2.  exclude Athena if not force iceberg (or make sure that staging destination was not truncated)
3. use some kind of simple dataset (not github - takes a lot of time)

@VioletM feel free to push this task to @rudolfix or @sh-rp or someone else...

dlt/common/destination/reference.py
VioletM: done

dlt/destinations/impl/dummy/configuration.py
VioletM: left both tests -- dummy based and with real destinations.
We need to implement the `should_truncate_table_before_load_on_staging_destination` in the dummy destination anyway.

VioletM:
@rudolfix 
> we need a proper test (look at test_stage_loading.py):

Added a test which runs on `all staging configurations`, taking about 10min, which is quite a lot. I can reduce the amount of all combinations between destinations and bucket types. It's nice that we're testing many combinations, but not necessary. Let me know if you think the test needs to be changed or even if you prefer to do it yourself.

> 1. make sure that staging destination is not truncated (flag at default) and truncated (flag to false). at the same time make sure that destination tables are NOT truncated

I think we should left default behavior as is -- truncated (flag to True by default). Otherwise, with the update users start to see increase in the staging usage

rudolfix (approved):
this is very good! thanks for the tests and docs update. your staging test is not marked as essential (OK!) so it won't run on a few destinations. nevertheless we can merge it and fix any remaining tests just before release

(I expect that athena non iceberg will fail here)

VioletM:
> (I expect that athena non iceberg will fail here)

Hm, but this test uses `destinations_configs(all_staging_configs=True)` and there is no athena config present there

docs/website/docs/dlt-ecosystem/staging.md
```diff
@@ -96,4 +100,21 @@ In essence, you need to set up two destinations and then pass them to `dlt.pipel
 
     Run the pipeline script as usual.
 
-> 💡 Please note that `dlt` does not delete loaded files from the staging storage after the load is complete.
+:::tip
```
sh-rp, line 103: I'd run this whole new section through chatgpt to grammar correct (or run the grammar checker on this page)

tests/load/pipeline/test_stage_loading.py
```diff
@@ -183,6 +190,50 @@ def test_staging_load(destination_config: DestinationTestConfiguration) -> None:
...
+    assert_load_info(info)
+
+    # load the data without truncating of the staging, should see two files on staging
+    pipeline.destination.config_params["truncate_tables_on_staging_destination_before_load"] = False
+    info = pipeline.run(resource)
+    assert_load_info(info)
+    # check there are two staging files
+    _, staging_client = pipeline._get_destination_clients(pipeline.default_schema)
+    with staging_client:
+        assert len(staging_client.list_table_files(table_name)) == 2  # type: ignore[attr-defined]
```
sh-rp, line 221: note for later: we should probably allow fs_client on the pipeline to also return the staging filesystem client with a flag.

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -503,6 +503,9 @@ def _should_autodetect_schema(self, table_name: str) -> bool:
             self.schema._schema_tables, table_name, AUTODETECT_SCHEMA_HINT, allow_none=True
         ) or (self.config.autodetect_schema and table_name not in self.schema.dlt_table_names())
 
+    def should_truncate_table_before_load_on_staging_destination(self, table: TTableSchema) -> bool:
```
sh-rp, line 506: imho having the same exact line in all implementations is not DRY. We can keep it like this for now, but I would rather implement this in the superclass, probably with hasattr and isinstance to get the config and verify the type.

tests/load/pipeline/test_stage_loading.py
```diff
@@ -183,6 +190,50 @@ def test_staging_load(destination_config: DestinationTestConfiguration) -> None:
     assert replace_counts == initial_counts
 
 
+@pytest.mark.parametrize(
+    "destination_config", destinations_configs(all_staging_configs=True), ids=lambda x: x.name
+)
+def test_truncate_staging_dataset(destination_config: DestinationTestConfiguration) -> None:
+    """This test checks if tables truncation on staging destination done according to the configuration.
```
sh-rp, line 197: the test is good, but to make it great we should also test wether keeping the staging files around will make the data be loaded again although it shouldn't. but for now i'd say it's good.

sh-rp (dismissed):
LGTM, please run the grammar checker on the docs and then you can merge.

tests/load/pipeline/test_stage_loading.py
VioletM: good point, should be easy to do

dlt/destinations/impl/bigquery/bigquery.py
rudolfix: did you see the previous version? at least here this is simple and mypy forces you to do overrides. previously you had to setup cooperative calling of super() and keep state in the class that should be an interface/trait. also if you forgot about that, the config would have no effect and you'll never know about it

Issue #1672, opened 2024-08-08
Move `filesystem ` source from verified-sources to dlt core


burnash:
### Background

The [`filesystem`](https://github.com/dlt-hub/verified-sources/tree/8ea2fe367fde7a539543c326b81576bfceba1381/sources/filesystem) source, currently lives in the verified-sources repository. It's a critical component for many users of the dlt who load data from cloud and local storage. Given its importance and frequent usage, integrating this source directly into the dlt core simplifies the user experience and makes dependency management easier.

### Proposal

Move the `filesystem` source from the verified-sources repository into the dlt core repository.

### Tasks

- [ ] Migrate code
- [ ] Migrate tests
- [ ] Update documentation: installation, usage, and examples. Add migration docs 
- [ ] Deprecate the `filesystem` source in the verified-sources repository

burnash:
@rudolfix please feel free to update the description

PR #1732, opened 2024-08-22
1672 - moves filesystem source

rudolfix:
### Description
1. keeps example pipelines and runs them in `tests/sources` as they are intended to run on local filesystem and duckdb so they are safe to be run without CI secrets
2. places tests in tests/load/sources/ because they depend on bucket configs that are available in load
3. all tests in (2) that instantiate `dlt.pipeline` use `DestinationConfig` scheme to run on all destinations
4. uses existing util functions and fixtures

Resolves: #1672


tests/load/sources/filesystem/test_filesystem_source.py
```diff
@@ -0,0 +1,273 @@
...
+    from dlt.extract.pipe_iterator import PipeIterator
+
+    # use pipe iterator to get items as they go through pipe
+    for pipe_item in PipeIterator.from_pipe(resource._pipe):
+        assert len(pipe_item.item) == 2
+        # no need to test more chunks
+        break
+
+
+@pytest.mark.parametrize(
```
sh-rp, line 257: with this setup you'll always run the example pipeline against duckdb, not against the currently active destination. I'm not quite sure how to inject the right destination for this test, maybe in the future we can make use of the updated dlt init command that is run before the tests to create a file with the right destination from which we can import this, not sure.

 I'd also suggest at least doing a very basic inspection of the loaded data by attaching to the pipeline, but maybe this is not needed..

sh-rp (changes requested):
To me it looks good with 2 questions:
* How do we test against all destinations properly (see comment below)
* Where do we specify the needed dependencies for the example pipeline (I will figure this out in my init PR, as suggested in slack I think I would rather use a templates folder that can contain all of this)

rudolfix:
> How do we test against all destinations properly (see comment below)
Right, I forgot about it. The tests are placed in `tests/load` so I'd just add our standard destination config parametrize to it. And they will be picked up automatically,

The example tests - they should run only on duckdb and local filesystem. And I will move them back to tests/sources

> Where do we specify the needed dependencies for the example pipeline (I will figure this out in my init PR, as suggested in slack I think I would rather use a templates folder that can contain all of this)
filesystem does not have additional requirements. I'd probably keep them together with the code as in verified sources. best if you propose something neat :)

sh-rp (approved):

dlt/sources/filesystem/readers.py
```diff
@@ -0,0 +1,129 @@
...
+
+    for file_obj in items:
+        # Here we use pandas chunksize to read the file in chunks and avoid loading the whole file
+        # in memory.
+        with file_obj.open() as file:
+            for df in pd.read_csv(file, **kwargs):
+                yield df.to_dict(orient="records")
+
+
+def _read_jsonl(items: Iterator[FileItemDict], chunksize: int = 1000) -> Iterator[TDataItems]:
```
VioletM, line 35: @rudolfix 
I'd suggest adding another reader for `.json` files. It's quite easy and it's being requested from time to time.

Issue #1627, opened 2024-07-23
Add sqlite as a dlt destination


Vasilije1990:
### Feature description

sqlite can be added as dlt destination 

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

I need an in-memory database with async support

### Proposed solution

Adding sqlite as a destination

### Related issues

https://news.ycombinator.com/item?id=39527441

rudolfix:
@Vasilije1990 (and anyone else interested) - is there anything special that you need:
1. what does async support mean? we do not have async destinations right now. they operate in batch mode. you could write custom async destination (reverse ETL) but this is not probably what you want
2. why SQL lite and not DuckDB?
3. sql lite is so different from other engines that we planned to do this in #21 via sql alchemy. that would also support 50+ other destinations. the blocker is lack of resources and until now there was not much interest in supporting regular dbmses as destinations

Vasilije1990:
1. DuckDB is not async, and I am using the destination as a metadata store for other services that use for example FastAPI Users as one
2. FastAPI Users requires async to work
3. That would work

rudolfix:
should be solved by #21

Issue #1329, opened 2024-05-07
support MySQL


taochen-ct:
### Feature description

when i use dlt on my project the source database is mysql, i can not extract data,  so request support MySQL

### Are you a dlt user?

Yes, I'm already a dlt user.

mirimgtec:
I think you can use command:
 dlt init sql_database duckdb
to see example for MySQL. dialect with SQLAlchemy library.
Now you can not treat MySQL as destination from out of the box.

PR #1734, opened 2024-08-23
Sqlalchemy destination

### LINKED ISSUE #21: implement sql alchemy destination
**Background**
We can handle many rdbmses with a single SQLAlchemy destination. In this ticket we want to implement a generic loader and test it with two engines: SQLLite and MySQL

**Requirements**

1. implement bulk inserts if possible
- we may need a new file writer from which sql alchemy inserts are the fastest
- some dialect support bulk inserts
https://stackoverflow.com/questions/3659142/bulk-insert-with-sqlalchemy-orm

2. sql alchemy has identifier length in each dialect
3. now destination capabilities may be dynamically modified. Use the connection string (that contains engine+dialect) to synthesize it.
4. if needed add `sql_alchemy_caps` to destination caps with additional settings (ie. if bulk inserts are supported or JSON fields)
- allow user to provide own sql_alchemy_caps
5. Use SQLALchemy to create and reflect databases. Implement `sql_client` interface (but probably form scratch).
6. write dispositions:
- implement append and replace write disposition
- merge in next ticket (I’m not even sure it is possible. But we are using only INSERT and DELETE so maybe?)

Tests:
- most of the tests that abstract sql_client and job_client should pass right away (ie. pipeline tests, state sync, drop command)
- sql_client tests should also pass

steinitzu:
**Draft** Sqlalchemy loader. Basic things work so far.   

### Description

* Mysql generally works. 
* Sqlite currently only works with dataset name `main`.   
* Support parquet and typed-jsonl file formats
* Implements `SqlJobClient` interface so most/all sql loader tests should run (many already run against mysql)

What's left:
- [ ] Override sql_client methods when sqlite is used. create/drop dataset works differently
- [ ] Limit query length and number of params sent per statement in load jobs
- [ ] Test bulk insert
- [ ] Test with sqlite and mysql (maybe other built in dialects that can run locally postgres/mssql)
- [ ] Test table reflection and table building logic
- [ ] Test sqlalchemy 1.4 and 2  in CI
- [ ] Lots of cleanup

### Related Issues
- Resolves #21 , #1627 , #1329


dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -0,0 +1,378 @@
...
+        dataset_name: str,
+        staging_dataset_name: str,
+        credentials: SqlalchemyCredentials,
+        capabilities: DestinationCapabilitiesContext,
+        engine_args: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        super().__init__(credentials.database, dataset_name, staging_dataset_name, capabilities)
+        self.credentials = credentials
+
+        if credentials.engine:
```
rudolfix, line 112: we could also reuse a single engine. you could look at DuckDbConnection which creates a single master connection and then borrows instances to clients.

hmmm we do not need to overcomplicate. I hope opening an engine is not that heavy

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -0,0 +1,378 @@
...
+        return str(sa.schema.CreateColumn(column).compile(self.engine))
+
+    def reflect_table(
+        self,
+        table_name: str,
+        metadata: Optional[sa.MetaData] = None,
+        include_columns: Optional[Sequence[str]] = None,
+    ) -> Optional[sa.Table]:
+        """Reflect a table from the database and return the Table object"""
+        if metadata is None:
```
rudolfix, line 318: there's no way to reflect many tables at once? still - this is good enough :)

dlt/destinations/impl/sqlalchemy/sqlalchemy_job_client.py
```diff
@@ -0,0 +1,503 @@
...
+        elif precision <= 32:
+            return sa.Integer()
+        elif precision <= 64:
+            return sa.BigInteger()
+        raise TerminalValueError(f"Unsupported precision for integer type: {precision}")
+
+    def _create_date_time_type(self, sc_t: str, precision: Optional[int]) -> sa.types.TypeEngine:
+        """Use the dialect specific datetime/time type if possible since the generic type doesn't accept precision argument"""
+        precision = precision if precision is not None else self.capabilities.timestamp_precision
+        if sc_t == "timestamp":
```
rudolfix, line 57: we support datetimes without timezones now see [#1492](https://github.com/dlt-hub/dlt/pull/1669)

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -0,0 +1,378 @@
...
+            tmp_metadata = sa.MetaData()
+            tbl = sa.Table(table_name, tmp_metadata, schema=self.dataset_name)
+        return self._make_qualified_table_name(tbl, escape)
+
+    def alter_table_add_column(self, column: sa.Column) -> None:
+        """Execute an ALTER TABLE ... ADD COLUMN ... statement for the given column.
+        The column must be fully defined and attached to a table.
+        """
+        # TODO: May need capability to override ALTER TABLE statement for different dialects
+        alter_tmpl = "ALTER TABLE {table} ADD COLUMN {column};"
```
rudolfix, line 293: there's no ALTER TABLE in SQLAlchemy? we are reflecting all tables before update schema. maybe we could reuse those tables to add columns and sync?

rudolfix (commented):
LGTM!

I was expecting that we need some kind of settings per dialect (ie `DialectCapabilities`) so we have optimized inserts. but it seems it is not necessary.

There are a lot of standard tests where this thing should work:
- restore state tests
- drop command / refresh mode tests
- all pipeline tests that do not requires merge write disposition

if we enable them we are good.

MERGE write disposition looks complicated, right? We surely can generate all required statements (we have only SELECT / INSERT / DELETE) but creating temporary tables looks like a stretch. Still we can do the Athena trick and create real tables for temporary data

A REPLACE mode that uses staging dataset would be also cool.

maybe those two could go to a separate ticket

dlt/destinations/impl/sqlalchemy/db_api_client.py
steinitzu: Yeah I was trying to keep it simple for now, but definitely would be nice to use the connection pool.   
I had in mind some generic solution for this cause it applies to other destinations too. E.g. if we add a "thread safe" capability then the loader can re-use the whole jobclient instance across jobs. In this case it wouldn't take much to make the client thread friendly, just a matter of moving the "current transaction" stuff to context vars.

dlt/destinations/impl/sqlalchemy/db_api_client.py
steinitzu: There is `metadata.reflect(only=[...table_names])`, but the problem with that is the whole operation fails if any of the tables don't exist. So it's tricky (unless we start using un-documented internals in sqla to hack something together).   
 But I *think* internally it handles one table at a time anyway so it shouldn't be much difference.

dlt/destinations/impl/sqlalchemy/db_api_client.py
steinitzu: There isn't unfortunately anything I could find, I dug through all the DDL generation stuff. I was planning to experiment with bringing in alembic for this if it's not too complicated, that should work at least with the built-in dialects.   

Sqlalchemy itself only handles creating tables, `Table().create()`, etc, that doesn't alter any existing tables.

rudolfix: OMG! I see

dlt/destinations/impl/sqlalchemy/sqlalchemy_job_client.py
```diff
@@ -0,0 +1,503 @@
...
+from dlt.common.json import json, PY_DATETIME_DECODERS
+from dlt.destinations.exceptions import DatabaseUndefinedRelation
+
+
+# from dlt.destinations.impl.sqlalchemy.sql_client import SqlalchemyClient
+from dlt.destinations.impl.sqlalchemy.db_api_client import SqlalchemyClient
+from dlt.destinations.impl.sqlalchemy.configuration import SqlalchemyClientConfiguration
+
+
+class SqlaTypeMapper:
```
rudolfix, line 35: TypeMapper got moved to `factory` and now has ABC class (there's one more method to detect incompatible data types but I think here it is just empty).

dlt/destinations/impl/sqlalchemy/sqlalchemy_job_client.py
```diff
@@ -0,0 +1,503 @@
...
+            from sqlalchemy.dialects.mysql import DOUBLE
+        return sa.Float(precision=53)  # Otherwise use float
+
+    def _to_db_decimal_type(self, column: TColumnSchema) -> sa.types.TypeEngine:
+        precision, scale = column.get("precision"), column.get("scale")
+        if precision is None and scale is None:
+            precision, scale = self.capabilities.decimal_precision
+        return sa.Numeric(precision, scale)
+
+    def to_db_type(self, column: TColumnSchema, table_format: TTableSchema) -> sa.types.TypeEngine:
```
rudolfix, line 100: there's PreparedTableSchema used in many places. this is a separate type on top of TTableSchema to make sure non-prepared tables do not leak to sql job client

rudolfix (changes requested):
please merge current devel. all big changes for 1.0 are in. this looks really good!

what about merges? IMO our standard set of sql jobs will work just fine with mysql and sqllite. maybe you could try that out?

dlt/destinations/impl/sqlalchemy/alter_table.py
```diff
@@ -0,0 +1,38 @@
+from typing import List
+
+import sqlalchemy as sa
+from alembic.runtime.migration import MigrationContext
```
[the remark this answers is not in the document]
steinitzu: Yeah would have liked to avoid this, but at least it was less complicated than I expected since alembic is not really used like this normally.

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -0,0 +1,415 @@
...
+        if self._current_connection is None:
+            self._current_connection = self.engine.connect()
+            if self.dialect_name == "sqlite":
+                self._sqlite_reattach_dataset_if_exists(self.dataset_name)
+        return self._current_connection
+
+    def close_connection(self) -> None:
+        if not self.external_engine:
+            try:
+                self.engine.dispose()
```
rudolfix, line 139: you do not close the active connection here
https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Engine.dispose

only connections in the pool will be closed by dispose. the checked out - not

IMO you must close the current connection

also do not dispose external engines.

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -0,0 +1,415 @@
...
+        engine_args: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        super().__init__(credentials.database, dataset_name, staging_dataset_name, capabilities)
+        self.credentials = credentials
+
+        if credentials.engine:
+            self.engine = credentials.engine
+            self.external_engine = True
+        else:
+            self.engine = sa.create_engine(
```
rudolfix, line 118: could we create engine when first connection is created? sometimes sql client is created to just use a few internal functions and never opens the connection

rudolfix (changes requested):
pls see comments on engine.dispose

steinitzu:
> @steinitzu all good but we do not need
> 
> ```
> self.engine.dispose()
> ```
> 
> we have null pool. so it will dispose of external engines which should not happen. WDYT?

Hmm I have the the flag for external engine so it's not closed:
https://github.com/dlt-hub/dlt/blob/sqlalchemy-loader/dlt/destinations/impl/sqlalchemy/db_api_client.py#L115-L117  

But yeah not sure if there's any need for it since the connection is closed anyway. Maybe in the special case where you override poolclass?

rudolfix:
> > @steinitzu all good but we do not need
> > ```
> > self.engine.dispose()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > we have null pool. so it will dispose of external engines which should not happen. WDYT?
> 
> Hmm I have the the flag for external engine so it's not closed: https://github.com/dlt-hub/dlt/blob/sqlalchemy-loader/dlt/destinations/impl/sqlalchemy/db_api_client.py#L115-L117
> 
> But yeah not sure if there's any need for it since the connection is closed anyway. Maybe in the special case where you override poolclass?

heh you are right. I overlooked the external flag. then all good

rudolfix (approved):
LGTM!

PR #1740, opened 2024-08-24
Feat/1711 create with not exists dlt tables

### LINKED ISSUE #1711: Prevent a race condition for the first run of pipelines that write in the same schema
### Feature description

When we have the following use case:
* several pipelines run in parallel
* they write to the same schema
* this is a first run

There could be a possibility of race conditions when pipelines are trying to create common tables:

```
_DLT_LOADS
_DLT_PIPELINE_STATE
_DLT_VERSION
``` 

This could be prevented with usage of sql statement with `IF EXISTS` clause.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Replication of big number of tables in parallel to the same dataset.

### Proposed solution

Using `IF EXISTS` clause when creating dlt internal tables

### Related issues

_No response_

rudolfix:
### Description
fixes #1711 

NOTE: this will prevent race conditions only if CREATE TABLE IF NOT EXISTS is atomic. ie. on postgres it is not if table contains partitions: https://stackoverflow.com/questions/77486178/how-to-prevent-create-table-if-not-exists-fail-with-duplicate-table

still the chances of a race are way smaller

also fixes two minor issues, see commit list below

### DIFF
--- a/dlt/common/destination/capabilities.py
+++ b/dlt/common/destination/capabilities.py
@@ -76,7 +76,7 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     # use naming convention in the schema
     naming_convention: TNamingConventionReferenceArg = None
     alter_add_multi_column: bool = True
-    create_table_not_exists: bool = True
+    supports_create_table_if_not_exists: bool = True
     supports_truncate_command: bool = True
     schema_supports_numeric_precision: bool = True
     timestamp_precision: int = 6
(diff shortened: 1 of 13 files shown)


dlt/common/destination/capabilities.py
```diff
@@ -76,6 +76,7 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     # use naming convention in the schema
     naming_convention: TNamingConventionReferenceArg = None
     alter_add_multi_column: bool = True
+    create_table_not_exists: bool = True
```
VioletM, line 79: I suggest to rename it that it's clear it for `_dlt` tables:
For ex, `create_dlt_table_if_not_exists`

VioletM (dismissed):

dlt/common/destination/capabilities.py
rudolfix: this is in destination caps. but I will rename it to make clear what it means

VioletM (approved):

PR #1742, opened 2024-08-25
Enable schema evolution for `merge` write disposition with `delta` table format

### LINKED ISSUE #1739: No schema evolution with `delta` table format on `filesystem` destination with `merge` write disposition
### dlt version

0.5.4a0

### Describe the problem

Schema evolution doesn't work for the `delta` table format on the `filesystem` destination when using the `merge` write disposition. New columns (that emerge after initial creation of the table) are disregarded—they don't end up in the table.

The `append` and `replace` dispositions don't have this issue.

### Expected behavior

_No response_

### Operating system

Windows

### Runtime environment

Local

### Python version

3.8

### dlt data source

_No response_

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

jorritsandbrink:
### Description

- enables schema evolution (adding new columns) for the `merge` write disposition with the `delta` table format
- increases minimum `deltalake` version to access `add_columns` method
- allows to pass a schema name to `get_delta_tables` for pipelines with multiple schemas (that may explain problem with "missing" delta tables)

### Related Issues

Fixes #1739


rudolfix (changes requested):
Thanks for working on this! It looks good! Two things:
- if arrow_ds is empty you do not evolve the schema. IMO that should happen. please add a test for it (`if arrow_ds.head(1).num_rows == 0:`)
- should we update all table schemas like in other destinations where it happens in `update_stored_schema`? if you agree let's create a ticket for that
- same thing for truncating tables before the load. this is actually used by `refresh` option

jorritsandbrink:
@rudolfix

> if arrow_ds is empty you do not evolve the schema. IMO that should happen. please add a test for it (if arrow_ds.head(1).num_rows == 0:)

Done.

> should we update all table schemas like in other destinations where it happens in update_stored_schema? if you agree let's create a ticket for that

Three options:
1. We let delta-rs do automatic schema evolution. 
    - This already works for `write_deltalake` (which we use for the `append` and `replace` dispositions).
    - This does [not](https://github.com/delta-io/delta-rs/issues/2282) yet work for `DeltaTable.merge` (which we use for the `merge` write disposition).
    - This does [not](https://github.com/delta-io/delta-rs/issues/2686) yet work for the "empty batch case".
2. We manually manage schema evolution.
   - In this case I think using `update_stored_schema` is a good idea.
3. Mix of 1 and 2.

We currently do 3. 1 is not possible yet, but might become possible when the linked tickets are done (they are already assigned, so could be soon). 2 is possible, but is a bigger burden on our side. Which has your preference?

> same thing for truncating tables before the load. this is actually used by refresh option

Okay, then we should probably use it.

rudolfix (dismissed):
OK this is top. let me add windows fix

rudolfix:
@jorritsandbrink 

So what I'd do:
in `update_stored_schema`
1. make sure that table prefix == table dir for delta (then we know that each table has a separate "folder"). then the folder structure is good and weird layouts are eliminated

in truncate / drop tables
1. disable for delta (ie. with an error message) OR
2. implement dropping and truncating tables properly (all `refresh` options should work after that)

migrating schema
You already have all the building blocks for (2) and it IMO makes sense to migrate tables before we start loading but the priority is low.

PR #1747, opened 2024-08-26
Update Snowflake docs

akelad:
Our docs contained the wrong information, stage files in Snowflake are kept by default: https://dlthub.com/docs/api_reference/destinations/impl/snowflake/factory#snowflake-objects

Noticed by a community member: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1724413082471379

### DIFF
--- a/docs/website/docs/dlt-ecosystem/destinations/snowflake.md
+++ b/docs/website/docs/dlt-ecosystem/destinations/snowflake.md
@@ -136,7 +136,12 @@ If you set the [`replace` strategy](../../general-usage/full-loading.md) to `sta
 recreated with a [clone command](https://docs.snowflake.com/en/sql-reference/sql/create-clone) from the staging tables.
 
 ## Data loading
-The data is loaded using an internal Snowflake stage. We use the `PUT` command and per-table built-in stages by default. Stage files are immediately removed (if not specified otherwise).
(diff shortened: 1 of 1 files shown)


akelad:
@VioletM as the Snowflake expert would be great if you can check that I've formatted the config right hehe

docs/website/docs/dlt-ecosystem/destinations/snowflake.md
```diff
@@ -136,7 +136,12 @@ If you set the [`replace` strategy](../../general-usage/full-loading.md) to `sta
 recreated with a [clone command](https://docs.snowflake.com/en/sql-reference/sql/create-clone) from the staging tables.
 
 ## Data loading
-The data is loaded using an internal Snowflake stage. We use the `PUT` command and per-table built-in stages by default. Stage files are immediately removed (if not specified otherwise).
+The data is loaded using an internal Snowflake stage. We use the `PUT` command and per-table built-in stages by default. Stage files are kept by default, unless specified otherwise via the `keep_staged_files` parameter:
```
VioletM, line 139: This is true only for Snowlake named or external stage (s3/azure/gcp). In this case, dlt will keep the latest files by default unless it says otherwise via `keep_staged_files` parameter. If people does not provide any stage to dlt, it uses table stage, and I don't know what happened to these files, but I don't think you can access them at all

akelad: ahh ok so I guess I should edit this to say the stage files are removed unless you explicitly define a stage?

VioletM (approved):

Issue #1725, opened 2024-08-22
Docs | New top-level structure for sources


VioletM:
- [ ] sources and destinations at the top level
- [ ] in sources there is a clear distinction between 3 main generic sources and other verified sources
- [ ] there is a link to Write custom source yourself tutorial

PR #1750, opened 2024-08-27
Move sources and destinations to the top level

VioletM:
closes [#1725](https://github.com/dlt-hub/dlt/issues/1725)


akelad:
The sources logo is the wrong colour in dark mode:

image

Also, it's still misaligned somehow in light mode 🤔 
image

I think no point in spending too much time on that yourself, either we get Anton to figure it out or we involve Elise or someone

mariarice15 (dismissed):
Looks good! Only thing is maybe move 'custom destination' into a tip, like it is under sources. Otherwise, it may look like the users have to pick one from the list, and we don't support anything else.

VioletM:
> I think no point in spending too much time on that yourself

I already spent some time on it, so I think I know how to change it :)

akelad (approved):

PR #1762, opened 2024-08-28
Feat/1679 make dlt init use core sources

sh-rp:
### Description
Updates the dlt init command to use the core source if present. It uses all the same mechanisms as before, so secrets extractions and destination replacements still work.

What this PR does so far:
- [ ] User core sources if present (filesystem for now)
- [ ] Move generic pipeline to core
- [ ] Allows for omitting core sources
- [ ] Updates wording in a hopefully useful way
- [ ] Add a note about setting up the custom destination in the correct way
 - [ ] Update the command that lists the available sources to show the core sources also
 
Follow up ToDos:
- [ ] Update generic pipeline template and add test for it (or figure out what templates there should be)
- [ ] Allow modifiers for sources, e.g. dlt init filesystem:az duckdb will create credentials for azure


dlt/cli/init_command.py
```diff
@@ -38,10 +41,16 @@
 from dlt.cli.exceptions import CliCommandException
 from dlt.cli.requirements import SourceRequirements
 
+
 DLT_INIT_DOCS_URL = "https://dlthub.com/docs/reference/command-line-interface#dlt-init"
 DEFAULT_VERIFIED_SOURCES_REPO = "https://github.com/dlt-hub/verified-sources.git"
 INIT_MODULE_NAME = "init"
 SOURCES_MODULE_NAME = "sources"
+SKIP_CORE_SOURCES_FOLDERS = [
```
rudolfix, line 49: should we declare this in `dlt.sources`?

dlt/cli/init_command.py
```diff
@@ -234,15 +243,30 @@ def init_command(
...
-    fmt.echo("Looking up the init scripts in %s..." % fmt.bold(repo_location))
-    clone_storage = git.get_fresh_repo_files(repo_location, get_dlt_repos_dir(), branch=branch)
-    # copy init files from here
-    init_storage = FileStorage(clone_storage.make_full_path(INIT_MODULE_NAME))
-    # copy dlt source files from here
-    sources_storage = FileStorage(clone_storage.make_full_path(SOURCES_MODULE_NAME))
-    # load init module and get init files and script
-    init_module = load_script_module(clone_storage.storage_path, INIT_MODULE_NAME)
+    # lookup core sources
+    local_path = Path(os.path.dirname(os.path.realpath(__file__))).parent / SOURCES_MODULE_NAME
```
rudolfix, line 247: if explict repo location is present, disable the local sources. that allows to still use the old version or your own fork of verified sources

dlt/sources/.gitignore
```diff
@@ -0,0 +1,10 @@
+# ignore secrets, virtual environments and typical python compilation artifacts
```
rudolfix, line 1: I'm not sure this will be distributed with the pypi package? should we place all non-importable code, templates and files in some kind of "data" folder?

rudolfix (commented):
looks good! should we place all templates and non-importable python code (ie demo pipelines in separate "data" folder)?

rudolfix:
@sh-rp let's move forward! I'm particularly interested how you'll deal with more complex templates. currently our demo pipelines are runnable code but in the core they do not need to be... so we do not need to use AST to rewrite them.

dlt/cli/init_command.py
sh-rp: I have added a flag for this now and amended some of the output to guide the user

dlt/cli/init_command.py
sh-rp: I have moved it to the same location where the verified sources config is for now

docs/website/docs/reference/command-line-interface.md
```diff
@@ -23,9 +23,9 @@ version if run again with existing `source` name. You are warned if files will b
 ### Specify your own "verified sources" repository.
 You can use `--location <repo_url or local folder>` option to specify your own repository with sources. Typically you would [fork ours](https://github.com/dlt-hub/verified-sources) and start customizing and adding sources ie. to use them for your team or organization. You can also specify a branch with `--branch <name>` ie. to test a version being developed.
 
-### List all verified sources
+### List all sources
 ```sh
-dlt init --list-verified-sources
+dlt init --list-sources
```
rudolfix, line 28: should we add new flags and some information on core and verified sources?

rudolfix (approved):
LGTM! we need a better place for the templates (`init` folder) but good for now.
you can add more info on core/verified sources to cli reference (ie. `omit` flag)

PR #1770, opened 2024-08-29
historic docs v2

sh-rp:
### Description
PR for adding historic docs without checking these historic docs into the repo. Mechanism:

* Discover all tags and find all the most recent minor versions (above a certain configurable threshold)
* Check out each tag in a separate folder and build the docs at that point in time and copy them into the main project
* Current repo version will be the devel version, so we should only deploy docs from devel.
* Will deploy the master branch as the version for the current highest release instead of the tag
* Prevents indexing by google via header tags
* Ensure algolia does not index historic versions (via algolia console, you can explicitely say to respect noindex header tags)

TODO after merging:
* Switch of devel deployment on netfliy, we might need to update the paths in the website that point the /docs/devel path to that deployment.


docs/website/tools/update_versions.js
```diff
@@ -0,0 +1,99 @@
...
+
+console.log(`Found ${versions.length} elligible versions`)
+if (versions.length < 5) {
+    console.error("Sanity check failed, not enough elligble version tags found")
+    process.exit(1)
+}
+
+// go through the versions and find all newest versions of minor versions
+const selectedVersions = ["master"];
+let lastVersion = versions[0];
```
rudolfix, line 46: can we add `0.5.4 (latest)` to the top tag?

docs/website/tools/update_versions.js
```diff
@@ -0,0 +1,99 @@
...
+fs.rmSync("versions.json", { force: true })
+
+fs.mkdirSync(VERSIONED_DOCS_FOLDER);
+fs.mkdirSync(VERSIONED_SIDEBARS_FOLDER);
+
+// check that checked out repo is on devel
+console.log("Checking branch")
+const branch = proc.execSync(`cd ${REPO_DIR} && git rev-parse --abbrev-ref HEAD`).toString().trim()
+
+if (branch != "devel") {
```
rudolfix, line 68: I think I do not understand this part. We should not use the REPO_DIR to build the current (development) docs. What IMO we should do:

- what we call `devel` we should build using the already checked out repo including uncommitted local changes. this will also allow to build previews for PRs (or this already works like that? because we still build local docs)
- the highest tag found should not be built from this tag but from `master`. this will allow to update docs in master branch and get then online without increasing the tag

rudolfix (changes requested):
Cool! Let's merge it and then we can enable it on netlify even before 1.0 release. Both master and devel branch changes should trigger the same build (as we do both devel and master + tags in each)

docs/website/tools/update_versions.js
sh-rp: Yes, this is how it works, this line only is a sanity check wether the repo could be checked out properly. the devel branch of the checked out repo never actually is built, check further down, it only ever builds stuff collected in the versions.

docs/website/tools/update_versions.js
```diff
@@ -0,0 +1,99 @@
...
+
+// const
+REPO_DIR = ".dlt-repo"
+REPO_DOCS_DIR = REPO_DIR + "/docs/website"
+REPO_URL = "https://github.com/dlt-hub/dlt.git"
+VERSIONED_DOCS_FOLDER = "versioned_docs"
+VERSIONED_SIDEBARS_FOLDER = "versioned_sidebars"
+
+// no doc versions below this version will be deployed
+MINIMUM_SEMVER_VERSION = "0.4.0"
```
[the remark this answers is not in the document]
sh-rp: could we keep it until the 1.0 release?

PR #1771, opened 2024-08-29
Don't use Custom Embedding Functions

### LINKED ISSUE #1765: LanceDB destination: can't query generated tables
### dlt version

0.5.3

### Describe the problem

# Problem
Vector search on a LanceDB table generated by dlt is broken. Here's a simple query:

```python
import os
import lancedb

dlt_lancedb_uri = os.environ["DESTINATION__LANCEDB__CREDENTIALS__URI"]
lancedb_con = lancedb.connect(dlt_lancedb_uri)
lancedb_table = lancedb_con.open_table("my_table")
query = lancedb_table.search("My very important question")
results = query.to_list()
```

### Expected behavior

Since I expect downstream users to import `lancedb` and not worry about how data is ingested, dlt should adopt a different strategy for embedding function registration:

1. Add to the docs the need to `import dlt.destinations.impl.lancedb.models` before querying data from lancedb
2. Have `dlt/destinations/impl/lancedb/__init__.py` import `dlt/destinations/impl/lancedb/models.py`. This should enable `dlt.destinations.lancedb` to be sufficient (I believe?)
3. Avoid needing a custom `PatchedOpenAIEmbeddings` and rely on natively-supported lancedb functions
4. Collaborate with `lancedb` to modify the stored `pyarrow.Schema`'s metadata to include the required module imports (i.e., the `dlt ` submodule). This would add the embedding function to the LanceDB registry at deserialization *before* trying to retrieve the function from the registry,

This error was painful to debug, because nothing points to `dlt` being the source. Renaming the registered function to `openai_dlt_patch` would be of great help

### Steps to reproduce

1. Configure the `lancedb` destination (credentials, embedding function, etc.)
2. Use the `lancedb` destination with the `lancedb_adapter` to ingest data
3. In a separate process (script, notebook, REPL, etc.), import `lancedb` only and access a generated table that has an embed column.
4. Query the table using lancedb's `.search()` (vector search)
5. It should fail saying `"openai_patched"` is not in registry

Fix
6. import `dlt.destinations.impl.lancedb.models`
7. retry steps 4 and 5 and it should now be working

By manually checking the LanceDB embedding function registry, you can see the `"openai_patched"` function defined by dlt being registered.

```python
import lancedb.embeddings.registry as embedding_registry_module

registry = embedding_registry_module.get_registry()
registry._functions  # dictionary of {func_name: func} of type Dict[str, Callable]
registry.get("openai_patched")
```

dlt code: https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/lancedb/models.py
lancedb code: https://github.com/lancedb/lancedb/blob/main/python/python/lancedb/embeddings/registry.py

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

Not relevant

### dlt destination

_No response_

### Other deployment details

dlt destination is `lancedb`

### Additional information

_No response_

Pipboyguy:
### Description

OpenAI embedding service doesn't accept empty string bodies. We used to deal with this by overriding the whole `OpenAIEmbedding` function. 

This caused more grief than it fixed since the LanceDB registry doesn't keep track of it well, with very finicky Arrow metadata parsing and de-serialisation.

We simplify this fix by simply replacing empty strings with a placeholder that should be very semantically dissimilar to 99.9% of queries. Ideally, the null strings' embedding vectors themselves should be pinned at the origin, but this should be handled by upstream LanceDB.   

The default vector column name is also changed to simply "vector" to coincide with LanceDB's default vector name to make onboarding and setup easier.

### Related Issues

- Fixes #1765

### Additional Context

See  https://github.com/lancedb/lancedb/issues/1577


dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -731,6 +720,19 @@ def run(self) -> None:
         with FileStorage.open_zipsafe_ro(self._file_path) as f:
             records: List[DictStrAny] = [json.loads(line) for line in f]
 
+        # Replace empty strings with placeholder string if OpenAI is used.
```
zilto, line 723: Can't tell the impact on performance, but I think it's a good fix until there's progress on the LanceDB issue! 

I don't know how frequently you'd hit an empty string when embedding, but it might be worth mentioning in the docs?

tests/load/lancedb/utils.py
```diff
@@ -52,7 +52,7 @@ def assert_table(
         "_dlt_id",
         "_dlt_load_id",
         dlt.config.get("destination.lancedb.credentials.id_field_name", str) or "id__",
-        dlt.config.get("destination.lancedb.credentials.vector_field_name", str) or "vector__",
+        dlt.config.get("destination.lancedb.credentials.vector_field_name", str) or "vector",
```
zilto, line 55: I think using `vector` is nice because it aligns with the lancedb defaults.

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -81,6 +78,7 @@
 
 TIMESTAMP_PRECISION_TO_UNIT: Dict[int, str] = {0: "s", 3: "ms", 6: "us", 9: "ns"}
 UNIT_TO_TIMESTAMP_PRECISION: Dict[str, int] = {v: k for k, v in TIMESTAMP_PRECISION_TO_UNIT.items()}
+EMPTY_STRING_PLACEHOLDER = "__EMPTY_STRING_PLACEHOLDER__"
```
rudolfix, line 81: use some random string. who knows what kind of tokenizer may be used against it... openAI may embed this as separate words

dlt/destinations/impl/lancedb/lancedb_client.py
rudolfix: @Pipboyguy didn't we switch the format to parquet? I think it is in PR that is still in review. anyway we'll be able to use `pa.compute`  to replace those soon

Pipboyguy: @rudolfix yes indeed, it does make it a bit tricky to implement a fix considering the switch in format.

Pipboyguy: @zilto agreed, will add a doc entry for this!

dlt/destinations/impl/lancedb/lancedb_client.py
Pipboyguy: Ahh good point! You're right I'll replace with randomly gen string

rudolfix (approved):
LGTM!

Issue #1749, opened 2024-08-27
abort load package and raise exception on terminal errors in jobs


rudolfix:
**Background**
by default `dlt` will move terminally failed jobs into `failed_jobs` folder and will not raise any exceptions. users can change `load.raise_on_failed_jobs=true` config option to abort the package and raise exception afterwards.

this PR makes this behavior default.

background: https://dlthub.com/docs/running-in-production/running#handle-exceptions-failed-jobs-and-retry-the-pipeline

**Requirements**
PR 1:
1. Switch `raise_on_failed_jobs` to true
2. Update run in production docs
3. Update failing tests. There may be a lot of them. Add unit tests to `test_dummy_client` (I hope we test this config flag somewhere in it)

PR 2:
4. Add a new cli/pipeline method to retry aborted package (moves failed jobs back to new). extend this command:
https://dlthub.com/docs/reference/command-line-interface#get-the-load-package-information
and add custom parser with  "retry-aborted". which:
- will remove "aborted" flag and move package from "completed" to "normalized"
- move all failed jobs back to new jobs and delete failed messages.

willi-mueller:
@rudolfix 
The only downside of this change might be that the default strategy for the `write_position = "replace"` would truncate the table and not add any data because the load package failed.

Could we change the default strategy to ' insert-from-staging ' to minimize the chances of users deleting their data on error?

Background: https://dlthub.com/docs/general-usage/full-loading#the-truncate-and-insert-strategy

PR #1781, opened 2024-09-03
Feat/1749 abort load package and raise exception on terminal errors in jobs

willi-mueller:
### Related Issues

- first part of #1749


docs/website/docs/running-in-production/running.md
```diff
@@ -259,9 +259,21 @@ def check(ex: Exception):
...
-You may inspect if the failed jobs are present by checking the load info as follows:
+If any job in the package **fails terminally** it will be moved to `failed_jobs` folder and assigned
+such status.
+By default, **an exceptions is raised** and on the first failed job, the load package will be aborted with `LoadClientJobFailed` (terminal exception).
+Such package  will be completed but its load id is not added to the `_dlt_loads` table.
+All the jobs that were running in parallel are completed before raising. The dlt state, if present, will not be visible to `dlt`.
+Here is an example `config.toml` to disable this behavior:
+
+```toml
+# you should really load just one job at a time to get the deterministic behavior
```
rudolfix, line 270: let's remove workers=1 thing

rudolfix (changes requested):
How we need to test it
1. there are tests in dummy client that are testing abort feature, there just remove the flag. it is by default on
2. make the other tests to pass by disabling the flag

look for `raise_on_failed_jobs` in the whole code
we use it to raise on load info, those are not needed anymore and can be removed (many of those)

I see that we set this flag in airflow helper. we should change the default there and also we never set it to false

there are other cases. you'll find them for sure

docs/website/docs/running-in-production/running.md
willi-mueller: Done

willi-mueller: Done

docs/website/docs/reference/performance_snippets/performance-snippets.py
```diff
@@ -179,9 +179,7 @@ async def _run_async():
                 loop.run_in_executor(executor, _run_pipeline, pipeline_1, async_table),
                 loop.run_in_executor(executor, _run_pipeline, pipeline_2, defer_table),
             )
-        # result contains two LoadInfo instances
-        results[0].raise_on_failed_jobs()
-        results[1].raise_on_failed_jobs()
+        # results contains two LoadInfo instances
```
willi-mueller, line 182: Without the references into the `results` list this example is less instructive. I am not sure whether to:
1. leave only the comment
2. leave the calls to `results[n].raise_on_failed_jobs()`
3. call some other method/property, e.g. print(results[0]) or `results[0].asdict()`

dlt/helpers/airflow_helper.py
```diff
@@ -86,7 +86,7 @@ def __init__(
         This happens **after all dlt loading jobs executed**. See more here: https://dlthub.com/docs/running-in-production/running#failed-jobs
 
         `abort_task_if_any_job_failed` will abort the other dlt loading jobs and fail the Airflow task in any of the jobs failed. This may put your warehouse in
-        inconsistent state so the option is disabled by default.
+        inconsistent state. See https://dlthub.com/docs/running-in-production/running#handle-exceptions-failed-jobs-and-retry-the-pipeline.
```
willi-mueller, line 89: Setting the destination to an inconsistent state sounds scary. However, it's a necessary inconvenience, correct?

tests/pipeline/test_pipeline.py
```diff
@@ -949,7 +951,7 @@ def fail_extract():
     assert py_ex.value.step == "extract"
 
     os.environ["COMPLETED_PROB"] = "0.0"
-    os.environ["RAISE_ON_FAILED_JOBS"] = "true"
+    os.environ["RAISE_ON_FAILED_JOBS"] = "true"  # TODO: why is this necessary?
```
willi-mueller, line 954: Without setting this env, the pipeline does not raise. However, it should be the default. Thus, the implementation must be incomplete.

rudolfix (approved):
LGTM!

PR #1782, opened 2024-09-03
controls row groups and empty tables

rudolfix:
### Description
See commit list and docs


docs/website/docs/dlt-ecosystem/file-formats/parquet.md
```diff
@@ -35,6 +35,7 @@ Under the hood, `dlt` uses the [pyarrow parquet writer](https://arrow.apache.org
 - `flavor`: Sanitize schema or set other compatibility options to work with various target systems. Defaults to None which is **pyarrow** default.
 - `version`: Determine which Parquet logical types are available for use, whether the reduced set from the Parquet 1.x.x format or the expanded logical types added in later format versions. Defaults to "2.6".
 - `data_page_size`: Set a target threshold for the approximate encoded size of data pages within a column chunk (in bytes). Defaults to None which is **pyarrow** default.
+- `row_group_size`: Set the number of rows in a row group - see remarks below, because `pyarrow` does not handle this setting like you would expect.
```
willi-mueller, line 38:
```suggestion
- `row_group_size`: Set the number of rows in a row group. [See here](#row-group-size) how this can optimize parallel processing of queries on your destination over the default setting of `pyarrow`.
```

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
```diff
@@ -76,3 +77,16 @@ You can generate parquet files without timezone adjustment information in two wa
 2. Set the **timestamp_timezone** to empty string (ie. `DATA_WRITER__TIMESTAMP_TIMEZONE=""`) to generate logical type without UTC adjustment.
 
 To our best knowledge, arrow will convert your timezone aware DateTime(s) to UTC and store them in parquet without timezone information.
+
+
+### Row group size
+`pyarrow` parquet writer writes each item (be it table or batch which is internally converted to table) in a separate row group. This may lead to many small row groups
+to be created which may not be optimal for certain query engines (ie. `duckdb` parallelizes on a row group). `dlt` allows to control the size of the row group by
+buffering and concatenating (0 copy) tables/batches before they are written. You can control the size by setting the buffer size
```
willi-mueller, line 85:
```suggestion
The `pyarrow` parquet writer writes each item, i.e. table or record batch, in a separate row group. 
This may lead to many small row groups which may not be optimal for certain query engines. For example, `duckdb` parallelizes on a row group.
`dlt` allows controlling the size of the row group by
buffering and concatenating tables and batches before they are written. The concatenation is done as a zero-copy to save memory.
You can control the memory needed by setting the count of records to be buffered as follows:
```

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
```diff
@@ -76,3 +77,16 @@ You can generate parquet files without timezone adjustment information in two wa
...
+```toml
+[extract.data_writer]
+buffer_max_items=1000000
+```
+Mind that we must hold the tables in memory. 1 000 000 rows in example above may take quite large amount of it.
```
willi-mueller, line 90:
```suggestion
Mind that `dlt` holds the tables in memory. Thus, 1,000,000 rows in the example above may consume a significant amount of RAM.
```

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
```diff
@@ -76,3 +77,16 @@ You can generate parquet files without timezone adjustment information in two wa
...
 To our best knowledge, arrow will convert your timezone aware DateTime(s) to UTC and store them in parquet without timezone information.
+
+
+### Row group size
+`pyarrow` parquet writer writes each item (be it table or batch which is internally converted to table) in a separate row group. This may lead to many small row groups
+to be created which may not be optimal for certain query engines (ie. `duckdb` parallelizes on a row group). `dlt` allows to control the size of the row group by
+buffering and concatenating (0 copy) tables/batches before they are written. You can control the size by setting the buffer size
+```toml
+[extract.data_writer]
+buffer_max_items=1000000
```
willi-mueller, line 88: I prefer the scientific notation when we've got a lot of zeros. Wdyt?
```suggestion
buffer_max_items=10e6
```

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
```diff
@@ -76,3 +77,16 @@ You can generate parquet files without timezone adjustment information in two wa
...
+```
+Mind that we must hold the tables in memory. 1 000 000 rows in example above may take quite large amount of it.
+
+`row_group_size` has limited utility with `pyarrow` writer. It will split large tables into many groups if set below item buffer size.
```
willi-mueller, line 92: This sentence is not clear to me yet. Is the instruction to the user something like the following?

```suggestion
For the `pyarrow` parquet writer, ensure to have`row_group_size >= buffer_max_items`. Otherwise, your destination might have more row groups than optimal.
```

willi-mueller (changes requested):
Reviewing the docs first to make sure I understood the topic.

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
rudolfix: ohhh so actually the reverse is true. `row_group_size < buffer_max_item` to have any effect. this is the core of the problem I'm fixing here. pyarrow will create row group of size of parquet table being written or smaller. btw. other, well desgined implementations allow to write batches to the same groups. not here

dlt/common/data_writers/buffered.py
```diff
@@ -117,8 +117,11 @@ def write_data_item(self, item: TDataItems, columns: TTableSchemaColumns) -> int
...
                 new_rows_count = 1
         self._buffered_items_count += new_rows_count
-        # flush if max buffer exceeded
-        if self._buffered_items_count >= self.buffer_max_items:
+        # flush if max buffer exceeded, the second path of the expression prevents empty data frames to pile up in the buffer
+        if (
+            self._buffered_items_count >= self.buffer_max_items
+            or len(self._buffered_items) >= self.buffer_max_items
+        ):
             self._flush_items()
```
willi-mueller, line 125: Could we eliminate the comment by refactoring it to a method?
```suggestion
        self.flush_if_max_buffer_exceeded()
```

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
```diff
@@ -76,3 +77,18 @@ You can generate parquet files without timezone adjustment information in two wa
...
+`dlt` allows controlling the size of the row group by
+buffering and concatenating tables and batches before they are written. The concatenation is done as a zero-copy to save memory.
+You can control the memory needed by setting the count of records to be buffered as follows:
+```toml
+[extract.data_writer]
+buffer_max_items=10e6
+```
+Mind that `dlt` holds the tables in memory. Thus, 1,000,000 rows in the example above may consume a significant amount of RAM.
+
+`row_group_size` has limited utility with `pyarrow` writer. It will split large tables into many groups if set below item buffer size.
```
willi-mueller, line 94: Got it now, thanks! Maybe we can give a recommendation like this:
```suggestion
Setting `row_group_size` has limited utility with the `pyarrow` parquet writer because large source tables can end up fragmented into too many groups.
Thus, we recommend setting `row_group_size < buffer_max_items` only when the write_disposition is `"replace"`.
For all other write dispositions, we recommend the default `row_group_size` to avoid fragmentation.
```

dlt/common/data_writers/buffered.py
willi-mueller: Similarly, `_update_row_count(item)` might be neat.

dlt/common/data_writers/writers.py
```diff
@@ -474,17 +474,33 @@ def write_header(self, columns_schema: TTableSchemaColumns) -> None:
...
+        tables = []
         for row in rows:
-            if not self.writer:
-                self.writer = self._create_writer(row.schema)
-            if isinstance(row, pyarrow.Table):
-                self.writer.write_table(row, row_group_size=self.parquet_row_group_size)
-            elif isinstance(row, pyarrow.RecordBatch):
-                self.writer.write_batch(row, row_group_size=self.parquet_row_group_size)
+            self.items_count += row.num_rows
+            if isinstance(row, pyarrow.RecordBatch):
```
willi-mueller, line 487: **Blocking:** I find this surprising How can a row be a RecordBatch or Table? How can a row have `num_rows`?
Could we call it `item` like in the docs?

dlt/common/data_writers/writers.py
```diff
@@ -474,17 +474,33 @@ def write_header(self, columns_schema: TTableSchemaColumns) -> None:
...
+                    batches = []
+                tables.append(row)
             else:
                 raise ValueError(f"Unsupported type {type(row)}")
-            # count rows that got written
-            self.items_count += row.num_rows
+        if batches:
+            tables.append(pyarrow.Table.from_batches(batches))
+
+        table = pyarrow.concat_tables(tables, promote_options="none")
```
willi-mueller, line 499: I would find it easier to understand if we extract this into a method: `self._concat_items(items)`

dlt/common/data_writers/writers.py
```diff
@@ -474,17 +474,33 @@ def write_header(self, columns_schema: TTableSchemaColumns) -> None:
...
+        if not self.writer:
+            self.writer = self._create_writer(table.schema)
+        # write concatenated tables, "none" options ensures 0 copy concat
```
willi-mueller, line 502: that's an important remark! Glad you put it!

willi-mueller (changes requested):
Great fix! I put some ideas that would help me read the code more easily and docs refactoring.
Just one blocking comment on a naming. Otherwise fine!

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
rudolfix: row groups do not map to write dispositions like this... I think this is only relevant to advanced users that optimize their parquet files for a particular query engine...

rudolfix:
@willi-mueller I fixed most of the suggested code changes and some docs changes... I think we are good for now. thanks, good review :)

PR #1800, opened 2024-09-11
Fix typo "frequenly"

ruudwelten:
### Description
Just a small typo fix: "frequenly" > "frequently". Would have combined multiple but this was the only one I found.


burnash (approved):
Thanks for spotting and fixing this @ruudwelten!

PR #1803, opened 2024-09-11
skips tables without jobs when merging delta tables

rudolfix:
### Description
1. it was assumed that all tables in table chain will have a job. that is not true for nested tables where child tables have no rows
2. schema evolution: presence is checked via names, not full schema comparison
3. "heavy" object are explicitly deleted. not sure if that helps a lot... internally delta does a lot of unnecessary conversions on the arrow objects

### DIFF
--- a/tests/load/pipeline/test_merge_disposition.py
+++ b/tests/load/pipeline/test_merge_disposition.py
@@ -9,6 +9,7 @@ import dlt
 
 from dlt.common import json, pendulum
 from dlt.common.configuration.container import Container
+from dlt.common.destination.utils import resolve_merge_strategy
 from dlt.common.pipeline import StateInjectableContext
 from dlt.common.schema.utils import has_table_seen_data
 from dlt.common.schema.exceptions import (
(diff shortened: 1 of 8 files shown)


tests/load/pipeline/test_merge_disposition.py
```diff
@@ -261,6 +264,143 @@ def r(data):
...
+        if destination_config.table_format == "delta":
+            # delta merges cannot delete from nested tables
+            assert table_counts == {
+                "parent": 3,  # id == 3 not deleted (not present in the data)
+                "parent__child": 3,  # child not deleted
+                "parent__child__grandchild": 3,  # grand child not deleted,
+            }
+        else:
+            assert table_counts == {
+                "parent": 3,  # id == 3 not deleted (not present in the data)
```
jorritsandbrink, line 344: Parent is not deleted but corresponding childs are. Seems odd; is this desired behavior?

jorritsandbrink (approved):
LGTM! Thanks for `delta` improvements.

tests/load/pipeline/test_merge_disposition.py
jorritsandbrink: You're right. Seems fine now that I look at it again.

PR #1818, opened 2024-09-16
incremental `scd2` with `merge_key`

### LINKED ISSUE #1789: Support incremental loading for scd2
### Feature description

Currently, the [scd2 write disposition](https://dlthub.com/docs/general-usage/incremental-loading#scd2-strategy) requires running a full extract of the source data in order to work; in other words, it does not support incremental loads.  The main justification for this seems to be that doing a full extract is the only way to detect hard deletes.  While true, there are plenty of situations where we would like to perform incremental scd2 loading where hard deletes are not a concern.  At the moment, the workaround we have is to do a regular merge incremental load, followed by a dbt snapshots.  But that effectively doubles the amount of storage needed in our warehouse and introduces an extra computational step.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

We have data in a SaaS product that contains large, mutable data and the vendor enforces rate limiting.  Doing daily or more frequent regular full extracts is not feasible or cost effective.  Having an incremental scd2 solution (even if it didn't manage hard deletes) would be very valuable to us.

### Proposed solution

The scd2 write disposition supports incremental merge, with the caveat that hard deletes are not supported when doing so.

### Related issues

_No response_

jorritsandbrink:
### Description
Adds `merge_key` support for `scd2` to enable incremental extracts.

When a `merge_key` is provided, absent records are only retired if their `merge_key` value is present in the source extract.

```py
@dlt.resource(
    merge_key="my_natural_key",
    write_disposition={"disposition": "merge", "strategy": "scd2"}
)
def my_incremental_dim_data():
    ...
```

### Related Issues

Closes #1789


tests/load/pipeline/test_scd2.py
```diff
@@ -717,6 +715,102 @@ def r(data):
...
+
+    # load 3 — natural key 2 is absent, natural key 1 has changed
+    dim_snap = [
+        {"nk": 1, "foo": "bar"},
+    ]
+    info = p.run(r(dim_snap), **destination_config.run_kwargs)
+    assert_load_info(info)
+    assert load_table_counts(p, "dim_test")["dim_test"] == 3
+    boundary_ts = get_load_package_created_at(p, info)
+    # natural key 1 should now have two records (one retired, one active)
```
sh-rp, line 770: we should also test here that when the row with natural key 1 reverts back to the first version (and has the same row hash as it had in the first load) it will still create a new entry.

dlt/common/schema/typing.py
```diff
@@ -231,15 +231,22 @@ class TWriteDispositionDict(TypedDict):
...
+class TMergeDispositionDict(TWriteDispositionDict):
     strategy: Optional[TLoaderMergeStrategy]
+
+
+class TScd2StrategyDict(TMergeDispositionDict, total=False):
     validity_column_names: Optional[List[str]]
     active_record_timestamp: Optional[TAnyDateTime]
     boundary_timestamp: Optional[TAnyDateTime]
     row_version_column_name: Optional[str]
+    retire_if_absent: Optional[bool]
```
sh-rp, line 243: maybe call this "retire_absent_rows"? then it is a bit more clear just from the naming what it does. wdyt?

tests/load/pipeline/test_scd2.py
jorritsandbrink: done

dlt/common/schema/typing.py
jorritsandbrink: Yes! Changed it.

dlt/common/schema/typing.py
```diff
@@ -231,15 +231,22 @@ class TWriteDispositionDict(TypedDict):
...
     strategy: Optional[TLoaderMergeStrategy]
+
+
+class TScd2StrategyDict(TMergeDispositionDict, total=False):
     validity_column_names: Optional[List[str]]
     active_record_timestamp: Optional[TAnyDateTime]
     boundary_timestamp: Optional[TAnyDateTime]
     row_version_column_name: Optional[str]
+    retire_absent_rows: Optional[bool]
+    natural_key: Optional[str]
```
sh-rp, line 244: this is no longer needed, is it?

dlt/extract/hints.py
```diff
@@ -452,10 +453,19 @@ def _merge_merge_disposition_dict(dict_: Dict[str, Any]) -> None:
...
-            dict_["x-boundary-timestamp"] = md_dict["boundary_timestamp"]
-        # add columns for `scd2` merge strategy
+
         if merge_strategy == "scd2":
+            md_dict = cast(TScd2StrategyDict, md_dict)
+            if "boundary_timestamp" in md_dict:
+                dict_["x-boundary-timestamp"] = md_dict["boundary_timestamp"]
+            if "retire_absent_rows" in md_dict:
+                dict_["x-retire-absent-rows"] = md_dict["retire_absent_rows"]
+            if "natural_key" in md_dict:
```
sh-rp, line 463: also no longer needed

docs/website/docs/general-usage/incremental-loading.md
```diff
@@ -300,6 +300,23 @@ pipeline.run(dim_customer())  # third run — 2024-04-10 06:45:22.847403
...
+        "disposition": "merge",
+        "strategy": "scd2",
+        "retire_absent_rows": False,
+    }
+)
+def dim_customer():
+    ...
+...
+```
+Using this setting, records are not retired in the destination if their corresponding natural keys are not present in the source extract. This allows for incremental extracts that only contain updated records. You need to specify the natural key as `merge_key` when `retire_absent_rows` is `False`. Compound natural keys are allowed and can be specified by providing a list of column names as `merge_key`.
```
sh-rp, line 318: @rudolfix we are using only the merge_key as opposed to a combination of primary key and merge key here, let us know if this is ok.

sh-rp (changes requested):
Small changes requested and one question that @rudolfix should answer.

dlt/common/schema/typing.py
jorritsandbrink: Whoops. Removed now.

dlt/extract/hints.py
jorritsandbrink: also removed

docs/website/docs/general-usage/incremental-loading.md
rudolfix: @jorritsandbrink @sh-rp I think there's some kind of conceptual confusion. `merge_key` is correct but it is not a natural key for the source data. it is the same `merge_key` we have in `delete-insert` strategy. 

Example 1: I load data day by day. I set the merge key to `day` column. I load the same day twice. Then I retire records only from the given day.
Example 2: I set the merge key to the same value as primary (natural) key (or to the content hash if no natural key is present). Then I have the behavior as I'd set the `retire_absent_rows` to False. (which is not needed anymore btw.)
Example 3: I update chunked documents and use merge key to be doc_id. Then I retire all the missing chunks for those doc ids, not touching the others.

My take:
1. drop the flag
2. if the merge key is present, always limit the set of records to retire
3. update the documentation to explain two basic cases (do not retire deleted records at all, retire deleted records only for given partition)

WDYT? IMO this is really powerful functionality if done this way

jorritsandbrink: I think you're right :) I'll adapt per your suggestions.

jorritsandbrink:
@rudolfix I've adapted the code per your suggestions. Can you review?

rudolfix (approved):
this is so good! thanks!

PR #1838, opened 2024-09-18
feat: filesystem delete old pipeline state files

### LINKED ISSUE #1657: compact dlt pipeline state table in filesystem destination
**Background**
We started to store pipeline state and schemas in `filesystem` destination to be able to restore them. In case of state table each load may produce a new file. We should truncate old states ie. to just keep 100 newest.

**Requirements**
1. when load is completed, delete old files in the `_dlt_pipeline_state` table
2. keep N last state files
3. delete only the state files that corresponds to finished loads (they have corresponding completed entry). this is to prevent a rate case when we have 100 unsuccessful partial loads and we delete the last right state
4. make this mechanism on by default and allow it to be disabled with a flag

donotpush:
### Description

Keep only the latest 100 pipeline state files

### Related Issues
- Resolves #1657


dlt/common/storages/configuration.py
```diff
@@ -145,6 +145,8 @@ class FilesystemConfiguration(BaseConfiguration):
     kwargs: Optional[DictStrAny] = None
     client_kwargs: Optional[DictStrAny] = None
     deltalake_storage_options: Optional[DictStrAny] = None
+    enable_state_cleanup: bool = True
```
sh-rp, line 148: This should be a config on the filesystem destination as it is not applied anywhere else. I would not have the "enable" flag, but interpret max_state_files=None as this feature being disabled.

dlt/common/storages/configuration.py
```diff
@@ -170,6 +172,9 @@ def on_resolved(self) -> None:
         if self.is_local_path(self.bucket_url):
             self.bucket_url = self.make_file_url(self.bucket_url)
 
+        if self.max_state_files <= 1:
```
sh-rp, line 175: I think it should be allowed for this to be 1, no?

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -520,6 +520,31 @@ def _get_state_file_name(self, pipeline_name: str, version_hash: str, load_id: s
...
+            ]
+
+            # sort state file info by load_id in descending order
+            state_file_info.sort(key=lambda x: x["load_id"], reverse=True)
+
+            # keeping only the most recent MAX_STATE_HISTORY files
+            files_to_delete = state_file_info[self.config.max_state_files :]
+
+            # delete the old files
+            for file_info in files_to_delete:
```
sh-rp, line 545: I think with this logic you are actually deleting the newest 100 files, no? These are the ones that should be kept.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -520,6 +520,31 @@ def _get_state_file_name(self, pipeline_name: str, version_hash: str, load_id: s
             f"{pipeline_name}{FILENAME_SEPARATOR}{load_id}{FILENAME_SEPARATOR}{self._to_path_safe_string(version_hash)}.jsonl",
         )
 
+    def _cleanup_pipeline_states(self, pipeline_name: str) -> None:
```
sh-rp, line 523: In the ticket it says that we should make sure to not delete state files attached to failed loads, but we are not saving state on failed loads, so we should be good here.

tests/load/pipeline/test_filesystem_pipeline.py
```diff
@@ -1254,6 +1255,64 @@ def my_resource_inc(prim_key=dlt.sources.incremental("id")):
...
+    os.environ["DESTINATION__FILESYSTEM__MAX_STATE_FILES"] = "1"
+
+    with pytest.raises(Exception, match="The max_state_files must be grater than 1."):
+        p = destination_config.setup_pipeline(
+            "p1", dataset_name=destination_config.destination_name
+        )
+        p.run([1, 2, 3])
+
+
+@pytest.mark.parametrize(
```
sh-rp, line 1273: I would parametrize DESTINATION__FILESYSTEM__MAX_STATE_FILES to [None, 10, 30] so see that the settings works

dlt/destinations/impl/filesystem/filesystem.py
donotpush: Yes, indeed. @rudolfix specified:

```
delete only the state files that corresponds to finished loads (they have corresponding completed entry). 
this is to prevent a rate case when we have 100 unsuccessful partial loads and we delete the last right state
```

will partial loads store a state file?

donotpush:
@sh-rp I have implemented the requested changes and added additional tests. I have also documented these tests.

The only divergence from your comments concerns the use of None for the integer max_state_files. I encountered some errors with None, so I decided to disable cleanup proces when max_state_files is set to 0 or negative values. This change is properly documented in the codebase.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -476,7 +476,9 @@ def _to_path_safe_string(self, s: str) -> str:
         """for base64 strings"""
         return base64.b64decode(s).hex() if s else None
 
-    def _list_dlt_table_files(self, table_name: str) -> Iterator[Tuple[str, List[str]]]:
+    def _list_dlt_table_files(
```
sh-rp, line 479: you have changed the bevahior of this function but not updated the other places where it is used, this will no longer just list the files of the current pipeline by default, so please double check that there are no suprising side effects :)

sh-rp (changes requested):
Looks very good, just one concern mentioned above.

dlt/destinations/impl/filesystem/filesystem.py
donotpush: The function behaviour hasn't change for the other places because `pipeline_name` is set to None by default.

And the condition will always be True when pipeline_name is None - returning all the files wihtout filtering. The functions `_iter_stored_schema_files` and `_list_dlt_table_files` aren't affected by it.

```
# Filters only if pipeline_name provided
if pipeline_name is None or fileparts[0] == pipeline_name:
    yield filepath, filepart
```

sh-rp (approved):

PR #1851, opened 2024-09-20
Fix/1849 Do Not Parse Ignored Empty Responses

### LINKED ISSUE #1849: Empty Response (e.g. HTTP 204) causes ignore response_actions to fail
### dlt version

1.0.0

### Describe the problem

Using the rest_api_source when specifying a response action such as `{"status_code": 204, "action": "ignore"}` if the response is empty this application fails as there is a log entry created which calls `response.json()`, this call fails throwing an exception which is unhandled. In my case this was when calling the Azure DevOps API. 

### Expected behavior

The empty response matching the status code and/or content should be ignored

### Steps to reproduce

Configure a `rest_api_source` to call an endpoint which returns no content and configure a `response_action` which should ignore it. Run the pipeline and it will fail with the error: `simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`

### Operating system

Linux, Windows

### Runtime environment

Local

### Python version

3.10

### dlt data source

rest_api_source

### dlt destination

DuckDB, Snowflake

### Other deployment details

_No response_

### Additional information

My current workaround is to override empty responses when they might occur by using the following function in the list of `response_actions` ahead of the desired ignore action:

```
def fixup_empty_response(response: Response) -> Response:
        if not response.text:
            response._content = "{}".encode("utf-8")
        return response
```
Then in my configuration I add:
```
{
...
"response_actions":[ fixup_empty_response, {"status_code": 204, "action": "ignore"}]
...
}
```

TheOneTrueAnt:
### Description
Simple fix to remove the call to `response.json()` when logging the execution of an `"ignore"` response action, instead calls `response.text`. Included a test based off of the existing one for ignoring 404 errors. 

### Related Issues

- Fixes #1849 

### Additional Context
Message lines were moved into a single line to meet with linter requirements.

### DIFF
--- a/dlt/sources/rest_api/config_setup.py
+++ b/dlt/sources/rest_api/config_setup.py
@@ -506,7 +506,7 @@ def _create_response_action_hook(
         elif action_type == "ignore":
             logger.info(
                 f"Ignoring response with code {response.status_code} "
-                f"and content '{response.json()}'."
+                f"and content '{response.text}'."
             )
             raise IgnoreResponseException
 
--- a/dlt/sources/rest_api/config_setup.py
+++ b/dlt/sources/rest_api/config_setup.py
@@ -505,8 +505,7 @@ def _create_response_action_hook(
                 hook(response)
         elif action_type == "ignore":
             logger.info(
-                f"Ignoring response with code {response.status_code} "
-                f"and content '{response.text}'."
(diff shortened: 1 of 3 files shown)


TheOneTrueAnt:
I realised that I'd neglected to run the linter _after_ writing the tests! Have tidied them up now!

PR #1854, opened 2024-09-21
Allows any duckdb version, fixes databricks az credentials

### LINKED ISSUE #1828: Support for DuckDB 1.1.0
### Feature description

Support for installing the latest versions of dlt and DuckDB.

Current error:
```
0.178 uv pip install --system -r requirements.txt
2.551   × No solution found when resolving dependencies:
2.551   ╰─▶ Because dlt[duckdb]==1.0.0 depends on duckdb>=0.6.1,<0.11 and
2.551       you require dlt[duckdb]==1.0.0, we can conclude that you require
2.551       duckdb>=0.6.1,<0.11.
2.551       And because you require duckdb==1.1.0, we can conclude that your
2.551       requirements are unsatisfiable.
```

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

I want to install the latest versions of dlt, DuckDB, dbt-core, and dbt-duckdb.

### Proposed solution

_No response_

### Related issues

I found this discussion:
https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1718957863295209

### LINKED ISSUE #1839: Feature: report missing "parent" in schema yaml
### Feature description

When I run the attached code, I get an error:

> Job for a_test__sub_record.1d2b10ead9.insert_values failed terminally in load 1726687518.354422 with message null value in column "_dlt_parent_id" of relation "a_test__sub_record" violates not-null constraint
DETAIL:  Failing row contains (a, PztIkGSe+oI/8Q, null, null, 2KkBaSEHNC3a/A).
. The package is aborted and cannot be retried.

The reason is that the "parent:" field in the schema yaml is missing.

Part of dlt knows that "sub_record" is a nested table (because it generates a table name `a_test__sub_record` with two underscores), but another part (I think it's `schema.utils.is_nested_table`) doesn't know it's a nested table, because the "parent" clause is missing in the schema.

Unpack with "tar xvfz test_files.tgz": [test_files.tgz](https://github.com/user-attachments/files/17050099/test_files.tgz)

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

My use case: I copy schema from the export schema to the import schema to control types, but I forgot to copy one line.

It took me a couple of hours to track this down, by stepping through the loading code in a debugger. It could be easier.

This is a "quality of life" issue. I've found dlt is working well, but if I do anything wrong, it can be hard to figure out what's happening.

### Proposed solution

The behavior I'd like is: dlt stops earlier (say, during schema parsing, or extract, instead of load), and says "a_test__sub_record is a nested table, but missing a parent declaration in the schema".

### Related issues

_No response_

### LINKED ISSUE #1853: Can't use AzureServicePrincipalCredentials with Databricks destination and Azure staging
### dlt version

1.0.0

### Describe the problem

Pipelines fail when using AzureServicePrincipalCredentials with Databricks destination and Azure staging.

[dlt/destinations/impl/databricks/databricks.py#L106](https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/databricks/databricks.py#L106) has:

```py
assert isinstance(staging_credentials, AzureCredentialsWithoutDefaults)
```

This means we can't use `AzureServicePrincipalCredentials`, which is not an instance of `AzureCredentialsWithoutDefaults`.

I believe this is a bug, because when I comment out the assertion, my pipeline succeeds.

### Expected behavior

My pipeline should succeed when using AzureServicePrincipalCredentials with Databricks destination and Azure staging.

### Steps to reproduce

1. Set up a pipeline using any source with Databricks destination and Azure staging using AzureServicePrincipalCredentials.
2. Run the pipeline
3. See it fails due to the assertion I linked above

I'm running this in a Databricks notebook.

### Operating system

Linux

### Runtime environment

Other

### Python version

3.10

### dlt data source

sql_database (Postgres)

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
fixes #1853 
fixes #1828 
fixes #1839


dlt/normalize/validate.py
```diff
@@ -41,3 +45,13 @@ def verify_normalized_table(
             f"`{table['table_format']}` for table `{table['name']}`. "
             "The setting will probably be ignored."
         )
+
+    parent_key = get_first_column_name_with_prop(table, "parent_key")
+    if parent_key and not is_nested_table(table):
```
sh-rp, line 50: we could consider cleaning up these kind of problems automatically, no?

sh-rp (approved):

dlt/normalize/validate.py
rudolfix: IMO in 90% this is user trying their own table structure...

PR #1855, opened 2024-09-22
allows to define hints for nested tables

### LINKED ISSUE #1647: Simplify schema modification of child tables
### Feature description

You should be able to use `apply_hints` on child tables just like you can on the parent tables. Right now there's no way to modify the schema of child tables without explicitly doing so in a schema.yml

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

> When child tables are generated all of the field types are text . This makes them essentially unusable for analysis/aggregation without further transformations. I also can't partition them.

Slack conversation: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1721944874168369

### Proposed solution

_No response_

### Related issues

_No response_

rudolfix:
### Description
fixes #1647

### DIFF
--- a/dlt/extract/extractors.py
+++ b/dlt/extract/extractors.py
@@ -130,7 +130,7 @@ class Extractor:
             # convert to table meta if created table variant so item is assigned to this table
             if meta.create_table_variant:
                 # name in hints meta must be a string, otherwise merge_hints would fail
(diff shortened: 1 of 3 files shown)


rudolfix:
continued in #2165

Issue #1131, opened 2024-03-22
disable rows deduplication if Incremental is attached to a resource with `merge` write disposition


rudolfix:
### Feature description

Incremental will remove duplicating rows (only with the same cursor field value, read the docs) based on content hash or primary key. This is not expected behavior when merge key is present - destination should be allowed to merge changes coming from the new records.
We want to disable "deduplication" in Incremental and expose destination to all the rows from data source.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Please see https://github.com/dlt-hub/dlt/issues/971#issuecomment-1983417044

### Proposed solution

1. * [x] disable deduplication by setting `primary_key` of Incremental to `()` (disable dedup) when resource has "merge" write disposition. we already propagate resource key to Incremental key so change should be easy
2. * [ ] keep explicitly set Incremental primary key (should already work like that)
3. * [ ] observe `apply_hints` changes in the resource (`_set_hints`) and if primary 
4. * [ ] make sure that using `with_hints` marker that is executed during resource execution is able to apply write disposition and disable the deduplication
5. * [x] also test for transformers that `bind()` the incremental late 
6. * [x] warn if dedup state grows too big ie. > 200 entries

willi-mueller:
Implementation strategy: 
1. read the write_disposition from the table_schema_template in DltResource._set_hints()
2. read the write_disposition from the hints_template in DltResourceHints._set_hints()
3. set the primary key of the incremental object to `()` because if it's a `Sequence` and emty it causes no deduplication.

PR #1877, opened 2024-09-26
logs warning if deduplication state is large

willi-mueller:
### Description

As a first step towards resolving this issue: warns if the deduplication state grows over a large count of hashes of the primary key.

### Related Issues

[#1131 ](https://github.com/dlt-hub/dlt/issues/1131)

### DIFF
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -532,10 +532,11 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
             # add directly computed hashes
             unique_hashes.update(transformer.unique_hashes)
             self._cached_state["unique_hashes"] = list(unique_hashes)
+            dedup_count = len(self._cached_state["unique_hashes"])
             DEDUP_WARNING_THRESHOLD = 200
-            if len(self._cached_state["unique_hashes"]) > 200:
+            if dedup_count > DEDUP_WARNING_THRESHOLD:
                 logger.warning(
-                    f"There are over {DEDUP_WARNING_THRESHOLD} records to be deduplicated because"
+                    f"There are {dedup_count} records to be deduplicated because"
                     f" they share the same primary key `{self.primary_key}`."
                 )
         return rows
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -532,7 +532,12 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
             # add directly computed hashes
             unique_hashes.update(transformer.unique_hashes)
             self._cached_state["unique_hashes"] = list(unique_hashes)
-
+            DEDUP_WARNING_THRESHOLD = 200
+            if len(self._cached_state["unique_hashes"]) > 200:
+                logger.warning(
+                    f"There are over {DEDUP_WARNING_THRESHOLD} records to be deduplicated because"
+                    f" they share the same primary key `{self.primary_key}`."
+                )
         return rows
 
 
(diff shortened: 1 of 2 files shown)


dlt/extract/incremental/__init__.py
```diff
@@ -502,7 +502,13 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
             # add directly computed hashes
             unique_hashes.update(transformer.unique_hashes)
             self._cached_state["unique_hashes"] = list(unique_hashes)
-
+            dedup_count = len(self._cached_state["unique_hashes"])
+            DEDUP_WARNING_THRESHOLD = 200
+            if dedup_count > DEDUP_WARNING_THRESHOLD:
```
rudolfix, line 507: this will flood user with messages. please compare the number of hashes before `unique_hashes.update(transformer.unique_hashes)` and after to see if threshold is crossed. so you display that only once

dlt/extract/incremental/__init__.py
```diff
@@ -502,7 +502,13 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
...
+                logger.warning(
+                    f"There are {dedup_count} records to be deduplicated because"
+                    f" they share the same primary key `{self.primary_key}`."
```
rudolfix, line 510: we need way better explanation here.
1. primary key is not known sometimes, then we hash the content. 
2. the root cause is not primary key. it is low resolution of cursor column. ie it is on a day. use cursor column name here

tests/extract/test_incremental.py
```diff
@@ -2586,3 +2586,24 @@ def updated_is_int(updated_at=dlt.sources.incremental("updated_at", initial_valu
...
+    ):
+        yield data_to_item_format(
+            item_type,
+            [{"id": i, "created_at": 1} for i in range(201)],
+        )
+
+    logger_spy = mocker.spy(dlt.common.logger, "warning")
+    p = dlt.pipeline(pipeline_name=uniq_id())
+    p.extract(some_data(1))
+    logger_spy.assert_any_call(
```
rudolfix, line 2606: make sure it is displayed only once here

rudolfix (changes requested):

dlt/extract/incremental/__init__.py
burnash: Resolved in dd66d42

dlt/extract/incremental/__init__.py
burnash: Resolved in dd66d42

tests/extract/test_incremental.py
burnash: Resolved in dd66d42

dlt/extract/incremental/__init__.py
```diff
@@ -118,6 +118,8 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
     EMPTY: ClassVar["Incremental[Any]"] = None
     placement_affinity: ClassVar[float] = 1  # stick to end
 
+    DEFAULT_DUPLICATE_CURSOR_WARNING_THRESHOLD: ClassVar[int] = 200
```
rudolfix, line 121: `Incremental` is also a dataclass. you should declare `uplicate_cursor_warning_threshold` under `lag` and this is a good place to attach default value so `DEFAULT_DUPLICATE_CURSOR_WARNING_THRESHOLD` is not longer needed.

tbh. my preference would be to not create a new field at all and just keep `DEFAULT_DUPLICATE_CURSOR_WARNING_THRESHOLD` if someone want to change that, one can do it for all instances by changing the classvar. which is IMO pretty cool. WDYT?

rudolfix (changes requested):
I'd like to simplify the code a little

dlt/extract/incremental/__init__.py
burnash: Good point, changed in ca3471c

rudolfix (approved):

PR #1888, opened 2024-09-27
move default pipelines of cores sources into source folders

sh-rp:


rudolfix (changes requested):
here we keep example pipelines with code of the sources. so still with  "production" code. maybe `dlt/sources/_demo_pipelines` would be a better place? also it does not look natural if pipeline is in the same folder as data source code...

rudolfix (approved):
LGTM!

Issue #971, opened 2024-02-15
DuckDB - append adds extra rows even when there's no updated data


karakanb:
### dlt version

0.4.4

### Describe the problem

The `append` disposition is supposed to not append if there's no new data and it works that way in postgres destination, but with duckdb it reinserts the data.

### Expected behavior

in the second iteration, the output should not have an extra row since there's no new data, however it appends anyway.

### Steps to reproduce

use the following code:

```python
import os
import dlt
from dlt.sources.credentials import ConnectionStringCredentials
from dlt.common import pendulum
import shutil

import duckdb
from ingestr.src.sql_database import sql_table


def load_standalone_table_resource() -> None:
    pipeline = dlt.pipeline(
        pipeline_name="output",
        destination=dlt.destinations.duckdb(
            credentials=uri,
        ),
        pipelines_dir="pipeline_dataappendprob",
        dataset_name="testschema_append",
    )

    # Run the resources together
    info = pipeline.run(
        sql_table(
            credentials=uri,
            schema="testschema_append",
            table="input",
            incremental=dlt.sources.incremental(
                "updated_at",
            ),
        ),
        dataset_name="testschema_append",
        table_name="output",
        write_disposition="append",
    )
    print(info)


def get_abs_path(relative_path):
    return os.path.abspath(os.path.join(os.path.dirname(__file__), relative_path))

abs_db_path = get_abs_path("sample_append.db")
uri = f"duckdb:///sample_append.db"
conn = duckdb.connect(abs_db_path)
def setup_db():
    try:
        shutil.rmtree(get_abs_path("pipeline_dataappendprob"))
    except:
        pass

    conn.execute("DROP SCHEMA IF EXISTS testschema_append CASCADE")
    conn.execute("CHECKPOINT")

    conn.execute("CREATE SCHEMA testschema_append")
    conn.execute("CREATE TABLE testschema_append.input (id INTEGER, val VARCHAR, updated_at DATE)")
    conn.execute("INSERT INTO testschema_append.input VALUES (1, 'val1', '2022-01-01'), (2, 'val2', '2022-01-02')")
    conn.execute("CHECKPOINT")


def run(iteration: int):
    print("running iteration", iteration)
    load_standalone_table_resource()

    print("--------------")
    print("source:")
    conn.sql("select * from testschema_append.input").show()
    print()
    print("destination:")
    conn.sql("select * from testschema_append.output").show()
    print("--------------")

setup_db()
run(1)
run(2)
```

[...]

PR #1892, opened 2024-09-28
Feat/1331 disables deduplication for incremental

### LINKED ISSUE #1131: disable rows deduplication if Incremental is attached to a resource with `merge` write disposition

willi-mueller:
### Description
This PR disables deduplication for the test case described by ingestr here: https://github.com/dlt-hub/dlt/issues/971#issuecomment-1983417044

### Related Issues

- Resolves #1131 

### Questions
@rudolfix 
1. I could not understand your points 2, 3, and 4 in the issue #1131 . Are they already implemented in this PR?
2. I am not sure if the last test `tests/extract/test_incremental.py::test_deduplication_on_write_disposition_not_merge` makes sense at all because on `write_disposition="replace"` the table is truncated before the load. I included it because the ticket speaks of the merge write disposition so I wanted to test the opposite too. Feel free to drop this commit.
3. Are the assertions on the incremental's `last_value` superfluous?
4. I could not find a way to test that the incremental returns all values without deduplication. Thus, I implemented the tests by making assertions on the loaded data. Is this strategy fine?

### TODO after merge

- [ ] notify the developers of ingestr


rudolfix:
we implemented smarter way to control deduplication: via open and closed ranges for incremental values. with open lower range the deduplication is automatically disabled (no overlap possible). so we do not need risky change of behavior for merge. thx!

willi-mueller:
@rudolfix amazing. I'm glad you found a different way that needs less changes! The code that's never shipped is the most bug-free ans fastest too! 😄

PR #1895, opened 2024-09-28
show full HTTP response on failure (Feat/1605)

### LINKED ISSUE #1605: show full response when printing failed api calls
### Feature description

If an API call fails, the cli calls the requests library's [raise_for_status](https://github.com/psf/requests/blob/main/src/requests/models.py#L997) method which prints the request's `status code` and `reason`. But not all APIs use the response's reason attribute. It would be helpful if we could instead either configure what response attributes get printed or at least print `response.text`. E.g. I was debugging a klaviyo connector and got the usual response message when my api call failed. The cause was that I hadn't configured the request header correctly but the only way I figured that out was by editing the raise_for_status method's code to look like this:
`if 400 <= self.status_code < 500:
            http_error_msg = (
                f"{self.status_code} Client Error: {reason} for url: {self.url}. Full response: {self.text}"
            )`

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

debugging failed klaviyo api calls

### Proposed solution

override the raise_for_status method? or maybe add a config that lets you specify which response attributes should be printed on failure

### Related issues

_No response_

willi-mueller:
### Description
This PR implements error logging of the full HTTP response on error when the `dlt.sources.helpers.requests.Session` is used.

Unless a custom session with different configuration is used, all HTTP responses with an error code 4xx will log an error. 

This includes:
1. the `RESTClient`
2. each call made by the `dlt.sources.rest_api`

### Related Issues

- Resolves #1605


burnash:
I propose an alternative solution to logging the full HTTP response on error: we can use a context manager `catch_http_error` that captures HTTP errors and allows for flexible and controlled error handling / debugging.

Example:
```py
import dlt
from dlt.sources.helpers.rest_client import paginate, catch_http_error


@dlt.resource(
    table_name="issues",
    write_disposition="replace",
    primary_key="id",
)
def get_issues():
    for page in paginate(
        "https://api.github.com/repos/dlt-hub/dlt/issues",
        headers={"Authorization": "Bearer invalid_token_here"},
    ):
        yield page


pipeline = dlt.pipeline(
    pipeline_name="github_issues_merge",
    destination="duckdb",
)

with catch_http_error() as http_error:
    load_info = pipeline.run(get_issues)

if http_error:
    print(f"HTTP {http_error.response.status_code}: {http_error.response.content}")
```

Pros:
- improved flexibility: the use can interact with captured HTTPError
- cleaner and safer logs
- selective usage: no need to alter the session

The rationale is that in my opinion accessing the HTTPError should be done on the higher level of abstraction than Requests's session as when the user is using the request/response directly they already have easy access to HTTP errors. 
So the idea of this change is to provide some way to extract HTTPError from the dlt exceptions.

adrianbr:
+1 this is needed for vibe coding

rudolfix:
I think what is in the PR is pretty good. We should just generalize the catcher to be able to find any exception in the tree and move it to utils. but my take here is that we do not solve typical user problems which is a better readability of exceptions coming from Pipeline. my take:
1. Pipeline class always raises `PipelineStepFailed` exception with a trace of a failed step. We can pretty print this.
2. We could use `get_exception_trace_chain` or `get_exception_traces` to get traces of the inner exceptions and print them. Those traces contain additional info like load ids, job ids, pipeline names etc. that we could pretty print
3. with log level on DEBUG or INFO we could print verbose information ie. for HTTPExceptions?
WDYT? @adrianbr @burnash if you agree I can push someone to implement it

burnash:
Hey @willi-mueller, thanks so much for this PR! I'm going to close it for now: initially I suggested and took a different approach here, but after consideration, I've decided to implement something similar to your suggestion in a different PR (#2867). Sorry about the confusion, and thank you again for your contribution!

PR #1904, opened 2024-10-01
[cli] add explicit paginators to rest_api template

AstrakhantsevaAA:
users have some issues with dlt init rest_api duckdb command, I set paginators explicitly, because they are not detected automatically for some reason. also reduced the amount of data for github api (request data for a week, not for month)  because of rate limits

### DIFF
--- a/dlt/sources/rest_api_pipeline.py
+++ b/dlt/sources/rest_api_pipeline.py
@@ -109,7 +109,7 @@ def load_github() -> None:
 
     load_info = pipeline.run(github_source())
     print(load_info)  # noqa: T201
-    print(pipeline.last_trace)
+    print(pipeline.last_trace)  # noqa: T201
 
 
 def load_pokemon() -> None:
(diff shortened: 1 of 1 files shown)


burnash:
If we want to explicitly define pagination, we'd need to update the tutoral as it says that we didn't explicitly specify pagination: https://dlthub.com/docs/tutorial/rest-api#pagination

AstrakhantsevaAA:
@burnash 
>Can you reproduce it? If yes, let's create an issue for this and fix it.

Yes, I can, unfortunately, for both sources :( Let's create an issue then. 

Screenshot 2024-10-01 at 15 48 57

burnash:
> Yes, I can, unfortunately, for both sources :( Let's create an issue then.

@AstrakhantsevaAA I think it's not an actual error but a not very good warning. Most likely we need to change the message. This `Fallback pagiantor` warning  comes from the child (dependent) resource and informs that there's no paginator on this page.

AstrakhantsevaAA:
@burnash yes, I see it is just a warning, but also it is the cause of the rate limit error. Not sure why it is happening, but if i add pagination explicitly then the data loads without any problem and rate limit errors. It seems like it keeps requesting one page until the rate limit error, so you can see the list of these warnings (with pagination it takes 1-2 requests).

created the issue: https://github.com/dlt-hub/dlt/issues/1915

PR #1914, opened 2024-10-02
fix: UUIDs are not an unknown data type (logging)

### LINKED ISSUE #1862: When ingesting a sql_database with a UUID column, getting a flood of warnings
### dlt version

1.0.0

### Describe the problem

Exactly what the title says.

### Expected behavior

No annoying warnings log messages.

### Steps to reproduce

- Create a table in any sql database with a UUID column type
- run any pipeline with any destination

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

built-in sql_database

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

The bug can be easily corrected by changing https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/sql_database/schema_types.py#L89

Before
```python
    if hasattr(sqltypes, "Uuid") and isinstance(sql_t, sqltypes.Uuid):
        # we represent UUID as text by default, see default_table_adapter
        col["data_type"] = "text"
    if isinstance(sql_t, sqltypes.Numeric):
        # check for Numeric type first and integer later, some numeric types (ie. Oracle)
        # derive from both
        # all Numeric types that are returned as floats will assume "double" type
        # and returned as decimals will assume "decimal" type
        if sql_t.asdecimal is False:
            col["data_type"] = "double"
```

After
```python
    if hasattr(sqltypes, "Uuid") and isinstance(sql_t, sqltypes.Uuid):
        # we represent UUID as text by default, see default_table_adapter
        col["data_type"] = "text"
    elif isinstance(sql_t, sqltypes.Numeric):
        # check for Numeric type first and integer later, some numeric types (ie. Oracle)
        # derive from both
        # all Numeric types that are returned as floats will assume "double" type
        # and returned as decimals will assume "decimal" type
        if sql_t.asdecimal is False:
            col["data_type"] = "double"
```

neuromantik33:
### Description
- Fixes #1862

### DIFF
--- a/dlt/sources/sql_database/schema_types.py
+++ b/dlt/sources/sql_database/schema_types.py
@@ -86,7 +86,7 @@ def sqla_col_to_column_schema(
     if hasattr(sqltypes, "Uuid") and isinstance(sql_t, sqltypes.Uuid):
         # we represent UUID as text by default, see default_table_adapter
         col["data_type"] = "text"
-    if isinstance(sql_t, sqltypes.Numeric):
+    elif isinstance(sql_t, sqltypes.Numeric):
         # check for Numeric type first and integer later, some numeric types (ie. Oracle)
         # derive from both
         # all Numeric types that are returned as floats will assume "double" type


burnash (approved):
Looks good, thanks @neuromantik33 for the fix!

PR #1919, opened 2024-10-03
Add engine_adapter_callback

umar-ahmed:
### Description

- Adds `engine_adapter_callback` that is executed after the SQLAlchemy engine is constructed
- This allows developers to apply any arbitrary statements before subsequent queries are issued
- In particular, I want to be able to run `SET workload = "olap";` to connect to a Planetscale DB without bumping into[ 100k rows returned system limits](https://planetscale.com/docs/reference/planetscale-system-limits#query-limits).

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/dlt/sources/sql_database/__init__.py
+++ b/dlt/sources/sql_database/__init__.py
@@ -19,6 +19,7 @@ from .helpers import (
     SqlTableResourceConfiguration,
     _detect_precision_hints_deprecated,
     TQueryAdapter,
+    TEngineAdapter,
 )
 from .schema_types import (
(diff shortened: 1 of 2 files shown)


rudolfix:
@umar-ahmed `sql_database` and `sql_table` accept Engine instance in credentials. So you can create your own, run anything on it and then pass it to dlt. OFC having this callback is more convenient because you do not need to handle credentials yourself. if you plan to continue this PR we will also need a simple test where you verify that adapter receives engine instance

PR #1923, opened 2024-10-03
Feat/1922 rest api source add mulitple path parameters

### LINKED ISSUE #1922: rest_api_source - support multiple resolve params from single parent
### Feature description

It would be great if the rest_api_client could support multiple `resolve` path parameters from a single parent. for example the following should be acceptable with both the `id` and `group` fields of a parent record being populated into the url in the expected places.

```
{
      "name": "user_details",
      "endpoint": {
          "path": "user/{user_id}/{group_id}",
          "params": {
              "user_id": {
                  "type": "resolve",
                  "field": "id",
                  "resource": "users",
              },
              "group_id": {
                  "type": "resolve",
                  "field": "group",
                  "resource": "users",
              },
          },
      },
  },
```

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

When working with the Azure DevOps API I wanted to be able to retrieve pipeline run timelines which ended up needing two parameters to be populated in url. As a workaround we used a map function to manually construct the url in the parent resource but this polutes the data model a bit.

### Proposed solution

Accept multiple path parameters from a single parent only. I'll submit a PR along with this issue containing a proposed solution.

### Related issues

_No response_

TheOneTrueAnt:
### Description
Adds support for  multiple path parameters to child resources in a rest_api_source.

### Related Issues
- Closes #1922

### Additional Context
Added some tests which seemed reasonable and ensured that it doesn't allow multiple parents.

### DIFF
--- a/dlt/sources/rest_api/__init__.py
+++ b/dlt/sources/rest_api/__init__.py
@@ -212,7 +212,7 @@ def create_resources(
     client_config: ClientConfig,
     dependency_graph: graphlib.TopologicalSorter,
     endpoint_resource_map: Dict[str, Union[EndpointResource, DltResource]],
-    resolved_param_map: Dict[str, Optional[ResolvedParam]],
+    resolved_param_map: Dict[str, Optional[List[ResolvedParam]]],
 ) -> Dict[str, DltResource]:
     resources = {}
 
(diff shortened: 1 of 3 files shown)


burnash (approved):
Very well structured PR, thank you very much @TheOneTrueAnt for the contribution.

TheOneTrueAnt:
Great news, no problem at all and thanks for merging it. Looking forward to the next release 😊

burnash:
@TheOneTrueAnt, we've just released dlt 1.2.0 featuring this change.

PR #1926, opened 2024-10-04
only truncate or delete from existing tables in refresh modes

sh-rp:
### Description
Only drop known and materialized tables from the destination, fixes a bunch of issues with the refresh mode.

### DIFF
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -649,12 +649,9 @@ WHERE """
         Delete all stored versions with the same name as given schema.
         Fails silently if versions table does not exist
         """
-        with suppress(DatabaseUndefinedRelation):
-            name = self.sql_client.make_qualified_table_name(self.schema.version_table_name)
-            (c_schema_name,) = self._norm_and_escape_columns("schema_name")
-            self.sql_client.execute_sql(
-                f"DELETE FROM {name} WHERE {c_schema_name} = %s;", schema.name
-            )
+        name = self.sql_client.make_qualified_table_name(self.schema.version_table_name)
+        (c_schema_name,) = self._norm_and_escape_columns("schema_name")
+        self.sql_client.execute_sql(f"DELETE FROM {name} WHERE {c_schema_name} = %s;", schema.name)
 
     def _update_schema_in_storage(self, schema: Schema) -> None:
         # get schema string or zip
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -19,6 +19,7 @@ from typing import (
 import zlib
 import re
 from contextlib import contextmanager
+from contextlib import suppress
 
 from dlt.common import pendulum, logger
 from dlt.common.json import json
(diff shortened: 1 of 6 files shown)


dlt/destinations/job_client_impl.py
```diff
@@ -197,8 +197,11 @@ def drop_tables(self, *tables: str, delete_schema: bool = True) -> None:
             delete_schema: If True, also delete all versions of the current schema from storage
         """
         with self.maybe_ddl_transaction():
-            self.sql_client.drop_tables(*tables)
-            if delete_schema:
+            # only run drop command on existing tables
+            existing_tables = self.get_known_existing_storage_tables()
+            tables_to_drop = set(existing_tables).intersection(set(tables))
```
sh-rp, line 202: this will only issue drop commands for tables that exist in the destination AND are known by dlt. This is a light change from before where any table could be dropped. Is this ok?

rudolfix (changes requested):
this is too brute force. reflecting schema is costly but this fix will only work on sql_job_client like destinations and can be done a little bit better.

please
1. first reproduce problem with a test (please check both drop resource and drop data)
2. then find a way to skip the drop on staging dataset if there's no staging dataset (client.is_storage_initialized is available, maybe this is enough)
3. all the DROP TABLE (and related) statements are supposed to skip silently when table does not exist

dlt/destinations/job_client_impl.py
```diff
@@ -645,11 +648,15 @@ def _row_to_schema_info(self, query: str, *args: Any) -> StorageSchemaInfo:
...
     def _delete_schema_in_storage(self, schema: Schema) -> None:
         """
-        Delete all stored versions with the same name as given schema
+        Delete all stored versions with the same name as given schema.
+        Fails silently if versions table does not exist
         """
-        name = self.sql_client.make_qualified_table_name(self.schema.version_table_name)
-        (c_schema_name,) = self._norm_and_escape_columns("schema_name")
-        self.sql_client.execute_sql(f"DELETE FROM {name} WHERE {c_schema_name} = %s;", schema.name)
+        with suppress(DatabaseUndefinedRelation):
```
sh-rp, line 654: this will not work when running inside a transaction (if supports_ddl_transaction is active), because the exception is raise when the transaction context exits. We could check wether the table exists before we try to delete from it. DELETE FROM IF EXISTS does not seem to be standardized in sql, so we'd probably need a couple of different implementations for each destination.

rudolfix: so we still need this "suppress"? checking for initialized dataset should be enough. OFC user can drop the table schema manually, then it will fail. 
my take: remove "suppress" and run tests again. you can restore the test client to the old form, in most cases we have transactions and as you say - this does not work anyway

rudolfix (changes requested):

rudolfix:
@sh-rp looks like some databases will raise when schema does not exist and we drop table in it with IF NOT EXISTS. my take is to fix the test here. that behavior should be presented by checking if storage is initialized

PR #1927, opened 2024-10-04
Support Spatial Types for PostGIS

### LINKED ISSUE #696: Support spatial / geo types
### Feature description

dlt should support geo types like shapes, geometries & geographies and different CRS as best as possible to allow data loading/transfer of such data in spatial databases like Postgis (Postgres addon), SpatiaLite (SQLite addon), BigQuery (only 2D geographies), H2, Oracle Spatial...

### Are you a dlt user?

### Use case

Natively, Postgres supports [basic geometries](https://www.postgresql.org/docs/current/datatype-geometric.html) and Postgis adds support for georeferenced geometries/geographies with different CRS (coordinate reference systems) & even 3D. BigQuery [supports 2D geographies](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions) and only in CRS WGS64/EPSG:4326. As different database systems (and coordinate reference systems) fit different usages, data is often transfered and transformed between systems for certain workloads. Loading spatial data is quite cumbersome because bad support from common libraries, different coordinate systems and expensive operations on geometries.

### Proposed solution

Support spatial data. In some directions data has to be converted to certain CRS or to geojson/WKT because the target system might not support it otherwise. These conversions can be lossy for example by small changes due to transformation and rounding of coordinates. Also different DBMS enforce validation ((counter-)clockwise polygons, self-intersections, touching points) differently, so sometimes not all geometries might be transferable.

### Related issues

_No response_

Pipboyguy:
### Description
This PR adds support for PostGIS GEOMETRY type in PostgreSQL. It introduces a logical type for GEOMETRY on binary or text fields.

Users can declare geometry columns via the `postgres_adapter`, enabling direct insertion from WKT, WKB and WKB hex representations, in addition to GeoPandas integration with automatic geometry column detection and conversion.

### Related Issues

- Closes #696

### Limitations

- `LinearRing` only works with wkb type (python binary type)
- EWKB (Extended WKB) is not yet supported
- geopandas can be supported as well, but will need more work

Typical usage pattern:

```py
postgres_adapter(geodata_resource, geometry=["geom"], srid=3857)
```


dlt/destinations/impl/postgres/postgres.py
```diff
@@ -130,21 +144,227 @@ def run(self) -> None:
...
-                "COPY %s (%s) FROM STDIN WITH (FORMAT CSV, DELIMITER '%s', NULL '',"
-                " %s ENCODING '%s')"
-                % (
-                    qualified_table_name,
-                    headers,
-                    sep,
-                    null_headers,
-                    csv_format.encoding,
-                )
+                f"COPY {qualified_table_name} ({headers}) FROM STDIN WITH (FORMAT CSV, DELIMITER"
```
burnash, line 147: 👍

tests/load/cases/loading/sample_geodata.xml
burnash, line 1: Is this file being used somewhere? (maybe implicitly?) I can't find the reference to it.

burnash (commented):
Thanks for the PR, @Pipboyguy 
I have a couple of questions, please see my comments.

tests/load/cases/loading/sample_geodata.xml
Pipboyguy: It isn't anymore, I wanted to use it for geopandas tests but ended up forgetting abuot it. Thanks for catching this! Removed

dlt/destinations/impl/postgres/postgres.py
```diff
@@ -130,21 +144,227 @@ def run(self) -> None:
...
+
+        if not value or value == "NULL":
+            return "NULL"
+        if value.startswith("E'") and value.endswith("'"):
+            return value[2:-1]
+
+        return value[1:-1] if value.startswith("'") and value.endswith("'") else value
+
+    @staticmethod
+    def _split_fragment_to_records(text: str) -> List[str]:
```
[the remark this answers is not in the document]
Pipboyguy: @burnash I looked into it and I don't believe there would be much benefit from bringing in sqlparse as a dependency. Looking at the docs, there's nothing to handle the complex nested parentheses needed for geometry type parsing like `ST_GeomFromText('POINT(1 1)')`. 

The only parentheses handling shown is the basic [`Parenthesis` ](https://sqlparse.readthedocs.io/en/latest/analyzing.html#sqlparse.sql.Parenthesis)token class, which can't distinguish between the outer VALUES parentheses and nested geometry function calls. The nested brackets would make the control flow just as complex as our current parser, if not more so when handling PostGIS-specific syntax. 

Thoughts?

dlt/common/data_writers/escape.py
```diff
@@ -56,7 +67,12 @@ def escape_postgres_literal(v: Any) -> Any:
     if isinstance(v, (list, dict)):
         return _escape_extended(json.dumps(v))
     if isinstance(v, bytes):
-        return f"'\\x{v.hex()}'"
+        if is_valid_wkb(v):
```
rudolfix, line 70: this is very impractical :/ we have a dependence on shapely and there may be false positives (there are no magic numbers in wkb). passing expected type to escape...literal will slow this down

we could use a special binary type but user would need to wrap binary anyway. so my take is to drop `wkb` and just keep `wkb_hex`.

rudolfix (changes requested):
this is really good!
1. my suggestion is that we drop wkb and just keep wkb_text. also move shapely to dev dependencies
2. IMO csv with WKT and WKB_HEX should work! did you try it? it'd be hard to understand why not....

tests/load/postgres/utils.py
```diff
@@ -0,0 +1,68 @@
...
+    Point,
+    LineString,
+    Polygon,
+    MultiPoint,
+    MultiLineString,
+    MultiPolygon,
+    GeometryCollection,
+    LinearRing,
+)
+from shapely.wkb import dumps as wkb_dumps  # type: ignore
```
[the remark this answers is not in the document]
Pipboyguy: Yes, it is used in `load.postgres.test_postgres_table_builder.test_geometry_types` to generate valid geometries. This should probably be a changed to a fixture if we use it again

dlt/destinations/impl/postgres/postgres.py
```diff
@@ -167,21 +158,30 @@ def __init__(
     def create_load_job(
         self, table: PreparedTableSchema, file_path: str, load_id: str, restore: bool = False
     ) -> LoadJob:
+        if any(
+            column.get(GEOMETRY_HINT) for column in table["columns"].values()
+        ) and not file_path.endswith("insert_values"):
+            # Only insert_values load jobs supported for geom types.
+            # TODO: This isn't actually true, can make it work with geoarrow!
+            raise TerminalValueError(
```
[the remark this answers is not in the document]
Pipboyguy: I added csv loader cases to the geo test, and it all passes. Your intuition was entirely correct!

rudolfix (approved):
LGTM!

PR #1931, opened 2024-10-07
Fix/1897 support https endpoints clickhouse

sh-rp:
### Description
See linked ticket


sh-rp:
@rudolfix the reporter of this ticket did not get back to me. I added some simple tests and a docs entry. I don't think we should deduplicate forward slashes as requested, afaik s3://bucket/folder/subfolder and s3://bucket/folder//subfolder are separate locationns

rudolfix (changes requested):
LGTM! but the test didn't run on CI. maybe it is not included? if this is like you intended then merge this

PR #1938, opened 2024-10-09
WIP: Pluggable Cli Commands

sh-rp:
### Description
This PR adds a plugin interface for CLI commands and converts all the existing commands to plugins.

Notes:

TODOs before merge:
* Check that the deploy commands behaves correctly when dependencies are missing, this is the only thing that significantly changed.
* Add a cli command to our test plugin


dlt/cli/_dlt.py
```diff
@@ -304,276 +124,23 @@ def main() -> int:
...
-        "load-package", help="Displays information on load package, use -v or -vv for more info"
-    )
-    pipe_cmd_package.add_argument(
-        "load_id",
-        metavar="load-id",
-        nargs="?",
-        help="Load id of completed or normalized package. Defaults to the most recent package.",
-    )
+    m = plugins.manager()
+    commands = cast(List[Type[SupportsCliCommand]], m.hook.plug_cli())
```
sh-rp, line 131: plugins get discovered here

dlt/cli/_dlt.py
```diff
@@ -304,276 +124,23 @@ def main() -> int:
...
 
-    subparsers.add_parser("telemetry", help="Shows telemetry status")
+    # install available commands
+    installed_commands: Dict[str, SupportsCliCommand] = {}
+    for c in commands:
```
sh-rp, line 135: each plugin gets instantiated and registered with the main parser, overwriting of existing commands is prevented. This approach is quite safe, but it will not allow to modify existing parsers or subparsers, this is to be discussed.

rudolfix: I'd like to be able to overwrite init command with my own. this prevents that? IMO we should allow rewriting

dlt/common/configuration/specs/pluggable_cli_command.py
```diff
@@ -0,0 +1,18 @@
+from typing import Protocol
+
+import argparse
+
+
+class SupportsCliCommand(Protocol):
```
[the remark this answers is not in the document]
rudolfix: hmmm this should go to `cli` module, ie. to where you keep the hook definition. this is a place for injectable contexts

rudolfix (changes requested):
please add a cli to dlt test plugin that adds commands and overwrites existing. tests are already set up.

should we have a module that imports all hooks? for the benefit of plugin makers that then can see all hooks are "external" protocols. we ourselves do not need to import it.

otherwise this is pretty good. let's do PR soon :)

sh-rp:
@rudolfix the requests are adressed. With regards to:
`should we have a module that imports all hooks? for the benefit of plugin makers that then can see all hooks are "external" protocols. we ourselves do not need to import it.`
Should we put it in dlt.plugins and decide on a naming scheme for these protocols? I think it would be nice to have one central place where you can import all the protocols from. let me know it would be an easy change.

dlt/cli/_dlt.py
rudolfix: I do not think it is implemented. IMO you should first deduplicate command list, keeping the last occurence. It is better not to add overwritten subparsers to the parser

dlt/cli/_dlt.py
```diff
@@ -304,276 +127,22 @@ def main() -> int:
...
+    # NOTE: plugin commands are added in reverse order so that the last added command (coming from external plugin)
+    # overwrites internal commands
+    commands.reverse()
 
-    subparsers.add_parser("telemetry", help="Shows telemetry status")
+    # install available commands
+    installed_commands: Dict[str, SupportsCliCommand] = {}
+    for c in commands:
+        command = c()
+        command_parser = subparsers.add_parser(command.command, help=command.help_string)
```
rudolfix, line 143: here... what if we overwrite `command`?

rudolfix (changes requested):
this is really good! pls. drop the WIP

only thing to consider: deduplicate commands before adding them to parser.

PR #1940, opened 2024-10-09
prefers uv over pip if found

rudolfix:
### Description
see commit list

### DIFF
--- a/dlt/common/runners/venv.py
+++ b/dlt/common/runners/venv.py
@@ -4,8 +4,9 @@ import shutil
 import venv
 import types
 import subprocess
-from typing import Any, List, Type
+from typing import Any, ClassVar, List, Type
 
+from dlt.common import known_env
 from dlt.common.exceptions import CannotInstallDependencies, VenvNotFound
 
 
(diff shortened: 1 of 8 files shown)


dlt/common/runners/venv.py
```diff
@@ -22,6 +23,8 @@ def post_setup(self, context: types.SimpleNamespace) -> None:
 class Venv:
     """Creates and wraps the Python Virtual Environment to allow for code execution"""
 
+    PIP_TOOL: ClassVar[str] = os.environ.get(known_env.DLT_PIP_TOOL, None)
```
burnash, line 26: Maybe DEPENDENCY_MANAGER? This communicates the purpose of the var more explicitly

burnash (changes requested):
I'd rename PIP_TOOL to something more clear and explicit, `DEPENDENCY_MANAGER`, or `PACKAGE_INSTALLER`

dlt/common/runners/venv.py
rudolfix: those are good points. but PIP means "Preferred Installer Program" so I think the name here... we can still change it before next release... maybe by expanding the acronym

PR #1941, opened 2024-10-09
Added extended jsonpath_ng parser

francescomucio:
### Description
This PR introduce the `jsonpath_ng` extended parser, which is more compliant with the JsonPath standard.

An example of the lack of features of the standard parser is here: [Filters not implmented](https://github.com/h2non/jsonpath-ng/issues/8)

### Additional Context
This was discussed in the slack community [here](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1728335837644989).

### DIFF
--- a/dlt/common/jsonpath.py
+++ b/dlt/common/jsonpath.py
@@ -3,8 +3,8 @@ from itertools import chain
 
 from dlt.common.typing import DictStrAny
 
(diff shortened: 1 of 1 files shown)


burnash (approved):
Hey @francescomucio thank you very much for the fix!

PR #1942, opened 2024-10-10
Update url in deploy-with-airflow-composer.md

FriedrichtenHagen:
### Description
The original urls is broken. I assume this is the correct page to link to?

---

just a small url


burnash (approved):
Thank you for the fix, @FriedrichtenHagen

PR #1945, opened 2024-10-10
dataset factory

sh-rp:
### Description
This PR is an example implementation of a dataset factory to build datasets without a pipeline object.

### DIFF
--- a/dlt/__init__.py
+++ b/dlt/__init__.py
@@ -42,6 +42,7 @@ from dlt.pipeline import (
 )
 from dlt.pipeline import progress
 from dlt import destinations
+from dlt.destinations.dataset import dataset
 
 pipeline = _pipeline
 current = _current
@@ -79,6 +80,7 @@ __all__ = [
     "TCredentials",
     "sources",
     "destinations",
+    "dataset",
 ]
 
 # verify that no injection context was created
--- a/dlt/__init__.py
+++ b/dlt/__init__.py
@@ -42,7 +42,7 @@ from dlt.pipeline import (
 )
 from dlt.pipeline import progress
 from dlt import destinations
-from dlt.destinations.dataset import dataset
+from dlt.destinations.dataset import dataset as _dataset
 
 pipeline = _pipeline
 current = _current
@@ -80,7 +80,7 @@ __all__ = [
     "TCredentials",
     "sources",
     "destinations",
-    "dataset",
+    "_dataset",
 ]
 
 # verify that no injection context was created
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -66,6 +66,8 @@ from dlt.common.exceptions import MissingDependencyException
 TDestinationConfig = TypeVar("TDestinationConfig", bound="DestinationClientConfiguration")
 TDestinationClient = TypeVar("TDestinationClient", bound="JobClientBase")
 TDestinationDwhClient = TypeVar("TDestinationDwhClient", bound="DestinationClientDwhConfiguration")
+TDatasetType = Literal["dbapi", "ibis"]
+
 
 DEFAULT_FILE_LAYOUT = "{table_name}/{load_id}.{file_id}.{ext}"
 
@@ -657,8 +659,8 @@ class JobClientBase(ABC):
 
 class WithStateSync(ABC):
     @abstractmethod
-    def get_stored_schema(self) -> Optional[StorageSchemaInfo]:
-        """Retrieves newest schema from destination storage"""
+    def get_stored_schema(self, any_schema_name: bool = False) -> Optional[StorageSchemaInfo]:
+        """Retrieves newest schema from destination storage, setting any_schema_name to true will return the newest schema regardless of the schema name"""
         pass
 
     @abstractmethod
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -659,8 +659,11 @@ class JobClientBase(ABC):
 
 class WithStateSync(ABC):
     @abstractmethod
-    def get_stored_schema(self, any_schema_name: bool = False) -> Optional[StorageSchemaInfo]:
-        """Retrieves newest schema from destination storage, setting any_schema_name to true will return the newest schema regardless of the schema name"""
+    def get_stored_schema(self, schema_name: str = None) -> Optional[StorageSchemaInfo]:
+        """
+        Retrieves newest schema with given name from destination storage
+        If no name is provided, the newest schema found is retrieved.
+        """
         pass
 
     @abstractmethod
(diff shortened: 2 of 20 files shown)


dlt/destinations/job_client_impl.py
```diff
@@ -397,14 +397,21 @@ def _from_db_type(
...
-    def get_stored_schema(self) -> StorageSchemaInfo:
+    def get_stored_schema(self, any_schema_name: bool = False) -> StorageSchemaInfo:
         name = self.sql_client.make_qualified_table_name(self.schema.version_table_name)
         c_schema_name, c_inserted_at = self._norm_and_escape_columns("schema_name", "inserted_at")
-        query = (
-            f"SELECT {self.version_table_schema_columns} FROM {name} WHERE {c_schema_name} = %s"
-            f" ORDER BY {c_inserted_at} DESC;"
-        )
-        return self._row_to_schema_info(query, self.schema.name)
+        if any_schema_name:
```
sh-rp, line 403: this can be compressed and also needs to be implemented for all destinations

tests/load/test_read_interfaces.py
```diff
@@ -212,6 +212,47 @@ def double_items():
     loads_table = pipeline._dataset()[pipeline.default_schema.loads_table_name]
     loads_table.fetchall()
 
+    # check dataset factory
```
sh-rp, line 215: we need proper tests to ensure that always the newest schema actually is selected, this is just basic test code to make sure it generally works

dlt/pipeline/pipeline.py
```diff
@@ -1717,10 +1718,11 @@ def __getstate__(self) -> Any:
         # pickle only the SupportsPipeline protocol fields
         return {"pipeline_name": self.pipeline_name}
 
-    def _dataset(self, dataset_type: Literal["dbapi", "ibis"] = "dbapi") -> SupportsReadableDataset:
+    def _dataset(self, dataset_type: TDatasetType = "dbapi") -> SupportsReadableDataset:
```
sh-rp, line 1721: it might be nice to also be able to a dataset for the staging destination if desired.

dlt/__init__.py
```diff
@@ -79,6 +80,7 @@
     "TCredentials",
     "sources",
     "destinations",
+    "dataset",
```
sh-rp, line 83: allows dlt.dataset(something). is this ok?

dlt/destinations/job_client_impl.py
sh-rp: also: is this signature ok, or do we want to add a new function for this? I'm also not sure about this "any_schema_name":.. :)

rudolfix (changes requested):
this is good! I think there's a big overlap with dataset and part of pipeline that does the same:
- keeping destinations (also staging - you should IMO include is as optional, sometimes we'll use ie. Athena to do Iceberg but actually open staging filesystem as a data lake)
- keeping a list of schemas on the dataset
- initializing configs, exposing various clients

do you think it makes sense to refactor pipeline right now?

do you think we could keep a dataset instance in the pipeline and just expose some methods from it

the standalone part looks good. I'm not sure if we should go for a single schema or for many schemas dataset?

dlt/__init__.py
rudolfix, line 83: let's export `_dataset` until we have stable interface.

dlt/common/destination/reference.py
```diff
@@ -657,8 +659,8 @@ def __exit__(
 
 class WithStateSync(ABC):
     @abstractmethod
-    def get_stored_schema(self) -> Optional[StorageSchemaInfo]:
-        """Retrieves newest schema from destination storage"""
+    def get_stored_schema(self, any_schema_name: bool = False) -> Optional[StorageSchemaInfo]:
```
rudolfix, line 662: I'd rather add a new method but it really do not fit here. this interface assumes that there's a known name of a schema.

my take would be to change signature to 
```
get_stored_schema(self, schema_name: str = None)
```
if None is specified, we load the newest schema, if name is provided we load the newest schema with given name

dlt/destinations/dataset.py
```diff
@@ -71,29 +76,100 @@ def _wrap(*args: Any, **kwargs: Any) -> Any:
...
 
     def __getattr__(self, table_name: str) -> SupportsReadableRelation:
         """access of table via property notation"""
         return self.table(table_name)
+
+
+def dataset(
+    destination: TDestinationReferenceArg,
+    dataset_name: str,
+    schema: Union[Schema, str, None] = None,
```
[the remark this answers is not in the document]
rudolfix: OK cool. but as discussed we'll need to implement dataset compatible with pipeline dataset (many schemas, different database layout: we support schema separation but it is rarely used)

tests/load/test_job_client.py
```diff
@@ -951,6 +952,73 @@ def _load_something(_client: SqlJobClientBase, expected_rows: int) -> None:
             )
 
 
+# NOTE: this could be folded into the above tests, but these only run on sql_client destinations for now
+# but we want to test filesystem and vector db here too
```
rudolfix, line 956: why can't you extend to other destinations with selectors below?

rudolfix: at least filesystem should be tested

tests/load/test_read_interfaces.py
```diff
@@ -212,6 +213,73 @@ def double_items():
...
+        dataset_name=pipeline.dataset_name,
+        schema=pipeline.default_schema_name,
+    )
+    assert dataset.schema.tables["items"]["write_disposition"] == "replace"  # type: ignore
+
+    # check that schema is not loaded when wrong name given
+    dataset = dlt.dataset(
+        destination=destination_for_dataset,
+        dataset_name=pipeline.dataset_name,
+        schema="wrong_schema_name",
```
rudolfix, line 243: we allow a new schema to be added right? that's why we do not raise when schema is not known? we'll need a better method of adding schemas to a dataset. also to sync schemas etc.

rudolfix (changes requested):
I'd change the WithState interface to be more explicit and also add schema tests for the filesystem

tests/load/test_job_client.py
sh-rp: The schema stuff for the filesystem currently is tested in this new function as well as the filesystem specific tests. I'd like to do a bigger PR soon where I harmonize all the tests that can be merged, but that is a bit of work and will also require some changes in the filesystem that are not at all related to this dataset stuff. Or is there some change here that in your view is not tested?

tests/load/test_read_interfaces.py
sh-rp: What actually happens here is that an empty schema is created as a standin, it does not do anything and also does not get saved anywhere. I can change that if you like, but afaik for now it should be ok.

dlt/__init__.py
sh-rp: done

tests/load/test_read_interfaces.py
rudolfix: no, that is OK. I just thing that all the code that interacts with destination and now is in the pipeline (ie. schema lists, schema storage etc. probably could go to Dataset at some point)

rudolfix (approved):
LGTM! ready for merge!

PR #1946, opened 2024-10-10
SQL Database: Support including/excluding NULL cursor values

### LINKED ISSUE #1935: sql_database source ignores nullable cursor values during incremental loading
### dlt version

1.1.0

### Describe the problem

I'm encountering an issue with the `sql_database` source while using incremental loading. Specifically, when trying to load rows from a table (in my example called `locations`) with a nullable `end_at` timestamp column (similar to a _sessions_ table where `end_at` remains `NULL` until the session ends), the query generated does not correctly handle the `NULL` values in the cursor column.

According to the [documentation on incremental loading](https://dlthub.com/docs/general-usage/incremental-loading#loading-when-incremental-cursor-path-is-missing-or-value-is-nonenull), it should be possible to include rows with `NULL` values in the cursor column by setting `on_cursor_value_missing="include"`. However, the SQL query generated is as follows:
```sql
SELECT locations.id, 
              locations.user_id,
              locations.begin_at,
              locations.end_at,
              locations."primary",
              locations.host,
              locations.campus_id 
FROM locations 
WHERE locations.end_at >= :end_at_1
```
This query **cannot** include any rows where `end_at` is `NULL`, which is the opposite of the expected behavior.

### Expected behavior

The `on_cursor_value_missing="include"` option should generate a query (from [TableLoader._make_query()](https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/sql_database/helpers.py#L77-L112)) that includes rows where `end_at` is `NULL`, without requiring manual intervention or query modification.

```sql
SELECT locations.id, 
       locations.user_id,
       locations.begin_at,
       locations.end_at,
       locations."primary",
       locations.host,
       locations.campus_id 
FROM locations 
WHERE locations.end_at IS NULL OR locations.end_at >= :end_at_1
```

### Steps to reproduce

- Create any table with a cursor column that is nullable
- Insert a row without assigning the latter column
- Attempt to perform an incremental load on it
- It should raise an error by default since `on_cursor_value_missing=raise` is set by default

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

`sql_database`

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

As a workaround, I manually modified the query using the `query_adapter_callback` parameter to include the rows with `NULL` `end_at` values. However, this feels more like a hack than a proper solution.
```python
def allow_pending_locations(query: Select, table: Table):
    """Bug fix to handle NULL end_at"""
    if not query._where_criteria:
        return query
    last_value = query._where_criteria[0].right.effective_value
    return table.select().where(
        or_(
            table.c.end_at.is_(None),
            table.c.end_at > last_value,
        )
    )
...
sql_table(query_adapter_callback=allow_pending_locations...)
```

steinitzu:
### Description

Respect the `on_cursor_value_missing` incremental setting in sqlalchemy source.
Generating WHERE clause equvialent to:

```sql
-- include None
WHERE cursor_col >= :last_value OR cursor_col IS NULL

-- exclude None
WHERE cursor_col >= :last_value AND cursor_col IS NOT NULL
```
I'm not sure if the `exclude` case needs to be handled but thought I'd include explicitly anyway in case any databases interpret `>= NULL` differently.

### Related Issues

- Resolves #1935

### DIFF
--- a/dlt/sources/sql_database/helpers.py
+++ b/dlt/sources/sql_database/helpers.py
@@ -13,8 +13,6 @@ from typing import (
 )
 import operator
 
-from sqlalchemy import and_, or_
-
 import dlt
 from dlt.common.configuration.specs import (
     BaseConfiguration,
@@ -38,7 +36,7 @@ from .schema_types import (
     TTypeAdapter,
 )
 
-from dlt.common.libs.sql_alchemy import Engine, CompileError, create_engine
+from dlt.common.libs.sql_alchemy import Engine, CompileError, create_engine, sa
 
 
 TableBackend = Literal["sqlalchemy", "pyarrow", "pandas", "connectorx"]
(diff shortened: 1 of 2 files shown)


dlt/sources/sql_database/helpers.py
```diff
@@ -13,6 +13,8 @@
 )
 import operator
 
+from sqlalchemy import and_, or_
```
rudolfix, line 16: could you import via our lib, not directly?

dlt/sources/sql_database/helpers.py
```diff
@@ -96,9 +100,15 @@ def _make_query(self) -> SelectAny:
             return query  # type: ignore[no-any-return]
 
         if self.last_value is not None:
-            query = query.where(filter_op(self.cursor_column, self.last_value))
+            where_and_clauses = [filter_op(self.cursor_column, self.last_value)]
```
rudolfix, line 103: queries are OK. should `exclude` option be independent from `if self.last_value is not None:`? I think so. we could skip NONE also on initial load

rudolfix (changes requested):
looks good! see comments for small improvements

rudolfix (approved):
LGTM!

PR #1951, opened 2024-10-14
cli exception handling

sh-rp:
### Description
This PR adds standardized exception handling for the cli plugins.

This includes:
- If a command executes without exception, by default there will be a return code of 0
- If an unknown exception occurs, the return code will be -1 and the default docs url for the command (or base docs url) will be printed out as guidance
- If a known error occurs, the command can throw a CliCommandException to control the return code, provide a custom docs url and determin wether there should be an exception raise if the debug flag is set to true

Notes:
The default return code for unknown exceptions always is -1, so some existing codes have changed so there is no overlap.

### DIFF
--- a/dlt/cli/_dlt.py
+++ b/dlt/cli/_dlt.py
@@ -1,11 +1,13 @@
 from typing import Any, Sequence, Type, cast, List, Dict
 import argparse
+import click
 
 from dlt.version import __version__
 from dlt.common.runners import Venv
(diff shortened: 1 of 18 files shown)


rudolfix (approved):
LGTM! thanks for fixing all standard commands

rudolfix (changes requested):
deploy tests are failing

PR #1957, opened 2024-10-15
feat: add incremental lag for datetime, int, and float cursors

### LINKED ISSUE #970: add lag / attribution window to incremental
**Background**
In many cases certain portion of data should be reacquired during incremental loading. Ie we want to always capture last 7 days of data when getting daily analytics report. Or we want to refresh slack message replies and we have a moving window of 7 days for that.
Technically we would always pass start_date +/- the lag to the function accepting incremental

**Requirements**
1. * [ ] add new optional filed to `Incremental` class that will hold the lag.
2. * [ ] lag should be expressed as float and will be interpreted depending on the type of the cursor. for datetimes it is a lag value in seconds, for any other type use + / - operator depending on the `last_value_func`
3. * [ ] we support only `min` and `max` for `last_value_func`. for customs we do not have "+" operator defined.

donotpush:
### Related Issues

- Fixes #970


rudolfix (changes requested):
@donotpush the way you solve it is good! if you apply `lag` to last_value prop all the code should work.

there are a few things that - mostly a result of not perfect code - that you need to pay attention:
- def parse_native_representation(self, native_value: Any) -> None: - you need to transfer lag manually
- make sure `merge` works (lag is copied via universal mechanism)
- we have `ensure_pendulum_datetime` that converts strings into datetimes (also int and floats) - maybe it is better to use that. you should also convert the result back to string in the same representation. or the comparison will stop working. tldr;> _apply_lag must return exactly the same type as in input
- we should also support date. then lag is in days

one more thing which IMO will be super helpful: allow to define lag in rest_api toolkint:
```class IncrementalConfig(IncrementalRESTArgs, total=False):
    start_param: str
    end_param: Optional[str]
```
you can add `lag` here and hopefully it will be passed to incremental

dlt/common/time.py
```diff
@@ -154,6 +155,29 @@ def ensure_pendulum_time(value: Union[str, datetime.time]) -> pendulum.Time:
     raise TypeError(f"Cannot coerce {value} to a pendulum.Time object.")
 
 
+def detect_datetime_format(value: str) -> Optional[str]:
+    format_patterns = {
```
rudolfix, line 159: this is cool, but if you go beyond just parsing ISO date time, please unit-test a few cases

tests/load/pipeline/test_pipelines.py
```diff
@@ -997,3 +997,126 @@ def events_timezone_unset():
...
+    We validate that the final dataset reflects the correct data updates, taking the lag into account.
+    """
+
+    dataset_name = f"{destination_config.destination_name}{uniq_id()}"
+    pipeline = destination_config.setup_pipeline("pipeline", dataset_name=dataset_name)
+
+    name = "events"
+
+    @dlt.resource(name=name, primary_key="id")
+    def r1(_=dlt.sources.incremental("created_at")):
```
rudolfix, line 1087: the concept of test is good. but IMO you should create just one resource with merge write disposition that you call with the lag you want from the very  beginning. this is how people will use it IMO.

emulate returning "updated_events" on the second call to it ie. via some kind of nonlocal flag that tells to add those events on the second time.

the results of the test look good but please test two additional things
- that you do not apply lag to "initial_value" (set it in incremental)
- there's IMO issue with internal deduplication. please test for updated_events that update id=3:
```
{
                "id": 3,
                "created_at": "2023-03-03T02:00:01Z",
                "event": "updated",
            }
```
```
IMO it will not be included into final result because we'll deduplicate it. you should IMO disable deduplication when lag is defined
```py
@property
    def deduplication_disabled(self) -> bool:
        """Skip deduplication when length of the key is 0"""
        return isinstance(self.primary_key, (list, tuple)) and len(self.primary_key) == 0
```

tests/load/pipeline/test_pipelines.py
```diff
@@ -997,3 +997,126 @@ def events_timezone_unset():
...
+    excluding `id=1` from the updates due to the lag.
+    """
+
+    dataset_name = f"{destination_config.destination_name}{uniq_id()}"
+    pipeline = destination_config.setup_pipeline("pipeline", dataset_name=dataset_name)
+
+    name = "items"
+
+    @dlt.resource(name=name, primary_key="id")
+    def r1(_=dlt.sources.incremental("id")):
```
rudolfix, line 1031: I have similar comments as to the test below:
1. use just one resource with "append" write disposition. you can return the same response each time
2. call the resource several time with different lags (use apply_hints to replace incremental)
3. test if expected elements got duplicated
4. IMO you'll have a deduplication bug that I describe above

rudolfix (changes requested):
please see my comments. on top of that:
1. I'll ask @burnash to help you with adding lag to rest api. Or maybe we'll propose a commit to speed it up. just fix the tests we have so we know all works
2. I'm not sure you test the lag on string and date fields? those should be quite simple tests ie. for append only resources that detect expected duplicates due to lag. see my comments

donotpush:
> * parse_native_representation

Added lag in parse_native_representation but I haven't test it, and also not tested merge.

donotpush:
> @donotpush the way you solve it is good! if you apply `lag` to last_value prop all the code should work.
> 
> there are a few things that - mostly a result of not perfect code - that you need to pay attention:
> 
> * def parse_native_representation(self, native_value: Any) -> None: - you need to transfer lag manually
> * make sure `merge` works (lag is copied via universal mechanism)
> * we have `ensure_pendulum_datetime` that converts strings into datetimes (also int and floats) - maybe it is better to use that. you should also convert the result back to string in the same representation. or the comparison will stop working. tldr;> _apply_lag must return exactly the same type as in input
> * we should also support date. then lag is in days
> 
> one more thing which IMO will be super helpful: allow to define lag in rest_api toolkint:
> 
> ```
>     start_param: str
>     end_param: Optional[str]
> ```
> 
> you can add `lag` here and hopefully it will be passed to incremental

I’ve added lag in parse_native_representation, but I haven’t tested it yet, nor have I tested the merge functionality. The rest of the implementation is complete, except for the rest_api.

tests/load/pipeline/test_pipelines.py
donotpush: Implemented your comments except for number 2:

- call the resource several time with different lags (use apply_hints to replace incremental)

I have used `@pytest.mark.parametrize("lag", [-1, 10])` to execute the code with different lags instead of using apply_hints.

Also facing the deduplication bug - more info above in the other comments.

dlt/common/time.py
donotpush: Added unit-test, and extended ISO support.

donotpush:
I have introduced a bug in IncrementalTransform in my last commit - fixing it.

donotpush:
Fixed the tests:

- test_pipeline_resource_incremental_datetime_lag
- test_pipeline_resource_incremental_int_lag

Please check them carefully because I didn't really manage at anytime to reproduce deduplication bug that you mentioned - I was wrong in my previous comments.

Even without modifying the `deduplication_disabled` code my tests pass which it is an indication to rever the code changes in `deduplication_disabled`:



Screenshot 2024-10-23 at 15 50 43

dlt/common/time.py
rudolfix: amazing :)

dlt/extract/incremental/__init__.py
```diff
@@ -284,11 +300,53 @@ def parse_native_representation(self, native_value: Any) -> None:
...
             self.allow_external_schedulers = merged.allow_external_schedulers
             self.row_order = merged.row_order
+            self._lag = merged.lag
             self.__is_resolved__ = self.__is_resolved__
         else:  # TODO: Maybe check if callable(getattr(native_value, '__lt__', None))
             # Passing bare value `incremental=44` gets parsed as initial_value
             self.initial_value = native_value
 
+    def _apply_lag(self, value: TCursorValue) -> TCursorValue:
+        if self._lag is None:
```
rudolfix, line 310: we should also skip functions that are not min / max and if we do so, we warn the user

tests/load/pipeline/test_pipelines.py
```diff
@@ -997,3 +997,158 @@ def events_timezone_unset():
...
+            yield from initial_entries
+
+    pipeline.run(items_resource)
+    is_second_run = True
+    pipeline.run(items_resource)
+    is_second_run = False
+    is_third_run = True
+    pipeline.run(items_resource)
+
+    expected_results = {
```
rudolfix, line 1068: looks that stuff is working but I'd prefer if you compare results to list of names, not just counts. you can order those by load_id and "id" to get deterministic order that corresponds to insert order

tests/load/pipeline/test_pipelines.py
```diff
@@ -997,3 +997,158 @@ def events_timezone_unset():
...
+            {
+                "id": 3,
+                "created_at": "2023-03-03T02:00:01Z",
+                "event": "3",
+            },  # the lag will be applied here
+        ]
+
+        # Updated entries in second call
+        updated_events = [
+            {"id": 1, "created_at": "2023-03-03T01:00:00Z", "event": "updated"},
```
rudolfix, line 1125: I'd prefer if you name it updated_1, updated_2 etc. so it is clearly visible which updated got it. way easier to verify the test

rudolfix (changes requested):
date format detection function is really cool! I think it will be useful in other places. we still need more tests. I expect this feature to be frequently used and all edge cases will be quickly explored. missing tests:
- can we test how `min` last value function behaves?
- please make sure lag does not impact initial values
- please test if lag is disabled for custom value functions
- can we test lag on a date and float?

also please move incremental tests to tests/extract/incremental. it is sufficient to just run them on duckdb which is available there

donotpush:
@rudolfix implemented all the requested changes.

Please review carefully, specially the end cases.

rudolfix (approved):
LGTM!
we still needs docs before we merge

I did a few fixes please check it out @donotpush 
- I moved lag functions to a separate module. makes incremental code simpler. allows to write unit tests for apply_lag for various edge cases (also done)
- `if self.end_value:` - this also includes "0" which is a valid end value. @donotpush it is OK to use those implicit casts to `bool` but you must always make sure this is exactly what you want. `lag==0` is OK to skip. `end_value==0` is a valid value and ie. we should skip lag also for that (it is tested)
- I refactored the code that makes sure that lagged values are within the range of `initial_value` by just using `last_value_func`.
- `type: ignore` - always give a reason in brackets
- added lag as field to dataclass, not via property
- extended TypedDict to support lag in rest-api

we'll merge this once docs are ready

PR #1966, opened 2024-10-18
Move `exclude_keys()` to `dlt.common.utils`

### LINKED ISSUE #1954: rest_api: move `exclude_keys` to `dlt.common.utils`
Suggestion to move the utility function `exclude_keys` to `dlt.common.utils` from `rest_api.utils`.

burnash:
Resolves #1954

### DIFF
--- a/dlt/common/utils.py
+++ b/dlt/common/utils.py
@@ -504,17 +504,18 @@ def without_none(d: Mapping[TKey, Optional[TValue]]) -> Mapping[TKey, TValue]:
     return {k: v for k, v in d.items() if v is not None}
 
 
-def exclude_keys(d: Mapping[str, Any], keys: Iterable[str]) -> Dict[str, Any]:
-    """Removes specified keys from a dictionary and returns a new dictionary.
+def exclude_keys(mapping: Mapping[str, Any], keys: Iterable[str]) -> Dict[str, Any]:
+    """Create a new dictionary from the input mapping, excluding specified keys.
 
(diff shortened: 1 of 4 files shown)


dlt/common/utils.py
```diff
@@ -504,6 +504,19 @@ def without_none(d: Mapping[TKey, Optional[TValue]]) -> Mapping[TKey, TValue]:
     return {k: v for k, v in d.items() if v is not None}
 
 
+def exclude_keys(d: Mapping[str, Any], keys: Iterable[str]) -> Dict[str, Any]:
```
VioletM, line 507: just a small suggestion: if we're moving this code to the core, maybe we should change the name of the first arg from `d` to something more meaningful? :) Like `data_dict` or `input_dict`

VioletM (dismissed):

dlt/common/utils.py
burnash: This is a very good point, thanks for pointing it out @VioletM, I named it `d` for consistency with some of the other functions from `dlt.common.utils`, e.g. `dict_remove_nones_in_place` or `without_none`. 
But if we look at how Python itself calls that argument, it should be called `mapping` instead since it operates on a mapping type. (examples: https://docs.python.org/3/library/stdtypes.html#dict, https://docs.python.org/3/library/stdtypes.html#str.format_map)
I'll update the PR

VioletM (approved):
👍

PR #1969, opened 2024-10-21
Excludes duplicate checks of docs/examples from mypy and flake8 checks in Makefile

burnash:


sh-rp (approved):

rudolfix (changes requested):
why we disable mypy and flake? there are several serious issues that it detects.
if you want untyped code allow Any in signatures and returns. this can be done in mypy.ini (per folder)

same for flake

burnash:
@rudolfix we're not disabling it, the PR removes checking `docs/example` twice in a row

rudolfix (approved):
LGTM!

PR #1991, opened 2024-10-25
Add open/closed range arguments for incremental

### LINKED ISSUE #1948: Support to customize incremental compare operators for sql_table
### Feature description

The sql_table source/resource is hard coded to use the operator ge (`>=`) when comparing data with incremental state (or le (`>=`) depending on the incremental function/sort order), we would like the option to configure it to use gt (`>`) / lt in stead.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

We have tables which are loaded seldom, but with large volumes, all with the same incremental key value. We would like to run dlt on top of these often, but not use resource re-loading the same data every time no new data has been added.

### Proposed solution

An option to the sql_table source/resource whether to use classic (e.g. operator.ge) or aggressive (operator.gt) incremental filtering on the extracted sql-table.

### Related issues

_No response_

steinitzu:
### Description
This allows configuring whether the incremental range is open or closed on both sides (start/end value).  
Translates to changing the operators between `> | >=` / `< | <=` in sql database source and in incremental filtering.   

Default is `start_range=closed` and `end_range=open`, meaning the exact initial value is included and the exact end value is excluded.    
With `start_range=open` you get `WHERE cursor > last_value` instead of `>=`.  
With `end_range=closed` you get `... cursor <= end_value` so non-overlapping chunks are still possible without the start_value deduplication logic.

### Related Issues

- Fixes #1948

### DIFF
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -184,7 +184,7 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
         self.start_out_of_range: bool = False
         """Becomes true on the first item that is out of range of `start_value`. I.e. when using `max` this is a value that is lower than `start_value`"""
 
-        self._transformers: Dict[str, IncrementalTransform] = {}
+        self._transformers: Dict[Type[IncrementalTransform], IncrementalTransform] = {}
         self._bound_pipe: SupportsPipe = None
         """Bound pipe"""
         self.range_start = range_start
@@ -201,24 +201,6 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
         if self._transformers:
             for transform in self._transformers.values():
                 transform.primary_key = value
-
-    def _make_transforms(self) -> None:
-        types = [("arrow", ArrowIncremental), ("json", JsonIncremental)]
-        for dt, kls in types:
-            self._transformers[dt] = kls(
-                self.resource_name,
-                self.cursor_path,
-                self.initial_value,
-                self.start_value,
-                self.end_value,
-                self.last_value_func,
-                self._primary_key,
-                set(self._cached_state["unique_hashes"]),
-                self.on_cursor_value_missing,
-                self.lag,
-                self.range_start,
-                self.range_end,
-            )
 
     @classmethod
     def from_existing_state(
@@ -503,7 +485,8 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
         )
         # cache state
         self._cached_state = self.get_state()
-        self._make_transforms()
+        # Clear transforms so we get new instances
+        self._transformers.clear()
         return self
 
     def can_close(self) -> bool:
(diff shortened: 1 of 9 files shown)


dlt/extract/incremental/__init__.py
```diff
@@ -111,6 +112,8 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
     row_order: Optional[TSortOrder] = None
     allow_external_schedulers: bool = False
     on_cursor_value_missing: OnCursorValueMissing = "raise"
+    range_start: TIncrementalRange = "closed"
```
rudolfix, line 115: do not forget docstrings and docs

rudolfix: also add this to `TIncrementalArgs` (so we can define those in REST API)

rudolfix: also when start_range is open - disable boundary deduplication. there's no reason to deduplicate. there's no boundary overlap

dlt/extract/incremental/transform.py
```diff
@@ -218,6 +227,8 @@ def __call__(
             ) from ex
         # new_value is "less" or equal to last_value (the actual max)
         if last_value == new_value:
+            if self.range_start == "open":
+                return None, False, False
```
rudolfix, line 231: right! maybe this disables deduplication as well...

dlt/extract/incremental/transform.py
```diff
@@ -311,13 +322,19 @@ def __call__(
 
         if self.last_value_func is max:
             compute = pa.compute.max
-            end_compare = pa.compute.less
-            last_value_compare = pa.compute.greater_equal
+            end_compare = pa.compute.less if self.range_end == "open" else pa.compute.less_equal
```
rudolfix, line 325: we should cache those when creating object (?)

dlt/sources/sql_database/helpers.py
```diff
@@ -87,11 +91,11 @@ def _make_query(self) -> SelectAny:
...
         if last_value_func is max:  # Query ordered and filtered according to last_value function
-            filter_op = operator.ge
-            filter_op_end = operator.lt
+            filter_op = operator.ge if self.range_start == "closed" else operator.gt
+            filter_op_end = operator.lt if self.range_end == "open" else operator.le
         elif last_value_func is min:
-            filter_op = operator.le
-            filter_op_end = operator.gt
+            filter_op = operator.le if self.range_start == "closed" else operator.lt
+            filter_op_end = operator.gt if self.range_end == "open" else operator.ge
```
rudolfix, line 98: we should document this in sql_database docs. we have separate chapter for incremental loading. 
for example if we load incrementally by id or high resolution timestamp or we do not expect stuff to be added (ie. if we have a cursor on day) it is better to keep range open (that disables deduplication and produces faster code)

rudolfix (changes requested):
this looks pretty complete to me! let's work on docs and a few proposed improments

dlt/extract/incremental/transform.py
steinitzu: Not completely. Added this check to the `deduplication_disabled` and a test for it, so it gets disabled now

dlt/extract/incremental/__init__.py
steinitzu: Added all here

dlt/sources/sql_database/helpers.py
steinitzu: Wrote a section on the sql docs, hope it's clear.

rudolfix (approved):
this is very good now!

PR #1992, opened 2024-10-25
Fix BigQueryLoadJob hiding root cause exception

### LINKED ISSUE #1989: BigQueryLoadJob exception formatting hides the initial error message
### dlt version

1.2.0

### Describe the problem

We get an error (totally valid) that the table we use as a destination is not found in BQ while trying to write data to this table. The error is stored in `_exception` field of an object of `BigQueryLoadJob` and looks like this:
```
DatabaseUndefinedRelation(NotFound('POST https://bigquery.googleapis.com/upload/bigquery/v2/projects/our-project/jobs?uploadType=resumable: Not found: Table <our_project>:<our_dataset>.<our_table> was not found in location EU'))
```

The code is trying to return structured exception from the job:
https://github.com/dlt-hub/dlt/blob/774ad5c841efe11411f37dc1fd9c8ac9e2877cc2/dlt/load/load.py#L409

Here is the implementation of `expection` method in `BigQueryLoadJob`:
https://github.com/dlt-hub/dlt/blob/774ad5c841efe11411f37dc1fd9c8ac9e2877cc2/dlt/destinations/impl/bigquery/bigquery.py#L126-L135

The problem is the `_exception` field of `job` object (inherited from `RunnableLoadJob`) contains the error but the `_bq_load_job` field is `None`. This leads to the situation that the initial error is hidden by not meaningful error 

```
<class 'AttributeError'> 'NoneType' object has no attribute 'error_result'
```

### Expected behavior

I expect to receive 
```
DatabaseUndefinedRelation(NotFound('POST https://bigquery.googleapis.com/upload/bigquery/v2/projects/our-project/jobs?uploadType=resumable: Not found: Table <our_project>:<our_dataset>.<our_table> was not found in location EU'))
```
or similar message as an error.

### Steps to reproduce

1. Setup some `dlt` pipeline using BigQuery as a destination.
2. Run the pipeline and check that destination table and `_dlt_` service tables were created.
3. Delete only destination table and keep all `_dlt_` service tables.
4. Run the pipeline again.

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

_No response_

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

_No response_

xneg:
…parents exception() method.

### Description
Explicitly check that _bg_load_job is not None otherwise fallback to parents exception() method.

### Related Issues
- Fixes #1989

### Additional Context
The detailed description is provided in the issue

### DIFF
--- a/dlt/destinations/impl/bigquery/bigquery.py
+++ b/dlt/destinations/impl/bigquery/bigquery.py
@@ -123,15 +123,17 @@ class BigQueryLoadJob(RunnableLoadJob, HasFollowupJobs):
                 )
 
     def exception(self) -> str:
-        return json.dumps(
-            {
-                "error_result": self._bq_load_job.error_result,
-                "errors": self._bq_load_job.errors,
-                "job_start": self._bq_load_job.started,
-                "job_end": self._bq_load_job.ended,
-                "job_id": self._bq_load_job.job_id,
-            }
-        )
+        if self._bq_load_job:
+            return json.dumps(
+                {
+                    "error_result": self._bq_load_job.error_result,
+                    "errors": self._bq_load_job.errors,
+                    "job_start": self._bq_load_job.started,
+                    "job_end": self._bq_load_job.ended,
+                    "job_id": self._bq_load_job.job_id,
+                }
+            )
+        return super().exception()
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


rudolfix (approved):
LGTM! thanks!

PR #1993, opened 2024-10-27
Added partial loading example

dat-a-man:
### Description

Adding partial loading example.


docs/examples/partial_loading/partial_loading.py
```diff
@@ -0,0 +1,179 @@
...
+                    "endpoint": {
+                        "path": f"player/{username}/games/{year}/{month_str}",  # API endpoint path
+                    },
+                    "primary_key": ["url"],  # Primary key to prevent duplicates
+                }
+            ],
+        )
+
+        # Fetch data from the API using the configured REST API source
+        data = list(rest_api_source(config))
```
AstrakhantsevaAA, line 63: we you turn source into list and then wrap it into resource again? just use rest_api_resource instead and use apply_hints()

AstrakhantsevaAA (approved):

PR #1994, opened 2024-10-27
loads secrets from colab userdata and steamlit + bugfixes

### LINKED ISSUE #1885: Bug in duckdb credentials resolution
When using two pipeline instances with a duckdb destination, the credentials are not resolved the right way. This does not seem to be the case for other destinations.

```
p1 = dlt.pipeline("p1", destination=duckdb(credentials="./duck1.duckdb"))
p2 = dlt.pipeline("p2", destination=duckdb(credentials="./duck2.duckdb"))

# both below lines print "duckdb:////Volumes/dlthub/dlt/duck2.duckdb"
# but the first one should be duck1.duckdb
print(p1.destination_client().config.credentials)
print(p2.destination_client().config.credentials)
```

rudolfix:
### Description
Fixes several small errors and other small improvements
fixes #1885 
See commit list

### DIFF
--- a/dlt/destinations/impl/duckdb/configuration.py
+++ b/dlt/destinations/impl/duckdb/configuration.py
@@ -173,13 +173,17 @@ class DuckDbCredentials(DuckDbBaseCredentials):
                 pipeline_path = explicit_path or DUCK_DB_NAME % self.bound_to_pipeline.pipeline_name
                 # if explicit_path was an absolute path it will be used
                 pipeline_path = os.path.join(initial_cwd, pipeline_path)
-                if not self.bound_to_pipeline.first_run:
-                    if not os.path.exists(pipeline_path):
-                        logger.warning(
-                            f"Duckdb attached to pipeline {self.bound_to_pipeline.pipeline_name} in"
-                            f" path {os.path.relpath(pipeline_path)} was could not be found but"
-                            " pipeline is marked as used.  so database is expected to be present."
-                        )
+            if not self.bound_to_pipeline.first_run:
+                if not os.path.exists(pipeline_path):
+                    logger.warning(
+                        f"Duckdb attached to pipeline {self.bound_to_pipeline.pipeline_name} in"
+                        f" path {os.path.relpath(pipeline_path)} was could not be found but"
+                        " pipeline has already ran. This may be a result of (1) recreating or"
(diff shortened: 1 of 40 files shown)


dlt/pipeline/pipeline.py
```diff
@@ -392,6 +390,12 @@ def drop(self, pipeline_name: str = None) -> "Pipeline":
             self.config,
             self.runtime_config,
         )
+        new_pipeline.activate()
```
sh-rp: I suppose there isn't  a way to replace this pipeline with the new one so that this is transparent to the user? I would probably expect to be able to do:
p.run(...)
p.drop()
p.run(...)

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -226,10 +211,18 @@ def __init__(
...
             credentials=credentials,  # type: ignore[arg-type]
             destination_name=destination_name,
             environment=environment,
         )
+        self.bound_to_pipeline = bound_to_pipeline
         self.create_indexes = create_indexes
+
+    def on_resolved(self) -> None:
+        # pass bound pipeline to duckdb credentials
+        # TODO: find a better way to pass and bind explicit pipeline context
```
sh-rp, line 230: I am also not sure, but maybe we should not have an "active pipeline" but rather some pipeline context object that is shared among the whole object tree of a given pipeline, similar to how other libraries do it. But I'm not sure what other consequences would come from not having active pipelines anymore. Also what happens if we actually start using dataset objects as "destinations" and share them between multiple pipelines, do we then have to replace this context object everytime? In this scenario the active pipeline probably is the better so solution..

sh-rp (dismissed):

dlt/destinations/impl/duckdb/configuration.py
rudolfix: the primary use of active pipeline is to have access to state, schemas, clients etc. in dlt source and resources. this is OK because we execute them from within the pipeline (except sources that may be instantiated before pipeline is created and activated, cost of freedom ;>)

dlt/pipeline/pipeline.py
rudolfix: I completely reworked this part. And now it works like you say :)

PR #2000, opened 2024-10-29
1990 - dataset columns select and limit

sh-rp:
### Description

Adds "limit", "select" and "head" methods to readablerelation to be able to limit the amount of rows returned as well as select a subset of columns.

The interface is done in an sqlalchemy chaining style. So to select the first 10 rows and two columns of a table and get that as an arrow table, you will write:

```
arrow_table = dataset.my_customers.select(["name", "birthdate"]).limit(10).arrow()
```

Questions:
* I am normalizing table_names and column_names from the user input now. I think this should be good. Lmk.
* I think it might be quite easy to add sorting, should I try?
* We could add support for getting the portion of a relation with a certain load id or list of load ids. This is native dlt stuff and would be cool to have, no?
* Maybe we could allow the user to manually add a where clause to the statement of a relation somehow. I have a feeling some will want to have filtering and full filtering support may be a bit tricky, but maybe it's not actually that tricky.


dlt/destinations/dataset.py
```diff
@@ -42,18 +58,56 @@ def __init__(
...
+        table_name = self._client.make_qualified_table_name(self._table_name)
+
+        maybe_limit_clause = ""
+        if self._limit:
+            maybe_limit_clause = f"LIMIT {self._limit}"
+
+        selector = "*"
+        if self._selected_columns:
+            selector = ",".join(
+                [self._naming.normalize_identifier(c) for c in self._selected_columns]
```
sh-rp, line 76: NOTE: do we need to escape identifiers here? The tests will show ... :)

rudolfix (approved):
LGTM
i did a small change to column selector to allow pandas-like index. also I fixed name norm

regarding improvements
- getting partitions of data: yes! we should support getting "new" data somehow... for example by writing back to `_load_ids` table with higher statuses
- WHERE clause: better integrate with ibis. and we have SQL already

PR #2004, opened 2024-10-30
ibis support - hand over credentials to ibis backend for a number of destinations

sh-rp:
### Description
Adds support for handing over our credentials to an ibis backend. Supported destinations are:

```
    "dlt.destinations.postgres",
    "dlt.destinations.duckdb",
    "dlt.destinations.motherduck",
    "dlt.destinations.filesystem",
    "dlt.destinations.bigquery",
    "dlt.destinations.snowflake",
    "dlt.destinations.redshift",
    "dlt.destinations.mssql",
    "dlt.destinations.synapse",
    "dlt.destinations.clickhouse", 
```

Discussion at ibis about using ibis only as a query builder: https://github.com/ibis-project/ibis/discussions/10452. There also is a question there about setting a default database (they call what we call dataset a database)


dlt/pipeline/pipeline.py
```diff
@@ -1729,3 +1734,11 @@ def _dataset(self, dataset_type: TDatasetType = "dbapi") -> SupportsReadableData
             schema=(self.default_schema if self.default_schema_name else None),
             dataset_type=dataset_type,
         )
+
+    def _ibis(self) -> IbisBackend:
```
sh-rp, line 1738: I don't think our original idea of exposing an ibis dataset via the same methods as the dbapi one makes sense, the interface of ibis is completely different to ours.

dlt/destinations/dataset.py
```diff
@@ -318,3 +321,29 @@ def dataset(
...
+
+    if destination.destination_type not in [
+        "dlt.destinations.postgres",
+        "dlt.destinations.duckdb",
+        "dlt.destinations.filesystem",
+    ]:
+        raise NotImplementedError()
+
+    ibis = ibis.connect(client.config.credentials.to_native_representation())
+    # NOTE: there seems to be no standardized way to set the current dataset / schema in ibis
```
sh-rp, line 346: I'm not sure where we should put the stuff that converts our credentials into something IBIS understands. `client.config.credentials.to_native_representation()` could work for many cases, but selecting the default dataset will probably be something destination specific..

sh-rp: Also for the filesystem destination this will work quite differently, because we first need to populate the duckdb database and then attach ibis to it.

dlt/pipeline/pipeline.py
rudolfix: but you can use overloads on Literal `dataset_type` 
```py
@overload
def _dataset(self, dataset_type: TDatasetType = "dbapi") -> SupportsReadableDataset:
   ...
@overload
def _dataset(self, dataset_type: TDatasetType = "ibis") -> IbisBackend:
   ...
 ```
 to return different types. hmmm but maybe you are right. we should implement readable dataset interface on ibis.
 
 so maybe we add `ibis()` on DBAPI dataset? that will return ibis backend? I do not think adding it on a relation makes sense

File: dlt/common/destination/reference.py [hunk 3]
```diff
@@ -493,19 +497,55 @@ def df(self, chunk_size: int = None) -> Optional[DataFrame]:
         """
         ...
 
-    def arrow(self, chunk_size: int = None) -> Optional[ArrowTable]: ...
+    # accessing data
+    def arrow(self, chunk_size: int = None) -> Optional[ArrowTable]:
+        """fetch arrow table of first 'chunk_size' items"""
+        ...
+
+    def ibis(self, chunk_size: int = None) -> Optional[IbisTable]: ...
```
rudolfix, line 505: I'd add it on `ReadableDBAPIDataset` and return implementation of it in `_dataset` of pipeline

dlt/destinations/dataset.py
```diff
@@ -42,18 +90,91 @@ def __init__(
         self.iter_arrow = self._wrap_iter("iter_arrow")  # type: ignore
         self.iter_fetch = self._wrap_iter("iter_fetch")  # type: ignore
 
+    # TODO: where should this go, should cursors support "native" ibis or do we do a conversion somewhere
+    def ibis(self, chunk_size: int = None) -> Optional[IbisTable]:
```
rudolfix, line 94: this does not make sense... we have stuff materialized. way better to just do df() which has data frame interface

dlt/common/destination/reference.py
```diff
@@ -493,19 +497,55 @@ def df(self, chunk_size: int = None) -> Optional[DataFrame]:
...
+
+    def iter_df(self, chunk_size: int) -> Generator[DataFrame, None, None]:
+        """iterate over data frames tables of 'chunk_size' items"""
+        ...
+
+    def iter_arrow(self, chunk_size: int) -> Generator[ArrowTable, None, None]:
+        """iterate over arrow tables of 'chunk_size' items"""
+        ...
+
+    def iter_ibis(self, chunk_size: int) -> Generator[IbisTable, None, None]: ...
```
rudolfix, line 515: how are you going to implement that? you'll need to create IbisTable on a chunk of data which is lazy evaluated. otherwise does not make sense

dlt/destinations/dataset.py
rudolfix: 1. part that creates backend should go to dlt/helpers/ibis
2. part that converts credentials should go to common/libs/ibis
here we can improve a lot. we already pollute credentials in common with concrete converting functions (like fsspec, delta etc.) we should create some universal mechanism to convert them.

dlt/destinations/dataset.py
```diff
@@ -74,6 +195,35 @@ def _wrap(*args: Any, **kwargs: Any) -> Any:
...
+        # NOTE: the line below will ensure that no unknown columns are selected if
+        # schema is known
+        rel.computed_schema_columns
+        return rel
+
+    def head(self) -> "ReadableDBAPIRelation":
+        return self.limit(5)
+
 
 class ReadableDBAPIDataset(SupportsReadableDataset):
```
rudolfix, line 228: let's move ibis() here. then we hand over the connection and return the ibis backend

rudolfix: can we move schema property to the Protocol?

dlt/destinations/dataset.py
```diff
@@ -178,3 +317,58 @@ def dataset(
...
+        # NOTE: we should probably have the option for the user to only select a subset of tables here
+        with sql_client as _:
+            sql_client.create_views_for_all_tables()
+        ibis = ibis.duckdb.from_connection(duck)
+
+    else:
+        raise NotImplementedError()
+
+    # NOTE: there seems to be no standardized way to set the current dataset / schema in ibis
+    ibis.raw_sql(f"SET search_path TO {dataset_name};")
```
rudolfix, line 361: why? you just do:
return ibis[dataset_name]
to select namespace.
https://duckdb.org/docs/guides/python/ibis.html

rudolfix (changes requested):

dlt/pipeline/pipeline.py
sh-rp: I like the overload, I did not know this was possible...

rudolfix:
also here you wrote a helper: https://github.com/dlt-hub/dlt/pull/1491/files

dlt/destinations/dataset.py
sh-rp: where does it say that on this page? In any case it does not work and I also looked for a bit before and could not find any standard way to select the current database (as it is called in ibis). The call the whole thing (in the case of duckdb the file) a catalog and what we call dataset is called database there. You can list the database but not set a default one.. At least I have not found a method.

dlt/destinations/dataset.py
sh-rp: I removed those

dlt/common/destination/reference.py
sh-rp: removed it

dlt/destinations/dataset.py
sh-rp: it seems like there is no way to consistently pre-select a database, and for example setting the schema on connection in snowflake for some reason also does not work, so I am opting now to not set any default schema/database on a connection and we need to tell the user to give the database name for listing and retrieving tables when using the ibis interfaces.

tests/load/test_read_interfaces.py
```diff
@@ -251,6 +251,56 @@ def test_db_cursor_access(populated_pipeline: Pipeline) -> None:
...
+            "_dlt_loads",
+            "_dlt_pipeline_state",
+            "_dlt_version",
+            "double_items",
+            "items",
+            "items__children",
+        ]
+    }
+
+    # TODO, ibis needs describe permission on snowflake?
```
sh-rp, line 298: final todo

rudolfix (changes requested):
very good. I'm for keeping `ibis` there if we are going to implement ibis-based dataset.

tests/load/test_read_interfaces.py
```diff
@@ -251,6 +251,71 @@ def test_db_cursor_access(populated_pipeline: Pipeline) -> None:
...
+    "populated_pipeline",
+    configs,
+    indirect=True,
+    ids=lambda x: x.name,
+)
+def test_ibis_dataset_access(populated_pipeline: Pipeline) -> None:
+    # NOTE: we could generalize this with a context for certain deps
+    import subprocess
+
+    subprocess.check_call(
```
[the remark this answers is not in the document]
sh-rp: it already is, but I forgot to add the extras, I just added those.

dlt/common/libs/ibis.py
```diff
@@ -0,0 +1,121 @@
...
+    # NOTE: Athena could theoretically work with trino backend, but according to
+    # https://github.com/ibis-project/ibis/issues/7682 connecting with aws credentials
+    # does not work yet.
+    # "dlt.destinations.athena",
+]
+
+
+def create_ibis_backend(
+    destination: TDestinationReferenceArg, dataset_name: str, client: JobClientBase
+) -> BaseBackend:
```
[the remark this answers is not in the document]
sh-rp: removed

dlt/destinations/dataset.py
```diff
@@ -310,3 +325,14 @@ def dataset(
     if dataset_type == "dbapi":
         return ReadableDBAPIDataset(destination, dataset_name, schema)
     raise NotImplementedError(f"Dataset of type {dataset_type} not implemented")
+
+
+# helpers
+def _get_client_for_destination(
```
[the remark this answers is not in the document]
sh-rp: I did this, let's see if the tests pass. I maybe we should also put the function for creating a dataset_name there?

rudolfix (approved):
LGTM!

PR #2016, opened 2024-11-02
Fix pagination issue in `JSONResponseCursorPaginator` with empty string cursor value

### LINKED ISSUE #2012: Empty String Cursor Value from Zoom API Leading to Pagination Issue in `JSONResponseCursorPaginator`
### dlt version

dlt 1.3.0

### Describe the problem

I encountered an issue when using the [JSONResponseCursorPaginator](https://github.com/dlt-hub/dlt/blob/75ae23f174e5357a01612e48fe8a475c2db9db35/dlt/sources/helpers/rest_client/paginators.py#L595) class in rest_client for pagination requests while interacting with Zoom's API. The `update_state()` method is implemented as follows:

https://github.com/dlt-hub/dlt/blob/75ae23f174e5357a01612e48fe8a475c2db9db35/dlt/sources/helpers/rest_client/paginators.py#L651-L654

The problem occurs when values is equal to `[""]`, resulting in self._next_reference being set to `""` instead of None. This leads to pagination requests continuing to the next page, whereas I expect it to be None to stop further requests.

### Expected behavior

The expected behavior is for `self._next_reference` to be set to None when values is `[""]`, thus preventing further pagination requests. This matches the anticipated termination condition logic.

### Steps to reproduce

1. Use the `JSONResponseCursorPaginator` to implement pagination requests.
2. When the value extracted from the JSON response using the specified path is `[""]`, observe the pagination behavior.
3. Note whether `self._next_reference` is set to `""` instead of None.

### Operating system

Linux

### Runtime environment

Kubernetes

### Python version

3.11

### dlt data source

ZOOM api, e.g. [GET /users](https://developers.zoom.us/docs/api/users/#tag/users/GET/users)

### dlt destination

Snowflake

### Other deployment details

_No response_

### Additional information

I plan to submit a PR to address this issue.

kang8:
### Description
Adjusts the `update_state()` method to set `_next_reference` to None
when the cursor value extracted from the JSON response is an empty
string, preventing unintended pagination requests.

### Related Issues

- Fixes #2012

### DIFF
--- a/dlt/sources/helpers/rest_client/paginators.py
+++ b/dlt/sources/helpers/rest_client/paginators.py
@@ -651,7 +651,7 @@ class JSONResponseCursorPaginator(BaseReferencePaginator):
     def update_state(self, response: Response, data: Optional[List[Any]] = None) -> None:
         """Extracts the cursor value from the JSON response."""
         values = jsonpath.find_values(self.cursor_path, response.json())
-        self._next_reference = values[0] if values else None
+        self._next_reference = values[0] if values and values[0] else None
 
     def update_request(self, request: Request) -> None:
         """Updates the request with the cursor query parameter."""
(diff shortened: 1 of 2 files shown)


burnash (approved):
Thanks you for the fix @kang8

willi-mueller:
Great job, thank you @kang8!

PR #2017, opened 2024-11-03
duckdb filesystem custom secrets

sh-rp:
### Description
* Use custom secrets location for fs tests
* Warn the user if they use the default location when creating persistent secrets
* Prevent persistent secrets for in memory database versions
* Add some secrets tests

Note:
I tried implementing changing the secrets location in create_authentication, but somehow I could not get it right. I think this solution with warning the user should be good for now.


dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -74,12 +73,27 @@ def drop_authentication(self, secret_name: str = None) -> None:
...
 
+        if persistent and self.memory_db:
+            raise Exception("Creating persistent secrets for in memory db is not allowed.")
+
+        current_secret_directory = self._conn.sql(
+            "SELECT current_setting('secret_directory') AS secret_directory;"
+        ).fetchone()[0]
+
+        # TODO: what is with windows?
+        is_default_secrets_directory = current_secret_directory.endswith("/.duckdb/stored_secrets")
```
rudolfix, line 88: not going to work because windows use backslashes. ie. 
```
D SELECT current_setting('secret_directory') AS secret_directory;
┌──────────────────────────────────────────┐
│             secret_directory             │
│                 varchar                  │
├──────────────────────────────────────────┤
│ C:\Users\rudolfix\.duckdb\stored_secrets │
└──────────────────────────────────────────┘
```

please use Path to compare paths

tests/load/filesystem/test_sql_client.py
```diff
@@ -297,11 +312,93 @@ def test_delta_tables(destination_config: DestinationTestConfiguration) -> None:
...
     )
 
 
+@pytest.mark.essential
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(all_buckets_filesystem_configs=True, bucket_subset=(AWS_BUCKET,)),
+    ids=lambda x: x.name,
+)
+def test_secrets_management(destination_config: DestinationTestConfiguration) -> None:
```
rudolfix, line 327: do we need this test in the load? if you move it to `tests/pipelines` it will execute on windows as well. then you'll see if you normalize the paths correctly

rudolfix (changes requested):

dlt/destinations/impl/filesystem/sql_client.py
sh-rp: done

tests/load/filesystem/test_sql_client.py
sh-rp: moved it

pytest.ini
```diff
@@ -1,7 +1,7 @@
 [pytest]
 pythonpath= dlt docs/website/docs
 norecursedirs= .direnv .eggs build dist
-addopts= -v --showlocals --durations 10
```
sh-rp, line 4: this is needed to be able to capture the logs with the pytest capfd

sh-rp: ok, somehow it does not work on ci...

pyproject.toml
```diff
@@ -89,6 +89,7 @@ alembic = {version = ">1.10.0", optional = true}
 paramiko = {version = ">=3.3.0", optional = true}
 sqlglot = {version = ">=20.0.0", optional = true}
 db-dtypes = { version = ">=1.2.0", optional = true }
+aiohttp = { version = ">=3.9", optional = true }
```
sh-rp, line 92: this seems to be needed for python 3.12

rudolfix (approved):
this is really good. but you add a lot of dependencies to common tests (filesystem, s3). better if we move `test_filesystem_sql_secrets` to `tests/load` where all deps are available. I'd like to keep deps in common tests minimal

rudolfix (changes requested):
this is really good. but you add a lot of dependencies to common tests (filesystem, s3). better if we move test_filesystem_sql_secrets to tests/load where all deps are available. I'd like to keep deps in common tests minimal

sh-rp:
@rudolfix done!

rudolfix (approved):
LGTM!

PR #2022, opened 2024-11-04
enable `delta` partitioning on arrow normalizer load id

jorritsandbrink:
- adds logic to handle [bug](https://github.com/delta-io/delta-rs/issues/2969) in delta-rs
- enables partitioning on load id column added by Arrow normalizer when using `delta` table format


rudolfix (dismissed):
LGTM!

rudolfix (approved):
LGTM!

Issue #2023, opened 2024-11-04
Code failing with dlt 1.3.0


senickel:
I have some code that worked fine with 1.2.0 and now fails with 1.3.0 with this error:
```
current_value, traces = _resolve_config_field(
^^^^^^^^^^^^^^^^^^^^^^
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/resolve.py", line 321, in _resolve_config_field
value, traces = _resolve_single_value(
^^^^^^^^^^^^^^^^^^^^^^
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/resolve.py", line 420, in _resolve_single_value
providers_context = container[PluggableRunContext].providers
~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/container.py", line 67, in __getitem__
item = spec()
^^^^^^
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/specs/pluggable_run_context.py", line 72, in __init__
self._plug(run_dir=None)
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/specs/pluggable_run_context.py", line 125, in _plug
m = plugins.manager()
^^^^^^^^^^^^^^^^^
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/plugins.py", line 41, in manager
return Container()[PluginContext].manager
~~~~~~~~~~~^^^^^^^^^^^^^^^
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/container.py", line 67, in __getitem__
item = spec()
^^^^^^
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/plugins.py", line 34, in __init__
load_setuptools_entrypoints(self.manager)
File "/bin/app/lib/python3.11/site-packages/dlt/common/configuration/plugins.py", line 51, in load_setuptools_entrypoints
if hasattr(dist, "name") and not dist.name.startswith("dlt-"):
^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'startswith' 
```
I apologize for not providing a minimum example yet. However, I wanted to report this right away.

burnash:
Fixed in #2024

PR #2024, opened 2024-11-04
fix: if name of distribution is None

senickel:
### Description
I am getting this [exception](https://github.com/dlt-hub/dlt/issues/2023) starting with dlt version 1.3.0. 
I do not know how to reproduce it but the underlying issue seems to be that the `name` attribute  of `Distribution` can be `None`.


rudolfix (approved):

sh-rp:
@senickel thanks for reporting and fixing this!

PR #2025, opened 2024-11-05
Fix the deprecation warning in `.common.configuration.container`

### LINKED ISSUE #2015: Deprecation Warning in Multithreading
Hello,
My log is littered with two repeating warnings:
```
/app/.venv/lib/python3.10/site-packages/dlt/common/configuration/container.py:95: DeprecationWarning: currentThread() is deprecated, use current_thread() instead
  if m := re.match(r"dlt-pool-(\d+)-", threading.currentThread().getName()):
```

```
/app/.venv/lib/python3.10/site-packages/dlt/common/configuration/container.py:95: DeprecationWarning: getName() is deprecated, get the name attribute instead
```

https://github.com/dlt-hub/dlt/blob/21c636388f197eac538bd7e262a342cb294752eb/dlt/common/configuration/container.py#L95

Is there a possible update to avoid this on Python 3.10?

burnash:
Fixes #2015

### DIFF
--- a/dlt/common/configuration/container.py
+++ b/dlt/common/configuration/container.py
@@ -92,7 +92,7 @@ class Container:
             return self.main_context
         else:
             # thread pool names used in dlt contain originating thread id. use this id over pool id
-            if m := re.match(r"dlt-pool-(\d+)-", threading.currentThread().getName()):
+            if m := re.match(r"dlt-pool-(\d+)-", threading.current_thread().name):
                 thread_id = int(m.group(1))
             else:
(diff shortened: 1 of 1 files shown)


sh-rp (commented):
looks good, seems to be around since before python 3.9

burnash:
> looks good, seems to be around since before python 3.9

Even earlier: https://docs.python.org/3.8/library/threading.html#threading.Thread 
So there's no need to handle anything for 3.8.

PR #2026, opened 2024-11-05
fix resource level max_table_nesting and normalizer performance tuning

sh-rp:
### Description
The resource level max_table_nesting settings where not passed down to child tables, this PR fixes this. I also completely rewrote the tests to be much more readable (imho), also they were not testing various cases although they were claiming to. 

b6c7d035-1ea6-46f9-a17b-24ea78c7e918

### DIFF
--- a/dlt/common/normalizers/json/relational.py
+++ b/dlt/common/normalizers/json/relational.py
@@ -96,7 +96,7 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
         # self.primary_keys = Dict[str, ]
 
     def _flatten(
-        self, table: str, dict_row: DictStrAny, _r_lvl: int
+        self, table: str, dict_row: DictStrAny, parent_path: Tuple[str, ...], _r_lvl: int
     ) -> Tuple[DictStrAny, Dict[Tuple[str, ...], Sequence[Any]]]:
         out_rec_row: DictStrAny = {}
         out_rec_list: Dict[Tuple[str, ...], Sequence[Any]] = {}
@@ -117,7 +117,7 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
                 # for lists and dicts we must check if type is possibly nested
                 if isinstance(v, (dict, list)):
                     if not self._is_nested_type(
-                        self.schema, table, nested_name, self.max_nesting, __r_lvl
+                        self.schema, table, nested_name, self.max_nesting, parent_path, __r_lvl
                     ):
                         # TODO: if schema contains table {table}__{nested_name} then convert v into single element list
                         if isinstance(v, dict):
@@ -268,7 +268,7 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
         schema = self.schema
         table = schema.naming.shorten_fragments(*parent_path, *ident_path)
         # flatten current row and extract all lists to recur into
-        flattened_row, lists = self._flatten(table, dict_row, _r_lvl)
+        flattened_row, lists = self._flatten(table, dict_row, parent_path, _r_lvl)
         # always extend row
         DataItemNormalizer._extend_row(extend, flattened_row)
         # infer record hash or leave existing primary key if present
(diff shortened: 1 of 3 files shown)


dlt/common/normalizers/json/relational.py
```diff
@@ -423,10 +423,21 @@ def _normalize_prop(
...
+        schema: Schema, table_name: str, parent_path: Tuple[str, ...]
+    ) -> Optional[int]:
+        """gets table nesting level, will inherit from parent if not set"""
+
+        # try go get table directly
         table = schema.tables.get(table_name)
+
+        # if table is not found, try to get it from root path
+        if not table and parent_path:
+            table = schema.tables.get(parent_path[0])
```
sh-rp, line 436: should we check intermediary tables here too? so if parent path is ["one", "two"] should we also check table "one__two" or is it always correct to use the top table? I'm not sure there is a case where there would be settings in a intermediary table.

rudolfix: we need only parent table. the overall fix is good but maybe we could optimize this block a little? it is always top level table that has this setting  so maybe we should pass root table as table_name here? then no changes are needed

dlt/common/normalizers/json/relational.py
```diff
@@ -96,7 +96,7 @@ def _reset(self) -> None:
         # self.primary_keys = Dict[str, ]
 
     def _flatten(
-        self, table: str, dict_row: DictStrAny, _r_lvl: int
+        self, table: str, dict_row: DictStrAny, parent_path: Tuple[str, ...], _r_lvl: int
```
rudolfix, line 99: this is probably the most often called function in our code. maybe we can skip passing additional argument here or do something else to make it faster?
- count the recursion level down (r_lvl) with the initial value initialized to max nesting per table. so we do not unnest when  below _r_lvl
- pass just parent table name, should be faster than tuple

rudolfix (changes requested):
this is good just look at performance again because function here are often called

sh-rp:
For the following code the time goes from 4.5s on devel to about 3.1 seconds on this branch with the normalization caching:
```

import json
import os
import time
from tests.common.utils import json_case_path
from dlt.common.schema import Schema

def rasa_event_bot_metadata():
    with open(json_case_path("rasa_event_bot_metadata"), "rb") as f:
        return json.load(f)
    
def norm():
    return Schema("default").data_item_normalizer  # type: ignore[return-value]

if __name__ == "__main__":
    
    payload = rasa_event_bot_metadata()

    for i in range(1):
        n = norm() 
        start = time.time()
        for i in range(1000):
            payload["_id"] = i
            list(n.normalize_data_item(payload, "load_id", "table"))
        print(f"Time taken: {time.time() - start}")


```

dlt/common/normalizers/json/relational.py
```diff
@@ -422,12 +427,39 @@ def _normalize_prop(
             validator_f=column_name_validator(schema.naming),
         )
 
+    #
+    # Cached helper methods for all operations that are called often
+    #
+    @staticmethod
+    @lru_cache(maxsize=None)
```
sh-rp, line 434: this adds a considerable speed boost. we could also consider adding caching support on the naming and not here so that all normalizers and other places can benefit, I'm not quite sure if there are other places where this gets called as often as here though.

rudolfix: this is interesting because this functions is already cached. are you using `snake_case`? convention? please look at the underlying code again

dlt/common/normalizers/json/relational.py
```diff
@@ -422,12 +427,39 @@ def _normalize_prop(
...
+    def _shorten_fragments(schema: Schema, *idents: str) -> str:
+        return schema.naming.shorten_fragments(*idents)
+
+    @staticmethod
+    @lru_cache(maxsize=None)
+    def _normalize_table_identifier(schema: Schema, table_name: str) -> str:
+        return schema.naming.normalize_table_identifier(table_name)
```
rudolfix, line 441: this is also cached already

rudolfix (commented):
please see my comments. let's check where the improvements are coming from, becasue I already cache ident normalizers...

sh-rp:
@rudolfix it's using snake_case naming. There are a few of operations that get called over and over again because they are not cached. For example if you call normalize_identifier, then this is not cached, only _normalize_identifier is. The snake_case class calls the super class where strip is called and some checks are made. The strip alone accounts for 0.2s of those 1.7s or so gains. Then for example in `shorten_fragments` `make_paths` is calculated over and over again which also accounts for a bit of the savings etc. I can try to move all the caching to the schema if you like.

rudolfix (approved):
OK! there are other naming conventions that are not cached so that will speed things up. All non-deterministic naming conventions will stop working but no one should write those :)

PR #2027, opened 2024-11-05
[PoC] - prevent leaking of secrets when running validate_dict

sh-rp:


dlt/common/utils.py
```diff
@@ -635,3 +635,42 @@ def __getattribute__(self, name: str) -> Any:
                 raise RuntimeError("This instance has been dropped and cannot be used anymore.")
 
     return DefunctClass
+
+
+def obfuscate_string(s: str) -> str:
+    """Obfuscates string by replacing some or all characters with asterisks"""
+    if len(s) < 6:
```
willi-mueller, line 642: I'd consider replacing this magic number with a constant and maybe use a single constant for both this place as well as here: https://github.com/dlt-hub/dlt/blob/c0e4795b2c6854238df7741f03e067fb789ce442/dlt/sources/rest_api/__init__.py#L449

willi-mueller (commented):
This implementation is nice! I think it would allow us to delete TODO comments, such as here: https://github.com/dlt-hub/dlt/blob/c0e4795b2c6854238df7741f03e067fb789ce442/dlt/sources/rest_api/__init__.py#L434

Just in case you have discussed this already, you could maybe update the discussion here with your results:
https://github.com/dlt-hub/dlt/pull/1797

Anton and I argued for the method this PR proposes, Marcin was against it.

dlt/common/utils.py
sh-rp: actually I think we can get rid of this mask_secrets code entirely, no?

sh-rp:
@willi-mueller what where the arguments against implementing it this way?

burnash (commented):
Looks good to me, two important from my point of view to consider:
1. Allow the user to disable the obfuscation via a setting in [runtime configuration](https://dlthub.com/docs/running-in-production/running#set-the-log-level-and-format)
2. Instead of partial masking of the secrets where we leave first and last character, do the total masking as @rudolfix initially suggested.

zilto:
Status on this PR? The obfuscation feature could be valuable when passing the dlt config to LLMs via the MCP

PR #2031, opened 2024-11-06
[WIP] Setting to control delta job count for each delta write

sh-rp:
…ble write

### DIFF
--- a/dlt/common/data_writers/buffered.py
+++ b/dlt/common/data_writers/buffered.py
@@ -113,6 +113,7 @@ class BufferedDataWriter(Generic[TWriter]):
         # rotate the file if max_bytes exceeded
         if self._file:
             # rotate on max file size
+            print(self._file.tell())
             if self.file_max_bytes and self._file.tell() >= self.file_max_bytes:
                 self._rotate_file()
             # rotate on max items
--- a/dlt/destinations/impl/filesystem/configuration.py
+++ b/dlt/destinations/impl/filesystem/configuration.py
@@ -22,6 +22,8 @@ class FilesystemDestinationClientConfiguration(FilesystemConfiguration, Destinat
     )
     current_datetime: Optional[TCurrentDateTime] = None
     extra_placeholders: Optional[TExtraPlaceholders] = None
+    delta_jobs_per_write: Optional[int] = None
+    """how many jobs to write in a single commit for Delta tables"""
 
     @resolve_type("credentials")
     def resolve_credentials_type(self) -> Type[CredentialsConfiguration]:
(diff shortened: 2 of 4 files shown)


rudolfix (changes requested):
this is not solving a root problem of a large file begin loaded at once.

AFAIK delta-rs is accepting arrow batches and for sure it is accepting data frames,

IMO
- keep the original code with a dataset
- add a setting ie batch size to split arrow dataset into batches. we should possibly use multiples of row groups as when we batch in normalizer
- you feed batches into delta

this way we conserve memory. tbh. we were sure that by using dataset we allow to do efficient batching inside the library. it seems the opposite is true

sh-rp:
@rudolfix the underlying issue in the rust implementation is being worked on here: https://github.com/delta-io/delta-rs/pull/2289. I'm not sure we should fix this ourselves... My worry when we feed in batches is, that if there job somehow aborts for transient reasons, then the whole reference job will be marked as unloaded and all rows of all files are loaded again on the next try and you can end up with double entries. if we split this up into multiple jobs, this will not happen, or if it does, the amount of double rows is less.

jorritsandbrink:
Regarding the suggested solution:
- `replace` write disposition: will it work?
- `upsert` strategy: will it catch duplicate `primary_key` values if they're in different jobs?
- might be slower

My take:
- current code does the "right" thing
- the `delta-rs` bug is part of their [Rust v1.0.0 Milestone](https://github.com/delta-io/delta-rs/milestone/9) that is due by November 20, so hopefully will be solved soon (one of their main committers self-assigned the issue a couple of days ago)
- we should not complicate our code base to work around the bug in the meantime

As a workaround for now, users can prevent memory issues in the `load` step by yielding a smaller data set during `extract`?

sh-rp:
@Gilbert09 is loading the table in multiple loads and option until the rust implementation is fixed? I tend to agree with @jorritsandbrink that the rust implementation should behave properly.

Gilbert09:
> replace write disposition: will it work?

No, this doesn't work right now, I'm having to force users to use `merge` for these large tables until the Rust implementation is fixed. 

> might be slower

It's certainly slower - there's a non-nominal cost in overheads coming from _somewhere_ between each job (likely a delta-rs overhead)


In theory, this doesn't actually need to be merged. I've solved half the issue by monkey patching our DLT implementation (see https://github.com/PostHog/posthog/pull/26040) - the real solution here is the rust implementation being fixed which looks like has some progress now

rudolfix:
> @rudolfix the underlying issue in the rust implementation is being worked on here: [delta-io/delta-rs#2289](https://github.com/delta-io/delta-rs/pull/2289). I'm not sure we should fix this ourselves... My worry when we feed in batches is, that if there job somehow aborts for transient reasons, then the whole reference job will be marked as unloaded and all rows of all files are loaded again on the next try and you can end up with double entries. if we split this up into multiple jobs, this will not happen, or if it does, the amount of double rows is less.

right! we'll need to use transactions then possibly accumulating all data again in the memory... waiting for delta-rs fix is probably the best way. should we close this PR

sh-rp:
closing as won't fix (or rather will be fixed by delta rs hopefully)

PR #2033, opened 2024-11-07
Incremental table hints and incremental in resource decorator

### LINKED ISSUE #1974: add incremental value info to table computed in the DltResource
**Background**
We do not annotate table schemas with Incremental settings if present. This annotations may be useful downstream for systems that will automatically generate pipelines that incrementally append or merge data and are without access to original Python code.

**Tasks**

1. standardize column annotation for Incremental. We should be able to recreate Incremental instance from those settings. Look what we already do when defining incremental for `rest_api` - we may reuse the same TypedDict.
2. When computing schema, create a proper annotation on the table. we allow only one incremental, so we need to update the code that diffs and merge tables so only one incremental is allowed.
3. For incremental that do not refer to a single field (simple json path) we obviously do not generate annotations. 
4. now it is trivial to add `incremental` to `dlt.resource` decorator. accept both incremental instance and new typed dict. there's existing code that will use such incremental to override the one defined in resource function arguments.

**Tests**
1. on top of usual tests, test also the updated decorator:
- do the value in decorator override the default in function signature
- is it passed to an argument with Incremental type but without value?
- what if there's no incremental argument?

steinitzu:
* Extract incremental settings to a dict in table schema
* Support passing incremental settings to @resource decorator

### Description
* Add `incremental` hint to table schema, containing all incremental settings used in extract
* Accept `incremental` in `@resource` decorator. Either as dict (`IncrementalArgs` typeddict or incremental instance)
* Decorator incremental is injected to resource func if argument is in signature
* Explicit argument overrides decorator incremental

### Related Issues

- Resolves #1974

### DIFF
--- a/dlt/extract/decorators.py
+++ b/dlt/extract/decorators.py
@@ -70,7 +70,7 @@ from dlt.extract.source import (
     TSourceFunParams,
 )
 from dlt.extract.resource import DltResource, TUnboundDltResource, TDltResourceImpl
-from dlt.extract.incremental import TIncrementalConfig
+from dlt.extract.incremental import TIncrementalConfig, Incremental
 
 
 @configspec
@@ -627,6 +627,9 @@ def resource(
         TDltResourceImpl instance which may be loaded, iterated or combined with other resources into a pipeline.
     """
 
+    if incremental is not None:
+        incremental = Incremental.ensure_instance(incremental)
+
     def make_resource(_name: str, _section: str, _data: Any) -> TDltResourceImpl:
         table_template = make_hints(
             table_name,
@@ -691,7 +694,7 @@ def resource(
         return _wrap
 
     def decorator(
-        f: Callable[TResourceFunParams, Any]
+        f: Callable[TResourceFunParams, Any],
     ) -> Callable[TResourceFunParams, TDltResourceImpl]:
         if not callable(f):
             if data_from:
@@ -1033,7 +1036,7 @@ TDeferredFunParams = ParamSpec("TDeferredFunParams")
 
 
 def defer(
-    f: Callable[TDeferredFunParams, TBoundItems]
+    f: Callable[TDeferredFunParams, TBoundItems],
 ) -> Callable[TDeferredFunParams, TDeferred[TBoundItems]]:
     @wraps(f)
     def _wrap(*args: Any, **kwargs: Any) -> TDeferred[TBoundItems]:
--- a/dlt/extract/decorators.py
+++ b/dlt/extract/decorators.py
@@ -655,6 +655,7 @@ def resource(
         )
 
         if incremental:
+            # Reset the flag to allow overriding by incremental argument
             resource.incremental._from_hints = False
         # If custom nesting level was specified then
         # we need to add it to table hints so that
--- a/dlt/extract/decorators.py
+++ b/dlt/extract/decorators.py
@@ -32,7 +32,6 @@ from dlt.common.reflection.spec import spec_from_signature
 from dlt.common.schema.utils import DEFAULT_WRITE_DISPOSITION
 from dlt.common.schema.schema import Schema
 from dlt.common.schema.typing import (
-    TColumnNames,
     TFileFormat,
     TWriteDisposition,
     TWriteDispositionConfig,
@@ -43,7 +42,8 @@ from dlt.common.schema.typing import (
 )
 from dlt.common.storages.exceptions import SchemaNotFoundError
 from dlt.common.storages.schema_storage import SchemaStorage
-from dlt.common.typing import AnyFun, ParamSpec, Concatenate, TDataItem, TDataItems
+from dlt.common.typing import AnyFun, ParamSpec, Concatenate, TDataItem, TDataItems, TColumnNames
+
 from dlt.common.utils import get_callable_name, get_module_name, is_inner_callable
 
 from dlt.extract.hints import make_hints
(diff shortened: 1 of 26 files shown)


dlt/extract/decorators.py
```diff
@@ -642,7 +648,9 @@ def make_resource(_name: str, _section: str, _data: Any) -> TDltResourceImpl:
             selected,
             cast(DltResource, data_from),
             True,
+            incremental=incremental,
```
rudolfix, line 651: IMO we should extend `make_hints` to accept incremental config. then we'll be able to use existing mechanism that injects incremental into functions

dlt/extract/incremental/__init__.py
```diff
@@ -578,6 +609,34 @@ def get_incremental_arg(sig: inspect.Signature) -> Optional[inspect.Parameter]:
                 break
         return incremental_param
 
+    @staticmethod
+    def inject_implicit_incremental_arg(
```
rudolfix, line 613: Why we need an explicit method to inject Incremental passed from the decorator? we already have one that is used when `apply_hints` is used. (via `_set_hints` and later `bind` in the incremental). is there a problem with the precedence ie. to make sure that default value had it? IMO we can ignore such things and rely on the precedence from `apply_hints`

dlt/extract/resource.py
```diff
@@ -110,6 +112,9 @@ def __init__(
         self._explicit_args: DictStrAny = None
         self.SPEC = SPEC
         self.source_name = None
+        self._initial_incremental: Optional[Incremental[Any]] = (
```
rudolfix, line 115: I really think we do not need it

dlt/extract/resource.py
```diff
@@ -122,6 +127,7 @@ def from_data(
         selected: bool = True,
         data_from: Union["DltResource", Pipe] = None,
         inject_config: bool = False,
+        incremental: Optional[TIncrementalConfig] = None,
```
rudolfix, line 130: pass it in `hints`

dlt/extract/resource.py
```diff
@@ -442,35 +449,51 @@ def add_step(
...
 
+    def set_incremental(
+        self, new_incremental: Incremental[Any], from_hints: bool = False
+    ) -> Optional[Incremental[Any]]:
+        """Set/replace the incremental transform for the resource.
+
+        Args:
+            new_incremental: The Incremental instance/hint to set or replace
+            from_hints: If the incremental is set from hints. Defaults to False.
+            fixed: The Incremental is fixed on the resource (I.e. set from `@resource` decorator)
```
rudolfix, line 460: this arg is not there

dlt/extract/resource.py
```diff
@@ -480,10 +503,39 @@ def _set_hints(
...
+        return table_schema
+
     def bind(self: TDltResourceImpl, *args: Any, **kwargs: Any) -> TDltResourceImpl:
         """Binds the parametrized resource to passed arguments. Modifies resource pipe in place. Does not evaluate generators or iterators."""
         if self._args_bound:
             raise TypeError(f"Parametrized resource {self.name} is not callable")
+
+        incremental = self.incremental
+
+        if self._initial_incremental and (
```
rudolfix, line 524: why we need to do it here? isn't it enough if we keep Incremental in thints (via update_hints?)

rudolfix (changes requested):
@steinitzu pls take a look at:
- IMO it is way better if we set Incremental on the column level in Schema and ignore all non-fitting cases
- I think the code that injects Incremental should use existing mechanism (apply_hints) OR I do not understand some edge cases you had

dlt/extract/incremental/__init__.py
steinitzu: Yeah might be getting into overly complicated workarounds here. I'll see about simplifying this.  
I got into issues where it's not possible to override the initial hint with an argument cause the hint always takes precedence. There were some cases as well where the hint does not get injected into the default incremental arg, it was lost when resource is cloned I think.

dlt/extract/hints.py
```diff
@@ -506,6 +513,22 @@ def _merge_merge_disposition_dict(dict_: Dict[str, Any]) -> None:
                 "row_key": False,
             }
 
+    @staticmethod
+    def _merge_incremental_column_hint(dict_: Dict[str, Any]) -> None:
+        incremental = dict_.pop("incremental")
+        if isinstance(incremental, Incremental):
+            incremental_hint = incremental.to_table_hint()
```
rudolfix, line 520: the ticket asked to put the full incremental info into table now when I think of it - it will be confusing. it describes the behavior of the data source, not the data in the actual table. my take is to just set incremental to `True` if it is present on a column,

it is sufficient for what we plan upstream. we could add row order later if we really need it (using `x-` annotation)

I do not thing this impacts this PR too much

dlt/extract/incremental/__init__.py
```diff
@@ -166,6 +178,17 @@ def __init__(
         self._bound_pipe: SupportsPipe = None
         """Bound pipe"""
 
+    def to_table_hint(self) -> Optional[IncrementalArgs]:
+        """Table hint is only returned when all properties are serializable"""
+        if self.last_value_func not in (min, max):
```
rudolfix, line 183: we we do not need to care. we just return True if we can localize incremental to a column

rudolfix (changes requested):
1. let's save just `incremental`: True into the schema. see my comment why
2. I added a test that makes sure we can add incremental during extraction (but limited...)
3. we should also add "primary_key" to IncrementalArgs. mind that empty key ([] or ()) is also a valid hint. (this key is used for boundary deduplication)

tests:
Do we detect nested incremental columns ie $.item.id will map into "item__id"? if you allow for that then please test it

rudolfix (approved):
LGTM!

PR #2046, opened 2024-11-10
leverage ibis expression for getting readablerelations

sh-rp:
### Description
Implements support for using ibis expressions to generate sql statements for relations. To this end this PR implements a new type of readable relation which gets used that wraps an ibis unboundtable expression, but still accesses data the old way.

Todos:
* Figure out typing
* Ensure all ibis features documented in https://ibis-project.org/tutorials/ibis-for-sql-users work.


sh-rp:
Another note to self, we probably need to run columns names through the normalizer. Or we assume the user will use normalized names as they are present in the schema when building these expressions.

tests/load/test_read_interfaces.py
```diff
@@ -467,3 +467,49 @@ def test_standalone_dataset(populated_pipeline: Pipeline) -> None:
...
+    total_records = _total_records(populated_pipeline)
+
+    items_table = dataset.table("items")
+    double_items_table = dataset.table("double_items")
+
+    # check full table access
+    df = items_table.df()
+    assert len(df.index) == total_records
+
+    df = double_items_table.df()
```
sh-rp, line 492: regular dlt dataset execution methods (df, arrow, iter_arrow...) work everywhere

rudolfix (commented):
this is really cool! IMO we should keep ibis as optional dependency (it only works with python 3.10+). so we have two options:
- separate relation
- enable the proxy behavior if ibis is found, if not we fallback to the current behavior

i'd probably go for the second option. I'm just a little bit worried about the typing

in both cases we should implement a few common expressions we already have in our existing relation (limit, head, column selection).

regarding the schema: column lineage you can do with sqlglot. it makes sense to invest a little bit of time to understand how it is done:
- https://github.com/tobymao/sqlglot/blob/main/posts/ast_primer.md (btw. it seems that we are not finding tables in expression in a correct way)
- https://sqlglot.com/sqlglot/lineage.html

we can add sqlglot as a regular dependency. and use it everywhere we have sql SELECT statements.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -0,0 +1,161 @@
...
+        """limit the result to 'limit' items"""
+        return self._proxy_expression_method("limit", limit)  # type: ignore
+
+    def head(self, limit: int = 5) -> "ReadableIbisRelation":
+        """limit the result to 5 items by default"""
+        return self._proxy_expression_method("head", limit)  # type: ignore
+
+    def select(self, *columns: str) -> "ReadableIbisRelation":
+        """set which columns will be selected"""
+        return self._proxy_expression_method("select", *columns)  # type: ignore
```
sh-rp, line 161: all the other ibis methods are not defined here yet. we'd need to add them to the interface and raise an error if they are called when the regular relation is returned I think. wdyt?

sh-rp:
@rudolfix the only thing still missing here is proper typing for this ibis wrapper. The "brute force" way of doing it would be to define all the methods of the ibis expression we can make use of on the SupportsReadableRelation Interface, forward those calls in the ibirelation like we do with limit etc, and raise on the regular relation. Alternatively I could do some kind of wildcard typing to make the linter shut up, but it would be less strict. Maybe you can give me your opinion, if you don't mind, then I'll just decide on one.

dlt/destinations/dataset/factory.py
```diff
@@ -0,0 +1,22 @@
...
+    TDatasetType,
+    TDestinationReferenceArg,
+)
+
+from dlt.common.schema import Schema
+
+from dlt.destinations.dataset.dataset import ReadableDBAPIDataset
+
+
+def dataset(
```
rudolfix, line 16: do you want to improve typing in this PR? if so we could
1. make `SupportsReadableDataset` generic where it is parametrized by relation type (bound to SupportsReadableRelation)
2. use overload on the literal `TDatasetType` when in auto/ibis you return SupportsReadableDataset parametrized by ReadableIbisRelation, otherwise by just SupportsReadableRelation
3. we can make (2) dynamic dependent if ibis is present

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -0,0 +1,161 @@
...
+        if attr is None:
+            raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
+
+        if not callable(attr):
+            return attr
+        return partial(self._proxy_expression_method, name)
+
+    def __getitem__(self, columns: Union[str, Sequence[str]]) -> "ReadableIbisRelation":
+        # casefold column-names
+        columns = [self.sql_client.capabilities.casefold_identifier(col) for col in columns]
```
rudolfix, line 146: this is column getter right? so you will receive ibis `Column` object(s) right? then maybe select them from the schema like we do in regular relation?

this will cover many cases already at make ibis backward compatible with base expression

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -0,0 +1,161 @@
...
+        return self.compute_columns_schema()
+
+    @columns_schema.setter
+    def columns_schema(self, new_value: TTableSchemaColumns) -> None:
+        raise NotImplementedError("columns schema in ReadableDBAPIRelation can only be computed")
+
+    def compute_columns_schema(self) -> TTableSchemaColumns:
+        """provide schema columns for the cursor, may be filtered by selected columns"""
+        # TODO: provide column lineage tracing with sqlglot lineage
+        return None
```
rudolfix, line 92: maybe we could try some simple implementation:
1. we preserve table schema until first expression is made
2. we preserve table schema for `head` `limit` and `select` method
3. maybe we also recognize a few exceptions that do not modify schema ie `filter` and free form filter expressions as here: https://ibis-project.org/tutorials/ibis-for-pandas-users#filtering-rows

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # try casefolded name for ibis columns access
+        if attr is None:
+            name = self.sql_client.capabilities.casefold_identifier(name)
+            attr = getattr(self._expression, name, None)
+
+        if attr is None:
+            raise AttributeError(f"'{self.__class__.__name__}' object has no attribute '{name}'")
+
+        if not callable(attr):
```
rudolfix, line 140: I think ibis supports a column getter with a dot notation. my take would be to 
- start supporting it upstream
- modify the schema accordingly

tests/load/test_read_interfaces.py
```diff
@@ -51,6 +52,14 @@ def _expected_chunk_count(p: Pipeline) -> List[int]:
 @pytest.fixture(scope="session")
 def populated_pipeline(request) -> Any:
     """fixture that returns a pipeline object populated with the example data"""
+
+    # ensure ibis is installed for these tests
+    import subprocess
+
+    subprocess.check_call(
```
rudolfix, line 59: this is hardcore. arent we able to use dependency group for ibis?

tests/load/test_read_interfaces.py
```diff
@@ -472,6 +436,138 @@ def test_schema_arg(populated_pipeline: Pipeline) -> None:
...
+
+    df = double_items_table.df()
+    assert len(df.index) == total_records
+
+    # check limit
+    df = items_table.limit(5).df()
+    assert len(df.index) == 5
+
+    # check chained expression with join, column selection, order by and limit
+    joined_table = (
```
rudolfix, line 469: you do not need to test all the expressions but you should test various expression forms:
1. those that return Table (like here)
2. those that select a column
3. a single column getter with . notation
4. a free form filters:
```py
expr = penguins.bill_length_mm > 37.0
```
5. materializations like their dataframe and arrow getters
6. there are also expressions that add and remove columns to schema
7. expressions that return Expr (not Table - those do some weird things, I'm not sure SQL can be generated for them)?

rudolfix (changes requested):
very cool! tldr;>
- we can track schema changes in 90% of cases easily (or at least as good as in upstream object)
- we can do better with typing but maybe in separate PR
- tests need to be better
- can we do without dynamic installation for ibis?

dlt/destinations/dataset/ibis_relation.py
sh-rp: I'm not sure what you mean, the column getter dot notation is already working, see the tests, I'm using it there all the time for filtering and joining etc.

tests/load/test_read_interfaces.py
sh-rp: I have added quite a few tests from the "ibis for sql users" page in the ibis docs. changed a few things in the ibisrelation implementation to make this work. I think most cases should actually work now. Let me know if you think something is missing there or you have an idea what to test next.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -0,0 +1,161 @@
...
+    "tsql": "postgres",
+    "databricks": "postgres",
+    "clickhouse": "postgres",
+    "redshift": "postgres",
+    "presto": "postgres",
+}
+
+
+# TODO: provide ibis expression typing for the readable relation
+class ReadableIbisRelation(BaseReadableDBAPIRelation):
```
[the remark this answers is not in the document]
sh-rp: No, there is nothing to mix in, I had already looked at many parts of the ibis definition. For now I have added a brute force typing so the warnings go away...

dlt/destinations/dataset/factory.py
sh-rp: Can you maybe explain how (3) works? I don't understand how mypy can lint something else if a certain library is installed..

dlt/destinations/dataset/ibis_relation.py
sh-rp: What do you mean with "free form filter"?

tests/load/test_read_interfaces.py
sh-rp: I basically want to avoid running some subset of tests only with ibis, but I'll just add a group for ibis and make it available for all destination tests.

tests/load/test_read_interfaces.py
sh-rp: I ammended it a bit more, I think it should be quite good now.

tests/load/test_read_interfaces.py
sh-rp: done

dlt/destinations/dataset/factory.py
sh-rp: Anyway as mentioned above, I have now brute forced typing, but i'm happy to improve this if I understand how to do (3)

dlt/common/libs/ibis.py
```diff
@@ -119,3 +137,37 @@ def create_ibis_backend(
         con = ibis.duckdb.from_connection(duck)
 
     return con
+
+
+def create_unbound_ibis_table(
```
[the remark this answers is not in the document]
sh-rp: I did that (see the commit) but there is one place in common where the IbisBackend typing is imported for the readable relation.

dlt/destinations/dataset/ibis_relation.py
sh-rp: order_by and limit also are included, I have done this now.

rudolfix (dismissed):
LGTM! we'll solve ibis typing somehow later

PR #2052, opened 2024-11-11
disable sftp + delta test

sh-rp:
### Description
The lancedb PR last week broke our CI. Fixes done here are:
* Make sure delta tests are not run on the sftp bucket
* revert a change to sequence_generator in our test utils which was breaking weaviate and qdrant tests
* Remove pyarrow 17 workflow as 17 is the default now
* Run common tests with pyarrow 15 and run the libs tests again with pyarrow 17
* Remove pyarrow17 test markers in some cases.


.github/workflows/test_common.yml
```diff
@@ -118,6 +118,9 @@ jobs:
       - name: Install pyarrow
         run: poetry install --no-interaction -E duckdb -E cli -E parquet --with sentry-sdk
 
+      - name: Ensure pyarrow 15
+        run: poetry run pip install pyarrow==15.0.2
+
```
rudolfix, line 123: OK! just make sure that this line:
```
poetry install --no-interaction -E duckdb -E cli -E parquet -E deltalake -E sql_database --with sentry-sdk,pipeline,sources
```
is not installing arrow 17 again. I see delta  is being installed so probably that happens... maybe we should split testing of the `lib` to a separate step where we install delta

rudolfix (changes requested):
I'm pretty sure we reinstall arrow 17 too early. pls see my comment

.github/workflows/test_common.yml
sh-rp: yeah good point. I think every call to poetry install has the potential to change this. I have updated the workflow not fix this, but I am not quite sure how you would set it up tbh.

rudolfix (approved):

PR #2066, opened 2024-11-14
edit snippet to make more runnable

kning:
### Description
Tried running the example as is and ran into a few issues:
* dlt[parquet] missing dependency
* missing secret (removed since it's not used in this example)
* removed serialized flag (this is kind of an advanced feature and not really needed here)
* added minimal reflection level, otherwise if you run this script twice in a row you get a "column constraint not supported" kind of error

Another thing I noticed is that the way the snippets are used in the example excludes information that prevents the code from being truly runnable e.g. `import modal` statement
image

At a high level, if this is my first time coming to this page, I'd expect to be able to copy / paste the code as is here and run it right away and given how the snippets are set up, that's not the case and I'm not sure how to make it so. Open to ideas!

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/docs/website/docs/walkthroughs/deploy-a-pipeline/deploy_snippets/deploy-with-modal-snippets.py
+++ b/docs/website/docs/walkthroughs/deploy-a-pipeline/deploy_snippets/deploy-with-modal-snippets.py
@@ -12,6 +12,7 @@ def test_modal_snippet() -> None:
         "dlt>=1.1.0",
         "dlt[duckdb]",  # destination
         "dlt[sql_database]",  # source (MySQL)
+        "dlt[parquet]",  # file format dependency 
         "pymysql",  # database driver for MySQL source
     )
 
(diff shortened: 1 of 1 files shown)


AstrakhantsevaAA:
@kning absolutely fair comments!

>removed serialized flag (this is kind of an advanced feature and not really needed here)

I had to add it to run tests in the snippet, otherwise modal doesn't allow to set @app.function not in the global namespace.

>import modal statement

You can modify snippet code in [there](https://github.com/dlt-hub/dlt/blob/e086bbcdefe3d832483d81f5e13e2e8e63f7b492/docs/website/docs/walkthroughs/deploy-a-pipeline/deploy_snippets/deploy-with-modal-snippets.py#L3)

kning:
added changes, is there a good way to see the preview? also what exactly happens in ci, does it run this code at all?

AstrakhantsevaAA:
>what exactly happens in ci

@kning also trying figure this out, let's create a separate branch and merge your changes there and run CI/CD

AstrakhantsevaAA:
@kning https://github.com/dlt-hub/dlt/pull/2079

PR #2067, opened 2024-11-15
`iceberg` table format support for `filesystem` destination

### LINKED ISSUE #1996: write iceberg tables on filesystem destination
**Background**
We aim to support backend and server-less write support for iceberg tables. We'd like to do that in similar way we do it to delta-tables: make `table_format` `iceberg` to be recognized by the filesystem destination. From the user PoV this means:
- writing and reading iceberg tables without query engine as a separate backend
- maintaining and evolving the schema without catalog as a separate backend

We want to use pyiceberg. This limits the write disposition to append and replace (until upsert is implemented). We also wont' support vacuum, compact or z-order ops on the tables.

**Tasks**
1. * [ ] we maintain a "technical" catalog: sqllite file per table. those files we store together with the data
2. * [ ] to write a table we lock the sqllite file with `TransactionalFile`, pull it locally, use with pyiceberg and then write it back.
3. * [ ] use `pyiceberg` to append, replace tables, create partitions, do schema evolution etc.
4. * [ ] support all buckets via fsspec
5. * [ ] like for delta, expose pyiceberg for a given table. read only (catalog without lock) and r/w with lock on catalog (maybe via context manager). this will allow people ie. to delete or rebuild partitions on a table.
6. * [ ] support `filesystem` `sql_client` to create views on ICEBERG via duckdb

jorritsandbrink:
closes #1996


pyproject.toml
```diff
@@ -134,7 +141,7 @@ sqlfluff = "^2.3.2"
 types-deprecated = "^1.2.9.2"
 pytest-console-scripts = "^1.4.1"
 pytest = "^7.0.0"
-mypy = "^1.10.0"
+mypy = ">=1.11.0,<1.13.0"
```
jorritsandbrink, line 144: - `mypy` upgrade needed to solve https://github.com/apache/iceberg-python/issues/768
- has `<1.13.0` requirement because `1.13.0` gives another error (I did not investigate that error thoroughly, rather just added the `<1.13.0` to unblock myself)
- I simply `ignore`d new lint errors arising due to version upgrade

dlt/common/configuration/specs/aws_credentials.py
```diff
@@ -77,6 +78,16 @@ def to_object_store_rs_credentials(self) -> Dict[str, str]:
...
         return creds
 
+    def to_pyiceberg_fileio_config(self) -> Dict[str, Any]:
+        return {
+            "s3.access-key-id": self.aws_access_key_id,
+            "s3.secret-access-key": self.aws_secret_access_key,
+            "s3.session-token": self.aws_session_token,
+            "s3.region": self.region_name,
+            "s3.endpoint": self.endpoint_url,
+            "s3.connect-timeout": 300,
```
jorritsandbrink, line 88: Perhaps timeout should be configurable.

dlt/common/libs/pyiceberg.py
```diff
@@ -0,0 +1,145 @@
...
+        catalog.create_table(
+            table_id,
+            schema=ensure_iceberg_compatible_arrow_schema(schema),
+            location=client.make_remote_url(table_path),
+        )
+
+    return catalog
+
+
+def get_iceberg_tables(
```
jorritsandbrink, line 97: Perhaps this function should not exist. Rather we could use the dlt dataset API. Same for `get_delta_tables`.

jorritsandbrink:
@rudolfix / @sh-rp this PR isn't there yet but can you give an intermediate review?

What still needs to happen:
- Azure/GCS support
- partitioning
- add `iceberg` to more of the existing `delta` tests
- probably some other things

Notes:
- I suggest we try adding partitioning in a follow-up PR. I looked at it but couldn't easily implement it:
    1. confusing API
    2. lack of documentation
    3. incorrect documentation
    4. bugs (or I am doing something wrong, which is not unlikely given 1/2/3)
- I faced some issues with dependency management; see `pyproject.toml`.

dlt/common/configuration/specs/aws_credentials.py
rudolfix: there's no way to pass fsspec filesystem instance, right?

rudolfix: ok location for tables is always `str` so the answer is no

dlt/common/libs/pyiceberg.py
```diff
@@ -0,0 +1,150 @@
...
+    if write_disposition == "append":
+        table.append(ensure_iceberg_compatible_arrow_data(data))
+    elif write_disposition == "replace":
+        table.overwrite(ensure_iceberg_compatible_arrow_data(data))
+
+
+def get_catalog(
+    client: FilesystemClient,
+    table_name: str,
+    schema: pa.Schema = None,
```
rudolfix, line 61: IMO the namespace corresponds to our dataset right? I'd allow to pass it here instead of hardcoding `DLT_ICEBERG_NAMESPACE`.

it looks like iceberg catalog has only two levels: namespaces and tables.

dlt/common/libs/pyiceberg.py
```diff
@@ -0,0 +1,150 @@
...
+        uri="sqlite:///:memory:",
+        **_get_fileio_config(client.config.credentials),
+    )
+    catalog.create_namespace(DLT_ICEBERG_NAMESPACE)
+
+    # add table to catalog
+    table_id = f"{DLT_ICEBERG_NAMESPACE}.{table_name}"
+    table_path = f"{client.dataset_path}/{table_name}"
+    metadata_path = f"{table_path}/metadata"
+    if client.fs_client.exists(metadata_path):
```
rudolfix, line 78: is it possible to split the function that returns a catalog (even per table so a new catalog is simply empty) and function that does table maintenance (evolves schema or creates empty table)

IMO the code structure would be
```py
with get_catalog() as catalog:
   create_or_evolve_table()
   write_data()
```

rudolfix: where catalog is abstract catalog class of pyiceberg

dlt/common/libs/pyiceberg.py
rudolfix: LGTM for now. we may accept `Dataset` instance instead of Pipeline later

rudolfix (commented):
IMO this is almost good.
- we need to decide if we want to save per table catalog.
- I'd enable iceberg scanner in duckdb/sql_client in filesystem. mostly to start running the same tests that use delta
- maybe a refactor of `get_catalog` I mentioned.

tests/load/pipeline/test_filesystem_pipeline.py
```diff
@@ -724,13 +769,20 @@ def delta_table(data):
...
-    expected_schema = ensure_delta_compatible_arrow_data(arrow_table).schema
-    assert actual.schema.equals(expected_schema)
-    expected_num_rows = 3 if write_disposition == "append" else 2
+    expected, actual = get_expected_actual(
+        "evolving_table", destination_config.table_format, arrow_table
+    )
+    assert actual.schema.equals(expected.schema)
+    if write_disposition == "append":
+        expected_num_rows = 3
+    elif write_disposition == "replace" and destination_config.table_format == "delta":
```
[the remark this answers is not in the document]
jorritsandbrink: okay, I created a bug [ticket](https://github.com/dlt-hub/dlt/issues/2092) for `delta`

dlt/common/configuration/specs/aws_credentials.py
jorritsandbrink: Right, I don't think you can pass an `fsspec` instance. `pyiceberg` uses two different filesystem backends (`fsspec` and `pyarrow`) and it seems they chose to have their own `FileIO` [config](https://py.iceberg.apache.org/configuration/#fileio) system on top of those.

tests/load/filesystem/test_credentials_mixins.py
```diff
@@ -170,15 +216,22 @@ def test_aws_object_store_rs_credentials(driver: str, fs_creds: Dict[str, Any])
...
 @pytest.mark.parametrize(
     "driver", [driver for driver in ALL_FILESYSTEM_DRIVERS if driver in ("gs")]
 )
-def test_gcp_object_store_rs_credentials(driver, fs_creds: Dict[str, Any]) -> None:
+@pytest.mark.parametrize("mixin", ALL_CREDENTIALS_MIXINS)
+def test_gcp_credentials_mixins(
+    driver, fs_creds: Dict[str, Any], mixin: Type[TCredentialsMixin]
+) -> None:
+    if mixin == WithPyicebergConfig:
+        pytest.skip("`WithPyicebergConfig` mixin currently not implemented for GCP.")
```
jorritsandbrink, line 233: Seems like `pyiceberg` only accepts OAuth authentication for GCP. I tried it with the OAuth client used for BigQuery tests (the one in `[destination.bigquery.credentials]`), but got a permissions error. Perhaps that client doesn't have read permissions on `gs://ci-test-bucket`. If we have (or create) an Oauth client with correct permissions, I can test it again and merge the code for GCP.

jorritsandbrink: Changed scopes on OAuth client. Works now.

rudolfix (changes requested):
code is good and ready to merge. we also have sufficient tests. we still need a little bit better docs.
1. `pyiceberg.io:__init__.py:348 Defaulting to PyArrow FileIO` maybe we could hide this log? we set log levels during tests in `pytest_configure`
2. ~there are some tests that do not work in https://github.com/dlt-hub/dlt/actions/runs/12096164907/job/33730150792?pr=2067#step:8:4106 pls take a look~ (they are coming from somewhere else)
3. we need help with improving the docs. my take would be to create a separate destination Delta / Iceberg where we could move most of the docs out of the filesystem WDYT?
- for both table formats you could write shortly how they are implemented
- in case of iceberg we also should describe how we write tables without catalog and mention the limitations (single writer!)
- we have a page where we enumerate table formats. I'll write some kind of introduction there next week

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -258,6 +258,13 @@ def create_views_for_tables(self, tables: Dict[str, str]) -> None:
             from_statement = ""
             if schema_table.get("table_format") == "delta":
                 from_statement = f"delta_scan('{resolved_folder}')"
+            elif schema_table.get("table_format") == "iceberg":
+                from dlt.common.libs.pyiceberg import _get_last_metadata_file
+
+                self._setup_iceberg(self._conn)
```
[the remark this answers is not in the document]
jorritsandbrink: Both issues mentioned in `_setup_iceberg` are still open. I just tried with `duckdb=1.1.3`, and it fails without `_setup_iceberg`.

jorritsandbrink:
@rudolfix I've addressed your feedback. Please have a look.

Only thing I don't understand are these failing tests:

```
FAILED tests/load/filesystem/test_credentials_mixins.py::test_gcp_credentials_mixins[WithObjectStoreRsCredentials-gs] - dlt.common.configuration.exceptions.ConfigValueCannotBeCoercedException: Configured value for field scopes cannot be coerced into type <class 'list'>
FAILED tests/load/filesystem/test_credentials_mixins.py::test_gcp_credentials_mixins[WithPyicebergConfig-gs] - dlt.common.configuration.exceptions.ConfigValueCannotBeCoercedException: Configured value for field scopes cannot be coerced into type <class 'list'>
FAILED tests/load/filesystem/test_gcs_credentials.py::test_explicit_filesystem_credentials - dlt.common.configuration.exceptions.ConfigValueCannotBeCoercedException: Configured value for field scopes cannot be coerced into type <class 'list'>
```

I can't reproduce them, it works on my machine.

It's strange because, looking at the logs, it looks as if `deserialize_value` is called with `key = 'scopes', value = 'None', hint = <class 'list'>`. Logically, this cannot happen because that function is only called `if value is not None`.

```py
def _resolve_config_field(
   ...
        # if value is resolved, then deserialize and coerce it
        if value is not None:
            # do not deserialize explicit values
            if value is not explicit_value:
                value = deserialize_value(key, value, inner_hint)
```

I'm probably overlooking something. Do you know what's going on?

rudolfix:
@jorritsandbrink that got fixed on another branch yesterday. so we are good. for some reason docs linting is not passing. pls check it out

jorritsandbrink:
> @jorritsandbrink that got fixed on another branch yesterday. so we are good. for some reason docs linting is not passing. pls check it out

@rudolfix fixed it by deleting `chess_games` folder in S3. 

Error was:

```
<class 'dlt.common.schema.exceptions.SchemaEngineNoUpgradePathException'>
E           In schema: chess_com_source: No engine upgrade path in schema chess_com_source from 11 to 10, stopped at 11. You possibly tried to run an older dlt version against a destination you have previously loaded data to with a newer dlt version.
```

on `_sync_destination` in `docs/examples/partial_loading/test_partial_loading.py::test_partial_loading`.

Don't think that error came from this branch.

rudolfix:
@jorritsandbrink now it all LGTM! just here:
https://github.com/dlt-hub/dlt/actions/runs/12125961933/job/33807050441?pr=2067#step:12:9511

I think we are importing iceberg to aggressively into filesystem. we should allow it to work both without delta and iceberg. please check it out!

jorritsandbrink:
@rudolfix I don't think we're importing `pyiceberg` too aggressively. Those errors arose because `pyiceberg` was accidentally removed from `pyproject.toml` when merging the latest version of `devel`. It has already been fixed and the latest checks don't contain those errors.

Issue #997, opened 2024-02-24
[chores] split long running destination tests for nightly builds


rudolfix:
There are several tests that run very slowly. Most of them are merge and replace tests with various configurations. They are especially slow on lakehouses. What we need to do

1. identify long running tests and move them to separate workflows that we will run nightly
2. make sure that workflows triggered by PR run within 1-2 hours
3. make sure that (2) is more or less covered. it would be good to keep single comprehensive merge and replace tests in there

PR #2082, opened 2024-11-21
feat(rest_api): custom client for specific resources

joscha:
### Description

This adds the ability to pass a custom `ClientConfig` to an `EndpointResource`.
The passed `ClientConfig` object will be merged with the global one from  the containing `@dlt.resource`.

#### Example

```py
{
    "client": {
        "base_url": "https://api.mything.com",
        # Default auth is using Bearer authentication
        "auth": {
            "type": "bearer",
            "token": dlt.secrets["your_api_token"],
        }
    },
    "resources": [
        "resource-using-bearer-auth", # <-- for example this resource
        {
            "name": "my-resource-with-special-auth",
            "client": {
                # However this resource is using HTTP Basic auth instead.
                "auth": HttpBasicAuth("user", dlt.secrets["your_basic_auth_password"]) 
            },
            # ...
        }
    ]
    # ...
}
```

In the corresponding [Slack thread](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1732109635628149) a question came up whether a custom authenticator could be used. The answer is yes, it would be possible to write custom authenticator which introspects the resource requested (base on the URL for example), however the implementation of such an authenticator would be a lot more involved, less obvious than just merging a configuration object and it would also not have an answer for differing `base_url`s , `headers` or other client-specific parameters for example.

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### Additional Context

See originating [Slack thread](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1732109635628149).

The reason for this feature is that some API endpoints, even though available under the same base URL and returning the same entities, may have different ways of accessing them. For example the [Affinity V2 API](https://developer.affinity.co/) uses an OpenAPI-based access model with [Bearer authentication](https://developer.affinity.co/#section/Getting-Started/Authentication). This part of the API is available under `https://api.affinity.co/v2/`.
The [first version of the API](https://api-docs.affinity.co/) is available under `https://api.affinity.co/` but uses [HTTP Basic Auth](https://api-docs.affinity.co/#authentication).
Entities from both APIs are using the same entity IDs and are related, thus must be specified in the same `@dlt.resource` in order to define those relations.
This is currently (`1.4.0`) not possible, due to the differing (incompatible) authentication methods.


dlt/sources/rest_api/__init__.py
```diff
@@ -263,12 +263,17 @@ def create_resources(
             incremental_cursor_transform,
         ) = setup_incremental_object(request_params, endpoint_config.get("incremental"))
 
+        merged_client_config: ClientConfig = {
```
joscha, line 266: configs get merged here, with the keys defined in the endpoint taking preference over the global one.

dlt/sources/rest_api/typing.py
```diff
@@ -200,14 +200,18 @@ class OAuth2ClientCredentialsConfig(AuthTypeConfig, total=False):
...
+class BaseClientConfig(TypedDict, total=False):
+    base_url: Optional[str]
     headers: Optional[Dict[str, str]]
     auth: Optional[AuthConfig]
     paginator: Optional[PaginatorConfig]
     session: Optional[Session]
 
 
+class ClientConfig(BaseClientConfig, total=False):
+    base_url: str  # type: ignore[misc]
```
joscha, line 212: Typing overrides are currently not yet supported (see https://github.com/python/mypy/issues/7435) thus the ignore here.
For the optional dict in each endpoint we need to ensure that all keys are optional.

tests/sources/rest_api/configurations/source_configs.py
```diff
@@ -395,6 +395,19 @@ def repositories():
...
     },
+    {
+        "client": {
+            "base_url": "https://test",
+        },
+        "resources": [
+            {
+                "name": "test",
+                "client": {
+                    "auth": HttpBasicAuth("", "BASIC_AUTH_TOKEN"),
```
joscha, line 406: example of a valid config with an `auth` override, such as what is needed for solving the issue described in the pull request.

dlt/sources/rest_api/__init__.py
```diff
@@ -405,14 +410,22 @@ def paginate_dependent_resource(
 
 def _validate_config(config: RESTAPIConfig) -> None:
     c = deepcopy(config)
-    client_config = c.get("client")
+    _mask_client_config_auth(c.get("client"))
+    resources = c.get("resources")
```
francescomucio, line 414: if you do `c.get("resources", [])`, you can skip the `if resources:`

dlt/sources/rest_api/__init__.py
```diff
@@ -405,14 +410,22 @@ def paginate_dependent_resource(
...
+    if resources:
+        for resource in resources:
+            if isinstance(resource, str) or isinstance(resource, DltResource):
```
francescomucio, line 417: I would go for the opposite check (is instance of `EndpointResource`?), is there a reason for testing like this

joscha: Yes, it's a `TypedDict` and the check is not (yet) supported on it :-/

francescomucio: sure, you could test only for `dict`, but that should be enough. if's it's a `dict` (even if an incorrect one) the code will try to mask the client field. Then the real validation will happen.

It is not super precise (a comment to mention that the actual test should be for the proper type, but python...), but a bit simpler to read.

Happy to hear other opinions @burnash

dlt/sources/rest_api/__init__.py
```diff
@@ -405,14 +410,22 @@ def paginate_dependent_resource(
...
+                continue
```
joscha, line 418:
```suggestion
    resources = c.get("resources", [])
    for resource in resources:
        if isinstance(resource, str) or isinstance(resource, DltResource):
            continue
```

dlt/sources/rest_api/__init__.py
joscha: changed in 1bf1fe766ab071f6a5df70f9434e0ae2b8c984ba

joscha:
@willi-mueller @burnash would you mind taking a look and give me feedback on this, please?

burnash:
Hi @joscha, thanks you for the contribution. I think this is a great idea. As you may have noticed, `rest_api` already creates a `RESTClient` instance for each of the resources under the hood.

One thing that I would like to discuss is that the PR extends to the resource configuration with the `client` key which, in turn, creates an additional way to configure the `paginator`. This is already configured via `endpoint`:

```python
{
    "client": {
        "base_url": "https://api.mything.com",
        "auth": {
            # ...
        },
        "paginator": "json_link"
    },
    "resources": [
        "resource-using-bearer-auth",
        {
            "name": "my-resource-with-special-auth",
            "client": {
                "auth": HttpBasicAuth("user", dlt.secrets["your_basic_auth_password"]),
                "paginator": "json_link"  # <-- now it could be configured here as well
            },
            "endpoint": {
                "paginator": "json_link"
            }
        }
    ]
}
```

I think it would be better to keep the configuration in one place. To achieve this, I see two options:
1. Move the `paginator` configuration to the resources's `client` from the `endpoint` configuration.
2. Do not add a `client` configuration to resources, but extend the `endpoint` configuration to accept `auth`.

Although I think the first option is more consistent with the global `client` configuration, I would prefer the second option because:
1. It is shorter for cases where the user needs to configure pagination (which happens often).
2. It is backward-compatible.

Your example would then look like this:
```python
{
    "client": {
        "base_url": "https://api.mything.com",
        "auth": {
            "type": "bearer",
            "token": dlt.secrets["your_api_token"],
        }
    },
    "resources": [
        "resource-using-bearer-auth",
        {
            "name": "my-resource-with-special-auth",
            "endpoint": {
                "path": "my-resource-with-special-auth",
                "auth": HttpBasicAuth("user", dlt.secrets["your_basic_auth_password"]),
                "paginator": "json_link",
            }
        }
    ]
}
```

What do you think?

joscha:
Hi @burnash,

I initially had only `auth` as it was enough for what I wanted to do initially but when I opened the PR I noticed that this would not actually allow a completely reconfigured rest client. Some APIs when migrating between versions may choose to use different header parameters and/or base_urls so I thought overwriting/merging the whole client object would be best, as this would first of all work for all eventualities but also in concept be more simple than just add one separate object and then later another one, etc. whilst having to stay backwards compatible.
What do you think about removing the paginator as a possible property for overrides of the client parameters? Would that be an option? I do generally think your outlined 1. is cleaner long-term. If we were to accept both we could get the benefit of the short notation. Given the endpoint is more deeply nested I think the behavior would be that the endpoint paginator would override resource paginator would override global client paginator.
It is quite some baggage to carry though, for the convenience of omitting one extra `client {}`.

francescomucio:
For my experience, changing the paginator is something more common than changing the authentication method for different endpoints of the same API.

I am in favour of the first option, but I can see how it can cause some confusion. In my opinion we should make clear how things are handled, possible options are:

1. the client in the resource cannot contain a client.paginator, but only an endpoint.paginator
2. the client in the resource can contain a client.paginator, but the endpoint.paginator wins.

So it will be like

`client.paginator` < `default_resources.client.paginator` < `default_resource.endpoint.paginator` < `resource.client.paginator` < `resource.endpoint.paginator`

What do you think?

burnash:
Thank you @joscha and @francescomucio for your thoughtful input and discussion on this PR. After considering all the points raised, I wanted to share my perspective.

I'm not fully convinced that adding a client configuration directly to the resource is the optimal approach. In practice, setting the paginator explicitly per endpoint is much more common than changing the authentication method within the same API. Introducing a client config at the resource level may lead to unnecessary ambiguity.

Regarding the headers, I believe these should be specified within the endpoint configuration. In fact, there's already a PR adding headers at the endpoint level https://github.com/dlt-hub/dlt/pull/2084 

As for the `base_url`, I'm not sure it's necessary within the resource configuration since `endpoint.path` can accept an absolute URL (I'd need update the documentation to reflect this). Putting headers and auth into the endpoint configuration would make it consistent with how the underlying `RESTClient` and `RESTClient.paginate` operate.

Here's my proposal for moving forward. 
1. Extend the endpoint configuration to accept `auth` which overrides the client's auth configuration for the endpoint.
2. Add headers configuration to the endpoint via https://github.com/dlt-hub/dlt/pull/2084 

Here's how the configuration would look:

```python
{
    "client": {
        "base_url": "https://api.mything.com",
        "auth": {
            "type": "bearer",
            "token": dlt.secrets["your_api_token"],
        }
    },
    "resources": [
        "resource-using-bearer-auth",
        {
            "name": "my-resource-with-special-auth",
            "endpoint": {
                "path": "my-resource-with-special-auth",
                "auth": HttpBasicAuth("user", dlt.secrets["your_basic_auth_password"]),
                "paginator": "json_link",
                "headers": {"Custom-Header": "value"}
            }
        }
    ]
}
```

@joscha, would you be open to updating the PR to reflect this approach?

joscha:
I can most definitely update the PR that way if that's what you wish. I think possibly spreading the individual properties of `Client` into the endpoint config is shortsighted as it will lead to an explosion of highly specialized properties spread into that object of whenever another is missing to support the next person's use case.
But I haven't DLT long enough to say whether that is likely or not, so given that, unless I hear back from you otherwise, I will update the PR later to use auth and headers.
I will also update the docs with the mention of the absolute URL as it is a related issue, I was not aware this was accepted before you wrote it.
Thank you!

joscha:
Purely in regards to mechanics, do you:
* Want me to wait until the open headers PR is decided and merged
* Open the PR anyway and we deal with conflicts
* Want me to base my PR off the headers PR

?

burnash:
Thank you @joscha for your understanding and flexibility! Let's proceed with adding just the auth configuration to the endpoint config for now. Since #2084 is already handling the headers implementation. I will help with the managing the conflicts.

joscha:
changes in @burnash. Decided to do the mention of being able to use an absolute URL in a separate pull request.

joscha:
I just saw that `verified-sources` seems to have a copy (?) of the rest client - see
https://github.com/dlt-hub/verified-sources/blob/3d2993a685e8c1d713a2283ac4025ca3903ed392/sources/rest_api/__init__.py#L249

1. is that intentional? (my assumption is yes, although none of the other verified sources seem to use it) 
2. is there a process that syncs the two? If yes, where can I find it? if no, why not?
3. if 2.) is "no", should I make the same change in this repository as well?

docs/website/docs/dlt-ecosystem/verified-sources/rest_api/basic.md
```diff
@@ -308,6 +308,32 @@ A resource configuration is used to define a [dlt resource](../../../general-usa
 - `include_from_parent`: A list of fields from the parent resource to be included in the resource output. See the [resource relationships](#include-fields-from-the-parent-resource) section for more details.
 - `processing_steps`: A list of [processing steps](#processing-steps-filter-and-transform-data) to filter and transform the data.
 - `selected`: A flag to indicate if the resource is selected for loading. This could be useful when you want to load data only from child resources and not from the parent resource.
+- `auth`: An optional `AuthConfig` instance. If passed, is used over the one defined in the [dlt resource](../../../general-usage/resource.md) definition. Example:
```
burnash, line 311: I think it should be `client` not, resource, e.g.:
```suggestion
- `auth`: An optional `AuthConfig` instance. If passed, is used over the one defined in the [client](#client) definition. Example:
```

dlt/sources/rest_api/__init__.py
```diff
@@ -410,6 +410,16 @@ def _validate_config(config: RESTAPIConfig) -> None:
         auth = client_config.get("auth")
         if auth:
             auth = _mask_secrets(auth)
+    resources = c.get("resources", [])
+    for resource in resources:
+        if isinstance(resource, str) or isinstance(resource, DltResource):
```
burnash, line 415: I think this can be simplified to:
```suggestion
        if isinstance(resource, (str, DltResource)):
```

dlt/sources/rest_api/__init__.py
```diff
@@ -410,6 +410,16 @@ def _validate_config(config: RESTAPIConfig) -> None:
...
+            continue
+        if "endpoint" in resource:
+            endpoint = resource.get("endpoint")
```
burnash, line 418: And this to:

```suggestion
        if endpoint := resource.get("endpoint"):
```

dlt/sources/rest_api/__init__.py
```diff
@@ -410,6 +410,16 @@ def _validate_config(config: RESTAPIConfig) -> None:
         auth = client_config.get("auth")
         if auth:
             auth = _mask_secrets(auth)
```
burnash, line 412: Unrelated to this PR: there's not much point in this assignment, but I'll fix that in a follow up PR

burnash (changes requested):
Thanks you @joscha 
Look good, there are a couple of small improvements needed, please the comments.

burnash:
Very good question.

> I just saw that `verified-sources` seems to have a copy (?) of the rest client - see https://github.com/dlt-hub/verified-sources/blob/3d2993a685e8c1d713a2283ac4025ca3903ed392/sources/rest_api/__init__.py#L249
> 
> 1. is that intentional? (my assumption is yes, although none of the other verified sources seem to use it)

Yes, it's intentional, we moved  rest_api, sql_database and filesystem sources to dlt core in [1.0.0](https://github.com/dlt-hub/dlt/releases/tag/1.0.0). Versions prior to 1.0.0 still reference these sources as being in the verified-sources repository when you execute `dlt init`.

> 2. is there a process that syncs the two? If yes, where can I find it? if no, why not?

There isn’t a formal sync mechanism in place. If someone is working with projects that require or benefit from the latest updates or additional features available in the newer versions of these resources, I recommend upgrading to dlt version 1.0.0 or later.

> 3. if 2.) is "no", should I make the same change in this repository as well?

As this is a feature PR, for now, there's no need to port these changes to verified-sources repo. For bugfixes it's different.

tests/sources/rest_api/integration/test_response_actions.py
```diff
@@ -316,3 +317,40 @@ def add_field(response: Response, *args, **kwargs) -> Response:
     mock_response_hook_2.assert_called_once()
 
     assert all(record["custom_field"] == "foobar" for record in data)
+
+
+def test_auth_overwrites_for_specific_endpoints(mock_api_server, mocker):
+    def custom_hook(response: Response, *args, **kwargs) -> Response:
+        assert response.request.headers["Authorization"] == f"Basic {base64.b64encode(b'U:P').decode('ascii')}"
```
joscha, line 324:
```suggestion
        assert (
            response.request.headers["Authorization"]
            == f"Basic {base64.b64encode(b'U:P').decode('ascii')}"
        )
```

dlt/sources/rest_api/__init__.py
joscha: removed in e2f6c8cf

tests/sources/rest_api/configurations/source_configs.py
```diff
@@ -395,6 +395,18 @@ def repositories():
             repositories(),
         ],
     },
+    {
+        "client": {"base_url": "https://github.com/api/v2"},
+        "resources": [
+            {
+                "name": "issues",
+                "endpoint": {
+                    "path": "dlt-hub/{repository}/issues/",
```
burnash, line 404: Looks like this line is causing the tests to fail as `{repository}` path parameter is missing.
```suggestion
                    "path": "user/repos",
```

joscha: ah, thank you!

joscha:
🎉 all good to go now

burnash (approved):
> 🎉 all good to go now

Yes, looks great, thank you!

joscha:
Awesome, thank you. Are there instructions on how to use the current devel branch at a given sha? The only mention of nightly builds I could find is here: https://github.com/dlt-hub/dlt/issues/997

burnash:
You can use [git+ prefix in pip](https://pip.pypa.io/en/stable/topics/vcs-support/), for example:

Install a devel branch:
```
pip install "git+https://github.com/dlt-hub/dlt.git@devel"
```

Install a devel branch with duckdb extra
```
pip install "git+https://github.com/dlt-hub/dlt.git@devel#egg=dlt[duckdb]"
```

Replace `devel` with an sha to get a specific commit.

Issue #2071, opened 2024-11-18
Enhanced parameter resolution for RESTAPIConfig  and REST API source


ArneDePeuter:
### Feature description

Hi!

Currently, in DLT, you can define and resolve parameters dynamically like this:

```python3
"endpoint": {
    "path": "issues/{issue_number}/comments",
    "params": {
        "issue_number": {
            "type": "resolve",
            "resource": "issues",
            "field": "number",
        }
    },
},
```

While this works for some use cases, it's limited in flexibility. Specifically, it does not support dynamic or more generic scenarios, such as resolving parameters for headers or request bodies.

### Are you a dlt user?

I'd consider using dlt, but it's lacking a feature I need.

### Use case

I need to run data extraction from an API that requires dynamic headers, following these steps:

Authenticate using a login endpoint to retrieve a base token.
- Retrieve a list of items using the base token.
- For each item in the list:
   - Fetch a token specific to that item.
   - Make further requests where the item-specific token is included in the headers.

Example Workflow:
1. POST /authentication -> Returns a base token
2. GET /items -> Uses the base token, returns a list of items
3. For each item:
   - POST /items/{item_id}/token -> Returns an item-specific token
   - GET /items/{item_id}/data -> Uses the item-specific token in headers
   
While this may reflect poor API design, it is a real-world scenario that requires more flexible parameter resolution capabilities.

### Proposed solution

EG for dynamic headers:
```python3
"endpoint": {
    "path": "issues/comments",
    "headers": {"token": "{token}"}
    "params": {
        "token": {
            "type": "resolve",
            "resource": "authenticate",
            "field": "token",
        }
    },
},
# or even with static parameters
"endpoint": {
    "path": "issues/comments",
    "headers": {"token": "{token}", "static_param":  region}
    "params": {
        "token": {
            "type": "resolve",
            "resource": "authenticate",
            "field": "token",
        }
    },
},
```

### Related issues

This feature could also hypothetically be extended to support dynamic bodies in the future. While this isn't a current issue for me, it may benefit other use cases where request bodies need to include dynamically resolved fields.

burnash:
Thanks for the detailed feature request, @ArneDePeuter, and for laying out the use case so clearly! This is a really interesting suggestion. We'll review this further and evaluate how it might be incorporated into rest_api source.

ArneDePeuter:
Hi @burnash, the PR referenced satisfies the feature. Hope its useful!

Issue #1978, opened 2024-10-22
rest_api: Enable resolve in query parameters


jlopezpena:
### Feature description

For REST APIs, it should be possible to dynamically populate query parameters based on values obtained from an existing resource.

This can be currently done **only** if the resolved parameter is part of the path, but not if it is used inside the query.

Eg, this works in a source resource definition:

```python
{
    "name": "seasons",
    "endpoint": {
        "path": "competitions/{competition_id}/seasons",
        "params": {
            "competition_id": {
                "type": "resolve",
                "resource": "competitions",
                "field": "id",
            }
        }
    },
    "include_from_parent": ["id"],
}
```
but this doesn't:
```python
{
    "name": "seasons",
    "endpoint": {
        "path": "competitions/seasons/?competition_id={competition_id}",
        "params": {
            "competition_id": {
                "type": "resolve",
                "resource": "competitions",
                "field": "id",
            }
        }
    },
    "include_from_parent": ["id"],
}
```
This forces creating a resource or transformer outside of the declarative API, using `@dlt.transformer` or `@dlt.resource` decorators, and makes for overall messy code

### Are you a dlt user?

I am scoping it for our company's use case

### Use case

It is a very common pattern to find REST APIs where the id's of documents coming from a query are used in subsequent queries. Sometimes the API defines this as a path parameter, and sometimes as a query parameter. This is just an implementation detail that should be abstracted away in `dlt`.

### Proposed solution

The resolve mechanism should also work for query parameters. An ideal solution would work either with explicity inclusion of the query parameter in the path (as in the example above) or (even better!) in a declarative format, with something like this:
```python
{
    "name": "seasons",
    "endpoint": {
        "path": "competitions/seasons/",
        "params": {
            "competition_id": {
                "type": "resolve",
                "resource": "competitions",
                "field": "id",
                "location": "query"  # <- This field would ensure the resolved parameter gets injected in the query
            }
        }
    },
    "include_from_parent": ["id"],
}
```

### Related issues

None found

joscha:
> ### Use case

We have a similar use-case:

The [Affinity CRM](https://affinity.co/) has three main entities: Persons, Companies and Opportunities.
The CRM allows to attach notes to each of these entities. They are always connected to at least one to many of each type of these entities, but never to different types. The way the API models this is described [here](https://api-docs.affinity.co/#get-all-notes):
[...]

PR #2084, opened 2024-11-21
FEAT: parameterised headers rest_api_source

ArneDePeuter:
### Description
This pull request addresses the feature request discussed in issue [#2071](https://github.com/dlt-hub/dlt/issues/2071). It introduces dynamic parameter resolution for headers, allowing both keys and values to be resolved dynamically.

### Key Features
- Dynamic Header Resolution: Supports resolving both header keys and values at runtime using specified parameters.
- Header params can also be bound to provided params.
- You can still define headers in the client, these will always be used. Endpoint specific parameters extend this dict.

### Usage
The feature is demonstrated with the following example:
```
{
    "name": "chicken",
    "endpoint": {
        "path": "chicken",
        "headers": {"{token}": "{token}", "num": "2"},
        "params": {
            "token": {
                "type": "resolve",
                "field": "token",
                "resource": "authenticate",
            },
        },
    },
}
```
### Tests
Comprehensive tests have been added to validate the implementation and ensure its correctness.


ArneDePeuter:
Excuse me for the failed checks, I'll have a look at it!

ArneDePeuter:
All checks should pass now, also added an extra test!

burnash:
Hey @ArneDePeuter thanks for the PR and for adding tests! I'll review it soon.

tests/sources/rest_api/test_rest_api_source.py
```diff
@@ -149,3 +151,166 @@ def test_dependent_resource(destination_name: str, invocation_type: str) -> None
...
+        "client": {"base_url": base_url, "headers": {"foo": "bar"}},
+        "resources": [
+            {
+                "name": "chicken",
+                "endpoint": {
+                    "path": "chicken",
+                    "headers": {
+                        "{token}": "{token}",
+                        "num": "2",
+                        "nested": {"nested": "{token}", "{token}": "other"},
```
burnash, line 284: @ArneDePeuter could you elaborate on the use case the nested headers?

ArneDePeuter: I added this for the sake of supporting every use case. After some research tough... headers are mostly flat key value pairs. So I also now question wether this is needed. 😬

ArneDePeuter:
@burnash I removed the recursive parameter resolution for headers, as this was not useful.

ArneDePeuter:
I now have following question. In: `dlt/sources/rest_api/__init__.py:366` would it be useful that resolved params gets extended with the incremental params? This way incrementals can be in the url as in the headers

Implemented as this for example:
```python
# dlt/sources/rest_api/__init__.py:366 
if incremental_object:
  params = _set_incremental_params(
      params,
      incremental_object,
      incremental_param,
      incremental_cursor_transform,
  )
  resolved_params.extend(params)

for item in items:
  formatted_path, formatted_headers, parent_record = process_parent_data_item(
      path, item, resolved_params, include_from_parent, headers
  )
```

I would like to hear your opinion on this. 😀

burnash:
> @burnash I removed the recursive parameter resolution for headers, as this was not useful.

Thanks for the update, @ArneDePeuter

> I now have following question. In: dlt/sources/rest_api/__init__.py:366 would it be useful that resolved params gets extended with the incremental params? This way incrementals can be in the url as in the headers.

I don't see a need to implement incremental in headers. I haven't seen a use case for incremental parametes in headers yet. If such a case exists, I would suggest implementing a dlt source using `RESTClient` rather than the rest_api source.

> Secondly, I find it confusing that parameter definitions are intertwined with query params. Is this something that has been brought up before?

Good point. I have already received this feedback from someone. The way the rest_api source defines parameters is loosely based on how parameters are defined in the OpenAPI specification. I don't find the current implementation too confusing, as the end user has control over how path parameters are named and resolved. They are also excluded from the query parameters. However, I'm open to suggestions on how to improve this.

If we extend `params` to include header params (as in this PR), it might become confusing. I was thinking about different ways to address this. I was considering different ways to address this. One option is to take inspiration from the suggested configuration for query string parameters in https://github.com/dlt-hub/dlt/issues/1978 and define the location of the parameter in a `location` key. For example, from your scenario:

```
"endpoint": {
    "path": "issues/comments",
    "headers": {"token": "{token}", "static_param":  region}
    "params": {
        "token": {
            "type": "resolve",
            "resource": "authenticate",
            "field": "token",
            "location": "header"
        }
    },
},
```

What do you think? Would you be willing to update this PR to handle the location key?

ArneDePeuter:
I added the location field. Although the current code is not super clean and it needs some more attention. 
The problem is that TypedDicts dont support default values. 
(Personally I think that Pydantic would be better for this use case (data validation) than TypedDicts and dataclasses.)

ArneDePeuter:
@burnash I’ve added the `location` parameter. By default, its value is set to `'path'`, ensuring backward compatibility. Hope you like it! The only supported locations currently are `'path'` and `'header'`.

burnash:
> Personally I think that this feature adds redundant information, I liked the fact that a param was not bound to a specific location. Params are unique by name so if users would want to make it location based they could use name conventions such as `'header_<paraname>'`. I feel like this location field adds a layer of complexity, I would like to hear your opinion on this.


@ArneDePeuter, sorry for the delayed response. I thought it through and I agree with your point: the location field does not really fix the problem while adding complexity. However, I still don't think putting everything in params is the ideal solution either.
I have a different idea on how to fix this, here's the proposal https://github.com/dlt-hub/dlt/issues/2190
It should make the interpolation consistent across path, query and header params. 
Let me know what you think! Sorry for giving you a wrong lead in the first place with the location field.
Once the proposal is finalized, I'll implement it first. After that, we'll see what is needed in this PR to merge it.

ArneDePeuter:
> > Personally I think that this feature adds redundant information, I liked the fact that a param was not bound to a specific location. Params are unique by name so if users would want to make it location based they could use name conventions such as `'header_<paraname>'`. I feel like this location field adds a layer of complexity, I would like to hear your opinion on this.
> 
> @ArneDePeuter, sorry for the delayed response. I thought it through and I agree with your point: the location field does not really fix the problem while adding complexity. However, I still don't think putting everything in params is the ideal solution either. I have a different idea on how to fix this, here's the proposal #2190 It should make the interpolation consistent across path, query and header params. Let me know what you think! Sorry for giving you a wrong lead in the first place with the location field. Once the proposal is finalized, I'll implement it first. After that, we'll see what is needed in this PR to merge it.

Hi @burnash,

Thank you so much for your patience, and I’m really sorry for the late response! 🙏

I wanted to share that I’ve ended up creating my own separate solution to address this issue, so I won’t have the motivation to contribute further here. That said, I think your approach is good! 

🌟 Wishing you the best of luck with implementing it, and I hope it turns out great! 🚀

Take care, and happy coding! 😊

riceyrice:
> I don't see a need to implement incremental in headers. I haven't seen a use case for incremental parametes in headers yet. If such a case exists, I would suggest implementing a dlt source using `RESTClient` rather than the rest_api source.

I have a use case that Zoho CRM API accepts `If-Modified-Since` header to filter to records updated since the given time. For now I'll look at using `RESTClient` to patch this in.

burnash:
@riceyrice awesome, thanks for the example and for proving me wrong :)

You can use incremental values in headers starting with dlt version 1.9.0 (the current stable release).

The devel branch in the docs has an updated [incremental section with an example](https://dlthub.com/docs/devel/dlt-ecosystem/verified-sources/rest_api/basic#in-request-headers):

```
{
    "path": "posts",
    "headers": {
        "X-Since-Timestamp": "{incremental.start_value}"  # In custom header
    },
    "incremental": {
        "cursor_path": "created_at",
        "initial_value": "2024-01-25T00:00:00Z",
    },
}
```

Let me know if that helps

PR #2099, opened 2024-11-26
docs: document that `path` can also be a URL

joscha:
via @burnash in https://github.com/dlt-hub/dlt/pull/2082#issuecomment-2498137202

> As for the base_url, I'm not sure it's necessary within the resource configuration since endpoint.path can accept an absolute URL (I'd need update the documentation to reflect this). Putting headers and auth into the endpoint configuration would make it consistent with how the underlying RESTClient and RESTClient.paginate operate.

### DIFF
--- a/dlt/sources/helpers/rest_client/client.py
+++ b/dlt/sources/helpers/rest_client/client.py
@@ -96,18 +96,18 @@ class RESTClient:
 
     def _create_request(
         self,
-        path: str,
+        path_or_url: str,
         method: HTTPMethod,
         params: Optional[Dict[str, Any]] = None,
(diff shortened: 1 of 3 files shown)


dlt/sources/helpers/rest_client/client.py
```diff
@@ -171,6 +171,8 @@ def paginate(
 
         Args:
             path (str): Endpoint path for the request, relative to `base_url`.
```
joscha, line 173: didn't touch this name, as public

burnash (approved):
Looks great, @joscha, thank you for the contribution!

PR #2100, opened 2024-11-27
allow to increase total count on most progress bars, fixes incorrect output in load stage

sh-rp:
### Description
Adds support for increasing total on progress bars.

### DIFF
--- a/dlt/common/runtime/collector.py
+++ b/dlt/common/runtime/collector.py
@@ -335,7 +335,8 @@ class AliveCollector(Collector):
             bar = self._bars[key] = bar.__enter__()
         # if message:
         #     bar.set_postfix_str(message)
-        bar(inc)
+        if inc > 0:
+            bar(inc)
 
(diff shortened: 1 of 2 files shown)


dlt/common/runtime/collector.py
```diff
@@ -333,19 +383,28 @@ def update(
             bar = alive_bar(total=total, title=desc, **self.alive_kwargs)
             self._bars_contexts[key] = bar
             bar = self._bars[key] = bar.__enter__()
+            self._bars_counts[key] = 0
+        else:
+            # TODO: implement once total change is supported
```
sh-rp, line 388: alive bars don't seem to support updating the total, you can create a new one and remember the old settings, but then a new bar appears in the ui, and the old half files one stays..

rudolfix: alive is just for fun. so OK :)

rudolfix (approved):
LGTM!

PR #2101, opened 2024-11-27
docs(rest_client): note about `data_selector`

joscha:
I personally think it's a bug and
```py
if self.data_selector:
    return self.extract_response(response, self.data_selector)
```
should be called automatically for all requests (`.get(...).json()`, ...) when given (1), however it would be a breaking change, so updating the documentation is the next best thing.
Might also be possibly worth deprecating the parameter from the global `RESTClient` constructor, given the fact that it is only used inside `.paginate(...)` (2).

Let me know if you want me to open a pull request for (1) and/or (2) instead of this one.

### DIFF
--- a/dlt/sources/helpers/rest_client/client.py
+++ b/dlt/sources/helpers/rest_client/client.py
@@ -58,7 +58,7 @@ class RESTClient:
         auth (Optional[AuthBase]): Authentication configuration for all requests.
         paginator (Optional[BasePaginator]): Default paginator for handling paginated responses.
         data_selector (Optional[jsonpath.TJsonPath]): JSONPath selector for extracting data from responses.
-            Only used when paginating. Use `extract_response` for other requests.
+            Only used in `paginate`.
         session (BaseSession): HTTP session for making requests.
         paginator_factory (Optional[PaginatorFactory]): Factory for creating paginator instances,
             used for detecting paginators.
--- a/dlt/sources/helpers/rest_client/client.py
+++ b/dlt/sources/helpers/rest_client/client.py
@@ -58,6 +58,7 @@ class RESTClient:
         auth (Optional[AuthBase]): Authentication configuration for all requests.
         paginator (Optional[BasePaginator]): Default paginator for handling paginated responses.
         data_selector (Optional[jsonpath.TJsonPath]): JSONPath selector for extracting data from responses.
+            Only used when paginating. Use `extract_response` for other requests.
         session (BaseSession): HTTP session for making requests.
         paginator_factory (Optional[PaginatorFactory]): Factory for creating paginator instances,
             used for detecting paginators.


burnash:
Hey @joscha you're absolutely right. I'd say it should not be in `RESTClient.__init__` indeed. I would opt for (2) as I don't think it's really needed for get() since the data (if it's a JSON response) can be retrieved directly from the resulted object. 
Thanks for spotting this!

joscha:
Do you want me to remove it from Rest client? It is a breaking change, albeit a super easy fix. Alternatively I could deprecate it instead and leave the comment in this PR.

burnash:
> Do you want me to remove it from Rest client? It is a breaking change, albeit a super easy fix. Alternatively I could deprecate it instead and leave the comment in this PR.

Let's deprecate it now (with the DeprecationWarning) and we'll remove it with the next major release.

joscha:
> > Do you want me to remove it from Rest client? It is a breaking change, albeit a super easy fix. Alternatively I could deprecate it instead and leave the comment in this PR.
> 
> Let's deprecate it now (with the DeprecationWarning) and we'll remove it with the next major release.

Okay. Will update the PR.

joscha:
Whilst doing the update I just realized that this was probably introduced for convenience. Like for example an API with lots of paginated endpoints that all follow the same pattern. It allows to do something like:

```py
rest_client = RESTClient(..., data_selector="data")

yield from rest_client.paginate("a")
yield from rest_client.paginate("b")
yield from rest_client.paginate("c")
...
yield from rest_client.paginate("x")
```
without having to pass the `data_selector="data"` to each of them:
```py
rest_client = RESTClient(...)

yield from rest_client.paginate("a", data_selector="data")
yield from rest_client.paginate("b", data_selector="data")
yield from rest_client.paginate("c", data_selector="data")
...
yield from rest_client.paginate("x", data_selector="data")
```


Which is definitely more brittle and less DRY.
Whilst it being a bit confusing to not be respected in get/... I actually do think after looking at the code, that there's validity in this convenience, so I'd propose to not deprecate it but leave the comment as-is. WDYT @burnash?

dlt/sources/helpers/rest_client/client.py
```diff
@@ -58,6 +58,7 @@ class RESTClient:
         auth (Optional[AuthBase]): Authentication configuration for all requests.
         paginator (Optional[BasePaginator]): Default paginator for handling paginated responses.
         data_selector (Optional[jsonpath.TJsonPath]): JSONPath selector for extracting data from responses.
+            Only used when paginating. Use `extract_response` for other requests.
```
burnash, line 61: Let's remove the suggestion to use `extract_response` as it could be actually simpler to directly get the data.
```suggestion
            Only used in `paginate`.
```

burnash:
> I actually do think after looking at the code, that there's validity in this convenience, so I'd propose to not deprecate it but leave the comment as-is.

Yes, also makes sense. Let's do that. I have a suggestion to adjust the wording though.

dlt/sources/helpers/rest_client/client.py
joscha: > Let's remove the suggestion to use extract_response as it could be actually simpler to directly get the data.

There's a lot of other lifting in that function though, I think it's actually quite good to use it, but I don't mind either way, committed.

PR #2105, opened 2024-11-28
🐛 Fix parquet layout example in the docs

trymzet:
### Description
Changes an example which doesn't work to one that does in this section https://dlthub.com/docs/dlt-ecosystem/destinations/filesystem#available-layout-placeholders


burnash (approved):
Thanks for the fix, @trymzet!

PR #2108, opened 2024-11-28
bug: `@dlt.transformer` ignores `write_disposition="replace"`

joscha:
### Description

When a transformer yields duplicates, they are not merged as defined in the given `write_disposition`.
See attached test, which is expected to pass, however it fails.

### Expected

multiple yields overwrite each other. Last yield wins.

Thus the expected resulting table should look like this:

| id | current_pass |
| -- | -- |
| 1 | 2 |
| 2 | 2 |

however, it looks like this:

| id | current_pass |
| -- | -- |
| 1 | 1 |
| 2 | 1 |
| 1 | 2 |
| 2 | 2 |

even though `primary_key` has been defined and `write_disposition` is `"replace"`

### DIFF
--- a/tests/load/pipeline/test_merge_disposition.py
+++ b/tests/load/pipeline/test_merge_disposition.py
@@ -1607,7 +1607,7 @@ def test_transformer_replace() -> None:
 
     current_page = 0
     @dlt.transformer(
-        data_from=gen_pages_of_ids, write_disposition="replace", primary_key="id", merge_key="id"
+        data_from=gen_pages_of_ids, write_disposition="replace", primary_key="id"
     )
     def transformed_ids(
         page: List[TDataItem],
--- a/tests/load/pipeline/test_merge_disposition.py
+++ b/tests/load/pipeline/test_merge_disposition.py
@@ -1,7 +1,7 @@
 from copy import copy
 import pytest
 import random
-from typing import List
+from typing import Iterable, List
 import pytest
 import yaml
 
@@ -18,8 +18,8 @@ from dlt.common.schema.exceptions import (
     CannotCoerceNullException,
 )
 from dlt.common.schema.typing import TLoaderMergeStrategy
-from dlt.common.typing import StrAny
-from dlt.common.utils import digest128
+from dlt.common.typing import StrAny, TDataItem
+from dlt.common.utils import digest128, uniq_id
 from dlt.common.destination import AnyDestination, DestinationCapabilitiesContext
 from dlt.common.destination.exceptions import DestinationCapabilitiesException
 from dlt.common.libs.pyarrow import row_tuples_to_arrow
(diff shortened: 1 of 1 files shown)


tests/load/pipeline/test_merge_disposition.py
```diff
@@ -1598,3 +1598,34 @@ def arrow_items(rows, schema_columns, timezone="UTC"):
...
+
+def test_transformer_replace() -> None:
+    @dlt.resource()
+    def gen_pages_of_ids() -> Iterable[TDataItem]:
+        yield [{"id": 1}, {"id": 2}]
+        yield [{"id": 1}, {"id": 2}]
+
+    current_page = 0
+    @dlt.transformer(
+        data_from=gen_pages_of_ids, write_disposition="replace", primary_key="id", merge_key="id"
```
joscha, line 1610:
```suggestion
        data_from=gen_pages_of_ids, write_disposition="replace", primary_key="id"
```

joscha:
> Hey @joscha , without looking at your test to deeply, are you maybe confusing write_disposition "replace" with "merge"? If you set it to replace, merge keys are not used, but the table is fully replaced with each pipeline run.

hi @sh-rp , updated the description and code to remove the merge key; it's the primary key I was after, sorry for the confusion. Would you mind looking at the code? I am also happy to pair/have a call/add more detail to it, if needed.

sh-rp:
@joscha the primary key also is not used when using replace. What exactly do you expect the primary key to do? If you need deduplication or merging, you'll have to use the merge write_disposition. In your example the transformer yields 4 rows alltogether and these seem to end up in the destination table, which is the correct behavior.

joscha:
> the primary key also is not used when using replace. What exactly do you expect the primary key to do?

I was under the impression the  primary key denotes uniqueness, do I have that wrong?

> If you need deduplication or merging, you'll have to use the merge write_disposition. In your example the transformer yields 4 rows alltogether and these seem to end up in the destination table, which is the correct behavior.

I expect a newer (later) record to enter the system replace an older (earlier) record if they have equal primary keys. So if I have a stream of data, that yields multiple of the same records, the database ends up with only one of them.

I.e.:
```jsonc
[
 { id: 1 },
 { id: 1 },
 { id: 1 },
 { id: 1 },
 { id: 1 },
 { id: 1 },
]
```
with `primary_key="id"` would yield 

| id |
| -- |
| 1  |

sh-rp:
The primary key is just a column hint, it does not do anything on its own. dlt does not enforce uniqueness, you can set it to translate to a primary key or unique key on some destinations which would make your example fail though because you'd be trying to insert two records with the same primary key.

If you want to have deduplication on the incoming records but still to a replace, you'll have to truncate the table before the pipeline run and use a merge write disposition with primary keys and possibly a dedup_sort hint: 

[create_authentication](https://dlthub.com/docs/general-usage/incremental-loading#delete-insert-strategy)

joscha:
Okay, thank you, will give that a try. This behaviour was not obvious to me. I think for two reasons:
* The notion of a primary key is overloaded; because of the name I automatically assumed it would be unique
* `"replace"` suggested to me it would work on a per-row basis, not only correspond to the whole table, thus I expected later records to overwrite earlier ones.

I wonder if it would make sense for me to open a pull request to the docs to add this information? It would have definitely helped me. WDYT?

I will try your suggestion above and close this issue in the meantime.

PR #2109, opened 2024-11-28
bug: `hint`s are retained across different yields

joscha:
### Description

I noticed in a real-life environment that mixing yields without markers and markers sometimes pushes data into the wrong table.
I was able to create a test case that reproduces it.

The expected outcome is 4 tables, looking like this:

`things`:
| id |
| -- |
| 1000 |
| 2000 |

`things_a`:
| id | my_resource_id | value |
| -- | -- | -- |
| a | 1000 | 1 |
| a | 2000 | 4 |

`things_b`:
| id | my_resource_id | value |
| -- | -- | -- |
| b | 1000 | 2 |
| b | 2000 | 5 |

`things_c`:
| id | my_resource_id | value |
| -- | -- | -- |
| c | 1000 | 3 |
| c | 2000 | 6 |

but instead `things_c` receives an additional item intended for the non-marked table `things`, so it looks like this:

`things_c`:
| id | my_resource_id | value |
| -- | -- | -- |
| c | 1000 | 3 |
| 2000 | `None` | `None` |
| c | 2000 | 6 |

### DIFF
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -2878,11 +2878,10 @@ def test_push_table_with_upfront_schema() -> None:
     info = copy_pipeline.run(data, table_name="events", schema=copy_schema)
     assert copy_pipeline.default_schema.version_hash != infer_hash
 
-def test_nested_inserts_correct_target() -> None:
-    @dlt.resource(
-        primary_key="id",
-        columns={"id": {"data_type": "bigint"}},
-    )
+
+@pytest.mark.parametrize("mark_main_resource", (True, False))
+def test_nested_inserts_correct_target(mark_main_resource: bool) -> None:
+    @dlt.resource()
     def my_resource():
         yield [
             {
@@ -2891,7 +2890,7 @@ def test_nested_inserts_correct_target() -> None:
                     {"id": "a", "value": 1},
                     {"id": "b", "value": 2},
                     {"id": "c", "value": 3},
-                ]
+                ],
             },
             {
                 "id": 2000,
@@ -2899,30 +2898,32 @@ def test_nested_inserts_correct_target() -> None:
                     {"id": "a", "value": 4},
                     {"id": "b", "value": 5},
                     {"id": "c", "value": 6},
-                ]
+                ],
             },
         ]
 
-    @dlt.transformer(
-        data_from=my_resource,
-        write_disposition="replace",
-        # parallelized=True,
-        primary_key="id",
-        merge_key="id"
-    )
+    @dlt.transformer(data_from=my_resource)
     def things(
         my_resources: List[TDataItem],
     ) -> Iterable[TDataItem]:
 
 
         for my_resource in my_resources:
-            fields: List[Dict] = my_resource.pop("fields")
-            yield my_resource
+            fields: List[TDataItem] = my_resource.pop("fields")
+            if mark_main_resource:
+                yield dlt.mark.with_hints(
+                    item=my_resource,
+                    hints=dlt.mark.make_hints(
+                        table_name="things",
+                    )
+                )
+            else:
+                yield my_resource
+
             for field in fields:
-                #id = field.pop("id")
-                id = field["id"]
-                table_name = f"things_{id}"
-                field = { "my_resource_id": my_resource["id"] } | field
+                my_id = field["id"]
+                table_name = f"things_{my_id}"
+                field = {"my_resource_id": my_resource["id"]} | field
                 yield dlt.mark.with_hints(
                     item=field,
                     hints=dlt.mark.make_hints(
@@ -2932,11 +2933,8 @@ def test_nested_inserts_correct_target() -> None:
                 )
 
     @dlt.source()
-    def my_source(
-    ) -> Sequence[DltResource]:
-        return (
-            things
-        )
+    def my_source() -> Sequence[DltResource]:
+        return [things]
 
     pipeline_name = "pipe_" + uniq_id()
     pipeline = dlt.pipeline(pipeline_name=pipeline_name, destination="duckdb")
@@ -2944,11 +2942,4 @@ def test_nested_inserts_correct_target() -> None:
     assert_load_info(info)
     rows = load_tables_to_dicts(pipeline, "things_c", exclude_system_cols=True)
     print(rows)
-    assert_data_table_counts(pipeline, {"things": 1, "things_a": 1, "things_b": 1, "things_c": 1 })
-    assert pipeline.last_trace.last_normalize_info.row_counts == {
-        "_dlt_pipeline_state": 1,
-        "things": 2,
-        "things_a": 2,
-        "things_b": 2,
-        "things_c": 2,
-    }
+    assert_data_table_counts(pipeline, {"things": 2, "things_a": 2, "things_b": 2, "things_c": 2 })
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -8,7 +8,7 @@ import random
 import shutil
 import threading
 from time import sleep
-from typing import Any, List, Tuple, cast
+from typing import Any, Dict, Iterable, List, Sequence, Tuple, cast
 from tenacity import retry_if_exception, Retrying, stop_after_attempt
 
 import pytest
@@ -35,7 +35,7 @@ from dlt.common.runtime.collector import LogCollector
 from dlt.common.schema.exceptions import TableIdentifiersFrozen
 from dlt.common.schema.typing import TColumnSchema
 from dlt.common.schema.utils import new_column, new_table
-from dlt.common.typing import DictStrAny
+from dlt.common.typing import DictStrAny, TDataItem
 from dlt.common.utils import uniq_id
 from dlt.common.schema import Schema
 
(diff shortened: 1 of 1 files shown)


joscha:
> is retained during extraction
yes! Can you point me to the code where that happens? I searched and couldn't find it.

> I'll need to verify if this is the correct behavior
I most definitely do not expect it and the examples in the documentation don't mention it either.
It also only happens on the very last insert. E.g. on Import of Y main rows with Y*X marked rows for other tables only the very last yield of Y will end up incrorrectly in the previously as X marked table.

> If you explicitely mark every row, then it will work.

It doesn't actually, I believe. I have tried this as the first attempt and I remember it not working, however figuring out where that row came from drove me a bit nuts, so I tried a lot of things and not very methodical. I will add a second test case to this pull request with an explicit mark for the parent resource.

sh-rp:
If you replace `yield my_resource` with:

```
            yield dlt.mark.with_hints(
                    item=my_resource,
                    hints=dlt.mark.make_hints(
                        table_name="things",
                    )
                )
```

you will get the expected result.

For where this problem originates this will be in the normalizer or the pipe iterator. If you can have a stab at it. We are currently very busy but should come back to this shortly, but you can always ask for help here :)

sh-rp:
PS: Yes, we consider this a bug.

joscha:
> you will get the expected result.

just added a second parameter to the test to make sure it works as intended: 3cff1a3f - it does, so \o/

> this will be in the normalizer or the pipe iterator.

OK, will have a short look if I can find it.

joscha:
```
things {'id': 1000}
things_a {'my_resource_id': 1000, 'id': 'a', 'value': 1}
things_b {'my_resource_id': 1000, 'id': 'b', 'value': 2}
things_c {'my_resource_id': 1000, 'id': 'c', 'value': 3}
None
things_c {'id': 2000} <---- HERE IS THE ISSUE, this should be `things`
things_a {'my_resource_id': 2000, 'id': 'a', 'value': 4}
things_b {'my_resource_id': 2000, 'id': 'b', 'value': 5}
things_c {'my_resource_id': 2000, 'id': 'c', 'value': 6}
```
via
```diff
diff --git a/dlt/extract/extractors.py b/dlt/extract/extractors.py
index 41d3035a..c62fa44e 100644
--- a/dlt/extract/extractors.py
+++ b/dlt/extract/extractors.py
@@ -135,6 +135,7 @@ class Extractor:
 
         if table_name := self._get_static_table_name(resource, meta):
             # write item belonging to table with static name
+            print(table_name, items)
             self._write_to_static_table(resource, table_name, items, meta)
         else:
             # table has name or other hints depending on data items
@@ -148,8 +149,10 @@ class Extractor:
 ```
shows that in subsequent runs after setting a hint on an item, the resource seems to store it.

The reason it stores it, is this call to `merge_hints`:
https://github.com/dlt-hub/dlt/blob/61c2ed96053bd02632b87e2c85fa940a91a9d03b/dlt/extract/extractors.py#L129

`merge_hints` calls `resource.apply_hints`, which then mutates the underlying resource. Any subsequent writes afterwards that do NOT bring it's own meta (e.g. `table_name` is empty) , will then fall back to the meta of the resource in https://github.com/dlt-hub/dlt/blob/61c2ed96053bd02632b87e2c85fa940a91a9d03b/dlt/extract/extractors.py#L153 and 💥 .

I believe this would possibly not only cause issues for table names, but also for all other metadata, if any of it would divert from what's on the resource. `table_name` is just the most obvious one.

There are various fixes that could be done, but I am unsure if there are any backwards-compatible ones if anyone is relying on this not very sane behavior, as someone could have used this as a feature like:

```py
yield item1
yield item2
yield dlt.mark.with_hints(item3, hints=dlt.mark.make_hints(table_name="xxx"))
yield item4
yield item5
yield item...
```
whoch would have yielded 2 items into the table defined in the resource and then all other subsequent items into table `xxx`.

joscha:
Something like this also fails:


```py
@dlt.transformer(
    table_name=lambda item: "xxxx"
)
def xxxx() -> Iterable[TDataItem]:
    yield {"id": 1}
    yield dlt.mark.with_hints(
        item={"id": 1000},
        hints=dlt.mark.make_hints(
            table_name="yyyy",
            write_disposition="merge",
            primary_key="id",
        ),
    )
```

with:
```
A set of table hints provided to the resource is inconsistent: Table name yyyy must be a function if any other table hint is a function
```
for the same reason - it tries to merge the hitns for the sub-table (`yyyy`) into the main resource and then 💥 because the main resource has a dynamic table name.
Workaround is `table_name=lambda item: "yyyy"` but that's besides the point; it's the same merge that causes the issue.

sh-rp:
@joscha thanks for the investigation, this makes sense! I have assigned the task to myself and hope I'll get to fixing it next week. Keep an eye on the pr :)

joscha:
I found another issue relating to columns, which is the same codepath:
```py
@dlt.resource()
def my_resource():
    yield dlt.mark.with_hints(
                            item={"hello": "..."},
                            hints=dlt.mark.make_hints(
                                table_name="a",
                                columns={
                                    "hello": {
                                        "data_type": "text"
                                    }
                                }
                            ),
                        )
    yield dlt.mark.with_hints(
                            item={"world": "..."},
                            hints=dlt.mark.make_hints(
                                table_name="b",
                            ),
                        )
```

should create:

`Table: a`
| hello |
| -- |
| ... |

`Table: b`
| world |
| -- |
| ... |


but creates:

Screenshot 2024-12-11 at 2 19 05 pm

(the column definition is retained on the pipeline for subsequent runs until another `hint` overwrites it.

joscha:
@sh-rp this bug makes dlt almost unusable when trying to yield data into different tables, I'd love to work towards a fix, can I assume that we agree this is a regression and hints should not be shared across different yields at all, e.g. the pipeline state must not be tainted by whatever hint is passed with a marked yield?

rudolfix:
@joscha we are a little in the area of undefined behavior. try to create table variants if you use dynamic table name (maybe we should set this automatically and for sure we should document it)
```py
def with_hints(
    item: TDataItems, hints: TResourceHints, create_table_variant: bool = False
) -> DataItemWithMeta:
    """Marks `item` to update the resource with specified `hints`.

    Will create a separate variant of hints for a table if `name` is provided in `hints` and `create_table_variant` is set.

    Create `TResourceHints` with `make_hints`.
    Setting `table_name` will dispatch the `item` to a specified table, like `with_table_name`
    """
    return DataItemWithMeta(HintsMeta(hints, create_table_variant), item)
```

rudolfix:
regarding the state: all the tables are generated by a single resource and will be marked as such. so they share a single state

joscha:
> @joscha what happens if you do

Still running; So far the only thing that I see is a constant stream of:
```
2024-12-12 13:18:11,692|[WARNING]|72231|8212723776|dlt|hints.py|_set_hints:389|A data item validator was created from column schema in companies for a table `<some_table>` variant. Currently such validator is ignored.
```

with `<some_table>` being one of the names of my dynamic tables.
Will report back once the run ist done.

joscha:
> @joscha what happens if you do
[...]
>                                 create_table_variant=True,

I added another parametrized test in e1cbe19.
Result is:
```
tests/pipeline/test_pipeline.py::test_columns_do_not_linger[True] {'a': [{'hello': '...'}], 'b': [{'world': '...'}]}
PASSED
tests/pipeline/test_pipeline.py::test_columns_do_not_linger[False] {'a': [{'hello': '...'}], 'b': [{'hello': None, 'world': '...'}]}
FAILED
```

so `create_table_variant=True` seems to alleviate the issue with the columns at the very least.

joscha:
I've been running this a while with table variants and I believe it is working as expected.
We definitely need to make
 > A data item validator was created from column schema in companies for a table `<some_table>` variant. Currently such validator is ignored.
configurable, so it's usually suppressed.

Also, I still think a fix is needed for the standard case. Or at the very very least extensive documentation both markdown and code around the current behavior.

sh-rp:
Hey @joscha, to summarize what the current state of this discussion is.

* We need to make sure that warnings don't pile up over and over again (you can create a ticket for this)
* We need to document the table variants stuff (this could also go into a ticket with an example of this case)

Is there anything else?

joscha:
> Is there anything else?

Only one extra item comes to mind: if we agree that yielding items with hints without `create_table_variant=True` mutates the resource state is not a bug, we should probably also add a couple tests in regards to it that assert this explicitly.

sh-rp:
@joscha what do you mean with mutating table state without this flag? I can't quite tell from the conversation above. As @rudolfix noted, there is one state section per resource and even if you yield to multiple different tables from this resource, the state is still managed in the same place.

joscha:
> what do you mean with mutating table state without this flag?

I mean that any item(s) you dispatch with a hint mutates the resource state unless you pass `create_table_variant=True`. I think it is not obvious at all that dispatching a single item with a hint will mean that any future yielded items will also be stored with that hint. imho we should:

1. either not share state at all
2. make `create_table_variant=True` the default
3. or document this very clearly in the docs

The order of this list is in order of preference / what seems to be the most logical. Please note that 1) and 2) are breaking backwards compatibility if anyone has been relying on this behaviour.

sh-rp:
@joscha yes, this is the correct behavior. I also just had a look at the section in the docs where this is explained, and it says there emitting hints with "with_hints" has the same effect as running apply_hints, so the behavior is documented correctly, just maybe not clear enough. If you'd like to improve this, you can update the docs and maybe add more tests, let me know. The default behavior should not change though. As @rudolfix mentions, there may be a few edge cases that are not properly defined or have unexpected results, so you're welcome to write some tests for those too. I'll leave it up to you if you want to work on this, any contributions are of course welcome :)

sh-rp:
PS: The reasoning for this behavior is also explained in this section of the docs.

joscha:
> just maybe not clear enough

Yeah, definitely not I think. Even you 

(via [see restore_from_destination=False,](https://github.com/dlt-hub/dlt/pull/2109#issuecomment-2511275022))

> PS: Yes, we consider this a bug.

seemed to be thrown at the default behavior as outlined in the ticket.
Given that now it seems it's not a bug, but expected behvior I will close this issue.

For any future readers coming across this issue, the solution (workaround) is to pass:

```py
create_table_variant=True
```
to every `with_hints` call.

PR #2117, opened 2024-12-03
fixes dlt init fails in Colab

rudolfix:
### Description
userdata is available when outside of Notebook but kernel is not so the code to get secrets fails


AstrakhantsevaAA (approved):
Thank you Marcin!

PR #2120, opened 2024-12-04
Add missing mention of the required `endpoint_url` config in GCS

trymzet:
### Description
This config is mentioned in [another place](https://dlthub.com/docs/dlt-ecosystem/destinations/filesystem#using-s3-compatible-storage) discussing S3-compatible storage, but omitted [here](https://dlthub.com/docs/dlt-ecosystem/destinations/clickhouse#using-google-cloud-or-s3-compatible-storage-as-a-staging-area) from the text (only included in the example).


burnash (approved):
Thanks for spotting this and for the fix, @trymzet

PR #2129, opened 2024-12-09
Fix validation error in for custom auth classes

### LINKED ISSUE #2122: Registering custom auth class leads to `DictValidationException` with REST API source
### dlt version

1.4.1

### Describe the problem

I'm trying to register a custom authentication class for the REST API source in order to use the dictionary configuration syntax, as described [here](https://dlthub.com/docs/dlt-ecosystem/verified-sources/rest_api/basic#authentication). However, when creating the source, I get a validation error because my new registered class is not part of the built-in ones, even though the `get_auth_class()` function returns the correct class.

This really neat feature was introduced in https://github.com/dlt-hub/verified-sources/pull/556 and would fit our use cases perfectly 😄 Any suggestions on what I'm doing wrong, @willi-mueller?

Note that using the class directly works:
```python
"auth": MorningstarAuth(username, password)
```
The issue seems to be related to `register_auth()` updating the `AUTH_MAP` constant which doesn't get used when the config dictionary is validated, hence leading to the `DictValidationException`.

### Expected behavior

After registering the custom auth class, I should be able to use the configuration syntax and get the same result as using the class directly.

### Steps to reproduce

```python
import dlt
from dlt.common.configuration.specs.base_configuration import configspec
from dlt.sources.helpers.rest_client.auth import AuthConfigBase
from dlt.sources.rest_api.config_setup import register_auth
from requests import PreparedRequest


@configspec
class MorningstarAuth(AuthConfigBase):
    username: str = ""
    password: str = ""

    def __call__(self, request: PreparedRequest) -> PreparedRequest:
        request.headers["Authorization"] = f"Bearer {self.password}"
        return request


register_auth("morningstar", MorningstarAuth)


@dlt.source
def morningstar_source():
    source_config: RESTAPIConfig = {
        "client": {
            "auth": {
                "type": "morningstar",
                "username": "some_username",
                "password": "some_password",
            },
            "base_url": "https://www.emea-api.morningstar.com/",
        },
        "resources": [{"name": "screener", "endpoint": {"path": "/ecint/v1/screener"}}],
    }
    yield from rest_api_resources(source_config)


source = morningstar_source()
```

I get the following error:

```bash
dlt.common.exceptions.DictValidationException: In path ./client: field 'auth' expects the following types: AuthConfigBase, Literal, BearerTokenAuth, BearerTokenAuthConfig, APIKeyAuth, ApiKeyAuthConfig, HttpBasicAuth, HttpBasicAuthConfig, OAuth2ClientCredentials, OAuth2ClientCredentialsConfig. Provided value {'type': 'morningstar', 'username': 's*****e', 'password': 's*****d'} with type 'dict' is invalid with the following errors:
For BearerTokenAuthConfig: In path ./client/auth: following required fields are missing {'token'}
For ApiKeyAuthConfig: In path ./client/auth: following required fields are missing {'api_key'}
For HttpBasicAuthConfig: In path ./client/auth: field 'type' with value morningstar is not one of: ['bearer', 'api_key', 'http_basic', 'oauth2_client_credentials']
For OAuth2ClientCredentialsConfig: In path ./client/auth: following required fields are missing {'access_token_url', 'client_secret', 'client_id'}
For AuthConfigBase: In path ./client: field 'auth' expects callable (function or class instance) but got  '{'type': 'morningstar', 'username': 's*****e', 'password': 's*****d'}'. Mind that signatures are not validated
For Literal: In path ./client: field 'auth' with value {'type': 'morningstar', 'username': 's*****e', 'password': 's*****d'} is not one of: ['bearer', 'api_key', 'http_basic', 'oauth2_client_credentials']
For BearerTokenAuth: In path ./client: field 'auth' expects callable (function or class instance) but got  '{'type': 'morningstar', 'username': 's*****e', 'password': 's*****d'}'. Mind that signatures are not validated
For APIKeyAuth: In path ./client: field 'auth' expects callable (function or class instance) but got  '{'type': 'morningstar', 'username': 's*****e', 'password': 's*****d'}'. Mind that signatures are not validated
For HttpBasicAuth: In path ./client: field 'auth' expects callable (function or class instance) but got  '{'type': 'morningstar', 'username': 's*****e', 'password': 's*****d'}'. Mind that signatures are not validated
For OAuth2ClientCredentials: In path ./client: field 'auth' expects callable (function or class instance) but got  '{'type': 'morningstar', 'username': 's*****e', 'password': 's*****d'}'. Mind that signatures are not validated
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

REST API

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

burnash:
Fixes #2122

### DIFF
--- a/dlt/common/typing.py
+++ b/dlt/common/typing.py
@@ -484,3 +484,18 @@ def copy_sig_any(
         return func
 
     return decorator
+
+
+def add_value_to_literal(literal: Any, value: Any) -> None:
+    """Extends a Literal at runtime with a new value.
+
+    Args:
+        literal (Type[Any]): Literal to extend
+        value (Any): Value to add
+
+    """
+    type_args = get_args(literal)
+
+    if value not in type_args:
+        type_args += (value,)
+        literal.__args__ = type_args
--- a/dlt/sources/rest_api/config_setup.py
+++ b/dlt/sources/rest_api/config_setup.py
@@ -20,6 +20,7 @@ from dlt.common import logger
 from dlt.common.configuration import resolve_configuration
 from dlt.common.schema.utils import merge_columns
 from dlt.common.utils import update_dict_nested, exclude_keys
+from dlt.common.typing import add_value_to_literal
 from dlt.common import jsonpath
 
 from dlt.extract.incremental import Incremental
@@ -64,6 +65,7 @@ from .typing import (
     ResponseActionDict,
     Endpoint,
     EndpointResource,
+    AuthType,
 )
 
 
@@ -152,6 +154,8 @@ def register_auth(
             "Your custom auth has to be a subclass of AuthConfigBase"
         )
     AUTH_MAP[auth_name] = auth_class
+
+    add_value_to_literal(AuthType, auth_name)
 
 
 def get_auth_class(auth_type: str) -> Type[AuthConfigBase]:
@@ -285,7 +289,7 @@ def build_resource_dependency_graph(
             resolved_param_map[resource_name] = None
             break
         assert isinstance(endpoint_resource["endpoint"], dict)
-        # connect transformers to resources via resolved params
+        # find resolved parameters to connect dependent resources
         resolved_params = _find_resolved_params(endpoint_resource["endpoint"])
 
         # set of resources in resolved params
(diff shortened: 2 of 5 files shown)


rudolfix (changes requested):
@burnash LGTM. we still miss a unit test for `add_value_to_literal` in `test_typing.py`. reason: this is hacky so we want to test it for all OS/python combinations

burnash:
@rudolfix I've added the test. As discussed imo the better solution is to create a special extensible type for this

PR #2131, opened 2024-12-10
convert add_limit to pipe step based limiting

sh-rp:
### Description

Up until now we were managing limits inside a somewhat conflated function that was wrapping generators. The problem was that limits where applied before incrementals where and that we had a certain amount of code duplication with respect to wrapping async iterators. This PRs solves this.


dlt/extract/utils.py
```diff
@@ -183,6 +183,17 @@ def check_compat_transformer(name: str, f: AnyFun, sig: inspect.Signature) -> in
     return meta_arg
 
 
+def wrap_iterator(gen: Iterator[TDataItems]) -> Iterator[TDataItems]:
```
sh-rp, line 186: pure iterators cannot be stopped, so we need to wrap them in a generator that I can control from the limit step.

dlt/extract/items.py
```diff
@@ -238,3 +238,23 @@ class ValidateItem(ItemTransform[TDataItem]):
...
+    def __init__(self, max_items: int) -> None:
+        self.max_items = max_items if max_items is not None else -1
+
+    def bind(self, pipe: SupportsPipe) -> "LimitItem":
+        self.gen = pipe.gen
+        self.count = 0
+        return self
+
+    def __call__(self, item: TDataItems, meta: Any = None) -> Optional[TDataItems]:
+        if self.count == self.max_items:
```
rudolfix, line 255: I'm not sure that is enough. we should close gen when we reach max items. but in this implementation we close gen at the end of the pipe. not all steps are sync steps, there are for example steps that yield. (or maybe even async steps, I do not remember). we should still return None when count > max_items.

I think we need to add more tests.
- what happens if we do add_yield_map?
- are we testing limit for async generators / iterators
- any differences for round robin / fifo?
- make sure that all expected items are pushed to transformer (this happens via special ForkStep)

rudolfix (changes requested):
this looks so simple now and I think implementation is right. two things:
- we have a test failing in a curious way. apparently we call rest API twice even if the limit is 1. why? we count items at the end of the pipe but there's just a single pipe. we must investigate
- by counting at end of the pipe we change the behavior. I think it makes sense... but maybe we can add sticky flag as an argument to add_limit? so people can still stick it to gen object and count unfiltered items as before?

dlt/extract/resource.py
```diff
@@ -214,31 +215,24 @@ def requires_args(self) -> bool:
...
     @validator.setter
     def validator(self, validator: Optional[ValidateItem]) -> None:
         """Add/remove or replace the validator in pipe"""
-        step_no = self._pipe.find(ValidateItem)
-        if step_no >= 0:
-            self._pipe.remove_step(step_no)
+        self._pipe.remove_by_type(ValidateItem)
         if validator:
-            self.add_step(validator, insert_at=step_no if step_no >= 0 else None)
+            self.add_step(validator)
```
sh-rp, line 235: NOTE: i remove inserting at the same position in favor of automatic resolution via placement affinity. I think this makes more sense, I can revert to the old behavior though.

dlt/extract/items.py
```diff
@@ -238,3 +240,49 @@ class ValidateItem(ItemTransform[TDataItem]):
     def bind(self, pipe: SupportsPipe) -> ItemTransform[TDataItem]:
         self.table_name = pipe.name
         return self
+
+
+class LimitItem(ItemTransform[TDataItem]):
+    placement_affinity: ClassVar[float] = 1.1  # stick to end right behind incremental
+
+    def __init__(
```
sh-rp, line 248: I have moved the time limit and rate limiting over from the other PR. I think the time limit is fairly uncontroversial, the rate limiting is a little bit sketchy. I think it would be really cool to implement this with this PR, but we could also add a kind of global rate limiting on the Pipeiterator level that gets handed over from the DltSource to the Pipeiterator and is applied in the _get_source_item function to only extract a new item from any pipe if a min amount of time has passed.

dlt/extract/items.py
```diff
@@ -238,3 +238,23 @@ class ValidateItem(ItemTransform[TDataItem]):
     def bind(self, pipe: SupportsPipe) -> ItemTransform[TDataItem]:
         self.table_name = pipe.name
         return self
+
+
+class LimitItem(ItemTransform[TDataItem]):
+    placement_affinity: ClassVar[float] = 1.1  # stick to end right behind incremental
```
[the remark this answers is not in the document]
sh-rp: I have not done this for now, but changed a few convenience methods. The incremental confused me a bit ;)

dlt/extract/items.py
sh-rp: I think this is all tested now, except for round robin and fifo, but I am quite sure that this will not make a difference, since rr and fifo only apply at the get_source_item level and there is no async stuff going on in the add_limit (it's all taken care of already in other places)

dlt/extract/items.py
sh-rp: update: only keeping the time limit here which is very straightforward to implement and I think pretty useful.

joscha:
related to #2142

joscha:
closes #2142

rudolfix (changes requested):
still missing two things:
1. actual test where limit is used with incremental to do the backfilling. my take is to add this to sql_database tests and (remember to add row_order)
2. having this example I'd add it to performance guide and say how to split large backfils. esp. mentioning that records should be ordered and not to retake data twice (ie. via WHERE clause)
3. on Py 3.11 some tests are not passing consistently. pls take a look

dlt/extract/items.py
```diff
@@ -238,3 +240,42 @@ class ValidateItem(ItemTransform[TDataItem]):
...
+        return self
+
+    def __call__(self, item: TDataItems, meta: Any = None) -> Optional[TDataItems]:
+        self.count += 1
+
+        # detect when the limit is reached, max time or yield count
+        if (
+            (self.count == self.max_items)
+            or (self.max_time and time.time() - self.start_time > self.max_time)
+            or self.max_items == 0
```
rudolfix, line 266: max_items == 0 is wrong argument. does it make sense to return 0 items? some kind of connection check?

rudolfix: nooo! actually it makes sense. you can check if we can connect and parse data from the source without loading anything.

sh-rp:
> still missing two things:
> 
>     1. actual test where limit is used with incremental to do the backfilling. my take is to add this to sql_database tests and (remember to add row_order)
> 
>     2. having this example I'd add it to performance guide and say how to split large backfils. esp. mentioning that records should be ordered and not to retake data twice (ie. via WHERE clause)
> 
>     3. on Py 3.11 some tests are not passing consistently. pls take a look

I have added a general test that combines incremental and add_limit. I will also add a nice example using the rfam database I think but will have to do this tomorrow.

dlt/extract/utils.py
rudolfix: OK I have afterthought here. This is applied unconditionally to all `gen` objects? also to simple lists when there's no Limit step?

if so this will slow down simple pipelines considerably. maybe we should wrap in `bind` of `Limit`?

tests/extract/test_incremental.py
```diff
@@ -3960,3 +3961,58 @@ def some_data(
...
+
+    p = dlt.pipeline(pipeline_name="incremtal_limit", destination="duckdb", dev_mode=True)
+
+    p.run(resource())
+
+    # check we have the right number of items
+    assert len(p.dataset().items.df()) == 10
+    assert resource_called == 10
+    # check that we have items 0-9
+    assert p.dataset().items.df().id.tolist() == list(range(10))
```
rudolfix, line 4000: does that work? not sure we have pandas installed when we run `extract` tests

rudolfix (changes requested):
I'm not sure we should always wrap iterators. See my comment

tests/extract/test_incremental.py
sh-rp: it passed in the last run, yes

sh-rp:
@rudolfix I have delayed the iterator wrapping to the LimitItem binding now, I agree that this is probably a good idea. That said, the typing/importing is a bit messy now imho..

tests/extract/test_incremental.py
```diff
@@ -3960,3 +3961,58 @@ def some_data(
...
+            nonlocal resource_called
+            resource_called += 1
+            yield {
+                "id": i,
+                "value": str(i),
+            }
+
+    resource.add_limit(10)
+
+    p = dlt.pipeline(pipeline_name="incremtal_limit", destination="duckdb", dev_mode=True)
```
joscha, line 3992:
```suggestion
    p = dlt.pipeline(pipeline_name="incremental_limit", destination="duckdb", dev_mode=True)
```

rudolfix (approved):
LGTM!

PR #2135, opened 2024-12-11
fix ibis az problems on linux

sh-rp:
### Description
Reverse order of doing things when creating the ibis backend to see wether we can fix the problems with azure.


rudolfix (changes requested):
I merged a hack into destinations.yaml that links certs in a proper place. IMO you should remove it and run the CI again

rudolfix:
@sh-rp now it fails again...

sh-rp:
¯\_(ツ)_/¯

rudolfix:
@sh-rp heh that was tricky. actually `SET` was executed on a clone of the connection, while `ibis` was using another clone. And `SET` is not shared. pls check my commit and review

sh-rp (commented):
looks good, the tests also pass, thanks for figuring this out!

rudolfix (approved):
LGTM!

PR #2141, opened 2024-12-12
make duckdb handle Iceberg table with nested types

jorritsandbrink:
Main things in this PR:
- workaround to read Iceberg tables with nested types in duckdb (https://github.com/duckdb/duckdb_iceberg/issues/47)
- fix to replace existing duckdb views for Iceberg tables or when the `abfss` scheme is used—views need replacement in these cases because their definitions reference specific files, and these references need to be updated when the source table changes
- switch to `dev_mode` in `tests/load/filesystem/test_sql_client.pytest_table_formats` to work around leftover data from previous tests when using `gs` scheme via S3 interface (https://github.com/dlt-hub/dlt/issues/2148)


rudolfix (approved):

rudolfix (approved):
LGTM!

PR #2142, opened 2024-12-12
docs: document that `parallelized=True` resources with `add_limit(x)` usually yield `x-1`

joscha:
Most of the time (80%ish or so) parallelized resources that are limited yield one item less.

### DIFF
--- a/tests/pipeline/test_resources_evaluation.py
+++ b/tests/pipeline/test_resources_evaluation.py
@@ -217,6 +217,19 @@ def test_limit_async_resource() -> None:
     assert len(result) == 13
 
 
+@pytest.mark.parametrize("parallelized", [False, True])
+def test_limit_sync_resource(parallelized: bool) -> None:
+    @dlt.resource(parallelized=parallelized)
+    def sync_resource1():
+        for i in range(1, 10):
+            yield i
+
+    limit = 5
+    result = list(sync_resource1().add_limit(limit))
+    allowed_result_range = range(limit - int(parallelized), limit + 1)
+    assert len(result) in allowed_result_range
+
+
 @pytest.mark.parametrize("parallelized", [True, False])
 def test_parallelized_resource(parallelized: bool) -> None:
     os.environ["EXTRACT__NEXT_ITEM_MODE"] = "fifo"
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_resources_evaluation.py
```diff
@@ -217,6 +217,19 @@ async def async_resource1():
...
+@pytest.mark.parametrize("parallelized", [False, True])
+def test_limit_sync_resource(parallelized: bool) -> None:
+    @dlt.resource(parallelized=parallelized)
+    def sync_resource1():
+        for i in range(1, 10):
+            yield i
+
+    limit = 5
+    result = list(sync_resource1().add_limit(limit))
+    allowed_result_range = range(limit - int(parallelized), limit + 1)
```
joscha, line 229: this is not ideal.
Ideal would be to make the parallel yields exact. I tried this for a few hours to no avail. Next best would be to run this test x times when `parallelized=True` and then ensure that the threshold of yields lies somewhere above 4.5 or so.

sh-rp:
@joscha thanks for this PR. If you look in the PR list, you can see that I have changed the add_limit implementation. You could check out that branch and see wether you are still seeing those problem there or maybe even add your test there. I think it might actually be resolved.

joscha:
Will give your branch a try!

sh-rp:
@joscha amazing, your helping out on this is is very much appreciated :)

joscha:
> @joscha amazing, your helping out on this is very much appreciated :)

I am glad it now produces exactly `limit` results to be honest. I was quite stumped for a bit, thought the REST api I was using had an issue. I might make sense to merge the test in this pull request after merging #2131, so we pick up regressions. As far as I can tell this is currently not asserted anywhere. With your changes in 3738c29b467f8a9e9807a0c9304f91c4e479f8c1 the code in this PR becomes:
```py
@pytest.mark.parametrize("parallelized", [False, True])
def test_limit_sync_resource(parallelized: bool) -> None:
    @dlt.resource(parallelized=parallelized)
    def sync_resource1():
        for i in range(1, 10):
            yield i

    limit = 5
    result = list(sync_resource1().add_limit(limit))
    assert len(result) == limit
```
which is precise enough to keep.

sh-rp:
@joscha, I think this case is already tested with the async resource test (as a parallized resource is just that under the hood) but I added this case to my PR. I'll close this one now :)

PR #2161, opened 2024-12-18
docs: remove outdated subsection

joscha:
### Description

References are described in the section right above.

### DIFF
--- a/docs/website/docs/general-usage/schema.md
+++ b/docs/website/docs/general-usage/schema.md
@@ -227,9 +227,6 @@ You are able to bring your own `row_key` by adding a `_dlt_id` column/field to y
 
 `merge` write disposition requires an additional nested reference that goes from **nested** to **root** table, skipping all parent tables in between. This reference is created by [adding a column with a hint](incremental-loading.md#forcing-root-key-propagation) `root_key` (named `_dlt_root_id` by default) to nested tables.
 
-### Table references
-You can annotate tables with table references. This feature is coming soon.
-
 ## Schema settings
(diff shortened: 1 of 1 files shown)


docs/website/docs/general-usage/schema.md
```diff
@@ -227,9 +227,6 @@ You are able to bring your own `row_key` by adding a `_dlt_id` column/field to y
 
 `merge` write disposition requires an additional nested reference that goes from **nested** to **root** table, skipping all parent tables in between. This reference is created by [adding a column with a hint](incremental-loading.md#forcing-root-key-propagation) `root_key` (named `_dlt_root_id` by default) to nested tables.
 
-### Table references
-You can annotate tables with table references. This feature is coming soon.
```
sh-rp, line 231: @joscha thanks for the PR. The section above describes nested references, not table references, if you read the whole references section you can see that this differentiation is done at the beginning of the section. We have a conceptual implementation of these table references but they are not available in the core library yet.

joscha: I see - should this be `####` then, to be one level below?

joscha: changed it in 3aaa33835c90cfddecc3a5414ba2e32ba2f2123c

sh-rp: no, not really, ### is correct.

joscha: OK, I don't think it makes sense:

PR #2162, opened 2024-12-18
Docs: fix credentials link (add slash)

AstrakhantsevaAA:
>there seems to be a difference in wether there is a slash at the end of the url


sh-rp (approved):
solid work :)

PR #2165, opened 2024-12-19
Apply hints for nested tables

### LINKED ISSUE #1647: Simplify schema modification of child tables

steinitzu:
### Description
**Draft** of nested table hints implementation:
 ```
apply_hints(path=['a', 'b', 'c'], columns=...)
```
Is working so far but there are some bugs and tests needed.

### Related Issues

- Resolves #1647

### CODE REVIEW

**Review by rudolfix — CHANGES_REQUESTED** [review]
Please see my suggestion how to deal with naming convention. Docs requirements are in the ticket.

**Comment by zilto** [comment]
The current implementation adds the tables to the schema (as tested), but it doesn't affect how the data is loaded. 

For example, the hints will appear in 
```python
pipeline.default_schema.tables.keys()
# ignoring the dlt tables
# 'nested_data', 'override_child_outer1', 'override_child_outer1_innerfoo','nested_data__outer1', 'nested_data__outer1__innerfoo'
```

Whereas the normalizer row counts show no ingested data for the tables
```python
pipeline.last_trace.last_normalize_info.row_counts
# 'nested_data': 2, 'nested_data__outer1': 2, 'nested_data__outer1__innerfoo': 2
```

I believe changes need to be made to `Extractor._write_to_dynamic_table()` and `_get_dynamic_table_name()` to push data to the right table. (`Extractor._write_to_static_table()` should rely on the explicitly provided table name).

The extractor would need to hold some mapping, but it could be more appropriate to move the logic to `dlt.common.normalizers.json.helpers` or to a `Schema` method?

rudolfix:
Relational normalizer follows its logic of creating nested tables and column names. it comes only from the data. there's no mechanism to rename those, except the root table name which the user **must** set.

`dlt` is data first, not schema first. it is counterintuitive if you chose to start your work with schema, not data.

I assume that in example you are giving, you used a custom table name for nested table. If this is not the case ping me on slack. maybe there's a bug somewhere

in the ticket above, there's a note:

> You still may allow users to specify table_name on the nested hint. If you do so, you'll need to modify the normalizer so it maps paths to those names. IMO this is for another ticket and bigger overhaul of the schema
prevent following to be set on nested table:
parent_table_name: TTableHintTemplate[str] = None,
incremental: TIncrementalConfig = None,

so I'd say we block setting table name on nested hints (also parent name and incremental do make sense)

zilto:
> We need a good example (in Examples). my take is: let's go for something advanced:
here's a test test_merge_on_keys_in_schema: this test models the nested json into a set of tables with natural primary keys (not dlt generated ones). it uses some tricks to modify the nested tables. now we should be able to use our nested hints to just decorate resource properly. you are also free to change this test first and run it!
the goal of the example is to show how you can custom model your nested tables.

Working on the requested documentation (#1647), I encounter many edge cases of the current implementation. My understanding is that our goal is for these two snippets to be equivalent

## Current test
From  `tests/load/pipeline/test_merge_disposition.py::test_merge_on_keys_in_schema`
```python
@dlt.source(schema=schema)
def ethereum(slice_: slice = None):
    @dlt.resource(
        table_name="blocks",
        write_disposition={"disposition": "merge", "strategy": merge_strategy},
    )
    def data():
        with open(
            "tests/normalize/cases/ethereum.blocks.9c1d9b504ea240a482b007788d5cd61c_2.json",
            "r",
            encoding="utf-8",
        ) as f:
            yield json.load(f) if slice_ is None else json.load(f)[slice_]

    # also modify the child tables (not nested)
    schema_ = dlt.current.source_schema()
    blocks__transactions = schema_.tables["blocks__transactions"]
    blocks__transactions["write_disposition"] = "merge"
    blocks__transactions["x-merge-strategy"] = merge_strategy  # type: ignore[typeddict-unknown-key]
    blocks__transactions["table_format"] = destination_config.table_format

    blocks__transactions__logs = schema_.tables["blocks__transactions__logs"]
    blocks__transactions__logs["write_disposition"] = "merge"
    blocks__transactions__logs["x-merge-strategy"] = merge_strategy  # type: ignore[typeddict-unknown-key]
    blocks__transactions__logs["table_format"] = destination_config.table_format

    return data
```

## Target API
```python
@dlt.source(schema=schema)
def ethereum(slice_: slice = None):
    @dlt.resource(
        table_name="blocks",
        write_disposition={"disposition": "merge", "strategy": "delete-insert"}
    )
    def data():
        with open(
            "./ethereum.blocks.9c1d9b504ea240a482b007788d5cd61c_2.json",
            "r",
            encoding="utf-8",
        ) as f:
            yield json.load(f) if slice_ is None else json.load(f)[slice_]

      
    data.apply_hints(
        path=["transactions"],
        write_disposition="merge",
        additional_table_hints={"x-merge-strategy": "delete-insert"},
        table_format=None,
    )

    data.apply_hints(
        path=["transactions", "logs"],
        write_disposition="merge",
        additional_table_hints={"x-merge-strategy": "delete-insert"},
        table_format=None,
    )
    
    return data
```

## Problem
Currently, inspecting the source via `ethereum().schema` gives different results. The Current Test shows the hints on the `blocks__transactions` and the `blocks__transactions__logs` tables. It is not the case for the Target API. 

When in the lifecycle should the hints be assigned? After `.apply_hints()`? After source creation? After first pipeline run?

### Value changes on comparison
When running the pipeline with the Target API, I encounter a very odd error
```
  File "/home/tjean/projects/dlthub/dlt/dlt/extract/extract.py", line 353, in _extract_single_source
    extractors[item_format].write_items(
  File "/home/tjean/projects/dlthub/dlt/dlt/extract/extractors.py", line 140, in write_items
    self._write_to_static_table(resource, table_name, items, meta)
  File "/home/tjean/projects/dlthub/dlt/dlt/extract/extractors.py", line 223, in _write_to_static_table
    items = self._compute_and_update_table(resource, table_name, items, meta)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tjean/projects/dlthub/dlt/dlt/extract/extractors.py", line 263, in _compute_and_update_table
    diff_table = utils.diff_table(self.schema.name, existing_table, computed_table)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tjean/projects/dlthub/dlt/dlt/common/schema/utils.py", line 461, in diff_table
    ensure_compatible_tables(schema_name, tab_a, tab_b, ensure_columns=False)
  File "/home/tjean/projects/dlthub/dlt/dlt/common/schema/utils.py", line 529, in ensure_compatible_tables
    raise TablePropertiesConflictException(
dlt.common.schema.exceptions.TablePropertiesConflictException: In schema: ethereum: Cannot merge partial tables into table `blocks__transactions` due to property `parent` with different values: "None" != "blocks"
```
When debugging the call to `ensure_compatible_tables()`, the values magically change before and after the comparison operator (see comments).

```python
def ensure_compatible_tables(
    schema_name: str, tab_a: TTableSchema, tab_b: TPartialTableSchema, ensure_columns: bool = True
) -> None:
    """Ensures that `tab_a` and `tab_b` can be merged without conflicts. Conflicts are detected when

    - tables have different names
    - nested tables have different parents
    - tables have any column with incompatible types

    Note: all the identifiers must be already normalized
    """
    if tab_a["name"] != tab_b["name"]:
        raise TablePropertiesConflictException(
            schema_name, tab_a["name"], "name", tab_a["name"], tab_b["name"]
        )
    table_name = tab_a["name"]
    # check if table properties can be merged
    # breakpoint()
    # tab_a.get("parent"): None
    # tab_b.get("parent"): None 
    if tab_a.get("parent") != tab_b.get("parent"):
        # breakpoint()
        # tab_a.get("parent"): None
        # tab_b.get("parent"): "blocks"
        raise TablePropertiesConflictException(
            schema_name, table_name, "parent", tab_a.get("parent"), tab_b.get("parent")
        )

    if not ensure_columns:
        return
```

dlt/extract/resource.py
```diff
@@ -466,7 +468,9 @@ def _set_hints(
             if table_schema_template.get("validator") is not None:
                 self.validator = table_schema_template["validator"]
 
-    def compute_table_schema(self, item: TDataItem = None, meta: Any = None) -> TTableSchema:
+    def compute_table_schema(
+        self, item: TDataItem = None, meta: Any = None, path: Tuple[str, ...] = None
```
rudolfix, line 472: we do not need to pass `path` here

rudolfix:
> Working on the requested documentation (https://github.com/dlt-hub/dlt/issues/1647), I encounter many edge cases of the current implementation. My understanding is that our goal is for these two snippets to be equivalent

right!
> When in the lifecycle should the hints be assigned? After .apply_hints()? After source creation? After first pipeline run?
1. the `schema` property of `source` is the input schema that got passed in decorator. no hints are applied to it
2. "Current test" does a hack and modifies it directly. That's why you see it, but it is not how it should work
3. To get a schema with applied hints you must call `source.discover_schema()`. You'll get a schema clone with all hints applied to it. Note: now I see that `compute_table_chain` is not used there. please fix it if you want to see all your tables in schema clone.
4. Extract phase uses source input schema, applies all hints to it and then merges changes into existing pipeline schema (from the previous runs). During extract - this all happens on clones, which get comitted to storage when `extract` method exists. TLDR;> you can get `pipeline.default_schema` after `extract` or full `run` and you should see all the changes
5. please commit code that raises this weird bug. I'll take  look. The "current test" is passing now.

this is pretty deep PR. we'll take time to iron this out. I still have some issues with user interface (ie. @dlt.resource should accept a list of nested resources so users do not need to do many `apply_hints`). we'll cover it at the end

dlt/common/schema/typing.py
```diff
@@ -262,6 +262,8 @@ class TTableReference(TypedDict):
 
 class _TTableSchemaBase(TTableProcessingHints, total=False):
     name: Optional[str]
+    path: Optional[Tuple[str, ...]]
```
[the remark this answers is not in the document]
zilto: The `path` kwarg is used in `DltResourceHints.apply_hints()` where it triggers `__getitem__` and `__setitem__` on the newly introduced `DltResourceHintsDict` (all in `hints.py`).

This is where the path is walked a first time and hints are stored. At that time, the `parent_table_name` is not set on `DltResourceHints`. Then, when generating the schema (at extract), we walk over stored hints.

## Illustration
```python
   data = [
            {
                "id": 1,
                "outer1": [
                    {"outer1_id": "2", "innerfoo": [{"innerfoo_id": "3"}]},
                ],
                "outer2": [
                    {"outer2_id": "4", "innerbar": [{"innerbar_id": "5"}]},
                ],
            }
    ]
    nested_resource = dlt.resource(data, name="nested_data")

    nested_resource.apply_hints(
        path=["outer1"],
        columns=[
            {"name": "outer1_id", "data_type": outer1_id_new_type}
        ],
    )
    nested_resource.apply_hints(
        path=["outer2", "innerbar"],
        columns=[
            {"name": "innerbar_id", "data_type": outer2_innerbar_id_new_type},
        ],
    )
```

The implementation implies:
- hints are applied to a single "sub path" or "nested table" at a time. Similarly, current hints are applied at a single level (the top level).
- requires to call `.apply_hints()` for each path to add hints to
- `path` specification doesn't depend on naming convention / normalization
Limitations:
- if the structure changes, the hints may fail "silently" (if structure changes, you should double check)
- related nested fields may be processed differently  e.g., should `receiver > address` have the same processing as `sender > address`

(I'm inheriting code, so I may not have the full context)

## Hints structure
We can think about this problem with how we represent hints. 

Currently, it's a flat mapping of top level keys AFAIK
```
{
  id: {name: id, data_type: text},
  outer1: {name: outer1, data_type: json},
  outer2: {name: outer2, data_type: json},
}
```

The PR uses a flat mapping with tuple keys to uniquely identify a subpath. This avoids recursion challenges
```
{
  id: {...},
  outer1: {...},
  outer2: {...},
  (outer1, ): {columns: {name: outer1_id, data_type: double}},
  (outer2, innerbar): {columns: {name: innerbar_id, data_type: bigint}},
}
```

## Suggestion: implicit path
What would the user API look like without the `path` arg? Using the `parent/table_parent_name` hint would be equivalent to specifying the last part of a `path`

```python
    nested_resource.apply_hints(
        table_parent_name="outer1",
        columns=[
            {"name": "outer1_id", "data_type": outer1_id_new_type}
        ],
    )
    nested_resource.apply_hints(
        table_parent_name="innerbar", 
        columns=[
            {"name": "innerbar_id", "data_type": outer2_innerbar_id_new_type},
        ],
    )
```
- two subpaths could have name collisions for `table_parent_name` 
- How does that work for deeper nesting?
- how does it related to naming convention / normalization? specifying `outer2__innerbar` seems worse than `[outer2, innerbar]`

Then, we would have to implicitly match fields stored in a flat map. We would have to merge parent and child hints because of name collision.

```
{
  id: {...},
  outer1:  {name: outer1, data_type: json},  # clashes with the nested table key
  outer2: {...},
  outer1: {columns: {name: outer1_id, data_type: double}},  # key clashes with the top-level key
  # becomes
  outer1: {name: outer1, data_type: json, nested: {columns: 
  innerbar_id: {name: innerbar_id, data_type: bigint}
}
```
Notes:
- allows bulk operations
- could be beneficial to ensure same hints for related fields
- doesn't work for other hints than `columns`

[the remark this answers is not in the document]
rudolfix: Here I was just commenting on `_TTableSchemaBase` which is a dictionary type that represents table schema: both root and nested tables. There we can just remove `path`. We do not need to modify this type in this PR.

We are discussing are resource hints. Resource may generate a root table (or many if table name is dynamic) and many nested tables. Resource hints will be applied to schemas of generated tables. So, of course, resource and table hints are similar concepts.

anyway, I'm for getting rid of `path` in `apply_hints`, Instead I'd accept:
```
nested_hints: Optional[Dict[Union[str, Tuple[str,...]], "TResourceHintsBase"]]
```
which is a dictionary off all nested hints **non recursive** so it just contains all nested paths.

We could also go for a recursive version, I'm not sure what's better for an user. A path as a tuple may not be very intuitive. But allows to skip certain nesting levels where we do not want to set anything.

The internal implementation will be pretty similar IMO. We may not need `DltResourceHintsDict` and `walk` method because user would provide the dict and there's no recursion.

* you can apply hint to any nesting level in single call
* we can reuse the same argument type in `dlt.resource` decorator (calls `apply_hints` internally)
* same update rules apply to hints on any nesting level (also code reuse)
* we already have a convenience method `make_hints` that creates `TResourceHints`, we could provide `make_nested_hints` that does not take `incremental` and some other arguments that cannot be present in nested tables, 

WDYT?

btw. `parent_table_name` won't be needed anymore in `apply_hints` and we'll deprecate it. it was never working as intended anyway. but

rudolfix:
@zilto I've fixed a few things in the normalizer and I've made a few changes, not sure that is enough for you to take over and finish it. pls take a look and decide. what happened:
- relational normalizer is able to break nested tables chain if it sees a potentially nested table that does not define a parent hint
- this creates several "root" tables coming from a single resource
- I keep nested hints on the ResourceHints (but not on a ResourceNesterHints - so it is not recursive) so we can use table variants and dynamic nested hints
- biggest change when creating tables from nested hints: "resource" and "parent" hints are added conditionally. we break the nested chain if nested table specifies primary/merge key
- I also disabled creating intermediate nested tables... you can restore it if you want
- now the ethereum tests are passing: transactions and logs tables are now top level (root) tables and they are properly merged - because the primary key is defined on them

dlt/extract/extractors.py
```diff
@@ -220,63 +220,80 @@ def _write_to_static_table(
...
-            computed_table["x-normalizer"] = {"evolve-columns-once": True}
-        existing_table = self.schema.tables.get(table_name, None)
-        if existing_table:
-            # TODO: revise this. computed table should overwrite certain hints (ie. primary and merge keys) completely
-            diff_table = utils.diff_table(self.schema.name, existing_table, computed_table)
-        else:
-            diff_table = computed_table
+        computed_tables = self._compute_tables(resource, items, meta)
+        # overwrite root table name (if coming from meta)
+        # TODO: remove below, remove root_table_name from arguments
```
zilto, line 255: Is the commented out code is ready to be removed?

dlt/extract/extractors.py
```diff
@@ -220,63 +220,80 @@ def _write_to_static_table(
...
+                self.schema.resolve_contract_settings_for_table(table_name, computed_table),
+            )
 
-        # apply contracts
-        diff_table, filters = self.schema.apply_schema_contract(
-            schema_contract, diff_table, data_item=items
-        )
+            # this is a new table so allow evolve once
+            if schema_contract["columns"] != "evolve" and self.schema.is_new_table(table_name):
+                computed_table["x-normalizer"] = {"evolve-columns-once": True}
```
zilto, line 269: sidenote: Reading through the codebase `"evolve-columns-once": True` is set *after* the table sees data for the first time. Maybe the name `"columns-evolved-once"` would better communicate that.

tests/common/normalizers/test_json_relational.py
```diff
@@ -748,38 +748,43 @@ def test_table_name_meta_normalized() -> None:
 def test_row_id_is_primary_key() -> None:
     # TODO: if there's a column with row_id hint and primary_key, it should get propagated
     pass
```
zilto, line 750: remove `pass` and `TODO` comment

tests/extract/test_decorators.py
```diff
@@ -320,27 +321,70 @@ def get_users():
...
+            "name": "campagins__steps__budgets",
+            "resource": "campaigns",
+        },
     ]
 
+    table_schemas = campaigns().compute_nested_table_schemas("campagins", Schema("default").naming)
+    assert table_schemas == expected_table_schemas
+
+    @dlt.transformer(nested_hints=nested_hints)
+    def campaign_detials(campaing):
```
zilto, line 378: typo `detials -> details` in several places

tests/extract/test_decorators.py
```diff
@@ -320,27 +321,70 @@ def get_users():
...
+        },
+        {
+            "table_format": "iceberg",
+            "columns": {"budget_id": {"name": "budget_id", "nullable": False, "primary_key": True}},
+            "name": "campagins__steps__budgets",
+            "resource": "campaigns",
+        },
     ]
 
+    table_schemas = campaigns().compute_nested_table_schemas("campagins", Schema("default").naming)
```
zilto, line 374: typo `campagins -> campaigns` in several places

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -88,40 +89,39 @@ def test_merge_on_keys_in_schema(
...
-        return data
+        return blocks()
 
     # take only the first block. the first block does not have uncles so this table should not be created and merged
     info = p.run(
         ethereum(slice(1)),
         **destination_config.run_kwargs,
     )
     assert_load_info(info)
+    print(p.default_schema.to_pretty_yaml())
```
zilto, line 124: remove `print()`

tests/pipeline/test_pipeline.py
```diff
@@ -3254,3 +3255,250 @@ def gen2():
...
+    normalize_info = p.normalize()
+    load_id = normalize_info.loads_ids[0]
+    norm_metrics = normalize_info.metrics[load_id][0]
+    for file_name, _ in norm_metrics["job_metrics"].items():
+        # always jsonl
+        assert file_name.endswith("jsonl")
+
+
+def test_nested_hints_write_disposition_append_merge() -> None:
+    # we can mix replace and append write disposition in nested tables
```
zilto, line 3283: design comment: do we want to allow this behavior? My main concern is backward compatibility once people hack around this feature. Maybe `pipeline.run()` can log a warning `WriteDispositionMismatch` or `UndefinedWriteDisposition` with message that says behavior might change in the future.

I would expect the default/intended behavior to be that the nested table always follows the root table behavior

zilto (commented):
I reviewed all files. I believe there are some typos and dev comments to remove.

I'm *very* excited about this feature. Awesome work!

tests/pipeline/test_pipeline.py
rudolfix: done! also test added that defines same write disposition and root table (which is allowed)

rudolfix (commented):
that was a good review thx! stuff fixed

zilto (approved):
All reviews were addressed. Looks good to me

rudolfix:
> All reviews were addressed. Looks good to me.
> Just fix failing tests. Seems related to comparing lists and item ordering. Could use sets instead

ah yeah I know exactly what happens here. this must pass. it checks the inferred column order...

PR #2168, opened 2024-12-20
[Experiment] extending features with addition pip package

sh-rp:
### Description

This example shows how to implement a basic version of a feature in one package and extending it in another package making it possible for the base lib to load the extended version if present. We use a factory for this and this will also allow us to upgrade typing to a new version if a certain lib is present.


sh-rp:
closing at this was just a poc

Issue #2169, opened 2024-12-20
Fix unexpected behavior when cursor fields are non-UTC datetime when using lag


hairrrrr:
### dlt version

1.5.0

### Describe the problem

Currently, if `lag` is applied to `incremental`, non-UTC datatime will be transformed to UTC datatime. If the `initial_value` is not a UTC datetime, `lagged_last_value` may be wrong in the string comparison (`max` or`min`).

More importantly, if the cursor field value is a non-UTC datatime, it will compare to a UTC `lagged_last_value`, which is `start_value` and `last_value` in `IncrementalTransform`.

Cause lag may modify the last_value in `IncrementalTransform`, it is not safe to write back state if rows are empty.

`datatime.strftime`(below Python 3.12.0) can not preserve the colon in the timezone when converting datetime objects back to strings. This may also cause an unexpected behavior:

```python
printf(max("2024-10-20T15:30:00-00:00", "2024-10-20T15:30:00-0030"))
# 2024-10-20T15:30:00-00:00 will be printed
```

I have added tests that can reproduce all the above cases, Let me know if there is anything to improve in the code.

### Steps to reproduce

The tests can cover all the cases.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

PR #2170, opened 2024-12-20
bug: support lag for non-UTC datetime cursor fields

hairrrrr:
This PR:

- Add `ensure_pendulum_datetime_non_utc` to parse datetime strings into non-UTC datetime objects.
- Add `_datetime_obj_to_str` to preserve the colon in the timezone when converting datetime objects back to strings.
- Skip writing back state if no valid rows are found for `last_value` in the transformer, which may otherwise cause incorrect behavior.

Fix issure: https://github.com/dlt-hub/dlt/issues/2169

Please let me know if there are any problems with my code. Thank you!

### DIFF
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -1,6 +1,7 @@
 import os
 from datetime import datetime  # noqa: I251
 from typing import Generic, ClassVar, Any, Optional, Type, Dict, Union, Literal, Tuple
+
 from typing_extensions import get_args
 
 import inspect
@@ -560,8 +561,14 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
         else:
             rows = self._transform_item(transformer, rows)
 
-        # write back state
+        # ensure last_value maintains forward-only progression when lag is applied
+        if self.lag and (cached_last_value := self._cached_state.get("last_value")):
+            transformer.last_value = self.last_value_func(
+                (transformer.last_value, cached_last_value)
+            )
+        # writing back state
         self._cached_state["last_value"] = transformer.last_value
+
         if not transformer.deduplication_disabled:
             # compute hashes for new last rows
             unique_hashes = set(
--- a/dlt/common/time.py
+++ b/dlt/common/time.py
@@ -1,6 +1,7 @@
 import contextlib
 import datetime  # noqa: I251
 import re
+import sys
 from typing import Any, Optional, Union, overload, TypeVar, Callable  # noqa
 
 from pendulum.parsing import (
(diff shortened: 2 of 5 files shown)


dlt/extract/incremental/__init__.py
```diff
@@ -560,8 +561,14 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
         else:
             rows = self._transform_item(transformer, rows)
 
-        # write back state
-        self._cached_state["last_value"] = transformer.last_value
+        # Skip writing back state if no valid rows for last_value in transformer may be wrong
+        if rows is not None:
```
rudolfix, line 565: this is really a good catch but the problem is deeper than this. we cannot let the state go back and indeed when there's a lag this can unfortunately happen (like in your test, where the "last value" record is missing in the second read).

so IMO this code will fix it
```py
         if self.lag and "last_value" in self._cached_state:
            transformer.last_value = self.last_value_func(transformer.last_value, self._cached_state["last_value"])
        self._cached_state["last_value"] = transformer.last_value
```
it uses last_value_func to make sure that "highest" last_value is preserved ("highest" per order established by the last_value_func)

`rows` is of TDataItems type: so it may be a single item or a list of items. items may be dicts, panda frames, arrow tables or batches.

so for example rows may be a pandas frame or a list of pandas frames. that's way the checks below are not sufficient. but as I said the problem is deeper (but solution is at the end simpler)

rudolfix (changes requested):
thanks for catching edge cases for the lag! the problem you've found is deeper than just empty results. please see my comments.

dlt/extract/incremental/__init__.py
hairrrrr: Thanks, @rudolfix! I think this modification aligns well with the dlt code style. As you mentioned, there might be deeper issues that I’m not aware of at the moment.

>so for example rows may be a pandas frame or a list of pandas frames. that's way the checks below are not sufficient. but as I said the problem is deeper (but solution is at the end simpler)

I have a question. If rows is an empty list of pandas DataFrames, I think the code `rows != []` will handle this condition. This is because of the following code:

```
        if isinstance(rows, list):
            rows = [
                item
                for item in (self._transform_item(transformer, row) for row in rows)
                if item is not None
            ]
        else:
            rows = self._transform_item(transformer, rows)
```

hairrrrr:
Hi, @rudolfix. I have resovled all the changes you ask! Please review again. Thank you!

rudolfix (approved):
LGTM!

jrhorne:
Apologies if this is the wrong place, but was reading through the threads on timezones for the lag parameter.

I'm trying to add a lag to an incremental loading that did not previously have a lag. So I changed:

``` 
  s3_alerts = filesystem(
      bucket_url=f"s3://{s3_bucket_name}/",
      file_glob=alert_glob_expression,
      extract_content=False,
      files_per_page=10000
  )

s3_alerts.apply_hints(incremental=dlt.sources.incremental("modification_date"))
```

to now have the hints be:
` s3_alerts.apply_hints(incremental=dlt.sources.incremental("modification_date", lag=3*24*60*60))`


Is this the correct way to do this for the s3 filesystem type? I'm now getting this traceback

``` 
File "/app/.venv/lib/python3.10/site-packages/dlt/extract/incremental/__init__.py", line 378, in last_value
    last_value = apply_lag(
  File "/app/.venv/lib/python3.10/site-packages/dlt/extract/incremental/lag.py", line 70, in apply_lag
    and last_value_func((initial_value, lagged_last_value)) == initial_value
TypeError: can't compare offset-naive and offset-aware datetimes
```

hairrrrr:
Hi @jrhorne,

This seems to be another timezone-related issue. Have you solved this problem? If not, perhaps you could create an issue with a simple reproduction example? I'll be happy to help if I can.

PR #2172, opened 2024-12-22
fix: unravel Optional to inner generic arg from instance

### LINKED ISSUE #2171: rest_api_resources exit early when joining external schedulers
### dlt version

1.5.0

### Describe the problem

Dynamically created rest_api resources from a `RestApiConfig` having an Incremental configuration exit prematurely when "joining" an external scheduler.  This seems to prohibit resources created through the `rest_api_resources` factory function from being able to use Airflow start/end inference for incremental parameters.

### Expected behavior

I would expect that resources created through `rest_api_resources` would behave similarly to resources defined by the more traditional `dlt.resource` decorator with respect to `allow_external_schedulers`.

The warning that I see in the logs is always:
```
2024-12-21 19:41:12,788|[WARNING]|188923|134449741906688|dlt|__init__.py|_join_external_scheduler:414|Specified Incremental last value type dlt.extract.incremental.Incremental[typing.Any] is not supported. Please use DateTime, Date, float, int or str to join external schedulers.(issubclass() arg 1 must be a class)
```

### Steps to reproduce

Here is an example test-case hitting an endpoint on the Gemini Rest API:

```python
from pathlib import Path
import os
from datetime import datetime
import dlt
from dlt.sources.rest_api import (
    rest_api_resources,
    RESTAPIConfig,
    ClientConfig,
    EndpointResource,
    Endpoint,
)
from dlt.sources.rest_api.typing import IncrementalConfig

DEST_PATH = Path.home() / "example"
os.environ["DESTINATION__FILESYSTEM__BUCKET_URL"] = f"file:///{DEST_PATH}"
os.environ["NORMALIZE__DATA_WRITER__DISABLE_COMPRESSION"] = "true"


@dlt.source
def gemini_btc():
    config = RESTAPIConfig(
        client=ClientConfig(base_url="https://api.gemini.com/v1"),
        resources=[
            EndpointResource(
                name="btcusd_trades",
                write_disposition="append",
                endpoint=Endpoint(
                    path="trades/btcusd",
                    incremental=IncrementalConfig(
                        allow_external_schedulers=True,
                        cursor_path="timestamp",
                        initial_value=datetime(2024, 12, 21).timestamp(),
                        start_param="timestamp",
                    ),
                ),
            ),
        ],
    )
    return rest_api_resources(config)


if __name__ == "__main__":
    pipeline = dlt.pipeline(
        pipeline_name="gemini_btc",
        dataset_name="gemini_btc",
        destination="filesystem",
    )
    pipeline.run(gemini_btc())
```

When I step through the pipeline run, I'll get to `dlt.common.typing.get_generic_type_argument_from_instance` which seems to handle cases where the `instance.__orig_class__` is an Incremental generic, but not an Optional[Incremental[T]], which is the actual type at this point.

### Operating system

Linux

### Runtime environment

Astronomer

### Python version

3.11

### dlt data source

Rest API

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

otosky:
Addresses resources created by the `rest_api_resources` factory function exiting prematurely when trying to _join_external_scheduler and not inferring start/end dates from Airflow.

### Related Issues

Closes #2171

### DIFF
--- a/dlt/common/typing.py
+++ b/dlt/common/typing.py
@@ -438,9 +438,8 @@ def get_generic_type_argument_from_instance(
     """
     orig_param_type = Any
     if cls_ := getattr(instance, "__orig_class__", None):
-        # unfurl Optional[Incremental[...]] to Incremental[...]
         if is_optional_type(cls_):
-            cls_ = get_args(cls_)[0]
+            cls_ = extract_inner_type(cls_)
 
         # instance of generic class
         pass
--- a/dlt/common/typing.py
+++ b/dlt/common/typing.py
@@ -438,11 +438,7 @@ def get_generic_type_argument_from_instance(
     """
     orig_param_type = Any
     if cls_ := getattr(instance, "__orig_class__", None):
-        if is_optional_type(cls_):
-            cls_ = extract_inner_type(cls_)
-
-        # instance of generic class
-        pass
+        cls_ = extract_inner_type(cls_)
     elif bases_ := get_original_bases(instance.__class__):
         # instance of class deriving from generic
         cls_ = bases_[0]
--- a/dlt/common/typing.py
+++ b/dlt/common/typing.py
@@ -438,6 +438,10 @@ def get_generic_type_argument_from_instance(
     """
     orig_param_type = Any
     if cls_ := getattr(instance, "__orig_class__", None):
+        # unfurl Optional[Incremental[...]] to Incremental[...]
+        if is_optional_type(cls_):
+            cls_ = get_args(cls_)[0]
+
         # instance of generic class
         pass
     elif bases_ := get_original_bases(instance.__class__):
--- a/tests/common/test_typing.py
+++ b/tests/common/test_typing.py
@@ -7,6 +7,7 @@ from typing import (
     Callable,
     ClassVar,
     Final,
+    Generic,
     List,
     Literal,
     Mapping,
@@ -48,7 +49,6 @@ from dlt.common.typing import (
     add_value_to_literal,
     get_generic_type_argument_from_instance,
 )
-from dlt.extract import Incremental
 
 
 class TTestTyDi(TypedDict):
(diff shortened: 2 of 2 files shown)


otosky:
This could also potentially be moved up the call stack into `IncrementalResourceWrapper.wrap`, such that `__orig_class__` is the actual `Incremental` type annotation and not wrapped in an `Optional`?

https://github.com/dlt-hub/dlt/blob/5a1cb69a31ffbe82cf0f00fb65b94e05e7310763/dlt/extract/incremental/__init__.py#L714

dlt/common/typing.py
```diff
@@ -438,6 +438,10 @@ def get_generic_type_argument_from_instance(
     """
     orig_param_type = Any
     if cls_ := getattr(instance, "__orig_class__", None):
+        # unfurl Optional[Incremental[...]] to Incremental[...]
+        if is_optional_type(cls_):
```
rudolfix, line 442: you could also look at `extract_inner_type` in `common/typing.py`. it will unfurl other wrappers (ie. Annotated) and also nested cases

tests/common/test_typing.py
```diff
@@ -310,3 +314,17 @@ def test_add_value_to_literal() -> None:
     add_value_to_literal(TestSingleLiteral, "green")
     add_value_to_literal(TestSingleLiteral, "blue")
     assert get_args(TestSingleLiteral) == ("red", "green", "blue")
+
+
+def test_get_generic_type_argument_from_instance() -> None:
+    # generic contains hint
+    instance = SimpleNamespace(__orig_class__=Incremental[str])
```
rudolfix, line 321: we are trying to keep the common module as the upstream dependency of all other top level modules. so it is better to not import from `extract` here. You could just use any standard generic type for testing

rudolfix (changes requested):
thanks for the fix and writing the test! pls see the review comments

tests/common/test_typing.py
otosky: Great point - updated in a4e5c11

dlt/common/typing.py
otosky: Nice! Got rid of the conditional check and just used `extract_inner_type` instead.

rudolfix (approved):
LGTM!

PR #2183, opened 2025-01-01
adds destination registry

rudolfix:
### Description
* adds destination registry
* splits reference.py into many files
* puts PipelineConfiguration into (pipelines, <pipeline name>) section
* allows to auto import sources


joscha:
This sounds great! Is there a design doc around this somewhere, outlining how it would work? Or a sample destination which uses a remote source (like a gh repo)?

rudolfix:
> This sounds great! Is there a design doc around this somewhere, outlining how it would work? Or a sample destination which uses a remote source (like a gh repo)?

this is a mechanism that will discover additional sources and destinations provided via dlt plugins. here's a followup PR where we implement that fully https://github.com/dlt-hub/dlt/pull/2236

basically a plugin (we use pluggy) registers itself in the search path for sources and destinations and then user can use those references everywhere we have `destination=""` . also the PR above allows to pass source name to dlt.run("source_name") instead of an instance... this is not documented because I do not consider this stable API.

we are doing a bigger documentation push, where those things get explained...

rudolfix:
moved to #2236

PR #2184, opened 2025-01-01
unifies handling of naive datetime for arrow and object incrementals

rudolfix:
### Description
unifies handling of naive datetime for arrow and object incrementals


sh-rp:
Maybe one note: if older versions of connector-x depended on this, maybe we should bump the min version? Or add a warning in the docs? I don't know how serious this problem was before.

rudolfix:
@sh-rp it is already bumped

PR #2185, opened 2025-01-02
Force use of typeddict from typingexentions, pin poetry in tests, simple disable python 3.8

sh-rp:
### Description
See attached ticket for description of what was fixed.

### DIFF
--- a/dlt/common/typing.py
+++ b/dlt/common/typing.py
@@ -45,6 +45,11 @@ from typing_extensions import (
 
 from typing_extensions import is_typeddict as _is_typeddict
 
+if sys.version_info >= (3, 12):
+    from typing import TypedDict  # noqa: I251
+else:
+    from typing_extensions import TypedDict  # noqa: I251
+
 try:
     from types import UnionType  # type: ignore[attr-defined]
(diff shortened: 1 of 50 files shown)


dlt/common/typing.py
```diff
@@ -45,6 +45,11 @@
 
 from typing_extensions import is_typeddict as _is_typeddict
 
+if sys.version_info >= (3, 12):
```
rudolfix, line 48: IMO it is safe to always import from typing_extensions. I do that ie. for Annotated. should we change this?

tox.ini
```diff
@@ -6,6 +6,8 @@ banned-modules = datetime = use dlt.common.pendulum
                  decimal.Decimal = use dlt.common.Decimal
                  open = use dlt.common.open
                  pendulum = use dlt.common.pendulum
+                 typing.TypedDict = use dlt.common.typing.TypedDict
```
rudolfix, line 9: this is cool!

rudolfix (changes requested):

rudolfix (approved):
LGTM!

PR #2047, opened 2024-11-10
[WIP] enables python 3.13

rudolfix:

PR #2194, opened 2025-01-07
drop python 3.8, enable python 3.13, and enable full linting for 3.12

sh-rp:
### Description

This PR does the following:
* Drop support for python3.8, also removes all special code and dependencies we added for pyhton3.8
* Enable linting for python3.12, all extras are now supported on python3.12
* Enable python3.13 and run common tests there. Use dlt-pendulum for now.

Based on https://github.com/dlt-hub/dlt/pull/2047

ToDo:
* Should dlt-pendulum be on our private pypi?


pyproject.toml
```diff
@@ -24,9 +29,13 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.8.1,<3.13"
+python = ">=3.9,<3.14"
 requests = ">=2.26.0"
-pendulum = ">=2.1.2"
+# TODO: pin to tag on dlt-pendulum repo
+dlt-pendulum = [
```
sh-rp, line 35: NOTE: this still requires cargo at this point

sh-rp: @rudolfix I have another question about this: should we not have a different package name for our pendulum? The way it is currently set up, dlt-pendulum is always used, also for python versions before 3.13, this means there will be package name collissions if users also have the regular pendulum in their dependencies already in a project. I'm not sure what happens in this case.

Also you mentioned that you made some changes they will not like and there is a change in the code where you removed the casting of pendulum.now() to a string, are the libraries now incompatible?

sh-rp: We are now using a properly build dlt-pendulum for python 3.13, so that part is solved. The tests will verify wether this string removal still works or not.

.github/workflows/lint.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-        python-version: ["3.8.x", "3.9.x", "3.10.x", "3.11.x"]
+        python-version: ["3.9.x", "3.10.x", "3.11.x", "3.12.x"]
```
sh-rp, line 29: make dev works for 3.12 now!

rudolfix (approved):
LGTM! just resolve all the conflicts...

deploy/dlt/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM alpine:3.15
+FROM python:3.11.11-alpine
```
sh-rp, line 1: Note: switched to a python image, I can probably revert this if we want to use a vanilla alpine here.

deploy/dlt/Dockerfile
```diff
@@ -17,16 +17,28 @@ ADD compiled_requirements.txt .
...
     rm get-pip.py &&\
-    pip3 install --upgrade setuptools wheel &&\
-    rm -r /usr/lib/python*/ensurepip &&\
-    pip3 install -r compiled_requirements.txt &&\
+    pip install --upgrade setuptools wheel pip &&\
+    rm -r /usr/lib/python*/ensurepip
+
+# install arrow 17.0.0, usually we would need apache-arrow-dev=17.0.0 but it is not available in alpine 3.20
+# adapt this version to the arrow version you need
+RUN git clone --no-checkout https://github.com/apache/arrow.git /arrow \
```
sh-rp, line 29: as noted in the comment, alpine always seems to come with one version of each package, i remember something like this from our last company, so I need to build arrow 17 here if I want to use it..

Makefile
```diff
@@ -102,11 +102,14 @@ publish-library: build-library
 	poetry publish
 
 test-build-images: build-library
-	# TODO: enable when we can remove special duckdb setting for python 3.12
+	# NOTE: poetry export does not work with our many different deps, we install a subset and freeze
 	# poetry export -f requirements.txt --output _gen_requirements.txt --without-hashes --extras gcp --extras redshift
-	# grep `cat compiled_packages.txt` _gen_requirements.txt > compiled_requirements.txt
+	poetry install --no-interaction -E gcp -E redshift -E duckdb
+	poetry run pip freeze > _gen_requirements.txt
```
sh-rp, line 108: this works well, poetry export was failing on several dependencies..

PR #2197, opened 2025-01-08
use a dummy operator at the start of parallel pipelines

### LINKED ISSUE #2196: Parallelize all sources in Airflow, including the first one
### Feature description

I would like to parallelize all sources in Airflow, including the first one in the list, PR incoming.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

The first source in a pipeline can take a long time to complete, right now in both parallel modes the first source is ran alone, this is time that could be spent running multiple sources.

### Proposed solution

Using a `DummyOperator` at the start, instead of splitting the first source and branching after it has run.

### Related issues

_No response_

alucryd:
### Description
Used a `DummyOperator` instead of the first source to parallelize all sources, including the first one.

### Related Issues

- Fixes #2196 

### Additional Context
The first source can take a long time to run, this can make pipelines faster by parallelizing even the first source.

### DIFF
--- a/dlt/helpers/airflow_helper.py
+++ b/dlt/helpers/airflow_helper.py
@@ -448,10 +448,10 @@ class PipelineTasksGroup(TaskGroup):
                 tasks = []
                 sources = data.decompose("scc")
                 t_name = self._task_name(pipeline, data)
-                start = make_task(pipeline, sources[0])
+                start = DummyOperator(task_id=f"{t_name}_start")
 
                 # parallel tasks
-                for source in sources[1:]:
+                for source in sources:
                     for resource in source.resources.values():
                         if resource.incremental:
                             logger.warn(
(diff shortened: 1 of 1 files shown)


rudolfix (changes requested):
@alucryd there's a reason to run first task and then all others in parallel: it will create initial schema in the database and standard dlt tables. all tasks share the same dataset.

if you still want to work on this PR then let's add new option to `add_run`: ie `dummy_task_first` and if set to True, do what do right now.

I do not want to change existing behavior, too many deployments that may rely on that are in production

alucryd:
> @alucryd there's a reason to run first task and then all others in parallel: it will create initial schema in the database and standard dlt tables. all tasks share the same dataset.
> 
> if you still want to work on this PR then let's add new option to `add_run`: ie `dummy_task_first` and if set to True, do what do right now.
> 
> I do not want to change existing behavior, too many deployments that may rely on that are in production

I see, thanks for the heads up, I don't have the full picture yet but I'm getting there. I ran this change in production and didn't run into any issue with a completely new datasource so I wrongly assumed it would be harmless.

I assume it would be too much work to split the schema and table creations and only run that in the first task?

In any case I'll add the proposed option and default it to false so it doesn't impact anyone.

sh-rp:
@prakharcode I will close this PR for no activity, if you'd like to continue it or provide a new one, please re-open this or create a new PR. Thanks :)

Issue #2201, opened 2025-01-09
dedup_sort behavior when column value is equal


guitcastro:
### Documentation description

Currently, the docs does not have any indication on what happens when `dedup_sort` column has the same value. I am running the pipeline where the same rows (no changes) are being override. 

What is the expected behavior? 

* New rows (from resource) takes precedence and override rows in destination 
* Rows with the same values might get ordered at random

### Are you a dlt user?

Yes, I'm already a dlt user.

rudolfix:
a followup ticket: https://github.com/dlt-hub/dlt/issues/2279

PR #2202, opened 2025-01-09
Deltalake dedup_sort support

guitcastro:
### Description

Deltalake is ignoring `dedup_sort` column and always overriding rows with matching ids. This PR added support to dedup_sort

### Additional Context

This is my first PR, and I am very unfamiliar with the code base. So, I expected that this PR might need a few interactions of reviews to get it right. Per example, I didn't not find an easy way to test it. I am very open to make changes if someone point's me the right direction.

### DIFF
--- a/dlt/common/libs/deltalake.py
+++ b/dlt/common/libs/deltalake.py
@@ -6,7 +6,7 @@ from dlt.common import logger
 from dlt.common.libs.pyarrow import pyarrow as pa
 from dlt.common.libs.pyarrow import cast_arrow_schema_types
 from dlt.common.schema.typing import TWriteDisposition, TTableSchema
-from dlt.common.schema.utils import get_first_column_name_with_prop, get_columns_names_with_prop
+from dlt.common.schema.utils import get_first_column_name_with_prop, get_columns_names_with_prop, get_dedup_sort_tuple
 from dlt.common.exceptions import MissingDependencyException
 from dlt.common.storages import FilesystemConfiguration
 from dlt.common.utils import assert_min_pkg_version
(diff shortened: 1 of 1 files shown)


rudolfix (commented):
@guitcastro right now `dedup_sort` has no function for `upsert` merge strategy. indeed we do not issue a warning when it is defined. it is just ignored.

we assume that in case of `upsert` the input dataset is deduplicated on primary_key. in that case, the target dataset is deduplicated as well.

we could indeed use `dedup_sort` to deduplicate the source. but your code is doing something else. #2201 does not help to understand your use case.

are you trying to use `dedup_sort` as some kind of record version? and skip the update on rows for which version is not changed?

that is typically done using `Incremental` during extract phase so records that are not changed are simply not present.

guitcastro:
@rudolfix 

The issue was indeed confused. English is not my native language and I opened the issues before digging in the code and didn't not updated it. Sorry about that. 

> we could indeed use dedup_sort to deduplicate the source. but your code is doing something else. https://github.com/dlt-hub/dlt/issues/2201 does not help to understand your use case.

I might have written the code wrong, but my intention is to deduplicate using dedup_sort.  It's working locally and is based on [this example from data brics docs](https://docs.databricks.com/en/delta/merge.html#data-deduplication-when-writing-into-delta-tables):   

```
deltaTable.alias("logs").merge(
    newDedupedLogs.alias("newDedupedLogs"),
    "logs.uniqueId = newDedupedLogs.uniqueId AND logs.date > current_date() - INTERVAL 7 DAYS") \
  .whenNotMatchedInsertAll("newDedupedLogs.date > current_date() - INTERVAL 7 DAYS") \
  .execute()
```

Could you point out what might be wrong with my code?

cheers!

rudolfix:
Your problems is: your source data contains duplicates and you want to get rid of them. correct? This should happen BEFORE you execute merge statement so:

```
def merge_delta_table(
    table: DeltaTable,
    data: Union[pa.Table, pa.RecordBatchReader],
    schema: TTableSchema,
) -> None:
```
you need to deduplicate `data` using pyarrow or duckdb.

current implementation in `dlt` assumes that your source data is deduplicated

rudolfix:
@guitcastro `dedup_sort` is used to deterministically deduplicate the source dataset. what you are asking for is that we support some kind of "version number" to detect changed records and skip unchanged when merging.

this is typically done via incremental loading: you remove all records with `modified_date` older than required in your `dlt.resource`.

still, the idea to filter updated records is pretty good. if you still want to work in this we can let users pass any update expression to the `merge_delta_table`:

1. if you have a dlt.resource you can use `apply_hints` with `additional_table_hints` to add any table level hint
2. make the `merge_delta_table` function to look for ie. `x-merge-update-filter` in `schemsa` that contains the expression you want to add to the update_all
3. if found just pass it to the update_all.
this will allow for any filter expression user may want to be passed.

WDYT? should be even less code than your PR has right now

Issue #2203, opened 2025-01-12
Deploy command fails even with extras installed


jonbiemond:
### dlt version

1.5

### Describe the problem

When running `dlt deploy` the command fails with the message below even though `dlt[cli]` is installed.
```bash
WARNING: Please install additional command line dependencies to use deploy command:
pip install "dlt[cli]"
We ask you to install those dependencies separately to keep our core library small and make it work everywhere.
NOTE: Please refer to our docs at 'https://dlthub.com/docs/walkthroughs/deploy-a-pipeline' for further assistance.
```

This is because `dlt[cli]` extras do not include all the dependencies needed by the deploy command. Namely the `pip` dependency.

### Steps to reproduce

1. Install `dlt[cli]` into a virtual environment without pip. For example using a package manager such as [`uv`](https://docs.astral.sh/uv/).
```bash
uv venv
source .venv/bin/activate
uv pip install "dlt[cli]"
```

2. Run the deploy command.
```bash
dlt deploy pipeline.py
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### Additional information

1. Do not hide the exception and replace it with a custom message. This makes it difficult to debug. Instead, include the exception message with the warning.
2. Include the `pip` dependency in the `cli` extras.

PR #2204, opened 2025-01-12
Display exception message on deploy command failure

jonbiemond:
### Description
When the deploy command fails due to a missing import, output the exception message to the terminal. This way the user has the necessary information to resolve the error.

Previous output:
```bash
WARNING: Please install additional command line dependencies to use deploy command:
pip install "dlt[cli]"
We ask you to install those dependencies separately to keep our core library small and make it work everywhere.
NOTE: Please refer to our docs at 'https://dlthub.com/docs/walkthroughs/deploy-a-pipeline' for further assistance.
```

After fix:
```bash
ERROR: No module named 'pipdeptree'
WARNING: Please install additional command line dependencies to use deploy command:
pip install "dlt[cli]"
We ask you to install those dependencies separately to keep our core library small and make it work everywhere.
NOTE: Please refer to our docs at 'https://dlthub.com/docs/walkthroughs/deploy-a-pipeline' for further assistance.
```

### Related Issues

Fixes (#2203)

### DIFF
--- a/dlt/cli/plugins.py
+++ b/dlt/cli/plugins.py
@@ -32,9 +32,9 @@ try:
         SecretFormats,
     )
 
-    deploy_command_available = True
-except ModuleNotFoundError:
-    deploy_command_available = False
+    deploy_command_import_exception = None
+except ModuleNotFoundError as exc:
+    deploy_command_import_exception = exc
 
 
 @plugins.hookspec()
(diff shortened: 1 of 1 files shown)


sh-rp:
@jonbiemond thanks for your PR. Does installing the "cli" extra not resolve this error? In my view the error message that is there now is correct.

jonbiemond:
@sh-rp It did not for me. You can see the details in the linked issue #2203 
Once #2207 is merged that should also be resolved.

The problem with the current error message is that it hides the underlying exception. This makes it very difficult to debug and resolve.

sh-rp:
Ok, I see. This is rather a bug in the extras (that you thankfully fixed!) than a bug here. I think if we merge your other PR, we do not need this change anymore. Alternatively you could write something like:

```
raise CliCommandException() from deploy_command_import_exception
```

And not print out this exception. Then the full trace will be visible when running the cli with the --debug flag.

PR #2211, opened 2025-01-14
add downloads badge

sh-rp:
### Description
 We now have 1m downloads per month, we should display this as a badge :)


AstrakhantsevaAA (approved):
Looks good!

PR #2213, opened 2025-01-14
#2110 adds several not found error message patterns to sqlalchemy destination

### LINKED ISSUE #2110: SAP HANA SQLAlchemy destination fails to synchronize
### dlt version

1.4.0

### Describe the problem

I'm trying to load data to a SAP HANA database using a SQLAlchemy engine as destination, defining it like so `engine = create_engine('hana+hdbcli://XXXXXX:XXXXXX@12.12.12.12:12345')`. The HANA instance is up and running, and credentials are correct. Source is a simple REST API. This is how I define my pipeline

```
pipeline = dlt.pipeline(
    pipeline_name='hana_load',
    destination=dlt.destinations.sqlalchemy(engine),
    dataset_name='MYSCHEMA',
)

load_info = pipeline.run(source)
```

Immediately, an exception is raised

```
(hdbcli.dbapi.Error) (362, 'invalid schema name: MYSCHEMA: line 2 col 6 (at pos 302)')
[SQL: SELECT myschema._dlt_pipeline_state.version, myschema._dlt_pipeline_state.engine_version, myschema._dlt_pipeline_state.pipeline_name, myschema._dlt_pipeline_state.state, myschema._dlt_pipeline_state.created_at, myschema._dlt_pipeline_state.version_hash, myschema._dlt_pipeline_state._dlt_load_id
FROM myschema._dlt_pipeline_state JOIN myschema._dlt_loads ON myschema._dlt_loads.load_id = myschema._dlt_pipeline_state._dlt_load_id
WHERE myschema._dlt_pipeline_state.pipeline_name = ? AND myschema._dlt_loads.status = ? ORDER BY myschema._dlt_loads.load_id DESC]
[parameters: ('quick_start', 0)]
(Background on this error at: https://sqlalche.me/e/14/dbapi)
```

which is expected since at this point, neither the schema nor the metadata tables have been created. This happens in `destinations/impl/sqlalchemy/sqlalchemy_job_client.py` in `get_stored_state()`. Same thing happens in `_get_stored_schema()`.

Surprisingly, this is not the case with a Postgres SQLAlchemy engine where the exception is not thrown, the pipeline runs without errors and data is loaded in the destination table. It is curious though that the code block within the context manager is never reached.

```
with self.sql_client.execute_query(q) as cur:
    row = cur.fetchone()    # <---- Never reached, even in Postgres case
    if row is None:
        return None

    # TODO: Decode compressed schema str if needed
    return StorageSchemaInfo.from_normalized_mapping(
        row._mapping, self.schema.naming  # type: ignore[attr-defined]
    )
```

Postgres and HANA SQLAlchemy engines use different drivers underneath (`psycopg2` and `hdbcli` respectively) so this might have some relation to the issue. I was able to bypass this by adding a try/except block in `get_stored_state()` and `_get_stored_schema()` but I'm not sure that this is the proper way to solve it. I haven't tested it with other RDBMS destinations (like MySQL).

### Expected behavior

No error thrown, like in the Postgres case or proper exception handling for the cases it throws.

### Steps to reproduce

- Create a SAP HANA and a Postgres SQLAlchemy engine like so 
`hana_engine = create_engine('hana+hdbcli://XXXXXX:XXXXXX@12.12.12.12:12345')`
`postgres_engine = create_engine('postgresql+psycopg2://XXXXXX:XXXXXX@13.13.13.13:5432/some_db')`

- Run this simple pipeline
```
pipeline = dlt.pipeline(
    pipeline_name='hana_load',
    destination=dlt.destinations.sqlalchemy(engine),
    dataset_name='MYSCHEMA',
)

load_info = pipeline.run(source)
```
- Expect an exception raised for the HANA case, but not for Postgres

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

A simple REST API

### dlt destination

SQLAlchemy (SAP HANA & Postgres)

### Other deployment details

_No response_

### Additional information

Adding a try/except block solves the issue but there might be a more elegant approach.

rudolfix:
### Description
There's no SQLAlchemy exception for an unknown entity/relation so we do pattern matching. This PR extends it to most popular dialects

fixes #2110

### DIFF
--- a/tests/load/filesystem/test_filesystem_common.py
+++ b/tests/load/filesystem/test_filesystem_common.py
@@ -157,6 +157,20 @@ def test_glob_files(with_gdrive_buckets_env: str, load_content: bool, glob_filte
     assert_sample_files(all_file_items, filesystem, config, load_content, glob_filter)
 
 
+@pytest.mark.parametrize("load_content", (True, False))
+def test_glob_single_file(with_gdrive_buckets_env: str, load_content: bool) -> None:
+    bucket_url = os.environ["DESTINATION__FILESYSTEM__BUCKET_URL"]
+    bucket_url, config, filesystem = glob_test_setup(
+        bucket_url, "standard_source/samples/met_csv/A803"
+    )
+    # use glob to get data
+    all_file_items = list(glob_files(filesystem, bucket_url, "A803_20230919.csv"))
+    assert len(all_file_items) == 1
+    assert (
+        all_file_items[0]["relative_path"] == all_file_items[0]["file_name"] == "A803_20230919.csv"
+    )
+
+
 def test_glob_overlapping_path_files(with_gdrive_buckets_env: str) -> None:
     bucket_url = os.environ["DESTINATION__FILESYSTEM__BUCKET_URL"]
     # "standard_source/sample" overlaps with a real existing "standard_source/samples". walk operation on azure
--- a/dlt/destinations/impl/sqlalchemy/db_api_client.py
+++ b/dlt/destinations/impl/sqlalchemy/db_api_client.py
@@ -435,17 +435,36 @@ class SqlalchemyClient(SqlClientBase[Connection]):
             return DatabaseUndefinedRelation(e)
         msg = str(e).lower()
         if isinstance(e, (sa.exc.ProgrammingError, sa.exc.OperationalError)):
-            if "exist" in msg:  # TODO: Hack
-                return DatabaseUndefinedRelation(e)
-            elif "unknown table" in msg:
-                return DatabaseUndefinedRelation(e)
-            elif "unknown database" in msg:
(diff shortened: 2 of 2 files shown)


tests/load/filesystem/test_filesystem_common.py
```diff
@@ -157,6 +157,20 @@ def test_glob_files(with_gdrive_buckets_env: str, load_content: bool, glob_filte
     assert_sample_files(all_file_items, filesystem, config, load_content, glob_filter)
 
 
+@pytest.mark.parametrize("load_content", (True, False))
```
sh-rp, line 160: this looks completely unrelated but also good :)

sh-rp (approved):
lgtm, I will assume you looked those up ;)

PR #2216, opened 2025-01-14
exp: pygwalker data explorer in streamlit

zilto:
### Description

Add a page to the streamlit dashboard available via `dlt pipeline NAME show`. It uses the recent `Pipeline.dataset` interface to pass a dataframe to `pygwalker`. 

https://github.com/user-attachments/assets/e8b6dca7-49b4-40a7-af39-fdb6469cf08e

### Goals

When first loading a pipeline, I want to know the properties of the data (null, distribution, outliers). It's reducing the friction of having to create temporary notebooks and install `ipykernel/jupyter` dependencies

### Implementation

- [pygwalker](https://github.com/Kanaries/pygwalker) is an open-source Python library that produces an interactive widget to explore dataframes (works in Jupyter, VSCode, Streamlit, Marimo) based on Vega. 

- Integration surface area is small; only assumes that you can pull the data locally via the `Pipeline.dataset` feature.
  - Deeper integration could use `Pipeline.dataset` to query data or pass an `sqlalchemy` connection to `pygwalker` to handle data loading.
  - pygwalker can use `duckdb` for efficient processing of data

### TODO
- add useful message for optional/missing dependencies

### DIFF
--- a/dlt/helpers/streamlit_app/blocks/explorer.py
+++ b/dlt/helpers/streamlit_app/blocks/explorer.py
@@ -1,0 +1,43 @@
+import dlt
+import pygwalker
+import streamlit as st
+from pygwalker.api.streamlit import StreamlitRenderer
+
+st.set_page_config(layout="wide")
+
+
+
+
+
+@st.cache_resource
+def pygwalker_renderer(pipeline_name: str, table_name: str) -> StreamlitRenderer:
+    pipeline = dlt.attach(pipeline_name)
+    dataset = pipeline.dataset()
+    df = dataset[table_name].df()
+    return StreamlitRenderer(
+        df,
+        kernel_computation=True,  # use duckdb under the hood
+        default_tab="vis"
+    )
+
+
+def select_table_name(pipeline: dlt.Pipeline) -> str:
+    current_schema = pipeline.default_schema
+    if schema_name := st.session_state.get("schema_name"):
+        current_schema = pipeline.schemas[schema_name]
+
+    selected_table = st.selectbox(
+        label="Active table",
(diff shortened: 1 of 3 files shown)


dlt/helpers/streamlit_app/blocks/explorer.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+
+
+
+@st.cache_resource
+def pygwalker_renderer(pipeline_name: str, table_name: str) -> StreamlitRenderer:
+    pipeline = dlt.attach(pipeline_name)
+    dataset = pipeline.dataset()
+    df = dataset[table_name].df()
```
sh-rp, line 16: this will load the full table into memory, is this desired? on a large table this will take a long time and at some point kill the host. is there some other way we could do this?

dlt/helpers/streamlit_app/pages/explorer.py
```diff
@@ -0,0 +1,17 @@
+import dlt
+import streamlit as st
+
+from dlt.helpers.streamlit_app.blocks.explorer import show_explorer
+from dlt.helpers.streamlit_app.blocks.menu import menu
+from dlt.helpers.streamlit_app.utils import render_with_pipeline
+
+
+def show(pipeline: dlt.Pipeline) -> None:
```
sh-rp, line 9: we should have at least simple tests for this page to make sure it renders.

sh-rp:
As discussed in the meeting today, we will keep this as a reference, alternatively if @zilto has the time, we can add a test and also derive the explorable view from an sql query which is set to a limit of 2000 items per default or something like that.

PR #2219, opened 2025-01-15
#2166 databricks direct loading

### LINKED ISSUE #2166: Load data into databricks without external staging and auth.
It seems that python sdk for databricks allows to upload files.

1. Research if it is possible to load files into tables like we do into BigQuery: when a local file may be copied into a table without any stage
2. If that does not work, research how to use Volumes on databricks to copy files there and use COPY INTO to move them into table.
3. If authentication is not configured, enable default credentials (ie. if present on serverless compute). You can take a look how `CredentialsWithDefault` is used (most implementations check if default credentials are present in def on_partial(self) -> None: but in your case you should to it in on_resolve when all fields holding credentials are empty)

Ideal scenario. when running in a Notebook, is that we can load a source (ie `rest_api`) without any additional configuration, staging or authorization - like we are able to do with duckdb

donotpush:
- Resolves #2166

### DIFF
--- a/dlt/destinations/impl/databricks/configuration.py
+++ b/dlt/destinations/impl/databricks/configuration.py
@@ -40,23 +40,30 @@ class DatabricksCredentials(CredentialsConfiguration):
             try:
                 # attempt notebook context authentication
                 from databricks.sdk import WorkspaceClient
+                from databricks.sdk.service.sql import EndpointInfo
 
                 w = WorkspaceClient()
-                notebook_context = (
-                    w.dbutils.notebook.entry_point.getDbutils().notebook().getContext()  # type: ignore[union-attr]
-                )
-                self.access_token = notebook_context.apiToken().getOrElse(None)
+                self.access_token = w.dbutils.notebook.entry_point.getDbutils().notebook().getContext().apiToken().getOrElse(None)  # type: ignore[union-attr]
 
-                self.server_hostname = notebook_context.browserHostName().getOrElse(None)
+                # pick the first warehouse on the list
+                warehouses: List[EndpointInfo] = list(w.warehouses.list())
+                self.server_hostname = warehouses[0].odbc_params.hostname
+                self.http_path = warehouses[0].odbc_params.path
             except Exception as e:
                 raise e
 
-            if not self.access_token or not self.server_hostname:
+            if not self.access_token:
                 raise ConfigurationValueError(
-                    "No valid authentication method detected. Provide either 'client_id' and"
-                    " 'client_secret' for OAuth, or 'access_token' for token-based authentication,"
-                    " and the server_hostname."
+                    "Databricks authentication failed: No valid authentication method detected."
+                    " Please provide either 'client_id' and 'client_secret' for OAuth, or"
+                    " 'access_token' for token-based authentication."
                 )
+
+        if not self.server_hostname or not self.http_path or not self.catalog:
+            raise ConfigurationValueError(
+                "Databricks authentication failed: 'server_hostname', 'http_path', and 'catalog'"
+                " are required parameters. Ensure all are provided."
+            )
 
     def to_connector_params(self) -> Dict[str, Any]:
         conn_params = dict(
--- a/dlt/destinations/impl/databricks/configuration.py
+++ b/dlt/destinations/impl/databricks/configuration.py
@@ -5,6 +5,7 @@ from dlt.common.typing import TSecretStrValue
 from dlt.common.configuration.specs.base_configuration import CredentialsConfiguration, configspec
 from dlt.common.destination.reference import DestinationClientDwhWithStagingConfiguration
 from dlt.common.configuration.exceptions import ConfigurationValueError
+from dlt.common.pipeline import get_dlt_pipelines_dir
 
 DATABRICKS_APPLICATION_ID = "dltHub_dlt"
 
@@ -24,6 +25,7 @@ class DatabricksCredentials(CredentialsConfiguration):
     """Additional keyword arguments that are passed to `databricks.sql.connect`"""
     socket_timeout: Optional[int] = 180
     user_agent_entry: Optional[str] = DATABRICKS_APPLICATION_ID
+    staging_allowed_local_path: Optional[str] = None
 
     __config_gen_annotations__: ClassVar[List[str]] = [
         "server_hostname",
@@ -35,6 +37,10 @@ class DatabricksCredentials(CredentialsConfiguration):
     ]
 
     def on_resolved(self) -> None:
+        # conn parameter staging_allowed_local_path must be set to use 'REMOVE volume_path' SQL statement
+        if not self.staging_allowed_local_path:
+            self.staging_allowed_local_path = get_dlt_pipelines_dir()
+
         if not ((self.client_id and self.client_secret) or self.access_token):
             try:
                 # attempt context authentication
@@ -80,6 +86,7 @@ class DatabricksCredentials(CredentialsConfiguration):
             access_token=self.access_token,
             session_configuration=self.session_configuration or {},
             _socket_timeout=self.socket_timeout,
+            staging_allowed_local_path=self.staging_allowed_local_path,
             **(self.connection_parameters or {}),
         )
 
--- a/dlt/destinations/impl/databricks/configuration.py
+++ b/dlt/destinations/impl/databricks/configuration.py
@@ -13,8 +13,8 @@ DATABRICKS_APPLICATION_ID = "dltHub_dlt"
 @configspec
 class DatabricksCredentials(CredentialsConfiguration):
     catalog: str = None
-    server_hostname: str = None
-    http_path: str = None
+    server_hostname: Optional[str] = None
+    http_path: Optional[str] = None
     access_token: Optional[TSecretStrValue] = None
     client_id: Optional[TSecretStrValue] = None
     client_secret: Optional[TSecretStrValue] = None
@@ -48,9 +48,9 @@ class DatabricksCredentials(CredentialsConfiguration):
                 # pick the first warehouse on the list
                 warehouses: List[EndpointInfo] = list(w.warehouses.list())
                 self.server_hostname = warehouses[0].odbc_params.hostname
-                #self.http_path = warehouses[0].odbc_params.path
-            except Exception as e:
-                raise e
+                self.http_path = warehouses[0].odbc_params.path
+            except Exception:
+                pass
 
             if not self.access_token:
                 raise ConfigurationValueError(
--- a/dlt/destinations/impl/databricks/configuration.py
+++ b/dlt/destinations/impl/databricks/configuration.py
@@ -37,19 +37,13 @@ class DatabricksCredentials(CredentialsConfiguration):
     def on_resolved(self) -> None:
         if not ((self.client_id and self.client_secret) or self.access_token):
             try:
-                # attempt notebook context authentication
+                # attempt context authentication
                 from databricks.sdk import WorkspaceClient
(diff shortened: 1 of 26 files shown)


rudolfix (changes requested):
On top of this review:
We should be able to run databricks without configured staging so we can enable many tests that will do that.

We have destination config in `destinations_configs` test util:
1. remove databricks here:
```
destination_configs += [
            DestinationTestConfiguration(destination_type=destination)
            for destination in SQL_DESTINATIONS
            if destination
            not in ("athena", "synapse", "databricks", "dremio", "clickhouse", "sqlalchemy")
        ]
```
2. add staging to this and move it in with other staging databricks setup
```
destination_configs += [
            DestinationTestConfiguration(
                destination_type="databricks",
                file_format="parquet",
                bucket_url=AZ_BUCKET,
            )
        ]
```
you can also ping me and I can prepare a valid commit

dlt/destinations/impl/databricks/configuration.py
```diff
@@ -37,10 +38,23 @@ class DatabricksCredentials(CredentialsConfiguration):
...
     def on_resolved(self) -> None:
         if not ((self.client_id and self.client_secret) or self.access_token):
-            raise ConfigurationValueError(
-                "No valid authentication method detected. Provide either 'client_id' and"
-                " 'client_secret' for OAuth, or 'access_token' for token-based authentication."
-            )
+            # databricks authentication: get context config
+            from databricks.sdk import WorkspaceClient
+
+            w = WorkspaceClient()
```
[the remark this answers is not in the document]
donotpush: makes sense, I haven't tested yet on the notebook, and you are also right about that exception context - I have to catch that exception, and add some tests.

donotpush:
> On top of this review: We should be able to run databricks without configured staging so we can enable many tests that will do that.
> 
> We have destination config in `destinations_configs` test util:
> 
> 1. remove databricks here:
> 
> ```
> destination_configs += [
>             DestinationTestConfiguration(destination_type=destination)
>             for destination in SQL_DESTINATIONS
>             if destination
>             not in ("athena", "synapse", "databricks", "dremio", "clickhouse", "sqlalchemy")
>         ]
> ```
> 
> 2. add staging to this and move it in with other staging databricks setup
> 
> ```
> destination_configs += [
>             DestinationTestConfiguration(
>                 destination_type="databricks",
>                 file_format="parquet",
>                 bucket_url=AZ_BUCKET,
>                 extra_info="az-authorization",
>             )
>         ]
> ```
> 
> you can also ping me and I can prepare a valid commit

I have managed to run tests without staging this way:


```
@pytest.mark.parametrize(
    "destination_config",
    destinations_configs(default_sql_configs=True, subset=("databricks",)),
    ids=lambda x: x.name,
)
def test_databricks_direct_load(destination_config: DestinationTestConfiguration) -> None:
    os.environ["DESTINATION__DATABRICKS__CREDENTIALS__CLIENT_ID"] = ""
    os.environ["DESTINATION__DATABRICKS__CREDENTIALS__CLIENT_SECRET"] = ""

    # direct_load
    os.environ["DESTINATION__DATABRICKS__CREDENTIALS__DIRECT_LOAD"] = "True"

    bricks = databricks()
    config = bricks.configuration(None, accept_partial=True)
    assert config.credentials.access_token

    dataset_name = "test_databricks_token" + uniq_id()
    pipeline = destination_config.setup_pipeline(
        "test_databricks_token", dataset_name=dataset_name, destination=bricks
    )

    info = pipeline.run([1, 2, 3], table_name="digits", **destination_config.run_kwargs)
    assert info.has_failed_jobs is False

    with pipeline.sql_client() as client:
        rows = client.execute_sql(f"select * from {dataset_name}.digits")
        assert len(rows) == 3
```

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -50,126 +50,214 @@ def __init__(
...
+        file_name = FileStorage.get_file_name_from_file_path(local_file_path)
+        file_format = ""
+        if file_name.endswith(".parquet"):
+            file_format = "parquet"
+        elif file_name.endswith(".jsonl"):
+            file_format = "jsonl"
+        else:
+            return "", file_name
+
+        volume_path = f"/Volumes/{self._sql_client.database_name}/{self._sql_client.dataset_name}/{self._sql_client.volume_name}/{time.time_ns()}"
```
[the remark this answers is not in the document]
donotpush: agree about the volume_name

I'll answer the path (time_ns) and file_name in another comment

donotpush:
Answering  (1) CREATE/DROP volume and (2) file name issues

(1) CREATE/DROP Volume

I agree that we can implement your proposed approach for CREATE/DROP VOLUME. However, it will be less efficient as the volume will remain until either the user decides to clean it or we implement a cleanup process (e.g., deleting files after copy statement).

The reason I introduced the create_volume and drop_volume methods is that the DatabricksLoadJob.run method is executed for every single file in parallel, treating each file as a separate COPY statement. This approach is inefficient for “direct loading” since we could load multiple files into a single table using one COPY statement. Additionally, the COPY statement from a volume isn’t file-based—the path must refer to a directory that contains multiple files. This design constraint necessitates creating a unique directory for each file in the current implementation.

Your proposal to run multiple CREATE VOLUME statements might not heavily impact ODBC performance, but it was problem for when I was creating the volume with the SDK (API rate limits) . While this might not be an issue with ODBC since it supports CREATE VOLUME IF NOT EXISTS, it will execute a lot of redudant statements but ODBC will handle it.

Another reason for introducing these methods was to allow the DROP VOLUME statement to be executed outside the run_pool or worker context, which is not feasible otherwise. That said, I’m flexible. If you prefer, I can remove these methods, execute CREATE VOLUME as many times as there are files in the pipeline, and leave the volumes as is, relying on the user for cleanup.

(2) file name issues

I spent an entire day troubleshooting this issue:

[COPY_INTO_SOURCE_SCHEMA_INFERENCE_FAILED] The source directory did not contain any parsable files of type PARQUET. Please check the contents of '/Volumes/test/test_databricks_tokenb3337f88ee667396b15f4e5b2dd5dbb0/_temp_load_volume/file_1737154045813408000'. The error can be silenced by setting 'spark.databricks.delta.copyInto.emptySourceCheck.enabled' to 'false'.


After extensive debugging, I discovered that Databricks runs everything on Spark, and files with names starting with _ or . are treated as hidden files. This caused the COPY command to fail. As a result, we need a way to rename files to avoid conflicts, ensuring file names never depend on the exact table name. For instance, it always failed on the DLT table _dlt_pipeline_state (file _dlt_pipeline_state.1b87dba623.0.parquet).

Since the COPY statement requires a directory (not a file path) and each directory can contain 1 to N files, our current design creates a directory per file. I’ve been using time.time_ns() to create unique directories inside the volume and a random UID as a hex code for the file name. Changing the file name also required determining the correct file extension or format.

To summarize:
	• For (1), we can remove the methods and use CREATE VOLUME IF NOT EXISTS. However, for large loads, this will generate many unnecessary SQL statements.
	•For (2), we need to rename files to avoid conflicts with Spark/Databricks limitations, and each COPY statement must point to a unique directory.

I’m open to adjusting the path format to something more standardized or continuing with random UIDs. Also, let me know your thoughts on deleting files from the volume after loading. Should we leave them for the user to clean, or implement automated cleanup?

donotpush:
@rudolfix I’ve implemented most of your recommendations, but I ran into an issue with enabling the non-staging tests. Unfortunately, your suggested approach didn’t work as expected, the destination gets added, but the files are not local. I’m unsure if it’s possible to achieve this by only changing the configuration without modifying how dlt.destination.databricks is instantiated in the tests

That said, I did add a new test that uses the local file successfully.

Let me know if you have further suggestions or ideas on this!

dlt/destinations/impl/databricks/configuration.py
```diff
@@ -37,9 +36,32 @@ class DatabricksCredentials(CredentialsConfiguration):
...
         if not ((self.client_id and self.client_secret) or self.access_token):
+            try:
+                # attempt notebook context authentication
+                from databricks.sdk import WorkspaceClient
+                from databricks.sdk.service.sql import EndpointInfo
+
+                w = WorkspaceClient()
+                self.access_token = w.dbutils.notebook.entry_point.getDbutils().notebook().getContext().apiToken().getOrElse(None)  # type: ignore[union-attr]
+
+                # pick the first warehouse on the list
```
rudolfix, line 47: please take a list of warehouses when `server_hostname` or `http_path` and then set only the empty. configuration takes precedence. and if we have both values we skip getting warehouses altogether

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -54,122 +53,221 @@ def run(self) -> None:
...
+        volume_database = self._sql_client.dataset_name
+        volume_name = "_dlt_staging_load_volume"
+
+        # create staging volume name
+        fully_qualified_volume_name = f"{volume_catalog}.{volume_database}.{volume_name}"
+        if self._job_client.config.staging_volume_name:
+            fully_qualified_volume_name = self._job_client.config.staging_volume_name
+            volume_catalog, volume_database, volume_name = fully_qualified_volume_name.split(".")
+
+        self._sql_client.execute_sql(f"""
```
rudolfix, line 128: create only unnamed volumes. if `staging_volume_name` assume that it exists. please also document this. mention that if you go for production it is better to use named volume

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -54,122 +53,221 @@ def run(self) -> None:
...
+            CREATE VOLUME IF NOT EXISTS {fully_qualified_volume_name}
+        """)
+
+        volume_path = f"/Volumes/{volume_catalog}/{volume_database}/{volume_name}/{time.time_ns()}"
```
rudolfix, line 132: maybe you should take `uniq_id()` from dlt common instead of time? for example on windows you get timer resolution of 20ms. you may get clashing names

rudolfix (changes requested):
I enables no staging mode as a default for databricks.
1. please update the documentation
2. is it hard to delete staging files from databrick volumes? I bet you can do that from SQL... we have similar functionality for snowflake. IMO should be pretty easy
otherwise LGTM! good job

dlt/destinations/impl/databricks/databricks.py
donotpush: done

dlt/destinations/impl/databricks/databricks.py
donotpush: done

dlt/destinations/impl/databricks/configuration.py
donotpush: done

donotpush:
@rudolfix 

Thanks for the review, and helping out with enabling the tests.

I've implemented all your recommendations.

- Docs, I've named this feature "Direct Load", feel free to suggest changes or change it directly, perhaps it is better to add it just as another staging flavour: databricks managed volumes. 

- Files deleted from volume right after loading using SQL: `REMOVE volume_path` - executing that command required a new conn parameter to be set: `self.staging_allowed_local_path = get_dlt_pipelines_dir()` - I didn't know which path to use, perhaps is better to use something different (it also works just setting a non-empty string), I don't see any data being generated in the provided path.

- About the tests, I think that a few aren't working propertly, for example `test_databricks_external_location` is now only running on destination "databricks-no-staging" (but files aren't local) instead of `databricks-*-staging-az-authorization`

rudolfix (approved):
LGTM! thanks for the tests and docs.

PR #2236, opened 2025-01-26
typed entity registries

rudolfix:
### Description
replaces #2183 

also
* cleansup dict arrow types for delta and iceberg
* bumps duckdb and iceberg scanner to nightly
* optimizes creating views in `sql_client` for filesystem
* allows passing string source references to `pipeline.run` on top of passing source instances


tests/extract/test_sources.py
```diff
@@ -942,6 +942,98 @@ async def r_async():
     assert async_list in allowed_results
 
 
+def test_limit_yield_cleanup() -> None:
```
sh-rp, line 945: this is a good test! i had briefly thought about wether the limit influences the incremental state but I think I forgot to add a test. There are a couple of print statements remaining further down here.

dlt/extract/reference.py
```diff
@@ -0,0 +1,229 @@
...
+
+    with_args = deprecated("Please use clone method instead", category=Dlt04DeprecationWarning)(
+        clone
+    )
+
+
+AnySourceFactory = SourceFactory[Any, DltSource]
+
+
+class SourceReference:
```
sh-rp, line 68: Maybe these classes are to different, but isn't there a common mechanism between the source and the destination reference that would warrant a mutual baseclass?

sh-rp:
Very cool PR! I read all of it but as always with these huge PRs can't promise to understand every part of it, but some semi related or even not related observations are:

* The filesystem destination is getting a bit crazy (not due to these changes) and I am wondering if we should split it up into the regular filesystem, delta and iceberg. Also from a user / sdk perspective I think we are abusing the schema with the table_format. Intuitively a destination should be delta and then all the stuff getting loaded should be delta tables without any additional hints. Then you should be able to add "delta" as destination instead of filesystem when defining the pipeline.

* Based on my thinking about having better insight into configuration errors, we could consider having a "DLT__" prefix for all environment variables that are consumed by dlt, this way it would be easier to give the user a report about which settings were found and which ones were actually used and which ones might be typos or outdated configs.

tests/extract/test_sources.py
rudolfix: I wrote this test because I realized that we cannot limit `pg_replication` this way which is doing `yield from` iterator but full iterators MUST be consumed. So we'd need to add some kind of groupings.

rudolfix:
@sh-rp I've got all tests fixed I think

PR #2242, opened 2025-01-27
Update branch naming guide in contributing doc to add a "keep" prefix

sh-rp:
### Description
Some branches we want to keep around for reference or for working on later but close the pr. For this we add a keep prefix.


rudolfix (approved):
LGTM! Also @burnash  should see it before merging

burnash (approved):

PR #2246, opened 2025-01-28
added exception for unsupported arrow types

### LINKED ISSUE #2223: Provide more detailed error reporting when DLT pipeline fails due to incorrect data type conversions
### Feature description

Scenario:

A dlt job failed because a column in the source is casted wrong (interval instead of timestamp).
dlt fails because the Pyarrow backend cannot convert a timestamp to a duration type.

```
Run started at 2025-01-16 14:30:14.498969+00:00 and FAILED in 30.31 seconds with 2 steps.
Step extract FAILED in 30.30 seconds.
Failed due to: Pipeline execution failed at stage extract when processing package 1737037814.5092769 with exception:

<class 'ValueError'>
duration[us]

Load package 1737037814.5092769 is NEW and NOT YET LOADED to the destination and contains no failed jobs

Step run FAILED in 30.31 seconds.
Failed due to: Pipeline execution failed at stage extract when processing package 1737037814.5092769 with exception:

<class 'ValueError'>
duration[us]

Load package 1737037814.5092769 is NEW and NOT YET LOADED to the destination and contains no failed jobs
```

The trace even if the config is set to `DEBUG` contains no insight into what column caused this failure. To debug this, one has to manually deselect tables/columns and re-run.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

Trying to find out, which column causes the job to fail.

### Proposed solution

In the trace, show which table/column caused the failure.

### Related issues

_No response_

zilto:
Fix #2223 by adding `UnsupportedArrowTypeException` with helpful error messages. This was implemented at the low-level with proper testing.

The type conversion function is used by higher level constructs:
- `dlt.extract.extractors.ArrowExtractor._compute_table()`
- `dlt.normalize.items_normalizers.ArrowItemsNormalizer._fix_schema_precisions()`
Reraising was necessary to include the column name and table name.

I didn't write tests for these two high-level object methods because I couldn't find straightforward tests to implement. If it's desirable, please point me in the right direction. The test should be simple to write


zilto:
The failing test in `common` for `python==3.13` and Windows seems unrelated to the changes here. Seems a bit esoteric to investigate

dlt/common/libs/pyarrow.py
```diff
@@ -400,10 +424,15 @@ def py_arrow_to_table_schema_columns(schema: pyarrow.Schema) -> TTableSchemaColu
     """
     result: TTableSchemaColumns = {}
     for field in schema:
+        try:
+            converted_type = get_column_type_from_py_arrow(field.type)
+        except UnsupportedArrowTypeException:
+            raise UnsupportedArrowTypeException(arrow_type=field.type, column_name=field.name)
```
rudolfix, line 430: we are raising same exception type to augment the info in it (column, table). maybe we should just set the relevant property of the existing exception and re-raise? then we have a cleaner exception trace (not many exception causing each other).

but I'm not sure here. just for your consideration.

rudolfix (changes requested):
LGTM! just see my review comment.

failing test looks like a fluke. I reran the test

dlt/common/libs/pyarrow.py
zilto: This exception can only originate from `get_column_type_from_py_arrow()`, but this function has no context about the column and table names, which are highly relevant for debugging.

I added `@setter` to the exception class to allow to modify the exception message in place without reraising

rudolfix (approved):
LGTM!

PR #2255, opened 2025-02-02
docs: Fix typo in file name in "Create a pipeline" walkthrough

lpillmann:
### Description
When following the walkthrough ["Create a pipeline" ](https://dlthub.com/docs/walkthroughs/create-a-pipeline), I got the error
```
python: can't open file '.../github_api_duckdb/github_api.py': [Errno 2] No such file or directory
```
when running:

```bash
python github_api.py
```

It turns out the command `dlt init github_api duckdb` creates the pipeline file with a slightly different name `github_api_pipeline.py`, so we need to adjust it.

### Related Issues
N/A

### Additional Context
N/A


burnash (approved):
Thank you for the fix, @lpillmann

PR #2259, opened 2025-02-03
more flexible toml resolvers

sh-rp:
### Description
Updates toml resolvers to have an arbitrary list of resolvable dirs and paths which can be subclassed and refined more.


dlt/common/configuration/providers/toml.py
```diff
@@ -179,36 +182,37 @@ def _read_toml_file(self, toml_path: str) -> tomlkit.TOMLDocument:
...
+            # merge all toml files into one
+            result_toml: Optional[tomlkit.TOMLDocument] = None
+            for path in toml_paths:
+                if (loaded_toml := self._read_toml_file(path)) is not None:
+                    if result_toml is None:
+                        result_toml = loaded_toml
+                    else:
+                        result_toml = update_dict_nested(loaded_toml, result_toml)
+
+            # if nothing was found, try to load from google colab or streamlit
```
sh-rp, line 199: this part has changed in a way that colab and streamlit secrets only are loaded if no toml files at all are found. In the old implementation a potential global setting was merged into these. I am not sure if this is ok, this should definately be clear before we merge this.

dlt/common/configuration/providers/toml.py
```diff
@@ -64,31 +63,36 @@ def __init__(
             name(str): name of the provider when registering in context
             supports_secrets(bool): allows to store secret values in this provider
             file_name (str): The name of `toml` file to load
-            settings_dir (str, optional): The location of `file_name`. If not specified, defaults to $cwd/.dlt
-            global_dir (bool, optional): Looks for `file_name` in global_dir (defaults to `dlt` home directory which in most cases is $HOME/.dlt)
+            resolvable_dirs (List[str], optional): A list of directories to resolve the file from, files will be merged into each other in the order the directories are specified. If not specified, defaults to [$cwd/.dlt, $HOME/.dlt]
```
rudolfix, line 66: pls change the docstring
- no optional
- does not default to anything

rudolfix (changes requested):

Issue #2264, opened 2025-02-04
Unexpected behavior when using both resolved and unresolved params


CharlesRngrd:
### dlt version

1.6.0

### Describe the problem

Hi,

In the resource bellow, I define 3 kings of params:
 
- pagination param
- resolved param
- unresolved param

```
{
    "name": "users",
    "selected": False,
    "endpoint": {
        "path": "/api/v4/users",
        "paginator": {"type": "page_number", "total_path": "totalPages"},
        "params": {
            "sortBy": "hiringDate",
            "desc": True,
            "displayArchivedUsers": False,
            "id_cofaq": {
                "type": "resolve",
                "resource": "companies",
                "field": "id",
            },
        },
        "data_selector": "content",
    },
```
},

The issue is that the unresolved params are not sent to the request according to the logs:

>  https://api.eurecia.com/eurecia/api/v4/users with params={'id_cofaq': 'COFAQ--------20191114-164321', 'page': 0}

Could you fix this please ?

### Steps to reproduce

Create a fake resource with both:

- pagination param
- resolved param
- unresolved param

Then run the resource and see in the logs which parameters are there.

### Operating system

Windows

### Runtime environment

Local

### Python version

3.12

burnash:
Hey @CharlesRngrd thanks for reporting this. Looking into it.

CharlesRngrd:
Ok no problem.
I hardcoded the parameter so I can wait.

burnash:
Closing this as resolved in #2210 

@CharlesRngrd thank you again for reporting this. The fix is available in the alpha release 1.8.0a0. We're going to release 1.8.0 very soon.

Please mind the updated syntax for query params. Your example from above would look like this:

```
{
    "name": "users",
    "selected": False,
    "endpoint": {
        "path": "/api/v4/users",
        "paginator": {"type": "page_number", "total_path": "totalPages"},
        "params": {
            "sortBy": "hiringDate",
            "desc": True,
            "displayArchivedUsers": False,
            "id_cofaq": "{resources.companies.id}",  # <-- note the new template placeholder 
        },
        "data_selector": "content",
    },
}
```

Let me know if this works for you.

CharlesRngrd:
@burnash Sorry for the delay, it works pretty well except I have an empty body issue with the new feature.
I will open a new issue for that.

PR #2268, opened 2025-02-04
rest_api: add tests for query string parameters

burnash:
The PR expands the offline integration tests for the rest_api source to verify that query string parameters are correctly sent to the server.

Related issue: #2264

### DIFF
--- a/tests/sources/rest_api/integration/test_offline.py
+++ b/tests/sources/rest_api/integration/test_offline.py
@@ -100,17 +100,127 @@ def test_load_mock_api(mock_api_server):
     )
 
 
-def test_single_resource_query_string_params(mock_api_server):
+@pytest.mark.parametrize(
+    "endpoint_params,expected_static_params",
+    [
+        # Static params only
+        pytest.param(
+            {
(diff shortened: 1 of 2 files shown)


tests/sources/rest_api/integration/test_offline.py
```diff
@@ -100,6 +100,297 @@ def test_load_mock_api(mock_api_server):
...
+                    "base_page": 1,
+                    "total_path": "total_pages",
+                },
+            },
+            "resources": [
+                {
+                    "name": "posts",
+                    "endpoint": {
+                        **endpoint_params,
+                        "response_actions": [response_hook],
```
rudolfix, line 218: this is pretty cool. but `mock_api_server` is just a requests mock and you can get a call history from it: https://requests-mock.readthedocs.io/en/latest/history.html

burnash: good point, thanks for the tip. fixed in 51d990392ddf28e6668d874df7f367247344e213

rudolfix (approved):
LGTM!

PR #2270, opened 2025-02-06
add workflow for running plus tests

sh-rp:
### Description
Runs some basic tests against a nightly build of dlt+. It also adds dlt_plus to the depedencies when linting docs snippets, so these should be properly linted now!


rudolfix (approved):
LGTM!

PR #2286, opened 2025-02-11
feat: support encoding custom types in `json` modules

joscha:
As per suggestion in https://github.com/dlt-hub/dlt/pull/2286#pullrequestreview-2616014683 this adds a new `json.set_custom_encoder(my_encoder)` feature that allows passing a callable to encode arbitrary data types when dumping json.

Usage:
```py
from dlt.common import json
from dlt.common.json import JsonSerializable
from pydantic import AnyUrl

def my_custom_encoder(obj: Any) -> JsonSerializable:
    if isinstance(obj, AnyUrl):
        # encodes the url as punycode string
        return str(obj)
    # Don't know this type, throw
    raise TypeError(repr(obj) + " is not JSON serializable")

json.set_custom_encoder(my_custom_encoder)
json.dumps(AnyUrl("https://£££.com"))  # "https://xn--9aaa.com/"
```

The custom encoder is run after all internal encoding cases have been exhausted. If `None` is returned from it, the serialization fails as previously.

-----
Previously, this was:
# fix: allows `AnyUrl` pydantic types

Without this change, serializing a model with [`AnyUrl`](https://docs.pydantic.dev/latest/api/networks/#pydantic.networks.AnyUrl) fails.

A sample pydantic model would be:
```py
class Resources(BaseModel):
    website: AnyUrl | None = None
    """
    Website URL.
    """
```

The error with such an entity yielded from a resource looks like this:

```
TypeError: Type is not JSON serializable: AnyUrl
```

Details:

```
    self._extract_single_source(
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/extract/extract.py", line 349, in _extract_single_source
    with self.manage_writers(load_id, source):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/ss2y42j0bjh7j2vlnxys606r3wz9r931-python3-3.12.8-env/lib/python3.12/contextlib.py", line 144, in __exit__
    next(self.gen)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/extract/extract.py", line 387, in manage_writers
    self.extract_storage.close_writers(load_id)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/extract/storage.py", line 78, in close_writers
    storage.close_writers(load_id, skip_flush=skip_flush)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/storages/data_item_storage.py", line 85, in close_writers
    writer.close(skip_flush=skip_flush)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/data_writers/buffered.py", line 176, in close
    self._flush_and_close_file(skip_flush=skip_flush)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/data_writers/buffered.py", line 260, in _flush_and_close_file
    self._flush_items(allow_empty_file)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/data_writers/buffered.py", line 250, in _flush_items
    self._writer.write_data(self._buffered_items)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/data_writers/writers.py", line 184, in write_data
    json.typed_dump(items, self._f)
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/json/_orjson.py", line 32, in typed_dump
    fp.write(typed_dumpb(obj, pretty=pretty))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/json/_orjson.py", line 36, in typed_dumpb
    return _dumps(obj, sort_keys, pretty, custom_pua_encode, orjson.OPT_PASSTHROUGH_DATETIME)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/joscha/dev/dlt-source-morphais/.devenv/state/venv/lib/python3.12/site-packages/dlt/common/json/_orjson.py", line 24, in _dumps
    return orjson.dumps(obj, default=default, option=options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Type is not JSON serializable: AnyUrl
```

Please let me know if this is an acceptable fix, then I will add a test


dlt/common/json/__init__.py
```diff
@@ -12,6 +12,11 @@
 except ImportError:
     PydanticBaseModel = None  # type: ignore[misc]
 
+try:
+    from pydantic import AnyUrl as PydanticAnyUrl
+except ImportError:
+    PydanticAnyUrl = None  # type: ignore[misc]
```
joscha, line 18: I cargo-culted this from above - I am assuming this is for the case when the Pydantic classes are not available and meant to gracefully degrade?

dlt/common/json/__init__.py
```diff
@@ -49,6 +54,8 @@ def custom_encode(obj: Any) -> str:
         return dataclasses.asdict(obj)  # type: ignore
     elif isinstance(obj, Enum):
         return obj.value  # type: ignore[no-any-return]
+    elif isinstance(obj, PydanticAnyUrl):
+        return str(obj) # will punycode encode the host if required
```
joscha, line 58: could also be:
```suggestion
        return obj.unicode_string() # will NOT punycode encode the host
```
which would be more in line with not touching the data inside DLT

rudolfix (changes requested):
there are too many different types that we could add this way. my take for this PR would be to add pluggable `custom_encde_extension` that is NULL by default and that you can set in `json` module if you need special types to be serialized. so you could add your special type handling you just plug in your encoder. such thing we are happy to merge. so
1. at the end of cutom_encode call `custom_encde_extension` (if not None) and return wha it returns or if it returns None, raise TypeError like previously
2. add tests along other json tests

Also mind that maybe this is not the place you really want to plug your encoders.

between extract and normalize (where all weird types happen) we use `custom_pua_encode` so 90% you want to plug your encoder there. lmk. glad to help

joscha:
> there are too many different types that we could add this way. my take for this PR would be to add pluggable `custom_encde_extension` that is NULL by default and that you can set in `json` module if you need special types to be serialized. so you could add your special type handling you just plug in your encoder. such thing we are happy to merge. so
> 1. at the end of cutom_encode call `custom_encde_extension` (if not None) and return wha it returns or if it returns None, raise TypeError like previously
> 2. add tests along other json tests
> 
> Also mind that maybe this is not the place you really want to plug your encoders.
> 
> between extract and normalize (where all weird types happen) we use `custom_pua_encode` so 90% you want to plug your encoder there. lmk. glad to help
> 

I think that makes sense. Do you want to give me an interface signature of your choosing with a 3 line example on how to use it and I'll implement it?

dlt/common/json/__init__.py
```diff
@@ -49,6 +54,8 @@ def custom_encode(obj: Any) -> str:
         return dataclasses.asdict(obj)  # type: ignore
     elif isinstance(obj, Enum):
         return obj.value  # type: ignore[no-any-return]
+    elif isinstance(obj, PydanticAnyUrl):
```
sh-rp, line 57: you need to also test wether pydanticanyurl is defined, see above how it is done with the Basemodel.

joscha: I can, but going by @rudolfix' comment the implementation of this might become a fully custom hook anyway, no?

dlt/common/json/__init__.py
```diff
@@ -126,6 +137,13 @@ def custom_pua_encode(obj: Any) -> str:
     elif isinstance(obj, Enum):
         # Enum value is just int or str
         return obj.value  # type: ignore[no-any-return]
+    elif _custom_encoder is not None:
+        ret = _custom_encoder(obj)
+        if ret is not None:
+            if not isinstance(ret, str):
+                print("raisins")
```
joscha, line 144:
```suggestion
```

dlt/common/json/_orjson.py
```diff
@@ -21,7 +27,8 @@ def _dumps(
         options |= orjson.OPT_INDENT_2
     if sort_keys:
         options |= orjson.OPT_SORT_KEYS
-    return orjson.dumps(obj, default=default, option=options)
+    print("_custom_encoder", _custom_encoder)
```
joscha, line 30:
```suggestion
```

tests/common/test_json.py
```diff
@@ -348,3 +358,67 @@ def test_load_and_compare_all_impls() -> None:
...
+    json_impl.set_custom_encoder(my_custom_encoder)
+
+    if json_impl is _simplejson:
+        with pytest.raises(TypeError, match="Custom encoder must return a string for PUA encoding"):
+            json_impl.typed_dumpb(AnyUrl("https://www.example.com"))
+    elif json_impl is _orjson:
+        # orjson abstracts away custom exceptions and wrap them in a nonserializable TypeError
+        with pytest.raises(TypeError, match="Type is not JSON serializable"):
+            json_impl.typed_dumpb(AnyUrl("https://www.example.com"))
+    assert calls == 1
```
joscha, line 424:
```suggestion
    assert calls == 1

```

dlt/common/json/__init__.py
```diff
@@ -126,6 +137,12 @@ def custom_pua_encode(obj: Any) -> str:
     elif isinstance(obj, Enum):
         # Enum value is just int or str
         return obj.value  # type: ignore[no-any-return]
+    elif _custom_encoder is not None:
+        ret = _custom_encoder(obj)
+        if ret is not None:
+            if not isinstance(ret, str):
+                raise TypeError("Custom encoder must return a string for PUA encoding")
```
joscha, line 144: I did this because of the signature of `custom_pua_encode` which is more narrow than the one from `custom_encode`.
We can also:
* allow two separate encoders, one for `custom_pua_encode` and one for `custom_encode`
* not throw here and just document the requirement; I can see that the `Enum` case above just ignores a potential any return value and objects with `_asdict` extension are actually not returning `str` either - is it possible the signature of `custom_pua_encode` is outdated?

tests/common/test_json.py
```diff
@@ -225,6 +226,15 @@ def test_json_pendulum(json_impl: SupportsJson) -> None:
 #     assert isinstance(delta, timedelta)
 #     print(str(delta.as_timedelta()))
 
+@pytest.fixture(autouse=True)
+def run_around_tests():
+    for impl in _JSON_IMPL:
+        # clear the encoder
+        impl.set_custom_encoder(None)
```
joscha, line 233: we could implement a `.clear_custom_encoder()` if you wanted, just holler.

tests/common/test_json.py
```diff
@@ -348,3 +358,67 @@ def test_load_and_compare_all_impls() -> None:
...
+    s = json_impl.dumps(x)
+    assert json_impl.loads(s) == {"url": "https://xn--9aaa.com/", "hello": "world"}
+
+@pytest.mark.parametrize("json_impl", _JSON_IMPL)
+def  test_serialize_custom_types_no_encoder(json_impl: SupportsJson) -> None:
+    with pytest.raises(TypeError, match="is not JSON serializable"):
+        roundtrip(json_impl)
+
+@pytest.mark.parametrize("json_impl", _JSON_IMPL)
+def  test_serialize_custom_types_noop_encoder(json_impl: SupportsJson) -> None:
```
joscha, line 377: this tests asserts happens when a custom encoder is set but returns `None`.

tests/common/test_json.py
```diff
@@ -348,3 +358,67 @@ def test_load_and_compare_all_impls() -> None:
...
+        if isinstance(obj, AnyUrl):
+            return str(obj)
+        return None
+
+    json_impl.set_custom_encoder(my_custom_encoder)
+    roundtrip(json_impl)
+    assert calls == 1
+
+@pytest.mark.parametrize("json_impl", _JSON_IMPL)
+def  test_serialize_custom_types_pua_string_only(json_impl: SupportsJson) -> None:
```
joscha, line 406: this tests that the json implementation throws if the custom encoder returns non-`str` on a pua encode; see above.

dlt/common/json/__init__.py
```diff
@@ -21,8 +21,15 @@
...
+Serializable = Union[str, Dict[Any, Any]]
 
-def custom_encode(obj: Any) -> Union[str, Dict[Any, Any]]:
+Encoder = Callable[[Any], Union[Serializable, None]]
+"""
+Custom user-defined encoder function type
+"""
+
+
+def custom_encode(_custom_encoder: Union[Encoder, None], obj: Any) -> Serializable:
```
rudolfix, line 32: for regular code, using `partial` and injecting the encoder into signature would be ideal. but this is one of the most intensive called functions in `dlt` so what about the following
- we keep the signature intact
- we check a global variable `_custom_encoder ` in line 60
- this variable is stored in **this** module (but maybe set via _orjson or _simplejeson - please figure out)

dlt/common/json/_orjson.py
```diff
@@ -21,7 +29,7 @@ def _dumps(
         options |= orjson.OPT_INDENT_2
     if sort_keys:
         options |= orjson.OPT_SORT_KEYS
-    return orjson.dumps(obj, default=default, option=options)
+    return orjson.dumps(obj, default=partial(default, _custom_encoder), option=options)
```
rudolfix, line 32: as mentioned we'll skip partial here so this function goes back to the original. just a question: why `typed_dumpb` got skipped? you want to apply encoding to pua encoder as well (which is used between extract and normalize stages)

dlt/common/libs/pyarrow.py
```diff
@@ -866,13 +867,17 @@ def convert_numpy_to_arrow(
                 try:
                     # the 3 types match those supported by `map_nested_in_place()`
                     if isinstance(value, (tuple, dict, list)):
-                        encoded_value = map_nested_in_place(custom_encode, value)
+                        encoded_value = map_nested_in_place(
```
rudolfix, line 870: if we go with global var this won't be needed

tests/common/test_json.py
```diff
@@ -226,6 +226,17 @@ def test_json_pendulum(json_impl: SupportsJson) -> None:
 #     print(str(delta.as_timedelta()))
 
 
+@pytest.fixture(autouse=True)
+def run_around_tests():
+    for impl in _JSON_IMPL:
+        # clear the encoder
+        impl.set_custom_encoder(None)
+    yield
```
rudolfix, line 234: you must put `yield` in try..finally... and release changes there. you may get exception form the code executed on yield

rudolfix (changes requested):
this is good! thanks for really good tests and docs. also you are correct that encoding function signature is outdated. please read my review.

one more thing: you should probably test your case with Pydantic. we have tests/pipeline/test_pipeline_extra.py where you can add end to end test

dlt/common/json/_orjson.py
joscha: > question: why `typed_dumpb` got skipped? you want to apply encoding to pua encoder as well (which is used between extract and normalize stages)

`typed_dumpb` delegates to `_dumps` which already curries here. If I were to curry there, then it would attempt to curry twice.

tests/common/test_json.py
joscha: changed in 659cc5c

dlt/common/libs/pyarrow.py
joscha: changed in 659cc5c

dlt/common/json/__init__.py
joscha: implemented in 659cc5c. One downside of the global var is that state can possibly leak, e.g. the custom encoder is not only set on the imported impl. I am unsure of how much of an issue that is, as most likely the `json` module won't be used as a utility, but only set up once before pipeline exectuon.

dlt/common/json/__init__.py
```diff
@@ -21,8 +21,15 @@
 
 TPuaDecoders = List[Callable[[Any], Any]]
 
+Serializable = Union[str, Dict[Any, Any]]
```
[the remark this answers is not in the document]
joscha: changed in 659cc5c

joscha:
@rudolfix I made changes regarding to your earlier feedback, PTAL

dlt/common/json/__init__.py
```diff
@@ -116,19 +140,24 @@ def custom_pua_encode(obj: Any) -> str:
     elif isinstance(obj, bytes):
         return _B64BYTES + base64.b64encode(obj).decode("ascii")
     elif hasattr(obj, "asdict"):
-        return obj.asdict()  # type: ignore
+        return obj.asdict()  # type: ignore[no-any-return]
```
joscha, line 143: these ignores can be more narrow now, as we allow a broader return value (dicts)

dlt/common/json/__init__.py
```diff
@@ -21,9 +21,28 @@
 
 TPuaDecoders = List[Callable[[Any], Any]]
 
+JsonSerializable = Union[str, Dict[str, Any]]
```
joscha, line 24: This type is a bit naive. Possibly it would be better to use something slightly more accurate like described in https://github.com/python/typing/issues/182

tests/common/test_json.py
```diff
@@ -348,3 +360,41 @@ def test_load_and_compare_all_impls() -> None:
...
+def test_serialize_custom_types_no_encoder(json_impl: SupportsJson) -> None:
+    with pytest.raises(TypeError, match="is not JSON serializable"):
+        roundtrip(json_impl)
+
+
+@pytest.mark.parametrize("json_impl", _JSON_IMPL)
+def test_serialize_custom_types_with_encoder(json_impl: SupportsJson) -> None:
+    calls = 0
+
+    def my_custom_encoder(obj: Any) -> Any:
```
joscha, line 391:
```suggestion
    def my_custom_encoder(obj: Any) -> JsonSerializable:
```

tests/pipeline/test_pipeline_extra.py
```diff
@@ -744,3 +749,23 @@ def events_timezone_unset():
...
+        raise TypeError(repr(obj) + " is not JSON serializable")
+
+    # get global custom encoder
+    current_custom_json_encoder = _custom_encoder
+    try:
+        json.set_custom_encoder(my_custom_encoder)
+        s = json.dumps(AnyUrl("https://£££.com"))
+        assert json_impl.loads(s) == "https://xn--9aaa.com/"
+    finally:
+        # revert whatever custom encoder was set before
```
joscha, line 770:
```suggestion
        # revert to whatever custom encoder was set before
```

rudolfix (commented):
this is really good now! I have one question in my review and then we can merge

dlt/common/json/__init__.py
```diff
@@ -21,9 +21,28 @@
...
+_custom_encoder: Union[None, JsonEncoder] = None
+"""
+Holds the custom encoder function, if set.
+This is used as a last-resort fallback for encoding objects.
+"""
+
+
+def _custom_encode(obj: Any) -> JsonSerializable:
     """Returns a JSON-serializable representation of `obj`"""
+    global _custom_encoder
```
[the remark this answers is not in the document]
joscha: gotcha; I am happy either way, let me know. I do think the marker makes it more readable and very clear in intent, but I am happy to follow your guidance.

rudolfix (changes requested):
all clear! please fix the tests and we can merge

tests/common/test_json.py
```diff
@@ -18,6 +18,7 @@
     _simplejson,
     SupportsJson,
     _DATETIME,
+    _custom_encoder,
```
joscha, line 21:
```suggestion
    _custom_encoder,
    JsonSerializable,
```

joscha:
> all clear! please fix the tests and we can merge

All green now @rudolfix

rudolfix (approved):
LGTM!

Issue #2224, opened 2025-01-17
Error when reading numeric datatype from Postgres: Could not convert Decimal('0.0') with type decimal.Decimal: tried to convert to double


PilotKresse:
### dlt version

1.5.0

### Describe the problem

Running the following snippet in DLT:

```python
pipe = dlt.pipeline(
    "test",
    dataset_name="test_pipeline",
    destination="duckdb",
)



r = sql_table(
    table="[REDACTED]",
    schema="[REDACTED]",
    backend="pyarrow"
)


pipe.run(r)
```

fails with error `Could not convert Decimal('0.0') with type decimal.Decimal: tried to convert to double`. 

When I run the pipeline with another table from the same database that does not contain columns of type numeric, the pipeline runs successfully. What is the problem here?

Using the sqlalchemy backend is not an option, because it is too slow.

### Steps to reproduce

Try to run above pipeline with a Postgres database (version 13) and a table containing columns of type numeric

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

sql_table

### dlt destination

DuckDB

rudolfix:
@PilotKresse could you past the stack trace? I'm not sure where the exceptions is coming from. May be SQL alchemy internals...

bgarcevic:
I have the same issue here is my stack trace from snowflake -> snowflake - It works with the sqlalchemy backend but not the pyarrow:

Traceback (most recent call last):
  File ".env\Lib\site-packages\dlt\extract\pipe_iterator.py", line 277, in _get_source_item
    pipe_item = next(gen)
                ^^^^^^^^^
  File ".env\Lib\site-packages\dlt\sources\sql_database\helpers.py", line 301, in table_rows
    yield from loader.load_rows(backend_kwargs)
  File ".env\Lib\site-packages\dlt\sources\sql_database\helpers.py", line 178, in load_rows
    yield from self._load_rows(query, backend_kwargs)
  File ".env\Lib\site-packages\dlt\sources\sql_database\helpers.py", line 200, in _load_rows
    yield row_tuples_to_arrow(
          ^^^^^^^^^^^^^^^^^^^^
  File ".env\Lib\site-packages\dlt\common\configuration\inject.py", line 247, in _wrap
    return f(*bound_args.args, **bound_args.kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #2295, opened 2025-02-12
Improved `"decimal"` to `pyarrow` conversion

zilto:
Better errors for decimal conversion to pyarrow and helpful messages. 

## Related issues
#2224 This might not fix their error, will need more info to repro database-specific issues.

The problem from the original author seems to be solvable with current dlt `precision` and `scale` hints when `data_type="decimal"`([reference](https://dlthub.com/docs/general-usage/schema/#tables-and-columns)). Catching the `pyarrow` exception and adding a helpful message should helpful future users.

Linking #2273 because it touches the same code path. Having pyarrow infer types as default would likely also solve this issue

## Changes
- `dlt.common.libs.pyarrow.row_tuples_to_arrow()` was refactored to flatten conversion logic. 
  - reducing the number of intermediary numpy and arrow references potentially improves memory usage
- `dlt.sources.sql_database.arrow_helpers.row_tuples_to_arrow()` mentions in the docstring that it should pickup the destination's capabilities. Despite the function signature, `caps` was never passed. I added a call to retrieve the current capabilities
- `Decimal` handling has comprehensive warnings, exceptions, and tests

## Design / future work
- `dlt` should more explicitly document if conversion to `pyrrow` will 1) coerce + warn or 2) raise. We could further leverage [pyarrow.compute.CastOptions](https://arrow.apache.org/docs/python/generated/pyarrow.compute.CastOptions.html#pyarrow.compute.CastOptions) to let pyarrow determine behavior
- Currently, setting hints `{"data_type": "decimal"}` without `scale` or `precision` strongly enforces defaults `(38, 9)`. Letting pyarrow handle conversion seems like a better default.
  - It's odd that the destination capabilities set precision and scale. If the data goes **to python**, can't we just infer it? 
- Inspecting `psycopg3` (dlt uses `psycopg2`) [shows that Postgres](https://github.com/psycopg/psycopg/blob/be4e811369e446188e3ffd6c000dd32eec594023/psycopg/psycopg/types/numeric.py#L274) uses binary or bytes representation for `Numeric` type.

### DIFF
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -815,20 +815,41 @@ def convert_numpy_to_arrow(
 
             elif (
                 "to utf8 using function cast_string" in error_msg
-                and dlt_data_type == "json"
+                and dlt_data_type in ("json", "text")
                 and pa.types.is_string(inferred_arrow_type)
             ):
                 # this is handled by fallback case 3
                 logger.warning(
-                    "Received `data_type='json'`, data requires serialization to string, slowing"
-                    " extraction. Cast the JSON field to STRING in your database system to improve"
-                    " performance. For example, create and extract data from an SQL VIEW that"
-                    " SELECT with CAST."
+                    f"Received `data_type='{dlt_data_type}'`, data requires serialization to"
+                    " string, slowing extraction. Cast the JSON field to STRING in your database"
+                    " system to improve performance. For example, create and extract data from an"
+                    " SQL VIEW that SELECT with CAST."
                 )
 
-    # case 2: encode Python types unsupported by Arrow. Simple types are converted to strings and complex types to common structures (dict, list)
+    # case 2: encode Sequence and Mapping types (list, tuples, set, dict, etc.) to JSON strings
+    # This logic needs to be before case 3, otherwise pyarrow might infer the deserialized JSON object as a `pyarrow.struct` instead of `pyarrow.string`
+    if arrow_array is None and dlt_data_type in (
+        "json",
+        "text",
+    ):  # depending on the backend, JSON columns are inferred as data_type="text"
+        json_serialized_values: list[Union[bytes, None]] = []
+        for value in column_data:
+            if value is None:
+                json_serialized_values.append(None)
+                continue
+            try:
+                json_serialized_values.append(json.dumpb(value))
+            except TypeError as e:
+                raise PyToArrowConversionException(
+                    data_type=dlt_data_type,
+                    inferred_arrow_type=inferred_arrow_type,
+                    details="dlt failed to a JSON-serializable type.",
+                ) from e
+
+        arrow_array = pa.array(json_serialized_values).cast(pa.string())
+
+    # case 3: encode Python types unsupported by Arrow. Simple types are converted to strings and complex types to common structures (dict, list)
     # This catches specialized SQL types like `Ranges`
-    # See case 3 for encoding Sequence and Mapping types as JSON
     if arrow_array is None and dlt_data_type is None:
         try:
             arrow_array = pa.array(column_data)
@@ -861,24 +882,6 @@ def convert_numpy_to_arrow(
                     ) from e
 
             arrow_array = pa.array(encoded_values)
-
-    # case 3: encode Sequence and Mapping types (list, tuples, set, dict, etc.) to JSON strings
-    if arrow_array is None and dlt_data_type == "json":
-        json_serialized_values: list[Union[bytes, None]] = []
-        for value in column_data:
-            if value is None:
-                json_serialized_values.append(None)
-                continue
-            try:
-                json_serialized_values.append(json.dumpb(value))
-            except TypeError as e:
-                raise PyToArrowConversionException(
-                    data_type=dlt_data_type,
-                    inferred_arrow_type=inferred_arrow_type,
-                    details="dlt failed to a JSON-serializable type.",
-                ) from e
-
-        arrow_array = pa.array(json_serialized_values).cast(pa.string())
 
     if arrow_array is None:
         raise PyToArrowConversionException(
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -11,19 +11,18 @@ from typing import (
     Iterable,
     Iterator,
     Sequence,
-    Tuple,
 )
 
 from dlt import version
 from dlt.common.exceptions import MissingDependencyException, DltException
 from dlt.common.schema.typing import C_DLT_ID, C_DLT_LOAD_ID, TColumnSchema, TTableSchemaColumns
-from dlt.common import logger, json
-from dlt.common.json import custom_encode, map_nested_in_place
-
+from dlt.common import logger
+from dlt.common.json import custom_encode
+from dlt.common.data_types.type_helpers import json_to_str
 from dlt.common.destination.capabilities import DestinationCapabilitiesContext
 from dlt.common.schema.typing import TColumnType
 from dlt.common.schema.utils import is_nullable_column
-from dlt.common.typing import StrStr, TFileOrPath
+from dlt.common.typing import StrStr, TFileOrPath, TDataItems
 from dlt.common.normalizers.naming import NamingConvention
 
 try:
@@ -103,6 +102,9 @@ class UnsupportedArrowTypeException(DltException):
         self._update_message()
 
 
+class PyToArrowConversionException(Exception):
+    pass
+
 def get_py_arrow_datatype(
     column: TColumnType,
     caps: DestinationCapabilitiesContext,
@@ -133,17 +135,18 @@ def get_py_arrow_datatype(
         precision, scale = column.get("precision"), column.get("scale")
         if (precision is None) and (scale is None):
             precision_tuple = caps.decimal_precision
-        elif (precision is None) and isinstance(scale, int):  # 2nd clause not requires, but it helps typechecker
-            precision_tuple = (caps.decimal_precision[0], scale)
+        elif (precision is None):
+            precision_tuple = caps.decimal_precision
             logger.warning(
                 f"Received decimal column hint `scale={scale}`, but `precision` not set. "
-                f"Will assume default destination capability `precision={caps.decimal_precision[0]}`"
+                f"Will assume default destination capability `(precision, scale) = {caps.decimal_precision}`"
             )
-        elif (scale is None) and isinstance(precision, int):
-            precision_tuple = (precision, caps.decimal_precision[1])
+        elif (scale is None):
+            # setting scale to 0 when unspecified is a common practice across databases 
+            precision_tuple = (precision, 0)
             logger.warning(
                 f"Received decimal column hint `precision={precision}`, but `scale` not set. "
-                f"Will assume default destination capability `scale={caps.decimal_precision[1]}`"
+                f"Will assume default destination capability `scale=0`"
             )
         else:
             precision_tuple = (precision, scale)
@@ -646,17 +649,11 @@ def concat_batches_and_tables_in_order(
     return pyarrow.concat_tables(tables, promote_options="none")
 
 
-def row_tuples_to_arrow(
-    rows: Sequence[Any], caps: DestinationCapabilitiesContext, columns: TTableSchemaColumns, tz: str
-) -> Any:
-    """Converts the rows to an arrow table using the columns schema.
-    1. Pivot rows into columns.
-    2. if `data_type` is available, use it to type cast the `pyarrow.array`
-    3. if `data_type` is missing,
-    2. Columns missing `data_type` will be inferred from the row data.
-    Columns with object types not supported by arrow are excluded from the resulting table.
+def transpose_rows_to_columns(rows: TDataItems, column_names: Iterable[str]) -> dict[str, Any]:  # dict[str, np.ndarray]
+    """Transpose rows (data items) into columns (numpy arrays). Returns a dictionary of {column_name: column_data}
+
+    Uses pandas if available. Otherwise, use numpy, which is slower     
     """
-    from dlt.common.libs.pyarrow import pyarrow as pa
     import numpy as np
 
     try:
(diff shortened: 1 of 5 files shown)


dlt/common/libs/pyarrow.py
```diff
@@ -654,104 +669,82 @@ def row_tuples_to_arrow(
...
+            arrow_array = pa.array(data.ravel(), type=pa.float64()).cast(get_py_arrow_datatype(column_schema, caps, tz), safe=False)
 
-    # If there are unknown type columns, first create a table to infer their types
-    if columnar_unknown_types:
-        new_schema_fields = []
-        for key in list(columnar_unknown_types):
-            arrow_col: Optional[pa.Array] = None
+        elif column_schema.get("data_type") is not None:
             try:
-                arrow_col = pa.array(columnar_unknown_types[key])
```
rudolfix, line 710: you removed this case. but we must handle this. it was added because there are NULL only columns for which we are not able to infer type. and those we remove because code downstream will not handle it.

you can do `remove_null_columns` at the end. but please add a test case for that!

dlt/common/libs/pyarrow.py
```diff
@@ -654,104 +669,82 @@ def row_tuples_to_arrow(
...
+            arrow_array = pa.array([
+                (
+                    None if s is None else 
+                    json.dumps(s) if not issubclass(type(s), set) else
+                    json.dumps(list(s))
+                )
+                for s in data.ravel()
+            ])
+
+        elif column_schema.get("data_type") in ("decimal", "wei") and issubclass(py_type, (str, float)):
```
rudolfix, line 694: here is code that fails for @bgarcevic . it seems that some dialects are mixing float and decimal values even if the data is declared decimals.
that is on top of dialects like Oracle (AFAIK) that say that column is decimal but still emit floats 🤯 

as mentioned below. try to create `arrow_array` first and then handle this via Exception. test cases:
- column declared as decimal but contains float
- as above but 10 first tows are NULL
- as above but contains both float and decimal values
- some of the decimal values have precision exceeding the destination precision (IMO arrow will fail on coercion)

out action in that case is to:
- cast floats to decimals like we do below (it may fail for the mixed case though...)
- convert all floats to decimals in Python and then try again
- in all cases where we lose precision (floats -> decimal, reduce decimal precision to fit into caps precision or when precision > decimal128 - another test case possible only with Postgres) we must warn the user

Also please handle the reverse situation

dlt/common/libs/pyarrow.py
```diff
@@ -654,104 +669,82 @@ def row_tuples_to_arrow(
...
+                if (
+                    column_schema.get("data_type") == "decimal"
+                    and (column_schema.get("precision") is None or column_schema.get("scale") is None)
+                ):
+                    msg += "\nWhen using `decimal` consider setting appropriate `precision` and `scale` hints. See reference: https://dlthub.com/docs/general-usage/schema/#tables-and-columns"
+                raise RuntimeError(msg) from e
 
+        else:
+            try:
+                flattened_data = data.ravel()  # use variable to avoid doing `.ravel()` twice if exception
```
rudolfix, line 719: you won't need that if you `ravel` all data at the beginning

rudolfix (changes requested):
thanks to refactoring the code. `row_tuples_to_arrow` is not only used by `sql_database` but also by data access and other internal functions that convert python lists into arrow tables. so it got way more important that it was. 

so if you started to refactor let's fix it a little bit and add more test cases. I left some proposal in the reviews. we'll need more tests as well. for example:
- lists/dicts/sets present in column declared as json, also when there are initial NULL values
- floats mixed with decimals also with NULL at start, when decimal is declared
- decimals with precision > decimal 128 being converted to arrow decimal

dlt/common/libs/pyarrow.py
```diff
@@ -131,11 +131,23 @@ def get_py_arrow_datatype(
         return pyarrow.string()
     elif column_type == "decimal":
         precision, scale = column.get("precision"), column.get("scale")
-        precision_tuple = (
-            (precision, scale)
-            if precision is not None and scale is not None
-            else caps.decimal_precision
-        )
+        if (precision is None) and (scale is None):
```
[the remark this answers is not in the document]
zilto: I implemented `3` and `4`, but didn't understand what logic to couple.

I see that `TypeMapperImpl` has methods `wei_precision` and `decimal_precision`, and I should retrieve that through the `DestinationCapabilitiesContext` in `dlt.common.libs.pyarrow.get_py_arrow_datatype()`?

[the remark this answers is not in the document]
rudolfix: > I see that TypeMapperImpl has methods wei_precision and decimal_precision, and I should retrieve that through the DestinationCapabilitiesContext in dlt.common.libs.pyarrow.get_py_arrow_datatype()?

those functions should do exactly the same you do below so I thought we could have common helper function. let's skip it for now. we have enough complexity already

dlt/common/libs/pyarrow.py
```diff
@@ -634,124 +649,151 @@ def concat_batches_and_tables_in_order(
...
+
+    dlt_data_type = column_schema.get("data_type")
+    inferred_arrow_type = get_py_arrow_datatype(column_schema, caps, tz) if dlt_data_type is not None else None
+    arrow_array = None
+
+    # base case (0): allow pyarrow to infer type, or create array of specified dlt specified type
+    # NOTE this is more restrictive than `pa.array(column_data).cast(inferred_arrow_type)`
+    try:
+        # type=None lets pyarrow infer the type from the data
+        arrow_array = pa.array(column_data, type=inferred_arrow_type)
```
rudolfix, line 687: this looks really good

IMO we can do another iteration
- we start exactly like here
- in case of any error we have 3 cases

(1) we know the inferred pyarrow type. in that case we let pyarrow to infer the type from the data and then we use cast to get the desired type
- this will handle floats to decimals
- this will handle decimal to decimal (if we allow for non safe casts)
- may handle many more cases
- can we make unsafe casts optional?
- if cast fails we are done. inferred type cannot be met so we raise exception

(2) we do not know desired type. in that case we let arrow to infer from the data
- here we are allowed to do fallback by converting into string
- the encode function does not return str in all cases. it has wrong signature :/ you can still use it but is some cases it returns dicts, then you need to dump the value explicitly

(3) the third case is when arrow fails on converint lists, dicts or sets, in that case we can dump all values with json parser.
- we can mix it with case (1) so we convert data before trying cast
- it looks to me that it replaces case (2) if we detect that lists are present. alternatively we can merge those but I'm not sure if this is the best idea (there are a lot of ifs when casting types)

dlt/common/libs/pyarrow.py
```diff
@@ -634,124 +649,151 @@ def concat_batches_and_tables_in_order(
...
+            f"Received `data_type={dlt_data_type}`. Failed to convert to arrow type {inferred_arrow_type}."
+        )
+
+    # case 1: coercing Python types unsupported by Arrow to a JSON friendly format
+    # e.g. dataclasses -> dict, UUID -> str
+    if arrow_array is None and dlt_data_type is None:
+        try:
+            arrow_array = pa.array([
+                None if item is None
+                else custom_encode(item)
```
rudolfix, line 709: see my comment below. custom encode has misleading signature (it returns json serializable type not always string)

rudolfix: feel free to change the signature...

dlt/common/libs/pyarrow.py
```diff
@@ -634,124 +649,151 @@ def concat_batches_and_tables_in_order(
...
+        logger.warning(
+            "Received `data_type='json'`, data requries serialization to string, slowing extraction. "
+            "Cast the JSON field to STRING in your database system to improve performance. "
+            "For example, create and extract data from an SQL VIEW that SELECT with CAST."
+        )
+        arrow_array = pa.array(
+            [
+                (
+                    None if item is None
+                    else json_to_str(list(item)) if issubclass(type(item), set)  # sets are not supported by json_to_str
```
rudolfix, line 732: this function should be called `typed_json_to_str` because before dumping it removes type encoding from str values. due to that it will be pretty slow. my take would be to use `json` directly, you can also experiment with `dumpb` and casting to str in arrow. this will be the fastest, we use `orjson` by default and it dumps to binary natively

dlt/common/libs/pyarrow.py
```diff
@@ -634,124 +649,151 @@ def concat_batches_and_tables_in_order(
...
         )
 
-    return pa.Table.from_pydict(columnar_known_types, schema=arrow_schema)
+    # case 3: backends store decimals in various format (binary, bytes, etc.) and returned Python values have inconsistent types
+    # `base case` fails because the Python type received is incompatible with the inferred pyarrow type (e.g., `pyarrow.decimal128()`) 
+    if arrow_array is None and dlt_data_type in ("decimal", "wei"):
+        # let pyarrow infer type (e.g., float, string) THEN cast it to the inferred type
+        tmp_array = pa.array(column_data)
+        try:
+            arrow_array = tmp_array.cast(inferred_arrow_type)
```
rudolfix, line 746: this is really cool. you didn't need `safe=True` here? you should be able to dump 64bit float into decimal so probably all right.

dlt/common/libs/pyarrow.py
```diff
@@ -634,124 +649,151 @@ def concat_batches_and_tables_in_order(
...
+
+    arrow_arrays = []
+    arrow_fields = []
+    for column_name, column_data in columnar.items():
+        column_schema = columns[column_name]
+
+        try:
+            arrow_array = convert_numpy_to_arrow(column_data, caps, column_schema, tz)
+        # TODO if converting to arrow fail, should we raise or skip column?
+        except PyToArrowConversionException as e:
```
rudolfix, line 777: we should definitely fail. what could be cool though if the exception we raise always has information about column name. maybe we can reuse the exception you worked on recently?

in that case user can remove data and try again

rudolfix (changes requested):
thx again for the refactor. it looks like we are close to something really decent and we'll fix a lot of edge cases!

also tests are looking good. we'll need maybe a few more use cases ie.
- data contains UUIDs (both for explicit text and no inferred type)
- data contains mixed types (just to see what happens, no inferred type should serialize everything to string)

dlt/common/libs/pyarrow.py
zilto: `safe=True` is the default. Setting `safe=False` is equivalent to setting `False` on [all of these settings](https://arrow.apache.org/docs/python/generated/pyarrow.compute.CastOptions.html#pyarrow.compute.CastOptions), which can influence date, time, int, float, decimal, and utf8 strings

zilto:
Linting seems to fail solely because of the removed annotation for `streamlit.runtime  # type: ignore`. Locally, mypy was complaining that I should remove it. Which is it?

dlt/common/libs/pyarrow.py
```diff
@@ -654,104 +742,210 @@ def row_tuples_to_arrow(
...
-                columnar_known_types[key] = arrow_col
-                new_schema_fields.append(
-                    pa.field(
-                        key,
-                        arrow_col.type,
-                        nullable=columns[key].get("nullable", True),
-                    )
-                )
+                    # the 3 types match those supported by `map_nested_in_place()`
+                    if isinstance(value, (tuple, dict, list)):
```
rudolfix, line 847: if we see dict or list we should do what we do in case (3) - json.dumpb. for tuples and sets - convert to list.

other values: as below (custom_encode but then encode to utf-8 so we have consistently bytes).

tldr;> dump nested types to json. or maybe I miss something here?

dlt/sources/sql_database/arrow_helpers.py
```diff
@@ -1,15 +1,14 @@
...
-
 from dlt.common.configuration import with_config
+from dlt.common.configuration.container import Container
 from dlt.common.destination import DestinationCapabilitiesContext
 
 
 @with_config
 def row_tuples_to_arrow(
     rows: Sequence[Any],
-    caps: DestinationCapabilitiesContext = None,
```
rudolfix, line 12: `with_config` decorator will inspect the signature of `row_tuples_to_arrow` and will inject `caps` automatically (caps derive from `ContainerInjectableContext` which request injection from `Container` which is also a config provider in `dlt`.)

to check that try wrap your test with
```py
with Container().injectable_context(destination_caps):
...
``
and you'll see `destination_caps` injected. (just explaining, see below)

btw. you code below is OK and we can keep it. just remove `@with_config` which is not needed and will slightly slower all calls to decorated function.

File: dlt/sources/sql_database/arrow_helpers.py [hunk 12]
```diff
@@ -20,5 +19,10 @@ def row_tuples_to_arrow(
     from dlt.common.libs.pyarrow import row_tuples_to_arrow as _row_tuples_to_arrow
 
     return _row_tuples_to_arrow(
-        rows, caps or DestinationCapabilitiesContext.generic_capabilities(), columns, tz
+        rows=rows,
+        caps=Container().get(DestinationCapabilitiesContext)
+        or DestinationCapabilitiesContext.generic_capabilities(),
+        columns=columns,
+        tz=tz,
+        safe_arrow_conversion=True,
```
rudolfix, line 27: OK! but if we keep with_config, `safe_arrow_conversion` (if present in the signature) would be configurable. just FYI. we can come back to that later

dlt/common/libs/pyarrow.py
```diff
@@ -103,6 +101,83 @@ def table_name(self, value: str) -> None:
         self._update_message()
 
 
+class PyToArrowConversionException(Exception):
```
rudolfix, line 104: please use `DltException` as base. People calling dlt methods should be able to filter exceptions coming from it.

rudolfix (changes requested):
code logic is OK now. thanks for working on `PyToArrowConversionException`. it is really cool!

there's a format error. I do `make format && make lint` before pushing to be sure all is good.

your mypy problem: it should not happen, we have
```
[mypy-streamlit]
ignore_missing_imports=true
```
.. and I do not see it on CI

rudolfix (approved):
LGTM! thanks for the tests!

PR #2291, opened 2025-02-12
Update dlt+ installation page

akelad:
### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

PR #2296, opened 2025-02-13
Docs: plus: adds AI Workflows section

burnash:


docs/website/docs/plus/features/ai.md
```diff
@@ -0,0 +1,159 @@
...
+# AI Workflows
+
+## Explore data in your dlt+ project with Claude Desktop
+
+This guide shows you how to connect your dlt+ project to Claude Desktop for data exploration. It uses the [Model Context Protocol (MCP)](https://modelcontextprotocol.io), an open protocol that allows large language models to retrieve information from external sources and use it in their responses.
+
+### Prerequisites
+- Python 3.10+
+-  A virtual environment (and all commands should be run from within it)
+- [Claude Desktop](https://claude.ai/download) installed
```
akelad, line 16: I think you can just refer to the installation guide of dlt+ as a prerequesite, currently in my PR: https://github.com/dlt-hub/dlt/pull/2291
And ofc still Claude Desktop. You can also just add the mcp support installation that you've added as a separate section here imo.

docs/website/docs/plus/features/ai.md
```diff
@@ -0,0 +1,159 @@
...
+title: AI Workflows
+description: Explore data in your dlt+ project with Claude Desktop using the Model Context Protocol
+keywords: [dlt+, Claude Desktop, MCP, Model Context Protocol]
+---
+
+# AI Workflows
+
+## Explore data in your dlt+ project with Claude Desktop
+
+This guide shows you how to connect your dlt+ project to Claude Desktop for data exploration. It uses the [Model Context Protocol (MCP)](https://modelcontextprotocol.io), an open protocol that allows large language models to retrieve information from external sources and use it in their responses.
```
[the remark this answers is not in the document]
akelad: Also based on the fact that I've removed the additional section header here, I think you can make all the sections below ## headings.

docs/website/docs/plus/features/ai.md
```diff
@@ -0,0 +1,159 @@
...
+
+When you click on the icon, you will see the "Available MCP Tools" popup with the tools description.
+
+![Claude Desktop available MCP tools](https://storage.googleapis.com/dlt-blog-images/plus/mcp/claude-desktop-available-tools.png)
+
+### Start chatting
+
+Now you can start chatting with Claude Desktop and ask it questions about the data in your dlt+ project.
+
+For example, we may ask "Which tables do I have in my pipeline?":
```
akelad, line 149: Maybe we should have a full list of available tools here/questions that work? Otherwise people don't know what they can ask Claude apart from this question

docs/website/docs/plus/features/ai.md
```diff
@@ -0,0 +1,159 @@
...
+- Python 3.10+
+-  A virtual environment (and all commands should be run from within it)
+- [Claude Desktop](https://claude.ai/download) installed
+
+### Install dlt+ with MCP support
+
+Make sure your virtual environment is activated, then:
+
+```bash
+pip install dlt-plus[mcp]
```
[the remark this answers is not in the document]
burnash: This is to make dlt-plus dependencies leaner by default, otherwise it brings a lot of sub-dependencies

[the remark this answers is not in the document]
akelad: Ok fair, makes sense

docs/website/docs/plus/features/ai.md
zilto: Good point! MCP have built-in `prompts` that we could use to register "questions that work". Users would be able to select them) We could also add a `help()` tool for the LLM to return a text describing its current capabilities.

docs/website/docs/plus/features/ai.md
burnash: Right, changed  in add522f931d7dd46a7882f3cec4ac210d3fcf0e9

burnash: I kept the `pip install dlt-plus[mcp]` in a separate section though. Then it's part of the prerequisites list, I believe, the list becomes too dense.

docs/website/docs/plus/features/ai.md
```diff
@@ -0,0 +1,159 @@
...
+```
+
+### Set up a test dlt+ project
```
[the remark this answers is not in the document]
burnash: Good point! I've tweaked the section to let users work with their own projects, but since the setup can be tricky, I think having a small example would be super helpful as a reference. Here the difference from tutorial (in my view) is that there's minimum explanation.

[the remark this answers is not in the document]
akelad: ok sounds good!

docs/website/docs/plus/features/ai.md
burnash: > Maybe we should have a full list of available tools here/questions that work? 

I don't want to put the full list here – it will likely change and become out of sync quickly. Instead I've added a tip to click on "available tools" button in Claude so the user can see what actual tools are available. (also added a sample list of questions).

@zilto, this is a very good point

akelad (approved):

PR #2314, opened 2025-02-17
fixes custom destination to go line by line in typed jsonl

sh-rp:
### Description
The custom destination would fail on multiline typed jsonl files. This was previously undetected, because orjsons typed_dump wraps after a large amount of items only.

### DIFF
--- a/dlt/destinations/job_impl.py
+++ b/dlt/destinations/job_impl.py
@@ -179,20 +179,21 @@ class DestinationJsonlLoadJob(DestinationLoadJob):
 
         # stream items
         with FileStorage.open_zipsafe_ro(self._file_path) as f:
-            encoded_json = json.typed_loads(f.read())
-            if isinstance(encoded_json, dict):
-                encoded_json = [encoded_json]
+            for line in f:
+                encoded_json = json.typed_loads(line)
+                if isinstance(encoded_json, dict):
+                    encoded_json = [encoded_json]
 
-            for item in encoded_json:
-                # find correct start position
-                if start_index > 0:
-                    start_index -= 1
-                    continue
-                # skip internal columns
-                for column in self._skipped_columns:
-                    item.pop(column, None)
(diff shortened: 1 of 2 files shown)


dlt/destinations/job_impl.py
```diff
@@ -179,20 +179,21 @@ def get_batches(self, start_index: int) -> Iterable[TDataItems]:
...
-                for column in self._skipped_columns:
-                    item.pop(column, None)
-                current_batch.append(item)
-                if len(current_batch) == self._config.batch_size:
-                    yield current_batch
-                    current_batch = []
-            yield current_batch
+            for line in f:
+                encoded_json = json.typed_loads(line)
+                if isinstance(encoded_json, dict):
```
rudolfix, line 184: is this to allow both typed jsonl (list only) and regular jsonl?

zilto: The changes make sense to me. 

AFAIK, before we had:
```python
encoded_json = json.typed_loads()   # expects a list

# special case if file contains a single JSONL record
if isinstance(encoded_json, dict): 
  encoded_json = [encoded_json] 
``` 

With the changes, the same lines have a slightly different meaning:
```python
for line in f:  # expects line to have a single JSONL record
   encoded_json = json.typed_loads(line)
   
   # base case where the line contains a single JSONL
   if isinstance(encoded_json, dict):
     encoded_json = [encoded_json]
```

zilto (approved):
LGTM. Looking at the `dlt.common.json` code, I understand that parsing the JSON file at once or line by line should have equivalent performance

dlt/destinations/job_impl.py
sh-rp: I don't remember, it was a while ago when I created this, but I'm pretty sure a test failed because there was a case where there was not a list but a dict json coming from the file..

PR #2316, opened 2025-02-17
Added a "LIMIT 1" clause to the fetch state query in BigQuery client

### LINKED ISSUE #2315: Unstable behaviour while fetching the pipeline state from BigQuery
### dlt version

1.4.0

### Describe the problem

We encountered very strange problem. While running as usual our pipelines against BigQuery we got an error:
`'google.api_core.exceptions.PermissionDenied'>\n403 request failed: the user does not have 'bigquery.readsessions.create' permission for 'projects/<our_project>'"`
Our service account has `BigQuery Job User` and it doesn't contain this permission.
But it worked before and it was still working for some similar pipelines with the same service account. 

### Expected behavior

`BigQuery Job User` is enough to fetch state from BigQuery and run pipelines.

### Steps to reproduce

These are not exact steps to reproduce but after debugging I have found this:
https://github.com/dlt-hub/dlt/blob/e677b33266c34cc1b3aab520a5b8182142bab93f/dlt/destinations/job_client_impl.py#L424-L430

While fetching state using this query solely depending on the pipeline name it can either succeed or fall. Weird thing that it is working when the query returns only 1 record (non-incremental table) and when query returned 619 records. But two other pipelines for which query returned 595 records - they fail.

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

_No response_

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

I don't think it's `dlt` issue. Something weird can happen under the hood of BigQuery. However, the problem can be easily fixed by adding a "LIMIT 1" clause to the original query. Overall, this change could slightly improve performance.

xneg:
### Description
 
Added a "LIMIT 1" clause to the fetch state query in BigQuery client since only the first record is used later in the code.

### Related Issues

- Fixes #2315

### DIFF
--- a/dlt/destinations/impl/bigquery/bigquery.py
+++ b/dlt/destinations/impl/bigquery/bigquery.py
@@ -48,8 +48,8 @@ from dlt.destinations.impl.bigquery.sql_client import BigQuerySqlClient, BQ_TERM
 from dlt.destinations.job_client_impl import SqlJobClientWithStagingDataset
 from dlt.destinations.job_impl import DestinationJsonlLoadJob, DestinationParquetLoadJob
 from dlt.destinations.job_impl import ReferenceFollowupJobRequest
-from dlt.destinations.sql_client import SqlClientBase
 from dlt.destinations.sql_jobs import SqlMergeFollowupJob
+from dlt.destinations.sql_client import SqlClientBase
 
 
 class BigQueryLoadJob(RunnableLoadJob, HasFollowupJobs):
--- a/dlt/destinations/impl/bigquery/bigquery.py
+++ b/dlt/destinations/impl/bigquery/bigquery.py
@@ -8,19 +8,20 @@ from google.api_core import retry
 from google.cloud import exceptions as gcp_exceptions
 from google.cloud.bigquery.retry import _RETRYABLE_REASONS
 
-from dlt.common import logger
+from dlt.common import pendulum, logger
 from dlt.common.destination import DestinationCapabilitiesContext, PreparedTableSchema
 from dlt.common.destination.client import (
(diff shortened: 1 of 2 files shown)


sh-rp:
I don't think this will work with all destinations, mssql has a different syntax for example.

xneg:
@sh-rp ah, you are right, thank you! I have to think about different solution.

xneg:
I couldn't think anything better than moving "LIMIT 1" clause to specific BigQuery client.

sh-rp:
@xneg sorry for not responding in more detail earlier, it just occured to me that we already have a solution for this, albeit not a very elegant one. Check the _limit_clause_sql function that has a special implementation for mssql and how it is used in the datasets. You can just use that to achieve what you want with a 2 line code change. Let me know if you need help.

xneg:
@sh-rp Thank you! It's good idea although a bit more than 2 lines of code. And thank you for hint with get_stored_schema, I applied changes there too.

rudolfix:
@sh-rp `ci from fork` label does not work in this repo :/ we only have it implemented on verified sources. so to test it we would need to change our workflows...

sh-rp:
@rudolfix I added ci from fork to our workflows a year ago or so and it is still there, but it does not work for some reason...

sh-rp:
@xneg I could not get our full test suite to run on your fork, so I have created a new PR. Thanks again for your work on this, we will credit you in the release for your contribution :)

xneg:
@sh-rp Sure thing! Thank you!
One question for future work: should I open PRs somehow differently? Should I recreate the fork?

PR #2329, opened 2025-02-18
Feat/2251-use vectorized scanner

### LINKED ISSUE #2251: Use USE_VECTORIZED_SCANNER=True for loading parquet files to Snowflake
### Feature description

Snowflake has a way of optimizing the loading of parquet data with `USE_VECTORIZED_SCANNER` option.

https://www.snowflake.com/en/engineering-blog/loading-terabytes-into-snowflake-speeds-feeds-techniques/
https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#label-use-vectorized-scanner

The feature request is to enable this by default (it's already marked as TODO: https://github.com/dlt-hub/dlt/blob/ae18acb96ce3d13c2dc9d60c59db5d3ffb9e8774/dlt/destinations/impl/snowflake/snowflake.py#L176)

Important note:
```
You can only enable the vectorized scanner if the following conditions are met:

The ON_ERROR option must be set to ABORT_STATEMENT or SKIP_FILE.

The other values, CONTINUE, SKIP_FILE_num, 'SKIP_FILE_num%' are not supported.
```

### Are you a dlt user?

Yes, I run dlt in production.

## Implementation details

1. * [ ] add an option to do that to SnowflakeConfiguration, disabled by default
2. * [ ] investigate if `test_parquet_loading` test works with option enabled, 
3. * [ ] write a specialized test for snowflake based on test above
4. * [ ] document the new option

anuunchin:
### Description
This PR enables the vectorized scanner for Parquet files when loading to Snowflake. the `ON_ERROR` setting defaults to `ABORT_STATEMENT`. 

### Related Issues

- Closes #2251

### DIFF
--- a/tests/load/pipeline/test_snowflake_pipeline.py
+++ b/tests/load/pipeline/test_snowflake_pipeline.py
@@ -1,5 +1,5 @@
+from copy import deepcopy
 import os
-import time
 import pytest
 from pytest_mock import MockerFixture
 
@@ -7,8 +7,6 @@ import dlt
 from dlt.common import pendulum
 from dlt.common.utils import uniq_id
 from dlt.destinations.exceptions import DatabaseUndefinedRelation
-from dlt.destinations.sql_client import SqlClientBase
-from typing import Any
 from dlt.load.exceptions import LoadClientJobFailed
 from dlt.pipeline.exceptions import PipelineStepFailed
 
@@ -16,10 +14,14 @@ from tests.load.pipeline.test_pipelines import simple_nested_pipeline
 from tests.load.snowflake.test_snowflake_client import QUERY_TAG
 from tests.pipeline.utils import assert_load_info, assert_query_data
 from tests.load.utils import (
+    TABLE_UPDATE_COLUMNS_SCHEMA,
+    assert_all_data_types_row,
     destinations_configs,
     DestinationTestConfiguration,
     drop_active_pipeline_data,
 )
+from tests.cases import TABLE_ROW_ALL_DATA_TYPES_DATETIMES
+
 
 # mark all tests as essential, do not remove
 pytestmark = pytest.mark.essential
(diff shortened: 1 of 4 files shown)


tests/load/pipeline/test_snowflake_pipeline.py
```diff
@@ -204,3 +206,73 @@ def test_char_replacement_cs_naming_convention(
...
+        """Fetch COPY INTO queries within the last minute containing the table name."""
+        query = """
+        SELECT QUERY_TEXT FROM TABLE(INFORMATION_SCHEMA.QUERY_HISTORY( \
+            END_TIME_RANGE_START => DATEADD(MINUTE, -5, CURRENT_TIMESTAMP), \
+            END_TIME_RANGE_END => CURRENT_TIMESTAMP \
+        )) WHERE QUERY_TEXT LIKE '%COPY INTO%'
+        """
+        return [row[0] for row in client.execute_sql(query) if table_name in row[0]]
+
+    os.environ["DESTINATION__SNOWFLAKE__USE_VECTORIZED_SCANNER"] = "TRUE"
```
sh-rp, line 234: A general approach: If you are testing a toggle as in this case, I think it makes sense to parametrize this. So you run the same block of code once with the property set to true and once with falls. Of course you will then also have different assertions at the end depending on which version is running, but that way you can ensure that the actual code that runs the pipeline is exactly the same, just once with the flag/toggle switched to on and once to off.

tests/load/pipeline/test_snowflake_pipeline.py
```diff
@@ -204,3 +206,66 @@ def test_char_replacement_cs_naming_convention(
...
+    os.environ["DESTINATION__SNOWFLAKE__USE_VECTORIZED_SCANNER"] = use_vectorized_scanner
+
+    pipeline, data = simple_nested_pipeline(
+        destination_config, f"vectorized_scanner_{use_vectorized_scanner}_{uniq_id()}", False
+    )
+    info = pipeline.run(data(), **destination_config.run_kwargs)
+    assert_load_info(info)
+
+    # Allow time for query history to update in Snowflake
+    time.sleep(2)
```
rudolfix, line 247: this is a good way to check if a proper query was generated. but there are other ways that IMO require way less work and are pretty standard
* refactor the code so you can unit test the essential function. this is already done with `gen_copy_sql`
* you can also using `Mocker` and `spy` to hijack the return value if the function value and check if VECTORIZE part is present
then you do not need to check Snowflake history - you can trust that the other tests are checking if the `gen_copy_sql` output is really executed

tldr;> nothing to change. but you can read more on mocking

tests/load/pipeline/test_snowflake_pipeline.py
```diff
@@ -204,3 +206,66 @@ def test_char_replacement_cs_naming_convention(
...
+        )) WHERE QUERY_TEXT LIKE '%COPY INTO%'
+        """
+        return [row[0] for row in client.execute_sql(query) if table_name in row[0]]
+
+    os.environ["DESTINATION__SNOWFLAKE__USE_VECTORIZED_SCANNER"] = use_vectorized_scanner
+
+    pipeline, data = simple_nested_pipeline(
+        destination_config, f"vectorized_scanner_{use_vectorized_scanner}_{uniq_id()}", False
+    )
+    info = pipeline.run(data(), **destination_config.run_kwargs)
```
rudolfix, line 243: here's the problem: `data()` returns a very simple set. what we need to do is to check if we can load all data types in both vectorization cases.  We have following test `test_all_data_types` (look for it). it is pretty crazy because it tests all our destinations. but you just need to take column_schemas, data_row = table_update_and_row() and load the row. make sure that nothing fails.

you could also port this part, but that's optional:
```py
with pipeline.sql_client() as sql_client:
        qual_name = sql_client.make_qualified_table_name
        db_rows = sql_client.execute_sql(f"SELECT * FROM {qual_name('data_types')}")
        assert len(db_rows) == 10
        db_row = list(db_rows[0])
        # parquet is not really good at inserting json, best we get are strings in JSON columns
        parse_json_strings = (
            destination_config.file_format == "parquet"
            and destination_config.destination_type in ["redshift", "bigquery", "snowflake"]
        )
        allow_base64_binary = (
            destination_config.file_format == "jsonl"
            and destination_config.destination_type in ["redshift", "clickhouse"]
        )
        allow_string_binary = (
            destination_config.file_format == "parquet"
            and destination_config.destination_type in ["clickhouse"]
        )
        # content must equal
        assert_all_data_types_row(
            db_row[:-2],
            parse_json_strings=parse_json_strings,
            allow_base64_binary=allow_base64_binary,
            allow_string_binary=allow_string_binary,
            timestamp_precision=sql_client.capabilities.timestamp_precision,
            schema=column_schemas,
        )
```
it verifies if we are gettig the right types back. AFAIK there was a small problem with JSON fields on Snowflake. so let;s check out

rudolfix (changes requested):
this good! thx for writing test and docs. please see my comments on testing all data types

anuunchin:
> lgtm! somehow the docs are not building, you'll have to investigate

There was an img not found error, so I rebased - should be fine now 👓

rudolfix (approved):
LGTM! and thanks for a good test and docs

PR #2337, opened 2025-02-20
Docs/dlt plus custom doc breadcrumbs

stantsvek:
### Description
Add dlt+ logo inside breadcrumbs and custom admonition before H1 for every DLT Plus page using [Swizzling](https://docusaurus.io/docs/2.x/swizzling#what-is-safe-to-swizzle)

image


VioletM:
Hi @stantsvek !
Is it possible to use lower letters in the header of the admonition?

dlt+ instead of DLT+

burnash (dismissed):
Looks very good, thank you @stantsvek!

burnash (approved):

PR #2343, opened 2025-02-24
adds support for http and path style for filesystem sql_client with s3

sh-rp:


smasyutin (approved):
It works with minio via http and `s3_url_style=path`.

Thank you @sh-rp !

PR #2364, opened 2025-03-01
synchronise pipeline protocol definitions and fix signature to use proper typing

### LINKED ISSUE #2363: Pipeline protocols are out of sync with implementation and do not play well with static type checkers.
### dlt version

1.7.0

### Describe the problem

We're using `pyright` in production. I noticed that `pyright` does not like the `pipeline.run(...)` [method](https://github.com/dlt-hub/dlt/blob/344f38d7fb6e1071ec31f3532e82da7cb58b629b/dlt/common/pipeline.py#L511) because the parameter types do not match their default values. For example, `destination: TDestinationReferenceArg = None` contains a type violation, because `None` cannot be assigned to `TDestinationReferenceArg`. We're adding `# type: ignore` comments to `pipeline.run(...)` calls because of this.

### Expected behavior

The solution is to wrap the types of all the named arguments in an `Optional`, so that `None` can be passed in by users without any errors from the static type checker. 

### Steps to reproduce

Try to define the following code snippet:
```python
def run(
    pipeline: SupportsPipeline,
    data: Any = None,
    *,
    destination: TDestinationReferenceArg | None = None
) -> LoadInfo:
        return pipeline.run(
            data,
            destination=destination
        )
```

and look for the following warning
```
Pyright: Argument of type "TDestinationReferenceArg | None" cannot be assigned to parameter "destination" of type "TDestinationReferenceArg" in function "run"
  Type "TDestinationReferenceArg | None" is not assignable to type "TDestinationReferenceArg"
    "None" is not assignable to "TDestinationReferenceArg"
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

I also noticed that the `run` [method](https://github.com/dlt-hub/dlt/blob/344f38d7fb6e1071ec31f3532e82da7cb58b629b/dlt/common/pipeline.py#L511) of the `SupportsPipeline` protocol is currently out of sync with the actual [implementation](https://github.com/dlt-hub/dlt/blob/344f38d7fb6e1071ec31f3532e82da7cb58b629b/dlt/pipeline/pipeline.py#L618). The same holds true for `__call__` in `SupportsPipelineRun`.

ubiquitousbyte:
### Description
Adds `Optional` to all arguments of `pipeline.run(...)` to make `None` values acceptable by all type checkers. 

### Related Issues

- Resolves #2363 

### Additional Context
Extracts the `pipeline.run(...)` arguments into a typed dictionary and uses `Unpack` to put them back in `SupportsPipeline`, `SupportsPipelineRun` and `Pipeline`. This way, the protocols and the implementation will always be in sync with each other. 

This also allows users of `dlt` to create custom wrappers around `run` with enhanced functionality. Here's an example of our custom `run` wrapper:

```python
def run(
    pipeline: SupportsPipeline,
    data: Any = None,
    *,
    destination: TDestinationReferenceArg | None = None,
    dataset_name: str | None = None,
    credentials: Any = None,
    table_name: str | None = None,
    write_disposition: TWriteDispositionConfig | None = None,
    columns: Sequence[TColumnSchema] | None = None,
    primary_key: TColumnNames | None = None,
    schema: Schema | None = None,
    loader_file_format: TLoaderFileFormat | None = None,
    schema_contract: TSchemaContract | None = None,
    state_store: PipelineStateStore | None = None,
) -> LoadInfo:
    with _maybe_reverse_etl_pipeline(pipeline, state_store) as pipeline:
        return pipeline.run(
            data,
            dataset_name=dataset_name,  # type: ignore
            destination=destination,  # type: ignore
            credentials=credentials,  # type: ignore
            table_name=table_name,  # type: ignore
            write_disposition=write_disposition,  # type: ignore
            columns=columns,  # type: ignore
            primary_key=primary_key,  # type: ignore
            schema=schema,  # type: ignore
            loader_file_format=loader_file_format,  # type: ignore
            schema_contract=schema_contract,  # type: ignore
        )
```

With the proposed changes here, we can refactor it into:
```python
def run(
    pipeline: SupportsPipeline,
    data: Any = None,
    *,
    state_store: PipelineStateStore | None = None,
    **kwargs: Unpack[SupportsPipelineRunArgs],
) -> LoadInfo:
    with _maybe_reverse_etl_pipeline(pipeline, state_store) as pipeline:
        return pipeline.run(data, **kwargs)
```


rudolfix (commented):
this looks really good! we were considering using Unpack here and in many different places (ie. to represent resource hints). the problem was that support in code editors was really bad:
- you could not see arguments in Unpack, just kwargs
- docstrings for an argument are not automatically displayed

I just checked VSCODE and it looks like it works! What editor are you on @ubiquitousbyte ? we need to check pycharm, cursor (is it VSCODE based?) to make sure we are not breaking dev experience

ubiquitousbyte:
I wasn't getting the docstrings either. Adding them to `SupportsPipelineRunArgs` has partially fixed this for me, i.e. they are rendered as can be seen in the attached image. I also get an error if I try to use an invalid type, indicating that type checking works. However, I am unable to get any doc strings for the `credentials`, `staging`, and `destination` arguments. :/
image

I am using the [Zed editor](https://zed.dev/) with `pyright` as a language server.

rudolfix:
Now I do not see any arguments at all... but there's still a chance. IMO if you convert `SupportsPipelineRun` into a `def` (function) with a all the arguments and docstrings you'll be able to use `dlt.common.typing:copy_sig` to copy the signature and docstring.

for methods (that take `self`) you may try to use `copy_sig_any` that should skip first parameter from copied signature. If that does not work, I'm pretty sure you can add another copy function that does that (using Concatenate).

see how methods above are used

rudolfix:
I'm closing this PR because there's no easy way to fix it. we did a lot of effort to correct typing in the resources and we may also come to this eventually

PR #2365, opened 2025-03-02
allows to apply table level hints via `make_hints`

rudolfix:
### Description
makes `additional_table_hints` part of `TResourceHints` which allows to update table-level hints dynamically


dlt/extract/hints.py
```diff
@@ -332,12 +336,14 @@ def apply_hints(
...
-                        t[k] = v  # type: ignore[literal-required]
+                if additional_table_hints:
+                    if t.get("additional_table_hints") is not None:
+                        for k, v in additional_table_hints.items():
+                            t["additional_table_hints"][k] = v
                     else:
-                        t.pop(k, None)  # type: ignore[misc]
-                t.pop("additional_table_hints", None)  # type: ignore
+                        t["additional_table_hints"] = additional_table_hints
+                else:
```
sh-rp, line 345: So if additional_table_hints is set to {} then they are fully popped and if there is at least one item, then it gets merged into the existing values? is this the correct behavior?

rudolfix: yes, all hints behave like this.

sh-rp (dismissed):

PR #2371, opened 2025-03-03
Fix/133 ibis iceberg delta

anuunchin:
### Description

Enable ibis expression support for iceberg and delta destinations which inherit from filesystem.

### Additional Context

UNDER CONSTRUCTION.


anuunchin:
I think, this should be uncommented in `pyproject.toml` as well, ibis doesn't have upper bounds for pyarrow now
```
# TODO: restore when ibis allows pyarrow 18
# [tool.poetry.group.ibis]
#optional = true

#[tool.poetry.group.ibis.dependencies]
# ibis-framework = { version = ">=9.0.0,<10.0.0", markers = "python_version >= '3.10'", extras = ["duckdb", "postgres", "bigquery", "snowflake", "mssql", "clickhouse"]}
```

dlt/common/destination/utils.py
```diff
@@ -210,7 +216,11 @@ def prepare_load_table(
         prep_table = fill_hints_from_parent_and_clone_table(tables, table)
         table_format = prep_table.get("table_format")
         # set table format if preferred
-        if destination_capabilities.preferred_table_format and not table_format:
+        if (
```
rudolfix, line 219: this is a good fix! but should be isolated to only `filesystem` destination. all other destinations follow (or will follow) the table format. look for `prepare_load_table` in `filesystem.py`. all destinations can customize how tables are prepared.

dlt/helpers/ibis.py
```diff
@@ -53,12 +60,18 @@ def create_ibis_backend(
...
     """Create a given ibis backend for a destination client and dataset"""
 
+    # ensure destination is a Destination instance
+    if not isinstance(destination, Destination):
+        destination = Destination.from_reference(destination)
+
     # check if destination is supported
-    destination_type = Destination.from_reference(destination).destination_type
-    if destination_type not in SUPPORTED_DESTINATIONS:
+    if not any(issubclass(destination.spec, spec_class) for spec_class in SUPPORTED_DESTINATIONS):
```
rudolfix, line 68: AFAK `issubclass` takes tuple of classes to check. so `any` wont be needed. pls check

rudolfix: also you could just raise in the last `else`, after all the possibilities are expired? in that case you may not need `SUPPORTED_DESTINATIONS` at all

tests/load/test_read_interfaces.py
```diff
@@ -748,7 +748,10 @@ def test_ibis_dataset_access(populated_pipeline: Pipeline) -> None:
     from dlt.helpers.ibis import SUPPORTED_DESTINATIONS
 
     # check correct error if not supported
-    if populated_pipeline.destination.destination_type not in SUPPORTED_DESTINATIONS:
+    if not any(
```
rudolfix, line 751: `SUPPORTED_DESTINATIONS` are misleading, now they contain configurations. maybe you should rename or get rid of this

rudolfix (changes requested):
looks good. we'll need to change how filesystem stored dlt tables at some point. not having the state and schemas in iceberg lake will soon be a problem

dlt/helpers/ibis.py
sh-rp: I think it would be cool to save the sqlglot dialect to be used in the destination capabilities, this will be very useful for the transformations stuff I am doing because I need to know which dialect to select there too. The default dialect in the generic_capabilities can be duckdb. Or is there a reason we should not have this info there?

rudolfix: let's do separate ticket

rudolfix (approved):
this is very good!

PR #2373, opened 2025-03-04
skips schema validation when loading from internal storage

### LINKED ISSUE #2356: Progressive slowdown and memory leak when creating multiple tables in separate pipeline runs
### dlt version

1.7.0

### Describe the problem

We are using dlt to extract data sequentially from 200+ different mongodb collections to duckdb/athena in separate pipeline runs. Even for similarly sized collections, we can see that the pipeline runs get significantly slower over time and that memory keeps increasing until the process is killed.

Under "Steps to reproduce" below I report a minimal example to reproduce the issue.

The memory leak reported in https://github.com/dlt-hub/dlt/issues/2221 might be related, although the issue I am reporting seems to be there in version 1.3.0 as well.

### Expected behavior

I would expect the running times of the different pipelines runs to be the same for similarly sized collections, and memory to remain stable. 

### Steps to reproduce

The issue can be easily reproduced by running the below script, that saves the same dummy data to separate tables (for `multi_table=True`). The issue is the same regardless of the destination ( I tried duckdb and filesystem).

```
import time
import dlt

multi_table = True

source = [{"col": 1}]

destination = dlt.destinations.duckdb("_pipeline_leak/data.duckdb")
destination = dlt.destinations.filesystem("/tmp/_dlt")
pipeline = dlt.pipeline(
    pipeline_name="my_pipeline",
    pipelines_dir="_pipeline_leak",
    destination=destination,
    dataset_name="my_data",
    dev_mode=True,
)

for i in range(2500):
    start_time = time.time()
    table_name = f"table_{i}" if multi_table else "table"
    load_info = pipeline.run(source, table_name=table_name, write_disposition="append")
    end_time = time.time()
    if i % 20 == 0:
        print(f"pipeline.run #{i} - took {end_time - start_time:.1f} seconds")
```

Running the above script you can see the following output

```
pipeline.run #0 - took 0.2 seconds
pipeline.run #20 - took 0.5 seconds
pipeline.run #40 - took 0.6 seconds
pipeline.run #60 - took 1.2 seconds
pipeline.run #80 - took 1.2 seconds
pipeline.run #100 - took 1.9 seconds
pipeline.run #120 - took 1.6 seconds
pipeline.run #140 - took 1.9 seconds
pipeline.run #160 - took 1.9 seconds
....
pipeline.run #2260 - took 26.9 seconds
pipeline.run #2280 - took 26.8 seconds
pipeline.run #2300 - took 26.8 seconds
pipeline.run #2320 - took 28.1 seconds
pipeline.run #2340 - took 27.6 seconds
pipeline.run #2360 - took 28.2 seconds
pipeline.run #2380 - took 28.5 seconds
pipeline.run #2400 - took 28.5 seconds
pipeline.run #2420 - took 29.1 seconds
pipeline.run #2440 - took 29.1 seconds
pipeline.run #2460 - took 28.9 seconds
```

The first time you run the pipeline it takes 0.2 seconds, the 2460th time you run it it takes 28.9 seconds.
Over time the memory usage increases by 4+ Gb

Image

Things I noticed:
1. The issue is not there if the data is appended to a unique table (`multi_table = False`)
2. Profiling the above script, you can see that the vast majority of the time is spent on `validation:verify_prop` called from `schema:from_dict`->`utils:validate_stored_schema`. This suggests something wrong there, possibly a variable that grows unboundedly 

Thanks for developing a great library!

### Operating system

Linux

### Runtime environment

Docker, Docker Compose

### Python version

3.10

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
1. fixes #2356 
2. makes schema caches private so they are discarded when relational normalizer is discarded
3. ~get_type_hints are super slow and were replaced with ___annotations__~ (not that easy)
4. schema validation is skipped on loading from pipeline schema storage (and kept for external sources)
5. also bumps `dlt` version

### DIFF
--- a/tests/common/normalizers/test_json_relational.py
+++ b/tests/common/normalizers/test_json_relational.py
@@ -422,7 +422,6 @@ def test_list_in_list() -> None:
 
     rows = list(schema.normalize_data_item(chats, "1762162.1212", "zen"))
     # both lists are json types now
-    print(rows)
     assert len(rows) == 3
     zen__webpath = [row for row in rows if row[0][0] == "zen__webpath"]
     assert all("list" in row[1] for row in zen__webpath)
--- a/tests/common/normalizers/test_json_relational.py
+++ b/tests/common/normalizers/test_json_relational.py
@@ -416,12 +416,13 @@ def test_list_in_list() -> None:
         "zen__webpath", parent_table_name="zen", columns=[{"name": "list", "data_type": "json"}]
     )
     schema.update_table(path_table)
+    # NOTE: we could reset normalizer in update_table so schema caches are removed. skip for now
+    schema.data_item_normalizer._reset()  # type: ignore[attr-defined]
     assert "zen__webpath" in schema.tables
-    # clear cache with json paths
-    normalize_helpers.is_nested_type.cache_clear()
 
     rows = list(schema.normalize_data_item(chats, "1762162.1212", "zen"))
     # both lists are json types now
+    print(rows)
     assert len(rows) == 3
     zen__webpath = [row for row in rows if row[0][0] == "zen__webpath"]
     assert all("list" in row[1] for row in zen__webpath)
--- a/.github/workflows/test_dbt_cloud.yml
+++ b/.github/workflows/test_dbt_cloud.yml
@@ -51,6 +51,8 @@ jobs:
 
       - name: Check out
         uses: actions/checkout@master
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
       - name: Setup Python
         uses: actions/setup-python@v4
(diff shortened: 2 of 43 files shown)


tests/common/normalizers/test_json_relational.py
```diff
@@ -416,12 +416,13 @@ def test_list_in_list() -> None:
...
     schema.update_table(path_table)
+    # NOTE: we could reset normalizer in update_table so schema caches are removed. skip for now
+    schema.data_item_normalizer._reset()  # type: ignore[attr-defined]
     assert "zen__webpath" in schema.tables
-    # clear cache with json paths
-    normalize_helpers.is_nested_type.cache_clear()
 
     rows = list(schema.normalize_data_item(chats, "1762162.1212", "zen"))
     # both lists are json types now
+    print(rows)
```
sh-rp, line 425: please remove :)

PR #2375, opened 2025-03-05
fix chess examples by using dlt rest client

mattseddon:
### Description

This PR fixes the errors in the chess examples from docs/examples.

Here is a recent CI failure: https://github.com/dlt-hub/dlt/actions/runs/13662290829/job/38196000155

### Additional Context

I have been poking around the repository and looking at bugs. I noticed that these examples are currently failing in the CI. The underlying issue is that the API is returning 404s for some of the requests made. If there is a better way to fix the examples then please LMK. I could not find an obvious way to add validation (i.e. if a player does not exist ignore them) in the docs.


mshirlaw:
LGTM ✅

sh-rp:
Hey @mattseddon, thanks your PR, much appreciated! Could you rebase your fork to the latest devel on this repo? The tests are not executing and I think the fix for this should be on our current devel branch.

mattseddon:
> Hey @mattseddon, thanks your PR, much appreciated! Could you rebase your fork to the latest devel on this repo? The tests are not executing and I think the fix for this should be on our current devel branch.

@sh-rp I'd assume that the `examples` tests are not running because they're all marked with the following:

```
@skipifgithubfork
@pytest.mark.forked
```

Unless something has changed?

Also, it looks like this change only addresses a transient issue with the `chess.com` API (the pipeline seems to be green again).

I think the reason that the change worked was that the `RESTClient` instantiates a new `Client` with `raise_for_status=False`. I tested by running the data into `duckdb`. When the API was misbehaving I got data like this:

image

Issue #2372, opened 2025-03-04
Extensions to ReadableDBAPIDataset and ReadableIbisRelation


sh-rp:
Goal:

We need a couple of extensions to our readable datasets to interact with the data in the context of dlt loads. We want to be able to retrieve all load ids that the dataset knows about, be able to sort by loading timestamp, be able to filter by success status and be able to filter rows of a table by selected load is. In short we need the following functions:

On `ReadableDBAPIDataset`:
* `def list_load_ids(status: Union[int, List[int]] = 0, limit: int = 10)` - should list all load ids of the dataset with the given `status` (may be list), ordered by load_id desc, and limited by `limit`. Be aware that the right schema needs to be selected
* `def last_load_id(status: Union[int, List[int]] = 0)` same as above but limited to 1. 

On `ReadableIbisRelation`:
* `filter_by_load_ids(load_ids: List[str])` - Should modify the ibis query so the result is filtered to only return rows of the given load_ids. In the case of child_tables, you'll need to join to the root table and filter by load_id there.
* `filter_by_last_load_id(status: Union[int, List[int]] = 0)` -  Filter by most recent load id that matches a given status.

I think it should be enough to support the above only if ibis datasets are available.

Further notes:
* Readable dataset docs are a good way to get started learning about accessing data in python: https://dlthub.com/docs/general-usage/dataset-access/dataset

zilto:
Closing as we're moving away from this structure

PR #2386, opened 2025-03-06
feat/23723 extend dataset querying API for incremental load

zilto:
Implements #2372 

## Description
This follows the specs. `last` was replaced by `latest`. Which is more consistent with the dlt language?

Currently, the internal of the ibis methods don't use proxying because the added layer make readability and debugging harder. I believe it's primarily meant for user-facing methods?

## Notes
I notice that the Ibis relationships use unbound tables. Is there a reason why we're not directly interacting with tables via the ibis connection?

### DIFF
--- a/dlt/common/schema/utils.py
+++ b/dlt/common/schema/utils.py
@@ -837,6 +837,24 @@ def get_root_table(tables: TSchemaTables, table_name: str) -> TTableSchema:
     return table
 
 
+def get_root_to_table_chain(tables: TSchemaTables, table_name: str) -> List[TTableSchema]:
+    """Return a list of tables ordered from root to child using the (row_key - parent_key) references.
+
+    Similar functions:
+        - `get_root_table()` returns the root of the specified child instead of the full chain
+        - `get_nested_tables()` returns all children of the specified table as a chain
+    """
+    chain: List[TTableSchema] = []
+
+    def _parent(t: TTableSchema) -> None:
+        chain.append(t)
+        if t.get("parent"):
+            _parent(tables[t["parent"]])
+
+    _parent(tables[table_name])
+    return chain[::-1]
+
+
 def get_nested_tables(tables: TSchemaTables, table_name: str) -> List[TTableSchema]:
     """Get nested tables for table name and return a list of tables ordered by ancestry so the nested tables are always after their parents
 
--- a/dlt/destinations/dataset/dataset.py
+++ b/dlt/destinations/dataset/dataset.py
@@ -1,3 +1,4 @@
+import textwrap
 from typing import Any, Union, TYPE_CHECKING, List
 
 from dlt.common.json import json
@@ -128,11 +129,14 @@ class ReadableDBAPIDataset(SupportsReadableDataset):
                 from dlt.helpers.ibis import create_unbound_ibis_table
                 from dlt.destinations.dataset.ibis_relation import ReadableIbisRelation
 
+                # TODO use the self.ibis() connection instead of `self.sql_client` and work against the Ibis Schema
+                # creating unbound tables would no longer be required and queries could be evaluated lazily
                 unbound_table = create_unbound_ibis_table(self.sql_client, self.schema, table_name)
                 return ReadableIbisRelation(  # type: ignore[abstract]
                     readable_dataset=self,
                     ibis_object=unbound_table,
                     columns_schema=self.schema.tables[table_name]["columns"],
+                    table_name=table_name,
                 )
             except MissingDependencyException:
                 # if ibis is explicitly requested, reraise
@@ -170,6 +174,29 @@ class ReadableDBAPIDataset(SupportsReadableDataset):
 
         # Execute query and build result dict
         return self(query)
+
+    def list_load_ids(self, status: Union[int, list[int]] = 0, limit: int = 10) -> list[str]:
+        """Return the list most recent `load_id`s in descending order.
+
+        If no `load_id` is found, return empty list.
+        """
+        status_value = (status,) if isinstance(status, int) else tuple(status)
+        # TODO protect from SQL injection
+        query = textwrap.dedent(f"""SELECT load_id 
+            FROM {self.schema.loads_table_name}
+            WHERE status IN {status_value}
+            ORDER BY load_id DESC
+            LIMIT {limit}""")
+        results = self.__call__(query=query).fetchall()
+        return [row[0] for row in results]
+
+    def latest_load_id(self, status: Union[int, list[int]] = 0) -> Union[str, None]:
+        """Return the latest `load_id`.
+
+        If no `load_id` is found, return None
+        """
+        results = self.list_load_ids(status=status, limit=1)
+        return results[0] if len(results) > 0 else None
 
     def __getitem__(self, table_name: str) -> SupportsReadableRelation:
         """access of table via dict notation"""
--- a/dlt/destinations/dataset/ibis_relation.py
+++ b/dlt/destinations/dataset/ibis_relation.py
@@ -1,10 +1,13 @@
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any, Union, Sequence
 
 from functools import partial
 
 from dlt.common.exceptions import MissingDependencyException
-from dlt.destinations.dataset.relation import BaseReadableDBAPIRelation
+from dlt.common.schema import Schema
+from dlt.common.schema.utils import get_root_table, get_root_to_table_chain
 from dlt.common.schema.typing import TTableSchemaColumns
+from dlt.destinations.dataset.relation import BaseReadableDBAPIRelation
 
 
 if TYPE_CHECKING:
@@ -54,11 +57,13 @@ class ReadableIbisRelation(BaseReadableDBAPIRelation):
         readable_dataset: ReadableDBAPIDataset,
         ibis_object: Any = None,
         columns_schema: TTableSchemaColumns = None,
+        table_name: str = None,
     ) -> None:
         """Create a lazy evaluated relation to for the dataset of a destination"""
         super().__init__(readable_dataset=readable_dataset)
         self._ibis_object = ibis_object
         self._columns_schema = columns_schema
+        self._table_name = table_name
 
     def query(self) -> Any:
         """build the query"""
@@ -84,6 +89,10 @@ class ReadableIbisRelation(BaseReadableDBAPIRelation):
     @columns_schema.setter
     def columns_schema(self, new_value: TTableSchemaColumns) -> None:
         raise NotImplementedError("columns schema in ReadableDBAPIRelation can only be computed")
+
+    @property
+    def table_name(self) -> str:
+        return self._table_name
 
     def compute_columns_schema(self) -> TTableSchemaColumns:
         """provide schema columns for the cursor, may be filtered by selected columns"""
@@ -173,6 +182,64 @@ class ReadableIbisRelation(BaseReadableDBAPIRelation):
             # here we just break the column schema inheritance chain
             return None
 
+    def _filter_nested_table(self, filtered_root_table: Any) -> Any:  # ibis.expr.types.Table
+        """Filter the current table based on a filtered root table.
+
+        This takes the Ibis Table expression `filtered_root_table`, which has filtered rows,
+        and propagate the selection of `_dlt_id` on the root table via `row_key -> parent_key`
+        recursively.
+
+        To visualize it:
+        ```
+        filtered_root_table = root.filter(root._dlt_load_id.isin(
+            ["foo", "bar"]
+        ))._dlt_id  # root row_key
+
+        child2.filter(child2._dlt_parent_id.isin(  # child2 parent_key
+            child1.filter(child1._dlt_parent_id.isin(  # child1 parent_key
+                filtered_root_table
+            ))._dlt_id  # child1 row_key
+        ))
+        ```
+        Takes as input and returns a `ibis.expr.types.Table`
+        """
+        from dlt.helpers.ibis import create_unbound_ibis_table
+
+        filtered_table = filtered_root_table
+        parent_row_key = None
+        for table in get_root_to_table_chain(self.schema.tables, self.table_name):
+            # the root table is already filtered, only set the parent_row_key
+            if parent_row_key is None:
+                parent_row_key = next(
+                    col_name
+                    for col_name, col in table["columns"].items()
+                    if col.get("row_key") is True
+                )
+                continue
+
+            parent_key = None
+            row_key = None
+            for col_name, col in table["columns"].items():
+                if col.get("parent_key") is True:
+                    parent_key = col_name
+                if col.get("row_key") is True:
+                    row_key = col_name
+
+            # should always match a column because `get_root_to_table_chain()` returns tables based on parent_key / row_key
+            assert parent_key is not None
+            assert row_key is not None
+
+            ibis_table = create_unbound_ibis_table(
+                self.sql_client, schema=self.schema, table_name=table["name"]
+            )
+            filter_clause = ibis_table[parent_key].isin(filtered_table[parent_row_key])
+            # TODO the only operation to proxy if required
+            filtered_table = ibis_table.filter(filter_clause)
+
+            parent_row_key = row_key
+
+        return filtered_table
+
     # forward ibis methods defined on interface
     def limit(self, limit: int, **kwargs: Any) -> "ReadableIbisRelation":
         """limit the result to 'limit' items"""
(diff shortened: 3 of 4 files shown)


dlt/destinations/dataset/dataset.py
```diff
@@ -171,6 +175,29 @@ def row_counts(
         # Execute query and build result dict
         return self(query)
 
+    def list_load_ids(self, status: Union[int, list[int]] = 0, limit: int = 10) -> list[str]:
```
sh-rp, line 178: This should return an SupportsReadableRelation, the user can convert this into a query string or run the .df() or arrow() functions on it. Check how it is done in the row_counts() method.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -186,6 +253,99 @@ def select(self, *columns: str) -> "ReadableIbisRelation":
...
+        load_ids = (
+            [
+                load_ids,
+            ]
+            if isinstance(load_ids, str)
+            else load_ids
+        )
+
+        root_table = get_root_table(self.schema.tables, self.table_name)
+        ibis_root_table = create_unbound_ibis_table(
```
sh-rp, line 269: You don't need to create any tables here, you should be able to get this via: self.readable_dataset["root_table_name"]

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -186,6 +253,99 @@ def select(self, *columns: str) -> "ReadableIbisRelation":
...
+            self.sql_client, schema=self.schema, table_name=root_table["name"]
+        )
+
+        # filter the root table
+        filtered_table = ibis_root_table.filter(ibis_root_table["_dlt_load_id"].isin(load_ids))
+        if root_table["name"] != self.table_name:
+            filtered_table = self._filter_nested_table(filtered_table)
+
+        # TODO use the proxies? it's a bit hard to do all that proxying
+        return self.__class__(
```
sh-rp, line 279: if you use the method above to get the root table, it will already be an ReadableIbisRelation, so you can just return it.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -173,6 +182,64 @@ def _get_filtered_columns_schema(self, columns: Sequence[str]) -> TTableSchemaCo
...
+    def _filter_nested_table(self, filtered_root_table: Any) -> Any:  # ibis.expr.types.Table
+        """Filter the current table based on a filtered root table.
+
+        This takes the Ibis Table expression `filtered_root_table`, which has filtered rows,
+        and propagate the selection of `_dlt_id` on the root table via `row_key -> parent_key`
+        recursively.
+
+        To visualize it:
+        ```
+        filtered_root_table = root.filter(root._dlt_load_id.isin(
```
sh-rp, line 194: I'm not quite sure what is going on here, but every child table has a dlt_root_id which points directly to the dlt_id of the root table, so you do not need to traverse the tablechain. All you really need to do is to join to the root table on dlt_root_id=dlt_load_id, join to the loads table and filter on status.

sh-rp (changes requested):
Thanks for the PR and good first approach! I think you can do it much easier, I left some comments. I understand that this proxy I have built is a bit hard to understand, but if you do it right you can just chain ibis expressions on to "self" and do not need to proxy anything. Also you don't need to process the full table chain for the derived tables.

dlt/destinations/dataset/ibis_relation.py
zilto: `readable_dataset` is passed at `__init__`, but it's not assigned to any attribute. Instead, it's passed to the `super().__init__()` and should be available through `self._dataset`

for my own learnings, any particular reason for this design?

sh-rp:
Even one more request: for all methods that filter by status, status should also be able to be "None", in that case no filtering on status happens and the loads table doesn't even get joined.

dlt/destinations/dataset/ibis_relation.py
zilto: I was under the impression that this should be possible, but based on the `populated_pipeline` found in existing tests, there's no such `_dlt_root_id` key.

```python
ibis_nested_rel = populated_pipeline.dataset(dataset_type="ibis").items__children

ibis_nested_rel.schema()
#  id             int64
#  _dlt_parent_id string
#  _dlt_list_idx  int64
#  _dlt_id        string>

ibis_nested_rel.schema.tables["items__children"]["columns"].keys()
# dict_keys(['id', '_dlt_parent_id', '_dlt_list_idx', '_dlt_id'])
```

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -85,6 +88,24 @@ def columns_schema(self) -> TTableSchemaColumns:
     def columns_schema(self, new_value: TTableSchemaColumns) -> None:
         raise NotImplementedError("columns schema in ReadableDBAPIRelation can only be computed")
 
+    @property
+    def table_name(self) -> str:
+        return self._table_name
+    
+    @cached_property
+    def load_table(self) -> "ReadableIbisRelation":
```
sh-rp, line 96: You should be able to get the loads table with self["_dlt_loads"] You don't need to construct any tables yourself, this class should be able to do all of this for you.

dlt/destinations/dataset/dataset.py
```diff
@@ -121,6 +119,49 @@ def _ensure_client_and_schema(self) -> None:
     def __call__(self, query: Any) -> ReadableDBAPIRelation:
         return ReadableDBAPIRelation(readable_dataset=self, provided_query=query)  # type: ignore[abstract]
 
+    def _filter_using_root_table(
```
sh-rp, line 122: I think the ticket maybe was not completely clear, but we only need these things for the ReadableIbisRelation. It will be quite a bit of work to get the pure sql statements to work on all destinations we support I think, so you can remove all the methods you defined on the non ibis relation, or rather throw an exception here that says Ibis is required to use these.

dlt/destinations/dataset/dataset.py
```diff
@@ -171,6 +213,78 @@ def row_counts(
         # Execute query and build result dict
         return self(query)
 
+    def list_load_ids(
```
sh-rp, line 216: see above

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -173,6 +194,94 @@ def _get_filtered_columns_schema(self, columns: Sequence[str]) -> TTableSchemaCo
...
+    ) -> "ReadableIbisRelation":
+        """Filter the root table and propagate to current table."""
+        # remove the proxy to facilitate transformations
+        if isinstance(values_to_include, ReadableIbisRelation):
+            values_to_include = values_to_include._ibis_object
+
+        # Case 1: if current table is root table, you can filter directly and exit early
+        dlt_root_table = get_root_table(self.schema.tables, self.table_name)
+        if dlt_root_table["name"] == self.table_name:
+            # doing ibis.expr.types.Table[COL_NAME] returns a column, but ReadableIbisRelation[COL_NAME] returns a table
```
sh-rp, line 210: I think you can access columns with self.columns(column_name) but I am not sure. In any case it would be nice to have a simple helper method for accessing columns so it is clearer what is going on.

sh-rp (changes requested):
Thanks for the further changes! I think this ticket is actually more complicated than I thought.

I think you can do something much shorter and nicer here. I will try to explain how I would do this and we can talk about it if it is unclear. In short, the resulting query that we want to have for brevity and performance reason would join the main table that we are querying to the loads table and then to ordering or filtering on the joined result. You do not want to be "joining" any table with "in" statements, this will be super slow. You can always inspect the result with self.query() to see what is going on. I would also create one function that joins the current table to the load table via any intermediary joins, you can probably make this recursive (meaning the function calls it self rather than going via a loop) but I am not sure if this will work. So the code would look something like this on the ReadableIbisRelation:

* `def _join_to_root_table(self) -> ReadableIbisRelation:` This method returns self if "parent" on the schema of this table is None, or returns a relation with a join to the root table which we then can further use to join to the loads_table. This is the method that can probably be done in a recursive way.

* `def _join_to_loads_table(self) -> ReadableIbisRelation:`: This method joins the current table to the loads table (it will also need to use the above methods) and return the result.

Now you can use the above methods to get an ibis relation that is joined to the loads table and do any other operations for the real task. You can filter on status, you can filter by load_id etc. and it will all operate on the joined table. The sql you want to have as an output would be something like this if you filter "table_name" by load_id greater than x via a root table.

`
SELECT table_name.col1, table_name.col2 FROM table_name JOIN root_table ON (table_name._dlt_root_id = root_table._dlt_load_id) JOIN _dlt_loads ON (root_table._dlt_load_id = _dlt_loads.load_id) WHERE _dlt_loads.load_id > x.
`

Let's not support the case where we don't have root_ids for now to make it easier, you can throw an exception with some help on how to create root ids.

zilto:
The change to `ReadableIbisRelation.__getitem__` triggers mypy because it doesn't respect the superclass signature. Let me know how I should approach this. Test coverage is not perfect, but I didn't want this PR to be a blocker for longer

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -173,6 +214,32 @@ def _get_filtered_columns_schema(self, columns: Sequence[str]) -> TTableSchemaCo
...
+        root_table_schema = get_root_table(self.schema.tables, self.table_name)
+        root_row_key = get_row_key(root_table_schema)
+        root_table = self._dataset.table(root_table_schema["name"])
+
+        joined_table = self.inner_join(
+            root_table.select(LOAD_ID_COL, root_row_key),
+            self[root_key] == root_table[root_row_key],
+        )
+        # `self` selects all columns from the original table
+        return joined_table.select(self, LOAD_ID_COL)  # type: ignore
```
sh-rp, line 241: Nice :)

dlt/common/schema/utils.py
```diff
@@ -829,6 +829,33 @@ def table_schema_has_type_with_precision(table: TTableSchema, _typ: TDataType) -
...
+def get_row_key(table: TTableSchema) -> str:
+    for col_name, col in table["columns"].items():
+        if col.get("row_key") is True:
+            return col_name
+    raise KeyError(
+        f"No row key found for table {table['name']}. This is likely a malformed table schema."
+    )
+
+
+def get_parent_key(table: TTableSchema) -> Optional[str]:
```
sh-rp, line 841: same as above

dlt/common/schema/utils.py
```diff
@@ -829,6 +829,33 @@ def table_schema_has_type_with_precision(table: TTableSchema, _typ: TDataType) -
...
+        if col.get("parent_key") is True:
+            return col_name
+    return None
+
+
+def is_root_table(table: TTableSchema) -> bool:
+    return table.get("parent") is None
+
+
+def get_root_key(table: TTableSchema) -> Optional[str]:
```
sh-rp, line 852: same as above

dlt/common/schema/utils.py
```diff
@@ -837,6 +864,24 @@ def get_root_table(tables: TSchemaTables, table_name: str) -> TTableSchema:
     return table
 
 
+def get_root_to_table_chain(tables: TSchemaTables, table_name: str) -> List[TTableSchema]:
```
sh-rp, line 867: i don't think you need this anymore, do you?

dlt/destinations/dataset/dataset.py
```diff
@@ -171,6 +169,40 @@ def row_counts(
...
+        self, status: Union[int, list[int], None] = 0, limit: Union[int, None] = 10
+    ) -> SupportsReadableRelation:
```
sh-rp, line 174: this should be implemented as ibis expression and raise if ibis is not installed

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -173,6 +214,32 @@ def _get_filtered_columns_schema(self, columns: Sequence[str]) -> TTableSchemaCo
             # here we just break the column schema inheritance chain
             return None
 
+    def _join_to_root_table(self) -> "ReadableIbisRelation":
+        """Join the current table to the root table. If the current table is root, it's a no-op."""
+        LOAD_ID_COL = "_dlt_load_id"
```
sh-rp, line 219: this constant already exists as C_DLT_LOAD_ID

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -173,6 +214,32 @@ def _get_filtered_columns_schema(self, columns: Sequence[str]) -> TTableSchemaCo
...
+
+        table_schema = self.schema.tables[self.table_name]
+        if is_root_table(table_schema):
+            return self
+
+        root_key = get_root_key(table_schema)
+        # TODO setup another case that traverse parent-row keys to join nested tables without root_key
+        if root_key is None:
+            raise KeyError(
+                "ReadableIbisRelation requires a `root_key` hint to join non-root tables. "
```
sh-rp, line 229: we should add a link to the transformation docs in the future for this exception.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -186,6 +253,76 @@ def select(self, *columns: str) -> "ReadableIbisRelation":
...
         return self._proxy_expression_method("select", *columns)  # type: ignore
 
+    # TODO ensure same defaults in ReadableDBAPIDataset and ReadableIbisRelation; and docstrings
+    def list_load_ids(
+        self, status: Union[int, list[int], None] = 0, limit: Optional[int] = None
+    ) -> "ReadableIbisRelation":
+        load_table = self._dataset.table(self.schema.loads_table_name)
+        if status is not None:
+            status = [status] if isinstance(status, int) else status
+            load_table = load_table.filter(load_table["status"].isin(status))
```
sh-rp, line 263: it would be interesting to find out wether `isin` with a one item list is the same speed as filter with that value. I'd hope the sql optimizer would find this automatically. Just a thought.

dlt/common/schema/utils.py
zilto: I'll remove it for now, but we might want it if we want to support these joins when `root_key=True` is missing

tests/load/test_read_interfaces.py
```diff
@@ -801,6 +800,235 @@ def test_ibis_dataset_access(populated_pipeline: Pipeline) -> None:
...
+@pytest.mark.no_load
+@pytest.mark.essential
+@pytest.mark.parametrize(
+    "populated_pipeline",
+    configs,
+    indirect=True,
+    ids=lambda x: x.name,
+)
+def test_ibis_filter_load_ids(populated_pipeline: Pipeline) -> None:
+    # TODO test filter when >1 successful load_id are present in destination
```
sh-rp, line 861: I agree with the comment, I add another load run that loads some stuff to the populated table in the setup method, you may need to update tests above

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -186,6 +253,76 @@ def select(self, *columns: str) -> "ReadableIbisRelation":
...
+
+        load_table = load_table.order_by(load_table["load_id"].desc())
+        # limit needs to be applied after sorting
+        if limit is not None:
+            load_table = load_table.limit(limit)
+
+        return load_table.load_id  # type: ignore
+
+    def latest_load_id(self, status: Union[int, list[int], None] = 0) -> "ReadableIbisRelation":
```
sh-rp, line 272: why not call the method above with a limit of one?

sh-rp (changes requested):
Thanks for the changes, it looks very cool already. What is missing:

* We should add the new functions to the docs
* You need to test with more than one successful load and ideally you would also have some orphan rows that are not matched to any load_id, but that is less important imho.
* I merged devel into this PR, so the tests are running, but not passing, check the results

zilto:
The docs linter is failing because methods are not available on non-Ibis datasets. Errors are not super helpful

```
lint_setup/lint_me/website/docs/general-usage/dataset-access/dataset_2.py:263: error: "SupportsReadableRelation" not callable  [operator]
lint_setup/lint_me/website/docs/general-usage/dataset-access/dataset_2.py:267: error: "SupportsReadableRelation" not callable  [operator]
lint_setup/lint_me/website/docs/general-usage/dataset-access/dataset_2.py:270: error: "SupportsReadableRelation" not callable  [operator]
lint_setup/lint_me/website/docs/general-usage/dataset-access/dataset_2.py:280: error: "SupportsReadableRelation" not callable  [operator]
lint_setup/lint_me/website/docs/general-usage/dataset-access/dataset_2.py:283: error: "SupportsReadableRelation" not callable  [operator]
lint_setup/lint_me/website/docs/general-usage/dataset-access/dataset_2.py:286: error: "SupportsReadableRelation" not callable  [operator]
lint_setup/lint_me/website/docs/general-usage/dataset-access/dataset_2.py:290: error: "SupportsReadableRelation" not callable  [operator]
```

rudolfix (changes requested):
This code does exactly what it should but IMO both implementation and user interface is too complicated.

From user perspective the dataset has two "ranges" of data
- all data (regular tables etc.)
- new data / incremental data (part of all data selected by a criteria set by transformation layer)

user should not be bothered with particular definition of the new data when implementing queries.

ideally what user needs to do is to do:
`github.issues.filter(type='new').df()` - all data
`github.issues.incremental().filter(type='new').df()` - incremental data

incremental information should be passed to dataset __init__ and carried to all relations. in case of `_dlt_load_id` it is super easy: it is just a list of load ids for which we should filter. that's it. `dataset` does not need to know any particular method to obtain this list.

my take:
1. all methods that obtain load ids from `_dlt_loads` table (with status, with most recent, with gt etc.) should be removed from here and placed in the transformation layer which knows which method to use. maybe we could use ANSI SQL to get those load ids
2. list of load ids will be small. but if we have 10k load ids, most database backends will create hashtable on `IN` clause. we do not need to worry
3. dataset should accept an instance of ie. `IncrementalFilter` class that puts it into incremental mode.

`IncrementalFilter` for `_dlt_load_id`:
1. applies to all tables/relations without any changes
2. keeps a list of load ids inside - created upfront by the transformation layer
3. implements actual `incremental()` that given a relation, returns filtered relation
- ideally works for both ibis and sqlglot/SQL text

this way we decouple dataset from a particular incremental implementation, provide for a simple UI and allow for any Incremental scheme including those that are different per table/relation

@zilto @sh-rp WDYT? I think 99% of the code could be kept. it just should go to specific classes or helper modules

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -154,6 +195,32 @@ def _get_filtered_columns_schema(self, columns: Sequence[str]) -> TTableSchemaCo
...
+                "Set `root_key=True` on the source or use `write_disposition='merge'`."
+            )
+        root_table_schema = get_root_table(self.schema.tables, self.table_name)
+        root_row_key: str = get_first_column_name_with_prop(
+            root_table_schema, column_prop="row_key"
+        )
+        root_table = self._dataset.table(root_table_schema["name"])
+
+        joined_table = self.inner_join(
+            root_table.select(C_DLT_LOAD_ID, root_row_key),
```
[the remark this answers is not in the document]
zilto: aaaah, likely the bug that made it fail for some tests. I properly normalized `C_DLT_LOAD_ID` in tests, but forgot about here

zilto:
In terms of user-facing API, we discussed 

```python
pipeline = dlt.pipeline(...)
dataset = pipeline.dataset()
reviews = dataset.reviews.incremental()
items = dataset.items
t = reviews.join(items, reviews.item_id == items.id)   
```

I believe `.incremental()` on a table (`dataset.reviews`) doesn't tell you that it's a filter / WHERE operation. It's also hard to distinguish from proxied Ibis methods.

## Suggestion
What about filtering the dataset directly? This `.filter()` method applies to the `_dlt_loads` table and is propagate to child tables (via root key, and future support for recursive). The convenience methods `.list_load_ids()` allow to retrieve the desired `load_id` values.

This would allow the following:

```python
pipeline = dlt.pipeline(...)
dataset = pipeline.dataset()
last_load_id = dataset.last_load_id 
reviews  = dataset.filter(last_load_id).reviews  #filtered
items = dataset.items  # not filtered
t = reviews.join(items, reviews.item_id == items.id)   
```

```python
pipeline = dlt.pipeline(...)
dataset = pipeline.dataset()
# .filter() returns a dataset with some attributes set
incremental_dataset = pipeline.dataset().filter(dataset.last_load_id)
reviews = incremental_dataset.reviews  #filtered
items = dataset.items  # not filtered
```

```python
pipeline = dlt.pipeline(...)
dataset = pipeline.dataset(filter="specific-load-id")
# both received the dataset filter
reviews = dataset.reviews  #filtered
items = dataset.items  # filtered
```

### Required changes
- `ReadableDBAPIRelation` (SQL and Ibis subclasses) can store a set of load_id as attribute (there's one object per table) 
- When `ReadableDBAPIRelation.query()` is called (i.e., execute the lazy query), filter the load table and apply to current table via join or filter
- `ReadableDBAPIDataset` gets a convenience method `list_load_ids()` that returns descending ordered list, can filter by status, and set limit 
- `ReadableDBAPIDataset` can store a set of load_id as attribute that's passed to the `ReadableDBAPIRelation` whenever `.table()` is used.

PR #2387, opened 2025-03-09
Updated the adjust a schema and sql configuration docs

dat-a-man:
### Description

Updated the caution about existing columns not modified automatically.


AstrakhantsevaAA (approved):
Good!

PR #2388, opened 2025-03-09
does not default to empty body in dependent resources

### LINKED ISSUE #2381: V1.8.0 - REST API source - Unexpected body in GET requests
### dlt version

1.8.0

### Describe the problem

Hi,

When I use the new syntax to add a param to the request, I face a 400 because dlt adds an empty body to the request.

Here is my resouce:

```python
{
    "name": "structures",
    "endpoint": {
        "path": "/api/v4/structures",
        "params": {
            "idCompany": "{resources.companies.id}",
        },
    },
},
```

And here is the request log with the empty body:

> {'method': 'GET', 'url': 'https://api.eurecia.com/eurecia/api/v4/structures?limit=1000&idCompany=COFAQ--------20191114-164321', 'headers': {'accept': 'application/json', 'token': '****'}, '_cookies': <RequestsCookieJar[]>, 'body': b'{}', 'hooks': {'response': []}, '_body_position': None}

As you can the, the previous resource didn't contain any body at all:

> {'method': 'GET', 'url': 'https://api.eurecia.com/eurecia/api/v4/companies?limit=1000', 'headers': {'accept': 'application/json', 'token': '****'}, '_cookies': <RequestsCookieJar[]>, 'body': None, 'hooks': {'response': []}, '_body_position': None}

Could you fix this please ?

### Expected behavior

Body should stay none by default

### Steps to reproduce

Setup a resource with the same pattern as bellow:

```
{
    "name": "structures",
    "endpoint": {
        "path": "/api/v4/structures",
        "params": {
            "idCompany": "{resources.companies.id}",
        },
    },
},
```

### Operating system

Windows

### Runtime environment

Local

### Python version

3.12

### dlt data source

_No response_

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
fixes #2381


dlt/sources/rest_api/config_setup.py
```diff
@@ -732,8 +732,10 @@ def process_parent_data_item(
         item, resolved_params, incremental, incremental_value_convert
     )
     expanded_path = expand_placeholders(path, params_values)
-    expanded_params = expand_placeholders(params or {}, params_values)
-    expanded_json = expand_placeholders(request_json or {}, params_values)
+    expanded_params = None if params is None else expand_placeholders(params, params_values)
```
burnash, line 735: I don't think there's need to do this for params, as

1) they get `{}` down the line in Requests's  if params are `None`:

```
        headers = {} if headers is None else headers
        params = {} if params is None else params
        hooks = {} if hooks is None else hooks
```

https://github.com/psf/requests/blob/5d31ddbf7b3f512dd3c61952d2c16e980a2b7dc1/src/requests/models.py#L274-L276

it's only json who gets the special treatment: 
https://github.com/psf/requests/blob/5d31ddbf7b3f512dd3c61952d2c16e980a2b7dc1/src/requests/models.py#L504-L515



2) `params` is not optional, but is (and [was](https://github.com/dlt-hub/dlt/blob/18ae4475e80cc0d88d1922a10a8b55021a45f966/dlt/sources/rest_api/__init__.py#L246)) an empty dict by default:
https://github.com/dlt-hub/dlt/blob/31907226025d52b83554d3b886e914fc8ef0675a/dlt/sources/rest_api/__init__.py#L243

burnash (approved):
Looks good. Thank you @rudolfix

Issue #2349, opened 2025-02-25
When using s3 as a staging destination for Redshift, allow users to specify region of s3 bucket


nsnider-fabric:
### Are you a dlt user?

Yes, I run dlt in production.

### Use case

When copying data from S3 to redshift, if the s3 bucket is in a different region than your redshift cluster, it is required to specify the REGION of the S3 bucket in your COPY statement. This is documented in AWS docs here - https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-source-s3.html#copy-region

With the current dlt <> redshift integration, there is no way for the user to specify the region in the COPY statement. This causes load jobs to fail permanently when the s3 bucket and redshift cluster are in different regions.

### Proposed solution

The COPY statement is templated here I believe: 

https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/redshift/redshift.py#L107

The goal would be to enable the user to configure the REGION with something like a `config.toml` or environment variables

sh-rp:
@djudjuu it's probably enough to add the region from the staging credentials in the `RedshiftCopyFileLoadJob` SQL statement, so a one line code change. For us to test it, you'd have to create an s3 bucket in a different region than our redshift database resides and see if it works, you can also ask the user to check it out, they'll have to specify a different region for the staging destination (filesystem) and end-destination (redshift)

yannik207:
hey @sh-rp ,
I updated the class locally and it seems to work for me.
Should the region have a default value like 'us-east-1' or what would you suggest?
How about error handling should this be implemented too?

I would be happy to contribute to this issue :)

PR #2389, opened 2025-03-09
Fix/add s3 region for redshift staging destination related to 2349

yannik207:
### Description
the pr is focusing on adding the s3 region for the copy into job

### Related Issues

- Fixes: https://github.com/dlt-hub/dlt/issues/2349

### Additional Context
Is the default value correct, or should I switch to another region or just do not use a default value

### DIFF
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -1,8 +1,8 @@
 import platform
 import os
+from dlt.common import logger
 
 from dlt.destinations.utils import is_compression_disabled
-from dlt.common import logger
 
 if platform.python_implementation() == "PyPy":
     import psycopg2cffi as psycopg2
@@ -75,32 +75,25 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
         self._sql_client = self._job_client.sql_client
         # we assume s3 credentials where provided for the staging
         credentials = ""
-        # get format
-        ext = os.path.splitext(self._bucket_path)[1][1:]
-
-        if self._staging_iam_role:
-            credentials = f"IAM_ROLE '{self._staging_iam_role}'"
-        elif self._staging_credentials and isinstance(
+        region = ""
+        if self._staging_credentials and isinstance(
             self._staging_credentials, AwsCredentialsWithoutDefaults
         ):
             aws_access_key = self._staging_credentials.aws_access_key_id
             aws_secret_key = self._staging_credentials.aws_secret_access_key
+            credentials = (
+                "CREDENTIALS"
+                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
+            )
+
             aws_region = self._staging_credentials.region_name
-
-            region_part = ""
             if aws_region:
-                if ext == "parquet":
-                    # Redshift doesn't support copying across regions for columnar data formats
-                    # https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html
-                    logger.warning("Redshift doesn't support copying Parquet files across regions. Region parameter will be ignored.")
-                else:
-                    region_part = f";region={aws_region}"
+                region = f"region '{aws_region}'"
 
-                credentials = (
-                    "CREDENTIALS"
-                    f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}{region_part}'"
-                )
-
+        if self._staging_iam_role:
+            credentials = f"IAM_ROLE '{self._staging_iam_role}'"
+        # get format
+        ext = os.path.splitext(self._bucket_path)[1][1:]
         file_type = ""
         dateformat = ""
         compression = ""
@@ -109,6 +102,11 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
             dateformat = "dateformat 'auto' timeformat 'auto'"
             compression = "" if is_compression_disabled() else "GZIP"
         elif ext == "parquet":
+            # Redshift doesn't support copying across regions for columnar data formats
+            # https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html
+            logger.warning("Redshift doesn't support copying Parquet files across regions. Region parameter will be ignored.")
+            region = ""
+
             file_type = "PARQUET"
             # if table contains json types then SUPER field will be used.
             # https://docs.aws.amazon.com/redshift/latest/dg/ingest-super.html
@@ -125,7 +123,8 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
                 {file_type}
                 {dateformat}
                 {compression}
-                {credentials} MAXERROR 0;""")
+                {credentials}
+                {region} MAXERROR 0;""")
 
 
 class RedshiftMergeJob(SqlMergeFollowupJob):
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -92,9 +92,9 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
             )
             if region_name:
                 credentials = (
-                "CREDENTIALS"
-                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key};region={region_name};'"
-            )
+                    "CREDENTIALS"
+                    f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key};region={region_name};'"
+                )
 
         # get format
         ext = os.path.splitext(self._bucket_path)[1][1:]
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -124,7 +124,7 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
                 {dateformat}
                 {compression}
                 {credentials}
-                {region} MAXERROR 0;""")
+                {region} MAXERROR 0;""")
 
 
 class RedshiftMergeJob(SqlMergeFollowupJob):
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -74,7 +74,7 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
 
     def run(self) -> None:
         self._sql_client = self._job_client.sql_client
-        # we ssume S3 credentials were provided for the staging
+        # we assume S3 credentials were provided for the staging
         credentials = ""
         if self._staging_iam_role:
             credentials = f"IAM_ROLE '{self._staging_iam_role}' REGION '{self._s3_region}'"
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -74,7 +74,7 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
 
     def run(self) -> None:
         self._sql_client = self._job_client.sql_client
-        # Assume S3 credentials were provided for the staging
+        # we ssume S3 credentials were provided for the staging
         credentials = ""
         if self._staging_iam_role:
             credentials = f"IAM_ROLE '{self._staging_iam_role}' REGION '{self._s3_region}'"
@@ -88,7 +88,7 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
                 f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key};region={self._s3_region}'"
             )
 
-        # Get format
+        # get format
         ext = os.path.splitext(self._bucket_path)[1][1:]
         file_type = ""
         dateformat = ""
@@ -99,12 +99,15 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
             compression = "" if is_compression_disabled() else "GZIP"
         elif ext == "parquet":
             file_type = "PARQUET"
+            # if table contains json types then SUPER field will be used.
+            # https://docs.aws.amazon.com/redshift/latest/dg/ingest-super.html
             if table_schema_has_type(self._load_table, "json"):
                 file_type += " SERIALIZETOJSON"
         else:
             raise ValueError(f"Unsupported file type {ext} for Redshift.")
 
         with self._sql_client.begin_transaction():
+            # TODO: if we ever support csv here remember to add column names to COPY
             self._sql_client.execute_sql(f"""
                 COPY {self._sql_client.make_qualified_table_name(self.load_table_name)}
                 FROM '{self._bucket_path}'
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -91,9 +91,10 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
                 f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}"
             )
             if region_name:
-                credentials += f"region={region_name};"
-
-            credentials += "'"  # Closing single quote
+                credentials = (
+                "CREDENTIALS"
+                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key};region={region_name};'"
+            )
 
         # get format
         ext = os.path.splitext(self._bucket_path)[1][1:]
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -104,7 +104,10 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
         elif ext == "parquet":
             # Redshift doesn't support copying across regions for columnar data formats
             # https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html
-            logger.warning("Redshift doesn't support copying Parquet files across regions. Region parameter will be ignored.")
+            logger.warning(
+                "Redshift doesn't support copying Parquet files across regions. Region parameter"
+                " will be ignored."
+            )
             region = ""
 
             file_type = "PARQUET"
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -65,28 +65,35 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
         file_path: str,
         staging_credentials: Optional[CredentialsConfiguration] = None,
         staging_iam_role: str = None,
-        s3_region: str = "us-east-1",  # Add region as a parameter
+        staging_region_name: str = None,
     ) -> None:
         super().__init__(file_path, staging_credentials)
         self._staging_iam_role = staging_iam_role
-        self._s3_region = s3_region  # Store region
+        self._region_name = staging_region_name
         self._job_client: "RedshiftClient" = None
 
     def run(self) -> None:
         self._sql_client = self._job_client.sql_client
-        # we assume S3 credentials were provided for the staging
+        # we assume s3 credentials where provided for the staging
         credentials = ""
         if self._staging_iam_role:
-            credentials = f"IAM_ROLE '{self._staging_iam_role}' REGION '{self._s3_region}'"
+            credentials = f"IAM_ROLE '{self._staging_iam_role}'"
+            if self._region_name:
+                credentials += f" REGION '{self._region_name}'"
         elif self._staging_credentials and isinstance(
             self._staging_credentials, AwsCredentialsWithoutDefaults
         ):
             aws_access_key = self._staging_credentials.aws_access_key_id
             aws_secret_key = self._staging_credentials.aws_secret_access_key
+            region_name = self._staging_credentials.region_name
             credentials = (
                 "CREDENTIALS"
-                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key};region={self._s3_region}'"
+                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}"
             )
+            if region_name:
+                credentials += f"region={region_name};"
+
+            credentials += "'"  # Closing single quote
 
         # get format
         ext = os.path.splitext(self._bucket_path)[1][1:]
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -103,7 +103,7 @@ class RedshiftCopyFileLoadJob(CopyRemoteFileLoadJob):
             compression = "" if is_compression_disabled() else "GZIP"
         elif ext == "parquet":
             # Redshift doesn't support copying across regions for columnar data formats
-            # https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html
+            # https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html  # noqa: E501
             logger.warning(
                 "Redshift doesn't support copying Parquet files across regions. Region parameter"
                 " will be ignored."
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -2,6 +2,7 @@ import platform
 import os
 
 from dlt.destinations.utils import is_compression_disabled
+from dlt.common import logger
 
 if platform.python_implementation() == "PyPy":
     import psycopg2cffi as psycopg2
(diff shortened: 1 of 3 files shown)


dlt/destinations/impl/redshift/redshift.py
```diff
@@ -65,25 +65,27 @@ def __init__(
...
+            credentials = f"IAM_ROLE '{self._staging_iam_role}' REGION '{self._s3_region}'"
         elif self._staging_credentials and isinstance(
             self._staging_credentials, AwsCredentialsWithoutDefaults
         ):
             aws_access_key = self._staging_credentials.aws_access_key_id
             aws_secret_key = self._staging_credentials.aws_secret_access_key
             credentials = (
                 "CREDENTIALS"
-                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
+                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key};region={self._s3_region}'"
```
djudjuu, line 88: question: the AwsCredentialsWithoutDefaults already has an attribute `region_name`. shouldnt we take that one then?

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -65,25 +65,27 @@ def __init__(
         file_path: str,
         staging_credentials: Optional[CredentialsConfiguration] = None,
         staging_iam_role: str = None,
+        s3_region: str = "us-east-1",  # Add region as a parameter
```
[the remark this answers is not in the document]
yannik207: Hey Julius,
Yeah that's what I asked in the issue #2349 :D 
but I think it make sense to swap it to not specifying it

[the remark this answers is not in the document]
yannik207: another suggestion might be to extract the region via boto3: something like this:

`import boto3
s3 = boto3.client("s3")
bucket_region = s3.get_bucket_location(Bucket=bucket_name)["LocationConstraint"]
`

dlt/destinations/impl/redshift/redshift.py
yannik207: yeah you are right.
I changed it and we can now use region from the AwsCredentialsWithoutDefaults if it is provisioned :)

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -76,15 +78,22 @@ def run(self) -> None:
...
         ):
             aws_access_key = self._staging_credentials.aws_access_key_id
             aws_secret_key = self._staging_credentials.aws_secret_access_key
+            region_name = self._staging_credentials.region_name
             credentials = (
                 "CREDENTIALS"
-                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
+                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}"
             )
+            if region_name:
```
djudjuu, line 93: lets not append the region to the credentials, but make it a separate {region} template variable. 
i guess if its "" it won't affect the query

yannik207: So you think it's harder to read/maintain or do you see any other reason? :)

dlt/destinations/impl/redshift/redshift.py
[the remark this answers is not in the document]
djudjuu: > Yeah that's what I asked in the issue #2349 :D
true :)
so let's not have a default value. 
I think its cleanes to make it optional string and load it from the config. We would then have to add it to RedshiftClientConfiguration-dataclass (redshift/configuration.py) and pass it along when the RedshiftCopyFileLoadJob gets instantiated in the client. We would then add a section to the docs, that warns about your issue, saying that you have to define the staging-region if you want to copy across regions.

either here: https://dlthub.com/docs/dlt-ecosystem/destinations/redshift#authentication-iam-role

I am wondering though if there is an even cleaner way....I have to dig a bit deeper and understand how credentials are resolved. I'll get back to you shortly

btw: the test we have to keep green is this one:
```
poetry run pytest tests/load/pipeline/test_pipelines.py::test_parquet_loading -k "redshift-parquet-staging-s3-role"
```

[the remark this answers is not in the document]
djudjuu: > import boto3 s3 = boto3.client("s3") bucket_region = s3.get_bucket_location(Bucket=bucket_name)["LocationConstraint"] 

hmm, my first reaction is that this would get messy. what if the role didnt get the right to query the region? I think its cleaner to keep it simple and have good docs for the config

dlt/destinations/impl/redshift/redshift.py
djudjuu: jup, just because its cleaner and technically not part of the credential

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -65,9 +65,11 @@ def __init__(
         file_path: str,
         staging_credentials: Optional[CredentialsConfiguration] = None,
         staging_iam_role: str = None,
+        staging_region_name: str = None,
```
djudjuu, line 68: this would have to be made part of the `RedshiftClientConfiguration` and then passed to the load-job

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -76,15 +78,23 @@ def run(self) -> None:
...
+            region_name = self._staging_credentials.region_name
             credentials = (
                 "CREDENTIALS"
-                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
+                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}"
             )
+            if region_name:
+                credentials = (
+                    "CREDENTIALS"
+                    f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key};region={region_name};'"
```
djudjuu, line 96: if region is unspecified, than there shouldnbt a "region=" at the end of the string (or tempvar)

dlt/destinations/impl/redshift/redshift.py
djudjuu: 2nd thought here: if we add to the configuration, we will have to add a lot of tests, i assume so maybe let me have a chat with my colleagues here
@sh-rp

dlt/destinations/impl/redshift/redshift.py
yannik207: but that's what I am doing if region_name is unspecified it will use the query template mentioned above. that's why I initially thought it would be easier to append the region if it is specified and to use two "different" templates :)

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -76,15 +78,23 @@ def run(self) -> None:
         credentials = ""
         if self._staging_iam_role:
             credentials = f"IAM_ROLE '{self._staging_iam_role}'"
+            if self._region_name:
+                credentials += f" REGION '{self._region_name}'"
```
sh-rp, line 82: @yannik207 it should be possible to use `self._staging_credentials.region_name`n here and not have to introduce a new setting, no? You should have this already set to the correct region since you need it to upload the files to the staging destination in the first place. So from my point of view we only need to add these two lines but insert from a different place. Let me know if I am missing something here.

sh-rp: PS: Adding the lines below also would be good.

dlt/destinations/impl/redshift/redshift.py
djudjuu: ok, lets not do the separate templateVar and just attach the region to the credentials

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -65,9 +65,11 @@ def __init__(
         file_path: str,
         staging_credentials: Optional[CredentialsConfiguration] = None,
         staging_iam_role: str = None,
+        staging_region_name: str = None,
     ) -> None:
         super().__init__(file_path, staging_credentials)
         self._staging_iam_role = staging_iam_role
+        self._region_name = staging_region_name
```
djudjuu, line 72: nitpick: should be `self._staging_region_name`

dlt/destinations/impl/redshift/redshift.py
djudjuu: okay so let's remove this line and always take the region from the credentials object

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -76,15 +78,23 @@ def run(self) -> None:
         credentials = ""
```
djudjuu, line 78: line 75 says it already: "we assume s3 credentials were provided"

its just not clear if they use iam-role or access+secret-key. the if-else here is a bit confusing.
i think  its safe to assume that its always AwsCredentials is always there here

djudjuu:
also please make a note in the docs-file
`dlt/docs/website/docs/dlt-ecosystem/destinations/redshift.md`  under the staging section that one has to set the 
`[destination.filesystem.credentials]` value in the config-toml if the bucket is in a different region than the redshift cluster

dlt/destinations/impl/redshift/redshift.py
yannik207: Hey @djudjuu && @sh-rp 
Sorry for asking again, but I'm a bit confused. Should we include the region when self._staging_iam_role is used? If not, where should the region value come from?

djudjuu: hey no worries,
the code is not making it obvious here, as we only support staging from s3, its safe to assume that the staging credentials will be there. 
so unless they're None (in which case staging is not being used and thus the region doesnt need to be specified) get the region from them and attach them to the credentials, which you can define either with iam_role or access_key&secret.

  try to get region from credentials
  if iam_role -> defined with iam_role ; region
  else -> define with accesskey&secret ; region

djudjuu: this CopyFileLoadJob can only be called from staging context:

image
and 
image

dlt/destinations/impl/redshift/redshift.py
yannik207: so switching from if-else to else? :)

dlt/destinations/impl/redshift/redshift.py
yannik207: > okay so let's remove this line and always take the region from the credentials object

The only thing that confuses me is the mentioned answer, because as far as I understood it now is that I should not remove the staging_region variable from the Load-Job.
I committed my solution and hopefully it is what you expected.

docs/website/docs/dlt-ecosystem/destinations/redshift.md
```diff
@@ -114,6 +114,8 @@ has_case_sensitive_identifiers=true
 
 Redshift supports s3 as a file staging destination. `dlt` will upload files in the parquet format to s3 and ask Redshift to copy their data directly into the db. Please refer to the [S3 documentation](./filesystem.md#aws-s3) to learn how to set up your s3 bucket with the bucket_url and credentials. The `dlt` Redshift loader will use the AWS credentials provided for s3 to access the s3 bucket if not specified otherwise (see config options below). Alternatively to parquet files, you can also specify jsonl as the staging file format. For this, set the `loader_file_format` argument of the `run` command of the pipeline to `jsonl`.
 
+**Note:** If the S3 bucket is in a different region than the Redshift cluster, you must set the [destination.filesystem.credentials] value in the config.toml file to ensure proper access.
```
djudjuu, line 117: maybe better:
>  you must set the `region_name` value in the `[destination.filesystem.credentials]`  section

djudjuu: Also, i've just found that REGION is not supported for parquet files, so lets add that you can not copy PARCET files across regions (atm)
https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html

yannik207: alright. just brainstorming how and where is the best solution to implement the logic, what's your opinion? :)

yannik207: My initial idea is to implement a check where, if file_type == "PARQUET" and region_name is set, an exception is raised (Exception("REGION is not supported for PARQUET files")). Does this approach make sense, or is there a better way to handle this?

djudjuu: tricky: 
we don't know the destination region , so we dont know if its actually is different from the staging
so if we...
- a) set the region by default, we will get an error if we're in the same region and copy parquet files
- b) we dont set the region, copy will fail if we're in different regions and copy anything non-columnar
 
i I say we set it if its there, unless file format is columnar, then we don't set it and log a warning (and also mention it in the docs)

djudjuu: ok, @yannik207 here is the new plan in pseudocode:


- get region from credentials (those will always be there, as will credentials.staging_region)
  - if the file format is parquet (see code line 90) set region to None and log a warning this might fail because redshift doesnt support copying across regions for columnar data formats. Also place comment with a link to the amazon docs above the code  that issues the warning 
- instantiate credentials either with iamrole or the other way, adding region if specified (personally, i think its less messy to just add the region as template var, but maybe you can make it look clean the other way too :)

the most important thing here is that the tests pass, ie both authentication ways work when the region is set and file-format is not parquet. atm the test that i linked above only tests parquet files, so we will need to adjust that one. you can try to do it if you want, otherwise I can work on it tomorrow.

thanks!

yannik207: Hey @djudjuu ,
first of all thanks for your patience really appreciate it!
I tried to implement a logic that I think mostly satisfies what you've designed with the pseudocode.

- I decided to go with the template var because I do not see a possibility to fill the region with None or leave it blank
- I do not 100% understand how to get the credentials if we use the IAM role 

thanks again for you help :)

yannik207: I want to avoid such a deeply nested if-else structure. so do you see a possibility to get the region from the credentials or does it maybe make sense to implement dedicated helper methods like: `_get_aws_credentials` && `_get_iam_role_credentials`?

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -74,20 +75,32 @@ def run(self) -> None:
...
             aws_access_key = self._staging_credentials.aws_access_key_id
             aws_secret_key = self._staging_credentials.aws_secret_access_key
-            credentials = (
-                "CREDENTIALS"
-                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
-            )
+            aws_region = self._staging_credentials.region_name
+
+            region_part = ""
+            if aws_region:
```
djudjuu, line 91: **issue**: this way you're only adding the region if not _staging_iam_role

imo both is okay:
```sh
COPY my_table
FROM 's3://my-bucket/path/'
CREDENTIALS 'aws_access_key_id=YOUR_KEY;aws_secret_access_key=YOUR_SECRET'
REGION 'us-east-1';
 ```
or 
```sh
COPY my_table
FROM 's3://my-bucket/path/'
CREDENTIALS 'aws_access_key_id=YOUR_KEY;aws_secret_access_key=YOUR_SECRET;region=us-east-1';
``` 
will work, so we really can just attach the region to the end the same way for both

yannik207: oh silly me! Now I get what you meant! thank you <3

djudjuu: cool! i imagined like this should work

```python
def run(self) -> None:
        self._sql_client = self._job_client.sql_client
        # we assume s3 credentials where provided for the staging
        credentials = ""
        # get format
        region_part = ""
        if self._staging_credentials and isinstance(
            self._staging_credentials, AwsCredentialsWithoutDefaults
        ):
            staging_region = self._staging_credentials.region_name
            region_part = f"region '{staging_region}'" if staging_region else ""
            aws_access_key = self._staging_credentials.aws_access_key_id
            aws_secret_key = self._staging_credentials.aws_secret_access_key
            credentials = (
                "CREDENTIALS"
                f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
            )

        if self._staging_iam_role:
            credentials = f"IAM_ROLE '{self._staging_iam_role}'"

        ext = os.path.splitext(self._bucket_path)[1][1:]
        file_type = ""
        dateformat = ""
        compression = ""
...
        if ext == "parquet":
               # Redshift doesn't support copying across regions for columnar data formats
               # https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html
              logger.warning("Redshift doesn't support copying Parquet files across regions. Region parameter will be ignored.")
            region_part = ""
....
sql = ...
                {credentials}
                {region_part}
                  MAXERROR 0;""")
 
```

yannik207: alright!
how should I proceed? Will you open your own pr? should I implement the logic? or what's your plan? :)

djudjuu: sorry, just saw this. pls do the changes to reshift.py file, i'll write the test.
the test will just cover the auth- with region-parameter for csv files WITHIN the same region

when we did both it would be cool if you could run it with your cross region setup and the two different auth ways

djudjuu: so i've pushed a test that checks if the copy command with region is working for iam and aws credentials (and is removed for parquet format.  so let me know if you have adjusted the redshift file and run the tests with the cross-region copy

tests/load/pipeline/test_redshift.py
```diff
@@ -41,3 +43,69 @@ def my_source() -> Any:
...
+    """
+    Tests if copy-command is constructed correctly for both iam-role and aws-credentials
+    when REGION is set. It should be set for jsonl, but unset for parquet
+    """
+    # initialize pipeline
+    pipeline : dlt.Pipeline = destination_config.setup_pipeline(
+        "redshift_region_test_" + uniq_id(), dataset_name="redshift_region_test_" + uniq_id()
+    )
+    # TODO assert that region is actually configured
+    # staging_credentials = ## how to get those??
```
djudjuu, line 89: @sh-rp can you help me get the region from the credentials here? I didnt manage to...

dlt/destinations/impl/redshift/redshift.py
yannik207: you okay with the new commit?

djudjuu: looks good! 
did you also test it in your cross region setup? 

one thing that remains is that I need a little help with one of my tests, I will ping my colleague about this.

You're really earning that contributor badge :)

yannik207: will test it in the evening and let you know :)

djudjuu: actually, i've made the tests a bit more exahustive. you can run them with
```
poetry run pytest tests/load/pipeline/test_redshift.py::test_copy_from_staging_with_region -x
```
before you do 
make dev
and then add the necessary credentials for your redshift and s3 bucket to 
tests/.dlt/secrets.toml (< you probably have to create this)

mine looks like:
```
[destination]
redshift.credentials="postgresql://loader:..../...

region_name = "eu-central-1"
aws_access_key_id="secret"
aws_secret_access_key="secret"

```

yannik207: hey @djudjuu ,
sorry got sick over the weekend.
yep seems to work for me :)

djudjuu:
@rudolfix  can you help me understand why those tests fail. It seems to me like in those tests there are no staging credentials set up. 

what I understand:
1. whenever we do staging for redshift, it has to be s3, as thats the only supported staging area
2. loadjobs can only be done from staging, so in RedshiftCopyFileLoadJob, we can safely assume that s3 was configured
( -> maybe not, this is probably where I am wrong?)

what I don't understand:
- from where which secrets are the failing CICD tests setting up the credentials/configurations
- are the CI/CD tests setting up a staging?
- why are unrelated tests failing: https://github.com/dlt-hub/dlt/actions/runs/14085468239/job/39449121322?pr=2389

where I am unhappy: redshift.pyL
image

- technically, this method reads like region and credentials can be empty strings, but only because we assume 1. and 2. are we sure that its not. 
the changes we did were quite minimal, only trying to redo credentials and the query command. 

#### on tests
-  I wanted to have test-cases for (columnar / non-columnar format) * (region specified / not specified) * (aws-creds / iam-role). would this be how you'd do it?

#### context
now we always try to read the AWS credentials
image

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -86,6 +86,12 @@ def run(self) -> None:
                 f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
             )
 
+            aws_region = self._staging_credentials.region_name
+            if aws_region:
+                region = f"region '{aws_region}'"
+
+        if self._staging_iam_role:
```
rudolfix, line 93: you changed the logic here so both IAM_ROLE and CREDENTIALS will be present. They are exclusive auth methods and in previous code IAM_ROLE took precedence. That must still be the case!

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -86,6 +86,12 @@ def run(self) -> None:
                 f" 'aws_access_key_id={aws_access_key};aws_secret_access_key={aws_secret_key}'"
             )
 
+            aws_region = self._staging_credentials.region_name
+            if aws_region:
```
rudolfix, line 90: why do you emit REGION only for CREDENTIALS authentication?
https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-source-s3.html

it is pretty clear that it is top level option in copy command. you should emit it always when present!

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -96,6 +102,14 @@ def run(self) -> None:
             dateformat = "dateformat 'auto' timeformat 'auto'"
             compression = "" if is_compression_disabled() else "GZIP"
         elif ext == "parquet":
+            # Redshift doesn't support copying across regions for columnar data formats
+            # https://docs.aws.amazon.com/redshift/latest/dg/copy-usage_notes-copy-from-columnar.html  # noqa: E501
+            logger.warning(
```
rudolfix, line 107: that's really true 🤯

djudjuu:
redshift does actually fail if you do copy command with region parameter for parquet

image

tests/load/pipeline/test_redshift.py
```diff
@@ -41,3 +45,71 @@ def my_source() -> Any:
...
+AWS_BUCKET = dlt.config.get("tests.bucket_url_s3", str)
+redshift_with_staging_configs = [
+    DestinationTestConfiguration(
+        destination_type="redshift",
+        staging="filesystem",
+        file_format="jsonl",
+        bucket_url=AWS_BUCKET,
+        extra_info="credential-forwarding",
+    ),
+    DestinationTestConfiguration(
```
rudolfix, line 59: All of the configurations except this one are already tested. I think you can just add this to staging destinations in `def destinations_configs(` and this will be picked up by many test cases

tests/load/pipeline/test_redshift.py
```diff
@@ -41,3 +45,71 @@ def my_source() -> Any:
...
+        destination_type="redshift",
+        staging="filesystem",
+        file_format="parquet",
+        bucket_url=AWS_BUCKET,
+        staging_iam_role="arn:aws:iam::267388281016:role/redshift_s3_read",
+        extra_info="s3-role",
+    ),
+]
+staging_regions = ["eu-central-1", ""]
```
rudolfix, line 77: yeah a real test would require us to setup cross-region bucket and load data from there. I think can stick to the setup we have

tests/load/pipeline/test_redshift.py
```diff
@@ -41,3 +45,71 @@ def my_source() -> Any:
...
+test_cases = list(itertools.product(redshift_with_staging_configs, staging_regions))
+
+
+@pytest.mark.parametrize("destination_config, staging_region", test_cases)
+def test_copy_from_staging_with_region(
```
rudolfix, line 82: this test is not really needed. we have plenty of tests like this and they will add `REGION` clause to the copy command. if you want to write a proper test:
1. we'd need to setup cross - region bucket (too much work)
2. you can write a unit test but that needs our code to be restructured. See how Snowflake destination is doing that (we separate sql generating part so it can be unit tested). ping me if you want to do it. but first make standard test to pass

dlt/destinations/impl/redshift/redshift.py
djudjuu: why?
if IAM_ROLE is defined, `credentials` will be overwritten to iam-string, so iam still takes precedence (if, if clause)

dlt/destinations/impl/redshift/redshift.py
djudjuu: because only the AswCredentials-objects read the region value from the secrets-toml and the comment above says its safe to "assume that s3 credentials were provided"

alternative1:

add staging_region to `RedshiftClientConfiguration` -> and pass it as an arg when the loadJob is created L190

I am assuming the user would put it in:
```
region_name="eu-central-1"
```

this was discussed previously but dave suggested to just read it from the credentials.

easier way is to just look for the region_name, but that still only works if credentials were instantiated:
```python
aws_region = (
            self._staging_credentials.get("region_name", "") if self._staging_credentials else ""
        )
        region = f"region '{aws_region}'" if aws_region else ""
```

tests/load/pipeline/test_redshift.py
djudjuu: done

PR #2396, opened 2025-03-11
feat: Expand sql table resource config

xneg:
### Description
It's possible to specify sql table using `config.toml` file ([see docs](https://dlthub.com/docs/dlt-ecosystem/verified-sources/sql_database/advanced#configuring-with-toml-or-environment-variables)).
```
[sources.sql_database.chat_message]
backend="pandas"
chunk_size=1000

[sources.sql_database.chat_message.incremental]
cursor_path="updated_at"
```
The problem is this setting-based approach allows only `append` strategy and doesn't allow to specify custom `primary` or `merge` keys.
While `primary` keys are reflected from source table it's not possible for views.
So this PR allows to specify `write_disposition`, `primary_key` and `merge_key` in `config.toml` file with some restrictions (see Additional Context):

```
[sources.sql_database.chat_message]
chunk_size=1000

[sources.sql_database.chat_message.incremental]
write_disposition.disposition = "merge"
primary_key = ["id", "updated_at"]
cursor_path="updated_at"
```

### Additional Context
There is a problem with not total alignment between `def resource(...)` method's arguments and fields in config classes.
Current config system doesn't support `Union` types (and probably it's not possible and there is no point) but this leads to the situation that we can't use plain string values:
```
[sources.sql_database.chat_message.incremental]
write_disposition= "merge"
primary_key = "id"
```
but we have to use more complex dict or lists values:

```
[sources.sql_database.chat_message.incremental]
write_disposition.disposition= "merge"
primary_key = ["id"]
```
If you have any better ideas how to realize this opportunity I'm open to adjust my PR or implement it different way.

### DIFF
--- a/dlt/sources/sql_database/__init__.py
+++ b/dlt/sources/sql_database/__init__.py
@@ -236,6 +236,9 @@ def sql_table(
     Returns:
         DltResource: The dlt resource for loading data from the SQL database table.
     """
+    # In case we get None from the config, we want to default to "append"
+    write_disposition = write_disposition if write_disposition else "append"
+
     _detect_precision_hints_deprecated(detect_precision_hints)
 
     if detect_precision_hints:
--- a/dlt/sources/sql_database/__init__.py
+++ b/dlt/sources/sql_database/__init__.py
@@ -174,6 +174,8 @@ def sql_table(
     resolve_foreign_keys: bool = False,
     engine_adapter_callback: Callable[[Engine], Engine] = None,
     write_disposition: TWriteDispositionConfig = "append",
+    primary_key: List[str] = None,
+    merge_key: List[str] = None,
 ) -> DltResource:
     """
     A dlt resource which loads data from an SQL database table using SQLAlchemy.
@@ -227,6 +229,9 @@ def sql_table(
             set transaction isolation level.
 
         write_disposition (TWriteDispositionConfig): write disposition of the table resource, defaults to `append`.
+        primary_key (List[str]): A list of column names that comprise a private key. Typically used with "merge" write disposition to deduplicate loaded data.
+        merge_key (List[str]): A list of column names that define a merge key. Typically used with "merge" write disposition to remove overlapping data ranges ie. to
+            keep a single record for a given day.
 
     Returns:
         DltResource: The dlt resource for loading data from the SQL database table.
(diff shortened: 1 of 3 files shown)


sh-rp:
Hi @xneg, thanks for this PR. It looks good so far, we would need to have some tests for this though. I think it would enough to write a test for various settings that go into the sql_table function and then check the returned DLTResource object if the hints where set correctly. You'll need to do it with and without a found table_obj.

dlt/sources/sql_database/helpers.py
```diff
@@ -408,3 +409,6 @@ class SqlTableResourceConfiguration(BaseConfiguration):
     defer_table_reflect: Optional[bool] = False
     reflection_level: Optional[ReflectionLevel] = "full"
     included_columns: Optional[List[str]] = None
+    write_disposition: Optional[TWriteDispositionDict] = None
+    primary_key: Optional[List[str]] = None
```
sh-rp, line 413: same comment about typing as above

tests/load/sources/sql_database/test_sql_database_source.py
```diff
@@ -271,6 +271,26 @@ def test_general_sql_database_config(sql_source_db: SQLAlchemySourceDB) -> None:
...
+    )
+    table = sql_table(
+        table="chat_message",
+        schema=sql_source_db.schema,
+        write_disposition="merge",
+        primary_key=["id"],
+        merge_key=["merge_id"],
+    )
+    # verify
+    assert table.write_disposition == "merge"
```
sh-rp, line 289: I think this is good enough, theoretically we could also test that the manually provided primary_key actually overwrites the one found in the reflection, if there is an easy way to do this, please do it, if not I think it is ok.

dlt/sources/sql_database/__init__.py
```diff
@@ -264,8 +272,15 @@ def sql_table(
     else:
         hints = {}
 
+    if primary_key:
```
sh-rp, line 275: maybe add a comment here why this is treated differently than the merge key: because we need to overwrite what is found in the reflection.

sh-rp (changes requested):
lgtm, just a few very small changes

tests/load/sources/sql_database/test_sql_database_source.py
```diff
@@ -271,6 +271,45 @@ def test_general_sql_database_config(sql_source_db: SQLAlchemySourceDB) -> None:
...
+        backend=backend,
+        schema=sql_source_db.schema,
+        write_disposition="merge",
+        reflection_level="full",
+        resolve_foreign_keys=True,
+        primary_key="created_at",
+    )
+
+    pipeline = make_pipeline("duckdb")
+    pipeline.extract(table)
```
djudjuu, line 308: I thought this way I was sure that the primary key got derived correctly, or should I just check on the hints?

sh-rp (approved):
lgtm

PR #2397, opened 2025-03-11
fix: better import exception for numpy

### LINKED ISSUE #2380: Missing numpy dependency when running with PyArrow >= 18
### dlt version

1.7

### Describe the problem

Because `PyArrow >= 18` moves `numpy` to optional runtime dependencies, when running pipelines with SQL sources and `pyarrow` backend, it'll fail with an import error

```
ModuleNotFoundError: No module named 'numpy'
```

### Expected behavior

Pipelines using `pyarrow` backend works with `pyarrow >= 18`

### Steps to reproduce

* Install dlt with `pyarrow >= 18`
* Run a dlt pipeline using SQL source and `pyarrow` backend

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12

### dlt data source

Any SQL source

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

zilto:
### Related Issues
Fixes #2380

### Context
When converting arbitrary row data to `pyarrow`, we try to use `pandas` to transpose row-wise data to column-wise data. If unavailable, we fallback to `numpy`

Since `pyarrow >= 18`, [numpy is no longer a dependency](https://github.com/apache/arrow/issues/25118). Our pyarrow code assumed that numpy was available. There's no error / change from related to our recent refactoring. 

### Problem
User reported (#2380) that a pipeline with SQL source fails with an import error when using `pyarrow >= 18` because of missing numpy dependency.

Note. the reported config `pyarrow==18, python==3.12` doesn't match the `pyproject.toml` constraints `pyarrow < 18` for `python < 3.13`. I would expect the package manager to enforce contraints and prevent this error. 

### Solution
Pyarrow is not a dependency of `dlt[sql_database]`, so numpy probably shouldn't be either. Also, except this code, only some LanceDB-related function seems to assume numpy to be available.

Exception handling now raises a message similar to when pyarrow is missing.
```
E               dlt.common.exceptions.MissingDependencyException: 
E               You must install additional dependencies to run dlt pyarrow helpers. If you use pip you may do the following:
E               
E               pip install "numpy"
E               
E               Numpy is required for this pyarrow operation
```

Unfortunately, it's hard to raise earlier than `DBApiCursor`'s arrow related methods

### DIFF
--- a/dlt/common/libs/numpy.py
+++ b/dlt/common/libs/numpy.py
@@ -1,6 +1,6 @@
 from dlt.common.exceptions import MissingDependencyException
 
 try:
-    import numpy
+    import numpy  # noqa: I251
 except ModuleNotFoundError:
-    raise MissingDependencyException("dlt Numpy Helpers", ["numpy"])
+    raise MissingDependencyException("dlt numpy Helpers", ["numpy"])
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -731,6 +731,13 @@ def transpose_rows_to_columns(
     Uses pandas if available. Otherwise, use numpy, which is slower
     """
     try:
+        from dlt.common.libs.numpy import numpy as np
+    except MissingDependencyException:
+        raise MissingDependencyException(
+            "dlt pyarrow helpers", ["numpy"], "Numpy is required for this pyarrow operation"
+        )
+
+    try:
         from pandas._libs import lib
 
         pivoted_rows = lib.to_object_array_tuples(rows).T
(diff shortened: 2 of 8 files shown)


sh-rp:
I think the correct fix might be creating a new arrow extra that has pyarrow and numpy as dependencies and using that extra in all other extra dependencies, but I am not sure. It looks like we only use numpy for for converting tuples to arrow tables, maybe this can be done in a fast way with some arrow interface on arrow 18, so we only need numpy for arrow versions where it is part of the dependencies.

zilto:
> I think the correct fix might be creating a new arrow extra that has pyarrow and numpy as dependencies and using that extra in all other extra dependencies, but I am not sure. 

I thought about this, but the other code paths that use `pyarrow` do not need to numpy. By searching for `np` and `numpy` in the codebase, it seems that only this pivot operation and LanceDB use numpy

rudolfix (changes requested):
1. we already have a helper for numpy that will raise user friendly exception (you may review it and add more info to it)
`from dlt.common.libs.numpy import numpy`
2. we need numpy only when there's no pandas so you can import in the `ImportError` handler (does not really matter AFAIK pandas has numpy in dependencies)
3. please add numpy to banned modules in `tox.ini`. this line slipped through when moving code from verified sources to here becasue direct numpy import want not banned
4. please mention numpy/panda in `sql_database` docs (it is already there but we may add a note about the newest arrow)

rudolfix (approved):
this is good and reduces entropy!

PR #2401, opened 2025-03-12
transformations tutorial

sh-rp:
### Description
A first tutorial for our dlt+ python based transformations

### DIFF
--- a/docs/website/docs/plus/getting-started/transformations.md
+++ b/docs/website/docs/plus/getting-started/transformations.md
@@ -357,7 +357,7 @@ def mutated_transformation(dataset: SupportsReadableDataset):
     return dataset.employees.mutate(static_column=1234)
 
 # we can supply a column hint for the static_column, which will work for sql transformations
-@dlt_plus.transform(transformation_type="sql", lineage_mode="best_effort", column_hints={"static_column": {"type": "double"}})
+@dlt_plus.transform(transformation_type="sql", lineage_mode="best_effort", columns={"static_column": {"type": "double"}})
 def mutated_transformation(dataset: SupportsReadableDataset):
     return dataset.employees.mutate(static_column=1234)
 
--- a/docs/website/docs/plus/getting-started/transformations.md
+++ b/docs/website/docs/plus/getting-started/transformations.md
@@ -1,0 +1,446 @@
+---
+title: Transformations tutorial
+description: Using the dlt+ transformations to transform data
+keywords: [transformations, dlt+, project]
+---
+
+### Topics covered in this tutorial
+
+This tutorial introduces you to the dlt+ transformations basics. You will learn how to:
+
+* transform existing tables with ibis expressions
+* sync datasets, remote to local, local to remote or remote to remote
+* use a local cache to run transformations across multiple datasets
+* retain full column level lineage during syncs and transformations
+
+### Topics covered in a future tutorial
+
+* Incremental transformations
+* Incremental and partial syncing
+* dbt based transformations
+* transformations dlt+ project and CLI integration
+
+## Prerequisites
+
+To follow this tutorial, make sure that:
+
+- dlt+ is set up according to the [installation guide](installation.md)
+- ibis-framework is installed in your environment
+- you're familiar with the [core concepts of dlt](../../reference/explainers/how-dlt-works.md)
+- you're familiar with our read access based on [ibis expressions](../../general-usage/dataset-access/dataset#modifying-queries-with-ibis-expressions)
+
+
+## Introductory notes
+
+To work through this tutorial, we will be using multiple local DuckDB instances to transform and sync data between them. In the real world, you will be able to use any destination/dataset that supports ibis expressions. We will also create an example dataset with a couple of tables to work with to demonstrate how you can make use of transformations.
+
+## Our example dataset
+
+Copy the pipeline below and run it in your environment:
+
+<details>
+
+<summary>Pipeline code</summary>
+
+
+```python
+
+import dlt
+import dlt_plus
+from dlt.common.destination.dataset import SupportsReadableDataset
+
+
+# Cities data
+@dlt.resource(write_disposition="replace")
+def offices():
+    yield [
+        {"id": 1, "city": "New York"},
+        {"id": 2, "city": "Berlin"},
+    ]
+
+@dlt.resource(write_disposition="replace")
+def departments():
+    yield [
+        {"id": 1, "department": "Engineering"},
+        {"id": 2, "department": "Marketing"},
+    ]
+
+# Employee data
+@dlt.resource(write_disposition="replace")
+def employees():
+    yield [
+        {"id": 1, "name": "Alice", "department_id": 1, "office_id": 1},
+        {"id": 2, "name": "Bob", "department_id": 2, "office_id": 2},
+        {"id": 3, "name": "Charlie", "department_id": 1, "office_id": 2},
+        {"id": 4, "name": "Diana", "department_id": 2, "office_id": 1},
+    ]
+
+# adds a custom hint to the name column, we will use this to test lineage later
+employees.apply_hints(columns={"name": {"x-pii": True}})  # type: ignore
+
+# Payroll data
+@dlt.resource(write_disposition="replace")
+def payroll():
+    yield [
+        {"id": 1, "employee_id": 1, "month": "2024-01", "amount": 7000},
+        {"id": 2, "employee_id": 1, "month": "2024-02", "amount": 7100},
+        {"id": 3, "employee_id": 2, "month": "2024-01", "amount": 5500},
+        {"id": 4, "employee_id": 2, "month": "2024-02", "amount": 5600},
+        {"id": 5, "employee_id": 3, "month": "2024-01", "amount": 5000},
+        {"id": 6, "employee_id": 3, "month": "2024-02", "amount": 5100},
+        {"id": 7, "employee_id": 4, "month": "2024-01", "amount": 6000},
+        {"id": 8, "employee_id": 4, "month": "2024-02", "amount": 6100},
+    ]
+
+# Create a DLT source with three resources
+@dlt.source
+def employee_data():
+    return offices(), employees(), payroll(), departments()
+
+# Create and run pipeline
+company_pipeline = dlt.pipeline(
+    pipeline_name="company_pipeline",
+    destination="duckdb",
+    dataset_name="company_dataset"
+)
+
+# Execute pipeline
+info = company_pipeline.run(employee_data())
+print(info)
+```
+</details>
+
+We have a straightforward company database with live employee master data containing names and payroll payments, as well as two tables for offices and departments. As always, you can inspect the contents of the dataset by running `dlt pipeline company_pipeline show` to open the dlt streamlit app.
+
+
+## A first few transformations 
+
+Transformations work very similarly to regular dlt.resources, which they are under the hood, and can be grouped into a regular `dlt.source`. Transformations take a populated dataset as input to build transformations from this source and move the computed data into the same or another dataset. If data stays within the same physical destination, you can execute the transformation as a SQL statement and no data needs to be moved to the machine where the dlt pipeline is running. For all other cases, data will be extracted the same way as with a regular dlt pipeline.
+
+Let's move some data from the company dataset into a fictional warehouse dataset. Let's assume we want the employee master data to be available in the warehouse dataset. We want to make sure not to include employee names in the warehouse dataset for privacy reasons.
+
+```python
+
+# get the company dataset
+company_dataset = company_pipeline.dataset()
+
+# print the employee table without the name column as arrow table
+print(company_dataset.employees.drop("name").arrow())
+
+# create the above select as a transformation in the warehouse dataset
+@dlt_plus.transform(transformation_type="python")
+def cleaned_employees(dataset: SupportsReadableDataset):
+    return dataset.employees.drop("name")
+
+# create the warehouse pipeline and run it:
+warehouse_pipeline = dlt.pipeline(
+    pipeline_name="warehouse_pipeline",
+    destination="duckdb",
+    dataset_name="warehouse_dataset"
+)
+
+# run the transformation
+info = warehouse_pipeline.run(cleaned_employees(company_dataset))
+print(info)
+
+```
+
+You can now inspect your warehouse pipeline with the dlt Streamlit app by running `dlt pipeline warehouse_pipeline show`. You should see the cleaned_employees table in the warehouse dataset, which is missing the name column.
+
+What is happening here?
+
+* We have defined one new transformation function with the `dlt.transform` decorator. This function does not yield rows but returns a single object. This can be a SQL query as a string, but preferably it is a ReadableIbisRelation object that can be converted into a SQL query by dlt. You can use all ibis expressions detailed in the dlt oss docs. Note that you do not need to execute the query by running `.arrow()` or `.iter_df()` or something similar on it.
+
+* The resulting table name is derived from the function name but may be set with the `table_name` parameter, like the `dlt.resource`.
+
+* For now, we always need to set a transformation type. This can be `python` or `sql`. `sql` transformations will be executed directly as SQL on the dataset and are only available for transformations where the source_dataset and destination_dataset are the same or on the same physical destination. More on this below.
+
+* All other resource parameters are available on the dlt.transform decorator, like `write_disposition`, `primary_key`, `columns`, etc. We will cover these in more detail below or in another tutorial.
+
+* Running the transformations is just like a regular pipeline run - you will have a complete schema and entries in the dlt_loads table.
+
+
+Here are a few more examples to illustrate what can be done with transformations using our company and warehouse pipelines:
+
+### Copy multiple tables at once into the warehouse dataset
+
+Copy two tables and clean up employee data for privacy:
+
+```python
+
+@dlt_plus.transform(transformation_type="python")
+def copied_employees(dataset: SupportsReadableDataset):
+    return dataset.employees.drop("name")
+
+@dlt_plus.transform(transformation_type="python")
+def copied_payroll(dataset: SupportsReadableDataset):
+    return dataset.payroll
+
+# we can group transformnations into a regular source
+def copy_source(dataset: SupportsReadableDataset):
+    return copied_employees(dataset), copied_payroll(dataset)
+
+# create the warehouse pipeline and run it:
+warehouse_pipeline.run(copy_source(company_dataset))
+
+```
+
+### Create a new joined and aggregated table
+
+We want a table in the warehouse that shows how many payments were made per employee in total. Fully in ibis expressions.
+
+```python
+
+@dlt_plus.transform(transformation_type="python")
+def payment_per_employee(dataset: SupportsReadableDataset):
+    employee_tables = dataset.employees
+    payroll_table = dataset.payroll
+    joined_table = employee_tables.join(payroll_table, employee_tables.id == payroll_table.employee_id)
+    return joined_table.group_by(joined_table.employee_id).aggregate(total_payments=joined_table.amount.sum())
+    
+
+# create the warehouse pipeline and run it:
+warehouse_pipeline.run(payment_per_employee(company_dataset))
+```
+
+
+Inspect your warehouse dataset with Streamlit - you will see the total amount per employee_id.
+
+### Create aggregated tables in the same dataset
+
+You can run a transformation against the same dataset that the source data comes from. This might be useful if you need some pre-computed stats (similar to a materialized view),
+or you need to do some cleanup after a load on the same dataset. Let's use the same transformation as above plus another one in a group. Since the source and destination datasets are the same, we can use SQL transformations, which are much faster.
+
+```python
+
+@dlt_plus.transform(transformation_type="sql", write_disposition="replace")
+def payment_per_employee_sql(dataset: SupportsReadableDataset):
+    employee_tables = dataset.employees
+    payroll_table = dataset.payroll
+    joined_table = employee_tables.join(payroll_table, employee_tables.id == payroll_table.employee_id)
+    return joined_table.group_by(joined_table.employee_id).aggregate(total_payments=joined_table.amount.sum())
+
+@dlt_plus.transform(transformation_type="python", write_disposition="replace")
+def rows_per_table(dataset: SupportsReadableDataset):
+    # this is a speical relation on the dataset which computes row_counts
+    return dataset.row_counts()
+
+
+# we can group transformnations into a regular source
+def stats_source(dataset: SupportsReadableDataset):
+    return rows_per_table(dataset), payment_per_employee_sql(dataset)
+
+# now run this transformation on the SAME pipeline
+company_pipeline.run(stats_source(company_dataset))
+```
+
+Check the company pipeline to see the aggregated tables in the same dataset.
+
+:::tip
+For the row_counts transformation, we use the Python transformation type, as dlt+ currently cannot compute the needed column hints for this table. More on SQL transformations below. You can observe that you can easily mix Python and SQL transformations in the same source.
+:::
+
+
+## A star schema example
+
+A common use case is to create a star schema with a fact table and a few dimension tables where each important column is at most one join away from the fact table. Let's create a star schema for our company dataset with payroll payments as facts and make the office and department tables directly joinable from the fact table:
+
+```python
+
+# let's create a new star warehouse for simplicity
+star_pipeline = dlt.pipeline(
+    pipeline_name="star_pipeline",
+    destination="duckdb",
+    dataset_name="star_dataset"
+)
+
+# the dimension tables can just be copied over, again we clean the personal data from the employees table
+
+@dlt_plus.transform(transformation_type="python")
+def employees(dataset: SupportsReadableDataset):
+    return dataset.employees.drop("name")
+
+@dlt_plus.transform(transformation_type="python")
+def departments(dataset: SupportsReadableDataset):
+    return dataset.departments
+
+@dlt_plus.transform(transformation_type="python")
+def offices(dataset: SupportsReadableDataset):
+    return dataset.offices
+
+# create the fact table from the payroll table with some joins
+@dlt_plus.transform(transformation_type="python")
+def payroll_facts(dataset: SupportsReadableDataset):
+    employee_tables = dataset.employees
+    payroll_table = dataset.payroll
+    joined_table = employee_tables.join(payroll_table, employee_tables.id == payroll_table.employee_id)
+    # we select the relevant columns, making department and office id directly available
+    return joined_table.select(
+        payroll_table.id,
+        payroll_table.employee_id,
+        payroll_table.month,
+        payroll_table.amount,
+        employee_tables.department_id,
+        employee_tables.office_id
+    )
+
+@dlt.source()
+def star_source(dataset: SupportsReadableDataset):
+    return payroll_facts(dataset),employees(dataset), departments(dataset), offices(dataset)
+
+
+# create the star schema in the star warehouse
+info = star_pipeline.run(star_source(company_dataset))
+print(info)
+
+# now that we have a nice star schema, presumable in a column oriented warehouse, we can aggregate
+# payments per department with only a single join:
+star_dataset = star_pipeline.dataset()
+fact_table = star_dataset.payroll_facts
+office_table = star_dataset.offices
+
+joined_table = fact_table.join(office_table, fact_table.office_id == office_table.id)
+print(joined_table.group_by(office_table.city).aggregate(total_payments=joined_table.amount.sum()).arrow())
+
+
+```
+
+:::tip
+For simply copying table data, you can also use the sync source demonstrated below. Also note that in future versions of the dlt+ transformation framework, there will be a template generator to create a star schema from a given dataset to give you a starting point.
+:::
+
+
+## Lineage
+
+dlt+ transformations contain a lineage engine that can trace the origin of columns resulting from transformations. You may have noticed that we added a custom hint to the name column in the employees table at the beginning of the page. This hint is a custom hint that we decided to add to all columns containing very sensitive data. Ideally, we would like to know which columns in a result are derived from columns containing sensitive data. dlt+ lineage will do just that for you. Let's run an aggregated join query into our warehouse again, but this time we will not drop the name column:
+
+```python
+
+@dlt_plus.transform(transformation_type="python")
+def payment_per_employee(dataset: SupportsReadableDataset):
+    employee_tables = dataset.employees
+    payroll_table = dataset.payroll
+    joined_table = employee_tables.join(payroll_table, employee_tables.id == payroll_table.employee_id)
+    return joined_table.group_by(joined_table.employee_id, joined_table.name).aggregate(total_payments=joined_table.amount.sum())
+
+# run the transformation
+info = warehouse_pipeline.run(payment_per_employee(company_dataset))
+print(info)
+
+```
+
+You can now inspect the schema of the warehouse and see that the name column of the aggregated table is also marked with our hint: `uv run dlt pipeline warehouse_pipeline schema`. dlt+ uses the same mechanism to derive precision and scale information for columns.
+
+You can control the lineage mode with the `lineage_mode` parameter on the dlt.transform decorator, which can be set to `strict`, `best_effort`, or `disabled`. The default is `best_effort`, which will try to trace the lineage for each column and set the derived hints but will not fail if this is not possible. SQL-based transformations that introduce non-traceable columns will require you to supply column type hints for any columns that could not be traced. For Python-based transformations, the column type can be derived from the incoming data. In `strict` mode, any transformation that introduces non-traceable columns will fail with a useful error. This mode is recommended if you must ensure which columns may contain sensitive data and need to prevent them from being unmarked. You can also disable lineage for a specific transformation by setting the `lineage_mode` to `disabled`. In this mode, the full schema will be derived from the incoming data and no additional hints will be added to the resulting dataset.
+
+### Untraceable columns
+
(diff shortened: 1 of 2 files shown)


docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+Let's move some data from the company dataset into a fictional warehouse dataset. Let's assume we want the employee master data to be available in the warehouse dataset. We want to make sure not to include employee names in the warehouse dataset for privacy reasons.
+
+```python
+
+# get the company dataset
+company_dataset = company_pipeline.dataset()
+
+# print the employee table without the name column as arrow table
+print(company_dataset.employees.drop("name").arrow())
```
djudjuu, line 128: I don't like the arrow table output. df() looks much nice imo
```suggestion
print(company_dataset.employees.drop("name").df())
```
image

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+@dlt_plus.transform(transformation_type="sql", write_disposition="replace")
+def payment_per_employee_sql(dataset: SupportsReadableDataset):
+    employee_tables = dataset.employees
+    payroll_table = dataset.payroll
+    joined_table = employee_tables.join(payroll_table, employee_tables.id == payroll_table.employee_id)
+    return joined_table.group_by(joined_table.employee_id).aggregate(total_payments=joined_table.amount.sum())
+
+@dlt_plus.transform(transformation_type="python", write_disposition="replace")
+def rows_per_table(dataset: SupportsReadableDataset):
+    # this is a speical relation on the dataset which computes row_counts
```
djudjuu, line 225:
```suggestion
    # this is a special relation on the dataset which computes row_counts
```

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+    return dataset.row_counts()
+
+
+# we can group transformnations into a regular source
```
djudjuu, line 229:
```suggestion
# we can group transformations into a regular source
```

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+print(info)
+```
+</details>
+
+We have a straightforward company database with live employee master data containing names and payroll payments, as well as two tables for offices and departments. As always, you can inspect the contents of the dataset by running `dlt pipeline company_pipeline show` to open the dlt streamlit app.
+
+
+## A first few transformations 
+
+Transformations work very similarly to regular dlt.resources, which they are under the hood, and can be grouped into a regular `dlt.source`. Transformations take a populated dataset as input to build transformations from this source and move the computed data into the same or another dataset. If data stays within the same physical destination, you can execute the transformation as a SQL statement and no data needs to be moved to the machine where the dlt pipeline is running. For all other cases, data will be extracted the same way as with a regular dlt pipeline.
```
djudjuu, line 118: this sentence is a bit circular: 
> Transformations take a populated dataset as input to build transformations from this source
not sure how to make it better though

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+Let's move some data from the company dataset into a fictional warehouse dataset. Let's assume we want the employee master data to be available in the warehouse dataset. We want to make sure not to include employee names in the warehouse dataset for privacy reasons.
```
djudjuu, line 120: suggestion: a bit more polished:
```suggestion
"Let's transfer some data from the company dataset into a fictional warehouse dataset. Let's assume we want the employee master data to be included in the warehouse, but the employee names should be excluded to ensure privacy."```

File: docs/website/docs/plus/getting-started/transformations.md [hunk 6]
```diff
@@ -0,0 +1,446 @@
...
+# Execute pipeline
+info = company_pipeline.run(employee_data())
+print(info)
+```
+</details>
+
+We have a straightforward company database with live employee master data containing names and payroll payments, as well as two tables for offices and departments. As always, you can inspect the contents of the dataset by running `dlt pipeline company_pipeline show` to open the dlt streamlit app.
+
+
+## A first few transformations
```
djudjuu, line 116: first few or few first?

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+# if we use best effort but switch the transformation type to sql, it will fail
+@dlt_plus.transform(transformation_type="sql", lineage_mode="best_effort")
+def mutated_transformation(dataset: SupportsReadableDataset):
+    return dataset.employees.mutate(static_column=1234)
+
+# we can supply a column hint for the static_column, which will work for sql transformations
+@dlt_plus.transform(transformation_type="sql", lineage_mode="best_effort", column_hints={"static_column": {"type": "double"}})
+def mutated_transformation(dataset: SupportsReadableDataset):
+    return dataset.employees.mutate(static_column=1234)
```
djudjuu, line 362: this one fails for me

image

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+```
+
+Check the company pipeline to see the aggregated tables in the same dataset.
+
+:::tip
+For the row_counts transformation, we use the Python transformation type, as dlt+ currently cannot compute the needed column hints for this table. More on SQL transformations below. You can observe that you can easily mix Python and SQL transformations in the same source.
+:::
+
+
+## A star schema example
```
djudjuu, line 244: **praise**: thats a helpful example, neatly packaged.

docs/website/docs/plus/getting-started/transformations.md
sh-rp: hm

sh-rp: should be `columns=`

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+# run it on the star pipeline
+info = star_pipeline.run(s)
+print(info)
+```
+
+Creating a sync source from any dataset without additional parameters will copy all rows of all tables and retain the full lineage of the data. Lineage is set to `strict` mode. If you want to copy just a few tables, you can do so by passing a list of tables to the sync source:
+
+```python
+# copy only the employees and payroll tables
```
djudjuu, line 402: **learning question**: whats the nicest way to actually reset the destination to verify the effect of those syncs? after the first one, the `offices` table is already there, so I am not really seeing a difference in pipeline state.
i've tried 
pipeline.run(refresh="drop_data"), but that doesnt actually affect the data at the destination only 
```python
                * `drop_data` - Wipe all data and resource state for all resources being processed. Schema is not modified.
```
(see here: https://github.com/dlt-hub/dlt/blob/74689ac2d7d71dec56abcb6822db1c81e2020c4d/dlt/pipeline/pipeline.py#L690)

so i've just gone to manually deleting the `star_pipeline` directory  in my `~/.dlt/pipelines` folder and the .duckdb in the repo...

anuunchin:
1.   Is the way the transformations are defined clear?
3.   Is it clear what lineage does?
4.   Is it clear how SQL and Python transformations differ, also from a technical point of view?
5.   Is it clear what the sync source does?
6.   Are there any obvious features missing? At the beginning of the tutorial, there is a list of planned features, but you can still list anything you think should be included.
7.   What would you add to this documentation?
8.   Did you find any bugs or unexpected behaviors?
9.   Which sections of the tutorial are too long or too brief?

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+    joined_table = employee_tables.join(payroll_table, employee_tables.id == payroll_table.employee_id)
+    return joined_table.group_by(joined_table.employee_id, joined_table.name).aggregate(total_payments=joined_table.amount.sum())
+
+# run the transformation
+info = warehouse_pipeline.run(payment_per_employee(company_dataset))
+print(info)
+
+```
+
+You can now inspect the schema of the warehouse and see that the name column of the aggregated table is also marked with our hint: `uv run dlt pipeline warehouse_pipeline schema`. dlt+ uses the same mechanism to derive precision and scale information for columns.
```
djudjuu, line 332: > . dlt+ uses the same mechanism to derive precision and scale information for columns.

its not clear to me what that means? Are numbers (e.g. in financial applications) affected by precision changes? what does scale mean here?

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+You can control the lineage mode with the `lineage_mode` parameter on the dlt.transform decorator, which can be set to `strict`, `best_effort`, or `disabled`. The default is `best_effort`, which will try to trace the lineage for each column and set the derived hints but will not fail if this is not possible. SQL-based transformations that introduce non-traceable columns will require you to supply column type hints for any columns that could not be traced. For Python-based transformations, the column type can be derived from the incoming data. In `strict` mode, any transformation that introduces non-traceable columns will fail with a useful error. This mode is recommended if you must ensure which columns may contain sensitive data and need to prevent them from being unmarked. You can also disable lineage for a specific transformation by setting the `lineage_mode` to `disabled`. In this mode, the full schema will be derived from the incoming data and no additional hints will be added to the resulting dataset.
```
djudjuu, line 334: this section is a bit long and would profit from being turned into some kind of a list or table or the like.

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+* transformations dlt+ project and CLI integration
+
+## Prerequisites
+
+To follow this tutorial, make sure that:
+
+- dlt+ is set up according to the [installation guide](installation.md)
+- ibis-framework is installed in your environment
+- you're familiar with the [core concepts of dlt](../../reference/explainers/how-dlt-works.md)
+- you're familiar with our read access based on [ibis expressions](../../general-usage/dataset-access/dataset#modifying-queries-with-ibis-expressions)
```
djudjuu, line 30: question: why 'read' access? that confuses me here....is that something other than using  ibis expressions to access the dataset? is it a different way to say this `methods like limit and select won't work. Include any filtering or column selection directly in your SQL query.`
after the reading the link, I still felt like I didnt check that box

djudjuu:
### general feedback

+I do like the feature a lot. its powerful and succint. `sync_source()` feels like magic! I never want to live without it again.

+the tutorial arrived swiftly and in a neat packaging (just kidding) 

+It was easy to do though, and the transformations interface is pretty clear.

+Its great that there are many hints about stuff that will be supported/come out in the future.  (although sometimes its in a tip, sometimes in the text, sometimes a bulleted list). Especially the star-schema template sounds like a really cool idea!

+the sql vs python section is pretty clear to me!

+doing the homework was fun! Its nice to accomplish that much with so few lines of code

-Tying into that, I always think the better the code-snippets that illustrate individual concepts tie into one another the better, kind of like hooking into the power of storytelling (if u get what i mean). (Counterexample, is when syncing with just a few tables, makes no visible change, because the table is already there.) It's not obvious how to do that, but i'd also volunteer to do it

-I am missing a general introductory definition of what transformations are 
 (my attempt:  operations that modify, enrich or restructure raw data using sql or python expressions) that act as close as possible to your data and are thus very fast. wrapped in the dlt_plus.transform they do....). Quite possibly its obvious to dlt-users, but I still think that being more self-contained knowledge-wise (vs requiring much implicit knowledge) is a good value, potentially for newcomers (if we aim to be the defacto standard) and possibly as well for LLMs.
 One easy sentence would be to mention where they go beyond what `@dlt.transformers` can define.

-I found no bugs, everything just worked, except when i tried to pipe a dlt_plus.transform into a dlt.transformer
(when I read that dlt_plus.transform with python is a resource under the hood I just had to try it) 
-> i tried to pipe a dlt_transform into a dlt.transformer but failed...
maybe I did something wrong, here is a gist with it: https://gist.github.com/djudjuu/428fb5567d3846d5692213ad8ff665a4

anuunchin:
1. Is the way the transformations are defined clear?
   - Yes, it is super clear.
2. Is it clear what lineage does?
   - I thought this was about `from where` a specific column stems, not which hints it inherits. Therefore, I was a bit confused.
3. Is it clear how SQL and Python transformations differ, also from a technical point of view?
   - Yes, it is clear as well.
4. Is it clear what the sync source does?
   - Yes.
5. Are there any obvious features missing? 
   - I don't really have a meaningful input on this.

### On homework:

I started doing the homework by defining two pipelines for secure and public data with dlt.yml. And then I was confused as to how datasets are defined, so I went to the source code and was able to do it. Then, for some reason the cache was created as duckdb database even though I seemed to have set the destination to filesystem - but I'm pretty sure it was because of some leftover metadata from previous tries. However, I wasn't 100% sure how to manually clean up everything and it just kept persisting. Then, I couldn't continue with the sync step because I realized the idea (presumably) was just to use python scripts, not necessarily the project yml combined with ci. After this realization, the homework seemed very clear.

> Might be a good idea to add very explicity examples for the whole dlt+ docs - but then again, I'm not used to yml projects.

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+    return dataset.employees.mutate(static_column=1234)
+
+# we can supply a column hint for the static_column, which will work for sql transformations
+@dlt_plus.transform(transformation_type="sql", lineage_mode="best_effort", columns={"static_column": {"type": "double"}})
+def mutated_transformation(dataset: SupportsReadableDataset):
+    return dataset.employees.mutate(static_column=1234)
+
+```
+
+## SQL vs Python transformations
```
akelad, line 366: Tbh I don't fully understand when you would use one over the other apart from speed. Maybe the entire tutorial should contain examples for when one makes sense over the other instead of listing reasons down here (I know there's an example above for an SQL transformation but its just a brief mention like "you could use that here").

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+* Are much faster than Python transformations and should be used if the use case allows for it
+* Only work if the full resulting schema is known ahead - this means lineage must pass in strict mode. Alternatively, you can supply column hints for columns that lineage cannot trace to the origin. This is the case, for example, if you add a static column in a .mutate() call on a ReadableIbisRelation
+* You cannot pipe the transformation into a dlt.transformer or do any kind of other data manipulation on dataitems, since the data never leaves the database
+
+
+## Sync Source
+As you have observed above, there are cases where you essentially would like to copy a table from one dataset to another but do not really need to build any SQL expressions, because you do not want to filter or join. Let's say you simply want to copy the full company dataset into the warehouse dataset without any kind of data cleanup - you can very easily do this with the sync source. Under the hood, the sync source is a regular dlt source with a list of `dlt.transform` objects that run this sync.
```
akelad, line 386: I think if we publish this as a public tutorial we should from the beginning describe the recommended methods for doing certain things. So use the `sync_source` function from the beginning of the tutorial

akelad (commented):
Full disclosure: I did not do the homework, only the tutorial 😄 

I left a lot of specific comments but some overall feedback:
* I think the fact that the entire example is moving (and modifying ofc) data from one duckdb dataset to another duckdb dataset makes it feel very much like a toy example. So when we publish this to the public maybe an example of actually moving it to a warehouse or filesystem or something would be cool
* I mentioned a few times that I'm not sure whether we're planning on releasing this tutorial to the general public. If we do, then I think we need to rewrite it a bit because there's a few confusing parts and sometimes too much unstructured text. It's not quite as easy to read as the other hackathons we've done. Ofc Solutions Eng is happy to help with that! 
* Naming of `@dlt_plus.transform` - I would consider making it `@dlt.transform` even if it's easily confusable with `@dlt.transformer`.

Also a general question - why would you do this instead of running transformations with dbt? Excuse my ignorance I am not a data engineer 🤠

rahuljo:
I also need to disclose that I've only done the tutorial and didn't have time to do the homework 😅 

What I liked:
1. I think the employees example is a very good choice for this tutorial, especially the star schema example and the lineage bit.
2. The sync source is really cool!  Very often I first load my data locally into duckdb and then to load it elsewhere I need to re-run the same code again, so the `sync_source` way of doing things would be pretty handy. Is there a reason why it's under `dlt_plus.transform` and not `dlt_plus.sources`?

What could be better:
1. I was able to follow the tutorial without problems, but I did find myself constantly asking what is the advantage of doing this over any of the other multiple ways to transform data in dlt (dbt, `sql_client`). Is it because using `@dlt_plus.transform` lets you do things like specify write dispositions and have your transformations run on cache? I would re-phrase the tutorial that doesn't just focus on how to do the transformation but also on the advantage of doing it this way.
2. +1 on Akela's point of adding output of the code snippets in the documentation to make it more readable. Especially in the star schema example, visualization for what the output looks like would make it a lot more understandable.

adrianbr (commented):
I was not able to run the tutorial, I tried on my local and on codespaces. error AttributeError: module 'dlt_plus' has no attribute 'transform'

I read the docs and the interface looks good but i can see nothing about inspecting lineage and how that looks

adrianbr (commented):
additional feedback after running 
- I also care about non-pii lineage for troubleshooting

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+You can now inspect your warehouse pipeline with the dlt Streamlit app by running `dlt pipeline warehouse_pipeline show`. You should see the cleaned_employees table in the warehouse dataset, which is missing the name column.
+
+What is happening here?
+
+* We have defined one new transformation function with the `dlt.transform` decorator. This function does not yield rows but returns a single object. This can be a SQL query as a string, but preferably it is a ReadableIbisRelation object that can be converted into a SQL query by dlt. You can use all ibis expressions detailed in the dlt oss docs. Note that you do not need to execute the query by running `.arrow()` or `.iter_df()` or something similar on it.
+
+* The resulting table name is derived from the function name but may be set with the `table_name` parameter, like the `dlt.resource`.
+
+* For now, we always need to set a transformation type. This can be `python` or `sql`. `sql` transformations will be executed directly as SQL on the dataset and are only available for transformations where the source_dataset and destination_dataset are the same or on the same physical destination. More on this below.
```
adrianbr, line 156: or on the same physical destination. ->  are the same or colocated and accessible from each other.

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+Python transformations
+* Can be run from any dataset to any destination dataset, you only need to ensure that the destination supports all column types you are using in the transformation or clean them up
+* Will run the select query on the source dataset but extract all resulting rows into the pipeline as Arrow tables / Parquet files and push them into the destination
+* Chunk size for Arrow table extract can be set with the `chunk_size` parameter on the dlt.transform decorator: `@dlt.transform(transformation_type="python", chunk_size=1000)`
+* When instantiated, behaves like a regular dlt.resource - you can pipe it into a dlt.transformer, add filters and maps with 'add_filter' and 'add_map' methods
+* Is generally much slower than pure SQL transformations
+* Will also work if not all resulting columns can be traced to origin columns (unless specified otherwise, more on lineage below)
+
+SQL transformations
+* Can only be run if the source and destination dataset are the same or on the same physical destination. Currently there is no proper check, so using SQL in a place where you should not will result in an obscure error
```
adrianbr, line 379: are the same or colocated and accessible from each other.

docs/website/docs/plus/getting-started/transformations.md
adrianbr: on syncs - how would I manage separate incremental strategies per table? maybe it's a good idea to show the alternative too

AstrakhantsevaAA (commented):
Super positive experience, everything work as expected (except sql transformation).
Tutorial is super clear. 
Would be great to know what features `dataset` has, what methods it has, I can get it from the code ofc, but from docs it's not clear. API reference or just enumerate main features.

SQL transformation doesn't work for me, but I'm pretty sure it worked a week ago. Idk how to fix it.
Left some comments.

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+- you're familiar with our read access based on [ibis expressions](../../general-usage/dataset-access/dataset#modifying-queries-with-ibis-expressions)
+
+
+## Introductory notes
+
+To work through this tutorial, we will be using multiple local DuckDB instances to transform and sync data between them. In the real world, you will be able to use any destination/dataset that supports ibis expressions. We will also create an example dataset with a couple of tables to work with to demonstrate how you can make use of transformations.
+
+## Our example dataset
+
+Copy the pipeline below and run it in your environment:
```
zilto, line 39: For first snippet, unfold the code and tell the user what file to put the code in. Make explicit how to run it (`python pipeline.py` or `uv run python pipeline.py`)

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+    dataset_name="company_dataset"
+)
+
+# Execute pipeline
+info = company_pipeline.run(employee_data())
+print(info)
+```
+</details>
+
+We have a straightforward company database with live employee master data containing names and payroll payments, as well as two tables for offices and departments. As always, you can inspect the contents of the dataset by running `dlt pipeline company_pipeline show` to open the dlt streamlit app.
```
zilto, line 113: Streamlit is not installed and not mentioned in the `Prerequisites` section

docs/website/docs/plus/getting-started/transformations.md
zilto, line 118: Do first, then give some light explanation, and use collapsible for details. In a tutorial, "how things work" is not the main concern. It's rather "get things done" or "how to do"

zilto: the transform on-backend vs in-memory is an important distinction. Should be at the beginning of its own paragraph instead of the middle

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+info = warehouse_pipeline.run(cleaned_employees(company_dataset))
+print(info)
+
+```
+
+You can now inspect your warehouse pipeline with the dlt Streamlit app by running `dlt pipeline warehouse_pipeline show`. You should see the cleaned_employees table in the warehouse dataset, which is missing the name column.
+
+What is happening here?
+
+* We have defined one new transformation function with the `dlt.transform` decorator. This function does not yield rows but returns a single object. This can be a SQL query as a string, but preferably it is a ReadableIbisRelation object that can be converted into a SQL query by dlt. You can use all ibis expressions detailed in the dlt oss docs. Note that you do not need to execute the query by running `.arrow()` or `.iter_df()` or something similar on it.
```
zilto, line 152: error in the text `@dlt.transform` should be `@dlt_plus.transform`

zilto: Remove use of negation "Note that you do not need" -> "you can use X if you want Y"

zilto: After the code snippet is a good place to highlight that `@dlt_plus.transform` is a custom `@dlt.resource` (instead of before the snippet)

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+Check the company pipeline to see the aggregated tables in the same dataset.
+
+:::tip
+For the row_counts transformation, we use the Python transformation type, as dlt+ currently cannot compute the needed column hints for this table. More on SQL transformations below. You can observe that you can easily mix Python and SQL transformations in the same source.
+:::
+
+
+## A star schema example
+
+A common use case is to create a star schema with a fact table and a few dimension tables where each important column is at most one join away from the fact table. Let's create a star schema for our company dataset with payroll payments as facts and make the office and department tables directly joinable from the fact table:
```
zilto, line 246: I don't know what a star schema and fact tables are. Instead of leading with terminology, explain "what & why" you're doing. Then, name the terms and give a useful link to the definition (wikipedia or a tech blog)

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+    return payroll_facts(dataset),employees(dataset), departments(dataset), offices(dataset)
+
+
+# create the star schema in the star warehouse
+info = star_pipeline.run(star_source(company_dataset))
+print(info)
+
+# now that we have a nice star schema, presumable in a column oriented warehouse, we can aggregate
+# payments per department with only a single join:
+star_dataset = star_pipeline.dataset()
```
zilto, line 298: This code snippt is confusing because you do Ibis code inside `@dlt_plus.transform` functions, but also outside of it. Mention explicitly "Doing a dlt+ transform allows to push data and create schema using dlt" and "in the main body, you can use the regular dlt dataset interface to manipulate loaded data (from a `dlt.resource` or `dlt.transform`).

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+# let's create a new star warehouse for simplicity
+star_pipeline = dlt.pipeline(
+    pipeline_name="star_pipeline",
+    destination="duckdb",
+    dataset_name="star_dataset"
+)
+
+# the dimension tables can just be copied over, again we clean the personal data from the employees table
+
+@dlt_plus.transform(transformation_type="python")
```
zilto, line 259: Explain why you need to wrap single table access in `@dlt_plus.transform`

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+```
+
+:::tip
+For simply copying table data, you can also use the sync source demonstrated below. Also note that in future versions of the dlt+ transformation framework, there will be a template generator to create a star schema from a given dataset to give you a starting point.
+:::
+
+
+## Lineage
+
+dlt+ transformations contain a lineage engine that can trace the origin of columns resulting from transformations. You may have noticed that we added a custom hint to the name column in the employees table at the beginning of the page. This hint is a custom hint that we decided to add to all columns containing very sensitive data. Ideally, we would like to know which columns in a result are derived from columns containing sensitive data. dlt+ lineage will do just that for you. Let's run an aggregated join query into our warehouse again, but this time we will not drop the name column:
```
zilto, line 315: Show first, then explain. Add a code snippet of the schema or a screenshot of something meaningful (Ibis has a built-in visualization feature)

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+### Untraceable columns
+
+In a couple of scenarios the current implementation is not able to trace the lineage of a column. In some cases you may opt to add a static value to an expression, which is not traceable by definitions. We are planning to allow to add exceptions to the lineage tracing for cases like these. Another case is where a column is the result of some operation on multiple source columns, such as concatenating the value of two separate columns or mathematical operations that use more than one source column. These cases are still being worked on and will be supported in future versions of dlt+.
```
zilto, line 338: I don't understand the issue with scalar / static values and multi-column operations. Both are supported by Ibis. Those can be included in lineage. If you have `col_foo`, `col_bar` and `col_baz = col_bar + 1`, you know `col_baz` depends on `col_bar` and scalar `1`

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+### Some examples from our company dataset
+
+```python
```
zilto, line 342: This would look better as "text -> snippet -> text -> snippet" rather than a long snippet + comments

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+# the following transformation will fail because the static_column is not traceable
+@dlt_plus.transform(transformation_type="python", lineage_mode="strict")
+def mutated_transformation(dataset: SupportsReadableDataset):
+    return dataset.employees.mutate(static_column=1234)
```
zilto, line 347: explain what `.mutate()` does (it adds a new column in this case)

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+# we can supply a column hint for the static_column, which will work for sql transformations
+@dlt_plus.transform(transformation_type="sql", lineage_mode="best_effort", columns={"static_column": {"type": "double"}})
+def mutated_transformation(dataset: SupportsReadableDataset):
+    return dataset.employees.mutate(static_column=1234)
+
+```
+
+## SQL vs Python transformations
+
+As you have seen above, you can choose to execute any transformation either as a Python or SQL transformation. For both transformation types, the basis is a decorated function that returns something that can be evaluated into a SQL query, either a string or a ReadableIbisRelation object. Both transformation types can be mixed and they both result in new tables in the destination dataset with a schema derived from the source dataset. The key differences are:
```
zilto, line 368: Be more opinionated: "you should use SQL / Ibis when you can because XYZ. If you want to write pandas / polars / arrow, you also". Switch ordering to have SQL first to reflect advice.

zilto, line 368: if `python` and `sql` actually mean `in-memory` vs `on-backend`, we should just say that. For example, Ibis *is* Python, but is on backend, like SQL.
    - Similarly, the Tip hint in `Create aggregated tables in the same dataset` says "you can mix SQL and Python". This statement is confusing because there's no SQL code `SELECT * FROM foo` in the snippet. It's all Python code.

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+### Create aggregated tables in the same dataset
+
+You can run a transformation against the same dataset that the source data comes from. This might be useful if you need some pre-computed stats (similar to a materialized view),
+or you need to do some cleanup after a load on the same dataset. Let's use the same transformation as above plus another one in a group. Since the source and destination datasets are the same, we can use SQL transformations, which are much faster.
+
+```python
+
+@dlt_plus.transform(transformation_type="sql", write_disposition="replace")
+def payment_per_employee_sql(dataset: SupportsReadableDataset):
```
zilto, line 217: I don't understand how you guarantee that `dataset.payroll` exists in this snippet.

This API seems generally hard to test because it takes the `dataset` state for granted. Wouldn't it be better to push tables "up" and write a idempotent and unit testable

```python
def payment_per_employee_sql(employees, payroll):
```
or add metadata to the decorator to do static validation before running the transform

With the decorator wrapping the function it can check table access and even validate some schema contracts / data validation before running `payment_per_employee`
```python
@dlt_plus.transform(..., deps=["employees", "payroll"])
def payment_per_employee(dataset):
   return ...
```

zilto: What's the return type of these functions? They return a `ReadableIbisRelation` object, but does the decorator wrap them into `dlt.Transform`?

I find dlt Source and Resource generally hard to unit test (harder than it should at least), but unit tests are even more important for transforms

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+@dlt_plus.transform(transformation_type="python")
+def copied_employees(dataset: SupportsReadableDataset):
+    return dataset.employees.drop("name")
+
+@dlt_plus.transform(transformation_type="python")
+def copied_payroll(dataset: SupportsReadableDataset):
+    return dataset.payroll
+
+# we can group transformnations into a regular source
+def copy_source(dataset: SupportsReadableDataset):
```
zilto, line 180: I understand that this function behaves like a `@dlt.source` that returns multiple `@dlt.resource`, but it's not obvious from the API given it doesn't have the distinctive `@dlt.source` (or any decorator)

docs/website/docs/plus/getting-started/transformations.md
zilto, line 386: Maybe `sync_dataset()` is a better name than `sync_source()`. You're not syncing a `@dlt.source` AFAIK, and the function receives a `Dataset` as input

zilto: Explain why you can't do this with native `dlt` (a lot of users seems to be chaining `sql_databse` source to other destinations).

docs/website/docs/plus/getting-started/transformations.md
```diff
@@ -0,0 +1,446 @@
...
+
+As you have seen above, you can choose to execute any transformation either as a Python or SQL transformation. For both transformation types, the basis is a decorated function that returns something that can be evaluated into a SQL query, either a string or a ReadableIbisRelation object. Both transformation types can be mixed and they both result in new tables in the destination dataset with a schema derived from the source dataset. The key differences are:
+
+Python transformations
+* Can be run from any dataset to any destination dataset, you only need to ensure that the destination supports all column types you are using in the transformation or clean them up
+* Will run the select query on the source dataset but extract all resulting rows into the pipeline as Arrow tables / Parquet files and push them into the destination
+* Chunk size for Arrow table extract can be set with the `chunk_size` parameter on the dlt.transform decorator: `@dlt.transform(transformation_type="python", chunk_size=1000)`
+* When instantiated, behaves like a regular dlt.resource - you can pipe it into a dlt.transformer, add filters and maps with 'add_filter' and 'add_map' methods
```
zilto, line 374: Users will ask "why should I do `@dlt_plus.transform` **after** loading data? why not do it in the initial pipeline?". We will need to highlight and clearly explain ELT vs ETL and **why**. Also, explain what's actually doable.

docs/website/docs/plus/getting-started/transformations.md
djudjuu: i agree. Worth being explicit here. Maybe have the same statement in different annotations mentioning in brackets where its executed.

sh-rp:
closing this hackathon, will keep the link to these comments in the transformations ticket for future improvement of the transformation docs!

PR #2404, opened 2025-03-13
"model" item_format support

sh-rp:
### Description

This PR introduces a new item-format "model" which is a sql select statement which will be inserted into a given table. For now only the duckdb destination supports this. Further destinations will be available after @anuunchin finished her research on this.

It is possible to mark an item with a HintsMeta to indicate the model item_type, which means the yielded string will be interpreted as a valid select statement for the targeted destination. During extraction each statement will be stored in its own job. 

A resource emitting a model query would look like this, given an input dataset. Columns need to be supplied so dlt can create / update the table:

```python
@dlt.resource()
def copied_table() -> Any:
    query = dataset["example_table"].limit(5).query()
    yield dlt.mark.with_hints(
        query, hints=make_hints(columns={...), data_item_format="model"
    )
```


dlt/common/data_writers/writers.py
```diff
@@ -175,6 +179,32 @@ def writer_spec(cls) -> FileWriterSpec:
...
+    @classmethod
+    def writer_spec(cls) -> FileWriterSpec:
+        return FileWriterSpec(
+            "model",
+            "model",
+            file_extension="model",
+            is_binary_format=False,
+            supports_schema_changes="True",
+            supports_compression=False,
+            # NOTE: we create a new model file for each sql row
```
sh-rp, line 203: I think it makes sense to have one model per file

dlt/destinations/impl/duckdb/factory.py
```diff
@@ -129,7 +129,7 @@ class duckdb(Destination[DuckDbClientConfiguration, "DuckDbClient"]):
     def _raw_capabilities(self) -> DestinationCapabilitiesContext:
         caps = DestinationCapabilitiesContext()
         caps.preferred_loader_file_format = "insert_values"
-        caps.supported_loader_file_formats = ["insert_values", "parquet", "jsonl"]
+        caps.supported_loader_file_formats = ["insert_values", "parquet", "jsonl", "model"]
```
sh-rp, line 132: For this first iteration we only support duckdb, the transformations can check the capabilities to figure out where pure sql may be used.

dlt/extract/hints.py
```diff
@@ -68,11 +69,17 @@ class TResourceHints(TResourceHintsBase, total=False):
 
 
 class HintsMeta:
-    __slots__ = ("hints", "create_table_variant")
+    __slots__ = ("hints", "create_table_variant", "data_item_format")
```
sh-rp, line 72: I opted to add a slot on the default hints, I could also create a subclass similar to the file import, but for me this solution somehow makes more sense, since it is kind of a hint, but does not need to go into the schema.

dlt/extract/utils.py
```diff
@@ -60,14 +62,21 @@
     pandas = None
 
 
-def get_data_item_format(items: TDataItems) -> TDataItemFormat:
+def get_data_item_format(items: TDataItems, meta: Any = None) -> TDataItemFormat:
```
sh-rp, line 65: Each row that has a special item type needs to have the meta set, similar to how the ImportFileMeta works. We could also store it so it works more like the other hints, but I think this should be ok. It's mostly going to be used from the transformations anyway.

dlt/extract/hints.py
rudolfix: I think you do not need to modify `HintsMeta`. Just create a container class in which you'll yield models ie
```py
class ModelStr(str):
   pass
```
and yield it. Are you 100% sure we do not need any model properties to be stored in the file? or we plan to pass those via TableSchema? ie. look here:
https://docs.getdbt.com/reference/model-configs (materialization)
https://docs.getdbt.com/docs/build/incremental-strategy (this is our write disposition + primary key) heh cool

dlt/extract/resource.py
```diff
@@ -81,7 +86,7 @@ def with_hints(
     Create `TResourceHints` with `make_hints`.
     Setting `table_name` will dispatch the `item` to a specified table, like `with_table_name`
     """
-    return DataItemWithMeta(HintsMeta(hints, create_table_variant), item)
+    return DataItemWithMeta(HintsMeta(hints or {}, create_table_variant, data_item_format), item)
```
rudolfix, line 89: why default {} value?

dlt/extract/utils.py
rudolfix: as mentioned, just check the item type

dlt/destinations/job_client_impl.py
```diff
@@ -279,6 +311,9 @@ def create_load_job(
         if SqlLoadJob.is_sql_job(file_path):
             # create sql load job
             return SqlLoadJob(file_path)
+        if ModelLoadJob.is_model_job(file_path):
```
rudolfix, line 314: hmmmm maybe we should register jobs via plugin as well?

rudolfix (commented):
looks cool! but IMO can be simplified

dlt/destinations/job_client_impl.py
```diff
@@ -113,6 +113,38 @@ def is_sql_job(file_path: str) -> bool:
...
+
+    def _insert_statement_from_select_statement(self, select_statement: str) -> str:
+        """
+        NOTE: Here we generate an insert statement from a select statement, this is the duckdb
+        dialect, we may be able to transpile with sqlglot for each destination or we need
+        to subclass and override this method.
+        """
+        sql_client = self._job_client.sql_client
+        name = sql_client.make_qualified_table_name(self._load_table["name"])
+        return f"INSERT INTO {name} {select_statement};"
```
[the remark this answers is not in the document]
sh-rp: Right now these inserts are done on the dataset that the jobs are initially loaded to, so if you have a sql transformation with write_disposition merge, it will insert the rows into the staging table and then create a regular merge jobs, so another sql job, that gets executed at the end. At least I think it should do this, I still have to write tests. I'd like to keep this PR as simple as possible for now and then migrate to more efficient jobs later.

sh-rp:
"Are you 100% sure we do not need any model properties to be stored in the file? or we plan to pass those via TableSchema?"
I'm not. Conceptually it would be cool if the model files only had the instructions of where to take the data from and the rest is in the schema, so it is like all the other job files. I was planning to add views in the transformations, I would've added a new table format for this here, but in your review you said we don't need them, so now this is not here. For merging and upserting we already have the right hints which are in the schema, so I don't think anything else is required. But I'm not 100% sure at this point :)

dlt/destinations/job_client_impl.py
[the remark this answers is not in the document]
rudolfix: heh right! but I see a way to skip the staging dataset...

dlt/common/storages/data_item_storage.py
```diff
@@ -27,8 +27,11 @@ def _get_writer(
         writer = self.buffered_writers.get(writer_id, None)
         if not writer:
             # assign a writer for each table
+            kwargs = {}
+            if self.writer_spec.file_max_items:
+                kwargs["file_max_items"] = self.writer_spec.file_max_items
             path = self._get_data_item_path_template(load_id, schema_name, table_name)
-            writer = BufferedDataWriter(self.writer_spec, path)
+            writer = BufferedDataWriter(self.writer_spec, path, **kwargs)
```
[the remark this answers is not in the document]
rudolfix: OK I get it now, but I think you could just take `file_max_items` from `writer_spec` in `BufferedDataWriter` . you do not need to pass any kwargs.

dlt/normalize/worker.py
```diff
@@ -111,7 +111,7 @@ def _get_items_normalizer(
             if item_format == "file":
```
rudolfix, line 111: I think `item_format_from_file_extension` should return `model` item format, not `file` for models. here this is not relevant but IMO may be in other places

sh-rp: It does already doesn't it? Or am I misunderstanding your comment.

dlt/common/storages/data_item_storage.py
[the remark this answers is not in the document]
sh-rp: The problem is, that if I set an explicit "None" as an argument, it will not take this value from the `BufferedDataWriterConfiguration`. It seems the config values in constructors are only set for arguments that do not have an explicit value set during instantiation, which makes sense to me. This also is true for a 'None' value, so I need to call the BufferedDataWriter.__init__ either with this argument or without it, but never with `None`.

dlt/destinations/job_client_impl.py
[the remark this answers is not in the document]
sh-rp: Yes of course, but imho out of scope for now, because I am actually trying to build the transformations in time. unless you insist, then I'll do it :)

tests/load/pipeline/test_model_item_format.py
```diff
@@ -0,0 +1,170 @@
+# test the sql insert job loader, works only on duckdb for now
```
sh-rp, line 1: I am not quite sure what else to test here, basically we are just testing wether these sql statements are created properly, all the column types etc should work since we are on the same destination and if they don't, there is not much we can do about it really, since we are not touching the data.

dlt/extract/hints.py
sh-rp: I have done it this way now.

dlt/extract/resource.py
sh-rp: i removed it, it was a leftover

dlt/common/data_writers/writers.py
anuunchin: @rudolfix  I don't quite understand what this convo is about 👀. If rephrased, is the the same as:

```
@rudolfix: One can yield many models (many insert statements) per table in a single run, so it's wrong to have one model (insert statement) per table
@sh-rp: No, it does make sense to have one model (insert statement) per table
```

dlt/common/typing.py
```diff
@@ -127,7 +127,9 @@ class SecretSentinel:
 VARIANT_FIELD_FORMAT = "v_%s"
 TFileOrPath = Union[str, PathLike, IO[Any]]
 TSortOrder = Literal["asc", "desc"]
-TLoaderFileFormat = Literal["jsonl", "typed-jsonl", "insert_values", "parquet", "csv", "reference"]
+TLoaderFileFormat = Literal[
+    "jsonl", "typed-jsonl", "insert_values", "parquet", "csv", "reference", "sql", "model"
+]
```
anuunchin, line 132: @sh-rp  why do we also add `"sql"` here? 🤔

PR #2419, opened 2025-03-18
feat(DLT-PLUS-105): factor table format filesystem destinations to less bound modules

jfahne:
### Description

Ahead of changes to allow more support for Apache Iceberg catalogs, refactor the table format filesystem destinations (Apache Iceberg and LFAI Delta currently) to be less coupled with bare (having no format specification) filesystem implementation details. The behavior of the filesystem destination has not been modified in anyway.

### Additional Context

This change is in part to demonstrate further simplifications that could be made throughout the project by preferring structural sub typing as described in [PEP 544](https://peps.python.org/pep-0544/) to compose more generic module level definitions. Today, the project makes use of multiple inheritance with a large amount of the parent class overridden by child implementations and interstitial layers of mixin classes which bind implementations excessively making changes difficult.

The example of note in this change is the elimination of the `TableFormatFilesystemLoadJob` which inherits from `FilesystemLoadJob` only to override almost all of the implementation of `FilesystemLoadJob` which itself inherits from `RunnableLoadJob`. This multiple inheritance makes it difficult to understand what the few stubs attached to `IcebergFilesystemLoadJob` do in context and difficult to identify the right layer at which to extend implementation in support of new features. Instead this change sacrifices a small amount of boilerplate added by eliminating the `TableFormatFilesystemLoadJob` for the gain of far fewer nested definitions, protocol definitions which describe the role that a load job or filesystem client instance plays in, for example, the construction of a remote path, and makes clear where to extend a class's implementation to enable new features.

After this change has been merged, it will be followed by refactoring specific to the Iceberg load job in furtherance of the dlt+ ticket for broader Iceberg catalog support.

### DIFF
--- a/dlt/destinations/impl/filesystem/table_format/common.py
+++ b/dlt/destinations/impl/filesystem/table_format/common.py
@@ -1,0 +1,116 @@
+import os
+from typing import Any, Final, Protocol, TYPE_CHECKING, Union
+
+from dlt.common.destination.client import JobClientBase
+from dlt.common.destination.typing import PreparedTableSchema
+from dlt.common.schema.utils import get_columns_names_with_prop
+
+if TYPE_CHECKING:
+    from dlt.common.libs.pyarrow import pyarrow as pa
+
+DLT_PYARROW_MODULE_NAME: Final = "pyarrow"
+PREPARED_TABLE_SCHEMA_PROPERTY_PARTITION: Final = "partition"
+LOAD_TABLE_WRITE_DISPOSITION_KEY: Final = "write_disposition"
+
+
+class HasSourceFilePaths(Protocol):
+    """Objects which have source data stored across one or more files."""
+
+    file_paths: list[str]
+
+
+class HasNamedLoadTable(Protocol):
+    """Objects which have a table name and schema into which data is loaded."""
+
+    _load_table: PreparedTableSchema
+
+    @property
+    def load_table_name(self) -> str: ...
+
+
+class HasTableDirectoryGetter(Protocol):
+    """Objects which allow retrieving the directory of a table stored in a filesystem."""
+
+    def get_table_dir(self, table_name: str) -> str:
+        """Retrieves the directory in a filesystem for a table object by name.
+        Args:
+            table_name: name of a table in a filesystem
+        Returns:
+            directory of table in filesystem as a string
+        """
+        raise NotImplementedError
+
+
+class HasRemotePathBuilder(Protocol):
+    """Objects which build a path in reference to a remote filesystem."""
+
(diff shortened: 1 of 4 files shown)


dlt/destinations/impl/filesystem/table_format/iceberg.py
```diff
@@ -0,0 +1,78 @@
...
+        write_iceberg_table(
+            table=self._iceberg_table(),
+            data=self.arrow_dataset.to_table(),
+            write_disposition=self._load_table[LOAD_TABLE_WRITE_DISPOSITION_KEY],
+        )
+
+    def _iceberg_table(self) -> "pyiceberg.table.Table":  # type: ignore[name-defined] # noqa: F821
+        from dlt.common.libs.pyiceberg import get_catalog
+
+        catalog = get_catalog(
```
rudolfix, line 64: our goal is to be able to plug any catalog here, not just use ephemeral one. this is actually hardcoded.

so the idea was to add a method:
```py
def get_catalog():
   from dlt.common.libs.pyiceberg import get_catalog

    return get_catalog(
            client=self._job_client,
            table_name=self.load_table_name,
            schema=self.arrow_dataset.schema,
            partition_columns=self._partition_columns,
        )
```
that you can override in dlt+ to plug any other catalog

dlt/destinations/impl/filesystem/table_format/common.py
```diff
@@ -0,0 +1,116 @@
...
+
+if TYPE_CHECKING:
+    from dlt.common.libs.pyarrow import pyarrow as pa
+
+DLT_PYARROW_MODULE_NAME: Final = "pyarrow"
+PREPARED_TABLE_SCHEMA_PROPERTY_PARTITION: Final = "partition"
+LOAD_TABLE_WRITE_DISPOSITION_KEY: Final = "write_disposition"
+
+
+class HasSourceFilePaths(Protocol):
```
rudolfix, line 16: I'm not sure I understand the simplification here. It generates a lot of duplicated code in concrete classes. This is probably a good abstraction but why not place this in a base class? I know you argued against it but here I have 5 protocols instead of single base class where all methods works together.
I'm saying this is wrong, I just need some clarifications

jfahne: The format specific classes only depend on `RunnableJob` and the definitions in the module's `common.py` file. Because Delta, Iceberg, and other table formats like Hudi are more dissimilar beyond their shared use of filesystems as an abstraction for a storage layer, I would anticipate that their implementations deviate more as features are added to dlt for them and the maintenance of a common parent that is less general than `RunnableJob` becomes cumbersome.

Issue #2399, opened 2025-03-12
Ibis insert statement research


sh-rp:
### TLDR
For the transformations we want to support a new loader_file_format called `model`. The loader files are the files that contain the data that will be loaded to the destination, the current ones we already have are `parquet` or `insert_values` which are a parquet file with a table / a list of sql insert statements containing values respectively. This new `model` format will contain an sql select clause as a string which we use as the source data for the table to be materialized. We need to know which of our sql destinations support these kind of inserts and which don't and which ones have specific quirks.

### What to do
Please find out if we can use our `ReadableIbisRelation` to create these statements. I think ibis supports this, but I am not sure how to do it and if it actually works on all of our destinations that support the `ReadableIbisRelation`. 

### Example
One example for duckdb. Copying one table fully to another would have the following syntax in duckdb. This will work if both tables exist and have the same schema, you can take that as a given for now.

`INSERT INTO dest_table_name SELECT * FROM source_table_name`

To get the select part from a given dlt dataset with populated tables you do:

`pipeline.dataset().source_table_name `

Which gives you an IbisRelation (if ibis is installed) that contains a select query for the whole table. You can inspect this query with:

`print(pipeline.dataset().source_table_name.query())`

Can you find an ibis command that will create an insert statement using this query and can you test it out on every destination to see wether it works? If you have built an ibis expression that has this insert, you can just execute it on the sql_client of the dataset to see what happens and if the result is alright:

```
d = pipeline.dataset()
select_relation = d.source_table_name
target_relation = d.target_table_name

# some magic happens here that produces:
insert_relation = 


with dataset.sql_client() as client:
   client.execute(insert_relation.query())
```

zilto:
The `ReadableDBAPIDataset.ibis()` returns an `IbisBackend`. They should all support the `IbisBackend.insert()` method ([duckdb ref](https://ibis-project.org/backends/duckdb#ibis.backends.duckdb.Backend.insert), [bigquery](https://ibis-project.org/backends/bigquery#ibis.backends.bigquery.Backend.insert))

sh-rp:
@zilto I want to generate an sql insert statement and not use the ibis backend.

zilto:
I figured that we wanted to use our own SQL mechanisms. 

[...]

PR #2420, opened 2025-03-18
Testing sqlglot's insert

anuunchin:
### Description

> Please check out #2399 where the motivation behind this PR is explained.

Here, we're just testing if an INSERT INTO statement created with the following steps works for different destinations:
- Assume two tables have the same schema
- Create a `SELECT * FROM` statement using an IbisRelation with something like:
     ```python
     dataset = pipeline.dataset()
     select_relation_query = dataset.table("select_from_table").query()  
     ```
- Make the `target_table` with:
     ```python
     target_table = client.make_qualified_table_name("insert_into_table")
     ```
- Make sqlglot create an `INSERT INTO ... SELECT * FROM ...` statement with something like:
     ```python
     import sqlglot.expressions as sge
     query = sge.insert(
        expression=select_relation_query,
        into=target_table  
        dialect=dialect,
    ).sql()
     ```
   > [The source code of insert](https://github.com/tobymao/sqlglot/blob/1d6218e386b3b1a5081272e179b8e48ec57153b4/sqlglot/expressions.py#L7494). Thanks to @zilto for pointing into this direction 🙏

### Related Issues

- Relates to #2399 

### Additional Context

Some destination don't really recognize double quotes ("") for identifiers, so there's a special handling in the test.

### DIFF
--- a/tests/load/test_insert_copy.py
+++ b/tests/load/test_insert_copy.py
@@ -9,20 +9,20 @@ from tests.load.utils import (
 
 
 DESTINATIONS: List[str] = [
-    #    "duckdb",
-    #    "athena",
-    #    "bigquery",
-    #    "clickhouse",
-    #    "databricks",
-    #    "motherduck",
-    #    "redshift",
-    #    "snowflake",
-    #    "sqlalchemy",
-    # NOTE: the following haven't been tested yet
-    #    "postgres"
-    #    "mssql",
-    #    "filesystem",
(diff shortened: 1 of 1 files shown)


anuunchin:
Filesystem is being skipped as of now

sh-rp:
@anuunchin really cool work, closing this as it is continued in https://github.com/dlt-hub/dlt/pull/2423

PR #2423, opened 2025-03-19
"model" item_format support takeover

anuunchin:
### Disclaimer:

This is a takeover PR of [this PR](https://github.com/dlt-hub/dlt/pull/2404).

### Description

This PR introduces a new item-format "model" which is a sql select statement which will be inserted into a given table. For now, all default sql destinations support it, except for Athena and Dremio. 

It is possible to mark an item with a HintsMeta to indicate the model item_type, which means the yielded string will be interpreted as a valid select statement for the targeted destination. During extraction each statement will be stored in its own job. 

A resource emitting a model query would look like this, given an input dataset. Columns need to be supplied so dlt can create / update the table:

```python
@dlt.resource()
def copied_table() -> Any:
    query = dataset["example_table"].limit(5).query()
    yield dlt.mark.with_hints(
        query, hints=make_hints(columns={...), data_item_format="model"
    )
```
Related to #2420, #2399.


tests/load/pipeline/test_model_item_format.py
```diff
@@ -0,0 +1,244 @@
...
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(
+        default_sql_configs=True,
+        exclude=["athena"],
+    ),
+    ids=lambda x: x.name,
+)
+def test_multiple_statements_per_resource(destination_config: DestinationTestConfiguration) -> None:
```
anuunchin, line 214: @sh-rp  This test fails for postgres because it's strict about enforcing unique constraints:

```
dlt.destinations.exceptions.DatabaseTerminalException: duplicate key value violates unique constraint "copied_table__dlt_id_key"
DETAIL:  Key (_dlt_id)=(L4WTgeyuBgTi/A) already exists.
```
On that note, I'm not sure what the purpose of this test is 👀, considering the comment implies that we're yielding two models to create 2 new tables - even though only one new table is created, namely the `copied_table`

rudolfix (changes requested):
This is really good! but there are a few conceptual problems we must address:
- in the simples case the model is just a string. and then we use destination dialect to parse it. this is fine
- but there will be cases when different dialect is used to write transformations and different to execute them. ie, stuff is created with duckdb and executed on snowflake. in that case we lose that information. 

My take: convert 
- ModelStr into named tuple (ie. SqlModel) (dialect, model) where dialect is optional
- allow to create it form sqlglot expression that will populate both elements
- file writer should store both dialect and model. my take:
first line: **dialect: name** is optional
next lines: actual model
this way we can store models with and without dialect.

file writer should do basic sanity checks:
- parse model using dialect
- if dialect not found, use destination dialect (you should have caps there)
- if no caps - ansi dialect is used
- after parsing you should have a list of columns which you must compare with a list of columns in dlt schema. raise on any mismatch. transformation layer must generate correct data here (with that, databricks problem should be detected early)

in model job: parse model file and use correct dialect to load it, then if different dialect was used - transpile it to destination dialect.

now a few other things:
- we should be able to enable Athena! the dialect that we need is Trino (models are queries, then we do DML - for that Trino is used)
- Athena actually has more engines. Hive is used for DDL. for that we'd need a separate setting in caps. just ignore it for now

sh-rp:
I am not convinced that anyone would be creating these statements on one database type and send them to another. You are building these sql strings with ibis expressions on a dlt dataset and this dataset will be the one you load to in the end, so the dialects will match. If we really want to keep the option to load any model file on any destination, I would suggest making it a requirement that the model files be in the sqlglot generic dialect, this way the ModelJob, which knows where it is being executed, may transpile it on the fly.

anuunchin:
@sh-rp @rudolfix  tried to implement the feedback except for the column checking part. I'm sure this is not there yet - I have the following questions:
- Is it going in the right direction?
- In sqlglot, there's no ansi dialect. If no dialect is passed, sqlglot uses its default dialect which is supposed to be as general as possible
- I'm still somewhat confused about `allow to create it form sqlglot expression that will populate both elements`. Therefore, the attempt to incorporate this feedback is probably wrong.
- I'm also inclined towards Dave's point that `not convinced that anyone would be creating these statements on one database type and send them to another. `

dlt/common/data_writers/writers.py
```diff
@@ -175,6 +180,36 @@ def writer_spec(cls) -> FileWriterSpec:
...
+
+    def write_data(self, items: Sequence[TDataItem]) -> None:
+        super().write_data(items)
+        self.items_count += len(items)
+        for item in items:
+            dialect = item.dialect or (self._caps.sqlglot_dialect)
+            query = item.query
+            parsed_query = sqlglot.parse_one(query, read=dialect)
+            normalized_query = parsed_query.sql(dialect=dialect)
+            self._f.write("dialect: " + (dialect or "") + "\n" + normalized_query + "\n")
```
rudolfix, line 197: this is good. I'll try to convince you that this makes sense

dlt/destinations/job_client_impl.py
```diff
@@ -113,6 +119,99 @@ def is_sql_job(file_path: str) -> bool:
...
+    A job to insert rows into a table from a model file which contains a single select statement
+    """
+
+    def __init__(self, file_path: str) -> None:
+        super().__init__(file_path)
+        self._job_client: "SqlJobClientBase" = None
+
+    def run(self) -> None:
+        with FileStorage.open_zipsafe_ro(self._file_path, "r", encoding="utf-8") as f:
+            select_dialect = f.readline().split(": ")[1].strip() or None
```
rudolfix, line 133: better to split on ":" without space. we should always assume that people with mess around with human readable files so they may ie. write their own

rudolfix: do you want to prevent getting dialect as "" in here? IMO if dialect is empty you should use dialect from caps

dlt/destinations/job_client_impl.py
```diff
@@ -113,6 +119,99 @@ def is_sql_job(file_path: str) -> bool:
...
+
+        normalized_select_query = parsed_select_query.sql(dialect=select_dialect)
+
+        destination_dialect = self._job_client.capabilities.sqlglot_dialect
+        transpiled_select_query = sqlglot.transpile(
+            normalized_select_query, read=select_dialect, write=destination_dialect
+        )[0]
+
+        # Apply case folding if needed
+        if self._job_client.capabilities.has_case_sensitive_identifiers:
```
rudolfix, line 192: a clear proof that true query normalization is needed in the normalization is needed. you can keep it for now but it will be replaced. or if you want to make future work easier try to fix it as follows
* you should **never** convert any identifiers ie to lower case in the original SELECT, you must respect naming convention that was originally used. it applies to input dataset. here we convert names to output dataset
* instead use the parsed query to find the top level SELECT and find column names in it. then apply aliases to them (SELECT column AS alias) that are properly case-folded. (via. self._sql_client.capabilities.casefold_identifier). also see what InsertValuesLoadJob does with INSERT column names
* note: you assume that proper normalization happened earlier. you just case-fold identifiers here.

dlt/destinations/job_client_impl.py
```diff
@@ -113,6 +119,99 @@ def is_sql_job(file_path: str) -> bool:
...
+
+        # Parse SELECT
+        parsed_select_query = sqlglot.parse_one(select_statement, read=select_dialect)
+
+        # Adjust table parts (catalog/db/this) based on dialect and catalog presence
+        for table in parsed_select_query.find_all(sqlglot.exp.Table):
+            parts = list(table.parts)
+            if target_catalog:
+                if len(parts) == 3:
+                    table.set("catalog", sqlglot.exp.to_identifier(target_catalog))
```
rudolfix, line 171: IMO if someone set the catalog we must keep it. it could be some kind of federated database

dlt/extract/extract.py
```diff
@@ -343,6 +343,9 @@ def _extract_single_source(
             "arrow": ArrowExtractor(
                 load_id, self.extract_storage.item_storages["arrow"], schema, collector=collector
             ),
+            "model": TextExtractor(
```
rudolfix, line 346: this will be `ModelExtractor` and it will contain logic to normalize and rewrite models (ie. to add `_dlt_load_id`. you can already rename that

tests/load/pipeline/test_model_item_format.py
```diff
@@ -0,0 +1,338 @@
...
+    other_dataset = other_pipeline.dataset()
+    other_dialect = other_pipeline.destination.capabilities().sqlglot_dialect
+
+    example_table_columns = other_dataset.schema.tables["example_table"]["columns"]
+
+    # for now, we need to make sure the same data exists in the destination
+    pipeline = destination_config.setup_pipeline(
+        pipeline_name="pipeline", dataset_name="test_model_item_format", dev_mode=False
+    )
+    pipeline.run(
```
rudolfix, line 77: please set up a pipeline with `dlt.sources.incremental`. IMO we are missing incremental `transform` for the `Model`. (it is the same mechanism we used to add extractor and storage above)

rudolfix (dismissed):
conceptually we are good to go. but IMO the model rewrite to convert this into INSERT must be changed.
I'll do followup ticket so we can handle normalization properly

dlt/destinations/job_client_impl.py
anuunchin: @rudolfix  but wouldn't that be odd in case the select statement comes from a different database and we need to transpile it in 🤔
```python
        transpiled_select_query = sqlglot.transpile(
            normalized_select_query, read=select_dialect, write=destination_dialect
        )[0]
```

sh-rp:
@anuunchin: I have one additional request. I think you are currently creating the following type of insert statements:

`INSERT INTO T1 SELECT (C1, C2) FROM T2`

This will insert the columns from the select statement in the order they come, but this causes problems in certain cases when creating new columns on ibis. It should work if the queries are like this:

```
INSERT INTO T1 (C1, c2)
SELECT C1, C2 FROM T2
```

Here you specify the target columns names also. It might be a bit tricky to implement but you can see if it maybe is easy to do.

anuunchin:
@rudolfix @sh-rp 
The following are added:
- column aliasing with casefolded names (which also addressed what @sh-rp  mentioned about columns in the previous comment)
- some minor things that were requested:
  - unnecessary parsing, reparsing removed
  - destination dialect is used to parse the query if no select dialect is provided
  - incremental fails, so there's a test for the future 👀

dlt/common/typing.py
```diff
@@ -129,7 +129,9 @@ class SecretSentinel:
 VARIANT_FIELD_FORMAT = "v_%s"
 TFileOrPath = Union[str, PathLike, IO[Any]]
 TSortOrder = Literal["asc", "desc"]
-TLoaderFileFormat = Literal["jsonl", "typed-jsonl", "insert_values", "parquet", "csv", "reference"]
+TLoaderFileFormat = Literal[
```
sh-rp, line 132: I think "sql" can be removed from this list again, this is a leftover from my very first version.

dlt/common/data_writers/writers.py
```diff
@@ -175,6 +180,36 @@ def writer_spec(cls) -> FileWriterSpec:
...
+    """Writes incoming items row by row into a text file and ensures a trailing ;"""
+
+    def write_header(self, columns_schema: TTableSchemaColumns) -> None:
+        pass
+
+    def write_data(self, items: Sequence[TDataItem]) -> None:
+        super().write_data(items)
+        self.items_count += len(items)
+        for item in items:
+            dialect = item.dialect or (self._caps.sqlglot_dialect)
```
sh-rp, line 193: I am not sure self._caps will be set if you just run the extract on a pipeline without having a destination set, which is possible. Maybe better check if _caps is set in this statement.

dlt/extract/hints.py
```diff
@@ -84,13 +88,32 @@ class TResourceHints(TResourceHintsBase, total=False):
 
 
 class HintsMeta:
-    __slots__ = ("hints", "create_table_variant")
+    __slots__ = ("hints", "create_table_variant", "data_item_format")
```
sh-rp, line 91: the data_item_format slot is also a leftover I think and can be removed.

dlt/extract/hints.py
```diff
@@ -84,13 +88,32 @@ class TResourceHints(TResourceHintsBase, total=False):
...
         self.hints = hints
         self.create_table_variant = create_table_variant
 
 
+class SqlModel(NamedTuple):
+    query: str
+    dialect: Optional[str] = None
+
+    @classmethod
+    def from_sqlglot(cls, query: str, dialect: Optional[str] = None) -> "SqlModel":
```
sh-rp, line 107: The correct name would be `from_query_string`, no? This is independent of sqlglot.

tests/load/pipeline/test_model_item_format.py
sh-rp: @anuunchin I think you can get around this by just stripping out all unique hints when forwarding the table schema to be created, no?

tests/load/pipeline/test_model_item_format.py
```diff
@@ -0,0 +1,313 @@
...
+        [{"a": 1, "b": "n", "c": True}, {"a": None, "b": "n", "c": False}],
+        table_name="example_table",
+    )
+    dataset = pipeline.dataset()
+
+    example_table_columns = dataset.schema.tables["example_table"]["columns"]
+
+    select_dialect = pipeline.destination.capabilities().sqlglot_dialect
+
+    # This should raise a column number mismatch error
```
sh-rp, line 235: If you are testing specific errors, it might make sense to inspect which exception exactly is raised and nested within PipelineStepFailed, so we are not shadowing this with some other problem.

tests/load/pipeline/test_model_item_format.py
```diff
@@ -0,0 +1,313 @@
...
+        )
+
+    with pytest.raises(PipelineStepFailed):
+        pipeline.run([partial_insert()])
+
+    # This should raise a type mismatch.
+    # SQLite (via SQLAlchemy) is lenient with types, so instead of a type error,
+    # we get a NOT NULL constraint violation when inserting None from "a" into "_dlt_id".
+    @dlt.resource()
+    def reversed_insert() -> Any:
```
sh-rp, line 253: if I understand this test correctly, you are testing what happens if the select statement returns the column names in a different order than they are in the schema. I thought that my requested change that you also implemented would make this work, so that an insert from select will always work if all columns are present regardless of the order, but maybe I am mistaken here and we can't do it.

tests/load/pipeline/test_model_item_format.py
sh-rp: PS: I think alternatively you can set create_indexes on the postgres config to false, then these constraints will also not be created.

tests/load/pipeline/test_model_item_format.py
sh-rp: @rudolfix what exactly is the task here? Incremental will not work with model jobs, should we have a good error or why should there be an incremental test?

sh-rp (changes requested):
A really cool PR! I have a few small cleanup comments, mostly stuff leftover from my work. And I have this additional request to see if it is possible to make queries work where the columns in the select statement are in a different order than in the table, your one test seems to suggest that this does not work at the moment. Maybe we can't do it, then we can keep it as is.

dlt/common/data_writers/writers.py
anuunchin: @sh-rp  oww, i didn't know that was possible. It's indeed erroring when no destination is set. I added a small test. Also noticed that
```python
        super().write_data(items)
        self.items_count += len(items)
```
the second line is unncesseary because the super write_data increments count already, so i removed it

tests/load/pipeline/test_model_item_format.py
anuunchin: @sh-rp  I removed this test and adjusted the `test_simple_model_jobs` one because similar logic was already partially there

anuunchin:
@sh-rp Thank you for the review! I incorporated the feedback. In a nutshell:
- Explicitly aliased columns are handled (There's a test as well)
- Case with no destination is handled (also with a test)
- Unnecessary test with partial/reversed insert is incorporated into a simpler test that had overlapping logic anyway 
👀

PR #2426, opened 2025-03-20
Added `dlt init --list-destinations`

zilto:
The command `dlt init <source>  <destination>` has a `--list-sources` that pulls from `dlt-hub/verified-sources`, but didn't have a `--list-destinations` flag. This addition improves the CLI experience and helps agents properly use this command.

The values for `--list-destinations` are taken from `dlt.destinations.__all__` which aims to match the logic used by the `dlt init` command.

### DIFF
--- a/docs/website/docs/reference/command-line-interface.md
+++ b/docs/website/docs/reference/command-line-interface.md
@@ -491,8 +491,8 @@ Creates a pipeline project in the current folder by adding existing verified sou
 
 **Usage**
 ```sh
-dlt init [-h] [--list-sources] [--location LOCATION] [--branch BRANCH] [--eject]
-    [source] [destination]
+dlt init [-h] [--list-sources] [--list-destinations] [--location LOCATION]
+    [--branch BRANCH] [--eject] [source] [destination]
 ```
 
 **Description**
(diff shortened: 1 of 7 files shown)


docs/website/docs/reference/command-line-interface.md
```diff
@@ -522,6 +522,7 @@ Inherits arguments from [`dlt`](#dlt).
 **Options**
 * `-h, --help` - Show this help message and exit
 * `--list-sources, -l` - Shows all available verified sources and their short descriptions. for each source, it checks if your local `dlt` version requires an update and prints the relevant warning.
+* `--list-destinations` - Shows the name of all core dlt destinations.
```
AstrakhantsevaAA, line 525: I was just passing by. But I have one ask, can we also have a short version of this command, something like `-ld`

sh-rp: One note from my side: I dlt+, i have `dlt destination list-available` and `dlt source list-available`. We could consider moving to these commands in the oss version too.

zilto: > One note from my side: I dlt+, i have `dlt destination list-available` and `dlt source list-available`. We could consider moving to these commands in the oss version too.

I think the command format `dlt source --flag` has better semantic than `dlt init --command`. Though, I'm trying to have this simple fix in to unblock AI / MCP features in a realistic timeframe

zilto:
Closing in favor of #2441

PR #2427, opened 2025-03-21
Optional engine_adapter_callback

michelzurkirchen:
### Description

The parameter `engine_adapter_callback` on the source `sql_database` wasn't optional, which caused an issue with type checking, as the parameter isn't actually required. This PR makes the parameter optional.

### Related Issues

This is a partial fix for [issue 2415](https://github.com/dlt-hub/dlt/issues/). It adresses the initial finding regarding `engine_adapter_callback`, it doesn't address the more general issue of the mypy setting `strict_optional=false`.


sh-rp:
@michelzurkirchen thanks :) merging this.

PR #2435, opened 2025-03-24
Fix: insert-from-staging replace strategy incorrectly using staging-optimized pattern

### LINKED ISSUE #2422: bigquery insert-from-staging replace strategy doesn't work
### dlt version

1.6.1

### Describe the problem

Hey, I'm loading data into bigquery and I want to use the
https://dlthub.com/docs/general-usage/full-loading#the-insert-from-staging-strategy
insert from staging strategy.
But when looking at the query that gets run on bigquery it runs 
```
DROP TABLE IF EXISTS `<proj_id>`.`<dataset>`.`<destination_table>`;

CREATE TABLE `<proj_id>`.`<dataset>`.`<destination_table>`` CLONE `<proj_id>`.`<dataset>`.`<destination_table>`;
```
So it looks like it still runs staging-optimized ?

### Expected behavior

I expect the query to look more like 
```
    INSERT INTO `{destination_table}` ({columns})
    SELECT {values}
    FROM `{staging_table}` s;
```

### Steps to reproduce

run a pipeline with a bigquery target and set `  "replace_strategy": "staging-optimized",` and then look at they queries that are run in your project history. 

### Operating system

macOS

### Runtime environment

Docker, Docker Compose

### Python version

3.12

### dlt data source

sql_table 

### dlt destination

Google BigQuery

### Other deployment details

Na

### Additional information

The workaround would be to first make dlt write out to a staging dataset  and then run something like 
this query yourself in some function afterwards
`    CREATE OR REPLACE TABLE `{destination_table}` AS
    SELECT *
    FROM `{staging_table}` s;`
Which I guess could work, but would be nicer if it worked as expected :)

anuunchin:
### Description
Previously, the `SqlStagingCopyFollowupJob` generated a clone sql for both the "staging-optimized" and "insert-from-staging" strategies. Now, its `generate_sql()` method receives the replace strategy from `SqlJobClientBase`'s `_create_replace_followup_jobs()` and correctly generates insert jobs based on that strategy. 
A relevant test has also been added.

### Related Issues

- Fixes #2422

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -441,7 +441,9 @@ class AthenaClient(SqlJobClientWithStagingDataset, SupportsStagingDestination):
         if self._is_iceberg_table(table_chain[0]):
             return [
                 SqlStagingCopyFollowupJob.from_table_chain(
-                    table_chain, self.sql_client, {"replace": True}
+                    table_chain,
+                    self.sql_client,
+                    {"replace": True, "replace_strategy": self.config.replace_strategy},
                 )
             ]
         return super()._create_replace_followup_jobs(table_chain)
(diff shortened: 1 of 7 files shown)


rudolfix (changes requested):
this is very good! but test looks like misconfigured

rudolfix (approved):
LGTM!

Issue #2398, opened 2025-03-12
rest_api: v1.8 GraphQL payload fails REST API config validation


davidcotton:
### dlt version

1.8.0, 1.8.1

### Describe the problem

dlt version 1.8.0 introduced a new (and very useful) [feature to extract query string and JSON parameters](https://github.com/dlt-hub/dlt/commit/644c04c36d14d80ea623b93f6f38fdde46c1c2d7#diff-06ab5847e7b22c83195c8dab991b10f9e1e4fb29fe9c20c9e29ad3381ae66af8) to define resource relationships. 

A side-effect of that change is that it limits the JSON data we can send in the request body, for example a POST request to a GraphQL API, as this change now validates the content of the JSON body. 

Prior to dlt 1.8 I could define a RESTAPIConfig like:
```
from dlt.sources.rest_api import RESTAPIConfig, rest_api_resources

config: RESTAPIConfig = {
    "client": {
        "base_url": "https://www.example.com",
    },
    "resources": [
        {
            "name": "artists",
            "endpoint": {
                "path": "",
                "method": "POST",
                "json": {
                    "query": """
                        query Artist {
                            artist(id: "123") {
                                id
                                name
                            }
                        }
                    """,
                },
                "data_selector": "data.artist",
            },
        },
    ],
}
yield from rest_api_resources(config)
```

However this now fails validation with a `ValueError: Expression '
    artist(id' defined in json is not valid. Valid expressions must start with one of: resources`

I think this is caused by t[his change](https://github.com/dlt-hub/dlt/commit/644c04c36d14d80ea623b93f6f38fdde46c1c2d7#diff-88519ac405535e9788510f3a539be736c0cd909ebb5924d9a95e9a13af3a644fR324-R325) now validating the JSON param.

Is there a way we could either disable validation on a given resource - or provide a JSON body that isn't validated? Possibly this is just stretching the REST API source too far and I should just build a GraphQL source?

### Steps to reproduce

```
from dlt.sources.rest_api import RESTAPIConfig, rest_api_resources

config: RESTAPIConfig = {
    "client": {
        "base_url": "https://www.example.com",
    },
    "resources": [
        {
            "name": "artists",
            "endpoint": {
                "path": "",
                "method": "POST",
                "json": {
                    "query": """
                        query Artist {
                            artist(id: "123") {
                                id
                                name
                            }
                        }
                    """,
                },
                "data_selector": "data.artist",
            },
        },
    ],
}
[...]

PR #2436, opened 2025-03-24
Fix/rest api handle expressions in graphql payload

francescomucio:
### Description
This is a fix to handle REST API POST requests to GraphQL endpoing

### Related Issues

- Fixes https://github.com/dlt-hub/dlt/issues/2398 

### Additional Context
Unfortunately to patch the current error, we had to use a workaround.

The current solution has two drawbacks:
- GraphQL queries can contain elements who look like placeholders (e.g. in the string `"query { shop { name } }", `{name}` is the name of a field returned by GraphQL endpoint, but it could also be the name of a params). For this reason we had to disable the method `_raise_if_any_not_in` is the call is to GraphQL endpoint.

- Not introduced in this PR, but GraphQL endpoints return valid JSON objects with an error message, in this case the REST API doesn't raise any error and insert the error message in the destination table. This is quite annoying. 

We need to have a better way to deal with GraphQL than just handle them as POST calls, we need to have an additional method or configuration to correctly handle placeholders and errors.

### DIFF
--- a/dlt/sources/rest_api/config_setup.py
+++ b/dlt/sources/rest_api/config_setup.py
@@ -322,8 +322,12 @@ def build_resource_dependency_graph(
         params_expressions = _find_expressions(endpoint_resource["endpoint"].get("params", {}))
         _raise_if_any_not_in(params_expressions, available_contexts, message="params")
 
-        json_expressions = _find_expressions(endpoint_resource["endpoint"].get("json", {}))
-        _raise_if_any_not_in(json_expressions, available_contexts, message="json")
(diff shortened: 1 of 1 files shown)


burnash:
This PR introduces special handling for GraphQL logic inside the rest_api source. This may work for now, but it would be cleaner and more maintainable to separate that GraphQL-specific functionality into a dedicated GraphQL source. That way, the rest_api source stays focused on REST use cases, and GraphQL logic can be handled in a context that better fits its unique needs.

PR #2437, opened 2025-03-25
rest_api: interpolate `json` and `headers` for top resources

### LINKED ISSUE #2398: rest_api: v1.8 GraphQL payload fails REST API config validation

burnash:
Fixes #2398

This PR:
- Runs `expand_placeholders()` on both `json` and `header` request parameters for top-level resources  
- Applies this regardless of whether `incremental_object` is defined, to properly handle escaped placeholders


dlt/sources/rest_api/__init__.py
```diff
@@ -307,25 +307,29 @@ def paginate_resource(
...
                         )
-                    # TODO: expand json as well. make sure you do not default to {} as this will
-                    #  generate empty body
-                    path = expand_placeholders(path, format_kwargs)
-                    params = expand_placeholders(params, format_kwargs)
+
+                # Always expand placeholders to handle escaped sequences
+                path = expand_placeholders(path, format_kwargs)
+                params = expand_placeholders(params, format_kwargs)
+                headers = (
```
FelixCaronBrp, line 329: No issue here, but expand_placeholders already returns None when the object is None, hence the additional if might be redundant.

burnash: Excellent observation, @FelixCaronBrp, thanks, fixed that in b89917c

FelixCaronBrp (approved):

dlt/sources/helpers/rest_client/client.py
```diff
@@ -99,8 +99,8 @@ def _create_request(
         self,
         path_or_url: str,
         method: HTTPMethod,
-        params: Optional[Dict[str, Any]] = None,
         headers: Optional[Dict[str, Any]] = None,
+        params: Optional[Dict[str, Any]] = None,
```
rudolfix, line 103: please move `headers` to the end

rudolfix (changes requested):

dlt/sources/helpers/rest_client/client.py
burnash: done in e1c4966

rudolfix (approved):
LGTM!

PR #2439, opened 2025-03-25
Reload module to update `SourceReference`

zilto:
When a lasting Python process wants to reload a `.py` module where a `Source` or `Resource` is defined, the current mechanism doesn't reload the reference from file.

This PR includes a test that should successfully pass.

Oddly, the test passes when using the debugger `pytest tests/path/to/file.py --pdb`, but fails when not using the debugger `pytest tests/path/to/file.py`. This likely has to do with core Python import mechanisms

### DIFF
--- a/dlt/common/reflection/ref.py
+++ b/dlt/common/reflection/ref.py
@@ -111,8 +111,11 @@ def object_from_ref(
     Returns: a tuple (typechecked attr, trace if error)
 
     """
+    # TODO consider using `module.submodule:object_in_module` notation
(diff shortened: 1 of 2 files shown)


zilto:
The fact that adding the debugger changed the behavior suggests that `locals()` and `globals()` scopes might be at play ([see reference](https://www.youtube.com/watch?v=dGNt88HMUpg)). I suggest we avoid hacking the Python import mechanisms

zilto:
Closing for now, but problem hasn't been fixed

PR #2444, opened 2025-03-26
update dlt and dlt+ installations

### LINKED ISSUE #2299: Update dlt and dlt+ installation docs
### Documentation description

We noticed while writing the dlt+ installation docs that the Python instructions for Windows might be wrong - why does it tell you to install pip only for Windows? We should investigate/fix that.

And also, we should probably change or (add to) the environment instructions to use uv, because the current ones are a very old way of creating environments.

### Are you a dlt user?

None

Soot3:
Only windows instructions had a line installing pip

Fixes #2299 @akelad 

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### Additional Context
Updated the python 3.10 installation for ubuntu as well. Previously it gave module not found errors. Didnt end up adding uv as it adds steps to the commands e.g pip install dlt will become uv pip install dlt

### DIFF
--- a/docs/website/docs/plus/getting-started/installation.md
+++ b/docs/website/docs/plus/getting-started/installation.md
@@ -61,6 +61,7 @@ After installing [Python 3.10 (64-bit version) for Windows](https://www.python.o
 
 ```sh
 C:\> pip3 install -U pip
+C:\> pip install uv
 ```
 
   </TabItem>
@@ -90,7 +91,7 @@ source .venv/bin/activate
   </TabItem>
   <TabItem value="macos">
 
-Create a new virtual environment in your working folder. This will create an `./venv` directory where your virtual environment will be stored:
+Create a new virtual environment in your working folder. This will create a `./venv` directory where your virtual environment will be stored:
 
 ```sh
 uv venv --python 3.10
@@ -105,7 +106,7 @@ source .venv/bin/activate
   </TabItem>
   <TabItem value="windows">
 
-Create a new virtual environment in your working folder. This will create an `./venv` directory where your virtual environment will be stored:
+Create a new virtual environment in your working folder. This will create a `./venv` directory where your virtual environment will be stored:
 
 ```bat
 C:\> uv venv --python 3.10
--- a/docs/website/docs/plus/getting-started/installation.md
+++ b/docs/website/docs/plus/getting-started/installation.md
@@ -40,7 +40,7 @@ You can install Python 3.10 with `apt`.
 ```sh
 sudo apt update
 sudo apt install python3.10
-sudo apt install python3.10-venv
+pip install uv
 ```
 
   </TabItem>
@@ -51,6 +51,7 @@ On macOS, you can use [Homebrew](https://brew.sh) to install Python 3.10.
 ```sh
 brew update
 brew install python@3.10
+brew install uv
 ```
 
   </TabItem>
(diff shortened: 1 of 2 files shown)


docs/website/docs/plus/getting-started/installation.md
```diff
@@ -51,6 +54,7 @@ On macOS, you can use [Homebrew](https://brew.sh) to install Python 3.10.
 ```sh
 brew update
 brew install python@3.10
+python3.10 -m ensurepip --upgrade
```
sh-rp, line 57: usually python should be available as the `python` command and you can use `pyenv` to manage multiple python versions and which one is active. Or is there a specific reason you add the python command with the version here?

docs/website/docs/reference/installation.md
```diff
@@ -23,9 +23,12 @@ If you have a different Python version installed or are missing pip, follow the
 You can install Python 3.10 with `apt`.
 
 ```sh
+sudo add-apt-repository ppa:deadsnakes/ppa
+sudo apt install -y software-properties-common
 sudo apt update
 sudo apt install python3.10
 sudo apt install python3.10-venv
+sudo apt install pip
```
sh-rp, line 31: I think pip should be installed automatically together with python, is that not the case?

docs/website/docs/plus/getting-started/installation.md
Soot3: @sh-rp the instructions came with a specific version range  (Python 3.9-3.13) so I just made sure all commands used the newly installed python rather than a potentially lower version

docs/website/docs/reference/installation.md
Soot3: Not always. And the main issue was that only the windows instructions installed pip explicitly.

akelad:
Hi @Soot3 thanks for the PR! I think there's a small misunderstanding with the issue though. There were two things in there:
* We wanted to validate whether you need to explicitly install pip on Windows (based on our own research, it seems that you do sometimes), the instructions for Mac/Ubuntu don't need to change.
* We should replace the instructions for a virtual environment to use `uv` instead. There's an example of how that might look like [here](https://rasa.com/docs/pro/installation/python#virtual-environment-setup)

So the only thing that needs to be done is the second part about the virtual environment - if you want to take that on, we'd be happy to accept that as a contribution :)

Soot3:
Okay then. Let me change the commit. Thank you @akelad

docs/website/docs/plus/getting-started/installation.md
```diff
@@ -51,6 +51,7 @@ On macOS, you can use [Homebrew](https://brew.sh) to install Python 3.10.
 ```sh
 brew update
 brew install python@3.10
+brew install uv
```
akelad, line 54:
```suggestion
pip install uv
```

akelad: you're also missing the `pip install uv` part for Windows - please add it!

akelad: Same comments for the dlt+ docs btw!

docs/website/docs/plus/getting-started/installation.md
```diff
@@ -74,46 +75,46 @@ This way, all the dependencies for your current project will be isolated from pa
 
   <TabItem value="ubuntu">
 
-Create a new virtual environment in your working folder. This will create an `./env` directory where your virtual environment will be stored:
+Create a new virtual environment in your working folder. This will create an `./venv` directory where your virtual environment will be stored:
```
akelad, line 78:
```suggestion
Create a new virtual environment in your working folder. This will create a `./venv` directory where your virtual environment will be stored:
```

akelad (commented):
Nice, thank you! I left some small comments, all of them also apply to the dlt+ page :)

Soot3:
@akelad  fixed. Thank you

akelad (approved):
looks good to me!

docs/website/docs/plus/getting-started/installation.md
```diff
@@ -60,6 +61,7 @@ After installing [Python 3.10 (64-bit version) for Windows](https://www.python.o
 
 ```sh
 C:\> pip3 install -U pip
+C:\> pip install uv
```
akelad, line 64:
```suggestion
C:\> pip3 install uv
```

docs/website/docs/reference/installation.md
```diff
@@ -45,6 +46,7 @@ After installing [Python 3.10 (64-bit version) for Windows](https://www.python.o
 
 ```sh
 C:\> pip3 install -U pip
+C:\> pip install uv
```
akelad, line 49:
```suggestion
C:\> pip3 install uv
```

akelad (approved):

akelad:
Thanks again for the submission @Soot3!

PR #2445, opened 2025-03-26
Fix failing bigquery test with new asserted error message

sh-rp:
### Description
An error message of bigquery has changed and we assert this in one of the tests. I have updated this.


burnash (approved):
Thanks @sh-rp

Issue #2360, opened 2025-02-28
JSONResponseCursorPaginator updateRequest method only works for GET requests


au-sebastian-alcaino:
### dlt version

1.7.0

### Describe the problem

JSONResponseCursorPaginator in its update_request method its only updating request.params which works for GET requests, but its missing updating request.json that is used by POST requests.
https://github.com/dlt-hub/dlt/blob/d8cfa393cc75d70100e660e210a0cd705d582467/dlt/sources/helpers/rest_client/paginators.py#L658

### Expected behavior

JSONResponseCursorPaginator should detect if params or json is being used, and update the json or params field accordingly.

### Steps to reproduce

Using this source configuration to talk to Jira api
```
pipeline = dlt.pipeline(
        pipeline_name=name,
        destination=dlt.destinations.postgres(
            {
                "host": os.environ["database_host"],
                "port": os.getenv("database_port", 5432),
                "database": os.environ["database_dbname"],
                "username": os.environ["database_username"],
                "password": os.environ["database_password"],
            }
        ),
        dataset_name="dset_name",
        progress="log",
    )

pipeline.run(rest_api_source({
            "client": {
                "base_url": f"https://{subdomain}.atlassian.net",
                "auth": HttpBasicAuth(email, api_token)
            },
            "resource_defaults": {
                "primary_key": "key",
                "write_disposition": "replace",
                "endpoint": {
                    "json": {
                        "maxResults": 100,
                    },
                },
            },
            "resources": [
                {
                    "name": "issues",
                    "endpoint": {
                        "path": "/rest/api/3/search/jql",
                        "paginator": JSONResponseCursorPaginator(cursor_path="nextPageToken",
                                                                 cursor_param="nextPageToken"),
                        "json": {
                            "fields": ["-comments"],
                            "expand": "fields,renderedFields",
                            "fieldsByKeys": True,
                            "jql": jql,
                        },
                        "method": "POST",
                        "data_selector": "issues",
                    },
                    "primary_key": "key"
                },
            ],
        }))
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

```python
rest_api_source({
            "client": {
                "base_url": f"https://{subdomain}.atlassian.net",
                "auth": HttpBasicAuth(email, api_token)
            },
            "resource_defaults": {
                "primary_key": "key",
                "write_disposition": "replace",
                "endpoint": {
                    "json": {
                        "maxResults": 100,
                    },
                },
            },
[...]

PR #2446, opened 2025-03-26
feat(paginator): enhance JSONResponseCursorPaginator to support cursor placement in request JSON body

kang8:
### Description

- Added `cursor_json` parameter to specify where to place the cursor in the JSON body.
- Updated documentation to reflect the new functionality and usage examples.
- Enhanced tests to cover new cursor handling scenarios in both query parameters and JSON body.

### Related Issues

- Closes: #2360


dlt/sources/helpers/rest_client/detector.py
```diff
@@ -180,7 +180,7 @@ def cursor_paginator_detector(response: Response) -> Tuple[JSONResponseCursorPag
     if not cursor_path_parts:
         return None, None
 
-    return JSONResponseCursorPaginator(cursor_path=".".join(cursor_path_parts)), 0.5
+    return JSONResponseCursorPaginator(cursor_path=".".join(cursor_path_parts), cursor_param='cursor'), 0.5
```
burnash, line 183: Why do you think this change is needed?

kang8:
I decided to create a dedicated JSONResponseBodyPaginator rather than extending JSONResponseCursorPaginator to handle request body pagination. This approach follows the Single Responsibility Principle - JSONResponseCursorPaginator remains focused on query parameter pagination, while JSONResponseBodyPaginator specifically handles cursor placement in the request body.

Benefits of this separation:

- Cleaner code with clear responsibility boundaries
- Better maintainability and testability
- More intuitive API for developers
- Preserves backward compatibility

Regarding the naming, I chose JSONResponseBodyPaginator to indicate it extracts cursor values from JSON responses and places them in request bodies. However, I'm open to alternative naming suggestions if there's a preference for different terminology.

kang8:
@burnash (sorry, for ping) I'd like to kindly check if you have had a chance to look at this PR.

I believe this implementation offers a clean solution to the issue #2360 by introducing a dedicated paginator for request body cursor pagination.

I'm happy to address any feedback or questions you might have about the approach. Thank you for your time!

burnash:
Hey @kang8, thanks for the contribution! Sorry I was away and couldn't review the first version of your PR sooner.

You're right, it's challenging to extend the `JSONResponseBodyPaginator` in a backward-compatible way due to the default value for `cursor_param`. I was also leaning toward creating a separate paginator, as you did in your latest commit. However, after further consideration, I think your initial approach might be preferable. That way, we maintain a single paginator handling cursor pagination, regardless of how it manages the token for the next page.

I believe it's still possible to preserve backward compatibility if `JSONResponseBodyPaginator` checks both the request method and the JSON path when setting the cursor in the payload. For example:

```python
# New: for POST requests with cursor_body_path set, add cursor to JSON body
if request.method == "POST" and self.cursor_body_path is not None:
    if request.json is None:
        request.json = {}
    self._set_value_at_path(request.json, self.cursor_body_path, self._next_reference)
else:
    # Previous code is backward compatible
    if request.params is None:
        request.params = {}

    request.params[self.cursor_param] = self._next_reference
```

Let me know if you see any edge cases or downsides.

kang8:
Thank you for your message.

I noticed you accidentally wrote "JSONResponseBodyPaginator" instead of "JSONResponseCursorPaginator" in your comment.

> I think your initial approach might be preferable.

I agree, and I've gone back to using the initial approach. I've made some adjustments to ensure backward compatibility while supporting the cursor_body_path parameter. I decided not to check for `request.method == "POST"` in the `update_request()` method, but instead enforced that only one of `cursor_param` or `cursor_body_path` can be set. This gives users more flexibility, as request bodies might be used with methods other than POST.

Finally, I really like the `cursor_body_path` naming and the `_set_value_at_path` design you suggested, so I've incorporated those into my implementation.

tests/sources/helpers/rest_client/test_client.py
```diff
@@ -400,21 +404,11 @@ def test_paginate_json_body_without_params(self, rest_client) -> None:
         # leave 3 pages of data
         posts_skip = (DEFAULT_TOTAL_PAGES - 3) * DEFAULT_PAGE_SIZE
 
-        class JSONBodyPageCursorPaginator(BaseReferencePaginator):
```
burnash, line 403: Very good!

burnash (approved):
Very good, thank you very much for the comprehensive contribution @kang8. Really appreciate docs and tests!

axelearning:
Thank you for implementing this feature! It's working perfectly !

PR #2456, opened 2025-03-29
Fix read parquet chunk size

diwu-sf:
### Description
Look at the actual documentation for `iter_batches` https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.iter_batches

The old code was incorrectly using a tiny 10 batch size thinking it was batches of parquet files to read, but the actual code was reading parquet files one by one, and iterating 10 records at a time from each file.

### Additional Context
Was seeing bad parquet performance, turns out the default chunk size was horribly inefficient.


rudolfix (approved):
LGTM! thanks!

PR #2462, opened 2025-03-31
rest_api: enhance placeholder expansion to preserve value types

### LINKED ISSUE #2458: rest_api: interpolation in JSON payload forces string type, breaking numeric fields
### Description

In the `rest_api` source, when interpolation is used in JSON payloads (e.g., `"json": {"post_id": "{resources.posts.id}"}`), the placeholder is always replaced with a string, even if the original value is numeric. This results in payloads like `{"post_id": "123"}` instead of `{"post_id": 123}`. For APIs expecting numeric types, this causes type mismatches and potential errors.

#### Proposed fix

Modify the `expand_placeholders` function to handle sole placeholders differently: if a value is exactly a placeholder (e.g., `"{resources.posts.id}"`), replace it with the actual value from `placeholders`, preserving its type (e.g., `123` as an integer).

burnash:
Fixes #2458


rudolfix (approved):
LGTM!

PR #2470, opened 2025-04-01
Fixes bigquery, ibis tests

anuunchin:
### Description
This PR tries to fix ci tests that are failing. 

1. `test_read_interfaces.py::test_ibis_dataset_access`

   - Root of cause: [This commit](https://github.com/ibis-project/ibis/commit/9b25ab1be5b43d8b1933a093500547710e122729) in the latest 10.4.0 ibis release, which refactored `get_schema` to use raw sql and introduced a dependency on the pg_enum system table, a table that doesn't exist in redshift, causing schema introspection to fail during `ibis.table(...)` calls.
  
2. `test_read_interfaces.py::test_ibis_expression_relation`

   - Root of cause: [This commit](https://github.com/ibis-project/ibis/commit/329ad7c1a1500b21c02b70dd06973fe1c89aa6a3) in the latest 10.4.0 ibis release, which basically replaces the CountStar syntax with `<table_name>_count`.

3. Bigquery error text change
   - Root of cause: A change in the error message. The fix is pretty straightforward.

### DIFF
--- a/dlt/helpers/ibis.py
+++ b/dlt/helpers/ibis.py
@@ -64,20 +64,11 @@ def create_ibis_backend(
         con = ibis.duckdb.from_connection(duck)
     elif issubclass(destination.spec, PostgresClientConfiguration):
         if destination.spec is RedshiftClientConfiguration:
-            # patch psycopg
-            try:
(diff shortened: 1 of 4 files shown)


rudolfix (approved):
LGTM!

PR #2478, opened 2025-04-02
types ibis relation in data access

rudolfix:
### Description
This PR adds typing to ibis relation so it looks like ibis Table when typechecking (and working in VSCode etc.)
- datasets are generic with relation type
- overloads are used to pick the right relation type when creating dataset or getting it from pipeline
- we derive ibis relaiton from ibis Table
- ibis backend helper is fixed to (1) set current schema so we can use dot notation to access tables (2) adds athena and databricks backends + tests
- fixes Athena to set catalog name correctly and removes location tag for iceberg (AWS fixed something apparently)

**to reviewer** part of the review is to checkout the code and see if expected typings and docstrings are visible (by dataset type and with ibis uninstalled)


zilto (commented):
This is great! All of the annotation are properly propagated via the connection `dataset.ibis()`. The annotation for `ReadableIbisRelation` have a few limitations (maybe not be worth the effort):

- `ReadableIbisRelation._ibis_object` has no type
- `dataset.table_name.col_name` has type `Any` instead of `ir.Column`
- `dataset.table_name["col_name"]` incorrectly returns a Table instead of a Column. Once we fix this, we'll want to properly type it.


## The code I used
I ran the chess pipeline then inspected in VSCode this file

```python
import dlt
import ibis
import ibis.expr.types as ir

pipeline = dlt.attach("chess_pipeline")
dataset = pipeline.dataset(dataset_type="ibis")

table = dataset.players_online_status
table2 = table[["username", "online_status"]]
table3 = table.select("username")
table4 = table["username", "online_status"]

table4 = table._ibis_object  # typing fails
table5 = table._ibis_object.select("username")  # typing fails

col1 = dataset.players_online_status["username"]  # should access a column (not a typing issue)
col2 = dataset.players_online_status.username  # typing limit

# raw ibis seems to work
con = dataset.ibis()
bound_table = con.table("players_online_status")
bound_table2 = con.table("players_online_status")[["username", "online_status"]]
bound_table3 = con.table("players_online_status").select("username")
bound_table4 = con.table("players_online_status")["username", "online_status"]

bound_col = con.table("players_online_status")["username"]  # properly typed
bound_col = con.table("players_online_status").username  # properly typed
```

rudolfix:
@zilto thanks for the review. you point to a few issues in directly adopting Table typing. we also had separate discussion on it. my take:
1. I will not derive `ReadableIbisRelation` from `ibis` `Table` when type checking. we can decide later what we do here
2. I will keep the other changes ie. adding more ibis backend, default datasets and dataset being generics with the relation type

PR #2480, opened 2025-04-03
Update docs for transaction handling - snowflake.md

sspaeti:
### Description
If you use Snowflake and have transaction settings not set to default, e.g., have set `ALTER SESSION SET AUTOCOMMIT = FALSE;`, dlt will not work properly and end up having an open session with uncommitted Inserts/merge statements.

I added a note to the docs with this PR to make that clear and not run into errors I did.

### Additional Context
Even better would be to explicitly set `ALTER SESSION SET AUTOCOMMIT = TRUE;` on the insert to destination schema as well as you do for inserting into Staging, where you added `ALTER SESSION SET AUTOCOMMIT = TRUE;` explicitly.

PS: I added the note to `Additional connection options`, not sure if that is the best place. Feel free to reorder or let me know where it would make more sense.

### DIFF
--- a/docs/website/docs/dlt-ecosystem/destinations/snowflake.md
+++ b/docs/website/docs/dlt-ecosystem/destinations/snowflake.md
@@ -132,6 +132,10 @@ This will set the timezone and session keep alive. Mind that if you use TOML, yo
 `"snowflake://loader/dlt_data?authenticator=oauth&timezone=UTC&client_session_keep_alive=true"`
 will pass `client_session_keep_alive` as a string to the connect method (which we didn't verify if it works).
 
+:::note
+Be aware that dlt is expecting the transaction behavior to `ALTER SESSION SET AUTOCOMMIT = TRUE;`. Meaning all DMLs are automatically committed or set to `ALTER SESSION SET AUTOCOMMIT = FALSE;` where needed by dlt. BUT, dlt will not set it explicitly to `TRUE` when needed; therefore, if you have set the transaction mode to `FALSE` on Snowflake, dlt will have opened COMMITs that will not be committed anymore. Please use the default behaviour of Snowflake to avoid running into issues.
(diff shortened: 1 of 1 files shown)


sspaeti:
Just FYI: I'm not sure if the `:::note` is not working, or the preview (above [link](https://deploy-preview-2480--dlt-hub-docs.netlify.app/) isn't correct_, but it does not show my additional note on the preview.

rudolfix:
@sspaeti how did you manage to change session settings? each loader job gets a separate connection which is closed after COPY/MERGE command. I'm asking because indeed we didn't start the transaction before SQL job and I changed that: we do it in `LoadJob` managed run method. MERGE/COPY SQL scripts are not affected.

sspaeti:
@rudolfix It was set at a account level (globally). As it can be set on three levels:

- Account (global) -> `alter account set autocommit = false;`
- User                    -> `alter user <username> set autocommit = false;`
- Session               -> `alter session set autocommit = false;`

Whereas session user and account settings overrule. See https://docs.snowflake.com/en/sql-reference/parameters#label-autocommit.

I believe this happens, because you only explicitly set the `ALTER SESSION SET AUTOCOMMIT = TRUE;` but not for `FALSE`. Meaning if someone has it the global/account settings set to `FALSE`, that's when we ran into the issue of open sessions. Small issues, but might be worth mentioning or make it explicit in both way? 🤔

rudolfix:
@sspaeti now I get it? should we just set autocommit explicitly to TRUE on opening the connection? we do that anyway after comitting/rollbacking the transaction.
if we do that, we just need to mention that we set it explicitly in the docs and user can still have those account settings, right?

rudolfix:
this PR will explicitly set autocommit on new connection https://github.com/dlt-hub/dlt/pull/2593

PR #2485, opened 2025-04-04
feat: Ibis execution method proxies added

zilto:
### Description
TL;DR allow `ReadableIbisRelation.execute()`, `.to_pandas()`, `.to_parquet()` etc.

When accessing a dataset using Ibis, you get an `ReadableIbisRelation` that's holds an `UnboundTable` object. This allows to build expressions and later compile them to SQL and execute them. 

These unbound tables and expressions can't be executed directly, which is impractical when iterating over data transformations.

This PR uses the `ReadableIbisRelation`'s reference to the dataset to create an Ibis connection on the fly and execute the unbound expression.

### DIFF
--- a/dlt/destinations/dataset/ibis_relation.py
+++ b/dlt/destinations/dataset/ibis_relation.py
@@ -24,6 +24,22 @@ TRANSPILE_VIA_DEFAULT = [
     "presto",
 ]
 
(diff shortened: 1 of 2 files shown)


zilto:
Other solutions preferred

PR #2496, opened 2025-04-08
Fix: sqlalchemy table names of type sqlalchemy.sql.elements.quoted_name causing deserialization error

### LINKED ISSUE #2452: SQLAlchemy object in schema YAML causes deserialization error
### dlt version

1.8.1

### Describe the problem

**Describe the bug**
When running a pipeline with `import_schema_path="pipelines/schemas/import"` set in the config.toml the resource fields of the import schema are not serialized correctly. 
This happens for every resource. 

```
resource: !!python/object/apply:sqlalchemy.sql.elements.quoted_name
- Partner
- null
```

**Screenshots**
Image

Slack thread with more context:
https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1739779561709789

### Expected behavior

The import schema should be serialized correctly.
  `  resource: Partner`

### Steps to reproduce

Run a pipeline, in my case im using the `sql_database()` verified resources against a Postgres DB.
It fails when encountering the SQLAlchemy object serialized in YAML:
Public Postgres: 
https://rnacentral.org/help/public-database

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

dlt verified source: `sql_database()`

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

Manually resolving the import schema fixes this.

anuunchin:
### Description
This PR simply force-casts sqlalchemy table names of type `sqlalchemy.sql.elements.quoted_name` to string. As a result, the following utility function in `dlt/common/schema/utils.py` stops serializing them as Python objects. 
```python
def to_pretty_yaml(stored_schema: TStoredSchema) -> str:
    return yaml.dump(stored_schema, allow_unicode=True, default_flow_style=False, sort_keys=False)
```

### Related Issues

- Fixes #2452 

### Additional Context
We could alternatively handle it in the utility function with something like:
```python
import yaml

def generic_str_subclass_representer(dumper, data):
    return dumper.represent_str(str(data))

# This will match all subclasses of str (including quoted_name)
yaml.add_multi_representer(str, generic_str_subclass_representer)
```
so that we handle all string subclasses, not just `sqlalchemy.sql.elements.quoted_name`.

### DIFF
--- a/dlt/sources/sql_database/__init__.py
+++ b/dlt/sources/sql_database/__init__.py
@@ -233,7 +233,7 @@ def sql_table(
         hints = {}
 
     return decorators.resource(
-        table_rows, name=table, write_disposition=write_disposition, **hints
+        table_rows, name=str(table), write_disposition=write_disposition, **hints
     )(
         engine,
         table_obj if table_obj is not None else table,  # Pass table name if reflection deferred
--- a/tests/load/sources/sql_database/test_sql_database_source.py
+++ b/tests/load/sources/sql_database/test_sql_database_source.py
@@ -3,6 +3,7 @@ from copy import deepcopy
 from typing import Any, Callable, cast, List, Optional, Set
 
 import pytest
+from pytest_mock import MockerFixture
 
 import dlt
 from dlt.common import json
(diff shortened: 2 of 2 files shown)


rudolfix (approved):
LGTM!

but to be 100% complete you should add a test that checks what we get when we do str on quoted identifier? do we get a quoted string? or not?

we do not have any tests for sqlalchemy helper but ideally we shoudl have them and see what quoted and unquoted strings return (claude will generate such code...)

rudolfix (approved):
this is double plus good now!

PR #2498, opened 2025-04-08
exp: DltBackend for ibis implemented

zilto:
Continuous work to improve the transformation engine.

## Requirements
- instantiate lazy tables from the dlt table schema; we don't want to query the destination for table information
- enable `ibis -> SQL string` conversion to use the dlt engine for execution
- try to support the full Ibis user-facing API / avoid proxying

## Intended user workflow
```python
pipeline = dlt.pipeline(...)
dataset = pipeline.dataset()
con = dataset.ibis()

customers = con.table("customers")
orders = con.table("orders")

query = (
  customers
  .join(
    orders.filter(orders.date > "2025-02-01"),
    customers.id == orders.customer_id,
    how="left",
  )
  .select("name", "address", "category", "cost")
  .group_by("category")
  .cost.sum()
)

# .execute() returns a pandas dataframe by default
result = query.execute()
# we can override for this to use dlt's implementation (not yet in the PR)
result.to_pyarrow()
```

## Technical implementation
- the `DltBackend` class is an Ibis backend. It's responsible for translating and executing Ibis expressions.
- the `DltBackend` subclasses the `SQLBackend` which implements almost all required methods through SQLGlot. The Ibis expression is compiled to an SQLGlot expression, and then compiled to an SQL string.
- no custom compiler is implemented, we just need to map dlt destinations to the right compiler. For example, `filesystem` destination has an SQLClient that uses Duckdb, so we compile `ibis -> sqlglot duckdb -> string`
- the SQL clients of the dlt destinations are used for execution

## Future work / Limitations
- I currently disabled UDF and extension support, but it should work
- Ibis has a hook system that we can use for logging, store things on the backend object, or even write to destination
- Ibis has a caching system using temporary tables that is useful for executing transformation DAGs

### DIFF
--- a/dlt/destinations/dataset/dataset.py
+++ b/dlt/destinations/dataset/dataset.py
@@ -42,12 +42,9 @@ class ReadableDBAPIDataset(SupportsReadableDataset):
 
     def ibis(self) -> IbisBackend:
         """return a connected ibis backend"""
-        from dlt.helpers.ibis import create_ibis_backend
+        from dlt.destinations.dataset.ibis_relation import DltBackend
 
-        return create_ibis_backend(
-            self._destination,
-            self._destination_client(self.schema),
-        )
+        return DltBackend.from_dataset(self)
 
     @property
     def schema(self) -> Schema:
--- a/dlt/destinations/dataset/ibis_relation.py
+++ b/dlt/destinations/dataset/ibis_relation.py
@@ -1,12 +1,37 @@
-from typing import TYPE_CHECKING, Any, Union, Sequence
-
+from __future__ import annotations
+import contextlib
 from functools import partial
+from typing import TYPE_CHECKING, Any, Union
 
+# ibis imports follow the convention used in the library
+import ibis
+import ibis.backends.sql.compilers as sc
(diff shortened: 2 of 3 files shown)


rudolfix (changes requested):
this LGTM and backend implementation goes in a good direction. My bigqest question is:
do we replace `ReadableIbisRelation` with ibis `Table`? I'm always overdoing so my take is that we should be able to derive from ibis `Table` and `BaseReadableDBAPIRelation`. maybe this is somehow possible? then we could have the same interface to
- generate text queries
- infer dlt schemas
- expose `dataset` as a property
- have our own data reading methods like `df()` (which I like way more than what ibis does)

WDYT?

zilto:
This draft PR made suggestions how to bind expressions to dlt data. Design session decided for another approach focusing on the `Relation` object. 

Useful code from this PR (e.g., getting the Ibis to SQLGlot compiler) was moved to this [branch/PR](https://github.com/dlt-hub/dlt/pull/2757).

zilto:
closing in favor of #3179

PR #2503, opened 2025-04-10
cli: `dlt ai IDE setup` command added

zilto:
This new CLI command allows to generate "rules files" for the user's project according to IDE-specific specs. This follows similar logic as `dlt init` command for core sources.

## Design choices
- The format `dlt ai IDE setup` was used, but it could be `dlt ai setup IDE`

## TODO
- improve rules files content
- support more IDEs
- add tests
- think about mechanism to "update" all rules files at once


rudolfix (changes requested):
command is good and well structured. but at the end @adrianbr needs to give it a test run.

there's one thing I'd change: please re-use the code that clones verified sources and take files from there. we have a place for it
https://github.com/dlt-hub/verified-sources/tree/master/init

right now it is not used (we moved init template to core code) so you can wipe it out and place the files there.

AFAIK @adrianbr also plans to put a few docs files that may be common to all editors. we'll know soon

zilto:
@rudolfix If I understand, you suggest placing `rules.md` files in `dlt-hub/verified-sources/init`. This directory is no longer used by `dlt`, so I could just used the `dlt` function "get files from verified-sources/init" instead of implementing new logic

Questions:
- wouldn't older `dlt` version still pull from `dlt-hub/verified-sources/init`? Deleting files would be a breaking change forcing version upgrade
- any reason to put this files in `dlt-hub/verified-sources` vs. `dlt-hub/dlt`? Easier to update the rules files? The downside I see is that rules should probably match the `dlt` version (easy to manually fix though)
- for code reuse, I looked at `dlt pipeline init` command it's not really reusable without refactoring. This PR implements the same logic but trimmed

rudolfix:
@zilto my goal here is to let @adrianbr and @AstrakhantsevaAA to quickly iterate the rules without re-releasing our library.

we did that with `init` repo (discontinued) and then with `verified-sources`. btw. this is also the reason I used a real running code without any templating and did all the changes by parsing Python to AST and rewriting scripts.. but I digress :)

Answers:
> wouldn't older dlt version still pull from dlt-hub/verified-sources/init? Deleting files would be a breaking change forcing version upgrade

version `0.5.x` did that and users have already instructions to use `--branch 0.5.x` (AFAIK) argument if they want to pull compatible version. and what I can see, the old docs are down anyway... so I bet this is not a problem

> any reason to put this files in dlt-hub/verified-sources vs. dlt-hub/dlt? Easier to update the rules files? The downside I see is that rules should probably match the dlt version (easy to manually fix though)

the idea was that verified sources is way smaller to clone locally and making PRs is easier. is it really like this? for init scripts it definitely was (there was just simple linting that checked if dlt command will run for all possible init variants)

backward compat: we planned to use version tags to pull always the right version of verified sources. but ultimately such strict biding was never necessary. sill users can pass their own location and branch to init command so they can run their own repos with their private code

> for code reuse, I looked at dlt pipeline init command it's not really reusable without refactoring. This PR implements the same logic but trimmed

I meant code that clones the verified source repo into know location. it is in `dlt.common.git` module. it does a few tricks to minimize the pulls, also you can 

tldr;> see first sentence of my comment.

dlt/cli/ai_command.py
```diff
@@ -0,0 +1,59 @@
...
+
+def ai_setup_command(ide: TSupportedIde = None) -> None:
+    """Get AI rules files into your local project for the selected IDE.
+    
+    Get the source and destination directories for the rules files.
+    Files found in the source directory will be copied into the destination directory.
+    """
+    specs = IDE_RULES_SPECS[ide]
+
+    src_storage = FileStorage(str(Path(dlt.__file__).parent.joinpath(RULES_TEMPLATE_RELATIVE_PATH)))
```
rudolfix, line 39: I'd go for something way dumber. you have
- /ai folder
- it has subfolders with names of the editors
- those subfolders contains files to copy into `run_dir`

the command looks if it has a subfolder with editor name and copies all the files

this way you can add and remove editors, add more files etc. without changes in the code. WDYT?

OFC you should not overwrite existing files.

dlt/cli/ai_command.py
```diff
@@ -0,0 +1,59 @@
...
+            continue
+        
+        shutil.copy2(src_file_path, dest_file_path)
+
+
+# TODO create a command to create a copy-pasteable MCP server config
+
+
+def mcp_command() -> None:
+    ...
```
rudolfix, line 59: this will be cool. questions:
- should we integrate it into cli that actually runs mcp? for example `dlt mcp info` would show it?
- should support both stdio and sse (with port) to make stuff easier?
a very top thing would be to somehow be aware of which editor is set up and display info for it without asking for a type

dlt/cli/plugins.py
```diff
@@ -645,6 +647,29 @@ def execute(self, args: argparse.Namespace) -> None:
...
+            title="Available subcommands", dest="operation", required=False
+        )
+        ai_subparsers.add_parser(
+            "setup", help="Generate IDE-specific configuration and rules files",
+        )
+        # TODO support MCP-proxy configuration
+        # ai_mcp_cmd = ai_subparsers.add_parser("mcp", help="Launch the dlt MCP server")
+    
+    def execute(self, args: argparse.Namespace) -> None:
+        ai_setup_command_wrapper(args.ide)
```
rudolfix, line 671: should we check if project is initialized already? ie. if `.dlt` folder exists (you check if run_context.settings exist)

zilto:
See this PR https://github.com/dlt-hub/verified-sources/pull/608 for related files

rudolfix (commented):
LGTM! is there a branch with the content in verified sources already?

rudolfix (commented):
this is really good. we still miss documentation. I'll come if we want to do something more that jus cli reference update

zilto:
> this is really good. we still miss documentation.
Ok, I'm not crazy. I **did** update the docs, but CI was failing because using `set` with argparse would make the CI check non-deterministic

PR #2507, opened 2025-04-10
Normalize model files

sh-rp:
### Description

Starting point of implementing normalization of model files.

In this PR a ModelNormalizer is added which needs to be implemented and tested. There is an attempt at adding a dlt_load_id, but at this point this will fail because of the way the insert statement is created in the load stage. Furthermore, regarding the dlt_load_id, the __call__ method of the normalizer will have to update the schema and return this schema update in cases where the dlt_load_id column is not yet part of the schema.

### DIFF
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -203,11 +203,9 @@ class ModelLoadJob(RunnableLoadJob, HasFollowupJobs):
             )
 
         # Get identifiers for the columns in the INSERT statement
-        # Normalizer aliased all selections, unless it was a star
+        # Normalizer aliased all selections
         columns = []
         for _, expr in enumerate(top_level_select.expressions):
-            if isinstance(expr, sqlglot.exp.Star):
-                break
             alias_name = expr.alias
             columns.append(sqlglot.exp.to_identifier(alias_name))
 
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -40,6 +40,7 @@ from dlt.common.schema.utils import (
     normalize_table_identifiers,
     version_table,
 )
+from dlt.common.utils import read_dialect_and_sql
 from dlt.common.storages import FileStorage
 from dlt.common.storages.load_package import LoadJobInfo, ParsedLoadJobFileName
 from dlt.common.schema import TColumnSchema, Schema, TTableSchemaColumns, TSchemaTables
@@ -147,10 +148,10 @@ class ModelLoadJob(RunnableLoadJob, HasFollowupJobs):
 
     def run(self) -> None:
         with FileStorage.open_zipsafe_ro(self._file_path, "r", encoding="utf-8") as f:
-            select_dialect = (
-                f.readline().split(":")[1].strip() or self._job_client.capabilities.sqlglot_dialect
+            select_dialect, select_statement = read_dialect_and_sql(
+                file_obj=f,
+                fallback_dialect=self._job_client.capabilities.sqlglot_dialect,  # caps are available at this point
             )
-            select_statement = f.read()
 
         sql_client = self._job_client.sql_client
         insert_statement = self._insert_statement_from_select_statement(
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -202,23 +202,14 @@ class ModelLoadJob(RunnableLoadJob, HasFollowupJobs):
                 " using a SELECT statement."
             )
 
-        # Casefold column names according to destination and wrap them in aliases
+        # Get identifiers for the columns in the INSERT statement
+        # Normalizer aliased all selections, unless it was a star
         columns = []
-        for i, expr in enumerate(top_level_select.expressions):
-            if isinstance(expr, sqlglot.exp.Column):
-                original_name = expr.name
-                alias_name = self._job_client.capabilities.casefold_identifier(original_name)
-                alias = sqlglot.exp.Alias(
-                    this=expr.copy(),
-                    alias=sqlglot.exp.to_identifier(alias_name),
-                )
-                top_level_select.expressions[i] = alias
-                columns.append(sqlglot.exp.to_identifier(alias_name))
-            elif isinstance(expr, sqlglot.exp.Alias):
-                alias_name = expr.alias
-                casefolded_alias = self._job_client.capabilities.casefold_identifier(alias_name)
-                expr.set("alias", casefolded_alias)
-                columns.append(sqlglot.exp.to_identifier(casefolded_alias))
+        for _, expr in enumerate(top_level_select.expressions):
+            if isinstance(expr, sqlglot.exp.Star):
+                break
+            alias_name = expr.alias
+            columns.append(sqlglot.exp.to_identifier(alias_name))
 
         # Build final INSERT
         query = sqlglot.expressions.insert(
--- a/dlt/normalize/items_normalizers.py
+++ b/dlt/normalize/items_normalizers.py
@@ -88,8 +88,13 @@ class ModelItemsNormalizer(ItemsNormalizer):
                 partition_by=None,
                 order=None,
             )
-            casted_row_num = sqlglot.exp.Cast(this=row_num, to=sqlglot.exp.DataType.build("TEXT"))
-            return sqlglot.exp.func("MD5", casted_row_num)
+            concat_expr = sqlglot.exp.func(
+                "CONCAT",
+                sqlglot.exp.Column(this=self._normalize_casefold(C_DLT_LOAD_ID)),
+                sqlglot.exp.Literal.string("-"),
+                row_num,
+            )
+            return sqlglot.exp.func("MD5", concat_expr)
         elif dialect == "sqlite":
             return sqlglot.exp.func(
                 "lower",
@@ -196,11 +201,6 @@ class ModelItemsNormalizer(ItemsNormalizer):
             outer_parsed_select (sqlglot.exp.Select): The parsed outer SELECT statement.
             columns (TTableSchemaColumns): The schema columns to match.
         """
-        if len(outer_parsed_select.selects) == 1 and isinstance(
-            outer_parsed_select.selects[0], sqlglot.exp.Star
-        ):
-            return None
-
         # Map alias name -> expression
         alias_map = {expr.alias.lower(): expr for expr in outer_parsed_select.selects}
 
@@ -295,7 +295,7 @@ class ModelItemsNormalizer(ItemsNormalizer):
         # The query is ensured to be a select statement upstream,
         # but we double check here
         if not isinstance(parsed_select, sqlglot.exp.Select):
-            raise ValueError("Only SELECT statements should reach the model normalizer.")
+            raise ValueError("Only SELECT statements should be used as SqlModel queries.")
 
         outer_parsed_select, needs_reordering = self._build_outer_select_statement(
             select_dialect, parsed_select, self.schema.get_table_columns(root_table_name)
--- a/dlt/normalize/items_normalizers.py
+++ b/dlt/normalize/items_normalizers.py
@@ -64,7 +64,7 @@ class ItemsNormalizer:
 
 
 class ModelItemsNormalizer(ItemsNormalizer):
-    def _add_or_replace_column(
+    def _adjust_wth_dlt_columns(
         self,
         parsed_select: sqlglot.exp.Select,
         column_name: Literal["_dlt_load_id", "_dlt_id"],
@@ -111,6 +111,7 @@ class ModelItemsNormalizer(ItemsNormalizer):
                 schema.naming,
             )
 
+            schema.update_table(partial_table, normalize_identifiers=False)
             table_updates = schema_update.setdefault(root_table_name, [])
             table_updates.append(partial_table)
 
@@ -130,21 +131,26 @@ class ModelItemsNormalizer(ItemsNormalizer):
         parsed_select = sqlglot.parse_one(select_statement, read=select_dialect)
         parsed_select = cast(sqlglot.exp.Select, parsed_select)
 
+        # normalize selected columns - basically all of them
+        schema = self.schema
+        table = schema.tables[root_table_name]
+        schema.update_table(table, normalize_identifiers=True)
+
         schema_updates = []
 
-        if self.config.json_normalizer.add_dlt_load_id:
+        if self.config.model_normalizer.add_dlt_load_id:
             dlt_load_id_alias = sqlglot.exp.Alias(
                 this=sqlglot.exp.Literal.string(self.load_id),
                 alias=sqlglot.exp.to_identifier("_dlt_load_id"),
             )
-            load_id_update = self._add_or_replace_column(
+            load_id_update = self._adjust_wth_dlt_columns(
                 parsed_select, "_dlt_load_id", dlt_load_id_alias, root_table_name
             )
             if load_id_update:
                 schema_updates.append(load_id_update)
 
         if (
-            self.config.json_normalizer.add_dlt_id
+            self.config.model_normalizer.add_dlt_id
             and self.config.destination_capabilities.sqlglot_dialect != "redshift"
         ):
             # TODO: redshift doesn't have an in-built function that generates unique values
@@ -158,7 +164,7 @@ class ModelItemsNormalizer(ItemsNormalizer):
                 alias=sqlglot.exp.to_identifier("_dlt_id"),
             )
 
-            dlt_id_update = self._add_or_replace_column(
+            dlt_id_update = self._adjust_wth_dlt_columns(
                 parsed_select, "_dlt_id", dlt_id_alias, root_table_name
             )
 
--- a/dlt/normalize/items_normalizers.py
+++ b/dlt/normalize/items_normalizers.py
@@ -64,40 +64,63 @@ class ItemsNormalizer:
 
 
 class ModelItemsNormalizer(ItemsNormalizer):
-    def _adjust_wth_dlt_columns(
+    def _adjust_with_dlt_columns(
         self,
         parsed_select: sqlglot.exp.Select,
-        column_name: Literal["_dlt_load_id", "_dlt_id"],
-        alias_expression: sqlglot.exp.Alias,
         root_table_name: str,
     ) -> Optional[TSchemaUpdate]:
-        """
-        Adds or replaces a column in the SELECT clause of a parsed SQL query and
-        returns a single TSchemaUpdate if a new column was added, otherwise None.
-        """
-        schema = self.schema
-        schema_update: TSchemaUpdate = {}
+        if len(parsed_select.selects) == 1 and isinstance(
+            parsed_select.selects[0], sqlglot.exp.Star
+        ):
+            logger.warning(
+                f"Skipping adding dlt columns to model query {parsed_select.sql()}: contains only a"
+                " star expression, so column names are unknown."
+            )
+            return None
 
-        column_replaced = False
-        for i, select in enumerate(parsed_select.selects):
-            # Check if the column already exists
-            select_alias = (
-                select.alias_or_name.lower()
-                if isinstance(select, sqlglot.exp.Alias)
-                else (
-                    select.output_name.lower() if isinstance(select, sqlglot.exp.Column) else None
-                )
+        schema_update: TSchemaUpdate = {}
+        schema = self.schema
+        dialect = self.config.destination_capabilities.sqlglot_dialect
+
+        # Build aliases based on config
+        dlt_columns: dict[str, Optional[sqlglot.exp.Alias]] = {}
+
+        if self.config.model_normalizer.add_dlt_load_id:
+            dlt_columns["_dlt_load_id"] = sqlglot.exp.Alias(
+                this=sqlglot.exp.Literal.string(self.load_id),
+                alias=sqlglot.exp.to_identifier("_dlt_load_id"),
             )
 
-            if select_alias == column_name:
-                # Replace the existing column
-                parsed_select.selects[i] = alias_expression
-                column_replaced = True
-                break
+        if self.config.model_normalizer.add_dlt_id and dialect != "redshift":
+            function_name = "generateUUIDv4" if dialect == "clickhouse" else "UUID"
+            dlt_columns["_dlt_id"] = sqlglot.exp.Alias(
+                this=sqlglot.exp.func(function_name),
+                alias=sqlglot.exp.to_identifier("_dlt_id"),
+            )
 
-        if not column_replaced:
-            # Append the column if it doesn't exist
-            parsed_select.selects.append(alias_expression)
+        # Replace if exists, otherwise append
+        for i, select in enumerate(parsed_select.selects):
+            for column_name, alias_expr in dlt_columns.items():
+                if alias_expr is None:
+                    continue
+                select_alias = (
+                    select.alias_or_name.lower()
+                    if isinstance(select, sqlglot.exp.Alias)
+                    else (
+                        select.output_name.lower()
+                        if isinstance(select, sqlglot.exp.Column)
+                        else None
+                    )
+                )
+                if select_alias == column_name:
+                    parsed_select.selects[i] = alias_expr
+                    dlt_columns[column_name] = None  # Mark as replaced
+
+        # Append any not-replaced aliases and update schema
+        for column_name, alias_expr in dlt_columns.items():
+            if alias_expr is None:
+                continue
+            parsed_select.selects.append(alias_expr)
 
             partial_table = normalize_table_identifiers(
                 {
@@ -110,14 +133,11 @@ class ModelItemsNormalizer(ItemsNormalizer):
                 },
                 schema.naming,
             )
-
             schema.update_table(partial_table, normalize_identifiers=False)
             table_updates = schema_update.setdefault(root_table_name, [])
             table_updates.append(partial_table)
 
-            return schema_update
-
-        return None
+        return schema_update if schema_update else None
 
     def __call__(self, extracted_items_file: str, root_table_name: str) -> List[TSchemaUpdate]:
         with self.normalize_storage.extracted_packages.storage.open_file(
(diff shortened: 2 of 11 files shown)


anuunchin:
`There is an attempt at adding a dlt_load_id, but at this point this will fail because of the way the insert statement is created in the load stage.` This has stopped failing (handled in [feat/2366-sql-jobs-3](https://github.com/dlt-hub/dlt/tree/feat/2366-sql-jobs-3) ), except for the case when the Select statement has a star schema.

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +63,120 @@ def __init__(
...
+                if isinstance(select, sqlglot.exp.Alias)
+                else (
+                    select.output_name.lower() if isinstance(select, sqlglot.exp.Column) else None
+                )
+            )
+
+            if select_alias == column_name:
+                # Replace the existing column
+                parsed_select.selects[i] = alias_expression
+                column_replaced = True
```
sh-rp, line 95: small note: you can just return None here, then you do not need column_replace and the if clause further down

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +63,120 @@ def __init__(
...
+        column_name: Literal["_dlt_load_id", "_dlt_id"],
+        alias_expression: sqlglot.exp.Alias,
+        root_table_name: str,
+    ) -> Optional[TSchemaUpdate]:
+        """
+        Adds or replaces a column in the SELECT clause of a parsed SQL query and
+        returns a single TSchemaUpdate if a new column was added, otherwise None.
+        """
+        schema = self.schema
+        schema_update: TSchemaUpdate = {}
```
sh-rp, line 79: I would raise here if a star select is detected. Add a new exception that explains why dlt_id and dlt_load_id can't be processed if not all column names are known.

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +63,120 @@ def __init__(
...
+                parsed_select, "_dlt_load_id", dlt_load_id_alias, root_table_name
+            )
+            if load_id_update:
+                schema_updates.append(load_id_update)
+
+        if (
+            self.config.json_normalizer.add_dlt_id
+            and self.config.destination_capabilities.sqlglot_dialect != "redshift"
+        ):
+            # TODO: redshift doesn't have an in-built function that generates unique values
```
sh-rp, line 150: They have these user defined functions in redshift, where you essentially can just use a python method. I'm not sure we will be able to get this to work with sqlglot though. If it does not, we should raise here in case adding dlt_id is set to true.

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +63,120 @@ def __init__(
...
+                file_obj=f,
+                fallback_dialect=self.config.destination_capabilities.sqlglot_dialect,  # caps are available at this point
+            )
+
+        parsed_select = sqlglot.parse_one(select_statement, read=select_dialect)
+        parsed_select = cast(sqlglot.exp.Select, parsed_select)
+
+        schema_updates = []
+
+        if self.config.json_normalizer.add_dlt_load_id:
```
sh-rp, line 135: I am not quite sure if part of this task is also to harmonize the normalizer behavior across loader_file_formats, if it is, we can ditch the `json_normalizer` and `parquet_normalizer` on `NormalizeConfiguration` and just use on normalizer config for all three normalizers. If we do not harmonize, then I'd instead add a `model_normalizer`config that has the same defaults as the `json_normalizer`.

dlt/normalize/items_normalizers.py
sh-rp: Another idea, that I maybe mentioned elsewhere: you may be able to get the row_number of the currently processing row and combine it with a hashfunction (redshift has both features afaik) to get a mechanism to create a unique id.

anuunchin:
Sqlite doesn't have an in built UUID, will probably resort to something similar to what's used in Redshift's case

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +63,217 @@ def __init__(
...
+    def __call__(self, extracted_items_file: str, root_table_name: str) -> List[TSchemaUpdate]:
+        with self.normalize_storage.extracted_packages.storage.open_file(
+            extracted_items_file, "r"
+        ) as f:
+            select_dialect, select_statement = read_dialect_and_sql(
+                file_obj=f,
+                fallback_dialect=self.config.destination_capabilities.sqlglot_dialect,  # caps are available at this point
+            )
+
+        parsed_select = sqlglot.parse_one(select_statement, read=select_dialect)
```
sh-rp, line 248: We should probably raise here if this was not a select statement. Then you can just remove all the other tests for star selects in the individual steps. I think supporting these is too complicated at this point.

tests/load/pipeline/test_model_item_format.py
```diff
@@ -405,3 +419,293 @@ def example_table() -> Any:
...
+        "a",
+        "b",
+        "_dlt_id",
+        "_dlt_load_id",
+    }
+
+    # Validate that each table has exactly one model job
+    assert count_job_types(pipeline) == {
+        target_table_name: {"model": 1},
+    }
```
sh-rp, line 492: should you not test the values of these rows? is the load_id set correctly and is the dlt_id unique?

tests/load/pipeline/test_model_item_format.py
```diff
@@ -405,3 +419,293 @@ def example_table() -> Any:
...
+        allow_base64_binary=destination_config.destination_type == "clickhouse",
+    )
+
+
+@pytest.mark.parametrize("tables_contract", ["freeze", "evolve"])
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(
+        default_sql_configs=True,
+        subset=["duckdb"],
```
sh-rp, line 554: good!

tests/normalize/test_model_item_normalizer.py
```diff
@@ -0,0 +1,356 @@
...
+    if dialect == "duckdb":
+        assert (
+            f"SELECT subquery.a AS a, subquery.b AS b, '{load_id}' AS _dlt_load_id, UUID() AS"
+            " _dlt_id FROM (SELECT a, b FROM my_table) AS subquery\n"
+            == normalized_select_query
+        )
+    elif dialect == "bigquery":
+        assert (
+            f"SELECT subquery.a AS a, subquery.b AS b, '{load_id}' AS _dlt_load_id, GENERATE_UUID()"
+            " AS _dlt_id FROM (SELECT a, b FROM my_table) AS subquery\n"
```
sh-rp, line 137: maybe we should use a different identifier than "subquery" in the normalizer, something more abstract so that we do not have collissions with anything the user may provide? maybe "_dlt_subquery" or something like this?

sh-rp (commented):
Since I opened the PR I can't request changes :) But I added some comments but have to review the tests more closely. The two bigger tasks in view are:

* Disallow any star selects, raise with a good exception
* Write a test that inserts from two tables with different but overlapping schemas into a third table, so that we have the case that the columns in the existing table schema are a superset of the columns coming from the incoming hints and sql expression.

Other observations so far:
* You either need to set "ci-full" on this PR to make all tests run on all destinations or make more of the tests "essential"
* We had this idea to not modify the query if re-ordering and changing the naming would have no effect, but I think that is not super important.

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +63,217 @@ def __init__(
...
+            dlt_columns[C_DLT_ID] = sqlglot.exp.Alias(
+                this=func_expr,
+                alias=sqlglot.exp.to_identifier(NORM_C_DLT_ID),
+            )
+
+        # Replace if dlt columns exist in the select statement, otherwise append
+        for i, select in enumerate(outer_parsed_select.selects):
+            select_alias = select.alias.lower()
+            for column_name, alias_expr in dlt_columns.items():
+                if alias_expr is None:
```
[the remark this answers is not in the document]
anuunchin: yes, when the a dlt column has already been replaced in the select statement

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +63,217 @@ def __init__(
...
+                    continue
+                if select_alias == column_name:
+                    outer_parsed_select.selects[i] = alias_expr
+                    dlt_columns[column_name] = None  # Mark as replaced
```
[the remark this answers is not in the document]
anuunchin: but that will change dict size during iteration - should i simplify the logic here instead

tests/normalize/test_model_item_normalizer.py
```diff
@@ -0,0 +1,371 @@
...
+    [True, False],
+)
+@pytest.mark.parametrize("caps", MODEL_CAPS, indirect=True, ids=DESTINATIONS_SUPPORTING_MODEL)
+def test_selected_column_names_reordering(
+    caps: DestinationCapabilitiesContext,
+    columns: List[str],
+    add_dlt_columns: bool,
+    monkeypatch,
+) -> None:
+    monkeypatch.setenv("NORMALIZE__MODEL_NORMALIZER__ADD_DLT_LOAD_ID", str(add_dlt_columns))
```
sh-rp, line 285: just a note: We have a fixture `preserve_environ` that restores the env variables back to what they were before the tests run so that you do not n

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +65,267 @@ def __init__(
...
+        Args:
+            dialect (str): The SQL dialect for which the UUID expression needs to be generated.
+
+        Returns:
+            sqlglot.exp.Expression: A SQL expression that generates a UUID for the specified dialect.
+        """
+        # NOTE: redshift and sqlite don't have an in-built uuid function
+        if dialect == "redshift":
+            row_num = sqlglot.exp.Window(
+                this=sqlglot.exp.Anonymous(this="row_number"),
```
sh-rp, line 87: I have a question about this: Is there some possibility of this method producing duplicates? Is this row_number the row_number of the current insert, or of the whole table? In both cases, you can produce scenarios where 2 rows will end up with the same row_number. In case one it will happen directly on the next load if this row_number starts at 0 again, and in case two it will happen if rows are for any reason deleted and you load again. 

Can we add a random seed (maybe the load id) to the row number so that this does not happen?

sh-rp: For redshift there also is this method: https://stackoverflow.com/questions/46578889/how-to-generate-12-digit-unique-number-in-redshift. I'm not sure how feasible this is though...

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +65,267 @@ def __init__(
...
+
+        Args:
+            select_dialect (str): The SQL dialect to use for parsing and formatting.
+            parsed_select (sqlglot.exp.Select): The parsed SELECT statement.
+            columns (TTableSchemaColumns): The schema columns to match.
+
+        Returns:
+            Tuple[sqlglot.exp.Select, bool]: The outer SELECT statement and a flag indicating if reordering is needed.
+        """
+        if len(parsed_select.selects) == 1 and isinstance(
```
sh-rp, line 242: do we need this here? This case is already caught in __call__ no?

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +65,267 @@ def __init__(
...
+        """
+        Reorders or adjusts the SELECT statement to match the schema:
+        1. Adds missing columns as NULL.
+        2. Removes extra columns not in the schema.
+
+        Args:
+            outer_parsed_select (sqlglot.exp.Select): The parsed outer SELECT statement.
+            columns (TTableSchemaColumns): The schema columns to match.
+        """
+        if len(outer_parsed_select.selects) == 1 and isinstance(
```
sh-rp, line 199: This is not needed anymore, we do not process star selects.

dlt/destinations/job_client_impl.py
```diff
@@ -201,23 +202,14 @@ def _insert_statement_from_select_statement(
...
-                    alias=sqlglot.exp.to_identifier(alias_name),
-                )
-                top_level_select.expressions[i] = alias
-                columns.append(sqlglot.exp.to_identifier(alias_name))
-            elif isinstance(expr, sqlglot.exp.Alias):
-                alias_name = expr.alias
-                casefolded_alias = self._job_client.capabilities.casefold_identifier(alias_name)
-                expr.set("alias", casefolded_alias)
-                columns.append(sqlglot.exp.to_identifier(casefolded_alias))
+        for _, expr in enumerate(top_level_select.expressions):
```
sh-rp, line 208: do we need this here?

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +65,267 @@ def __init__(
...
+                file_obj=f,
+                fallback_dialect=self.config.destination_capabilities.sqlglot_dialect,  # caps are available at this point
+            )
+
+        parsed_select = sqlglot.parse_one(select_statement, read=select_dialect)
+
+        # The query is ensured to be a select statement upstream,
+        # but we double check here
+        if not isinstance(parsed_select, sqlglot.exp.Select):
+            raise ValueError("Only SELECT statements should reach the model normalizer.")
```
sh-rp, line 298: This error will never come if users to our transformations, but only if they manually yield models as you do in your tests. It would be better to have an error message that reflects this. Something like "The SqlModel you have supplied does not seem to be a SELECT statement." or something like that.

tests/load/pipeline/test_model_item_format.py
```diff
@@ -47,19 +55,106 @@
...
+    ids=[f"query-{q.strip().split()[0].lower()}" for q in UNSUPPORTED_MODEL_QUERIES],
+)
+def test_model_writer_with_non_select_query(unsupported_model_query: str, mocker) -> None:
+    try:
+        mocker_file = io.StringIO()
+        writer = ModelWriter(mocker_file)
+
+        mock_item = [MagicMock(dialect=None, query=unsupported_model_query)]
+        writer.write_data(mock_item)
+        pytest.fail(
```
sh-rp, line 136: testing exceptions can be done with the `pytest.raises` contextmanager.

tests/load/pipeline/test_model_item_format.py
```diff
@@ -155,16 +252,16 @@ def test_simple_model_jobs(destination_config: DestinationTestConfiguration) ->
     Test creating SQL model jobs for various scenarios:
     - Copying a table without a specific column with a row limit.
     - Reversing the column order in the output table with a row limit.
-    - Copying the entire table with a row limit.
     """
     # populate a table with two columns each with 10 items and retrieve dataset
     pipeline = destination_config.setup_pipeline(
-        f"test_model_item_format_{uniq_id()}", dev_mode=False
+        f"test_simple_model_jobs_{uniq_id()}", dev_mode=False
```
sh-rp, line 258: if you set dev_mode to true, you probably do not need the uniq_id()

tests/load/pipeline/test_model_item_format.py
```diff
@@ -47,19 +55,106 @@
...
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(
+        default_sql_configs=True,
+        subset=DESTINATIONS_SUPPORTING_MODEL,
+    ),
+    ids=lambda x: x.name,
+)
+def test_star_select(destination_config: DestinationTestConfiguration) -> None:
```
sh-rp, line 76: My suggestion would be to test these kind of things completely without a pipeline load as unit tests. If you only want to check wether star selects yield an error, you can simply take the modelitem normalizer, give it a sql string and see if it errors the right way. It is much much faster to test and you can test many kinds of queries to be sure everything works. Have a look at these lineage tests: https://github.com/dlt-hub/dlt/pull/2528/files#diff-2708d00deace203cf7e6a1256b7c3934d1985c0272c452fe07415d2a527aca88R58.

tests/load/pipeline/test_model_item_format.py
```diff
@@ -236,20 +323,137 @@ def copied_table() -> Any:
...
+    pipeline.run(
+        [{"a": i, "b": i + 10} for i in range(5)],
+        table_name="example_table_ab",
+        table_format=destination_config.run_kwargs["table_format"],
+    )
+
+    pipeline.run(
+        [{"a": i, "c": i + 20} for i in range(5)],
+        table_name="example_table_ac",
+        table_format=destination_config.run_kwargs["table_format"],
```
sh-rp, line 420: why not **run_kwargs?

tests/load/pipeline/test_model_item_format.py
```diff
@@ -236,20 +323,137 @@ def copied_table() -> Any:
...
+    assert df[casefold("b")].dropna().sum() == 59  # -1 + 11 + 12 + 13 + 14 + 15
+    assert df[casefold("c")].dropna().sum() == 109  # -1 + 21 + 22 + 23 + 24 + 25
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destination_configs,
+    ids=lambda x: x.name,
+)
+def test_model_from_joined_table(destination_config: DestinationTestConfiguration):
```
sh-rp, line 406: My idea for this test is to test the following:
* Create the two tables that you created
* Then run a transformation where you first copy `example_table_ab` into a new table, and then copy `example_table_ac` into the same table, without a join and without a prior pipeline run to somehow create the schema of this new table. I want to see if this works.

tests/load/pipeline/test_model_item_format.py
```diff
@@ -425,3 +633,381 @@ def example_table() -> Any:
...
+    load_id = load_info.loads_ids[0]
+    casefolder = pipeline.sql_client().capabilities.casefold_identifier
+    result_items = dataset[target_table_name].df()[casefolder(drop_column)].to_list()
+
+    if drop_column == "_dlt_load_id":
+        assert all(
+            item == load_id for item in result_items
+        ), f"All values should match _dlt_load_id={load_id}"
+    elif drop_column == "_dlt_id":
+        assert len(result_items) == len(set(result_items)), "Values in _dlt_id must be unique"
```
sh-rp, line 723: I don't think this tests the uniqueness of the dlt_id does it? Maybe it does.. :)

tests/normalize/test_model_item_normalizer.py
```diff
@@ -0,0 +1,371 @@
...
+    aliases = [select.alias for select in parsed_norm_select_query.selects]
+    assert aliases == [
+        caps.casefold_identifier(col) for col in ["a", "b", "c", "d", "_dlt_load_id", "_dlt_id"]
+    ]
+
+
+@pytest.mark.parametrize("caps", MODEL_CAPS, indirect=True, ids=DESTINATIONS_SUPPORTING_MODEL)
+def test_select_column_missing_in_schema(
+    caps: DestinationCapabilitiesContext, model_normalize: Normalize
+) -> None:
```
sh-rp, line 349: This is a cool test / behavior, but I wonder if this will ever happen that we have a select where columns are missing in the schema. Anyway, let's keep it!

dlt/normalize/items_normalizers.py
```diff
@@ -52,6 +65,267 @@ def __init__(
...
+        alias_map = {expr.alias.lower(): expr for expr in outer_parsed_select.selects}
+
+        # Build new selects list in correct schema column order
+        new_selects = []
+        for col in columns:
+            lower_col = col.lower()
+            expr = alias_map.get(lower_col)
+            if expr:
+                new_selects.append(expr)
+            # If there's no such column select, just put null
```
sh-rp, line 214: nice! there is one additional thing I think we need: If we put null for a column, we should check the schema to see if this column is `nullable`. If it is not, raise with a useful error and test this.

tests/normalize/test_model_item_normalizer.py
```diff
@@ -0,0 +1,371 @@
...
+    select_dialect, normalized_select_query, _ = extract_normalize_retrieve(
+        model_normalize, model, schema, "my_table", dialect
+    )
+
+    assert select_dialect == dialect
+
+    # Parse the normalized model query to extract the aliases of the outer select
+    parsed_norm_select_query = sqlglot.parse_one(normalized_select_query, read=dialect)
+    assert isinstance(parsed_norm_select_query, sqlglot.exp.Select)
+    aliases = [select.alias for select in parsed_norm_select_query.selects]
```
sh-rp, line 340: we should also test that the non existing aliases are set to null, no?

tests/normalize/test_model_item_normalizer.py
```diff
@@ -0,0 +1,371 @@
...
+def test_simple_model_normalizing(
+    caps: DestinationCapabilitiesContext, model_normalize: Normalize
+) -> None:
+    """
+    This test demonstrates how a model with simple sql query is transformed by the normalizer for each destination capabilities.
+    Explicit queries are used to clearly show the expected transformation process.
+    """
+    # TODO: tests for sqlalchemy dialects
+    dialect = caps.sqlglot_dialect
+    model = SqlModel.from_query_string(query="SELECT a, b FROM my_table", dialect=dialect)
```
sh-rp, line 125: why not use a query here where all the model normalizer features are used? It's just slightly more complex, but you could test:
* Adding of missing column which is set to null
* normalizing of at least one column name
* re-ordering columns to fit the provided schema.

wdyt?

model = SqlModel.from_query_string(query="SELECT b, A_a FROM my_table", dialect=dialect)
schema = create_schema_with_complete_columns("my_table", "text", ["a_a", "b", "c"])

expected (more or less):

f"SELECT _dlt_subquery.a_a AS a_a, _dlt_subquery.b AS b, NULL as c, '{load_id}' AS _dlt_load_id,"
 " UUID() AS _dlt_id FROM (SELECT a, b FROM my_table) AS _dlt_subquery\n"

tests/normalize/test_model_item_normalizer.py
```diff
@@ -0,0 +1,371 @@
+from typing import Iterator, List, Tuple
```
sh-rp, line 1: all tests in this file that do not actually access a destination should get the essential decorator. and at least one test that does a load should also get it, ideally a test that tests all normalizer features.

tests/normalize/test_model_item_normalizer.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+MODEL_CAPS = [get_caps(dest_name=dest) for dest in DESTINATIONS_SUPPORTING_MODEL]
+
+
+@pytest.fixture
+def caps(request) -> Iterator[DestinationCapabilitiesContext]:
+    _caps = request.param()
+    # If it's a destination with staging (Athena, Dremio)
+    # force the model in caps
```
sh-rp, line 40: I think you can remove this if you merge in devel, no?

sh-rp:
@anuunchin really nice, I think we are almost there! What I would additionally maybe consider, is to create a test setup, similar to the one where you test normalization for all dialects, that runs the normalizer for many different queries for the duckdb dialect to see if edge cases are covered. This test should not do any loads, but just get a schema and a query and some settings (put dlt columns or not) and you assert the normalized query. These will run super fast and you can test lots of different things. We should test that:

* common table expressions work: `WITH (query) SELECT ...`
* nested star selects should probably work (not sure): `SELECT name FROM (SELECT * FROM customer)`
* Star selects on joins should not work: `SELECT t1.*, t2.name FROM (t1 join t2)
* Test different kinds of re-ordering also when joining

tests/normalize/test_model_item_normalizer.py
anuunchin: 😮

dlt/destinations/job_client_impl.py
anuunchin: Did you mean the `_` and `enumeration` or generally the fact that we're extracting column names? I think we need the columns otherwise the insert will not have explicit columns - which is fine, but not ideal I think - or no 👀

tests/load/pipeline/test_model_item_format.py
anuunchin: adjusted where applicable, the reason was that the laoder file format for the iceberg athena test config was resolving to `parquet` and i was explicitly passing only table format 👀

tests/load/pipeline/test_model_item_format.py
anuunchin: but why? 👀  the set length would be different if there are duplicates in the list or am i tripping

tests/load/pipeline/test_model_item_format.py
anuunchin: owwww

dlt/normalize/items_normalizers.py
anuunchin: oww 🫡

tests/normalize/test_model_item_normalizer.py
anuunchin: 🫡

tests/normalize/test_model_item_normalizer.py
anuunchin: I understood that by `at least one test that does a load` you meant the tests in the `test_model_item_format` since all the tests in `test_model_item_normalizer` don't load. So I adjusted the `test_model_item_format.test_simple_model_jobs` to cover some additinal cases and marked it as essential along with the one that tests write dispositions as i think that one's important 👀

tests/normalize/test_model_item_normalizer.py
anuunchin: I'm assuming you meant
```
        _caps.update({"preferred_loader_file_format": "model"})
        _caps.update({"supported_loader_file_formats": ["model"]})
```
this part, The reason I have this is that i'm getting caps directly from DESTINATIONS_SUPPORTING_MODEL -> for dremio and athena, we need merge caps with filesystem -> I just adjusted this fixture so it's clearer

rudolfix (changes requested):
looks excellent! what we still need to do is to align this with other normalizers. full alignment was planned for later (#1317 ). what we need to do though:
1. we are able to generate only random `_dlt_id` this is almost inconsequential for models which have no nested links. use `get_root_row_id_type` to get the type of _dlt_id and if it is not random we must raise or skip _dlt_id to be created
2. btw. I'm not sure arrow normalizer does the above. I think not 🤯 we'll take care of that later
3. what if _dlt_id already exists in the model? normalizer should leave existing one. this is called "bring your own dlt_it" :)
4. IMO we should disable _dlt_id by default but enable _dlt_load_id. I plan to do that for arrow as well (this will be breaking change)

again! this is good

dlt/destinations/job_client_impl.py
```diff
@@ -201,23 +202,14 @@ def _insert_statement_from_select_statement(
...
+            if isinstance(expr, sqlglot.exp.Star):
```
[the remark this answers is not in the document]
sh-rp: I meant that we don't really need to check for star selects at all, because they should not pass the normalizer stage, but we can also keep it.

tests/load/pipeline/test_model_item_format.py
sh-rp: You are right, I thought the pd to_list would give a list of dictionaries, but it is giving a list of strings, to this works!

dlt/normalize/items_normalizers.py
```diff
@@ -88,8 +88,13 @@ def _uuid_expr_for_dialect(self, dialect: str) -> sqlglot.exp.Expression:
                 partition_by=None,
                 order=None,
             )
-            casted_row_num = sqlglot.exp.Cast(this=row_num, to=sqlglot.exp.DataType.build("TEXT"))
-            return sqlglot.exp.func("MD5", casted_row_num)
+            concat_expr = sqlglot.exp.func(
+                "CONCAT",
+                sqlglot.exp.Column(this=self._normalize_casefold(C_DLT_LOAD_ID)),
```
sh-rp: cool!

tests/normalize/test_model_item_normalizer.py
```diff
@@ -138,9 +138,19 @@ def test_star_select_rejection(
     with pytest.raises(
         NormalizeJobFailed,
         match=r"Model queries using a star \(`\*`\) expression cannot be normalized\.",
-    ) as py_exc:
+    ):
         _, _, _ = extract_normalize_retrieve(model_normalize, model, schema, "my_table", dialect)
 
+    # Ensure star selects with explicit columns are also rejected
+    another_model = SqlModel.from_query_string(query="SELECT *, NULL AS c FROM my_table")
```
sh-rp, line 145: ah ok :)

sh-rp (commented):
approved from my side with some notes

PR #2510, opened 2025-04-10
Fix clickhouse sql syntax for sentinel table creation

sh-rp:
### Description
This small PR aligns the sql syntax for creating the sentinel table on clickhouse with what is specified in the official docs (also chatgpt agrees...). For our clickhouse instance, both syntax works, I checked that on the console and verified that in both cases the correct column has the primary key constraint. We have a use reporting that the old syntax does not work, I'm assuming they are on a different clickhouse version, I can't find any mention of this changing in the clickhouse changelogs for the last 18 months though.


rudolfix (approved):
LGTM!

PR #2511, opened 2025-04-10
Preserve type of emtpy lists in incremental to allow to materialize empty resources

sh-rp:
### Description
See very nice bug report


sh-rp:
Note: We should check add_map, add_limit and add_filter too, also in async, I think they all might end up removing this info..

rudolfix (approved):
LGTM

Issue #2502, opened 2025-04-09
SQLGlot powered Schema discovery for ReadableRelations ("basic lineage")


sh-rp:
**TLDR**

For any class that implements `SupportsReadableRelation` we want to be able to compute the column schema if requested. Right now we already have a `column_schema` property that kind works in some cases when we just do a subselection of columns or a limit or something like that, but it does not work for more complex sql queries / ibis expressions. This is needed to be able to create and update tables for transformations before the transformations are executed.

Since we have the full input schema for each dataset, we should be able to compute the resulting columns_schema by running some form of lineage on the sql query and looking up the columns in the input schema. A simple version of this exists in the dlt+ repo, maybe this is the way, maybe some completely different way is better.

**Additional Reading**
* SQLGlot ast / lineage primer: https://github.com/tobymao/sqlglot/blob/main/posts/ast_primer.md
* Columsn discovery implementation in dlt+: https://github.com/dlt-hub/dlt-plus/pull/217/files#diff-b1f6ef92887216758374b380dc3aa203edec8a88cdc596f7edc1cb0b1e3a9c22

zilto:
released via transformation release

PR #2512, opened 2025-04-11
feat/2502 `sqlglot` lineage

zilto:
This follows #2502 [WIP]

This uses SQLGlot's [built-in type annotation and propagation](https://github.com/tobymao/sqlglot/blob/17f7eaff564790b1fe7faa414143accf362f550e/sqlglot/optimizer/annotate_types.py#L135). Essentially it does:
1. Get the dlt schema associated with the dataset
2. Map dlt types to destination types
3. Create a SQLGlot schema using the destination types
4. Parse the query to SQL AST
5. Quality and type-annotate the AST
6. Collect the columns specified in the final SELECT clause
7. Convert the SELECT columns SQLGlot types back to dlt types 

## TODOs
- support joins between two separate datasets, this will probably require catching incoming tables in the args in the proxy and checking wether they are from another dataset, if so, extend the sqlglot schema.
- Ensure case sensitive columns keep their case in lineage, it seems this is not the case at the moment. check  NormalizationStrategy in sqlglot
- Snowflake lineage does not work properly, right now we use duckdb instead, because for example on star selects, snowflake lineage does not seem to know what to do

### DIFF
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -96,11 +96,11 @@ jobs:
         run: poetry install --no-interaction --with sentry-sdk
 
       - run: |
-          poetry run pytest tests/common tests/normalize tests/reflection tests/plugins tests/load/test_dummy_client.py tests/extract/test_extract.py tests/extract/test_sources.py tests/pipeline/test_pipeline_state.py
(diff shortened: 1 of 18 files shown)


sh-rp:
closed in favor of continued work in https://github.com/dlt-hub/dlt/pull/2528

Issue #2483, opened 2025-04-04
Incremental Load with Lag throws TypeError: "can't compare offset-naive and offset-aware datetimes"


epDugas:
### dlt version

1.9.0

### Describe the problem

When specifying lag on an incremental load, the second run raises:
`TypeError: can't compare offset-naive and offset-aware datetimes`

Debugging shows that in method:
`_apply_lag_to_datetime`, when `value` is a Pendulum DateTime object with `tzinfo=datetime.timezone.utc`, the operation in the return statement yields a Pedulum DateTime with no timezone info, when `value - delta` is performed.

```python
return value - delta if last_value_func is max else value + delta
``` 

This value is returned from `_apply_lag_to_value` to function `apply_lag`, and the if statement in that function raises the ValueError when the `last_value_func` is called, because lagged_last_value has had timezone info removed.

```python
 if (
        initial_value is not None
        and last_value_func((initial_value, lagged_last_value)) == initial_value
    ):
```

Additional Info:
When the pipeline state is retrieved, the pendulum datetime objects have `tzinfo=datetime.timezone.utc`, instead of `TimeZone('UTC")`:

![Image](https://github.com/user-attachments/assets/d8a1e1d5-964a-411f-a9a0-938bad9fbac2)

StackTrace:
```
File ".../dlt/extract/incremental/lag.py", line 75, in apply_lag
    and last_value_func((initial_value, lagged_last_value)) == initial_value
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: can't compare offset-naive and offset-aware datetimes
```

### Expected behavior

Lag to work with UTC datetimes values.

### Steps to reproduce

Using shopify verified source, use `lag` parameter. Perform two loads, the second run should fail with the above details. 

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

dlt verified source - shopify

NOTE: Using databricks destination, not available in drop-down.

### Additional information

I am able to reproduce in a test, in the verified sources repo, in `test_shopify_source.py`.

epDugas:
I did some more debugging. Changing the following resolved the issue, not saying that is the right fix, as I am unaware of unintended consequences of that change, just thought it would be helpful information. If that change is acceptable, or if there is some other preferred method of resolving, let me know and I will create a PR. 

Location: `dlt.common.json.__init__.py`
Function:

```python
[...]

PR #2514, opened 2025-04-11
use pendulum.parse instead of fromisoformat when deserializing pua json

sh-rp:
### Description

We have a bug, where the pua_decoder creates a `datetime` object from a json string which looses timezone information on `deepcopy` (repro script attached). This means every time we modify the pipeline state, all timezone informations of all datetime entries are dropped which leads to problems in incrementals. This PR fixes this at least for when pendulum is installed, there is a new test that verifies this too.

pendulum.DateTime.fromisoformat was used before, pendulum.parse is used now. You can see that during a deepcopy the isoformat parsed datetime object will loose its timezone.

```
some_vals = {
    "nested": {
        "parse": pendulum.parse('2024-01-01T00:00:00+00:00'),
        "isoformat": pendulum.DateTime.fromisoformat('2024-01-01T00:00:00+00:00')
    }
}

print(type(some_vals["nested"]["parse"]))
print(type(some_vals["nested"]["isoformat"]))

print(some_vals)
print(deepcopy(some_vals))
```

### DIFF
--- a/dlt/common/json/__init__.py
+++ b/dlt/common/json/__init__.py
@@ -96,7 +96,8 @@ def _datetime_decoder(obj: str) -> datetime:
         # Backwards compatibility for data encoded with previous dlt version
         # fromisoformat does not support Z suffix (until py3.11)
         obj = obj[:-1] + "+00:00"
-    return pendulum.DateTime.fromisoformat(obj)
+    # TODO: should we check that this is in fact a datetime?
+    return pendulum.parse(obj)
 
 
 # define decoder for each prefix
--- a/dlt/common/json/__init__.py
+++ b/dlt/common/json/__init__.py
@@ -91,13 +91,16 @@ _TIME = chr(PUA_START + 7)
 PUA_START_UTF8_MAGIC = _DECIMAL.encode("utf-8")[:2]
 
 
-def _datetime_decoder(obj: str) -> datetime:
+def _datetime_decoder(obj: str) -> pendulum.DateTime:
     if obj.endswith("Z"):
         # Backwards compatibility for data encoded with previous dlt version
         # fromisoformat does not support Z suffix (until py3.11)
         obj = obj[:-1] + "+00:00"
-    # TODO: should we check that this is in fact a datetime?
-    return pendulum.parse(obj)
+    # tz=None sets no timezone if if it not specified on string
+    dt = pendulum.parse(obj, tz=None)
+    if not isinstance(dt, pendulum.DateTime):
+        raise ValueError(f"Expected pendulum.DateTime, got {type(dt)}")
+    return dt
 
 
 # define decoder for each prefix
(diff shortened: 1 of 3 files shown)


rudolfix (changes requested):
heh very interesting. my problem with pendulum.parse is that it also parses string "now" as now() 🤯 and it should use this iso function internally. I bet this has something to do with rust speedup being used.

maybe we should try to parse using Python datetime?

epDugas:
> This looks like a bug in pendulum and actually we also get the same problem outside of our code coming from inside arrow if we use these pendulum objects, in one of the tickets the repro for this is as follows:
> 
> ```
> assert apply_lag(
>     1, pendulum.DateTime.fromisoformat("2023-03-02T01:15:29+00:00"), pendulum.DateTime.fromisoformat("2023-03-02T01:15:30+00:00"), max
> ) == pendulum.DateTime.fromisoformat("2023-03-02T01:15:29+00:00")
> ```
The issue (I believe), is that calling `fromisoformat` on `pendulum.DateTime` calls it's base class' implementation of the `fromisoformat` function. The base class `datetime` returns: `cls(...)`, which returns a type of `pendulum.DateTime`, but without the appropriate pendulum timezone properties set (it only contains `datetime.tzinfo`, not the pedulum `timezone_name` for example). I would suspect that is not a pendulum bug, since it did not implement the `fromisoformat` function of it's base class, and it is just an artifact of the late binding nature of the call to `cls(...)` from the `datetime.fromisoformat` function? 

[reproduce exception in apply_lag](https://github.com/dlt-hub/dlt/issues/2483#issuecomment-2795746228)

sh-rp:
@epDugas to me this is a bug if pendulum subclasses another class and re-implements the deepcopy in such a way that the behavior of the superclass is broken. But that is probably hairsplitting ;)

Additional notes: If we decide to keep pendulum, we not only need to fix our parsing code but also all docs examples where the "wrong" way of parsing pendulum dates is used, because this problem will surface outside of dlt code too, as the apply_lag example shows.

rudolfix:
Looks like there were attempts to fix that in pendulum:
```py
def __add__(self, other: datetime.timedelta) -> Self:
        if not isinstance(other, datetime.timedelta):
            return NotImplemented

        caller = traceback.extract_stack(limit=2)[0].name
        if caller == "astimezone":
            return super().__add__(other)

        return self._add_timedelta_(other)
```
`astimezone` is one of the functions that destroy timezone info. (and it still does, even with this fix). also this destroys tzinfo
```py
pendulum.DateTime(2025, 4, 27, tzinfo=timezone.utc) + timedelta(1)
```
maybe pendulum does not understand what `timezone.utc` is? anyway this does not look like as an easy fix and the number of such differences will only increase

@sh-rp my take is to fix the thing above like you do and we'll have to remove pendulum ASAP from our codebase and docs. we can keep some semi-public function in the library and we'll need to convert pendulum into regular datetimes, too many people is using it with our lib

rudolfix (approved):

PR #2525, opened 2025-04-15
Added info about dlt's internal tables

dat-a-man:
### Description

Updated section reference destination tables. Added info about dlt's internal tables


AstrakhantsevaAA (approved):
Very good Aman!

Issue #2493, opened 2025-04-08
extract and normalize Model data item


rudolfix:
**Background**
#2404 adds new data item (Model, currently SQL only). This data item must be properly extracted and normalized. Conceptually a Model must go through exactly the same treatment as Arrow table (or other tabular data).
From this perspective, it is just a table but not yet materialized.
- _dlt_load_id must be added to the model per normalizer settings. model needs to be rewritten. canonical place is to do it in normalizer.
- models may evolve and new columns are added, also apply data contracts there
- normalize table and column names

**Tasks**
1. * [ ] fix the arrow normalizer: `normalize_py_arrow_item` will not replace `_dlt_load_id` column if it exists. if requested, we must replace the existing column with right load id. this is actually a bug. python dict normalizer overwrites load ids and this is the desired behavior
2. * [ ] write a function that does the same as `normalize_py_arrow_item` but to a Model (via sqlglot)
- selected column names will be normalized according to `naming`
 - elected column will be reordered according to `columns` (maybe we do not need to do that: model insert statement will direct data to physical table columns, parquet insert does not do that on some backend)
 - empty columns will be inserted if they are missing, types will be generated using `caps`
 - add `_dlt_load_id` column if it is missing and `load_id` is provided
3. * [ ] We do not need to normalize in extract phase. we do that for arrow to avoid rewrites. here we can just implement it in `normalize`. Implement `ModelItemsNormalizer` that does the same as `ArrowItemsNormalizer` and wire it into existing normalize worker logic
- schema evolution: to infer output schema we need a model + input schema. this will be supplied by "lineage" and passed as HintMeta to the extractor. extractor will update schema and apply data contracts. **so schema evolution is done**
- in normalizer we must normalize identifiers in the models AND verify if all coluns in the model are in schema (to verify lineage module)
- implementation hint: wrap model into a SELECT statement that will convert identifiers to the output schema naming convention. **as mentioned we cannot change names in the original model (yielded by the user)**
- 4. * [ ] Make the `json normalizer` respect the `add_dlt_load_id` and `add_dlt_id` settings.

**Tests**
1. * [ ] add to data contract tests, data contracts must be applied no matter the item type
2. * [ ] write test for a model with all possible data types (maybe we can do it using the same fixture as other data items)

sh-rp:
@anuunchin for reference, the code for normalizing arrow tables is in def `normalize_py_arrow_item(` in `dlt/common/libs/pyarrow.py`

PR #2526, opened 2025-04-15
normalize_py_arrow_item now replaces load id column with the right one

anuunchin:
### Description
This PR addresses the first task in #2493, which says:

> fix the arrow normalizer: `normalize_py_arrow_item` will not replace _dlt_load_id column if it exists. if requested, we must replace the existing column with right load id. this is actually a bug. python dict normalizer overwrites load ids and this is the desired behavior

Now, the arrow normalizer replaces the load id even if the column exists. An additional test is added to `test_arrow_sources.py`.

### Related Issues

- Relates to #2493

### DIFF
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -394,7 +394,7 @@ def should_normalize_arrow_schema(
     columns: TTableSchemaColumns,
     naming: NamingConvention,
     add_load_id: bool = False,
-) -> Tuple[bool, Mapping[str, str], Dict[str, str], Dict[str, bool], bool, TTableSchemaColumns]:
+) -> Tuple[bool, Mapping[str, str], Dict[str, str], Dict[str, bool], TTableSchemaColumns]:
     """Figure out if any of the normalization steps must be executed. This prevents
     from rewriting arrow tables when no changes are needed. Refer to `normalize_py_arrow_item`
     for a list of normalizations. Note that `column` must be already normalized.
@@ -415,16 +415,6 @@ def should_normalize_arrow_schema(
     dlt_id_col = naming.normalize_identifier(C_DLT_ID)
     dlt_columns = {dlt_load_id_col, dlt_id_col}
 
-    # Do we need to add a load id column?
-    if add_load_id and dlt_load_id_col in columns:
-        try:
-            schema.field(dlt_load_id_col)
-            needs_load_id = False
-        except KeyError:
-            needs_load_id = True
-    else:
-        needs_load_id = False
-
     # remove all columns that are dlt columns but are not present in arrow schema. we do not want to add such columns
     # that should happen in the normalizer
     columns = {
@@ -433,18 +423,27 @@ def should_normalize_arrow_schema(
         if name not in dlt_columns or name in rev_mapping
     }
 
+    try:
+        has_dlt_load_id_field = schema.field(dlt_load_id_col) is not None
+    except KeyError:
+        has_dlt_load_id_field = False
+
     # check if nothing to rename
     skip_normalize = (
         (list(rename_mapping.keys()) == list(rename_mapping.values()) == list(columns.keys()))
         and not nullable_updates
-        and not needs_load_id
+        and (
+            (not add_load_id)
+            # ... OR we *do* need to add it, but we already have it in the schema,
+            # yet it is not listed among the new columns to add.
+            or (add_load_id and dlt_load_id_col not in columns and has_dlt_load_id_field)
+        )
     )
     return (
         not skip_normalize,
         rename_mapping,
         rev_mapping,
         nullable_updates,
-        needs_load_id,
         columns,
     )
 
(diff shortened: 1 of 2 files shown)


tests/pipeline/test_arrow_sources.py
```diff
@@ -610,3 +610,45 @@ def some_data():
 
         # Normalized file has _dlt_load_id
         assert pa.compute.all(pa.compute.equal(tbl["_dlt_load_id"], load_id)).as_py()
+
+
+def test_replace_existing_dlt_load_id():
```
sh-rp, line 615: Another test idea: you could create a matrix with 2 dimensons: input data has this column or not on one, and setting is set to true vs false on the other, then you basically get 4 tests for free and you may be even able to remove the test above, or fuse the two.

sh-rp (approved):
LGTM!

dlt/common/libs/pyarrow.py
```diff
@@ -466,55 +455,66 @@ def normalize_py_arrow_item(
     5. Add `_dlt_load_id` column if it is missing and `load_id` is provided
     """
     schema = item.schema
-    should_normalize, rename_mapping, rev_mapping, nullable_updates, needs_load_id, columns = (
+    should_normalize, rename_mapping, rev_mapping, nullable_updates, columns = (
         should_normalize_arrow_schema(schema, columns, naming, load_id is not None)
     )
-    if not should_normalize:
+    if not should_normalize and not load_id:
```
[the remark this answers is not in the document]
rudolfix: another reason why it makes sense: we may handle `_dlt_id` the same way. in fact it is already handled in the normalize step (we add it when streaming data so we do not overload the memory)

dlt/common/libs/pyarrow.py
```diff
@@ -466,55 +455,66 @@ def normalize_py_arrow_item(
...
-            pyarrow.field(
-                naming.normalize_identifier(C_DLT_LOAD_ID),
-                load_id_type,
-                nullable=False,
-            )
+        dlt_load_id_field = pyarrow.field(
+            naming.normalize_identifier(C_DLT_LOAD_ID), load_id_type, nullable=False
         )
-        new_columns.append(pyarrow.array([load_id] * item.num_rows, type=load_id_type))
+        dlt_load_id_column = pyarrow.array([load_id] * item.num_rows, type=load_id_type)
```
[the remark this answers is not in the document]
rudolfix: actually we can do it right now. it is easy. and I see we have converting code in delta and pyiceberg that works on chunks. so it is way better if we do it. code below works:
```py
import pyarrow as pa
import numpy as np

# Step 1: Create the dictionary value array.
# We define a dictionary that contains just one unique string.
dictionary = pa.array(["the_string"])

# Step 2: Create an indices array with one million entries.
# All indices are zero, pointing to the only entry in our dictionary.
indices = pa.array(np.zeros(1_000_000, dtype='int32'))

# Step 3: Construct the dictionary array using the indices and dictionary.
dict_array = pa.DictionaryArray.from_arrays(indices, dictionary)

# (Optional) Create another column, for example, an integer column.
# This column simply holds the sequence 0, 1, 2, ..., 999999.
other_column = pa.array(np.arange(1_000_000))

# Step 4: Create a PyArrow table with the dictionary column and the other column.
table = pa.Table.from_arrays([dict_array, other_column], names=['dict_col', 'int_col'])

# Display the table schema and the first few rows.
print("Table Schema:")
print(table.schema)

print("\nFirst 5 rows:")
print(table.slice(0, 5))

```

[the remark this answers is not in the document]
rudolfix: OFC LLM used `int32` to create in memory index 🤯 use `int8` (there are no bit flags in arrow)

dlt/common/libs/pyarrow.py
```diff
@@ -641,10 +613,20 @@ def add_constant_column(
         value: The value to fill the new column with
         index: The index at which to insert the new column. Defaults to -1 (append)
     """
-    field = pyarrow.field(name, pyarrow.dictionary(pyarrow.int8(), data_type), nullable=nullable)
+    try:
+        from dlt.common.libs.numpy import numpy as np
+    except MissingDependencyException:
```
rudolfix, line 618: why do you re-raise the same exception but with different message? if you want to be more specific please write exactly why we need numpy :) also see below

rudolfix (changes requested):
we still need some refactoring. the test is very good! (but IDK how it is passing with the current code, maybe I'm missing something)

dlt/common/libs/pyarrow.py
anuunchin: oww, haha, the double exception is indeed weird ...

dlt/extract/extractors.py
```diff
@@ -398,6 +398,42 @@ def _apply_contract_filters(
...
+            dlt_load_id_col_name = self.naming.normalize_identifier(C_DLT_LOAD_ID)
+
+            idx = item.schema.get_field_index(dlt_load_id_col_name)
+            # if the column already exists and is in columns, get rid of it
+            if idx != -1 and dlt_load_id_col_name in columns:
+                item = pyarrow.remove_columns(item, dlt_load_id_col_name)
+
+            # get pyarrow.string() type
+            pyarrow_string = pyarrow.get_py_arrow_datatype(
+                columns[dlt_load_id_col_name],
```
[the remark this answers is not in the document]
anuunchin: I found out the dlt load id column was always in `column`, because we update the schema with _copute_and_update_tables before writing item. I removed the load id logic from there

dlt/common/libs/pyarrow.py
```diff
@@ -530,22 +514,57 @@ def normalize_py_arrow_item(
...
     # create desired type
     return item.__class__.from_arrays(new_columns, schema=pyarrow.schema(new_fields))
 
 
+def add_dlt_load_id_column(
+    item: TAnyArrowItem,
+    columns: TTableSchemaColumns,
+    caps: DestinationCapabilitiesContext,
+    naming: NamingConvention,
+    add_load_id: bool,
```
rudolfix, line 526: why do you need this? this just skips the whole code inside. so just don't call this function instead

dlt/common/libs/pyarrow.py
```diff
@@ -530,22 +514,57 @@ def normalize_py_arrow_item(
...
+            "UTC",  # ts is irrelevant to get pyarrow string, but it's required...
+        )
+
+        # add the column with the new value at previous index or append
+        item = add_constant_column(
+            item=item,
+            name=dlt_load_id_col_name,
+            data_type=pyarrow_string,
+            value=replacement_load_id,
+            nullable=False,
```
rudolfix, line 561: take nullability from column schema

dlt/common/libs/pyarrow.py
```diff
@@ -669,11 +688,24 @@ def add_constant_column(
...
         value: The value to fill the new column with
         index: The index at which to insert the new column. Defaults to -1 (append)
+    Note:
+        This function creates a dictionary field for the new column, which is memory-efficient
+        when the column contains a single repeated value.
+        The column is created as a DictionaryArray with int8 indices.
     """
-    field = pyarrow.field(name, pyarrow.dictionary(pyarrow.int8(), data_type), nullable=nullable)
+    dictionary = pyarrow.array([value], type=data_type)
+    zero_buffer = pyarrow.py_buffer(b"\x00" * item.num_rows)
```
rudolfix, line 697: this is allocating bytes on Python heap and this array can be garbage collected. There's a high chance that arrow will create a duplicate buffer out of that. So we still are not 0 copy and our exercise here is to reduce memory usage.

Arrow does not have a python interface to operate on mem buffers. Because they are C and those are just part of this language. You can try this:
```
import pyarrow as pa, ctypes
buf = pa.allocate_buffer(size, resizable=False)
ctypes.memset(buf.address, 0, size)   # one pass, no extra allocation
```
unfortunatelly `allocate_buffer` does not guarantee zeroed memory so we need to memset. we can't get better than this

rudolfix (changes requested):

rudolfix (approved):
LGTM!

PR #2535, opened 2025-04-19
Added section using xmin for Change Data Capture (CDC) to pg_replication docs

### LINKED ISSUE #2376: support postgres xmin replication
### Feature description

We have a number of postgres sources that cannot be configured for CDC and many tables are large and don't have a suitable incremental/cursor column (ie. last_changed).

Am currently using airbyte for most our streams as they support cdc, xmin and user supplied cursor column. It would be nice if the postgres source supported all three as well (if I understand correctly it only supports cdc, you have to use sql_database for user supplied and xmin isn't supported).

I think I can use `query_adapter_callback` to implement xmin (I know this can wrap but it's a work-around until I can get last_changed implemented on our source tables) but it would be nice if the postgres source supported all three (or I've misread the documentation, in which case maybe it's not as clear as it could be?)

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

I'm currently using a mix of airbyte and dlt as neither cover all my use cases.

### Proposed solution

_No response_

### Related issues

_No response_

dat-a-man:
### Description

Added section using xmin for Change Data Capture (CDC) to pg_replication docs

### Related Issues

- Fixes #2376 
- Closes #2376
- Resolves #2376

### DIFF
--- a/docs/website/docs/dlt-ecosystem/verified-sources/pg_replication.md
+++ b/docs/website/docs/dlt-ecosystem/verified-sources/pg_replication.md
@@ -275,3 +275,27 @@ If you wish to create your own pipelines, you can leverage source and resource m
     
    Similarly, to replicate changes from selected columns, you can use the `table_names` and `include_columns` arguments in the `replication_resource` function.
 
+## Optional: Using `xmin` for Change Data Capture (CDC)
+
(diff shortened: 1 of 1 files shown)


AstrakhantsevaAA (approved):
Great!

PR #2556, opened 2025-04-24
Enabling 'model' loader_file_format for athena, synapse and dremio

### LINKED ISSUE #2554: Enable 'model' loader_file_format for athena, synapse and dremio
### TLDR

Currently these three destinations do not support the `model` `loader_file_format` for various reasons. There is a good chance we can get them to run, this needs to be investigated.

### Synapse
Was not tested at the time of implementation, so I think we skipped it. Probably easy to enable?

### Athena
Model files should work with iceberg tables unless there is some other component missing. This is a destination where 'model' would then be the only supported loader_file_format because loading otherwise requires staging, so we'll have to see that the model files actually get run on the correct destination. And in the tests we will have to set the `table_format` to `iceberg` for the decorated transformation functions.

### Dremio
Should work, same situation as above where this would be the only loader_file_format. Also needs investigation.

anuunchin:
### Description

This PR adds `model` loader file format support for Athena. (Dremio and Synapse  support is under construction).
For Athena, what's currently happening is that the load step skips staging if model format is present.

### Related Issues

- Closes #2554


dlt/pipeline/pipeline.py
```diff
@@ -577,15 +578,39 @@ def load(
...
+        else:
+            load_step = Load(
+                self._destination,
+                staging_destination=self._staging,
+                collector=self.collector,
+                is_storage_owner=False,
+                config=load_config,
+                initial_client_config=client.config,
+                initial_staging_client_config=staging_client.config if staging_client else None,
+            )
```
anuunchin, line 613: @sh-rp I don't think this is the correct location to be doing this (if correct logic at all), maybe it should happen inside the load step? 👀

tests/load/pipeline/test_model_item_format.py
```diff
@@ -226,11 +246,19 @@ def copied_table() -> Any:
...
+            "copied_table_no_b": {"model": 1},
+            "reversed_table": {"model": 1},
+            "copied_table": {"model": 1},
+        }
+    else:
+        assert count_job_types(pipeline) == {
+            "copied_table_no_b": {"model": 1, "sql": 1},
+            "reversed_table": {"model": 1, "sql": 1},
+            "copied_table": {"model": 1, "sql": 1},
+        }
```
anuunchin, line 261: @sh-rp  This also needs to be addressed I think - for uniformity: athena should also have only one model job (but I'm not 100% sure how this will affect athena in general)

dlt/common/data_writers/writers.py
```diff
@@ -783,6 +783,13 @@ def resolve_best_writer_spec(
             except DataWriterNotFound:
                 pass
 
+    # as a last resort, if the item_format is "model"
+    if item_format == "model":
+        try:
+            return DataWriter.class_factory(item_format, item_format, ALL_WRITERS).writer_spec()
+        except DataWriterNotFound:
+            pass
+
```
anuunchin, line 792: @sh-rp  also don't think it's correct to be forcing modelwriter here. What happens is that this func receives parquet as the only possible_file_format and the preferred format. The logic should be handled upstream so that this func receives correc arguments 🤔

tests/load/pipeline/test_model_item_format.py
```diff
@@ -43,17 +45,23 @@
...
+    )
     dataset = pipeline.dataset()
 
     select_dialect = pipeline.destination.capabilities().sqlglot_dialect
 
     example_table_columns = dataset.schema.tables["example_table"]["columns"]
 
     # TODO: incremental is not supported for models yet
-    @dlt.resource()
+    @dlt.resource(table_format="iceberg")
```
sh-rp, line 64: I'd only set this for athena, or alternatively just run the athena version with iceberg enabled for all tables, you can filter out the other one. It just should not be set for all other destination tests.

sh-rp:
Additional notes (WIP):
* We need to set at least one of the model loader tests as essential (search for the keyword in the tests). Tests marked with this marker are run on every CI execution. Select the test that combines the test of the most features into one.

sh-rp (changes requested):
This is a good first attempt! I did some digging in the code, and the main problem is, that at the moment the whole job type resolution and loader system assumes, that all jobs always go to either the staging or the final destination and are not mixed up. It will be a bigger task to refactor this and make sure that it does not break anything, so for now we will do this solution:
* Remove your changes to pipeline.py and writers.py
* In athena, add the model to the supported staging formats list and remove them from the main list
* In the generic caps add the model to the list of supported formats dlt/common/destination/capabilities.py
* In the loader where it is decided which job gets executed on which destination, route model jobs to the main destination: in `def is_staging_destination_job` return false if type is model

Then it will work.

Unwanted Sideeffects for this solution are:
* You need a staging destination configured to run model jobs
* The capabilities are not quite correct now since it looks like the models are executed on the staging destination
* You will be able to send models to a filesystem destination, which will just upload the model file there.

You can create a new ticket for us to fix this, part of this ticket could also be to support user side reference jobs. The reference jobs are special jobs that are created by dlt to insert data from staging destinations into the final destination, some people might want to do this manually so we could allow it.

tests/load/pipeline/test_model_item_format.py
```diff
@@ -270,15 +288,15 @@ def test_write_dispositions(
     example_table_columns = dataset.schema.tables["example_table_1"]["columns"]
     # In Databricks, Ibis adds a helper column to emulate offset, causing a schema mismatch
     # when the query attempts to insert it. We explicitly select only the expected columns.
-    relation = (
-        dataset["example_table_2"].order_by("a").limit(7, offset=3)[example_table_columns.keys()]
-    )
+    relation = dataset["example_table_2"].order_by("a").limit(7)[example_table_columns.keys()]
```
anuunchin, line 291: @sh-rp  NOTE: I removed offset from this relation due to the issue with synapse (I think we should detect offsets in queries for synapse and somehow implement rewrites like the code in job_client_imp that - but I'm not sure if it should be prioritized 👀)

sh-rp: You could replace the offset by filter(a>3) for this test and keep the rest of the assertions the same, but that is up to you :)

sh-rp (changes requested):
Very cool, thanks :) You can just remove this example sql code and maybe put this into a new ticket with a note to fix limit on synapse. I would not say that this has high priority, we should probably put a note in the docs somewhere. Also you can rebase on devel (or merge in) the tests should run faster.

sh-rp (approved):
nice :)

PR #2558, opened 2025-04-25
make it easier to customize range paginator, and support sending via JSON body

diwu-sf:
I have an endpoint that uses `PagePaginator` but it's a `POST` API and I need to inject the value `{ 'page': page_num }` into the JSON payload.

The way I implement this is make a `CustomPageNumberPaginator` and override a single `update_request()`:
```py
def update_request(self, request: Request):
  super().update_request(request)
  request.json[self.param_name] = self.current_value
```

This works fine if I fix some of the redundant code in DLT's `RangePaginator` to get rid of duplications in `init_request` and `update_request`

For some endpoints that need POST, I also need the ability to send data via the JSON payload instead of query params, similar to this: https://dlthub.com/docs/dlt-ecosystem/verified-sources/rest_api/basic#via-json-body

I have a `POST .../search` endpoint that wants the page number to be in the payload, with something that looks like:
```
{
  query: ...,
  page: ...,
}
```

### DIFF
--- a/dlt/sources/helpers/rest_client/paginators.py
+++ b/dlt/sources/helpers/rest_client/paginators.py
@@ -140,10 +140,7 @@ class RangePaginator(BasePaginator):
     def init_request(self, request: Request) -> None:
         self._has_next_page = True
         self.current_value = self.initial_value
-        if request.params is None:
-            request.params = {}
-
-        request.params[self.param_name] = self.current_value
+        self.update_request(request)
 
     def update_state(self, response: Response, data: Optional[List[Any]] = None) -> None:
         if self._stop_after_this_page(data):
@@ -389,10 +386,6 @@ class OffsetPaginator(RangePaginator):
         )
         self.limit_param = limit_param
         self.limit = limit
-
-    def init_request(self, request: Request) -> None:
-        super().init_request(request)
-        request.params[self.limit_param] = self.limit
 
     def update_request(self, request: Request) -> None:
         super().update_request(request)
--- a/dlt/sources/helpers/rest_client/paginators.py
+++ b/dlt/sources/helpers/rest_client/paginators.py
@@ -98,6 +98,7 @@ class RangePaginator(BasePaginator):
         total_path: Optional[jsonpath.TJsonPath] = None,
         error_message_items: str = "items",
         stop_after_empty_page: Optional[bool] = True,
+        via_json_body: bool = False,
     ):
         """
         Args:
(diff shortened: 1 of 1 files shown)


diwu-sf:
@burnash hi, what do you think about adding this support to the range paginators?
This kind of thing would be useful for POST rest endpoints (and potentially for graphQL queries as well)

burnash:
Hey @diwu-sf. Thanks for the PR and sorry for the late reply. I can see the need for JSON body pagination with POST requests.
How often do you run into APIs that need page number pagination params in the JSON body? I'm trying to figure out if this should be built into the core library or if a simple custom paginator makes more sense. Could you share the docs / reference for the API you're trying to query?

Also FYI - JSONResponseCursorPaginator already does JSON body pagination with cursor_body_path, though it's meant for cursors not page numbers: but it could work if you get the next page number as a part of your response.

diwu-sf:
This is the API that has the pagination in the body:
https://api.govspend.com/docs/?url=/docs#/
`POST /runSavedSearch`

I did have to make a custom paginator for now to make this work, which is basically a copy paste of the normal range paginator, just setting the page number into the JSON body instead. There was a lot of exact duplication of logic, hence I wanted to see if we can just support the toggle in the base implementation to put the pagination into the body

burnash:
Thanks for the detailed explanation and the API reference, @diwu-sf. You've convinced me that we need to support POST for range paginators. 

However, I'd like to make the implementation consistent with how `JSONResponseCursorPaginator` already handles this. Currently, it uses `cursor_body_path` parameter to specify where in the JSON body to place the cursor value, which is more flexible than a boolean flag suggested in this PR.

Here's what I think we should do:
1. Follow the existing pattern and instead of `via_json_body` use something like `param_body_path` in RangePaginator and respective params in PageNumberPaginator and OffsetPaginator.
2. This new parameter should support nested paths just like cursor paginator supports "cursor.next", e.g. we should support nested paths like "meta.pagination.page".
3. When param_body_path is None, use query parameters (existing behavior)

What do you think about this approach? Would you be interested in updating the PR to follow this pattern?

burnash:
Hey @diwu-sf, just following up to see if you'd like to continue working on this PR or if you'd prefer that I take it from here. Thanks again for your effort!

burnash:
I'm going to close this PR for now as the proposed solution doesn't address nested paths for the placement of params in th JSON body, and it's also not consistent with how `JSONResponseCursorPaginator` already handles similar functionality. 

Thanks again @diwu-sf  for identifying this and proposing a solution. If you'd like to continue working on this with the revised approach, please feel free to open a new PR. Otherwise, I may pick this up in the future when I have bandwidth.

PR #2566, opened 2025-04-28
Simplify pipeline test utils

sh-rp:
### Description
We have some cryptic code in our pipeline load test utils. We needed this to abstract getting row counts and table contents across different destination types, but now we have the nice dataset which also supports filessystem, delta and iceberg, so we can probably just use that, this PR will attempt to remove a bunch of cold code.


tests/load/utils.py
```diff
@@ -606,7 +598,7 @@ def destinations_configs(
                     bucket_url=bucket,
                     extra_info=bucket,
                     supports_merge=False,
-                    file_format="parquet",
+                    file_format="jsonl",  # keep jsonl as default, test utils are setup for this
```
sh-rp, line 601: Note for review: I changed the default format to jsonl so we save more core in the test utils for the sftp special case, I didn't see any downside to this.

tests/load/pipeline/test_open_table_pipeline.py
```diff
@@ -505,7 +506,8 @@ def two_part_empty():
         assert load_table_counts(pipeline, "users")["users"] == 0
         assert destination_config.destination_type == "filesystem"
     except IcebergViewException:
-        assert destination_config.destination_type != "filesystem"
+        pass
```
djudjuu, line 509: same?

dlt/helpers/ibis.py
```diff
@@ -109,11 +109,14 @@ def _ignore_hstore(conn: Any, name: Any) -> Any:
         con = ibis.snowflake.connect(
             schema=dataset_name, **sn_credentials, create_object_udfs=False
         )
-    elif issubclass(destination.spec, MsSqlClientConfiguration):
+    elif issubclass(destination.spec, MsSqlClientConfiguration) and not issubclass(
```
djudjuu, line 112: note: not sure whats going on here, pls verify yourself (just to say I wont invest time to understand this now. hard to say why I would expect this to be a correct change without more context.)

dlt/helpers/ibis.py
```diff
@@ -109,11 +109,14 @@ def _ignore_hstore(conn: Any, name: Any) -> Any:
...
+        destination.spec, SynapseClientConfiguration
+    ):
         from dlt.destinations.impl.mssql.mssql import MsSqlJobClient
 
         assert isinstance(client, MsSqlJobClient)
         ms_credentials = client.config.credentials.to_native_representation()
+        ms_credentials = ms_credentials.replace("synapse://", "mssql://")
```
djudjuu, line 119: same here? seems like you're replacing any unexpected credential prefixes that would still be thre

dlt/helpers/ibis.py
```diff
@@ -139,14 +142,14 @@ def _ignore_hstore(conn: Any, name: Any) -> Any:
...
         )
-    elif issubclass(destination.spec, DatabricksClientConfiguration):
-        from dlt.destinations.impl.databricks.databricks import DatabricksClient
-
-        bricks_client = cast(DatabricksClient, client)
-        con = ibis.databricks.connect(
-            **bricks_client.config.credentials.to_connector_params(),
-            schema=bricks_client.sql_client.dataset_name,
-        )
+    # elif issubclass(destination.spec, DatabricksClientConfiguration):
```
djudjuu, line 145: commented out on purpose?

tests/pipeline/utils.py
```diff
@@ -3,7 +3,8 @@
 import random
 from os import environ
 import io
-
+import os
```
djudjuu, line 6: this is indeed a file that deserve the name `utils`.
It also twists my brain to look at it. I wouldn't count on me detecting any bugs in your changes here

tests/load/lancedb/utils.py
```diff
@@ -28,6 +28,7 @@ def assert_unordered_dicts_equal(
     assert dict_set1 == dict_set2, "Lists contain different dictionaries"
 
 
+# TODO: merge with assert_table in main pipeline utils...
```
djudjuu, line 31: question: leaving this for another PR?

tests/load/utils.py
```diff
@@ -573,14 +573,6 @@ def destinations_configs(
 
     # add local filesystem destinations if requested
     if local_filesystem_configs:
-        destination_configs += [
-            DestinationTestConfiguration(
-                destination_type="filesystem",
-                bucket_url=FILE_BUCKET,
-                file_format="insert_values",
```
djudjuu, line 580: why remove this?

djudjuu (approved):
very good improvement: with the helper functions a lot of places become much more readable!
all the methods in `tests/pipeline/utils.py` seem very handy and it's nice to know they exist. I couldnt go over all of them to confirm you didnt mess it up, but if they produce the same results as before (in the tests) then that's a good indicator I guess.

i left a few comments in places that looked like maybe you forgot some code there or commented something out, just to make sure you intend to merge this.

otherwise, its approved from me. If we want to be extra sure, someone else needs to also look at `tests/pipeline/utils.py`

tests/pipeline/utils.py
```diff
@@ -308,148 +237,112 @@ def assert_only_table_columns(
...
-            [
-                f"SELECT '{name}' as name, COUNT(1) as c FROM {q_name}"
-                for name, q_name in zip(table_names, qualified_names)
-            ]
-        )
-        with c.execute_query(query) as cur:
-            rows = list(cur.fetchall())
-            return {r[0]: r[1] for r in rows}
+def load_table_counts(p: dlt.Pipeline, *table_names: str) -> DictStrAny:
+    """Returns row counts for `table_names` as dict, if no table names are given, all data tables are counted"""
```
djudjuu, line 243: learning question (feel free to ignore): does this have to be DictStrAny?
why not dict str int?

tests/pipeline/utils.py
```diff
@@ -114,150 +115,63 @@ def assert_load_info(info: LoadInfo, expected_load_packages: int = 1) -> None:
...
-    iceberg_table_names = [
-        table_name
-        for table_name in table_names
-        if get_table_format(client.schema.tables, table_name) == "iceberg"
-    ]
-    if len(iceberg_table_names) > 0:
-        from dlt.common.libs.pyiceberg import get_iceberg_tables
-
-        iceberg_tables = get_iceberg_tables(p, *table_names, schema_name=schema_name)
+    result: Dict[str, List[Dict[str, Any]]] = {}
```
djudjuu, line 152: this type is a bit complex. 
this is my understanding:

```
{
    "users": [
        {"id": 1, "name": "Alice", "age": 30},
        {"id": 2, "name": "Bob", "age": 25},
    ],
    "orders": [
        {"order_id": 101, "user_id": 1, "amount": 250.5},
        {"order_id": 102, "user_id": 2, "amount": 100.0},
    ],
}
```
Outer dictionary:
- "users" and "orders" are table names.
Inner list:
Each list contains dictionaries representing rows in the respective table.
Inner dictionaries:
Keys are column names (id, name, age, etc.).
Values are the corresponding data for each column.

so can we add some intermediary types, that make live easier here?

```
# Represents a single row in a table (column name -> value)
TableRow = Dict[str, Any]

# Represents all rows in a table (list of rows)
TableRows = List[TableRow]
```

tests/load/pipeline/test_open_table_pipeline.py
```diff
@@ -490,7 +490,8 @@ def two_part_empty():
         assert destination_config.destination_type == "filesystem"
     except IcebergViewException:
         # currently duckdb does not allow to create views on empty Iceberg tables
-        assert destination_config.destination_type != "filesystem"
+        # assert destination_config.destination_type != "filesystem"
+        pass
```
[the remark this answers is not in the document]
sh-rp: the assert did not make any sense imho

tests/load/pipeline/test_open_table_pipeline.py
sh-rp: same as above

dlt/helpers/ibis.py
sh-rp: ah, yes, this can go, thanks!

tests/pipeline/utils.py
```diff
@@ -114,150 +115,63 @@ def assert_load_info(info: LoadInfo, expected_load_packages: int = 1) -> None:
...
 
 
-def _is_filesystem(p: dlt.Pipeline) -> bool:
+def _is_sftp(p: dlt.Pipeline) -> bool:
     if not p.destination:
         return False
-    return p.destination.destination_name == "filesystem"
+    return (
+        p.destination.destination_name == "filesystem"
+        and p.destination_client().config.protocol == "sftp"  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
sh-rp: If it is a filesystem, then a protocol will always be there!

tests/pipeline/utils.py
sh-rp: OK :)

tests/load/lancedb/utils.py
sh-rp: yes, correct

tests/load/utils.py
sh-rp: Because i had special code testing the insert_values and these are implicitly tested on many destinations

tests/pipeline/utils.py
sh-rp: Yeah, good point, I'm just using it because this type has always been here ;)

tests/pipeline/utils.py
```diff
@@ -114,150 +185,109 @@ def assert_load_info(info: LoadInfo, expected_load_packages: int = 1) -> None:
...
-            f_q_table_name = c.make_qualified_table_name(table_name)
-            query = f"SELECT {query_columns} FROM {f_q_table_name}"
-            with c.execute_query(query) as cur:
-                for row in list(cur.fetchall()):
-                    table_rows.append(dict(zip(columns, row)))
-        result[table_name] = table_rows
+        relation = p.dataset(schema=schema_name)[table_name]
+        columns = list(relation.columns_schema.keys())
+        for row in relation.fetchall():
+            result[table_name] = result.get(table_name, []) + [dict(zip(columns, row))]
```
djudjuu, line 290: this row is why I love python :heart:

tests/pipeline/utils.py
```diff
@@ -268,36 +298,89 @@ def load_tables_to_dicts(
...
+    """
+
+    def dict_to_tuple(d):
+        # Sort items to ensure consistent ordering
+        return tuple(sorted(d.items()))
+
+    counter1 = Counter(dict_to_tuple(d) for d in actual)
+    counter2 = Counter(dict_to_tuple(d) for d in expected)
+
+    assert counter1 == counter2
```
djudjuu, line 366: praise: nice improvement, faster, better error-msg I suppose and checks row-counts too

tests/pipeline/utils.py
```diff
@@ -308,157 +391,201 @@ def assert_only_table_columns(
...
 
 def assert_schema_on_data(
     table_schema: TTableSchema,
     rows: List[Dict[str, Any]],
     requires_nulls: bool,
     check_nested: bool,
 ) -> None:
-    """Asserts that `rows` conform to `table_schema`. Fields and their order must conform to columns. Null values and
-    python data types are checked.
+    """Validate that *rows* conform to *table_schema*.
```
djudjuu, line 571: praise: I love the doc-strings.

djudjuu (approved):
good stuff. Love the doc-strings too, I imagine this would make it much more likely that someone would use these helpers themselves

tests/pipeline/utils.py
djudjuu: just for fun: here is the same in java:
```
// Result map: table name -> list of rows (each row is a map of column -> value)
        Map<String, List<Map<String, Object>>> result = new HashMap<>();

        // Table name
        String tableName = "users";

        // Create a new row as a map (column -> value)
        Map<String, Object> newRow = new HashMap<>();
        for (int i = 0; i < columns.size(); i++) {
            newRow.put(columns.get(i), row.get(i));
        }

        // Add the new row to the result map
        result.putIfAbsent(tableName, new ArrayList<>()); // Ensure the table exists in the map
        result.get(tableName).add(newRow);

        // Print the result
        System.out.println(result);
   ```

File: tests/pipeline/utils.py [hunk 16]
```diff
@@ -308,157 +391,201 @@ def assert_only_table_columns(
...
+        table_schema (TTableSchema): Table schema as returned by
+            ``dlt.common.schema``.
+        rows (List[Dict[str, Any]]): Table rows to validate.
+        requires_nulls (bool): When ``True`` all nullable columns must appear at
+            least once with a ``None`` value in *rows*.
+        check_nested (bool): Whether to parse JSON strings and validate nested
+            structure types.
+
+    Raises:
+        AssertionError: If any of the schema constraints are violated.
```
anuunchin, line 588: if the `json.loads()` fails for ckeck_nested it will raise a different error - or is it an irrelevant detail 👀

sh-rp: It will, but I would say that is not really relevant, theoretically many types of exceptions could arise from calling this function, so I would document the ones that are important in the context. If json.loads fails, it will be quite clear what happened I think.

tests/pipeline/utils.py
```diff
@@ -72,13 +113,32 @@ def item(n):
...
+
+    Yields:
+        Any: Dynamically created resources named ``resource_<n>``.
+    """
     for n in range(many):
         yield dlt.resource(run_deferred(iters), name="resource_" + str(n))
 
 
 @dlt.resource(table_name="users")
 def users_materialize_table_schema():
```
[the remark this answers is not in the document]
sh-rp: For the time being we do not have super strict typing on the tests folder, so many functions to not have return types and the linter will pass..

dlt/common/destination/client.py
```diff
@@ -652,8 +653,14 @@ def get_open_table_catalog(self, table_format: TTableFormat, catalog_name: str =
         """Gets the catalog that keeps tables' metadata. Currently only pyiceberg Catalog is supported"""
 
     @abstractmethod
-    def get_open_table_location(self, table_format: TTableFormat, table_name: str) -> str:
-        """Computes location in which table metadata is stored. Does not verify if table exists."""
+    def get_open_table_location(
```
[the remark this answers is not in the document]
rudolfix: maybe we can add path separator at the end if this is a folder? I thought `filesystem` does that?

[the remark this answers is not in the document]
rudolfix: we already implement this interface for pyiceberg, there it is irrelevant

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -894,14 +896,17 @@ def get_open_table_catalog(self, table_format: TTableFormat, catalog_name: str =
...
+    ) -> Tuple[str, bool]:
         """All tables have location, also those in "native" table format."""
         folder = self.get_table_dir(table_name)
+        prefix = self.get_table_prefix(table_name)
         location = self.make_remote_url(folder)
         if self.config.is_local_filesystem and os.name == "nt":
             # pyiceberg cannot deal with windows absolute urls
             location = location.replace("file:///", "file://")
-        return location
+        return location, (folder == prefix or (folder + os.path.sep) == prefix)
```
rudolfix, line 909: maybe `make_remote_url` removes trailing separator from `folder`?

but overall `filesystem` does not enforce the convention that folder should always end with separator. maybe we should unit test a few functions there to make sure they follow this convention

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -168,12 +170,17 @@ def _escape_column_name(col_name: str) -> str:
...
-            supports_wildcard_notation = not self.is_abfss
+            supports_wildcard_notation = not self.is_abfss and is_folder
 
             resolved_files_string = f"'{table_location}/**/*.{first_file_type}'"
+
             if not supports_wildcard_notation:
+                # list_table_files returns a list of absolute paths but without scheme
                 files = self.remote_client.list_table_files(table_name)
-                resolved_files_string = ",".join(map(lambda f: f"'{protocol}://{f}'", files))
+                if protocol == "file":
```
rudolfix, line 180: was there something wrong with leaving the protocol for "file"? maybe on windows? I'm asking because I recently spent a lot of time fixing this place. but you seem to be right here!

btw. where we really lack docstiring is FsClient. does `list_table_files` return absolute path? with scheme or not?

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -168,12 +170,17 @@ def _escape_column_name(col_name: str) -> str:
             first_file_type = self.get_file_format(table_schema)
 
             # build files string
-            supports_wildcard_notation = not self.is_abfss
+            supports_wildcard_notation = not self.is_abfss and is_folder
```
rudolfix, line 173: does this duckdb limitation apply to all protocols? did you check it? could be useful

rudolfix (changes requested):

dlt/destinations/impl/filesystem/sql_client.py
sh-rp: It does, and that is also reflected in the statement. supports_wildcard_notation always if false if we do not have a folder, regardless of the protocol.

dlt/destinations/impl/filesystem/filesystem.py
sh-rp: I am ensuring now that this function returns a url with a trailing slash if it is a folder (it has to be removed in some places downstream again for things to work)

PR #2567, opened 2025-04-28
Fix: drop corresponding staging table when original table is dropped

### LINKED ISSUE #2534: Resetting a pipeline resource should also drop staging table
### dlt version

1.9.0

### Describe the problem

Selectively resetting a resource using `dlt pipeline drop <resource>` currently only drops the final table but not the staging table of the resource, which can cause load failure if the schema is changed in an incompatible way.

Manually dropping the staging table will trigger a different error when `dlt` attempts to truncate the staging table.

### Expected behavior

* Staging table should be dropped and recreate when a pipeline's resource is reset.
* DLT should not run truncate command if the staging table doesn't exist

### Steps to reproduce

1. Create a dlt pipeline
2. Drop a resource with `dlt pipeline drop <resource>`
3. Change the type of a column in the resource and reload
4. An error is raised because of incompatible data when loading staging table.

### Operating system

Linux

### Runtime environment

Local

### Python version

3.13

### dlt data source

_No response_

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
### Description
This PR simply drops staging tables when the `drop_tables` function is called. Previously, the drop command didn't drop the staging tables, which was causing issues when the resource was being loaded with an updated schema.

### Related Issues

- Fixes #2534

### DIFF
--- a/dlt/load/utils.py
+++ b/dlt/load/utils.py
@@ -131,15 +131,17 @@ def init_client(
             )
         )
 
-        with job_client.with_staging_dataset():
-            _init_dataset_and_update_schema(
-                job_client,
-                expected_update,
-                staging_tables | {schema.version_table_name},  # keep only schema version
-                staging_tables,  # all eligible tables must be also truncated
-                staging_info=True,
-                drop_tables=drop_table_names,  # try to drop all the same tables on staging
-            )
+        # if there are tables to drop, we should also drop them in the staging dataset
+        if staging_tables or drop_table_names:
+            with job_client.with_staging_dataset():
+                _init_dataset_and_update_schema(
+                    job_client,
+                    expected_update,
+                    staging_tables | {schema.version_table_name},  # keep only schema version
+                    staging_tables,  # all eligible tables must be also truncated
+                    staging_info=True,
+                    drop_tables=drop_table_names,  # try to drop all the same tables on staging
+                )
 
     return applied_update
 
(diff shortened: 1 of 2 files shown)


sh-rp (changes requested):
In case I am not missing something, there is a much simpler way to drop the staging tables. Have a look at the `drop_storage` function in the job_client_impl. It will drop the main storage (which is the main dataset) and then drop the staging storage if it exists by switchting to the staging dataset with the contextmanager. It also suppresses `DatabaseUndefinedRelation` which is raised if these tables do not exist. I think you can do exactly the same thing for `drop_tables` in `job_client_impl`.

dlt/load/utils.py
```diff
@@ -131,7 +131,8 @@ def init_client(
             )
         )
 
-        if staging_tables:
+        # if there are tables to drop, we should also drop them in the staging dataset
+        if staging_tables or drop_table_names:
             with job_client.with_staging_dataset():
```
anuunchin, line 136: Basically, what was happening is that when the client was initialized from the drop command, there were no staging tables available and the old `if staging_tables` condition could never became true. The thing is that the func `_extend_tables_with_table_chain` that returns `staging_tables` uses the schema that only contains dlt tables when we're dropping. I'm not sure if this is intentional  :eyes:

rudolfix: it is a bug. like 99% :)

rudolfix (approved):
LGTM! test is good. if all drop and staging tests will pass we are good here

rudolfix (approved):
LGTM!

PR #2570, opened 2025-04-28
Fix/2486 fixes mssql datetime precision

### LINKED ISSUE #2486: sql_database MSSQL conversion error with datetime
### dlt version

1.9.0

### Describe the problem

Trying to export SQL table from MSSQL using pyarrow backend gives a conversion error for datetime columns. I am using PYODBC with SQL Server Driver 2017.

```
<class 'dlt.extract.exceptions.ResourceExtractionError'>
In processing pipe RISK_HEADERS: extraction of resource RISK_HEADERS in generator table_rows caused an exception: Conversion to arrow failed for field `RNWDATE` with dlt hint `data_type=timestamp` and `inferred_arrow_type=timestamp[ns]` This data type seems currently unsupported by dlt. Please open a GitHub issue
```

Max datetime in this column is "2918-08-01 00:00:00.000", would this cause issues being so far in the future? I have tried using table_adapter_callback to amend these values but I am either making a mistake or it isnt the solution,

```
    def _table_adapter_callback(table: sa.Table) -> sa.Table:

        for column in table.columns.values():            
            if isinstance(column.type, sa.DateTime):
                sa.update(table).where(column >= "2050-01-01 00:00:00.000").values(column.name="2050-01-01 00:00:00.000")

        return table
```

### Expected behavior

_No response_

### Steps to reproduce

```
pipeline = dlt.pipeline(
        pipeline_name="pipeline_dlt_test",
        destination="mssql",
        dataset_name="dlt_test"
    )

    source = sql_database(
        credentials=credentials,
        chunk_size=50000,
        include_views=True,
        schema="dbo",
        backend="pyarrow",
        table_names=["..."]
            
    )

    # Run the resources together
    info = pipeline.extract(source, write_disposition="replace")
    print(info)
```

### Operating system

Windows

### Runtime environment

Local

### Python version

3.12

### dlt data source

Microsoft SQL Server
sql_database

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
fixes #2486 and adds some test cases for naive datetimes

### DIFF
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -843,22 +843,28 @@ def convert_numpy_to_arrow(
                     data_type=dlt_data_type,
                     inferred_arrow_type=inferred_arrow_type,
                     details=(
-                        "Insufficient decimal precision. Consider setting `precision` and `scale`"
-                        " hints: https://dlthub.com/docs/general-usage/schema/#tables-and-columns"
+                        f"Insufficient decimal precision {error_msg}. Consider setting `precision`"
+                        " and `scale` hints:"
+                        " https://dlthub.com/docs/general-usage/schema/#tables-and-columns"
                     ),
                 ) from e
 
             elif (
-                "to utf8 using function cast_string" in error_msg
-                and dlt_data_type in ("json", "text")
-                and pa.types.is_string(inferred_arrow_type)
-            ):
+                ("to utf8 using function cast_string" in error_msg and dlt_data_type == "text")
+                or dlt_data_type == "json"
+            ) and pa.types.is_string(inferred_arrow_type):
                 # this is handled by fallback case 3
                 logger.warning(
                     f"Received `data_type='{dlt_data_type}'`, data requires serialization to"
                     " string, slowing extraction. Cast the JSON field to STRING in your database"
                     " system to improve performance. For example, create and extract data from an"
                     " SQL VIEW that SELECT with CAST."
+                )
+            else:
+                raise PyToArrowConversionException(
+                    data_type=dlt_data_type,
+                    inferred_arrow_type=inferred_arrow_type,
+                    details=f"This conversion is currently unsupported by dlt ({error_msg})",
                 )
 
     # case 2: encode Sequence and Mapping types (list, tuples, set, dict, etc.) to JSON strings
@@ -888,10 +894,10 @@ def convert_numpy_to_arrow(
     if arrow_array is None and dlt_data_type is None:
         try:
             arrow_array = pa.array(column_data)
-        except (pa.ArrowInvalid, pyarrow.ArrowTypeError):
+        except (pa.ArrowInvalid, pyarrow.ArrowTypeError) as e:
             logger.warning(
-                "Type can't be inferred by `pyarrow`. Values will be encoded as in a loop, slowing"
-                " extraction."
+                f"Type can't be inferred by `pyarrow` {e.args[0]}. Values will be encoded as in a"
+                " loop, slowing extraction."
             )
             encoded_values: list[Union[None, Mapping[Any, Any], Sequence[Any], str]] = []
             for value in column_data:
@@ -917,13 +923,6 @@ def convert_numpy_to_arrow(
                     ) from e
 
             arrow_array = pa.array(encoded_values)
-
-    if arrow_array is None:
-        raise PyToArrowConversionException(
-            data_type=dlt_data_type,
-            inferred_arrow_type=inferred_arrow_type,
-            details="This data type seems currently unsupported by dlt. Please open a GitHub issue",
-        )
 
     return arrow_array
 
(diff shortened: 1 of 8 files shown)


tests/load/sources/sql_database/mssql_source.py
```diff
@@ -0,0 +1,161 @@
...
+from sqlalchemy import text
+from sqlalchemy.exc import SQLAlchemyError
+import mimesis
+
+from dlt.common.pendulum import pendulum, timedelta
+from dlt.common.utils import uniq_id
+from dlt.sources.credentials import ConnectionStringCredentials
+
+
+class MSSQLSourceDB:
```
sh-rp, line 16: I would've probably created a generic source db that uses sqlalchemy and maybe sqlmodel so we can create an example dataset on any database including an abstraction for manipulating rows. We already have more or less the exact same code for postgres. But that is something for the future maybe.

dlt/extract/incremental/transform.py
```diff
@@ -113,17 +120,37 @@ def __call__(
     ) -> Tuple[bool, bool, bool]: ...
 
     @staticmethod
-    def _adapt_if_datetime(row_value: Any, last_value: Any) -> Any:
-        # For datetime cursor, ensure the value is a timezone aware datetime.
-        # The object saved in state will always be a tz aware pendulum datetime so this ensures values are comparable
+    def _adapt_timezone(row_value: datetime, cursor_value: datetime, cursor_value_name: str) -> Any:
```
sh-rp: This is smart! If we have incoming rows with varying timezone awareness there will be a big mess though, right? Maybe we should somehow raise in the normalizer when we detect this. Or is there a mechanism somewhere? I have not seen this coming up in the community, but if there is unstructured data being loaded this could always be a possibility.

dlt/sources/sql_database/helpers.py
```diff
@@ -109,6 +113,14 @@ def __init__(
     def _make_query(self) -> SelectAny:
         table = self.table
         query = table.select()
+
+        # use limit step to limit the query
+        if self.limit:
+            # limit works with chunks (by default)
+            limit = self.limit.limit(self.chunk_size)
+            if limit is not None:
+                query = query.limit(limit)
```
sh-rp, line 122: Just out of curiosity, is it faster to apply a limit on the query even when you are not retrieving all rows from the cursor?

dlt/destinations/impl/bigquery/factory.py
```diff
@@ -69,6 +71,27 @@ def ensure_supported_type(
...
+        if precision and precision != self.capabilities.timestamp_precision:
+            logger.warn(
+                f"BigQuery does not support custom precision for column '{column_name}' in"
+                f" table '{table_name}'. Will use default precision."
+            )
+
+        # TIMESTAMP is always timezone-aware in BigQuery
+        # DATETIME is always timezone-naive in BigQuery
+        # NOTE: we disable DATETIME because it does not work with parquet
+        return "TIMESTAMP" if timezone else "TIMESTAMP"
```
sh-rp: Is this line correct? The condition does not change anything..

dlt/extract/incremental/lag.py
```diff
@@ -1,57 +1,70 @@
...
+            value,
             f"Lag is not supported for cursor type: {type(value)} with last_value_func:"
-            f" {last_value_func}. Strings must parse to DateTime or Date."
+            f" {last_value_func}. Strings must parse to DateTime or Date.",
         )
 
     return value
 
 
 def _apply_lag_to_datetime(
```
sh-rp, line 56: should be `apply_lag_to_date` I think

tests/load/sources/sql_database/mssql_source.py
rudolfix: right! we can have a few tables with standard types that should work on all source databases. for that we can extract a few standard tests.

tables with specific datatypes could at least have standardized names. but this is significant amount of work. I did a first step by enabling more source databases here.

dlt/extract/incremental/transform.py
rudolfix: huh! why this comment is on outdated code? Anyway - if tz-awareness changes on cursor column, the comparison will fail and exact step will abort. we normalize data after extract phase. so user should use `map` to normalize this data.

incremental is used mostly for sql databases and rest apis so this problem is not popping up - like you say

dlt/sources/sql_database/helpers.py
rudolfix: this is very dependent on particular implementation. what really makes a difference is an index on cursor column. then query engine can stream that data in chunks without scanning. in that case `limit` will not change much AFAIK. it however allows to load data from connectorx in chunks

dlt/destinations/impl/bigquery/factory.py
rudolfix: it is but I'll remove it to make it 100% clear. DATETIME does not work on bigquery in practice

dlt/extract/incremental/lag.py
rudolfix: yeah. typing is wrong.

Issue #2547, opened 2025-04-23
Ensure factory functions of core sources produce nice default values


sh-rp:
### TLDR
After implementing #2531 we should ensure that the defaults created for the sql_database, rest_api and filesystem sources provide a good starting point for engineers to implement their own settings.

### Rest API
The rest api does not play very nice with the automated value discovery, this is due to the nested dictionary structure. It would be nice to have something that corresponds to this jaffle shop config:

```py
source = rest_api_source(
    {
        "client": {
            "base_url": "https://jaffle-shop.dlthub.com/api/v1/",
                "paginator": {
                    "type": "header_link",
                },
            },
            "resources": [
                "customers",
                "products",
            ],
        },
    )
```

in the dlt.yml as the result of an instantiation.

### SQL Database

As a starting point, let's try to replicate it so it will sync the familes from the genome database, see the example pipeline script for this. I think we could also host a jaffle shop postgres database for a few bucks a months, but that would be the next step, first let's try to get the code in order.

### Filesystem

We could probably add a filesystem that just points to a local folder. This is tbd.

rudolfix:
REST API has a different signature than above and it will be easier to provide defaults:
```py
@decorators.source
def rest_api(
    client: ClientConfig = dlt.config.value,
    resources: List[Union[str, EndpointResource, DltResource]] = dlt.config.value,
    resource_defaults: Optional[EndpointResourceBase] = None,
) -> List[DltResource]:
```
but those defaults will look nice in `yaml` format. if we will put them into `toml` config it won't look good and will be something that is not even documented.
Maybe: if we generate a sample configuration and it does not contain any secrets then we should place it in `config.toml` and for now in OSS we skip that?

right now I get something like this:
```toml
[sources.rest_api]
resources = ["a", "b", "c"] # please set me up!
[sources.rest_api.client] # please set me up!
key = "value"
```

**SQL Database**
Here we have two options:
1. generate config like below
```toml
[sources.sql_database.credentials]
drivername = "drivername" # please set me up!
database = "database" # please set me up!
password = "password" # please set me up!
username = "username" # please set me up!
host = "host" # please set me up!
port = 0 # please set me up!
```
but with placeholders that make it completely clear that they are wrongly configured. I also had an idea that we can use something like
`drivername = "#please set me up1"
and if we see this in `resolve_confg` we treat that as unresolved variable and issue a special warning. WDYT?

btw. in config.toml I get
```toml
[sources.sql_database]
[...]

Issue #2531, opened 2025-04-17
Sane defaults in config and secrets


djudjuu:
### Feature description

we should provide default values for configs and secrets that are only used in the config and secrets tomls on instantiation and never directly in the code. 

### Why we need it

For example when instantiating the sql_database as a source, the defaults that are written into the toml file will look like they make sense but will raise obscure errors when called. It would be nice to have some defaults there that connect to some public database which the user then may change. We do not want these defaults to be defaults in the code though, that does not make sense. The same is true for the rest_api.

### Implementation details

We currently have a `__config_gen_annotations__` class var on our configuration classes which will be used by the code of the init cli functions to determine which configuration params of a Configuration should be written as example value into the toml files. For now this is just a list of variables which are marked as vars that should go into the `toml` file. This is done in `write_spec` in `config_toml_writer.py`. We can change the `__config_gen_annotations__` so it may also be a dictionary. If it is a dictionary then the keys are the name of the config var and the value is the default value to be used when writing the `toml` files. If the value is None, the old mechanism of discovering the config value is used. Explicit default values should still override the setting in `__config_gen_annotations__`.

Furthermore we will have to extend the `source` and `resource` decorators to take this dict, so it will be used when creating a new source from the command line. For this the source decorator should have a new parameter `_config_gen_annotations` which needs to be forwarded to `spec_from_signature` where the Configuration class for the source is built and then needs to be set there. This way the above mechanism SHOULD also work for decorated sources.

djudjuu:
@rudolfix 
>  do we want to start generating config.toml as well in OSS? 
I am pretty sure, this was always the case! `dlt init sql_database duckdb`  would generate the example script and all sources/resources that were imported in there would write the required configs into the config.toml

![Image](https://github.com/user-attachments/assets/e4ee9ffe-7e72-4219-8346-4c9093b395fd)

PR #2571, opened 2025-04-29
WIP: generate example configs via source decorator

djudjuu:
referencing:  
- https://github.com/dlt-hub/dlt/issues/2531#issuecomment-2825361442
- https://github.com/dlt-hub/dlt/issues/2547


dlt/sources/rest_api/__init__.py
```diff
@@ -63,7 +63,28 @@
...
+@decorators.source(
+    example_values={
+        "client": {"base_url": "https://jaffle-shop.scalevector.ai/api/v1/"},
+        # "resources": ["customers", "orders"],
+        # or
+        "resources": [
+            {
+                "name": "orders",
+                "endpoint": {
+                    "path": "orders",
```
djudjuu, line 75: I don't see a nice solution way of putting this into the `config.toml file`
there also are plenty of other even longer examples in the comment below, so there is some duplication (and docstrings can not be templated, so we can not reuse it in there)

dlt/sources/rest_api/__init__.py
```diff
@@ -63,7 +63,31 @@
...
-@decorators.source
+@decorators.source(
+    example_values={
+        "client": {
+            "base_url": "https://jaffle-shop.scalevector.ai/api/v1/",
+            "paginator": "header_link"
+            },
+        "resources": ["customers", "orders"],
+        # or
+        # "resources": [
```
djudjuu, line 74: I don't see a nice solution way of putting this into the config.toml file.
there also are plenty of other even longer examples in the comment below, so there is some duplication.

maybe we need a dlthub just for rest-api configs

PR #2575, opened 2025-04-29
Experiment: Test disable dropping of datasets in tests

sh-rp:
### Description
This is an experiment to see if we can save CI time by not dropping datasets in a pytest fixtures but maybe in a nightly workflow run. These nightly runs would not be concurrent but only run once, so they could also be executed from some github runner on a small DO droplet to note waste CI minutes.

Things to check:

* Execution time (Athena w/o iceberg is at around 70-80 minutes for essential tests, with iceberg around 170 minutes)
* Test collissions. It is conceivable that some tests use the same dataset name without dev mode and we see collissions, so if tests fail, investigate!


sh-rp:
Closing. This change does not appear to speed up CI at all. We will still create a cleanup script for our destinations and then apply this change, but it does not change much.

Issue #2540, opened 2025-04-22
Transformation and Lineage follow-up work


sh-rp:
A few tasks to do after the first round of schema discovery and transformations is done:

- [x] Implement `allow_unknown_columns` and `allow_anonymous_columns` in lineage
- [x] See if we can get native snowflake and `clickhouse` lineage to work, right now we are using duckdb to help out. Same for `sqlalchemy`, but there I don't see a way of doing in in another way.
- [x] Figure out which hints to keep after schema discovery, right now all are kept which is not correct. Probably we can detect if it is a one to one table copy, then keep all hints, if it is some derived version of the table, at least remove `primary_key` and `merge_keys`, but maybe also others. We need a proposal here.
- [x] Implement function `is_same_physical_destination` for SupportsReadableDataset so we can give good errors if users try to run sql transformations on datasets that are not on the same location.
- [x] Implement and test correct exception classes when returning incompatible from transformation function.
- [ ] Extend dataset so it that it is aware of incremental data. The constructor could take a load_id which we consider to be the last load id that was fully processed, on instantiated the dataset would store this last_load_id and also store the newest completed load by accessing the destinations. Then this instance is able to return incremental records by selecting rows with load_id larger than last_load_id and smaller or equals last_successful_load. There is a kind of prototype of this in the transformation incremental tests.
- [ ] Make `chunk_size` for python based transformations configurable. Either by having an own config implementation or inheriting normalizer settings.
- [x] We have an issue on the sqlgot repo for figuring out why snowflake column annotation does not work in lineage: https://github.com/tobymao/sqlglot/issues/5098
- [ ] Add tests with special naming to see wether identifier normalization works correctly
- [x] Add tests with all data-types
- [ ] Convert sqlglot schema creation functions to a dlt `TypeMapper`

Docs:
- We have some very nice feedback on the transformations tutorial on this older hackathon PR: https://github.com/dlt-hub/dlt/pull/2401

Maybe:
- [ ] Make multidataset transformations work. We would need to be able to do lineage on a combined schema. it was working in the prototype, but we changed the lineage code a bit. Probably we would need to retain all involved schemas in the ibis proxy. There is a commented out test in the transformation tests.
- [x] Add some kind of scalar method to the ibis proxy that returns the scalar value for max or sum queries that only have one row with one column.

sh-rp:
closing in favor of https://github.com/dlt-hub/dlt/issues/2776

PR #2577, opened 2025-04-29
feat: 2540 lineage `allow_unknown_columns` and `allow_anonymous_columns`

zilto:
This PR follows #2540. Mainly, it adds exception handling and test cases for intended behavior.

A few recommended changes to discuss:
- for the API of `compute_columns_schema()`, I recommend renaming `allow_unknown_columns` to `infer_schema: bool`. This would match the SQLGlot API that it uses under the hood.
 ```py
 def compute_columns_schema(
    sql_query: str,
    allow_unknown_columns: bool = True,
    allow_anonymous_columns: bool = True,
    allow_fail: bool = True,
) -> TTableSchemaColumns:
 ```
 - please review test cases carefully. The `allow_unknown_columns` can behave in odd ways.
 - Is it ok if the returned dlt schema has `data_type` missing?


dlt/transformations/lineage.py
```diff
@@ -163,6 +180,10 @@ def create_sqlglot_schema(
 
 
 # TODO should we raise an exception for anonymous columns?
+# TODO rename `allow_unknown_columns` to `infer_sqlglot_schema`;  this would match the SQLGlot `infer_schema` kwarg;
+# even if `allow_unknown_columns=True`, some queries haved undetermined final columns
+# TODO rename `allow_fail` to `allow_partial` or `allow_incomplete`. This would ensure a dictionary is always returned
+# even in cases of SQL parsing errors, missing table reference, unresolved `SELECT *`
```
sh-rp, line 186: @zilto you can do those two changes now if you like.

sh-rp: I'd prefer `allow_partial`

dlt/transformations/lineage.py
```diff
@@ -171,14 +192,68 @@ def compute_columns_schema(
...
+
+        raise LineageFailedException(
+            f"Parsed SQL query is not a SELECT statement. Received SQL expression of type {expression.type}."
+        )
+    
+    if allow_anonymous_columns is False:
+       for col in expression.selects:
+           if col.output_name == "":
+               raise LineageFailedException(
+                   "Found anonymous column in SELECT statement. Use `allow_anonymous_columns=True` for permissive handling.\n"
```
sh-rp, line 221: we'll have to write nice docs for this so people understand what is going on :)

tests/transformations/test_lineage.py
```diff
@@ -0,0 +1,302 @@
...
+                "table_1": {
+                    "col_varchar": sge.DataType.build("VARCHAR", dialect=dialect),
+                    "col_bool": sge.DataType.build("BOOLEAN", dialect=dialect),
+                }
+            }
+        }
+    )
+
+    select_query = "SELECT col_unknown FROM table_unknown;"
+    # TODO note that it doesn't contain a `data_type` field. Should we enforce a default?
```
sh-rp, line 138: I would not put a data_type if we do not know it, so this code is ok I think.

sh-rp (commented):
@zilto very nice PR! You can make the suggested arg changes if you like and please ensure the transformations tests run (testing duckdb is enough) before merging this into the transformations branch. Please ensure that the lineage is called with the correct parameters from inside the transformations function:
* infer_sqlglot_schema=True (we should make this configurable probably)
* allow_partial=False
* allow_anonymous_colums=False (we need a very helpful exception that explains that aliases should be used)

dlt/transformations/lineage.py
```diff
@@ -171,14 +192,68 @@ def compute_columns_schema(
...
-        raise TypeError(
-            f"Received SQL expression of type {expression.type}. Only SELECT queries are accepted."
+
+        raise LineageFailedException(
+            f"Parsed SQL query is not a SELECT statement. Received SQL expression of type {expression.type}."
+        )
+    
+    if allow_anonymous_columns is False:
+       for col in expression.selects:
+           if col.output_name == "":
```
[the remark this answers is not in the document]
zilto: correct, but the output name is generated later by `qualify()`. 

The current implementation is:
- `allow_anonymous_columns is False` don't allow unnamed columns in user-provided query. Raise an exception.
- `allow_anonymous_columns is True` allow unnamed columns in user-provided query and infer their name

PR #2581, opened 2025-04-30
Speedup CI: Cache google secrets

sh-rp:
### Description
This PR aims to speed up CI runs by caching google secrets. Previously google secrets was hit about 15 times per test, this patches the secrets provider to cache the first hit during the entire pytest session which works for our specific use case. Exemptions from this patch for the tests for the Provider are included.


rudolfix (approved):
LGTM!

PR #2587, opened 2025-05-02
feat: 2540 - SQLGlot type mapping

zilto:
This PR follows #2540. It focuses on building robust type mapping between SQLGlot and dlt. This would allow to reliably propagate type-related hints (nullable, precision, scale, timezone) via SQLGlot.

This is WIP, details to come. Ignore the file `dlt/transformations/infer_hints.py` for now.
## Notes
- This relates to #2498. The code could potentially be simplified to use the Ibis `SQLGlotBackend`


dlt/common/libs/sqlglot.py
```diff
@@ -0,0 +1,437 @@
...
+
+
+def _build_parameterized_sqlglot_type(
+    dlt_type: TDataType,
+    nullable: Optional[bool] = None,
+    precision: Optional[int] = None,
+    scale: Optional[int] = None,
+    timezone: Optional[bool] = None,
+) -> DataType:
+    hints: TColumnType = {}
```
sh-rp, line 283: this hints dict is never read

dlt/common/libs/sqlglot.py
```diff
@@ -0,0 +1,437 @@
...
+            precision=precision,
+            scale=scale,
+            timezone=timezone,
+            nullable=nullable,
+        )
+    )
+
+
+# TODO decide default behavior: use specific types (e.g., DECIMAL128)
+# or parameterized types like DECIMAL(precision,scale)
```
sh-rp, line 222: my intuition would be to used the parametrized ones, but it's not clear to me what the drawbacks are? you mention above that they are not as reliable? if we use the non-parametrized ones, what do we do when we encounter a precision that is not available as its "own" type?

dlt/common/libs/sqlglot.py
```diff
@@ -0,0 +1,437 @@
...
+    return sqlglot_type
+
+
+def get_metadata(sqlglot_type: sge.DataType) -> TColumnSchema:
+    """Get a metadata dictionary from the SQLGlot DataType object."""
+    metadata = sqlglot_type.meta
+    return _filter_dlt_hints(metadata)  # type: ignore[arg-type]
+
+
+def _filter_dlt_hints(hints: TColumnSchema) -> TColumnSchema:
```
sh-rp, line 420: cool :)

dlt/common/libs/sqlglot.py
zilto: For now, I made the "parameterized type" the default code path. The "named type" approach is not used anywhere, but I kept it in the code for now.

dlt/common/libs/sqlglot.py
```diff
@@ -0,0 +1,480 @@
...
+    "timestamp": DataType.Type.TIMESTAMP,
+    "bigint": DataType.Type.BIGINT,
+    "binary": DataType.Type.VARBINARY,
+    "time": DataType.Type.TIME,
+    "decimal": DataType.Type.DECIMAL,
+    "wei": DataType.Type.UINT256,
+}
+
+
+# TODO should we raise errors on type conversion or silently convert to correct values?
```
sh-rp, line 151: Can you give an example for an error? I'm not quite clear when one would appear.

sh-rp (changes requested):
@zilto this looks very good so far. I thought I had also added a comment about testing what happens wrt mathematical operations between to columns that have a different precision or scale, but I can't see this comment here so I don't know where it went, maybe slack :) 

What I would do as next steps:

1) Add some tests that test what happens when precision and scale collide with mathematical operations for all datatypes that have precision or scale. Probably there are some useful tests we can do with datetime types. Maybe you can ask ChatGpt. The main goal is to find out how SQLGlot behaves and add notes into the docs so users know how to resolve problems and also understand why they are having problems if row data_types don't match. I would run those for all destinations (see the load/test_lineage.py tests. Although with your current changes, I think the outcome will probably be the same for each destination? We should test this. 

2) Rebase to or merge from the current transformations branch so the conflicts are gone, and then point the base branch to be devel (we will change it back later) This way all tests will be run and you can see wether there are any problems either in my or your part of the code that we need to address.

zilto:
I created a diagram to ground the discussion (could cleanup for the docs).

image

Note that the dlt schema for `avg_sales` table is not derived from any physical data

## Testing precision and scale handling
> Add some tests that test what happens when precision and scale collide with mathematical operations for all datatypes that have precision or scale. [...] 

IMO, what is asked here is outside the scope (and represent enormous amount of work). 

The SQLGlot function [annotate_types()](https://github.com/tobymao/sqlglot/blob/17f7eaff564790b1fe7faa414143accf362f550e/sqlglot/optimizer/annotate_types.py#L28) that we leverage already does the heavy-lifting of managing type promotion for ALL SQL operations. It accepts a `dialect` kwarg to enable backend-specific type handling (bigquery, presto, duckdb, etc.). The `dlt -> destination` mapping is very loose, but the `SQLGlot -> destination` mapping should be exact. Else, it's probably a bug that we can raise to the SQLGlot team. 
 
OTOH, the current challenge with our approach is that we have two paths:
- our engine: `sales (dlt) -> sales (sqlglot) -> infer query ->  avg_sales (sqlglot) -> avg_sales (dlt)`
- the destination: `sales (dlt) -> sales (dest) -> execute query -> avg_sales (destinat) -> avg_sales (dlt)`

### example
Here's a real example (didn't test for it):
```python
sql = "SELECT AVG(foo) as avg_foo FROM table"

# inference path
{"name": "foo", "data_type": "bigint", "precision": 0}   # dlt
DataType.Type.BIGINT(precision=0)  # SQLGlot
# infer SQL
DataType.Type.DOUBLE  # SQLGlot
{"name": "avg_foo", "data_type": "double"}

# execution path with duckdb backend
# DuckDB converts precision=0 to DECIMAL
{"name": "foo", "data_type": "bigint", "precision": 0}   # dlt
DECIMAL(precision=0)  # duckdb destination
# execute SQL
DECIMAL(precision=?)  # duckdb destination
{"name": "avg_foo", "data_type": "decimal", "precision": ?}  # dlt
```
Lessons:
- in the **inference path**, we expect SQLGlot to automatically convert our `BIGINT(precision=0)` to `DECIMAL(precision=0)` if we specify `dialect="duckdb"`
- because we use a "no round-trip" design, this is the best we can get 


## document behavior
> The main goal is to find out how SQLGlot behaves and add notes into the docs so users know how to resolve problems and also understand why they are having problems if row data_types don't match.

What types of problem do you expect? What does it means to have "row data_types that don't match"?

Since we have the **inference path** and the **execution path**, we currently guarantee explicitly:
1. `dlt.resource` matches the `dlt.Schema` it produces
2. `dlt.Schema` should describe the destination schema (the mapping is loose and not thoroughly tested)
3. `dlt.Schema` to `sqlglot.Schema` is thoroughly tested; our test suite describe the intended behavior
4. `sqlglot.Schema` to `dlt.Schema` is thoroughly tested; our test suite describe the intended behavior
5. SQLGlot query inference result schema should match destination query execution result schema

The main issues we can expected are related to *5.*:
- does the input `dlt.Schema -> sqlglot.Schema` adequately represents `destination schema -> sqlglot.Schema`?
- is the backend `dialect` properly set?
  - we should distinguish that users can write their SQL query in `dialect="duckdb"`, but executing that query on BigQuery requires setting `annotate_types(..., dialect="bigquery")` 
- inference get messy if there are multiple backends involved (e.g., you have a DAG where you expect certain SQL query on Snowflake then on DuckDB);

sh-rp:
@zilto thanks for your detailed thoughts! You are exactly right, my request was a bit to brief, I should've explained in a bit more detail. The problems that could arise (and we should at least have some understanding and simple tests about) are the ones where the type and precision inference that is done by sqlglot differs from what the database expects when running the query. 

So maybe you are right and these tests do not need to be in this PR but rather added into the transformations tests where we run the full thing: one or a few transformations that, for example, combine doubles with different precisions and see what the outcome is. Like I said above, the goal is probably not to fully prevent these problems, but to be aware of them and explain them in the docs and what the user can do (supply their own hints). Maybe we get lucky and it all just works, who knows :)

In the case where we have multiple backends (e.g. running a transformation on duckdb but loading into snowflake) we run our regular normalizer step and I think it should work or you would get a useful error from dlt. But we can also test for this.

PS: I meant "column data types", I really should read my posts again before I hit send ;)

PR #2594, opened 2025-05-03
slightly clarify clickhouse docs

sh-rp:
### Description
Fixes and improves clickhouse configuration docstring and docs page


rudolfix (commented):
@sh-rp I think what is confusing here is that `port` is not used at all in code base. Or maybe I'm wrong?

sh-rp:
I'll just merge this :)

PR #2597, opened 2025-05-04
list secrets in vault config provider to avoid calls to backend

rudolfix:
### Description
* adds option t list secrets upfront so we call backend only to fetch existing secrets
* refactors how vaults are configured
* documents google secret vault
* uses global secret cache in tests

### DIFF
--- a/docs/website/docs/walkthroughs/add_credentials.md
+++ b/docs/website/docs/walkthroughs/add_credentials.md
@@ -168,4 +168,4 @@ to Secrets backend.
 
 ## Retrieving credentials from other vault types
 Subclass `VaultDocProvider` and implement methods to fetch a secret and (optionally) to list secrets then
-[register subclass as custom provider](../../../examples/custom_config_provider/).
+[register subclass as custom provider](../../examples/custom_config_provider).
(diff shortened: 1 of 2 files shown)


dlt/common/configuration/providers/vault.py
```diff
@@ -123,13 +127,37 @@ def _lookup_paths(pipeline_name_: str, known_section_: str) -> None:
...
         pass
 
     def _update_from_vault(
         self, full_key: str, key: str, hint: type, pipeline_name: str, sections: Tuple[str, ...]
     ) -> None:
         if full_key in self._vault_lookups:
             return
+
+        # list the vault on first access if enabled
+        if self.list_secrets and self._available_keys is None:
```
sh-rp, line 149: cool :)

tests/conftest.py
```diff
@@ -34,6 +31,26 @@ def initial_providers(self) -> List[ConfigProvider]:
...
+    def _list_vault(self):
+        key_ = "__list_vault"
+        if key_ not in CACHED_GOOGLE_SECRETS:
+            CACHED_GOOGLE_SECRETS[key_] = super()._list_vault()
+        return CACHED_GOOGLE_SECRETS[key_]
+
+
+from dlt.common.configuration.providers import google_secrets
+
+google_secrets.GoogleSecretsProvider = CachedGoogleSecretsProvider  # type: ignore[misc]
```
sh-rp, line 52: I don't think this will work this way. I tried something like that when I was testing and it did not work, and I just checked out your branch and the cachedprovider is not used unfortunately..

sh-rp (approved):

Issue #2529, opened 2025-04-17
AWS Athena Information Schema Error with Large Table Lists in dlt Pipeline


amirdataops:
### dlt version

1.8.0

### Describe the problem

We encountered an issue in our dlt pipeline when querying AWS Athena's INFORMATION_SCHEMA.COLUMNS for multiple tables. The error we faced was:

Error listing table columns for catalog awsdatacatalog: Detected Iceberg type table without metadata location. Please make sure an Iceberg-enabled compute engine such as Athena or EMR Spark is used to create the table, or the table is created by using the Iceberg open source AWS library iceberg-aws.

This error happens due to the query executed against INFORMATION_SCHEMA:

```
sql:
SELECT table_name, column_name, data_type, is_nullable
    FROM INFORMATION_SCHEMA.COLUMNS
WHERE table_schema = '{my_table_schema_name}' AND table_name IN ({127 elements}) 
ORDER BY table_name, ordinal_position
```

The query works without issues when the number of elements in the IN clause is less than or equal to 100. However, with larger lists (e.g., 127 elements), the query takes around 20 minutes to complete, leading to pipeline failure.

To mitigate this, we adjusted the INFO_TABLES_QUERY_THRESHOLD parameter by forking the dlt library and setting the value from 1000 to 40, thus avoiding the fallback mechanism defined in the library's code:

```
python
if (
    len(name_lookup) > self.INFO_TABLES_QUERY_THRESHOLD
    or len(",".join(folded_table_names)) > self.capabilities.max_query_length / 2
):
    logger.info(
        "Fallback to query all columns from INFORMATION_SCHEMA due to limited query length"
        " or table threshold"
    )
    folded_table_names = []
```

This workaround resolved the error.

### Expected behavior

The query against INFORMATION_SCHEMA.COLUMNS should operate efficiently regardless of the number of table names passed in the IN clause. The pipeline should successfully retrieve metadata without fallback or excessive execution delays.

### Steps to reproduce

1. Use AWS Athena as the SQL backend for the dlt pipeline.
2. Attempt a query similar to the following:
```
sql
SELECT table_name, column_name, data_type, is_nullable
FROM INFORMATION_SCHEMA.COLUMNS
WHERE table_schema = '{my_table_schema_name}' AND table_name IN ({127 elements}) 
ORDER BY table_name, ordinal_position

```
3. Observe the error or extremely delayed query execution if the number of elements in the IN clause exceeds 100.
[...]

PR #2600, opened 2025-05-05
 2529-INFO_TABLES_QUERY_THRESHOLD-as-paramterer-from-config

amirdataops:
### Description
The pull request introduces a new configuration parameter (info_tables_query_threshold) to set the threshold for info tables queries in the DestinationClientConfiguration. This aims to make the query threshold dynamic and configurable, instead of using a hard-coded value.

### Related Issues

- Fixes [#2529](https://github.com/dlt-hub/dlt/issues/2529)

### Additional Context

Added the info_tables_query_threshold parameter in the DestinationClientConfiguration class in client.py.
Updated the SqlJobClientBase class (job_client_impl.py) to fetch and apply the threshold dynamically from the configuration during initialization. Ensured backward compatibility by setting a default threshold value of 1000.


djudjuu:
I linted the code for you, (`make format && make lint-fix`) otheewise this looks good. lets see what the test say.

one comment: Wouldn't it also make sense to actually set the default value for the `AthenaClientConfiguration` to something lower than 100 then?

dlt/common/destination/client.py
```diff
@@ -150,6 +150,9 @@ class DestinationClientConfiguration(BaseConfiguration):
     credentials: Optional[CredentialsConfiguration] = None
     destination_name: Optional[str] = None  # name of the destination
     environment: Optional[str] = None
+    info_tables_query_threshold: int = dataclasses.field(
```
rudolfix, line 153: you can just set `1000` directly. it is immutable so no need to use a factory. also pls move it to `DestinationClientDwhConfiguration` - not all destinations are SQL based, not all need this config

dlt/destinations/impl/athena/athena.py
```diff
@@ -178,6 +178,9 @@ def __init__(
         capabilities: DestinationCapabilitiesContext,
     ) -> None:
         super().__init__(None, dataset_name, staging_dataset_name, capabilities)
+        self.INFO_TABLES_QUERY_THRESHOLD = 80
```
rudolfix, line 181: this is hacky. please add `info_tables_query_threshold` to `AthenaClientConfiguration` and just use it directly. remove `INFO_TABLES_QUERY_THRESHOLD` from the whole codebase

tests/load/pipeline/test_pipelines.py
```diff
@@ -848,23 +848,23 @@ def test_query_all_info_tables_fallback(destination_config: DestinationTestConfi
...
-        info = pipeline.run([1, 2, 3], table_name="digits_2", **destination_config.run_kwargs)
-        assert_data_table_counts(pipeline, {"digits_1": 3, "digits_2": 3})
+    info = pipeline.run([1, 2, 3], table_name="digits_1", **destination_config.run_kwargs)
+    assert_load_info(info)
+    # create empty table
+    client: SqlJobClientBase
+    # we must add it to schema
+    pipeline.default_schema._schema_tables["existing_table"] = new_table("existing_table")
+    with pipeline.destination_client() as client:  # type: ignore[assignment]
+        client.INFO_TABLES_QUERY_THRESHOLD = 0
```
rudolfix, line 858: please use env variable to mock config value

rudolfix (changes requested):
@djudjuu code does what it should but is too hacky. please see review.

also we need docs:
- in athena docs: mention that we have this limit (80)
- in destination docs describe this config option.

in essence we read full information schema and filter tables in code when threshold is exceeded.

rudolfix:
@djudjuu maybe you need to merge current devel here. then tests should run. if not we'll take a look

rudolfix (approved):
LGTM!

PR #2612, opened 2025-05-07
feat: 2540 - incremental datasets

zilto:
This PR leverages adds incremental filters to `ReadableDBAPIDataset` and `ReadableDBAPIRelation`. 

Filters are SQL queries that filter a `target` table based on the `_dlt_load` table (`_dlt_load_id`, load date, load status). It leverages SQLGlot to write queries that are portable across destination backends. 

## Potential API
Here are some potential user-facing API. One or more could be implemented. The real convenience is that it hides the table joins

### 1. Dataset instantiation
  ```python
  latest_dataset = ReadableDBAPIDataset(dataset_name="foo", load_id=[...])
  
  @dlt.transformations()
  def customer_transactions(dataset):
    latest_customers = latest_dataset.customers
    latest_transactions = latest_dataset.transactions
    ...
    return ...
  ```
  
### 2. Dataset-level generic filter
  ```python
  dataset = ReadableDBAPIDataset(dataset_name="foo")
  
  @dlt.transformations()
  def customer_transactions(dataset):
    all_customers = dataset.customers
    latest_transactions = dataset.filter_loads(load_ids=[...], status=..., ).transactions
    ...
    return ...
    
  # also
  latest_dataset = dataset.filter_loads(...)
  latest_customers = latest_dataset.customers
  latest_transactions = latest_dataset.transactions
  ```
  
### 3. Dataset-level named filters
  ```python
  dataset = ReadableDBAPIDataset(dataset_name="foo")
  
  @dlt.transformations()
  def customer_transactions(dataset):
    all_customers = dataset.customers
    latest_transactions = dataset.latest_load.transactions   # dataset property
    ...
    return ...
  ```

### 4. Relation-level named filters
  ```python
  dataset = ReadableDBAPIDataset(dataset_name="foo")
  
  @dlt.transformations()
  def customer_transactions(dataset):
    all_customers = dataset.customers
    all_transactions = dataset.transactions
    latest_transactions = all_transactions.latest_load   # relation property
    ...
    return ...
  ```
  
### 5. Relation-level generic filters
  ```python
  dataset = ReadableDBAPIDataset(dataset_name="foo")
  
  @dlt.transformations()
  def customer_transactions(dataset):
    all_customers = dataset.customers
    all_transactions = dataset.transactions
    latest_transactions = all_transactions.filter_loads(...)
    ...
    return ...
  ```
  
## Implementation
The filter expression can be stored as a SQLGlot expression. This can be composed as a Common Table Expression (CTE) when querying the `ReadableRelation`. Here's some pseudo code

```python
dataset = ReadableDBAPIDataset()

dataset = dataset.filter(status=...)
# internally
load_id_expr = dataset._list_load_ids(status=...)
dataset._filter = load_id_expr

latest_customers = dataset.table("customers")
df = latest_customer.select("id", "name").df()
#  internally
customers_relation = dataset.table("customers", cte=dataset._filter)
select_query = sge.select("id", "name")
# ...
full_query = select_query.with(customers_relation.cte)
df = execute(full_query.sql())
```
  
## Thoughts
I have a general preference for dataset levels filters. This forces the user to do the filtering / selection at the beginning of the query. There's no performance hit if using SQLGlot or Ibis because the full expression can be later optimized

### DIFF
--- a/dlt/destinations/dataset/relation.py
+++ b/dlt/destinations/dataset/relation.py
@@ -123,13 +123,10 @@ class BaseReadableDBAPIRelation(SupportsReadableRelation, WithSqlClient):
             query = sqlglot.transpile(query, read=dialect, write="duckdb")[0]
             dialect = "duckdb"
 
-        # TODO: maybe store the SQLGlot schema on the dataset
         # TODO: support joins between datasets
-        d = self._dataset
-        sqlglot_schema = lineage.create_sqlglot_schema(d.schema)
         return lineage.compute_columns_schema(
             query,
-            sqlglot_schema,
+            self._dataset.sqlglot_schema,
             dialect,
             infer_sqlglot_schema=infer_sqlglot_schema,
             allow_anonymous_columns=allow_anonymous_columns,
@@ -150,10 +147,11 @@ class ReadableDBAPIRelation(BaseReadableDBAPIRelation):
         self,
         *,
         readable_dataset: "ReadableDBAPIDataset",
-        provided_query: Any = None,
+        provided_query: Union[str, sge.Select] = None,
         table_name: str = None,
         limit: int = None,
         selected_columns: Sequence[str] = None,
+        load_ids: Union[Sequence[str], Set[str], None] = None,
     ) -> None:
         """Create a lazy evaluated relation to for the dataset of a destination"""
 
@@ -168,12 +166,24 @@ class ReadableDBAPIRelation(BaseReadableDBAPIRelation):
         self._table_name = table_name
         self._limit = limit
         self._selected_columns = selected_columns
+        self._load_ids = set(load_ids) if load_ids else set()
 
     # TODO can we assume this returns `str`?
     def query(self) -> Any:
         """build the query"""
-        if self._provided_query:
+        if isinstance(self._provided_query, str):
             return self._provided_query
+        elif isinstance(self._provided_query, sge.Select):
+            return self._provided_query.sql()
+        else:
+            expr = self.__prepare_relation_expr()
+            return expr.sql()
+
+    def sql(self) -> str:
+        if isinstance(self._provided_query, str):
+            return self._provided_query
+        elif isinstance(self._provided_query, sge.Select):
+            return self._provided_query.sql()
         else:
             expr = self.__prepare_relation_expr()
             return expr.sql()
--- a/dlt/destinations/dataset/relation.py
+++ b/dlt/destinations/dataset/relation.py
@@ -2,6 +2,7 @@ from typing import Any, Generator, Sequence, Type, Union, TYPE_CHECKING
 from contextlib import contextmanager
 
 import sqlglot
+import sqlglot.expressions as sge
 
 from dlt.common.destination.dataset import (
     SupportsReadableRelation,
@@ -168,37 +169,34 @@ class ReadableDBAPIRelation(BaseReadableDBAPIRelation):
         self._limit = limit
         self._selected_columns = selected_columns
 
+    # TODO can we assume this returns `str`?
     def query(self) -> Any:
         """build the query"""
-        # TODO reimplement this using SQLGLot instead of passing strings
         if self._provided_query:
             return self._provided_query
+        else:
+            expr = self.__prepare_relation_expr()
+            return expr.sql()
 
-        dataset_schema = self._dataset.schema
-
-        table_name = self.sql_client.make_qualified_table_name(
-            dataset_schema.naming.normalize_path(self._table_name)
+    def __prepare_relation_expr(self) -> sge.Expression:
+        qualified_table_name = self.sql_client.make_qualified_table_name(
+            self._dataset.schema.naming.normalize_path(self._table_name)
         )
-
-        maybe_limit_clause_1 = ""
-        maybe_limit_clause_2 = ""
-        if self._limit:
-            maybe_limit_clause_1, maybe_limit_clause_2 = self.sql_client._limit_clause_sql(
-                self._limit
-            )
-
-        selector = "*"
         if self._selected_columns:
-            selector = ",".join(
-                [
-                    self.sql_client.escape_column_name(
-                        dataset_schema.naming.normalize_tables_path(c)
-                    )
-                    for c in self._selected_columns
-                ]
+            selection = (
+                self.sql_client.escape_column_name(
+                    self._dataset.schema.naming.normalize_tables_path(c)
+                )
+                for c in self._selected_columns
             )
+        else:
+            selection = ["*"]
 
-        return f"SELECT {maybe_limit_clause_1} {selector} FROM {table_name} {maybe_limit_clause_2}"
+        query = sge.select(*selection).from_(qualified_table_name)
+        if isinstance(self._limit, int):
+            query = query.limit(self._limit)
+
+        return query
 
     def __copy__(self) -> Self:
         return self.__class__(
--- a/dlt/common/libs/sqlglot.py
+++ b/dlt/common/libs/sqlglot.py
@@ -428,7 +428,7 @@ def _filter_dlt_hints(hints: TColumnSchema) -> TColumnSchema:
         # those are directly expressed via the DataType and Column
         if k in DATA_TYPE_HINTS:
             continue
-        
+
         if k in ALLOWED_HINTS:
             final_hints[k] = v
         elif k.startswith(CUSTOM_HINT_PREFIX):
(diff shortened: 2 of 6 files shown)


dlt/destinations/dataset/relation.py
```diff
@@ -168,37 +169,34 @@ def __init__(
...
-            dataset_schema.naming.normalize_path(self._table_name)
+    def __prepare_relation_expr(self) -> sge.Expression:
+        qualified_table_name = self.sql_client.make_qualified_table_name(
+            self._dataset.schema.naming.normalize_path(self._table_name)
         )
-
-        maybe_limit_clause_1 = ""
-        maybe_limit_clause_2 = ""
-        if self._limit:
-            maybe_limit_clause_1, maybe_limit_clause_2 = self.sql_client._limit_clause_sql(
```
sh-rp, line 186: you can check wether the `limit_clause_sql` is being used in any other place in the codebase, most likely it is not, so you can remove it from all the clients!

zilto: As far as I could tell, this tuple was required only because of the `TOP` statement which needs to be in the `SELECT` portion

sh-rp:
@zilto contrary to the ticket I originally created, I think we need a solution where the load_id or bounds of load_ids that are considered "new" are known on instantiation of the dataset. The reason for this is, that when we run transformations that have some kind of awareness of what is the "new" or "incremental" part of the dataset must be the same for every transformation execution. If we base this on sql_queries that look up these load_ids in the dataset, it could happen that new data arrives while these transformations are executed, and some operate with newer load_ids than others. I hope it is clear what I mean by this.

So in my view a good interface for a first load_id based incremental dataset would be:

```
incremental_dataset = ReadableDBAPIDataset(dataset_name="foo", last_processed_load_id: str, newest_loaded_load_id: str)

# get full items table
incrementat_dataset.items.arrow()

# only get increment part of items table
incremental_dataset.items.new.arrow()
```

Where the new portion of the table is filtered by `item._dlt_load_id > last_processed_load_id` and `item._dlt_load_id <= last_processed_load_id`.

The `last_processed_load_id` could be omitted and taken from the loads table at instantiation of the dataset, but NOT during execution or extraction of the transformation.

This way all transformations that are run are guaranteed to run on the same incremental data. We also need to make sure that we can get the full table as well as the incremental version of any table from the same dataset instance. Some transformations will join incremental data to a full table (think orders joined to customers).

rudolfix (commented):
there's a lot of good stuff here! but what we are lacking is understanding how the most common case of incremental loading will work. we started to discuss it. we have some prior work in our dbt packages but it was never completed.
1. let's do architecture call on this. I'll prepare some input
2. I still have substantial feedback to the user interface and code organizations. but maybe let's do (1) first

also @sh-rp pls recall all the dbt stuff with processed load ids

zilto:
## Problem
For a v1, we should focus on this simple incremental scenario
> "My pipeline runs several time (EL). I want to make sure that all loaded packages are also transformed (T)"

This scenario assumes:
- The goal is to apply T to all loaded packages
- EL and T can happen at different times. 
- We can multiple EL packages processed in a single T execution (e.g., ingest hourly and transform daily)
- We ignore temporal dimension (i.e., dates, timestamps)
- The transformation code / SQL query remains the same

## Solution
Add a  `_dlt_transforms` table that tracks successful transforms. It is keyed on `(transformation_name, _dlt_load_id)`. A transformation is successful only once it has `status==0`. We should track failed T executions in this table if `dev_mode=False`.

When the user wants to Transform data, we can directly lookup `_dlt_transforms` and check if it needs to run or no-op.

image

### User code
Key points for this simple scenario:
- pass load ids to the `dataset()` instantiator
- the `dataset` stores the received load ids in `_dlt_transforms` metadata table
- tables accessed via the `dataset` are globally filtered by default

```python
@dlt.transformation
def sales_by_city(dataset):
  customers = dataset.table("customers")
  sales = dataset.table("sales")
  ...
  return result
  
pipeline = dlt.pipeline(...)
load_info = pipeline.run(sql_database(...))
dataset = pipeline.dataset(load_ids=load_info.loads_ids)
pipeline.run(sales_by_city(dataset))
```
This simple approach seems intuitive. When `dataset(load_ids=None)`, we process all available load_ids and mark them as processed in `_dlt_transforms`

### Accessing full tables in incremental loads
This is a common pattern
> Data in the `sales` table is loaded hourly, but not the `customers` table. When doing Transform, we want to join the incremental `sales` with the full `customers` table. 

From a user perspective, it makes more sense to specify when they want to retrieve the `full` table than have to specify when they want the `incremental` one (otherwise, you wouldn't be using incremental via `dataset(load_ids=...)`.

```python
@dlt.transformation
def sales_by_city(dataset):
  customers = dataset.table("customers", full=True)
  sales = dataset.table("sales")
  ...
  return result
  
pipeline = dlt.pipeline(...)
load_info = pipeline.run(sql_database(...))
dataset = pipeline.dataset(load_ids=load_info.loads_ids)
pipeline.run(sales_by_city(dataset))
```

If the user accesses the "full dataset" via `pipeline.dataset()` instead of the incremental dataset, the `full` kwarg in `dataset.table("customers", full=True)` is a no-op. This seems less ambiguous than having `dataset.table("customers", incremental=True)` when the full dataset is select, but should be equivalent.

### Backfills
For "backfills", we delete records in `_dlt_transforms` to force the engine to re-process some load packages. We should really use SCD2 instead of deleting records to track backfills or "retrigger"

## Future work
### Delete loaded transform
It would be useful to have utilities to delete all transformation records associated with a `_dlt_load_id` (this is also true for resources)

### Time-based incremental
To support time-based incremental, we filter the `_dlt_loads` table based on date. Then, we join with `_dlt_transforms` to determine which `load_ids` to select. The same logic allows for time-based backfills.

Note, there's ambiguity between "incremental based on the time the data was loaded by dlt" and "incremental based on the timestamp of the record (i.e., a date column)"

### Run `resource` and `transformation` in the same pipeline
The following API should be possible. We internally process all resources, then apply the transformations with the most recent load_id
```python
pipeline = dlt.pipeline(...)
pipeline.run(sql_database(...), customers_by_city)  # transformation is not called `()`
```

### Fingerprinting transform
All the mechanisms above assume the transformation query doesn't change. Now image there was a bug in `sales_by_city()` and we just fixed it. We would need to drop the `_dlt_transforms` and the associated rows **for that specific transform** before executing the new query. This could be automated if we find a diff between queries

rudolfix:
@zilto we'll surely need such table at least to track "package lineage" form input dataset to output dataset (where, as you mention, single transformation package may process many input packages).

what I do not get is why you place this table in the input dataset. my top level thinking here:
- input dataset should not be aware that any transformations run on it. running transformations should not change input dataset
- in many cases such dataset will be physically read only from the POV of transformation layer.

there are many other reasons not to track loads in the input dataset ie. to be able to run the same transformation source (group) many times but to different datasets/destinations (ie. testing, ci, different settings)

it may sound like a heresy but I'd go for pipeline state for the initial implementation.
1. the only thing you need to store in the state is the max(`inserted_at`) from `_dlt_loads` in the input dataset
2. when you run the pipeline you retrieve the previous max_inserted_at from the state
3. you select the load ids to process: `SELECT load_id, inserted_at from _dlt_loads WHERE inserted_at > ?max_inserted_at AND ?inserted_at < now()  AND status = 0 and schema_name = ?dataset_schema_name` this also ensures that we work with datasets shared by many pipelines and sources/schemas
4. `now()` in the query above becomes new state. we do not even need access to output dataset to get those load_ids (if we persist pipeline working folder)
```py
table = new_table(
        LOADS_TABLE_NAME,
        columns=[
            {"name": "load_id", "data_type": "text", "nullable": False},
            {"name": "schema_name", "data_type": "text", "nullable": True},
            {"name": "status", "data_type": "bigint", "nullable": False},
            {"name": "inserted_at", "data_type": "timestamp", "nullable": False},
            {
                "name": "schema_version_hash",
                "data_type": "text",
                "nullable": True,
            },
        ],
    )
```

that's it, we do not need to implement anything else. some remarks:
- this obviously deals with `refresh`, `drop` and `dev_mode` automatically. so full refresh is handled
- we should store `max_inserted_at` in the resource state so we can drop/refresh single resources!
- complication: avoid races. when we run pipeline, all resources must receive the same `now()`

hmmm this looks too simple :) what I'm missing here?

**next steps**
1. implement append only `_dlt_transformation` in the output dataset
2. create `dbt` scaffolding package
3. discuss user interface

zilto:
closing in favor of #2716

zilto:
closing in favor of https://github.com/dlt-hub/dlt/pull/2716

PR #2614, opened 2025-05-07
docs: Fix incorrect nesting in secrets.toml

agrueneberg:
There is one `credentials` too many.


sh-rp:
@agrueneberg thanks :)

PR #2616, opened 2025-05-08
Feat/2609-clickhouse-precision-integer-mapping

### LINKED ISSUE #2609: ClickHouse: Support precision-based integer type mapping
### Feature description

Currently, `bigint` maps to `Int64` and doesn't take `precision` into account. A quick glance suggests that all that is needed is overriding `to_db_integer_type`?

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

- Reduce data storage requirements

### Proposed solution

- Use `precision` to create smaller integer values

### Related issues

_No response_

hsm207:
### Description

Add support for all of [Clickhouse's integer precision](https://clickhouse.com/docs/sql-reference/data-types/int-uint#integer-ranges).

### Related Issues

- Fixes #2609
- Closes #2609
- Resolves #2609

### Additional Context
This PR gives users the option to map `bigint` to Int8, Int16, etc using the `precision` hint to optimise storage.

Currently, `bigint` only maps to Int64.

### DIFF
--- a/dlt/destinations/impl/clickhouse/factory.py
+++ b/dlt/destinations/impl/clickhouse/factory.py
@@ -10,6 +10,7 @@ from dlt.common.data_writers.escape import (
 )
 from dlt.common.destination import Destination, DestinationCapabilitiesContext
 
+from dlt.common.destination.typing import PreparedTableSchema
 from dlt.common.exceptions import TerminalValueError
 from dlt.common.schema.typing import TColumnSchema, TColumnType
 from dlt.destinations.type_mapping import TypeMapperImpl
@@ -92,7 +93,7 @@ class ClickHouseTypeMapper(TypeMapperImpl):
 
         return super().from_destination_type(db_type, precision, scale)
 
-    def to_db_integer_type(self, column: TColumnSchema, table: Any = None) -> str:
+    def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema = None) -> str:
         """Map integer precision to the appropriate ClickHouse integer type."""
         precision = column.get("precision")
         if precision is None:
--- a/dlt/destinations/impl/clickhouse/factory.py
+++ b/dlt/destinations/impl/clickhouse/factory.py
@@ -10,6 +10,7 @@ from dlt.common.data_writers.escape import (
 )
 from dlt.common.destination import Destination, DestinationCapabilitiesContext
 
+from dlt.common.exceptions import TerminalValueError
 from dlt.common.schema.typing import TColumnSchema, TColumnType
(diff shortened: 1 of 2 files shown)


hsm207:
@rudolfix  I noticed the pipeline failed on the following lines:

* dlt/helpers/ibis.py:24

* tests/load/test_read_interfaces.py:626

These are both mypy errors ([unused-ignore]) caused by # type: ignore comments that no longer appear necessary. Since I didn't modify these files in this PR, just wanted to flag it for visibility. 

Let me know if you'd like me to clean them up as part of this PR!

dlt/destinations/impl/clickhouse/factory.py
```diff
@@ -91,6 +91,23 @@ def from_destination_type(
 
         return super().from_destination_type(db_type, precision, scale)
 
+    def to_db_integer_type(self, column: TColumnSchema, table: Any = None) -> str:
```
agrueneberg, line 94: Have you looked into `PreparedTableSchema` instead of `Any` here?

dlt/destinations/impl/clickhouse/factory.py
```diff
@@ -91,6 +91,23 @@ def from_destination_type(
...
+            return "Int64"
+        if precision <= 8:
+            return "Int8"
+        if precision <= 16:
+            return "Int16"
+        if precision <= 32:
+            return "Int32"
+        if precision <= 64:
+            return "Int64"
+        raise ValueError(
```
agrueneberg, line 107: Implementations for other databases tend to raise a `TerminalValueError` here.

hsm207: good catch!

hsm207 (commented):
> Have you looked into `PreparedTableSchema` instead of `Any` here? 

Thanks for catching this!

dlt/destinations/impl/clickhouse/factory.py
hsm207: Thanks for catching this

rudolfix (changes requested):
code that generates precision integers LGTM!
still we need to do a few things:
1. @hsm207 please resubmit this PR directly to `dlt` repo: then all CI will run for you without any special labels etc. you have write rights
2. when you do that you should get errors in clickhouse table builder tests (where Int64 was always generated, no matter the type). this needs to be correcred
3. you should recognize new types in reverse operation: `from_destination_type` - at least mark them as `bigint`. currently they will be inferred as text

thanks for contributing!

hsm207:
resubmitted in #2627

PR #2619, opened 2025-05-09
Detect wether query just filters rows or is more complex with sqlglot

### LINKED ISSUE #2557: Detect wether query just filters rows with sqlglot
### TLDR

For deciding which hints to preserve during our lineage / column schema detection, we need to know if a given sql query is just filtering a subset of rows with all columns from a given table or doing something more advanced (subset of columns, joins, aggregates..). For the simple filtering we can keep all hints, for the more advanced queries we need to remove primary_key and unique and other hints, since we don't really know how those columns should be used.

### The main task
We need a small sqlglot based util that can analyze a given query and decide wether this is a simple query or a complex query.

### Examples for a table with three columns col1, col2, col3
`select * from table` -> simple, only one table is accessed and all columns are there
`select col1, col2 from table` -> complex as not all columns are there
`select t1.col1 as blah, t1.col2 as blubb, t1.col3 from table as t` -> simple, only one table, all columns are there
`select * from t1 join t2 on t1.id =t2.t1_id `-> complex, it's a join
`select * from t1 where t1.id > 5 offset 5 limit 10` -> simple, only one table, all columns are there
`select *, static_val = "123" from table` -> debatable, but I would say simple, only one table is accessed and all columns are there.

anuunchin:
### Description
This is a research PR with an implementation of a utility function that analyzes a select query and detects whether it's complex or not. 

### Related Issues

- Resolves #2557 

### Additional Notes:

- A union all statement is parsed as a Union expression rather than a select one.
- Queries with CTEs are flagged as complex no matter the underlying structure.

### DIFF
--- a/dlt/common/libs/sqlglot.py
+++ b/dlt/common/libs/sqlglot.py
@@ -1,14 +1,12 @@
-from typing import Optional
+from typing import Optional, Union, Set
 
 from dlt.common.utils import without_none
 from dlt.common.exceptions import MissingDependencyException, TerminalValueError
 from dlt.common.schema.typing import TColumnType, TDataType, TColumnSchema
 
-try:
-    import sqlglot.expressions as sge
-    from sqlglot.expressions import DataType, DATA_TYPE
-except ModuleNotFoundError:
-    raise MissingDependencyException("dlt sqlglot helpers", ["sqlglot"])
+import sqlglot.expressions as sge
+from sqlglot.expressions import DataType, DATA_TYPE
+from sqlglot.optimizer.scope import build_scope
 
 
 SQLGLOT_TO_DLT_TYPE_MAP: dict[DataType.Type, TDataType] = {
@@ -544,3 +542,86 @@ def _filter_dlt_hints(hints: TColumnSchema) -> TColumnSchema:
             final_hints[k] = v
 
     return final_hints  # type: ignore[return-value]
+
+
+def query_is_complex(
+    parsed_select: Union[sge.Select, sge.Union],
+    columns: Set[str],
+) -> bool:
+    """
+    Return **True** unless the query is provably “simple”.
+    A *simple* query
+    1. references **exactly one** physical table,
+    2. contains no complex constructs (CTEs, sub-queries, derived tables,
+       unions, window functions, GROUP BY, DISTINCT, etc.),
+    3. projects either
+         • a plain/qualified star with only constant literals after it, or
+         • the full, explicit list of *all* columns with only constant
+           literals after it.
+    Anything we cannot *prove* to be simple is conservatively flagged
+    as complex.
+    """
+    root_scope = build_scope(parsed_select)
+    tables: list[sge.Table] = []
+    non_literal_cols: dict[str, int] = {}
+    star_present = False
+    for node in root_scope.walk():
+        if isinstance(node, sge.Table):
+            tables.append(node)
+        if isinstance(node, sge.Window):
+            return True
+        if isinstance(node, sge.SetOperation):
+            return True
+        if isinstance(node, sge.With):
+            return True
+        if isinstance(node, sge.Distinct):
+            return True
+        if isinstance(node, sge.Group):
+            return True
+        if isinstance(node, sge.DPipe):
+            return True
+        # Detect a star in the outermost projection list.
+        # A star can be either “*” or a qualified “t.*”.
+        # In the AST:
+        #   • bare  “*”    Star -> SELECT
+        #   • qual. “t.*”  Star -> Column -> SELECT
+        if isinstance(node, sge.Star) and (
+            node.parent == parsed_select or (node.parent and node.parent.parent == parsed_select)
+        ):
+            star_present = True
+        # Detect plain (aliased) column references that belong to the outermost projection list.
+        if isinstance(node, sge.Column) and not isinstance(node.this, sge.Star):
+            # it's an (aliased) column in our outer select
+            if node.parent == parsed_select or (
+                isinstance(node.parent, sge.Alias) and node.parent_select == parsed_select
+            ):
+                non_literal_cols[node.name] = non_literal_cols.get(node.name, 0) + 1
+        # An aliased expression that is *not* just a column or a
+        # literal disqualifies the query from being “simple
+        if isinstance(node, sge.Alias):
+            if not (isinstance(node.this, sge.Column) or isinstance(node.this, sge.Literal)):
+                return True
+    if len(tables) > 1:
+        return True
+    if star_present and not non_literal_cols:
+        return False
+    non_literal_cols_names = set(non_literal_cols.keys())
+    # if any column is selected twice, it is complex
+    if any(count > 1 for count in non_literal_cols.values()):
+        return True
+    # if selected columns are exactly the same as the table columns, it is simple
+    if non_literal_cols_names == columns:
+        return False
+    # if newly added columns are all exclusively dlt columns, it is simple
+    if (
+        all(col_name.startswith("_dlt_") for col_name in non_literal_cols_names - columns)
+        and non_literal_cols_names - columns != set()
+    ):
+        return False
+    # if missing columns are all exclusively dlt columns, it is simple
+    if (
+        all(col_name.startswith("_dlt_") for col_name in columns - non_literal_cols_names)
+        and columns - non_literal_cols_names != set()
+    ):
+        return False
+    return True
--- a/tests/load/pipeline/test_query_complexity_analyzer.py
+++ b/tests/load/pipeline/test_query_complexity_analyzer.py
@@ -1,0 +1,171 @@
+from typing import Any, Dict, List, Union
+import pytest
+from tests.load.utils import destinations_configs
+import sqlglot
+
+VARIOUS_QUERIES: List[Dict[str, Union[str, bool]]] = [
+    {"description": "star select", "query": "SELECT * FROM my_table", "complex": False},
+    {
+        "query": "SELECT t.* FROM my_table AS t",
+        "complex": False,
+        "description": "star select + aliased table",
+    },
+    {
+        "query": "SELECT * FROM my_table ORDER BY col1",
+        "complex": False,
+        "description": "star select + order",
+    },
+    {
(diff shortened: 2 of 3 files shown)


dlt/common/utils.py
```diff
@@ -661,3 +666,89 @@ def is_typeerror_due_to_wrong_call(exc: Exception, func: AnyFun) -> bool:
 removeprefix = getattr(
     str, "removeprefix", lambda s_, p_: s_[len(p_) :] if s_.startswith(p_) else s_
 )
+
+
+def query_is_complex(
+    parsed_select: Union[sqlglot.exp.Select, sqlglot.exp.Union],
+    columns: Set[str],
```
anuunchin, line 673: This will probably need to be TTableSchemaColumns 👀

dlt/common/utils.py
```diff
@@ -661,3 +666,89 @@ def is_typeerror_due_to_wrong_call(exc: Exception, func: AnyFun) -> bool:
...
+        return True
+
+    # star + only literals
+    if star_present and not non_literal_cols:
+        return False
+
+    if non_literal_cols == columns:
+        return False
+
+    return True
```
anuunchin, line 754: Another thing I tried here is using sqlglot's diff function which outputs what needs to be done to create a given query from the base query. In our case, the base query could be `SELECT * FROM my_table`, or `SELECT <all_columns explicitly> FROM my_table` and we can check if the required actions involve removing or adding a column and some other allowed actions  - but this I realized might not be as straightforward as simply going through the parsed query

tests/load/pipeline/test_query_complexity_analyzer.py
```diff
@@ -0,0 +1,174 @@
...
+        "complex": False,
+        "description": "full select + literal",
+    },
+    {
+        "query": (
+            "WITH my_table AS (SELECT * FROM my_table WHERE col1 = True) SELECT * FROM my_table;"
+        ),
+        "complex": True,
+        "description": "cte + one table ref",
+    },
```
anuunchin, line 72: this query is currently resolved as complex, ideally it should be simple, but I'm wondering if we allow CTEs at all

dlt/common/utils.py
```diff
@@ -661,3 +666,89 @@ def is_typeerror_due_to_wrong_call(exc: Exception, func: AnyFun) -> bool:
...
+
+    Args:
+        parsed_select (sqlglot.exp.Select): The parsed SELECT statement.
+        columns Set[str]: Column names of the table.
+
+    Returns:
+        bool: Whether a query is considered complex.
+    """
+    # 1. If more than one table is referenced -> complex
+    tables = {table for table in parsed_select.find_all(sqlglot.exp.Table)}
```
zilto, line 696: I remembered reading in the [SQGlot AST primer](https://github.com/tobymao/sqlglot/blob/main/posts/ast_primer.md#walk-methods) that `find()`, `find_all()` and `walk()` are not always reliable

> Here's a common pitfall of the walk methods:
> ```py
> ast.find_all(exp.Table)
> ```
> At first glance, this seems like a great way to find all tables in a query. However, Table instances are not always tables in your database. Here's an example where this fails:
> ```py
> ast = parse_one("""
> WITH x AS (
>   SELECT a FROM y
> )
> SELECT a FROM x
> """)
> 
> # This is NOT a good way to find all tables in the query!
> for table in ast.find_all(exp.Table):
>     print(table)
> 
> # x  -- this is a common table expression, NOT an actual table
> # y
> ```

The post follows with how to use the `Scope` object to traverse the AST. At each "node" of the scope (think of traversing a graph), you can directly query the `.stars`, `.tables`, etc. properties to retrieve what's in the scope.

anuunchin: @zilto Do we take into account only physical tables? I just thought if the query already has multiple table expressions, no matter whether each of them corresponds to an actual table or not, we still consider the query complex - because it's complex in terms of syntax if not in terms of semantics 👀

anuunchin: Added the sqlglot example just in case: 🫡 
```
    {
        "query": "WITH temp_table AS (SELECT * FROM my_table) SELECT * FROM temp_table;",
        "complex": True,
        "description": "cte + two table refs",
    },
    {
        "query": "WITH x AS (SELECT a FROM y) SELECT a FROM x",
        "complex": True,
        "description": "cte + two table refs + sqlglot example",
    },
```

zilto: I meant to say that the resource I linked proposes an approach that could simplify our code. Manipulating instances of the [Scope class](https://github.com/tobymao/sqlglot/blob/56da9629899e72ab1e15cfc45ede838c4c38c16e/sqlglot/optimizer/scope.py#L27) might be more convenient than manipulating raw `Expression` objects. `Scope` has many useful properties

Probably something like this:

```python
from sqlglot.optimizer.scope import build_scope

ast = parse_one("""
WITH x AS (
  SELECT a FROM y
)
SELECT a FROM x
""")

root = build_scope(ast)
# `.traverse()` recursively walks the graph
for scope in root.traverse():
    if scope.is_cte:
       return True
       
    if scope.is_derived_table:
       return True
       
    if scope.is_subquery:
       return True
       
    if scope.is_union:
       return True
       
    if scope.pivots:
       return True
       
    if scope.joint_hints:
       return True

   # and more
```

tests/load/pipeline/test_query_complexity_analyzer.py
sh-rp: Let's keep it like this for now.

tests/load/pipeline/test_query_complexity_analyzer.py
```diff
@@ -0,0 +1,180 @@
...
+    {
+        "query": "SELECT * FROM my_table UNION ALL SELECT * FROM my_table",
+        "complex": True,
+        "description": "union",
+    },
+]
+
+
+@pytest.mark.parametrize(
```
sh-rp, line 148: I think you can remove this. It is enough to test with duckdb, alternatively you would have to supply example queries for each dialect and our assumption should be that sqlglot works the same for each language (and gives good errors if a feature is not available for one of the languages)

dlt/common/utils.py
```diff
@@ -661,3 +665,168 @@ def is_typeerror_due_to_wrong_call(exc: Exception, func: AnyFun) -> bool:
...
+            ):
+                return True
+
+    if len(tables) > 1:
+        return True
+
+    if star_present and not non_literal_cols:
+        return False
+
+    if non_literal_cols == columns:
```
sh-rp, line 829: I think we should also return `False` if the missing columns are one or many of the `_dlt` columns. So the user may drop the dlt_id or dlt_load_id and still retain all hints. This thought just occured to me. The condition can be col_name.lower().starts_with("_dlt") to detect if the missing column should not lead to a complex query.

dlt/common/utils.py
```diff
@@ -661,3 +665,168 @@ def is_typeerror_due_to_wrong_call(exc: Exception, func: AnyFun) -> bool:
 removeprefix = getattr(
     str, "removeprefix", lambda s_, p_: s_[len(p_) :] if s_.startswith(p_) else s_
 )
+
+
+def query_is_complex(
```
sh-rp, line 670: I think this should go into `dlt/common/libs/sqlglot.py`, this is where @zilto has put the sqlglot stuff on the transformations branch.

dlt/common/utils.py
sh-rp: I'm not quite sure, maybe this will be an SQLGlotSchema, but let's keep a list of known columns for now and we can later change it when we use this code in the transformations work.

dlt/common/utils.py
sh-rp: Additionally I think this will fail if all columns are there but in a different order, so you should test that too.

tests/load/pipeline/test_query_complexity_analyzer.py
```diff
@@ -0,0 +1,180 @@
+from typing import Any, Dict, List, Union
+import pytest
+
+from tests.load.utils import destinations_configs
+
+import sqlglot
+
+
+VARIOUS_QUERIES: List[Dict[str, Union[str, bool]]] = [
+    {"description": "star select", "query": "SELECT * FROM my_table", "complex": False},
```
sh-rp, line 10: I would add the following tests:
* Select all columns but in a different order than they appear in the original table (should **not** be marked as complex)
* Select all columns but one column twice (should be marked as complex)
* Select in a way that all column names from the orignal column are there, but actually one of the columns is an alias of a different column: SELECT col1, col2, col2 as col3" should be marked as complex.

dlt/common/utils.py
```diff
@@ -661,3 +665,168 @@ def is_typeerror_due_to_wrong_call(exc: Exception, func: AnyFun) -> bool:
...
+
+    root_scope = build_scope(parsed_select)
+
+    tables: list[sqlglot.exp.Table] = []
+    non_literal_cols: set[str] = set()
+
+    star_present = False
+
+    for node in root_scope.walk():
+        if isinstance(node, sqlglot.exp.Table):
```
sh-rp, line 784: would it maybe make since to have a whitelist of expressions (ones that pass this test) rather than a blacklist? just an idea.

sh-rp (changes requested):
This is very good. I think you can keep the scope approach and fix the few things I have highlighted.

dlt/common/utils.py
anuunchin: @sh-rp  This will potentially be a very long list. i.e the WHERE statement in SELECT * FROM t1 WHERE t1.id > 5 OFFSET 5 LIMIT 10 includes the following nodes:

        sqlglot.exp.Where,
        sqlglot.exp.And,
        sqlglot.exp.GT,
        sqlglot.exp.Not,
        sqlglot.exp.Is,
        sqlglot.exp.Null,

but then again, the blacklist will porbably also include a lot more additional things 👀

sh-rp (approved):
very nice, thank you :)

dlt/common/libs/sqlglot.py
```diff
@@ -7,6 +7,7 @@
 try:
```
sh-rp: we can move this try catch block now, sqlglot is now a main dependency of dbt and will always be availble

PR #2623, opened 2025-05-12
[WIP] feat/2531 sane defaults in config and secrets

djudjuu:
`poetry run dlt init filesystem filesystem` to try it out

## the good
- once a flavor is given, its easy to use it to select the correct type from a Union to write down example values

## the bad
 - getting the flavors out of the code (to let the user choose) is messy, because there are many type-resolving functions that are all not aware of the flavor-logic (which needs to be used somewhere deep inside when a union-type-hint is resolved into a single hint
 
 
## open:
 - bucket_url is just of type string, so its not easy to show the different ways it can be set up, even thought that would most interesting for filesystem, as it is used to determine the protocol and thus the expected credentials
 
## thoughts:
 - our examples/flavors should be in line with the docs. I think they are quite good. 
 https://dlthub.com/docs/dlt-ecosystem/verified-sources/filesystem/basic
 
 ## open questions / todo
 - how does it look to pass the flavors from the decorators through to the config_gen_annotations

### DIFF
--- a/dlt/cli/config_toml_writer.py
+++ b/dlt/cli/config_toml_writer.py
@@ -130,7 +130,8 @@ def write_values(
     toml: TOMLContainer,
     values: Iterable[WritableConfigValue],
     overwrite_existing: bool,
-    config_flavor: str = None,
+    source_flavor: str = None,
+    destination_flavor: str = None,
 ) -> None:
     # TODO: decouple writers from a particular object model ie. TOML
     for value in values:
@@ -143,6 +144,7 @@ def write_values(
             else:
                 toml_table = toml_table[section]  # type: ignore
 
+        config_flavor = source_flavor if "sources" in value.sections else destination_flavor
         write_value(
             toml_table,
             value.name,
--- a/dlt/cli/config_toml_writer.py
+++ b/dlt/cli/config_toml_writer.py
@@ -59,6 +59,7 @@ def write_value(
     overwrite_existing: bool,
     default_value: Any = None,
     is_default_of_interest: bool = False,
+    config_flavor: str = None,
 ) -> None:
     # skip if table contains the name already
     if name in toml_table and not overwrite_existing:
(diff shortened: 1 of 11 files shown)


dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -24,6 +24,20 @@ class FilesystemDestinationClientConfiguration(FilesystemConfigurationWithLocalF
...
+        "minio": {},
+        "gcs": {},
+        "gcp-oauth2": {},
+        "az-token": {},
+        "az-client": {},
+        "az-token": {},
+        "sftp": {},
+        "sftp-password": {},
+    }
+
```
djudjuu, line 40: this is ugly. its a hack because I couldnt find a smarter way to get the flavors out of the code in order to ask the user to select them

dlt/cli/init_command.py
```diff
@@ -746,3 +778,49 @@ def init_pipeline_at_destination(
...
+
+
+# def inform_about_config_flavor_message(required_secrets: Sequence[WritableConfigValue], section: str) -> None:
+#     # get all unique secret hints and print their config flavor menu's
+#     possible_choices = set(secret.hint for secret in required_secrets if section in secret.sections)
+#     for hint in possible_choices:
+#         print_config_flavor_menu_from_union(hint)
+
+
+def check_for_config_flavors(config_values: Dict[str, WritableConfigValue], section: str) -> str:
```
djudjuu, line 790: this is hacky code. I thought it would be simpler to find out if a config-hint class has flavors.

dlt/cli/config_toml_writer.py
```diff
@@ -99,7 +101,15 @@ def write_spec(
...
 
         # check if field is of particular interest and should be included if it has default
-        is_default_of_interest = name in config.__config_gen_annotations__
+        fields_to_include = config.__config_gen_annotations__
+        if isinstance(config.__config_gen_annotations__, dict):
+            if config_flavor:
+                fields_to_include = config.__config_gen_annotations__.get(config_flavor, {})
+                default_value = config.__config_gen_annotations__.get(config_flavor, {}).get(
+                    name, default_value
+                )
```
djudjuu, line 110: easy way to use the example values if flavor is known

PR #2630, opened 2025-05-12
QoL: improved `__repr__()` for public interface

zilto:
This PR adds tailor-made `__repr__()` for the public interface. 

### Description
Python objects have 2 two primary string render methods:
- `__repr__()` should give a representation of the Python object. It's mostly used in interactive sessions, notebooks, and debugger sessions. Ideally, copy pasting the `__repr__()` output should allow to instantiate the same object
- `__str__()` provides a useful user-facing representation. It is triggered by `print()`

### Notes
- Objects like `DltResource` and `DltSource` should not be instantiated directly. Therefore, the `__repr__()` renders their decorator instantiator `@dlt.resource()` and `@dlt.source()`
  - Alternative: render `dlt.resource(..., kwarg1="foo", kwarg2="bar")`. The user can then replace `...` with data or a callable. This might give a better user error
  - OTOH, `@dlt.resource(...)` is the kind of code we would want LLMs to write
- Using `< >` to delimit `__repr__()` is common, but it makes nested representations harder to read e.g.,
  ```py
  duckdb_destination.__repr__()
  # <dlt.destinations.duckdb(config_providers=[<EnvironmentProvider(...)>, <TomlProvider(...)>)>
  ```
- I didn't exactly follow the "everything is copy-pastable convention". For example, I substituted the full `columns=...` kwargs by `columns={...}` when it is present. 
- We should give attention to the order of the rendered kwargs with the most useful / identifiable at the beginning
- Default values and `None` are often explicitly ignored to reduce visual noise


dlt/common/configuration/accessors.py
```diff
@@ -13,6 +13,16 @@
 
 
 class _Accessor(abc.ABC):
+    # TODO should represent the resolution order
+    def __repr__(self) -> str:
+        s = "dlt.secrets("
```
zilto, line 18: oops, this should properly distinguish `dlt.secrets` and `dlt.config`

dlt/pipeline/pipeline.py
```diff
@@ -920,6 +920,23 @@ def last_trace(self) -> PipelineTrace:
             return self._last_trace
         return load_trace(self.working_dir)
 
+    def __repr__(self) -> str:
```
sh-rp, line 923: We have a similar system for saving pipeline state in `_props_to_state` of the pipeline, but unfortunately the vars it produces do not match the pipeline constructor. You could use it though to extract the relevant variables into the pipeline state object and build a __repr__ string that can actually be executed:

dlt.pipeline(
destination=dlt.destinations.<destination_type>(<destination_name),
...)

I think the pipeline class is the only one where a "proper" repr string that you can run with eval can be created, and maybe this even makes sense if people want to create and inspect a pipeline and its dataset from a stacktrace. Wdyt? The test of this would also run the eval to check that it works. Or is this overkill?

zilto: I think the `__repr__` should be optimized for interactive sessions (python debugger, ipython, notebooks, etc.). It should minimally tell the user the "identity of the object" and some key properties.

After a quick review, most Python libraries do not provide `__repr__` that can be evaluated (numpy, pandas, scikit-learn, pyarrow, airflow,  requests). I think keeping the `< >` in the `__repr__` string highlights that they can't be evaluated (i.e., if the string could be evaluated, I would strip the `< >`)

dlt/common/schema/schema.py
```diff
@@ -676,6 +676,16 @@ def tables(self) -> TSchemaTables:
     def settings(self) -> TSchemaSettings:
         return self._settings
 
+    def __repr__(self) -> str:
+        kwargs = {
+            "name": self.name,
+            "version": self.version,
+            "data_tables": self.data_table_names(),
```
sh-rp, line 683: I think I would just do "tables" and list all tables here

sh-rp (changes requested):
This is very cool, I have a few small notes and question, I would also add a `__repr__` for destinations and datasets, these also are important entities that may show up in traces. Let me know what you think.

sh-rp (approved):
approved, we will do dataset in a follow-up

PR #2633, opened 2025-05-13
QoL: warning with hint to provide data types for columns with exclusively None values

### LINKED ISSUE #2632: Raise an appropriate warning when a column has exclusively None values and is essentially dropped
Since `dlt` is data-driven, it cannot infer the data type for a column if all values are `None`, unless the user explicitly provides a type hint. This *silent dropping* has caused confusion for users before.

Because it is impossible to infer the data type from `None`, we now raise an explicit warning that lists the columns which could not be typed and were therefore *silently dropped*.

The warning should come from the normalizers.

anuunchin:
### Description
This PR adds warning to normalizers if the data type of a specific column couldn't be inferred due to all values being None. 
The warning includes guidance to explicitly provide type hints.

### Related Issues

- Resolves #2632


zilto:
The pyarrow data loader also has logic to skip null columns. 

- **from SQL to arrow**. [This code](https://github.com/dlt-hub/dlt/blob/a0c6ea17bed178bb941ca1343fe2759a6603ec52/dlt/common/libs/pyarrow.py#L641) skips null columns when loading an SQL source if the `data_type` isn't specified by the user and can't be inferred.

- **from data items to arrow** (i.e., resource). [This code](https://github.com/dlt-hub/dlt/blob/a0c6ea17bed178bb941ca1343fe2759a6603ec52/dlt/common/libs/pyarrow.py#L1022) skips columns that only include nulls and don't have an explicit `data_type` because we can't infer the dlt `data_type`.

Testing around this PR should probably check the different load format (items, pyarrow, model). 

(I don't remember the exact order of steps, so maybe this comment is irrelevant)

tests/libs/pyarrow/test_pyarrow.py
```diff
@@ -446,3 +454,44 @@ def test_is_arrow_item(pa_type: Any) -> None:
...
+
+    if is_none:
+        logger_spy.assert_called_once()
+        expected_warning = (
+            "Null type(s) detected for column(s) ['col1'] in table 'my_resource' of schema"
+            " 'my_source'. Unless type hint(s) are provided, the column(s) will not be loaded."
+        )
+        assert expected_warning in logger_spy.call_args_list[0][0][0]
+    else:
+        logger_spy.assert_not_called()
```
anuunchin, line 497: I'm not sure if the tests should go where they currently are 👀

rudolfix:
This is tricky ticket because we modify really old code. You do changes in right places but I'll do that more systematically. We could add "none" type to `dlt` (Python has one). But that is too complicated so instead:
1. I'd still create columns for null values but without data types. those are called unbounded columns and will not be materialized in the destination. but still they can have all the annotations etc. Unbounded columns are implemented and tested
2. if we create column via normalizer as above we add special marker (`x-normalizer-seeb-null-first`: true)

@zilto comments: 
1. IMO we should let NULL columns so this needs to be removed https://github.com/dlt-hub/dlt/blob/a0c6ea17bed178bb941ca1343fe2759a6603ec52/dlt/common/libs/pyarrow.py#L1022 -downstream code will deal with this
2. this hopefully will not affect us: https://github.com/dlt-hub/dlt/blob/a0c6ea17bed178bb941ca1343fe2759a6603ec52/dlt/common/libs/pyarrow.py#L641

Implementation
1. handle arrow NULL type in `get_column_type_from_py_arrow` and return unbounded column (but with our marker)
3. create unbounded column in `schema.py`
```py
def _coerce_null_value(
        self, table_columns: TTableSchemaColumns, table_name: str, col_name: str
    ) -> None:
        """Raises when column is explicitly not nullable"""
        if col_name in table_columns:
            existing_column = table_columns[col_name]
            if not utils.is_nullable_column(existing_column):
                raise CannotCoerceNullException(self.name, table_name, col_name)
```
here's a snippet that does it:
```py
inferred_column = self._infer_column(
                col_name, v, data_type=col_type
            )
```
(obviously v is None and col type is type(None) so you need to fix _infer_column

4. warn all columns without data types and our marker: `def verify_normalized_table(` in normalize. this is central point that every item_normalizer works with

Tests:
On top of tests that you have - we need a test for `sql_database` source where NULL only column is passed along and warned at the end. maybe you can add column to `_make_precision_table` that has always NULL in mock data?
also remember that you must use "minimal" reflection level: otherwise dlt will lock the data type from sql alchemy

tests/load/sources/sql_database/test_sql_database_source.py
```diff
@@ -852,6 +853,41 @@ def test_all_types_no_precision_hints(
     )
 
 
+@pytest.mark.parametrize("backend", ["pyarrow", "sqlalchemy"])
```
anuunchin, line 856: pandas packend automatically treats None, or N/A values as text 👀

dlt/common/destination/utils.py
```diff
@@ -226,7 +227,9 @@ def prepare_load_table(
         if table_format == "native":
             # no special table format, use whatever is supported
             prep_table["table_format"] = None
-
+        # remove incomplete columns
+        for column, _ in find_incomplete_columns(table):
+            prep_table["columns"].pop(column["name"], None)
```
rudolfix: why do you need `pop(..., None)` here? column with `name` 100% exists. were you getting errors?

dlt/common/schema/schema.py
```diff
@@ -780,11 +781,19 @@ def set_schema_contract(self, settings: TSchemaContract) -> None:
...
-        column_schema = TColumnSchema(
-            name=k,
-            data_type=data_type or self._infer_column_type(v, k),
-            nullable=not self._infer_hint("not_null", k),
-        )
+        # return unbounded table
+        if v is None and data_type is None:
+            column_schema = TColumnSchema(
+                name=k,
+                nullable=True,
```
rudolfix, line 804: IMO this should be `not self._infer_hint("not_null", k)` as well. and if it is `False` then raise `raise CannotCoerceNullException(self.name, table_name, col_name)`

rudolfix: also unit test this in common schema tests

dlt/extract/extractors.py
```diff
@@ -340,10 +340,9 @@ def write_items(self, resource: DltResource, items: TDataItems, meta: Any) -> No
         static_table_name = self._get_static_table_name(resource, meta)
         items = [
             # 3. remove columns and rows in data contract filters
-            # 2. Remove null-type columns from the table(s) as they can't be loaded
-            self._apply_contract_filters(
-                pyarrow.remove_null_columns(tbl), resource, static_table_name
-            )
+            # NOTE: We don't remove null-type columns from the table(s) with pyarrow.remove_null_columns(tbl).
+            # The removal is handled in the normalizer.
+            self._apply_contract_filters(tbl, resource, static_table_name)
```
rudolfix, line 343: this is good!

dlt/normalize/items_normalizers.py
```diff
@@ -677,15 +687,16 @@ def __call__(self, extracted_items_file: str, root_table_name: str) -> List[TSch
             must_rewrite = pyarrow.should_normalize_arrow_schema(
```
rudolfix: `remove_null_columns_from_schema` should be done inside `should_normalize_arrow_schema` and only one `must_rewrite` should be returned

dlt/normalize/items_normalizers.py
```diff
@@ -579,6 +583,9 @@ def _write_with_dlt_columns(
             for batch in pyarrow.pq_stream_with_new_columns(
                 f, new_columns, row_groups_per_read=self.REWRITE_ROW_GROUPS
             ):
+                if contains_null_cols:
+                    batch = pyarrow.remove_null_columns(batch)
```
rudolfix: `remove_null_columns` should be done in `normalize_py_arrow_item`.
thanks to that we'll do it already in the extractor (which uses the same normalize function). this is how it was supposed to work.

in the extractor:
1. we extract dlt schema first so we'll include incomplete table on which we'll warn in normalizer
2. then we'll drop columns before we write arrow in the extractor
3. then in normalizer all is normalized and we do not need to rewrite

dlt/normalize/normalize.py
```diff
@@ -159,6 +159,26 @@ def map_single(self, schema: Schema, load_id: str, files: Sequence[str]) -> TWor
...
+            logger.info(
+                f"Table {table_name} has seen data for the first time with load id {load_id}"
+            )
+            x_normalizer["seen-data"] = True
+
+        # Handle column-level x-normalizer
+        # drop seen-null-first flag if data type was set
+        for column in table_schema.get("columns", {}).values():
+            col_x_normalizer = column.setdefault("x-normalizer", {})
+            if col_x_normalizer.get("seen-null-first") and "data_type" in column:
```
rudolfix, line 177: this is good!

tests/common/schema/test_inference.py
```diff
@@ -373,12 +373,14 @@ def __call__(self) -> Any:
...
+def test_coerce_new_null_value(schema: Schema) -> None:
     row = {"timestamp": None}
     new_row, new_table = schema.coerce_row("event_user", None, row)
+    # No new rows, but new column in schema
     assert "timestamp" not in new_row
-    # columns were not created
-    assert new_table is None
+    assert "data_type" not in new_table["columns"]["timestamp"]
+    assert new_table["columns"]["timestamp"]["nullable"] is True
+    assert new_table["columns"]["timestamp"]["x-normalizer"]["seen-null-first"] is True
```
rudolfix, line 383: please also test if we raise when a column is supposed to be not null. we should raise. see above. `_dlt_id` is supposed to be not null. (schema should contain this):
```
"not_null": [
            "_dlt_id",
            "_dlt_root_id",
            "_dlt_parent_id",
            "_dlt_list_idx",
            "re:^_dlt_load_id$",
        ],
```)

File: tests/libs/pyarrow/test_pyarrow.py (hunk 1 above)
Comment (rudolfix) [reply tests/libs/pyarrow/test_pyarrow.py]: here you should just unit-test `get_column_type_from_py_arrow` and this other function that you added. this test can be moved to `test_pipeline_extra` where we can use resources and have arrow installed.

here we write simple unit test with minmum other imports

Comment (rudolfix) [reply tests/libs/pyarrow/test_pyarrow.py]: also you should get warning in `normalize` not `extract`. is it the reason the test is diabled?

File: tests/load/pipeline/test_arrow_loading.py [hunk 10]
```diff
@@ -296,3 +300,57 @@ def some_data():
...
     assert_load_info(info)
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, default_staging_configs=True, subset=["duckdb"]),
+    ids=lambda x: x.name,
+)
+@pytest.mark.parametrize("is_none", [True, False])
+def test_warning_from_arrow_normalizer_on_null_column(
```
rudolfix, line 310: this is also a good place to run arrow tests. another good place is `test_pipeline_extra`. those tests we also run on mac on windows

tests/load/pipeline/test_arrow_loading.py
```diff
@@ -296,3 +300,57 @@ def some_data():
...
+
+    pipeline = destination_config.setup_pipeline("arrow_" + uniq_id())
+
+    pipeline.extract(my_source())
+    pipeline.normalize()
+
+    if is_none:
+        logger_spy.assert_called_once()
+        expected_warning = (
+            "The column col1 in table my_resource did not receive any data during this load."
```
rudolfix, line 347: this is good but maybe too good :) I'd just check a substring. small wording changes should not fail this test. or? up to you

tests/pipeline/test_arrow_sources.py
```diff
@@ -72,7 +72,10 @@ def some_data():
         normalized_bytes = f.read()
 
         # Normalized is linked/copied exactly and should be the same as the extracted file
-        assert normalized_bytes == extracted_bytes
+        # NOTE: this is only true if there are no null columns,
+        # as we remove null columns in the normalizer now.
+        # Previously, the null columns were removed in the extractor.
+        assert normalized_bytes != extracted_bytes
```
rudolfix: the old behavior will be back again.

rudolfix (changes requested):
just a few small things to fix. really good tests. thanks!

dlt/extract/extractors.py
```diff
@@ -340,10 +340,9 @@ def write_items(self, resource: DltResource, items: TDataItems, meta: Any) -> No
         static_table_name = self._get_static_table_name(resource, meta)
         items = [
             # 3. remove columns and rows in data contract filters
-            # 2. Remove null-type columns from the table(s) as they can't be loaded
-            self._apply_contract_filters(
-                pyarrow.remove_null_columns(tbl), resource, static_table_name
-            )
+            # NOTE: We don't remove null-type columns from the table(s) with pyarrow.remove_null_columns(tbl).
```
rudolfix: comment no longer valid. we remove them in the extract to save time

dlt/normalize/items_normalizers.py
```diff
@@ -538,7 +538,11 @@ class ArrowItemsNormalizer(ItemsNormalizer):
     REWRITE_ROW_GROUPS = 1
 
     def _write_with_dlt_columns(
-        self, extracted_items_file: str, root_table_name: str, add_dlt_id: bool
+        self,
+        extracted_items_file: str,
+        root_table_name: str,
+        add_dlt_id: bool,
+        #        contains_null_cols: bool = False,
```
rudolfix: remove

Issue #2605, opened 2025-05-06
Credentials with placeholder values should emit warnings during resolution


djudjuu:
### Feature description

If a value in `config.toml` or `secrets.toml` looks like it wasn't configured, we should emit a warning.

### Are you a dlt user?

None

### Use case

If someone tries to set up a project we want them to get up and running very smoothly

### Proposed solution

if writing a configuration/secret, we could use values that make it clear that they don't occur in production.
and if those values get encountered during resolution log a warning

something like this:

```python
MOON_LANDING = pendulum.datetime(1969, 7, 21, 2, 56, 0, tz='UTC')
TYPE_EXAMPLES = {
    "text": "<configure me>",
    "timestamp": MOON_LANDING.to_iso8601_string(),
    "date": MOON_LANDING.to_date_string(),
    "json_sequence": ["a", "b", "c"],
}

....
def generate_typed_example(name: str, hint: AnyType) -> Any:
    inner_hint = extract_inner_hint(hint)
    try:
        sc_type = py_type_to_sc_type(inner_hint)
        if sc_type == "text":
            return TYPE_EXAMPLES["text"]
        if sc_type == "bigint":
            return 0
        if sc_type == "double":
            return 1.0
        if sc_type == "bool":
            return True
        if sc_type == "json":
            if is_subclass(inner_hint, C_Sequence):
                return TYPE_EXAMPLES["json_sequence"]
            else:
                table = tomlkit.table(False)
                table["key"] = "value"
                return table
        if sc_type == "timestamp":
            return TYPE_EXAMPLES["timestamp"]
        if sc_type == "date":
            return TYPE_EXAMPLES["date"]
        if sc_type in ("wei", "decimal"):
            return "1.0"
        raise TypeError(sc_type)
    except TypeError:
        return name
```

PR #2636, opened 2025-05-14
warn when resolving configs or secrets with placeholder values

djudjuu:
fixes: #2531 , #2605 

after discussion:

- complex configs/secrets (json) will not be written
- other types will have configure me or a special timestamp that we can compare against (moon landing)
- when such a value is encounted during config-resolution we emit warning with details


dlt/cli/config_toml_writer.py
```diff
@@ -36,10 +36,12 @@ def generate_typed_example(name: str, hint: AnyType) -> Any:
             return True
         if sc_type == "json":
             if is_subclass(inner_hint, C_Sequence):
-                return ["a", "b", "c"]
+                return ["customers", "orders", "items"]
```
sh-rp, line 39: This seems very specific to the rest_api (same below). This might totally mess up instantiation of verified sources, have you thought about that?

djudjuu: I checked against all verified sources and only chess is using this (atm)
to instantiate the secret_dict

djudjuu: which, to be fair, looks weird:

```toml
[sources.chess.secret_dict] # fill this in!
base_url = "https://jaffle-shop.scalevector.ai/api/v1" # config object example

[sources.chess.secret_dict.paginator] # nested config object example
type = "header_link"
```

djudjuu:
alternative:

```python
        if sc_type == "json":
            if is_subclass(inner_hint, C_Sequence):
                return ["<configure me>"]
            else:
                table = tomlkit.table(False)
                table["key"] = "<configure me>"
                table["nested_key"] = {"subkey": "<configure me>"}
                return table
 ```

to get

```toml

[sources.rest_api]
resources = ["<configure_me>"] # fill this in!
[sources.rest_api.client] # fill this in!
key = "<configure me>"

[sources.rest_api.client.nested_key]
subkey = "<configure me>"
```

and for chess:
```

[sources.chess.secret_dict] # fill this in!
key = "<configure me>"

[sources.chess.secret_dict.nested_key]
subkey = "<configure me>"

```

dlt/cli/config_toml_writer.py
sh-rp: @djudjuu this code applies to all configs, also configs for sources that user have written themselves, so we do not know when they will be applied. If you have discussed something else with @rudolfix then please disregard my comment, but I would really not put rest_api defaults into generic code. I thought we would use templates in dlt+, and I think we can probably skip the config toml for sources in the core repo, no? We use those templates there, or is there a reason to keep them?

djudjuu: jup, this is disregarded already, I guess I could have deleted the comment. we will just not write values in OSS for complex types
(and do templates in dlt+)

dlt/cli/config_toml_writer.py
```diff
@@ -26,25 +33,16 @@ def generate_typed_example(name: str, hint: AnyType) -> Any:
...
-                return ["a", "b", "c"]
-            else:
-                table = tomlkit.table(False)
-                table["key"] = "value"
-                return table
-        if sc_type == "timestamp":
-            return pendulum.now().to_iso8601_string()
-        if sc_type == "date":
-            return pendulum.now().date().to_date_string()
+            return "json"
```
sh-rp, line 45: does this produce a useful result? if you don't want anything to be written for json (which I guess makes sense) then I think you need to skip this value somehow.

djudjuu: first i was raising an not implemented error here, but I thought its cleaner to produce something valid here and deal with it in the place that this function is used.
 (its only one place)
image

dlt/cli/config_toml_writer.py
```diff
@@ -14,6 +14,13 @@
 from dlt.common.data_types import py_type_to_sc_type
 from dlt.common.typing import AnyType, is_optional_type, is_subclass
 
+MOON_LANDING = pendulum.datetime(1969, 7, 21, 2, 56, 0, tz="UTC")
```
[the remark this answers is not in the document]
djudjuu: hehe....space is always cool

sh-rp (approved):
LGTM

you still could consider adding a secrets section for sources that do not provide any secrets we can print (such as the rest_api) just that the user knows where to put stuff, but I'll leave it up to you.

rudolfix (changes requested):
overall LGTM! but we need a few changes

rudolfix (approved):
LGTM!

PR #2638, opened 2025-05-14
Feat: Support clustered tables with custom column order in BigQuery destination

### LINKED ISSUE #2545: BigQuery destination does not respect column clustering order
### dlt version

1.9

### Describe the problem

In BigQuery, when clustering tables, `dlt` is currently using the schema's column order instead of the column order specified by the users.

This behavior can negatively affect the performance of table cluster, and should be fixed.

https://cloud.google.com/bigquery/docs/clustered-tables#cluster_column_ordering

### Expected behavior

When creating BigQuery clustered table, `dlt` should use the exact column order as the list passed in by the users.

### Steps to reproduce

1. Configure a BQ destination with a table containing 2 columns: `a` `b`
2. Set clustering setting for this table as `[b, a]`
3. dlt create the table clustered by `[a, b]`

Expected behavior: 
The table should be clustered by `[b, a]`

### Operating system

Linux

### Runtime environment

Local

### Python version

3.13

### dlt data source

Any

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

_No response_

hsm207:
### Description
When user passes `cluster=['b', 'a']`, the big query table will be clustered by column `b` first, then `a`. 

### Related Issues

- Fixes #2545
- Closes #2545
- Resolves #2545

### Additional Context
See [Introduction to clustered tables](https://cloud.google.com/bigquery/docs/clustered-tables)


dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -287,12 +288,22 @@ def _get_table_update_sql(
...
+            # This is required for backward compatibility and to pass tests like
+            # 'test_create_table_with_partition_and_cluster' in test_bigquery_table_builder.py,
+            # which set clustering only via per-column hints and expect the CLUSTER BY clause.
+            cluster_cols_from_columns = [
+                self.sql_client.escape_column_name(c["name"])
+                for c in new_columns
+                if c.get("cluster") or c.get(CLUSTER_HINT, False)
+            ]
+            if cluster_cols_from_columns:
+                sql[0] += "\nCLUSTER BY " + ", ".join(cluster_cols_from_columns)
```
hsm207, line 306: @rudolfix 

Do we want to keep this branch to support the old behavior?

My thoughts:
Looking at the `test_create_table_with_partition_and_cluster` test, it's ambiguous to deduce the intended clustering column order if users only set per-column "cluster" hints. The resulting order in the CLUSTER BY clause will just follow the order of columns in the schema, not necessarily the user's intent. If users directly modify the schema or set per-column hints, it's not possible to reliably track or enforce the intended clustering order without introducing more complexity.

Should we encourage users to always use the `bigquery_adapter`'s `cluster` argument for explicit ordering, and discourage direct schema editing for clustering?

rudolfix: btw. I bet we'll have similar requests for primary keys, merge keys etc. where we also follow column ordering... maybe I'll add it to our QoL project

rudolfix (changes requested):
this is good! and I like the test. see my comments

hsm207 (commented):
> here we have small code duplication
> `sql[0] += "\nCLUSTER BY " + ", ".join(cluster_list)` and column name escaping (self.sql_client.escape_column_name) could be present only once and you just create lists for main and fallback scenario 

good catch!

dlt/destinations/impl/bigquery/bigquery.py
hsm207: @rudolfix I've made the deprecation and updated the docs.

Please have a look and let me know if need needs adjusting. I tried to keep it consistent with other deperactions.

rudolfix (approved):
LGTM!

PR #2639, opened 2025-05-14
Docs: addition of normalizer behaviour to transformations docs

anuunchin:
### Description
This PR adds some relevant snippets to the transformations docs. As outlined by @sh-rp, this covers:
- What are the defaults wrt adding load_id and dlt_id
- How is the dlt_id create in sql transformations (special db sql functions)
- When the transformations are python based, this behaves like a regular non transformations pipeline (your code won't even be used in that case)
-  What happens if the user supplies a dlt_id in the transformation?

### DIFF
--- a/docs/website/docs/general-usage/transformations/index.md
+++ b/docs/website/docs/general-usage/transformations/index.md
@@ -174,10 +174,8 @@ Additionally, column names are normalized according to the naming schema selecte
 This allows dlt to maintain data lineage and enables features like incremental loading and merging, even when working with raw SQL queries.
 
 :::info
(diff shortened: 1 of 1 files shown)


docs/website/docs/general-usage/transformations/index.md
```diff
@@ -162,6 +177,7 @@ This allows dlt to maintain data lineage and enables features like incremental l
 This feature is planned but not yet implemented. Currently, you need to explicitly include these columns in your SQL queries if you want to use them for tracking purposes.
```
sh-rp, line 177: you can remove this part, since we will have it soon :)

sh-rp (approved):

PR #2640, opened 2025-05-14
Feat: Add configuration propagation to deltalake.write_deltalake (#2629)

### LINKED ISSUE #2629: Make delta destination configurable (pass more parameters to underlying write_deltalake)
### Feature description

I would love to configure my delta destination in more detail. I was missing this feature when I wanted to create a new data integration that writes to a delta destination but with change data capture enabled (https://docs.delta.io/latest/delta-change-data-feed.html). In order to have that work, I need to also configure the `minWriterVersion` in the delta writer protocol. There are more config options that are not exposed through `dlt`.

DLT writes to delta through the `testy/.venv/lib/python3.12/site-packages/dlt/common/libs/deltalake.py::write_delta_table` function. This calls the underlying `deltalake.write_deltalake` function by providing a subset of arguments.

**dlt's wrapper func**
```
def write_delta_table(
    table_or_uri: Union[str, Path, DeltaTable],
    data: Union[pa.Table, pa.RecordBatchReader],
    write_disposition: TWriteDisposition,
    partition_by: Optional[Union[List[str], str]] = None,
    storage_options: Optional[Dict[str, str]] = None,
) -> None:
    """Writes in-memory Arrow data to on-disk Delta table.

    Thin wrapper around `deltalake.write_deltalake`.
    """
    write_deltalake(  # type: ignore[call-overload]
        table_or_uri=table_or_uri,
        data=ensure_delta_compatible_arrow_data(data, partition_by),
        partition_by=partition_by,
        mode=get_delta_write_mode(write_disposition),
        schema_mode="merge",  # enable schema evolution (adding new columns)
        storage_options=storage_options,
        engine="rust",  # `merge` schema mode requires `rust` engine
    )
```

**actual deltalake.write_deltalake func:**
```
def write_deltalake(
    table_or_uri: Union[str, Path, DeltaTable],
    data: Union[
        "pd.DataFrame",
        ds.Dataset,
        pa.Table,
        pa.RecordBatch,
        Iterable[pa.RecordBatch],
        RecordBatchReader,
        ArrowStreamExportable,
    ],
    *,
    schema: Optional[Union[pa.Schema, DeltaSchema]] = None,
    partition_by: Optional[Union[List[str], str]] = None,
    mode: Literal["error", "append", "overwrite", "ignore"] = "error",
    file_options: Optional[ds.ParquetFileWriteOptions] = None,
    max_partitions: Optional[int] = None,
    max_open_files: int = 1024,
    max_rows_per_file: int = 10 * 1024 * 1024,
    min_rows_per_group: int = 64 * 1024,
    max_rows_per_group: int = 128 * 1024,
    name: Optional[str] = None,
    description: Optional[str] = None,
    configuration: Optional[Mapping[str, Optional[str]]] = None,
    schema_mode: Optional[Literal["merge", "overwrite"]] = None,
    storage_options: Optional[Dict[str, str]] = None,
    partition_filters: Optional[List[Tuple[str, str, Any]]] = None,
    predicate: Optional[str] = None,
    target_file_size: Optional[int] = None,
    large_dtypes: bool = False,
    engine: Literal["pyarrow", "rust"] = "rust",
    writer_properties: Optional[WriterProperties] = None,
    custom_metadata: Optional[Dict[str, str]] = None,
    post_commithook_properties: Optional[PostCommitHookProperties] = None,
    commit_properties: Optional[CommitProperties] = None,
) -> None:
```

**Proposal**
Provide the `configuration` option to config, that are passed to the underlying `write_deltalake` function to give the user more fine grained control over how to configure the delta destination.
An alternative would be to expose a `kwargs` object that is passed to `write_deltalake` while protecting defaults (e.g. `engine="rust"`. I will start with the initial approach (only expose `configuration`. If it is desired to make it fully flexible, let me know and I can take the `kwargs` approach. For now, that scope is a bit bigger, so I'll start small.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

Have more fine grained control over the `delta` destination. Currently, there is no direct way of activating change data capture, deletion vectors, setting the reader/writer protocol versions, etc.

### Proposed solution

Provide the `configuration` option to config, that are passed to the underlying `write_deltalake` function to give the user more fine grained control over how to configure the delta destination.
An alternative would be to expose a `kwargs` object that is passed to `write_deltalake` while protecting defaults (e.g. `engine="rust"`. I will start with the initial approach (only expose `configuration`. If it is desired to make it fully flexible, let me know and I can take the `kwargs` approach. For now, that scope is a bit bigger, so I'll start small.

### Related issues

_No response_

gaatjeniksaan:
### Description
Currently there is no way to configure the `delta` table destination. DLT uses a thin wrapper (`write_delta_table`) around the `deltalake.write_deltalake` function, but uses the default `configuration=None`. So currently there is no direct way of configuring the resulting delta table. This PR changes that by exposing a new config field `deltalake_configuration` that is passed through.

### Related Issues

- Fixes #...
- Closes #...
- Resolves #2629 

### Additional Context
As described in the issue, there is an alternative approach of simply allowing a `kwargs` parameter that can be constructed that will be passed to the `write_deltalake` call, but that scope is slightly bigger, and the current approach is more consistent with `storage_options`.

In my previous PR I had unrelated failing tests, and I do so again, but the stuff that I've touched seems to work, so I think it's all good from my side, but I don't have had enough exposure to the project to really assess that correctly.

### DIFF
--- a/dlt/common/libs/deltalake.py
+++ b/dlt/common/libs/deltalake.py
@@ -1,3 +1,4 @@
+from collections.abc import Mapping
 from typing import Optional, Dict, Union, List
 from pathlib import Path
 
@@ -95,6 +96,7 @@ def write_delta_table(
     write_disposition: TWriteDisposition,
     partition_by: Optional[Union[List[str], str]] = None,
     storage_options: Optional[Dict[str, str]] = None,
+    configuration: Optional[Mapping[str, Optional[str]]] = None,
 ) -> None:
     """Writes in-memory Arrow data to on-disk Delta table.
 
@@ -108,6 +110,7 @@ def write_delta_table(
         schema_mode="merge",  # enable schema evolution (adding new columns)
         storage_options=storage_options,
         engine="rust",  # `merge` schema mode requires `rust` engine
+        configuration=configuration,
     )
 
 
--- a/dlt/common/storages/configuration.py
+++ b/dlt/common/storages/configuration.py
@@ -16,7 +16,7 @@ from dlt.common.configuration.specs import (
     SFTPCredentials,
 )
 from dlt.common.exceptions import TerminalValueError
-from dlt.common.typing import DictStrAny, get_args
+from dlt.common.typing import DictStrAny, DictStrOptionalStr, get_args
 from dlt.common.utils import digest128
 
 
@@ -184,6 +184,7 @@ class FilesystemConfiguration(BaseConfiguration):
     kwargs: Optional[DictStrAny] = None
     client_kwargs: Optional[DictStrAny] = None
     deltalake_storage_options: Optional[DictStrAny] = None
+    deltalake_configuration: Optional[DictStrOptionalStr] = None
 
     @property
     def protocol(self) -> str:
--- a/dlt/common/typing.py
+++ b/dlt/common/typing.py
@@ -88,6 +88,7 @@ CallableAny = NewType("CallableAny", Any)  # type: ignore[valid-newtype]
 NoneType = type(None)
 DictStrAny: TypeAlias = Dict[str, Any]
 DictStrStr: TypeAlias = Dict[str, str]
+DictStrOptionalStr: TypeAlias = Dict[str, Optional[str]]
 StrAny: TypeAlias = Mapping[str, Any]  # immutable, covariant entity
 StrStr: TypeAlias = Mapping[str, str]  # immutable, covariant entity
 StrStrStr: TypeAlias = Mapping[str, Mapping[str, str]]  # immutable, covariant entity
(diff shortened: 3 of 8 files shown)


rudolfix (approved):
LGTM! thanks for the docs and tests!
We could pass this configuration per table via table hints in the schema, but that's more work

rudolfix:
> @rudolfix thanks for reviewing! Shall I open a new issue and go down the table hints route as well? Wouldn't mind taking that work as well

that would be cool! please tag me in the ticket before you start. also feel free to change our delta helper functions. you'll need to pass prepared TTableSchema there to access hints.

PR #2656, opened 2025-05-17
Psycopg2SqlClient: accept extra options

nicob3y:
### Description

Add ability to add Postgres runtime server config via Psycopg2SqlClient.

Example of PostgresCredentials usage with "my_setting" server setting :
```
from dlt.destinations.impl.postgres.configuration import PostgresCredentials

creds = PostgresCredentials("postgresql://user:password@host:port/db?options=-cmy_setting=value")
```

### Additional Context

I needed to transmit a parameter (such as a session option) to the PostgreSQL server, but the current implementation of Psycopg2SqlClient did not allow me to do this easily.

### DIFF
--- a/dlt/destinations/impl/postgres/sql_client.py
+++ b/dlt/destinations/impl/postgres/sql_client.py
@@ -45,9 +45,13 @@ class Psycopg2SqlClient(SqlClientBase["psycopg2.connection"], DBTransaction):
         self.credentials = credentials
 
     def open_connection(self) -> "psycopg2.connection":
+        options=f"-c search_path={self.fully_qualified_dataset_name()},public"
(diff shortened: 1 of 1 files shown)


rudolfix (changes requested):
code looks good! Are you able to mention this feature in our docs? we mention connection string in `postgres.md` close to the top of the file

rudolfix:
@nicob3y we could not get linter fixes and docs update to it so we are closing the PR for now. feel free to re-open. I think that having this option makes sense but we need docs and a simple test to merge it....

nicob3y:
@rudolfix I can't reopen this PR so I'm creating a new one

PR #2657, opened 2025-05-17
Fixing docs broken link

chethanuk:
### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/docs/website/docs/running-in-production/alerting.md
+++ b/docs/website/docs/running-in-production/alerting.md
@@ -66,7 +66,7 @@ for package in load_info.load_packages:
                 )
             )
 ```
-Refer to this [example](../examples/chess_production/) for a practical application of the method in a production environment.
+Refer to this [example](../../../examples/chess_production/chess_production.py) for a practical application of the method in a production environment.
 
 Similarly, Slack notifications can be extended to include information on pipeline execution times, loading durations, schema modifications, and more. For comprehensive details on configuring and sending messages to Slack, please read [here](./running#using-slack-to-send-messages).
 


burnash:
Hi @chethanuk, thank you for spotting this. From what I can see the issue here is different, it appears to be an issue with Docusaurus or docs deployment: the link is only broken when there's a trailing slash in the URL path. 

E.g. : 
- https://dlthub.com/docs/running-in-production/alerting - "examples" links properly
- https://dlthub.com/docs/running-in-production/alerting/ - "examples" is not found.

I'll take a closer look at the config to see what's wrong.

sh-rp:
@chethanuk Thanks for the PR.  I just had a look and it appears to me that this link works without problems on all major browsers (Firefox, Chrome and Safari), so I am not sure what you are trying to fix here? Let me know what the problem is that you see, otherwise I'll close this PR.

sh-rp:
@chethanuk the problem here is that we are using a relative link in the url and it resolves differently depending on a trailing slash on the alert page as @burnash mentioned. The correct fix would be to turn this into an absolute url, relative to the base url of the docs page, but not to link to the python file. Let me know if you'd like to change this, otherwise I'd close this PR.

sh-rp:
> Hi @chethanuk, thank you for spotting this. From what I can see the issue here is different, it appears to be an issue with Docusaurus or docs deployment: the link is only broken when there's a trailing slash in the URL path.
> 
> E.g. :
> 
>     * https://dlthub.com/docs/running-in-production/alerting - "examples" links properly
> 
>     * https://dlthub.com/docs/running-in-production/alerting/ - "examples" is not found.
> 
> 
> I'll take a closer look at the config to see what's wrong.

I don't think this is a big problem. All the internal links resolve to pages without the trailing slash as far as I can see, so this case should not happen very often?

burnash:
> All the internal links resolve to pages without the trailing slash as far as I can see, so this case should not happen very often?

I suspect this is may be either external links or search traffic. Somehow people get into a page with a trailing slash.

burnash:
Closing this for now.

PR #2662, opened 2025-05-19
dlt marimo app pre-release version

sh-rp:
### Description
Implements a marimo app as replacement of and improvement upon the streamlit app.

### Preview
Check out this branch and run the app with `dlt studio`

### TODO

- [ ] Write tests, both unit and e2e
- [x] Re-activate disabled tests
- [x] finish trace view
- [ ] add hackable cell
- [ ] create ejectable / hackable app?


dlt/helpers/studio/utils.py
```diff
@@ -0,0 +1,255 @@
...
+
+    Args:
+        row_id (str): The id of the row.
+        name (str): The name of the column.
+        __ (Any): The value of the cell.
+
+    Returns:
+        Dict[str, str]: The css style of the cell.
+    """
+    style = {"background-color": "white" if (int(row_id) % 2 == 0) else "#f4f4f9"}
```
zilto, line 252: image



This issue is worsened because the user can't toggle between light vs. dark mode. The app takes the system color by default

tests/helpers/studio/test_utils.py
```diff
@@ -0,0 +1,699 @@
...
+import dlt
+from dlt.common.schema import Schema
+from dlt.common.storages import FileStorage
+from dlt.common.pendulum import pendulum
+from dlt.helpers.studio.config import StudioConfiguration
+from dlt.helpers.studio import utils
+from dlt.sources._single_file_templates.fruitshop_pipeline import fruitshop as fruitshop_source
+
+
+pytest.skip(allow_module_level=True)
```
sh-rp, line 19: Note: fix, review and activate these tests

rudolfix (changes requested):
approved for alpha release! just move the dev deps to a group like in the review comment

anuunchin:
This is so cool! 🤩  I will definitely be using this from now on, the only things I was a bit confused about is:
- in `Pipeline Loads`, I don't really understand what the `Cache query results` and `Limit to 1000 rows` options are doing 🤔

PR #2664, opened 2025-05-19
Fix:  Expanding star selections

### LINKED ISSUE #2660: Expand sql star queries to explicitly list all columns in transformations
For the model normalizer to work properly, we need to ensure that star selects that either come from ibis expressions or from "manually entered" sql queries are expanded to list all columns. For this we need to modify the transformation_function on the transformation branch:

1. Find the branch of the code where we have the query as well as all column names. And we know that we are processing a "model" transformation type. This is here: https://github.com/dlt-hub/dlt/pull/2528/files#diff-487a015741a437c30f34bb4124364122502a03f908e1f7572b4b542ef4fa1abeR167
2. Modify the `select_query` to expand the top level star select if there is one. You have all columns in var `all_columns`.
3. Done :)

You may be able to do this with the sqlglot query optimizer: https://sqlglot.com/sqlglot/optimizer/optimizer.html by using the qualify rule. You'll need to supply the the full sqlglot schema for the dataset we are processing which you can get with  `lineage.create_sqlglot_schema`. I woud do a simple test first to see if this approach works.

anuunchin:
### Description
The current model normalizer doesn't allow star expressions. However, the `def query()` method in `ibis_relation.py` uses `ibis.to_sql()` which internally detects if a select is a star selection and returns a `*`. The internal logic that automatically returns a `*` directly depends on 
on the following line in https://github.com/ibis-project/ibis/blob/eb7e98494eca0d81aa54d8891e03c8e53185c28c/ibis/backends/sql/compilers/base.py#L1301:

```python
            if op.is_star_selection() and (not qualified or self.supports_qualify):
                fields = [STAR]
```
This PR temporarily disables the detection of star schemas with a context manager when `def query()` is used.

> Note that in order for this to work, the table expression should explicitly include all columns. Therefore, `transformation.py` is adjusted accordingly.

### Related Issues

- Fixes #2660


sh-rp (changes requested):
Thanks for figuring this out :) I did a quick check though with sql_glot and the optimizer, and I think we can actually do this in a more clean way directly on the query, check out this snippet:

```
import sqlglot
from sqlglot.optimizer.qualify import qualify
schema = {"tbl": {"col": "INT", "col2": "STRING"}}
expression = sqlglot.parse_one("SELECT * FROM tbl")
print(qualify(expression, schema=schema).sql())
```

The optimizer will expand the stars for us by default. Let me know if this works. We have the code to get the current sqlglot schema in `lineage.py` (which we will probably have to adjust once @zilto s work is merged, but I will do this.)

sh-rp:
this work has gone into https://github.com/dlt-hub/dlt/pull/2528, will close this PR

PR #2671, opened 2025-05-22
Upsert merge strategy for iceberg

### LINKED ISSUE #2549: enable upsert for filesystem / iceberg destination
**Background**
With `upsert` added to `pyiceberg` we can implement the same `upsert` mechanism as in `delta`. Porting the code should be trivial.

**Tasks**
1. * [ ] port `delta` upsert code into `pyiceberg`
2. * [ ] add `upsert` to merge strategy selector in `factory` of `filesystem`
3. * [ ] that should enable upsert tests for iceberg. we do not need any new tests here

anuunchin:
### Description
This PR enables the "upsert" merge write disposition for Iceberg. The functionality is similar to how delta works. Underneath, Iceberg's upsert table is used.

### Related Issues

- Resolves #2549 

### Additional Context (for failing tests)
1.  The test `test_merge_on_keys_in_schema_nested_hints` would likely fail with maximum recursion error, as Iceberg's upsert has a limitation mentioned [here](https://github.com/apache/iceberg-python/issues/1759). This is apparently resolved in pyiceberg 0.9.1, but I'm still getting a sigill error with 0.9.1 and maximum recursion error with 0.9.0, the reason for which I'm unsure about.
    -  **Possible solution**: The issue can be addressed with batching the tables with size 1000 each.

2.  The test `test_table_format_schema_evolution` is failing with `Data type struct<a: int64> is not supported in join non-key field json`  because iceberg's upsert uses arrow's hash join which doesn't handle nested types. The issue was addressed with this [PR](https://github.com/apache/iceberg-python/pull/1878) in pyiceberg 0.9.1, but with 0.9.1 I get a different error saying `Unexpected physical type FIXED_LEN_BYTE_ARRAY for decimal(6, 4), expected INT32`. The latter issue was also addressed in this [PR](https://github.com/apache/iceberg-python/pull/1983) in pyiceberg, but it's not yet released. Therefore, the issue should go away with pyiceberg's next release. (I tested with the newest code and it works 👀 )
    -  **Possible solution**: We wait for the next pyiceberg release and upgrade to the latest version.

3. The tests `test_pipeline_load_parquet` and `test_table_format_child_tables ` fail because iceberg's upsert strictly forbids duplicates in input data as well as target table - if the key columns are not unique inside the source it raises an error. Delta doesn't even seem to be working properly 👀 .
   - **Possible solution**: We keep it explicit in the docs and leave it to the user to follow the uniqueness requirement + adjust the failing tests.


dlt/common/libs/pyiceberg.py
```diff
@@ -63,6 +64,34 @@ def write_iceberg_table(
...
+    schema: TTableSchema,
+) -> None:
+    """Merges in-memory Arrow data into on-disk Iceberg table."""
+    strategy = schema["x-merge-strategy"]  # type: ignore[typeddict-item]
+    if strategy == "upsert":
+        # evolve schema
+        with table.update_schema() as update:
+            update.union_by_name(ensure_iceberg_compatible_arrow_schema(data.schema))
+
+        if "parent" in schema:
```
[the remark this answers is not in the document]
rudolfix: `upsert` is using deterministic `row_key` that is computed from `primary_key` if the root table

docs/website/docs/dlt-ecosystem/destinations/iceberg.md
```diff
@@ -120,4 +120,14 @@ The [S3-compatible](./filesystem.md#using-s3-compatible-storage) interface for G
 The `az` [scheme](./filesystem.md#supported-schemes) is not supported when using the `iceberg` table format. Please use the `abfss` scheme. This is because `pyiceberg`, which dlt used under the hood, currently does not support `az`.
 
 ## Table format `merge` support
-The `merge` write disposition is not supported for Iceberg and falls back to `append`. If you're interested in support for the `merge` write disposition with Iceberg, check out [dlt+ Iceberg destination](../../plus/ecosystem/iceberg.md).
+The [`upsert`](../../general-usage/merge-loading.md#upsert-strategy) merge strategy is supported for `iceberg`. This strategy requires that the input data contains no duplicate rows based on the key columns, and that the target table also does not contain duplicates on those keys.
```
rudolfix: please warn about batching

dlt/common/libs/pyiceberg.py
```diff
@@ -63,6 +64,42 @@ def write_iceberg_table(
...
+            join_cols = [get_first_column_name_with_prop(schema, "unique")]
+        else:
+            join_cols = get_columns_names_with_prop(schema, "primary_key")
+
+        for rb in data.to_batches(max_chunksize=1_000):
```
rudolfix, line 101: we should revise batching when this
https://github.com/apache/iceberg-python/pull/1817
is implemented.
1. maybe the problem is gone
2. if not - we at least do it in a single transaction

pyproject.toml
```diff
@@ -105,7 +105,10 @@ db-dtypes = { version = ">=1.2.0", optional = true }
 # https://github.com/apache/airflow/issues/28723
 # pyiceberg = { version = ">=0.7.1", optional = true, extras = ["sql-sqlite"] }
 # we will rely on manual installation of `sqlalchemy>=2.0.18` instead
-pyiceberg = { version = ">=0.9.0", optional = true }
+# replace pyiceberg's version with the one released after 0.9.1
+pyiceberg = { git = "https://github.com/apache/iceberg-python.git", rev = "260ef54e3920d435ae3b2ccda090e66f9c1ac015", optional = true }
+# pyiceberg = { version = ">=0.9.1" , optional = true }
```
rudolfix: this needs to be a release version, dependency on PR cannot be released to PyPI

tests/load/pipeline/test_merge_disposition.py
```diff
@@ -817,6 +820,9 @@ def test_merge_no_merge_keys(destination_config: DestinationTestConfiguration) -
     ids=lambda x: x.name,
 )
 def test_pipeline_load_parquet(destination_config: DestinationTestConfiguration) -> None:
+    import os
+
+    os.environ["LOAD__WORKERS"] = "1"
```
rudolfix: do we need this? is it preventing sigkills?

rudolfix (dismissed):
LGTM! but you need to solve the conflict and merge devel...

PR #2676, opened 2025-05-23
Add support for creating custom integer-range partition table in BigQuery

### LINKED ISSUE #2516: Add better support for BigQuery partitions
### Feature description

dlt has a `biqquery_adaptor` that accepts a partition column name, this is rather limiting, big query supports different partition types (i.e. timestamp | date | datetime | int64) at different granularities (i.e. hour | day | month | year) or integer ranges (i.e. start + end + interval), along with options to raise an exception if a query doesn't filter by the partition key, and/or to expire partitions after n days etc.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

This is important for cost management, I've observed a large step change in BigQuery costs using incremental merge replication on large tables - using proper partitioning rectifies this but for all but in most cases I've had to manually create the destinations tables in order to properly partition.

### Proposed solution

One option is to change the type of the `partition` argument to  a dict similar to dbt (https://docs.getdbt.com/reference/resource-configs/bigquery-configs) rather than a column name and generate the `PARTITION BY` clause

The other option is to leave `partition` as a str, but accept the BigQuery syntax literally ie.

```
DATE_TRUNC(transaction_date, MONTH)
  OPTIONS (
    partition_expiration_days = 3,
    require_partition_filter = TRUE)
```

see https://cloud.google.com/bigquery/docs/creating-partitioned-tables

### Related issues

_No response_

hsm207:
### Description

Add full support for the `RANGE_BUCKET` partition expression in BigQuery

### Related Issues

- Fixes #2516
- Closes #2516
- Resolves #2516

### Additional Context

* [Create an integer-range partitioned table](https://cloud.google.com/bigquery/docs/creating-partitioned-tables#create_an_integer-range_partitioned_table)
* [CREATE TABLE syntax](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-definition-language#create_table_statement)

### DIFF
--- a/dlt/destinations/adapters.py
+++ b/dlt/destinations/adapters.py
@@ -3,7 +3,7 @@
 from dlt.destinations.impl.weaviate.weaviate_adapter import weaviate_adapter
 from dlt.destinations.impl.qdrant.qdrant_adapter import qdrant_adapter
 from dlt.destinations.impl.lancedb import lancedb_adapter
-from dlt.destinations.impl.bigquery.bigquery_adapter import bigquery_adapter
+from dlt.destinations.impl.bigquery.bigquery_adapter import bigquery_adapter, bigquery_partition
 from dlt.destinations.impl.synapse.synapse_adapter import synapse_adapter
 from dlt.destinations.impl.clickhouse.clickhouse_adapter import clickhouse_adapter
 from dlt.destinations.impl.athena.athena_adapter import athena_adapter, athena_partition
@@ -14,6 +14,7 @@ __all__ = [
     "qdrant_adapter",
     "lancedb_adapter",
     "bigquery_adapter",
+    "bigquery_partition",
     "synapse_adapter",
     "clickhouse_adapter",
     "athena_adapter",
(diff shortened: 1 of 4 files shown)


hsm207:
> You still store SQL statements in `dlt` schema (correct me if I'm wrong!)

I am not sure, I'm using the `PartitionTransformation` approach like in `athena_adapter`. I thought this is what is meant by a typed dict approach.

> ... but the above is way better.

Do you mean doing something like this?

```python
class RangeBucketPartition(TypedDict):
    type: Literal["range_bucket"]
    column_name: str
    start: int
    end: int
    interval: int
    
class bigquery_partition:
    """Helper class to generate BigQuery partition transformations."""

    @staticmethod
    def range_bucket(
        column_name: str,
        start: int,
        end: int,
        interval: int = 1,
    ) -> RangeBucketPartition:
        """Partition by an integer column with the specified range."""
        return {
            "type": "range_bucket",
            "column_name": column_name,
            "start": start,
            "end": end,
            "interval": interval,
        }
```

and when we have a method in `bigquery_partition` that does something like:

```python
def render_partition_sql(partition: PartitionTransformationDict) -> str:
    if partition["type"] == "range_bucket":
        return (
            f"RANGE_BUCKET({partition['column_name']}, "
            f"GENERATE_ARRAY({partition['start']}, {partition['end']}, {partition['interval']}))"
        )
```

PR #2678, opened 2025-05-23
fix/2677: remove recursive filewatching

### LINKED ISSUE #2677: docs: failure to launch local dev server
### dlt version

1.10

### Describe the problem

From the directory `docs/website`, calling `npm run start` to build the docs fails.

The stack trace points to "too many file watchers"
```
[1]     at async Promise.all (index 0) {
[1]   errno: -28,
[1]   syscall: 'watch',
[1]   code: 'ENOSPC',
[1]   path: '/home/tjean/projects/dlthub/dlt/docs/website/docusaurus.config.js',
[1]   filename: '/home/tjean/projects/dlthub/dlt/docs/website/docusaurus.config.js'
[1] }
[1] [INFO] Docusaurus version: 3.7.0
[1] Node version: v22.12.0
[1] docusaurus start exited with code 1
--> Sending SIGTERM to other processes..
[0] node tools/preprocess_docs.js --watch exited with code SIGTERM
```

## Investigation
- The error is a general node message related to too many watchers. A [quick fix](https://stackoverflow.com/questions/55763428/react-native-error-enospc-system-limit-for-number-of-file-watchers-reached) is increasing the number of watchers allowed on the system. This is not a sustainable solution
- This [error is documented in VSCode](https://code.visualstudio.com/docs/setup/linux#_visual-studio-code-is-unable-to-watch-for-file-changes-in-this-large-workspace-error-enospc), which also uses node. Freeing some watchers on the system could solve the docs issue, but it's not a sustainable solution
- The `start` command in `package.json` seems to call another node process with `--watch` flag (could this double the number of watchers?) git blame points to [this commit](https://github.com/dlt-hub/dlt/commit/c0735acbe471953c971bb2b19a6b3e7fb76bb99b) 4 months ago
  - the `--watch` flag triggers watching recursively the full `docs/website` directory, including the files it generates...
    ```ts
    watch("../docs", { recursive: true, filter: /\.(py|toml|md)$/  }, function(evt, name) {
    ```
- We recently enabled rendering multiple dlt versions of the docs via [this commit](https://github.com/dlt-hub/dlt/commit/63de7c5cc38307a4149ab0704a94cb4139da61d0) 2 months ago
- The current setup breaks `npm run docusaurus start`. The error implicitly complains that `process_docs.js` wasn't run.
  ```
  Error: The docs folder does not exist for version "current". A docs folder is expected to be found at docs_processed.
  ```
  - running `npm run preprocess-docs` once fixes the issue

### Related problem
Because `preprocess_docs.js` regenerates all files, the docusaurus watcher recompiles everything. This makes the dev update loop much slower

## Potential solution
I suspect this new issue has to do with multiple versions of the docs being watched multiple times. If `preprocess_docs.js` watches, it produces a bunch of files under `docs_processed/` that are then watched by  docusaurus. 

- A quick fix is to use docusaurus start --poll which polls docs for changes at a fix rate instead of using watchers

- The most robust solution would be to migrate the bespoke preprocess_docs.js  to a [docusaurus plugin](https://docusaurus.io/docs/api/plugin-methods#example) with a lifecyle method loadContent to generate markdown files (this would cut watchers in half) (edited)

### Operating system

Linux

zilto:
Partially fixes #2677. 

Nonetheless, docs updates are slow because `tools/preprocess_docs.js` updates all files in `docs_processed/`. Then, Docusaurus thinks it needs to recompile everything.

### DIFF
--- a/docs/website/tools/preprocess_docs.js
+++ b/docs/website/tools/preprocess_docs.js
@@ -420,7 +420,7 @@ processDocs()
 if (process.argv.includes("--watch")) {
   console.log(`Watching...`)
   let lastUpdate = Date.now();
-  watch("../", { recursive: true, filter: /\.(py|toml|md)$/  }, function(evt, name) {
+  watch("./docs", { recursive: true, filter: /\.(py|toml|md)$/  }, function(evt, name) {
       // break update loop
       if (Date.now() - lastUpdate < 500) {
           return;


sh-rp (approved):
Approved, I think we / I should change the script that we have an initial "update docs" that wipes the output folder and rebuilds all files, and the watcher then should not wipe anything and just update the changed file. Should not be hard to do but for now I suppose this fix is goo.

PR #2685, opened 2025-05-26
adds parquet support to postgres via adbc

rudolfix:
### Description
adds parquet support to postgres via adbc


dlt/destinations/impl/postgres/factory.py
```diff
@@ -129,7 +133,14 @@ def _raw_capabilities(self) -> DestinationCapabilitiesContext:
         # https://www.postgresql.org/docs/current/limits.html
         caps = DestinationCapabilitiesContext()
         caps.preferred_loader_file_format = "insert_values"
-        caps.supported_loader_file_formats = ["insert_values", "csv", "model"]
+        try:
+            # supports adbc for direct parquet loading
+            import adbc_driver_postgresql.dbapi
+
+            caps.supported_loader_file_formats = ["insert_values", "csv", "parquet", "model"]
+        except ImportError:
```
zilto, line 141: The way this is currently setup, running the test `tests/load/pipeline/test_arrow_loading.py::test_load_arrow_item` raises an inexact error.  It doesn't mention that  adbc driver might be missing.

> Unsupported file format parquet in destination postgres. Supported formats: insert_values, csv, model. If postgres supports loading data via staging bucket, more formats may be available.

from
```python
# pipeline.py

    def _verify_destination_capabilities(
        self,
        caps: DestinationCapabilitiesContext,
        loader_file_format: TLoaderFileFormat,
    ) -> None:
        # verify loader file format
        if loader_file_format and loader_file_format not in caps.supported_loader_file_formats:
>           raise DestinationIncompatibleLoaderFileFormatException(
                self._destination.destination_name,
                (self._staging.destination_name if self._staging else None),
                loader_file_format,
                set(caps.supported_loader_file_formats),
            )
```

It seems odd that `Capabilities` are conditional and current venv instead of static. Instead of having the check on `Pipeline._verify_destination_capabilities()`. We could have a method on `DestinationCapabilitiesContext`

```python
class DestinationCapabilitiesContext:
  # ...
  def validate_property(self, property: str, value: Any) -> bool: 
     ...
```

It would be impossible for another developer to guess that `capabilities.supported_loader_file_format` is not a static value and it would be hard to catch

dlt/destinations/impl/postgres/postgres.py
```diff
@@ -51,6 +51,37 @@ def generate_sql(
...
+
+    def run(self) -> None:
+        self._config = self._job_client.config
+
+        from dlt.common.libs.pyarrow import pq_stream_with_new_columns
+        from dlt.common.libs.pyarrow import pyarrow
+        import adbc_driver_postgresql.dbapi as adbapi
+
+        def _iter_batches() -> Iterator[pyarrow.RecordBatch]:
+            for table in pq_stream_with_new_columns(self._file_path, ()):
```
zilto, line 67: Instead of referencing `self` in a nested function, this seems a bit more readable

```python
        # this takes an argument now
        def _iter_batches(file_path) -> Iterator[pyarrow.RecordBatch]:
            for table in pq_stream_with_new_columns(file_path, ()):
                yield from table.to_batches()

        with adbapi.connect(
            self._config.credentials.to_native_representation()
        ) as conn, conn.cursor() as cur:
            rows = cur.adbc_ingest(
                self.load_table_name,
                _iter_batches(self._file_path),  # pass `self` value here
```

zilto (changes requested):
All tests passed successfully. However, the `DestinationCapabilitiesContext` with imports inside a `try/except` seems risky. Our current approach can't provide useful exceptions

dlt/destinations/impl/postgres/factory.py
rudolfix: right! there was already a mechanism to do that (selectors) pls take a look. also there's a direct explanation when `parquet` is not available

zilto (dismissed):
Previous reviews were addressed. LGTM

I didn't pull the branch to try the feature. I expect the new tests in CI to be sufficient

zilto (approved):

PR #2689, opened 2025-05-27
Replace def _limit_clause_sql with sqlglot

anuunchin:
### Description

This PR combines two tasks:

1. This PR replaces the _limit_clause_sql function with sqlglot. Also aims to adjust the following functions involving manual query building:
    - `def row_counts()` in `dlt/destinations/dataset/dataset.py`
    - `def query()` in `dlt/destinations/dataset/relation.py`
    - `def get_stored_schema()` in `dlt/destinations/job_client_impl.py`
    - `def get_stored_schema_by_hash()` in `dlt/destinations/job_client_impl.py`
    - `def get_stored_state()` in `dlt/destinations/job_client_impl.py`

2. Adding `def where()` and `def order_by()` to `ReadableDBAPIRelation`, the task description of which is defined in #2777


sh-rp:
Just a note, things that could additionally be replace is the sql code in "row_counts" and where we get state and schema from the destination.

dlt/destinations/impl/mssql/sql_client.py
```diff
@@ -187,6 +187,3 @@ def _make_database_exception(cls, ex: Exception) -> Exception:
     @staticmethod
     def is_dbapi_exception(ex: Exception) -> bool:
         return isinstance(ex, pyodbc.Error)
-
-    def _limit_clause_sql(self, limit: int) -> Tuple[str, str]:
```
sh-rp, line 211: Good riddance :)

dlt/destinations/job_client_impl.py
```diff
@@ -523,36 +520,137 @@ def _from_db_type(
...
-        state_table = self.sql_client.make_qualified_table_name(self.schema.state_table_name)
-        loads_table = self.sql_client.make_qualified_table_name(self.schema.loads_table_name)
-        c_load_id, c_dlt_load_id, c_pipeline_name, c_status = self._norm_and_escape_columns(
-            "load_id", C_DLT_LOAD_ID, "pipeline_name", "status"
+        state_table_name = self.schema.state_table_name
+        loads_table_name = self.schema.loads_table_name
+        dataset_name = self.sql_client.dataset_name
+        catalog_name = self.sql_client.catalog_name(escape=False)
+
+        if self.config.destination_type == "clickhouse":
```
sh-rp: Just on a general note, the destination specific implementations should never leak into the base class. The right way to do it would be to have the correct implementation in the subclass. But with the changes I propose below, this problem should be gone in any case.

dlt/destinations/job_client_impl.py
```diff
@@ -523,36 +520,137 @@ def _from_db_type(
...
+            dataset_name = catalog_name
+            catalog_name = None
+
+        c_schema_name, c_inserted_at = [
+            self.schema.naming.normalize_path(p) for p in ["schema_name", "inserted_at"]
+        ]
+
+        table_expr = sqlglot.exp.Table(
+            this=sqlglot.exp.to_identifier(table_name, quoted=True),
+            db=sqlglot.exp.to_identifier(dataset_name, quoted=True),
```
sh-rp: If we use this new function to normalize queries (explained above), we will not have to add db and catalog everywhere.

dlt/destinations/dataset/relation.py
```diff
@@ -205,29 +207,47 @@ def _query(self) -> Any:
 
         dataset_schema = self._dataset.schema
 
-        table_name = self.sql_client.make_qualified_table_name(
-            dataset_schema.naming.normalize_path(self._table_name)
+        table_name = dataset_schema.naming.normalize_tables_path(self._table_name)
```
sh-rp: There is a change now on devel that runs the result of the query method through a qualifying and normalization step, have a look at what is there now, and you will see that you will not have to do any destination specific stuff or add database and catalog information here.

sh-rp (changes requested):
Thanks for the work on this, it's a good first start :) There has been new work on the transformations that makes this much easier. I also think the ultimate goal of this work is to remove as much handwritten sql from our codebase as possible (we have a lot of it in the sql_jobs file which handles merging, replacing and appending), but this is for future tickets.

Firstly I think you have to wait just a little longer until we have the upcoming transformations PR merged, as there are some more change in there. After that you can do the following:

* Have a look at the `def normalize_query` function that @rudolfix wrote. This will create a query with identifiers in the "dlt schema" space into a query with the correct fully qualified identifiers of the destination. It needs the sql_client and the sql_glot schema for this which you have in all the places where you are writing these queries. You can also look at this on the transformations decoupling branch which is more up to date.
* Update your queries to not add catalogs, databases or any aliases but rather run them through this normalize function. They should become much easier and all of the destination specific code should go away (clickhouse)
* I think we could add a dedicated file at `dlt/destinations/queries.py` where all these queries live and can be unit tested independently of any actual loading happening! The "normalize_query" function can go there too imho.
* Add unit tests for the queries. We can talk about what exactly needs to be tested here.

dlt/destinations/job_client_impl.py
anuunchin: 🫡

dlt/destinations/dataset/dataset.py
```diff
@@ -227,26 +228,30 @@ def row_counts(
...
-                query += f" WHERE {dlt_load_id_col} = '{load_id}'"
-            queries.append(query)
 
-        query = " UNION ALL ".join(queries)
+        union_all_expr = None
+
+        for table_name in selected_tables:
+            counts_expr = build_row_counts_expr(
+                table_name=table_name,
+                dlt_load_id_col=dlt_load_id_col,
```
zilto, line 250: This uses `dlt_load_id_col`, but the value seems to be `None` if not entering the `if load_id:` clauses (line 232). I think you would need to move `dlt_load_id_col = ...` (line 233) outside of this scope?

dlt/destinations/dataset/relation.py
```diff
@@ -1,29 +1,64 @@
...
 from dlt.common.schema.typing import TTableSchemaColumns
 from dlt.common.typing import Self
 from dlt.common.exceptions import TypeErrorWithKnownTypes
 from dlt.transformations import lineage
 from dlt.destinations.sql_client import SqlClientBase, WithSqlClient
-from dlt.destinations.dataset.exceptions import (
-    ReadableRelationHasQueryException,
-)
-from dlt.destinations.dataset.utils import normalize_query
+from dlt.destinations.queries import normalize_query, build_select_expr
```
zilto, line 33: I like the import path and name `dlt.destinations.queries` for this purpose :)

dlt/destinations/dataset/relation.py
```diff
@@ -1,29 +1,64 @@
...
 
 if TYPE_CHECKING:
     from dlt.destinations.dataset.dataset import ReadableDBAPIDataset
 else:
     ReadableDBAPIDataset = Any
 
 
+class FilterOp(Enum):
```
zilto, line 44: Clean use of enum. Enums are good for internal components whereas `Literal` are better for user-facing API. For example, the user can write `do_filter(op="EQ")` without having to do `from dlt ... import FilterOp` then `do_filter(op=FilterOp.EQ)`

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
         limit: int = None,
         selected_columns: Sequence[str] = None,
         normalize_query: bool = True,
+        sqlglot_expression: sge.Select = None,
```
zilto: We can use `@typing.overload` to add more signatures and indicate that `BaseReadableDBAPIRelation` should be initialized with `table_name` OR `provided_query` OR `sqlglot_expression` (receiving more than one of those has undefined behavior and should error) (LLMs are pretty good a generating these `@overload`

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
+                ),
+            )
+            if limit:
+                self._sqlglot_expression = self._sqlglot_expression.limit(limit)
 
-        selected_columns = (
-            self._selected_columns
-            if self._selected_columns
-            else dataset_schema.get_table_columns(self._table_name).keys()
+    def _query(self) -> Any:
```
zilto: Type annotate `sge.Select` instead of `Any`

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
+        rel._sqlglot_expression = rel._sqlglot_expression.limit(limit)
         return rel
 
     def select(self, *columns: str) -> Self:
-        if self._provided_query:
-            raise ReadableRelationHasQueryException("select")
         rel = self.__copy__()
         rel._selected_columns = columns
+        new_proj = [sge.Column(this=sge.to_identifier(col, quoted=True)) for col in columns]
+        rel._sqlglot_expression.set("expressions", new_proj)
```
zilto: I believe there's a bug here? (maybe I'm wrong).  The intended behavior should be in the tests

```python
# table has columns: foo, bar, baz

relation = ...
relation.select("foo")  # returns table: foo; Good
relation.select("foo", "bar").select("bar")  # returns table: bar; Good
relation.select("foo").select("bar")   # returns table: bar; Bad
relation.select("bar").select("foo", "baz")  # returns table: foo, baz; Bad
``` 

I labeled cases as `Bad` because they don't match what chaining `.select()` on Ibis and SQLGlot expressions does. I think doing `rel._sqlglot_expression.select()` (multiple Select clauses) is ok because it matches the code the user wrote (even if it's bad). 

We can probably add some good try/except to catch SQLGlot exceptions.

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
+        if isinstance(operator, str):
+            try:
+                operator = FilterOp[operator.upper()]
+            except KeyError:
+                raise ValueError(
+                    f"Invalid operator '{operator}'. Expected one of:"
+                    f" {[op.name.lower() for op in FilterOp]}"
+                )
+
+        def _build_typed_literal(v: Any, sqlglot_type: sge.DataType) -> sge.Expression:
```
zilto: Fix type annotation If `sqlglot_type` can be `None` (as suggested by line 348)

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
+        column_name: str,
+        operator: Union[FilterOp, str],
+        value: Any,
+    ) -> Self:
+        if isinstance(operator, str):
+            try:
+                operator = FilterOp[operator.upper()]
+            except KeyError:
+                raise ValueError(
+                    f"Invalid operator '{operator}'. Expected one of:"
```
zilto: Use backticks \` in exception. In `dlt.common.exceptions`, there are no `ValueErrorWithKnownValues` and `TypeErrorWithKnownTypes` that standardize this type of message

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
+            condition = condition_cls(this=column, expression=value_expr)
+
+        rel = self.__copy__()
+        rel._sqlglot_expression = rel._sqlglot_expression.where(condition)
+        return rel
+
+    def filter(  # noqa: A003
+        self,
+        column_name: str,
+        operator: Union[FilterOp, str],
```
zilto: For a user-facing interface, using `Literal` would be better than `Union[Enum, str]`

dlt/destinations/job_client_impl.py
```diff
@@ -408,8 +403,11 @@ def create_load_job(
     def complete_load(self, load_id: str) -> None:
         name = self.sql_client.make_qualified_table_name(self.schema.loads_table_name)
         now_ts = pendulum.now()
+        load_table_columns = ", ".join(
+            self.sql_client.escape_column_name(col) for col in self.loads_table_schema_columns
+        )
         self.sql_client.execute_sql(
-            f"INSERT INTO {name}({self.loads_table_schema_columns}) VALUES(%s, %s, %s, %s, %s);",
+            f"INSERT INTO {name}({load_table_columns}) VALUES(%s, %s, %s, %s, %s);",
```
zilto, line 414: Can probably replace these few lines with an SQLGlot INSERT statement. The trick is:

1. write the SQL `str`
2. parse with SQLGlot
3. print the SQLGlot representation
4. paste the representation back into the code

You should have something starting with `exp.Insert`

dlt/destinations/job_client_impl.py
```diff
@@ -524,36 +522,111 @@ def _from_db_type(
...
-                f"SELECT {self.version_table_schema_columns} FROM {name}"
-                f" ORDER BY {c_inserted_at} DESC;"
-            )
-            return self._row_to_schema_info(query)
-        else:
-            query = (
-                f"SELECT {self.version_table_schema_columns} FROM {name} WHERE {c_schema_name} = %s"
-                f" ORDER BY {c_inserted_at} DESC;"
+        table_name = self.schema.version_table_name
+        c_schema_name, c_inserted_at = [
```
zilto: list comprehension + unpacking seem unnecessary; split this into two lines; (I see how this comes from the previous code)

dlt/destinations/job_client_impl.py
```diff
@@ -524,36 +522,111 @@ def _from_db_type(
...
+        return self._row_to_schema_info(normalized_query)
 
     def get_stored_state(self, pipeline_name: str) -> StateInfo:
-        state_table = self.sql_client.make_qualified_table_name(self.schema.state_table_name)
-        loads_table = self.sql_client.make_qualified_table_name(self.schema.loads_table_name)
-        c_load_id, c_dlt_load_id, c_pipeline_name, c_status = self._norm_and_escape_columns(
-            C_DLT_LOADS_TABLE_LOAD_ID, C_DLT_LOAD_ID, "pipeline_name", "status"
+        state_table_name = self.schema.state_table_name
+        loads_table_name = self.schema.loads_table_name
+        c_load_id, c_dlt_load_id, c_pipeline_name, c_status = [
```
zilto: split assignment into multiple lines

dlt/destinations/job_client_impl.py
```diff
@@ -524,36 +522,111 @@ def _from_db_type(
...
+            self.schema.naming.normalize_path(p)
+            for p in ["load_id", C_DLT_LOAD_ID, "pipeline_name", "status"]
```
zilto: "load_id"  now has a constant `C_DLT_LOADS_TABLE_LOAD_ID` in `dlt.common.schema.typing`

dlt/destinations/job_client_impl.py
```diff
@@ -524,36 +522,111 @@ def _from_db_type(
...
+            self.schema.naming.normalize_path(p) for p in ["schema_name", "inserted_at"]
+        ]
+
+        select_expr = build_select_expr(
+            table_name=table_name,
+            selected_columns=list(self.version_table_schema_columns.keys()),
+            quoted_identifiers=True,
+        )
+
+        order_expr = sqlglot.exp.Ordered(
```
zilto: In this file, there's several places where `sqlglot.exp` could be replaced by `sge` (already imported in this file)

dlt/destinations/job_client_impl.py
```diff
@@ -408,8 +403,11 @@ def create_load_job(
     def complete_load(self, load_id: str) -> None:
         name = self.sql_client.make_qualified_table_name(self.schema.loads_table_name)
         now_ts = pendulum.now()
+        load_table_columns = ", ".join(
+            self.sql_client.escape_column_name(col) for col in self.loads_table_schema_columns
```
zilto, line 411: (not specific to this line) The majority of the logic of the methods on `SqlJobClientBase` should be moved to `dlt.destinations.queries`. 

1. `dlt.destinations.queries` contains `completed_load(load_id)`, `get_stored_schema(schema_name)`, `get_stored_state(pipeline_name)`, etc.; These functions returns SQLGlot objects
2. Methods in this module only import the queries, generate the destination-specific version, and execute the query on the client

A good way to spot these refactors is to check where the `self` is used. It's always a single line to access the `sql_client`. Doing this small refactor helps us keep the library clean

dlt/destinations/job_client_impl.py
```diff
@@ -837,9 +923,11 @@ def _commit_schema_update(self, schema: Schema, schema_str: str) -> None:
         now_ts = pendulum.now()
         name = self.sql_client.make_qualified_table_name(self.schema.version_table_name)
         # values =  schema.version_hash, schema.name, schema.version, schema.ENGINE_VERSION, str(now_ts), schema_str
+        version_table_columns = ", ".join(
+            self.sql_client.escape_column_name(col) for col in self.version_table_schema_columns
+        )
         self.sql_client.execute_sql(
-            f"INSERT INTO {name}({self.version_table_schema_columns}) VALUES (%s, %s, %s, %s, %s,"
-            " %s);",
+            f"INSERT INTO {name}({version_table_columns}) VALUES (%s, %s, %s, %s, %s, %s);",
```
zilto, line 943: Like above, try to use SQLGlot's Insert expression

dlt/transformations/lineage.py
```diff
@@ -41,29 +42,35 @@ def create_sqlglot_schema(
...
+        column_mapping = map_columns(columns)
         if column_mapping:
-            # table_name = sql_client.make_qualified_table_name_path(table_name, quote=False, casefold=False)[-1]
             sqlglot_schema[table_name] = column_mapping
 
+    if schema.state_table_name not in sqlglot_schema:
+        state_columns = pipeline_state_table()["columns"]
+        column_mapping = map_columns(state_columns)
+        if column_mapping:
+            sqlglot_schema[schema.state_table_name] = column_mapping
```
zilto: Don't assign directly. You can use `sqlglot.schema.Schema.add_table()`

zilto (changes requested):
Ok, I took my time with this review, but everything is great! There are a few clean up changes and slight refactors requested. Good job!

dlt/destinations/dataset/dataset.py
sh-rp: if you look at `build_row_counts_expr` this looks correct to me. you either provide both values or none.

dlt/destinations/dataset/dataset.py
```diff
@@ -227,26 +228,30 @@ def row_counts(
...
+            )
+            if union_all_expr is None:
+                union_all_expr = counts_expr
+            else:
+                union_all_expr = union_all_expr.union(counts_expr, distinct=False)
 
+        dialect = self.destination_client.capabilities.sqlglot_dialect
+        query = union_all_expr.sql(dialect)
         # Execute query and build result dict
         return self(query)
```
sh-rp: I think you can return a `ReadableDBAPIRelation` with the `sqlglot_expression` set to the expression you have built. Then the normalization and rendering of the query will happen the same way as with all other `ReadableDBAPIRelation` relations and this query can also be loaded in a transformation.

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
         limit: int = None,
         selected_columns: Sequence[str] = None,
```
sh-rp: we don't need selected columns and limit anymore, you can store this info in the query object.

sh-rp: You can also rename "provided_query" to "query" and accept both a string or a sqlglot query and react accordingly in the init function. wdyt?

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
-        if self._limit:
-            maybe_limit_clause_1, maybe_limit_clause_2 = self.sql_client._limit_clause_sql(
-                self._limit
+        self._sqlglot_expression: sge.Select = None
+        if sqlglot_expression:
+            self._sqlglot_expression = sqlglot_expression
+        elif provided_query:
+            self._sqlglot_expression = cast(
+                sge.Select,
+                parse_one(
```
sh-rp: you can do `maybe_parse` here, and it will either take the expression if it is one, or parse a string.

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
+        else:
+            self._sqlglot_expression = build_select_expr(
+                table_name=table_name,
+                selected_columns=(
+                    list(selected_columns)
+                    if selected_columns
+                    else list(self._dataset.schema.get_table_columns(table_name).keys())
+                ),
+            )
+            if limit:
```
sh-rp: this should be done in the `limit` function and then a new releation object with the sqlglot expression with limit applied should be returned there.

dlt/destinations/dataset/relation.py
sh-rp: We specified it differently in the ticket, but we could also just raise if you try to chain selects, there is no reason to do it for this simple relation.

dlt/destinations/dataset/relation.py
sh-rp: I would also suggest to move this literal helper function to `dlt/common/libs/sqlglot.py`, it might be useful elsewhere!

dlt/destinations/dataset/relation.py
```diff
@@ -213,72 +243,150 @@ def __init__(
...
+
+        sqlgot_type = to_sqlglot_type(
+            dlt_type=self.columns_schema[column_name].get("data_type"),
+            precision=self.columns_schema[column_name].get("precision"),
+            timezone=self.columns_schema[column_name].get("timezone"),
+            nullable=self.columns_schema[column_name].get("nullable"),
+        )
+
+        value_expr: Union[sge.Expression, sge.Tuple]
+        if isinstance(value, Iterable) and not isinstance(value, (str, bytes)):
```
sh-rp: this part can also be in this helper function mentioned above

dlt/destinations/dataset/relation.py
sh-rp: I would also put a literal type on the operator so you have good autocomplete in the editor. This is what you mean @zilto , right?

dlt/destinations/job_client_impl.py
sh-rp: I don't think you need to do this at all if you send the query through the "normalize_query" function at the end. You can use the non-normalized identifiers everywhere.

dlt/destinations/job_client_impl.py
sh-rp: same comment as above, not needed at all.

dlt/destinations/job_client_impl.py
sh-rp: I agree, all this query code should go into the queries file

dlt/destinations/job_client_impl.py
```diff
@@ -573,16 +646,29 @@ def _norm_and_escape_columns(self, *columns: str) -> Iterator[str]:
...
-            f" {table_name} WHERE {c_version_hash} = %s {maybe_limit_clause_2};"
+        where_condition = sqlglot.exp.EQ(
+            this=sqlglot.exp.Column(this=sqlglot.exp.Identifier(this=c_version_hash, quoted=True)),
+            expression=sqlglot.exp.Literal.string(version_hash),
         )
-        return self._row_to_schema_info(query, version_hash)
+
+        select_expr = select_expr.where(where_condition).limit(1)
+
+        dialect = self.sql_client.capabilities.sqlglot_dialect
```
sh-rp: this whole code block is repeated 4 times, you could create a helper function on the jobclient that renders a sqlglot query object into the the current dialect.

tests/destinations/test_readable_dbapi_dataset.py
```diff
@@ -94,18 +93,19 @@ def test_copy_and_chaining(dataset_type: TDatasetType) -> None:
...
-    assert relation._table_name == relation2._table_name
     assert relation._provided_query == relation2._provided_query
     assert relation._selected_columns == relation2._selected_columns
+    assert relation._sqlglot_expression == relation2._sqlglot_expression
 
     # test copy while chaining limit
     relation3 = relation2.limit(22)
     assert relation2 != relation3
-    assert relation2._limit != relation3._limit
+    assert relation2._sqlglot_expression != relation3._sqlglot_expression
```
sh-rp, line 106: good :)

tests/destinations/test_readable_dbapi_dataset.py
```diff
@@ -148,19 +148,19 @@ def test_computed_schema_columns(dataset_type: TDatasetType) -> None:
...
 
-    with pytest.raises(ReadableRelationHasQueryException):
-        relation.limit(5)
+    query = relation.limit(5)._query()
+    assert "SELECT * FROM something LIMIT 5" == query
 
-    with pytest.raises(ReadableRelationHasQueryException):
-        relation.head()
+    query = relation.head()._query()
+    assert "SELECT * FROM something LIMIT 5" == query
```
sh-rp: I think we were planning to not have star selects here or do is there a reason for this? I think it should select all columns explicitely by default.

tests/load/test_read_interfaces.py
```diff
@@ -677,6 +677,66 @@ def test_column_selection(populated_pipeline: Pipeline) -> None:
...
 
+@pytest.mark.no_load
+@pytest.mark.essential
+@pytest.mark.parametrize(
+    "populated_pipeline",
+    configs,
+    indirect=True,
+    ids=lambda x: x.name,
+)
+def test_order_by(populated_pipeline: Pipeline) -> None:
```
sh-rp, line 716: i would also check chained orderings, so order by two different columsn consecutively.

sh-rp (changes requested):
Very cool PR so far, but we can make it even a bit better :)

dlt/destinations/dataset/relation.py
anuunchin: this is good to know

dlt/destinations/dataset/relation.py
anuunchin: This war probably meant as str 👀

dlt/destinations/dataset/dataset.py
anuunchin: owwwwww

tests/load/test_read_interfaces.py
```diff
@@ -677,6 +680,115 @@ def test_column_selection(populated_pipeline: Pipeline) -> None:
...
+    assert asc_ids == list(range(20))
+
+    desc_ids = [row[0] for row in table_relationship.order_by("id", "desc").limit(20).fetchall()]
+    assert desc_ids == list(range(total_records - 1, total_records - 21, -1))
+
+    chained = [row[0] for row in table_relationship.order_by("id").limit(5).select("id").fetchall()]
+    assert chained == list(range(5))
+
+    @dlt.resource
+    def chained_order():
```
sh-rp, line 730: the pipelines dataset are not wiped between tests to make them faster, so adding a new resource here has broken other tests.

anuunchin:
The relation logic was moved to #2814.
The logic replacing query strings to sqlglot expr to #2824

PR #2700, opened 2025-05-30
docs: fix invalid bigquery reference in athena destination

goober:
### Description
Fixes a tiny comment where `bigquery` was referenced instead of `athena`


sh-rp:
Hey @goober, thanks for finding this :)

PR #2703, opened 2025-05-30
motherduck destination config improvement: uppercase env var

djudjuu:
actually there was no real bug and the code worked, but it didnt match the dlt-convention around casing env vars. this pr just fixes this

### DIFF
--- a/tests/e2e/helpers/studio/test_e2e.py
+++ b/tests/e2e/helpers/studio/test_e2e.py
@@ -193,7 +193,7 @@ def test_page_loads(page: Page):
     _go_home(page)
     page.get_by_role("link", name="never_run_pipeline").click()
 
-    expect(page.get_by_text(app_strings.sync_status_error_text)).to_be_visible()
+    expect(page.get_by_text(app_strings.sync_status_success_text.split("from")[0])).to_be_visible()
     expect(page.get_by_text("_storage/.dlt/pipelines/never_run_pipeline")).to_be_visible()
 
     # check schema info (this is the yaml part)


rudolfix (changes requested):
in motherduck docs the env var is lowercare
https://motherduck.com/docs/key-tasks/authenticating-and-connecting-to-motherduck/authenticating-to-motherduck/

so maybe we should check both lower and upper case vars.

sh-rp (approved):
lgtm

rudolfix (approved):

sh-rp:
@rudolfix and @djudjuu  actually I have one question about this change. If I undertand the code / credentials correctly, the motherduck credentials assume that motherduck will automatically pick up the correct env var if it is set. So the tests we have here do not confirm that it will work, right? So it would be good to test out the connection to motherduck manually just to see wether an uppercase token works.

djudjuu:
so i used a simple script to test it and it works, we could also just add it as a real test:
```
def test_motherduck_client_compatibility() -> None:
    os.environ["HOME"] = "/tmp"
    import duckdb
    import dlt

    credentials = dlt.secrets.get(
        "destination.motherduck.credentials", expected_type=MotherDuckCredentials
    )
    assert credentials.password
    # os.environ[token_key] = credentials.password
    os.environ["MOTHERDUCK_TOKEN"] = credentials.password

    # Create connection to the default database
    conn = duckdb.connect("md:my_db")

    # Run query
    # conn.sql("CREATE TABLE items (item VARCHAR, value DECIMAL(10, 2), count INTEGER)")
    # conn.sql("INSERT INTO items VALUES ('jeans', 20.0, 1), ('hammer', 42.2, 2)")
    # res = conn.sql("SELECT * FROM items")

    # Close the connection
    conn.close()
```

Issue #2704, opened 2025-05-30
bug: `ReadableIbisRelation` mutates dataset  `Schema`


zilto:
### dlt version

1.11.0

### Describe the problem

Accessing a non-existent attribute on `ReadableIbisRelation` (i.e., `pipeline.dataset(dataset_type="ibis")`) returns an unbound empty Ibis table and mutates `Schema._schema_mapping` stored on the `ReadableDBAPIDataset` object. (see simple repro).

The bug is not present for `ReadableRelation`

### Expected behavior

It should raise that the attribute doesn't exist. Accessing an attribute should not mutate state.

### Steps to reproduce

```python
pipeline = dlt.pipeline("bugged", destination="duckdb")
dataset = pipeline.dataset(dataset_type="ibis")

schema_v0 = dataset.schema
schema_v0.tables.keys()   # ["_dlt_version", "_dlt_loads"]
schema_v0.data_tables()  #[ ]
schema_v0.dlt_tables()  # dict for `_dlt_version` and `_dlt_loads`

# BUG
dataset.foo
# or
dataset.table("foo")
schema_v1 = dataset.schema
schema_v0.tables.keys()   # ["_dlt_version", "_dlt_loads", "foo"]
schema_v0.data_tables()  # [ ]
schema_v0.dlt_tables()  # dict for `_dlt_version` and `_dlt_loads`
```

### Additional information

There are a lot of complexities with `ReadableRelation` objects. In particular, circular references between datasets and relations, type annotations, proxying Ibis methods, etc. I have ideas for simpler designs, but I would need to understand the initial requirements that lead to current implementation

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

zilto:
bugged feature deprecated

PR #2705, opened 2025-06-01
[transformations] decouples sqlglot lineage and schema generation from destination identifiers

rudolfix:
### Description
1. Makes schemas generated by lineage and ibis modules to use dlt schema identifiers without any table name expansion (clickhouse), case folding (snowflake) etc.
2. this makes queries and expressions destination agnosting.
3. identifier normalization, expansions and qualification of tables, star expansion etc. is done after sqlglot schema generation and lineage
4. all queries are aliased (top select) so dlt identifiers are preserved - all the identifier adaptation to destination is hidden in the query
5. I allowed for UNION selects to work with lineage (row counts didn't)
6. I also added native cursor for arrow in Snowflake

Tests
1. I was able to enable all destinations for lineage and read interface tests. all queries are destination agnostic. column selection uses schema identifiers etc. so many exceptions are removed
2. several tests will fail. I will need help with fxing them. I'm pretty sure we can enable more test cases now
3. Snowflake native cursor does not let setting chunk size. some test must be fixed for sure

### DIFF
--- a/dlt/common/destination/dataset.py
+++ b/dlt/common/destination/dataset.py
@@ -39,14 +39,12 @@ class SupportsReadableRelation:
     sql glot query analysis and lineage. dlt hints for columns are kept in some cases. Refere to <docs-page> for more details.
     """
 
-    def query(self, qualified: bool = False) -> Any:
-        """Returns the sql query that represents the relation
-
-        Args:
-            qualified (bool, optional): Whether to return the qualified query. Defaults to False.
+    def query(self) -> Any:
+        """Returns the sql query that represents the relation. The query will be qualified, quoted and escaped
+           according to a SQL dialect that the destination uses.
 
         Returns:
-            Any: The sql query that represents the relation
+            Any: The qualified sql query that represents the relation
         """
         raise NotImplementedError("Query is not supported for this relation")
 
--- a/dlt/common/destination/dataset.py
+++ b/dlt/common/destination/dataset.py
@@ -306,7 +306,7 @@ class SupportsReadableDataset(Generic[TReadableRelation], Protocol):
         data_tables: bool = True,
         dlt_tables: bool = False,
         table_names: List[str] = None,
-        load_id: str = None
+        load_id: str = None,
     ) -> SupportsReadableRelation:
         """Returns the row counts of the dataset
 
--- a/dlt/destinations/dataset/dataset.py
+++ b/dlt/destinations/dataset/dataset.py
@@ -157,16 +157,25 @@ class ReadableDBAPIDataset(SupportsReadableDataset[ReadableIbisRelation]):
 
     def __call__(self, query: Any) -> ReadableDBAPIRelation:
         # TODO: accept other query types and return a right relation: sqlglot (DBAPI) and ibis (Expr)
+        # TODO: parse query as ibis relation, however ibis will quote that query when compiling to sql
         return ReadableDBAPIRelation(readable_dataset=self, provided_query=query)
 
     def table(self, table_name: str) -> ReadableIbisRelation:
         # we can create an ibis powered relation if ibis is available
         relation: BaseReadableDBAPIRelation
         if self._dataset_type == "ibis":
+            from dlt.common.schema.utils import new_table
             from dlt.helpers.ibis import create_unbound_ibis_table
             from dlt.destinations.dataset.ibis_relation import ReadableIbisRelation
 
-            unbound_table = create_unbound_ibis_table(self.sql_client, self.schema, table_name)
+            # allow to create empty tables without schema to unify behavior with default relation
+            schema = self.schema
+            if table_name not in schema.tables:
+                schema.update_table(new_table(table_name))
+
+            unbound_table = create_unbound_ibis_table(
+                schema, self.sql_client.dataset_name, table_name
+            )
             relation = ReadableIbisRelation(
                 readable_dataset=self,
                 ibis_object=unbound_table,
@@ -212,8 +221,8 @@ class ReadableDBAPIDataset(SupportsReadableDataset[ReadableIbisRelation]):
         queries = []
         for table in selected_tables:
             query = (
-                f"SELECT '{table}' as table_name, COUNT(*) as row_count FROM"
-                f" {self.sql_client.make_qualified_table_name(table)}"
+                f"SELECT '{table}' as table_name, COUNT(1) as row_count FROM"
+                f" {self.sql_client.make_qualified_table_name(table, quote=False, casefold=False)}"
             )
             if load_id:
                 query += f" WHERE {dlt_load_id_col} = '{load_id}'"
--- a/dlt/destinations/dataset/ibis_relation.py
+++ b/dlt/destinations/dataset/ibis_relation.py
@@ -2,8 +2,6 @@ from typing import TYPE_CHECKING, Any, Union, Sequence
 from functools import partial
 import sqlglot
 
-from dlt.common.schema.typing import TTableSchemaColumns
-
 from dlt.destinations.dataset.relation import BaseReadableDBAPIRelation
 
 
@@ -35,22 +33,27 @@ class ReadableIbisRelation(BaseReadableDBAPIRelation):
         self._ibis_object = ibis_object
 
     def _query(self) -> Any:
-        """build the query"""
         from dlt.helpers.ibis import ibis
 
-        target_dialect = self._dataset.sql_client.capabilities.sqlglot_dialect
+        ibis_expr = self._ibis_object
+        caps = self._dataset.sql_client.capabilities
+        target_dialect = caps.sqlglot_dialect
 
         # render sql directly if possible
+        # NOTE: ibis is optimized for reading a real schema from db and pushing back optimized sql
+        #  - it quotes all identifiers, there's no option to get unqoted query
+        #  - it converts full lists of column names back into star
+        #  - it optimizes the query inside, adds meaningless aliases to all tables etc.
         if target_dialect not in TRANSPILE_VIA_DEFAULT:
             if target_dialect == "tsql":
                 # NOTE: Ibis uses the product name "mssql" as the dialect instead of the official "tsql".
-                return ibis.to_sql(self._ibis_object, dialect="mssql")
+                return ibis.to_sql(ibis_expr, dialect="mssql")
             else:
-                return ibis.to_sql(self._ibis_object, dialect=target_dialect)
+                return ibis.to_sql(ibis_expr, dialect=target_dialect)
 
         # here we need to transpile to ibis default and transpile back to target with sqlglot
         # NOTE: ibis defaults to the duckdb dialect, if a dialect is not passed
-        sql = ibis.to_sql(self._ibis_object)
+        sql = ibis.to_sql(ibis_expr)
         sql = sqlglot.transpile(sql, write=target_dialect)[0]
         return sql
 
@@ -69,16 +72,6 @@ class ReadableIbisRelation(BaseReadableDBAPIRelation):
             for k, v in kwargs.items()
         }
 
-        # casefold string params, we assume these are column names
-        args = tuple(
-            self.sql_client.capabilities.casefold_identifier(arg) if isinstance(arg, str) else arg
-            for arg in args
-        )
-        kwargs = {
-            k: self.sql_client.capabilities.casefold_identifier(v) if isinstance(v, str) else v
-            for k, v in kwargs.items()
-        }
-
         # Call it with provided args
         result = method(*args, **kwargs)
 
@@ -89,11 +82,6 @@ class ReadableIbisRelation(BaseReadableDBAPIRelation):
         """Wrap all callable attributes of the expression"""
 
         attr = getattr(self._ibis_object, name, None)
-
-        # try casefolded name for ibis columns access
-        if attr is None:
-            name = self._dataset._sql_client.capabilities.casefold_identifier(name)
-            attr = getattr(self._ibis_object, name, None)
 
         if attr is None:
             raise AttributeError(
@@ -107,12 +95,7 @@ class ReadableIbisRelation(BaseReadableDBAPIRelation):
         return partial(self._proxy_expression_method, name)
 
     def __getitem__(self, columns: Union[str, Sequence[str]]) -> "ReadableIbisRelation":
-        # casefold column-names
-        if isinstance(columns, str):
-            columns = self.sql_client.capabilities.casefold_identifier(columns)
-        elif isinstance(columns, Sequence):
-            columns = [self.sql_client.capabilities.casefold_identifier(col) for col in columns]
-        else:
+        if not isinstance(columns, (str, Sequence)):
             raise TypeError(
                 f"Invalid argument type: {type(columns).__name__}, requires a sequence of column"
                 " names Sequence[str] or a single column name str"
--- a/dlt/destinations/dataset/relation.py
+++ b/dlt/destinations/dataset/relation.py
@@ -1,19 +1,20 @@
-from typing import Any, Generator, Sequence, Type, Union, TYPE_CHECKING
+from typing import Any, Dict, Generator, Optional, Sequence, Tuple, Type, Union, TYPE_CHECKING
 from contextlib import contextmanager
 
 import sqlglot
+import sqlglot.expressions as sge
 
 from dlt.common.destination.dataset import (
     SupportsReadableRelation,
 )
-from dlt.destinations.sql_client import SqlClientBase, WithSqlClient
+
 from dlt.common.schema.typing import TTableSchemaColumns
 from dlt.common.typing import Self
-from dlt.transformations import lineage
 
+from dlt.transformations import lineage
+from dlt.destinations.sql_client import SqlClientBase, WithSqlClient
 from dlt.destinations.dataset.exceptions import (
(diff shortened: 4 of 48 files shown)


dlt/destinations/dataset/dataset.py
```diff
@@ -157,16 +157,25 @@ def _ensure_schema(self) -> None:
...
     def table(self, table_name: str) -> ReadableIbisRelation:
         # we can create an ibis powered relation if ibis is available
         relation: BaseReadableDBAPIRelation
         if self._dataset_type == "ibis":
+            from dlt.common.schema.utils import new_table
             from dlt.helpers.ibis import create_unbound_ibis_table
             from dlt.destinations.dataset.ibis_relation import ReadableIbisRelation
 
-            unbound_table = create_unbound_ibis_table(self.sql_client, self.schema, table_name)
+            # allow to create empty tables without schema to unify behavior with default relation
```
sh-rp: I think this should fail. We know the full schema and the ibis relation should not allow the users to create relations for tables that do not exist. Or what is the exact purpose of this?

dlt/destinations/dataset/dataset.py
```diff
@@ -157,16 +157,25 @@ def _ensure_schema(self) -> None:
...
+            schema = self.schema
+            if table_name not in schema.tables:
+                schema.update_table(new_table(table_name))
```
sh-rp: I also think the schema of the dataset should be immutable, no? Adding tables to the schema just because they were accessed (but don't actually exist) also does not make sense.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -69,16 +72,6 @@ def _proxy_expression_method(self, method_name: str, *args: Any, **kwargs: Any)
             for k, v in kwargs.items()
         }
 
-        # casefold string params, we assume these are column names
```
sh-rp, line 72: getting rid of this is very cool!

dlt/destinations/impl/snowflake/sql_client.py
```diff
@@ -16,18 +16,31 @@
...
+    def iter_df(self, chunk_size: int) -> Generator[DataFrame, None, None]:
+        # full frame
+        if not chunk_size:
+            yield self.native_cursor.fetch_pandas_all()
+            return
+        # iterate
+        # NOTE: no way to impact chunk size
+        yield from self.native_cursor.fetch_pandas_batches()
+
+    def iter_arrow(self, chunk_size: int) -> Generator[ArrowTable, None, None]:
```
sh-rp, line 36: cool!

dlt/transformations/lineage.py
```diff
@@ -30,47 +30,42 @@
...
-    dataset_catalog = sql_client.make_qualified_table_name_path(None, escape=False)
-    if len(dataset_catalog) == 2:
-        catalog, database = dataset_catalog
-        nested_schema = {catalog: {database: sqlglot_schema}}
-    else:
-        (database,) = dataset_catalog
-        nested_schema = {database: sqlglot_schema}  # type: ignore
-
-    # NOTE: normalize=False is important as we already have normalized identifiers
+    nested_schema = {dataset_name: sqlglot_schema}
```
sh-rp, line 68: Are you sure there are no cases where we need the catalog level? In snowflake for example you can join tables across databases (which is catalog here) and you would need all three nesting levels in the query as well as the lineage. We do not support cross dataset transformations right now because I wanted to not expand the scope, but we will soon imho.

dlt/transformations/lineage.py
```diff
@@ -136,15 +131,24 @@ def compute_columns_schema(
             schema=sqlglot_schema,
             dialect=dialect,
             infer_schema=infer_sqlglot_schema,
+            quote_identifiers=False,
+            expand_stars=True,
```
sh-rp, line 136: This is true by default

dlt/helpers/ibis.py
```diff
@@ -195,35 +193,19 @@ def _ignore_hstore(conn: Any, name: Any) -> Any:
...
-    table_path = sql_client.make_qualified_table_name_path(table_name, escape=False)
-
-    catalog = None
-    if len(table_path) == 3:
-        catalog, database, table = table_path
-    else:
-        database, table = table_path
-
     # create unbound ibis table and return in dlt wrapper
-    unbound_table = ibis.table(schema=ibis_schema, name=table, database=database, catalog=catalog)
```
sh-rp, line 227: I wrote a comment further below, I think we should keep all three levels. There are databases where you can join across catalogs and I don't think we should drop this.

dlt/transformations/transformation.py
```diff
@@ -140,25 +141,31 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
...
             )
-
-        select_query = transformation_result.query(qualified=True)
         # compute lineage
         computed_columns: TTableSchemaColumns = {}
         all_columns: TTableSchemaColumns = columns or {}
         # strict lineage!
-        computed_columns = transformation_result.compute_columns_schema(
+        # TODO: make it a public method
+        # TODO: why schema inference and anonymous columns are wrong? we do not want columns without
```
sh-rp, line 157: what do you mean with this comment?

dlt/transformations/transformation.py
```diff
@@ -140,25 +141,31 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
...
+        computed_columns, parsed_query, _ = transformation_result._compute_columns_schema(
             infer_sqlglot_schema=False,
             allow_anonymous_columns=False,
             allow_partial=False,
         )
+        # TODO: expose a method to render query
+        select_query = parsed_query.sql(
+            dialect=transformation_result.sql_client.capabilities.sqlglot_dialect
+        )
+        # TODO: why? don't we prevent empty column schemas above?
```
sh-rp, line 167: The idea is, that there may be cases where the lineage fails and can't figure out columns, so the user can supply their own hints. We can just prevent transformations where we can't automatically discover the columns, but I think having a documented fallback is better. Before this I had this setting for "strict" lineage which would control this, but I removed it because you didn't like it (I still think it could be useful but we can wait for user feedback on this)

docs/website/docs/dlt-ecosystem/destinations/motherduck.md
```diff
@@ -54,7 +54,7 @@ Motherduck now supports configurable **access tokens**. Please refer to the [doc
 
 You can pass token in a native Motherduck environment variable:
 ```sh
-export motherduck_token='<token>'
+export MOTHERDUCK_TOKEN='<token>'
```
sh-rp, line 57: Hehe, this PR is completely hijacked now. I think we should probably keep it small caps, as linux and mac have case sensitive env var.s

dlt/transformations/lineage.py
```diff
@@ -30,47 +30,42 @@
 
 def create_sqlglot_schema(
```
sh-rp, line 31: The reason I was using `make_qualified_table_name_path` and thus relying on the destination is because of the expanded table names, where the tables are called `dataset__table1`. I agree with you and @zilto that it would be nicer to have the sqlglot / lineage completely decoupled from the destination. However if we leave it like this, we will loose the ability to do cross dataset queries / transformation on any destination that has these expanded table names, because there can be tables with the same name in both virtual datasets and our code will not be able to see the difference. Currently the lineage does not support cross dataset transformation, but it will be quite easy for @zilto to add and then we will see this problem. We could however decide to not support these on these destinations in favor or simpler and more maintainable code.

sh-rp: @zilto for reference what I am talking about here are our "virtual datasets" as implemented, for better or worse, on clickhouse: https://dlthub.com/docs/dlt-ecosystem/destinations/clickhouse#datasets.

dlt/destinations/dataset/relation.py
```diff
@@ -52,31 +56,83 @@ def sql_client(self) -> SqlClientBase[Any]:
...
+            if is_casefolding:
+                orig_selects = {}
+                for i, proj in enumerate(parsed_query.selects):
+                    orig_selects[i] = proj.name or proj.args["alias"].name
+
+            # case fold all identifiers and quote
+            for node in parsed_query.walk():
+                if isinstance(node, sge.Table):
+                    # expand named of known tables. this is currently clickhouse things where
+                    # we use dataset.table in queries but render those as dataset___table
```
sh-rp: as mentioned further up, expanding here is really cool, but probably to late in the process for the multi dataset usecase.

dlt/destinations/dataset/relation.py
```diff
@@ -52,31 +56,83 @@ def sql_client(self) -> SqlClientBase[Any]:
...
+                    if self._sqlglot_schema.column_names(node):
+                        expanded_path = self.sql_client.make_qualified_table_name_path(
```
sh-rp: How will this work in a multi-dataset query? How to determine which client to use? I think this needs more thought.

dlt/transformations/lineage.py
sh-rp: After thinking and discussing this, maybe this will work. But I need to fix some other things to verify. I'll keep this as a reference.

dlt/transformations/lineage.py
```diff
@@ -64,6 +64,9 @@ def create_sqlglot_schema(
             # table_name = sql_client.make_qualified_table_name_path(table_name, quote=False, casefold=False)[-1]
             sqlglot_schema[table_name] = column_mapping
 
+    # NOTE: is this the correct way? We need the shortened datasetname in the schema, as this will be in the query..
+    dataset_name = schema.naming.normalize_identifier(dataset_name)
```
sh-rp: I am not sure if this is right. In any case, the actual materialized dataset name may have been shortened, so we need to adapt the sqlglot schema to use the correct one.

dlt/destinations/dataset/dataset.py
zilto: I believe [it already does this](https://github.com/dlt-hub/dlt/issues/2704). I thought it was a bug

sh-rp: Yeah, I feel like something is not quite right there..

dlt/helpers/ibis.py
rudolfix: I know that we clarified this. but user queries use dataset names and identifiers in dlt schema. they are destination agnostic

dlt/transformations/lineage.py
rudolfix: right - but not clear when you look at the code. it is important to see that star selects are expanded

dlt/transformations/transformation.py
rudolfix: yeah you are right! but I think we misconfigured something here or I do not understand what those settings mean:
```
infer_sqlglot_schema=False,
allow_anonymous_columns=False,
allow_partial=False,
```
what I expect to have here is a list of columns as found in the query that include:
- columns with data type 
- columns without data types (incomplete columns)
- anonymous columns with names generated (by sqlglot?)

so actually we do something lax.

then we should apply columns supplied by the user to see if we have something missing. but that should not happen here. it should happen in the normalize at the very end of processing.

we should review settings above. also it would be good to improve docstrings a little

dlt/transformations/transformation.py
```diff
@@ -140,25 +141,30 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
...
+        computed_columns, _ = transformation_result._compute_columns_schema(
             infer_sqlglot_schema=False,
             allow_anonymous_columns=False,
             allow_partial=False,
         )
+        select_dialect = datasets[0].sql_client.capabilities.sqlglot_dialect
+        select_query = transformation_result.normalized_query.sql(dialect=select_dialect)
+
+        # TODO: why? don't we prevent empty column schemas above?
         all_columns = {**computed_columns, **(columns or {})}
```
rudolfix, line 168: don't do it here. dlt will apply columns in extract.

dlt/transformations/transformation.py
```diff
@@ -175,9 +181,8 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
                     + "Please run with strict lineage or provide data_type hints "
                     + f"for following columns: {unknown_column_types}",
                 )
-
             yield dlt.mark.with_hints(
-                SqlModel(select_query, dialect=datasets[0].sql_client.capabilities.sqlglot_dialect),
+                SqlModel(select_query, dialect=select_dialect),
```
rudolfix, line 185: I can't attach comments to surrounding lines so I'll handle it here:
1. do not raise `UnknownColumnTypesException` here. raise it in the normalizer in the respective items normalizer. at this point you do not know if columns will not come ie. from imported yaml schema or dynamically from the user
2. `dlt.mark.with_hints(SqlModel(...)` I'd rather not use this but pass hints as property of `SqlModel` and take it out in extract. this will allow users to use `dlt.mark` if they want
3. `dlt.mark.with_hints(chunk, hints=make_hints(columns=all_columns)` -> `just re-yield chunks. we do not have any additional info that is not handled by extract already)

docs/website/docs/dlt-ecosystem/destinations/motherduck.md
rudolfix: ehhhh can we remove this commit and rewrite branch? this fix is wrong...

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -35,22 +33,27 @@ def __init__(
         self._ibis_object = ibis_object
 
     def _query(self) -> Any:
```
rudolfix, line 35: how we generate raw query (`_query`) is wrong in both relation types and generates a lot of unnecessary code.
'_query` should return parsed sqlglot query. then

1. we can drop 90% of code in ibis relation. ibis `to_sql` actually creates sqlglot query inside. just do the same here. it is 2 lines of code IMO. no transpilation needed. no ifs for mssql etc.
2. we can simplify default relation. just construct sqlglot expression. we can most probably remove `sql_client` limit clause. we do not longer need it

dlt/transformations/lineage.py
```diff
@@ -82,8 +78,9 @@ def compute_columns_schema(
     infer_sqlglot_schema: bool = True,
```
rudolfix, line 78: having `dialect` here raises a fundamental question in which dialect we expect our users to write those "destination-agnostic" queries? right now it is a destination dialect I think it is a practical default.

my take: add `query_dialect` argument to `dlt.transformation` to be able to fix ie. `duckdb` dialect even if on production we'll transpile it to snowflake query in the `load.py`.

this we should consider in separate ticket

dlt/common/destination/dataset.py
```diff
@@ -263,11 +271,12 @@ def dataset_name(self) -> str:
             str: The name of the dataset
         """
 
-    def __call__(self, query: Any) -> SupportsReadableRelation:
+    def __call__(self, query: Any, normalize_query: bool = True) -> SupportsReadableRelation:
```
rudolfix, line 274: I get why we need it. I think it makes sense but it is a power user feature. maybe `raw_query=False` is a better name? this is how such feature is often called.
1. if it is True we do not qualify the query
2. we do not compute columns schema
3. in transformations we'll require a full schema to be provided by users via columns
makes sense to me

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -35,22 +33,27 @@ def __init__(
         self._ibis_object = ibis_object
```
rudolfix, line 33: why we do not accept `normalize_query`?

dlt/destinations/dataset/relation.py
```diff
@@ -166,12 +180,28 @@ def compute_columns_schema(
...
+        if self._columns_schema is None:
+            self._columns_schema, self._qualified_query = self._compute_columns_schema()
+        return self._columns_schema
 
     @columns_schema.setter
     def columns_schema(self, new_value: TTableSchemaColumns) -> None:
         raise NotImplementedError("Columns Schema may not be set")
 
+    @property
+    def qualified_query(self) -> sge.Query:
```
rudolfix, line 192: IMO we just have `query()` in user-facing api. no one will understand the difference. should we hide it until we know that we want it?

rudolfix (commented):
I have a few ideas how we can simplify this even more

dlt/destinations/dataset/ibis_relation.py
sh-rp: We do not need an ibis relation for the raw query, none of the ibis methods are used. If you use __call__ on the dataset, you always get the normale ReadableRelation with a fixed sql string that you cannot change.

dlt/destinations/dataset/relation.py
sh-rp: yes, I did that

dlt/common/destination/dataset.py
sh-rp: I actually had called it raw_query before :) I reverted it back more or less now.

dlt/destinations/dataset/ibis_relation.py
sh-rp: 1. Done
2. This is a task for @anuunchin  :)

dlt/transformations/transformation.py
sh-rp: I removed it.

PR #2708, opened 2025-06-02
Feat/add athena dabatase location option

### LINKED ISSUE #2706: Add Database Location Option for Athena
### Feature description

Add a config option for Athena database location.

### Are you a dlt user?

None

### Use case

I've been having trouble with Athena destination creating databases and then not having access to those databases somehow. Creating the databases ahead of time with the correct location seems to solve the issue.

### Proposed solution

Add a optional config for database location to the Athena config. If that config option is specified, add a [location clause ](https://docs.aws.amazon.com/athena/latest/ug/create-database.html) to the create database command. I'll attach a PR to this shortly.

### Related issues

_No response_

eric-pinkham-rw:
### Description
Adds a option to set the database location for Athena by updating the [location clause](https://docs.aws.amazon.com/athena/latest/ug/create-database.html#parameters) if provided.

This been tested and works as expected against our Athena instance.

### Related Issues
- Resolves #2706


rudolfix (dismissed):
LGTM!

rudolfix:
@eric-pinkham-rw there's still linting problem `make format` will fix it. also it would be great if you could merge the current devel branch into it... we had a broken test in `common` that got merged. sorry for the trouble :)

PR #2710, opened 2025-06-03
docs: rest_api: clarify dlt resource and rest_api specific parameters

### LINKED ISSUE #2698: Declarative api config for resource should have table_name property
### Feature description

As of today in api config (declarative) there is `name` param for the `resources`:
```
source = rest_api_source(
...
"resources": [
                {
                    "name": f"financials_{symbol}",  # this will be used as a table name and resource name
                    "endpoint": {
...)
```
There is no possibility to configure desired table name. Normally we may it is possible, but not if we want to make it in declarative way. So, my proposition is:
```
source = rest_api_source(
...
"resources": [
                {
                    "name": f"financials_{symbol}",  # this will be used as resource name and table name if table_name param is not set
                    "table_name": "raw_table",       # table name where the data should be ingested
                    "endpoint": {
...)
```

### Are you a dlt user?

Yes, community version

### Use case

I have incremental load from the api per some filter. The data is identical and the source is the same, so it should go to the same table, but the incremental is different per `symbol`. That's why resource name should be controlled separately from the table name.

### Proposed solution

Add new param `table_name`

### Related issues

_No response_

burnash:
Resolves #2698

### DIFF
--- a/docs/website/docs/dlt-ecosystem/verified-sources/rest_api/basic.md
+++ b/docs/website/docs/dlt-ecosystem/verified-sources/rest_api/basic.md
@@ -288,6 +288,7 @@ This is a list of resource configurations that define the API endpoints to be lo
 
 A resource configuration is used to define a [dlt resource](../../../general-usage/resource.md) for the data to be loaded from an API endpoint. When defining the resource you may specify:
 - dlt resource parameters, for for example:
+    - `name`: The name of the resource. This is also used as the table name in the destination unless overridden by the `table_name` parameter.
     - `write_disposition`: The write disposition for the resource.
     - `primary_key`: The primary key for the resource.
     - `table_name`: Override the table name for this resource.


docs/website/docs/dlt-ecosystem/verified-sources/rest_api/basic.md
```diff
@@ -286,15 +286,23 @@ This is a list of resource configurations that define the API endpoints to be lo
...
-- `write_disposition`: The write disposition for the resource.
-- `primary_key`: The primary key for the resource.
-- `include_from_parent`: A list of fields from the parent resource to be included in the resource output. See the [resource relationships](#include-fields-from-the-parent-resource) section for more details.
-- `processing_steps`: A list of [processing steps](#processing-steps-filter-and-transform-data) to filter and transform your data.
-- `selected`: A flag to indicate if the resource is selected for loading. This could be useful when you want to load data only from child resources and not from the parent resource.
-- `auth`: An optional `AuthConfig` instance. If passed, is used over the one defined in the [client](#client) definition. Example:
+A resource configuration is used to define a [dlt resource](../../../general-usage/resource.md) for the data to be loaded from an API endpoint. When defining the resource you may specify:
+- dlt resource parameters, for for example:
```
AstrakhantsevaAA, line 290: can you also include "name" as a resource name, to avoid confusions in the future

burnash: updated

PR #2714, opened 2025-06-04
test destination maintenance script

sh-rp:
### Description

All of our tests clean up after they run on our destinations, but there are cases where the pytest teardown gets interrupted and orphan datasets remain. This PR adds a script that we can run on a schedule that cleans up these datasets every night. For some destinations, we know if for sure for athena and the glue catalog, there is a massive slowdown if many datasets are present.


sh-rp:
Closing since nothing is going forward here for now

PR #2715, opened 2025-06-04
qol: exception formatting

zilto:
### Description
The majority of exceptions use f-strings and interpolate variables into the exception message. However, they don't highlight when values are interpolated nor name the entities. This makes some user-facing message confusing.

Example
```python
# current exception
f"A data source {func_name} of a transformer {resource_name} is an undecorated function."
# could render as
"A data source source of a transformer transformer is an undecorated function." 

# after changes
"A data source `source` of a transformer `transformer` is an undecorated function."
```

### Changes
- Use backticks \` to enclose reference to code objects and interpolated variables
- Use f-string formating `{variable_name=:}` for clearer exception messages. This also simplifies maintenance when renaming args
  ```python
  column_mode = "freeze"
  # from
  f"{column_mode} column mode not implemented for Pydantic validation"
  # to 
  f"`{column_mode=:}` not implemented for Pydantic validation"
  # interpolates to
  "`column_mode='freeze'` not implemented for Pydantic validation"
  ``` 
- Add to `dlt.common.exceptions` the `TypeErrorWithKnownTypes` and `ValueErrorWithKnownValues` to streamline a two common error messages
  ```python
  compression = "foo"
  # from
  ValueError('The argument `compression` must have one of the following values: "auto", "enable", "disable".')
  # to
  ValueErrorWithKnownValues("compression", compression, ["auto", "enable", "disable"])
  # renders as
  'Received invalid value `compression="foo"`.  Valid values are: ["auto", "enable", "disable"]'
  ```

### DIFF
--- a/dlt/common/configuration/exceptions.py
+++ b/dlt/common/configuration/exceptions.py
@@ -147,7 +147,7 @@ class ConfigValueCannotBeCoercedException(ConfigurationValueError):
         self.field_value = field_value
         self.hint = hint
         super().__init__(
-            f"Configured value for field {field_name} cannot be coerced into type {str(hint)}"
+            f"Configured value for field `{field_name}` cannot be coerced into type `{str(hint)}`"
         )
 
 
--- a/dlt/common/configuration/exceptions.py
+++ b/dlt/common/configuration/exceptions.py
@@ -122,6 +122,9 @@ class UnmatchedConfigHintResolversException(ConfigurationException):
             f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}`"
             " but these fields are not defined in the spec.\nWhen using @resolve_type() decorator,"
             f" Add the fields with 'Any' or another common type hint, example:\n\n{example}"
+            f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}` but"
+            " these fields are not defined in the spec.\nWhen using `@resolve_type()` decorator, Add"
+            f" the fields with 'Any' or another common type hint, example:\n\n{example}"
         )
         super().__init__(msg)
 
@@ -131,8 +134,7 @@ class FinalConfigFieldException(ConfigurationException):
 
     def __init__(self, spec_name: str, field: str) -> None:
         super().__init__(
-            f"Field `{field}` in spec `{spec_name}` is final but is being changed by a config"
-            " provider"
+            f"Field `{field}` in spec `{spec_name}` is final but is being changed by a config provider"
         )
 
 
@@ -144,7 +146,7 @@ class ConfigValueCannotBeCoercedException(ConfigurationValueError):
         self.field_value = field_value
         self.hint = hint
         super().__init__(
-            f"Can't coerce value for config field `{field_name}` into type {str(hint)}"
+            f"Configured value for field {field_name} cannot be coerced into type {str(hint)}"
         )
 
 
@@ -172,7 +174,7 @@ class ConfigFieldMissingTypeHintException(ConfigurationException):
         self.field_name = field_name
         self.typ_ = spec
         super().__init__(
-            f"Field {field_name} on configspec {spec} does not provide required type hint"
+            f"Field `{field_name}` on configspec `{spec}` does not provide required type hint"
         )
 
 
@@ -183,7 +185,7 @@ class ConfigFieldTypeHintNotSupported(ConfigurationException):
         self.field_name = field_name
         self.typ_ = spec
         super().__init__(
-            f"Field {field_name} on configspec {spec} has hint with unsupported type {typ_}"
+            f"Field `{field_name}` on configspec `{spec}` has hint with unsupported type `{typ_}`"
         )
 
 
@@ -192,7 +194,7 @@ class ValueNotSecretException(ConfigurationException):
         self.provider_name = provider_name
         self.key = key
         super().__init__(
-            f"Provider {provider_name} cannot hold secret values but key {key} with secret value is"
+            f"Provider `{provider_name}` cannot hold secret values but key `{key}` with secret value is"
             " present"
         )
 
@@ -211,10 +213,9 @@ class InvalidNativeValue(ConfigurationException):
         self.inner_exception = inner_exception
         inner_msg = f" {self.inner_exception}" if inner_exception is not ValueError else ""
         super().__init__(
-            f"{spec.__name__} cannot parse the configuration value provided. The value is of type"
-            f" {native_value_type.__name__} and comes from the"
-            " {embedded_sections} section(s). Value may be a secret and is not shown. "
-            f"Details: {inner_msg}"
+            f"`{spec.__name__}` cannot parse the configuration value provided. The value is of type "
+            f"`{native_value_type.__name__}` and comes from the sections `{embedded_sections}` "
+            f"Value may be a secret and is not shown. Details: {inner_msg}"
         )
 
 
@@ -224,20 +225,20 @@ class ContainerInjectableContextMangled(ContainerException):
         self.existing_config = existing_config
         self.expected_config = expected_config
         super().__init__(
-            f"When restoring context {spec.__name__}, instance {expected_config} was expected,"
-            f" instead instance {existing_config} was found."
+            f"When restoring context `{spec.__name__}`, instance `{expected_config}` was expected,"
+            f" instead instance `{existing_config}` was found."
         )
 
 
 class ContextDefaultCannotBeCreated(ContainerException, KeyError):
     def __init__(self, spec: Type[Any]) -> None:
         self.spec = spec
-        super().__init__(f"Container cannot create the default value of context {spec.__name__}.")
+        super().__init__(f"Container cannot create the default value of context `{spec.__name__}`.")
 
 
 class DuplicateConfigProviderException(ConfigProviderException):
     def __init__(self, provider_name: str) -> None:
         super().__init__(
             provider_name,
-            f"Provider with name {provider_name} already present in ConfigProvidersContext",
+            f"Provider with name `{provider_name}` already present in `ConfigProvidersContext`",
         )
--- a/dlt/common/configuration/exceptions.py
+++ b/dlt/common/configuration/exceptions.py
@@ -33,8 +33,8 @@ class ConfigProviderException(ConfigurationException):
 class ConfigurationWrongTypeException(ConfigurationException):
     def __init__(self, _typ: type) -> None:
         super().__init__(
-            f"Invalid configuration instance type `{_typ}`. Configuration instances must derive from"
-            " BaseConfiguration and must be decorated with @configspec."
+            f"Invalid configuration instance type `{_typ}`. Configuration instances must derive"
+            " from BaseConfiguration and must be decorated with @configspec."
         )
 
 
@@ -48,12 +48,11 @@ class ConfigFieldMissingException(KeyError, ConfigurationException):
         super().__init__(spec_name)
 
     def __str__(self) -> str:
-        msg = (
-            f"Missing fields in configuration: {str(self.fields)}"
-            f" {self.spec_name}\n"
-        )
+        msg = f"Missing fields in configuration: {str(self.fields)} {self.spec_name}\n"
         for f, field_traces in self.traces.items():
-            msg += f"\tfor field `{f}` the following (config providers, keys) were tried in order:\n"
+            msg += (
+                f"\tfor field `{f}` the following (config providers, keys) were tried in order:\n"
+            )
             for tr in field_traces:
                 msg += f"\t\t({tr.provider}, {tr.key})\n"
 
@@ -69,7 +68,10 @@ class ConfigFieldMissingException(KeyError, ConfigurationException):
                 msg += f"Provider `{provider.name}` loaded values from locations:\n{locations}\n"
 
             if provider.is_empty:
-                msg += f"WARNING: provider `{provider.name}` is empty. Locations (i.e., files) are missing or empty.\n"
+                msg += (
+                    f"WARNING: provider `{provider.name}` is empty. Locations (i.e., files) are"
+                    " missing or empty.\n"
+                )
 
         # check if entry point is run with path. this is common problem so warn the user
         main_path = main_module_file_path()
@@ -94,9 +96,7 @@ class ConfigFieldMissingException(KeyError, ConfigurationException):
                         " but run your script from some other folder, secrets/configs will not be"
                         " found\n"
                     )
-        msg += (
-            "Learn more: https://dlthub.com/docs/general-usage/credentials/\n"
-        )
+        msg += "Learn more: https://dlthub.com/docs/general-usage/credentials/\n"
         return msg
 
     def attrs(self) -> Dict[str, Any]:
@@ -119,9 +119,9 @@ class UnmatchedConfigHintResolversException(ConfigurationException):
             f">>>    {name}: Any" for name in field_names
         )
         msg = (
-            f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}` but"
-            " these fields are not defined in the spec.\nWhen using @resolve_type() decorator, Add"
-            f" the fields with 'Any' or another common type hint, example:\n\n{example}"
+            f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}`"
+            " but these fields are not defined in the spec.\nWhen using @resolve_type() decorator,"
+            f" Add the fields with 'Any' or another common type hint, example:\n\n{example}"
         )
         super().__init__(msg)
 
@@ -131,7 +131,8 @@ class FinalConfigFieldException(ConfigurationException):
 
     def __init__(self, spec_name: str, field: str) -> None:
         super().__init__(
-            f"Field `{field}` in spec `{spec_name}` is final but is being changed by a config provider"
+            f"Field `{field}` in spec `{spec_name}` is final but is being changed by a config"
+            " provider"
         )
 
 
--- a/dlt/common/configuration/exceptions.py
+++ b/dlt/common/configuration/exceptions.py
@@ -121,10 +121,10 @@ class UnmatchedConfigHintResolversException(ConfigurationException):
         msg = (
             f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}`"
             " but these fields are not defined in the spec.\nWhen using @resolve_type() decorator,"
+            f" Add the fields with 'Any' or another common type hint, example:\n\n{example}The"
+            f" config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}` but"
+            " these fields are not defined in the spec.\nWhen using `@resolve_type()` decorator,"
             f" Add the fields with 'Any' or another common type hint, example:\n\n{example}"
-            f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}` but"
-            " these fields are not defined in the spec.\nWhen using `@resolve_type()` decorator, Add"
-            f" the fields with 'Any' or another common type hint, example:\n\n{example}"
         )
         super().__init__(msg)
 
@@ -134,7 +134,8 @@ class FinalConfigFieldException(ConfigurationException):
 
     def __init__(self, spec_name: str, field: str) -> None:
         super().__init__(
-            f"Field `{field}` in spec `{spec_name}` is final but is being changed by a config provider"
+            f"Field `{field}` in spec `{spec_name}` is final but is being changed by a config"
+            " provider"
         )
 
 
@@ -194,8 +195,8 @@ class ValueNotSecretException(ConfigurationException):
         self.provider_name = provider_name
         self.key = key
         super().__init__(
-            f"Provider `{provider_name}` cannot hold secret values but key `{key}` with secret value is"
-            " present"
+            f"Provider `{provider_name}` cannot hold secret values but key `{key}` with secret"
+            " value is present"
         )
 
 
@@ -213,9 +214,9 @@ class InvalidNativeValue(ConfigurationException):
         self.inner_exception = inner_exception
         inner_msg = f" {self.inner_exception}" if inner_exception is not ValueError else ""
         super().__init__(
-            f"`{spec.__name__}` cannot parse the configuration value provided. The value is of type "
-            f"`{native_value_type.__name__}` and comes from the sections `{embedded_sections}` "
-            f"Value may be a secret and is not shown. Details: {inner_msg}"
+            f"`{spec.__name__}` cannot parse the configuration value provided. The value is of type"
+            f" `{native_value_type.__name__}` and comes from the sections `{embedded_sections}`"
+            f" Value may be a secret and is not shown. Details: {inner_msg}"
         )
 
 
--- a/dlt/common/configuration/exceptions.py
+++ b/dlt/common/configuration/exceptions.py
@@ -23,8 +23,6 @@ class ConfigurationValueError(ConfigurationException, ValueError):
 class ContainerException(DltException):
     """base exception for all exceptions related to injectable container"""
 
-    pass
-
 
 class ConfigProviderException(ConfigurationException):
     def __init__(self, provider_name: str, *args: Any) -> None:
@@ -35,7 +33,7 @@ class ConfigProviderException(ConfigurationException):
 class ConfigurationWrongTypeException(ConfigurationException):
     def __init__(self, _typ: type) -> None:
         super().__init__(
-            f"Invalid configuration instance type {_typ}. Configuration instances must derive from"
+            f"Invalid configuration instance type `{_typ}`. Configuration instances must derive from"
             " BaseConfiguration and must be decorated with @configspec."
         )
 
@@ -51,13 +49,13 @@ class ConfigFieldMissingException(KeyError, ConfigurationException):
 
     def __str__(self) -> str:
         msg = (
-            f"Following fields are missing: {str(self.fields)} in configuration with spec"
+            f"Missing fields in configuration: {str(self.fields)}"
             f" {self.spec_name}\n"
         )
         for f, field_traces in self.traces.items():
-            msg += f'\tfor field "{f}" config providers and keys were tried in following order:\n'
+            msg += f"\tfor field `{f}` the following (config providers, keys) were tried in order:\n"
             for tr in field_traces:
-                msg += f"\t\tIn {tr.provider} key {tr.key} was not found.\n"
+                msg += f"\t\t({tr.provider}, {tr.key})\n"
 
         from dlt.common.configuration.container import Container
         from dlt.common.configuration.specs import PluggableRunContext
@@ -68,15 +66,10 @@ class ConfigFieldMissingException(KeyError, ConfigurationException):
         for provider in providers.providers:
             if provider.locations:
                 locations = "\n".join([f"\t- {os.path.abspath(loc)}" for loc in provider.locations])
-                msg += (
-                    f"Provider {provider.name} used following locations to load"
-                    f" values:\n{locations}\n"
-                )
+                msg += f"Provider `{provider.name}` loaded values from locations:\n{locations}\n"
+
             if provider.is_empty:
-                msg += (
-                    f"WARNING: provider {provider.name} is empty. Locations (ie. files) may not"
-                    " exist or may be empty.\n"
-                )
+                msg += f"WARNING: provider `{provider.name}` is empty. Locations (i.e., files) are missing or empty.\n"
 
         # check if entry point is run with path. this is common problem so warn the user
         main_path = main_module_file_path()
@@ -102,8 +95,7 @@ class ConfigFieldMissingException(KeyError, ConfigurationException):
                         " found\n"
                     )
         msg += (
-            "Please refer to https://dlthub.com/docs/general-usage/credentials/ for more"
-            " information\n"
+            "Learn more: https://dlthub.com/docs/general-usage/credentials/\n"
         )
         return msg
 
@@ -127,7 +119,7 @@ class UnmatchedConfigHintResolversException(ConfigurationException):
             f">>>    {name}: Any" for name in field_names
         )
         msg = (
-            f"The config spec {spec_name} has dynamic type resolvers for fields: {field_names} but"
+            f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}` but"
             " these fields are not defined in the spec.\nWhen using @resolve_type() decorator, Add"
             f" the fields with 'Any' or another common type hint, example:\n\n{example}"
         )
@@ -139,7 +131,7 @@ class FinalConfigFieldException(ConfigurationException):
 
     def __init__(self, spec_name: str, field: str) -> None:
         super().__init__(
-            f"Field {field} in spec {spec_name} is final but is being changed by a config provider"
+            f"Field `{field}` in spec `{spec_name}` is final but is being changed by a config provider"
         )
 
 
@@ -151,7 +143,7 @@ class ConfigValueCannotBeCoercedException(ConfigurationValueError):
         self.field_value = field_value
         self.hint = hint
         super().__init__(
-            "Configured value for field %s cannot be coerced into type %s" % (field_name, str(hint))
+            f"Can't coerce value for config field `{field_name}` into type {str(hint)}"
         )
 
 
@@ -169,7 +161,7 @@ class ConfigFileNotFoundException(ConfigurationException):
     """thrown when configuration file cannot be found in config folder"""
 
     def __init__(self, path: str) -> None:
-        super().__init__(f"Missing config file in {path}")
+        super().__init__(f"Missing config file in `{path}`")
 
 
 class ConfigFieldMissingTypeHintException(ConfigurationException):
--- a/dlt/common/configuration/exceptions.py
+++ b/dlt/common/configuration/exceptions.py
@@ -121,9 +121,6 @@ class UnmatchedConfigHintResolversException(ConfigurationException):
         msg = (
             f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}`"
             " but these fields are not defined in the spec.\nWhen using @resolve_type() decorator,"
-            f" Add the fields with 'Any' or another common type hint, example:\n\n{example}The"
-            f" config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}` but"
-            " these fields are not defined in the spec.\nWhen using `@resolve_type()` decorator,"
             f" Add the fields with 'Any' or another common type hint, example:\n\n{example}"
         )
         super().__init__(msg)
--- a/dlt/common/data_writers/exceptions.py
+++ b/dlt/common/data_writers/exceptions.py
@@ -37,7 +37,8 @@ class DestinationCapabilitiesRequired(DataWriterException, ValueError):
     def __init__(self, file_format: TLoaderFileFormat):
         self.file_format = file_format
         super().__init__(
-            f"Writer for `{file_format=:}` requires destination capabilities which were not provided."
+            f"Writer for `{file_format=:}` requires destination capabilities which were not"
+            " provided."
         )
 
 
@@ -59,7 +60,8 @@ class FileSpecNotFound(KeyError, DataWriterNotFound):
         self.file_format = file_format
         self.data_item_format = data_item_format
         super().__init__(
-            f"Can't find a file writer for spec with `{file_format=:}` and `{data_item_format=:}` where the full spec is `{spec}`"
+            f"Can't find a file writer for spec with `{file_format=:}` and `{data_item_format=:}`"
+            f" where the full spec is `{spec}`"
         )
 
 
@@ -84,5 +86,6 @@ class InvalidDataItem(DataWriterException):
         self.file_format = file_format
         self.data_item_format = data_item_format
         super().__init__(
-            f"A data item of type {data_item_format=:} cannot be written as `{file_format}: {details}`"
+            f"A data item of type {data_item_format=:} cannot be written as `{file_format}:"
+            f" {details}`"
         )
--- a/dlt/common/data_writers/exceptions.py
+++ b/dlt/common/data_writers/exceptions.py
@@ -12,7 +12,7 @@ class InvalidFileNameTemplateException(DataWriterException, ValueError):
     def __init__(self, file_name_template: str):
         self.file_name_template = file_name_template
         super().__init__(
-            f"Wrong file name template {file_name_template}. File name template must contain"
+            f"Wrong file name template `{file_name_template}`. File name template must contain"
             " exactly one %s formatter"
         )
 
@@ -20,7 +20,7 @@ class InvalidFileNameTemplateException(DataWriterException, ValueError):
 class BufferedDataWriterClosed(DataWriterException):
     def __init__(self, file_name: str):
         self.file_name = file_name
-        super().__init__(f"Writer with recent file name {file_name} is already closed")
+        super().__init__(f"Writer with recent file name `{file_name}` is already closed")
 
 
 class FileImportNotFound(DataWriterException, FileNotFoundError):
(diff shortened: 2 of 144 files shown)


rudolfix:
I briefly looked at the failed tests and except:
`SELECT  o,t,h,e,r,_,d,e,c,i,m,a,l FROM items` which fails the lineage test I do not see anything suspicious.

`FAILED tests/destinations/test_utils.py::test_get_resource_for_adapter - Failed: DID NOT RAISE <class 'ValueError'>`
this is pretty weird, You added new `ValueError` but derived correctly so all tests should still work...

zilto:
> FAILED tests/destinations/test_utils.py::test_get_resource_for_adapter - Failed: DID NOT RAISE <class 'ValueError'>
this is pretty weird, You added new ValueError but derived correctly so all tests should still work...

That's a fun one. it seems that the code doesn't `raise` any exception (took me several minute reading the code)! 

```python
    if isinstance(data, DltSource):
        if len(data.selected_resources.keys()) == 1:
            return list(data.selected_resources.values())[0]
        else:
            # no `raise` keyword
            ValueError(
                "You are trying to use an adapter on a `DltSource` with multiple resources. You can"
                " only use adapters on: pure data, a `DltResouce` or a `DltSource` with a single"
                " `DltResource`."
            )
```
Good reminder for writing unit tests: `red -> green -> refactor`. This test never failed!

dlt/common/configuration/exceptions.py
```diff
@@ -23,8 +23,6 @@ class ConfigurationValueError(ConfigurationException, ValueError):
 class ContainerException(DltException):
     """base exception for all exceptions related to injectable container"""
```
djudjuu, line 25: learning question: why not pass?

dlt/common/configuration/exceptions.py
```diff
@@ -50,14 +48,13 @@ def __init__(self, spec_name: str, traces: Mapping[str, Sequence[LookupTrace]])
         super().__init__(spec_name)
 
     def __str__(self) -> str:
-        msg = (
-            f"Following fields are missing: {str(self.fields)} in configuration with spec"
-            f" {self.spec_name}\n"
-        )
+        msg = f"Missing fields in configuration: {str(self.fields)} {self.spec_name}\n"
```
djudjuu, line 51: nitpick: why not "...for spec`{self.spec}`?

dlt/common/configuration/exceptions.py
```diff
@@ -127,9 +119,12 @@ def __init__(self, spec_name: str, field_names: Sequence[str]) -> None:
...
         msg = (
-            f"The config spec {spec_name} has dynamic type resolvers for fields: {field_names} but"
-            " these fields are not defined in the spec.\nWhen using @resolve_type() decorator, Add"
-            f" the fields with 'Any' or another common type hint, example:\n\n{example}"
+            f"The config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}`"
+            " but these fields are not defined in the spec.\nWhen using @resolve_type() decorator,"
+            f" Add the fields with 'Any' or another common type hint, example:\n\n{example}The"
+            f" config spec `{spec_name}` has dynamic type resolvers for fields: `{field_names}` but"
+            " these fields are not defined in the spec.\nWhen using `@resolve_type()` decorator,"
+            f" Add the fields with 'Any' or another common type hint, example:\n\n{example}"
```
djudjuu, line 124: issue: this looks like an accidental copy

tests/sources/rest_api/configurations/test_resolve_config.py
```diff
@@ -429,7 +429,7 @@ def test_one_resource_cannot_bind_two_parents(
         rest_api_resources(config)
 
     error_msg = str(exc_info.value)
-    assert "Multiple parent resources for user_details:" in error_msg
+    assert "Multiple parent resources for `user_details`" in error_msg
```
djudjuu: issue: not quite the error message, should be 
"...resources for resource `user_details`..."

dlt/common/data_writers/exceptions.py
```diff
@@ -50,8 +51,7 @@ def __init__(self, file_format: TLoaderFileFormat, data_item_format: str):
         self.file_format = file_format
         self.data_item_format = data_item_format
         super().__init__(
-            f"Can't find a file writer for file format {file_format} and item format"
-            f" {data_item_format}"
+            f"Can't find a file writer for `{file_format=:}` and item format `{data_item_format=:}`"
```
djudjuu, line 54: praise: another cool thing I learned from reviewing your PRs, the {var:=} extension.

dlt/common/data_writers/writers.py
```diff
@@ -153,12 +154,12 @@ class ImportFileWriter(DataWriter):
 
     def write_header(self, columns_schema: TTableSchemaColumns) -> None:
         raise NotImplementedError(
-            "ImportFileWriter cannot write any files. You have bug in your code."
+            "`ImportFileWriter` cannot write any files. You have bug in your code."
```
djudjuu: I can hear this in Marcin's voice :)
nitpick: should be "a bug"...although its funny and i'd like it

dlt/common/libs/pyarrow.py
```diff
@@ -812,8 +812,7 @@ def transpose_rows_to_columns(
         logger.info(
             "Pandas not installed, reverting to numpy.asarray to create a table which is slower"
         )
-        pivoted_rows = np.asarray(rows, dtype="object", order="k").T  # type: ignore[call-overload]
-
+        pivoted_rows = np.asarray(rows, dtype="object", order="K").T
     return {
```
djudjuu, line 839: praise: nice fix! apparently 'k' didn't do anything, but maybe it then defaulted to K...
was there a no test for this? maybe we should consider adding it, if it's easy

dlt/common/libs/pydantic.py
```diff
@@ -221,11 +221,11 @@ def apply_schema_contract_to_model(
     """
     if data_mode == "evolve":
         # create a lenient model that accepts any data
-        model = create_model(model.__name__ + "Any", **{n: (Any, None) for n in model.__fields__})  # type: ignore[call-overload, attr-defined]
+        model = create_model(model.__name__ + "Any", **{n: (Any, None) for n in model.model_fields})  # type: ignore
```
djudjuu, line 224: learning question: why not `__pydantic_fields__`

 __pydantic_fields__: ClassVar[Dict[str, FieldInfo]]  # noqa: UP006
    """A dictionary of field names and their corresponding [`FieldInfo`][pydantic.fields.FieldInfo] objects.
    This replaces `Model.__fields__` from Pydantic V1.

dlt/common/runtime/run_context.py
```diff
@@ -103,13 +103,13 @@ def import_run_dir_module(run_dir: str) -> ModuleType:
         run_dir = os.path.abspath(run_dir)
         base_dir = os.path.basename(run_dir)
         if not base_dir:
-            raise ImportError(f"run dir {run_dir} looks like filesystem root")
```
djudjuu, line 106: nitpick: could be `{rundir:=}`

dlt/common/schema/schema.py
```diff
@@ -358,8 +358,8 @@ def apply_schema_contract(
                         existing_table,
                         schema_contract,
                         data_item,
-                        f"Trying to create new variant column {column_name} to table"
-                        f" {table_name} but data_types are frozen.",
+                        f"Can't add create variant column `{column_name} for table `{table_name}`"
```
djudjuu: issue: add create -> create

dlt/common/schema/schema.py
```diff
@@ -1143,7 +1143,7 @@ def _replace_and_apply_naming(
             if not table_name.startswith(self._dlt_tables_prefix):
                 raise SchemaCorruptedException(
                     self.name,
-                    f"A naming convention {self.naming.name()} mangles _dlt table prefix to"
+                    f"A naming convention `{self.naming.name()}` mangles `_dlt` table prefix to"
```
djudjuu, line 1168: issue: A table {table_name} is missing the backticks

dlt/common/schema/utils.py
```diff
@@ -440,7 +441,7 @@ def merge_columns(
     * incomplete columns in `columns_a` that got completed in `columns_b` are removed to preserve order
     """
     if columns_partial is False:
-        raise NotImplementedError("columns_partial must be False for merge_columns")
+        raise NotImplementedError("`columns_partial` must be `False` for `merge_columns`")
```
djudjuu: nitpick: I don't think the error message is helpful here, as columns_partial being false is exactly what raises the error here, isn't it

djudjuu: not a problem this pr created , but it could fix it

dlt/common/storages/exceptions.py
```diff
@@ -68,31 +69,30 @@ def __init__(self, load_id: str, supported_formats: Iterable[str], wrong_job: st
...
+            f"Package with `{load_id=:}` is already completed, but another complete was requested"
         )
 
 
 class LoadPackageNotCompleted(LoadStorageException):
     def __init__(self, load_id: str) -> None:
         self.load_id = load_id
         super().__init__(
-            f"Package with load id {load_id} is not yet completed, but method required that"
+            f"Package with `{load_id=:}` is not yet completed, but method required that"
```
djudjuu: nitpick: `but method required that` reads like its missing what method. maybe this could be `but the calling method required it to be`

dlt/common/storages/normalize_storage.py
```diff
@@ -65,7 +65,7 @@ def migrate_storage(self, from_version: Version, to_version: Version) -> None:
                     self.storage.storage_path,
                     from_version,
                     to_version,
-                    f"There are extract files in {NormalizeStorage.EXTRACTED_FOLDER} folder."
+                    f"There are extract files in folder `{NormalizeStorage.EXTRACTED_FOLDER}`."
                     " Storage will not migrate automatically duo to possible data loss. Delete the"
```
djudjuu: type: duo -> due

dlt/common/storages/transactional_file.py
```diff
@@ -114,8 +114,8 @@ def _sync_locks(self) -> t.List[str]:
             output.append(name)
         if not output:
             raise RuntimeError(
-                f"When syncing locks for path {self.path} and lock {self.lock_path} no lock file"
-                " was found"
+                f"Lock syncing failed. No lock file found for path `{self.path}` and lock"
```
djudjuu, line 117: praise: I like this format much better. is there a wor for it? 
in emails or conversations, I call it `frontloading`

dlt/common/validation.py
```diff
@@ -99,7 +99,7 @@ def verify_prop(pk: str, pv: Any, t: Any) -> None:
                 if len(failed_validations) == len(union_types):
                     type_names = [get_type_name(ut) for ut in union_types]
                     msg = (
-                        f"field '{pk}' expects the following types: {', '.join(type_names)}."
+                        f"field '{pk}' expects the following types: {type_names}."
                         f" Provided value {pv} with type '{type(pv).__name__}' is invalid with the"
```
djudjuu: ' -> `

dlt/common/validation.py
```diff
@@ -121,12 +121,12 @@ def verify_prop(pk: str, pv: Any, t: Any) -> None:
...
                 raise DictValidationException(
-                    f"field '{pk}' with value {pv} is not one of: {a_l}", path, t, pk, pv
+                    f"field `{pk}={pv}` is not one of: {a_l}", path, t, pk, pv
                 )
         elif t in [int, bool, str, float]:
             if not isinstance(pv, t):
                 raise DictValidationException(
-                    f"field '{pk}' with value {pv} has invalid type '{type(pv).__name__}' while"
+                    f"field `{pk}={pv}` has invalid type '{type(pv).__name__}' while"
                     f" '{t.__name__}' is expected",
```
djudjuu: ' -> `

djudjuu: and two below too

dlt/common/validation.py
```diff
@@ -136,7 +136,7 @@ def verify_prop(pk: str, pv: Any, t: Any) -> None:
         elif is_typeddict(t):
             if not isinstance(pv, dict):
                 raise DictValidationException(
-                    f"field '{pk}' with value {pv} has invalid type '{type(pv).__name__}' while"
+                    f"field '{pk}' with value `{pv}` has invalid type '{type(pv).__name__}' while"
                     f" '{get_type_name(t)}' is expected",
```
djudjuu:
```suggestion
                    f" `{get_type_name(t)}` is expected",
```

dlt/common/validation.py
```diff
@@ -173,7 +173,7 @@ def verify_prop(pk: str, pv: Any, t: Any) -> None:
             for d_k, d_v in pv.items():
                 if not isinstance(d_k, str):
                     raise DictValidationException(
-                        f"field '{pk}' with key {d_k} must be a string", path, t, pk, d_k
+                        f"field '{pk}' with key `{d_k}` must be a string", path, t, pk, d_k
```
djudjuu:
```suggestion
                        f"field `{pk}` with key `{d_k}` must be a string", path, t, pk, d_k
```

dlt/common/validation.py
```diff
@@ -161,7 +161,7 @@ def verify_prop(pk: str, pv: Any, t: Any) -> None:
         elif is_dict_generic_type(t):
             if not isinstance(pv, dict):
                 raise DictValidationException(
-                    f"field '{pk}' with value {pv} has invalid type '{type(pv).__name__}' while"
+                    f"field '{pk}' with value `{pv}` has invalid type '{type(pv).__name__}' while"
```
djudjuu:
```suggestion
                    f"field `{pk}` with value `{pv}` has invalid type `{type(pv).__name__}` while"
```

dlt/common/validation.py
```diff
@@ -190,7 +190,7 @@ def verify_prop(pk: str, pv: Any, t: Any) -> None:
             else:
                 raise DictValidationException(
                     f"field '{pk}' expects callable (function or class instance) but got "
```
djudjuu:
```suggestion
                    f"field `{pk}` expects callable (function or class instance) but got "
```

dlt/common/validation.py
```diff
@@ -190,7 +190,7 @@ def verify_prop(pk: str, pv: Any, t: Any) -> None:
...
-                    f" '{pv}'. Mind that signatures are not validated",
+                    f"'{pv}'. Mind that signatures are not validated",
```
djudjuu:
```suggestion
                    f"`{pv}`. Mind that signatures are not validated",
```

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -97,8 +99,8 @@ def __getattr__(self, name: str) -> Any:
     def __getitem__(self, columns: Union[str, Sequence[str]]) -> "ReadableIbisRelation":
         if not isinstance(columns, (str, Sequence)):
             raise TypeError(
-                f"Invalid argument type: {type(columns).__name__}, requires a sequence of column"
-                " names Sequence[str] or a single column name str"
+                "Expected a sequence of column names `Sequence[str]` or a single column `str`. "
```
djudjuu, line 100: nit: I guess this could be `TypeErrorWithKnownTypes` too

dlt/destinations/dataset/relation.py
```diff
@@ -280,17 +282,10 @@ def select(self, *columns: str) -> Self:
...
-        elif isinstance(columns, Sequence):
-            return self.select(*columns)
-        raise TypeError(
-            f"Invalid argument type: {type(columns).__name__}, requires a sequence of column names"
-            " Sequence[str]"
-        )
+        if not isinstance(columns, Sequence):
+            raise TypeErrorWithKnownTypes("columns", columns, ["Sequence[str]"])
+
+        return self.select(*columns)
```
djudjuu: issue: check is instance of Sequence, but error message reports `Sequence[str]`

dlt/destinations/impl/athena/factory.py
```diff
@@ -112,7 +112,7 @@ def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema =
         elif precision <= 64:
             return "bigint"
         raise TerminalValueError(
-            f"bigint with {precision} bits precision cannot be mapped into athena integer type"
+            f"bigint with `precision={precision}` can't be mapped to athena integer type"
```
djudjuu:
```suggestion
            f"bigint with `{precision:=}` can't be mapped to athena integer type"
```

dlt/destinations/impl/clickhouse/factory.py
```diff
@@ -110,7 +110,7 @@ def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema =
         if precision <= 64:
             return "Int64"
         raise TerminalValueError(
-            f"bigint with {precision} bits precision cannot be mapped into ClickHouse integer type"
+            f"bigint with {precision=:} can't be mapped to ClickHouse integer type"
```
djudjuu:
```suggestion
            f"bigint with `{precision=:}` can't be mapped to ClickHouse integer type"
```

dlt/destinations/impl/databricks/factory.py
```diff
@@ -80,7 +80,7 @@ def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema =
         if precision <= 64:
             return "BIGINT"
         raise TerminalValueError(
-            f"bigint with {precision} bits precision cannot be mapped into databricks integer type"
+            f"bigint with {precision=:} can't be mapped to Databricks integer type"
```
djudjuu:
```suggestion
            f"bigint with `{precision=:}` can't be mapped to Databricks integer type"
```

dlt/destinations/impl/destination/configuration.py
```diff
@@ -35,6 +35,6 @@ def ensure_callable(self) -> None:
             or self.destination_callable is dummy_custom_destination
         ):
             raise ConfigurationValueError(
-                f"A valid callable was not provided to {self.__class__.__name__}. Did you decorate"
-                " a function @dlt.destination correctly?"
+                f"A valid callable was not provided to `{self.__class__.__name__}`. Did you"
+                " decorate a function @dlt.destination correctly?"
```
djudjuu: better:
did you correctly decorate a function with `@dlt.destination`?

dlt/destinations/impl/duckdb/factory.py
```diff
@@ -76,7 +76,7 @@ def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema =
         elif precision <= 128:
             return "HUGEINT"
         raise TerminalValueError(
-            f"bigint with {precision} bits precision cannot be mapped into duckdb integer type"
+            f"bigint with {precision=:} can't be mapped to DuckDB integer type"
```
djudjuu:
```suggestion
            f"bigint with `{precision=:}` can't be mapped to DuckDB integer type"
```

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -872,7 +872,7 @@ def load_open_table(self, table_format: TTableFormat, table_name: str, **kwargs:
             return DeltaTable(table_location, storage_options=storage_options)
         else:
             raise NotImplementedError(
-                f"Cannot load tables in {table_format} format in filesystem destination."
+                f"Can't load tables using {table_format=:} using `filesystem` destination."
```
djudjuu:
```suggestion
                f"Can't load tables using `{table_format=:}` with `filesystem` destination."
```
or other way. but 2x using sounds odd. (also backticks)

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -32,10 +33,10 @@ def __init__(
         persist_secrets: bool = False,
     ) -> None:
         if remote_client.config.protocol not in SUPPORTED_PROTOCOLS:
-            raise NotImplementedError(
-                f"Protocol {remote_client.config.protocol} currently not supported for"
-                f" FilesystemSqlClient. Supported protocols are {SUPPORTED_PROTOCOLS}."
+            raise ValueErrorWithKnownValues(
```
djudjuu: question: going from notimplemented to valueError. on purpose? if some tests fail unexpectcly, maybe you need to revert

dlt/destinations/impl/lancedb/lancedb_client.py
```diff
@@ -216,7 +216,7 @@ def write_records(
                 ).when_matched_update_all().when_not_matched_insert_all().execute(records)
         else:
             raise DestinationTerminalException(
-                f"Unsupported write disposition {write_disposition} for LanceDB Destination - batch"
+                f"Unsupported {write_disposition=:} for LanceDB Destination - batch"
```
djudjuu:
```suggestion
                f"Unsupported `{write_disposition=:}` for LanceDB Destination - batch"
```

dlt/destinations/impl/mssql/factory.py
```diff
@@ -65,9 +65,7 @@ def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema =
             return "int"
         elif precision <= 64:
             return "bigint"
-        raise TerminalValueError(
-            f"bigint with {precision} bits precision cannot be mapped into mssql integer type"
-        )
+        raise TerminalValueError(f"bigint with {precision=:} can't be mapped to MSSQL integer type")
```
djudjuu:
```suggestion
        raise TerminalValueError(f"bigint with `{precision=:}` can't be mapped to MSSQL integer type")
```

dlt/destinations/impl/postgres/factory.py
```diff
@@ -70,7 +70,7 @@ def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema =
         elif precision <= 64:
             return "bigint"
         raise TerminalValueError(
-            f"bigint with {precision} bits precision cannot be mapped into postgres integer type"
+            f"bigint with {precision=:} can't be mapped to PostgreSQL integer type"
```
djudjuu:
```suggestion
            f"bigint with `{precision=:}` can't be mapped to PostgreSQL integer type"
```

dlt/destinations/impl/postgres/factory.py
```diff
@@ -94,8 +94,8 @@ def to_db_datetime_type(
                 timestamp += f" ({precision})"
             else:
                 raise TerminalValueError(
-                    f"Postgres does not support precision '{precision}' for '{column_name}' in"
-                    f" table '{table_name}'"
+                    f"Postgres doesn't support {precision=:} for timestamp column `{column_name}`"
```
djudjuu, line 99:
```suggestion
                    f"Postgres doesn't support `{precision=:}` for timestamp column `{column_name}`"
```

dlt/destinations/impl/redshift/factory.py
```diff
@@ -97,7 +97,7 @@ def to_db_integer_type(self, column: TColumnSchema, table: PreparedTableSchema =
         elif precision <= 64:
             return "bigint"
         raise TerminalValueError(
-            f"bigint with {precision} bits precision cannot be mapped into postgres integer type"
+            f"bigint with {precision=:} can't be mapped to Redshift integer type"
```
djudjuu:
```suggestion
            f"bigint with `{precision=:}` can't be mapped to Redshift integer type"
```

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -53,14 +53,13 @@ def on_resolved(self) -> None:
...
+                    "Make sure that `private_key` in dlt recognized format is at"
+                    f" `{self.private_key_path}`. Note that binary formats are not supported"
                 )
         if not self.password and not self.private_key and not self.authenticator:
             raise ConfigurationValueError(
-                "Please specify password or private_key or authenticator fields."
-                " SnowflakeCredentials supports password, private key and authenticator based (ie."
-                " oauth2) authentication and one of those must be specified."
+                "`SnowflakeCredentials` requires one of the following to be specified: `password`,"
+                " `private_key`, `authenticator`."
```
djudjuu: question: not mentioning the oauth2 example? i get that its more uniform that way, but also less informative. as a noob I lean for the former version.

dlt/destinations/impl/snowflake/utils.py
```diff
@@ -180,7 +181,7 @@ def gen_copy_sql(
             f"ENCODING = '{csv_config.encoding}'",
         ]
     else:
-        raise ValueError(f"{loader_file_format} not supported for Snowflake COPY command.")
+        raise ValueError(f"{loader_file_format=:} not supported for Snowflake COPY command.")
```
djudjuu:
```suggestion
        raise ValueError(f"`{loader_file_format=:}` not supported for Snowflake COPY command.")
```

dlt/destinations/impl/sqlalchemy/type_mapper.py
```diff
@@ -31,7 +31,7 @@ def _db_integer_type(self, precision: Optional[int]) -> sa.types.TypeEngine:
             return sa.Integer()
         elif precision <= 64:
             return sa.BigInteger()
-        raise TerminalValueError(f"Unsupported precision for integer type: {precision}")
+        raise TerminalValueError(f"Unsupported {precision=:} for integer type.")
```
djudjuu:
```suggestion
        raise TerminalValueError(f"Unsupported `{precision=:}` for integer type.")
```

dlt/destinations/exceptions.py
```diff
@@ -38,7 +38,7 @@ def __init__(
         self.dataset_name = dataset_name
         self.inner_exc = inner_exc
         super().__init__(
-            f"Connection with {client_type} to dataset name {dataset_name} failed. Please check if"
+            f"Connection with `{client_type}` to `{dataset_name=:}` failed. Please check if"
```
djudjuu:
```suggestion
            f"Connection with `{client_type:=}` to `{dataset_name=:}` failed. Please check if"
```

dlt/destinations/exceptions.py
```diff
@@ -50,8 +50,8 @@ def __init__(self, client_type: str, dataset_name: str) -> None:
         self.client_type = client_type
         self.dataset_name = dataset_name
         super().__init__(
-            f"Connection with {client_type} to dataset {dataset_name} is closed. Open the"
-            " connection with 'client.open_connection' or with the 'with client:' statement"
+            f"Connection with `{client_type}` to `{dataset_name=:}` is closed. Open the"
```
djudjuu:
```suggestion
            f"Connection with `{client_type:=}` to `{dataset_name=:}` is closed. Open the"
```

dlt/extract/exceptions.py
```diff
@@ -60,9 +60,9 @@ def __init__(self, pipe_name: str, func_name: str, sig: Signature, call_error: s
         self.sig = sig
         super().__init__(
             pipe_name,
-            f"Unable to call {func_name}: {call_error}. The mapping/filtering function"
-            f" {func_name} requires first argument to take data item and optional second argument"
-            f" named 'meta', but the signature is {sig}",
+            f"Unable to call `{func_name}`: {call_error}. The mapping/filtering function"
+            f" `{func_name}` requires first argument to take data item and optional second argument"
+            f" named 'meta', but the signature is `{sig}`",
```
djudjuu:
```suggestion
            f" named `meta`, but the signature is `{sig}`",
```

dlt/extract/exceptions.py
```diff
@@ -273,9 +275,9 @@ def __init__(
         self.func_name = func_name
         self.sig = sig
         msg = (
-            f"The {kind} {resource_name} is parametrized and expects following arguments: {sig}."
-            f" Did you forget to bind the {func_name} function? For example from"
-            f" `source.{resource_name}.bind(...)"
+            f"The `{kind}` `{resource_name}` is parametrized and expects following arguments:"
+            f" {sig}. Did you forget to bind the {func_name} function? For example from"
```
djudjuu:
```suggestion
            f" `{sig}`. Did you forget to bind the `{func_name}` function? For example from"
```

dlt/extract/exceptions.py
```diff
@@ -297,16 +299,18 @@ class DataItemRequiredForDynamicTableHints(DltResourceException):
     def __init__(self, resource_name: str) -> None:
         super().__init__(
             resource_name,
-            f"""An instance of resource's data required to generate table schema in resource {resource_name}.
-        One of table hints for that resource (typically table name) is a function and hint is computed separately for each instance of data extracted from that resource.""",
+            "An instance of resource's data required to generate table schema in resource"
```
djudjuu: bonus section: it's late so I have trouble improving this, but I know its formulated a bit confusingly

dlt/extract/pipe.py
```diff
@@ -206,7 +206,8 @@ def remove_step(self, index: int) -> None:
         if index == self._gen_idx:
             raise CreatePipeException(
                 self.name,
-                f"Step at index {index} holds a data generator for this pipe and cannot be removed",
+                f"Step at index `{index}` holds a data generator for this pipe and cannot be"
```
djudjuu:
```suggestion
                f"Step at `{index:=}` holds a data generator for this pipe and cannot be"
```

dlt/extract/pipe.py
```diff
@@ -358,7 +359,7 @@ def _verify_head_step(self, step: TPipeStep) -> None:
         if not isinstance(step, (Iterable, Iterator, AsyncIterator)) and not callable(step):
             raise CreatePipeException(
                 self.name,
-                "A head of a resource pipe must be Iterable, Iterator, AsyncIterator or a Callable",
+                "A head of a resource pipe must be: [Iterable, Iterator, AsyncIterator, Callable]",
```
djudjuu, line 361: praise: I agree with [...] bein more user-friendly than ",".join([...]), because I like bracketing for the implicit context

dlt/helpers/dbt_cloud/client.py
```diff
@@ -101,8 +101,8 @@ def trigger_job_run(
         """
         if not (self.account_id and job_id):
             raise InvalidCredentialsException(
-                f"account_id and job_id are required, got account_id: {self.account_id} and job_id:"
-                f" {job_id}"
+                "`account_id` and `job_id` are required. "
+                f"Got `account_id={self.account_id} and `job_id={job_id}`"
```
djudjuu:
```suggestion
                f"Got `account_id={self.account_id} and `{job_id:=}`"
```

dlt/helpers/dbt_cloud/client.py
```diff
@@ -133,8 +133,8 @@ def get_run_status(self, run_id: Union[int, str]) -> Dict[Any, Any]:
         """
         if not (self.account_id and run_id):
             raise InvalidCredentialsException(
-                f"account_id and run_id are required, got account_id: {self.account_id} and run_id:"
-                f" {run_id}."
+                "`account_id` and `run_id` are required. "
+                f"Got `account_id={self.account_id} and `run_id={run_id}`"
```
djudjuu:
```suggestion
                f"Got `account_id={self.account_id} and `{run_id:=}`"
```

dlt/load/exceptions.py
```diff
@@ -45,8 +45,8 @@ def __init__(
         self.supported_types = supported_file_format
         self.file_path = file_path
         super().__init__(
-            f"Loader does not support writer {file_format} in  file {file_path}. Supported writers:"
-            f" {supported_file_format}"
+            f"Loader does not support writer for `{file_format=:}` in  file `{file_path}`."
+            f" Supported writers: {supported_file_format}"
```
djudjuu:
```suggestion
            f" Supported writers: `{supported_file_format}`"
```

dlt/load/exceptions.py
```diff
@@ -56,21 +56,21 @@ def __init__(self, table_name: str, write_disposition: str, file_name: str) -> N
         self.write_disposition = write_disposition
         self.file_name = file_name
         super().__init__(
-            f"Loader does not support {write_disposition} in table {table_name} when loading file"
-            f" {file_name}"
+            f"Loader does not support` {write_disposition=:}` in table `{table_name}` when loading"
```
djudjuu:
```suggestion
            f"Loader does not support `{write_disposition=:}` in table `{table_name}` when loading"
```

dlt/pipeline/configuration.py
```diff
@@ -60,4 +60,4 @@ def on_resolved(self) -> None:
 def ensure_correct_pipeline_kwargs(f: AnyFun, **kwargs: Any) -> None:
     for arg_name in kwargs:
         if not hasattr(PipelineConfiguration, arg_name) and not arg_name.startswith("_dlt"):
-            raise TypeError(f"{f.__name__} got an unexpected keyword argument '{arg_name}'")
+            raise TypeError(f"`{f.__name__}` got an unexpected keyword argument '{arg_name}'")
```
djudjuu:
```suggestion
            raise TypeError(f"`{f.__name__}` got an unexpected keyword argument `{arg_name}`")
```

dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -174,15 +174,15 @@ def _stop_after_this_page(self, data: Optional[List[Any]] = None) -> bool:
 
     def _handle_missing_total(self, response_json: Dict[str, Any]) -> None:
         raise ValueError(
-            f"Total {self.error_message_items} is not found in the response in"
-            f" {self.__class__.__name__}. Expected a response with a '{self.total_path}' key, got"
-            f" {response_json}"
+            f"Total `{self.error_message_items}` not found in the response in"
+            f" `{self.__class__.__name__}`.Expected a response with a '{self.total_path}' key, got"
```
djudjuu:
```suggestion
            f" `{self.__class__.__name__}`.Expected a response with a `{self.total_path}` key, got"
```

dlt/sources/rest_api/config_setup.py
```diff
@@ -219,27 +216,28 @@ def setup_incremental_object(
...
         if isinstance(param_config, Incremental):
             if param_config.end_value is not None:
                 raise ValueError(
-                    f"Only initial_value is allowed in the configuration of param: {param_name}. To"
-                    " set end_value too use the incremental configuration at the resource level."
-                    " See"
-                    " https://dlthub.com/docs/dlt-ecosystem/verified-sources/rest_api/basic#incremental-loading"
+                    "Only `initial_value` is allowed in the configuration of param:"
+                    f" `{param_name}`. "
+                    "To set end_value too use the incremental configuration at the resource level. "
```
djudjuu:
```suggestion
                    "To set `end_value` too use the incremental configuration at the resource level. "
```

dlt/sources/rest_api/config_setup.py
```diff
@@ -479,7 +480,7 @@ def _bind_path_params(resource: EndpointResource) -> None:
                 raise ValueError(
                     f"The path '{path}' defined in resource '{resource['name']}' contains a"
```
djudjuu:
```suggestion
                    f"The path `{path}` defined in resource `{resource['name']}` contains a"
```

dlt/sources/rest_api/config_setup.py
```diff
@@ -711,7 +711,7 @@ def _expressions_to_resolved_params(expressions: Set[str]) -> List[ResolvedParam
         parts = expression.strip().split(".", maxsplit=2)
         if len(parts) != 3:
             raise ValueError(
-                f"Invalid definition of {expression}. Expected format:"
+                f"Invalid definition of `{expression}`. Expected format:"
                 " 'resources.<resource>.<field>'"
```
djudjuu, line 716: nit: this file has a bunch of places where ' is still used. mostly in paths like these.

djudjuu (commented):
hey thierry,
what a ride. Reading nothing but exception messages is kinda fun. never did that before.

I found only two or three places that really need changes imo (prefixed my comment with `issue:`, the rest is just a bunch  of small things to adjust like missing backticks or ' that failed to get replaced.
I corrected or marked many of them, not all, but I think the codebase will be more unified for sure after this PR. (achieving 100% similarity wouldnt be a wise use of our time for sure)

the other main issue is that you probably need to adjust a couple more test-messages to match the new messages.

I'm on vacation until next week, so I won't mark this as changes requested to no tblock this from being merged becasue I am slow to respond, but my requested changes would be: the  issues i marked and make the tests pass. then its green from me.

dlt/common/configuration/exceptions.py
zilto: The `pass` keyword is not needed since there's a docstring.

## details

These are valid ways to define a "empty" class or function in Python (an "empty" function returns `None`):

```python
# with a pass
class MyClass:
  pass
  
def my_fn():
  pass
  
# with a docstring
class MyClass:
  """This is my class"""

def  my_fn():
  """This is my function"""
  
# with ellipsis
class MyClass:
  ...

def my_fn():
  ...

# ellipsis is typically used inline
class MyClass: ...

def my_fn(): ...
```

Those are all equivalent and can be combined. Personally, I find `pass` to be the most ambiguous because it can be used for other purposes in loops.

 "Empty" classes and functions are mostly used in typing stubs and `@typing.overload` to define type signatures. 

```python
# dlt.common.destination.dataset

class SupportsReadableRelation:
    # ...
    @overload
    def __getitem__(self, column: str) -> Self: ...

    @overload
    def __getitem__(self, columns: Sequence[str]) -> Self: ...

    def __getitem__(self, columns: Union[str, Sequence[str]]) -> Self:
        """Returns a new relation with the given columns selected.

        Args:
            columns (Union[str, Sequence[str]]): The columns to select.

        Returns:
            Self: The relation with the columns selected.
        """
        raise NotImplementedError("`__getitem__()` method is not supported for this relation")
```

dlt/common/libs/pyarrow.py
zilto: oh, I completely missed this. Good catch!

The numpy docs version `2.2 (latest)` and `1.26` indicate that `order="K"` is the default value. Then, I think the change from `"k"` (did nothing and defaulted to `"K"`) to `"K"` is safe

dlt/common/storages/transactional_file.py
zilto: I don't know, but "frontloading" is a good name for it! I don't have a rigorous process, but I try to write exceptions:

1. what's wrong in the fewest words possible. Here, it would be even better to raise a `LockSyncingError`
2. explain why the error is raised
3. include hints to start debugging
4. make it explicit when you interpolate a variable
5. keep long variables at the end (e.g., `self.path`)

Didn't want to rewrite all exceptions, but this one was hiding the key message at the very end "no lock file was found"

dlt/extract/exceptions.py
zilto: you're totally right

dlt/common/libs/pydantic.py
zilto: I have no strong opinion. It's just that VSCode showed a deprecation notice for `model.__fields__` in the IDE with message

```shell
The getter for property "__fields__" is deprecated
  The `__fields__` attribute is deprecated, use `model_fields` instead.Pylance
```

My guess would be that `__pydantic_fields__` is more of an internal attribute than a public-facing object

dlt/common/schema/utils.py
zilto: I'm unsure. I think it's legacy code from a refactoring. The developer didn't want to change the function signature by removing `columns_partial` (this would break downstream code), but instead added an explicit exception with a more useful error message.

Tip: when using git or GitHub, you can use `git blame` to blame who made the changes :). It points to specific commit where you'll find a useful commit message or PR hopefully.

dlt/destinations/dataset/ibis_relation.py
zilto: In this case, I kept the longer message to explain that you could be passing "many columns" or "a single column"

dlt/destinations/dataset/relation.py
zilto: It's inexpensive to check that `columns` is a `Sequence`, but checking `Sequence[str]` would require checking that all items in columns are `str`. This would be written as

```python
# reads as "if columns is not a sequence" and "if any item is not a string"
if not isinstance(columns, Sequence) and any(not isinstance(i, str) for i in columns):
```

dlt/destinations/impl/filesystem/sql_client.py
zilto: Good catch! You're right, I will revert to be safer.

PR #2718, opened 2025-06-05
transformations - updates

sh-rp:
### Description

This PR includes the following changes:
* Convert the readable dataset to handle sqlglot expressions where possible without intermediary steps where sql strings are generated and re-parsed.
 * Changes the transformation function to expect generators also for sql queries and relations
 *  Add a configuration option to force materialization (this is very useful for testing, but also may come in handy when doing data quality checks that require extraction later)
 * Uses regular relations as dataitem object, adds new interface to mark an object that may compute its schema_hints
 * Moves handling of hints for sql transformations into the extractor.
 * Adds support for working with unbound ibis tables as well as native support for unbound ibis expressions for the __call__ method of the dataset and transformations
 * Adds support for adding resource hints to pyarrow items that can be consumed by the ArrowExtractor
 * Removes ibis relation and consolidates dataset and relation implementations into one class each
 * Updates protocols for relation and dataset and makes these the publicly facing interface for these classes.
 * Updated docs

Open questions:
* The whole handling of hints needs to be reviewed and probably improved. I am unwrapping and wrapping dataitems in the transformations now and this feels more like a workaround at this point. Also I am not sure if the place where the hints get merged (user supplied hints vs hints from the relation) is right, I am not 100% sure what I am doing there.

### DIFF
--- a/dlt/common/destination/dataset.py
+++ b/dlt/common/destination/dataset.py
@@ -47,10 +47,12 @@ TFilterOperation = Literal["eq", "ne", "gt", "lt", "gte", "lte", "in", "not_in"]
 class DataAccess(Protocol):
     """Common data access protocol shared between dbapi cursors and relations"""
 
-    columns_schema: TTableSchemaColumns
-    """Returns the expected columns schema for the result of the relation. Column types are discovered with
-    sql glot query analysis and lineage. dlt hints for columns are kept in some cases. Refere to <docs-page> for more details.
-    """
+    @property
+    def columns_schema(self) -> TTableSchemaColumns:
+        """
+        Returns the expected columns schema for the result of the relation. Column types are discovered with
+        sql glot query analysis and lineage. dlt hints for columns are kept in some cases. Refere to <docs-page> for more details.
+        """
 
     def df(self, chunk_size: int = None) -> Optional[DataFrame]:
         """Fetches the results as arrow table. Uses the native pandas implementation of the destination client cursor if available.
@@ -275,6 +277,18 @@ class DBApiCursor(DataAccess):
 
     native_cursor: "DBApiCursor"
     """Cursor implementation native to current destination"""
+
+    @property
+    def columns_schema(self) -> TTableSchemaColumns:
+        return self._columns_schema
+
+    @columns_schema.setter
+    def columns_schema(self, value: TTableSchemaColumns) -> None:
+        self._columns_schema = value
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._columns_schema = None
 
     def execute(self, query: AnyStr, *args: Any, **kwargs: Any) -> None:
         """Execute a query on the cursor"""
--- a/dlt/destinations/dataset/relation.py
+++ b/dlt/destinations/dataset/relation.py
@@ -169,10 +169,6 @@ class ReadableDBAPIRelation(Relation, WithSqlClient):
             self._columns_schema, self.__qualified_query = self._compute_columns_schema()
         return self._columns_schema
 
-    @columns_schema.setter
-    def columns_schema(self, new_value: TTableSchemaColumns) -> None:
-        raise NotImplementedError("Columns Schema may not be set")
-
     @property
     def schema(self) -> TTableSchema:
         computed_columns, _ = self._compute_columns_schema(


dlt/transformations/transformation.py
```diff
@@ -149,38 +149,21 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
...
+            infer_sqlglot_schema=True,
+            allow_anonymous_columns=True,
+            allow_partial=True,
         )
         select_dialect = datasets[0].sql_client.capabilities.sqlglot_dialect
-        select_query = transformation_result.normalized_query.sql(dialect=select_dialect)
+        select_query = transformation_result.query()
 
-        # TODO: why? don't we prevent empty column schemas above?
         all_columns = {**computed_columns, **(columns or {})}
```
sh-rp: I have added a test for columns merging, and the columns yielded from here overwrite all incoming columns definitions from the decorator, which I think generally makes sense in dlt, but not for the transformations. I would maybe keep this topic open for now until we decide how this should work exactly.

tests/transformations/test_lineage.py
```diff
@@ -31,7 +33,6 @@ def sqlglot_schema() -> SQLGlotSchema:
 QUERY_KNOWN_TABLE_STAR_SELECT = "SELECT * FROM table_1"
 QUERY_UNKNOWN_TABLE_STAR_SELECT = "SELECT * FROM table_unknown"
 QUERY_ANONYMOUS_SELECT = "SELECT LEN(col_varchar) FROM table_1"
-QUERY_GIBBERISH = "%&/ GIBBERISH"
```
sh-rp, line 34: this can not happen anymore, as non parseable sql strings are already caught in the dataset.

tests/transformations/test_transformations.py
```diff
@@ -49,6 +49,43 @@ def copied_purchases(dataset: SupportsReadableDataset[Any]) -> Any:
...
+
+    assert fruit_p.default_schema.tables["inventory"]["columns"]["price"]["precision"] == 10
+    assert fruit_p.default_schema.tables["inventory"]["columns"]["price"]["scale"] == 2
+
+    # we can now transform this table twice, one with changed hints and once with the original hints
+    @dlt.transformation()
+    def inventory_original(dataset: SupportsReadableDataset[Any]) -> Any:
+        return dataset["inventory"]
+
+    @dlt.transformation(columns={"price": {"precision": 20, "scale": 2}})
```
sh-rp: I think this is the correct behavior, we should allow to set the precision for an existing column here. Or would you rather do this in the function body? I am not sure.

docs/website/docs/general-usage/transformations/index.md
```diff
@@ -110,6 +110,44 @@ If you prefer to write your queries in SQL, you can omit ibis expressions by sim
...
+Given a table of the name `my_table` with the columns `id` and `value` on `duckdb`, on a dataset with the name `my_dataset`, loaded into a dataset named `transformed_dataset`, the following query:
+
+```sql
+SELECT id, value FROM table
+```
+
+Will be translated to
+
+```sql
+INSERT INTO
```
sh-rp: this is the actually generated insert statement. I realized we were not quoting identifiers, so I added this. We are doing 2 subqueries here now, one stemming from the normalization in the extract and one from @anuunchin normalizer work. We can probably still optimize here?

tests/normalize/test_model_item_normalizer.py
```diff
@@ -469,52 +469,54 @@ def test_dlt_column_addition_configs(
...
+            == 'SELECT _dlt_subquery."a" AS "a", _dlt_subquery."b" AS "b",'
+            f" '{another_load_id}' AS"
+            ' "_dlt_load_id" FROM (SELECT a, b, MY_CUSTOM_FUNC() AS _dlt_id FROM my_table) AS'
+            " _dlt_subquery\n"
         )
     elif not add_dlt_load_id and add_dlt_id:
         assert (
             normalized_query
-            == 'SELECT _dlt_subquery."a" AS a, _dlt_subquery."b" AS b, UUID() AS _dlt_id FROM'
+            == 'SELECT _dlt_subquery."a" AS "a", _dlt_subquery."b" AS "b", UUID() AS "_dlt_id" FROM'
```
sh-rp, line 508: the model item normalizer will now always use quoted identifiers on the outer queries, regardless wether the inner select has quoted identifiers. Since all queries arriving here now go through query normalization in the dataset where quotes are applied, this should be fine acutally.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -32,30 +33,13 @@ def __init__(
...
-
-        # here we need to transpile to ibis default and transpile back to target with sqlglot
-        # NOTE: ibis defaults to the duckdb dialect, if a dialect is not passed
-        sql = ibis.to_sql(ibis_expr)
-        sql = sqlglot.transpile(sql, write=target_dialect)[0]
-        return sql
+    def _query(self) -> sge.Query:
+        from dlt.helpers.ibis import duckdb_compiler
+
+        select_query = duckdb_compiler.to_sqlglot(self._ibis_object)
```
sh-rp: For converting an ibis expression into sqlglot, you always need a compiler which is destination dependent. I think the compiler will fail if you use function that are known not to exist on a specific destination, and certain materializations will also occur in this step. If you have an expression that counts rows, the alias for the result will be created here, which will be `CountStar(*)` for duckdb, but which is not a valid identifier for bigquery for example (see comment in query normalizer)

dlt/destinations/dataset/utils.py
```diff
@@ -144,6 +145,8 @@ def normalize_query(
                 if len(expanded_path) == 3:
                     if node.db != expanded_path[0]:
                         node.set("catalog", sqlglot.to_identifier(expanded_path[0], quoted=False))
+        if isinstance(node, sge.Alias):
+            node.set("alias", naming.normalize_identifier(node.alias))
```
sh-rp: Here I make sure aliases that are created by the user or automatically by ibis in the compile step above adhere to the naming convention of the current destination. The alternative would be to use a different compiler for every destination. Normalizing the alias in the ModelNormalizer step is too late, since this code is also used for just accessing the data. Without this change, `def test_row_counts` will fail for bigquery because an alias with invalid symbols remains in the query. All other destinations seem to accept this.

sh-rp: I have reverted this change and am now using the bigquery compiler only for bigquery destinations. Maybe we should be using the correct compiler for each destination type, but it seems duckdb works for all. I am not sure about this one.

dlt/destinations/dataset/ibis_relation.py
```diff
@@ -32,30 +25,21 @@ def __init__(
...
-        sql = ibis.to_sql(ibis_expr)
-        sql = sqlglot.transpile(sql, write=target_dialect)[0]
-        return sql
+    def _query(self) -> sge.Query:
+        from dlt.helpers.ibis import duckdb_compiler, bigquery_compiler
+
+        # NOTE: We can use the duckdb compiler for all dialects except for bigquery
+        # as bigquery is more strict about identifier naming and will not accept
+        # identifiers generated by the duckdb compiler for anonymous columns
+        destination_dialect = self._dataset.sql_client.capabilities.sqlglot_dialect
```
rudolfix: I'll keep the old notes:
        # NOTE: ibis is optimized for reading a real schema from db and pushing back optimized sql
        #  - it quotes all identifiers, there's no option to get unqoted query
        #  - it converts full lists of column names back into star
        #  - it optimizes the query inside, adds meaningless aliases to all tables etc.

for example: I bet that BigQuery problem is quotation problem where `"` are not accepted

dlt/destinations/dataset/relation.py
```diff
@@ -62,19 +62,23 @@ def sql_client(self) -> SqlClientBase[Any]:
...
+        else:
+            query = self._normalized_query
+
+        if not isinstance(query, sge.Query):
+            raise ValueError(
+                f"Query `{query}` received for `{self.__class__.__name__}`. "
+                "Must be an SQL SELECT statement."
+            )
+
+        return query.sql(dialect=self._dataset.sql_client.capabilities.sqlglot_dialect)
```
rudolfix: why `provided_dialect` is not used here?

dlt/destinations/dataset/relation.py
```diff
@@ -29,19 +29,19 @@ def __init__(
...
         self._dataset = readable_dataset
-        self._should_normalize_query: bool = normalize_query
+        self._execute_raw_query: bool = execute_raw_query
 
         # derived / cached properties
         self._opened_sql_client: SqlClientBase[Any] = None
         self._columns_schema: TTableSchemaColumns = None
-        self._qualified_query: sge.Query = None
-        self._normalized_query: sge.Query = None
+        self.__qualified_query: sge.Query = None
```
rudolfix, line 103: is this intended? names will be mangled which is maybe good because we implement `__getitem__` to get column names here

dlt/destinations/dataset/utils.py
```diff
@@ -108,6 +108,7 @@ def normalize_query(
     sqlglot_schema: SQLGlotSchema,
     qualified_query: sge.Query,
     sql_client: SqlClientBase[Any],
+    naming: NamingConvention,
```
rudolfix: not needed anymore. and I think it should not be needed here

dlt/transformations/transformation.py
rudolfix: w8: you are overwriting `computed_columns` with `columns` which come from the decorator. so how are they not overwritten at the end?

dlt/transformations/transformation.py
```diff
@@ -150,38 +150,21 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
...
-            ]
-
-            if unknown_column_types:
-                raise UnknownColumnTypesException(
-                    resource_name,
-                    "For sql transformations all data_types of columns must be known. "
-                    + "Please run with strict lineage or provide data_type hints "
-                    + f"for following columns: {unknown_column_types}",
-                )
             yield dlt.mark.with_hints(
```
rudolfix: this is something I wrote in the old PR: could we place `all_column` in the `SqlModel`? or better: could we place a relation in it? then:
* user is able to use `dlt.mark` on the model (ie. their own custom hints or table name)
* `all_columns = {**computed_columns, **(columns or {})}` is not needed. columns will be applied by `compute_table_schema` in `DltResource`
* you can use relation/computed_columns in SqlModel to apply to the resource in extractor. Look how we call this method
```py
root_table_schema = resource.compute_table_schema(items, meta)
```
and 
```py
def compute_table_schema(self, item: TDataItem = None, meta: Any = None) -> TTableSchema:
        """Computes the table schema based on hints and column definitions passed during resource creation.
        `item` parameter is used to resolve table hints based on data.
        `meta` parameter is taken from Pipe and may further specify table name if variant is to be used
        """
```
so SqlModel will be passed there and you can use it to extract computed hints.

Or you can implement `_compute_schema` on `ModelExtractor` for fully custom logic

rudolfix (changes requested):

dlt/transformations/transformation.py
sh-rp: I am not sure what you are asking. My goal would be, that if the user sets a column hin on the transformation decorator, this should overwrite whatever is discovered during lineage. At least I think it should, but I can also just remove this line and users will have to deal with this in some other way if they need to change precision or something like that.

dlt/destinations/dataset/relation.py
sh-rp: Yes, so we can have a private _normalized_query property (defined further below)

dlt/transformations/transformation.py
sh-rp: I kind of did that now, but I also have to unwrap and rewrap the dataitem in the transform function and I am not fully convinced that the way I did it is the right way.. Also I don't think it works for materialized queries...

dlt/transformations/transformation.py
```diff
@@ -2,45 +2,60 @@
...
-from dlt.extract.exceptions import (
-    CurrentSourceNotAvailable,
-)
+from dlt.transformations.configuration import TransformationConfiguration
+from dlt.common.utils import get_callable_name
+from dlt.extract.exceptions import CurrentSourceNotAvailable
+from dlt.extract.pipe_iterator import DataItemWithMeta
+
+
+class MaterializableSqlModel(SqlModel, WithComputableHints):
```
sh-rp: Note: find better name, or merge this with relation.

dlt/transformations/transformation.py
```diff
@@ -66,40 +81,70 @@ def make_transformation_resource(
...
+            meta = original_first_item.meta
+            unwrapped_item = original_first_item.data
+
+        # catch the two cases where we get a relation from the transformation function
+        # NOTE: we only process the first item, all other things that are still in the generator are ignored
+        if isinstance(unwrapped_item, BaseReadableDBAPIRelation):
+            relation = unwrapped_item
+        # we see if the string is a valid sql query, if so we need a dataset
+        elif isinstance(unwrapped_item, str):
+            try:
```
sh-rp: this will allow a malformed sql query to be interpreted as string, possibly not what we want here, maybe single strings should always strictly be interpreted as sql query and fail this transformation, otherwise the user will have very strange results.

dlt/transformations/transformation.py
```diff
@@ -108,88 +153,35 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
...
+            if meta:
+                yield DataItemWithMeta(meta, sql_model)
+            else:
+                yield sql_model
         else:
-            # NOTE: dataset will not execute query over unknown tables or columns
-            for chunk in datasets[0](select_query).iter_arrow(chunk_size=config.buffer_max_items):
-                yield dlt.mark.with_hints(chunk, hints=make_hints(columns=all_columns))
+            for chunk in relation.iter_arrow(chunk_size=config.buffer_max_items):
+                yield dlt.mark.with_hints(chunk, hints=sql_model.compute_hints())
```
sh-rp: here we discard the hints coming from the outside, for the sqlmodel they are merged during extract, but since we have pure arrow tables here, we can't do it. We had this idea of putting our schema into the user data area of the arrow table. I have not looked at this yet, but should we do it?

dlt/extract/extractors.py
```diff
@@ -129,6 +129,11 @@ def __init__(
 
     def write_items(self, resource: DltResource, items: TDataItems, meta: Any) -> None:
         """Write `items` to `resource` optionally computing table schemas and revalidating/filtering data"""
+
+        if isinstance(items, WithComputableHints):
```
sh-rp: NOTE: this place and the place below should be reviews, I am not sure wether this is the correct place to do this..

dlt/extract/extractors.py
```diff
@@ -337,6 +342,19 @@ def _retrieve_normalize_config(self) -> ItemsNormalizerConfiguration:
         )
 
     def write_items(self, resource: DltResource, items: TDataItems, meta: Any) -> None:
+        items_list = items if isinstance(items, list) else [items]
+
+        # extract resource hints from metadata if available
+        for item in items_list:
```
sh-rp: same as right above

dlt/extract/hints.py
```diff
@@ -99,9 +102,32 @@ def __init__(
...
+class WithComputableHints:
+    """
+    A class that can compute a schema from a data item.
+    """
+
+    def compute_hints(self) -> TResourceHints:
+        pass
+
+
+class SqlModel:
```
sh-rp, line 105: This is still used in the normalizer phase to read and write the models.

dlt/helpers/ibis.py
```diff
@@ -209,3 +211,52 @@ def create_unbound_ibis_table(schema: Schema, dataset_name: str, table_name: str
     unbound_table = ibis.table(schema=ibis_schema, name=table_name, database=dataset_name)
 
     return unbound_table
+
+
+def get_compiler_for_dialect(dialect: str) -> SQLGlotCompiler:
```
sh-rp: Ideally ibis expressions would be convertable into sqlglot without any notion of the current dialect, but at least for bigquery this does not work for count queries. When converting from ibis to sqlglot, aliase name are rendered and for an expression of for example "dataset.items.count()" it will render something like this when converted to sql:

```sql
SELECT COUNT(*) as `COUNT(*)` FROM `items`
```

and count(*) is not a valid identifier in bigquery. If you rend this with the bigquery compiler it does something like this:

```sql
SELECT COUNT(*) as `count_star` FROM `items`
```

which works.

tests/transformations/test_transformation_decorator.py
```diff
@@ -44,19 +50,7 @@ def transform(dataset: SupportsReadableDataset[Any]) -> Any:
...
-    def transform(dataset: SupportsReadableDataset[Any]) -> Any:
-        return {"some": "data"}
-
-    with pytest.raises(PipelineStepFailed) as excinfo:
-        p.run(transform(dlt.dataset(dlt.destinations.duckdb("input_data"), "dataset_name")))
-
-    assert "Please either return a valid sql string or" in str(excinfo.value)
-
-
+@pytest.mark.skip(reason="TODO: fix this test")
```
sh-rp, line 55: This is hard to catch now, since we have to execute the original function to detect which type (transformation or regular resource) is in the decorated function

dlt/helpers/ibis.py
```diff
@@ -209,3 +212,52 @@ def create_unbound_ibis_table(schema: Schema, dataset_name: str, table_name: str
     unbound_table = ibis.table(schema=ibis_schema, name=table_name, database=dataset_name)
 
     return unbound_table
+
+
+def get_compiler_for_dialect(dialect: TSqlGlotDialect) -> SQLGlotCompiler:
```
rudolfix, line 217: this is how ibis does it:
```py
    import ibis.backends.sql.compilers as sc

    try:
        compiler_provider = getattr(sc, dialect)
    except AttributeError as e:
        raise ValueError(f"Unknown dialect {dialect}") from e

    if (compiler := getattr(compiler_provider, "compiler", None)) is None:
        raise NotImplementedError(f"{compiler_provider} is not a SQL backend")

    out = compiler.to_sqlglot(expr.unbind(), **kwargs)
```

dlt/destinations/dataset/dataset.py
```diff
@@ -328,3 +330,54 @@ def __str__(self) -> str:
...
+            return table
+
+        # fallback to the standard dbapi relation
+        return ReadableDBAPIRelation(
+            readable_dataset=self,
+            table_name=table_name,
+        )
+
+
+class ReadableIbisDataset(BaseReadableDBAPIDataset[ReadableIbisRelation]):
```
rudolfix: my take is to completely remove it. maybe there's some user code that uses `ibis_relation` but I think is better to announce the change cleanup our code now.

dlt/destinations/dataset/dataset.py
```diff
@@ -32,19 +33,24 @@
...
+    from dlt.helpers.ibis import Table as IbisTable
+    from dlt.helpers.ibis import Expr as IbisExpr
 else:
     IbisBackend = Any
+    IbisTable = Any
+    IbisExpr = Any
 
 
-class ReadableDBAPIDataset(SupportsReadableDataset[ReadableIbisRelation]):
+class BaseReadableDBAPIDataset(
```
rudolfix: we do not need generic relations and datasets anymore! I think this is the biggest win from removing `ibis_relation`

dlt/common/destination/dataset.py
```diff
@@ -9,24 +9,29 @@
...
     from dlt.common.libs.pandas import DataFrame
     from dlt.common.libs.pyarrow import Table as ArrowTable
-    from dlt.helpers.ibis import BaseBackend as IbisBackend
+    from dlt.helpers.ibis import BaseBackend as IbisBackend, Table as IbisTable, Expr as IbisExpr
 else:
     DataFrame = Any
     ArrowTable = Any
     IbisBackend = Any
+    IbisTable = Any
+    IbisExpr = Any
```
rudolfix, line 40: we do not declare several methods below like `select` or `order_by`. why?

btw. `where` is not correct. `filter` is used in ibis, duckdb 
also we should support column selectors that is compatible with this spec. I think we have one compatible with ibis?

rudolfix: maybe we should rename those to `Relation` and `Dataset`? but then we need to use them in all user-facing methods and particular implementations remain hidden. WDYT?

dlt/common/destination/dataset.py
```diff
@@ -37,18 +42,22 @@ class SupportsReadableRelation:
     sql glot query analysis and lineage. dlt hints for columns are kept in some cases. Refere to <docs-page> for more details.
     """
 
-    def query(self, pretty: bool = False) -> Any:
+    def query(self, pretty: bool = False) -> str:
```
rudolfix: this is probably better than having `Generic` relation. Here we basically say that all queries can be expressed as sql

dlt/common/destination/dataset.py
```diff
@@ -272,17 +283,43 @@ def dataset_name(self) -> str:
             str: The name of the dataset
         """
 
-    def __call__(self, query: Any, normalize_query: bool = True) -> SupportsReadableRelation:
+    def __call__(
+        self,
+        query: Union[str, sge.Select, IbisExpr],
+        query_dialect: TSqlGlotDialect = None,
+        execute_raw_query: bool = False,
```
rudolfix: this is needed by a few crazy tests. better if we use `__execute_raw_query` field that you can set in the test.

also: should we have an alias `query()`?

dlt/destinations/dataset/dataset.py
```diff
@@ -177,14 +176,31 @@ def _ensure_schema(self) -> None:
...
+            query_dialect=query_dialect,
+            execute_raw_query=execute_raw_query,
         )
 
-    def table(self, table_name: str) -> ReadableIbisRelation:
+    @overload
+    def table(self, table_name: str) -> TReadableRelation: ...
+
+    @overload
+    def table(self, table_name: str, table_type: Literal["ibis"]) -> IbisTable: ...
```
rudolfix, line 199: having the same API to create relation and query is painful. but we already agreed to do it. re. implementation - I'll probably move generation of queries/expressions to `Schema` where they belong.

dlt/destinations/dataset/dataset.py
```diff
@@ -177,14 +176,31 @@ def _ensure_schema(self) -> None:
...
+
+    @overload
+    def table(self, table_name: str, table_type: Literal["relation"]) -> TReadableRelation: ...
+
+    def table(
+        self, table_name: str, table_type: Literal["relation", "ibis"] = None
+    ) -> Union[TReadableRelation, IbisTable]:
```
rudolfix: I'd just use `Any` here.

dlt/destinations/dataset/factory.py
```diff
@@ -16,7 +18,7 @@ def dataset(
     dataset_name: str,
     schema: Union[Schema, str, None] = None,
     dataset_type: Literal["ibis"] = "ibis",
-) -> ReadableDBAPIDataset: ...
+) -> ReadableIbisDataset: ...
```
rudolfix: this is user facing api? So `Dataset` type would be the best

dlt/destinations/dataset/factory.py
```diff
@@ -43,4 +45,40 @@ def dataset(
     schema: Union[Schema, str, None] = None,
     dataset_type: TDatasetType = "auto",
```
rudolfix: dataset type should go. maybe we can add warnings in DBAPIRelation when unknown method is used? but tbh. I would not worry too much. I doubt ibis relation was used a lot.

dlt/destinations/dataset/relation.py
```diff
@@ -257,53 +285,56 @@ class ReadableDBAPIRelation(BaseReadableDBAPIRelation):
     def __init__(
```
rudolfix: IMO those overloads do not make sense now

dlt/transformations/transformation.py
```diff
@@ -66,40 +64,72 @@ def make_transformation_resource(
...
-        config: TransformationConfiguration = get_fun_last_config(func) or get_fun_spec(func)()  # type: ignore[assignment]
-
+        # Collect all datasets from args and kwargs
         all_arg_values = list(args) + list(kwargs.values())
+        datasets: List[BaseReadableDBAPIDataset[Any]] = [
+            arg for arg in all_arg_values if isinstance(arg, BaseReadableDBAPIDataset)
+        ]
+
+        # get first item from gen and see what we're dealing with
+        gen = func(*args, **kwargs)
```
rudolfix: allow resources that return as well

dlt/transformations/transformation.py
```diff
@@ -66,40 +64,72 @@ def make_transformation_resource(
...
+        original_first_item = next(gen)
+
+        # unwrap if needed
+        meta = None
```
rudolfix: this looks complicated. why can't we iterate the generator and decide for each item separately:
```
for item in gen:
    # decide what to do with item
    yield something
```

dlt/transformations/transformation.py
```diff
@@ -108,88 +138,42 @@ def transformation_function(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
...
+                yield relation
         else:
-            # NOTE: dataset will not execute query over unknown tables or columns
-            for chunk in datasets[0](select_query).iter_arrow(chunk_size=config.buffer_max_items):
-                yield dlt.mark.with_hints(chunk, hints=make_hints(columns=all_columns))
+            from dlt.common.libs.pyarrow import add_arrow_metadata
+
+            serialized_hints = json.dumps(relation.compute_hints())
+            for chunk in relation.iter_arrow(chunk_size=config.buffer_max_items):
+                updated_chunk = add_arrow_metadata(
```
rudolfix: we need to test if metadata survives incremental. o3 says it should , except a few weird cases

dlt/extract/extractors.py
rudolfix: this is not bad. but we have `DltTransformationResource` and IMO we could use this fact to create custom `compute_table_schema` that will:
* know about Relation and Arrow with metadata
* will detect `item` type, extract and apply hints

this may require a small refactor, the order of hints should be as follows (higher means precedence)
- hints coming from `column` or existing schema
- hints coming from meta
- hints coming from data item

as a bonus: we could add a few convenience methods to `DltTransformationResource`
- relation - will evaluate `gen` (via existing iter) and return first relation from it
- `compute_table_schema` if no item is provided would do the above and compute schema using first item in `gen`

rudolfix (commented):
code is pretty close. I still have doubts if our data access interfaces are clear enough
- we should indicate which classes are user facing and name them `Dataset` and `Relation`
- our top level relation assumes that queries are SQL and have a dialect.... I want to look at them again

I think we should get rid of ibis relation even if this is not backward compatible.

dlt/common/destination/dataset.py
```diff
@@ -272,37 +279,80 @@ def dataset_name(self) -> str:
             str: The name of the dataset
         """
 
-    def __call__(self, query: Any, normalize_query: bool = True) -> SupportsReadableRelation:
+    @property
+    def open_table_client(self) -> SupportsOpenTables:
```
sh-rp: I am not sure this should be here, I added it because it was used in a test..

dlt/common/destination/dataset.py
sh-rp: we have where and filter, I completely updated these protocols, I think they are much better now.

dlt/destinations/dataset/dataset.py
sh-rp: done!!

dlt/helpers/ibis.py
sh-rp: I'm not sure what to do with this snippet? I think we are doing exactly the same. Getting the right compiler for the dialect and compiling an unbound table (which it already is)

dlt/common/destination/dataset.py
sh-rp: This is now hidden in the relation interface anyway.

sh-rp: Maybe the correct name for this would be `to_sql()`?

dlt/destinations/dataset/dataset.py
sh-rp: I also don't really like it.. I was more in favor of dataset.items_table.ibis() or something like that. But with the overloads at least the typing is safe.

dlt/common/destination/dataset.py
sh-rp: I did both (except for just one underscore in the raw query thing)

dlt/transformations/transformation.py
sh-rp: I'm not quite sure what you mean, but at the moment the arrow metadata is an internal function that should not be used by the user. We could also attach it to every arrow table that gets yielded from a relation, regardless wether it is in the transformation context or not, but then the user could change the table and we would have a missmatching schema.

dlt/transformations/transformation.py
sh-rp: I did that and some other improvements, I'm not sure I can make it any easier, but maybe you have an idea.

dlt/transformations/transformation.py
sh-rp: done

dlt/common/data_writers/writers.py
```diff
@@ -191,16 +191,9 @@ def write_header(self, columns_schema: TTableSchemaColumns) -> None:
...
-            query = item.query
-            parsed_query = sqlglot.parse_one(query, read=dialect)
-
-            # Ensure the parsed query is a SELECT statement
-            if not isinstance(parsed_query, sqlglot.exp.Select):
-                raise ValueError("Only SELECT statements are allowed to write model files.")
-
-            normalized_query = parsed_query.sql(dialect=dialect)
-            self._f.write("dialect: " + (dialect or "") + "\n" + normalized_query + "\n")
+            dialect = item.query_dialect() or (self._caps.sqlglot_dialect if self._caps else None)
```
rudolfix: is it possible to not have `query_dialect`?

dlt/common/destination/dataset.py
```diff
@@ -1,3 +1,6 @@
+from types import TracebackType
```
rudolfix, line 1: do we need this?

dlt/common/destination/dataset.py
rudolfix: it should be where `destination_client` is. maybe we promote those to the public interface later

dlt/common/destination/dataset.py
```diff
@@ -272,37 +279,98 @@ def dataset_name(self) -> str:
...
+
+        Returns:
+            SupportsOpenTables: The open table client for the dataset
+        """
+        ...
+
+    def __call__(
+        self,
+        query: Union[str, sge.Select, IbisExpr],
+        query_dialect: TSqlGlotDialect = None,
```
rudolfix, line 330: why we need to specify dialect? so we can use other dialect than default dataset dialect (one from the destination)?

dlt/common/libs/pyarrow.py
```diff
@@ -1047,3 +1047,25 @@ class NameNormalizationCollision(ValueError):
     def __init__(self, reason: str) -> None:
         msg = f"Arrow column name collision after input data normalization. {reason}"
         super().__init__(msg)
+
+
+def add_arrow_metadata(
```
rudolfix, line 1082: there are several function in that module that will remove this metadata but don't have to. basically all that recreate schemas from list of columns and arrays. also method that creates arrow schema from columns does not set this metadata. maybe it should? at least we need to mark it as tech debt

dlt/destinations/dataset/dataset.py
```diff
@@ -193,26 +210,16 @@ def table(self, table_name: str) -> ReadableIbisRelation:
...
-            )
-
-        return relation  # type: ignore[return-value]
+        # fallback to the standard dbapi relation
+        return ReadableDBAPIRelation(
+            readable_dataset=self,
+            table_name=table_name,
+        )
 
     def row_counts(
```
rudolfix, line 224: what should we do with this? I was thinking if we could inject additional methods into dataset? like query packages ie. dataset.stats.row_count. or maybe we move this to queries? this is not public interface now, right?

dlt/destinations/dataset/dataset.py
```diff
@@ -255,13 +262,13 @@ def row_counts(
             else:
                 union_all_expr = union_all_expr.union(counts_expr, distinct=False)
 
-        return self(query=union_all_expr)
+        return self.query(query=union_all_expr)
 
-    def __getitem__(self, table_name: str) -> ReadableIbisRelation:
+    def __getitem__(self, table_name: str) -> ReadableDBAPIRelation:
```
rudolfix: why this is not `Relation`?

dlt/destinations/dataset/relation.py
```diff
@@ -21,80 +24,159 @@
...
+        return self._wrap_iter("iter_df")(*args, **kwargs)
+
+    def iter_arrow(self, *args: Any, **kwargs: Any) -> Any:
+        return self._wrap_iter("iter_arrow")(*args, **kwargs)
+
+    def iter_fetch(self, *args: Any, **kwargs: Any) -> Any:
+        return self._wrap_iter("iter_fetch")(*args, **kwargs)
 
+    @property
+    def columns_schema(self) -> TTableSchemaColumns:
```
rudolfix, line 167: hmmm maybe we can convert this into `schema`? and return `TTableSchema`? why? to propagate table hints as well. I'm pretty sure we'll have to change that soon

dlt/destinations/dataset/relation.py
```diff
@@ -21,80 +24,159 @@
...
+        if self._columns_schema is None:
+            self._columns_schema, self.__qualified_query = self._compute_columns_schema()
+        return self._columns_schema
+
+    @columns_schema.setter
+    def columns_schema(self, new_value: TTableSchemaColumns) -> None:
```
rudolfix: can we still set it?

dlt/extract/hints.py
```diff
@@ -55,11 +56,13 @@
 from dlt.extract.items_transform import ValidateItem
 from dlt.extract.utils import ensure_table_schema_columns, ensure_table_schema_columns_hint
 from dlt.extract.validation import create_item_validator
-from dlt.common.data_writers import TDataItemFormat
 
 import sqlglot
 
 
+DLT_HINTS_METADATA_KEY = "dlt_resource_hints"
```
rudolfix, line 63: not used?

dlt/helpers/ibis.py
rudolfix: what I mean here is that `ibis` has a convention to get compiler by name:
```py
compiler_provider = getattr(sc, dialect)
```
which should replace all that ifs below

dlt/common/destination/dataset.py
```diff
@@ -208,44 +228,31 @@ def __getitem__(self, columns: Union[str, Sequence[str]]) -> Self:
...
-
-        Args:
-            attr (str): The attribute to get.
-
-        Returns:
-            Any: The attribute of the relation
-        """
-        raise NotImplementedError("`__getattr__()` method is not supported for this relation")
 
     def __copy__(self) -> Self:
```
rudolfix, line 265: also __str__ and __repr__ should be declared I think?

dlt/destinations/dataset/relation.py
```diff
@@ -394,5 +425,24 @@ def __getitem__(self, columns: Sequence[str]) -> Self:
...
+            "Relation"
+            f" query:\n{indent(query_expr.sql(dialect=self.query_dialect(), pretty=True), prefix='  ')}\n"
+        )
+        msg += "Columns:\n"
+        for column in self.columns_schema.values():
+            # TODO: show x-annotation hints
+            msg += f"{indent(column['name'], prefix='  ')} {column['data_type']}\n"
+        return msg
+
+    def __copy__(self) -> Self:
```
rudolfix, line 495: OK this is more or less what __repr__ should show:
- dataset
- query as text (SQL)

dlt/transformations/transformation.py
```diff
@@ -1,52 +1,80 @@
...
+    from dlt.common.libs.pyarrow import pyarrow
+except (ImportError, MissingDependencyException):
+    pyarrow = None
 
 
 class DltTransformationResource(DltResource):
     def __init__(self, *args: Any, **kwds: Any) -> None:
         super().__init__(*args, **kwds)
 
+    def compute_table_schema(self, item: TDataItem = None, meta: Any = None) -> TTableSchema:
```
rudolfix, line 61: this is good. but we have one more complication. `extract` is caching static tables (where schema will not mutate). we check this like:
```py
def _get_static_table_name(self, resource: DltResource, meta: Any) -> Optional[str]:
        if resource._table_name_hint_fun:
```
maybe we should make this interface explicit? `has_static_hints() -> bool` and for transformations it is always false.

also this should be possibly fixed:
```py
def _get_dynamic_table_name(self, resource: DltResource, item: TDataItem) -> str:
        return self._normalize_table_identifier(resource._table_name_hint_fun(item))
```
so we can hide `._table_name_hint_fun` completely

dlt/transformations/transformation.py
```diff
@@ -66,130 +94,110 @@ def make_transformation_resource(
...
+                for chunk in relation.iter_arrow(chunk_size=config.buffer_max_items):
+                    yield add_arrow_metadata(chunk, {DLT_HINTS_METADATA_KEY: serialized_hints})
+
+        # support both generator and function
+        gen_or_item = func(*args, **kwargs)
+        iteratable_items = gen_or_item if isinstance(gen_or_item, Iterator) else [gen_or_item]
+
+        for item in iteratable_items:
+            if transformation_processed:
+                raise TransformationException(
```
rudolfix: heh why we should block it? maybe single query per transformation is a good thing but I can totally see people yielding model and a few arrow tables with intermediary data. or dispatch models to tables via dynamic table names. so they will yield many models.

ie. a model per each type of event in github

rudolfix (commented):
OK! this is pretty minor stuff now. I have one question for execution order. We execute models randomly right? so if I access data from output dataset (ie. dlt.current.pipeline.dataset()) I may get partially modified destination data.

maybe we should have a mode that we always use staging dataset? it seems we just need append strategy

rudolfix:
I will review tests later

dlt/common/data_writers/writers.py
sh-rp: What we can do is to always save all queries in the duckdb dialect for example and transpile them to the right dialect before execution. This was actually the original implementation but by your request we are now saving the queries in the dialect of the dataset they where created on. Or do you just want to get rid of this method? Anyway, either way is fine by me, let me know.

dlt/common/destination/dataset.py
sh-rp: it is used in the signature of the `__exit__` method

dlt/common/destination/dataset.py
sh-rp: Yes, correct. I wrote that in the docstring too, you can use duckdb dialect for mssql for example (mssql does not have offset limit but a more complicated notation for example). We can also remove it, but I think it could be useful. We only use it when parsing the query into the sqlglot expression.

dlt/destinations/dataset/dataset.py
sh-rp: Row counts is a public interface and it is documented already since we released the dataset. But I agree that it would be nice to have a place for special queries such as @zilto will need for the incremental datasets where you can filter by load_id etc. Maybe let's keep it here for now and move it once we know how to access these special queries?

dlt/extract/hints.py
sh-rp: it is used in `dlt/extract/hints.py`

dlt/common/libs/pyarrow.py
sh-rp: Ok, I will not do this now, but have started a ticket for initial transformation release follow up work: https://github.com/dlt-hub/dlt/issues/2827

dlt/destinations/dataset/dataset.py
sh-rp: I changed it for all places where a relation is returned

dlt/destinations/dataset/relation.py
sh-rp: Since this is defined by schema + query, I am not sure why we would want to set this? If you want to supply additional hints, you do this on the decorator of the transformation or with meta. Or what is the exact ask?

dlt/destinations/dataset/relation.py
sh-rp: Right now I am using this: `def compute_hints(self) -> TResourceHints:`, but we could change this to a schema property? The colums_schema has been on the relation since we created it and was inherited from the common interface with the cursor.

dlt/common/destination/dataset.py
sh-rp: I was not sure, but o4 thinks it it is not needed unless we are doing something special with those functions (we are not). They are part of the base object and are always there.

sh-rp: Copy on the other hand is not guaranteed to exist, so it should be here.

dlt/destinations/dataset/relation.py
sh-rp: I have converted it to a schema property now that also is on the protocol. I need to convert it back to a hints format though to be able to use it in the transform resource.

dlt/helpers/ibis.py
sh-rp: done, does not replace all ifs, but is shorter now :)

sh-rp:
> I will review tests later

There are a bunch of relevant tests in `tests/load/pipeline/test_model_item_format.py` which can probably be moved to transformation tests.

dlt/destinations/dataset/relation.py
sh-rp: I added those including tests

dlt/destinations/dataset/relation.py
rudolfix: I hope we can remove it!

dlt/common/destination/dataset.py
rudolfix: right! I think they are inherited from the Object

dlt/transformations/transformation.py
sh-rp: this was blocked because the hints did not update and I did not know why, now I do and this works and I allow it an uncommented the test :)

dlt/transformations/transformation.py
sh-rp: I did this, but we still need `_table_name_hint_fun` in the extractor to get the table name. I also tried to set the has_dynamic_hints to true if there are other dynamic hints, this works but it would change the extractor behavior. I am not sure if there maybe is a bug in there from before? It will only start writing to dynamic tables and updating hints if the table name is dynamic, this does not seem right..

sh-rp:
> OK! this is pretty minor stuff now. I have one question for execution order. We execute models randomly right? so if I access data from output dataset (ie. dlt.current.pipeline.dataset()) I may get partially modified destination data.
> 
> maybe we should have a mode that we always use staging dataset? it seems we just need append strategy

I would add a note in the docs, I think it's quite unlikely that users would use tables from the output dataset as input in the same load.

dlt/transformations/transformation.py
rudolfix: this is correct - table name must be dynamic for other hints to be dynamic. but I can't find where it was enforced...

dynamic hints can be pretty slow so the idea was:
- if only table name is dynamic we can still cache hints per table name
- if anything more is dynamic then we must always (for each item) recompute the hints

I'll take look at the code

tests/transformations/test_transformations.py
```diff
@@ -96,3 +167,119 @@ def extract_test(dataset: SupportsReadableDataset[Any]) -> Any:
...
+        "transform_without_generator": 13,
+    }
+
+    assert get_job_types(dest_p) == {"transform_without_generator": {"model": 1}}
+
+
+# @pytest.mark.skip("TODO: hints are not being merged correctly")
+
+
+# TODO: support multiple transformations in a single function
```
rudolfix: I think it is done now

tests/transformations/test_transformations.py
```diff
@@ -96,3 +167,119 @@ def extract_test(dataset: SupportsReadableDataset[Any]) -> Any:
...
+    dest_p.run(transform_without_generator(fruit_p.dataset()))
+
+    assert load_table_counts(dest_p, "transform_without_generator") == {
+        "transform_without_generator": 13,
+    }
+
+    assert get_job_types(dest_p) == {"transform_without_generator": {"model": 1}}
+
+
+# @pytest.mark.skip("TODO: hints are not being merged correctly")
```
rudolfix: to be removed?

tests/transformations/test_transformations.py
```diff
@@ -96,3 +167,119 @@ def extract_test(dataset: SupportsReadableDataset[Any]) -> Any:
...
+    ids=lambda x: x.name,
+)
+def test_multiple_transformations_in_function(
+    destination_config: DestinationTestConfiguration,
+) -> None:
+    fruit_p, dest_p = setup_transformation_pipelines(destination_config)
+    fruit_p.run(fruitshop_source())
+
+    @dlt.transformation()
+    def multiple_transformations(dataset: Dataset) -> Any:
```
rudolfix, line 262: would be cool to enumerate resource without pipeline and also to have a mixed easy and eager transformations yielded

tests/transformations/test_transformations.py
```diff
@@ -96,3 +167,119 @@ def extract_test(dataset: SupportsReadableDataset[Any]) -> Any:
...
+
+    @dlt.transformation()
+    def materializable_sql_model(dataset: Dataset) -> Any:
+        purchases = dataset.table("purchases", table_type="ibis")
+        customers = dataset.table("customers", table_type="ibis")
+        yield purchases.join(customers, purchases.customer_id == customers.id)[
+            ["id", "customer_id", "inventory_id", "quantity", "name"]
+        ]
+
+    model = list(materializable_sql_model(fruit_p.dataset()))[0]
```
rudolfix, line 212: do we have a specific test for resource iteration?

rudolfix (commented):
LGTM! just a few more questions

dlt/extract/extractors.py
```diff
@@ -211,7 +211,11 @@ def _write_to_dynamic_table(self, resource: DltResource, items: TDataItems, meta
             if table_name in self._filtered_tables:
                 continue
             # allow to cache dynamic table hints if only table name is dynamic
-            if table_name not in self._table_contracts or resource._table_has_other_dynamic_hints:
```
sh-rp, line 208: @rudolfix I had to revert his change that you made, because yielding multiple transformations with different schemas would not work without it. Relevant test is: `def test_multiple_transformations_in_function`. Maybe you should review this once again.

tests/transformations/test_transformations.py
sh-rp: Resources are enumerated and schema access is tested in `test_extract_without_source_name_or_pipeline` and `test_materializable_sql_model` and `test_enumerate_and_access_relation`.

tests/transformations/test_transformations.py
sh-rp: I'm not sure what this means, this is tested in the line above, no?

tests/transformations/test_transformations.py
sh-rp: For the mixed type function a small fix in the normalizer was needed that prevented mixed loader_file_type tables in general

input_data
zilto, line 1: I believe this shouldn't be here?

dlt/common/destination/dataset.py
```diff
@@ -132,31 +118,27 @@ def iter_fetch(self, chunk_size: int) -> Generator[List[Tuple[Any, ...]], Any, A
...
     def fetchone(self) -> Optional[Tuple[Any, ...]]:
         """Fetches the first item as a python tuple. Uses the native dbapi fetchone implementation of the destination client cursor.
 
         Returns:
             Optional[Tuple[Any, ...]]: A python tuple with the first item of the query results.
         """
-        raise NotImplementedError("`fetchone()` method is not supported for this relation")
+
+
+class Relation(DataAccess):
```
zilto, line 132: AFAIU, `DataAccess` and `Relation` are implemented as abstract base classes (`abc.ABC`) rather than duck-typing (`Protocol`). The class `Relation` inheriting `DataAccess` protocol should implement the methods directly (otherwise, make `Relation` the protocol).

For instance, the type inspection from the IDE complains that all methods don't return anything while they indicate they do.

image

What's the practical use case here? I thought IbisRelation was removed to have a single `Dataset` and `Relation` implementation. This has the benefit of shedding this abstract machinery

dlt/extract/extractors.py
rudolfix: right! but this prevents caching of the computed dynamic table hints which is 99% of cases we handle. my take:
convert `_table_has_other_dynamic_hints` into `has_other_dynamic_hints` which for transformation resources is always True and for others it falls back to the current behavior

dlt/common/data_writers/writers.py
rudolfix: my question was: can `item.query_dialect()` NOT return a dialect in current impl? so we can remove expression after "or"

dlt/transformations/transformation.py
rudolfix: this background why we do the schema caching and when

dlt/destinations/dataset/relation.py
rudolfix: ping! can we remove it?

dlt/common/destination/dataset.py
sh-rp: We have interfaces (protocols or abstract classes) in common and implementations in the other parts. This enables us to have these interfaces as return types in places where the implementation is not known or can not be imported due to circular dependencies

sh-rp: Also wrt to the protocol vs abc: Ideally the Relation would also be a Protocol, the problem is, that I am checking variables for types in the transformations to see what the transformations are yielding, and you can't check for protocols (or you can with runtime_checkable decorator but apparently that is very slow). I also don't see any problems in my IDE, maybe we are using different linters? The CI passes.

zilto: > This enables us to have these interfaces as return types in places where the implementation is not known or can not be imported due to circular dependencies

For this purpose, we should simply use 
```python
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
  # import from other modules
```

We should gradually introduce levels of abstraction. If we have a single `dlt.Dataset` and `dlt.Relation`, the additional layers make reading and editing the code harder IMO. For instance, I still struggle with the many layers of proxying and wrapping on `Relation.df()`, which could be written in a very straightforward manner

PR #2719, opened 2025-06-05
feat(athena): apply lakeformation tags on database

### LINKED ISSUE #2683: Add support for lakeformation tags
### Feature description

When using the athena destination, I would like to be able to add a lakeformation tag to the resource. 

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

We are running dlt orchestrated by dagster on aws ecs, using the iceberg format in s3 together with athena and glue. We want to expose data using lakeformation permissions only and it would be nice to not have to tag tables after they've been ingested.

### Proposed solution

I think the tags should be common per destination and could be set as destination config as a json string

[destination.athena]
lakeformation_tags="[{"tag1": "value1", "tag2": "value2"}]"

### Related issues

N/A

hagelborn:
### Description

AWS lakeformation can be used to manage permissions to data. This introduced an optional lakeformation config for the athena destination which can be used to apply a set of lakeformation tags to the glue database that the pipeline is loading to. All tables in the database will inherit the same tags. 

The implementation is heavily inspired by the dbt-athena adapter, and could be extended to allow resource specific lf tags.

The test requires running with iam principals that are lake formation admin, and that lakeformation has been activated in the aws account.

### Related Issues

- Resolves #2683


dlt/destinations/impl/athena/athena.py
```diff
@@ -19,7 +20,16 @@
 from contextlib import contextmanager
 from pendulum.datetime import DateTime, Date
 from datetime import datetime  # noqa: I251
-
+from threading import Lock
+
+import boto3
+from mypy_boto3_lakeformation import LakeFormationClient
```
rudolfix: those are needed to be present only when type checking. so if you could import them conditionally:
```py
if TYPE_CHECKING:
    ...
else:
     LakeFormationClient = Any
```
then we do not need to dstribute stubs via PyPI. they are not needed at runtime (I hope)

dlt/destinations/impl/athena/athena.py
```diff
@@ -431,6 +521,11 @@ def _get_table_update_sql(
                         LOCATION '{location}';""")
         return sql
 
+    def complete_load(self, load_id: str) -> None:
+        super().complete_load(load_id)
+        if self.config.lakeformation_config is not None:
```
rudolfix, line 543: FYI: in that case if `manage_lf_tags` fails, it won't be retried because the load package is completed above. Is adding those tags "best effort" operation or should be considered as part of "transaction" which is loading of the whole package?

pyproject.toml
```diff
@@ -83,6 +83,7 @@ pipdeptree = {version = ">=2.9.0,<2.10", optional = true}
 # pip is used by pipdeptree but not listed in its dependencies
 pip = {version = ">=23.0.0", optional = true}
 pyathena = {version = ">=2.9.6", optional = true}
+boto3-stubs = {extras = ["lakeformation"], version = "^1.38.28", optional = true}
```
rudolfix: if we use stubs only for the type checking you should move them to dev dependency group and remove from the extras

tests/load/athena_iceberg/test_lakeformation.py
```diff
@@ -0,0 +1,295 @@
...
+from dlt.common.utils import uniq_id
+from dlt.destinations.impl.athena.configuration import LakeformationConfig
+
+IAM_ARN = boto3.client("sts").get_caller_identity()["Arn"]
+ATHENA_WORKGROUP = dlt.config.get("tests.athena_workgroup", str)
+QUERY_RESULT_BUCKET = dlt.config.get("tests.query_results_bucket", str)
+
+# These tests require Lakeformation permissions to be able to run.
+# The tests have only been run as lakeformation admin so far, since the fixtures require
+# permissions to create new tags, and grant lakeformation permissions to the current role
```
rudolfix, line 22: yeah we'll need to change the permissions of our test user or use a separate one here. we'll figure that out

rudolfix (changes requested):
@hagelborn LGTM! and thanks for the tests. I think we could avoid distributing stubs with dlt package. pls. see review.

also we did some changes to lockfile recently. so it conflicts with your PR. you'll need to fix it.

thanks again

dlt/destinations/impl/athena/athena.py
hagelborn: I think a best effort approach is most reasonable here, as retrying probably wouldn't be able to achieve anything. It would only fail if lakeformation permissions don't exist.

That said maybe catching any exception during `manage_lf_tags` would make sense so the pipeline ends successfully?

hagelborn:
@rudolfix Thanks for your comments! I've updated and put the boto stubs in the dev group instead and made the type checking optional.

I rebased my branch on `devel` and relocked poetry.lock. When running the tests with `make test-common` suddenly some unrelated tests failed. The same tests fail for me on the `devel` branch. Is something broken upstream? I can see the tests are failing in CI too. Let me know if there's anything else I can do!

rudolfix:
@hagelborn we've again changed our ci and dependency system (we use `uv` now). I merged devel into your branch and added a few fixes to the tests. please see #2808 

you can rebase/merge it into this branch if you want this branch to be merged. we can also continue on the new branch. ping me what you pick

all tests are passing now and code itself is fine!

hagelborn:
@rudolfix Thanks for picking it up and I'm happy to see `uv` finally! :D 
We can continue on your branch and close this PR. The only remaining thing I can think of is making sure that the lakeformation specific tests work in your test pipeline, which would require setting up the IAM role to run the test.

rudolfix:
> @rudolfix Thanks for picking it up and I'm happy to see `uv` finally! :D We can continue on your branch and close this PR. The only remaining thing I can think of is making sure that the lakeformation specific tests work in your test pipeline, which would require setting up the IAM role to run the test.

this actually works, we've added permissions to lakeformation and a few tweaks to your test so they can be selected by a workflow for particular destination
https://github.com/dlt-hub/dlt/actions/runs/15862899932/job/44738233698?pr=2808#step:10:45

PR #2720, opened 2025-06-05
QoL: Friendly drop command 🐕‍🦺 

### LINKED ISSUE #1195: Make a drop command more user friendly
The current drop command requires additional credentials setting. We want to make it more user friendly: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1712059067366979

We have another report from the community that it is not clear which database the drop command is actually accessing, it is in this thread here: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1714050325808869. From my perspective the bottom line should be, that all cli commands that modify the destination in some way should print a message stating that that the credentials are not taken from the code (but rather toml file) and then also print the connection string and dataset to this destination, without the credentials, for the user to review.

anuunchin:
### Description
This PR aims to make the drop cli command friendly with appropriate warnings. 

- [x] Warn the user that they env vars or config files are used and inform that they should run the cli command from the correct location.
- [x] Warn the user that they are in the wrong repository when credentials can't be found. Currently the useful message is only given when a pipeline script is explicitly ran, but not when cli commands are used.
- [x] Warn the user about the tables that are to be dropped - with the location, name and any other relevant info. (I think this already implemented by @sh-rp 👀 )
- [x] When dlt-plus is installed, the drop command doesn't work with error `DropCommand.__init__() got an unexpected keyword argument 'project'`.

### Related Issues

- Fixes #1195


dlt/common/configuration/exceptions.py
```diff
@@ -78,8 +78,27 @@ def __str__(self) -> str:
...
 
-        # check if entry point is run with path. this is common problem so warn the user
         main_path = main_module_file_path()
+        # check if entry point is from cli command
+        if main_path and main_path.endswith("dlt"):
+            from dlt.common.runtime import run_context
+            import dlt
+
+            settings = run_context.active().settings_dir
+            abs_main_dir = dlt.current.pipeline().state["_local"]["initial_cwd"]
```
rudolfix: this is good but we can make it way better. `initial_cwd` is not 100% reliable and we have `run_context`. I'd do the following: whenever pipeline step is completed in `Pipeline` class we store run_context info into the local state (you have a method to set such state) ie.
```py
{"last_run_context": "settings_dir": run_context.active().settings_dir, "local_dir": ... }
```

then here we can compare current settings dir and last run settings dir and show this warning if they are really different

best place place to do that is in extract/normalize/load/sync_destination just before function returns.
we do that here because we have a writable state at that point so we do not need to load/save it again

rudolfix: `dlt.current.pipeline()` is not always available. if it is not - skip all the checks

rudolfix: you have two pipeline commands to read/write local state:
`get_local_state_val`
`set_local_state_val`

dlt/common/configuration/exceptions.py
```diff
@@ -78,8 +78,27 @@ def __str__(self) -> str:
                     " exist or may be empty.\n"
                 )
 
-        # check if entry point is run with path. this is common problem so warn the user
         main_path = main_module_file_path()
+        # check if entry point is from cli command
+        if main_path and main_path.endswith("dlt"):
```
rudolfix: we should run this check in any case

dlt/cli/pipeline_command.py
```diff
@@ -348,6 +349,16 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
                     fmt.warning(warning)
             return
 
+        fmt.echo(
```
rudolfix: mhmmm maybe we should show this thing only when setting dirs differ? so exactly the same condition we have in config exeception?

dlt/pipeline/pipeline.py
```diff
@@ -1053,6 +1059,21 @@ def get_local_state_val(self, key: str) -> Any:
             state = self._get_state()
         return state["_local"][key]  # type: ignore
 
+    def _update_last_run_context(self) -> None:
+        """
+        Persist the directory context in local state.
+        Safe-no-op if the pipeline is not active or if run_context cannot be resolved.
+        """
+        try:
+            ctx = dlt.current.run_context()
```
rudolfix: this must be always available. I plan to read this when we initialize the pipeline (in dlt.pipeline) and pass it to the class... tldr;> you do not need exception handler

dlt/pipeline/pipeline.py
```diff
@@ -1053,6 +1059,21 @@ def get_local_state_val(self, key: str) -> Any:
...
+        except Exception:
+            return
+
+        self.set_local_state_val(
+            "last_run_context",
+            {"settings_dir": ctx.settings_dir, "local_dir": os.path.abspath(ctx.local_dir)},
```
rudolfix: please also store `run_dir` - this is `cwd` in OSS, we'll need it to raise warnings correctly (in dlt+ it is not `cwd`)

dlt/pipeline/pipeline.py
```diff
@@ -595,6 +597,7 @@ def load(
             info: LoadInfo = self._get_step_info(load_step)
 
             self.first_run = False
+            self._update_last_run_context()
```
rudolfix, line 600: now when I look at all the changes I think there are only two places we should update last run context:
- here
- in `_sync_destination` where we set `first_run` to `False`

Also let's handle `last_run_context` in the same way as `first_run`. Pipeline class has a mechanism that automatically saves and restores state from fields on the class, 
you can check how `first_run` is implemented and add `last_run_context` in the same way:
- define new typed dict to store `last_run_context` and add it to `TPipelineLocalState`
- also add it to Pipeline as a field. (like `first_run`).
- `_update_last_run_context` should just set property on the pipeline
- `_state_to_props` and `_props_to_state` will automatically save and restore it
tbh. I forgot about this mechanism.

dlt/common/configuration/exceptions.py
```diff
@@ -73,29 +73,34 @@ def __str__(self) -> str:
                     " missing or empty.\n"
                 )
 
-        # check if entry point is run with path. this is common problem so warn the user
         main_path = main_module_file_path()
-        if main_path and main_path.endswith(".py"):
+
+        # check if entry point is run with path or from cli command. This is common problem so warn the user
+        if main_path and (main_path.endswith(".py") or main_path.endswith("dlt")):
```
rudolfix: We have two cases here to detect
1. pipeline context exists and we can get `last_run_context` from it. then we warn the user if settings_dir is different in pipeline than in current run context (it should be always available). the error message is the same as you wrote in the drop command.
2. else: the check that was already there. it was mostly warning if user wrote a pipeline, saved it into folder but then run it from a different cwd (for a first time)

dlt/pipeline/pipeline.py
rudolfix: make sure you store all paths as abspath

dlt/common/configuration/exceptions.py
```diff
@@ -73,29 +73,34 @@ def __str__(self) -> str:
...
             from dlt.common.runtime import run_context
+            import dlt
+
+            last_ctx = dlt.current.pipeline().get_local_state_val("last_run_context")
+            last_ctx_local_dir = last_ctx.get("local_dir")
```
rudolfix: mind that `last_ctx` may be None here! (if pipeline never fully ran). with changes I proposed below you'll be able to get this as a pipeline instance field. still it may be `None`

dlt/cli/pipeline_command.py
```diff
@@ -359,6 +361,22 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
                     fmt.warning(warning)
             return
 
+        if p.get_local_state_val("last_run_context").get("local_dir") != os.getcwd():
```
rudolfix: this good! but it is IMO safer to compare pipeline `settings_dir` with current run context `settings_dir` (abs paths!). see my comment for the configuration exception.

or maybe I miss something here? `local_dir` is indeed the same as cwd in OSS and in OSS those are equivalent with the check above. but IMO we are warning if settings location changed

rudolfix: also `p.get_local_state_val("last_run_context")` may be none.

rudolfix (changes requested):
we can simplify the pipeline code - there's a mechanism in there that saves and restores state into pipeline props automatically. ping if something is not clear in the review

dlt/pipeline/pipeline.py
anuunchin: Would having the `_update_last_run_context` only in load and _sync_destination mean that the last run context will become unreliable if a (power) user runs the load, normalize and extract parts separately? 👀

dlt/pipeline/pipeline.py
anuunchin: I think `run_dir` would be redundant 👀  because there's `local_dir` already? Or am I missing something

dlt/cli/pipeline_command.py
anuunchin: Should `settings_dir` not stay relative because we have `local_dir` which is absolute? 👀

anuunchin: Also, I don't get how the settings location would change, considering we have the pipeline dir and the .dlt folder will always be on the same level with it? 🤔

dlt/pipeline/pipeline.py
rudolfix: this is the best approximation to get "correct" run context of the pipeline. we do not want to store any run content but only those associated with successful pipeline runs. From that POV it is additional information to `first_run` flag. And in all other checks/warning you do, you assume that pipeline context is correct.

rudolfix: I do not want to extend the scope of this ticket anymore, but what we should really do is to store run context when Pipeline object is created and write the local state from it, not from the current(). We also should activate pipeline context when pipeline is activated (`activate()`). This may be pretty important if we run several pipelines in parallel.

rudolfix (commented):
state handling code is good! also pls see answers in review

rudolfix:
@anuunchin please look at the changes. what else we need:
- test for local run context will fail (I added a few more props). it needs to be fixed
- I extracted the method that generates warning messages in configuration exception so it can be (+-) unit tested.

you can emulate running script vs. running with command line.
A. if we run as a cli, 3rd warning should be always present
B. if we run from the same run_dir (cwd) as pipeline script, no warning should be visible
C. if we run from other dir: warning depends if the pipeline previously ran or not

pls try to test B and C

PR #2728, opened 2025-06-06
Update github workflow setup

sh-rp:
### Description
We need to be more economical with our CI minutes. Proposed changes are:

- [x] Only run common tests tests if linter passes
- [x] Only run destination tests if common tests pass
- [x] Try to convert remote and possible local destinations into a matrix job so there is less margin of error

TODO before finalizing

- [x] * Remove -m "smoke" flag which was only added to speed up the test process
- [x] * Re-enable commented out tests in common
- [x] * Figure out correct concurrency settings.

Review checklist
- Are all destinations tested and are the correct tests selected in the output?
- Is the order of execution of the jobs correct? (first lint, then common tests, then all the other stuff)
- Is the setup clear on first read? If not, explain what is confusing so I we can add comments


.github/workflows/test_destinations_local.yml
```diff
@@ -4,40 +4,76 @@
...
+            post_install_commands: "poetry run pip install sqlalchemy==2.0.18" # minimum version required by `pyiceberg`
+            additional_tests: "poetry run pytest tests/cli"
+
+          - name: postgres, duckdb
+            destinations: "[\"postgres\", \"duckdb\"]"
+            filesystem_drivers: "[\"memory\", \"file\"]"
+            extras: "postgres postgis parquet duckdb cli filesystem"
+            needs_postgres: true
+
+          # Clickhouse OSS (TODO: test with minio s3)
```
sh-rp, line 44: clickhouse oss is disabled for now, it would be good to make it work without any remote services, for example test it with s3 using minio

.github/workflows/test_destinations_remote.yml
```diff
@@ -0,0 +1,175 @@
...
+              destinations: "[\"synapse\"]"
+              filesystem_drivers: "[\"memory\"]"
+              extras: "synapse parquet"
+              pre_install_commands: "sudo ACCEPT_EULA=Y apt-get install --yes msodbcsql18"
+
+            # Postgres and Redshift (used to be test_destinations.yml)
+            - name: postgres and redshift
+              # TODO: check wether duckdb and all filesystem drivers need to be here, if so, explain why
+              destinations: "[\"redshift\", \"postgres\", \"duckdb\", \"dummy\"]"
+              filesystem_drivers: "[\"memory\", \"file\", \"r2\", \"s3\", \"gs\", \"az\", \"abfss\", \"gdrive\"]" # excludes sftp
```
sh-rp: I don't think we need all these drivers here, I took this from `test_destinations.yml`, so it is a pretty ancient part of this setup and I was not sure if there are any combinations of the above destinations that need these bucket settings, but afaik `filesystem_drivers` only applies to filesystem tests, right?

.github/workflows/test_destinations_remote.yml
```diff
@@ -0,0 +1,175 @@
...
+              extras: "postgres redshift postgis s3 gs az parquet duckdb"
+              with: ",adbc"
+
+            # Qdrant (disabled, TODO: explain why)
```
sh-rp: this workflow currently is disabled in the repo. I'd like to add a note here why exactly (because I don't know)

.github/workflows/main.yml
```diff
@@ -0,0 +1,102 @@
+name: dlt
+
+on:
+  pull_request:
+    branches: [ master, devel ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
rudolfix, line 8: so this workflow will cancel all the workflows it started?

.github/workflows/main.yml
```diff
@@ -0,0 +1,102 @@
...
+  # lint will check this and if lint is not run, no other tests will be run
+  get_docs_changes:
+    name: docs changes
+    uses: ./.github/workflows/get_docs_changes.yml
+
+  # if the PR is from a fork, we need to authorize the secrets access and remote destinations run with a lable
+  authorize_run_from_fork:
+    name: check if fork and if so wether secrets are available
+    # run when label is assigned OR when we are not a fork
+    if: ${{ github.event.label.name == 'ci from fork' || (github.event.pull_request.head.repo.full_name == github.repository && (github.event.action == 'opened' || github.event.action == 'synchronize'))}}
```
rudolfix, line 23: now the `pull_request_target` is deactivated so this authorization is defunct (github.event.pull_request.head.repo.full_name == github.repository as always true)

we can enable PRs from forks in a separate ticket

.github/workflows/main.yml
```diff
@@ -0,0 +1,102 @@
...
+  test_common:
+    name: run common tests on all python versions and OSes
+    needs: lint
+    uses: ./.github/workflows/test_common.yml
+
+  #
+  # Destination and Sources local tests, do not provide secrets
+  # Other tests that do not require remote connections
+  #
+  test_destinations_local:
```
rudolfix, line 43: those run in parallel right?

.github/workflows/main.yml
```diff
@@ -0,0 +1,102 @@
...
+    name: run local destination tests without secrets
+    needs: test_common
+    uses: ./.github/workflows/test_destinations_local.yml
+
+  test_sources_local:
+    name: run local source tests without secrets
+    needs: test_common
+    uses: ./.github/workflows/test_sources_local.yml
+
+  test_plus:
```
rudolfix, line 53: maybe those 3 workflows should only depend on lint?

.github/workflows/test_destinations_local.yml
rudolfix: hmmmm why? it is only using local filesystem. I'm running tests locally on it with network disconnected. we have compose file for it:
```
services:
  clickhouse:
    image: clickhouse/clickhouse-server
    ports:
      - "9000:9000"
      - "8123:8123"
    environment:
      - CLICKHOUSE_DB=dlt_data
      - CLICKHOUSE_USER=loader
      - CLICKHOUSE_PASSWORD=loader
      - CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT=1
    volumes:
      - clickhouse_data:/var/lib/clickhouse/
      - clickhouse_logs:/var/log/clickhouse-server/
    restart: unless-stopped
    healthcheck:
      test: [ "CMD", "wget", "--no-verbose", "--tries=1", "--spider", "http://localhost:8123/ping" ]
      interval: 3s
      timeout: 5s
      retries: 5


volumes:
  clickhouse_data:
  clickhouse_logs:
```

.github/workflows/test_destinations_local.yml
```diff
@@ -67,13 +104,60 @@ jobs:
...
+        if: ${{ matrix.needs_ftp }}
+        
+      - name: Start dremio
+        run: docker compose -f "tests/load/dremio/docker-compose.yml" up -d
+        if: ${{ matrix.needs_dremio }}
+
+      - run: |
+          docker compose -f "tests/load/clickhouse/clickhouse-compose.yml" up -d
+          echo "Waiting for ClickHouse to be healthy..."
+          timeout 30s bash -c 'until docker compose -f "tests/load/clickhouse/clickhouse-compose.yml" ps | grep -q "healthy"; do sleep 1; done'
```
rudolfix: btw. you start clickhouse here so it works locally. or not?

.github/workflows/test_destinations_local.yml
```diff
@@ -4,40 +4,77 @@
...
-  # NOTE: this workflow can't use github secrets!
-  # DLT_SECRETS_TOML: ${{ secrets.DLT_SECRETS_TOML }}
 
   # RUNTIME__SENTRY_DSN: https://6f6f7b6f8e0f458a89be4187603b55fe@o1061158.ingest.sentry.io/4504819859914752
   RUNTIME__LOG_LEVEL: ERROR
   RUNTIME__DLTHUB_TELEMETRY_ENDPOINT: ${{ secrets.RUNTIME__DLTHUB_TELEMETRY_ENDPOINT }}
-  ACTIVE_DESTINATIONS: "[\"duckdb\", \"postgres\", \"filesystem\", \"weaviate\", \"qdrant\"]"
-  ALL_FILESYSTEM_DRIVERS: "[\"memory\", \"file\", \"sftp\"]"
-
-  DESTINATION__WEAVIATE__VECTORIZER: text2vec-contextionary
```
rudolfix, line 27: this is not needed anymore?

.github/workflows/test_destinations_local.yml
```diff
@@ -4,40 +4,77 @@
...
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
-
 env:
-  # NOTE: this workflow can't use github secrets!
-  # DLT_SECRETS_TOML: ${{ secrets.DLT_SECRETS_TOML }}
 
   # RUNTIME__SENTRY_DSN: https://6f6f7b6f8e0f458a89be4187603b55fe@o1061158.ingest.sentry.io/4504819859914752
   RUNTIME__LOG_LEVEL: ERROR
   RUNTIME__DLTHUB_TELEMETRY_ENDPOINT: ${{ secrets.RUNTIME__DLTHUB_TELEMETRY_ENDPOINT }}
```
rudolfix: this won't be available on forks. we have this endpoint hardcoded in `conftest`. so we can remove it everywhere I think

.github/workflows/test_destinations_remote.yml
```diff
@@ -0,0 +1,176 @@
...
+            # Filesystem
+            - name: filesystem
+              destinations: "[\"filesystem\", \"dummy\"]"
+              # note that all buckets are enabled for testing
+              filesystem_drivers: "[\"memory\", \"file\", \"r2\", \"s3\", \"gs\", \"az\", \"abfss\", \"gdrive\"]" # excludes sftp which is run in local tests
+              extras: "gs s3 az parquet duckdb filesystem deltalake pyiceberg"
+              post_install_commands: "poetry run pip install sqlalchemy==2.0.18" # minimum version required by `pyiceberg`
+
+            # LanceDB
+            - name: lancedb
```
rudolfix, line 77: lance is local I think, you can move it to local pipelines

.github/workflows/test_destinations_remote.yml
rudolfix: those are left there after I extracted filesystem to a separate workflow. I think you can leave memory and file. let's see what happens

.github/workflows/test_destinations_local.yml
```diff
@@ -4,40 +4,77 @@
...
+            destinations: "[\"filesystem\", \"weaviate\", \"qdrant\"]"
+            filesystem_drivers: "[\"memory\", \"file\", \"sftp\"]"
+            extras: "parquet cli filesystem qdrant weaviate deltalake pyiceberg sftp"
+            needs_weaviate: true
+            needs_qdrant: true
+            needs_ftp: true
+            post_install_commands: "poetry run pip install sqlalchemy==2.0.18" # minimum version required by `pyiceberg`
+            additional_tests: "poetry run pytest tests/cli"
+
+          - name: postgres, duckdb
```
rudolfix: we were also running `dummy` destination

.github/workflows/test_destinations_remote.yml
```diff
@@ -0,0 +1,176 @@
...
+              extras: "lancedb parquet"
+              post_install_commands: "poetry run pip install openai"
+
+            # Motherduck
+            - name: motherduck
+              destinations: "[\"motherduck\"]"
+              filesystem_drivers: "[\"memory\"]"
+              extras: "motherduck s3 gs az parquet"
+
+            # MSSQL
```
rudolfix, line 89: mssql was always running a full suite. this one is fast. we should do the same for remote postgres

.github/workflows/test_destinations_remote.yml
rudolfix: I think we lost our test account. we only test local version now

.github/workflows/test_destinations_local.yml
```diff
@@ -88,45 +172,40 @@ jobs:
...
+        if: ${{ matrix.post_install_commands }}
 
       # always run full suite, also on branches
       - name: Run tests Linux
         run: |
           eval "$(ssh-agent -s)"
-          poetry run pytest tests/load --ignore tests/load/sources
-          poetry run pytest tests/cli
-        env:
-          DESTINATION__POSTGRES__CREDENTIALS: postgresql://loader:loader@localhost:5432/dlt_data
```
rudolfix, line 120: where did those go?

.github/workflows/main.yml
sh-rp: yes it will, it's pretty cool!

.github/workflows/main.yml
sh-rp: yes, all tests that have their conditions met start immediately, so you have a bunch of tests running in parallel which also include matrixes.

.github/workflows/test_destinations_local.yml
sh-rp: it has moved to the example secrets file: `tests/.dlt/dev.secrets.toml` and this file is then copied to secrets.toml. this is also very useful for local development

.github/workflows/test_destinations_local.yml
sh-rp: to this new file, see comment above. there is one step in the workflow that copies these back in.

.github/workflows/test_destinations_local.yml
sh-rp: I think it uses the local clickhouse but depends on s3, az etc. are you sure you can run the full test suite without network connection? I will uncomment it again and have another look tomorrow, but I think it will not work without secrets.

.github/workflows/main.yml
sh-rp: I changed it

.github/workflows/test_destinations_remote.yml
sh-rp: the tests need openai credentials probably for some embeddings or something like that?

.github/workflows/test_destinations_local.yml
sh-rp: not in the local tests, but I moved it here now from the remote tests where it also was before (in test_destinations.yml)

.github/workflows/test_destinations_local.yml
sh-rp: https://github.com/dlt-hub/dlt/actions/runs/15567110480/job/43836196990?pr=2728
you can see here that it uses s3 etc for those tests. we could maybe just run jsonl and parquet locally..

PR #2729, opened 2025-06-06
added constants for `load_id` col in `_dlt_loads` table

zilto:
### Description
The column `load_id` in the table `_dlt_loads` matches the column `_dlt_load_id` in data tables. However, this value wasn't explicitly encoded as a constant. This PR makes this more obvious and robust.

Motivation: I wanted to set a constant because work on transformations will frequently require joining columns `load_id` and `_dlt_load_id`.


rudolfix (approved):
LGTM!

PR #2730, opened 2025-06-07
fixes leaking datasets tests

rudolfix:
### Description
1. stores references to all active pipelines (on demand) and drops them after each test
2. properly mocks `local_dir` in tests so local files are automatically in `_storage` (no more *duckdb) databases after tests
3. removed concurrent access to resolved config traces (parallel tests were sometimes failing)
4. improves the code that opens duckdb and moterduck connection (opened connections were leaked when setting configs on connection failed etc.) 
5. finally added a full set of settings for duckdb connection, added a way to move connection to ibis with settings applied

more in commits


dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -92,14 +94,12 @@ def open_connection(self) -> duckdb.DuckDBPyConnection:
         if first_connection:
             # TODO: we need to frontload the httpfs extension for abfss for some reason
             if self.is_abfss:
-                self._conn.sql("INSTALL https; LOAD httpfs;")
+                self._conn.sql("INSTALL httpfs; LOAD httpfs;")
```
sh-rp, line 96: oh :)

tests/utils.py
```diff
@@ -167,15 +167,50 @@ def test_storage() -> FileStorage:
 
 
 @pytest.fixture(autouse=True)
-def autouse_test_storage() -> FileStorage:
-    return clean_test_storage()
+def autouse_test_storage(request) -> Optional[FileStorage]:
```
sh-rp, line 172: this is maybe out of scope, but all of these fixtures need good names and docstrings and should be added to the docstring linter, I often look up what they do exactly in code if I need to disable something for testing. This one is a good example, if you never looked it up the name `autouse_test_storage` gives no real hint about what it does.

tests/load/utils.py
```diff
@@ -740,41 +739,52 @@ def destinations_configs(
...
+    Container()[PipelineContext].enable_activation_history = True
     yield
-    if "no_load" in request.keywords:
-        return
     try:
-        drop_active_pipeline_data()
+        if "no_load" in request.keywords:
+            # always deactivate
+            Container()[PipelineContext].deactivate()
+            Container()[PipelineContext].clear_activation_history()
```
sh-rp, line 753: hm, maybe we want to keep the activation history here so we can clean up with `drop_active_pipeline_data` after a bunch of "no_load" tests have run without having to store the pipelines?

dlt/common/storages/configuration.py
```diff
@@ -324,3 +326,119 @@ def make_file_url(local_path: str) -> str:
         netloc is never set. UNC paths are represented as file://host/path
         """
         return make_fsspec_url("file", local_path, None)
+
+
+@dataclass
+class WithLocalFiles(BaseConfiguration):
+    """Mixin to BaseConfiguration that shifts relative locations into `local_dir` and allows for a few special locations.
+    :pipeline: in the pipeline working folder
```
sh-rp, line 334: I know this was already like this, but why not "pipeline_working_dir"? This way it is clear what this variable means.

dlt/destinations/impl/filesystem/sql_client.py
rudolfix: they were aliases it seems!

tests/utils.py
rudolfix: right! we'll dedicate a week to refactor our utils and release them as OSS lib. version in dlt+ has nice docstrings btw.

tests/load/utils.py
rudolfix: I added this because most of `no_load` tests produce empty or fake pipelines ie. with destinations that are not instantiated. so they were leaking into subsequent tests and breaking them on cleanup. I'll keep it util we have a good case

dlt/common/storages/configuration.py
rudolfix: hmmm you can add QoL ticket for that. I like :pipeline: because it is like :memory:

PR #2741, opened 2025-06-11
add marimo flag to show command

sh-rp:
### Description
We removed the studio command, but now it is really hard to start the marimo preview app, since the notebook file is somewhere in the package. This PR adds a `marimo` flag to the `pipeline show` command to make it easier to access this preview.


rudolfix (approved):
LGTM!

PR #2744, opened 2025-06-11
fix: text and binary precision via sqlglot

zilto:
## Description
The SQLGlot libs module had a remaining breakpoint() and didn't support dlt data types text and binary. This is now fixed and implemented

Previous issue got mangled #2713


rudolfix (approved):
LGTM indeed!

PR #2748, opened 2025-06-11
feat: SQLGlot for DDL parsing

zilto:
## Description
This maps dlt hints to SQLGlot DDL expressions. This internal feature could enable future usage:

### Table discovery
To enable `@dlt.transformation`, it's useful to know about tables manages outside of dlt.

It would make sense to discover from the `dlt.Dataset` API to add tables to its schema. Discovery should be limited to the same "database/dataset" namespace on the destination. Retrieving DDL from destination is a very cheap operation.

We would need a new hint for external tables.

tentative API.
```python
dataset = dlt.Dataset(...)
table_foo = dataset.table("foo")  # regular access
table_bar = dataset.table("bar")  # raise TableNotExist

dataset.discover()  # all tables within "database/dataset" namespace
dataset.discover("bar")  # single specific table
dataset.discover(["bar", "baz"])  # multiple specific tables 
dataset.discover(pattern="ba.*")  # unknown number using regex

table_bar = dataset.table("bar")  # succeeds because table is now in schema
```
This effect is persisted to the schema storage and the destination.

### Table creation
We have a lot of templated SQL strings for destination clients. We also use destination-specific SDK libraries.

Generating DDL from dlt hints could simplify internals and help with maintenance. This potentially simplifies supporting and testing new SQL backends.

More importantly, `dlt -> DDL` allows more powerful user configuration (e.g., create an index).

### DIFF
--- a/dlt/common/libs/sqlglot.py
+++ b/dlt/common/libs/sqlglot.py
@@ -1,12 +1,27 @@
-from typing import Optional
+from typing import Optional, Union
 
+from dlt.common.schema import Schema
+from dlt.common.schema.utils import apply_defaults
 from dlt.common.utils import without_none
 from dlt.common.exceptions import MissingDependencyException, TerminalValueError
-from dlt.common.schema.typing import TColumnType, TDataType, TColumnSchema
+from dlt.common.schema.typing import (
+    COLUMN_PROPS,
+    ColumnPropInfos,
+    TColumnType,
+    TDataType,
+    TColumnSchema,
+    TStoredSchema,
+    TTableSchemaColumns,
+    TTableSchema,
+    TTableReference,
+)
 
 try:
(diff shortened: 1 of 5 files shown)


zilto:
This is an improvement over SQLAlchemy reflection because it doesn't require the existence of an SQLAlchemy driver for the destination

rudolfix:
@zilto I'll close this PR for now and leave the branch. the amount of conflicts is probably impossible to resovle

zilto:
To be clear, I think this work is super relevant for lineage and DAG resolution (which we discussed separately). I never updated it because it was never reviewed (chicken & egg problem).

AFAIU, it was never reviewed because it was assumed to be a duplicate of `sqlalchemy` reflection.

PR #2749, opened 2025-06-11
docs: plus: add merge strategies for dlt+ Iceberg destination

burnash:


rudolfix (changes requested):
- I'd mention that we are not using `pyiceberg` for upsert but we implement our own algorithm which deletes and inserts data.
- I'd mention that we stress tested both options up to tens of millions of rows and we didn't observe any problems. memory and time increase linearly with the updated dataset size
- you do not mention that we cannot detect orphaned nested tables with `upsert`. you can find more info on that in OSS delta (hopefully)
- so we support merge keys on `upsert`? that would be good

docs/website/docs/plus/ecosystem/iceberg.md
```diff
@@ -314,6 +314,47 @@ For more information, refer to the [Lakekeeper section above](#lakekeeper-catalo
 
 All [write dispositions](../../general-usage/incremental-loading.md) are supported.
 
+## Merge strategies
```
VioletM, line 317: It's a bit confusing why for the Iceberg destination, we repeat merge strategies info. Are they different than for any other destination? Is it different for only the `upsert`? 
Also an explanation on how we should use the merge key.

So overall I would add something like: 

```
## Write dispositions

All [write dispositions](../../general-usage/incremental-loading.md) are supported. 

### Merge strategies

As any other destination, Iceberg destination supports two merge strategies when using `write_disposition="merge"`: 
* `delete-insert`
* `upsert`

But the implementation is slightly different for the Iceberg destination.

### Delete-insert strategy

The `delete-insert` strategy first deletes existing rows matching the key columns, then inserts new data. The `delete-insert` strategy loads data to a staging dataset, deduplicates the staging data if a `primary_key` is provided, deletes the data from the destination using `merge_key` and `primary_key`, and then inserts the new records. 

```py
@dlt.resource(
    write_disposition={"disposition": "merge", "strategy": "delete-insert"},
    primary_key="id",
    table_format="iceberg"
)
def my_resource():
    yield [{"id": 1, "name": "Alice"}, {"id": 2, "name": "Bob"}]

#### Key characteristics
- Requires primary key or merge key columns
- Supports only regular tables (nested tables not supported)
- Automatically deduplicates data based on primary keys
- Uses single Iceberg transaction for delete and insert operations

### Upsert strategy

The `upsert` strategy performs is similar to delete-insert strategy with two differences:
- It supports nested tables
- It does not perform automatic deduplication

```py
@dlt.resource(
    write_disposition={"disposition": "merge", "strategy": "upsert"},
    primary_key="id",
    table_format="iceberg"
)
def my_upsert_resource():
    yield [{"id": 1, "name": "Alice Updated"}, {"id": 3, "name": "Charlie"}]

```

docs/website/docs/plus/ecosystem/iceberg.md
```diff
@@ -314,6 +314,47 @@ For more information, refer to the [Lakekeeper section above](#lakekeeper-catalo
...
+- Requires primary key or merge key columns
+- Supports only regular tables (nested tables not supported)
+- Automatically deduplicates data based on primary keys
+- Uses single Iceberg transaction for delete and insert operations
+
+### Upsert strategy
+
+The `upsert` strategy performs is similar to delete-insert strategy with two differences:
+- It supports nested tables
+- It does not perform automatic deduplication
```
VioletM: Not sure what does that means. Isn't it supposed to use `primary_key` only? How then it will insert two records with the same `primary_keys`?

docs/website/docs/plus/ecosystem/iceberg.md
burnash: > It's a bit confusing why for the Iceberg destination, we repeat merge strategies info. Are they different than for any 
> other destination? 
> As any other destination, Iceberg destination supports two merge strategies when using [...]

Yes, some destination do not support some of the merge strategies, that's why we list what is supported.

docs/website/docs/plus/ecosystem/iceberg.md
burnash: Yes, it won't deduplicate by design. The two records will be inserted as is.

burnash: See https://dlthub.com/docs/general-usage/merge-loading#upsert-versus-delete-insert for the dedup

rudolfix (approved):
LGTM!

PR #2754, opened 2025-06-12
QoL: Drop column

### LINKED ISSUE #1153: Add support for dropping column in destination & dlt schema
### Feature description

Extend the cli drop command to allow the user to drop columns in the destination. Sometimes columns are created by mistake, and manually dropping them in the destination kind of works, but they will re-appear in the next load when dlt syncs the internal dlt schema with the destination schema.

### Are you a dlt user?

None

### Use case

_No response_

### Proposed solution

_No response_

### Related issues

_No response_

anuunchin:
### Description
This PR aims to enable the drop command on column level.

### Related Issues

- Resolves #1153


anuunchin:
A better solution is being implemented in #2909

PR #2755, opened 2025-06-12
Psycopg2SqlClient: accept extra options

nicob3y:
see #2656

### Description

Add ability to add Postgres runtime server config via Psycopg2SqlClient.

Example of PostgresCredentials usage with "my_setting" server setting :
```
from dlt.destinations.impl.postgres.configuration import PostgresCredentials

creds = PostgresCredentials("postgresql://user:password@host:port/db?options=-cmy_setting=value")
```

### Additional Context

I needed to transmit a parameter (such as a session option) to the PostgreSQL server, but the current implementation of Psycopg2SqlClient did not allow me to do this easily.

### DIFF
--- a/dlt/destinations/impl/postgres/sql_client.py
+++ b/dlt/destinations/impl/postgres/sql_client.py
@@ -45,9 +45,13 @@ class Psycopg2SqlClient(SqlClientBase["psycopg2.connection"], DBTransaction):
         self.credentials = credentials
 
     def open_connection(self) -> "psycopg2.connection":
+        options=f"-c search_path={self.fully_qualified_dataset_name()},public"
+        extra_options = self.credentials.query.get("options")
+        if extra_options:
+            options = f"{extra_options} {options}"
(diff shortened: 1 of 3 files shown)


nicob3y:
As it is, the "search_path" setting cannot be customized; do you think this could be useful?

rudolfix (dismissed):
LGTM!

rudolfix:
> As it is, the "search_path" setting cannot be customized; do you think this could be useful?

we add search path to the pipeline's dataset (corresponding to database schema in postgres). I think you can change it after connection? I do not expect many people wanting to change that

rudolfix:
@nicob3y looks like there's a small problem in the test: https://github.com/dlt-hub/dlt/actions/runs/15621365134/job/44797749605?pr=2755#step:14:940

nicob3y:
I think it's better now :-)

rudolfix (approved):
LGTM!

PR #2737, opened 2025-06-09
feat/explains partition and split loading

rudolfix:
### Description
This PR documents and tests two methods of splitting pipeline runs in smaller chunks that may be useful for backfilling:
- partition loading: where source data is partitioned into several ranges that may be independently loaded
- split loading: where source data is split into several sequential ranges using limit item

On top of that a few improvements are added:
- filesystem source follows `sort_order` and may be used for split loading
- `LimitItem` allows to count rows (not only yields/batches)
- several improvements to root_key propagation: (1) simplified implementation (2) allows to use parent_key if nesting level < 2 (3) does not enable root_key on scd2 (4) additional tests and docs

see commit log for more

PR #2756, opened 2025-06-12
mssql and snowflake bugfixes + bumps to 1.12.0

### LINKED ISSUE #2752: Slow merge SQL query in Snowflake
### dlt version

1.11.0

### Describe the problem

When running the dlt pipeline with the following conditions:
* Snowflake as a destination
* merge write disposition with `delete-insert` strategy
* both primary_key and merge_key are specified

dlt executes the following SQL query:
```
DELETE FROM "SCHEMA_1"."TABLE_1" as d WHERE EXISTS (SELECT 1 FROM "SCHEMA_1_STAGING"."TABLE_1" as s WHERE d."PK_1" = s."PK_1" AND d."PK_2" = s."PK_2" OR d."MERGE_KEY" = s."MERGE_KEY");
```
The `OR` operation is very(!) slow on the Snowflake side.

### Expected behavior

SQL Query optimization, for example using two statements instead of one, but keeping them in the same transaction.

```
DELETE FROM "SCHEMA_1"."TABLE_1" as d WHERE EXISTS (SELECT 1 FROM "SCHEMA_1_STAGING"."TABLE_1" as s WHERE d."PK_1" = s."PK_1" AND d."PK_2" = s."PK_2");
DELETE FROM "SCHEMA_1"."TABLE_1" as d WHERE EXISTS (SELECT 1 FROM "SCHEMA_1_STAGING"."TABLE_1" as s WHERE d."MERGE_KEY" = s."MERGE_KEY");
```

### Steps to reproduce

```
import dlt

# Get the incremental value from the target
pipeline = dlt.pipeline(
    destination=destination, 
    pipeline_name="test_pipeline", 
    dataset_name="schema_1",
    progress="log",
)
incremental = dlt.sources.incremental(
    cursor_path="incremental_key",
    primary_key=()
)
resource = dlt.sources.sql_database.sql_table(
    credentials=source,
    table="table_1",
    schema="schema_1",
    incremental=incremental,
)
resource.apply_hints(merge_key="merge_key")
resource.add_limit(1)

pipeline.run(resource, write_disposition="merge")
```

### Operating system

Linux

### Runtime environment

Docker, Docker Compose

### Python version

3.10

### dlt data source

_No response_

### dlt destination

Snowflake

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
* fixes mssql sql_client (from #2737 ) - recordset was not drained from messages
* fixes #2752 (n^2 complexity detected, query never exits on large datasets)
* adds flag to skip deduplication on `delete-insert` (to further decrease merge cost)
* fixes sqlalchemy merge
* splits `filesystem` ci run per bucket
* decreases the number of essential tests
* bumps to 1.12.0


dlt/destinations/impl/sqlalchemy/merge_job.py
```diff
@@ -120,6 +120,7 @@ def gen_merge_sql(
         )
 
         dedup_sort = get_dedup_sort_tuple(root_table)  # column_name, 'asc' | 'desc'
+        skip_dedup = root_table.get("x-stage-data-deduplicated", False)
```
sh-rp, line 123: at some point we should collect these hints in a const file and not use magic strings everywhere :)

dlt/cli/pipeline_command.py
```diff
@@ -114,9 +114,15 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
...
         if command_kwargs.get("marimo"):
+            import os
+
+            from dlt.common.utils import custom_environ
+            from dlt.common.known_env import DLT_DATA_DIR
+
             from dlt.helpers.studio.runner import run_studio
 
-            run_studio()
+            with custom_environ({DLT_DATA_DIR: os.path.dirname(p.pipelines_dir)}):
```
sh-rp, line 124: does this work if we start marimo with `subprocess.run`? I thought I had tested this and the env set in python does not get carried over to other processes.

.github/workflows/test_destinations_remote.yml
```diff
@@ -66,11 +66,25 @@ jobs:
               extras: "databricks s3 gs az parquet"
 
             # Filesystem
-            - name: filesystem
+            - name: filesystem_s3_local
               destinations: "[\"filesystem\"]"
               # note that all buckets are enabled for testing
```
sh-rp, line 71: these comments are not really up to date now, we could also mention the sftp exclusion at the top of this block

tests/load/pipeline/test_refresh_modes.py
```diff
@@ -585,9 +588,7 @@ def test_refresh_staging_dataset(destination_config: DestinationTestConfiguratio
 
 @pytest.mark.parametrize(
     "destination_config",
-    destinations_configs(
-        default_sql_configs=True, default_staging_configs=True, all_buckets_filesystem_configs=True
-    ),
+    destinations_configs(default_sql_configs=True, all_buckets_filesystem_configs=True),
     ids=lambda x: x.name,
 )
 @pytest.mark.parametrize("refresh", ["drop_source", "drop_resource", "drop_data"])
```
sh-rp, line 594: judging from the explanation of this test, it seems like we do not need it at all but have to test our internal interfaced downstream better (or maybe we don't trust those tests at all? :) )

dlt/cli/pipeline_command.py
rudolfix: hmmm AFAIK env is carried over. I have tests for that. maybe you should use our `Venv.current()` to start processes. then it is carried over.

PR #2759, opened 2025-06-13
docs: decrease docsearch pageRank for generated api docs

burnash:


burnash:
Implemented using [Record Extractor](https://docsearch.algolia.com/docs/record-extractor/#boost-search-results-with-pagerank).

PR #2765, opened 2025-06-16
open correct page in marimo app for "pipeline show" command

sh-rp:
This PR opens the correct page for the marimo app when using the "pipeline show" command.


rudolfix (approved):
LGTM!

PR #2766, opened 2025-06-16
migrate to uv

sh-rp:
### Description
This PR will migrate the project away from poetry to uv.

Notes:
* `uv lock --upgrade-package hexbytes==0.3.1` will lock a package to a certain version in the lockfile (as long as in allowed range of pyproject)

ToDo before merge:
- [x] Review old and new pyproject file properly
- [x] Fix windows tests on python 3.11 (they are disabled at the moment)
- [x] Delete `poetry.lock` file

Review Notes:
- [ ] Please check all the notes in this PR there are a few small questions
- [ ] Building the package works, but maybe review the build settings properly to be sure
- [ ] the uv.lock file now contains the newest version (as defined in the bounds of the pyproject toml) of each package, some I needed to downgrade but overall the locked versions are pretty up to date. As far as I remember this is ok and in cases were we need to test a specific version (for example sqlalchemy) this is specified in the workflow file.
- [ ] Readme and contributing files are updated
- [ ] We have two "regressions" with uv: no shell as in poetry, which probably is good, and uv version cannot bump to prerelease versions, which also is not really an issue.
- [ ] There is a poetry-deps.sh file in tools, I assume we can delete it?


.github/workflows/test_common.yml
```diff
@@ -23,33 +23,36 @@ jobs:
...
             shell: bash
           
           # windows tests
           - os: windows-latest
-            python-version: "3.11.x"
+            python-version: "3.11"
             shell: cmd
             pytest_args: '-m "not forked"'
+            # NOTE: remove this before uv migration merge
+            continue-on-error: true
```
sh-rp: NOTE: windows with python 3.11 needs to be fixed (there is some stuff with paths and on 3.13 it works for some reason?) and then this flag needs to be removed.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -287,6 +287,12 @@ def __init__(
         self.bucket_path = (
             config.make_local_path(config.bucket_url) if self.is_local_filesystem else fs_path
         )
+
+        # NOTE: we need to make checksum validation optional for boto to work with s3 compat mode
+        # https://www.beginswithdata.com/2025/05/14/aws-s3-tools-with-gcs/
+        os.environ["AWS_REQUEST_CHECKSUM_CALCULATION"] = "when_required"
```
sh-rp, line 293: you can set this directly on the boto client I think, but I am not sure how to do this. For now these flags work..

dlt/common/runners/venv.py
```diff
@@ -188,6 +188,9 @@ def install_deps(context: types.SimpleNamespace, dependencies: List[str]) -> Non
                 "--prerelease",
                 "if-necessary-or-explicit",
             ]
+            # on windows we need to copy the dependencies to the venv instead of linking
+            if os.name == "nt":
```
sh-rp, line 192: Note: review this

pyproject.toml
```diff
@@ -21,299 +26,313 @@ classifiers = [
...
+    "pyarrow>=14.0.0 ; python_version < '3.13'",
+    "pyarrow>=18.0.0 ; python_version >= '3.13'",
+    "numpy>=1.21,<2.0 ; python_version >= '3.9' and python_version < '3.12'",
+    "numpy>=1.26,<2.0 ; python_version >= '3.12' and python_version < '3.13'",
+    "numpy>=2.0.0 ; python_version >= '3.13'",
+    "pandas>2.1 ; python_version >= '3.12'",
+    "pandas<2.1 ; python_version < '3.12'",
+    # keep last duckdb version that works with azure and iceberg correctly
+    # NOTE, we need to allow duckdb 0.8 for venv tests with old dlt versions
+    "duckdb>=0.8,<1.2.1",
```
sh-rp: duckdb 0.8 is allowed here for the venv tests with old dlt versions to pass. maybe we could use a newer dlt version?

pyproject.toml
```diff
@@ -21,299 +26,313 @@ classifiers = [
...
-tenacity = ">=8.0.2"
-jsonpath-ng = ">=1.5.3"
-fsspec = ">=2022.4.0"
-packaging = ">=21.1"
-pluggy = ">=1.3.0"
-win-precise-time = {version = ">=1.4.2", markers="os_name == 'nt' and python_version < '3.13'"}
-sqlglot = ">=23.0.0"
-rich-argparse = ">=1.6.0"
-
-psycopg2-binary = {version = ">=2.9.1", optional = true}
```
sh-rp, line 61: this whole section of optional dependencies is now moved to the extras and groups, there is a section at the bottom for global constraints where I added some dependencies that where duplicates. We could move all constraints there but it is a bit of work, if you think I should do this, let me know.

sh-rp: this comment is outdated, there is no way in uv to create global dependency bounds that are respected by all extras and by all installers. uv internally the approach I had here before works, but not if you the install with poetry.

PR #2769, opened 2025-06-17
alway test docs snippets, not only after authorization

sh-rp:
### Description
This fixes the CI setup to always check the docs snippets, even if the PR is from a fork.


rudolfix (approved):
LGTM!

PR #2775, opened 2025-06-18
docs: correct question in "Choosing a write disposition" section

axelearning:
## Description
Fixed incorrect question asking in [Incremental loading](https://dlthub.com/docs/general-usage/incremental-loading).
In the section [Choosing a write disposition](https://dlthub.com/docs/general-usage/incremental-loading) the question:
```
Can I extract it incrementally from the source? 
```
should be :
```
Do you need to track history of change ?
```


sh-rp:
@axelearning, thanks for finding this!

Issue #2701, opened 2025-05-30
Marimo App ToDos


sh-rp:
A few smaller things to fix / update before the final release of the marimo app

UI / Usage improvements
- [ ] Handle incomplete columns and tables correctly

Other improvements
- [x] start marimo app in the same process instead of calling subprocess
- [x] replace dlt pipeline show command with marimo app, add flag to choose between streamlit and marimo, add console output to make users aware of this update
- [ ] Show exceptions info in trace section

Testing
- [x] unit test all helpers
- [x] make e2e tests load data, use the query history, fold out all trace views and do a basic check on the results
- [ ] optional: write basic unit tests for marimo cells

Docs and strings
NOTE: The overall goal should be that the app has good enough ui and descriptions that it is mostly self documenting.
- [ ] Add simple docs page, possibly mention the app in other parts of the docs
- [ ] Explain all table and column hints in the app

zilto:
I suggest moving `tests/e2e` elsewhere. The dependency on `pytest-playwright` is 43mb. Then it requires installing all chromium shells of size 163Mb, 101Mb, 91Mb, 93Mb. 

This adds complexity to local developer setups running unit tests and will likely slowdown CI (we could cache the dependencies, but we'll need to update from time to time). 

Also, it's likely that the marimo app is on a code path that will be updated unfrequently

sh-rp:
closing in favor of https://github.com/orgs/dlt-hub/projects/9/views/1?pane=issue&itemId=123732627&issue=dlt-hub%7Cdlt%7C2985

PR #2778, opened 2025-06-18
marimo app updates

sh-rp:
### Description

Make marimo app editable
 * when --edit flag is present, marimo app and marimo app styles are now ejected to the cwd
 * in non edit mode, the stock app is started, but if there is an ejected app in the current directory, this one gets started

Other changes
 * cli docs are updated
 * app css file was not part of package, it was added
 * app is started in process now with internal marimo api
 * Caching of queries slightly changed for better ui
 * Incomplete columns are not displayed in the table overview
 * Query in data explorer mode is now pretty printed
 * Incomplete tables are not displayed in the table overview

https://github.com/dlt-hub/dlt/issues/2701

### DIFF
--- a/dlt/cli/plugins.py
+++ b/dlt/cli/plugins.py
@@ -217,8 +217,8 @@ Requires `streamlit` to be installed in the current environment: `pip install st
             default=False,
             action="store_true",
             help=(
-                "Launch marimo app preview instead of streamlit. Will launch ejected version of app"
-                " (created with --edit flag) if it exists in the current directory."
+                "Launch marimo app instead of streamlit. Will launch editable version of app"
+                " (created with the --edit flag) if it exists in the current directory."
             ),
         )
         show_cmd.add_argument(
@@ -226,9 +226,9 @@ Requires `streamlit` to be installed in the current environment: `pip install st
             default=False,
             action="store_true",
             help=(
-                "Ejects marimo app preview to current directory (if it does not exist there yet)"
-                " and launches it in edit mode (only works together with --marimo flag, otherwise"
-                " it will be ignored)"
+                "Creates editable version of marimo app in current directory if it does not exist"
+                " there yet and launches it in edit mode. Only works when using the marimo app"
+                " (--marimo flag)."
             ),
         )
         pipeline_subparsers.add_parser(
--- a/dlt/helpers/studio/runner.py
+++ b/dlt/helpers/studio/runner.py
@@ -1,7 +1,7 @@
 import os
-import pkg_resources
+from importlib.resources import files
 from typing import Any
-
+from pathlib import Path
 from dlt.common.exceptions import MissingDependencyException
 
 # keep this, will raise if user tries to run studio without dependencies
@@ -37,13 +37,13 @@ def run_studio(pipeline_name: str = None, edit: bool = False) -> None:
     if edit and not ejected_app_exists:
         with open(dlt_app.__file__, "r", encoding="utf-8") as f:
             app_code = f.read()
-            with open(ejected_app_path, "w", encoding="utf-8") as f:
-                f.write(app_code)
-        css_file_path = pkg_resources.resource_filename("dlt", f"helpers/studio/{STYLE_FILE_NAME}")
+        with open(ejected_app_path, "w", encoding="utf-8") as f:
+            f.write(app_code)
+        css_file_path = Path(files("dlt.helpers.studio") / STYLE_FILE_NAME)  # type: ignore
         with open(css_file_path, "r", encoding="utf-8") as f:
             css_content = f.read()
-            with open(os.path.join(os.getcwd(), ejected_css_path), "w", encoding="utf-8") as f:
-                f.write(css_content)
+        with open(os.path.join(os.getcwd(), ejected_css_path), "w", encoding="utf-8") as f:
+            f.write(css_content)
         ejected_app_exists = True
 
     # set current pipeline
--- a/docs/website/docs/reference/command-line-interface.md
+++ b/docs/website/docs/reference/command-line-interface.md
@@ -206,8 +206,8 @@ Inherits arguments from [`dlt pipeline`](#dlt-pipeline).
 
 **Options**
 * `-h, --help` - Show this help message and exit
-* `--marimo` - Launch marimo app preview instead of streamlit. will launch ejected version of app (created with --edit flag) if it exists in the current directory.
-* `--edit` - Ejects marimo app preview to current directory (if it does not exist there yet) and launches it in edit mode (only works together with --marimo flag, otherwise it will be ignored)
+* `--marimo` - Launch marimo app instead of streamlit. will launch editable version of app (created with the --edit flag) if it exists in the current directory.
+* `--edit` - Creates editable version of marimo app in current directory if it does not exist there yet and launches it in edit mode. only works when using the marimo app (--marimo flag).
 
 </details>
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -316,7 +316,7 @@ include = [
     "README.md",
     "dlt/sources/pipeline_templates/.gitignore",
     "dlt/sources/pipeline_templates/.dlt/config.toml",
-    "dlt/helpers/studio/app_style.css"
+    "dlt/helpers/studio/dlt_app_styles.css"
 ]
 
 [tool.hatch.build.targets.wheel]


tests/helpers/studio/test_utils.py
```diff
@@ -1,695 +1,495 @@
 import os
```
sh-rp, line 1: Review note: The old version was generated by ai and not active (marked with pytest.skip). The new version I also generated a first pass with AI but completely reworked it. It does not completely test all cases, but the coverage is good enough for now imho.

dlt/cli/plugins.py
```diff
@@ -216,15 +216,19 @@ def configure_parser(self, pipe_cmd: argparse.ArgumentParser) -> None:
             "--marimo",
             default=False,
             action="store_true",
-            help="Launch marimo app preview instead of streamlit",
+            help=(
+                "Launch marimo app preview instead of streamlit. Will launch ejected version of app"
```
zilto: - why "launch marimo app preview"? Maybe just "launch marimo app" ?
- remove the word "eject/ejected" and align with the arg name "edit" (which matches the marimo terms). 

Revised
> Launch marimo app instead. If an editable version of the app (created with `--edit` arg) is found in the current directory, launch it instead.

dlt/cli/plugins.py
```diff
@@ -216,15 +216,19 @@ def configure_parser(self, pipe_cmd: argparse.ArgumentParser) -> None:
...
+            ),
         )
         show_cmd.add_argument(
             "--edit",
             default=False,
             action="store_true",
             help=(
-                "Launch marimo app preview in edit mode (only works together with --marimo flag,"
-                " otherwise it will be ignored)"
+                "Ejects marimo app preview to current directory (if it does not exist there yet)"
```
zilto: As above, align wording "eject" and "edit". Revised:

> Create and launch an editable copy of the marimo app to the current directory. Only works when both `--marimo --edit` flags are present.

Can we actively guard for incorrect args ? (e.g., `dlt pipeline foo show --edit`) shows warning `"--edit flag ignored. It only works when --marimo is also present"). Then, we can remove that info from the docstring

dlt/helpers/studio/dlt_app.py
```diff
@@ -284,7 +284,7 @@ def section_browse_data_table_list(
                     dlt_pipeline.dataset()
                     .table(_table_name)
                     .limit(1000 if dlt_restrict_to_last_1000.value else None)
-                    .query()
+                    .query(pretty=True)
```
zilto, line 287: An alternative to complete explicit `SELECT a, b, c FROM table` queries (with potentially a lot of aliases in Ibis), we can simply write `FROM table`. This is supported by the DuckDB dialect and our SQLGlot code should automatically handle it

dlt/helpers/studio/runner.py
```diff
@@ -15,15 +17,61 @@
         " ibis-framework.",
     )
 
+EJECTED_APP_FILE_NAME = "dlt_app.py"
```
zilto, line 20: rename EJECTED_APP to EDITABLE_APP

dlt/helpers/studio/runner.py
```diff
@@ -1,4 +1,6 @@
-import subprocess
+import os
+import pkg_resources
```
zilto: `pkg_resources` is deprecated (I'm guessing it's AI suggestion).  Use `importlib.resources` instead

dlt/helpers/studio/runner.py
```diff
@@ -15,15 +17,61 @@
...
+    from marimo._server.file_router import AppFileRouter
+
+    ejected_app_path = os.path.join(os.getcwd(), EJECTED_APP_FILE_NAME)
+    ejected_css_path = os.path.join(os.getcwd(), STYLE_FILE_NAME)
+    ejected_app_exists = os.path.exists(ejected_app_path)
 
-    studio_cmd = ["marimo", "run" if not edit else "edit", app.__file__]
+    # when editing, eject the app with styles to the cwd if not present already
+    if edit and not ejected_app_exists:
+        with open(dlt_app.__file__, "r", encoding="utf-8") as f:
```
zilto, line 38: `mode="r"` and `encoding="utf-8" are unnecessary because they're default.

dlt/helpers/studio/runner.py
```diff
@@ -15,15 +17,61 @@
...
+            app_code = f.read()
+            with open(ejected_app_path, "w", encoding="utf-8") as f:
```
zilto: indentation error? probably doesn't need to be nested; `encoding="utf-8"` is default value

dlt/helpers/studio/runner.py
```diff
@@ -15,15 +17,61 @@
...
+                f.write(app_code)
+        css_file_path = pkg_resources.resource_filename("dlt", f"helpers/studio/{STYLE_FILE_NAME}")
+        with open(css_file_path, "r", encoding="utf-8") as f:
```
zilto, line 43: same as above

dlt/helpers/studio/utils.py
```diff
@@ -217,7 +220,7 @@ def clear_query_cache(pipeline: dlt.Pipeline) -> None:
     get_query_result.cache_clear()
     get_loads.cache_clear()
     get_schema_by_version.cache_clear()
-    get_row_counts.cache_clear()
+    # get_row_counts.cache_clear()
```
zilto, line 223: should this be commented out or removed?

docs/website/docs/reference/command-line-interface.md
```diff
@@ -206,8 +206,8 @@ Inherits arguments from [`dlt pipeline`](#dlt-pipeline).
 
 **Options**
 * `-h, --help` - Show this help message and exit
-* `--marimo` - Launch marimo app preview instead of streamlit
-* `--edit` - Launch marimo app preview in edit mode (only works together with --marimo flag, otherwise it will be ignored)
+* `--marimo` - Launch marimo app preview instead of streamlit. will launch ejected version of app (created with --edit flag) if it exists in the current directory.
```
zilto: Should be re-generated after updating the CLI `.py` file

tests/helpers/studio/test_all_cells.py
```diff
@@ -41,8 +41,8 @@ def test_run_all_cells():
     Runs al cells with basic values, StopException is allowed
     """
     cells: List[Cell] = []
-    for item in dir(app):
-        c = getattr(app, item)
+    for item in dir(dlt_app):
```
zilto, line 44: Seems reasonable to [run the notebook as a script](https://docs.marimo.io/guides/scripts/). I think this test will be inconsistent if we were to mark some cells as "non-reactive"

pyproject.toml
```diff
@@ -316,6 +316,7 @@ include = [
     "README.md",
     "dlt/sources/pipeline_templates/.gitignore",
     "dlt/sources/pipeline_templates/.dlt/config.toml",
+    "dlt/helpers/studio/app_style.css"
```
zilto: Name is out of date. This PR renames the file and the runner calls it `dlt_app_styles.css`.

dlt/helpers/studio/runner.py
```diff
@@ -15,15 +17,61 @@
...
+        cli_args["pipeline"] = pipeline_name
 
-    subprocess.run(studio_cmd)
+    #
+    session_mode = SessionMode.RUN if not edit else SessionMode.EDIT
+
+    # app file
+    app_file_path = dlt_app.__file__ if not ejected_app_exists else ejected_app_path
+
+    start(
```
zilto, line 60: Awesome! A QoL issue with the previous approach was that `CTRL+C` to exit would not gracefully exit. I assume this fixes it?

dlt/helpers/studio/runner.py
```diff
@@ -1,4 +1,6 @@
-import subprocess
+import os
```
zilto, line 1:
```suggestion
from pathlib import Path
from importlib.resources import files
from typing import Any, Union

from dlt.common.exceptions import MissingDependencyException

# keep this, will raise if user tries to run studio without dependencies
try:
    import marimo
    import pandas
    import ibis
except ModuleNotFoundError:
    raise MissingDependencyException(
        "dlt marimo app",
        ["marimo", "pandas", "ibis-framework"],
        "the dlt marimo app requires additional dependencies, such as marimo, pandas and"
        " ibis-framework.",
    )

EDITABLE_APP_FILE_NAME = "dlt_app.py"
STYLE_FILE_NAME = "dlt_app_styles.css"


def create_editable_copy_and_get_path() -> Path:
    """Copy dlt's source code locally to create an editable notebook.
    The function returns the path to the local notebook.

    If the notebook already exists locally, just return the path without overriding.
    """
    editable_app_path = Path.cwd() / EDITABLE_APP_FILE_NAME
    if not editable_app_path.exists():
        app_source = Path(dlt_app.__file__).read_text()
        editable_app_path.write_text(app_source)
    
        app_css = Path(files("dlt.helpers.studio") / STYLE_FILE_NAME).read_text()
        (Path.cwd() / STYLE_FILE_NAME).write_text(app_css)

    return editable_app_path


def run_studio(pipeline_name: str = None, edit: bool = False) -> None:
    from dlt.helpers.studio import dlt_app

    # NOTE: we are using thinternal marimo server to run the dlt app, this might break
    from marimo._server.model import SessionMode
    from marimo._server.start import start
    from marimo._server.file_router import AppFileRouter

    app_file_path = create_editable_copy_and_get_path() if edit else dlt_app.__file__
```

dlt/helpers/studio/dlt_app.py
sh-rp: You could add this to the marimo app todo ticket (https://github.com/dlt-hub/dlt/issues/2701), for now I don't want to change this as it works well and it would entail some work to have a simplified version that is separate from our regular query rendering.

PR #2783, opened 2025-06-18
feat() add crlf support for csv exports

7amza79:
# Add CRLF Support for CSV Exports

## Description
This PR adds support for Windows-style line endings (CRLF) in CSV exports, while maintaining the default UNIX-style line endings (LF). This enhancement improves compatibility with Windows-based systems and tools that expect CRLF line endings.

## Changes
- Added `line_ending` configuration option to CSV writers with two possible values:
  - `lf` (default): Uses UNIX-style line endings
  - `crlf`: Uses Windows-style line endings
- Updated both `CsvWriter` and `ArrowToCsvWriter` to support line ending configuration
- Added comprehensive test coverage for both line ending types
- Updated documentation to reflect the new configuration option

## Technical Details
- Added `CsvLineEnding` type to configuration specs
- Implemented custom CSV dialect for Python's csv writer to handle CRLF
- Added version-aware implementation for PyArrow CSV writer (supports line endings from version 14.0)
- Added configuration options in both TOML and environment variables

## Testing
- Added unit tests for both `CsvWriter` and `ArrowToCsvWriter`
- Verified correct line ending generation for both LF and CRLF
- Ensured no mixed line endings in output files
- Added tests for both direct data writing and Arrow table conversion

## Documentation
- Updated CSV format documentation to include the new `line_ending` option
- Added examples for both TOML and environment variable configurations

## Configuration Example
```toml
[NORMALIZE]
data_writer__line_ending="crlf"
```

Or via environment variable:
```bash
NORMALIZE__DATA_WRITER__LINE_ENDING=crlf
```

## Checklist
- [x] Added new configuration option
- [x] Implemented support in both CSV writers
- [x] Added comprehensive tests
- [x] Updated documentation
- [x] Verified backward compatibility
- [x] Added examples for configuration

### DIFF
--- a/dlt/common/data_writers/buffered.py
+++ b/dlt/common/data_writers/buffered.py
@@ -46,6 +46,7 @@ class BufferedDataWriter(Generic[TWriter]):
         file_max_bytes: int = None,
         disable_compression: bool = False,
         _caps: DestinationCapabilitiesContext = None,
+        **writer_kwargs,
     ):
         self.writer_spec = writer_spec
         if self.writer_spec.requires_destination_capabilities and not _caps:
@@ -77,6 +78,7 @@ class BufferedDataWriter(Generic[TWriter]):
         self._created: float = None
         self._last_modified: float = None
         self._closed = False
+        self._writer_kwargs = writer_kwargs
         try:
             self._rotate_file()
         except TypeError:
@@ -243,7 +245,7 @@ class BufferedDataWriter(Generic[TWriter]):
                     self._file = self.open(self._file_name, "wb")  # type: ignore
                 else:
                     self._file = self.open(self._file_name, "wt", encoding="utf-8", newline="")
-                self._writer = self.writer_cls(self._file, caps=self._caps)  # type: ignore[assignment]
+                self._writer = self.writer_cls(self._file, caps=self._caps, **self._writer_kwargs)  # type: ignore[assignment]
                 self._writer.write_header(self._current_columns)
             # write buffer
             if self._buffered_items:
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,7 @@
 [pytest]
 pythonpath= dlt docs/website/docs
-norecursedirs= .direnv .eggs build dist
-addopts= -v --showlocals --durations 10
+norecursedirs= .direnv .eggs build dist .git .tox .env .venv venv env
+addopts= -v --showlocals --durations 10 --log-cli-level=INFO
 xfail_strict= true
 log_cli= 1
 log_cli_level= INFO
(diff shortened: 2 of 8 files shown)


dlt/common/data_writers/buffered.py
```diff
@@ -46,6 +46,7 @@ def __init__(
         file_max_bytes: int = None,
         disable_compression: bool = False,
         _caps: DestinationCapabilitiesContext = None,
+        **writer_kwargs,
```
rudolfix: I do not see this used. do you need it for testing? We pass `csv` options via config injection (you do it correctly below)

pytest.ini
```diff
@@ -1,14 +1,17 @@
 [pytest]
 pythonpath= dlt docs/website/docs
-norecursedirs= .direnv .eggs build dist
-addopts= -v --showlocals --durations 10
+norecursedirs= .direnv .eggs build dist .git .tox .env .venv venv env
+addopts= -v --showlocals --durations 10 --log-cli-level=INFO
```
rudolfix: please do not change the log level

rudolfix (changes requested):
this looks pretty good! pls see my comments. thx for the tests!

rudolfix:
@7amza79 please also fix linter: `make format` and `make lint`

burnash:
Hey @7amza79, would you like to continue working on this PR or you'd prefer that we take it from here? Thanks again for your effort!

7amza79:
@burnash I 've noticed that pyArrow didn't support CRLF writing neither so I've "abandonned" working on this implementation and used pandas instead.  So I think you can take it on from here if this "feature" is interesting for the product (which I think it is :D )

dlt/common/destination/configuration.py
```diff
@@ -10,6 +10,7 @@ class CsvFormatConfiguration(BaseConfiguration):
     delimiter: str = ","
     include_header: bool = True
     quoting: CsvQuoting = "quote_needed"
+    lineterminator: str = "\n"
```
rudolfix, line 13: this make sense. but will be almost impossible to specify via environment variables and bash. (you'll need to put actual CR / LF characters there). is it easy?

if not I'd add `on_resolved` handler to this config and "parse" literals r"\n" and r"\r" and convert them to actual codes. this is also good place to raise on wrong value

burnash: yeah, good point

burnash: let me try it, should't be impossible

burnash: not impossible:

```sh
LT=$'\r\n' python -c "import os; char = os.environ['LT']; print(char == '\r\n')"
True
```

good catch anyway as I need to update the docs

Issue #957, opened 2024-02-11
publish a stable dependence freeze for core library


rudolfix:
**Background**
We should start publishing a freeze of core library dependencies on which we run our test. Such dependencies should help users to build stable python environment. We recently had orjson crash which is a good example of why stable envs are important.
We already have a procedure to build dlt containers using the frozen deps so there are ways to export things.

Implementer should also research how to export such stable envs (besides containers?) ie. pip-tools, pip-compile, requirements.in etc. a little research is needed

**Requirements**
1. * [ ] publish freeze (possibly on make lint). The freeze should not change as long as packages don't
2. * [ ] add information that freeze exists ie. in Troubleshooting or in the instructions where we say how to build the environment (as an option)

PR #2787, opened 2025-06-19
run all common tests with resolution-lowest on uv sync

sh-rp:
### Description
We found a problem with our sqlglot dependency, so we now can use uv to install the lowest available dependencies that are allowed in pyproject and see wether at least our common tests pass.


zilto:
Solving this problem should allow to close this issue: #957

sh-rp:
> Solving this problem should allow to close this issue: #957

We can do that in another PR. The referenced ticket will probably not export lowest dependencies, but what is locked in the uv file.

pyproject.toml
```diff
@@ -68,7 +68,7 @@ gcp = [
 bigquery = [
     "grpcio>=1.50.0",
     "google-cloud-bigquery>=2.26.0",
-    "pyarrow>=14.0.0; python_version < '3.13'",
+    "pyarrow>=16.0.0; python_version < '3.13'",
```
sh-rp, line 71: we need this for "append_column" on batches in pyarrow. The alternative would be to change the implementation, but pyarrow 16 is over 1 year old, so I think we can bump it?

tests/common/test_time.py
```diff
@@ -164,7 +164,8 @@ def test_datetime_to_timestamp_helpers(
         ("2024-W42-5", "%Y-W%W-%u"),  # Week-based date with day
         ("2024-293", "%Y-%j"),  # Ordinal date
         ("20241020", "%Y%m%d"),  # Compact date format
-        ("202410", "%Y%m"),  # Compact year and month format
+        # TODO: fix for lowest direct resolution
+        # ("202410", "%Y%m"),  # Compact year and month format
```
sh-rp: Note to self: fix before review

pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 ]
 dependencies = [
     "requests>=2.26.0",
-    "pendulum>=2.1.2",
```
rudolfix: why we need to bump it? which tests are not passing. some people are still on 2.0... because it does not contain several "hacks" that 3.0 has. maybe we could just specify 3.0.0 for python > 3.11?

pyproject.toml
rudolfix: I'm sure I already bumped this. well... not :)

PR #2674, opened 2025-05-23
[Databricks destination] Adding comment and tags for table and columns and applying primary and foreign key constraints in Unity Catalog

bayees:
### Description

Create comments and tags for tables and columns in Unity Catalog tables. Also adding primary key and foreign key constraints using the primary_key and references hints.

I have added the create_indexes to enable primary and foreign key to keep backwards compatible

There is a new Databricks adapter to use the new x-databricks-cluster, x-databricks-table-comment, x-databricks-table-tags, x-databricks-column-comment, x-databricks-column-tags hints.

The table and column comments using description hints if those are available, but overrides those if the hints from the databricks adapter i given.

### Related Issues

- Closes #2625

PR #2791, opened 2025-06-19
(feat) allows to add SQL statements to schema migration executed after tables were created/altered

rudolfix:
### Description
Allows to override `_build_schema_update_sql` in sql job client. The intended usage is to call super class to collect sql statements schema changes and then add additional statements (ie. foreign references) that must be executed after all tables were created/altered.


dlt/destinations/job_client_impl.py
```diff
@@ -719,8 +732,15 @@ def _get_table_update_sql(
     def _get_constraints_sql(
         self, table_name: str, new_columns: Sequence[TColumnSchema], generate_alter: bool
     ) -> str:
+        """Creates or alters additional constraints that are not inlined into CREATE/ALTER TABLE statements"""
         return ""
 
+    def _get_table_post_update_sql(self, partial_table: TPartialTableSchema) -> List[str]:
+        """Generates SQL statements executed after all tables are migrated ie. containing foreign reference.
```
burnash:
```suggestion
        """Generates SQL statements executed after all tables are migrated i.e. containing foreign reference.
```

dlt/destinations/job_client_impl.py
```diff
@@ -668,6 +670,14 @@ def _build_schema_update_sql(
                 # keep only new columns
                 partial_table["columns"] = {c["name"]: c for c in new_columns}
                 schema_update[table_name] = partial_table
+                post_sql_statements = self._get_table_post_update_sql(partial_table)
+                for sql in post_sql_statements:
+                    if not sql.endswith(";"):
+                        sql += ";"
+                    post_sql_updates.append(sql)
```
burnash, line 677: It's best to extract this into a utility method, can do as a follow-up PR.

burnash (approved):
Looks good, there's a small code repetition we can address this in a follow up PR.

burnash:
@rudolfix and we'd need a test for it.

rudolfix:
> @rudolfix and we'd need a test for it.

right! hopefully #2674 will use the new override so for now tests there suffice

PR #2796, opened 2025-06-22
(QoL):adds `str` and `repr` to dataset and relation

rudolfix:
### Description
it is easier to work with datasets and relations in a notebook:
* adds `__repr__` to dataset
* adds str to dataset (display table list)
* adds str to relation (display expression schema)

### DIFF
--- a/dlt/destinations/dataset/relation.py
+++ b/dlt/destinations/dataset/relation.py
@@ -1,3 +1,4 @@
+from textwrap import indent
 from typing import Any, Dict, Generator, Optional, Sequence, Tuple, Type, Union, TYPE_CHECKING
 from contextlib import contextmanager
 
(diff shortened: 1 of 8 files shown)


dlt/destinations/dataset/relation.py
```diff
@@ -202,6 +205,17 @@ def normalized_query(self) -> sge.Query:
             )
         return self._normalized_query
 
+    def __str__(self) -> str:
```
sh-rp, line 208: why not just render and output the pretty formatted sql query here? or is this too abstract?

sh-rp (dismissed):

dlt/destinations/dataset/relation.py
rudolfix: right! we'll do this better. maybe like ibis: you show the source table, transformation and output table. this is just a quick update for 1.12.2

PR #2800, opened 2025-06-23
Fix: Null column type not inferred info/warning floods the output

anuunchin:
### Description
This PR collect the columns for which the type could not be inferred and uses one `UnboundColumnWIthoutTypeException` per table. Previously it was one `UnboundColumnWIthoutTypeException` per column, and the stdout was flooded with warning messages if there are too many of such columns. 

### Additional context
Since this is not a critical warning, I suggest that we move this to info 👀 - which will underneath involve just having an info message instead of an Exception object.  For now, the `UnboundColumnWIthoutTypeException` is kept.

### DIFF
--- a/dlt/normalize/validate.py
+++ b/dlt/normalize/validate.py
@@ -38,18 +38,23 @@ def verify_normalized_table(
     2. Raise `UnboundColumnException` on incomplete non-nullable columns (e.g. missing merge/primary key)
     3. Log warning if table format is not supported by destination capabilities
     """
+    null_first_cols = []
+
     for column, nullable in find_incomplete_columns(table):
         exc = UnboundColumnException(schema.name, table["name"], column)
         if nullable:
             # warn if null column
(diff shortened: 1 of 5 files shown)


dlt/normalize/validate.py
```diff
@@ -38,19 +38,24 @@ def verify_normalized_table(
...
         exc = UnboundColumnException(schema.name, table["name"], column)
         if nullable:
             # warn if null column
             seen_null_first = column.get("x-normalizer", {}).get("seen-null-first")
             if seen_null_first is True:
-                null_exc = UnboundColumnWithoutTypeException(schema.name, table["name"], column)
-                logger.warning(str(null_exc))
+                null_first_cols.append(column)
             else:
                 logger.warning(str(exc))
```
rudolfix: shoudl we handle this warning in the same way? multiple columns at once?

also a few lines below there's a typo `dtl schema` could you fix it?...

anuunchin: 🫡

PR #2806, opened 2025-06-24
add filesystem copy helper

sh-rp:
### Description
This PR implements a copy / import helper for the filesystem. When used, all files iterated by the filesystem resource are forwarded unchanged into the destination filesystem and no dlt metadata tables are created.

TODO

- [ ] Implement test stubs
- [ ] set up ci to test filesystem source from all buckets (credentials "problem" is already solved in the copy tests)
- [ ] Update docs to explain how to use this feature

### DIFF
--- a/dlt/sources/filesystem/copy.py
+++ b/dlt/sources/filesystem/copy.py
@@ -3,38 +3,50 @@ import tempfile, os
 from urllib.parse import urlparse
 import dlt
 
-from typing import Union, List, Any, overload, Iterator
+from typing import Any, Iterator
 
 from dlt.common.storages.fsspec_filesystem import FileItemDict
 from dlt.common.pipeline import SupportsPipeline
 from dlt.common.utils import custom_environ
 
+from dlt.common.storages.configuration import FilesystemConfiguration
+
 from dlt.extract import DltResource
 
+__source_name__ = "filesystem"
 
-@dlt.transformer()
-def import_transformer(items: Iterator[FileItemDict], temp_dir: str) -> Any:
+
+def _import_files(items: Iterator[FileItemDict], temp_dir: str) -> Any:
     for item in items:
         if isinstance(item, FileItemDict):
-            url = urlparse(item["file_url"])
             # local files can be imported directly
-            if url.scheme == "file":
+            if FilesystemConfiguration.is_local_path(item["file_url"]):
                 local_file_path = item.local_file_path
             else:
-                # remote files need to be downloaded to a temp directory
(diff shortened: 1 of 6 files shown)


dlt/sources/filesystem/copy.py
```diff
@@ -0,0 +1,96 @@
...
+                os.makedirs(local_file_dir, exist_ok=True)
+                with open(local_file_path, "wb") as f:
+                    if file_content := item.get("file_content", None):
+                        f.write(file_content)
+                    else:
+                        f.write(item.read_bytes())
+        # for string we assume it is a local file path
+        # TODO: does this make sense?
+        elif isinstance(item, str):
+            local_file_path = item
```
sh-rp, line 45: we should raise here is there is a dataitem that we can't use

rudolfix: this should never happen... `filesystem` yields `FileItemDict`. on anything else we should raise...

dlt/sources/filesystem/copy.py
```diff
@@ -0,0 +1,96 @@
...
+
+        ext = os.path.splitext(local_file_path)[1][1:]
+
+        # TODO: should we raise? Should it be configurable?
+        if ext not in [
```
rudolfix, line 50: hmmm maybe we can copy anything that is in the glob? but then we need more hack in filesystem to allow any file type

dlt/sources/filesystem/copy.py
```diff
@@ -0,0 +1,96 @@
...
+
+def copy_files(
+    items: DltResource,
+    pipeline: SupportsPipeline,
+    run_kwargs: Any = None,
+) -> None:
+    """Copy a file from the source to the destination.
+
+    Args:
+        resource: The source to copy from.
```
rudolfix, line 72: outdated

dlt/sources/filesystem/copy.py
```diff
@@ -0,0 +1,96 @@
...
+        # pipe through transformer and forward table name
+        table_name = items.table_name
+        items = items | dlt.transformer(table_name=table_name)(_import_files).bind(
+            temp_dir=temp_dir
+        )
+    else:
+        raise ValueError("Invalid resource parameter type: " + type(items))
+
+    with custom_environ({"DESTINATION__EXPERIMENTAL_EXCLUDE_DLT_TABLES": "True"}):
+        pipeline.run(items, **(run_kwargs or {}))
```
rudolfix, line 96: you can set this directly on the `pipeline.destination`.config_params no need to mock env variable.

sh-rp:
closing for later continuation

PR #2818, opened 2025-06-28
(chore) improves pool executor behaviors

rudolfix:
### Description
* warns if thread pool does not shut down in time (previously: locking)
* will fallback to spawn process start method if orchestrator is detected (should prevent prefect from locking in)
* uses `contextvar` to propagate exec pipe context (same mechanism for async and thread pool, removes global vars)
* source state handling moved from common to extract
* `dlt.state()` was removed (was deprecated in 0.4, should be removed in 1.0)
* changes some import to allow install wasm/pyodide


dlt/common/runners/pool_runner.py
```diff
@@ -38,10 +40,102 @@ def submit(self, fn: Callable[P, T], /, *args: P.args, **kwargs: P.kwargs) -> Fu
...
+        )
+
+        for m_ in [
+            is_running_in_airflow_task,
+            is_running_in_dagster_task,
+            is_running_in_prefect_flow,
+        ]:
+            if m_():
+                logger.info(
+                    f"Switching pool start method to `spawn` because `{m_.__name__}` is True"
```
djudjuu, line 124: praise: nice way of using the functions name. very concise. doesnt look like a human wrote it to me, but actually very nice.

dlt/common/runtime/exec_info.py
```diff
@@ -104,14 +110,41 @@ def is_airflow_installed() -> bool:
...
+    # module must be imported in dagster task
+    if dagster_ := sys.modules.get("dagster._core.execution.context.compute"):
+        try:
+            return dagster_.current_execution_context.get() is not None
+        except Exception:
+            pass
+    return False
+
+
+def is_running_in_prefect_flow() -> bool:
```
djudjuu, line 126: i have run my prefect playground parallelization example with this branch installed as dlt and it correctly detects prefect and sets the method to spawn, so that it no longer hangs during parallel process normalization

tests/common/runners/test_runners.py
```diff
@@ -167,3 +169,40 @@ def test_pool_runner_process_methods_configured(method) -> None:
...
+
+_tls = threading.local()
+
+
+def lock_del_task():
+    """Returns promptly, but its local object's __del__ blocks for 10 s."""
+
+    class Blocker:
+        def __del__(self):
+            time.sleep(2)
```
djudjuu, line 184: typo: should be 10

tests/common/runners/test_runners.py
```diff
@@ -167,3 +169,40 @@ def test_pool_runner_process_methods_configured(method) -> None:
...
     assert [v[0] for v in r.rv] == list(range(4))
+
+
+import threading
+
+_tls = threading.local()
+
+
+def lock_del_task():
+    """Returns promptly, but its local object's __del__ blocks for 10 s."""
```
djudjuu:
```suggestion
    """Returns promptly, but its local object's __del__ blocks for 2 s."""
```

pyproject.toml
```diff
@@ -46,7 +46,9 @@ dependencies = [
     "gitpython>=3.1.29",
     "pytz>=2022.6",
     "giturlparse>=0.10.0",
-    "orjson>=3.6.7,<4,!=3.9.11,!=3.9.12,!=3.9.13,!=3.9.14,!=3.10.1 ; platform_python_implementation != 'PyPy'",
+    "orjson>=3.6.7,<4,!=3.9.11,!=3.9.12,!=3.9.13,!=3.9.14,!=3.10.1 ; platform_python_implementation != 'PyPy' and sys_platform != 'emscripten'",
```
djudjuu, line 49: this is here on purpose because you've branched off of it?

uv.lock
```diff
@@ -2,11 +2,31 @@ version = 1
 revision = 1
 requires-python = ">=3.9.2, !=3.9.7, <3.14"
 resolution-markers = [
-    "python_full_version >= '3.13'",
-    "python_full_version == '3.12.*'",
-    "python_full_version == '3.11.*'",
-    "python_full_version == '3.10.*'",
-    "python_full_version < '3.10'",
+    "python_full_version >= '3.13' and platform_python_implementation == 'emscripten' and sys_platform != 'emscripten'",
```
djudjuu, line 5: same here I assume..

I still a bit confused around lockfile and pyproject.toml. first time that I work on a library, so I am not fully clear how it is best for our users / reproducible builds / ci/cds...

dlt/extract/concurrency.py
```diff
@@ -111,7 +115,9 @@ def submit(self, pipe_item: ResolvablePipeItem) -> TItemFuture:
         if isinstance(item, Awaitable):
             future = asyncio.run_coroutine_threadsafe(item, self._ensure_async_pool())
         elif callable(item):
-            future = self._ensure_thread_pool().submit(item)
+            # pass pipe context to thread pool, happens automatically for coroutines
+            ctx = contextvars.copy_context()
+            future = self._ensure_thread_pool().submit(ctx.run, item)
```
djudjuu, line 120: learning question: we don't have a test to reflect that, do we?

djudjuu (approved):
LGTM.
I am not sure I would have caught a bug here, as this is the first real time where I needed to think longer about threads, loops, futures and context boundaries.

but i definitely learned something and now have a better understanding of parallelism-paradigms

dlt/common/runtime/exec_info.py
rudolfix: yeah! I also ran some prefect tests. but finally I didn't include them. I hope our dlt+ tests will suffice

pyproject.toml
rudolfix: I'm separating `emscripten` (WASM) orjson dependency from all other system and architectures. pyodide has only 3.10.1 available which is buggy and crashes so I do not allow it in production. it will be good enough for WASM

dlt/extract/concurrency.py
rudolfix: yes we could have a synthetic test if context vars are really passed. but I expect that CPython team does that :)

we have really many tests for our use case. We use `contextvars` to pass "Pipe" instance to the executing task and this instance is used to discover the resource and its related state.
`test_resource_state_in_pipeline` check is context is propagated for different cases of tasks: single threaded, multi threaded and async

PR #2820, opened 2025-06-30
update lancedb orphan deletion mechanism

djudjuu:
## issue  1 orphan deletion
since some version lancedb's `tbl.merge_insert(merge_key).when_not_matched_by_source_delete(filter_condition
                ).execute(compatible_records)`

requires the source and target schema to align. 

-> implemented a mechanism to modify the records to be deleted to match the target schema. 
-> wrote some unit tests

## issue 2: api_key as hardcoded values no longer allowed in newer lancedb versions

instead they should be given to the registry, BUT that doesnt really work either.
Our mechanism to parse them from the environemt does work though

things i tried:
```python
        self.model_func = self.registry.get(embedding_model_provider).create(
            name=self.config.embedding_model,
            max_retries=self.config.options.max_retries,
            # api_key=self.config.credentials.api_key,  
            # >> ValueError: Sensitive key 'api_key' cannot be set to a hardcoded valueregistry error

            # api_key="$var:embedding_api_key",
            # api_key="$var:api_key",

            # both lead to
            # >> 401Error code: 401 - {'error': {'message': 'Incorrect API key provided: $var:api_key
            **({"host": embedding_model_host} if embedding_model_host else {}),
        )
        self.vector_field_name = self.config.vector_field_name

```
looking at the code, I think self.model_func iniitalized here is only used to read some metadata from to preserve to the schema, but doesnt connect ot the lancedb. 
=> so i think its not needed to pass it.

- tested it with local lancedb and lance-cloud (credentials in 1password)-dev

## which version to use?

after 0.22.0, 0.23.0 and 0.24.0 everything works, below there are different errors. 
during orphan deletion with embeddings, during sync, when trying to delete orphans with chunks... 

whats the plan here? should I try to make everything work for every single version?

also, to what should the deps for lancedb extra be pinned?

### resrouces

to run tests locally:
image


pyproject.toml
```diff
@@ -155,7 +155,7 @@ dremio = [
     "pyarrow>=18.0.0 ; python_version >= '3.13'",
 ]
 lancedb = [
-    "lancedb>=0.8.2 ; python_version < '3.13'",
+    "lancedb==0.24.0 ; python_version < '3.13'",
```
sh-rp: you should probably just bump the version in the lockfile, not here

djudjuu:
image

dlt/destinations/impl/lancedb/schema.py
```diff
@@ -83,6 +93,9 @@ def make_arrow_table_schema(
 
 
 def arrow_datatype_to_fusion_datatype(arrow_type: TArrowSchema) -> str:
+    # TODO: this is not tested. test add columns for all data types
```
djudjuu: test first!...only after test fails, this might need to be redone

dlt/destinations/impl/lancedb/utils.py
```diff
@@ -68,6 +78,82 @@ def fill_empty_source_column_values_with_placeholder(
...
+    try:
+        if write_disposition in ("append", "skip", "replace"):
+            tbl.add(records)
+        elif write_disposition == "merge":
+            # LanceDB requires identical schemas for when_not_matched_by_source_delete
+            # The source data schema must exactly match the target table schema (column names,
+            # order, and types). Only after 22. does it work with chunks and embeddings
+
+            target_schema = tbl.schema
+            # TODO: pass config here take vector column name
```
djudjuu: from lanceClientConfig name of column vector might change

dlt/destinations/impl/lancedb/jobs.py
```diff
@@ -0,0 +1,162 @@
...
+
+        super().__init__(file_path)
+        self._job_client: "LanceDBClient" = None
+        self.references = ReferenceFollowupJobRequest.resolve_references(file_path)
+
+    def run(self) -> None:
+        db_client: DBConnection = self._job_client.db_client
+        table_lineage: TTableLineage = [
+            TableJob(
+                # TODO: prepare table
```
djudjuu: preparation means: 
prepare table proparges default settings
schema should go through prepare

djudjuu:
one tests still fails on 
```sh
 uv run pytest tests/load/lancedb/test_merge.py::test_lancedb_root_table_remove_orphaned_records_with_real_embeddings
```

```sh
      
E           <class 'dlt.load.exceptions.LoadClientJobRetry'>
E           Job with `job_id=document.4e6a4e72d9.reference` had 5 retries which is a multiple of `max_retry_count=5`. Exiting retry loop. You can still rerun the load package to retry this job.

 Last failure message was: lance error: LanceError(IO): External error: LanceError(Arrow): Invalid argument error: Attempt to merge two RecordBatch with different sizes: 23 != 21, /root/.cargo/registry/src/index.crates.io-6f17d22bba15001f/lance-table-0.26.0/src/utils/stream.rs:46:25, /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/task/poll.rs:290:44
```

=> works when running lancedb==0.24.0 and 23...throws on 22

dlt/destinations/impl/lancedb/schema.py
djudjuu: added unit tests, they seem to work

dlt/destinations/impl/lancedb/utils.py
```diff
@@ -68,6 +78,82 @@ def fill_empty_source_column_values_with_placeholder(
...
+    Raises:
+        ValueError: If the write disposition is unsupported, or `id_field_name` is not
+            provided for update/merge operations.
+    """
+
+    try:
+        tbl = db_client.open_table(table_name)
+        tbl.checkout_latest()
+    except ValueError as e:
+        # TODO: why we wrap it? ValueError now means that we miss a table
```
djudjuu: just removed it and all tests kept passing

dlt/destinations/impl/lancedb/utils.py
djudjuu: done

dlt/destinations/impl/lancedb/schema.py
djudjuu: actually one fails for Time64Type(time64[us])

djudjuu: so `datetime.time(random.randrange(0, 24))` -> pa.Time64Type aka  Time64Type(time64[us]) (would be TIME)

which is not supported by lancedb:


lance error: Invalid user input: Unsupported data type: Time(None, None). Supported types: ["int [unsigned]", "tinyint [unsigned]", "smallint [unsigned]", "bigint [unsigned]", "float", "double", "string", "binary", "date", "timestamp(precision)", "datetime(precision)", "decimal(precision,scale)", "boolean"], /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lance-datafusion-0.29.0/src/planner.rs:520:17

tests/load/lancedb/test_schema_alignment.py
```diff
@@ -0,0 +1,281 @@
...
+        assert "decimal" in actual_columns, (
+            "Expected 'decimal' column to be present in destination table. Actual columns:"
+            f" {actual_columns}"
+        )
+
+
+# @pytest.mark.skip(reason="not completely understood what should happen")
+# this fails not sure why
+@pytest.mark.parametrize("object_format", ["object", "pandas", "arrow-table"])
+def test_missing_column_in_second_load(object_format: TestDataItemFormat) -> None:
```
djudjuu: error
2025-07-03 12:26:04,556|[WARNING]|203593|130097760614208|dlt|load.py|complete_jobs:430|Job for all_types_table.41059b8a6b.reference retried in load 1751538360.4558077 with message lance error: LanceError(Schema): Schema error: No field named target__dlt_load_id. Valid fields are "?table?".string, "?table?".float, "?table?".int, "?table?".datetime, "?table?".bool, "?table?".pre_normalized_column, "?table?".time, "?table?".binary, "?table?".decimal, "?table?".date., /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lance-0.30.0/src/dataset/write/merge_insert.rs:651:21 message:

tests/load/lancedb/test_schema_alignment.py
```diff
@@ -0,0 +1,281 @@
...
+        # Get the actual table schema from the destination
+        actual_columns = set(tbl.schema.names)
+
+        # Check that the decimal column is missing
+        assert "decimal" not in actual_columns
+
+
+@pytest.mark.xfail("normalizer issue?")
+def test_json_nesting_evolution() -> None:
+    """Test that json nesting evolution is handled correctly."""
```
djudjuu, line 301: error:
```
  raise PipelineStepFailed(
                self, "load", load_step.current_load_id, l_ex, step_info
            ) from l_ex
E           dlt.pipeline.exceptions.PipelineStepFailed: Pipeline execution failed at `step=load` when processing package with `load_id=1751538644.2943308` with exception:
E           
E           <class 'dlt.common.destination.exceptions.DestinationTransientException'>
E           lance error: Invalid user input: Column json__a already exists in the dataset, /root/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lance-0.30.0/src/dataset/schema_evolution.rs:149:21
```

tests/load/lancedb/test_schema_alignment.py
rudolfix: it happens because `_dlt_load_id` is not added to arrow table that is yielded. 
https://dlthub.com/docs/dlt-ecosystem/verified-sources/arrow-pandas#add-_dlt_load_id-and-_dlt_id-to-your-tables
this is required by "remove orphan" job

you need to enable this behavior. what we need to do here:
1. document this in docs
2. check this is `verify_schema` (for root tables, if remove orphan tables is enabled and `_dlt_load_id` is not present - raise an exception with an explanation)
3. parametrize this test so it removes orphans or not

tests/load/lancedb/test_schema_alignment.py
rudolfix: this is a  bug. I fixed it.

dlt/destinations/impl/lancedb/schema.py
```diff
@@ -83,6 +93,9 @@ def make_arrow_table_schema(
...
+    # o3 can generate this function from https://datafusion.apache.org/user-guide/sql/data_types.html
+    # also it may be possible to use readable names of arrow types
     type_map = {
         pa.bool_(): "BOOLEAN",
         pa.int64(): "BIGINT",
```
rudolfix: we are not testing other int sizes? because we should convert precision types as well. I've got something like this from o3
```
import pyarrow as pa
from pyarrow import types as pat

def arrow_to_datafusion(pa_type: pa.DataType) -> str:
    """
    Convert a pyarrow DataType instance to the canonical SQL-type string
    accepted by Apache DataFusion (see
    https://datafusion.apache.org/user-guide/sql/data_types.html).

    Parameters
    ----------
    pa_type : pyarrow.DataType
        The PyArrow type to convert.

    Returns
    -------
    str
        A SQL type string that DataFusion recognises.

    Raises
    ------
    NotImplementedError
        If the Arrow type has no current mapping in DataFusion’s SQL layer.
    """
    # --- Primitive & character ---

    if pat.is_boolean(pa_type):
        return "BOOLEAN"                                             # :contentReference[oaicite:0]{index=0}

    if pat.is_string(pa_type):
        return "VARCHAR"                                             # CHAR/VARCHAR/TEXT/STRING → Utf8 :contentReference[oaicite:1]{index=1}
    if pat.is_large_string(pa_type):
        return "VARCHAR"

    # --- Signed integers ---
    if pa_type == pa.int8():   return "TINYINT"                      # :contentReference[oaicite:2]{index=2}
    if pa_type == pa.int16():  return "SMALLINT"
    if pa_type == pa.int32():  return "INT"
    if pa_type == pa.int64():  return "BIGINT"

    # --- Unsigned integers ---
    if pa_type == pa.uint8():  return "TINYINT UNSIGNED"
    if pa_type == pa.uint16(): return "SMALLINT UNSIGNED"
    if pa_type == pa.uint32(): return "INT UNSIGNED"
    if pa_type == pa.uint64(): return "BIGINT UNSIGNED"

    # --- Floating point & decimal ---
    if pa_type == pa.float32(): return "FLOAT"
    if pa_type == pa.float64(): return "DOUBLE"

    if pat.is_decimal(pa_type):
        # works for Decimal128 & Decimal256
        return f"DECIMAL({pa_type.precision}, {pa_type.scale})"      # :contentReference[oaicite:3]{index=3}

    # --- Date & time ---
    if pat.is_date32(pa_type) or pat.is_date64(pa_type):
        return "DATE"                                                # :contentReference[oaicite:4]{index=4}

    if pat.is_time(pa_type):
        # DataFusion maps Time64(Nanosecond) to TIME; Time32* also works.
        return "TIME"                                                # :contentReference[oaicite:5]{index=5}

    if pat.is_timestamp(pa_type):
        # DataFusion’s SQL grammar simply uses TIMESTAMP; precision/zone
        # differences are preserved inside Arrow and don’t need to be in
        # the SQL type string.
        return "TIMESTAMP"                                           # :contentReference[oaicite:6]{index=6}

    if pat.is_interval(pa_type):
        return "INTERVAL"                                            # :contentReference[oaicite:7]{index=7}

    # --- Binary blobs ---
    if pat.is_binary(pa_type) or pat.is_large_binary(pa_type):
        return "BYTEA"                                               # :contentReference[oaicite:8]{index=8}

    # --- Unsupported / complex / nested ---
    unsupported_families = (
        pat.is_null, pat.is_fixed_size_binary, pat.is_list,
        pat.is_large_list, pat.is_struct, pat.is_map,
        pat.is_duration, pat.is_union, pat.is_dictionary, pat.is_run_end_encoded
    )
    if any(check(pa_type) for check in unsupported_families):
        raise NotImplementedError(
            f"Arrow type {pa_type} currently has no SQL equivalent in DataFusion."
        )

    # Fallback
    raise NotImplementedError(
        f"Unhandled Arrow type: {pa_type!r}"
    )
```

tests/load/lancedb/test_schema_alignment.py
djudjuu: rant: remove_orphans is built so badly!! NO_REMOVE_ORPHANS = x-remove-orphans defaults to False!
so what happends if i set the hint x-remove-orphans = False!!?

djudjuu: status: i have added code that emits a warning if there is a job with merge-key or remove-orphans (it would be better to check write_disposition but somehow I couldnt access it easily and also the test didnt work, so maybe I was wrong conceptually

test is failing for object and df:

image

dlt/destinations/impl/lancedb/schema.py
djudjuu: actually this whole method can be removed, because we can pass the pa.type directly to lancedb when creating a new column.

we could have a test whether that works though by slowly adding all those datatypes to a schema.
should be failry simple to adapt the other test

djudjuu:
remaning todos:

- [x]  write unit tests for different arrow precision types
 --> uintX -> intX
 floatX -> float64
 rest is preserved
 
image
I assume thats okay
test to run is `uv run pytest tests/load/lancedb/test_schema_alignment.py::test_arrow_precision_types`

- [x] mention that TIME type is not supported -> actually time does work now, probably because pa.types are used directly and not fusion datatypes
    - [ ]  ~~raise that TIME is not supported~~
> OK! we need to exclude TIME from our test, document that it does not work and ideally add make type mapper to raise a proper exception so def verify_supported_data_types( would catch it

- [x]  prepare table during remove orphan job
- [x]  specify the capabilities precision -> molka did it!
   - [x] check whether factory is correct
   - lancedb only supports decimal38, our factory is correct
>  we need to set the correct values in capabilities in factory.py so we do not produce too precise numbers

- [x]  raise exception with good explanation when orphans should get deleted and no _dlt_load_id is set

tests/load/lancedb/test_schema_alignment.py
```diff
@@ -0,0 +1,362 @@
...
+        actual_columns = set(tbl.schema.names)
+
+        # Check that the decimal column was added
+        assert "decimal" in actual_columns, (
+            "Expected 'decimal' column to be present in destination table. Actual columns:"
+            f" {actual_columns}"
+        )
+
+
+def test_arrow_precision_types():
```
djudjuu, line 157: @rudolfix is this expected or a bug?

docs/website/docs/dlt-ecosystem/destinations/lancedb.md
```diff
@@ -275,13 +279,5 @@ This is because PyArrow tables are immutable, so adding fields requires creating
 For huge tables, this may impact performance and memory usage since the full table must be loaded into memory to add the new fields.
 Keep these considerations in mind when working with large datasets and monitor memory usage if adding fields to sizable existing tables.
 
-### Null string handling for OpenAI embeddings
-
-OpenAI embedding service doesn't accept empty string bodies. We deal with this by replacing empty strings with a placeholder that should be very semantically dissimilar to 99.9% of queries.
-
-If your source column (column which is embedded) has empty values, it is important to consider the impact of this. There might be a _slight_ chance that semantic queries can hit these empty strings.
-
-We reported this issue to LanceDB: https://github.com/lancedb/lancedb/issues/1577.
```
djudjuu, line 284: I checked and this was closed

djudjuu:
- I've ran the tutorial from the side too to make sure it works (had to prevent the sentinelTable from being by every load-job thread).
- all of the above todos are done: https://github.com/dlt-hub/dlt/pull/2820#issuecomment-3032944406
- I am not sure about what should be in the `uv.lock`-file
- i enabled full-ci in between because i thought somehoe the test were not running
=> [this](https://github.com/dlt-hub/dlt/actions/runs/16075859496/job/45372659337?pr=2820
) test in the CI fails, I dont understand it
 
- some small todos from your initial review remain but I didnt get to them and you said they were not urgent

tests/load/lancedb/test_unit_test_add_vector_column.py
```diff
@@ -0,0 +1,301 @@
+import pyarrow as pa
+from datetime import datetime, date
+from decimal import Decimal
+
+from dlt.destinations.impl.lancedb.schema import add_vector_column
+
+
+def test_add_vector_column_basic_types() -> None:
```
rudolfix: I'm not sure what we are testing here. We have 4 cases to test:
- both schemas have vector column
- none has
- two other possibilities

the test should make sure that output schema has a vector column (added or kept) exactly at the index it was there originally. checking nulls is a nice addon :)

that should have end to end test as well where user adds vector in the input data. I wrote one: test_bring_your_own_vector I think it is enough for now. It would be good to have a test where we add vector ourselves but also ask one to be added with lance_adapter. but IMO we do not have time to do that now

rudolfix (changes requested):
this is pretty good! a lot of stuff got fixed. I've added a few last mile improvements to make lance to behave more like ducjdb: you can pass lance client instead of credentials. I also added test for passing own vector column (which resurfaced a few other problems - got fixed)

please take a look at my comment. we need better test in one place

rudolfix (approved):
LGTM!

docs/tools/prepare_examples_tests.py
```diff
@@ -11,7 +11,7 @@
 
 # settings
 SKIP_FOLDERS = ["archive", ".", "_", "local_cache"]
-SKIP_EXAMPLES: List[str] = []
+SKIP_EXAMPLES: List[str] = ["custom_destination_lancedb"]
```
djudjuu, line 14: this was needed because the script was just hanging up while trying to connect to the local lancedb (spotify.db)
image
so the lint-and-test-exapmle job would not finish but keep on running

PR #2823, opened 2025-06-30
Add remaining paramiko connect params to SFTP filesystem

### LINKED ISSUE #2822: Add `auth_strategy` support for SFTP client
### Feature description

Per https://docs.paramiko.org/en/stable/api/client.html, I would like to expose the `auth_strategy` parameter to use when using sftp filesystem

### Are you a dlt user?

### Use case

_No response_

### Proposed solution

Change dlt.common.configuration.specs.sftp_crendentials.SFTPCredentials to take in the `auth_strategy` param.

### Related issues

_No response_

AyushPatel101:
### Description
Docs - https://docs.paramiko.org/en/3.3/api/client.html#paramiko.client.SSHClient.connect

`paramiko.client.SSHClient.connect` has the option to provide a `pkey`, `disabled_algorithms`, `transport_factory`, `auth_strategy` params and I would like to expose those parameters for use.

### Related Issues
- Closes #2822


AyushPatel101:
Maybe @sh-rp if you have a chance to review

AyushPatel101:
Sorry to keep bothering, but would really appreciate a review on this as it is blocking my current workflow. Maybe @rudolfix ?

dlt/common/typing.py
```diff
@@ -89,6 +99,7 @@
 DictStrAny: TypeAlias = Dict[str, Any]
 DictStrStr: TypeAlias = Dict[str, str]
 DictStrOptionalStr: TypeAlias = Dict[str, Optional[str]]
+DictStrListStr: TypeAlias = Dict[str, List[str]]
```
sh-rp: We want to start getting rid of these kind of types, I think you can remove this and use `Dict[str, List[str]]` directly in the other places.

dlt/common/typing.py
```diff
@@ -65,6 +66,15 @@
...
 
+if TYPE_CHECKING:
+    try:
+        from paramiko import Transport
+    except ImportError:
+        Transport = Any  # type: ignore[misc, assignment]
+else:
+    Transport = Any
+
+SFTPTransportFactory: TypeAlias = Callable[[socket.socket], Transport]
```
sh-rp: There are types specific to the sftp_credentials, you should define them in the `sftp_credentials` class and not import anything from paramiko here.

sh-rp (changes requested):
Hey @AyushPatel101, sorry for the long wait, we were very busy! Thanks for your PR, I just have small change requests wrt typing, and please update the filesystem docs page with the new config parameters.

AyushPatel101:
@sh-rp requesting re-review whenever you get a chance

tests/load/filesystem_sftp/test_filesystem_sftp.py
```diff
@@ -12,6 +11,10 @@
 
 from tests.load.utils import ALL_FILESYSTEM_DRIVERS
 
+from paramiko.auth_strategy import Password
```
sh-rp: these paramiko imports will have to go below the pytest.skip directive, paramiko is not installed for the other destinations tests, we test dependency isolation, and pytest will fail I think. Please push this change and I added the tag for running the full ci, then we can merge.

sh-rp:
Thanks :)

PR #2824, opened 2025-07-01
Chore: Replacing query strings with sqlglot expressions

anuunchin:
### Description
This PR replaces manual query strings with sqlglot expressions where appropriate. `def count_rows()` was already replaced with sqlglot in #2814.


dlt/destinations/queries.py
```diff
@@ -33,7 +69,15 @@ def normalize_query(
         if isinstance(node, sge.Table):
             # expand named of known tables. this is currently clickhouse things where
             # we use dataset.table in queries but render those as dataset___table
-            if sqlglot_schema.column_names(node):
+            # Check if table exists in schema (not just if it has columns)
+            try:
```
sh-rp, line 75: in which case does this happen? are we building queries against tables that are not in the schema yet, maybe on sync_destinations?

dlt/destinations/queries.py
```diff
@@ -121,3 +165,259 @@ def build_select_expr(
...
+    columns: List[str],
+    quoted_identifiers: bool = True,
+) -> sge.Insert:
+    """Builds a SQL Insert expression with placeholders."""
+    table_expr = sge.Table(this=sge.to_identifier(table_name, quoted=quoted_identifiers))
+
+    columns_expr = [sge.to_identifier(col, quoted=quoted_identifiers) for col in columns]
+
+    placeholders_expr = sge.Values(
+        expressions=[sge.Tuple(expressions=[sge.Placeholder() for _ in enumerate(columns)])]
```
sh-rp: Maybe it is possible to set `Literal.string('%s')` instead of `sge.Placeholder()` here and skip the replacing of placeholders via regex? Did you try that?

dlt/destinations/job_client_impl.py
```diff
@@ -522,37 +534,61 @@ def _from_db_type(
...
-            query = (
-                f"SELECT {self.version_table_schema_columns} FROM {name}"
-                f" ORDER BY {c_inserted_at} DESC;"
-            )
-            return self._row_to_schema_info(query)
-        else:
-            query = (
-                f"SELECT {self.version_table_schema_columns} FROM {name} WHERE {c_schema_name} = %s"
-                f" ORDER BY {c_inserted_at} DESC;"
+    def _render_sql(
```
sh-rp: I think I'd move this into the queries file, then it can be used from other places too if needed.

dlt/transformations/lineage.py
```diff
@@ -34,38 +35,43 @@ def create_sqlglot_schema(
...
+    sqlglot_schema_dict = {}  # MappingSchema(empty_schema, normalize=False)
+
+    for table_name in schema.tables.keys():
+        # skip not materialized columns
+        columns = schema.get_table_columns(table_name, include_incomplete=False)
+        column_mapping = map_columns(columns)
+        if column_mapping:
+            sqlglot_schema_dict[table_name] = column_mapping
+
+    if schema.state_table_name not in sqlglot_schema_dict:
```
sh-rp, line 67: When exactly do we need this, also during sync_destination in cases where we do not have local data?

dlt/destinations/queries.py
anuunchin: When we trigger this function to create a table in `def _make_create_table()` which is called by `def _get_table_update_sql()` in `job_client_impl.py`

dlt/destinations/queries.py
anuunchin: this is genius

dlt/transformations/lineage.py
anuunchin: when the schema is first initialized - at that point the state table is not part of the tables and is only accesssible with schema.state_table_name

PR #2832, opened 2025-07-02
deploy marimo wasm playground to github pages and add to docs

sh-rp:
### Description
This PR attempts to deploy our own marimo notebooks to github pages.

* Example deployment here: https://dlt-hub.github.io/dlt/playground/
* Example of iframe on our docs: https://deploy-preview-2832--dlt-hub-docs.netlify.app/docs/devel/tutorial/playground


dlt/common/runners/pool_runner.py
```diff
@@ -170,8 +170,18 @@ def run_pool(
         )
 
     # start pool
-    pool = create_pool(config)
```
rudolfix, line 173: makes sense! but I'd go even further and make the check below in:
```py
def create_pool(config: PoolRunnerConfiguration) -> Executor:
```
and pick NullExecutor if threading is not available

dlt/common/runtime/telemetry.py
```diff
@@ -25,6 +26,10 @@ def start_telemetry(config: RuntimeConfiguration) -> None:
     if _TELEMETRY_STARTED:
         return
 
+    # telemetry uses threading
+    if not platform_supports_threading():
```
rudolfix: my take: change `RuntimeConfiguration` `on_resolved` and disable telemetry there.

rudolfix (changes requested):
Could you add a test for both changes? Also we need to align @AstrakhantsevaAA and @burnash how we want to link playground to our docs

PR #2833, opened 2025-07-02
Enhances ReadableDBAPIRelation: min/max, filter with expression

### LINKED ISSUE #2831: Enhancing ReadableDBAPIRelation: min/max, filter with expression
The `RedableDBAPIRelation` needs further enhancing as part of the transformations work:

- [ ] enable min on selects with a single column (so we can do scalar()
```python
def max(self) -> Self:
    if len(self._sqlglot_expression.selects) != 1:
        raise ValueError("Only allowed on selects with one colum")    
    current_select = self._sqlglot_expression.selects[0]
    max_expr = sge.Max(this=current_select) 
    rel = self.__copy__()
    rel._sqlglot_expression.selects = [max_expr]
```
> The user does: 
> ```
> dataset.items.select(["col1]").max()
> dataset.items.select(["col1"]).max.scalar()
> ```
- [ ] enable max as above
- [ ] overload filter to any SQL expression can be passed (as a string)
Something like:
```Python
def where(self, expressions: exprOrStr) -> Self:
   parsed_expr = maybe_parse(expressions)
   ... 
   rel = self.__copy__()
   rel._sqlglot_expression = rel._sqlglot_expression.where(condition)
```
> The user does: 
> ```
> expr = parse_one("col1 > 1")
> dataset.items.filter("col1 > 1")
> dataset.items.filter(expr)
> ```
> Where either string or a direct sqlglot expression can be passed.

anuunchin:
### Description
Adds `def max()`, `def min()` and accepts expressions in `def where()`.

### Related Issues

- Resolves #2831


dlt/common/destination/dataset.py
```diff
@@ -171,18 +172,52 @@ def select(self, *columns: str) -> Self:
...
+    def min(self) -> Self:  # noqa: A003
+        """Returns a new relation with the MIN aggregate applied.
+        Exactly one column must be selected.
+
+        Returns:
+            Self: The relation with the MIN aggregate expression.
+        """
+
+    @overload
+    def where(self, *, expr_or_str: SqlglotExprOrStr) -> Self: ...
```
sh-rp: I would change it so we allow two overloads: one is the old notation and the other is the new notation but the first var will take string or expression an the others are not used. you can call it column_or_expr or something like that.

sh-rp (changes requested):
This looks very good already. Could you do the following:
* Update the docs
* see if you can use your new min/max in `tests/load/transformations/test_incremental_transforms.py` for all tests? I think we should be able to replace all ibis code there.
* Also one other thought: Can you check that the same where expression as sql string works for all destination if we use quotes? I think the quoting character varies from destination to destination and it would be cool to be able to run the same code against all destination. For this, where you have `filtered_items_sql = items.where(expr_or_str="id < 10").fetchall()`, maybe also do a test for filtering by load_id which is a string and see if it passes everywhere.

PR #2835, opened 2025-07-03
Fix: saving compressed load files with .gz extension

### LINKED ISSUE #925: Save compressed load files with `.gz` extension
### Problem description

Load files are always saved with the extension of the file format, regardless of whether compression is enabled.    
I.e. `s3://path/to/load/file.jsonl` may or may not be compressed.   
This causes issues with e.g. databricks loader which can't parse gzip files without `.gz` extension so compression must be disabled.   Possibly affects snowflake and redshift too, I think we assume json files are always compressed.

### Solution

The data writer should add `.gz` extension when compression is enabled.  
There are few places where we parse filenames to detect format, usually relying on `filename.endswith('.<file_format>')` or `os.path.splitext`. These need to be refactored, we need a clean way to get format/compression from filename.

anuunchin:
### Description
This PR adds the `.gz` extension to compressed files based on a new config `disable_extension` that is set to `True` by default.

In a nutshell:
- `is_compression_disabled()` is removed and destination load jobs have access to `BufferedDataWriterConfiguration`. 
- Since the task involves getting rid of tricks like `FileStorage.is_gzipped()`, solely relying on  `BufferedDataWriterConfiguration` is not sufficient when it comes to imported files, since they must not be compressed in any situation. For this reason, a new context class `FileImportContext` was introduced that tracks whether a file is an imported file and, thus, does not require compression. The `is_compressed_file` attribute of `FileImportContext` is then accessed by the duckdb and clickhouse load jobs to correctly set the compression type.

### Related Issues

- Resolves #925


dlt/common/data_writers/buffered.py
```diff
@@ -176,6 +180,11 @@ def close(self, skip_flush: bool = False) -> None:
             self._flush_and_close_file(skip_flush=skip_flush)
             self._closed = True
 
+    @property
+    def _is_compression_enabled(self) -> bool:
+        """Returns True if compression is enabled for this writer"""
+        return self.writer_spec.supports_compression and self.open == gzip.open
+
```
anuunchin: This might be redundant. We can just check is `self.open` is `gzip_open`

dlt/destinations/impl/duckdb/sql_client.py
```diff
@@ -516,7 +516,7 @@ def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DB
                 continue
             schema = table.db
             # add only tables from the dataset schema
-            if schema or schema.lower() != self.dataset_name.lower():
```
anuunchin: I'm not 100% sure if this makes sense because it runs in every case except one corner-case with empty schema and empty dataset_name 🤔

tests/common/storages/test_compressed_files.py
anuunchin, line 1: This test is not exhaustive, since some writers are missing, but i don't think it's a problem for the first round of reviews 👀

sh-rp:
I did not read the PR, but just from a user / compatibility perspective, I think dlt should behave the exact same way with regards to adding these extensions after updating if you do not change any settings. This means:
* adding the gzip extension must be configurable
* This setting must be set to off by default
* It would be nicer to have it switched on by default, but then we would need to make this change in dlt 2.0 and mark it as breaking

sh-rp:
@anuunchin what I would do is the following: Use the .gz extension by default everywhere because it makes stuff simpler and then remove the .gz extensions for the cases where the filesystem is the main destination and the user has selected to not keep the extension there (which should be the default). So as we discussed:
* Always add or keep .gz extension in the DataWriter if compression is enabled, so after extracing and normalizing we have .gz on files that are compressed
* When importing files, probe the file and add .gz if compressed
* Add a new setting to the filesystem destination that controls wether the files should have a .gz compression if compressed, it be set to remove this extension by default to ensure backward compatibility. This setting should probably be ignored if the filesystem is used as the staging destination.
* If this setting is set, rename files that are copied to buckets or the local filesystem to not have this extension while actually running the copy job.

Things to check:
* This approach assumes that the users are ok with having the .gz extension on files in their staging buckets. Please confirm with @rudolfix that we can do this.
* Check if this has any influence on delta and iceberg. I think we output parquet files in the normalizer for both of these destinations, so we should be good there
* Athena (without iceberg) uses the filesystem destination as staging, but uses those exact files for backing the tables with data. In this case, the above setting should be respected.

dlt/common/data_writers/buffered.py
```diff
@@ -148,8 +147,10 @@ def import_file(
         if with_extension:
             spec = self.writer_spec._replace(file_extension=with_extension)
         with self.alternative_spec(spec):
-            self._rotate_file()
+            self._rotate_file(is_imported_file=True)
```
sh-rp: check wether the file is compressed before you rotate the file, set this new parameter mentioned below to true or false accordingly, then you do not have to mess with the filename at all here as it is set properly in the rotate_file method.

dlt/common/data_writers/buffered.py
```diff
@@ -226,11 +227,20 @@ def _buffer_items_with_row_count(self, item: TDataItems) -> int:
                 new_rows_count = 1
         return new_rows_count
 
-    def _rotate_file(self, allow_empty_file: bool = False) -> DataWriterMetrics:
+    def _rotate_file(
```
sh-rp, line 232: I would add a parameter `mark_compressed` instead of `is_imported_file`. if set to None, take the should_compress value, if set to true of false, respect that setting instead to decide wether to add this extension or not.

dlt/common/storages/load_package.py
```diff
@@ -156,14 +156,20 @@ class ParsedLoadJobFileName(NamedTuple):
     file_id: str
     retry_count: int
     file_format: TJobFileFormat
+    has_compression_ext: bool = False
```
sh-rp: I'd calls this `is_compressed`

dlt/common/storages/load_package.py
```diff
@@ -156,14 +156,20 @@ class ParsedLoadJobFileName(NamedTuple):
...
 
     def job_id(self) -> str:
         """Unique identifier of the job"""
+        if self.has_compression_ext:
```
sh-rp: i'd write this like the function below, first construct the basename, then add the extension, then you don't have the one part double.

dlt/common/storages/load_package.py
```diff
@@ -156,14 +156,20 @@ class ParsedLoadJobFileName(NamedTuple):
...
+            return f"{self.table_name}.{self.file_id}.{self.file_format}.gz"
         return f"{self.table_name}.{self.file_id}.{self.file_format}"
 
     def file_name(self) -> str:
         """A name of the file with the data to be loaded"""
-        return f"{self.table_name}.{self.file_id}.{int(self.retry_count)}.{self.file_format}"
+        base_name = f"{self.table_name}.{self.file_id}.{int(self.retry_count)}.{self.file_format}"
+        if self.has_compression_ext:
+            return f"{base_name}.gz"
+        return base_name
```
sh-rp: just as an idea, you can write this in two lines:
```
compression_extension = ".gz" if self.is_compressed else ""
return f"{self.table_name}.{self.file_id}.{int(self.retry_count)}{compression_extension}"
```

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -83,21 +84,24 @@ def run(self) -> None:
             bucket_url = urlparse(bucket_path)
             bucket_scheme = bucket_url.scheme
 
-        ext = cast(SUPPORTED_FILE_FORMATS, os.path.splitext(file_name)[1][1:].lower())
-        clickhouse_format: str = FILE_FORMAT_TO_TABLE_FUNCTION_MAPPING[ext]
-
         compression = "auto"
 
+        file_format, compression_ext = get_file_format_compression(file_name)
```
sh-rp: again, I would call this `is_compressed`, this is what it signifies

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -218,18 +218,12 @@ def _handle_staged_file(self) -> tuple[str, str, str, str]:
     def _determine_source_format(
         self, file_name: str, orig_bucket_path: str
     ) -> tuple[str, str, bool]:
-        if file_name.endswith(".parquet"):
-            return "PARQUET", "", False
+        file_format, _ = get_file_format_compression(file_name)
 
-        elif file_name.endswith(".jsonl"):
-            if not is_compression_disabled():
-                raise LoadJobTerminalException(
```
sh-rp, line 242: this exception is removed, please add it back in for the cases where this file is compressed. or is this disabled somewhere?

dlt/destinations/impl/duckdb/duck.py
```diff
@@ -32,10 +33,12 @@ def run(self) -> None:
...
+
+        file_format, _ = get_file_format_compression(self._file_path)
+        if file_format == "parquet":
             source_format = "read_parquet"
             options = ", union_by_name=true"
-        elif self._file_path.endswith("jsonl"):
+        elif file_format in ["jsonl", "typed-jsonl"]:
             # NOTE: loading JSON does not work in practice on duckdb: the missing keys fail the load instead of being interpreted as NULL
             source_format = "read_json"  # newline delimited, compression auto
             options = ", COMPRESSION=GZIP" if FileStorage.is_gzipped(self._file_path) else ""
```
sh-rp: this should not be here anymore, you should get the compression status from the filename, no?

dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -25,6 +25,8 @@ class FilesystemDestinationClientConfiguration(FilesystemConfigurationWithLocalF
     """Maximum number of pipeline state files to keep; 0 or negative value disables cleanup."""
     always_refresh_views: bool = False
     """Always refresh table scanner views by setting the newest table metadata or globbing table files"""
+    legacy_compression_without_ext: bool = False
```
sh-rp: Did we not decide in the discussion to not have this anymore? I think we always add this extension now.

dlt/destinations/job_impl.py
```diff
@@ -79,7 +79,8 @@ def job_id(self) -> str:
 
 class ReferenceFollowupJobRequest(FollowupJobRequestImpl):
     def __init__(self, original_file_name: str, remote_paths: List[str]) -> None:
-        file_name = os.path.splitext(original_file_name)[0] + "." + "reference"
+        job_info = ParsedLoadJobFileName.parse(original_file_name)
```
sh-rp, line 82: maybe ParsedLoadJobFileName could have a method `to_reference_file_name` that does this? And you can use it here?

dlt/destinations/path_utils.py
```diff
@@ -139,7 +139,10 @@ def prepare_params(
     if job_info:
         table_name = job_info.table_name
         file_id = job_info.file_id
-        ext = job_info.file_format
+        if job_info.has_compression_ext:
```
sh-rp: more or less same as above, maybe ParsedLoadJobFileName could have a method "full_extension" which does this little code block (and explains what it does)

dlt/destinations/path_utils.py
```diff
@@ -241,7 +244,10 @@ def create_path(
 
     # if extension is not defined, we append it at the end
     if "ext" not in placeholders:
-        path += f".{job_info.file_format}"
+        if job_info.has_compression_ext:
```
sh-rp: you can use the above. helper here too then.

dlt/destinations/path_utils.py
```diff
@@ -286,3 +292,20 @@ def get_table_prefix_layout(
         )
 
     return prefix
+
+
+def get_file_format_compression(file_path: str) -> Tuple[str, bool]:
```
sh-rp: i'd call it `get_file_format_and_compression` then the name is clearer.

sh-rp: Also we need some dedicated unit tests for this

docs/website/docs/dlt-ecosystem/destinations/filesystem.md
```diff
@@ -436,6 +436,27 @@ disable_compression=true
 
 - To decompress a `gzip` file, you can use tools like `gunzip`. This will convert the compressed file back to its original format, making it readable.
 
+### Compressed file extensions
+
+Starting from dlt version 1.14.0, compressed files automatically receive a `.gz` extension. Previously, compressed files were stored without the `.gz` extension.
```
sh-rp: probably version 1.15 (same is further below)

sh-rp (changes requested):
Thanks for the updates, it is already looking good :) I have a bunch of changes that are related to making it clearer what is going on with better naming and less double code. 

About maintaining backwards compatibility: I think the decision was to not have this setting but rather to maintain a version marker in the init file of the filesystem destination, and always add this extension for new filesystem datasets and maintain the old behavior for existing ones. Right now the init marker is just an empty file. Going forward, it should contain something l like "{'version': 1}" or similar. Depending on what it says in this file, the extensions will be added or not. I'll add details later today.

dlt/destinations/impl/filesystem/configuration.py
sh-rp: Ok, after rereading the discussion, I think we decided to keep this

sh-rp:
After more discussion, I think we will need to maintain a filesystem versioning. So the proposal is this:
* Remove this flag that governs wether to add the extension or not
* Version the filesystem, we keep the version in the base init file. If this file is empty, we assume its content to be `{'version': 1}` if it does not exist on initialize_storage, we set it to `{'version': CURRENT_VERSION}` which is a constant that is 2. If we encounter anything else in this file (non empty and parsing it does not have a version 1 or 2) we raise.
* An initialized filesystem instance should always know its current version, for now we do not do any migrations or anything like that, older datasets will have version 1, newer ones have version 2.
* datasets with version 1 will continue to save the files without this extension. 
* datasets with version 2 will add this extension
* We need to test this behavior with an older dlt version

dlt/destinations/impl/databricks/databricks.py
anuunchin: databricks itself does support compressed jsonl files, but looks like it fails to handle compressed files without the .gz extension. So we don't need this anymore

dlt/destinations/impl/filesystem/sql_client.py
```diff
@@ -218,9 +216,16 @@ def _escape_column_name(col_name: str) -> str:
...
+                        f" {{{column_types}}})"
                     )
 
+                if (
+                    table_name not in dlt_table_names
+                    and self.remote_client.get_storage_version() == 1
+                    and not is_compression_disabled()
+                ):
+                    from_statement = from_statement[:-1] + ", compression = 'gzip')"
+
```
anuunchin: We still use this `is_compression_disabled()` trick because solely knowing the storage version is not enough as the files inside it may or may not be compressed. Without this `if condition`, the duckdb reader won't work with filesystem datasets of version 1. 

PS: I put this as separately as possible, so that we can maybe remove this in the future 👀

dlt/common/data_writers/buffered.py
```diff
@@ -226,11 +229,23 @@ def _buffer_items_with_row_count(self, item: TDataItems) -> int:
...
-        self._file_name = (
-            self.file_name_template % new_file_id() + "." + self.writer_spec.file_extension
-        )
+
+        base_filename = self.file_name_template % new_file_id()
+        file_extension = self.writer_spec.file_extension
+
+        # Add .gz extension if compression is enabled and is not an imported file
+        # or if marked as a compressed file
+        if (mark_compressed is None and self.should_compress) or mark_compressed is True:
```
sh-rp: Not super important, but why not simply `if mark_compressed or self.should_compress:` ?

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -98,6 +102,12 @@ def run(self) -> None:
         # trade offs between different options and decided on this.
         # remote_path = f"{client.config.protocol}://{posixpath.join(dataset_path, destination_file_name)}"
         remote_path = self.make_remote_path()
+        if (
+            not self._job_client.config.as_staging_destination
+            and self._job_client.get_storage_version() == 1
```
sh-rp: This will access the init file and get the storage version every time a file is loaded, you should cache this on the job_client instance, it will be very bad for performance on buckets otherwise.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -304,6 +314,7 @@ def __init__(
         self._sql_client: SqlClientBase[Any] = None
         # iceberg catalog
         self._catalog: Any = None
+        self._init_path: str = None
```
sh-rp: this does not seem to be used, you can probably remove it?

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -354,10 +365,46 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
...
+
         # we mark the storage folder as initialized
         self.fs_client.makedirs(self.dataset_path, exist_ok=True)
         self.fs_client.touch(self.pathlib.join(self.dataset_path, INIT_FILE_NAME))
 
+        # we set the version to "2" to indicate that .gz extension is added by default
+        # version "1" corresponds to an empty init file and indicates that .gz is not added to compressed files
+        self.fs_client.write_text(
+            self.pathlib.join(self.dataset_path, INIT_FILE_NAME),
+            json.dumps({"version": 2}),
```
sh-rp: Here, always `CURRENT_VERSION` should be used, which is 2 at the moment but will be higher possibly later.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -354,10 +365,46 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
             logger.info(f"Will truncate tables {truncate_tables}")
             self.truncate_tables(list(truncate_tables))
 
+        # if the init file already exists, we return
+        if self.fs_client.exists(self.pathlib.join(self.dataset_path, INIT_FILE_NAME)):
```
sh-rp: you are creating the path to the init file multiple times, you can do it once, this reduces duplication.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -354,10 +365,46 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
...
 
+        # we set the version to "2" to indicate that .gz extension is added by default
+        # version "1" corresponds to an empty init file and indicates that .gz is not added to compressed files
+        self.fs_client.write_text(
+            self.pathlib.join(self.dataset_path, INIT_FILE_NAME),
+            json.dumps({"version": 2}),
+            encoding="utf-8",
+        )
+
+    def get_storage_version(self) -> int:
```
sh-rp: This is very good! We just need to cache the version on the client which should be easy. it will not change during the life of the client.

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -354,10 +365,46 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
...
+        except (KeyError, orjson.JSONDecodeError) as exc:
+            raise ValueError(f"Invalid content in {init_path}: {version_info_str!r}") from exc
+
+        if version not in SUPPORTED_VERSIONS:
+            raise ValueError(
+                f"Unsupported filesystem version {version!r} "
+                f"(supported: {sorted(SUPPORTED_VERSIONS)})"
+            )
+
+        return int(version)
```
sh-rp: I don't think you need to cast to int here, do you?

dlt/destinations/path_utils.py
```diff
@@ -286,3 +286,20 @@ def get_table_prefix_layout(
         )
 
     return prefix
+
+
+def get_file_format_and_compression(file_path: str) -> Tuple[str, bool]:
```
sh-rp, line 291: Actually, now that I think of it, this function more or less does the same as `ParsedLoadJobFileName`, I think you could use that everywhere really...

sh-rp: Or maybe I am mistaken?

docs/website/docs/dlt-ecosystem/destinations/filesystem.md
```diff
@@ -436,6 +436,10 @@ disable_compression=true
 
 - To decompress a `gzip` file, you can use tools like `gunzip`. This will convert the compressed file back to its original format, making it readable.
 
+:::note
+Starting with dlt version 1.16.0, compressed `csv` and `jsonl` files automatically include a `.gz` extension to reflect their gzip-compressed format. In versions prior to 1.16.0, compressed files were saved without the `.gz` extension. If you have a dataset created with an earlier version (e.g., 1.15.0 or below), dlt will automatically detect the older format and preserve the original naming (without `.gz`) for that dataset. New datasets created with 1.16.0 or later will include the `.gz` extension by default.
```
sh-rp: We may release this change with 1.15, so you'll have to update this text.

dlt/common/data_writers/buffered.py
anuunchin: owww, right

anuunchin: wait, then it would fail for imported but not compressed files, because self.should_compress may still be true

dlt/destinations/impl/filesystem/filesystem.py
anuunchin: owww, makes sense

dlt/destinations/impl/filesystem/filesystem.py
anuunchin: json loads returns Any type, but I'll remove the ugly casting and put it a bit above with `version: int = ...`

dlt/destinations/path_utils.py
anuunchin: PrasedLoadJobFileName isn't able to handle staged storage files (`{load_id}.{file_id}.{ext}`) 👀

anuunchin: Adjusted the function description and removed it from places where ParsedLoadJobFileName.parse() can be used - I think this makes more sense and conveys the difference

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -354,9 +379,46 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
...
             self.truncate_tables(list(truncate_tables))
 
+        # if the init file already exists, we return
+        if self.fs_client.exists(self.init_file_path):
+            return
+
         # we mark the storage folder as initialized
         self.fs_client.makedirs(self.dataset_path, exist_ok=True)
-        self.fs_client.touch(self.pathlib.join(self.dataset_path, INIT_FILE_NAME))
+        self.fs_client.touch(self.init_file_path)
```
rudolfix: you do not need to touch anymore

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -228,9 +229,10 @@ def create_load_job(
                         " `write_disposition='append'`. Resource received"
                         f" `write_disposition={table['write_disposition']}`"
                     )
-                if file_path.endswith(".jsonl"):
+                parsed_file = ParsedLoadJobFileName.parse(file_path)
+                if parsed_file.file_format in ["jsonl", "typed-jsonl"]:
```
rudolfix, line 233: this is way better now!

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -98,6 +102,12 @@ def run(self) -> None:
         # trade offs between different options and decided on this.
         # remote_path = f"{client.config.protocol}://{posixpath.join(dataset_path, destination_file_name)}"
         remote_path = self.make_remote_path()
+        if (
+            not self._job_client.config.as_staging_destination
+            and self._job_client.storage_version == 1
+            and remote_path.endswith(".gz")
```
rudolfix, line 108: how does it work? if this is our backward compat, we should keep the old layout (no `.gz`) forever - even if we upgrade to version 2, 3 etc.. or maybe I do not understand how it works...

What you need to do is:
- store both initial version and current version in "init" file. initial version is never upgraded and you can use it here
- OR store "disable_gz_extension" flag in "init" and use it to disable gz.
not sure what's better. I'd probably go for the initial version

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -354,9 +379,46 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
             logger.info(f"Will truncate tables {truncate_tables}")
             self.truncate_tables(list(truncate_tables))
 
+        # if the init file already exists, we return
+        if self.fs_client.exists(self.init_file_path):
```
rudolfix, line 380: this is not future proof. here you should read the version and if the version is different from the CURRENT_VERSION you should execute migration process. right now our migration is just to write a full version file into `init`.

look how `VersionedStorage` works. it is for local filesystem but pattern should be the same

rudolfix: for example if we decide to store our dlt tables in the same way as data tables we will write a migration which always ends with writing new version as a kind of commit

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -354,9 +379,46 @@ def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
...
+            return 1
+
+        try:
+            version: int = json.loads(version_info_str)["version"]
+        except (KeyError, orjson.JSONDecodeError) as exc:
+            raise ValueError(
+                f"Invalid content in {self.init_file_path}: {version_info_str!r}"
+            ) from exc
+
+        if version not in SUPPORTED_VERSIONS:
```
rudolfix: this not supported migration. exception is already in the core. see `VersionedStorage`

dlt/destinations/impl/filesystem/filesystem.py
rudolfix: (if I'm right here, our tests are lacking)

tests/pipeline/test_dlt_versions.py
```diff
@@ -220,6 +224,53 @@ def test_pipeline_with_dlt_update(test_storage: FileStorage) -> None:
...
+            # ensure duckdb reader handles compressed files without .gz extension
+            with pipeline.sql_client() as sql_client:
+                with sql_client.execute_query("SELECT * FROM issues") as curr:
+                    assert 40 == len(curr.df())
+
+            # ensure storage version 1, aka, the legacy version where .gz is not added
+            with pipeline.destination_client() as client:
+                assert isinstance(client, FilesystemClient)
+                assert client.get_storage_version() == 1
+
```
rudolfix, line 317: now write one more file, make sure still no gzip an you can query everything.

also
1. use both jsonl and csv in the pipeline. (maybe you need to modify pipeline script)
2. we need a whole new case: the one above had the files compressed before upgrade. now we need a case where files were NOT compressed before upgrade. both should work. I think you can just parametrize the test above but make sure that files are stored really uncompressed by reading them directly

tests/load/utils.py
```diff
@@ -476,15 +476,13 @@ def destinations_configs(
                 file_format="jsonl",
                 bucket_url=AWS_BUCKET,
                 extra_info="s3-authorization",
-                disable_compression=True,
```
rudolfix, line 479: why this is removed?

dlt/destinations/impl/filesystem/sql_client.py
rudolfix: kudos for spotting this

tests/load/utils.py
anuunchin: Because databricks does support compressed jsonl files:

Previously, dlt didn't support compressed jsonl files for Databricks and raised, but this is not true. Databricks is able to handle compressed files, but they need to have the .gz extension. This is now resolved and we don't need to turn of compression in test configs

rudolfix (approved):
LGTM!

PR #2839, opened 2025-07-04
Chore: Pyiceberg's python contsraint moved from project wide constraints

anuunchin:
### Description
`pyproject.toml` excluded python version 3.9.7 just because pyiceberg doesn't support it.
This was moved to pyiceberg specifically.

### DIFF
--- a/dlt/common/libs/pyiceberg.py
+++ b/dlt/common/libs/pyiceberg.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from typing import Dict, Any, List, Optional
 
 from fsspec import AbstractFileSystem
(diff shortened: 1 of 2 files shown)


dlt/common/libs/pyiceberg.py
```diff
@@ -20,6 +21,12 @@
 
 from dlt.destinations.impl.filesystem.filesystem import FilesystemClient
 
+if sys.version_info[:3] == (3, 9, 7):
```
sh-rp: Imho the uv or poetry dependency system should warn you if you try to install pyiceberg on 3.9.7, even if it is a transitive dependency, so I don't think this is needed.

pyproject.toml
```diff
@@ -173,7 +173,7 @@ sqlalchemy = [
     "alembic>1.10.0",
 ]
 pyiceberg = [
-    "pyiceberg>=0.9.1",
+    "pyiceberg>=0.9.1 ; python_version != '3.9.7'",
```
sh-rp: same comment as above. you can see in their project file that this python version is disabled: https://github.com/apache/iceberg-python/blob/main/pyproject.toml

anuunchin: owww, this makes total sense now, defo felt like i was going way overboard

sh-rp (approved):

PR #2842, opened 2025-07-07
Add DuckLake setup section to documentation

molkazhani2001:
Added DuckLake setup instructions to the MotherDuck destination docs

### DIFF
--- a/docs/website/docs/dlt-ecosystem/destinations/motherduck.md
+++ b/docs/website/docs/dlt-ecosystem/destinations/motherduck.md
@@ -90,6 +90,16 @@ CREATE SECRET my_secret IN MOTHERDUCK (
   REGION 'my-bucket-region'
 );
 ```
+
+**3. Configure `secrets.toml` in your dlt project**
+Your `secrets.toml` only needs to reference the DuckLake database and your service token:
+```toml
+[destination.motherduck.credentials]
+database = "my_ducklake"
+password = "<your token here>"
+```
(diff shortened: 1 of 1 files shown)


docs/website/docs/dlt-ecosystem/destinations/motherduck.md
```diff
@@ -68,6 +68,30 @@ More in Motherduck [documentation](https://motherduck.com/docs/key-tasks/authent
...
+**2. Register S3 Credentials with a MotherDuck Secret**
+```sql
+CREATE SECRET my_secret IN MOTHERDUCK (
+  TYPE S3,
+  KEY_ID 'my_s3_access_key',
+  SECRET 'my_s3_secret_key',
+  REGION 'my-bucket-region'
+);
+```
+With this setup, you can now load data into DuckLake tables using dlt, just like with any DuckDB destination through MotherDuck.
```
rudolfix, line 103: Please add an example of `secrets.toml` with the database configured. The point to show that you do not need to pass s3 credentials on `dlt` side

jukiewiczm:
@rudolfix / @molkazhani2001 

Do I understand correctly that in order to load new data using this DuckLake/Motherduck integration, the data still needs to go through Motherduck in a regular insert statement?

I guess it would be pretty cool to be able to upload the file directly to the bucket, and just call a metadata update in Motherduck. Seems like this could mitigate the problem of Motherduck having trouble ingesting a lot of files in parallel.

EDIT: After giving it a bit more thought I guess that would probably be a rather niche feature and probably quite hard to do safely and properly, so consider this as a loose discussion point.

Issue #2782, opened 2025-06-18
dlt.config.get() reads from secrets.toml instead of config.toml


rasnes:
### dlt version

1.12.0

### Describe the problem

When using separate `config.toml` and `secrets.toml` files as recommended in the documentation, `dlt.config.get()` returns values from `secrets.toml` instead of `config.toml`. This means configuration and secrets are not properly separated, and `dlt.config.get()` and `dlt.secrets.get()` return identical results.

This issue has been observed across multiple dlt versions (1.5.0, 1.11.0, and 1.12.0) and represents a potential security vulnerability where secrets could inadvertently leak through configuration access patterns.

### Expected behavior

- `dlt.config.get("sources.d365")` should return non-sensitive configuration from `.dlt/config.toml`
- `dlt.secrets.get("sources.d365")` should return sensitive credentials from `.dlt/secrets.toml`
- These two calls should return different data based on their respective files

**Current (incorrect) behavior:**
- Both `dlt.config.get()` and `dlt.secrets.get()` return the same data from `secrets.toml`
- `config.toml` appears to be ignored entirely

### Steps to reproduce

Create the following file structure:

**.dlt/config.toml:**
```toml
[sources.d365]
tenant_id = "test-tenant-id"
api_url = "https://example.com/api"
```

**.dlt/secrets.toml:**
```toml
[sources.d365]
client_id = "secret-client-id"
client_secret = "secret-client-secret"
```

**test_config.py:**
```python
import dlt

print(f"dlt version: {dlt.__version__}")

# This should return config from config.toml
config_data = dlt.config.get("sources.d365", {})
print(f"Config data: {config_data}")

# This should return credentials from secrets.toml  
secrets_data = dlt.secrets.get("sources.d365", {})
print(f"Secrets data: {secrets_data}")

# These should be different, but currently return identical results
print(f"Are they identical? {config_data == secrets_data}")
```

**Expected output:**
```
Config data: {'tenant_id': 'test-tenant-id', 'api_url': 'https://example.com/api'}
Secrets data: {'client_id': 'secret-client-id', 'client_secret': 'secret-client-secret'}
Are they identical? False
```

**Actual output:**
```
Config data: {'client_id': 'secret-client-id', 'client_secret': 'secret-client-secret'}
Secrets data: {'client_id': 'secret-client-id', 'client_secret': 'secret-client-secret'}
Are they identical? True
```

### Operating system

macOS

### Runtime environment

Virtual Machine

### Python version

[...]

PR #2853, opened 2025-07-07
fix: 2782 name collision `dlt.config.get()` and `dlt.secrets.get()` accessors

zilto:
## Description
This partially fixes the issue #2782. The repro reported in the initial issue is now a test case

```toml
# .dlt/config.toml
[sources.d365]
tenant_id = "test-tenant-id"
api_url = "https://example.com/api"
```

```toml
# .dlt/secrets.toml
[sources.d365]
client_id = "secret-client-id"
client_secret = "secret-client-secret"
```

```python
# test_config.py
import dlt

expected_config = {"tenant_id": "test-tenant-id", "api_url": "https://example.com/api"}
expected_secrets = {"client_id": "secret-client-id", "client_secret": "secret-client-secret"}

# BEFORE
dlt.config.get("sources.d365")  != expected_config  # BUG
dlt.config.get("sources.d365")  == expected_secrets  # BUG
dlt.secrets.get("sources.d365")  == expected_secrets   # OK

# NOW
dlt.config.get("sources.d365")  == expected_config  # OK
dlt.secrets.get("sources.d365")  == expected_config   # OK
```

## Changes
- The current changes are minimal and only involves changing the order used to load the default providers. - It **may** break pipelines, but these would be relying on undefined internal behavior.
- These changes should have no effect on use of `dlt.config.value` and `dlt.secrets.value` which are the main usage pattern that's promoted (vast majority)

## Additional Context

However, there's a remaining issue with the following pattern (continuing previous snippets)
```python
# config accesses secrets
dlt.secrets.get("sources.d365.client_secret") == expected_secrets["client_secret"]
dlt.config.get("sources.d365.client_secret") == expected_secrets["client_secret"]

# secrets only accesses secrets
dlt.config.get("sources.d365.tenant_id") == expected_config["tenant_id"]
dlt.secrets.get("sources.d365.tenant_id") is None
```
While this would be a simple fix, it's likely to break existing pipelines and configurations. In particular, this would potentially affect users who built custom `ConfigProvider` or derived from existing classes.

### DIFF
--- a/dlt/common/runtime/run_context.py
+++ b/dlt/common/runtime/run_context.py
@@ -28,8 +28,8 @@ def get_default_initial_providers(
 ) -> list[ConfigProvider]:
     providers = [
         EnvironProvider(),
-        ConfigTomlProvider(settings_dir, global_dir),
         SecretsTomlProvider(settings_dir, global_dir),
+        ConfigTomlProvider(settings_dir, global_dir),
     ]
     return providers
 
--- a/tests/common/configuration/test_accessors.py
+++ b/tests/common/configuration/test_accessors.py
@@ -7,11 +7,7 @@ import dlt
 from dlt.common import json
 from dlt.common.configuration.exceptions import ConfigFieldMissingException
 
-from dlt.common.configuration.providers import (
-    EnvironProvider,
-    ConfigTomlProvider,
-    SecretsTomlProvider,
-)
+from dlt.common.configuration.providers import EnvironProvider
 from dlt.common.configuration.providers.toml import (
     CONFIG_TOML,
     SECRETS_TOML,
@@ -325,19 +321,25 @@ def test_explicit_get_resolution_order(tmp_path):
     # patch the `.dlt` to use the `tmp_path/.dlt` instead of the `tests/.dlt` loaded when
     # launching the pytest run
     with reset_providers(str(dlt_dir)):
-        assert dlt.config.get("sources.d365") == expected_config
+        # TODO dlt==2.0; I would expect this:
+        # assert dlt.config.get("sources.d365") == expected_config
+        assert dlt.config.get("sources.d365") == expected_secrets  # odd
         assert dlt.secrets.get("sources.d365") == expected_secrets
 
-        assert dlt.config.get("sources") == {"d365": expected_config}
+        # TODO I would expect this:
+        # assert dlt.config.get("sources") == {"d365": expected_config}
+        assert dlt.config.get("sources") == {"d365": expected_secrets}  # odd
         assert dlt.secrets.get("sources") == {"d365": expected_secrets}
 
-        # TODO dlt==2.0; dlt.config.get() current allows to retrieve secrets
-        # this behavior is undesirable, but this would be a breaking change
+        # TODO dlt==2.0; I would expect this:
+        # assert dlt.config.get("sources.d365.client_id") is None
+        assert dlt.config.get("sources.d365.client_id") == expected_secrets["client_id"]  # odd
         assert dlt.secrets.get("sources.d365.client_id") == expected_secrets["client_id"]
-        assert dlt.config.get("sources.d365.client_id") == expected_secrets["client_id"]
 
-        assert dlt.secrets.get("sources.d365.client_secret") == expected_secrets["client_secret"]
+        # TODO dlt==2.0; I would expect this:
+        # assert dlt.config.get("sources.d365.client_secret") is None
         assert dlt.config.get("sources.d365.client_secret") == expected_secrets["client_secret"]
+        assert dlt.secrets.get("sources.d365.client_secret") == expected_secrets["client_secret"]
 
         # `dlt.secrets.get()` cannot access config value
         assert dlt.config.get("sources.d365.tenant_id") == expected_config["tenant_id"]
--- a/tests/common/configuration/test_toml_provider.py
+++ b/tests/common/configuration/test_toml_provider.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, Type
 import datetime  # noqa: I251
 from unittest.mock import Mock
 
+import tomlkit
 import dlt
 from dlt.common import pendulum, json
 from dlt.common.configuration import configspec, ConfigFieldMissingException, resolve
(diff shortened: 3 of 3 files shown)


tests/common/configuration/test_toml_provider.py
```diff
@@ -95,7 +96,9 @@ def test_config_provider_order(toml_providers: ConfigProvidersContainer, environ
     def single_val(port=None):
         return port
 
+    # TODO there's a problem here
     # secrets have api.port=1023 and this will be used
+    assert dlt.secrets.get("api.port") == 1023
     assert single_val(dlt.secrets.value) == 1023
```
zilto, line 122: Those are the annoying lines. After changing the order of the `default_initial_providers()` to solve name collision with `dlt.secrets.get()` creates the same name collision for `dlt.secrets.value`.

This test fails. The `config.toml` includes value 1024 and `secrets.toml` has value 1023
- line 101: `dlt.secrets.get()` will correctly get the expected value `1023`
- line 102: `dlt.secrets.value` will incorrectly get 1024 (AssertionError)

dlt/common/runtime/run_context.py
```diff
@@ -23,6 +23,17 @@
 DOT_DLT = os.environ.get(known_env.DLT_CONFIG_FOLDER, ".dlt")
 
 
+def get_default_initial_providers(
+    settings_dir: str, global_dir: Optional[str] = None
+) -> list[ConfigProvider]:
+    providers = [
+        EnvironProvider(),
+        ConfigTomlProvider(settings_dir, global_dir),
```
rudolfix: it is a breaking change if you swap it. also it is not necessary...

tests/common/configuration/test_accessors.py
```diff
@@ -287,3 +288,57 @@ def test_config_accessor_repr() -> None:
...
+    with reset_providers(str(dlt_dir)):
+        assert dlt.config.get("sources.d365") == expected_config
+        assert dlt.secrets.get("sources.d365") == expected_secrets
+
+        assert dlt.config.get("sources") == {"d365": expected_config}
+        assert dlt.secrets.get("sources") == {"d365": expected_secrets}
+
+        # TODO dlt==2.0; dlt.config.get() current allows to retrieve secrets
+        # this behavior is undesirable, but this would be a breaking change
+        assert dlt.secrets.get("sources.d365.client_id") == expected_secrets["client_id"]
```
rudolfix, line 337:
```py
def _get_value(self, field: str, type_hint: Type[Any] = None) -> Tuple[Any, List[LookupTrace]]:
```
`dlt.secrets.get()` uses simplified resolution method. it finds a dictionary for node `sources.d365` in `secrets.toml` and returns it without trying to resolve the other values.

this is because we do not know the type so simplified resolved does not know what else to look for. you can fix cases like above by merging dictionaries found in all providers (in correct order so earlier provider overrides later/lower provider)

i'm also fine if we do "won't fix". at least we'll have a few additional tests

tests/common/configuration/test_toml_provider.py
rudolfix: this happens because here we have a real collision and the resolution depends on provider precedence. 
tests/common/cases/configuration/.dlt/secrets.toml
vs.
tests/common/cases/configuration/.dlt/config.toml

tests/common/configuration/test_accessors.py
zilto: > you can fix cases like above by merging dictionaries found in all providers (in correct order so earlier provider overrides later/lower provider)

Given the default order `Environment -> Secrets -> Config`, wouldn't this approach still reveal secrets on name collision?

### issue
The issue reported by the user is a larger design decision that we can't change:
- If you do `dlt.config`, it includes providers: environment, secrets, config
- If you do `dlt.secrets`, it includes providers: environment, secrets

The behavior desired by the  is that:`dlt.config` can never retrieve values from `secrets.toml`. I believe others will have this intuition

dlt/common/runtime/run_context.py
```diff
@@ -23,6 +23,17 @@
...
 
+def get_default_initial_providers(
+    settings_dir: str, global_dir: Optional[str] = None
+) -> list[ConfigProvider]:
+    providers = [
+        EnvironProvider(),
+        SecretsTomlProvider(settings_dir, global_dir),
+        ConfigTomlProvider(settings_dir, global_dir),
+    ]
+    return providers
```
zilto, line 34: Latest commit reverts order to `master/devel`

zilto, line 34: Latest commit reverts order to `master/devel`

tests/common/configuration/test_accessors.py
```diff
@@ -287,3 +284,63 @@ def test_config_accessor_repr() -> None:
...
+
+        # TODO dlt==2.0; I would expect this:
+        # assert dlt.config.get("sources.d365.client_secret") is None
+        assert dlt.config.get("sources.d365.client_secret") == expected_secrets["client_secret"]
+        assert dlt.secrets.get("sources.d365.client_secret") == expected_secrets["client_secret"]
+
+        # `dlt.secrets.get()` cannot access config value
+        assert dlt.config.get("sources.d365.tenant_id") == expected_config["tenant_id"]
+        assert dlt.secrets.get("sources.d365.tenant_id") is None
```
zilto, line 346: @rudolfix I didn't fully understand your previous comments. Please write `assert` statements of what you expect. To me, the current passing assertions are bizarre.

tests/common/configuration/test_toml_provider.py
```diff
@@ -395,26 +418,43 @@ def test_write_value(toml_providers: ConfigProvidersContainer) -> None:
...
+
+    provider: SecretsTomlProvider = toml_providers.providers[2]  # type: ignore[assignment]
+    provider._config_doc = {}
+    provider.set_value(key, dataclasses.asdict(creds), None)
+    resolved_config = resolve.resolve_configuration(WithCredentialsConfiguration(), sections=(key,))
+
+    # TODO this should return an anonymized string if hinted with `TSecretValue`
+    # this currently leaks the content of `tests/common/cases/configuration/.dlt/secrets.toml`
+    assert resolved_config.credentials.secret_value == "2137"
+    assert resolved_config.credentials.secret_value != secret_value
```
zilto, line 438: Unexpected assertion. Please comment

tests/common/configuration/test_toml_provider.py
```diff
@@ -95,7 +97,28 @@ def test_config_provider_order(toml_providers: ConfigProvidersContainer, environ
...
+    # Likely scenario:
+    # 1. someone adds a new config point
+    # 2. it collides with an existing secret (no one will review all secrets on every config edit)
+    # 3. secret is leaked
+    assert config_raw["api"]["port"] == 1024  # type: ignore[index]
+    assert dlt.config.get("api.port") == 1023  # this gets the secrets value
+    assert single_val(dlt.config.value) == 1023  # this gets the secrets value
+
+    assert secrets_raw["api"]["port"] == 1023  # type: ignore[index]
+    assert dlt.secrets.get("api.port") == 1023
```
zilto, line 121: Unexpected assertions. Please comment

tests/common/configuration/test_toml_provider.py
```diff
@@ -395,26 +418,43 @@ def test_write_value(toml_providers: ConfigProvidersContainer) -> None:
...
-
-        # use dataclass to dict to recursively convert base config to dict
-        import dataclasses
-
-        provider.set_value("written_creds", dataclasses.asdict(creds), None)
-        # resolve config
-        resolved_config = resolve.resolve_configuration(
-            WithCredentialsConfiguration(), sections=("written_creds",)
-        )
-        assert resolved_config.credentials.secret_value == "***** ***"
```
zilto, line 417: The new tests `test_set_spec_value_on_secret_provider` and `test_set_spec_value_on_config_provider` should cover the same behavior as this, but it doesn't. In this old test, it suggests that

```python
# config
resolved_config.credentials.secret_value == "***** ***"

# secrets
resolved_config.credentials.secret_value == "***** ***"
```

But that is not the case. Properly isolating the test shows that
```python
# config
resolved_config.credentials.secret_value == "***** ***"

# secrets
# plain text secret, picked up from `secrets.toml`
# it's somehow not overwritten by the body of the test function
resolved_config.credentials.secret_value == "2137"
```

tests/common/configuration/test_toml_provider.py
```diff
@@ -395,26 +418,43 @@ def test_write_value(toml_providers: ConfigProvidersContainer) -> None:
...
+    creds.credentials = credentials
+
+    provider: ConfigTomlProvider = toml_providers.providers[1]  # type: ignore[assignment]
+    provider._config_doc = {}
+    provider.set_value(key, dataclasses.asdict(creds), None)
+
+    # TODO this should raise that `dlt.config.get()` doesn't allow access to secrets
+    # with pytest.raises(ValueNotSecretException):
+    resolved_config = resolve.resolve_configuration(WithCredentialsConfiguration(), sections=(key,))
+    assert resolved_config.credentials.secret_value == secret_value
```
zilto, line 457: Should this raise?

zilto:
Closing. Will re-open when a clear fix is decided

PR #2865, opened 2025-07-11
fix filesystem config section

sh-rp:
### Description
Fixes the configuration resolution for the filesystem readers and adds basic tests for all sources

### DIFF
--- a/tests/sources/filesystem/test_config_sections.py
+++ b/tests/sources/filesystem/test_config_sections.py
@@ -1,0 +1,18 @@
+import pytest
+
+from dlt.sources.filesystem import filesystem, read_parquet
+from dlt.common.configuration.exceptions import ConfigFieldMissingException
+
+
+def test_config_sections_resolution():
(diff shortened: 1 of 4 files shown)


tests/sources/filesystem/test_config_sections.py
```diff
@@ -0,0 +1,18 @@
...
+def test_config_sections_resolution():
+    filesystem_resource = filesystem(file_glob="**/*.parquet")
+    filesystem_pipe = filesystem_resource | read_parquet()
+
+    with pytest.raises(ConfigFieldMissingException) as exc_info:
+        list(filesystem_pipe)
+
+    # NOTE: we check that the first trace related to filesystem has the correct
+    # sections set
+    assert list(exc_info.value.traces.values())[1][0].key.startswith(
```
sh-rp, line 16: this is not the nicest test, I would've prefered to inspect the sections attribute of the traces, but they are not set and I, for now, did not invest the time to find out why and what thedifference between explicit sections and embedded sections..

PR #2867, opened 2025-07-11
rest_api: Redact secrets in logs, add configurable response body in errors

### LINKED ISSUE #2855: Exhance error messages in RestAPI Client
### Feature description

Directly log or return the entire response body from a REST request that causes the pipeline step to fail during extraction. Debugging the RestAPI sources is a bit of a pain right now, particularly since the request or response objects aren't available in the debugger and the large volume of nested decorators abstract any calls to perform the actual lookup, so producing a breakpoint on the response requires drilling into the internals of the source. 

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

Most of our DLT jobs are RestAPI source based and occasionally see transient issues that cause API responses to vary from expectations. The logs produced in a crash are not always helpful as they can mask the actual response. For example if we pull an API endpoint for a vendor and our contract changes so we no longer can hit a particular endpoint, we start seeing 'status:400 bad request' responses but no details. It requires using Postman with the returned full URL to get the actual message body that says why. 

Also during development it can be difficult to understand what's happening. One endpoint requires a list of columns we want returned from the response as a query parameter. I missed a comma when adding a new column and the following response was dumped in the exception.

```
File "/Users/.../Development/ourpath/.venv/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 497, in extract
    raise PipelineStepFailed(
dlt.pipeline.exceptions.PipelineStepFailed: Pipeline execution failed at `step=extract` when processing package with `load_id=1751987315.654639` with exception:

<class 'dlt.extract.exceptions.ResourceExtractionError'>
In processing pipe `mypipe`: extraction of resource `myresource` in `generator` `paginate_resource` caused an exception: 400 Client Error: Bad Request for url: https://endpoint.com/WITHMASSIVEQUERYSTRING....
python-BaseException
``

Running the long bad url in Postman the body of the message has:
`Invalid Column: column_1column2`

Which immediately made me know what the problem was that two columns weren't separated correctly. 
```
### Proposed solution

Currently it looks like requests `raise_for_status` function is used to trigger the alerts which prevents custom responses. I would propose checking the status_code for 4XX/5XX responses and raising the error in DLT so that the Response body can be returned as well. 

### Related issues

Additionally I will file another issue as this creates a *security* issue because API_KEY based authentication with the location as Query Params dumps the Request URL to console logs during a failure. This results in the API key being plaintext in logs.

### LINKED ISSUE #2856: Secrets written to logs with RestSource + API Keys in Query string
### dlt version

dlt 1.12.3

### Describe the problem

When using the APIKey based authentication, for the RestSource connector, where the location is 'query' any response that causes `raise_for_status()` to raise an exception dumps the URL of the full Request into the application logs. 

```
    auth = APIKeyAuth(name="api_key", api_key=secret_manager('api_key'), location="query")
```

Example dump:
```
  File "/Users/me/Development/app/.venv/lib/python3.11/site-packages/dlt/pipeline/pipeline.py", line 497, in extract
    raise PipelineStepFailed(
dlt.pipeline.exceptions.PipelineStepFailed: Pipeline execution failed at `step=extract` when processing package with `load_id=1751987266.2779489` with exception:

<class 'dlt.extract.exceptions.ResourceExtractionError'>
In processing pipe `mypipeline`: extraction of resource `myresource` in `generator` `paginate_resource` caused an exception: 400 Client Error: Bad Request for url: https://domain.com/endpoint/long&query=string&example&api_key=123456ABCDEFGHIJOK
```

### Expected behavior

Ideally the logs would redact the API key. 

### Steps to reproduce

Create a REST API source where the authentication method uses API Key auth, with the location as 'query'. Perform a bad request and observe logs. 

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

RestSource

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

Likely requires not using `raise_for_status()` and instead handling the 4XX/5XX error codes directly. This would also allow for the error response's body to be logged like in feature request #2855

burnash:
This PR implements:
- secret redaction in logs/errors
- adds configurable response body display to errors (`http_show_error_body`)

Resolves #2855 
Fixes #2856


dlt/sources/helpers/rest_client/client.py
```diff
@@ -143,6 +142,18 @@ def _send_request(self, request: Request, **kwargs: Any) -> Response:
         send_kwargs.update(**kwargs)  #  type: ignore[call-arg]
         return self.session.send(prepared_request, **send_kwargs)
 
+    def _log_request(self, request: Request, prepared_url: str) -> None:
+        # XXX: Use logger.isEnabledFor(logging.DEBUG) once dlt logger is fixed
```
rudolfix, line 151: should we fix something in the `logger` module?

dlt/sources/rest_api/__init__.py
```diff
@@ -55,15 +56,7 @@
 
 PARAM_TYPES: List[ParamBindType] = ["incremental", "resolve"]
```
rudolfix, line 57: we need to cleanup `__init__` those constants are not user-facing IMO

tests/sources/helpers/rest_client/test_client.py
```diff
@@ -544,3 +544,162 @@ def test_overwrite_headers(self, mocker, **kwargs) -> None:
         )
         response = rest_client.get("https://my.url", headers={"Accept": "my/mimetype"})
         assert response.json() == expected
+
+
+class TestSecretRedaction:
```
rudolfix, line 549: this is the only place we have tests in class layout. probably because they got contributed to verified sources and moved here?

dlt/sources/helpers/rest_client/client.py
burnash: Yes, it'd be good, currently logger proxies all module calls with:

```python
def __getattr__(name: str) -> LogMethod:
    """Forwards log method calls (debug, info, error etc.) to LOGGER"""

    def wrapper(msg: str, *args: Any, **kwargs: Any) -> None:
        if LOGGER:
            # skip stack frames when displaying log so the original logging frame is displayed
            stacklevel = 2
            if name == "exception":
                # exception has one more frame
                stacklevel = 3
            getattr(LOGGER, name)(msg, *args, **kwargs, stacklevel=stacklevel)

    return wrapper
```

so when I call logger.isEnabledFor() it fails with `stacklevel` is not an argument. I think we may want to scope the proxying.

tests/sources/helpers/rest_client/test_client.py
burnash: Yes, I think so. Do you see any downside to this structure?

burnash: Paginator tests are also groupped with classes, https://github.com/dlt-hub/dlt/blob/devel/tests/sources/helpers/rest_client/test_paginators.py
I think I put it together this way

rudolfix (approved):
LGTM

PR #2868, opened 2025-07-11
removes init files from dlt tables in filesystem

rudolfix:
### Description
* removes init files from dlt tables in filesystem
* creates folders for tables only if there's a schema migration (note: on buckets create folder is a noop)
* handles response/request validation in botocore in the credentials
* fixes a bug in synthesizing unpickler that could not handle enums (it is here because I've got a test error in dbt runner and I discovered that)


dlt/common/runners/synth_pickle.py
```diff
@@ -8,7 +8,14 @@
...
 
-class MissingUnpickledType:
+class _ClassMeta(type):
+    """Allows for type to have attributes that are not defined in the class."""
+
+    def __getattr__(cls, name: str) -> Any:
+        return name
+
+
+class MissingUnpickledType(object, metaclass=_ClassMeta):
```
anuunchin, line 18: This is good to know

anuunchin (approved):

PR #2870, opened 2025-07-13
fix: typing for updated datasets and relations `Protocols` 

zilto:
### Description
Follow-up to [this discussion](https://github.com/dlt-hub/dlt/pull/2718#discussion_r2196295544) 

- As mentioned, protocols need to directly inherit from `typing.Protocol` ([source](https://mypy.readthedocs.io/en/stable/protocols.html#defining-subprotocols-and-subclassing-protocols), reported by pylance in the screenshots I shared in the previous discussion). 
- mypy didn't fail because `Relation` was inadvertently not a `Protocol`. Similarly, some tests were making assumptions against `Relation` with `isinstance()` calls (which wouldn't be possible if it was a proper protocol). 
- `DataAccess` was implementing method such as `__init__` and property setters, which is not possible on a `Protocol`. To maintain the desired approach, I created a `DBApiCursorProtocol` that is inherited by the abstract base class `DBApiCursor`.
- I removed the uncessary `else` block after `if TYPE_CHECKING`
- import `from __future__ import annotations` to avoid circular imports

### Additional context
I still in favor of removing this abstraction layer as long as we have a single set of `Dataset` and `Relation` implementations. Having to modify their superclass / protocols makes maintenance more complex

### DIFF
--- a/dlt/common/destination/dataset.py
+++ b/dlt/common/destination/dataset.py
@@ -1,44 +1,35 @@
+from __future__ import annotations
+
+import abc
 from types import TracebackType
-
-
 from typing import (
-    Optional,
-    Sequence,
-    Union,
-    List,
-    Any,
-    Generator,
     TYPE_CHECKING,
-    Protocol,
-    Tuple,
-    AnyStr,
-    Literal,
     overload,
+    Any,
+    AnyStr,
+    Generator,
+    Literal,
+    Optional,
+    Protocol,
+    Sequence,
     Type,
+    Union,
 )
 
-from sqlglot.schema import Schema as SQLGlotSchema
 import sqlglot.expressions as sge
+from sqlglot.schema import Schema as SQLGlotSchema
 from sqlglot.expressions import ExpOrStr as SqlglotExprOrStr
 
-from dlt.common.typing import Self, Generic, TypeVar
+from dlt.common.typing import Self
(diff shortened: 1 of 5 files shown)


dlt/common/destination/dataset.py
```diff
@@ -1,44 +1,35 @@
...
+from dlt.common.schema.typing import TTableSchema, TTableSchemaColumns
 from dlt.common.libs.sqlglot import TSqlGlotDialect
-from dlt.common.schema.typing import TTableSchema
 
 if TYPE_CHECKING:
     from dlt.common.libs.pandas import DataFrame
     from dlt.common.libs.pyarrow import Table as ArrowTable
     from dlt.helpers.ibis import BaseBackend as IbisBackend, Table as IbisTable, Expr as IbisExpr
     from dlt.common.destination.client import SupportsOpenTables
-else:
```
rudolfix, line 35: indeed it was not necessary. it looks like at runtime we have literal string values as types... probably `ForwardRef` in `typing` module

rudolfix:
@sh-rp should look here as well!

sh-rp:
Thanks @zilto for fixing this! My plan was to allow any object that conforms to "Relation" to be able to be consumed in the transformations, but probably nobody is going to implement their own Relation anyway at this point so for me this is good.

sh-rp (approved):

PR #2873, opened 2025-07-13
fixes range_start=open in incremental

rudolfix:
### Description
Setting `range_start` to `open` in `Incremental` class could skip additional items in the range. `pyarrow` and `pandas` incrementals were not accepted.

more tests for open/closed ranges were added


dlt/extract/incremental/transform.py
```diff
@@ -239,39 +252,43 @@ def __call__(
             ) from ex
         # new_value is "less" or equal to last_value (the actual max)
         if last_value == new_value:
-            if self.range_start == "open":
-                # We only want greater than last_value
-                return None, False, False
             # use func to compute row_value into last_value compatible
             processed_row_value = last_value_func((row_value,))
+
+            if self.range_start == "open" and processed_row_value == self.start_value:
```
rudolfix, line 253: here's the fix

dlt/extract/incremental/transform.py
```diff
@@ -129,8 +130,7 @@ def _adapt_if_datetime(row_value: Any, last_value: Any) -> Any:
             row_value = pendulum.instance(row_value).in_tz("UTC")
         return row_value
 
-    @property
-    def deduplication_disabled(self) -> bool:
+    def compute_deduplication_disabled(self) -> bool:
```
burnash, line 128: Should we leave a deprecated alias here?

dlt/extract/incremental/transform.py
```diff
@@ -97,10 +98,10 @@ def compute_unique_value(
...
     ) -> str:
         try:
-            assert not self.deduplication_disabled, (
-                f"{self.resource_name}: Attempt to compute unique values when deduplication is"
-                " disabled"
-            )
+            # assert not self.deduplication_disabled, (
+            #     f"{self.resource_name}: Attempt to compute unique values when deduplication is"
+            #     " disabled"
+            # )
```
burnash:
```suggestion
```

dlt/extract/incremental/__init__.py
```diff
@@ -201,7 +201,7 @@ def primary_key(self, value: str) -> None:
         self._primary_key = value
         if self._transformers:
             for transform in self._transformers.values():
-                transform.primary_key = value
+                transform._primary_key = value
```
burnash: Could you elaborate why property setter is bypassed here? No sure how likely is the case where primary_key is modified after Incremental has been created but this could result in `transformer.boundary_deduplication` not recomputed. As far as I can see it's only done in transformer's primary_key setter.

rudolfix: it is a bug. thanks for spotting this. it is a result of automatic refactoring and surely we do not have a right test. I'll fix it now

dlt/extract/incremental/transform.py
rudolfix: transforms are internal object that users are not interacting with. we do not need to deprecate.

PR #2876, opened 2025-07-14
Add workspace extra and rename marimo app to "pipeline dashboard"

sh-rp:
### Description
This PR:
* Adds a new workspace extra, reduces the mount of dependencies in the marimo group and uses this extra in tests
* Renames the dlt marimo app to `pipeline dashboard` and updates the cli commands and docs accordingly
* Renames all files and symbols still referring to `studio` to `dashboard`

Test checklist:
* Unit tests pass
* e2e tests pass
* Manual check of cli commands for starting and editing dashboard work


pyproject.toml
```diff
@@ -181,6 +181,16 @@ pyiceberg = [
 postgis = [
     "psycopg2-binary>=2.9.1"
 ]
+workspace = [
+    "duckdb<1.2.1",
```
rudolfix: is there a problem with newer version? we should be compatible with our duckdb extra. so the newest one should be allowed (I think it is >0.9)

pyproject.toml
```diff
@@ -181,6 +181,16 @@ pyiceberg = [
...
+    "ibis-framework[duckdb, postgres, bigquery, snowflake, mssql, clickhouse, databricks]>=10.5.0 ; python_version >= '3.10' and python_version < '3.13'",
+    "ibis-framework[duckdb, postgres, bigquery, snowflake, clickhouse, databricks]>=10.5.0 ; python_version >= '3.13'",
+    "pyarrow>=16.0.0; python_version < '3.13'",
+    "pyarrow>=18.0.0 ; python_version >= '3.13'",
+    "pandas>2.1 ; python_version >= '3.12'",
+    "pandas<2.1 ; python_version < '3.12'",
+    "marimo>=0.13.6",
```
rudolfix: we need 0.14+I think?

pyproject.toml
```diff
@@ -181,6 +181,16 @@ pyiceberg = [
 postgis = [
     "psycopg2-binary>=2.9.1"
 ]
+workspace = [
+    "duckdb<1.2.1",
+    "ibis-framework[duckdb, postgres, bigquery, snowflake, mssql, clickhouse, databricks]>=10.5.0 ; python_version >= '3.10' and python_version < '3.13'",
```
rudolfix: it is just ibis with duckdb. everything else people need to figure out

rudolfix (changes requested):
extra needs a little tweaking

pyproject.toml
sh-rp: The newer versions have this iceberg bug, no? But I put it to >0.9 now.

pyproject.toml
sh-rp: I think I built this with 0.13, but I upgraded it anyway.

sh-rp:
common test pass (except for one exception on 3.11, but the relevant e2e tests pass on other versions) so merging as discussed

PR #2879, opened 2025-07-14
fix: patch connectorx timestamp handling; upgrade from `0.4.1` to `0.4.3` 

zilto:
## Description
Upgrade connectorx dependency from `<=0.4.1` to `>=0.4.3`.  There were 2 breaking changes in `0.4.2` and `0.4.3`:
  - the removal of `return_type="arrow2"` (the default in dlt)
  - changed timestamp time zone format from UTC to offset-based.

This PR fixes both issues.

### DIFF
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -1102,7 +1102,7 @@ def add_arrow_metadata(
 
 
 def _has_offset_timezones(item: Union[pyarrow.Table, pyarrow.RecordBatch]) -> bool:
-    """Eager check to identify it the table contains fields with offset-based timezones."""
+    """Eager check to identify if the table contains fields with offset-based timezones."""
     for field in item.schema:
         if getattr(field.type, "tz", "").startswith(("+", "-")):
             return True
@@ -1150,10 +1150,8 @@ def _convert_offset_timezones_field(field: pyarrow.Field) -> pyarrow.Field:
     return field
 
 
-def _convert_offset_timezones_table(
-    item: Union[pyarrow.Table, pyarrow.RecordBatch]
-) -> Union[pyarrow.Table, pyarrow.RecordBatch]:
+def _convert_offset_timezones_table(item: TAnyArrowItem) -> TAnyArrowItem:
     """Convert fields that use offset-based timezones to IANA timezones"""
     new_fields = [_convert_offset_timezones_field(field) for field in item.schema]
-    new_schema = pyarrow.schema(new_fields)
+    new_schema = pyarrow.schema(new_fields, metadata=item.schema.metadata)
     return item.cast(new_schema)
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -1101,16 +1101,24 @@ def add_arrow_metadata(
         return pyarrow.RecordBatch.from_arrays(item.columns, schema=new_schema)
 
 
-def _has_offset_timezones(item: Union[pyarrow.Table, pyarrow.RecordBatch]) -> bool:
-    """Eager check to identify if the table contains fields with offset-based timezones."""
-    for field in item.schema:
-        if getattr(field.type, "tz", "").startswith(("+", "-")):
+def _type_has_offset_timezone(type_: pyarrow.DataType) -> bool:
+    tzinfo = getattr(type_, "tz", None)
+    # check if `tz` attribute is a string if found
+    if isinstance(tzinfo, str):
+        # do a check against `tz` values that are strings
+        if tzinfo.startswith(("+", "-")):
             return True
     return False
 
 
-# NOTE could use an explicit mapping of ~30 offset to IANA timezones
-# would be less error-prone and easy to maintain.
+def _table_has_offset_timezones(item: TAnyArrowItem) -> bool:
+    """Eager check to identify if the table contains fields with offset-based timezones."""
+    for field in item.schema:
+        if _type_has_offset_timezone(field.type):
+            return True
+    return False
+
+
 def _offset_to_iana(offset_tzinfo: str) -> str:
     """Convert an offset-based time zone to IANA time zone.
 
@@ -1125,9 +1133,14 @@ def _offset_to_iana(offset_tzinfo: str) -> str:
     hours, _, minutes = offset_tzinfo[1:].partition(":")
     sign = "-" if offset_tzinfo[0] == "+" and hours != "00" else "+"
     if int(minutes) != 0:
-        raise ValueError(
-            "Received time zone defined as an offset with fractional hours. This is currently not"
-            f" supported. Time zone: {offset_tzinfo}"
+        from dlt.common import logger
+
+        # rounding logic: if minutes are <=30, truncate minutes; if >30, increment hour
+        hours = int(hours) + round(int(minutes) / 60)
+        logger.warn(
+            f"dlt had to round and convert time zone from `{offset_tzinfo}` to `'Etc/GMT{sign}{int(hours)}'`."
+            " If you're using `connectorx`, switch to another backend to avoid conversion errors."
+            " Read more: https://github.com/apache/arrow/issues/47043"
         )
     return f"Etc/GMT{sign}{int(hours)}"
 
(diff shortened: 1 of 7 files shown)


dlt/common/libs/pyarrow.py
```diff
@@ -1099,3 +1099,61 @@ def add_arrow_metadata(
         return pyarrow.Table.from_arrays(item.columns, schema=new_schema)
     else:  # RecordBatch
         return pyarrow.RecordBatch.from_arrays(item.columns, schema=new_schema)
+
+
+def _has_offset_timezones(item: Union[pyarrow.Table, pyarrow.RecordBatch]) -> bool:
+    """Eager check to identify it the table contains fields with offset-based timezones."""
```
anuunchin:
```suggestion
    """Eager check to identify if the table contains fields with offset-based timezones."""
```

tests/load/sources/sql_database/test_helpers.py
anuunchin, line 1: wondering if we're importing this file somewhere and using the additional imports here 👀

dlt/sources/sql_database/helpers.py
```diff
@@ -338,7 +338,7 @@ def _unwrap(table: TDataItem) -> TDataItem:
         column = pc.replace_with_mask(
             column,
             pc.equal(column, "null").combine_chunks(),
-            pa.scalar(None, pa.large_string()),
```
anuunchin, line 341: This is good to know, I wasn't aware of pa.large_string() at all 👀

dlt/common/libs/pyarrow.py
```diff
@@ -1099,3 +1099,61 @@ def add_arrow_metadata(
...
+            new_tzinfo = _offset_to_iana(tzinfo)
+            new_type = pyarrow.timestamp(field.type.unit, tz=new_tzinfo)
+            new_field = pyarrow.field(field.name, type=new_type)
+            return new_field
+
+    return field
+
+
+def _convert_offset_timezones_table(
+    item: Union[pyarrow.Table, pyarrow.RecordBatch]
```
anuunchin: Just remembered we have `TAnyArrowItem = Union[pyarrow.Table, pyarrow.RecordBatch]`

dlt/common/libs/pyarrow.py
```diff
@@ -1099,3 +1099,61 @@ def add_arrow_metadata(
...
+
+def _convert_offset_timezones_field(field: pyarrow.Field) -> pyarrow.Field:
+    """Convert an offset-based timezone to IANA timezone for a single field.
+    This is a no-op for non-timestamp fields. Timestamp unit is preserved.
+    """
+    if not pyarrow.types.is_timestamp(field.type):
+        return field
+
+    tzinfo = getattr(field.type, "tz", None)
+    if isinstance(tzinfo, str):
```
anuunchin: A bit above, I noted down the following for myself
```
if getattr(field.type, "tz", "")
```
which I thought is very elegant - or is it not since here `None` is used instead 🤔

anuunchin (changes requested):

dlt/common/libs/pyarrow.py
```diff
@@ -1099,3 +1099,61 @@ def add_arrow_metadata(
...
+
+    return field
+
+
+def _convert_offset_timezones_table(
+    item: Union[pyarrow.Table, pyarrow.RecordBatch]
+) -> Union[pyarrow.Table, pyarrow.RecordBatch]:
+    """Convert fields that use offset-based timezones to IANA timezones"""
+    new_fields = [_convert_offset_timezones_field(field) for field in item.schema]
+    new_schema = pyarrow.schema(new_fields)
```
rudolfix: this is not preserving existing pyarrow schema metadata. there are many other functions destroying metadata in this module.  I'd at least write a TODO here because we'll be cleaning this module and testing for metadata propagation soon.

dlt/common/libs/pyarrow.py
```diff
@@ -1099,3 +1099,61 @@ def add_arrow_metadata(
...
+
+    Facts:
+    - Time zones go from `-12:00` to `+14:00`
+    - Neutral time zone is `+00:00`
+    - The sign used in UTC is reversed for GMT; for example: UTC+05:00 == Etc/GMT-5
+    ref: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones
+    """
+    hours, _, minutes = offset_tzinfo[1:].partition(":")
+    sign = "-" if offset_tzinfo[0] == "+" and hours != "00" else "+"
+    if int(minutes) != 0:
```
rudolfix, line 1135: really? so IST (India) is not supported? `UTC +05:30` this is super common I suppose, most of people live in India ). Nepal has UTC +5:45. is this a limitation of the c library underneath Arrow?

rudolfix (changes requested):
This looks pretty good and it makes sense to fix this problem in CsvWriter which looks the only one that is affected. Looking at Connectorx code they will produce +00:00 only so we are safe (in practice).

should we bump connectorx here or in a separate PR?
are you sure that timezones like IST (+5:30) are not supported?

thanks for good tests!

dlt/common/libs/pyarrow.py
zilto: The problem:
- `pyarrow` / Arrow C doesn't support offset-based timezones whether it's `+05:00` or `"+05:45"`  
- `pyarrow` / Arrow C suports the IANA database which includes "named offsets" in 1h increments `Etc/GMT-2`, `Etc/GMT-8`. There is no "named offset" `Etc/GMT+5:45` 
- To convert `"+05:45"` produced by Arrow Rust to an IANA value, we would need to a Canonical time zone like `Asia/Kathmandu` ([see wikipedia](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones))

While `Etc/GMT-8` is future-proof and reliable option, `Asia/Kathmandu` (+5:45) is a political / social construct that **could change**. If we want to do this, I would create an explicit mapping of `offset -> IANA` that's easy to inspect and maintain. We have to carefully pick timezones that don't have daily saving time for all offsets

We're dealing with these complexities because `connectorx` [toggled off chrono-tz](https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#timezone-representation) on the Arrow Rust side, which mangles timestamps moving between Arrow C and Arrow Rust. Robust solutions would be

- Wait for offset-based timestamps  in Arrow C. Since my issue on `pyarrow`, there has been activity on this [older PR](https://github.com/apache/arrow/pull/12865)
- Fork and build a distribution of `connectorx` that [reverts this change](https://github.com/sfu-db/connector-x/pull/743/files#diff-a4cb550048acd55ca4ee7d3696496f776cd8e9561b6f959d44dff6bbcc731087)

dlt/common/libs/pyarrow.py
zilto: the logic you're suggesting is slightly different. 

### Current code
My goal is to guard against `tz` values that are not strings; it's not to guard against `None` when `tz` is missing.
```python
# check if `pyarrow.types` object has `tz` attribute
# else None
tzinfo = getattr(field.type, "tz", None)
# check if `tz` attribute is a string if found
if isinstance(tzinfo, str):
  # do a check against `tz` values that are strings
  if tzinfo.startswith(("+", "-")):
     ...
```

### Suggested code
This code handles `tz` values that are `str` and missing `tz` attribute; it doesn't account for `tz` that is not a `str`.

```python
# check if `pyarrow.types` object has `tz` attribute
# else empty string
tzinfo = getattr(field.type, "tz", "")
# do a check against `tz` value
# if `tzinfo` is not a `str`,  raise an error because method isn't available
if tzinfo.startswith(("+", "-")):
``` 

### Conclusion
Might be far-fetched, but these extra cautions are particularly relevant when integrating with 3rd party code. "For my code to work with pyarrow code, this must be a string"

dlt/common/libs/pyarrow.py
zilto: Opened a separate issue #2920

rudolfix:
@zilto the end goal is to enable the newest version of connectorx in `dlt` (no `arrow2`) and allow to get correct decimals. timezone is to me a edge case that we need to solve and I'm ok with something practical

1. `connectorx` forces "+00:00" as the timezone for all DateTime instances: https://github.com/sfu-db/connector-x/blob/08406fe172c54af456723f5c1ab1f161ccb84321/connectorx/src/destinations/arrow/arrow_assoc.rs#L242 so to solve timestamp problem we need to swap this tz with "UTC" (this is what they did before). you can cleanup this directly in TableLoader. 
2. we can disable this cleanup per arrow version (when proper fix is available)
3. arrow will fix csv writer at some point so I do not think we should keep you "hack". If there's no fix ever I'd see justification in keeping it. If not - our timezone problem is temporary and to simple for it.

tldr;> I suggest to move on with the core issue, you can keep your existing hack or do a simpler one.

zilto:
Closed in favor of #2933. We will revisit the time zone conversion issue between Arrow C and Arrow Rust when `pyarrow` releases their fix

PR #2882, opened 2025-07-14
Fix logger.isEnabledFor() TypeError

burnash:
Fixes a bug where `logger.isEnabledFor()` would raise `TypeError: got an unexpected keyword argument 'stacklevel'.` 
The logger proxy was incorrectly adding `stacklevel` to all method calls, but utility methods like isEnabledFor() don't accept this parameter.


rudolfix (dismissed):
LGTM! let's merge after 0.14.0 release

PR #2886, opened 2025-07-15
basic runner docs

djudjuu:


docs/website/docs/plus/features/pipeline-runner.md
```diff
@@ -0,0 +1,189 @@
+---
+title: "Dlt+ Runner"
```
VioletM: Regarding the location:
We already have a place in our docs for runners: https://dlthub.com/docs/plus/production/runners. You can move this doc there. If you rename the existing file from `runners.md` to `pipeline-runner.md` make sure you add it to `sidebar.js`

docs/website/docs/plus/features/pipeline-runner.md
```diff
@@ -0,0 +1,189 @@
...
+- **Automatic retry policies** with configurable backoff strategies
+- **Clean state management** to ensure consistent pipeline runs
+
+## Configuration
+
+The Runner is configured through the `run_config` section in your `dlt.yml` file.
+
+### Complete configuration example
+
+```yaml
```
VioletM: I know it's not super convenient, but we usually provide several tabs where it's relevant for configuration: with yaml, with toml, and in Python/env vars. Example: https://dlthub.com/docs/plus/ecosystem/iceberg#polaris-catalog

docs/website/docs/plus/features/pipeline-runner.md
```diff
@@ -0,0 +1,189 @@
...
+        max_attempts: 5
+        multiplier: 2
+        min: 1
+        max: 10
+      retry_pipeline_steps: ["load"]
+```
+
+## Run from clean folder
+
+The `run_from_clean_folder` parameter controls whether the pipeline starts from a clean state before running.
```
VioletM: The parameter name could be confusing for the people who don't have a deep understanding of how dlt works. Do you think we can still change the name? For example, `fresh_run` or `force_resync` or `sync_from_destination`?

docs/website/docs/plus/features/pipeline-runner.md
```diff
@@ -0,0 +1,189 @@
...
+The dlt+ runner behaves differently from `pipeline.run()` when there exists pending data in the pipeline's working directory:
+`pipeline.run()` will load only the pending data instead and needs to be invoked again for the given data.
+The dlt+ runner will also try finalizing the pending data, applying the retry policy and the trace settings, and then also
+try running with the given data.
+:::
+
+
+## Trace storage
+
+The `store_trace_info` parameter enables automatic storage of the pipelines trace, which can be used for for debugging, auditing, and monitoring pipeline runs.
```
VioletM: Should we describe what exactly is in the traces? Also, you can add a link to the Pydantic description of the traces

VioletM (changes requested):
Looks good! Left some comments :) 

Also, please, run the [grammar checker](https://github.com/dlt-hub/dlt/blob/devel/docs/tools/fix_grammar_gpt.py) if you haven't yet (I can't tell if you did cause I'm not very strong in grammar hehe)!

docs/website/docs/plus/features/pipeline-runner.md
djudjuu: Hmm, i guess we could, as the runner is released pretty silently for now and probably nobody is using it. But I am not convinced that your suggestions are better.. lets chat

docs/website/docs/plus/features/pipeline-runner.md
djudjuu: that's not supported atm. The three options only work if the config is in a @with_config / @config decorator. WE have an issue for this though. https://github.com/dlt-hub/dlt-plus/issues/314

djudjuu: .env and config.toml are not supported because the runner is not using @config  yet. but this has at least an issue: https://github.com/dlt-hub/dlt-plus/issues/314

docs/website/docs/plus/features/pipeline-runner.md
djudjuu: done

docs/website/docs/plus/production/pipeline-runner.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Configuration
+
+The runner is configured through the `run_config` section of a pipeline in your `dlt.yml` file.
+For direct access, you can also import and use the runner directly via the [Python interface](./pipeline-runner.md#python-api) in your code.
+Configuration via environment variables or `config.toml` is still under development.
+
+### Complete configuration example
+
+```yaml
```
VioletM, line 69: I still think that it would be good to at least have a tab with Python API, just the same example that you have below. Project is a quite unique feature and not everyone will be using it. I just want users to understand right away, that it's not nessesary.

docs/website/docs/plus/production/pipeline-runner.md
```diff
@@ -0,0 +1,198 @@
+---
+title: "Pipeline Runner"
```
VioletM:
```suggestion
title: "Pipeline runner"
```
we should use the Sentence case (https://www.notion.so/dlthub/Documentation-Writing-Guide-87f4fcc32655460c83bcaf7787d11e67?source=copy_link#7c4f89a24e854722b7ed973d73641f2e)

docs/website/docs/plus/production/pipeline-runner.md
djudjuu: ok, we can do tabs that show the same in yaml and python

docs/website/docs/plus/production/observability.md
```diff
@@ -1,12 +1,70 @@
...
+        pass
+
+# attach your collector to the pipeline
+pipeline.collector = MyLogCollector()
+
+load_info = pipeline.run(my_source)
+print(load_info)
+```
+
 # Observability
```
VioletM: I think it should be a header of the second type and maybe we can change the name a bit

```suggestion
## Additional features
```

Otherwise there is still the dlt+ callout
image

VioletM (dismissed):
Looks great! Only one small thing left

PR #2888, opened 2025-07-15
Add extra_credentials to clickhouse destination configuration

### LINKED ISSUE #2887: Add extra_credentials to clickhouse configuration use with s3
### Feature description

Update the clickhouse destination configuration and execution to allow usage of the `extra_credentials` arguments to the s3 function. This is recommended approach per the clickhouse [docs ](https://clickhouse.com/docs/sql-reference/table-functions/s3#using-s3-credentials-clickhouse-cloud).

There is a similar feature in the redshift destination

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

The current clickhouse implementation does not allow for passing `extra_credentials` to the s3 function which is used in the insert statements. We would prefer to follow the docs and use the assume role functionality outlined in the clickhouse docs.

When staging files with s3, the s3 credentials are typically passed to the destination and the credentials the dlt app users should allow for more access to s3 buckets versus the role assigned to the clickhouse cloud instance.

### Proposed solution

1. Add a new attribute named `extra_credentials` to the clickhouse credentials configuration.
2. Prefer this function for auth inplace of the aws access key and secret

### Related issues

_No response_

warje:
### Description
Adds a new configuration option to the clickhouse destination named `extra_credentials` which will map to the `extra_credentials` argument to the s3 function. This allows dlt to use role based authentication in alignment with the clickhouse docs

### Related Issues

- Fixes #2887 

### Additional Context
Currently using this change in our staging and production environments.

### DIFF
--- a/tests/load/clickhouse/test_clickhouse_table_function.py
+++ b/tests/load/clickhouse/test_clickhouse_table_function.py
@@ -65,9 +65,7 @@ def test_clickhouse_s3_table_function(
         creds_dict["s3_extra_credentials"] = s3_extra_credentials
 
     config = resolve_configuration(
-        ClickHouseClientConfiguration(
-            credentials=ClickHouseCredentials(creds_dict)
-        )
+        ClickHouseClientConfiguration(credentials=ClickHouseCredentials(creds_dict))
     )
     staging_credentials = AwsCredentials(
         aws_access_key_id=aws_access_key_id,
@@ -87,7 +85,7 @@ def test_clickhouse_s3_table_function(
     )
 
     expected = (
-        f"s3('https://bucket-name.s3.amazonaws.com/path/to/data'"
+        "s3('https://bucket-name.s3.amazonaws.com/path/to/data'"
         f",{expected_auth},'parquet','auto','gz')"
     )
     assert result == expected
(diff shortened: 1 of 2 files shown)


dlt/destinations/impl/clickhouse/configuration.py
```diff
@@ -33,6 +33,8 @@ class ClickHouseCredentials(ConnectionStringCredentials):
     """Timeout for establishing connection. Defaults to 10 seconds."""
     send_receive_timeout: int = 300
     """Timeout for sending and receiving data. Defaults to 300 seconds."""
+    extra_credentials: Optional[Dict[str, str]] = None
```
sh-rp: I would call this setting s3_extra_credentials so it is clear to what it applies.

sh-rp (changes requested):
Hey, thanks for this PR :) I have one note about the naming of the setting and two requests:
* Please add a simple test where you check the building of these strings depending on different settings
* Please update the clickhouse docs page and include information about this new setting

warje:
> Hey, thanks for this PR :) I have one note about the naming of the setting and two requests:
> 
>     * Please add a simple test where you check the building of these strings depending on different settings
> 
>     * Please update the clickhouse docs page and include information about this new setting

Thank you for the review and for your work on the project.

I have made the requested changes which includes tests and doc updates.

Regarding the unit tests. I refactored the `run` method a little to make it easier to test this specific change without running a pipeline.

I gave the unit tests a good shot, but honestly got a little confused with the test project structure specifically in the `loads` directory so if you have advise or guidance as far as what I can do differently, I would appreciate to hear it.

PR #2891, opened 2025-07-16
feat: autocompletion added for dataset and relation

zilto:
This adds tab completion for `dlt.Dataset` and `dlt.Relation`. It respectively suggests table names and column names.

## Related
I had to modify `__getattr__` on `ReadableDBAPIDataset` class because it was breaking the `_ipython_key_completions_`. Its implementation effectively prevented any use of `getattr()` to dynamically retrieve attributes or methods by name. My fix involved manually searching the object's `__dict__`.

### DIFF
--- a/dlt/destinations/dataset/dataset.py
+++ b/dlt/destinations/dataset/dataset.py
@@ -78,7 +78,8 @@ class ReadableDBAPIDataset(Dataset):
 
     @property
     def tables(self) -> list[str]:
-        return list(self.schema.tables.keys())
+        # return only completed tables
+        return self.schema.data_table_names() + self.schema.dlt_table_names()
 
     def _ipython_key_completions_(self) -> list[str]:
         return self.tables
@@ -209,13 +210,14 @@ class ReadableDBAPIDataset(Dataset):
     def table(self, table_name: str, table_type: Literal["ibis"]) -> IbisTable: ...
 
     def table(self, table_name: str, table_type: Literal["relation", "ibis"] = "relation") -> Any:
-        # dataset only provides access to tables known in dlt schema, direct query may cirumvent this
-        if table_name not in self.schema.tables.keys():
+        # dataset only provides access to tables known in dlt schema, direct query may circumvent this
+        available_tables = self.tables
+        if table_name not in available_tables:
             # TODO: raise TableNotFound
             raise ValueError(
                 f"Table `{table_name}` not found in schema `{self.schema.name}` of dataset"
                 f" `{self.dataset_name}`. Available table(s):"
-                f" {', '.join(self.schema.tables.keys())}"
+                f" {', '.join(available_tables)}"
             )
 
         if table_type == "ibis":
--- a/tests/destinations/test_readable_dbapi_dataset.py
+++ b/tests/destinations/test_readable_dbapi_dataset.py
@@ -98,12 +98,10 @@ def test_copy_and_chaining() -> None:
         ),
     )
 
-    dataset.schema.tables["items"] = {
-        "columns": {
-            "one": {"data_type": "text", "name": "one"},
-            "two": {"data_type": "json", "name": "two"},
-        }
-    }
+    dataset.schema.tables["items"] = new_table(
+        "items",
+        columns=[{"data_type": "text", "name": "one"}, {"data_type": "json", "name": "two"}],
+    )
 
     # create relation and set some stuff on it
     relation = cast(ReadableDBAPIRelation, dataset.items)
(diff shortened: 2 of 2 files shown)


dlt/destinations/dataset/relation.py
```diff
@@ -464,17 +471,18 @@ def scalar(self) -> Any:
     def __getitem__(self, columns: Sequence[str]) -> Self:
         # NOTE remember that `issubclass(str, Sequence) is True`
         if isinstance(columns, str):
+            columns = [columns]
```
rudolfix: yeah! I wanted to raise this bug as well

rudolfix: I added a test for that

dlt/destinations/dataset/relation.py
```diff
@@ -181,6 +181,13 @@ def schema(self) -> TTableSchema:
     @schema.setter
     def schema(self, new_value: TTableSchema) -> None:
         raise NotImplementedError("Schema may not be set")
+    
+    @property
+    def columns(self) -> list[str]:
```
rudolfix, line 186: NOTE: this is part of implementation and not exposed to the user. we can keep it here but at some point we should decide which props to promote to public interface

dlt/destinations/dataset/dataset.py
```diff
@@ -266,11 +274,18 @@ def row_counts(
...
+        if table_name not in self.tables:
+            raise KeyError(f"Table `{table_name}` not found on dataset. Available tables: `{self.tables}`")
 
-    def __getattr__(self, table_name: str) -> Relation:
-        """access of table via property notation"""
         return self.table(table_name)
+    
+    def __getattr__(self, name: str) -> Any:
+        """Retrieve a `Relation` via `__getitem__` if standard `__getattr__` returns `None`."""
+        attribute = self.__dict__.get(name, None)
```
rudolfix: `__getattr__` is called only when `name` is not in `__dict__` so you have none here every time. Old implementation was not raising AttributeError but ValueError (via `self.table`) which IMO was the cause of your problems.

rudolfix (dismissed):
LGTM but please read the comment because I added a few fixes.

this is working very well in Jupyter Notebook now and I'll try to merge this ASAP if tests will pass

dlt/destinations/dataset/dataset.py
```diff
@@ -76,6 +76,13 @@ def schema(self) -> Schema:
             self._ensure_schema()
         return self._schema
 
+    @property
+    def tables(self) -> list[str]:
```
sh-rp, line 80: I think we should only list tables that are complete here, trying to fetch data from tables that are not materialized will give an error. But I am not sure.

dlt/destinations/dataset/relation.py
```diff
@@ -182,6 +182,13 @@ def schema(self) -> TTableSchema:
     def schema(self, new_value: TTableSchema) -> None:
         raise NotImplementedError("Schema may not be set")
 
+    @property
+    def columns(self) -> list[str]:
+        return list(self.schema.get("columns", {}).keys())
+
+    def _ipython_key_completions_(self) -> list[str]:
```
sh-rp, line 189: same comment as above, only complete columns should be returned here.

dlt/destinations/dataset/relation.py
zilto: It's a purposeful addition. `.columns` is common across `pd.DataFrame`, `pl.DataFrame`, `pyarrow.Table`, etc.

dlt/destinations/dataset/dataset.py
zilto: is that something I can check on the `dlt.Schema` ?

dlt/destinations/dataset/dataset.py
zilto: do you have a suggestion for the fix? Without the line I've added, the completion mechanism fails to call `Dataset._ipython_key_completions_()`. 

It has to do with `getattr(dataset_obj, "_ipython_key_completions_")` failing inside IPython (i.e., code we can't change). I haven't looked into it more

zilto:
...

dlt/destinations/dataset/relation.py
zilto: Here, we're retrieving columns mentioned on the relation, not physical columns. For example, if you have a `SELECT foo AS bar FROM table`, calling `.columns` on the relation should return `bar` despite not existing on `table`.

It's different from completion on dataset where you made a good point that only want actually available tables

Issue #2611, opened 2025-05-07
chore: refactor `dataset` to be top level


zilto:
### Feature description

Currently, the code dataset-related code is split across:
- `dlt/destinations/dataset/`
- `dlt/common/destination/dataset.py`
- `tests/load/test_read_interfaces.py`

## Proposed solution
Refactor content to be stored under `dlt/dataset` and `tests/dataset`. This will help development and maintenance as the relationships between pipelines, resources, transformations, datasets, and dlt+ projects evolve.

To maintain backwards compatibility:
- initially, the file `dlt.dataset.__init__.py` will import content from `dlt.destinations.dataset.` et al.
- a deprecation warning is triggered when importing `dlt.destinations.dataset` directly
- new code will be added to `dlt.dataset` instead of older locations
- on `dlt == 2.0.0`, we can remove the older code

## Impact
- Moving library code and changing the import paths could impact downstream usages of internals (especially `dlt_plus`). This can be done with backward compatibility
- Moving tests should have no impact

PR #2896, opened 2025-07-16
feat: top-level `dlt.Dataset` and `dlt.Relation`

zilto:
## Summary
This PR introduces the top-level private module `dlt._dataset` which centralizes content previously in `dlt.common.destination.dataset` and `dlt.destinations.dataset`. It renames classes `ReadableDBAPIDataset` and `ReadableDBAPIRelation` to `Dataset` and `Relation`.

## Related
- #2611 
- #2654 

## Goal
- Provide more readable import paths and naming for public interface. (both for humans and LLMs)
- Clarify public vs. private interface and trim `Protocol` abstractions
- Entirely backwards compatible with appropriate deprecation warnngs.

## Changes
- Moved content to `dlt._dataset` such that old classes point to new classes `OldClass -> NewClass`. This means implementation will match, but we'll be able to deprecate `OldClass` when we want to
- Add the top-level public factory `dlt.dataset()` to instantiate dataset instead of `dlt.destinations.dataset.dataset()`. It's equiavalent to `dlt.Dataset()`. Though, this separate interface could allow us to hide more implementation details in the future.
- Remove `Dataset(Protocol)` and `Relation(Protocol)`. It simplifies the code and the typing. Including a `Protocol` doesn't reduce our maintainer burden; it increases it. Having both `Relation(Protocol)` and `ReadableDBAPIRelation(Relation)` means that I can't update `ReadableDBAPIRelation` in ways that break the protocol, but I can't update the `Relation` protocol either if someone depends on it. Then, if no user depends on the `Relation`, why include it in the first place? The numerous `type: ignore` and `cast()` calls are code smells that it introduces a lot of complexity
- No test logic was changed; only the necessary changes to import, `typing.cast()` calls, and `# type: ignore` were modified to comply with linting rules

### Notes
- Now is the best time to make these changes before users relying on these objects more
- Large numbers of `type: ignore` or linting skips should be challenged during PR reviews


zilto:
Both errors seem related to the chess pipeline

sh-rp:
I haven't reviewed it yet, but this is very cool! What I'd say for now:
* I'd say we want to keep the `Relation` and `Dataset` protocols as public interfaces. The idea here is, that there can be multiple implementations for them, the most obvious case is where we have an ad-hoc Dataset or Relation that is not backed by data previously loaded with dlt but created on the file from a folder of files. There will be a few differences, for example the schema will probably calculated from what duckdb discovered in this folder etc. But you can use them like any other dataset to transform or select data.
* I think we should create some kind of doc that defines how we want to organize our files and bit by bit (maybe starting with this) change the repo to it. One big part would be this private public separation, but also other rules that you have mentioned in your other ticket about what may be in init files and more.

WDYT?

sh-rp:
PS: We already have a `SupportsPipeline` protocol, so maybe our default dataset implementation could be named as you suggest, but we keep the protocols and call them `SupportsDataset` and `SupportsRelation`.

rudolfix:
OK I see that @sh-rp wants to keep protocols because there's a high chance we reimplement them. Indeed we could implement `Dataset` that wraps hugginface dataset and I'd like to do that soon honestly. let's talk then

zilto:
There seems to be a lot of concepts overlapping, at the cost of clarity. A `dlt.Dataset` is a "`dlt.Schema` + a destination". It's a very light abstraction for data access. In fact, the data access is defined by the destination (e.g., SQL client) and is decoupled from the `dlt.Dataset`. Under that definition, I can't imagine any other implementation.

## Data discovery
> the most obvious case is where we have an ad-hoc Dataset or Relation that is not backed by data previously loaded with dlt but created on the file from a folder of files

That is why I implemented SQLGlot DDL. It's a very lightweight way to "external tables" not managed by dlt. Marcin pointed out that sqlalchemy reflection achieves similar goals. In all cases, it's a **destination** feature, not an alternative dataset implementation.

## Alternative datasets
>  Indeed we could implement Dataset that wraps hugginface dataset and I'd like to do that soon honestly.

Despite being called a "HuggingFace Dataset", it would be very confusing to make a `dlt.Dataset` implementation. I'm guessing that the set of features you want could be implemented as a HuggingFace Hub source and destination. 

## Public vs. private
> Why you name this module _dataset and not dataset? do you want to rename other modules? (ie. _pipeline)? there are actually public interfaces in dataset so IDK we we make it private

Currently, I'm only export `dataset()`, `Dataset()`, `Relation()`. `dlt._dataset.__init__` contains the definition and `dlt.__init__` imports from there the public interfaces. We can open up the namespace `dlt.dataset.`, but it could class with a top-level `dlt.dataset()` function

# Keeping it stupid simple
dlt already has too many `m-to-n` problems. We should have unified interfaces for our key entrypoints `dlt.Pipeline`, `dlt.Dataset`,  and `dlt.Relation`. I understand the role and the value of `Protocol`, and `abc.ABC`, but lets add them once it serves a use case instead of early optimization. I'm also afraid of making these protocols public.

zilto:
Closing in favor of #2983

PR #2905, opened 2025-07-17
fix: use modern string formatting to set dataset name

julienschuermans:
### Description

No need to use [% formatting](https://docs.python.org/3/tutorial/inputoutput.html#old-string-formatting), we can [avoid runtime errors](https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting) by adopting `.format` instead.

This can avoid unexpected pipeline failures during the `"sync"` stage, I encountered this quite often:

> "Pipeline execution failed at `step=sync` with exception:\n\n<class 'TypeError'>\nnot all arguments converted during string formatting",

### DIFF
--- a/dlt/pipeline/pipeline.py
+++ b/dlt/pipeline/pipeline.py
@@ -1535,7 +1535,7 @@ class Pipeline(SupportsPipeline):
 
         # normalizes the dataset name using the dataset_name_layout
         if self.config.dataset_name_layout:
-            new_dataset_name = self.config.dataset_name_layout % new_dataset_name
+            new_dataset_name = self.config.dataset_name_layout.format(new_dataset_name)
         return new_dataset_name
 
     def _set_default_schema_name(self, schema: Schema) -> None:


sh-rp:
@julienschuermans thanks for the PR, we can't merge this though, as this will be a breaking change and stop existing pipelines that have the `dataset_name_layout` applied with `%` placeholders. Is this change for some reason important for your project?

zilto:
Would be useful if you could open an issue describing the problem you hint at!
> This can avoid unexpected pipeline failures during the "sync" stage, I encountered this quite often:
> `"Pipeline execution failed at step=sync with exception:\n\n<class 'TypeError'>\nnot all arguments converted during string formatting",`

Will close this PR as not planned for the reason @sh-rp  gave. Happy to discuss the issue and try to find an alternative solution

julienschuermans:
Thanks! I opened an issue here: https://github.com/dlt-hub/dlt/issues/2906

PR #2909, opened 2025-07-21
QoL: Destructive schema sync after manual column dropping

anuunchin:
### Description
This PR adds a new pipeline function that syncs the dlt schema with the destination (not vice versa) by removing a column from the schema if that column has been manually deleted in the destination.

> The motivation behind this is that rather than offering a cli command that drops columns - where we need to have separate `drop_columns` functions due to dialect differences and thus adds additional maintenance overhead - we delegate the dropping part to the user and instead allow them to sync the dlt schema in those scenarios.

### Related PRs:
#2754 

### Further:

This should be extended to table drop syncs as well.

### Note:

This is essentially solving the problem when the user manually drops things in the destination and the dlt pipeline breaks.


dlt/common/destination/client.py
```diff
@@ -539,6 +539,37 @@ def update_stored_schema(
...
+            schema_columns = self.schema.get_table_columns(table_name)
+            dropped_columns = set(schema_columns.keys()) - set(actual_columns)
+            if dropped_columns:
+                for dropped_col in dropped_columns:
+                    if schema_columns[dropped_col].get("increment"):
+                        logger.warning(
+                            "An incremental field is being removed from schema."
+                            "You should unset the"
+                            " incremental with `incremental=dlt.sources.incremental.EMPTY`"
+                        )
```
anuunchin: I just realized the incremental setting is also saved outside of the schema, so when the user
- manually drops a column from destination with an incremental setting
- syncs the schema destructively

They will need to also unset the incremental with `apply_hints(incremental=dlt.sources.incremental.EMPTY)`

dlt/destinations/impl/dummy/dummy.py
```diff
@@ -160,6 +160,10 @@ def update_stored_schema(
             )
         return applied_update
 
+    def _get_actual_columns(self, table_name: str) -> List[str]:
+        schema_columns = self.schema.get_table_columns(table_name)
+        return list(schema_columns.keys())
+
```
anuunchin: We just get columns from schema because it's a dummy

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -468,6 +468,26 @@ def update_stored_schema(
...
+        elif self.is_open_table("delta", table_name):
+            from dlt.common.libs.deltalake import get_table_columns as get_delta_table_columns
+
+            delta_table = self.load_open_table("delta", table_name)
+            return get_delta_table_columns(delta_table)
+
+        else:
+            schema_columns = self.schema.get_table_columns(table_name)
+            return list(schema_columns.keys())
+
```
anuunchin: It's impossible to get the actual columns from filesystem files without table format unless we read entire files. Also, it's unlikely that the user manually deletes specific columns from filesystem files I think 👀 . Therefore, we should raise NotImplemented instead of doing basically nothing.

dlt/common/destination/client.py
```diff
@@ -539,6 +539,37 @@ def update_stored_schema(
...
+        """
+        for table in self.schema.data_tables():
+            table_name = table["name"]
+
+            actual_columns = self._get_actual_columns(table_name)
+            schema_columns = self.schema.get_table_columns(table_name)
+            dropped_columns = set(schema_columns.keys()) - set(actual_columns)
+            if dropped_columns:
+                for dropped_col in dropped_columns:
+                    if schema_columns[dropped_col].get("increment"):
```
rudolfix: this is good and we need to keep this warning

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -468,6 +468,26 @@ def update_stored_schema(
         # externally changed
         return applied_update
 
+    def _get_actual_columns(self, table_name: str) -> List[str]:
+        """Get actual column names from files in storage for regular (non-delta/iceberg) tables
+        or column names from schema"""
+
+        if self.is_open_table("iceberg", table_name):
```
rudolfix: this should be merged into `get_storage_tables` which is already present in the filesystem. it will be really helpful.

note: both are operating on Arrow schemas so look how `get_storage_table` works in lance... you do not need to implement TypeMapper just look at lancedb implementation

dlt/common/destination/client.py
```diff
@@ -539,6 +539,37 @@ def update_stored_schema(
             )
         return expected_update
 
+    def update_stored_schema_destructively(
```
rudolfix: this is overall a good idea to add to JobClient but we should be more selective. Look at my top level review. We need 1. mixin class 2. it needs to be optional

rudolfix (changes requested):
This is very good idea but we need to approach it in more systematic way:

1. (almost) all of our destinations have 
```py
def get_storage_table(self, table_name: str) -> Tuple[str, TTableSchemaColumns]:
```
and/or
```py
def get_storage_tables(
        self, table_names: Iterable[str]
    ) -> Iterable[Tuple[str, TTableSchemaColumns]]: 
```
implemented. this will reflect the storage to get table schema out of it. you can use it to compare with the pipeline schema.

2. let's formalize it: add mixin class like `WithTableReflection` in the same manner `WithStateSync` is done. `get_storage_tables` is the more general method so you can add only this to the mixin
3. Now add this mixing to all `JobClientBase` implementations for which you want to support our new sync schema.

When the above is done we are able to actually compute schema diff.

Top level interface:
1. we have `sync_schema` that will do a regular schema migration (add missing columns and tables in the destination`
2. we need another method which is the reverse: it will delete columns and tables in the schema that are not present in the destination and then do the schema sync above
3. the method above should have a dry run mode - where we do not change the pipeline schema and we do not sync it
4. it should make sure if `destination_client()` implements `WithTableReflection` before continuing
5. it should allow to select tables to be affected

when this is done we can think about extending cli ie `dlt pipeline <name> schema` command

dlt/common/schema/schema.py
```diff
@@ -463,6 +463,26 @@ def drop_tables(
                 self.data_item_normalizer.remove_table(table_name)
         return result
 
+    def drop_columns(self, table_name: str, column_names: Sequence[str]) -> TPartialTableSchema:
```
rudolfix, line 425: this modifies table in-place (no copy is created). if so then you can keep it here. just mention it in docstring

dlt/common/schema/schema.py
```diff
@@ -463,6 +463,26 @@ def drop_tables(
...
+        dropped_col_schemas: TTableSchemaColumns = {}
+
+        for col in column_names:
+            col_schema = self._schema_tables[table["name"]]["columns"].pop(col)
+            dropped_col_schemas[col] = col_schema
+
+        table["columns"] = dropped_col_schemas
+        return table
+
+    def get_child_tables(self, table_name: str) -> List[TTableSchema]:
```
rudolfix: child tables linked via parent are called "nested tables". we already have `get_nested_tables` in utils. and in general pure schema operation are moved to utils to not grow this file too much

nested tables vs. classical references - there's mode in schema.md

dlt/common/destination/client.py
```diff
@@ -608,6 +726,22 @@ def get_stored_state(self, pipeline_name: str) -> Optional[StateInfo]:
         pass
 
 
+class WithTableReflection(ABC):
+    @abstractmethod
+    def get_storage_tables(
```
rudolfix, line 628: yeah this is really good idea to abstract a mixin. but the docstrings are for SQL job client implementation and you should move it there. here describe what is the role of the interface (to reflect existing tables. etc.)

dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -174,7 +175,9 @@ def gen_key_table_clauses(
         return sql
 
 
-class BigQueryClient(SqlJobClientWithStagingDataset, SupportsStagingDestination):
+class BigQueryClient(
+    SqlJobClientWithStagingDataset, SupportsStagingDestination, WithTableReflection
```
rudolfix: just add `WithTableReflection` to `SqlJobClient` where reflection method is implemented, then you do not need to do it for all destinations

dlt/common/destination/client.py
```diff
@@ -539,6 +542,121 @@ def update_stored_schema(
             )
         return expected_update
 
+    def update_dlt_schema(
```
rudolfix: this is good but it should not be a part of the very base job client. I'd probably move it to dlt/destination/utils as a regular function that instead of `self` takes `WithTableReflection` instance and `Schema` instance to be aligned

dlt/common/destination/client.py
```diff
@@ -539,6 +542,121 @@ def update_stored_schema(
...
+
+        Args:
+            table_names (Iterable[str], optional): Check only listed tables. Defaults to None and checks all tables.
+
+        Returns:
+            Optional[TSchemaTables]: Returns an update that was applied to the schema.
+        """
+        from dlt.destinations.sql_client import WithSqlClient
+
+        if not (isinstance(self, WithTableReflection) and isinstance(self, WithSqlClient)):
```
rudolfix: move this to pipeline.py

dlt/common/destination/client.py
```diff
@@ -539,6 +542,121 @@ def update_stored_schema(
...
+
+            if not dropped_col_names:
+                return {}
+
+            partial_table: TPartialTableSchema = {"name": table_name, "columns": {}}
+
+            for esc_name in dropped_col_names:
+                orig_name = escaped_to_original[esc_name]
+
+                # Athena doesn't have dlt columns in actual columns. Don't drop them anyway.
```
rudolfix: I do not get this...

dlt/common/destination/client.py
```diff
@@ -539,6 +542,121 @@ def update_stored_schema(
...
+                if disregard_dlt_columns and orig_name in [C_DLT_ID, C_DLT_LOAD_ID]:
```
rudolfix: `C_DLT_ID` must be normalized (but not quoted)

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -430,6 +469,45 @@ def verify_schema(
             raise exceptions[0]
         return loaded_tables
 
+    def _diff_between_actual_and_dlt_schema(
```
rudolfix: why this code is duplicated? are there some subtle differences in sql destinations vs. filesystem? then it is better to fix them. I think we want to implement reflection interface for lancedb as well (in the future). so this util method above should do

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -370,7 +376,40 @@ def get_storage_tables(
...
+                    # If it's an open table, only actually exsiting columns
+                    if self.is_open_table("iceberg", table_name):
+                        from dlt.common.libs.pyiceberg import (
+                            get_table_columns as get_iceberg_table_columns,
+                        )
+
+                        iceberg_table = self.load_open_table("iceberg", table_name)
+                        actual_column_names = get_iceberg_table_columns(iceberg_table)
+
+                        col_schemas = {
```
rudolfix: LGTM but I'm not sure what other implementations do? I think SQL implementation will returns all columns in the table. also those unknown by dlt. it is never tested. but if so we should keep this behavior

I know that for filesystem (parquet/json) we can't do that. but here we can

rudolfix: both schemas are basically arrow schemas and we have all converison functions

dlt/destinations/impl/filesystem/filesystem.py
rudolfix: in the base class this is an inner function so this one won't be called at all?

dlt/common/destination/client.py
```diff
@@ -539,6 +542,121 @@ def update_stored_schema(
...
+                table_drops.pop(table_name)
+                logger.warning(
+                    f"Removing table '{table_name}' from the dlt schema would leave orphan"
+                    f" table(s): {'.'.join(repr(t) for t in orphaned_table_names)}. Drop these"
+                    " child tables in the destination and sync the dlt schema again."
+                )
+
+        # 3. If it's not a dry run, we actually drop fromt the dlt schema
+        if not dry_run:
+            for table_name in table_drops:
```
rudolfix: schema has `drop_table` it does not drop nested tables which is what you want here I think (you force user to drop the full tree)

dlt/pipeline/pipeline.py
```diff
@@ -1036,6 +1039,24 @@ def sync_schema(self, schema_name: str = None) -> TSchemaTables:
             client.initialize_storage()
             return client.update_stored_schema()
 
+    @with_schemas_sync
+    def sync_dlt_schema(
```
rudolfix: let's clean up the names

- deprecate `sync_schema`. above add `sync_schema_to_destination`
- name this this `sync_schema_from_destination`
(or use different but clear names)

dlt/pipeline/pipeline.py
```diff
@@ -1036,6 +1039,24 @@ def sync_schema(self, schema_name: str = None) -> TSchemaTables:
...
+            raise PipelineConfigMissing(
+                self.pipeline_name,
+                "default_schema_name",
+                "load",
+                "Pipeline contains no schemas. Please extract any data with `extract` or `run`"
+                " methods.",
+            )
+        schema = self.schemas[schema_name] if schema_name else self.default_schema
+        with self._get_destination_clients(schema)[0] as client:
+            client.initialize_storage()
```
rudolfix: this method should not modify destination. `use is_storage_initialized` and raise exception if not.

dlt/pipeline/pipeline.py
```diff
@@ -1036,6 +1039,24 @@ def sync_schema(self, schema_name: str = None) -> TSchemaTables:
...
+            return client.update_dlt_schema(table_names=table_names, dry_run=dry_run)
```
rudolfix: here you should get reflection mixin (isinstance) and pass it to the util function I mentioned above. is mixin is not available, display clear explanation (that destination of given type can't reflect tables)

rudolfix (changes requested):
some changes needed

dlt/common/schema/schema.py
anuunchin: Oww, this makes total sense 🙏 - definitely had a feeling I was reinventing the wheel, but was looking for a function with the wrong keyword "child"

dlt/destinations/impl/bigquery/bigquery.py
anuunchin: oww right, WithTableReflection is already in SqlJobClient (wanted a straightforward inheritance going on so I could directly jump to references 👀 )

dlt/destinations/impl/filesystem/filesystem.py
anuunchin: No differences at all, this is a mistake from my side when I was moving around functions - this is removed

dlt/common/destination/client.py
anuunchin: owwww

dlt/pipeline/pipeline.py
anuunchin: Oww, right this makes sense

dlt/pipeline/pipeline.py
anuunchin: This is, indeed, much cleaner -> I removed the unnecessary abstract methods

dlt/common/destination/client.py
anuunchin: but dlt column names will always appear as they are in the dlt schema 👀

dlt/common/destination/client.py
anuunchin: Agree that it's confusing, just updated the docstring

dlt/destinations/impl/filesystem/filesystem.py
anuunchin: owww, makes sense, I overlooked the existence of `py_arrow_to_table_schema_columns`

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -478,7 +488,34 @@ def get_storage_tables(
...
+
+                    else:
+                        logger.warning(
+                            f"Table '{table_name}' does not use a table format and does not support"
+                            " true schema reflection. Returning column schemas from the dlt"
+                            " schema, which may be stale if the underlying files were manually"
+                            " modified. "
+                        )
+                        yield (table_name, self.schema.get_table_columns(table_name))
+
```
anuunchin, line 518: Just realized that for parquet files we can also just use pyarrow and read actual metadata 👀 , but I still don't think people drop columns in parquet files...

dlt/common/schema/schema.py
```diff
@@ -367,14 +367,26 @@ def drop_tables(
...
-        # TODO: make sure all nested tables to table_names are also dropped
+        # TODO: make sure all nested tables to table_names are also dropped,
         for table_name in table_names:
             table = self.get_table(table_name)
             if table and (not seen_data_only or utils.has_table_seen_data(table)):
                 result.append(self._schema_tables.pop(table_name))
                 self.data_item_normalizer.remove_table(table_name)
         return result
 
+    def drop_columns(self, table_name: str, column_names: Sequence[str]) -> TPartialTableSchema:
```
rudolfix, line 425: ohhh I think I implemented this one

dlt/destinations/impl/filesystem/filesystem.py
rudolfix: parquet files are immutable. you'd be reproducing what ideberg/delta is doing

dlt/destinations/utils.py
```diff
@@ -292,3 +298,163 @@ def get_deterministic_temp_table_name(table_name: str, op: str) -> str:
...
+    This function identifies columns that exist in the dlt schema but are missing from the actual
+    destination table. It's used during schema synchronization to detect when columns may have
+    been dropped from the destination and need to be removed from the dlt schema as well.
+
+    However, dlt internal columns (_dlt_id, _dlt_load_id) are treated specially because:
+
+    1. Users rarely drop dlt internal columns manually, and if they did,
+       dlt cannot recover from this situation anyway.
+
+    2. Athena has a constraint where dlt columns exist in the data but not in the table metadata:
```
rudolfix: this looks like a bug! or columns starting with `_` are not visible in INFORMATION SCHEMA?

dlt/destinations/utils.py
```diff
@@ -292,3 +298,163 @@ def get_deterministic_temp_table_name(table_name: str, op: str) -> str:
...
+                client,
+                schema,
+                table_name,
+                set(actual_col_schemas.keys()),
+            )
+            if partial_table:
+                column_drops[table_name] = partial_table
+
+    # 2. For entire table drops, we make sure no orphaned tables remain
+    for table_name in table_drops.copy():
```
rudolfix, line 364: new `drop_tables` in schema will also drop nested tables. do we need to do it here? or maybe part of the code can be reused?

rudolfix (changes requested):
this is really good! you need to resolve conflicts and maybe simplify code that deals with nested tables.

maybe we could document this? not sure where in the documentation it should go

dlt/destinations/impl/filesystem/filesystem.py
anuunchin: owww, right, that makes sense

dlt/destinations/utils.py
anuunchin: I separated the common part into a separate function

dlt/destinations/utils.py
anuunchin: hmmm, I wrote the test wrong and then blamed it on Athena's quirk here

dlt/common/destination/client.py
```diff
@@ -622,6 +626,20 @@ def get_stored_state(self, pipeline_name: str) -> Optional[StateInfo]:
...
+    @abstractmethod
+    def get_storage_tables(
+        self, table_names: Iterable[str]
+    ) -> Iterable[Tuple[str, TTableSchemaColumns]]:
+        """Retrieves table and column information for the specified tables.
+
+        Returns an iterator of tuples (table_name, columns_dict) where columns_dict
+        contains column schemas for existing tables, or is empty for non-existent tables.
+        Implementations use database introspection (INFORMATION_SCHEMA, table reflection) or file metadata.
+        """
```
rudolfix, line 642: Important note is missing: that column names are returned as in INFORMATION SCHEMA (database casing) and when comparing to dlt schema, columns in dlt schema must be casefolded. here's original docstring

The column names are returned as in information schema. To match those with columns in existing table, you'll need to use
        `schema.get_new_table_columns` method and pass the correct casing. Most of the casing function are irreversible so it is not
        possible to convert identifiers into INFORMATION SCHEMA back into case sensitive dlt schema.

dlt/destinations/utils.py
```diff
@@ -292,3 +298,152 @@ def get_deterministic_temp_table_name(table_name: str, op: str) -> str:
 
     op_name = f"{table_name}_{op}"
     return f"{op_name}_{NamingConvention._compute_tag(op_name, 0.001)}"
+
+
+class WithTableReflectionAndSql(WithTableReflection, WithSqlClient):
+    pass
+
+
+def _diff_between_actual_and_dlt_schema(
```
rudolfix: this is good but:
1. name is misleading, `get_removed_table_columns`
2. please look at function in `Schema` which does the reverse
```py
    def get_new_table_columns(
        self,
        table_name: str,
        existing_columns: TTableSchemaColumns,
        case_sensitive: bool,
        include_incomplete: bool = False,
    ) -> List[TColumnSchema]:
        """Gets new columns to be added to `existing_columns` to bring them up to date with `table_name` schema.
        Columns names are compared case sensitive by default. `existing_column` names are expected to be normalized.
        Typically they come from the destination schema. Columns that are in `existing_columns` and not in `table_name` columns are ignored.

        Optionally includes incomplete columns (without data type)"""
```
- try to extract get_new_table_columns here, you do not need  sql_client (just case folding function)
- model `get_removed_table_columns`

that will be nice code cleanup

dlt/destinations/utils.py
```diff
@@ -292,3 +298,152 @@ def get_deterministic_temp_table_name(table_name: str, op: str) -> str:
...
+    Example:
+        If dlt schema has [user_id, name, _dlt_id, _dlt_load_id] but destination
+        INFORMATION_SCHEMA only shows [user_id, name], this function would return
+        an empty dict (assuming disregard_dlt_columns=True) rather than suggesting
+        the dlt columns should be dropped.
+    """
+    col_schemas = schema.get_table_columns(table_name)
+
+    # Map escaped (like actual_col_names) -> original names (what appears in the dlt schema)
+    escaped_to_dlt = {
```
rudolfix: code is correct but you escape dlt names to destination names here (and this is correct)

dlt/destinations/utils.py
```diff
@@ -292,3 +298,152 @@ def get_deterministic_temp_table_name(table_name: str, op: str) -> str:
...
+
+    if not possibly_dropped_col_names:
+        return {}
+
+    partial_table: TPartialTableSchema = {"name": table_name, "columns": {}}
+
+    for esc_name in possibly_dropped_col_names:
+        name_in_dlt = escaped_to_dlt[esc_name]
+
+        if disregard_dlt_columns and name_in_dlt in [C_DLT_ID, C_DLT_LOAD_ID]:
```
rudolfix: never use any raw string to compare with dlt or destination identifiers. `schema` has `.startswith(self._dlt_tables_prefix)` to recognize dlt entities. please refactor it into a method on schema (`is_dlt_entity`) and replace all usages of `_dlt_tables_prefix` with it, also here

dlt/destinations/utils.py
```diff
@@ -292,3 +298,152 @@ def get_deterministic_temp_table_name(table_name: str, op: str) -> str:
...
+    Returns:
+        Optional[TSchemaDrop]: Returns the update that was applied to the schema.
+    """
+    tables = table_names if table_names else schema.data_table_names()
+
+    table_drops: TSchemaDrop = {}  # includes entire tables to drop
+    column_drops: TSchemaDrop = {}  # includes parts of tables to drop as partial tables
+
+    # 1. Detect what needs to be dropped
+    actual_table_col_schemas = dict(client.get_storage_tables(tables))
```
rudolfix: better to pass this to function, not the client

dlt/destinations/utils.py
```diff
@@ -292,3 +298,152 @@ def get_deterministic_temp_table_name(table_name: str, op: str) -> str:
...
+                f"An incremental field {name_in_dlt} is being removed from schema."
+                "You should unset the"
+                " incremental with `incremental=dlt.sources.incremental.EMPTY`"
+            )
+        partial_table["columns"][name_in_dlt] = col_schema
+
+    return partial_table if partial_table["columns"] else {}
+
+
+def update_dlt_schema(
```
rudolfix: `sync_schema_from_storage_schema` is I think better

rudolfix (commented):
this is really good and is doing a lot of refactoring which we need. but we need even more :)

also when this is finished we need two new cli commands that will use it:
dlt pipeline fruitshop schema [foo] sync-to-destination (attach to pipeline, use old sync method)
dlt pipeline fruitshop schema [foo] sync-from-destination (the reverse operation, with optional sync back to destination)

dlt pipeline fruitshop schema [foo] should still show schema (you can add "show" as well)

heh a lot of work... sorry for that
is dry run possible for the destructive sync?

dlt/destinations/utils.py
anuunchin: I see how the comment is confusing - fixed it

dlt/destinations/utils.py
anuunchin: moved the reverse logic function to the schema - it's definitely cleaner

dlt/common/destination/client.py
anuunchin: added an explanation with an example here (so that it has slightly more info than the sql implementation)

anuunchin:
Closing due to inactivity

PR #2912, opened 2025-07-21
Docs: Some additions to the filesystem gdrive source

### LINKED ISSUE #2907: Filesystem source from google drive: the file path specification doesn't work
### dlt version

1.14.1

### Describe the problem

This doesn't work
```python
filesystem_source = filesystem(
    bucket_url="gdrive://{folder_id}/test_file.csv",
    #file_glob="*.csv"
) | read_csv_duckdb()

pipeline = dlt.pipeline(pipeline_name="my_pipeline", destination="duckdb", dev_mode=True)

info = pipeline.run(filesystem_source)
print(info)

ds = pipeline.dataset()
rel = ds["_read_csv_duckdb"].df()
print(rel)
```

but this does:
```python
filesystem_source = filesystem(
    bucket_url="gdrive://{folder_id},
    file_glob="*.csv"
) | read_csv_duckdb()

pipeline = dlt.pipeline(pipeline_name="my_pipeline", destination="duckdb", dev_mode=True)

info = pipeline.run(filesystem_source)
print(info)

ds = pipeline.dataset()
rel = ds["_read_csv_duckdb"].df()
print(rel)
```

Additionally, from docs, it's not clear that the file in google drive must be explicitly shared with the service account.

### Expected behavior

The file extension should work, or docstring and docs should be corrected.

### Steps to reproduce

See code above.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

Filesystem + google drive

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
### Description
This PR just adds a note that if the bucket url is a specific file path, the file_glob needs to be set to an empty string. The issue was raised in community, which I opened here #2907. 

Additionally, an explicit step to share the relevant folders/files with the service account is added to the setup part, as the user was confused since the docs don't mention it.

### Related Issues

- Resolves #2907

### DIFF
--- a/docs/website/docs/dlt-ecosystem/verified-sources/filesystem/basic.md
+++ b/docs/website/docs/dlt-ecosystem/verified-sources/filesystem/basic.md
@@ -106,6 +106,7 @@ To get GCS/GDrive access:
 4. In IAM & Admin > Service Accounts, find your account, click the three-dot menu > "Manage Keys" >
    "ADD KEY" > "CREATE" to get a JSON credential file.
 5. Grant the service account appropriate permissions for cloud storage access.
+6. In the case of GDrive, share the respective folders/files with the service account.
 
 For more info, see how to
 [create a service account](https://support.google.com/a/answer/7378726?hl=en).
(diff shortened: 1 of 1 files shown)


VioletM (approved):
Amazing and short!

PR #2915, opened 2025-07-22
QoL: improve DataValidationError output:  use identifying columns if present

### LINKED ISSUE #1772: Show offending item in logs when schema_contract is set to "freeze"
### Feature description

When using `schema_contract="freeze"`, if a pipeline load fails because a new record is loaded that doesn't match the original type of the column, the offending record/item should be printed in the logs so that the user can figure out what to do about it. 

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Right now if you set the `schema_contract` to freeze and the pipeline is trying to load a new record that has a different type you get a log along the lines of:
```
Trying to create new variant column something__v_text to table my_table but data_types are frozen..
```
Even with debug mode, dlt doesn't show you which item has caused this, which make sit very hard to debug.

Slack convo: https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1724789620799849

### Proposed solution

_No response_

### Related issues

_No response_

djudjuu:
- Fixes #1772

### DIFF
--- a/dlt/common/schema/exceptions.py
+++ b/dlt/common/schema/exceptions.py
@@ -166,8 +166,7 @@ class DataValidationError(SchemaException):
 
         Schema, table and column names are given as a context and full `schema_contract` and causing `data_item` as an evidence.
         """
-        msg = ""
-        msg += f"Table: `{table_name}` "
+        msg = f"Table: `{table_name}` "
         if column_name:
             msg += f"Column: `{column_name}`"
         msg = (


dlt/common/schema/exceptions.py
```diff
@@ -178,6 +178,25 @@ def __init__(
...
+            and hasattr(data_item, "get")
+            and table_schema
+            and hasattr(table_schema, "get")
+        ):
+            identifier_columns = [
+                x.get("name")
+                for x in table_schema.get("columns", {}).values()
+                if x.get("primary_key") or x.get("merge_key") or x.get("unique")
+            ]
+            if identifier_columns:
```
djudjuu, line 190: i decided to not show any information about the data item if there is no info on what columns are relevant.
I was thinking to maybe do `str(data_item.get("first_column_name")[:50]` or something like that but it seemed a bit risky to me. what if the column is super huge or is bytes or whatnot?

dlt/common/schema/exceptions.py
```diff
@@ -178,6 +178,25 @@ def __init__(
             + f" . Contract on `{schema_entity}` with `{contract_mode=:}` is violated. "
             + (extended_info or "")
         )
+        data_item_str = ""
+        if (
+            data_item
+            and hasattr(data_item, "get")
+            and table_schema
+            and hasattr(table_schema, "get")
+        ):
```
djudjuu, line 184: too cautious? can I assume? could be more elegant to wrap this in try except....

dlt/common/schema/exceptions.py
```diff
@@ -178,6 +178,25 @@ def __init__(
             + f" . Contract on `{schema_entity}` with `{contract_mode=:}` is violated. "
```
anuunchin, line 175: not relevant specifically to this line, but I just noticed that the error message is:
```
dlt.common.schema.exceptions.DataValidationError: In schema `test_pipeline_1`: In Schema: `test_pipeline_1` Table: `my_other_resource`  . Contract on `tables` with `contract_mode=freeze` is violated. Can't add table `my_other_resource` because `tables` are frozen.
```

where `In schema: ` is duplicated, because base exception `SchemaException` also includes it. Maybe we should remove the duplicate here

dlt/common/schema/exceptions.py
anuunchin: I was under the impression that there is no such thing as being `too cautious` 👀  in SE

dlt/common/schema/exceptions.py
```diff
@@ -178,6 +178,25 @@ def __init__(
...
+                for x in table_schema.get("columns", {}).values()
+                if x.get("primary_key") or x.get("merge_key") or x.get("unique")
+            ]
+            if identifier_columns:
+                data_item_str += "Offending data item: "
+                data_item_keys = [
+                    f"{column}: {data_item.get(column)}" for column in identifier_columns
+                ]
+                data_item_str += ", ".join(data_item_keys)
+                msg += f" {data_item_str}"
```
anuunchin, line 196: I wanted to see what the message was with:

```python
@dlt.resource
def my_resource():
    for i in range(20):
        yield {"id": i, "age": 58}

@dlt.resource
def my_other_resource():
    for i in range(20):
        yield {"id": i, "age": 58, "name": "on"}


pipeline = dlt.pipeline(pipeline_name="test_pipeline_1", destination="duckdb", progress="log")
load_info = pipeline.run([my_resource(), my_other_resource()], write_disposition="replace", schema_contract={"columns": "freeze"})
print(load_info)
```

To `my_other_resource` I added another column `{"rand": 9}` and reran. As a result, the message I get is:

```
dlt.common.schema.exceptions.DataValidationError: In schema `test_pipeline_1`: In Schema: `test_pipeline_1` Table: `my_other_resource` Column: `rand` . Contract on `columns` with `contract_mode=freeze` is violated. Can't add table column `rand` to table `my_other_resource` because `columns` are frozen. Offending data item: _dlt_id: HXxljcwl2CGb2g
```

Which seems not to be the intention 👀

djudjuu: it kind of is. I don't want to show the entire object, because who knows how big that would be....
it doesnt show the id, because id is not a primary_key. 
i agree that like this it's not super helpful. maybe I can include a message about the load_package that this _dlt_id can be found in

djudjuu: wait. "column `rand`" wtf?

djudjuu: actually, its already there.
```
Job for `job_id=my_other_resource.dbb0a215e6.typed-jsonl` failed terminally in load with `load_id=1753871107.1320758` with message: In schema `test_pipeline_3`: In Table: `my_other_resource` Column: `rand` . Contract on `columns` with `contract_mode=freeze` is violated. Can't add table column `rand` to table `my_other_resource` because `columns` are frozen. Offending data item: _dlt_id: 3d6wzXJxGsKqSQ.
```

dlt/common/schema/exceptions.py
djudjuu: good one! done

dlt/common/schema/exceptions.py
```diff
@@ -167,8 +167,6 @@ def __init__(
         Schema, table and column names are given as a context and full `schema_contract` and causing `data_item` as an evidence.
         """
         msg = ""
```
anuunchin: Do we need the `msg = ""` part 👀

anuunchin: since we directly do
```python
msg += f"Table: `{table_name}` "
```

dlt/common/schema/exceptions.py
anuunchin: I don't know why I was confused by the _dlt_id the other day - Now that I think of it, intuitively speaking, users should check that item in the extracted package

dlt/common/schema/exceptions.py
djudjuu: good one.

PR #2917, opened 2025-07-22
restclient: json param range paginator

### LINKED ISSUE #2910: json body param for PageNumberPaginator
### Feature description

Add a way to paginate a "page number" API pagination type sending the next page as json body field instead of query param.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

For a real world usage of the Rest API source, I needed a PageNumber paginator, but instead of sending the "next page" as query param, the API required it in the POST request body.

### Proposed solution

I propose to add a `page_body_path` argument to `PageNumberPaginator`, similar to the `curson_body_path` used in `JSONResponseCursorPaginator`

### Related issues

_No response_

Giackgamba:
### Description

Add `*_body_path`parameters for both `RangePaginator`s (`PageNumberPaginator` and `OffsetPaginator`. This solves cases where the parameters are needed in the json body of a -usually- `POST` request.

### Related Issues

- Fixes #2910

### Additional Context
Looks like some tests are failing, but I think they are not related to this PR.

### DIFF
--- a/dlt/sources/helpers/rest_client/paginators.py
+++ b/dlt/sources/helpers/rest_client/paginators.py
@@ -92,7 +92,6 @@ class RangePaginator(BasePaginator):
     def __init__(
         self,
         param_name: Optional[str],
-        body_path: Optional[jsonpath.TJsonPath],
         initial_value: int,
         value_step: int,
         base_index: int = 0,
@@ -101,11 +100,15 @@ class RangePaginator(BasePaginator):
         error_message_items: str = "items",
         stop_after_empty_page: Optional[bool] = True,
         has_more_path: Optional[jsonpath.TJsonPath] = None,
+        param_body_path: Optional[jsonpath.TJsonPath] = None,
     ):
         """
         Args:
             param_name (str): The query parameter name for the numeric value.
                 For example, 'page'.
+            param_body_path (jsonpath.TJsonPath): The JSONPath expression specifying
+                where to place the numeric parameter in the request JSON body.
+                Defaults to `None`.
             initial_value (int): The initial value of the numeric parameter.
             value_step (int): The step size to increment the numeric parameter.
             base_index (int, optional): The index of the initial element.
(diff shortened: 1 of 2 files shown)


burnash:
Thank you for contributing, @Giackgamba! I'll check the PR out shortly.

dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -91,7 +91,8 @@ class RangePaginator(BasePaginator):
 
     def __init__(
         self,
-        param_name: str,
+        param_name: Optional[str],
+        body_path: Optional[jsonpath.TJsonPath],
```
burnash: 1. `body_path` needs to be moved to keyword arguments from positional arguments, otherwise, it'll break for users who passed positional arguments. (Ex: RangePaginator("page", ...) → now expects param_name, body_path)
2. It's best to use a more specific name (e.g. in line with `page_body_path` -> `param_body_path`) to make it even clearer.

dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -303,7 +321,8 @@ def __init__(
         self,
         base_page: int = 0,
         page: int = None,
-        page_param: str = "page",
+        page_param: Optional[str] = None,
+        page_body_path: Optional[jsonpath.TJsonPath] = None,
```
burnash: Please put `page_body_path` as a last keyword argument. I overlooked this early on when there were not so many arguments, but now I plan to gradually enforce keyword-only arguments to all of paginators.

dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -442,8 +474,10 @@ def __init__(
         self,
         limit: int,
         offset: int = 0,
-        offset_param: str = "offset",
-        limit_param: str = "limit",
+        offset_param: Optional[str] = None,
+        offset_body_path: Optional[jsonpath.TJsonPath] = None,
+        limit_param: Optional[str] = None,
+        limit_body_path: Optional[jsonpath.TJsonPath] = None,
```
burnash: Here as well, please put new `*_body_path` arguments at the end.

burnash (changes requested):
Looks very good! Thank you @Giackgamba for the tests and documentation. I have a couple of comments regarding the order of arguments. Please take a look.

PR #2924, opened 2025-07-23
restclient: misc Paginators improvements

burnash:
A follow up to #2910:

- Added keyword-only arguments separator to REST client paginators (RangePaginator, PageNumberPaginator, OffsetPaginator)
- Changed body paths from TJsonPath to str type in jsonpath module
- Added parameter validation 
- Added new tests for paginators

### DIFF
--- a/dlt/sources/helpers/rest_client/paginators.py
+++ b/dlt/sources/helpers/rest_client/paginators.py
@@ -101,13 +101,13 @@ class RangePaginator(BasePaginator):
         stop_after_empty_page: Optional[bool] = True,
         *,
         has_more_path: Optional[jsonpath.TJsonPath] = None,
-        param_body_path: Optional[jsonpath.TJsonPath] = None,
+        param_body_path: Optional[str] = None,
     ):
         """
         Args:
             param_name (str): The query parameter name for the numeric value.
                 For example, 'page'.
-            param_body_path (jsonpath.TJsonPath): The JSONPath expression specifying
+            param_body_path (str): The dot-separated path specifying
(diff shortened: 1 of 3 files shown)


dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -141,6 +143,10 @@ def __init__(
                 "One of `total_path`, `maximum_value`, `has_more_path`, or `stop_after_empty_page`"
                 " must be provided."
             )
+
+        if (not param_name and not param_body_path) or (param_name and param_body_path):
```
rudolfix: this `bool(param_name) == bool(param_body_path)` works as well

dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -482,9 +495,10 @@ def __init__(
         total_path: Optional[jsonpath.TJsonPath] = "total",
         maximum_offset: Optional[int] = None,
         stop_after_empty_page: Optional[bool] = True,
+        *,
```
rudolfix, line 498: why do we need this? to indicate rarely used arguments?

rudolfix (commented):
only minor comments. you can ignore them or merge...

dlt/sources/helpers/rest_client/paginators.py
burnash: No, I should have done this before to make it easier to do backward compatible changes. And I'd plan to tighten it to make everything kw only except limit and offest

PR #2929, opened 2025-07-24
feat: `Schema.to_dbml()`

zilto:
### Description
This allows to call `pipeline.default_schema.to_dbml()` and generate a string that can be rendered in a DBML frontend

image

This supports:
- data and dlt tables
- table metadata
- column metadata
- user-defined references
- root-child references
- parent-child references
- internal dlt tables references
- group by dlt tables by resource (might not be displayed in some DBML frontend)

### Future work
- leverage `_from_dbml_table`, `_from_dbml_column` and `_from_dbml_reference` to create two-way communication synchronization between dlt and DBML formats
- build widget to render DBML
- adding "table discovery" using SQLAlchemy reflection or  [SQLGlot DDL to dlt](https://github.com/dlt-hub/dlt/pull/2748), we could view entire destination schemas
- currently `Table.properties` and `Column.properties` are used internally to store metadata on the PyDBML object. They are not rendered in the DBML string. Though, [we can enable rendering](https://github.com/Vanderhoof/PyDBML/blob/master/docs/properties.md). This would allow to view and edit custom dlt hints
- modify `TTableReference` to include optional `from_table` and `label` fields. This would simplify code and add semantics to the relationship
- more tests for new utils functions
- more tests for schema conversion


dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+def _to_dbml_column(hints: TColumnSchema) -> Column:
+    """Convert a dlt column schema to a dbml column
+
+    `properties` are not rendered; we use them to store the original hints.
+
+    NOTE: dbml column has boolean properties and doesn't support `None`.
+    For example, we can't disambiguate `unique=False` in DBML from
+    `unique=False or `unique=None` in dlt.
```
rudolfix: there's just one property where `None` is something more than undefined (is actual values). it is one of x-props that we use for scd-2. 

we can also use a convention like `x-none` will translate to Python None.

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+    return Column(
+        name=hints["name"],
```
rudolfix: I'd do two improvements there.

you can automate Column generation by enumerating hints and creating entry for each in Column:
- you have definitions of known hints in `ColumnPropInfos` in `schema.typing` module. 
- those definitions contain default value. you do not need to do `hints.get("unique") is True` where `True` is hardcoded. see how `has_default_column_prop_value` works
- for all `x-hints` (not in the dict above) `None` is the default. but you do not need to worry - `Schema` calls `remove_defaults` in `to_dict` so default values are not saved
- anything that cannot be represented as property goes to `properties` (ie. all dicts like `x-normalize`)

this let us add new hints (or promote existing one to "official") by just declaring them in `ColumnPropInfos`. I still want to create a real registry of props ie. with propagation strategy etc.

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+
+    return output
+
+
+def _to_dbml_column(hints: TColumnSchema) -> Column:
+    """Convert a dlt column schema to a dbml column
+
+    `properties` are not rendered; we use them to store the original hints.
+
+    NOTE: dbml column has boolean properties and doesn't support `None`.
```
rudolfix: w8 so we cant have string props? we for example have `dedup_sort`: asc / desc

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+    return Column(
+        name=hints["name"],
+        type=hints["data_type"],
+        unique=hints.get("unique") is True,
+        not_null=hints.get("nullable") is False,
+        pk=hints.get("primary_key") is True,
+        note=hints.get("description"),
+        # we stringify because of `pydbml` constraints
+        properties=_stringify_dict(hints),
```
rudolfix: we should try to minimize this one to get as human readable text file as possible.

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+    )
+
+
+def _from_dbml_column(column: Column) -> TColumnSchema:
+    """Convert a dbml column to a dlt column schema"""
+    original_hints = _destringify_dict(column.properties)
+    retrieved_hints = {
+        "name": column.name,
+        "data_type": column.type,
+        "unique": None if original_hints.get("unique") is None else column.unique,
```
rudolfix: I think a dict of renames dlt <-> dbml would be good

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+    """Convert a dlt table schema to a dbml table
+
+    `properties` are not rendered; we use them to store the original hints.
+    """
+    columns = [_to_dbml_column(col) for col in table["columns"].values()]
+    return Table(
+        name=table["name"],
+        columns=columns,
+        note=table.get("description"),
+        properties=_stringify_dict(table),
```
rudolfix: so we stringify the whole table including columns? could we just stringify table props that are not a part of `TTableSchema`. procedure could be as in case of column: properties contain dlt hints that could not be represented as dbml props.

this will improve human readability of it.

dlt/common/schema/schema.py
```diff
@@ -737,6 +737,26 @@ def to_pretty_yaml(
...
+        self,
+        include_dlt_tables: bool = True,
+        include_internal_dlt_ref: bool = True,
+        include_parent_child_ref: bool = True,
+        include_root_child_ref: bool = True,
+        group_by_resource: bool = False,
+    ) -> str:
+        from dlt.helpers.dbml import schema_to_dbml
+
+        dbml_schema = schema_to_dbml(
```
rudolfix, line 761:
```py
d = self.to_dict(
            remove_defaults=remove_defaults, remove_processing_hints=remove_processing_hints
        )
```
is a crucial part of save process. it makes sure
- that schema version and hash are bumped and synced
- that various default values and Nones are removed
- that non-serializable and transient part of the schema are removed

`remove_defaults: bool = True, remove_processing_hints: bool = False` are part of the standard interface. they are used by schema storage ie. to export schemas to `export` folder (processing hints will be removed)

tldr;> make your convert function to work on `TStoredSchema`, not on `Schema` to get clean and serializable representation of `Schema`

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+TDBMLReferenceCardinality = Literal["<", ">", "-", "<>"]
+
+
+def _stringify_dict(d: Mapping[str, Any]) -> dict[str, str]:
+    return {k: str(v) for k, v in d.items()}
+
+
+def _destringify_dict(d: Mapping[str, str]) -> dict[str, Union[bool, None, str]]:
```
rudolfix, line 52: Is this for human readability? stringify with json is not enough?

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,390 @@
...
+    include_parent_child_ref: bool = True,
+    include_root_child_ref: bool = True,
+    group_by_resource: bool = False,
+    **kwargs: Any,
+) -> Database:
+    """Convert a dlt.Schema object to a PyDBML representation."""
+
+    dbml_schema = Database()
+
+    _add_tables(schema, dbml_schema, include_dlt_tables=include_dlt_tables)
```
rudolfix, line 437: if you use `TStoredSchema` (see above) you'll see certain schema props that should be saved. ie version and version hash. does `dbml` have a version?

tests/helpers/test_dbml.py
```diff
@@ -0,0 +1,628 @@
...
+    _from_dbml_table,
+    _to_dbml_reference,
+    _from_dbml_reference,
+    _group_tables_by_resource,
+)
+
+
+@pytest.fixture
+def example_schema() -> dlt.Schema:
+    return dlt.Schema.from_dict(
```
rudolfix, line 27: please take a look at tests/common/cases - there you'll see a few really complex schemas ie `ethereum` one. you can also all the one below to cases

tests/helpers/test_dbml.py
```diff
@@ -0,0 +1,628 @@
...
+
+    output = export_to_dbml(example_schema, path=None)
+    loaded_dbml = PyDBML(output)
+
+    assert isinstance(output, str)
+    # for some reason, the PyDBML objects are not directly equal
+    assert expected_output.dbml == loaded_dbml.dbml
+
+
+def test_export_to_dbml_to_file(
```
rudolfix: cool test but also try to import-export all the schema cases I mentioned abve and compare them because they should be identical. basically import-export should be identity function

rudolfix (changes requested):
this is good! thanks for writing test.

besides the review comment, please try to integrate this new format with a few dlt functions. should be fairly easy:

1. automatic import /export functionality
https://dlthub.com/docs/walkthroughs/adjust-a-schema

look for `schema_storage` module and:

`def _export_schema(` will get you started

2. `cli` support
- `PipelineCommand` has schema export function
- `SchemaCommand` is basically schema converter cli

dlt/helpers/dbml.py
zilto: You're right! I should use `{k: json.dumps(v) ...}` and `{k: json.loads(v) ...}` instead

It's not for human-readability. The constraint is that `pydbml.classes.Column().properties` and `pydbml.classes.Table().properties` only accept `dict[str, str]`.

dlt/helpers/dbml.py
zilto: I now clarified the comment. 

No. I wanted to note that attributes of `pydbml.classes.Column` and `pydbml.classes.Table` are booleans and don't accept `None`. (For example, `Column(not_null=...)` has to be `True` or `False`. 

Both classes support arbitrary fields (that don't collide) via `properties`.
```python
Column(
  name="foo",
  type="string",
  not_null=True,
  properties={"dedup_sort": "asc", "x-pii": True},
)
```

Then, in the DBML renderer, you can decide to include/exclude `properties` field in the export.

dlt/helpers/dbml.py
zilto: In the DBML renderer, you can decide to include/exclude properties field in the export.

We can store as much metadata as we want here.

dlt/helpers/dbml.py
zilto: Yes, you're right! Good catch

dlt/helpers/dbml.py
zilto: DBML allows "settings", which are key-value pairs outside of the first-class properties (primary key, nullable, etc.). Currently, most frontends I tried will error on custom settings. Though there's an open issue about it: https://github.com/holistics/dbml/issues/31

- [Column-level settings](https://dbml.dbdiagram.io/docs/#column-settings)
- [Table-level settings](https://dbml.dbdiagram.io/docs/#table-settings)
- [Project-level properties](https://dbml.dbdiagram.io/docs/#project-definition): this does currently support arbitrary key-value, but nothing is displayed

### TODO
We should not render custom dlt hints until custom settings are supported. When it is supported, we can simply toggle `allow_properties=True` in `PyDBML`

tests/helpers/test_dbml.py
zilto: I prefer the readability of having `tests/helpers/test_dbml.py` being self-contained. There's no need to switch files and I can modify `example_schema` to match my test requirements without contaminating other tests. 

Once we have two-sync implemented (`schema_to_dbml()` and `dbml_to_schema()`) we could parameterize the associated tests

tests/helpers/test_dbml.py
zilto: agreed. Will implement when two sync  is added.

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,521 @@
...
+__all__ = (
+    "schema_to_dbml",
+    "export_to_dbml",
+)
+
+# dlt sets no `data_type` field for columns that haven't seen data yet.
+# Since DBML doesn't support empty `type` field, we set a constant
+# NOTE should we use `UNKNOWN` or `None` or something else?
+UNKNOWN_DATA_TYPE = "UNKNOWN"
+TDBMLReferenceCardinality = Literal["<", ">", "-", "<>"]
```
rudolfix, line 45: see comment in #2941

docs/website/docs/general-usage/dataset-access/view-dlt-schema.md
```diff
@@ -0,0 +1,837 @@
+---
+title: Review dlt schema
+description: View your dlt schema via files, CLI, static and interactive diagram
+keywords: [schema, dataset, view, dbml, graphviz]
+---
+
+# Review dlt schema
+
+During the first `dlt.Pipeline` run, dlt produces a `dlt.Schema` from the data processed. This schema tells how data was stored on destination.
```
rudolfix, line 9: here you show how to review schema programmatically. there are two other ways:
- automatic import/export (adjust-a-schema.md)
- command line: you can display default schema in pipeline in any format

I'd add those as reference or as a tip.

dlt/helpers/dbml.py
```diff
@@ -0,0 +1,521 @@
...
+        type=column_hints.get("data_type", UNKNOWN_DATA_TYPE),
+        unique=bool(column_hints.get("unique")),
+        not_null=not bool(column_hints.get("nullable", True)),
+        pk=bool(column_hints.get("primary_key")),
+        note=column_hints.get("description"),
+        properties=None if properties == {} else _stringify_dict(properties),
+    )
+
+
+def _from_dbml_column(column: Column) -> TColumnSchema:
```
rudolfix, line 109: do you want to keep import code in this PR? that's OK if you do a followup quickly

dlt/helpers/dbml.py
zilto: This is not my design choice. It's the PyDBML API: https://github.com/Vanderhoof/PyDBML/blob/5c6a9173580d7273a6799cdd725eabd3a2e76911/pydbml/_classes/reference.py#L25

dlt/helpers/dbml.py
zilto: These functions are required for testing. "DBML to DLT" is not the scope of this PR and should be done in a follow-up (I would wait to see if DBML supports custom properties in the upcoming months)

zilto:
> When I remove curly braces all works. Maybe you could change the renderer behavior? it looks like the newest syntax is not widely supported
> 
> Extensions tested
> vscode-dbml: highlighting OK but does not understand curly braces.
> DBML Entity-Relationship Diagrams visualizer: could not make it work :/ and it looks really good. maybe you can try it out?
> DBML Live Preview: works after curly braces are converted

PyDBML [states syntax is valid for `v3.9.5`](https://github.com/Vanderhoof/PyDBML?tab=readme-ov-file#dbml-parser-for-python) with [current version being `3.13`](https://github.com/holistics/dbml/tags). The DBML output works on dbdiagram(.)io (by the creators of the specs), ChartDB, the VSCode extension `bocovo.dbml-erd-visualizer` (16k stars). Also the extension you mention `vscode-dbml` is labeled as unmaintained.

rudolfix (approved):
LGTM! amazing work!

Issue #2669, opened 2025-05-22
ConnectorX not supporting decimals/numeric


Rhiyo:
### Documentation description

Current DLT docs give a warning explaining that ConnectorX will convert decimals to floats and lose precision. As of the latest connectorx version this has been rectified. Would be good to update the docs and any dependencies to support this.

- DLT Docs (now semi-outdated) https://dlthub.com/docs/dlt-ecosystem/verified-sources/sql_database/configuration#connectorx
- PR That fixed this in ConnectorX https://github.com/sfu-db/connector-x/pull/806

### Are you a dlt user?

Yes, I'm already a dlt user.

zilto:
To give you an update, we're looking to upgrade to the latest connectorx (`0.4.3`), but are currently investigating [an issue](https://github.com/sfu-db/connector-x/issues/811) with pyarrow

zilto:
documentation was updated

Issue #2661, opened 2025-05-19
ConnectorX 'arrow2' ValueError when running SQL database pipeline


dat-a-man:
### dlt version

1.11.0

### Describe the problem

Running a SQL pipeline using dlt's `sql_database` source results in the following error due to the use of an unsupported `return_type` in ConnectorX:
ValueError: arrow2

This is triggered when calling `connectorx.read_sql(...)`.

### To Reproduce

1. Create a SQL source using dlt's `sql_database` helpers.
2. Use a pipeline to run extraction.
3. ConnectorX 0.4.2 is installed.
4. Run: `python3 sql_database_pipeline.py`

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12

### dlt data source

sql_database

### dlt destination

Google BigQuery

### Full Stack Trace
```
(.venv) amanguptanalytics@sandbox-vm:~/16_test_sling/tests/dlthub$ python3 sql_database_pipeline.py 
psutil dependency is not installed and mem stats will not be available. add psutil to your environment or pass dump_system_stats argument as False to disable warning.
/home/amanguptanalytics/16_test_sling/tests/.venv/lib/python3.11/site-packages/google/cloud/bigquery/client.py:595: UserWarning: Cannot create BigQuery Storage client, the dependency google-cloud-bigquery-storage is not installed.
  warnings.warn(
----------------------------- Extract sql_database -----------------------------
Resources: 0/6 (0.0%) | Time: 0.00s | Rate: 0.00/s

----------------------------- Extract sql_database -----------------------------
Resources: 0/6 (0.0%) | Time: 0.03s | Rate: 0.00/s

Traceback (most recent call last):
  File "/usr/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/amanguptanalytics/16_test_sling/tests/.venv/lib/python3.11/site-packages/dlt/extract/utils.py", line 260, in _parallel_gen
    return next(gen)  # type: ignore[call-overload]
           ^^^^^^^^^
  File "/home/amanguptanalytics/16_test_sling/tests/.venv/lib/python3.11/site-packages/dlt/sources/sql_database/helpers.py", line 303, in table_rows
    yield from loader.load_rows(backend_kwargs)
[...]

PR #2933, opened 2025-07-25
fix: replace `arrow2` with `arrow` backend for `connectorx`

zilto:
### Description
`Connectorx` dropped support for `arrow2` backend in version `0.4.2`. This was the default version used in `dlt`, creating a breaking change. 

issues: #2661 #2669 
related PR fixing timestamp issue: #2879 

### Changes
- `arrow` is now the default backend in `dlt`
- two small changes because of disparities between `arrow` and  `arrow2` behavior
- relax connectorx constraints in `pyproject.toml`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,7 +236,7 @@ dev = [
 # NOTE: those dependencies are used to test built in sources
 sources = [
     "connectorx>=0.3.3 ; python_version >= '3.9'",
-    "connectorx>=0.4.0,<0.4.2 ; python_version >= '3.10'",
+    "connectorx>=0.4.0 ; python_version >= '3.10'",
     "pymysql>=1.1.0,<2",
     "openpyxl>=3,<4",
     "mimesis>=7.0.0,<8",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -235,8 +235,8 @@ dev = [
 
 # NOTE: those dependencies are used to test built in sources
 sources = [
-    "connectorx>=0.3.3 ; python_version >= '3.9'",
-    "connectorx>=0.4.0,<0.4.2 ; python_version >= '3.10'",
+    "connectorx>=0.4.2 ; python_version >= '3.9'",
```
rudolfix: looks like support for Python 3.9 got dropped (`uv` does not resolve this dep on ci). my take: exclude Python 3.9. AFAIK we are not testing core source on it

rudolfix (changes requested):
LGTM! please look at Python 3.9 uv resolve problem.

decimals should now be present in the data returned from connectorx. this will be visible on `minimal` reflection level in test_reflection_levels. On higher levels we coerce to database schema so also previously we had decimals there (albeit converted from doubles in case of connectorx/pandas)

pyproject.toml
zilto: The issue with `uv` is the global setting  `requires-python = ">=3.9.2, <3.15"`. I'm surprised this applied to optional groups too.

I added back support for `connectorx<0.4.2`. After all, previous versions also supported the `arrow` backend

rudolfix (approved):
LGTM!

PR #2935, opened 2025-07-27
skips inferring incomplete column when already incomplete

### LINKED ISSUE #2928: Memory spike caused by not-incremental handling of schema updates
### dlt version

1.14.1

### Describe the problem

In [worker.py](https://github.com/dlt-hub/dlt/blob/devel/dlt/normalize/worker.py#L254) schema updates are saved per single record, without any intermediate simplification/deduplication.
This can bring to cases when you have a big schema, for a decent number of records, and eat a lot of RAM. 
In my case, for 570k records of 600 something columns, the schema updates object was taking up about 20gb of RAM. 

Image

There are a few cases reported in Slack of memory spikes in the normalize step, I believe this is a potential cause. 

### Expected behavior

dlt handles gracefully large amounts of records by incrementally checking for schema changes (or at least, aggregates them at some level)

### Steps to reproduce

Pick any large source, run it from scratch (with no initial schema) and check memory usage and how many schema updates are detected

### Operating system

macOS

### Runtime environment

Kubernetes

### Python version

3.12

### dlt data source

Custom Salesforce source based on the community one.

### dlt destination

Snowflake

### Other deployment details

I've run this both in a Kubernetes cluster and locally, both reproduce the issue.

### Additional information

I think a quick workaround could be to deduplicate the schema updates at least within the single file: 

```python
schema_updates.extend(deduplicate_partial_updates(partial_updates))
```

with 

```python
def deduplicate_partial_updates(
    partial_updates: List[TSchemaUpdate],
) -> List[TSchemaUpdate]:
    consolidated_update = {}
    for update in partial_updates:
        for table, records in update.items():
            if table not in consolidated_update:
                consolidated_update[table] = []
            for record in records:
                if record not in consolidated_update[table]:
                    consolidated_update[table].append(record)
    return [consolidated_update]
```

i have tried to be as less intrusive as possible in the dlt code, but you could expand this method (for example to make it work across different files). 

Before the change:

Image

After: 

Image

rudolfix:
### Description
fixes #2928

### DIFF
--- a/dlt/common/schema/schema.py
+++ b/dlt/common/schema/schema.py
@@ -833,11 +833,16 @@ class Schema:
         if existing_column and utils.is_complete_column(existing_column):
             if not utils.is_nullable_column(existing_column):
                 raise CannotCoerceNullException(self.name, table_name, col_name)
-        elif not existing_column:
-            inferred_unbounded_col = self._infer_column(
-                k=col_name, v=None, data_type=None, table_name=table_name
-            )
-            return inferred_unbounded_col
+        else:
+            # generate unbounded column only if it does not exist or it does not
+            # contain seen null
+            if not existing_column or not existing_column.get("x-normalizer", {}).get(
+                "seen-null-first"
+            ):
+                inferred_unbounded_col = self._infer_column(
+                    k=col_name, v=None, data_type=None, table_name=table_name
+                )
+                return inferred_unbounded_col
         return None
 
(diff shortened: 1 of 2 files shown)


anuunchin (dismissed):
Agh, my bad, this is a clear logic bug I introduced

rudolfix:
@anuunchin need another review.

PR #2936, opened 2025-07-28
fix: restclient: handle null data in response

### LINKED ISSUE #2932: JSONResponseCursorPaginator breaks on empty page
### dlt version

1.14.1

### Describe the problem

I'm extracting data from a REST API where I use `JSONResponseCursorPaginator` to handle pagination.

Here is an example:

First call

```
{
    "users": [...], // 10 items;
     "next_token": "..."
}
```

Second call with next_token in the headers

```
{
   "users": [...], // 4 items
   "next_token": "..."
}
```

At this point there is no more data to retrieve. But since the API gives a `next_token` in the response, dlt performs a last call.

Third call
```
{
  "users": null,
  "next_token": null
}
```

and `"users": null` is breaking the pipeline.

At this point, dlt should know that the last HTTP call is just an empty page and is not relevant data to load

### Expected behavior

_No response_

### Steps to reproduce

See description

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

REST API

### dlt destination

Snowflake

### Other deployment details

_No response_

### Additional information

_No response_

burnash:
Fixes #2932

### DIFF
--- a/dlt/sources/helpers/rest_client/client.py
+++ b/dlt/sources/helpers/rest_client/client.py
@@ -282,6 +282,8 @@ class RESTClient:
         logger.info(
             f"Extracted data of type {type(data).__name__} from path {data_selector}{length_info}"
         )
+        if data is None:
+            return []
         # wrap single pages into lists
         if not isinstance(data, list):
             data = [data]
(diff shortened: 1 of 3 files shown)


dlt/sources/helpers/rest_client/client.py
```diff
@@ -282,6 +282,8 @@ def extract_response(self, response: Response, data_selector: jsonpath.TJsonPath
         logger.info(
             f"Extracted data of type {type(data).__name__} from path {data_selector}{length_info}"
         )
+        if data is None:
```
rudolfix, line 285: maybe we should log the fact that None was found and replaced? this is a little hacky... and user should know

PR #2937, opened 2025-07-28
AI Command: extended with IDEs

anuunchin:
This PR adjust the ai command so that it is able to copy rule files based on the IDE. This PR is related to [this PR](https://github.com/dlt-hub/verified-sources/pull/637) in verified sources that uses continue's rules command to create ide specific rules. 

Additionally includes a small test to check if the rules are copied to the correct location.


dlt/cli/ai_command.py
```diff
@@ -1,57 +1,85 @@
...
+    "amp": "AGENT.md",
+    "codex": "AGENT.md",
+    "cody": ".sourcegraph",
+    "cursor": ".cursor",
+    "continue": ".continue",
+    "cline": ".clinerules",
+    "claude": "CLAUDE.md",
+    "windsurf": ".windsurf",
+    "copilot": ".github",
+}
```
anuunchin: This repetition should probably be properly dealt with

dlt/cli/ai_command.py
```diff
@@ -1,57 +1,85 @@
...
+
+SUPPORTED_IDES: Set[TSupportedIde] = set(get_args(TSupportedIde))
 VERIFIED_SOURCES_AI_BASE_DIR = "ai"
 
-# TODO Claude Desktop: rules need to be named `CLAUDE.md`, allow command to append to it
+# TODO Claude: rules need to be named `CLAUDE.md`, allow command to append to it
 # TODO Continue: rules need to be in YAML file, allow command to properly edit it
 # TODO generate more files based on the specifics of the source README and the destination
```
anuunchin, line 28: @zilto  i don't fully understand the third TODO 👀, Also it seems like continue rules are just .md files, or am I missing something

tests/cli/test_ai_command.py
```diff
@@ -0,0 +1,34 @@
...
+        return
+
+    if ide == "copilot":
+        expected_location /= "instructions"
+    elif ide in {"cursor", "continue", "windsurf"}:
+        expected_location /= "rules"
+
+    assert expected_location.is_dir()
+    md_files = list(expected_location.glob("*.mdc" if ide == "cursor" else "*.md"))
+    assert md_files
```
zilto: Thanks for adding a test. I think it could be easier to read if it was parameterized as:

```python
@pytest.mark.parametrize(
  ("ide", "expected_relative_path"),
  (
      ("cursor", "./.cursor/rules/dlt.mdc"),
      ("claude", "./CLAUDE.md"),
      ("cody", "./.sourcegraph/dlt.rule.md"),
      # etc. 
  )
)
def test_ai_setup_command(ide: TSupportedIde, expected_relative_path: str) -> None
```

[Zen #2](https://peps.python.org/pep-0020/#the-zen-of-python): Explicit is better than implicit

## Making future changes easy
Otherwise, there's a bit of a chicken-and-egg problem, but it would be nice to duplicate the test on `verified-sources`. Example scenario we would want to catch:

1. The Copilot rule format is updated; rules CLI is updated accordingly
2. `dlt ai setup` still retrieves successfully from `verified-sources` and this test succeeds, but rules are unusable in Copilot
3. we manually run `transpile-rules` in `verified-sources` to update the Copilot format
4. now, `dlt ai setup` fails to retrieve from `verified-sources` and this test fails
5. we update the CLI behavior **and the test** (it makes sense because our expectations also changed)
6. now things work

### Conclusions
- We can't automatically know when Copilot update its format
- When we update the rules in `verified-sources`, we should check if it doesn't break `dlt ai setup`; We want to know if we need to update `dlt`
- We expect to make changes to the test `test_ai_setup_command()` because our expectations will change. Therefore, explicit mapping will simplify maintenance vs `if/else` conditions

dlt/cli/ai_command.py
```diff
@@ -75,7 +103,9 @@ def ai_setup_command(
             % (fmt.bold(location), fmt.bold(branch or "<default>"))
         )
         return
-    src_dir = Path(src_storage.make_full_path(VERIFIED_SOURCES_AI_BASE_DIR)) / ide
+
+    suffix = LOCATION_IN_BASE_DIR.get(ide)
```
zilto: Rename variable from `suffix` to `sub_path`.  

In pathlib, [suffix refers to file extensions](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix). For example, `".json"` in `"path/to/file/schema.json"`. It can be accessed via `Path.suffix` and modified with `Path.with_suffix()`

Here, it seems to be a subpath with the `.joinpath()` or `/` operation.

dlt/cli/ai_command.py
```diff
@@ -1,57 +1,85 @@
...
+        relative_path = src_path.relative_to(src_root)
+        dest_file_path = dest_root / relative_path
+
         if dest_file_path.exists():
             if warn_on_overwrite:
                 fmt.warning(f"Existing rules file found at {dest_file_path.absolute()}; Skipping.")
             continue
 
-        copied_files.append(src_sub_path.name)
+        dest_file_path.parent.mkdir(parents=True, exist_ok=True)
```
zilto, line 58: Nice simplification

zilto (approved):
LGTM. There are some minor improvements suggested to the test cases and a variable name

dlt/cli/ai_command.py
anuunchin: oww

tests/cli/test_ai_command.py
anuunchin: woah, i wasn't aware of Python poetry 👀

zilto (approved):
Looks good to me!

tests/cli/test_ai_command.py
```diff
@@ -0,0 +1,46 @@
...
+        ("windsurf", ".windsurf/rules/", False),
+        ("copilot", ".github/instructions/", False),
+    ),
+)
+def test_ai_setup_command(
+    ide: TSupportedIde, expected_relative_path: str, is_single_file: bool
+) -> None:
+    with set_working_dir(TEST_STORAGE_ROOT):
+        ai_command.ai_setup_command(
+            ide=ide, location=DEFAULT_VERIFIED_SOURCES_REPO, branch="feat/continue-rules"
```
rudolfix: please remove this branch before merging. we need to merge verified sources first

tests/cli/test_init_command.py
```diff
@@ -617,6 +633,54 @@ def test_incompatible_dlt_version_warning(repo_dir: str, project_files: FileStor
...
+            init_command.init_command("dlthub:github", "duckdb", vibe_repo_dir)
+            _out = buf.getvalue()
+
+    assert "dlt will generate useful project rules tailored to your assistant/IDE." in _out
+    assert f"adding {ide_choice} rules, code snippets and docs" in _out
+    # Check that vibe source files are copied correctly
+    assert "file(s) supporting github were copied:\ngithub-docs.yaml\n" in _out
+
+
+def test_init_all_vibe_sources_together(vibe_repo_dir: str, project_files: FileStorage) -> None:
```
rudolfix, line 654: there are little chances this test will work. just one broken source in llm context will stop it. we'd need to add this linting in the vibe repo... my take: select 20 random vibe sources and hardcode them here.

rudolfix (changes requested):
LGTM! please fix the last test

tests/cli/test_init_command.py
```diff
@@ -617,6 +633,65 @@ def test_incompatible_dlt_version_warning(repo_dir: str, project_files: FileStor
...
+    assert f"adding {ide_choice} rules, code snippets and docs" in _out
+    assert "file(s) supporting github were copied:\ngithub-docs.yaml\n" in _out
+
+
+def test_init_all_vibe_sources_together(vibe_repo_dir: str, project_files: FileStorage) -> None:
+    # we test 20 hardcoded sources, use this to get all sources instead
+    # vibe_source_candidates = [*get_source_candidates(vibe_repo_dir, source_type="vibe")]
+    random_vibez = [
+        "news_api",
+        "alpaca",
```
rudolfix, line 659: now I want to see what alpaca API does :)

PR #2940, opened 2025-07-29
fix: missing `__repr__` for `@dlt.transformation`

zilto:
Added back the method from #2757 with related tests


rudolfix (approved):
LGTM! I'll cherry pick this to our dlt+ move branch

PR #2941, opened 2025-07-29
feat: add `label`, `table`, and `cardinality` fields to `TTableReference`

zilto:
Added fields on `TTableReference` that allows users to specify more metadata about the reference. 

Why:
- having `table` (left) in addition to `referenced_table` (right) specified directly on the `TTableReference` simplifies some logic.
- `cardinality` and `label` are just additional metadata, which can also be used in visualizations
- `table` would also allow `TTableReference` to be stored top-level on `dlt.Schema` (instead of inline on the `TTableSchema`)

### DIFF
--- a/dlt/common/schema/typing.py
+++ b/dlt/common/schema/typing.py
@@ -13,7 +13,7 @@ from typing import (
     NewType,
     Union,
 )
-from typing_extensions import Never
+from typing_extensions import Never, NotRequired
 
 from dlt.common.data_types import TDataType
 from dlt.common.normalizers.typing import TNormalizersConfig
@@ -276,14 +276,43 @@ TWriteDispositionConfig = Union[
 ]
 
 
+TReferenceCardinality = Literal["-", "<", ">", "<>"]
+"""Represents cardinality between `column` (left) and `referenced_column` (right)
+`-`: one-to-one
+`<`: one-to-many
+`>`: many-to-one
+`<>`: many-to-many
+
+Note that `column <> referenced_column` is equivalent to specifying
+both `column < referenced_column` and `column > referenced_column` 
+"""
+
+
 class TTableReference(TypedDict):
     """Describes a reference to another table's columns.
     `columns` corresponds to the `referenced_columns` in the referenced table and their order should match.
     """
 
+    label: NotRequired[str]
+    """Label to describe the relation 'liked'."""
+
+    cardinality: NotRequired[TReferenceCardinality]
+    """Cardinality of the relationship between `table.column` (left) and `referenced_table.referenced_column` (right)."""
+
+    table: NotRequired[str]
+    """Name of the table.
+    When `TTableReference` is defined on a `TTableSchema` (i.e., "inline reference"), the `table`
+    value is determined by `TTableSchema["name"]`
+    """
+
     columns: Sequence[str]
+    """Name of the column(s) from `table`"""
+
     referenced_table: str
+    """Name of the referenced table"""
+
     referenced_columns: Sequence[str]
+    """Name of the columns(s) from `referenced_table`"""
 
 
 TTableReferenceParam = Sequence[TTableReference]


dlt/common/schema/typing.py
```diff
@@ -276,14 +276,43 @@ class TScd2StrategyDict(TMergeDispositionDict, total=False):
 ]
 
 
+TReferenceCardinality = Literal["-", "<", ">", "<>"]
+"""Represents cardinality between `column` (left) and `referenced_column` (right)
+`-`: one-to-one
```
rudolfix, line 281: this is cool notation, but is it commonly used? I've never seen it. It will be to impossible to understand this without referring to docs: both for humans an LLMs. To me text is OK. You can also use UML:
1..1 1..* *..1
or
1:n n:1 1:1

note that many to many requires intermediate join table. that will never happen in our schema or in any physical database diagram.

you can also have 0..* (zero to many) indicating that parent table is allowed to have no children so LEFT JOIN is allowed.

cardinality is pretty deep when you dig into it... tldr;> I'll probably go for `one-to-one` style and add `zero-to-many` (and vice versa) if your aim here is to generate correct joins

dlt/common/schema/typing.py
```diff
@@ -276,14 +276,43 @@ class TScd2StrategyDict(TMergeDispositionDict, total=False):
...
+"""
+
+
 class TTableReference(TypedDict):
     """Describes a reference to another table's columns.
     `columns` corresponds to the `referenced_columns` in the referenced table and their order should match.
     """
 
+    label: NotRequired[str]
+    """Label to describe the relation 'liked'."""
```
rudolfix, line 297: I do not understand this comment :)  is it a custom annotation? you can always add as many `x-annotation-` as you want.

dlt/common/schema/typing.py
```diff
@@ -276,14 +276,43 @@ class TScd2StrategyDict(TMergeDispositionDict, total=False):
...
+
+    cardinality: NotRequired[TReferenceCardinality]
+    """Cardinality of the relationship between `table.column` (left) and `referenced_table.referenced_column` (right)."""
+
+    table: NotRequired[str]
```
rudolfix, line 302: OK. but now you need to validate this in your dbml PR

rudolfix (changes requested):
besides review: maybe we should add an example on how to use table references with dlt transformations? in EL pipelines the need for table references is minimal and we are not even validating them. but in T you do more modelling so adding those makes sense. we can validate them in model normalizer...

dlt/common/schema/typing.py
zilto: Example:
- you have table `Users` and `Posts`
- you have a  `Reference` that specifies  `User.id < Posts.user_id`
- you have a `label="liked"` for that `Reference`, giving the semantic meaning "(User, liked, Post)".

This would be displayed on an edge in a schema diagram. These triplets could also enabled graph database support.

dlt/common/schema/typing.py
zilto: The idea would be that `remove_processing_hints` converts all `TTableReference` to the same form. Either

```python
# inline definition (current)
table_name = "users"
table_schema = schema["tables"][table_name"] 
reference = table_schema["references"][0]

column_name = reference["columns"]
other_table_name = reference["referenced_table"]
other_column_name = reference["referenced_columns"]
# reference.get("table") is None
``` 

or

```python
# long form
references = schema["references"]
reference = references[0]

table_name = reference["table"]
column_name = reference["columns"]
other_table_name = reference["referenced_table"]
other_column_name = reference["referenced_columns"]
```


This is inspired by DBML which has 3 definitions: [long form, short form, and inline](https://dbml.dbdiagram.io/docs/#relationships--foreign-key-definitions)

zilto:
Converting to draft until I implement the `remove_processing_hints` [described here](https://github.com/dlt-hub/dlt/pull/2941#discussion_r2248328299)

dlt/common/schema/typing.py
sh-rp: Yeah, it is called "label" or "edge label" in graph databases. I'd just improve the comment so it is clear what is meant by this.

zilto:
closing because of inactivity; will reopen once actively being developed

PR #2951, opened 2025-07-31
2946 sqlalchemy destination fixes

### LINKED ISSUE #2946: engine_version type mismatch in _dlt_version table with sqlalchemy and trino
### dlt version

1.14.1

### Describe the problem

When trying to load into trino (starburst galaxy) using DLT, I get the following error:
```
dlt.destinations.exceptions.DatabaseTransientException: (trino.exceptions.TrinoUserError) TrinoUserError(type=USER_ERROR, name=TYPE_MISMATCH, message="line 1:20: Insert query has mismatched column types: Table: [bigint, bigint, timestamp(6) with time zone, varchar, varchar, varchar], Query: [integer, varchar(2), timestamp(6) with time zone, varchar(11), varchar(44), varchar(2836)]", query_id=20250730_144907_47540_6tedy)
[SQL: INSERT INTO dlt_testing_20250730024816._dlt_version (version, engine_version, inserted_at, schema_name, version_hash, schema) VALUES (?, ?, ?, ?, ?, ?)]
[parameters: (2, '11', DateTime(2025, 7, 30, 14, 49, 7, 909157, tzinfo=Timezone('UTC')), 'dlt_testing', 'xSp3/lxPcJQFGzqDx//3kKjFTwgox304QK4hpLTPEdg=', '{big schema string}')]
```
The mismatch to me looks like `engine_version` being a string in the insert statement. I see that it is [being cast to string](https://github.com/dlt-hub/dlt/blob/8aea94997561505c3266b512b5a32df49234740a/dlt/destinations/impl/sqlalchemy/sqlalchemy_job_client.py#L249) in `sqlalchemy_job_client.py` but is [defined as an int](https://github.com/dlt-hub/dlt/blob/8aea94997561505c3266b512b5a32df49234740a/dlt/common/schema/utils.py#L946) in what looks to me like the common schema for `_dlt_version`. 

### Expected behavior

That the `_dlt_version` loads successfully

### Steps to reproduce

Here's a minimal script:
```
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "dlt[sqlalchemy]",
#     "trino[sqlalchemy]",
# ]
# ///

from typing import Any, Dict, Iterator

import dlt


@dlt.resource(name="sample_data")
def generate_sample_data() -> Iterator[Dict[str, Any]]:
    """
    A simple data source that generates sample records.
    """
    # Generate some sample data
    sample_records = [
        {"id": 1, "name": "Alice", "department": "Engineering"},
        {"id": 2, "name": "Bob", "department": "Marketing"},
        {"id": 3, "name": "Carol", "department": "Sales"},
    ]

    for record in sample_records:
        yield record


def run_pipeline():
    """
    Main function to run the dlt pipeline with sqlalchemy destination.
    """
    pipeline = dlt.pipeline(
        pipeline_name="sample_sqlalchemy_pipeline",
        destination="sqlalchemy",
        dataset_name="sample_dataset",
        progress=dlt.progress.log(
            log_period=10,
        ),
    )

    load_info = pipeline.run(generate_sample_data())


if __name__ == "__main__":
    run_pipeline()
```
my secrets.toml:
```
[destination.sqlalchemy.credentials]
database = # database -> catalog
username = 
password = 
host = 
port = 443
drivername = "trino"
```

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12

### dlt data source

dummy data

### dlt destination

_No response_

### Other deployment details

Original issue encountered in dagster, reproduced with example above.
The galaxy catalog is using iceberg by default

### Additional information

I was able to get the load to succeed by removing the cast on engine version: https://github.com/dlt-hub/dlt/blob/8aea94997561505c3266b512b5a32df49234740a/dlt/destinations/impl/sqlalchemy/sqlalchemy_job_client.py#L249

rudolfix:
### Description
fixes #2946 

with several limitations coming from Trino dialect.

* normalizes `;` usage in all destinations (; are added when executing many from list)
* adds type mappers for mysql, mssql and trino
* json field support in mssql destination
* fixes clickhouse temp table: from Memory to MergeTree (Memory not shared across nodes)
* fixes bug in sql_client (duckdb) - see commit list for more

### DIFF
--- a/tests/.dlt/config.toml
+++ b/tests/.dlt/config.toml
@@ -13,7 +13,3 @@ bucket_url_r2="s3://dlt-ci-test-bucket"
 bucket_url_gdrive="gdrive://15eC3e5MNew2XAIefWNlG8VlEa0ISnnaG"
 bucket_url_sftp="sftp://localhost/data"
 memory="memory:///m"
-
-[destination.databricks]
-# Add PRIMARY KEY and FOREIGN KEY constraints to tables
-create_indexes=true
--- a/tests/load/databricks/test_databricks_adapter.py
+++ b/tests/load/databricks/test_databricks_adapter.py
@@ -1,9 +1,10 @@
-from typing import Iterator, Dict, Any, cast
+from typing import Iterator, Dict
 import pytest
 
 import dlt
 from dlt.common.utils import uniq_id
 from dlt.destinations.adapters import databricks_adapter
+from dlt.destinations import databricks
 from dlt.destinations.impl.databricks.databricks_adapter import (
     CLUSTER_HINT,
 )
@@ -24,7 +25,9 @@ pytestmark = pytest.mark.essential
 def test_databricks_hints(
     destination_config: DestinationTestConfiguration,
 ) -> None:
-    pipeline = destination_config.setup_pipeline(f"databricks_{uniq_id()}", dev_mode=True)
+    pipeline = destination_config.setup_pipeline(
+        f"databricks_{uniq_id()}", dev_mode=True, destination=databricks(create_indexes=True)
+    )
 
     @dlt.resource(
         columns={"some_int": {"data_type": "bigint", "nullable": False}},
@@ -196,7 +199,9 @@ def test_databricks_adapter_special_characters(
     destination_config: DestinationTestConfiguration,
 ) -> None:
     """Test that special characters in comments and tags are properly escaped"""
-    pipeline = destination_config.setup_pipeline(f"databricks_{uniq_id()}", dev_mode=True)
+    pipeline = destination_config.setup_pipeline(
+        f"databricks_{uniq_id()}", dev_mode=True, destination=databricks(create_indexes=True)
+    )
 
     @dlt.resource(
         columns={"some_int": {"data_type": "bigint", "nullable": False}},
--- a/tests/plugins/test_plugin_discovery.py
+++ b/tests/plugins/test_plugin_discovery.py
@@ -81,11 +81,17 @@ def test_run_context_passthrough() -> None:
     context = run_context.active()
     assert context.name == "dlt-test"
 
-    container = Container()
-    container[PluggableRunContext].reload(context.run_dir, dict(passthrough=True))
+    try:
+        container = Container()
+        container[PluggableRunContext].reload(context.run_dir, dict(passthrough=True))
 
-    context = run_context.active()
-    assert context.name == "dlt"
+        context = run_context.active()
+        assert context.name == "dlt"
+
+    finally:
+        container[PluggableRunContext].reload(context.run_dir, dict(passthrough=False))
+        context = run_context.active()
+        assert context.name == "dlt-test"
 
 
 def test_import_references() -> None:


dlt/common/schema/utils.py
```diff
@@ -963,7 +963,12 @@ def loads_table() -> TTableSchema:
     table = new_table(
         LOADS_TABLE_NAME,
         columns=[
-            {"name": C_DLT_LOADS_TABLE_LOAD_ID, "data_type": "text", "nullable": False},
+            {
+                "name": C_DLT_LOADS_TABLE_LOAD_ID,
+                "data_type": "text",
+                "nullable": False,
+                "precision": 64,
```
sh-rp, line 970: this is not going to cause problems for existing datasets right? I just checked `get_new_table_columns` on the Schema and this should be safe.

dlt/destinations/impl/sqlalchemy/alter_table.py
```diff
@@ -26,20 +29,48 @@ def flush(self) -> None:
...
+            )
+            self.ctx = None
+            self.ops = None
 
     def add_column(self, table_name: str, column: sa.Column, schema: str) -> None:
-        self.ops.add_column(table_name, column, schema=schema)
+        if self.ops:
+            # Use alembic if available
+            self.ops.add_column(table_name, column, schema=schema)
+        else:
```
sh-rp, line 59: Do we have a test for this somewhere?

docs/website/docs/dlt-ecosystem/destinations/sqlalchemy.md
```diff
@@ -130,14 +130,81 @@ with engine.connect() as conn:
 ```
 
 ## Notes on other dialects
-We tested this destination on **mysql** and **sqlite** dialects. Below are a few notes that may help enabling other dialects:
+We tested this destination on **mysql**, **sqlite** and **mssql** dialects. Below are a few notes that may help enabling other dialects:
```
sh-rp, line 133: Later on in this PR there are some trino specific things like the specialized type mapper. Did you test that against a trino database and should we mention this here?

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -184,10 +184,10 @@ def _new_temp_table_name(cls, table_name: str, op: str, sql_client: SqlClientBas
             )
 
     @classmethod
-    def _to_temp_table(cls, select_sql: str, temp_table_name: str) -> str:
+    def _to_temp_table(cls, select_sql: str, temp_table_name: str, unique_column: str) -> str:
         return (
             f"DROP TABLE IF EXISTS {temp_table_name} SYNC; CREATE TABLE {temp_table_name} ENGINE ="
-            f" Memory AS {select_sql}"
+            f" MergeTree PRIMARY KEY {unique_column} AS {select_sql}"
```
sh-rp, line 190: Just out of curiousity, is this a performance increase or why did we need this?

sh-rp: Ah, it says in the pr comment, ok :)

dlt/common/runners/synth_pickle.py
```diff
@@ -24,18 +25,17 @@ class SynthesizingUnpickler(pickle.Unpickler):
     """Unpickler that synthesizes missing types instead of raising"""
 
     def find_class(self, module: str, name: str) -> Any:
+        full_name = f"{module}.{name}"
+        module_obj = sys.modules[__name__]
+        with contextlib.suppress(AttributeError):
+            return getattr(module_obj, full_name)
```
sh-rp, line 31: I'm not really sure what is going on here, but shouldn't this be `getattr(module_obj, name)`? Or can you get the class with the full name from a module object? Maybe there should be a test.

dlt/destinations/impl/sqlalchemy/alter_table.py
rudolfix: we have a test for the alembic branch. tests for fallback are on Trino. there are so many weird problems with it that I disabled them. but they were passing. this is best effort for now.

dlt/common/schema/utils.py
rudolfix: existing schemas will not be updated automatically. you need to create schema migration for that

docs/website/docs/dlt-ecosystem/destinations/sqlalchemy.md
rudolfix: I answered above. What we need is a real Trino destination. sqlalchemy is just a nice trick where data is loaded with insert statements (and we do not support upsert)

dlt/common/runners/synth_pickle.py
rudolfix: this is pretty well tested. this unpickler works like this:
- tries to unpickle with a real class
- if that does not work: synthesizes object from pickle dump on the fly and caches it under full name in the current module

PR #2959, opened 2025-08-04
feat: `dlt.Schema.to_dot()` graphviz export

zilto:
This follows #2929 and adds DOT export.

### Design
- the DOT is produced by handling strings instead of using `graphviz`. This allows to make it 0 dependency. This is particularly useful because the `graphviz` Python library depends on graphviz system installs
- add `_repr_html_` to `dlt.Schema`, this method is automatically picked up by notebook renderers. This method wraps the DOT string into an HTML that loads JS bundles from the web to render the viz. This won't work offline
- add `render_schema_with_graphviz()`. This requires `graphviz` dependency and allows to render the `dlt.Schema` as svg, png, jpeg, pdf, etc.


rudolfix (approved):
LGTM! I'm not sure how many schema converters we'll need in the future but maybe we need to make them pluggable ie. via our existing `pluggy` and standard interface with schema import/export (it is almost standard btw.) 
also some settings could be configurable ie. html template and colors. this is easy to add and we can do it in the future

rudolfix (approved):
LGTM!

PR #2960, opened 2025-08-05
feat: Smart `dlt.Relation.join()`

zilto:
This adds convenience methods `dlt.Relation(...).join()` and `dlt.Relation(...).join_child(...)` to simplify joins between tables that have `TTableReference` on `dlt.Schema`.

#### Design decisions
- Joins for parent-child or root-child relationships are likely always `LEFT` joins.
- Joins for arbitrary references defaults to `INNER`, but could be any type
- For parent-child / root-child relationships, trim the name of the table based on their common prefix
- Namespace all columns with origin table name
- Add "metadata columns" at the beginning of the table
- Special renaming for metadata columns
- `.join()` can receive:
   - `dlt.Relation` to join on (if column is unambiguous) 
   - `"table_name"` (if column is unambiguous)
   - `"table_name.column_name"` in case of multiple references between tables

#### Open questions
- What is the public API? Is it useful to have `.join()` and `.join_parent_child()` separately?

### DIFF
--- a/dlt/destinations/dataset/relation.py
+++ b/dlt/destinations/dataset/relation.py
@@ -18,6 +18,7 @@ from contextlib import contextmanager
 from textwrap import indent
 from dlt.common.utils import simple_repr, without_none
 
+import sqlglot
 from sqlglot import maybe_parse
 from sqlglot.optimizer.merge_subqueries import merge_subqueries
 from sqlglot.expressions import ExpOrStr as SqlglotExprOrStr
@@ -27,6 +28,7 @@ import sqlglot.expressions as sge
 from dlt.common.destination.dataset import Relation, TFilterOperation
 
 from dlt.common.libs.sqlglot import to_sqlglot_type, build_typed_literal, TSqlGlotDialect
+from dlt.common.schema.utils import get_first_column_name_with_prop
 from dlt.common.schema.typing import TTableSchemaColumns, TTableSchema
 from dlt.common.typing import Self, TSortOrder
 from dlt.common.exceptions import ValueErrorWithKnownValues
@@ -86,7 +88,7 @@ class ReadableDBAPIRelation(Relation, WithSqlClient):
         self,
         *,
         readable_dataset: "ReadableDBAPIDataset",
-        query: Optional[Union[str, sge.Query, IbisExpr]] = None,
+        query: Optional[Union[str, sge.Select, IbisExpr]] = None,
         query_dialect: Optional[str] = None,
         table_name: Optional[str] = None,
         _execute_raw_query: bool = False,
@@ -100,8 +102,8 @@ class ReadableDBAPIRelation(Relation, WithSqlClient):
         # derived / cached properties
         self._opened_sql_client: SqlClientBase[Any] = None
         self._columns_schema: TTableSchemaColumns = None
-        self.__qualified_query: sge.Query = None
-        self.__normalized_query: sge.Query = None
+        self.__qualified_query: sge.Select = None
+        self.__normalized_query: sge.Select = None
 
         # parse incoming query object
         self._sqlglot_expression: sge.Query = None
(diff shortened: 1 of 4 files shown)


dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+            if "." in other:
+                other, *_ = other.split(".")
+            # TODO simplify this by removing protocol typing
+            other_rel = cast(ReadableDBAPIRelation, self._dataset.table(other))
+        elif isinstance(other, ReadableDBAPIRelation):
+            other_rel = other
+        else:
+            raise TypeError
+
+        current_table_name = rel._sqlglot_expression.find(sge.From).name
```
rudolfix, line 460: will it work with multiple join? ie. I have nesting level 2 and I use `join` twice. hmmm I think so. or multi joins are out of scope?

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+        current_table_schema = self._dataset.schema.tables[current_table_name]
```
rudolfix, line 461: you should take column names from the (qualified) relation's query. so you can identify whose columns are coming from `current_table_name`, you should also recover aliases here to create column alias in final SELECT

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+        references = current_table_schema.get("references")
+        if references is None:
+            raise KeyError(f"Not references found on schema for table `{current_table_name}`")
+
+        other_table_name = other_rel._sqlglot_expression.find(sge.From).name
```
rudolfix, line 466: same thing

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+            # other_col_name=other_col_name,
+            references=references,
+        )
+        join_condition = _create_join_condition_from_reference(current_table_name, reference)
+
+        meta_cols_aliases = []
+        current_cols_aliases = []
+        other_cols_aliases = []
+
+        for col in current_table_schema["columns"]:
```
rudolfix, line 480: here you are comparing normalized identifiers with non normalized constants (`C_DLT_LOAD_ID`). here's how we identify `dlt` columns and tables in filesystem:
```py
def is_dlt_identifier(self, ident: str) -> bool:
        return ident.startswith(self.schema._dlt_tables_prefix)
```
maybe it makes sense to add it to `Schema` object

dlt/destinations/queries.py
```diff
@@ -125,3 +131,66 @@ def build_select_expr(
     select_expr = sge.Select(expressions=columns_expr).from_(table_expr)
 
     return select_expr
+
+
+def _create_column_alias(
```
rudolfix, line 136: yes this is good! but please use naming convention to create "compound" identifiers ie. this will split parts into single identifiers and then create compound one (takes into account that parts may be already compound)
```
def normalize_alias(c: DbtGeneratorConfig, *parts: str) -> str:
    return c.naming.normalize_path(c.naming.make_path(*parts))
```

dlt/destinations/queries.py
```diff
@@ -125,3 +131,66 @@ def build_select_expr(
...
+        return " AND ".join(join_conditions)
+
+
+def _get_valid_reference(
+    *,
+    other_table_name: str,
+    other_column_name: Optional[str] = None,
+    references: TTableReferenceParam,
+) -> TTableReference:
+    for reference in references:
```
rudolfix, line 175: we have pretty good reference validation in `dbt_generator` I linked it in other comment. also I think you have something similar in `dbml` generator?

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+            if col in [C_DLT_LOAD_ID]:
+                continue
+
+            alias = _create_column_alias(current_table_name, col)
```
rudolfix, line 484: yeah this is good but you should use current table alias not name to generate column alias. this will let you to create joins of any depth

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+            if col in [C_DLT_ID]:
+                meta_cols_aliases.append(alias)
+            else:
+                current_cols_aliases.append(alias)
+
+        for col in other_table_schema["columns"]:
+            if col in [C_DLT_LOAD_ID]:
+                continue
+
+            alias = _create_column_alias(other_table_name, col)
```
rudolfix, line 494: same

dlt/destinations/dataset/relation.py
rudolfix: this will let you create joins on relations that do not select all columns

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+
+        parent_cols_aliases = []
+        for col in parent_table_schema["columns"]:
+            if col in ["_dlt_id", "_dlt_load_id"]:
+                continue
+
+            alias = _create_column_alias(parent_table_name, col)
+            parent_cols_aliases.append(alias)
+
+        _, _, child_table_part = child_table_name.removeprefix(parent_table_name).partition("__")
```
rudolfix, line 568: you should use `break_path` and `make_path` in `schema.naming`.

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+        join_condition = _create_join_condition(
+            table=parent_table_name,
+            column=parent_row_key,
+            other_table=child_table_name,
+            other_column=child_parent_key,
+        )
+
+        parent_cols_aliases = []
+        for col in parent_table_schema["columns"]:
+            if col in ["_dlt_id", "_dlt_load_id"]:
```
rudolfix, line 562: please compare only normalize identifiers. it may make sense to cache `_dlt_id` and `_dlt_load_id` in schema along a few others that are already there:
```py
self._dlt_tables_prefix = to_naming.normalize_table_identifier(DLT_NAME_PREFIX)
self.version_table_name = to_naming.normalize_table_identifier(VERSION_TABLE_NAME)
self.loads_table_name = to_naming.normalize_table_identifier(LOADS_TABLE_NAME)
self.state_table_name = to_naming.normalize_table_identifier(PIPELINE_STATE_TABLE_NAME)
```

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+                other_cols_aliases.append(alias)
+
+        rel._sqlglot_expression = (
+            sge.select(
+                *meta_cols_aliases,
+                *current_cols_aliases,
+                *other_cols_aliases,
+            )
+            .from_(current_table_name)
+            .join(other_table_name, on=join_condition, join_type=how)
```
rudolfix, line 507: what happens if `other` relation already has a join? ie. `users.join(orders.join(items)`? is sqlglot able to generate proper aliasing itself? if not you'll need to assign aliases to the tables ie `orders__items`. same table may be joined several times. also you remove prefixes from nested tables. that may create conflicts.

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
...
+        child_cols_aliases = []
+        for col in child_table_schema["columns"]:
+            if col in [child_parent_key, "_dlt_list_idx", "_dlt_id"]:
+                continue
+
+            alias = _create_column_alias(child_table_name, col, prefix=child_table_part)
+            child_cols_aliases.append(alias)
+
+        meta_cols_aliases = (
+            _create_column_alias(parent_table_name, "_dlt_load_id", prefix="", separator=""),
```
rudolfix, line 578: this is done so query looks nicer?

dlt/destinations/dataset/relation.py
```diff
@@ -431,6 +439,160 @@ def where(
         rel._sqlglot_expression = rel._sqlglot_expression.where(condition)
         return rel
 
+    def join(
```
rudolfix, line 442: We already have join generator which works pretty well for generating `dbt` packages. It supports both regular and nested references and also joins of unlimited depth (via table and column aliasing). Maybe some code or ideas could be helpful here?

here we render fact table: https://github.com/dlt-hub/dlt-plus/blob/devel/packages/dlt_plus/dlt_plus/dbt_generator/render.py#L72

this procedure renders fact table (indicated by config): it will create all possible join (regular and nested), also recursively up to a certain depth. your case is simpler (just one join) but the resultant relation should allow to add more joins (both on the same level and joins of join). or this is out of scope?

1. first we collect all references, validate them, and convert nested (automated) references into regular references (`get_rendered_tables`)
2. then required tables are added, here the crucial part is how aliasing is done. all tables except top (fact) table will be aliased uniquely. see https://github.com/dlt-hub/dlt-plus/blob/devel/packages/dlt_plus/dlt_plus/dbt_generator/render.py#L85 and https://github.com/dlt-hub/dlt-plus/blob/devel/packages/dlt_plus/dlt_plus/dbt_generator/render.py#L113
3. all columns that are added to the joined relation are uniquely aliased as well: https://github.com/dlt-hub/dlt-plus/blob/devel/packages/dlt_plus/dlt_plus/dbt_generator/render.py#L143 table aliases, not names are used to make sure that columns in SELECT list are unique

dlt/destinations/dataset/relation.py
zilto: This work was based on the incremental loading PRs were you suggested to not do recursive joins and only support immediate `parent-child` and `root-child` joins.

For this PR I believe we could support
```python
"""
tables with `root_key=True`:
- orders
- orders__items
- orders__items__details
- customers  # with ref `orders.customer_id == customers.id`
"""

# parent-child
dataset.table("orders").join("orders__items")
# LEFT JOIN orders._dlt_id = orders__items._dlt_parent_id
dataset.table("orders__items").join("orders__items__details")
# LEFT JOIN orders__items._dlt_id = orders__items__details._dlt_parent_id

# root-descendant
dataset.table("orders").join("orders__items__details")
# LEFT JOIN orders._dlt_id = orders__items__details._dlt_root_id

# relationship
dataset.table("orders").join("customers")
# INNER JOIN orders.customer_id = customers.id
dataset.table("customers").join("orders")
# INNER JOIN customer.id = orders.customer_id
```

Currently excluded:
```python
# recursive parent-child
# recursive API
dataset.table("orders").join("orders__items").join("orders__items__details")

# using a list API is easier 
dataset.table("orders").join(["orders__items", "orders__items__details"])
```
This has new implications
- We're keeping columns from 3 tables instead of 2
- It's different from `root-descendant` because we are including intermediary parents which have no descendant rows.
- Providing convenience table/column aliases is much harder for recursive API because we need to introspect the previous query that renamed columns

zilto:
Closing in favor of the two linked issues #3402 #3403

PR #2966, opened 2025-08-05
fix: avoid private interfaces; explicit compiler mapping

zilto:
### Description
The `dlt.helpers.ibis` module had two instances of accessing private modules of 3rd libraries:
- `sqlglot.expressions` without importing it
- dynamically retrieved `ibis.backends.sql.compilers.NAME` (`duckdb`, `postgres`, etc.)

Changes:
- use public interfaces from sqlglot and ibis
- remove dynamic mapping `dialect -> compiler`; changes to either library could break our code.
- add tests for the mapping `dialect -> compiler` to make explicit when the default `DuckDBCompiler` is expected to be retrieved
- the explicit mapping adds support for `presto -> TrinoCompiler`, `spark -> PySparkCompiler` and `spark2 -> PySparkCompiler`, which was previously unsupported

### Related
- adds features previously implemented in #2498

### DIFF
--- a/dlt/helpers/ibis.py
+++ b/dlt/helpers/ibis.py
@@ -1,6 +1,6 @@
 from typing import cast, Any
 
-from dlt.common.exceptions import MissingDependencyException
+from dlt.common.exceptions import MissingDependencyException, ValueErrorWithKnownValues
 from dlt.common.destination import TDestinationReferenceArg, Destination
 from dlt.common.destination.client import JobClientBase
 from dlt.common.schema import Schema
@@ -22,6 +22,7 @@ from dlt.destinations.impl.synapse.configuration import SynapseClientConfigurati
 try:
     import ibis
     import sqlglot
+    import sqlglot.expressions as sge
     from ibis import BaseBackend, Expr, Table
     import ibis.backends.sql.compilers as sc
     from ibis.backends.sql.compilers.base import SQLGlotCompiler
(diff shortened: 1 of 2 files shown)


dlt/helpers/ibis.py
```diff
@@ -214,28 +215,54 @@ def create_unbound_ibis_table(schema: Schema, dataset_name: str, table_name: str
...
+    elif dialect == "mysql":
+        compiler = sc.MySQLCompiler()
+    elif dialect == "oracle":
+        compiler = sc.OracleCompiler()
+    elif dialect == "postgres":
+        compiler = sc.PostgresCompiler()
+    elif dialect == "presto":
+        compiler = sc.TrinoCompiler()
+    elif dialect == "prql":
+        compiler = sc.DuckDBCompiler()
```
anuunchin: Is this one just, like, a fallback? 👀  wondering why we shouldn't leave it to the last else condition

dlt/helpers/ibis.py
```diff
@@ -214,28 +215,54 @@ def create_unbound_ibis_table(schema: Schema, dataset_name: str, table_name: str
...
+    elif dialect == "redshift":
+        compiler = sc.PostgresCompiler()
+    elif dialect == "risingwave":
+        compiler = sc.RisingWaveCompiler()
+    elif dialect == "snowflake":
+        compiler = sc.SnowflakeCompiler()
+    # NOTE I'm unsure if both `spark` and `spark2` are supported by the same compiler
```
anuunchin, line 246: A question for understanding - are we assuming here that spark should be backwards compatible with spark2 and, thus, the same compiler?

anuunchin (changes requested):
🧠

dlt/helpers/ibis.py
zilto: good catch!

dlt/helpers/ibis.py
zilto: Ok, so the logic can be confusing here. So I'll detail it for legacy purposes.

## Ibis internals
Ibis works as follow:
1. (ibis) write ibis expressions
2. (ibis) get the compiler associated with the backend (BigQuery, DuckDB, etc.) 
3. (ibis) use the compiler to convert the ibis expression to an sqlglot expression
4. (sqlglot) transpile the sqlglot expression to an SQL string of the target dialect (e.g., use the right quotations)

Ibis exposes "Ibis expression to SQL string", but doesn't expose "Ibis expression to SQLGlot expression". This is why dlt code does some digging.

## `dlt.Relation`
In the most basic case, a `dlt.Relation` does:
1. (dlt) give SQL query (string or sqlglot) to `dlt.Relation`
2. (dlt) identify the dialect of the destination
3. (sqlglot) maybe parse the SQL query
4. (sqlglot) transpile to destination dialect

### `dlt.Relation` with Ibis
Things are a bit more complicated when using ibis with `dlt.Relation`
1. (dlt) give an ibis expression to `dlt.Relation`
2. (dlt) identify the dialect of the destination
3. (dlt) retrieve the compiler from ibis to sqlglot (this PR)
4. (ibis) compile the ibis expression to an sqlglot expression
5. (sqlglot) transpile to destination dialect

## Our logic
For  `dlt.Relation` with ibis, we are determining the ibis to sqlglot compiler using `dlt destination -> sqlglot dialect -> ibis compiler`. This is confusing. How does `spark` and `spark2` have the same `PySpark` compiler, don't they have different syntax?

The thing is, the sqlglot dialect used to retrieve the compiler (e.g., `spark` or `spark2`) has no impact on the compile operation from Ibis to SQLGlot. What we'll see is that `PySpark` compiler will always translate to "a single specific spark SQL" THEN other dlt code will transpile "this single specific spark SQL" to either `spark` or `spark2`

A more appropriate logic would be:
1. (dlt) give an ibis expression to `dlt.Relation`
2. (dlt) identify the dialect of the destination (for sqlglot transpile)
3. (dlt) identify the compiler for the destination (for ibis to sqlglot compile)
4. (ibis) compile the ibis expression to an sqlglot expression
5. (sqlglot) transpile to destination dialect

This way we decouple `dlt destination -> sqlglot dialect` and `dlt destination -> ibis compiler`. This is how it was originally implemented in #2498, but it was discarded

anuunchin (approved):

PR #2967, opened 2025-08-05
repo: use `ruff check` for linting

zilto:
This uses `ruff` for linting (keeps `black` for formatting). I configured using rules defined in `tox.ini`. Then, I added exclusions such that `make lint` would succeed.

## Next steps
- use `ruff format` instead of `black`. Formatting **will not** break any code, but it will produce a large diff because of [know deviations between ruff and black](https://docs.astral.sh/ruff/formatter/black/). I would do that in a follow-up PR and add the commit to `.git-blame-ignore-revs`.
- move all configs to `pyproject.toml` (e.g.,`pytest.init`, `mypy.ini`) and delete unused files
- discuss additional linting rules to apply (sorting imports, remove unused imports, etc.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -355,6 +355,7 @@ include = [
 ]
 exclude = [
     "tests/reflection/module_cases/syntax_error.py",
+    "docs/website/docs/general-usage/transformations/transformation-snippets.py",
 ]
 
 [tool.ruff.lint]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -343,6 +343,44 @@ include = [
 requires = ["hatchling"]
 build-backend = "hatchling.build"
 
+
+[tool.ruff]
+line-length = 100
+target-version = "py39"
+src = ["dlt"]
+include = [
+    "dlt/**/*.py",
+    "tests/**/*.py",
+    "docs/**/*.py",
+]
+exclude = [
+    "tests/reflection/module_cases/syntax_error.py",
+]
+
+[tool.ruff.lint]
+preview = true
+ignore = [
+    # from tox.ini
+    "E1",
+    "E2",
+    "E3",
+    "E4",
+    "F401",
+    "W391",
+    "W292",
+    "E501",
+    # additional ignores to appease ruff
+    "E713",
+    "E721",
+    "E731",
+    "E741",
+    "E999",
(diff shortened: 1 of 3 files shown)


sh-rp (changes requested):
This is cool! What we additionally should do (and possibly it is already considered in some cases):
* Make sure functionality of flake8 provided by the installed plugins is covered in ruff (bugbear, tidy-imports, encodings, builtins, print)
* Ensure the banned imports are still banned, check `banned-modules` in tox.ini
* See if we can also lint docstrings the same way we did with ruff, if not, keep flake8 for that
* Remove unneeded dependencies (flake8 and plugins)
* Remove flake8 directives from the code (you can find them by searching flake8 in the codebase)
* Clean up or remove tox.ini file

pyproject.toml
```diff
@@ -341,6 +341,44 @@ include = [
...
+[tool.ruff]
+line-length = 100
+target-version = "py39"
+src = ["dlt"]
+include = [
+    "dlt/**/*.py",
+    "tests/**/*.py",
+    "docs/**/*.py",
+]
+exclude = [
```
sh-rp, line 356: we should exclude `docs/website/.dlt-repo` here too

zilto: I added the exclusion. Though, this directory didn't exist in my local installation. Is it dynamically created?

zilto:
I updated the `Makefile` and ruff exclusion to match with the recent transformations move

zilto:
Now, we have `ruff` and `flake8` installed. No changes were made to the existing linting / formatting setup. I just added `ruff check` (linting) and started reproducing our flake8 config inside Ruff.

## Next steps
1. Finish configuring `ruff` inside `pyproject.toml` to match the flake8 config.
2. Remove `flake8` checks from the `Makefile` and delete associated configs
3. Decide together what linting rules to enforce and enable them all (e.g., sort imports)
4. Exclude all files (or even modules) that have tons of linting issues; then we can progressively fix them and remove them from the exclude list
5. Investigate the docs testing and linting utils and ensure nothing is broken there. It currently behaves independently of the Python linting and should be unified.
6. (optional) replace `black` with `ruff` this will create a large diff given they have slightly different behaviors

PR #2968, opened 2025-08-05
helper to filter dlt-changes out of schema updates

djudjuu:


dlt/common/schema/utils.py
```diff
@@ -890,6 +891,52 @@ def get_data_and_dlt_tables(tables: TSchemaTables) -> tuple[list[TTableSchema],
     return data_tables, dlt_tables
 
 
+def changes_without_dlt_changes(
```
rudolfix: this name was correct in context of callback function. here it just removed "dlt entities" from a list of tables. so (1) it needs a different name (2) it makes sense to split it into many functions. see my general review

dlt/common/schema/utils.py
```diff
@@ -913,6 +914,90 @@ def get_data_and_dlt_tables(tables: TSchemaTables) -> tuple[list[TTableSchema],
     return data_tables, dlt_tables
 
 
+def is_dlt_table_or_column(name: str, dlt_prefix: str = DLT_NAME_PREFIX) -> bool:
```
rudolfix: we still need a small refinement. the point is to never take non normalized identifier for comparison. 
1. that should be clear in arg names ie (normalized_name). 
2. do not take default parameter, require normalized prefix
3. update docstrings


it makes sense to extract this small util:
```
self._dlt_tables_prefix = to_naming.normalize_table_identifier(DLT_NAME_PREFIX)
```
as a separate method that returns prefix based on naming convention

please do not take prefix in the method below as well

tests/common/schema/test_schema_utils.py
```diff
@@ -0,0 +1,142 @@
...
+    assert ethereum_schema.loads_table_name not in filtered_names
+    assert "blocks" in filtered_names
+
+    blocks = _get_table(filtered, "blocks")
+    block_column_names = set(blocks["columns"])
+    assert "_dlt_load_id" not in block_column_names
+    assert "_dlt_id" not in block_column_names
+
+
+def test_exclude_dlt_entities_honors_custom_prefix(ethereum_schema: Schema) -> None:
```
rudolfix: you do not need this test this. all tests could be parametrized like this:
```
@pytest.mark.parametrize(
    "name_normalizer_ref",
    (
        "tests.common.cases.normalizers.title_case",
        "tests.common.cases.normalizers.sql_upper",
        "tests.common.cases.normalizers.snake_no_x",
    ),
)
def test_references_parameterized_by_naming(schema: dlt.Schema, name_normalizer_ref: str) -> None:
    schema._normalizers_config["names"] = name_normalizer_ref
    schema.update_normalizers()
```
and then you can use naming convention for tests. you'll also see why default prefix should not be passed

PR #2972, opened 2025-08-06
Add table_loader_class parameter and create BaseTableLoader class

### LINKED ISSUE #2736: support custom backends, row loaders etc. via custom TableLoader in `sql_database`
**Background**
We provide several ways to customize `sql_database` via callback that let you ie. to rewrite the queries. Here we are allowing for a deep customization by subclassing `TableLoader` that is responsible for execution of the query over already reflected SqlAlchemy table. Overriding this class will allow:
- implementing additional backends (ie. `adbc` fast arrow readers or ibis backend)
- implementing custom row readers ie. with pagination (#2432)
- implementing own retry mechanisms
- very deep customization of generated queries

**Implementation**
1. * [ ] design an internal interface that users are supposed to override in subclasses
2. * [ ] test the interface so we do not break users' code by chaning it
3. * [ ] allow to: easy add new backends and override row loading
4. * [ ] provide two examples in tests and docs: paginated loading + adbc backend for postgres / snowflake (we may merge it into OSS if works well)

lucargir:
### Description
This PR creates an abstract class for the current TableLoader called BaseTableLoader with the existing methods `make_query` and `load_rows`. The initialisation of the original TableLoader is moved to the base class. A new parameter called `table_loader_class` is added to `sql_table` and `sql_database`.

### Related Issues
- Closes #2736


rudolfix:
@lucargir got implemented here: https://github.com/dlt-hub/dlt/pull/3595/ thanks for your help!!

PR #2978, opened 2025-08-07
avoid setting "None" string for aws session token

### LINKED ISSUE #2407: [Athena Destination]: Invalid Security Token with Profile Name in 1.3.0+
### dlt version

1.3.0+

### Describe the problem

I am experiencing difficulties with the Athena destination in versions 1.3.0 and above of the library. The code I am using worked perfectly in version 1.2.0, but after upgrading, I encounter an error related to invalid security tokens when using AWS credentials specified by profile name. If I provide aws_access_key_id and aws_secret_access_key it works but problem is when I specify profile_name or use AWS_PROFILE env variable.

```
An error occurred (UnrecognizedClientException) when calling the StartQueryExecution operation: The security token included in the request is inval
```

### Expected behavior

_No response_

### Steps to reproduce

- Set up a .dlt/secrets.toml file with the following content:

```
[sources.credentials]
drivername = "postgresql+psycopg2"
database = "main"
host = "host"
port = 5432
username = "username"
password = "password"

[destination.filesystem]
bucket_url = "s3://bucket-destination"

profile_name = "default"

[destination.athena]
query_result_bucket = "s3://bucket-results"
athena_work_group = "workgroup"

[destination.athena.credentials]
profile_name = "default"
```

- Use the following code to run the pipeline:

```
from dlt.sources.sql_database import sql_table

if __name__ == "__main__":
    import dlt

    @dlt.source(name="public", max_table_nesting=0)
    def main_public_test():
        tbl = sql_table(
            table="currencies",
            backend="pyarrow",
            defer_table_reflect=True,
        )
        tbl.apply_hints(table_format="iceberg")
        return [tbl]

    try:
        info = dlt.pipeline(
            pipeline_name="main",
            dataset_name="x_test",
            destination="athena",
            progress="log",
            dev_mode=True,
        ).run(main_public_test())
    except Exception as e:
        print(e)
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

Postgres

### dlt destination

AWS Athena / Glue Catalog

### Other deployment details

_No response_

### Additional information

- This issue does not occur in version 1.2.0; it works fine when I downgrade.
- I have verified that the credentials are correct and that the same AWS SDK versions (botocore and boto3) are being used.
- I do not see any relevant changes in the release notes for version 1.3.0 that could explain this behavior.

tpulmano:
### Description
When using AWS Profiles and retrieving credentials from boto I noticed that it can have AWS access keys but not a session token (`None`). 

If we assign the session token to the string `"None"` boto/Pyathena thinks a session token is being used for credentials and we end up with the error: 
```
dlt.destinations.exceptions.DatabaseTransientException: 
   An error occurred (UnrecognizedClientException) when calling the StartQueryExecution operation: The security token included in the request is invalid
```

### Related Issues
Fixes #2407 

### Additional Context
I didn't run into this issue when providing access keys to dlt only when using profiles.

### DIFF
--- a/dlt/common/configuration/specs/aws_credentials.py
+++ b/dlt/common/configuration/specs/aws_credentials.py
@@ -183,7 +183,8 @@ class AwsCredentials(AwsCredentialsWithoutDefaults, CredentialsWithDefault):
             return None
         self.aws_access_key_id = default.access_key
         self.aws_secret_access_key = TSecretStrValue(default.secret_key)
-        self.aws_session_token = TSecretStrValue(default.token)
+        if default.token is not None:
+            self.aws_session_token = TSecretStrValue(default.token)
         return default
 
     def to_native_credentials(self) -> Optional[Any]:


dlt/common/configuration/specs/aws_credentials.py
```diff
@@ -183,7 +183,8 @@ def _from_session(self, session: Any) -> Any:
             return None
         self.aws_access_key_id = default.access_key
         self.aws_secret_access_key = TSecretStrValue(default.secret_key)
-        self.aws_session_token = TSecretStrValue(default.token)
+        if default.token is not None:
+            self.aws_session_token = TSecretStrValue(default.token)
```
sh-rp: Let's change this to:

`self.aws_session_token = cast(TSecretStrValue, default.token)`

You can do the same for the line above!

sh-rp (changes requested):
Thanks for finding this! I'd solve it in a slightly different way, then we can merge.

tpulmano:
Ah, thanks for the tip. I updated the PR.

Issue #2934, opened 2025-07-27
"log" is not exported from module "dlt.pipeline.progress"


jonathanunderwood:
### dlt version

1.12.3

### Describe the problem

Creating a simple file:
```python
from dlt.pipeline.progress import log
```
and running pyright against this:
```bash
❯ poetry run pyright test.py
/redacted/test.py
  /redacted/test.py:1:35 - error: "log" is not exported from module "dlt.pipeline.progress"
    Import from "dlt.common.runtime.collector" instead (reportPrivateImportUsage)
1 error, 0 warnings, 0 informations
```

### Expected behavior

No pyright errror should result. 

### Steps to reproduce

See above.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

Not relevant.

jonathanunderwood:
The same error arises for 

```python
from dlt.progress import log
```

rudolfix:
@djudjuu pls take a look at this

zilto:
@rudolfix This is caused by name shadowing. The type checker will complain despite imports working.

```python
# dlt/__init__.py
# `dlt.pipeline` refers to `dlt/pipeline/__init__.py`, which contains `pipeline`
from dlt.pipeline import pipeline as _pipeline

# the name `dlt.pipeline` is set to `dlt.pipeline.pipeline`
pipeline = _pipeline

# export `dlt.pipeline` (which is `dlt.pipeline.pipeline`),
# making `dlt.pipeline` (`dlt/pipeline/__init__.py` inacessible)
__all__ = (
  ...,
  "pipeline"
)
```
The name `progress` is from `dlt.pipeline` (the module at `dlt/pipeline/__init__.py`), which becomes inacessible. This causes oddities like
```python
from dlt import progress  # works
# start a new clean Python session
import dlt.progress  # fails; this exists at dlt.pipeline.progress
``` 

### Note
To debug this, you must restart Python each time to avoid import side-effects. Once you import anything from `dlt`, Python will look at `sys.modules["dlt"]` first.

djudjuu:
I've looked into this, and for the moment I am not seeing a quick way out of it. I'll investigate more but I don't have the capacity to prioritize this high atm

jonathanunderwood:
> [@rudolfix](https://github.com/rudolfix) This is caused by name shadowing. The type checker will complain despite imports working.
> 
> # dlt/__init__.py
[...]

PR #2983, opened 2025-08-08
fix: top level relation

zilto:
## Related
- partial reimplementation of #2896 using more detailed commits and `git mv`

> - call it dlt.dataset instead of dlt._dataset
>   - (this will clash with top-level dlt.dataset() the dlt.Dataset factory; similar to the dlt.pipeline() issue we currently have #2934 )
> - keep the protocols as SupportsDataset  and SupportsRelation
> - use git mv
>   - I don't know if this will work, given the files still exist
> There will be no blockers?


zilto:
Closing because `git mv` doesn't improve GitHub diff view

PR #2984, opened 2025-08-09
Feat/2976 snowpark container auth

### LINKED ISSUE #2976: Support Snowflake-provided OAuth token in Snowpark Container Services
### Background 

When a dlt pipeline runs inside Snowpark Container Services, Snowflake [mounts an OAuth access token](https://docs.snowflake.com/en/developer-guide/snowpark-container-services/additional-considerations-services-jobs#connecting-with-a-snowflake-provided-oauth-token) at `/snowflake/session/token` and exposes `SNOWFLAKE_HOST` / `SNOWFLAKE_ACCOUNT` env vars. A standard connector call then looks like:

```python
def get_login_token():
    with open('/snowflake/session/token', 'r') as f:
        return f.read()

conn = snowflake.connector.connect(
    host = os.getenv('SNOWFLAKE_HOST'),
    account = os.getenv('SNOWFLAKE_ACCOUNT'),
    token = get_login_token(),
    authenticator = 'oauth'
)
```

### Task

Extend dlt’s Snowflake destination to auto-detect and use that token so no extra credentials are required in-container.

### Implementation

1. Update `SnowflakeCredentials.on_resolved` so if `/snowflake/session/token` exists and not explicitly passed via `token` param, we use it along with env's `SNOWFLAKE_HOST` and `SNOWFLAKE_ACCOUNT`
2. Do not require username argument
3. Update to_connector_params to use token, host and account
4. Include tests 
5. Update docs with an example

mohamedmeqlad99:
### Description
Added support for automatic OAuth token authentication when running inside Snowpark Container Services. This change enables dlt to:

- Auto-detect OAuth token from `/snowflake/session/token`
- Use `SNOWFLAKE_HOST` and `SNOWFLAKE_ACCOUNT` environment variables
- Support token-based auth without requiring username
- Maintain backward compatibility with existing auth methods

Changes include:
- Updated `SnowflakeCredentials` class
- Added comprehensive test coverage
- Updated Snowflake configuration documentation

### Related Issues

- Closes #2976

### Additional Context

Testing:
- Added unit tests in `test_snowflake_snowpark_token.py`
- Verified auto-detection works as expected
- Confirmed explicit credentials take precedence
- Validated connector parameter generation

Documentation:
- Added new section for Snowpark Container Services auth
- Included usage examples and override instructions

### DIFF
--- a/dlt/destinations/impl/snowflake/configuration.py
+++ b/dlt/destinations/impl/snowflake/configuration.py
@@ -1,4 +1,5 @@
 import dataclasses
+import os
 from pathlib import Path
 from typing import Final, Optional, Any, Dict, ClassVar, List
 
@@ -48,6 +49,23 @@ class SnowflakeCredentials(ConnectionStringCredentials):
                 setattr(self, param, self.query.get(param))
 
     def on_resolved(self) -> None:
+        # Auto-detect Snowpark Container Services token
+        token_path = "/snowflake/session/token"
+        if (
+            not self.token
+            and os.path.exists(token_path)
+        ):
+            try:
+                with open(token_path, "r") as f:
+                    self.token = f.read().strip()
+                # Use env vars if host not set
+                self.host = self.host or os.getenv("SNOWFLAKE_HOST") or os.getenv("SNOWFLAKE_ACCOUNT")
+                self.authenticator = self.authenticator or "oauth"
+            except Exception as ex:
+                raise ConfigurationValueError(
+                    f"Failed to read Snowpark Container Services token: {ex}"
+                )
+
         if self.private_key_path:
             try:
                 self.private_key = Path(self.private_key_path).read_text("ascii")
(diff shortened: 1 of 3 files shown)


dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -70,21 +90,31 @@ def get_query(self) -> Dict[str, Any]:
...
+        if self.token:
+            conn_params = dict(
+                query,
+                token=self.token,
+                account=self.host,
+                authenticator=self.authenticator or "oauth",
+                database=self.database,
+            )
+            # Remove user/password when using token auth
+            conn_params.pop("user", None)
```
burnash, line 108: This may break other oauth flows https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect#connecting-with-oauth where `user` is actually required.

burnash:
Thanks again for the PR. Closing it now due to inactivity

PR #2993, opened 2025-08-13
PoC: add pytest dlt package with useful helpers 

sh-rp:
### Description

We have a bunch of very useful testing helpers which we would like to share as a separate package to pypi. This PR servers as a starting point to discuss the repository structure with this package.

### Things to decide (with initial suggestions)
- [ ] The name of the test package (pytest-dlt)
- [ ] The versioning and release scheme (1-1 with the main dlt package)
- [ ] License (same as dlt package)
- [ ] Location of package (see current repo, it is in a folder and included in the main workspace via the uv sources section)
- [ ] Folder / module structure within this new test package (no proposals yet)


zilto:
Awesome feature (eager to use it myself). Regarding naming, the convention suggests `dlt-pytest` instead of `pytest-dlt` (see results for "pytest" on pypi)

sh-rp:
> Awesome feature (eager to use it myself). Regarding naming, the convention suggests `dlt-pytest` instead of `pytest-dlt` (see results for "pytest" on pypi)

chat-gpt says the convention is pytest first, but I am not sure that that is true. dlt-pytest sounds nicer to me too.

zilto:
I might have been wrong though. This [pytest docs page](https://docs.pytest.org/en/stable/reference/plugin_list.html#plugin-list) lists a lot of `pytest-PROJECT_NAME` plugins.

Though, if the number of `dlt` related projects grow, it can be useful to keep track of them using the `dlt-` prefix.

I'm ok with either

sh-rp:
Closing, so we may pick this up later

PR #2995, opened 2025-08-14
Jinso o fix/cors playground

Jinso-o:
### Description
Updated API to jaffelshop

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/docs/notebooks/playground/playground.py
+++ b/docs/notebooks/playground/playground.py
@@ -58,9 +58,7 @@ def connect(pipeline):
 @app.cell(hide_code=True)
 def tests(pipeline):
     # NOTE: this cell is only needed for testing this notebook on ci
-    count = pipeline.dataset().users.df().shape[0]
-    if count != 10:
-        raise AssertionError(f"expected 10 rows, got {count}")
+    assert pipeline.dataset().users.df().shape[0] == 10
     return
 
 
--- a/docs/notebooks/playground/playground.py
+++ b/docs/notebooks/playground/playground.py
@@ -17,7 +17,6 @@ async def initialize():
         await micropip.install("dlt[duckdb]")
         await micropip.install("pandas")
         # dependencies needed for ibis
-        await micropip.install("sqlite3")
         await micropip.install("ibis-framework[duckdb]")
 
     return sys, mo
--- a/docs/notebooks/playground/playground.py
+++ b/docs/notebooks/playground/playground.py
@@ -24,40 +24,40 @@ async def initialize():
 
 
 @app.cell
-def run(dlt):
+def run():
     import dlt, requests
-    
 
-    
     @dlt.resource(table_name="users")
     def users():
-        resp = requests.get("https://dummyjson.com/users", timeout=30)
-        yield from resp.json()["users"]
+        yield from requests.get("https://jsonplaceholder.typicode.com/users").json()
 
-    pipe = dlt.pipeline("dummy_users", "duckdb")
-    info = pipe.run(users())
-    print(info)
+    pipeline = dlt.pipeline(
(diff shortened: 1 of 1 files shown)


docs/notebooks/playground/playground.py
```diff
@@ -28,38 +27,40 @@ def run():
     import dlt
     import requests
 
-    @dlt.resource(table_name="users")
-    def users():
-        yield requests.get("https://jsonplaceholder.typicode.com/users").json()
+    @dlt.resource(name="customers")
+    def fetch_customers():
+        response = requests.get("https://jaffle-shop.dlthub.com/api/v1/customers")
+        yield from response.json()
```
AstrakhantsevaAA:
```suggestion
        yield response.json()
```

docs/notebooks/playground/playground.py
```diff
@@ -28,38 +27,40 @@ def run():
...
+    pipeline.dataset().customers.df()
     return
 
 
 @app.cell
 def connect(pipeline):
     # NOTE: This line allows your data to be explored in the marimo datasources which is the third item from the top in the left sidebar
     con = pipeline.dataset().ibis()
-    return
+    return (con,)
```
AstrakhantsevaAA, line 57: why do you return (con,) here?

docs/notebooks/playground/playground.py
```diff
@@ -28,38 +27,40 @@ def run():
...
 
 
 @app.cell(hide_code=True)
 def tests(pipeline):
     # NOTE: this cell is only needed for testing this notebook on ci
-    assert pipeline.dataset().users.df().shape[0] == 10
+    assert pipeline.dataset().customers.df().shape[0] > 0
```
AstrakhantsevaAA: how many customers this endpoint returns? can you use exact number here?

AstrakhantsevaAA (changes requested):

docs/notebooks/playground/playground.py
Jinso-o: affle-shop API (used in the playground) returns 100 customers. Therefore for the maintenance and sustainability I assumed it would be better to use `>0` then being specific with `== 100` as jaffle shop might be updated later. If you recommend me to be more specific I would then use range however would it be better to keep simplicity in the code?

AstrakhantsevaAA: It returns 100 because of default page_size=100 
https://jaffle-shop.dlthub.com/docs#/customers/get_customers_api_v1_customers_get

I also think we should use rest_api source, check the example here: https://www.notion.so/dlthub/OSS-One-Pager-benefit-to-user-version-1-25a9fb8e23cf800aa35acbe0a28b995d?source=copy_link#25a9fb8e23cf8054a7e9c4dda5044c64

docs/notebooks/playground/playground.py
sh-rp: I think marimo adds this if you are in edit mode

AstrakhantsevaAA: @Jinso-o just double check if it's not breaking anything

Jinso-o: will do!

Jinso-o: It's basically marimo's way of saying "make this variable available to other cells and the UI"!

docs/notebooks/playground/playground.py
Jinso-o: Should I comment the details? or should I leave it so that the playground users can play with it?

AstrakhantsevaAA: it looks good, leave it

PR #2997, opened 2025-08-15
fix: improve type hints for dataset and relation

zilto:
Changes:
- cleanup imports (`if TYPE_CHECKING` block, remove unused, `from __future__ import annotations`)
- use public imports (`dlt.Schema` instead of `dlt.common.schema.Schema`, `dlt.Dataset` instead of `dlt.dataset.dataset.Dataset`)
- move content of `dlt/dataset/factory.py` to `dlt/dataset/dataset.py`
- remove redundant `cast()`


rudolfix (approved):
LGTM!

rudolfix (approved):
LGTM!

Issue #2709, opened 2025-06-03
DuckLake support


adrianbr:
# Specs from rudolfix
(written by zilto, adapted from: https://github.com/dlt-hub/dlt/issues/2709#issuecomment-2958847943)

## Tasks
Create a new `ducklake` destination; it should inherit a lot of features from `duckdb`destination.

### Configuration
To allow to configure the catalog (a database) and storage (a filesystem), `DuckLakeCredentials` should derive from duckdb configuration and have the following signature

```python
@configspec(init=False)
class DuckLakeCredentials(DuckDbBaseCredentials):
    drivername: Final[str] = dataclasses.field(  # type: ignore
        default="ducklake", init=False, repr=False, compare=False
    )
    username: str
    password: TSecretStrValue = None
    database: str  # the name of the ducklake; required by DuckLakeSqlClient
    catalog: ConnectionStringCredentials  # for catalog; like postgres
    storage: FilesystemConfiguration  # for data;
```
Users will be able to configure the catalog and the storage from their `config.toml` and `secrets.toml`

### Resolve catalog and storage secrets
The duckdb connection needs credentials to the catalog ([postgres example](https://ducklake.select/docs/stable/duckdb/usage/choosing_a_catalog_database#postgresql)) and the storage ([supported storage](https://ducklake.select/docs/stable/duckdb/usage/choosing_storage))
- for storage, the duckdb destination already has a feature to get secrets from `FilesystemConfiguration` (i.e., the mechanism that allows to query S3 with duckdb)
- for catalog, we need to implement the function to get secrets from `ConnectionStringCredentials` 

### Configure DuckDB instance to support DuckLake
The class `DuckDbBaseCredentials` allows to set extensions, pragmas, global config, and local config. This allows to load the ducklake extension, but not to install it.

- `DuckLakeCredentials` should inherit from `DuckDbBaseCredentials`, but enforce the `ducklake` extension to be installed
- Set the current database to the ducklake name ([details here](https://github.com/dlt-hub/dlt/issues/2709#issuecomment-2948445610)).
- A lot of filesystem are [supported as duckdb extensions](https://ducklake.select/docs/stable/duckdb/usage/choosing_storage). `httpfs` supports all S3-compliant APIs. Also, it [supports python fsspec](https://duckdb.org/docs/stable/guides/python/filesystems.html). We can progressively add support here

### Out-of-scope
[...]

Issue #1692, opened 2024-08-14
support staging for duckdb and motherduck


rudolfix:
**Background**
duckdb/motherduck will load data from buckets so it is possible to use staging destination with them. This may decrease loading times esp. in case of Motherduck that can load s3 backend to backend.

**Requirements**
1. * [ ] pass credentials to duckdb and motherduck for all 3 buckets. (there's a mechanism to create such credentials)
2. * [ ] make sure you create credentials IN MOTHERDUCK on Motherduck
3. * [ ] if credentials for a given bucket exist, use them (read about scoping of secrets, we may also allow to configure secrets by name in config)
4. * [ ] create TEMPORARY credentials
5. * [ ] update COPY job to allow uris corresponding to buckets

PR #3015, opened 2025-08-19
feat: `ducklake` destination

zilto:
### Related Issues

- issue: https://github.com/dlt-hub/dlt/issues/2709

### Core changes
1. created new connection pool class and separated borrow/return conn from duckdb credentials.
2. connection pool can create new connections for every thread. ducklake does not work with default parallellism (the other thread always locks)
3. changes in derivation structure of `DuckDbBaseCredentials`: it is decoupled from `ConnectionStringCredentials` and used to configure connection options, pragmas, extensions and host connection pool. reason: ducklake does not need `ConnectionStringCredentials` at the top level. it only needs `catalog_name`. Overall this looks much nicer
4. connection pool still is still part of `DuckDbBaseCredentials`, Reason: this is the only singleton persisted during the whole load step. Correct implementation: allow for connection pool as top level entity. For now: good enough
5. a few bugs and undefined behaviors came out in configuration resolver. reason: ducklake configuration is complicated and modular ie. for the first time 3rd and 4th config embedding level goes to production (ducklake configuration -> ducklake credentials -> filesystem configuration -> filesystem credentials). fixes:
- first unresolved embedded configuration was stopping the resolution immediately. now all configurations are resolved till the end. example: catalog credentials were missing but filesystem credentials were configured. with the bug filesystem credentials were never resolved and for both defaults were used.
- as a consequence lookup traces that are collected are not recursive. previously only lookup traces from failing embedded configuration were collected. now we collect all of them
- error message is IMO way nicer and shows embedded traces. also shows the successfully resolved fields so user sees that actually something worked and something not
6. there's a mechanism in `dlt` that points all relative paths to data (ie. duckdb database, local filesystem data). I cleaned it up and make it easy to propagate redirects to embedded configurations (ie. now both catalog and storage are redirected). Still I'm not happy with how it is implemented... I'll give it another run in the future
7. There are docs for ducklake. Reading docs will help with the review


dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -202,6 +204,7 @@ def is_partial(self) -> bool:
         return self.database == ":pipeline:"
 
     def on_resolved(self) -> None:
+        # TODO Why don't we support `:memory:` string?
```
rudolfix: we support it. you can pass duckdb instance instead of credentials and destination factory will use it:
https://dlthub.com/docs/dlt-ecosystem/destinations/duckdb#destination-configuration (those docs will benefit from better section titles)

`:memory:` database is wiped out when connection is closed. during the loading the connection will be opened and closed several times. ie. to migrate schemas. and at the end all the data will be lost because we close all connection when loader exits

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -19,6 +19,8 @@
 DUCK_DB_NAME_PAT = "%s.duckdb"
 
 
+# NOTE duckdb extensions are only loaded when using the dlt cursor. They are not
+# loaded when using the native connection (e.g., when passing it to Ibis)
```
rudolfix: there's a mechanism to load extensions at start. it could be made easier for implementers but right now you can update `extensions` in `on_resolve` of DuckLakeCredentials(DuckDbBaseCredentials) (that you implement below).

some docs: https://dlthub.com/docs/dlt-ecosystem/destinations/duckdb#additional-configuration

another option you have is to subclass sql_client. see the base class.
```py
class DuckDbSqlClient(SqlClientBase[duckdb.DuckDBPyConnection], DBTransaction):
    dbapi: ClassVar[DBApi] = duckdb

    def __init__(
        self,
        dataset_name: str,
        staging_dataset_name: str,
        credentials: DuckDbBaseCredentials,
        capabilities: DestinationCapabilitiesContext,
    ) -> None:
        super().__init__(None, dataset_name, staging_dataset_name, capabilities)
        self._conn: duckdb.DuckDBPyConnection = None
        self.credentials = credentials
        # set additional connection options so derived class can change it
        # TODO: move that to methods that can be overridden, include local_config
        self._pragmas = ["enable_checkpoint_on_shutdown"]
        self._global_config: Dict[str, Any] = {
            "TimeZone": "UTC",
            "checkpoint_threshold": "1gb",
        }

    @raise_open_connection_error
    def open_connection(self) -> duckdb.DuckDBPyConnection:
        self._conn = self.credentials.borrow_conn(
            pragmas=self._pragmas,
            global_config=self._global_config,
            local_config={
                "search_path": self.fully_qualified_dataset_name(),
            },
        )
        return self._conn
```
and inject extensions on init or when connection is being opened

dlt/destinations/impl/duckdb/sql_client.py
```diff
@@ -546,3 +558,39 @@ def __del__(self) -> None:
         if self.memory_db:
             self.memory_db.close()
             self.memory_db = None
+
+
+def _install_extension(duckdb_sql_client: DuckDbSqlClient, extension_name: LiteralString) -> None:
```
rudolfix: mhmmm I think the code that adds extensions in `borrow_conn` will suffice. if not we can move those utils there?

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+# TODO support connecting to a snapshot
+
+
+@configspec(init=False)
+class DuckLakeCredentials(DuckDbCredentials):
+    def __init__(
+        self,
+        # TODO how does duckdb resolve the name of the database to the name of the dataset / pipeline
```
rudolfix: here's something that I may not fully grasp. but `DuckLakeCredentials` will create `:memory:` instance
- to which you attach `catalog` below
- to which you attach `storage`
- that gets configured with extensions and settings in `DuckLakeCredentials` (self)
- and this instance `DuckLakeCredentials` is used to borrow_con

so what should assume `dataset_name` here? catalog database if it is dukcdb? pls see below

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+        ducklake_name: str = "ducklake",
+        *,
+        catalog_database: Optional[Union[ConnectionStringCredentials, DuckDbCredentials]] = None,
```
rudolfix: postgres, mysql, duckdb, motherduck are all `ConnectionStringCredentials` so maybe that's enough to put here

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+        storage: Optional[FilesystemConfiguration] = None,
+        attach_statement: Optional[str] = None,
+    ) -> None:
+        if catalog_database is not None:
```
rudolfix: this should be executed in `on_resolve` handler. at this moment configuration will be injected from providers and you are able to do a last second checks

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+            raise NotImplementedError
+
+        if storage is not None:
+            raise NotImplementedError
+
+        import duckdb
+
+        super().__init__(duckdb.connect(":memory:"))
```
rudolfix: OK I see what you mean with `:memory:` what we should do here:
* `database` is set by the user and is a name of duck lake. no need to create new field (`ducklake_name`) for that (user can specify `ducklake://my_name` - like for other duckdb-based destinations)
* you should implement
```
def _conn_str(self) -> str:
    return ":memory:"
```
here. see how motherduck destination does it.

the end result is: 
1. `database` holds ducklake name which is IMO aligned with how `duckdb` and `motherduck` credentials work
2. you will open memory database to which ducklake will be attached

dlt/destinations/impl/ducklake/sql_client.py
```diff
@@ -0,0 +1,27 @@
...
+        capabilities: DestinationCapabilitiesContext,
+    ) -> None:
+        super().__init__(dataset_name, staging_dataset_name, credentials, capabilities)
+
+    # TODO support connecting to a snapshot
+    @raise_open_connection_error
+    def open_connection(self) -> DuckDBPyConnection:
+        """Ensure the `ducklake` extension is loaded. This needs to be done on every connection"""
+        super().open_connection()
+        self._conn.execute("LOAD ducklake;")
```
rudolfix: as mentioned above: you can add this to `extensions` in config and make `borrow_conn` to load this extension

dlt/destinations/impl/ducklake/sql_client.py
```diff
@@ -0,0 +1,27 @@
...
+        self._conn.execute(self.credentials.attach_statement)
```
rudolfix: the "problem" with ibis is that we do use `open_connection` to get connection that is passed to `ibis` but we use `borrow_conn` directly. This can be easily changed (connection ownership got solved by `move_conn`). I'm happy to do that

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def fingerprint(self) -> str:
+        """Returns a fingerprint of user access token"""
+        # TODO is this most appropriate?
+        if self.credentials and self.credentials.password:
+            return digest128(self.credentials.password)
+        return ""
+
+    def on_resolved(self) -> None:
+        local_db = self.make_location(
```
rudolfix: here we are redirecting files that are relative paths to a `local_dir`. so:
```
self.credentials.catalog_database.database =  self.make_location(
            self.credentials.catalog_database.database, DUCKLAKE_NAME_PATTERN
        )
```
should place the catalog database in `local_dir`. you should enable that only for `duckdb` drivername though. (and sqllite if you want to support it)

dlt/destinations/impl/ducklake/configuration.py
rudolfix: you can use `drivername` to distinguish them

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+
+@configspec(init=False)
+class DuckLakeCredentials(DuckDbCredentials):
+    def __init__(
+        self,
+        # TODO how does duckdb resolve the name of the database to the name of the dataset / pipeline
+        ducklake_name: str = "ducklake",
+        *,
+        catalog_database: Optional[Union[ConnectionStringCredentials, DuckDbCredentials]] = None,
+        storage: Optional[FilesystemConfiguration] = None,
```
rudolfix: hmmm we call it `filesystem` in `iceberg` but I think you are right here... and we should deprecate iceberg setting

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+            raise NotImplementedError
+
+        if storage is not None:
+            raise NotImplementedError
+
+        import duckdb
+
+        super().__init__(duckdb.connect(":memory:"))
+
+        self.ducklake_name = ducklake_name
```
rudolfix: 1. all the config fields will be set by the `resolve` module. you do not need to do it yourself
2. it is very rare you need to implement `__init__` on dataclass. to do post resolve checks and modifications, you use `on_resolve`
3. IMO you do not need this `__init__`. Look at the comment below on how to get in memory instance while keeping configured `database`

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @staticmethod
+    def build_attach_statement(
+        *,
+        ducklake_name: str,
+        catalog_database: Union[ConnectionStringCredentials, str],
+        storage: Optional[FilesystemConfiguration] = None,
+    ) -> str:
+        # TODO resolve ConnectionStringCredentials; duckdb has its own format
+        attach_statement = f"ATTACH IF NOT EXISTS 'ducklake:{ducklake_name}.ducklake'"
```
rudolfix: I've looked briefly here
https://ducklake.select/docs/stable/duckdb/usage/choosing_a_catalog_database

and it seems each catalog database has slightly different format when connection. so you'll need to use drivername to generate right statement
- actual implementation will work for duckdb (and sqllite if you add `sqllite:`) prefix
- postgres: it is a custom mapping from `ConnectionStringCredentials` to what ducklake means

so there's no universal mapping from `ConnectionStringCredentials` to duckdb attach which I mentioned in the ticket.

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+# TODO maybe this needs to be parameterized by `catalog` and `storage` implementation
+def _get_ducklake_capabilities() -> DestinationCapabilitiesContext:
+    caps = DestinationCapabilitiesContext()
+    caps = _set_duckdb_raw_capabilities(caps)
+    return caps
+
+
+# TODO support connecting to a snapshot
```
rudolfix: that would be amazing but we can do that later. snapshots mean reproducible local environments that you can get with 0 copy

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+        *,
+        ducklake_name: str,
+        catalog_database: Union[ConnectionStringCredentials, str],
+        storage: Optional[FilesystemConfiguration] = None,
+    ) -> str:
+        # TODO resolve ConnectionStringCredentials; duckdb has its own format
+        attach_statement = f"ATTACH IF NOT EXISTS 'ducklake:{ducklake_name}.ducklake'"
+        if storage:
+            # TODO handle storage credentials by creating secrets
+            attach_statement += f" (DATA_PATH {storage.bucket_url})"
```
rudolfix: you should pass `storage` to `create_secret` before you attach (after you open the connection)

tests/load/ducklake/test_ducklake_client.py
```diff
@@ -0,0 +1,150 @@
...
+from tests.utils import (
+    TEST_STORAGE_ROOT,
+    patch_home_dir,
+    autouse_test_storage,
+    preserve_environ,
+    wipe_pipeline,
+)
+
+
+def test_native_duckdb_workflow(tmp_path):
```
rudolfix, line 18: makes sense to do a few "smoke tests". the next step would be to enable ducklake to be tested for exactly the same tests as duckdb using ie. local duckdb as catalog and local filesystem as storage.

let's do another iteration of this ticket and then I'll look at this. I was able to do the same with `iceberg` destination so I'm pretty sure it will work

dlt/destinations/impl/ducklake/configuration.py
```diff
@@ -0,0 +1,115 @@
...
+                storage=self.storage,
+            )
+
+    @attach_statement.setter
+    def attach_statement(self, value: str) -> None:
+        self._attach_statement = value
+
+
+# TODO add connection to a specific snapshot
+# TODO does it make sense for ducklake to have a staging destination?
```
rudolfix: good point see here: https://github.com/dlt-hub/dlt/issues/1692

dlt/destinations/impl/ducklake/configuration.py
rudolfix: if that does not work, I can help with debugging

dlt/destinations/impl/ducklake/factory.py
```diff
@@ -0,0 +1,65 @@
...
+    def spec(self) -> type:
+        return DuckLakeClientConfiguration
+
+    @property
+    def client_class(self) -> type:
+        from dlt.destinations.impl.ducklake.ducklake import DuckLakeClient
+
+        return DuckLakeClient
+
+    def _raw_capabilities(self) -> DestinationCapabilitiesContext:
```
rudolfix, line 86: note: ducklake will support `upsert` (MERGE INTO) so we can enable this strategy to see if it works

rudolfix (changes requested):
this looks really good! here's summary of my suggestion:

- simplify ducklake credentials class (ie. remove `__init__`, implement `_conn_str()`
- load extensions in `borrow_conn`
- we'll need to tweak how connections are opened in ibis handover (but that's easy)

dlt/destinations/impl/ducklake/configuration.py
zilto: The format used in ducklake is
```
ducklake:{drivername}:{connection_args}
```
`connection_args` can be a file path (duckdb, sqlite) or key-value pairs (postgres, myssql)

So you get 
```
# duckdb is implicit
'ducklake:my_duckdb.duckdb'
# sqlite
'ducklake:sqlite:my_sqlite.db' 
# postgres
'ducklake:postgres:host=... port=... password=...'
```

You can also define secrets in ducklake such that 
```
# create secret `my_postgres` with connection info
'ducklake:my_postgres'
```

dlt/destinations/impl/ducklake/configuration.py
zilto: For the default case, here's what I'm currently aiming for:

```python
pipeline = dlt.pipeline("jaffle_shop", destination="ducklake")
pipeline.run(...)
```

- a `duckdb` instance is created in `:memory:`; we call it the `ducklake_client`
- the `ducklake_client` installs the `ducklake` extension for duckdb (needs to be done once per system)
- the `ducklake_client` uses the `ATTACH` command to load a `catalog` and `storage`
- the `catalog` is a duckdb instance on disk (with extension `.ducklake` instead of `.duckdb` by convention)
- the default `storage` is completely handled by DuckDB / DuckLake

The outcome is 
```
|- pipeline.py
|- jaffle_shop.ducklake  # catalog file (if duckdb or sqlite)
|- jaffle_shop.ducklake.files/  # storage
   |- main/  # schema level
      |- customers/  # table level
          |- data.parquet  # data
      |- orders/
```

### Design
- The `DuckLakeCredentials` inherits from `DuckDbCredentials` and the "main" credentials are used to define the `ducklake_client` 
- We always use an in-memory DuckDB connection for the `ducklake_client`

rudolfix:
there were pretty complicated issues with parallel loading, also depended on catalog types. I underestimated the effort a little when writing original ticket... luckily it seems that most of the work is done. below is a description of changes I've made. also see the commit log:

1. created new connection pool class and separated borrow/return conn from duckdb credentials. it got too convoluted
2. connection pool can create new connections for every thread. ducklake does not work with default paralellism (the other thread always locks)
3. enabled standard destination tests (well, they were already enabled), just 2-3 tests are not passing, to be investigated
4. if duckdb and sqllite are configured as catalogs, load jobs are sequential
5. I added ibis handling. seems to work.
6. support for sqllite, duckdb, postgres (with a simpe test). emits storage secrets (no tests), mysql support (no tests)

what is left:
1. tests: best if we run ducklake tests again as separate remote tests with remote postgres catalog + do a few smoke tests on supported buckets
2. motherduck catalog is not supported
3. open table interface is not supported (ie. to compact the tables)
4. I left most the commented code and module docstrings. may need cleanup
5. we are missing docs

let's touch base how and when to continue here

docs/website/docs/dlt-ecosystem/destinations/duckdb.md
```diff
@@ -280,13 +282,16 @@ dest_ = dlt.destinations.duckdb(
...
+
 
-### dbt support
+## dbt support
 This destination [integrates with dbt](../transformations/dbt/dbt.md) via [dbt-duckdb](https://github.com/jwills/dbt-duckdb), which is a community-supported package. The `duckdb` database is shared with `dbt`. In rare cases, you may see information that the binary database format does not match the database format expected by `dbt-duckdb`. You can avoid this by updating the `duckdb` package in your `dlt` project with `pip install -U`.
 
 NOTE: extensions, pragmas and configs are not propagated from `dlt` configuration to the dbt profile at this moment.
 
-### Syncing of `dlt` state
+###Syncing of `dlt` state
```
burnash:
```suggestion
### Syncing of `dlt` state
```

dlt/common/configuration/specs/connection_string_credentials.py
```diff
@@ -84,8 +84,11 @@ def _serialize_value(v_: Any) -> str:
...
-        url = url._replace(query=None)
-        # we only have control over netloc/path
-        return url.render_as_string(hide_password=True)
+        try:
+            url = self.to_url()
+            # do not display query. it often contains secret values
+            url = url._replace(query=None)
+            # we only have control over netloc/path
+            return url.render_as_string(hide_password=True)
+        except Exception:
```
burnash, line 93: Should the exception be more specific here? Could swallowing all exceptions hide potential issues?

dlt/destinations/impl/ducklake/sql_client.py
```diff
@@ -0,0 +1,158 @@
...
+        # creates a separate connection for each sql_client
+        try:
+            if not self.credentials.storage.is_local_filesystem:
+                self.create_secret(
+                    self.credentials.storage.bucket_url, self.credentials.storage.credentials
+                )
+            # NOTE: database must be detached otherwise it is left in inconsistent state
+            # TODO: perhaps move attach/detach to connection pool
+            self._conn.execute(self.attach_statement)
+            self._conn.execute(f"USE {self.credentials.catalog_name};")
```
burnash, line 60: Should this use `escape_identifier()` instead of passing a bare catalog_name?

dlt/destinations/impl/ducklake/sql_client.py
```diff
@@ -0,0 +1,158 @@
...
+        ):
+            if protocol in ["gs", "gcs"]:
+                logger.warning(
+                    "For gs/gcs access via duckdb please use the gs/gcs s3 compatibility"
+                    "layer if possible. "
+                    "Falling back to fsspec which degrades scanning performance."
+                )
+                self._register_filesystem(fsspec_from_config(self.credentials.storage)[0], "gcs")
+            else:
+                raise ValueError(f"Cannot create secret or register filesystem for `{protocol=:}`")
```
burnash, line 115: The `:` after the = is an empty format specifier. I wonder why is it needed here?

dlt/common/configuration/exceptions.py
```diff
@@ -41,26 +60,65 @@ def __init__(self, _typ: type) -> None:
...
         self.fields = list(traces.keys())
-        super().__init__(spec_name)
-
-    def __str__(self) -> str:
-        msg = f"Missing fields in configuration: {str(self.fields)} {self.spec_name}\n"
-        for f, field_traces in self.traces.items():
+        super().__init__(self.spec_name)
+
+    @staticmethod
+    def _print_traces(
```
burnash, line 72: The name `_print_traces` is misleading here as it doesn't actually print anything rather it returns a string, also "traces" is vague - it doesn't indicate this is for error reporting. I'd suggest `_build_error_message`

dlt/destinations/impl/ducklake/sql_client.py
```diff
@@ -0,0 +1,158 @@
...
+
+from dlt.common.storages.configuration import FileSystemCredentials
+from dlt.common.storages.fsspec_filesystem import fsspec_from_config
+from dlt.destinations.impl.duckdb.configuration import DuckDbCredentials
+from dlt.destinations.impl.duckdb.sql_client import DuckDBDBApiCursorImpl, DuckDbSqlClient
+from dlt.destinations.impl.ducklake.configuration import DuckLakeCredentials
+from dlt.destinations.sql_client import raise_open_connection_error
+
+
+class DucklakeDBApiCursorImpl(DuckDBDBApiCursorImpl):
```
burnash, line 16: Minor: inconsistent naming: `DucklakeDBApiCursorImpl` vs `DuckLake*` elsewhere.

zilto:
Adding here for legacy: `sqlite3` in Python has deep-rooted issues with concurrency. 

> Currently, the sqlite3 DB-API 2.0 attribute 'threadsafety' is hard-coded to 1, meaning "threads may share the module, but not connections". This is not always true, since it depends on the default SQLite threaded mode, selected at compile-time with the SQLITE_THREADSAFE define.

The parameter `SQLITE_THREADSAFE` can't be changed for `Python < 3.11` ([reference](https://bugs.python.org/issue45613))

dlt/common/configuration/specs/connection_string_credentials.py
zilto: @rudolfix You should fix directly because I don't know what type of exceptions you expect

dlt/destinations/impl/ducklake/sql_client.py
zilto: You're right that it's better to escape the identifier. We should also improve assertions to check that `catalog_name` is a valid DuckDB SQL identifier here. i.e., there's no funny character  in `catalog_name`

Issue #3011, opened 2025-08-19
Incomplete Fix for Stage Bucket URL Handling -> Relates Change (PR #2342)


aiAdrian:
### dlt version

1.9 ...  1.15

## Incomplete Fix for Stage Bucket URL Handling
The fix implemented in #2342 does not fully address all edge cases related to stage_bucket_url parsing.

Currently, the logic only checks whether the stage_bucket_url consists of {scheme} and {netloc} and assumes an empty path (e.g., az://my_net_loc/). This works fine for URLs without a subfolder. However, if the user specifies a subfolder (e.g., az://my_net_loc/prefix/), the current implementation causes a "file not found" error when accessing Azure.

### 💡 Suggested Fix
The issue lies in the logic at [utils.py#L94:](https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/snowflake/utils.py#L94)

```python 
# current implementation 
relative_url = parsed_file_url.path.lstrip("/")

# If stage bucket URL has a path, remove it from the beginning of the relative URL because this is already specified
# in the Snowflake stage location.
if stage_bucket_url:
    parsed_bucket_url = urlparse(stage_bucket_url)
    stage_bucket_path = parsed_bucket_url.path.lstrip("/")
    if stage_bucket_path:
        relative_url = relative_url.removeprefix(stage_bucket_path)
```

### fixed version 
```python 
relative_url = parsed_file_url.path.lstrip("/")

# If stage bucket URL has a path, remove it from the beginning of the relative URL because this is already specified
# in the Snowflake stage location.
if stage_bucket_url:
    parsed_bucket_url = urlparse(stage_bucket_url)
    stage_bucket_path = parsed_bucket_url.path.lstrip("/")
    if parsed_bucket_url.netloc and parsed_bucket_url.scheme:
        relative_url = relative_url.removeprefix(f"{parsed_bucket_url.scheme}://{parsed_bucket_url.netloc}/")
    elif stage_bucket_path:
        relative_url = relative_url.removeprefix(stage_bucket_path)

```

### Benefit
This change ensures that URLs with non-empty paths (i.e., subfolders) are correctly handled, avoiding file access errors in Azure and improving compatibility with Snowflake stage locations.

Let me know if you'd like me to submit this as a pull request as well!

### Steps to reproduce

Create a stage_bucket_url az blob storage stage with a subfolder 

#### Test Case 1
- Install dlt 1.8 
[...]

Issue #2342, opened 2025-02-24
Snowflake COPY INTO cannot use Snowflake S3 Stage


stevenayers:
### dlt version

1.7.0

### Describe the problem

Unless your Snowflake S3 stage is the root of an S3 bucket, [gen_copy_sql](https://github.com/dlt-hub/dlt/blob/eac0204e467a615493b83937c07d1204c3b3c967/dlt/destinations/impl/snowflake/snowflake.py#L102-L203) interpolates the relative path of your staged files incorrectly.

The ingestion will fail, because we're using `parsed_file_url.path.lstrip('/')` to get the path of the staging file, which doesn't take into account that the snowflake s3 stage may make up part of the `parsed_file_url.path`.

https://github.com/dlt-hub/dlt/blob/eac0204e467a615493b83937c07d1204c3b3c967/dlt/destinations/impl/snowflake/snowflake.py#L131-L134

**Example Code:**

```python
import os
import dlt
from dlt.destinations import snowflake, filesystem

MY_STAGE_PATH = "s3://my-bucket-name/dlt/"

@dlt.resource
def jaffle_shop():
    for i in range(10):
        yield {"id": i, "name": f"This is item {i}"}

destination = snowflake(
    credentials={
        "host": os.environ["SNOWFLAKE_ACCOUNT"],
        "warehouse": os.environ["SNOWFLAKE_WAREHOUSE"],
        "database": f"{os.environ['SNOWFLAKE_DATABASE']}",
        "role": os.environ["SNOWFLAKE_ROLE"],
        "username": os.environ["SNOWFLAKE_USER"],
        "password": os.environ["SNOWFLAKE_PASSWORD"],
    },
    stage_name="STAGING.DLT" # Set the path on this to MY_STAGE_PATH
)

staging = filesystem(
    bucket_url=MY_STAGE_PATH,
    credentials={
        "aws_access_key_id": os.environ["AWS_ACCESS_KEY_ID"],
        "aws_secret_access_key": os.environ["AWS_SECRET_ACCESS_KEY"],
        "region_name": os.environ["AWS_REGION"],
    },
)

pipeline = dlt.pipeline(
    pipeline_name="test",
    destination=destination,
    staging=staging,
    dataset_name="my_dataset",
)

pipline.run(jaffle_shop)

```

The staging files are saved to: `s3://my-bucket-name/dlt/my_dataset/jaffle_shop/`  ✅
[...]

PR #3016, opened 2025-08-20
fix: Incomplete Fix for Stage Bucket URL Handling -> Relates Change (PR #2342

aiAdrian:
Fixes: *Incomplete Fix for Stage Bucket URL Handling* -> Relates Change (PR #2342)

### Description
Fixes the issue  https://github.com/dlt-hub/dlt/issues/3011
- Currently, the logic only checks whether the stage_bucket_url consists of {scheme} and {netloc} and assumes an empty path (e.g., az://my_net_loc/). This works fine for URLs without a subfolder. However, if the user specifies a subfolder (e.g., az://my_net_loc/subfolder/), the current implementation causes a "file not found" error when accessing Azure. 
This change/pull request resolves the problem and improves the usability.

### Related Issues
- https://github.com/dlt-hub/dlt/issues/2342

### DIFF
--- a/dlt/destinations/impl/snowflake/utils.py
+++ b/dlt/destinations/impl/snowflake/utils.py
@@ -94,7 +94,9 @@ def gen_copy_sql(
         if stage_bucket_url:
             parsed_bucket_url = urlparse(stage_bucket_url)
             stage_bucket_path = parsed_bucket_url.path.lstrip("/")
-            if stage_bucket_path:
+            if parsed_bucket_url.netloc and parsed_bucket_url.scheme:
+                relative_url = relative_url.removeprefix(f"{parsed_bucket_url.scheme}://{parsed_bucket_url.netloc}/")
+            elif stage_bucket_path:
                 relative_url = relative_url.removeprefix(stage_bucket_path)
 
         from_clause = f"FROM @{stage_name}"


dlt/destinations/impl/snowflake/utils.py
```diff
@@ -94,7 +94,9 @@ def gen_copy_sql(
         if stage_bucket_url:
             parsed_bucket_url = urlparse(stage_bucket_url)
             stage_bucket_path = parsed_bucket_url.path.lstrip("/")
-            if stage_bucket_path:
+            if parsed_bucket_url.netloc and parsed_bucket_url.scheme:
+                relative_url = relative_url.removeprefix(f"{parsed_bucket_url.scheme}://{parsed_bucket_url.netloc}/")
```
burnash, line 98: Duplicating [my comment](https://github.com/dlt-hub/dlt/issues/3011#issuecomment-3263960582):

This code most likely has no effect as `relative_url` is already `path` from parsed. See `relative_url = parsed_file_url.path.lstrip("/")` and `parsed_file_url = urlparse(file_url)` from above

burnash (changes requested):
I believe this fix is wrong. See my comments and please share your config in the issue so we can help: #3011

Issue #2969, opened 2025-08-06
Move dlt educational courses into core repo


sh-rp:
### TLDR

We have two really cool educational courses in our dlthub-education repository: https://github.com/dlt-hub/dlthub-education:

* dlt fundamentals: https://github.com/dlt-hub/dlthub-education/tree/main/courses/dlt_fundamentals_dec_2024
* dlt advanced: https://github.com/dlt-hub/dlthub-education/tree/main/courses/dlt_advanced_2025

It would be cool to store them in our OSS repo, this way 
* they should be easier to find for llm crawlers, 
* we can lint them and keep them up to date
* we can create marimo versions 
* we can launch colabs and possibly molabs (the marimo colab version) directly from the repo links.
* The users will also be able to run them locally if they want to.

### Implementation

Let's implement this in three phases, one PR each:

#### 1. Copy colabs to oss repo

- [ ] Copy all colabs of the two courses (we can add more later) to the core repo as IPYNB files. I would propose to add them to `docs/education/dlt-fundamentals-course` and `docs/education/dlt-advanced-course`. Give them useful filenames derived from the topics.
- [ ] Create new docs pages where these are linked. Please talk to @AstrakhantsevaAA about where they go in the docs structure. Maybe `getting started -> education`? Those pages should look like the ones in the repo above with a list of links to directly open the colabs, but these links should open the colabs directly from the github version. (Example link: https://colab.research.google.com/github/dagworks-inc/hamilton/blob/main/examples/caching/tutorial.ipynb)
- [ ] Add instructions on how to run these colabs locally with jupyter
- [ ] You may need to exclude iypthon files from our max file size checker

#### 2. Quality assurance and updated links

- [ ] Add linting to the colabs by creating a lint-notebooks command to the main makefile which lints and typechecks them. We need quite lenient linter rules, not much should have to be changed. We want to catch breaking changes that happen in the core library so that our notebooks stay up to date. @zilto knows how to lint ipython notebooks.
- [ ] Try to update all places where we link to our old colabs to use these new github links. This would be in the education repository and also in our learnworlds platform (talk to @AstrakhantsevaAA about this)

adrianbr:
Comment on 
"Add instructions on how to run these colabs locally with jupyter"
I think we should provide a colab link too, otherwise we force data engineer learners into a workflow and tooling they might not be happy to use. Running jupyter notebooks is a nice extra option but doesn't replace the ease of diving in of colabs

Comment on 
[...]

PR #3017, opened 2025-08-20
Docs: Education notebooks formatted and linted

anuunchin:
This PR formats and lints education content and is the second PR in the series of PRs related to moving educational notebooks to OSS.

### Linting: 
- The tool called `nbqa` was used that creates temporary `.py` files from the `.ipynb` ones, so that flake and mypy can properly handle notebook files.

### Formatting:
- The formatting is done with `black` which has a `jupyter` extension.

> pyproject.toml and uv.lock are adjusted.

Relates to #2969.


uv.lock
zilto, line 1: I think we need to exclude `uv.lock` changes from this PR and produce the new lock when merging `devel` into `main`. Otherwise, it will create issues for branches of `devel`

zilto (changes requested):
Remove `uv.lock`. Otherwise, everything seems good

uv.lock
sh-rp: Why do you think we need it? You can resolve the problems locally if there is a lockfile conflict, it should not be a problem, it happens quite often.

sh-rp (approved):

sh-rp (approved):

Issue #3007, opened 2025-08-18
feat(dashboard): create reusable marimo widgets


zilto:
Currently, the "pipeline dashboard" is a marimo app that works "end-to-end" and is shipped as is. It includes:
- a pipeline state viewer
- sql data browser
- load package explorer
- and more

These components could be packages as functions that take a `dlt.Pipeline`, `dlt.Schema`, or destination as input. 

## Outcome
Users can import the widgets they want / need in their custom data apps / dashboard / reports they build at work. 

## Benefits / value
- proportional to the number of users of dlt + marimo + this specific feature
- small / medium value; nice for users, but also improves maintenability for devs

## Implementation
- this implementation will be specific to marimo
- determine the right namespace `dlt.helpers.marimo.widgets` ?
- write unit tests for widgets
- (optional) update the pipeline dashboard to use the widgets
  - this could class with existing tests, but could also benefit from more modular unit tests

zilto:
Implemented the mechanism to create widgets; 

Next steps:
- implement more widgets
- reimplement dashboard as a set of widgets

PR #3021, opened 2025-08-21
feat: dlt widgets for marimo

zilto:
This introduces mechanisms to build reusable widgets for marimo notebooks.

related: #3007 

## Changes
- Add `dlt.helpers.marimo` which contains widgets and utils
- Add a `load_package_viewer` and a `schema_widget` widgets
- Add a `README` tutorial on how to add dlt widgets; this is internal for now, but can be a useful resource for end users
- Added tests for some utilities

## How it works
A "dlt widget" is a full marimo notebook. This notebook can be called from another notebook using the native marimo feature [`app.embed()`](https://docs.marimo.io/api/app/?h=embed#marimo.App.embed). The `render()` function provides an interface that's easier to use for end users.

Currently, widgets don't output data. It is possible and easy to enable this. Enabling data output means we're creating a new public interface and have a "contract" to our users regarding backwards compatibility. We should enable this only once we're confident (and improve our testing around interfaces)

It is possible to get the widgets output by doing
```python
#%% cell 1
from dlt.helpers.marimo import render, load_package_inspector

#%% cell 2
# this is regular usage
await render(load_package_inspector)

#%% cell 3
# this gives data access; we offer no guarantee
result = await load_package_inspector.render()
result.output  # this renders the interactive widget

#%% cell 4
result.defs  # this returns data from the widget; will update as you click
``` 

## Future work
Pass arguments to the "dlt widget" marimo notebook; currently, `widget.embed()` doesn't support passing data. It can be achieved in hacky ways, but don't know how durable or safe it is. To discuss with the marimo team


rudolfix (approved):
LGTM!
I've merged devel here and tests are passing now

regarding INSERT VALUES - I'll get rid of this format in favor of jsonl and parquet + execute_many. INSERT VALUEs was an idea to generate a file in normalize stage and then just feed it into sql database without further parsing. but is seems execute many may be faster

PR #3029, opened 2025-08-24
Feature: Introduce support of http based resources for fs source

### LINKED ISSUE #3028: Introduce a http based resources support by filesystem source
### Feature description

Hi there!

I would like to extend filesystem source to add a support of http resources.  There are some public resources exposed by public CDNs which are accessible over https. They are quite limited in their capabilities, but it's enough to put `bucket_url` and fixed rule for globbing in order to download a particular file. `fsspec` provides `HTTPFilesystem` which can be leveraged to support the feature. I made a [PR](https://github.com/fsspec/filesystem_spec/pull/1909) to provide `Last-Modifed` timestamp required by filesystem source. 

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Example: [NYC Trip Data](https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2025-01.parquet)

### Proposed solution

1. Update ffspec HTTPFileSystem to provide "Last-Modified" date, if it's possible.
2. Update filesystem source

### Related issues

https://github.com/fsspec/filesystem_spec/pull/1909

TheLazzziest:
### Description

This PR introduces native support for HTTP-based data sources, such as those provided by CDNs like CloudFront. By leveraging fsspec's HTTPFileSystem, the filesystem source can now read remote files directly from http:// and https:// URLs without a prior local download. This enhancement makes it possible to easily ingest data from a wide range of public datasets.

### Related Issues

* Resolves #3028 

### Additional Context

A key challenge addressed in this implementation is the lack of a standard Last-Modified timestamp in HTTPFileSystem, which is crucial for dlt's filesystem source. To handle this, I added a [PR](https://github.com/fsspec/filesystem_spec/pull/1909) introducing Last-Modified property for http based resources ensuring the conformance with the main implementation.

### DIFF
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -131,7 +131,7 @@ jobs:
         if: matrix.python-version != '3.14.0-beta.4'
 
       - name: Install pipeline and sources dependencies
-        run: uv sync ${{ matrix.uv_sync_args }} --extra duckdb --extra cli --extra parquet --extra deltalake --extra sql_database --group sentry-sdk --group pipeline --group sources --group ibis
+        run: uv sync ${{ matrix.uv_sync_args }} --extra http --extra duckdb --extra cli --extra parquet --extra deltalake --extra sql_database --group sentry-sdk --group pipeline --group sources --group ibis
         if: matrix.python-version != '3.14.0-beta.4'
 
       - name: Run extract and pipeline tests
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,9 @@ az = [
 sftp = [
     "paramiko>=3.3.0"
 ]
+http = [
+    "aiohttp>3.9.0"
+]
 snowflake = ["snowflake-connector-python>=3.5.0"]
 motherduck = [
     "duckdb>=0.9",
--- a/uv.lock
+++ b/uv.lock
@@ -2140,6 +2140,9 @@ gcp = [
 gs = [
     { name = "gcsfs" },
 ]
+http = [
+    { name = "aiohttp" },
+]
 lancedb = [
     { name = "lancedb", marker = "python_full_version < '3.13'" },
     { name = "pyarrow" },
@@ -2336,6 +2339,7 @@ requires-dist = [
     { name = "adlfs", marker = "extra == 'az'", specifier = ">=2024.7.0" },
     { name = "adlfs", marker = "extra == 'clickhouse'", specifier = ">=2024.7.0" },
     { name = "adlfs", marker = "extra == 'synapse'", specifier = ">=2024.7.0" },
+    { name = "aiohttp", marker = "extra == 'http'", specifier = ">3.9.0" },
     { name = "alembic", marker = "extra == 'sqlalchemy'", specifier = ">1.10.0" },
     { name = "botocore", marker = "extra == 'athena'", specifier = ">=1.28" },
     { name = "botocore", marker = "extra == 'filesystem'", specifier = ">=1.28" },
@@ -2430,7 +2434,7 @@ requires-dist = [
     { name = "weaviate-client", marker = "extra == 'weaviate'", specifier = ">=3.22" },
     { name = "win-precise-time", marker = "python_full_version < '3.13' and os_name == 'nt'", specifier = ">=1.4.2" },
 ]
-provides-extras = ["gcp", "bigquery", "postgres", "redshift", "parquet", "duckdb", "ducklake", "filesystem", "s3", "gs", "az", "sftp", "snowflake", "motherduck", "cli", "athena", "weaviate", "mssql", "synapse", "qdrant", "databricks", "clickhouse", "dremio", "lancedb", "deltalake", "sql-database", "sqlalchemy", "pyiceberg", "postgis", "workspace", "dbml"]
+provides-extras = ["gcp", "bigquery", "postgres", "redshift", "parquet", "duckdb", "ducklake", "filesystem", "s3", "gs", "az", "sftp", "http", "snowflake", "motherduck", "cli", "athena", "weaviate", "mssql", "synapse", "qdrant", "databricks", "clickhouse", "dremio", "lancedb", "deltalake", "sql-database", "sqlalchemy", "pyiceberg", "postgis", "workspace", "dbml"]
 
 [package.metadata.requires-dev]
 adbc = [{ name = "adbc-driver-postgresql", specifier = ">=1.6.0" }]


rudolfix (changes requested):
this looks really good, also thanks for improving upstream libraries. A few improvements are still needed in order to merge it.

1. please do not modify `parse_iso_like_datetime` it really should parse iso-like strings and is used mostly in the normalizer. if you need to parse other file format, please add another function there and use it to parse last modified values. also `pendulum.parse` parses `now` string as now() 🤯 so we do not use it. for HTTPS headers it may be ok though

2. there are a few crucial `glob` tests for `filesystem` in `test_filesystem_source.py`. you can easily see how test cases are created for available buckets. you should be able to add httpfs to it and create files with test cases on the endpoint you configured right now

pls tell me if you have time to do the above. we could take this PR over at some point but it will take some time

TheLazzziest:
Hi @rudolfix! Thank you for the feedback. I have time to get these changes done.

> there are a few crucial glob tests for filesystem in test_filesystem_source.py. you can easily see how test cases are created for available buckets. you should be able to add httpfs to it and create files with test cases on the endpoint you configured right now

I saw the examples for other buckets (e.g.: `test_file_list`), however CDNs restrict file globbing if it's not a direct path to a specific file. In this regard, dlt can only fetch a particular file rather going over directories to find all of them. Thus, adding a http bucket to the rest of the testcases will force me to update all the tests. That's why I added a separate test rather trying to update the current ones. Does it sound reasonable to you or I am missing something ?

TheLazzziest:
Hi @rudolfix,

While working on the HTTP filesystem changes, I've found  a bug with `posixpath.join` in [glob_files](https://github.com/dlt-hub/dlt/blob/devel/dlt/common/storages/fsspec_filesystem.py#L358) function.

When `HTTPFileSystem` is configured with a trailing slash in `bucket_url` and a leading slash in `file_glob`, `posixpath.join` incorrectly ignores the `bucket_url`, leading to a `FileNotFoundError`.

**Example:**
- **Failing Configuration:**
```python
filesystem(bucket_url="https://d37ci6vzurychx.cloudfront.net", file_glob="/trip-data/yellow_tripdata_2025-01.parquet")`
```
- **Working Configuration:** 
```python
filesystem(bucket_url="https://d37ci6vzurychx.cloudfront.net/trip-data", file_glob="yellow_tripdata_2025-01.parquet")`
```

This behavior is difficult to debug due to a lack of validation on the resource path structure. So, my suggestion would be to replace the current path normalization block in glob_files with the FilesystemConfiguration object.

This change would delegate all path-rectifying work to the configuration object itself, streamlining the code and making the behavior much more predictable. Moreover, it would bring several key benefits: 

* **Fail-fast approach**: Errors could be caught during the initial configuration phase rather than during data processing.
* **Improved debugging**: We'd get more descriptive error messages, making it much faster to pinpoint configuration issues.
* **Developer guardrails**: It would introduce clear guardrails for developers, preventing common path-related mistakes.
* **Better fsspec integration**: The implementation would delegate more responsibility to fsspec, aligning more closely with its intended design (if that makes sense from your end)."

I'm happy to implement the fix, but I wanted to get your thoughts on the best way to handle this. Let me know if you agree with my assessment, so I can make another issue for that later.

TheLazzziest:
@rudolfix , I took the courage to do some refactoring of the datetime parsing and conversion logic while I was working on parsing changes — mainly in `dlt/common/time.py` and related typing. My main goals were to simplify the code, reduce duplication, and make the type handling more consistent and explicit. Let me know what you think! :pray:

dlt/common/storages/fsspec_filesystem.py
```diff
@@ -63,6 +63,7 @@ class FileItem(TypedDict, total=False):
     "adl": lambda f: ensure_pendulum_datetime(f["LastModified"]),
     "az": lambda f: ensure_pendulum_datetime(f["last_modified"]),
     "gcs": lambda f: ensure_pendulum_datetime(f["updated"]),
+    "https": lambda f: ensure_pendulum_datetime(f["Last-Modified"]),
```
rudolfix: please use your custom parsing function here instead modification of `ensure_pendulum_datetime`.

rudolfix: you can use `pendulum.parse` if you want

rudolfix (changes requested):
A few things to note:
1. to complete this PR you just need to use a proper parser in the place I indicate in comment. 
2. your `time.py` changes: in the meantime we rewrote this module, added very detailed tests for all the functions. this was necessary to properly handle naive timestamps (#2570 ) Also your changes modify our iso parser and allow for many other formats. I explained why we do not want to do that in my previous comment. My suggestion is to rollback this commit. If you want to simplify new `time.py` you can go for it. (but in a separate PR :))
3. `glob` issue: the case you describe looks inconvenient but this is how Python glob works as well. if your glob patter is absolute path, it will overwrite `root_dir` - in this case `bucket_url`. My take is that we do not fix there. We could also raise or warn if `file_glob` is absolute.
4 regarding tests. OK! let's go for simplified tests. just note that Python has built in http server that support listing. you could open it over the same folder we copy for the local glob tests.  GPT-5 generates pretty good fixture...

TheLazzziest:
Hey @rudolfix!

Sorry for the long delay, but I've finally pushed all the requested changes for the HTTP feature. Everything should be ready for your review now.

While you're checking those out, I wanted to quickly touch the documentation part. Should I include the docs updates in this PR?

My plan is to add a section for `Public CDNs` because I think it's a super simple and common use case. I was thinking of adding two things:

* Update the Introduction Page: Add a new tab for `Public CDN` (using the NYC taxi dataset as the example) in the "Load data with dlt" section. It's simple, requires no extra dependencies, and is a daily task for tons of devs.
* Add a `Public CDN` Page to Verified Sources: This page would explain the technical requirements needed to use this source type (like the `Last-Modified` constraint). However, I guess there won't be much to explain. As an alternative, we can also extend [cloud storage](https://dlthub.com/docs/tutorial/filesystem) section. 

Thus, these changes should improve the traction among the newcomers and help onboard people faster. Does that sound right, or do you have a better idea for where this info should go? Let me know so I can jump on the docs!

tests/load/sources/filesystem/test_filesystem_source.py
```diff
@@ -300,6 +300,32 @@ def _get_files() -> DltResource:
...
+)
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True),
+    ids=lambda x: x.name,
+)
+def test_http_filesystem(
+    public_http_server, bucket_url: str, destination_config: DestinationTestConfiguration
+):
+    public_resource = filesystem(bucket_url=bucket_url, file_glob="parquet/mlb_players.parquet")
```
rudolfix: does globbing work now? or this is still single file?

TheLazzziest: This is still a single file. Listing is [disabled](https://github.com/dlt-hub/dlt/pull/3029/files#diff-9e08e8c7769db7b58089fd7659d75ca96df40bb7b5d50299d3a30abd54da9ad6R156) .

TheLazzziest: > if we change anything that is user facing then we always improve documentation

That makes perfect sense,  the docs should follow!

My suggestion for the documentation came directly from talking with a developer who uses dlt in production. I noticed that using S3-compatible storage felt a bit different from hitting a direct HTTP endpoint. Even working with a standard REST API is not very close to this case. 

Basically, my thinking is that the developer experience for this new feature is subtle, and the case is not immediately obvious. I am not trying to push the idea, rather to share the reasoning behind the suggestion.

dlt/common/storages/configuration.py
```diff
@@ -137,7 +137,20 @@ def _make_file_url(scheme: str, fs_path: str, bucket_url: str) -> str:
     return uri
 
 
-MAKE_URI_DISPATCH = {"az": _make_az_url, "file": _make_file_url, "sftp": _make_sftp_url}
+def _make_http_url(schema: str, fs_path: str, bucket_url: str) -> str:
+    parsed_http_url = urlparse(fs_path)
+    if not parsed_http_url.scheme:
+        parsed_http_url.scheme = schema
```
rudolfix: `urlparse` returns a named tuple so this line will fail (also mypy complains). you can pass default schema directly into `urlparse` (https://docs.python.org/3/library/urllib.parse.html) or use `_replace` method.

TheLazzziest: Thanks for spotting that! You're totally right—I missed running the full `lint` rule before pushing, and Mypy definitely caught me. :pray: It took a significant portion of time to finish, so I was running the linters only on specific, smaller blocks to speed up development.

TheLazzziest: BTW, Have you considered using `pre-commit` on this matter ?

It could really help with validation during development enforcing the same level of quality on every commit, and preventing those simple errors from ever getting pushed.

PR #3036, opened 2025-08-26
Small dashboard fixes

sh-rp:
### Description
This PR has the following fixes and imprvovements:
* Do not fail if destination credentials are not available
* Do not order columns in schema browser by name, but rather keep them in the schema order
* Move dashboard tests out of github common workflows

### DIFF
--- a/dlt/helpers/dashboard/utils.py
+++ b/dlt/helpers/dashboard/utils.py
@@ -19,6 +19,8 @@ from dlt.common.pipeline import get_dlt_pipelines_dir
 from dlt.common.schema import Schema
 from dlt.common.storages import FileStorage
 from dlt.common.destination.client import DestinationClientConfiguration
+from dlt.common.exceptions import DltException
+from dlt.common.configuration.exceptions import ConfigFieldMissingException
 
 from dlt.helpers.dashboard import dlt_dashboard, ui_elements as ui
 from dlt.helpers.dashboard.config import DashboardConfiguration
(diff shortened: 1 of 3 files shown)


dlt/helpers/dashboard/utils.py
```diff
@@ -517,9 +518,13 @@ def get_local_data_path(pipeline: dlt.Pipeline) -> str:
     """Get the local data path of a pipeline"""
     if not pipeline.destination or not pipeline.default_schema_name:
         return None
-    config = pipeline._get_destination_clients()[0].config
-    if isinstance(config, WithLocalFiles):
-        return config.local_dir
+    try:
+        config = pipeline._get_destination_clients()[0].config
```
rudolfix, line 522: there's now a better way to get local_dir. stuff below is available in local state of the pipeline
```py
def _update_last_run_context(self) -> None:
        """
        Persist the directory context in local state.
        Safe-no-op if the pipeline is not active or if run_context cannot be resolved.
        """
        self.first_run = False
        self.last_run_context = {
            "settings_dir": os.path.abspath(self.run_context.settings_dir),
            "local_dir": os.path.abspath(self.run_context.local_dir),
            "run_dir": os.path.abspath(self.run_context.run_dir),
            "uri": self.run_context.uri,
        }
```
it is filled only after successful run of the pipeline. 

but we can keep the above., I plan to do a really good pipeline.attach which will restore the full run_context and inject all collected config and secrets via special config provider

rudolfix (approved):
LGTM!

PR #3048, opened 2025-09-02
Fix/67 normalizer child table behavior

sh-rp:
### Description

This PR fixes the behavior of the normalizer when encountering columns that have None or primitive values set that were previously detected as child tables. The new behavior is:

* If None is encountered, column is discarded
* If primitive value is encountered, it is converted to a list

Tests are pending.

### DIFF
--- a/dlt/common/normalizers/json/helpers.py
+++ b/dlt/common/normalizers/json/helpers.py
@@ -87,18 +87,6 @@ def is_nested_type(
     return data_type == "json"
 
 
-def is_column_nested_table_in_schema(
-    schema: Schema, table_name: str, column_name: str, path: Tuple[str, ...]
-) -> bool:
-    """Tells if the column points to a previously detected nested table in the schema."""
-    potential_nested_table_name = schema.naming.normalize_tables_path(
-        schema.naming.make_path(*path, column_name)
-    )
-    if schema.tables.get(potential_nested_table_name):
-        return True
-    return False
-
-
 def should_be_nested(schema: Schema, table_name: str) -> bool:
     """Tells if table should be nested or should be a root table. All tables created by the normalizer
     are nested, defined tables are checked.
--- a/dlt/common/normalizers/json/helpers.py
+++ b/dlt/common/normalizers/json/helpers.py
@@ -87,6 +87,18 @@ def is_nested_type(
     return data_type == "json"
 
 
+def is_column_nested_table_in_schema(
+    schema: Schema, table_name: str, column_name: str, path: Tuple[str, ...]
+) -> bool:
+    """Tells if the column points to a previously detected nested table in the schema."""
+    potential_nested_table_name = schema.naming.normalize_tables_path(
+        schema.naming.make_path(*path, column_name)
+    )
+    if schema.tables.get(potential_nested_table_name):
+        return True
+    return False
+
+
 def should_be_nested(schema: Schema, table_name: str) -> bool:
     """Tells if table should be nested or should be a root table. All tables created by the normalizer
     are nested, defined tables are checked.
--- a/dlt/common/normalizers/json/relational.py
+++ b/dlt/common/normalizers/json/relational.py
@@ -130,15 +130,16 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
         )
 
     def _flatten(
-        self, table: str, dict_row: DictStrAny, _r_lvl: int, ident_path: Tuple[str, ...] = None
+        self, table: str, dict_row: DictStrAny, _r_lvl: int, parent_path: Tuple[str, ...] = None
     ) -> Tuple[DictStrAny, Dict[Tuple[str, ...], Sequence[Any]]]:
         out_rec_row: DictStrAny = {}
         out_rec_list: Dict[Tuple[str, ...], Sequence[Any]] = {}
-        ident_path = ident_path or ()
+        parent_path = parent_path or ()
+        ident_path = parent_path + (table,)
 
         def norm_row_dicts(dict_row: StrAny, __r_lvl: int, path: Tuple[str, ...] = ()) -> None:
             # full ident path is the incoming ident path plus the nested path of this "internal" traversal
-            full_ident_path = path + ident_path if path else ident_path
+            full_ident_path = ident_path + path if path else ident_path
 
             for k, v in dict_row.items():
                 if k.strip():
@@ -159,7 +160,7 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
                         v = [v]
                         is_list_or_dict = True
 
-                # detect if the column is a nested table
+                # detect if the column is a nested type
                 nested_name = norm_k if path == () else self._shorten_fragments(*path, norm_k)
 
                 # for lists and dicts we must check if type is possibly nested
@@ -284,7 +285,7 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
     ) -> TNormalizedRowIterator:
         table = self._shorten_fragments(*parent_path, *ident_path)
         # flatten current row and extract all lists to recur into
-        flattened_row, lists = self._flatten(table, dict_row, _r_lvl, ident_path)
+        flattened_row, lists = self._flatten(table, dict_row, _r_lvl, parent_path)
         # always extend row
         DataItemNormalizer._extend_row(extend, flattened_row)
 
--- a/dlt/common/normalizers/json/relational.py
+++ b/dlt/common/normalizers/json/relational.py
@@ -125,44 +125,23 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
         self._should_be_nested = lru_cache(maxsize=None)(
             partial(normalize_helpers.should_be_nested, self.schema)
         )
-        self._is_column_nested_table_in_schema = lru_cache(maxsize=None)(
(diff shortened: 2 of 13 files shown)


dlt/common/normalizers/json/relational.py
```diff
@@ -125,23 +125,46 @@ def _reset(self) -> None:
...
             for k, v in dict_row.items():
                 if k.strip():
                     norm_k = self._normalize_identifier(k)
                 else:
                     # for empty keys in the data use _
                     norm_k = self.EMPTY_KEY_IDENTIFIER
+
+                is_list_or_dict = isinstance(v, (dict, list))
+                if not is_list_or_dict and self._is_column_nested_table_in_schema(
+                    table, norm_k, full_ident_path
```
sh-rp: should we put norm_k or k here? I am not sure.

dlt/common/normalizers/json/helpers.py
```diff
@@ -87,6 +87,18 @@ def is_nested_type(
...
 
 
+def is_column_nested_table_in_schema(
+    schema: Schema, table_name: str, column_name: str, path: Tuple[str, ...]
+) -> bool:
+    """Tells if the column points to a previously detected nested table in the schema."""
+    potential_nested_table_name = schema.naming.normalize_tables_path(
+        schema.naming.make_path(*path, column_name)
+    )
+    if schema.tables.get(potential_nested_table_name):
```
sh-rp: should this also verify that this table really is the child of a certain table, or do we just assume that if the name matches, it is the right child table?

tests/common/normalizers/test_json_relational.py
```diff
@@ -945,3 +951,98 @@ def add_dlt_root_id_propagation(norm: RelationalNormalizer) -> None:
...
+    assert rows[1][0] == ("t1__nested", "t1")
+    assert rows[1][1].items() > {"other": "nested_blah"}.items()
+
+    assert rows[2][0] == ("t1__nested__subnested", "t1__nested")
+    assert rows[2][1].items() > {"value": "singleton"}.items()
+
+    assert rows[3][0] == ("t1__nested", "t1")
+    assert rows[3][1].items() > {"other": "blah_the_second_nested_one"}.items()
+
+    assert rows[4][0] == ("t1__nested__subnested__subsubnested", "t1__nested")
```
sh-rp: this parent path is not right, I still have to investigate this.

dlt/common/normalizers/json/relational.py
```diff
@@ -125,23 +125,44 @@ def _reset(self) -> None:
         self._should_be_nested = lru_cache(maxsize=None)(
             partial(normalize_helpers.should_be_nested, self.schema)
         )
+        self._is_column_nested_table_in_schema = lru_cache(maxsize=None)(
```
rudolfix: I'm not sure caching works. what if nested table is created during normalization. does it always happen first?

dlt/common/normalizers/json/relational.py
```diff
@@ -125,23 +125,44 @@ def _reset(self) -> None:
...
 
     def _flatten(
-        self, table: str, dict_row: DictStrAny, _r_lvl: int
+        self, table: str, dict_row: DictStrAny, _r_lvl: int, full_path: Tuple[str, ...] = None
     ) -> Tuple[DictStrAny, Dict[Tuple[str, ...], Sequence[Any]]]:
         out_rec_row: DictStrAny = {}
         out_rec_list: Dict[Tuple[str, ...], Sequence[Any]] = {}
 
         def norm_row_dicts(dict_row: StrAny, __r_lvl: int, path: Tuple[str, ...] = ()) -> None:
+            # full nested path is the incoming ident path plus the nested path of this "internal" traversal
```
rudolfix: this is not a right place IMO. the fix should happen in `coerce_row` in `item_normalizer`. here you are coercing types. also this code runs for every single item in the normalizer (maybe it is time to rewrite it in rust...) to cover a super rate edge case.

in coerce_row you need to act only if you infer a new column. if you infer a new column fron a None, check if parent exists, if it does then discard.

the question is what we do with other simple values... if we assume that `str` is not coercable to `list[str]` then we just let this column to be created (this is what happens now).
same thing if we get a list of values but column exists for simple type.  then we let normalizer to create table

anuunchin:
@rudolfix hopefully this is closer to what you had in mind 👀 

A special logic was added to `def _coerce_null_value`, which checks the immediate child tables. 
If the child table with the name `<parent_tbl_name>__<col_name>` exists, then we don't infer column schema.

dlt/normalize/items_normalizers.py
```diff
@@ -595,6 +596,16 @@ def _coerce_null_value(
     ) -> Optional[TColumnSchema]:
         """Raises when column is explicitly not nullable or creates unbounded column"""
         existing_column = table_columns.get(col_name)
+        # If it exists as a direct child table, don't infer
+        if not existing_column:
+            direct_children = get_nested_tables(
+                self.schema._schema_tables, table_name, max_nesting=1, include_self=False
+            )
+            direct_child = next(
```
rudolfix: here's one unfixable problem: table names are created out of the not normalized identifier and that information is destroyed in the relational normalizer. still what you do here is probably the best. just don't use `endswith`. use naming.break_path() to get the last path component and compare it with col_name

also add a test with really long column names. see `naming.max_length`. I think we use postgres caps so it will be 64. you'll see that your detection does not work. (tell me if it does)

rudolfix: one more thing: this is really complicated, slow operation and this place may be called pretty frequently. you can extract this check to a cached function. see how relational.py does it
```
self._shorten_fragments = lru_cache(maxsize=None)(
            partial(normalize_helpers.shorten_fragments, self.naming)
        )
```
it looks to me that caching is safe. if table will be created during normalization, we'll have a column already in the root table - and we'll remove that column in the other PR you do

dlt/common/schema/utils.py
```diff
@@ -900,19 +900,41 @@ def get_root_table(tables: TSchemaTables, table_name: str) -> TTableSchema:
     return table
 
 
-def get_nested_tables(tables: TSchemaTables, table_name: str) -> List[TTableSchema]:
+def get_nested_tables(
```
rudolfix, line 903: do we have unit tests for that function? you should add unit tests for new params.

dlt/common/schema/utils.py
```diff
@@ -900,19 +900,41 @@ def get_root_table(tables: TSchemaTables, table_name: str) -> TTableSchema:
...
+    table_name: str,
+    max_nesting: Optional[int] = None,
+    include_self: Optional[bool] = True,
+) -> List[TTableSchema]:
     """Get nested tables for table name and return a list of tables ordered by ancestry so the nested tables are always after their parents
 
     Note that this function follows only NESTED TABLE reference typically expressed on _dlt_parent_id (PARENT_KEY) to _dlt_id (ROW_KEY).
+
+    Args:
+        max_nesting: If specified, limits the depth of nesting. 0 = only the root table, 1 = root + direct children, etc.
```
rudolfix: what about other args?

dlt/normalize/items_normalizers.py
anuunchin: `here's one unfixable problem: table names are created out of the not normalized identifier and that information is destroyed in the relational normalizer. ` -> maybe using the full ident_paths from the relational normalizer a solution? 👀

dlt/common/normalizers/json/relational.py
```diff
@@ -280,7 +280,7 @@ def _normalize_row(
 
         # yield parent table first
         should_descend = yield (
-            (table, self._shorten_fragments(*parent_path)),
+            (table, self._shorten_fragments(*parent_path), () if is_root else ident_path),
```
rudolfix: OK I always wanted to do that... if we are returning paths then just return a single tuple. I think you can just yield parent path + ident path (single tuple) and do the shortening at the other end. the biggest amount of work is to fix the tests... that's why I didn't do it

rudolfix: w8. so it it looks `ident_path` is a full path. right? then it is the only thing you need to yield.

dlt/normalize/items_normalizers.py
```diff
@@ -590,11 +602,37 @@ def _infer_column(
...
+
     def _coerce_null_value(
         self, table_columns: TTableSchemaColumns, table_name: str, col_name: str
     ) -> Optional[TColumnSchema]:
         """Raises when column is explicitly not nullable or creates unbounded column"""
         existing_column = table_columns.get(col_name)
+        # If it exists as a direct child table, don't infer
+        if not existing_column:
+            # Use cached table existence check to avoid expensive repeated lookups
+            full_ident_path = self._full_ident_path_tracker.get(table_name)
```
rudolfix, line 631: LGTM

rudolfix (changes requested):
we could merge this. but we can simplify `relational` even more. it makes sense. it is very cpu intensive. tbh. whole normalizer code should pass original non normalized tuples along. and we should store them in schema so we do not lose information. this is clearly 2.0 feature but we are getting there

dlt/common/normalizers/json/relational.py
anuunchin: agh, that's true

anuunchin: went with parent_path + ident_path because ident_path is not the full path

rudolfix (changes requested):
this is really good! but IMO we need one more end to end test ie. in `tests/pipeline/test_schema_contracts.py` where we reproduce original user problem with contract triggering on None

rudolfix (approved):
LGTM!

PR #3052, opened 2025-09-02
add up to date check for uv lockfile as first lint step

sh-rp:
### Description
We need to ensure that the lockfile is up to date when dependencies change. This PR adds a linter step to do this.


zilto (dismissed):
Awesome!

Issue #2564, opened 2025-04-25
user friendly `PipelineStepFailed` message


rudolfix:
**Background**
Any exception when pipeline executes is wrapped in `PipelineStepFailed`. It contains a lot of useful information, also depending on the step (ie. load_id, resource name that was the cause of exception, job name that failed to load etc.)
However current error message does not show that and user must dig through extremely deep stack to get any informaiton.

Also such trace is not added to pipeline trace so post-mortem inspection is not possible,

Our goal here is to
1. Improve error message by providing relevant details (step name, load_id, resource name, job name etc.)
2. Finding relevant exceptions in stack trace and showing them in error message. In particular we should look at rest_api, sql_database and filesystem common exceptions
3. Intelligently showing content of failed data items and `requests` reponse content (like pytest does maybe)
4. We can reuse existing stack trace crawlers (in common utils) to create traces to investigate and save for post mortem analysis (we could also save data item or request reponse content)
5. Depending on the pipeline step and relevant exceptions, propose follow up actions: ie. how to check pipeline trace or package with load id or a job

djudjuu:
keep last 10 traces, useful for marimo app

PR #3055, opened 2025-09-03
Dashboard updates and fixes

sh-rp:
### Description
 This PR fixes issues of the new dashboard:
 * Databrowser was broken for child-tables
 * Moves a lot of code into utils
 * Adds a bunch of example pipelines with different errors and problems and runs most of the utils with all of them
 * Fixes a ton of edge cases and small bugs
 * Improves error messaging for when destination is not available.

### DIFF
--- a/dlt/common/utils.py
+++ b/dlt/common/utils.py
@@ -326,6 +326,11 @@ def map_nested_values_in_place(
     return _nested
 
 
+# keep old name for backwards compatibility
+# dlt+ needs to be updated
+map_nested_in_place = map_nested_values_in_place
+
+
 def map_nested_keys_in_place(
     func: AnyFun,
     _nested: TAny,


dlt/helpers/dashboard/dlt_dashboard.py
```diff
@@ -617,16 +624,17 @@ def section_trace(
...
-                mo.accordion(
-                    {
-                        strings.trace_show_raw_trace_text: mo.ui.code_editor(
-                            json.dumps(dlt_trace, pretty=True),
-                            language="json",
-                        )
-                    }
-                )
-            )
+            # TODO: trace is not json serializable. I'd store it as such if that was easy
```
sh-rp: I'm sending this to the platform though, I'll have to check what I do there to make it work, maybe as_dict or something.

tests/helpers/dashboard/example_pipelines.py
```diff
@@ -0,0 +1,177 @@
+#
```
sh-rp, line 1: @rudolfix here I am creating a few example pipelines from the current codebase (I don't want to use serialized state or schemas so we always test with the current dlt). Maybe you can think of one or two more edge case pipelines or things we can do here to test all the utils.

sh-rp:
@rudolfix could you check `def open_local_folder(folder: str) -> None:` on your windows machine and see why it not works? Your feedback suggests the branching in the if statement is not right for windows, I took this from somewhere :)

dlt/helpers/dashboard/dlt_dashboard.py
sh-rp: I have now some code for sanitizing this which is tested for all the example pipelines and it is displayed as an "object" view in the dashboard now which is quite cool too.

dlt/helpers/dashboard/utils.py
```diff
@@ -39,6 +45,20 @@
 #
 
 
+def _exception_to_string(exception: Exception) -> str:
+    """Convert an exception to a string"""
```
sh-rp, line 51: which other exceptions can be encountered that we can convert to a better version on the dashboard UI when connecting to a destination or instantiating the destination client?

rudolfix: maybe you could look at `get_exception_traces` (trace module) and `get_exception_trace_chain` and use them?
I want to use them for exception trace enrichement. for example it extracts
```
# extract special attrs
            if k in ["load_id", "pipeline_name", "source_name", "resource_name", "job_id"]:
                trace[k] = v  # type: ignore[literal-required]
```

One that is missing is `PipelineStepFailed` which must be inspected for direct __cause__ (or two)

tldr;> we are going to work on clear exception traces (ie. remove all this wrapped stuff, surface user code) also #2564

dlt/helpers/dashboard/utils.py
```diff
@@ -301,6 +325,22 @@ def create_column_list(
...
 
 
+def get_source_and_resouce_state_for_table(
+    table: TTableSchema, pipeline: dlt.Pipeline, schema_name: str
+) -> Tuple[str, str, str]:
+    if "resource" not in table:
+        return None, None, None
+
+    # get source and resource state for correct resources from pipeline
+    _converted_state = json.loads(json.dumps(pipeline.state))
```
rudolfix: this is wrong. there are functions that will access source/resource state and you do not need to hardcode any key names.
look in extract/state.py

dlt/helpers/dashboard/utils.py
```diff
@@ -301,6 +325,22 @@ def create_column_list(
...
+    _source_state = _converted_state.get("sources", {}).get(schema_name, {})
+
+    _resources_state = _source_state.pop("resources", {})
+    _resource_state = _resources_state.get(table["resource"], {})
+
+    return table["resource"], yaml.safe_dump(_source_state), yaml.safe_dump(_resource_state)
```
rudolfix: why are we using yml? there's this Python object viewer I think?

dlt/helpers/dashboard/utils.py
rudolfix: also why do you destroy Python types in state. state is typed. may contain dates. why not display them?

dlt/helpers/dashboard/utils.py
```diff
@@ -311,17 +351,49 @@ def clear_query_cache(pipeline: dlt.Pipeline) -> None:
...
     # get_row_counts.cache_clear()
 
 
-@functools.cache
-def get_query_result(pipeline: dlt.Pipeline, query: str) -> pd.DataFrame:
+def get_default_query_for_table(
+    pipeline: dlt.Pipeline, schema_name: str, table_name: str, limit: bool
+) -> Tuple[str, str, str]:
+    try:
+        _dataset = cast(ReadableDBAPIDataset, pipeline.dataset(schema=schema_name))
```
rudolfix: we do not have `ReadableDBAPIDataset`. There's just `Dataset`. this is deprecated

dlt/helpers/dashboard/utils.py
```diff
@@ -637,7 +747,9 @@ def _humanize_datetime_values(c: DashboardConfiguration, d: Dict[str, Any]) -> D
     def _humanize_datetime(dt: Union[str, int]) -> str:
```
rudolfix, line 747: we have `humanize` module in the dependencies. maybe you could use it

dlt/helpers/dashboard/utils.py
```diff
@@ -549,6 +631,34 @@ def build_pipeline_link_list(
     return link_list
 
 
+def get_state_as_yaml_string(pipeline: dlt.Pipeline) -> str:
+    """Get the state of a pipeline as a yaml string"""
+    return yaml.safe_dump(json.loads(json.dumps(pipeline.state)))
```
rudolfix: I think you should not remove data types here. Why are you using yaml for display? this object tree is way better

dlt/helpers/dashboard/utils.py
```diff
@@ -549,6 +631,34 @@ def build_pipeline_link_list(
...
+
+
+def _remove_non_primitives(obj: Any) -> Any:
```
rudolfix: there's function `map_nested_in_place` that does recursive map but for values. maybe we could have recursive map for keys? and a mapping function should convert non-primitive keys to strings.

tuples should have a nice str representation

tests/helpers/dashboard/example_pipelines.py
rudolfix: this is really a good start and I can write down many more cases. for each case we should also describe behavior of dashboard app and affected widgets.

ie. a pipeline that fails in on particular job retry:
- what should be displayed on top of the app? (exception, warning that pipeline has pending package, show failed job, show retry message)
- what pipeline loads widget should show?
- what pipeline trace widgets should show?

possibly each widgets in the dashboard has a set of sanity checks that we could query to display on top of the page. So for example exception is coming from trace widget. failed job info from loads widget etc. let's think about it

tests/helpers/dashboard/example_pipelines.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def create_success_pipeline_duckdb(pipelines_dir: str = None, db_location: str = None):
+    """Create a test pipeline with in memory duckdb destination, properties see `run_success_pipeline`"""
+    import duckdb
+
+    pipeline = dlt.pipeline(
+        pipeline_name=SUCCESS_PIPELINE_DUCKDB,
+        pipelines_dir=pipelines_dir,
+        destination=dlt.destinations.duckdb(
```
rudolfix, line 99: we need `dummy+duckdb" destination. dummy client can generate very weird edge cases but it does not load data. maybe it could derive from duckdb so we could generate edge cases ie. with partially loaded data?

dlt/helpers/dashboard/utils.py
sh-rp: This block is only used for enriching the messages to the user if the connection to the destination fails, not generic dlt exceptions that happen during other operations.

dlt/helpers/dashboard/utils.py
sh-rp: I am using the functions in extract/state now, but since there is no active source, I still need to provide the name.

I have also changed all places where state is displayed (state section and the state are in the data explorer) to use mo.json instead of yaml notation.

dlt/helpers/dashboard/utils.py
sh-rp: done

dlt/helpers/dashboard/utils.py
sh-rp: will update this after https://github.com/dlt-hub/dlt/pull/3059 is merged.

dlt/helpers/dashboard/utils.py
sh-rp: yes. removed this in favor of the json object display.

dlt/helpers/dashboard/utils.py
sh-rp: done

tests/helpers/dashboard/example_pipelines.py
sh-rp: sounds good, I will create a ticket with follow up tasks and states.

dlt/helpers/dashboard/utils.py
sh-rp: I have updated this. Tuples are now rendered as their repr, which looks like this ('table_1', 'table_2') which is pretty clear imho.

dlt/common/utils.py
```diff
@@ -294,9 +294,16 @@ def clone_dict_nested(src: TDict) -> TDict:
     return update_dict_nested({}, src, copy_src_dicts=True)  # type: ignore[return-value]
 
 
-def map_nested_in_place(func: AnyFun, _nested: TAny, *args: Any, **kwargs: Any) -> TAny:
+def map_nested_in_place(
```
rudolfix: this looks good but this function is used in many CPU intensive modules. and I see you are cloning dicts and that makes this function heavy. way better to have separate key modify function.

dlt/extract/state.py
```diff
@@ -30,7 +30,7 @@ def _sources_state(pipeline_state_: Optional[TPipelineState] = None, /) -> DictS
     return sources_state_
 
 
-def source_state() -> DictStrAny:
+def source_state(source_state_key: Optional[str] = None) -> DictStrAny:
```
rudolfix, line 33: 👍

tests/helpers/dashboard/conftest.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+import tempfile
+
+from tests.helpers.dashboard.example_pipelines import (
```
rudolfix, line 4: do we have conftest  in helpers? we should at least have the same fixtures imported as in `tests/pipeline`

rudolfix:
> @rudolfix could you check `def open_local_folder(folder: str) -> None:` on your windows machine and see why it not works? Your feedback suggests the branching in the if statement is not right for windows, I took this from somewhere :)

I still see this
```
on_click handler for button (button()) raised an Exception:
  File "/home/rudolfix/src/dlt/.venv/lib/python3.11/site-packages/marimo/_plugins/ui/_impl/input.py", line 1257, in _convert_value
    return self._on_click(self._value)  # type: ignore[no-untyped-call]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/marimo_2437/__marimo__cell_Hbol_.py", line 49, in <lambda>
    on_click=lambda _: utils.open_local_folder(dlt_pipeline.working_dir),
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rudolfix/src/dlt/dlt/helpers/dashboard/utils.py", line 611, in open_local_folder
    subprocess.run(["xdg-open", folder], check=True)
  File "/usr/lib/python3.11/subprocess.py", line 546, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/subprocess.py", line 1022, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib/python3.11/subprocess.py", line 1899, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'xdg-open'
```

rudolfix (approved):
LGTM!

PR #3056, opened 2025-09-03
Feat: dataset access telemetry

### LINKED ISSUE #3045: add telemetry for anonymous dataset
**Background**
We are pretty good at tracking pipeline runs. Each step collects run metrics and basic pipeline info.
https://github.com/dlt-hub/dlt/blob/devel/dlt/pipeline/track.py#L96

Now we need to start tracking data usage which means tracking dataset / relations. Here I see following options:

1. track dataset instantiation via factory function (`dlt.dataset()`). This will also cover `pipeline.dataset()`
2. track first access to destination data (ie. `def _get_destination_client(self, schema: Schema) -> JobClientBase:`)
3. track relation materialization (ie `query` or similar)

My preference is go to for (2) because:
- it minimizes number of events compared to tracking destinations or all query materializations
- it sends telemetry when another external connection is made so it is not visible
- if connection to destination is made it typically means some data got materialized so this event is meaningful

**Requirements**
This is for option (2):
1. * [ ] instrument the right method, look at `with_telemetry` but you'll need your own handling (see `def track(event_category: TEventCategory, event_name: str, properties: DictStrAny) -> None:`)
2. * [ ] we need new event category! `data_access` seems right. event name  `connect`
3. * [ ] data collection:
"success": did connection succeeded
        "destination_name": pipeline.destination.destination_name if pipeline.destination else None,
        "destination_type": pipeline.destination.destination_type if pipeline.destination else None,
        "dataset_name_hash": digest128(pipeline.dataset_name) if pipeline.dataset_name else None,
        "default_schema_name": (
            digest128(schema_name)
        ),

4. * [ ] obviously we need tests for it and docs update in https://dlthub.com/docs/reference/telemetry

anuunchin:
This PR adds telemetry for anonymous data access events when destination clients are requested. As of now, it involves the `def with_dataset_access_telemetry()` decorator that is used with `def _get_destination_clients()`, mimicking the existing `def with_telemetry()`. A test is added in `test_telemetry.py`.

Resolves #3045


dlt/pipeline/pipeline.py
```diff
@@ -1319,6 +1320,7 @@ def _extract_source(
 
         return load_id
 
+    @with_dataset_access_telemetry()
```
anuunchin: This might appear too invasive, maybe there should be another function that makes it appear less so 🤔

dlt/common/runtime/telemetry.py
```diff
@@ -112,6 +114,65 @@ def _track(success: bool) -> None:
...
+
+    return decorator
+
+
+def with_dataset_access_telemetry() -> Callable[[TFun], TFun]:
+    """Track dataset access but only once per pipeline instance"""
+
+    def decorator(f: TFun) -> TFun:
+        def _wrap(*f_args: Any, **f_kwargs: Any) -> Any:
+            pipeline = f_args[0] if f_args else None
```
rudolfix: not 100% sure but you could take `locals()` here and there you'll find `self` (pipeline instance) and `schema`. `self` must be present. then your code is less hacky

dlt/common/runtime/telemetry.py
```diff
@@ -112,6 +114,65 @@ def _track(success: bool) -> None:
...
+            schema: Union[Schema, str, None] = f_kwargs.get("schema") or (
+                f_args[1] if len(f_args) > 1 else None
+            )
+
+            def _track_dataset_access_once(success: bool) -> None:
+                if pipeline and not getattr(pipeline, "_dataset_access_tracked", False):
+                    with contextlib.suppress(Exception):
```
rudolfix: OK I think we do not really need decorator at all. this is reproducing too much of logic that is in `def dataset(self, schema: Union[Schema, str, None] = None)` method.

my take: just send event after:
```py
dlt.dataset(
            self._destination,
            self.dataset_name,
            schema=schema,
        )
```
decorator is good if we plan to decorate many functions with it. right now it is just one function and it won't generalize to anything else.

dlt/common/runtime/telemetry.py
```diff
@@ -112,6 +114,65 @@ def _track(success: bool) -> None:
...
+                                if pipeline.destination
+                                else None
+                            ),
+                            "dataset_name_hash": (
+                                digest128(pipeline.dataset_name) if pipeline.dataset_name else None
+                            ),
+                            "default_schema_name_hash": digest128(schema_name),
+                        }
+                        # resolve runtime config and init telemetry
+                        if not _TELEMETRY_STARTED:
```
rudolfix: you do not need to do it. this already happened just before Pipeline object is instantiated

dlt/common/runtime/telemetry.py
```diff
@@ -112,6 +114,65 @@ def _track(success: bool) -> None:
...
+                                pipeline.destination.destination_name
+                                if pipeline.destination
+                                else None
+                            ),
+                            "destination_type": (
+                                pipeline.destination.destination_type
+                                if pipeline.destination
+                                else None
+                            ),
+                            "dataset_name_hash": (
```
rudolfix: also hash the pipeline name

dlt/common/runtime/telemetry.py
rudolfix: also: extract tracking itself to `track.py` (creating and sending props). look at `on_end_trace_step`: we have code duplication

rudolfix (changes requested):

dlt/common/runtime/telemetry.py
anuunchin: went with using bound arguments like in `with_telemetry`, otherwise `locals()` gives the `f_args` and `f_kwargs` which doesn't help in getting rid of the hacky stuff 👀 - unless I understood the comment incorrectly

dlt/common/runtime/telemetry.py
anuunchin: owww

dlt/common/runtime/telemetry.py
```diff
@@ -115,3 +119,22 @@ def _track(success: bool) -> None:
...
+        schema_name = schema.name
+    elif isinstance(schema, str):
+        schema_name = schema
+
+    props = {
+        "success": success,
+        **_build_base_props(pipeline=pipeline),
+        # The schema may differ from the default pipeline schema
+        "requested_schema_name_hash": digest128(schema_name) if schema_name else None,
+    }
```
anuunchin: @rudolfix  just realized, we should probably also include what schema name is actually being passed to `pipeline.dataset()`

dlt/common/runtime/telemetry.py
```diff
@@ -2,18 +2,22 @@
...
 from dlt.common.configuration import resolve_configuration
 from dlt.common.runtime.anon_tracker import (
     TEventCategory,
     init_anon_tracker,
     disable_anon_tracker,
     track,
 )
+from dlt.common.schema import Schema
+from dlt.common.pipeline import SupportsPipeline
+from dlt.pipeline.track import _build_base_props
```
rudolfix: pls do not import into common from downstream modules. if you have to do that, it means something is wrong with code org. Also modules should know as little as possible about other modules. Why telemetry should be aware of Schema and Pipeline. this module just sends dictionaries to an endpoint... (and I still thing it knows too much)

dlt/common/runtime/anon_tracker.py
```diff
@@ -16,7 +16,7 @@
 
 from dlt.version import __version__
 
-TEventCategory = Literal["pipeline", "command", "helper"]
+TEventCategory = Literal["pipeline", "command", "helper", "data_access"]
```
rudolfix: OK so we moved the event to `pipeline` so we do not need additional category. we just need additional name. ie. `access_dataset`

dlt/common/runtime/telemetry.py
```diff
@@ -115,3 +119,22 @@ def _track(success: bool) -> None:
         return _wrap  # type: ignore
 
     return decorator
+
+
+def with_dataset_access_telemetry(
```
rudolfix: this is good name for a decorator and this is a regular function. so the name should be an action ie. track_pipeline_create_dataset or smth

as mentioned above: this function fits track module better

dlt/common/runtime/telemetry.py
```diff
@@ -115,3 +119,22 @@ def _track(success: bool) -> None:
...
+    pipeline: SupportsPipeline, schema: Union[Schema, str, None], success: bool
```
rudolfix: you should just take schema name as parameter

dlt/common/runtime/telemetry.py
```diff
@@ -115,3 +119,22 @@ def _track(success: bool) -> None:
...
+    elif isinstance(schema, str):
+        schema_name = schema
+
+    props = {
+        "success": success,
+        **_build_base_props(pipeline=pipeline),
+        # The schema may differ from the default pipeline schema
+        "requested_schema_name_hash": digest128(schema_name) if schema_name else None,
+    }
+    track("data_access", "connect", props)
```
rudolfix: yeah let's do "pipeline", "access_dataset"

rudolfix (changes requested):
almost there

dlt/common/runtime/telemetry.py
anuunchin: `Also modules should know as little as possible about other modules. Why telemetry should be aware of Schema and Pipeline. this module just sends dictionaries to an endpoint...`  - this is a conceptual understanding point I was definitely missing - duly noted 🫡

dlt/common/runtime/telemetry.py
anuunchin: aaha, I get it now why `track.py` was specifically mentioned earlier ✅

dlt/pipeline/track.py
```diff
@@ -78,6 +78,35 @@ def on_start_trace_step(
...
+        "destination_type": pipeline.destination.destination_type if pipeline.destination else None,
+        "pipeline_name_hash": digest128(pipeline.pipeline_name),
+        "dataset_name_hash": digest128(pipeline.dataset_name) if pipeline.dataset_name else None,
+        "default_schema_name_hash": (
+            digest128(pipeline.default_schema_name) if pipeline.default_schema_name else None
+        ),
+    }
+
+
+def on_first_dataset_access(
```
anuunchin, line 95: @rudolfix  maybe naming this function like the other functions in this file is a good approach? 👀

rudolfix (approved):
LGTM!

PR #3059, opened 2025-09-03
feat(dataset): simplify public interface for `dlt.Dataset` and `dlt.Relation`

zilto:
This PR aims to simplify the classes `dlt.Dataset` and `dlt.Relation`. It should introduce no behavior changes and minimal backwards incompatibility. 

I tried to provide a clean commit history to facilitate PR review

Changes:
- remove `SupportsRelation` and `SupportsDataset` protocols
- remove `# type: ignore` and `cast()` associated with protocol use
- moved some private utility methods to utility function (dependency inversion)
- some arguments are moved to keywords only
- updated impacted tests and docs snippets


tests/load/test_read_interfaces.py
```diff
@@ -571,25 +571,28 @@ def test_sql_queries(populated_pipeline: Pipeline) -> None:
...
     # raw query has no aliases
     assert (
-        join_relationship._sqlglot_expression.sql("duckdb").replace(dataset_name, "dataset_name")
+        join_relationship.sqlglot_expression.sql("duckdb").replace(dataset_name, "dataset_name")
         == "SELECT i.id, di.double_id FROM dataset_name.items AS i JOIN dataset_name.double_items"
         " AS di ON (i.id = di.id) WHERE i.id < 20 ORDER BY i.id ASC"
     )
 
+    # TODO move these tests to the `dlt.destination.queries::normalize_query()`
+    # TODO modify `dlt.dataset.lineage::compute_columns_schema()` to return the normalized query instead of a tuple
```
sh-rp, line 574: please don't forget to move or update these tests

dlt/dataset/relation.py
```diff
@@ -60,49 +56,36 @@ def __init__(
...
-            self._sqlglot_expression = maybe_parse(
-                query,
-                dialect=query_dialect or self.query_dialect(),
-            )
-        else:
-            self._sqlglot_expression = build_select_expr(
-                table_name=table_name,
-                selected_columns=list(self._dataset.schema.get_table_columns(table_name).keys()),
-            )
+        self._schema: Optional[TTableSchemaColumns] = None
```
sh-rp, line 87: Imho this should be a TTableSchema, there are scenarios where we would also use table level hints, for example if the relation is just a Select * from {origin_table}. This is not implemented yet, but will be useful for synching datasets fully and incrementally.

dlt/dataset/dataset.py
```diff
@@ -87,126 +114,114 @@ def sqlglot_schema(self) -> SQLGlotSchema:
...
 
     @overload
-    def table(self, table_name: str, table_type: Literal["relation"]) -> SupportsRelation: ...
+    def table(self, table_name: str, table_type: Literal["relation"]) -> dlt.Relation: ...
 
     @overload
-    def table(self, table_name: str, table_type: Literal["ibis"]) -> IbisTable: ...
+    def table(self, table_name: str, table_type: Literal["ibis"]) -> ir.Table: ...
+
+    # TODO remove `table_type` argument. Instead, `dlt.Relation()` should have `.to_ibis()` method
```
sh-rp, line 218: I think I agree with this.

dlt/dataset/relation.py
zilto: Good point. Will work that in.

dlt/dataset/dataset.py
zilto: will do in a follow-up PR because it's non-trivial to correctly bind tables

dlt/common/libs/utils.py
```diff
@@ -38,3 +40,26 @@ def load_open_tables(
             schema_open_tables = [t for t in schema_open_tables if t in tables]
 
         return {name: client.load_open_table(table_format, name) for name in schema_open_tables}
+
+
+def is_instance_lib(obj: Any, *, class_ref: str) -> bool:
```
sh-rp, line 45: this is cool! maybe a few tests would be good as we have them for most helpers. Maybe in a follow-up ticket.

dlt/dataset/dataset.py
```diff
@@ -39,46 +36,76 @@ def __init__(
         dataset_name: str,
         schema: Union[dlt.Schema, str, None] = None,
     ) -> None:
-        # provided properties
-        self._destination = Destination.from_reference(destination)
-        self._provided_schema = schema
+        self._destination_reference = destination
```
sh-rp, line 39: this is never used, are there any plans for it?

dlt/dataset/dataset.py
```diff
@@ -283,24 +306,32 @@ def __getitem__(self, table_name: str) -> SupportsRelation:
         except ValueError as exc:
             raise KeyError(table_name, str(exc))
 
+    # TODO this creates all sorts of bugs in dynamic scenarios where `getattr()` is used
```
sh-rp, line 308: What is an example of such a bug? ibis does the same and it makes for a quite nice notation imho. I don't think we want to get rid of this.

dlt/dataset/dataset.py
```diff
@@ -87,126 +114,114 @@ def sqlglot_schema(self) -> SQLGlotSchema:
...
+        this value is the "output dialect" in the context of SQLGlot transpilation.
+        """
         return self.sql_client.capabilities.sqlglot_dialect
 
     @property
     def dataset_name(self) -> str:
+        """Name of the dataset"""
         return self._dataset_name
 
+    # TODO why do we need `_opened_sql_client` and `_sql_client`? One seems used by
```
sh-rp, line 128: opened_sql_client is used by the dataset context manager to be able to run multiple queries without closing the connection. you can check the dataset read tests to see how this is done, it is also explained in the docs. You can also read the cursor method in the Relation to see how these clients are treated differently.

dlt/dataset/dataset.py
```diff
@@ -362,3 +391,66 @@ def dataset(
...
+        destination_dataset_name=dataset.dataset_name,
+    )[0]
+
+
+def get_dataset_sql_client(dataset: dlt.Dataset) -> SqlClientBase[Any]:
+    client = get_dataset_destination_client(dataset)
+    if isinstance(client, WithSqlClient):
+        return client.sql_client
+    else:
+        raise Exception(
```
sh-rp, line 409: We should use `SqlClientNotAvailable` here.

sh-rp: Or maybe a variation, since that exception is specific to pipelines.

dlt/dataset/relation.py
```diff
@@ -145,35 +128,78 @@ def iter_arrow(self, *args: Any, **kwargs: Any) -> Any:
...
     @property
     def columns_schema(self) -> TTableSchemaColumns:
-        if self._columns_schema is None:
-            self._columns_schema, self.__qualified_query = self._compute_columns_schema()
-        return self._columns_schema
+        """dlt columns schema. Convenience method for `dlt.schema["columns"]`"""
+        return self.schema.get("columns", {})
 
     @property
     def schema(self) -> TTableSchema:
```
sh-rp, line 137: maybe this should be called `.table_schema` actually. wdyt? but we can also keep schema.

dlt/dataset/relation.py
```diff
@@ -145,35 +128,78 @@ def iter_arrow(self, *args: Any, **kwargs: Any) -> Any:
     def iter_fetch(self, *args: Any, **kwargs: Any) -> Any:
         return self._wrap_iter("iter_fetch")(*args, **kwargs)
 
+    # TODO did `.columns_schema` previously use (infer_sqlglot_schema=False, allow_partial=False)
```
sh-rp: IIRC this was because we assumed that our transformation lineage might not work for certain columns. Did you test this new code with running the transformations tests in dlt+ against it? If this passes, we can keep it this way.

dlt/dataset/relation.py
```diff
@@ -182,17 +208,16 @@ def sql_client(self) -> SqlClientBase[Any]:
     def sql_client_class(self) -> Type[SqlClientBase[Any]]:
         return self._dataset.sql_client_class
 
-    #
-    # Cursor Management
-    #
+    # TODO this probalby shouldn't be public; it should be accessed via the dataset or
```
sh-rp: You are right, you can make it private.

dlt/dataset/relation.py
```diff
@@ -232,87 +260,43 @@ def to_sql(self, pretty: bool = False, _raw_query: bool = False) -> str:
 
         return query.sql(dialect=self.query_dialect(), pretty=pretty)
 
+    # TODO this name is confusing; this returns the dialect of the destination;
```
sh-rp: you can rename it if you like

dlt/dataset/relation.py
```diff
@@ -430,12 +441,27 @@ def filter(  # noqa: A003
...
+            Self: A new Relation with the WHERE clause applied.
+        """
         if not operator and not value:
             return self.where(column_or_expr=column_or_expr)
         assert isinstance(column_or_expr, str)
         return self.where(column_or_expr=column_or_expr, operator=operator, value=value)
 
+    # NOTE The naming is ambiguous whether this returns a Relation or executes the query
+    # I suggest renaming `.to_scalar()`
     def scalar(self) -> Any:
```
sh-rp: we could name it `fetchscalar`, then it would fit our naming schema while indicating that something is being fetched from the destination.

tests/load/test_read_interfaces.py
sh-rp: you can remove these here now

sh-rp (changes requested):
Very cool PR, thanks for cleaning up a lot of things. I have some smaller annotations and you should make sure these changes run with the transformations, then we can merge this.

dlt/dataset/dataset.py
zilto: I have no immediate use, but it's practical to save the values used to construct the object. For instance, we could save the `self._destination_reference` instead of the `self._destination` when pickling the `dlt.Dataset` and we would be able to reconstruct it. It's also helpful for debugging

dlt/dataset/dataset.py
zilto: ## IDE support and type hinting

For instance, I just renamed `.scalar()` to `.fetchscalar()`. This is the only notation that fails to provide introspection and autocompletion

```python
dataset.items.scalar()  #<- can't introspect
dataset.table("items").scalar()  #<- method doesn't exist
dataset["items"].scalar()  #<- method doesn't exist
```

It also prevented me from efficiently renaming `scalar()`to `fetchscalar()` in all tests where with was used via the IDE

## Name collision

There's shadowing name collision. If you have a table named `schema`, `tables`, `dataset_name`, etc., it will return the method instead of data. This means that us changing *any* method on the class (addition, removal, renaming) would be a breaking change. 

To avoid avoid name collisions, users are forced to use `.table()` or `["table"]` for these specific tables and therefore should default to using only these two access methods.

## Single way to do things

I'm a big fan of Zen 13: `There should be one-- and preferably only one --obvious way to do it`. In the same realm, pandas just introduced the `pd.col()` selection (created by polars) instead of `df.column_name`, `df["column_name"]`, `df[["column_name"]]`

dlt/dataset/relation.py
zilto: I was on the fence on naming. Since a `Relation` maps to a table, it felt odd to have a `Table.table_schema` and `Table.schema`. It seems slightly better to have `Table.schema` and `Table.columns_schema`. Also, it mirrors the names of the typing `TTableSchema` and `TTableSchemaColumns`

dlt/dataset/dataset.py
zilto: I just moved the code existing code. I would keep this as future work. The exceptions under `dlt.common.destinations.exceptions` seems specific to pipeline runs / data loading and might not apply here.

dlt/dataset/relation.py
zilto: I'm currently running some tests. It seems that some `Dataset` imports need to be updated. I'm the process of opening a PR on dlt-plus

sh-rp (approved):

sh-rp:
NOTE: merging as failing tests are flukes as verified locally. One last commit is pushed that fixes a repr test.

PR #3060, opened 2025-09-04
Improved pipeline attach command and Dashboard launcher extensions

sh-rp:
### Description

This PR 
* allows to sync a pipeline from destination if it is not found locally by the attach command.
* Allows to set the port and host when launching the dashboard programmatically

### DIFF
--- a/dlt/pipeline/__init__.py
+++ b/dlt/pipeline/__init__.py
@@ -260,8 +260,8 @@ def attach(
             raise CannotRestorePipelineException(
                 pipeline_name,
                 p.working_dir,
-                f"Local pipeline not in {p.working_dir} found and no destination provided to"
-                " restore from.",
+                f"the pipeline was not found in {p.working_dir} found and no destination was"
+                " provided to restore from.",
             )
         p.sync_destination()
 
@@ -272,8 +272,8 @@ def attach(
             raise CannotRestorePipelineException(
                 pipeline_name,
                 p.working_dir,
-                f"Local pipeline not in {p.working_dir} found and provided destination and dataset"
-                " do not contain state for this pipeline.",
+                f"the pipeline was not found in {p.working_dir} found and provided destination and"
+                " dataset do not contain state for this pipeline.",
             )
         return p
 
--- a/tests/load/pipeline/test_pipelines.py
+++ b/tests/load/pipeline/test_pipelines.py
@@ -291,7 +291,7 @@ def test_attach_edgecases(destination_config: DestinationTestConfiguration) -> N
     p._wipe_working_folder()
     with pytest.raises(CannotRestorePipelineException) as exc_info:
         dlt.attach("test_attach_edgecases")
-    assert "no destination provided to restore from" in str(exc_info.value)
+    assert "no destination was provided to restore from" in str(exc_info.value)
     # no working folder left behing
     assert not p._pipeline_storage.has_folder("")
 


dlt/pipeline/__init__.py
```diff
@@ -193,6 +196,8 @@ def attach(
     destination: TDestinationReferenceArg = None,
     staging: TDestinationReferenceArg = None,
     progress: TCollectorArg = _NULL_COLLECTOR,
+    dataset_name: str = None,
+    sync_if_missing: bool = False,
     **injection_kwargs: Any,
 ) -> Pipeline:
```
sh-rp, line 199: I'm not sure about the changes to `def attach`. Maybe we should have something like `attach_remote` with a reduced arg set that does this and leave the attach unchanged.

dlt/cli/deploy_command_helpers.py
```diff
@@ -349,11 +368,33 @@ def parse_pipeline_info(visitor: PipelineScriptVisitor) -> List[Tuple[str, Optio
...
-                        "The value of 'pipeline_name' argument in call to `dlt_pipeline` cannot be"
+                        "The value of 'pipeline_name' argument in call to `dlt.pipeline` cannot be"
                         f" determined from {ast.unparse(p_d_node).strip()}. Pipeline working dir"
                         " will be found. Pass it directly with --pipeline-name option.",
                     )
-            pipelines.append((pipeline_name, pipelines_dir))
+            if extended_info:
+                d_t_node = call_args.arguments.get("destination")
+                if d_t_node:
+                    destination = evaluate_node_literal(d_t_node)
```
sh-rp: this does not work for any destination that is not a string literal..

dlt/cli/command_wrappers.py
```diff
@@ -154,10 +156,43 @@ def schema_command_wrapper(file_path: str, format_: str, remove_defaults: bool)
...
+            if len(possible_pipelines) > 1:
+                fmt.warning(
+                    "Multiple pipelines found in script, attaching pipeline"
+                    f" {pipeline_info['pipeline_name']}"
+                )
+
+            pipeline_name = pipeline_info["pipeline_name"]
+            pipelines_dir = pipeline_info["pipelines_dir"]
+
+            dlt.attach(
```
sh-rp: What can happen here is, that a user wants to open the pipeline as defined in the script, but there already is the state of some other pipeline with the same name on the local machine and that one is opened.

dlt/cli/command_wrappers.py
```diff
@@ -154,10 +156,43 @@ def schema_command_wrapper(file_path: str, format_: str, remove_defaults: bool)
 
 
 @utils.track_command("dashboard", True)
-def dashboard_command_wrapper(pipelines_dir: Optional[str], edit: bool) -> None:
+def dashboard_command_wrapper(
```
sh-rp: alternatively do changing this top level dashboard command, we could also allow the pipeline command to not take a pipeline name but a script file. But that is probably quite confusing.

dlt/cli/deploy_command_helpers.py
```diff
@@ -349,11 +368,33 @@ def parse_pipeline_info(visitor: PipelineScriptVisitor) -> List[Tuple[str, Optio
...
+                    if destination is None:
+                        raise CliCommandInnerException(
```
rudolfix: just warn here to be backward compatible

dlt/cli/deploy_command_helpers.py
rudolfix: here we could also parse destination factory. but IMO we should not invest too much in AST right now

dlt/cli/command_wrappers.py
```diff
@@ -154,10 +156,43 @@ def schema_command_wrapper(file_path: str, format_: str, remove_defaults: bool)
...
+    pipelines_dir: Optional[str], edit: bool, pipeline_script_path: Optional[str]
+) -> None:
     from dlt.helpers.dashboard.runner import run_dashboard
 
-    run_dashboard(pipelines_dir=pipelines_dir, edit=edit)
+    # if a pipeline script path is provided, we need to parse out pipeline info from script and sync it
+    pipeline_name: str = None
```
rudolfix: I get what you do here. but this code should be executed by `deploy` command (or not at all because deploy command has access to pipeline state and trace). I was possibly not specific when we discussed that:

1. workspace dashboard and report notebook should attach in the same way `dlt.attach(pipeline_name)` should be enough in both cases
2. it is the task of deployment script to generate additional information (form AST/state/trace) and add it to the job package. here it may just emit env variables:
```
PIPELINES__<pipeline_name>__DESTINATION_TYPE=...
PIPELINES__<pipeline_name>__DESTINATION_NAME=...
PIPELINES__<pipeline_name>__DATASET_NAME=...
...
```
attach will see it automatically even without those parameters being passed. look at the code

now the big question is how we gather this parameters. if you are against runtime information like using state or trace then we'll invest in AST parsing. but IMO it will never be as good

dlt/pipeline/__init__.py
```diff
@@ -216,24 +221,36 @@ def attach(
         staging or injection_kwargs.get("staging_type", None),
         destination_name=injection_kwargs.get("staging_name", None),
     )
+
+    pipeline_kwargs = {
```
rudolfix: I think there's a cleaner way. will leave a comment separately

dlt/pipeline/__init__.py
```diff
@@ -217,24 +218,30 @@ def attach(
...
-        False,  # always False as dev_mode so we do not wipe the working folder
-        progress,
-        True,
-        last_config(**injection_kwargs),
-        runtime_config,
-    )
-    # set it as current pipeline
-    p.activate()
-    return p
+    try:
```
rudolfix, line 224: please allow for explicit dataset name in args.

dlt/pipeline/__init__.py
```diff
@@ -217,24 +218,30 @@ def attach(
...
+            True,
+            last_config(**injection_kwargs),
+            runtime_config,
+        )
+        # set it as current pipeline
+        p.activate()
+        return p
+    except CannotRestorePipelineException:
+        # we can try to sync a pipeline with the given name
+        p = pipeline(pipeline_name, pipelines_dir, destination=destination, staging=staging)
```
rudolfix: we can attempt destination sync only if `destination` is set. otherwise rer-aise the exception.

dlt/pipeline/__init__.py
```diff
@@ -217,24 +218,30 @@ def attach(
...
+        p.sync_destination()
```
rudolfix, line 266: note: this can raise `PipelineStepFailed` if destination state is for some reason broken. this is OK to do

rudolfix: if p.first_run is True - means that there's no remote state. in that case you should wipe the pipeline working dir that got created by `dlt.pipeline` and reraise original exception

rudolfix: see my review message. it looks like it is already implemented in `pipeline_command`

PR #3063, opened 2025-09-05
Docs: Forcing root key propagation section improved

### LINKED ISSUE #3062: Propagating field: _dlt_root_id with replace after table already created in BigQuery causing tables to be updated instead of dropped/created.
### dlt version

1.15.0

### Describe the problem

We have a pipeline where we first used `replace` as the strategy for updating tables in BigQuery. However, now we are doing incremental updates and want to switch to `merge`, but we can't do so due to the following error:
```
Query error: Cannot add required fields to an existing schema. (field: _dlt_root_id) at [1:1]
```
So we decided to go back to using replace and to propagate the root key by setting `root_key=True`. Although with this we get the same error even when setting.
```
[destination]
replace_strategy = "staging-optimized"
```
According to the docs it should drop and re-create the tables in BigQuery. Could this be a bug with dlt because based on the docs it doesn't seem like it should be trying to update the existing table with this replace strategy enabled.

### Expected behavior

 https://dlthub.com/docs/general-usage/merge-loading#forcing-root-key-propagation
https://dlthub.com/docs/general-usage/full-loading#choosing-the-correct-replace-strategy-for-your-full-load

### Steps to reproduce

Run a `replace` without `root_key=True`, then again with. (Destination BigQuery)

### Operating system

macOS, Linux

### Runtime environment

Other

### Python version

3.13

### dlt data source

@dlt.source(root_key=True)

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
This PR adds an additional step to the [`Forcing root key propagation`](https://dlthub.com/docs/general-usage/merge-loading#forcing-root-key-propagation) section in `merge-loading.md`. The initial version suggests that this would work:

```py
import dlt

@dlt.resource
def my_table():
    yield {
        "id": 1,
        "name": "anuun",
        "age": 10, 
        "siblings": [
            {
                "id": 2,
                "name": "s",
            },
            {
                "id": 3,
                "name": "b",
            }
        ]
      }


@dlt.source()
def my_source():
    return [my_table()]

pipeline = dlt.pipeline(
    "test_pipeline",
    destination="duckdb",
    dataset_name="test_dataset",
)

the_source = my_source()

load_info = pipeline.run(the_source, write_disposition="replace")

the_source.root_key = True

load_info = pipeline.run(the_source, write_disposition="replace")
print(load_info)

```
However, this fails with the warning:
```
2025-09-05 16:11:07,691|[WARNING]|20067|8291859136|dlt|job_client_impl.py|_check_table_update_hints:769|Column(s) ['"_dlt_root_id"'] with NOT NULL are being added to existing table my_table__siblings. If there's data in the table the operation will fail.
```
as well as the error:
```
Parser Error: Adding columns with constraints not yet supported
```

The reason for this is self-explanatory and stems from the fact that dlt attempts to add the `_dlt_root_id` column with a `NOT NULL` constraint to an existing schema.

The solution would be to drop the sources/resources with the drop command, rerun the pipeline with "replace" and root key enabled. This is now mentioned in the docs. 

Resolves #3062

### DIFF
--- a/docs/website/docs/general-usage/merge-loading.md
+++ b/docs/website/docs/general-usage/merge-loading.md
@@ -220,52 +220,51 @@ Indexing is important for doing lookups by column value, especially for merge wr
 
 ### Forcing root key propagation
 
-The merge write disposition requires that the `_dlt_id` (`row_key`) of the root table be propagated to nested tables. This concept is similar to a foreign key but always references the root (top level) table, skipping any intermediate parents. We call this the `root key`. Root key propagation is automatically enabled for all tables that have the `merge` write disposition set. We do not enable it everywhere because it takes up additional storage space. Nevertheless, in some cases, you may want to permanently enable root key propagation.
+The merge write disposition requires that the `_dlt_id` (`row_key`) of the root table be propagated to nested tables. This concept is similar to a foreign key but always references the root (top level) table, skipping any intermediate parents. We call this the `root key`.
 
-**Note**: If you enable `root_key=True` on an existing source that previously ran without it, you'll need to drop and recreate the affected tables since the `_dlt_root_id` column cannot be added to existing tables that contain data.
+`Root key` propagation is automatically enabled for all tables that have the `merge` write disposition set from the beginning. We do not always enable it by default because it takes up additional storage space. Nevertheless, in some cases, you may want to permanently enable `root key` propagation.
 
-For example, suppose you have a source that doesn't use the merge write disposition and doesn't have `root_key` enabled:
+To enable `root key` propagation on an existing source or resource, you must drop and recreate its tables, since the `_dlt_root_id` column cannot be added to tables that already contain data.
+
+For example, suppose you used the [Facebook Ads](../dlt-ecosystem/verified-sources/facebook_ads.md) verified source, where the `merge` write disposition and `root key` are not enabled by default, to load the `ads` resource:
 ```py
 pipeline = dlt.pipeline(
-    pipeline_name='facebook_insights',
+    pipeline_name='facebook_ads_pipeline',
     destination='duckdb',
-    dataset_name='facebook_insights_data',
-    dev_mode=True
+    dataset_name='facebook_ads_data',
 )
-fb_ads = facebook_ads_source()
+my_facebook_ads = facebook_ads_source()
+pipeline.run(my_facebook_ads.with_resources("ads"))
 ```
 
-If you want to change the write disposition of the `ads` resource to use merge, you must first drop the existing resource tables from the destination:
+If you want to change the `ads` resource to use `merge`, you must first drop the existing resource tables from the destination:
 
 ```sh
-dlt pipeline facebook_insights ads
+dlt pipeline facebook_ads_pipeline drop ads
 ```
 
-This command removes the `ads` table and all its nested tables from the destination, allowing them to be recreated with the new schema that includes the `_dlt_root_id` column.
+This command removes the `ads` table and all its nested tables from the destination, allowing them to be later recreated with a schema that includes the `_dlt_root_id` column.
 
-Next, enable root key propagation and run with replace disposition, followed by merge:
+Next, enable `root key` propagation and run the pipeline once with `replace`, followed by `merge`:
 
 ```py
 pipeline = dlt.pipeline(
-    pipeline_name='facebook_insights',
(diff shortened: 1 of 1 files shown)


docs/website/docs/general-usage/merge-loading.md
```diff
@@ -242,7 +265,7 @@ fb_ads.ads.bind(states=("PAUSED", ))
 info = pipeline.run(fb_ads.with_resources("ads"), write_disposition="merge")
 ```
 
-In the example above, we enforce the root key propagation with `fb_ads.root_key = True`. This ensures that the correct data is propagated on the initial `replace` load so the future `merge` load can be executed. You can achieve the same in the decorator `@dlt.source(root_key=True)`.
+In the example above, we enable root key propagation with `fb_ads.root_key = True` and run the pipeline once with `replace` write disposition. This ensures that the schema includes the `_dlt_root_id` column when the tables are recreated with `replace`. Once this column is present, subsequent `merge` loads can be executed successfully. You can also enable root key propagation by decorating the source with `@dlt.source(root_key=True)`
```
anuunchin: @rudolfix  This works, but it may be better if dlt is able to automatically handle a scenario where a `_dlt_root_id` column is being added to a schema with `replace` write disposition, but then again this seems to be a pretty good solution 👀

docs/website/docs/general-usage/merge-loading.md
```diff
@@ -220,7 +220,30 @@ Indexing is important for doing lookups by column value, especially for merge wr
...
+For example, suppose you have a source that doesn't use the merge write disposition and doesn't have `root_key` enabled:
+```py
+pipeline = dlt.pipeline(
+    pipeline_name='facebook_insights',
+    destination='duckdb',
+    dataset_name='facebook_insights_data',
+    dev_mode=True
+)
+fb_ads = facebook_ads_source()
+```
```
zilto: I would trim the snippet to a minimum and add a bit more context relevant to `root_key`.

```py
@dlt.source
def facebook(): ...
   @dlt.resource
   def ads(...): ...
     
   return [ads(...), ...]

pipeline = dlt.pipeline("facebook", destination="duckdb")
```

Also, use consistent naming in related text. We currently have `facebook_ads_source()`, `fb_ads`,  `ads`, `facebook_insights`

docs/website/docs/general-usage/merge-loading.md
```diff
@@ -220,7 +220,30 @@ Indexing is important for doing lookups by column value, especially for merge wr
...
+
+If you want to change the write disposition of the `ads` resource to use merge, you must first drop the existing resource tables from the destination:
+
+```sh
+dlt pipeline facebook_insights ads
```
zilto: I don't think this is the intended drop command; ensure consistent naming

docs/website/docs/general-usage/merge-loading.md
```diff
@@ -242,7 +265,7 @@ fb_ads.ads.bind(states=("PAUSED", ))
 info = pipeline.run(fb_ads.with_resources("ads"), write_disposition="merge")
 ```
```
zilto, line 263: Trim the above example by removing all the code that isn't specifically relevant here:
- consistent naming with other snippets and text
- use a mock `@dlt.source` decorator to make the point more readable 
  ```py
  @dlt.source(root_key=True, write_disposition="replace")
  def facebook():
     @dlt.resource
     def ads(...): ...
     
     return [ads(...), ...]
  ```
- The `fb_ads.ads.bind(states=("DISAPPROVED", ))` is very confusing and don't seem specifically relevant to make the point. It's not about the facebook source
- don't assign `pipeline.run()` output to `info` if the variable is not used.

zilto (changes requested):
The update looks good. Most of my requested changes are about streamlining the already existing content. Sorry for upscoping

docs/website/docs/general-usage/merge-loading.md
anuunchin: agh, this is a dumb mistake ...

docs/website/docs/general-usage/merge-loading.md
anuunchin: It's a verified source, I'll adjust so that it's clear it's a verified source 👀

anuunchin: But also included this feedback 🫡

Issue #3072, opened 2025-09-09
allow custom metrics to be added to dlt resources and collect in extract info


rudolfix:
**Background**
We want resources and transform steps to be able to generate custom metrics and then collect them at the end of `extract` step when `ExtractInfo` is generated. Examples
1. `Incremental` generating a metric of total items vs. filtered out items and a metric related to key granularity (number of unique hashes in state) so we can detect common pitfalls ie. Incremental that everytime filters through a whole dataset or has low granularity cursor column
2. quality checks / validations that run after the `pipeline.run` in an additional `extract` step with `pipeline.dataset()` as input and store results in metrics dict

Those metrics will be visible in trace and ultimately may be used by Workspace Dashboard and other tools. Note that metrics are only stored in a trace and in ExtractInfo object and should be visible on `DltResource` object after pipeline has run.

We want a nice user interface for it:
- metrics are just a dictionary that is scoped to a resource. resource state is similar. user just sets values in it.
- users can get writable dict via `dlt.current.resource_metrics()`
- users implementing custom `ItemTransform` have access to `metrics` property of that item that will be collected
- we add `add_metrics` to `DltResource` which will add a transform step used to generate metrics from flowing data (this should be recommended way)

**Tasks**
We can split this into a few PRs:

PR 1:
implement custom metrics on a `DltResource`
1. * [ ] add `metrics` prop on `DltResource` and return it via `dlt.current.resource_metrics()`.
2. * [ ] add `metrics` prop on `ItemTransform`
3. * [ ] When extract step is completed, `extract` module must collect metrics from all steps of all participating resources. `_compute_metrics`method is a good place to do this. just merge all dictionaries: from `metrics` in `DltResource`, job metrics in `resource_metrics` and step metrics.
4. * [ ] Make sure those get into ExtractInfo
5. * [ ] Write tests where we check that metrics are collected and available, that they are in extract info and persisted in the schema
6. * [ ] Write tests to see what happens when users loads trace with dlt (what will be the shape of the data). mind that we have trace tests that also are checked against trace schema. find them and add metrics to this tests
7. * [ ] Write a chapter on custom metrics in `reource.md`

PR 2:
practical implementation:
1. * [ ] update `Incremental` class so it collects custom metrics:
- `unfiltered_items_count`: number of items entering the step (before incremental filtlers them out). remember to count rows like `LimitItem` does
- `unfiltered_batches_count`: number of batches (pages) entering the step
[...]

PR #3078, opened 2025-09-11
Feat: allowing custom metrics to be added to dlt resources and transform steps

anuunchin:
This PR is the first part of the custom metrics feature (#3072).  It provides a possibility to add a custom metric in dlt resources and transform steps. In a nutshell:

- The implementation adds a `metrics` prop to both `DltResource` and `ItemTransform` classes that return writable dictionaries.
- The`metrics` in `DltResource` itself can be accessible with `dlt.current.resource_metrics()`
- A new `ResourceMetrics` of TypedDict was created to structure metrics into three categories: 
   - `writer_metrics` (existing file performance data),
   - `custom_metrics` (user-defined)
   - `step_metrics` (from transform steps). T
The `_compute_metrics()` method in the extract module was adjusted to collect and merge all three metric types into the `ExtractInfo` within `resource_metrics`.

First PR in #3072

### DIFF
--- a/dlt/common/metrics.py
+++ b/dlt/common/metrics.py
@@ -23,13 +23,29 @@ class DataWriterMetrics(NamedTuple):
         return NotImplemented
 
 
-class DataWriterAndCustomMetrics(NamedTuple):
-    file_path: str
-    items_count: int
-    file_size: int
-    created: float
-    last_modified: float
+class DataWriterAndCustomMetrics(DataWriterMetrics):
     custom_metrics: Dict[str, Any]
+
+    def __new__(
+        cls,
+        file_path: str,
+        items_count: int,
+        file_size: int,
+        created: float,
+        last_modified: float,
+        custom_metrics: Dict[str, Any] = None,
+    ) -> "DataWriterAndCustomMetrics":
+        self = super(DataWriterAndCustomMetrics, cls).__new__(
+            cls, file_path, items_count, file_size, created, last_modified
+        )
+        self.custom_metrics = custom_metrics or {}
+        return self
+
+    def _asdict(self) -> Dict[str, Any]:
+        """Override _asdict to include custom_metrics in serialization"""
+        result = super()._asdict()
+        result["custom_metrics"] = self.custom_metrics
+        return result
 
 
 class StepMetrics(TypedDict):
--- a/dlt/extract/items_transform.py
+++ b/dlt/extract/items_transform.py
@@ -32,7 +32,16 @@ ItemTransformFunctionNoMeta = Callable[[TDataItem], TAny]
 ItemTransformFunc = Union[ItemTransformFunctionWithMeta[TAny], ItemTransformFunctionNoMeta[TAny]]
 
 
-class ItemTransform(ABC, Generic[TAny]):
+class BaseItemTransform:
+    def __init__(self) -> None:
+        self._custom_metrics: Dict[str, Any] = {}
+
+    @property
+    def custom_metrics(self) -> Dict[str, Any]:
+        return self._custom_metrics
+
+
+class ItemTransform(BaseItemTransform, ABC, Generic[TAny]):
     _f_meta: ItemTransformFunctionWithMeta[TAny] = None
     _f: ItemTransformFunctionNoMeta[TAny] = None
 
@@ -40,6 +49,7 @@ class ItemTransform(ABC, Generic[TAny]):
     """Tell how strongly an item sticks to start (-1) or end (+1) of pipe."""
 
     def __init__(self, transform_f: ItemTransformFunc[TAny]) -> None:
+        super().__init__()
         # inspect the signature
         sig = inspect.signature(transform_f)
         # TODO: use TypeGuard here to get rid of type ignore
@@ -47,7 +57,6 @@ class ItemTransform(ABC, Generic[TAny]):
             self._f = transform_f  # type: ignore
         else:  # TODO: do better check
             self._f_meta = transform_f  # type: ignore
-        self._custom_metrics: Dict[str, Any] = {}
 
     def bind(self: "ItemTransform[TAny]", pipe: SupportsPipe) -> "ItemTransform[TAny]":
         return self
@@ -56,13 +65,6 @@ class ItemTransform(ABC, Generic[TAny]):
     def __call__(self, item: TDataItems, meta: Any = None) -> Optional[TDataItems]:
         """Transforms `item` (a list of TDataItem or a single TDataItem) and returns or yields TDataItems. Returns None to consume item (filter out)"""
         pass
-
-    @property
-    def custom_metrics(self) -> Dict[str, Any]:
-        """Customizable resource metrics"""
-        if not hasattr(self, "_custom_metrics"):
-            self._custom_metrics = {}
-        return self._custom_metrics
 
 
 class FilterItem(ItemTransform[bool]):
@@ -158,6 +160,7 @@ class LimitItem(ItemTransform[TDataItem]):
     def __init__(
         self, max_items: Optional[int], max_time: Optional[float], count_rows: bool
     ) -> None:
+        BaseItemTransform.__init__(self)
         self.max_items = max_items if max_items is not None else -1
         self.max_time = max_time
         self.count_rows = count_rows
--- a/tests/extract/test_extract.py
+++ b/tests/extract/test_extract.py
@@ -612,8 +612,7 @@ def test_materialize_table_schema_with_pipe_items():
 
 
 @pytest.mark.parametrize(
-    "with_custom_metrics",
-    [True, False],
+    "with_custom_metrics", [True, False], ids=["with_custom_metrics", "without_custom_metrics"]
 )
 def test_resource_custom_metrics(extract_step: Extract, with_custom_metrics: bool) -> None:
     """Ensure that custom metrics from resources are collected and transform steps are available in extract info"""
@@ -684,8 +683,7 @@ def test_resource_custom_metrics(extract_step: Extract, with_custom_metrics: boo
 
 
 @pytest.mark.parametrize(
-    "with_custom_metrics",
-    [True, False],
+    "with_custom_metrics", [True, False], ids=["with_custom_metrics", "without_custom_metrics"]
 )
 def test_resource_step_custom_metrics(extract_step: Extract, with_custom_metrics: bool) -> None:
     """Ensure that custom metrics from both resources and their transform steps are collected and merged"""
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -43,7 +43,7 @@ from dlt.common.incremental.typing import (
     TIncrementalRange,
 )
 from dlt.extract.items import SupportsPipe, TTableHintTemplate
-from dlt.extract.items_transform import ItemTransform
+from dlt.extract.items_transform import BaseItemTransform, ItemTransform
 from dlt.extract.state import resource_state
 from dlt.extract.incremental.transform import (
     JsonIncremental,
@@ -644,6 +644,7 @@ class IncrementalResourceWrapper(ItemTransform[TDataItem]):
         Args:
             primary_key (TTableHintTemplate[TColumnKey], optional): A primary key to be passed to Incremental Instance at execution. Defaults to None.
         """
+        BaseItemTransform.__init__(self)
         self.primary_key = primary_key
         self.incremental_state: IncrementalColumnState = None
         self._allow_external_schedulers: bool = None
--- a/dlt/extract/validation.py
+++ b/dlt/extract/validation.py
@@ -9,7 +9,7 @@ except ModuleNotFoundError:
 from dlt.common.typing import TDataItems
 from dlt.common.schema.typing import TAnySchemaColumns, TSchemaContract, TSchemaEvolutionMode
 from dlt.extract.items import TTableHintTemplate
-from dlt.extract.items_transform import ValidateItem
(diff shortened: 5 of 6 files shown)


dlt/common/metrics.py
```diff
@@ -23,6 +25,55 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
...
+
+    @property
+    def custom_metrics(self) -> Mapping[str, Any]:
+        return MappingProxyType(self._custom)
+
+    def _asdict(self) -> Dict[str, Any]:
+        d = self.writer_metrics._asdict()
+        d.update(self._custom)
+        return d
+
```
anuunchin, line 43: Maybe there should be a custom __repr__ implementation, so the custom metrics are printed on the same level as the writer metrics, not within "_custom" 👀

dlt/common/metrics.py
```diff
@@ -23,6 +25,55 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
...
+        cls,
+        writer_metrics: DataWriterMetrics,
+        resource_metrics: Dict[str, Any] = None,
+        step_metrics: Dict[str, Any] = None,
+    ) -> "DataWriterAndCustomMetrics":
+        custom = {}
+        if resource_metrics:
+            custom.update(resource_metrics)
+        if step_metrics:
+            custom.update(step_metrics)
```
anuunchin: Maybe we should add a prefix, so that custom fields can never overlap with fields from the data writer metrics 👀

docs/website/docs/general-usage/resource.md
```diff
@@ -693,6 +693,49 @@ pipeline.run(
...
+# Access custom metrics from last trace
+trace = pipeline.last_trace
+load_id = load_info.loads_ids[0]
+resource_metrics = trace.last_extract_info.metrics[load_id][0]["resource_metrics"]["rows_table"]
+
+print(f"Custom metrics: {resource_metrics.custom_metrics}")
+```
+
+Custom metrics are included in pipeline traces and cannot use reserved keys (`file_path`, `items_count`, `file_size`, `created`, `last_modified`). Use them with [pipeline trace loading](../running-in-production/running.md#inspect-and-save-the-load-info-and-trace) for pipeline monitoring.
+
```
anuunchin, line 748: I think I couldn't really convey the benefits of using custom metrics here - but my understanding is that the resource.md will evolve as I go to the next PRs in the ticket 👀

tests/extract/test_extract.py
```diff
@@ -609,3 +609,60 @@ def empty_list(
...
+        "YieldMapItem": {},
+    }
+
+    assert all_resource_metrics["resource_with_other_metrics"].custom_metrics == {
+        "custom_count": 3,
+        "random_constant": 251.3,
+        "random_nested": {"value": 4, "unit": None},
+        "LimitItem": {},
+        "MapItem": {},
+    }
```
anuunchin: Maybe they should be accessible one level higher as an attribute:

```python
all_resource_metrics["resource_with_other_metrics"].custom_count
```

dlt/common/metrics.py
```diff
@@ -23,6 +25,55 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
         return NotImplemented
 
 
+@dataclass(frozen=True)
+class DataWriterAndCustomMetrics:
```
anuunchin: This is probably overkill, but I didn't want to touch the existing `DataWriterMetrics` 👀

dlt/extract/items_transform.py
```diff
@@ -55,6 +57,13 @@ def __call__(self, item: TDataItems, meta: Any = None) -> Optional[TDataItems]:
         """Transforms `item` (a list of TDataItem or a single TDataItem) and returns or yields TDataItems. Returns None to consume item (filter out)"""
         pass
 
+    @property
+    def metrics(self) -> Dict[str, Any]:
+        """Customizable resource metrics"""
+        if not hasattr(self, "_metrics"):
```
rudolfix: is there a reason? you create attribute in `__init__` so it should work

dlt/extract/resource.py
```diff
@@ -139,6 +140,7 @@ def __init__(
         self._explicit_args: DictStrAny = None
         self.source_name = None
         self._parent: DltResource = None
+        self._metrics: Dict[str, Any] = {}
```
rudolfix: pls. find `def _clone(` method and make note that we are not cloning metrics (does not make sense IMO)

dlt/extract/extract.py
```diff
@@ -251,7 +253,15 @@ def _compute_metrics(self, load_id: str, source: DltSource) -> ExtractMetrics:
         }
         # aggregate by resource name
         resource_metrics = {
-            resource_name: sum(map(lambda pair: pair[1], metrics), EMPTY_DATA_WRITER_METRICS)
+            resource_name: DataWriterAndCustomMetrics.from_writer_metrics(
+                writer_metrics=sum(map(lambda pair: pair[1], metrics), EMPTY_DATA_WRITER_METRICS),
+                resource_metrics=source.resources[resource_name].metrics,
```
rudolfix: I think for custom metrics we should just collect the information that user created and keep it in trace. If those were automatic metrics I'd probably keep everything separate but here if user really wants that, s/he can just embed them in custom dicts. so. For example if we instrument incremental we'll keep those metrics as {"incremental": {...metric..}} so they have a separate node.

tldr;>
1. merge resource and step dicts using `update_dict_nested` in order of pipe execution (resource first, then step in order)
2. pass it to DataWriterAndCustomMetrics

dlt/common/metrics.py
rudolfix: this is good idea but just derive from `DataWriterMetrics` and keep it as named tuple

dlt/common/metrics.py
```diff
@@ -23,6 +25,55 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
...
+    writer_metrics: DataWriterMetrics
+    _custom: Dict[str, Any] = field(default_factory=dict)
```
rudolfix: we just need `custom_metrics`: Dict[str, Any] here. and we do not need to put `writer_metrics` in container property. So we have backward compatible interface. And less tables when you load this with dlt as trace

dlt/common/metrics.py
```diff
@@ -23,6 +25,55 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
...
+
+    @classmethod
+    def from_writer_metrics(
```
rudolfix: you won't need that. NamedTuple creates constructor and repr

dlt/common/metrics.py
```diff
@@ -23,6 +25,55 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
...
+    ) -> "DataWriterAndCustomMetrics":
+        custom = {}
+        if resource_metrics:
+            custom.update(resource_metrics)
+        if step_metrics:
+            custom.update(step_metrics)
+        return cls(writer_metrics, custom)
+
+    @property
+    def file_path(self) -> str:
```
rudolfix: OK! but now you'll have it automatically

dlt/common/metrics.py
```diff
@@ -23,6 +25,55 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
...
+
+    @property
+    def last_modified(self) -> float:
+        return self.writer_metrics.last_modified
+
+    @property
+    def custom_metrics(self) -> Mapping[str, Any]:
+        return MappingProxyType(self._custom)
+
+    def _asdict(self) -> Dict[str, Any]:
```
rudolfix, line 44: this is good, let's merge all props together for loading. but I doubt that you can override any methods in NamedTuple so you'll need to merge props in `def asdict(self) -> DictStrAny:` of `ExtractInfo`

docs/website/docs/general-usage/resource.md
rudolfix: the biggest benefit is what you mention at the end: those values go into trace and will be available in various observability tool. we'll add them to the dashboard for example, after this ticket is merged. 

metrics can be useful when some information is lost ie. if we yield rows conditionally it may make sense to include total number of tested rows. in your example if we yield only numbers below threshold you could return total count as custom metric

tests/extract/test_extract.py
```diff
@@ -609,3 +609,60 @@ def empty_list(
...
+        custom_metrics["random_nested"] = {"value": 4, "unit": None}
+        yield [{"id": 1}, {"id": 2}]
+
+    resource_with_other_metrics.add_limit(10)
+    resource_with_other_metrics.add_map(lambda x: x)
+
+    source = DltSource(
+        dlt.Schema("metrics"), "module", [resource_with_metrics(), resource_with_other_metrics()]
+    )
+    load_id = extract_step.extract(source, 20, 1)
```
rudolfix, line 658: please check if `metrics` property got set on the resources, should be present after extract

rudolfix: hmmm maybe let's rename them to `custom_metrics`

tests/extract/test_extract.py
```diff
@@ -609,3 +609,60 @@ def empty_list(
...
+    step_info = extract_step.get_step_info(MockPipeline("buba", first_run=False))  # type: ignore[abstract]
+
+    all_resource_metrics = step_info.metrics[load_id][0]["resource_metrics"]
+    assert "resource_with_metrics" in all_resource_metrics
+    assert "resource_with_other_metrics" in all_resource_metrics
+
+    assert all_resource_metrics["resource_with_metrics"].custom_metrics == {
```
rudolfix: please also test
1. resource with steps that generate metrics
2. resource that does not generate any custom metrics. in that case dict should be empty and serialized (`asdict`) should not contain any additional properties

dlt/common/metrics.py
rudolfix: I actually looked at trace schema contract. It is better if we keep custom metrics in a separate `custom` dict for loading, this becomes easier to manage. for example we can mark `custom` field as `json` in our trace contract. let's try to do that

tests/pipeline/test_pipeline_trace.py
```diff
@@ -714,6 +718,53 @@ def test_last_pipeline_step_trace_returns_latest() -> None:
...
+    trace_pipeline = dlt.pipeline(
+        pipeline_name="test_custom_metrics_trace_schema", destination=dummy(completed_prob=1.0)
+    )
+    trace_pipeline.run([trace], table_name="trace")
+    inferred_schema = trace_pipeline.default_schema
+    resource_metrics_table_cols = inferred_schema.get_table(
+        "trace__steps__extract_info__resource_metrics"
+    )["columns"]
+
+    assert "custom_count" in resource_metrics_table_cols
```
rudolfix: OK with all the changes, those columns should have `custom__` prefix

docs/website/docs/general-usage/resource.md
```diff
@@ -693,6 +693,49 @@ pipeline.run(
 
 The `with_name` method returns a deep copy of the original resource, its data pipe, and the data pipes of a parent resource. A renamed clone is fully separated from the original resource (and other clones) when loading: it maintains a separate [resource state](state.md#read-and-write-pipeline-state-in-a-resource) and will load to a table.
 
+## Collect custom metrics in resources
```
rudolfix: `## Collect custom metrics` is sufficient

dlt/extract/resource.py
anuunchin: aha, i wasn't aware of this - this is good to know

dlt/extract/items_transform.py
```diff
@@ -55,6 +57,13 @@ def __call__(self, item: TDataItems, meta: Any = None) -> Optional[TDataItems]:
         """Transforms `item` (a list of TDataItem or a single TDataItem) and returns or yields TDataItems. Returns None to consume item (filter out)"""
         pass
 
+    @property
+    def custom_metrics(self) -> Dict[str, Any]:
+        """Customizable resource metrics"""
+        if not hasattr(self, "_custom_metrics"):
+            self._custom_metrics = {}
+        return self._custom_metrics
+
```
anuunchin: @rudolfix  I think we need 
```
        if not hasattr(self, "_custom_metrics"):
            self._custom_metrics = {}
```

for now because some child classes implement their own init and don't have this 👀

docs/website/docs/general-usage/resource.md
anuunchin: I improved the exmaple -> i think it's much better because it uses an example that's already in the docs and gets a "real" metric which is not in the trace 👀

dlt/common/metrics.py
```diff
@@ -23,6 +23,15 @@ def __add__(self, other: Tuple[object, ...], /) -> Tuple[object, ...]:  # type:
         return NotImplemented
 
 
+class DataWriterAndCustomMetrics(NamedTuple):
```
rudolfix: You could derive from `DataWriterMetrics`. is there a reason not to do that?

dlt/extract/items_transform.py
rudolfix: OK for now. you can also add it at class level (together with `_f`) but set it to None not `{}` otherwise all instances will share it. the correct way is to fix our init: possibly via another base class

rudolfix (changes requested):
two minor changes. after that you can merge. this is very good

dlt/extract/items_transform.py
anuunchin: owww

PR #3080, opened 2025-09-11
feat(notebook): more widgets!

zilto:
Refined the way to create reusable widget and packaged it under [mowidgets](https://github.com/zilto/mowidgets). 

Changes:
- simplifies how to create widgets; instructions updated
- added a widget that shows a summary table of the pipelines on your local machine
- added a test to ensure we don't remove / rename the public widgets we released

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2143,7 +2143,7 @@ lancedb = [
 ]
 marimo = [
     { name = "marimo" },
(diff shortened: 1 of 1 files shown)


zilto:
closing for now as it's currently not planned. Alternatively, the `dlt.Dataset` should have a reference to the pipelines that write to it. 

edit: oops, wrong PR. Will reopen with update content anyways

PR #3083, opened 2025-09-12
repo(ci): disable docker container autorestart

zilto:
`dlt` uses several containers in its test setup. Most define the policy `restart: unless-stopped` in their docker-compose.

This means these containers are started when booting the developer's machine and run most of the time in the background. This has a noticeable impact on CPU during intensive usage. The developer is required to manually spin down containers each time via `docker container stop $(docker ps -q -a)` 

## Changes
- set policy `restart: "on-failure"` which doesn't start containers on boot.
- remove the `version:` field; this has been deprecated and ignored for some time
- set `container_name` with a `dlt_test_` prefix to facilitate container management and prevent spawning duplicate containers

## Future steps
- some docker compose use docker networks and named volumes; those can be harder to cleanup. Would be useful to minimize resource usage of dev machines or provide cleanup scripts

### DIFF
--- a/tests/load/clickhouse/docker-compose.yml
+++ b/tests/load/clickhouse/docker-compose.yml
@@ -1,7 +1,8 @@
----
 services:
   clickhouse:
     image: clickhouse/clickhouse-server
+    container_name: dlt_test_clickhouse
+    restart: on-failure
     ports:
       - "9000:9000"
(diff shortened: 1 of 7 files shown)


sh-rp:
@zilto thanks! This PR makes total sense.

PR #3086, opened 2025-09-13
fix: convert local file path to posix before PUT to Databricks destination

### LINKED ISSUE #3085: DatabricksLoadJob fails during _handle_local_file_upload due to Windows local path
### dlt version

1.15.0-1.16.0

### Describe the problem

When working on Windows, attempting to load local files to a Databricks volume fails with the traceback revealing the following exception:

`databricks.sql.exc.Error: Local file operations are restricted to paths within the configured staging_allowed_local_path`

### Expected behavior

The state file and extracted data files are successfully uploaded to the designated Databricks staging volume.

### Steps to reproduce

Attempt to run the [Direct Load (Databricks Managed Volumes)
](https://dlthub.com/docs/dlt-ecosystem/destinations/databricks#direct-load-databricks-managed-volumes) example from the docs on a Windows machine.

### Operating system

Windows

### Runtime environment

Local

### Python version

3.12

### dlt data source

https://pokeapi.co/api/v2/

### dlt destination

_No response_

### Other deployment details

Cannot upgrade to more recent Python version as we are bound to the same version that is being used by the latest LTS Databricks runtime due to company policy.

### Additional information

When debugging the underlying connector library that throws the exception I noticed that the local path that is compared with the allowed paths is malformed. I suspect that the connector library expects a POSIX path and that the backslashes in Windows paths are inadvertently being considered as escape characters.

Connector library: databricks-sql-connector==3.7.4

The function is Cursor._handle_staging_operation in `databricks\sql\client.py`.

I fixed the issue by altering the following line in the Databricks destination module:

https://github.com/dlt-hub/dlt/blob/5236e15f2a4f3de50758d4fe42ec6df7ce2b664c/dlt/destinations/impl/databricks/databricks.py#L139

change to

```python
        posix_path = pathlib.Path(local_file_path).as_posix()
        self._sql_client.execute_sql(f"PUT '{posix_path}' INTO '{volume_file_path}' OVERWRITE")
```

This fixed the issue for me. If this fix is OK, let me know and I will create a PR.

AndreiBondarenko:
### Description

Fixing local file upload to Databricks destination from Windows local.

Underlying connector expects POSIX path (undocumented). Backslashes in Windows path treated as escape characters, causing malformed paths which fail the allowed local paths check.

Fixed by converting to POSIX path before interpolation of PUT ... INTO command used for upload.

### Related Issues

- Fixes #3085 

### Additional Context

Verified fix using personal Databricks workspace as destination.

### DIFF
--- a/dlt/destinations/impl/databricks/databricks.py
+++ b/dlt/destinations/impl/databricks/databricks.py
@@ -137,7 +137,9 @@ class DatabricksLoadJob(RunnableLoadJob, HasFollowupJobs):
         volume_path = f"/Volumes/{volume_catalog}/{volume_database}/{volume_name}/{uniq_id()}"
         volume_file_path = f"{volume_path}/{volume_file_name}"
 
-        posix_path = Path(local_file_path).as_posix() # backslash in Windows local path causes issues with PUT command
+        posix_path = Path(
+            local_file_path
+        ).as_posix()  # backslash in Windows local path causes issues with PUT command
         self._sql_client.execute_sql(f"PUT '{posix_path}' INTO '{volume_file_path}' OVERWRITE")
 
         from_clause = f"FROM '{volume_path}'"


AndreiBondarenko:
Please provide guidance if additional tests are required.

sh-rp:
@AndreiBondarenko thanks for the PR! I think we can probably merge as if the tests pass. Please make sure that the linter step passes, you need to format the code.

AndreiBondarenko:
@sh-rp Fixed the formatting. My bad! Lint step succeeded locally.

sh-rp:
@AndreiBondarenko I will test this locally on a windows machine and then probably merge!

sh-rp:
NOTE: tested locally on windows machine that this was not working as reported by the user and this change fixes it. Thanks again for the contribution.

PR #3091, opened 2025-09-15
Improve pipeline dashboard test coverage

sh-rp:
### Description
This PR improves test coverage for pipeline dashboard

* Enabling the e2e tests for the dashboard by fixing issues with async and paths, also disables ibis section opening as this locks the duckdb file forever and tests stop
* Re-enables macos e2e tests by downgrading to macos 14 (instead of 15.6) On MacOS 15 marimo would not start in the background with the & operator in the shell command of the workflow.
* Basic tests for functions in ui elements utils


tests/e2e/helpers/dashboard/test_e2e.py
```diff
@@ -90,6 +96,11 @@ def broken_resource():
 #
 
 
+def _normpath(path: str) -> str:
```
zilto, line 100: The built-in [pathlib](https://docs.python.org/3/library/pathlib.html) module should solve your problem. By creating a `pathlib.Path()` and converting it to a string, it should comply to the OS it runs on

```python
unix_string_path = "/path/to/my/dir"
path = pathlib.Path(unix_string_path)
os_compliant_path = str(path)
```

tests/helpers/dashboard/test_ui_elemens.py
zilto, line 1: typo: in file name `test_ui_elemens.py -> test_ui_elements.py`

tests/e2e/helpers/dashboard/test_e2e.py
```diff
@@ -1,4 +1,6 @@
 from typing import Any, Literal
+import asyncio  #
```
zilto: remove `#`

PR #3092, opened 2025-09-16
feat: `Dataset.write()`

zilto:
## Motivation
Users of `dlt.Dataset` want a simple way to write data back to the dataset.

Use cases:
- manually review data and push corrected records
- simple way to add records if you don't have access to the original `dlt.Pipeline` used to create the dataset

## Specs
- Look at `WritableDataset.save()` from `dlt-plus`
- Add `Dataset.write()` in `dlt` (this aligns with `pipeline.run()` operation)
  - Alternatives: `.write_to()`, `.load_into()`, `.load_table()`
- create an internal `dlt.Pipeline` named `_dlt_dataset_{dataset_name}`
- find a way for the internal pipeline to use the `dlt.Schema` from the `dlt.Dataset` instance; this way, this schema should evolve when `Dataset.load()` is used
- potential API
  ```python
  def load(
    self: dlt.Dataset,
    data: TDataItems,
    table_name: str,
    write_disposition: TWriteDisposition = "append",
    normalize: bool = False,
  ) -> LoadInfo: ...
  ```
  - `write_disposition` is useful to determine if we should append or modify existing records
  - `normalize` allows the user to decide to enable normalization (which might create more tables)
- can accept a `dlt.Relation` as input

## Out of scope
- `Dataset.load()` doesn't have to support 1-to-1 the `dlt.Pipeline.run()` method; if user needs full range of config, then they should create a pipeline

### DIFF
--- a/dlt/common/schema/schema.py
+++ b/dlt/common/schema/schema.py
@@ -1128,3 +1128,12 @@ class Schema:
         del state["naming"]
         del state["data_item_normalizer"]
         return state
+    
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, Schema):
+            raise NotImplementedError(f"Equality between `dlt.Schema` object and {type(other).__name__} is not supported.")
+
+        return self.version_hash == other.version_hash
+    
+    def __hash__(self) -> int:
+        return hash(self.version_hash)
--- a/dlt/common/schema/schema.py
+++ b/dlt/common/schema/schema.py
@@ -1128,12 +1128,14 @@ class Schema:
         del state["naming"]
         del state["data_item_normalizer"]
         return state
-    
+
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, Schema):
-            raise NotImplementedError(f"Equality between `dlt.Schema` object and {type(other).__name__} is not supported.")
+            raise NotImplementedError(
+                f"Equality between `dlt.Schema` object and {type(other).__name__} is not supported."
+            )
 
         return self.version_hash == other.version_hash
-    
+
     def __hash__(self) -> int:
         return hash(self.version_hash)
--- a/dlt/dataset/dataset.py
+++ b/dlt/dataset/dataset.py
@@ -193,16 +193,15 @@ class Dataset:
         This method uses a full-on `dlt.Pipeline` internally. You can retrieve this pipeline
         using `Dataset.get_write_pipeline()` for complete flexibility.
         """
-        resource = _data_to_resource(
-            data,
-            schema=self.schema,
-            table_name=table_name,
-            dataset_name=self.dataset_name,
-            write_disposition=write_disposition,
-        )
         internal_pipeline = _get_internal_pipeline(self.dataset_name, destination=self._destination)
         # TODO should we try/except this run to gracefully handle failed writes?
-        info = internal_pipeline.run([resource], schema=self.schema)
+        info = internal_pipeline.run(
+            data,
+            dataset_name=self.dataset_name,
+            table_name=table_name,
+            schema=self.schema,
+            write_disposition=write_disposition,
+        )
         # maybe update the dataset schema
         self._schema = internal_pipeline.default_schema
         return info
@@ -488,24 +487,6 @@ def _get_dataset_schema_from_destination_using_dataset_name(
                 schema = dlt.Schema.from_stored_schema(json.loads(stored_schema.schema))
 
     return schema
-
-# TODO move this to `dlt.extract`. Maybe there's an existing function?
-def _data_to_resource(
-    data: TDataItems,
-    *,
-    schema: dlt.Schema,
-    table_name: str,
-    dataset_name: str,
-    write_disposition: TWriteDisposition,
-) -> DltResource:
-    """Create a resource from a `data` argument."""
-    table_schema = schema.tables.get(table_name)
-    hints: TResourceHints = (
-        table_schema
-        if table_schema  # type: ignore
-        else dlt.mark.make_hints(table_name=table_name, write_disposition=write_disposition)
-    )
-    return DltResource.from_data(data, name=table_name, section=dataset_name, hints=hints)
 
 
 def _get_internal_pipeline(dataset_name: str, destination: AnyDestination) -> dlt.Pipeline:
--- a/dlt/dataset/dataset.py
+++ b/dlt/dataset/dataset.py
@@ -15,8 +15,6 @@ from dlt.common.destination.client import JobClientBase, SupportsOpenTables, Wit
 from dlt.common.typing import Self, TDataItems
 from dlt.common.schema.typing import C_DLT_LOAD_ID, TWriteDisposition
 from dlt.common.pipeline import LoadInfo
-from dlt.extract.hints import TResourceHints
-from dlt.sources import DltResource
 from dlt.common.utils import simple_repr, without_none
 from dlt.destinations.sql_client import SqlClientBase, WithSqlClient
 from dlt.dataset import lineage
@@ -174,7 +172,7 @@ class Dataset:
         helpful if we want to run sql queries without extracting the data
         """
         return is_same_physical_destination(self, other)
-    
+
     # TODO explain users can inspect `_dlt_loads` table to differentiate data originating
     # from `pipeline.run()` or `dataset.write()`
     def get_write_pipeline(self) -> dlt.Pipeline:
@@ -489,7 +487,9 @@ def _get_dataset_schema_from_destination_using_dataset_name(
     return schema
 
 
-def _get_internal_pipeline(dataset_name: str, destination: AnyDestination) -> dlt.Pipeline:
+def _get_internal_pipeline(
+    dataset_name: str, destination: TDestinationReferenceArg
+) -> dlt.Pipeline:
     """Setup the internal pipeline used by `Dataset.write()`"""
     pipeline = dlt.pipeline(
         pipeline_name=_INTERNAL_DATASET_PIPELINE_NAME_TEMPLATE.format(dataset_name=dataset_name),
--- a/dlt/dataset/dataset.py
+++ b/dlt/dataset/dataset.py
@@ -12,9 +12,11 @@ from dlt.common.libs.sqlglot import TSqlGlotDialect
 from dlt.common.json import json
 from dlt.common.destination.reference import AnyDestination, TDestinationReferenceArg, Destination
 from dlt.common.destination.client import JobClientBase, SupportsOpenTables, WithStateSync
-from dlt.common.schema import Schema
-from dlt.common.typing import Self
-from dlt.common.schema.typing import C_DLT_LOAD_ID
+from dlt.common.typing import Self, TDataItems
+from dlt.common.schema.typing import C_DLT_LOAD_ID, TWriteDisposition
+from dlt.common.pipeline import LoadInfo
+from dlt.extract.hints import TResourceHints
+from dlt.sources import DltResource
 from dlt.common.utils import simple_repr, without_none
 from dlt.destinations.sql_client import SqlClientBase, WithSqlClient
 from dlt.dataset import lineage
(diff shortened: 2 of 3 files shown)


rudolfix (commented):
My take on the interface.
1. `write` is OK!
2. make it a clearly specialized method to write a data chunk: current simple interface is good.
3. allowing `Relation` as data is a good idea. so there's certain symmetry (querying works with relations). in that case you can make `table_name` optional (relation has table name AFAIK?)

Implementation details:
my take would be to make internal pipeline used in `write` as invisible as possible:
1. disable destination sync, state sync and schema evolution (a total freeze on a table via contract)
2. possibly use `pipelines-dir` to hide it from command line and dashboard.

in essence we pretend that this pipeline does not exist

WDYT?

a helper class to get write pipeline is cool. I'd make it a public helper method. another helper method would be to convert Dataset into Source but that's another problem

1. I'd keep the internal pipeline that just loads data super simple

zilto:
> my take would be to make internal pipeline used in write as invisible as possible
> possibly use pipelines-dir to hide it from command line and dashboard.
> in essence we pretend that this pipeline does not exist

I changed the internal pipeline to be a context manager that uses a temporary directory as `pipelines_dir`

> disable destination sync, state sync and schema evolution (a total freeze on a table via contract)

I don't know exactly what I need to change / configure for destination and state sync (it doesn't seem to be in the kwargs for `dlt.pipeline()` and `pipeline.run()`).

For schema evolution, users should be able to modify schema. For example, someone wants to add a column or cast types. Though, I would have frozen schema as default and require users to explicitly change it.

tests/dataset/test_dataset_write.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def test_dataset_write():
+    dataset_name = "foo"
+    destination = dlt.destinations.duckdb(duckdb.connect())
+    dataset = dlt.dataset(destination, dataset_name)
+    table_name = "bar"
+    items = [{"id": 0, "value": "bingo"}, {"id": 1, "value": "bongo"}]
+
+    # TODO this is currently odd because the tables exists on the `Schema`
```
sh-rp, line 79: Schemas create the default tables on initialization. We could consider changing this, but that is the reason why they will always exist on any Schema instance regardless of wether anything was materialized.

dlt/dataset/dataset.py
```diff
@@ -170,6 +175,46 @@ def is_same_physical_destination(self, other: dlt.Dataset) -> bool:
...
 
+    # TODO explain users can inspect `_dlt_loads` table to differentiate data originating
+    # from `pipeline.run()` or `dataset.write()`
+    @contextmanager
+    def write_pipeline(self) -> Generator[dlt.Pipeline, None, None]:
+        """Get the internal pipeline used by `Dataset.write()`.
+
+        Passing a `pipelines_dir` allows you to set a
+        """
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
sh-rp: I think we should not run in this in a temporary directory but give the pipeline a predictable name and store it with the other pipeline metadata, this way the user can debug the run like any other pipeline run. This is up for debate though.

dlt/dataset/dataset.py
```diff
@@ -170,6 +175,46 @@ def is_same_physical_destination(self, other: dlt.Dataset) -> bool:
...
+                dataset_name=self.dataset_name, destination=self._destination, pipelines_dir=tmp_dir
+            )
+            yield pipeline
+
+    def write(
+        self,
+        data: TDataItems,
+        *,
+        table_name: str,
+        write_disposition: TWriteDisposition = "append",
```
sh-rp: I think the `columns` and `loader_file_format` args from the run method would also be good canditates here. You can also consider a `pipeline_kwargs` argument that get's forwarded to the internal pipeline instantiation. But maybe we do not need this and can add it if requested.

dlt/common/schema/schema.py
```diff
@@ -1128,3 +1128,14 @@ def __getstate__(self) -> Any:
         del state["naming"]
         del state["data_item_normalizer"]
         return state
+
+    def __eq__(self, other: Any) -> bool:
```
sh-rp, line 1132: Cool!

tests/dataset/test_dataset_write.py
```diff
@@ -0,0 +1,93 @@
+import pathlib
```
sh-rp, line 1: We need tests for writing into tables that already exist, and reading back from those tables with our database reader methods to see whether the schema was updated properly and works.<

We should also make sure that all the args provided to the write methods are forwarded properly.

dlt/dataset/dataset.py
```diff
@@ -170,6 +175,46 @@ def is_same_physical_destination(self, other: dlt.Dataset) -> bool:
...
+    # TODO explain users can inspect `_dlt_loads` table to differentiate data originating
+    # from `pipeline.run()` or `dataset.write()`
+    @contextmanager
+    def write_pipeline(self) -> Generator[dlt.Pipeline, None, None]:
+        """Get the internal pipeline used by `Dataset.write()`.
+
+        Passing a `pipelines_dir` allows you to set a
+        """
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            pipeline = _get_internal_pipeline(
```
sh-rp: You need to forward the staging destination here too, which probably includes knowing the staging destination on the dataset already. Alternatively one would have to provide the staging_destination via the run_kwargs. For working within notebooks where you often get the dataset from a pipeline instace it seems to me it would be good to always have it set on a dataset when you get it from the pipeline.

tests/dataset/test_dataset_write.py
zilto: I think the current behavior is ok and I wouldn't change it. Just wanted to leave a note in the test because the assertion could be surprising.

dlt/dataset/dataset.py
```diff
@@ -170,6 +175,46 @@ def is_same_physical_destination(self, other: dlt.Dataset) -> bool:
...
+            info = internal_pipeline.run(
+                data,
+                dataset_name=self.dataset_name,
+                table_name=table_name,
+                schema=self.schema,
+                write_disposition=write_disposition,
+            )
+
+        # maybe update the dataset schema
+        self._schema = internal_pipeline.default_schema
```
rudolfix: For sure - you want to see new tables right away, I'd extract this to a method and verify if we need to purge any cached objects (like sql_client which also holds schema)

rudolfix (commented):
We can continue with existing concept and implementation. Even if use case is simple (mostly interactive saving or data frames or relations) the underlying machinery is complex.
- in most cases we'll create new tables (ie. saving checks, saving "drafts" to be used later) - that will evolve and change schema. each load also generates load ids. **in that case I'll make internal pipeline visible in traces and also in workspace**. This is also the default behavior. We can change that later
- current implementation is good and we mostly need more tests.

Note that setting write disposition in `run` will change the schema.  My take would be to drop it from the arguments and replace with `overwrite` flag. Internally:
- if True we use `refresh="drop_resources"` to replace both table schema and data
- if False we fallback to the write_disposition that is already defined in the schema (append for new tables).
- if we need truncate mode, `overwrite` can be None | truncate | replace or similar
in all cases we use `refresh` flag to not change schema

having destructive overwrite IMO better fits interactive/exploratory work where destructive schema changes may be common at the beginning.

as a followup we should enable saving relations. ie. by finishing https://github.com/dlt-hub/dlt/pull/3272

another note: schema contract should apply. so if creating new tables or columns is blocked - this pipeline should comply.

tests:
- we need to test all modes (append, overwrite)
- create new tables, write to existing tables
- internal pipeline should be able to create new dataset - this is "scratchpad" case when data is saved aside or for experiments. should be tested
- curious how this behaves with schema contracts

@zilto @djudjuu my 2 cents here. I think Julius can take it over when we have go

djudjuu:
just had a meeting with marcin. the plan is now:
- lets focus on Dataset.write() as a function to create standalone Datasets (vs. using the dataset from the pipeline)
- signature is Dataset.write(data, table_name: str, overwrite:bool)
- overwrite=True, respect the write disposition of the underlying data
- overwrite=False does refresh="drop_resources", writes new schema into dlt_versions
- document in docs: that if take the Dataset from the pipeline and introduce changes to the schema (new table (append) or overwrite) they need to do pipeline.sync_destination() to pull the changes into the pipeline

##Follow-up:
- pipeline sync should be aware of: schema needs to be decoupled from state-sync
atm: new schemas won't automatically be picked up by a pipeline because pipeline compares local state to remote state (and state doesnt reference schema (dlt_version -table)

dlt/common/schema/schema.py
```diff
@@ -1192,3 +1128,14 @@ def __getstate__(self) -> Any:
...
```
zilto, line 1132: @djudjuu make sure to exclude this code. After research, this would be a significant breaking change. It would change the behavior of 

```python
schema1 = ...
schema2 = ...

schema1 == schema2
set([schema1, schema2])
{schema1: "foo", schema2: "bar"}
```
Also, it's an invalid `__hash__` method since the value is mutable


(you can ask an LLM for details)

dlt/dataset/dataset.py
zilto, line 78: @djudjuu thing to investigate: at the end of `Dataset.write()`, you could set `._schema = None`. This would force the next `Dataset.schema` call to sync the schema.

Follow-up issue: maybe the schema of the `dlt.Dataset` is not the same as the internal pipeline's schema name

dlt/dataset/dataset.py
```diff
@@ -170,6 +178,62 @@ def is_same_physical_destination(self, other: dlt.Dataset) -> bool:
...
+        The resulting load packages can be inspected in the pipeline's working directory which is
+        named "_dlt_dataset_{dataset_name}".
+        This directory will be wiped before each `write()` call.
+        """
+        with self.write_pipeline() as internal_pipeline:
+            # drop all load packages from previous writes
+            # internal_pipeline._wipe_working_folder()
+            internal_pipeline.drop()
+
+            # get write dispostion for existing table from schema (or "append" if table is new)
```
djudjuu: what about other resource level hints that make sure that the merging is done appropriately, e.g. see the docs [here](https://dlthub.com/docs/general-usage/merge-loading#example-with-merge-key-and-non-boolean-delete-column)

I don't think those will automatically be taken from the schema (easy to test though)

djudjuu:
also: `pipeline.sync_destination()` only syncs if the remote state version has changed (which it hasn't after `dataset.write()` in which case it calls `get_schema_from_destnation(always_download=False)` which just looks up the schema in the local storage (where it hasnt changed)
-> maybe we should introduce a `force_download`-flag to the sync_destination call :thinking: otherwise it won't happen

dlt/pipeline/pipeline.py
```diff
@@ -1658,7 +1665,7 @@ def managed_state(self, *, extract_state: bool = False) -> Iterator[TPipelineSta
             with self._container.injectable_context(StateInjectableContext(state=state)):
                 yield state
         except Exception:
-            backup_state = self._get_state()
+            yackup_state = self._get_state()
```
djudjuu, line 1668: backup!

zilto:
closing for now as it's currently not planned. Alternatively, the `dlt.Dataset` should have a reference to the pipelines that write to it.

djudjuu:
leaving some thoughts that came up while working on this:

if `dataset()` is taken from a pipeline, then `Dataset.write()` will change the schema (if somebody for example wants to correct a certain row with `write_disposition=merge`. The new schema is references in the dlt_version_tables (with a higher version than the pipeline schema).

BUT: the pipeline-state doesnt automatically look up the schema with the highest version. atm, it looks if the local state version hash matches remote state version hash (this didnt change, as pipeline state wasnt affected) and next pipeline run that creates a new schema generates a new schema with the same version number.

there can be made some changes so that `pipeline.sync_destination()` (or some other method) fetches the latest schema, but it still breaks the idea that there is just one update path to schema.*

SO: my conclusion was that `Dataset.write()` should be documented as a standalone way to write to the Dataset and the docs should warn about getting it from the pipeline having consequences.

PR #3093, opened 2025-09-16
feat: extend `TTableReference`

zilto:
This is a follow-up to #2941. I accidentally forced push to the branch while the PR was closed, which prevents me from reopening it :/ 

## Since last review
- add "0 to 1" and "0 to many" cardinality
- added a `.references` property to `dlt.Schema`
- use a single cardinality direction, i.e., kept "1 to many" and removed "many to 1"
- update docstrings

## Open question
Should `dlt.Schema.references` also return root, parent, and  `_dlt` tables references? Those are already well-defined and can be created by utilities in `dlt.common.schema.utils`. It could be strange that `dlt.Schema.references` returns a non-empty list while the schema object (i.e., the dictionary) includes no references (because they are implicit).

I would be in favor for including them. They would have the right `cardinality` and `label` (root, parent, _dlt) set out of the box 
  
## Motivation
- This will simplify the work on "smart joins" #2960
- The `table` field on `TTableReference` will make the object self-contained. Currently, you need to maintain a dictionary of `{table_name: TTableReference(...)}` for any meaningful usage
- The `label` field will add context for "semantic layer" features and for LLMs
- The `label` could enable graph database destinations

## Next steps
- Add a top-level namespace `references` to `dlt.Schema`. This would probably require a schema migration?

### DIFF
--- a/tests/common/schema/test_references.py
+++ b/tests/common/schema/test_references.py
@@ -11,8 +11,6 @@ from dlt.common.schema.utils import (
     create_version_and_loads_schema_name_reference,
 )
 
-from tests.utils import preserve_environ, patch_home_dir, autouse_test_storage
-
 
 @pytest.fixture
 def schema() -> dlt.Schema:


rudolfix (changes requested):
LGTM!

question:
1. Should `references` also include nested references via `parent`?

a test for `reference` property is missing

dlt/common/schema/typing.py
```diff
@@ -276,14 +276,88 @@ class TScd2StrategyDict(TMergeDispositionDict, total=False):
...
+
+    referenced_columns: Sequence[str]
+    """Name of the columns(s) from `referenced_table`"""
+
+
+TInlineTableReference = TTableReference
+
+
+# Compared to `TTableReference` or `TInlineTableReference`, `table` is required
+class TStandaloneTableReference(TypedDict, total=False):
```
rudolfix: you could deduplicate code by creating `_TTableReferenceBase` base that contains all common props beside `table` and then just add it in `TTableReference` and `TStandaloneTableReference`. or is there a problem with type checker if you do this?

dlt/common/schema/schema.py
```diff
@@ -605,6 +607,42 @@ def tables(self) -> TSchemaTables:
...
+                pass
+
+            try:
+                root_ref = utils.create_root_child_reference(self.tables, table_name)
+                all_references.append(cast(TStandaloneTableReference, root_ref))
+            except ValueError:
+                pass
+
+            try:
+                load_table_ref = utils.create_root_child_reference(self.tables, table_name)
```
rudolfix: this line is identical to `root_ref = utils.create_root_child_reference(self.tables, table_name)` so it will just duplicate entry in `all_references`? should you use `schema.loads_table_name` here? also foreign key is `_dlt_load_id` so this reference can be hardcoded (providing that `loads_table_name` exists and `_dlt_load_id` is in the `table`)

tests/common/schema/test_references.py
```diff
@@ -0,0 +1,437 @@
...
+    create_version_and_loads_hash_reference,
+    create_version_and_loads_schema_name_reference,
+)
+
+from tests.utils import preserve_environ, patch_home_dir, autouse_test_storage
+
+
+@pytest.fixture
+def schema():
+    return dlt.Schema.from_dict(
```
rudolfix, line 17: up to you, but I save those as yaml files for example
```py
with open("tests/common/cases/schemas/eth/ethereum_schema_v11.yml", "r", encoding="utf-8") as f:
        schema = dlt.Schema.from_dict(yaml.safe_load(f))
```

dlt/common/schema/schema.py
rudolfix: ok you have `create_load_table_reference` I think you wanted to use it here. this also means that tests are incomplete

dlt/common/schema/utils.py
```diff
@@ -996,25 +1019,53 @@ def create_parent_child_reference(tables: TSchemaTables, table_name: str) -> TTa
...
-        raise ValueError(f"Column `{C_DLT_LOAD_ID}` not found for `table_name={table['name']}`")
+        raise ValueError(
+            f"Table `{table['name']}` is not a root table and has no `{C_DLT_LOAD_ID}` column."
+        )
 
     return TTableReference(
+        label="_dlt_load",
+        cardinality="zero_to_many",
+        table=table["name"],
         columns=[C_DLT_LOAD_ID],
```
rudolfix: please never use non-normalized names for tables and columns. you can just paste `schema.naming` here and normalize those constants

tests/common/schema/test_references.py
rudolfix: end to end test is missing: schema.references(self) should produce all expected references that you enumerate below

rudolfix: to weed off naming convention problems: parametrize your test with naming convention. one that certainly produces mixed case identifiers:

```py
@pytest.fixture
def ci_gcp_client(empty_schema: Schema) -> BigQueryClient:
    empty_schema._normalizers_config["names"] = "tests.common.cases.normalizers.title_case"
    empty_schema.update_normalizers()
```

rudolfix (changes requested):
cool PR! but it still requires work on the tests and a few bugs will IMO come out when you improve them

dlt/common/schema/typing.py
zilto: I intended to have `TStandaloneReference` have `Required` type annotations. When using `Required` / `NotRequired`, you can't inherit and override these annotations via inheritance. Eventually, I dropped the usage of `Required` / `NotRequired` because they are poorly supported for Python <3.11 even through `typing_extensions`.

I think your solution is the correct one:
- `_TTableReferenceBase` has common fields (no `total=False` clause, no `Required/NotRequired`)
- `TInlineTableReference` has more optional fields 
- `TStandaloneTableReference` specifies all the required fields
- `TTableReference` aliases to `TInlineTableReference` for backwards compatibility

dlt/common/schema/schema.py
zilto: oops, mistake after copy-pasting

PR #3095, opened 2025-09-17
cli: updated error in the dlt pipeline show command

burnash:
When I run `dlt pipeline my_pipeline show` and I miss dependencies I get this message:

```
You must install additional dependencies to run `dlt pipeline dashboard`. If you use pip you may do the following:

pip install "dlt["workspace"]"

the dlt dashboard requires additional dependencies which you may install with the dlt workspace extra.
```

1. This may be confusing as it looks like command mismatch. The user runs `dlt pipeline my_pipeline show` but the error message refers to dlt pipeline dashboard.
2. The text after the pip command is redundant.

This PR updates error message in `dlt pipeline ... show` command:
- correctly reference the "show" command; 
- removed redundant text about additional dependencies

### DIFF
--- a/dlt/helpers/dashboard/runner.py
+++ b/dlt/helpers/dashboard/runner.py
@@ -6,6 +6,17 @@ from typing import Any
 from pathlib import Path
 from dlt.common.exceptions import MissingDependencyException
 
+
+def _detect_dashboard_command() -> str:
+    command = sys.argv[1]
+    if command == "pipeline":
+        return f"dlt pipeline {sys.argv[2]} show"
+    elif command == "dashboard":
+        return "dlt dashboard"
+    else:
+        raise ValueError(f"Invalid command: {command}")
+
+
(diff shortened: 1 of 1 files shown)


dlt/helpers/dashboard/runner.py
```diff
@@ -13,10 +13,9 @@
     import ibis
 except ModuleNotFoundError:
     raise MissingDependencyException(
-        "dlt pipeline dashboard",
+        "dlt pipeline [pipeline_name] show",
```
burnash:
```suggestion
        "dlt dashboard or dlt pipeline [pipeline_name] show",
```

burnash:
@sh-rp added a small detection function, not it should use the correct command for both options. Could you please take a look?

sh-rp (approved):
LGTM, it's a bit hacky but I think we can do this for now.

PR #3100, opened 2025-09-18
Run common and dashboard tests also with newest available allowed packages for all deps

sh-rp:
### Description
We have tests to test lower bounds of our allowed dependencies. We also want to see if a newer version of any or our dependencies (as allowed in pyproject.toml) breaks anything in dlt.

### DIFF
--- a/tests/common/schema/test_coercion.py
+++ b/tests/common/schema/test_coercion.py
@@ -32,9 +32,10 @@ def test_coerce_type_to_text() -> None:
     # bytes to text (base64)
     assert coerce_value("text", "binary", b"binary string") == "YmluYXJ5IHN0cmluZw=="
     # HexBytes to text (hex with prefix)
-    assert (
-        coerce_value("text", "binary", HexBytes(b"binary string")) == "0x62696e61727920737472696e67"
-    )
+    assert coerce_value("text", "binary", HexBytes(b"binary string")) in [
+        "0x62696e61727920737472696e67",
+        "62696e61727920737472696e67",
+    ]
 
     # Str enum value
     class StrEnum(Enum):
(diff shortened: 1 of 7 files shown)


tests/common/schema/test_coercion.py
```diff
@@ -32,9 +32,10 @@ def test_coerce_type_to_text() -> None:
     # bytes to text (base64)
     assert coerce_value("text", "binary", b"binary string") == "YmluYXJ5IHN0cmluZw=="
     # HexBytes to text (hex with prefix)
-    assert (
-        coerce_value("text", "binary", HexBytes(b"binary string")) == "0x62696e61727920737472696e67"
-    )
+    assert coerce_value("text", "binary", HexBytes(b"binary string")) in [
```
sh-rp, line 35: A newer version of hexbytes returns the hexcode without "0x". We have a couple of options:
* Keep as is, this means users will see other values in their destination after upgrading this lib
* Normalize result in normalizer so the result is the same regardless of lib version (kind of a breaking change, either for users on the old version or the new version)
* Update min version of hexbytes to the version with the new output. Also a kind of breaking change.

Imho this is probably an edge case. I think I would probably keep it as is.

rudolfix: HexBytes was used to make sure that 0x is present... so if that got removed it does not make sense to use it. My take: write our own HexBytes that alwasy prefixes and parses the prefix. it should be pretty simple to do.

for now I'd merge this ticket

rudolfix (approved):
@sh-rp let's merge it. I always wanted to test of max dependencies too.

re. HexBytes: see my comment I bet @alkaline-0 can port this to dlt and remove the dependency :)

PR #3101, opened 2025-09-18
Updates CONTRIBUTING.md and README.md to remove outdated information and add more info

sh-rp:
### Description

Updates CONTRIBUTING.md:
* Formatting and grammar checking
* Remove outdated information about testing 
* Add information about CI setup
* Update most testing sections with tips and tricks

Updates README.md: 
* Add links to release list, wasm playground and code examples section on website


CONTRIBUTING.md
```diff
@@ -49,176 +45,243 @@ When you're ready to contribute, follow these steps:
 3. Write your code and tests.
 4. Lint your code by running `make lint` and test common modules with `make test-common`.
 5. If you're working on destination code, contact us to get access to test destinations.
-6. If you have added, removed or updated dependencies in the `pyproject.toml`, make sure `uv.lock` is up to date by running `uv lock`. If you are merging in upstream changes from the devel branch and get a conflict on the lockfile, it's best to take the devel version and re-run `uv lock` to re-apply your changes.
-7. Create a pull request targeting the **devel** branch of the main repository.
-
-**Note:** for some special cases, you'd need to contact us to create a branch in this repository (not fork). See below.
+6. If you’ve added, removed, or updated dependencies in `pyproject.toml`, make sure `uv.lock` is up to date by running `uv lock`.
```
anuunchin, line 48: I understood that if `uv.lock` changes even though there are no new dependencies it's still fine - maybe we should mention this somewhere 👀

CONTRIBUTING.md
```diff
@@ -49,176 +45,243 @@ When you're ready to contribute, follow these steps:
...
 
-### Active branches
+**Note:** In some special cases, you’ll need us to create a branch in this repository (not in your fork). See below.
 
-We use **devel** (which is our default Github branch) to prepare a next release of `dlt`. We accept all regular contributions there (including most of the bugfixes).
+### Active Branches
 
-We use **master** branch for hot fixes (including documentation) that needs to be released out of the normal schedule.
+- **devel** (default GitHub branch): Used to prepare the next release of `dlt`. We accept all regular contributions here (including most bug fixes).  
+- **master**: Used for hotfixes (including documentation) that must be released outside of the normal schedule.
```
anuunchin, line 57: omg, i somehow wasn't aware that we have an actual schedule 👀  - should we mention what the schedule is?

CONTRIBUTING.md
```diff
@@ -49,176 +45,243 @@ When you're ready to contribute, follow these steps:
...
+
+- You can also select which destination tests to run using the `ACTIVE_DESTINATIONS` and `ALL_FILESYSTEM_DRIVERS` environment variables. The former selects destinations to use, while the latter determines which buckets to use for the filesystem destination and staging destinations. For example, the command `ACTIVE_DESTINATIONS='["duckdb", "filesystem"]' ALL_FILESYSTEM_DRIVERS='["memory", "file"]' uv run pytest tests/load` will run all loader tests on DuckDB and the filesystem (in-memory filesystem and local files). You can see these environment variables being used in our workflow setup.
+
+
 
 ## Local Development
 
-Use Python 3.9 for development, as it's the lowest supported version for `dlt`. You can select (and if needed download) the python version you need with `uv venv --python 3.11.6`, [uv python version docs](https://docs.astral.sh/uv/concepts/python-versions/#managed-and-system-python-installations).
+Use Python 3.9 for development, as it is the lowest supported version. You can select (and download if necessary) the version with:
```
anuunchin, line 232: Should we replace 3.9 with 3.10 right away? As @zilto  mentioned, understood that it won't be managed very soon and therefore we would stop supporting it as well 👀

README.md
```diff
@@ -107,6 +107,8 @@ We suggest that you allow only `patch` level updates automatically:
 * Using the [Compatible Release Specifier](https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release). For example **dlt~=1.0** allows only versions **>=1.0** and less than **<1.1**
 * Poetry [caret requirements](https://python-poetry.org/docs/dependency-specification/). For example **^1.0** allows only versions **>=1.0** to **<1.0**
```
anuunchin, line 108: Do we still use poetry somewhere 👀  or should we remove this part

anuunchin (changes requested):
🧠

CONTRIBUTING.md
zilto: `uv.lock` should only change if you update `pyproject.toml`. The lockfile guarantees that "the code works with these exact dependencies" and allows devs to work using the same deps.

Now, imagine you have in pyproject.toml `duckdb >= 1.0.0; < 2.0.0`. When you install or build `dlt` locally via `pip install .`, you will get the latest `duckdb` (let's say 1.4.0) which may be different from the `uv.lock` which has 1.3.2. You don't need to update the `uv.lock` because any version between 1 and 2 should be valid.

^to avoid this situation, you should `uv sync` instead of pip installing (this is what `make dev` does)

README.md
sh-rp: This section is about installing dlt in your project, so it does not apply to using uv in our repo, but to any tool the end-user uses to manage dependencies, it could be plain pip, uv or poetry for example.

CONTRIBUTING.md
sh-rp: Probably not, since we kind of have a schedule but often don't manage to adhere to it :)

CONTRIBUTING.md
sh-rp: I think i would do this all in one go. Mid october 3.9. will be deprecated and 3.14 will be released and we should do a PR that handles both.

PR #3111, opened 2025-09-22
Fix: `sql_client.raise_database_error` creates circular `__cause__` dependency

### LINKED ISSUE #3102: `sql_client.raise_database_error` creates circular `__cause__` dependency
### dlt version

`>= 1.15.0` (first version I've noticed it, could be earlier)

### Describe the problem

I have a scheduler that, when exception occurs, traverses through the exceptions `__cause__`s for error handling (trying to find the root cause of the exception).
What I've noticed in newer version of `dlt` is that it falls into an infinite recursion loop, likely because the aforementioned `raise_database_error` creates a recursive dependency.

I believe an equivalent of:
```python
try:
    ...
except Exception as ex:
    raise ex from ex
```

is happening, and given [this line](https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/duckdb/sql_client.py#L386) it does seem like a possibility.

### Expected behavior

Exceptions without circular `__cause__` dependencies

### Steps to reproduce

I found this in one of the prod workflows hence don't have a reproducible example at hand. I might try creating one if needed but I believe it's clear enough from the explanation and can be fixed without a repro example, by inspecting the code + implementing a proper unit test.

### Operating system

Linux

### Runtime environment

Other

### Python version

3.12

### dlt data source

_No response_

### dlt destination

MotherDuck

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
This PR simply stops the `raise_database_error` from creating circular __cause__ dependencies. An appropriate minimal test is added.

Resolves #3102.

### DIFF
--- a/dlt/destinations/sql_client.py
+++ b/dlt/destinations/sql_client.py
@@ -446,7 +446,10 @@ def raise_database_error(f: TFun) -> TFun:
             return (yield from f(self, *args, **kwargs))
         except Exception as ex:
             db_ex = self._make_database_exception(ex)
-            raise db_ex.with_traceback(ex.__traceback__) from ex
+            if db_ex is ex:
+                raise db_ex.with_traceback(ex.__traceback__)
+            else:
+                raise db_ex.with_traceback(ex.__traceback__) from ex
 
     @wraps(f)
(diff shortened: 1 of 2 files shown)


rudolfix (approved):
LGTM!

PR #3117, opened 2025-09-23
Feat: Custom metrics in the incremental transform

anuunchin:
This PR adds the following three custom metrics to the incremental transform:

- `unfiltered_items_count`
- `unfiltered_batches_count`
- `unique_hashes_count`

Specifically, in the `call` dunder of the `Incremental` class. 

> Since it is the incremental wrapper is the step that is used when metrics are computed the custom_metrics property of the wrapper just returns the custom metrics from the incremental itself.

An appropriate test was added.

### DIFF
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -618,10 +618,6 @@ class Incremental(
         # writing back state
         self._cached_state["last_value"] = transformer.last_value
 
-        initial_hash_list = self._cached_state.get("unique_hashes")
-        initial_hash_count = len(initial_hash_list) if initial_hash_list else 0
-        self.custom_metrics["initial_unique_hashes_count"] = initial_hash_count
-
         if transformer.boundary_deduplication:
             # compute hashes for new last rows
             # NOTE: object transform uses last_rows to pass rows to dedup, arrow computes
@@ -630,6 +626,10 @@ class Incremental(
                 transformer.compute_unique_value(row, self.primary_key)
                 for row in transformer.last_rows
             )
+            initial_hash_list = self._cached_state.get("unique_hashes")
+            initial_hash_count = len(initial_hash_list) if initial_hash_list else 0
+            self.custom_metrics["initial_unique_hashes_count"] = initial_hash_count
+
             # add directly computed hashes
             unique_hashes.update(transformer.unique_hashes)
             self._cached_state["unique_hashes"] = list(unique_hashes)
@@ -637,6 +637,8 @@ class Incremental(
             self.custom_metrics["final_unique_hashes_count"] = final_hash_count
 
             self._check_duplicate_cursor_threshold(initial_hash_count, final_hash_count)
+        else:
+            self._cached_state["unique_hashes"] = []
         return rows
 
     def _check_duplicate_cursor_threshold(
--- a/tests/extract/test_incremental.py
+++ b/tests/extract/test_incremental.py
@@ -4444,34 +4444,39 @@ def test_custom_metrics_in_incremental() -> None:
     load_id = _run_with_items([{"id": 3, "value": "3"}, {"id": 4, "value": "4"}], False)
     _assert_custom_metrics(load_id, 5, 4, 1, 1, 1)
 
-    # 4. run with duplicate cursor field values, but different hashes, as batch
+    # 4. run with duplicate cursor field values, but different hashes, as a single batch
     load_id = _run_with_items(
         [{"id": 5, "value": "5.1"}, {"id": 5, "value": "5.2"}, {"id": 5, "value": "5.3"}], True
     )
     _assert_custom_metrics(load_id, 8, 5, 1, 3, 3)
 
-    # 5. run with the same values as batch from previous run, but with no boundary deduplication
+    # 5. reset incremental with no boundary deduplication (primary_key=()) and run with the same values
+    # from previous run, should be loaded as a single batch with 3 items
     resource_with_metrics.apply_hints(
         incremental=dlt.sources.incremental(cursor_path="id", initial_value=-1, primary_key=())
     )
     load_id = _run_with_items(
         [{"id": 5, "value": "5.1"}, {"id": 5, "value": "5.2"}, {"id": 5, "value": "5.3"}], True
     )
-    _assert_custom_metrics(load_id, 3, 1, 3, 0, 3)
+    _assert_custom_metrics(load_id, 3, 1, 0, 0, 3)
 
-    # 6. run with two new items as a single batch
-    load_id = _run_with_items([{"id": 6, "value": "6.1"}, {"id": 6, "value": "6.2"}], True)
-    _assert_custom_metrics(load_id, 5, 2, 3, 0, 2)
+    # 6. run with one old and one new item as a single batch (still no boundary deduplication)
+    # should be loaded as a single batch with 2 items
+    load_id = _run_with_items([{"id": 5, "value": "5.1"}, {"id": 6, "value": "6.1"}], True)
+    _assert_custom_metrics(load_id, 5, 2, 0, 0, 2)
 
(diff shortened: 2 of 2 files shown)


dlt/extract/incremental/__init__.py
```diff
@@ -787,6 +797,13 @@ def allow_external_schedulers(self, value: bool) -> None:
         if self._incremental:
             self._incremental.allow_external_schedulers = value
 
+    @property
```
anuunchin, line 825: Will we ever collect metrics in the wrapper separately? 🤔  This is probably wrong, but I'm not 100% sure how the logic should work: 

Maybe, the wrapper's custom metrics should be merged with the metrics of self._incremental 👀  @rudolfix

rudolfix: this makes sense. but overall wrapped is a complication. wrapper is used to be an immutable object that is attached to incremental argument in the resource function as default. but I think we could get rid of this somehow

dlt/extract/incremental/__init__.py
```diff
@@ -567,6 +569,15 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
         if rows is None or (isinstance(rows, list) and len(rows) == 0):
```
rudolfix, line 593: metrics should be populated in this case as well.

dlt/extract/incremental/__init__.py
```diff
@@ -567,6 +569,15 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
...
             return rows
 
+        # collect metrics
+        self.custom_metrics["unfiltered_items_count"] = self.custom_metrics.get(
```
rudolfix: you can create metrics collection in `__init__`. I'd also type the metric dict with `TypedDict`. this will also solve the problem above

dlt/extract/incremental/__init__.py
```diff
@@ -567,6 +569,15 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
...
+            "unfiltered_items_count", 0
+        ) + count_rows_in_items(rows)
+        self.custom_metrics["unfiltered_batches_count"] = (
+            self.custom_metrics.get("unfiltered_batches_count", 0) + 1
+        )
+        self.custom_metrics["unique_hashes_count"] = len(self.get_state().get("unique_hashes", []))
```
rudolfix: 1. hash metrics should be updated after step runs. see `if transformer.boundary_deduplication:` it already counts the hashes. makes sense to store initial and final value see code
2. your code always create empty list, if statement would be better

dlt/extract/incremental/__init__.py
```diff
@@ -588,6 +618,10 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
         # writing back state
         self._cached_state["last_value"] = transformer.last_value
 
+        initial_hash_list = self._cached_state.get("unique_hashes")
+        initial_hash_count = len(initial_hash_list) if initial_hash_list else 0
+        self.custom_metrics["initial_unique_hashes_count"] = initial_hash_count
+
```
anuunchin: @rudolfix either I understood the purpose wrong, or `initial_unique_hashes_count` is somewhat unnecessary if it just gets the unique hashes from the pipeline state 👀

tests/extract/test_incremental.py
```diff
@@ -4372,3 +4372,106 @@ def resource(
...
+        incremental=dlt.sources.incremental(cursor_path="id", initial_value=-1, primary_key=())
+    )
+    load_id = _run_with_items(
+        [{"id": 5, "value": "5.1"}, {"id": 5, "value": "5.2"}, {"id": 5, "value": "5.3"}], True
+    )
+    _assert_custom_metrics(load_id, 3, 1, 3, 0, 3)
+
+    # 6. run with two new items as a single batch
+    load_id = _run_with_items([{"id": 6, "value": "6.1"}, {"id": 6, "value": "6.2"}], True)
+    _assert_custom_metrics(load_id, 5, 2, 3, 0, 2)
```
anuunchin: @rudolfix  my intuition here was to get 0 as the initial_unique_hashes_count, but since it's retrieved from the state it makes sense. However, if the boundary deduplication is off and unique hashes becomes 0, shouldn't this also reset the hash count in state ? 👀

dlt/extract/incremental/__init__.py
rudolfix: I do not understand this.
1. We have `final_unique_hashes_count` in state after the pipeline runs
2. State and trace are stored/sent to different places. Metrics are part of trace and will be sent to some alerting systems where state is not available

dlt/extract/incremental/__init__.py
```diff
@@ -588,6 +618,10 @@ def __call__(self, rows: TDataItems, meta: Any = None) -> Optional[TDataItems]:
         # writing back state
         self._cached_state["last_value"] = transformer.last_value
 
+        initial_hash_list = self._cached_state.get("unique_hashes")
```
rudolfix: This should happen after:
`if transformer.boundary_deduplication:`

but you are right regarding edge case in the comment in the test: we should reset unique hashes:
```py
if transformer.boundary_deduplication:
...
else:
.... # reset uniqye hashes
```

dlt/extract/incremental/__init__.py
```diff
@@ -625,7 +659,7 @@ def _check_duplicate_cursor_threshold(
 TIncrementalConfig = Union[Incremental[Any], IncrementalArgs]
 
 
-class IncrementalResourceWrapper(ItemTransform[TDataItem]):
+class IncrementalResourceWrapper(ItemTransform[TDataItem, IncrementalCustomMetrics]):
```
rudolfix, line 664: Yeah I didn't think about that... But yeah now we must make ItemTransform Generic over metric type

tests/extract/test_incremental.py
rudolfix: yes, it should see my comment above

tests/extract/test_incremental.py
```diff
@@ -4372,3 +4372,106 @@ def resource(
...
+    resource_with_metrics.incremental.primary_key = "id"
+    load_id = _run_with_items({"id": 7, "value": "7"}, True)
+    _assert_custom_metrics(load_id, 6, 3, 3, 1, 1)
+
+    # 8. run with None within a batch -> should increment unfiltered_items_count
+    load_id = _run_with_items([None, {"id": 8, "value": "8"}], True)
+    _assert_custom_metrics(load_id, 8, 4, 1, 1, 1)
+
+    # 9. run with None as a single batch -> should not increment unfiltered_items_count
+    load_id = _run_with_items([None, {"id": 9, "value": "9"}], False)
```
rudolfix: you pass `False` as `as_batch` is this really correct?

anuunchin: I don't think this is correct nor incorrect -> The only reason I put `False`, that is, "not as batch" here, is just because the previous few runs were run as batch. TLDR, for demonstration purposes, so the next person understand what's going on 👀 


OR, did you mean the assertions are incorrect? Which would mean you're suggesting that if
```
[None, {"id": 9, "value": "9"}]
```
is not passed as a single batch, 
then we expect the total unfiltered items count, as well as the unfiltered batches count, each to be incremented by 2, not 1?  👀

dlt/extract/incremental/__init__.py
anuunchin: indeed, i get it now 🫡

tests/extract/test_incremental.py
anuunchin: okay, I agree that it doesn't make sense because None as a single batch doesn't even reach incremental, removed ✅

PR #3122, opened 2025-09-24
QoL: accept destination name as shorthand form of destination 

### LINKED ISSUE #3109: accept destination name as shorthand form of destination
**Background**
When defining pipeline
```py
pipeline = dlt.pipeline("pipeline", destination="duckdb")
```
we use destination types as shorthands ie. `duckdb`. In that case the configuration entry looks like:
```toml
[destination.duckdb]
database="local.db"
```
This becomes a problem when moving from local to production because code above contains hardcoded pipeline type. One of our guiding principle is that the same code works locally, in tests and in production. This is not the case here. see ie. https://dlthub.com/docs/devel/walkthroughs/share-a-dataset

We have a mechanism to do that: named destinations. 
https://dlthub.com/docs/devel/general-usage/destination#pass-explicit-parameters-and-a-name-to-a-destination
but it still requires to use destination factories.

The idea here is to extend shorthand notation to recognize named destinations. So for example:
```py
pipeline = dlt.pipeline("pipeline", destination="warehouse")
```

will resolve
```toml
[destination.warehouse]
destination_type="duckdb"
database="local.db"
```
and
```toml
[destination.warehouse]
destination_type="bigquery"
project_id="...."

```
Note that we need `destination_type` to be able to resolve such configuration. see below

**Implementation**
We need a fallback in `Destination::from_reference` that will try to resolve `destinaton_type` config field if a regular reference was not found. In example above:
`destination.warehouse.destination_type` should be resolved and if that succeeded, `from_reference` should be called again, this time passing `destination_name` and `destination_type`

obviously
- if `destination_name` is provided, no fallback is executed
- if `ref` contains dots it cannot be a valid name so no fallback

anuunchin:
This PR extends the `Destination.from_reference` so that if it encounters an `UnknownDestinationModule` exception, it tries to resolve the destination type from configs based on the destination name and then use the resolved type to build the destination. An appropriate test is added. 

See ticket for more context, but the main reason for this is the following:

As of now, when moving from local to production, people still need to change the destination in the code itself: 
For example, locally they run with duckdb:
```py
pipeline = dlt.pipeline("pipeline", destination="duckdb")
```
Then, in production, they need to run with Bigquery and therefore need to adjust:
```py
pipeline = dlt.pipeline("pipeline", destination="bigquery")
```

This PR enables the user just to set a destination name and control type in configs. Say, the user has:
```py
pipeline = dlt.pipeline("pipeline", destination="warehouse") 
```
They don't need to change the actual pipeline code and can simply control the destination type from configs:
```toml
[destination.warehouse]
#destination_type="duckdb" 
destination_type="bigquery"
```

Resolves #3109

### DIFF
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -279,8 +279,8 @@ class Destination(ABC, Generic[TDestinationConfig, TDestinationClient]):
         destination_type: str = dlt.config.get(f"destination.{ref}.destination_type")
         if destination_type:
             try:
-                return DestinationReference.from_name(
-                    destination_type=destination_type,
+                return DestinationReference.from_reference(
+                    ref=destination_type,
                     destination_name=ref,
                     credentials=credentials,
                     environment=environment,
@@ -295,13 +295,9 @@ class Destination(ABC, Generic[TDestinationConfig, TDestinationClient]):
                 ref, credentials, destination_name, environment, **kwargs
             )
         except UnknownDestinationModule as e:
-            raise UnknownDestinationModule(
-                ref=e.ref,
-                qualified_refs=e.qualified_refs,
-                traces=e.traces,
-                destination_type=destination_type,
-                named_dest_attempted=True,
-            ) from e.__cause__
+            e.destination_type = destination_type
+            e.named_dest_attempted = True
+            raise e
 
 
 class DestinationReference:
@@ -410,27 +406,6 @@ class DestinationReference:
         if environment:
             kwargs["environment"] = environment
         return factory(**kwargs)
-
-    @classmethod
-    def from_name(
-        cls,
-        destination_type: str,
-        destination_name: str,
-        credentials: Optional[Any] = None,
-        environment: Optional[str] = None,
-        **kwargs: Any,
-    ) -> Optional[AnyDestination]:
-        """Instantiate destination from a destination type and name.
-        This method creates a named destination by using the destination type
-        as a factory and instantiating it with the provided destination name.
-        """
-        return DestinationReference.from_reference(
-            ref=destination_type,
-            credentials=credentials,
-            destination_name=destination_name,
-            environment=environment,
-            **kwargs,
-        )
 
     @staticmethod
     def normalize_type(destination_type: str) -> str:


dlt/common/destination/client.py
```diff
@@ -142,6 +142,11 @@ def from_normalized_mapping(
         )
 
 
+@configspec
+class DestinationTypeConfiguration(BaseConfiguration):
+    destination_type: Optional[str] = None
+
+
```
anuunchin: @rudolfix  I'm not sure if this is the proper way to do resolve the destination_type 👀

tests/common/destination/test_reference.py
```diff
@@ -308,6 +308,48 @@ def test_import_destination_config() -> None:
...
+            ref=f"dlt.destinations.{destination_type}", destination_name=destination_name
+        )
+        assert dest.destination_type == "dlt.destinations.duckdb"
+        assert dest.destination_name == destination_name
+
+        dest = Destination.from_reference(ref=f"dlt.destinations.{destination_type}")
+        assert dest.destination_type == "dlt.destinations.duckdb"
+        assert dest.destination_name == "duckdb"
+
+
```
anuunchin, line 384: @rudolfix  perhaps there should also be an end-to-end test changing destination type in configs (simulating going from local to prod) 👀

dlt/common/destination/client.py
rudolfix: yes this is good. the only problem I see is that it will be aware of
`destination.destination_type` and `destination_type` configs as well. but that's OK for now. I plan to add the traces list to resolved configuration object so we can filter those out

dlt/common/destination/reference.py
```diff
@@ -267,9 +271,30 @@ def from_reference(
...
+                ref, credentials, destination_name, environment, **kwargs
+            )
+        except UnknownDestinationModule:
+            if destination_name or "." in ref:
+                raise
+
+            resolved_config = resolve_configuration(
+                DestinationTypeConfiguration(),
+                sections=(known_sections.DESTINATION, ref),
+                accept_partial=True,
```
rudolfix: you do not need this. you marked the field as Optional so it will not raise

dlt/common/destination/client.py
anuunchin: 🫡

dlt/common/destination/reference.py
anuunchin: owwr, right 🫡

zilto:
This is a very cool feature! Having named destinations mirrors defining objects in Python nicely.

Though, this implementation can produce problems for our future selves.

## Problem
1. It's 2025-12-01, Alex creates a named destination using this feature called "aws-lake". It works as expected and retrieves the `destination_type="iceberg"` from its `config.toml`
2. It's 2026-01-01 and AWS announces their new open-table format AWS Lake
3. dltHub adds support for AWS Lake with shorthand notation `aws-lake`
4. Alex's pipeline that had named destination `aws-lake` silently broke because of the dltHub latest update

## Solution
We need to explicitly namespace named destinations. Could be `destination="dlt.aws-lake"` or whatever. This way, new destination never collide with custom names.

Also, it will improve readability for users who are not aware of named destinations. For instance, looking at `destination="aws-lake"`, I would go browse the dlt docs and would nothing and scratch my head.

rudolfix:
@zilto OK I see the point. I'm not so much afraid of name clashes but making it explicit that destination string is a name makes sense.

what about `name:aws-lake`?

dot notation is already taken to refer to destinations from outside of `dlt` ie `my_module.bigquery` will try to instantiate `bigquery` factory class from `my_module`. so `dlt.aws-lake` is just invalid reference.

@anuunchin I'll do review for other parts of code. if we'll go with this change it'll get even simpler.

rudolfix (changes requested):
I do not see end to end test. Where you create actual pipeline that uses destination name during instantiation.

> pipeline using destination name that is configured and not configured

you can add this test to `tests/pipeline/test_pipelines.py`. you have `config.toml` in `tests/.dlt/config.toml` to add named destination there. it is ok to test on filesystem or duckdb

zilto:
> @zilto OK I see the point. I'm not so much afraid of name clashes but making it explicit that destination string is a name makes sense.
> what about name:aws-lake?

I think it makes sense! Maybe `"@aws-lake"` would look prettier. Any special character should work.

## `dlt.destination` entity
> I really don't want to upscope this change, but these notes seem relevant for future work


### Currently
We don't have a named destination constructor. Typically, you would use:

```python
destination = dlt.destinations.duckdb(**kwargs)
```

Though, you could dig into internal methods to do 

```python
from dlt.common.destination.reference import Destination

destination = Destination.from_reference("duckdb")
```

### Changes
The current changes muddy the logic of `Destination.from_reference()` because we're effectively passing a `destination_name` as arg, not the `reference`.

Maybe we add another static method `Destination.from_name()`. 

Maybe we could introduce a simpler constructor like

```python
# though `dlt.destination` is already taken by `@dlt.destination`
# other kwargs are automatically filled from config if empty
destination = dlt.destination("local_db")
```
That would allow the cleanest API with the best IDE support
```python
# this shows it uses the `dlt.destination()` resolution logic
dlt.pipeline("foo", destination=dlt.destination("local_db"))
# OR this is unambiguously a "named destination"
# it won't clash with `"duckdb"`
dlt.pipeline("foo", destination_name="local_db")
```


Thinking more about this, it suggests that we want to add a "named destination" constructor. Currently, we have named pipelines and named datasets:

```python
pipeline = dlt.pipeline("foo")
# destination, dataset
dataset = dlt.dataset("duckdb", "foo")
# now we could even
dataset = dlt.dataset("@local_db", "foo")
```

dlt/common/destination/reference.py
```diff
@@ -267,9 +271,29 @@ def from_reference(
...
+            if not destination_type:
+                raise
+
+            return DestinationReference.from_reference(
+                ref=resolved_config.destination_type,
+                credentials=credentials,
+                destination_name=ref,
+                environment=environment,
+                **kwargs,
+            )
```
zilto: The static method `.from_reference()` already has a lot of logic. If we want to keep a single entrypoint, I suggest having

```python
@staticmethod
def from_name(...) -> DestinationReference:
  # the new logic
  
@staticmethod
def from_reference(...) -> DestinationReference:
   if ref.startswith("@"):
      return DestinationReference.from_name(...)
      
   # old logic   
     
```

rudolfix:
I think it is good idea to add `destination = dlt.destination("local_db")` constructor. We can still implement it together with the decorator. There's already `dlt.resource` that can be used both as decorator and resource constructor from any data - this won't be unusual. It will be pretty easy to implement that. Additional benefit is to remove `Destination.from_reference` from user facing docs. This constructor should handle all the documented cases. Let's do it in a separate ticket. (I already have something that works, typechecks and has proper docstrings)

Now I'm not sure I really want to have `@name`. It seems that `@` prefix is some kind of a new convention that we must explain and I do not see any other applications.  After some deliberations I think we should just go with `name` without any prefix:

shorthand reference (type) ie. `bigquery` is also a destination name if explicit name is not provided. For example:

```toml
[destination.bigquery]
```
and
```toml
[destination.bigquery]
destination_type="bigquery"
```
are equivalent and

```toml
[destination.warehouse]
destination_type="bigquery"
```
just follows the same pattern. So I think using `destination="warehouse"` is not confusing or less confusing that new `@warehouse`.

The name clash (`aws-lake`) can be solved by checking if named destination config exists before resolving shorthand type

TLDR;> @anuunchin 

* `dlt.destination` moved to a separate ticket
* for shorthand (no dot) names we look for named destination first in the config and if not found we fallback to shorthand type reference.
* now we can do procedure similar to this:
```py
@staticmethod
def from_name(...) -> DestinationReference:
  # the new logic
  
@staticmethod
def from_reference(...) -> DestinationReference:
   if ref.startswith("@"):
      return DestinationReference.from_name(...)
      
   # old logic   
```
because we check named destination first so maybe the code simplifies.

zilto:
@rudolfix I agree with the design choices! I think it's the cleanest option given it doesn't introduce anything new. 

I want to point out the PR should additionally:
1. update the docs on config resolution order
2. check that the config resolution exceptions properly tells the user the resolution order:
   - "bigquery named destination -> biquery destination type`
   - "warehouse named destination -> None"

dlt/common/destination/reference.py
```diff
@@ -267,9 +272,37 @@ def from_reference(
...
 
-        return DestinationReference.from_reference(
-            ref, credentials, destination_name, environment, **kwargs
-        )
+        # If destination_name is not provided and ref does not contain dots
+        # try to resolve named destination with destination type
+        named_dest_error = None
+        if not destination_name and "." not in ref:
+            try:
+                return DestinationReference.from_name(
```
rudolfix: I think it is too complicated. What we need to do (assuming we try the named option first).
1. try to resolve the name into type: `configured_destination_type = dlt.secrets.get("destination.{ref}.destination_type")` IMO we do not need to resolve full config. we can be very opinionated and require that full layout is used.
2. if resolves, pass it as ref to a regular `DestinationReference.from_reference` (and OFC input `ref` as `destination_name`). 

am I missing something here?

our complication is an user friendly exception message if ref not resolves at the end and if `ref` does not have a `.`
case 1: 
standard message: we were not able to resolve `configured_destination_type`: we keep current message but add a note that if `ref` is a named destination you must add `destination_type=...` to `destination.{ref}` basically instructing user what to do to fix the problem

case 2:
we were able to resolve `configured_destination_type` but it contains wrong type :) then we need to add a note where the resolved ref came from

case 3. (optional) the thing that you pointed out: if we resolved `configured_destination_type` but the original `ref` shadows something in the registry `DesintationReference.DESTINATIONS.get(ref_)` we can issue a warning. but maybe that's too much for this PR

anuunchin: Using `dlt.secrets.get("destination.{ref}.destination_type")` is definitely way cleaner 🫡 , I removed the unnecessary config class and the new exception. Things seem to be simpler now + error message is way more understandable (left the optional one for the next PR because docs mention the behavior)

dlt/common/destination/reference.py
```diff
@@ -267,9 +268,40 @@ def from_reference(
...
+            except Exception:
+                pass
+
+        # Then, try to resolve as a shorthand destination ref
+        try:
+            return DestinationReference.from_reference(
+                ref, credentials, destination_name, environment, **kwargs
+            )
+        except UnknownDestinationModule as e:
+            raise UnknownDestinationModule(
```
rudolfix: you could just set the property `named_dest_attempted` on existing exception and re-reraise. we do it in a few places.

dlt/common/destination/reference.py
```diff
@@ -411,3 +410,57 @@ def from_reference(
         if environment:
             kwargs["environment"] = environment
         return factory(**kwargs)
+
+    @classmethod
+    def from_name(
```
rudolfix: this function is just calling `from_reference` with a little bit different argument order right? we removed config check to `Destination` class. so I think we can just remove this method

dlt/common/destination/reference.py
anuunchin: owwwww, right

dlt/common/destination/reference.py
anuunchin: hmmmm, thought we would still have it in the follow-up PR 👀  - but removed anyway 🫡

PR #3131, opened 2025-09-24
Fix: The child table column remains in the schema as a partial column with seen-null-first=True

### LINKED ISSUE #3125: The child table column remains in the schema as a partial column with seen-null-first=True
### dlt version

Latest

### Describe the problem

When a column has only None values initially, it is saved in the schema as a partial column with `x-normalizer`.`seen-null-first` set to `True`.  In the end, when data arrives as a nested structure in this column, a child table is created, but the partial column schema that was created is not removed from the schema.

### Expected behavior

The column schema should be removed if it was created as a child table.

### Steps to reproduce

This test should pass:
```py
def test_empty_column_later_becoming_child_table_removed() -> None:
    name = "schema_test" + uniq_id()
    p = dlt.pipeline(
        pipeline_name=name,
        destination=dummy(completed_prob=1),
        export_schema_path=EXPORT_SCHEMA_PATH,
    )
    test = p.default_schema.naming.max_length
    @dlt.resource(table_name="my_table")
    def nested_data():
        nested_example_data = EXAMPLE_DATA[0]
        nested_example_data["children"] = None
        yield nested_example_data
    p.run(nested_data())
    @dlt.resource(table_name="my_table")
    def nested_data():
        nested_example_data = EXAMPLE_DATA[0]
        nested_example_data["children"] = [{"id": 2, "name": "Max"}, {"id": 3, "name": "Julia"}]
        yield nested_example_data
    p.run(nested_data())
    export_schema = _get_export_schema(name)
    assert "children" not in export_schema.tables["my_table"]["columns"]
    assert "my_table__children" in export_schema.tables
    assert "children" not in p.default_schema.tables["my_table"]["columns"]
    assert "my_table__children" in p.default_schema.tables
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.10

### dlt data source

Affects all sources

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
This PR adjusts the `_normalize_chunk` function of the json normalizer the following way:

- The schema_update is checked for child table updates.
- If a child table update corresponds to a previously created partial column, that partial column is removed from parent.

A test with long column names was added.

Resolves #3125


dlt/normalize/items_normalizers.py
```diff
@@ -518,6 +519,20 @@ def _normalize_chunk(
...
+                if not is_nested_table(update):
+                    continue
+
+                parent_name = update.get("parent")
+                parent_columns = self.schema.get_table_columns(parent_name, include_incomplete=True)
+
+                column = parent_columns.get(last_ident_path)
+                if column and utils.has_seen_null_first_hint(column):
+                    parent_columns.pop(last_ident_path)
+
```
anuunchin: This location seems to be the only location that makes sense to do this as of now because outside of this we don't have access to full ident paths 👀

dlt/normalize/normalize.py
```diff
@@ -173,11 +173,10 @@ def clean_x_normalizer(self, load_id: str, table_name: str, table_schema: TTable
         # Handle column-level x-normalizer
         # drop seen-null-first flag if data type was set
         for column in table_schema.get("columns", {}).values():
-            col_x_normalizer = column.setdefault("x-normalizer", {})
-            if col_x_normalizer.get("seen-null-first") and "data_type" in column:
-                col_x_normalizer.pop("seen-null-first", None)
-            if not col_x_normalizer:
-                column.pop("x-normalizer", None)
+            if has_seen_null_first_hint(column) and "data_type" in column:
```
rudolfix: OK so we do same thing in two places. maybe move this to the worker as well?

dlt/normalize/items_normalizers.py
```diff
@@ -518,6 +519,20 @@ def _normalize_chunk(
             except StopIteration:
                 pass
             signals.raise_if_signalled()
+
+        for table_name, updates in schema_update.items():
```
rudolfix: pls extract this code to a separate method and comment what it does

tests/pipeline/test_import_export_schema.py
```diff
@@ -246,3 +247,66 @@ def nested_data():
...
+    }
+    assert (
+        export_schema.tables[nested_tbl_name]["columns"]
+        .get(norm_col_name_b)["x-normalizer"]
+        .get("seen-null-first", False)
+    )
+
+    p.run(nested_data(with_grandchild=True))
+
+    export_schema = _get_export_schema(name)
```
rudolfix, line 314: please comment this test a little. we expect `norm_col_name_b` to be removed, right? explain why

rudolfix (changes requested):
forgot to request changes

dlt/normalize/normalize.py
anuunchin: 🫡 , I put the entire seen-null-first cleanup logic in a separate function as suggested -> that simplified things a lot more

tests/pipeline/test_import_export_schema.py
anuunchin: 🫡

dlt/normalize/normalize.py
anuunchin: I actually now think meddling with the schema in the normalizer is wrong in case we have multiple workers 👀

rudolfix (approved):
LGTM! thanks for fixing this. we need a better way to handle those. we create columns on null only to report them to the user at the end of the normalization. there must be a simpler way where we do not need to correct all edge cases with nested data.

PR #3134, opened 2025-09-25
restclient: add support for data parameter in RESTClient and rest_api

### LINKED ISSUE #2289: Unify the behavior of POST method to follow Requests library's POST method - add data as an argument
### Feature description

From the [dlt docs](https://dlthub.com/docs/general-usage/http/rest-client#restclientget-and-restclientpost-methods)

> These methods work similarly to the [get()](https://docs.python-requests.org/en/latest/api/#requests.get) and [post()](https://docs.python-requests.org/en/latest/api/#requests.post) functions from the Requests library.

However, to my knowledge passing in data as an argument is currently not possible. Therefore sending not-form-encoded data in the body of Request is not possible.

When trying to fetch data from an API that expects json payload to be of type string (not form-encoded) the following approach fails. 

```
import json

PAYLOAD = json.dumps(
	{
		"SearchKeys": {
			"CriteriaKey": "SearchCriteria:-998670234;;Filters:371857150;Sort:-1587235143",
			"ScratchpadKey": "a7f9b70c-e563-4ab3-b9ce-22f61e45236d"
		},
		"Criteria": {
			"ClassName": "resource_name",
			"IncludeWorkflow": False
		},
		"Filtering": {
			"Items": []
		},
		"Paging": {
			"Page": 1,
			"PageSize": 4
		},
		"Sorting": {
			"Items": [
				{
					"PropertyName": "description",
					"Direction": 1
				}
			]
		}
	}
)

source_config: RESTAPIConfig = {
"client": {
	"base_url": base_url,
	"auth": auth
},

"resources": 
	[
		{
			"name": "rest_list_all_posts",
			"table_name": "test_table",
			"endpoint": {
				"data_selector": "$", # Get the full response
				"path": "/search/",
				"paginator": page_paginator,
				"method": "POST",
				"json": PAYLOAD,
				
			},
		},
	]
}
```
Error trace:
```
For Endpoint: In path ./resources[0]/endpoint: field 'json' with value {"SearchKeys": {"CriteriaKey": "SearchCriteria:-998670234;;Filters:371857150;Sort:-1587235143", "ScratchpadKey": "a7f9b70c-e563-4ab3-b9ce-22f61e45236d"}, "Criteria": {"ClassName": "resource_name", "IncludeWorkflow": false}, "Filtering": {"Items": []}, "Paging": {"Page": 1, "PageSize": 4}, "Sorting": {"Items": [{"PropertyName": "description", "Direction": 1}]}} has invalid type 'str' while 'dict' is expected
```

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

From [requests docs](https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests)

> There are times that you may want to send data that is not form-encoded. If you pass in a string instead of a dict, that data will be posted directly. 

> For example, the GitHub API v3 accepts JSON-Encoded POST/PATCH data:
```
import json

url = 'https://api.github.com/some/endpoint'
payload = {'some': 'data'}

r = requests.post(url, data=json.dumps(payload))
```

Implementing this would enhance the ability to fetch data from different kinds of APIs

### Proposed solution
Unify the behavior of POST method to follow the requests library POST method by adding data as an argument. Alternatively allow json payload to be of type string (not form-encoded) using json.dumps() in endpoint configuration. 

### Related issues

_No response_

burnash:
The PR adds Python Request-like `data` param to RESTClient and rest_api

Closes #2289


dlt/sources/helpers/rest_client/__init__.py
```diff
@@ -19,6 +19,8 @@ def paginate(
     paginator: Optional[BasePaginator] = None,
     data_selector: Optional[jsonpath.TJsonPath] = None,
     hooks: Optional[Hooks] = None,
+    *,
```
burnash, line 22: Here and below: this `*` should have been here from the start: let's move it up with the next major release.

dlt/sources/helpers/rest_client/client.py
```diff
@@ -110,9 +110,15 @@ def _create_request(
         headers: Optional[Dict[str, Any]] = None,
         params: Optional[Dict[str, Any]] = None,
         json: Optional[Dict[str, Any]] = None,
+        data: Optional[Any] = None,
         auth: Optional[AuthBase] = None,
         hooks: Optional[Hooks] = None,
     ) -> Request:
+        if json and data:
```
burnash, line 117: Requests allows both json and data at the same time but prefers data. I've decided to simplify this and raise.

dlt/sources/rest_api/typing.py
```diff
@@ -250,6 +250,7 @@ class Endpoint(TypedDict, total=False):
     method: Optional[HTTPMethodBasic]
     params: Optional[Dict[str, Union[ResolveParamConfig, IncrementalParamConfig, Any]]]
     json: Optional[Dict[str, Any]]
+    data: Optional[Any]
```
burnash, line 253: Here it should be something like `Optional[Union[str, Dict[str, Any], List[Tuple[str, Any]]]]` but DictValidation breaks badly. Open to fix it in a follow up pr.

rudolfix (approved):
this is really good! I can't find any flaws in the code, also the test coverage is really good

PR #3135, opened 2025-09-25
Remove obsolete instruction from `CONTRIBUTING.md`

jorritsandbrink:
Looks to me like `make shell` is no longer a thing, so I removed its mention from `CONTRIBUTING.md`.
image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ Thank you for considering contributing to **dlt**! We appreciate your help in ma
 1. Fork the `dlt` repository and clone it to your local machine.
 2. Install `uv` with `make install-uv` (or follow the [official instructions](https://docs.astral.sh/uv/getting-started/installation/)).
 3. Run `make dev` to install all dependencies, including development ones.
-4. Activate your virtual environment with `make shell` and start working, or prepend all commands with `uv run` to run them within the uv environment. `uv run` is encouraged, it automatically keeps your project dependencies up to date.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -34,7 +34,6 @@ Thank you for considering contributing to **dlt**! We appreciate your help in ma
 1. Fork the `dlt` repository and clone it to your local machine.
 2. Install `uv` with `make install-uv` (or follow the [official instructions](https://docs.astral.sh/uv/getting-started/installation/)).
 3. Run `make dev` to install all dependencies, including development ones.
-4. Activate your virtual environment with `make shell` and start working, or prepend all commands with `uv run` to run them within the uv environment. `uv run` is encouraged, it automatically keeps your project dependencies up to date.
```
sh-rp, line 37: The stuff about uv is still acurate and should not be deleted. I would probably keep this info as is and re-add a shell command which activates the virtual env: `source .venv/bin/activate`

sh-rp (changes requested):

CONTRIBUTING.md
jorritsandbrink: I don't think that works:

image

Maybe that's why you removed that command in https://github.com/dlt-hub/dlt/pull/2787/commits/55f61e6d67c73524af9ec894f115da1e505ede8c?

I chose to update the instruction for venv activation instead.

sh-rp: Ah yes, ok :)

sh-rp (approved):

PR #3137, opened 2025-09-25
[Databricks destination] Feat/2863 databricks table optimization

### LINKED ISSUE #2863: [Databricks destination] Support for table optimization techniques in Databricks: partitioning, liquid clustering, data-skipping
### Feature description

Databricks supports different techniques to optimize data-retrieval, query latency, and make queries more efficient in general.
Some of the most common optimization techniques are:
1. Partitioning: it implies using `PARTITIONED BY (col_1, col_2)` in the table DDL statement - a similar concept to what is done in athena with the partition concept can be used
2. Liquid Clustering: the modern way to deal with partition and z-ordering in Databricks. It imply to pass `cluster by (col_1, col_2)` - can be achieve in a similar way as partitioning. It's also possible to have automatic clustering, which can be achieved via `CLUSTER BY AUTO` - that is really relevant in case we want to let Databricks do the heavy lifting for us.
3. data-skipping: more can be found [here](https://docs.databricks.com/aws/en/delta/data-skipping), it ca be defined as a table property; therefore to implement data-skipping  the destination must be extend to support setting table properties of whatever sort

In general this feature request will allow dltHub to have more control on how Delta Tables are written. 

### Are you a dlt user?

Yes, I'm already a dlt user. - Run multiple PoC in the past, and now I'm currently pushing for dltHub adoption in the new working place

### Use case

Optimize access pattern for tables in the raw layer, making queries faster and reducing cost.

### Proposed solution

For each optimization techniques, I already provided some hints on how the feature can be implement.

### Related issues

This issue: https://github.com/dlt-hub/dlt/pull/2674 is related

bayees:
### Description
Adds clustering, partitioning, table properties, and Apache Iceberg support to the Databricks destination adapter.

  New Parameters:
  - cluster: Column list for column clustering or "AUTO" for automatic optimization
  - partition: Table partitioning by specified columns
  - table_properties: Delta Lake optimization settings (TBLPROPERTIES)
  - table_format: Support for "DELTA" and "ICEBERG" table formats

  Key Features:
  - Performance optimization through clustering and partitioning for downstream queries
  - Ability to set data skipping with delta.dataSkippingStatsColumns

### Related Issues

- Closes #2863

### DIFF
--- a/dlt/destinations/impl/databricks/databricks_adapter.py
+++ b/dlt/destinations/impl/databricks/databricks_adapter.py
@@ -99,14 +99,17 @@ def databricks_adapter(
                 "delta.deletedFileRetentionDuration",
                 "delta.enableChangeDataFeed",
                 "delta.columnMapping.mode",
-                "delta.appendOnly"
+                "delta.appendOnly",
             ]
 
             for prop_key in table_properties.keys():
-                if any(prop_key.startswith(delta_prop) or prop_key == delta_prop for delta_prop in delta_only_props):
+                if any(
+                    prop_key.startswith(delta_prop) or prop_key == delta_prop
+                    for delta_prop in delta_only_props
+                ):
                     raise ValueError(
-                        f"Table property '{prop_key}' is Delta Lake specific and not supported with ICEBERG tables. "
-                        f"Remove this property when using table_format='ICEBERG'."
+                        f"Table property '{prop_key}' is Delta Lake specific and not supported with"
+                        " ICEBERG tables. Remove this property when using table_format='ICEBERG'."
                     )
 
     if cluster:
@@ -119,7 +122,8 @@ def databricks_adapter(
                 cluster = [cluster]
             if not isinstance(cluster, list):
                 raise ValueError(
-                    "`cluster` must be a list of column names, a single column name as a string, or 'AUTO'."
+                    "`cluster` must be a list of column names, a single column name as a string, or"
+                    " 'AUTO'."
                 )
             for column_name in cluster:
                 additional_column_hints[column_name] = {"name": column_name, CLUSTER_HINT: True}  # type: ignore[typeddict-unknown-key]
@@ -215,8 +219,8 @@ def databricks_adapter(
             # Validate value types
             if not isinstance(value, (str, int, bool, float)):
                 raise ValueError(
-                    f"Table property value for key '{key}' must be a string, integer, boolean, or float. "
-                    f"Got {type(value).__name__}."
+                    f"Table property value for key '{key}' must be a string, integer, boolean, or"
+                    f" float. Got {type(value).__name__}."
                 )
 
         additional_table_hints[TABLE_PROPERTIES_HINT] = table_properties
--- a/dlt/destinations/impl/databricks/databricks_adapter.py
+++ b/dlt/destinations/impl/databricks/databricks_adapter.py
@@ -9,34 +9,48 @@ from dlt.destinations.impl.databricks.typing import TDatabricksTableSchemaColumn
 
 
 CLUSTER_HINT: Literal["x-databricks-cluster"] = "x-databricks-cluster"
+PARTITION_HINT: Literal["x-databricks-partition"] = "x-databricks-partition"
+TABLE_FORMAT_HINT: Literal["x-databricks-table-format"] = "x-databricks-table-format"
 TABLE_COMMENT_HINT: Literal["x-databricks-table-comment"] = "x-databricks-table-comment"
 TABLE_TAGS_HINT: Literal["x-databricks-table-tags"] = "x-databricks-table-tags"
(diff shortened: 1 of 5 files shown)


bayees:
@burnash Hi Burnash, I am adding you because you where so kind to review at my last PR.

dlt/destinations/impl/databricks/databricks_adapter.py
```diff
@@ -9,34 +9,48 @@
 
 
 CLUSTER_HINT: Literal["x-databricks-cluster"] = "x-databricks-cluster"
+PARTITION_HINT: Literal["x-databricks-partition"] = "x-databricks-partition"
```
rudolfix: there's already an "official" hint for column-based partitioning: `partition`. please use it.
FYI: we are working on unifying partition hint on the table level so the column order and transform functions (ie year(column)) can be defined

dlt/destinations/impl/databricks/databricks_adapter.py
```diff
@@ -9,34 +9,48 @@
...
+TABLE_FORMAT_HINT: Literal["x-databricks-table-format"] = "x-databricks-table-format"
```
rudolfix: there's already a mechanism to specify table format on the resource level: https://dlthub.com/docs/dlt-ecosystem/destinations/delta-iceberg#set-table-format (example)

1. set the preferred and supported table formats in destination capabilities of databricks (in `factory.py`
```py
preferred_table_format: TTableFormat = ...
supported_table_formats: Sequence[TTableFormat] = ...
```
2. use `table.get("table_format")` to get table format when creating table schema

This is more powerful than using adapter:
- behavior is standardized and the same for many destinations
- you can set it on `pipeline.run` for ad hoc loads
- you can force all resources (including dlt tables) to use ie. iceberg
```py
destination=dlt.destinations.databricks(preferred_table_format="iceberg")
```

dlt/destinations/impl/databricks/databricks_adapter.py
```diff
@@ -53,21 +67,78 @@ def databricks_adapter(
...
 
+    # Handle table format
+    if table_format not in ["DELTA", "ICEBERG"]:
+        raise ValueError("`table_format` must be either 'DELTA' or 'ICEBERG'.")
+
+    # Store table format as a table hint
+    additional_table_hints[TABLE_FORMAT_HINT] = table_format
+
+    # Validate Iceberg-specific constraints
+    if table_format == "ICEBERG":
```
rudolfix: Oh OK, this you'd need to do when creating table... because `table_format` is not available at this point. and convert error to warnings (that given prop has no effect on ICEBERG) so we do not fail the load

other alternative is to implement
```py
def verify_schema(
        self, only_tables: Iterable[str] = None, new_jobs: Iterable[ParsedLoadJobFileName] = None
    ) -> List[PreparedTableSchema]:
```
on `DatabricksClient` but this check will happen in load phase so the user experience is still not as good as failing early or ignoring with warning.

tests/load/databricks/test_databricks_adapter.py
```diff
@@ -237,3 +240,352 @@ def demo_source():
...
+        ) as cur:
+            assert cur.fetchone()[0] == 5
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["databricks"]),
+    ids=lambda x: x.name,
+)
+def test_databricks_adapter_iceberg_format(
```
rudolfix, line 478: I just wonder if ICEBERG supports the same data types as delta. We could run this test on one of the fixtures that has all data types supported by `dlt`. Also we have a mechanism that can creates different destination data types based on `table_format`. However I think this test is sufficient and if there's a need we can improve type mapping later.

rudolfix (changes requested):
This is really cool! Thanks for writing good tests and updating docs. We have standard ways to set the table format and partitioning. Please see my review

tests/load/databricks/test_databricks_adapter.py
bayees: Parquet is the only file format that Databricks supports for Iceberg. So I believe that the current type mapper is sufficient. The only datatype not supported is `time`. I have added a test for iceberg in supplement to the test running in test_parquet_loading.

dlt/destinations/impl/databricks/databricks_adapter.py
bayees: Changed to the official partition flag

dlt/destinations/impl/databricks/databricks_adapter.py
bayees: Changed to the table_format flag

dlt/destinations/impl/databricks/databricks_adapter.py
bayees: I have implemented a suggestion to fix this.

PR #3138, opened 2025-09-26
docs: rest_api: fix rest_api_source parameters

burnash:


sh-rp (dismissed):

docs/website/docs/dlt-ecosystem/verified-sources/rest_api/advanced.md
```diff
@@ -11,8 +11,9 @@ keywords: [rest api, restful api, headers, response actions, advanced configurat
 - `section`: An optional section name in the configuration file.
 - `max_table_nesting`: Sets the maximum depth of nested tables above which the remaining nodes are loaded as structs or JSON.
 - `root_key` (bool): Enables merging on all resources by propagating the root foreign key to nested tables. This option is most useful if you plan to change the write disposition of a resource to disable/enable merge. Defaults to False.
+- `schema`: An explicit `Schema` instance to be associated with the source. If not present, `dlt` creates a new `Schema` object with provided `name`. If such `Schema` already exists in the same folder as the module containing the decorated function, such schema will be loaded from file.
```
VioletM: Could you mention the full path to the class the user would need to initialize?

VioletM (changes requested):

docs/website/docs/dlt-ecosystem/verified-sources/rest_api/advanced.md
burnash: Do you mean `dlt.Schema`?

VioletM: If it's what's expected, then yes! I don't think a lot of people worked directly with `dlt.Schema` so we need to mention that :)

VioletM:
```suggestion
- `schema`: An explicit `dlt.Schema` instance to be associated with the source. If not present, `dlt` creates a new `Schema` object with the provided `name`. If such `dlt.Schema` already exists in the same folder as the module containing the decorated function, such schema will be loaded from file.
```

PR #3144, opened 2025-09-29
Fix: Special handling of numeric type for oracle

### LINKED ISSUE #3133: Oracle NUMBER type incorrectly inferred as double
### dlt version

1.16.0

### Describe the problem

The oracle NUMBER type is used to represent both integer values as well as floating point values depending on the configuration. See https://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#i22289 for more details.

SQL Alchemy also models this by subclassing sqlalchemy.dialects.oracle.NUMBER from both sqlalchemy.types.Numeric and sqlalchemy.types.Integer. types. See https://docs.sqlalchemy.org/en/20/dialects/oracle.html#sqlalchemy.dialects.oracle.NUMBER

Dlt assumes that all oracle NUMBER types are of the Numeric variant and converts it to a double or decimal. It even mentions that it does this in the relevant code https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/sql_database/schema_types.py#L114

### Expected behavior

Instead of assuming that all NUMBER types are of the Numeric variant, dlt should check if NUMBER type is either the Numeric variant or the Integer variant.

In most cases this isn't that big of a deal, but in some larger values cannot be represented as a double while they can be represented as a bigint which becomes a problem during the conversion to arrow:
```
>>> import pyarrow as pa
>>> pa.array([9309935020231023], type="int64")
<pyarrow.lib.Int64Array object at 0x7b16939b51e0>
[
  9309935020231023
]
>>> pa.array([9309935020231023], type="float64")
  File "<stdin>", line 1, in <module>
  File "pyarrow/array.pxi", line 375, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 45, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Integer value 9309935020231023 is outside of the range exactly representable by a IEEE 754 double precision value
```

### Steps to reproduce

Successful pipeline:
1. Create an oracle table with a column with type NUMBER(17)
2. Insert some integers into the column
3. Run a pipeline to extract the table using the sql_table resource
4. Explore the dlt schema and notice that the column's data type is double

With the pyarrow error on top:
1. Create an oracle table with a column with type NUMBER(17)
2. Insert some integers into the column, including a very big number like 9309935020231023
3. Run a pipeline to extract the table using the sql_table resource with the pyarrow backend
4. Exception

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12

### dlt data source

sql_database(
        credentials
        schema="my_oracle_schema",
        defer_table_reflect=True,
        chunk_size=10000,
        table_names=["my_oracle_table"],
        backend="pyarrow",
    )

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

The workaround is to use a simple type adapter, but this should work out of the box

```
from sqlalchemy.dialects.oracle import NUMBER

def type_adapter_callback(sql_type):
    if isinstance(sql_type, NUMBER):
        return sql_type._type_affinity
    return sql_type  # Use default detection for other types
```

anuunchin:
This PR adjusts `sqla_col_to_column_schema` in `dlt/sources/sql_database/schema_types.py` so that it correctly infers the `bigint` dlt type from numeric columns from oracle that represent integers. Previously, they were inferred as "double", which was causing issues downstream, when pyarrow tried to build float values for very large integer numbers. An appropriate test is added.

Resolves #3133

### DIFF
--- a/dlt/sources/sql_database/schema_types.py
+++ b/dlt/sources/sql_database/schema_types.py
@@ -112,21 +112,25 @@ def sqla_col_to_column_schema(
         # we represent UUID as text by default, see default_table_adapter
         col["data_type"] = "text"
     elif isinstance(sql_t, sqltypes.Numeric):
-        # check for Numeric type first and integer later, some numeric types (ie. Oracle)
-        # derive from both
-        # all Numeric types that are returned as floats will assume "double" type
-        # and returned as decimals will assume "decimal" type
-        if sql_t.asdecimal is False:
-            col["data_type"] = "double"
-        else:
-            col["data_type"] = "decimal"
-            if sql_t.precision is not None:
+        # Oracle's Numeric type with integer affinity (e.g., NUMBER(17,0)) should be bigint
+        if hasattr(sql_t, "_type_affinity") and sql_t._type_affinity is sqltypes.Integer:
+            col["data_type"] = "bigint"
+            if add_precision and sql_t.precision is not None:
                 col["precision"] = sql_t.precision
-                # must have a precision for any meaningful scale
-                if sql_t.scale is not None:
-                    col["scale"] = sql_t.scale
-                elif sql_t.decimal_return_scale is not None:
-                    col["scale"] = sql_t.decimal_return_scale
+        else:
+            # all Numeric types that are returned as floats will assume "double" type
+            # and those that are returned as decimals will assume "decimal" type
+            if sql_t.asdecimal is False:
+                col["data_type"] = "double"
+            else:
+                col["data_type"] = "decimal"
+                if sql_t.precision is not None:
+                    col["precision"] = sql_t.precision
+                    # must have a precision for any meaningful scale
+                    if sql_t.scale is not None:
+                        col["scale"] = sql_t.scale
+                    elif sql_t.decimal_return_scale is not None:
+                        col["scale"] = sql_t.decimal_return_scale
     elif isinstance(sql_t, sqltypes.SmallInteger):
         col["data_type"] = "bigint"
         if add_precision:
--- a/tests/sources/sql_database/test_schema_types.py
+++ b/tests/sources/sql_database/test_schema_types.py
@@ -1,6 +1,13 @@
-import sqlalchemy as sa
+import pytest
+from typing import Optional, Union
 
-from dlt.sources.sql_database.schema_types import get_table_references
+import pyarrow as pa
+import sqlalchemy as sa
+from sqlalchemy.dialects.oracle import NUMBER
+from sqlalchemy.sql.type_api import TypeEngine
+
+from dlt.common.data_types import TDataType
+from dlt.sources.sql_database.schema_types import get_table_references, sqla_col_to_column_schema
 
 
 def test_get_table_references() -> None:
@@ -116,3 +123,49 @@ def test_get_table_references() -> None:
 
     refs = get_table_references(child)
     assert refs == []
+
+
+@pytest.mark.parametrize(
+    "oracle_type,expected_type,expected_precision,expected_scale,test_value",
+    [
+        (NUMBER(), "bigint", None, None, 123456789),
+        (NUMBER(precision=17), "bigint", 17, None, 9309935020231023),
+        (NUMBER(precision=17, scale=0), "bigint", 17, None, 9309935020231023),
+        (NUMBER(precision=10, scale=2), "decimal", 10, 2, 12345.67),
+    ],
+    ids=["NUMBER", "NUMBER(17)", "NUMBER(17,0)", "NUMBER(10,2)"],
+)
+def test_oracle_number_type_inference(
+    oracle_type: TypeEngine,
+    expected_type: TDataType,
+    expected_precision: Optional[int],
+    expected_scale: Optional[int],
+    test_value: Union[int, float],
+) -> None:
+    """Test Oracle NUMBER type inference to prevent PyArrow conversion errors.
+
+    Oracle NUMBER types can represent both integers and decimals based on their scale:
+    - NUMBER with scale=0 or no scale → should be inferred as 'bigint'
+    - NUMBER with scale>0 → should be inferred as 'decimal'
+
+    Previously, all Oracle NUMBER types were incorrectly inferred as 'double',
+    causing PyArrow conversion failures for large integers like 9309935020231023
+    that cannot be precisely represented as float64.
+    """
+    sql_col = sa.Column("test_col", oracle_type, nullable=True)
+    column_schema = sqla_col_to_column_schema(sql_col, reflection_level="full_with_precision")
+
+    assert column_schema["data_type"] == expected_type
+    if expected_precision is not None:
+        assert column_schema.get("precision") == expected_precision
+    if expected_scale is not None and expected_type == "decimal":
+        assert column_schema.get("scale") == expected_scale
+
+    # the inferred bigint type should work with PyArrow
+    if expected_type == "bigint" and test_value == 9309935020231023:
+        pa_array = pa.array([test_value], type="int64")
+        assert pa_array[0].as_py() == test_value
+        with pytest.raises(
+            pa.ArrowInvalid, match="Integer value 9309935020231023 is outside of the range"
+        ):
+            pa.array([test_value], type="float64")


dlt/sources/sql_database/schema_types.py
```diff
@@ -112,21 +112,25 @@ def sqla_col_to_column_schema(
...
-        # derive from both
-        # all Numeric types that are returned as floats will assume "double" type
-        # and returned as decimals will assume "decimal" type
-        if sql_t.asdecimal is False:
-            col["data_type"] = "double"
-        else:
-            col["data_type"] = "decimal"
-            if sql_t.precision is not None:
+        # Oracle's Numeric type with integer affinity (e.g., NUMBER(17,0)) should be bigint
+        if hasattr(sql_t, "_type_affinity") and sql_t._type_affinity is sqltypes.Integer:
```
sh-rp: as discussed, we are using an internal property here, possibly we can use a public interface of sqlalchemy, if not, we just need to add a comment that we stay aware of this. Maybe also double check that the tests for this are run with sqlalchemy 1.4 and 2.0.

anuunchin: replaced this part with an explicit comparison with oracle's Number type -> I think that's much clearer
🫡  also double checked that the test runs with sqlalchemy <2.0 and >2.0

dlt/sources/sql_database/schema_types.py
```diff
@@ -112,11 +113,14 @@ def sqla_col_to_column_schema(
         # we represent UUID as text by default, see default_table_adapter
         col["data_type"] = "text"
     elif isinstance(sql_t, sqltypes.Numeric):
-        # check for Numeric type first and integer later, some numeric types (ie. Oracle)
-        # derive from both
+        # special handling for oracle NUMBER types
+        if isinstance(sql_t, NUMBER) and sql_t.scale is None or sql_t.scale == 0:
```
rudolfix: I'm not sure this can be implemented this way. What about NUMERIC with 128bit precision and scale 0? We can't convert it into integer. This will work in Python (unlimited integers) but won't work with arrow which will not coerce such data.

we do not have oracle in our test setup so we cannot check it ourselves. My take would be to ask @jflipts to take a look at this. btw. in the issue a different method is used (`._type_affinity`)

dlt/sources/sql_database/schema_types.py
```diff
@@ -112,11 +113,14 @@ def sqla_col_to_column_schema(
...
+            col["data_type"] = "bigint"
+            if add_precision and sql_t.precision is not None:
```
rudolfix: I'm not sure it is correct. If there's a precision in numeric: is it a decimal precision or binary precision (integer) when we deal with integer? probably the safest was would be to not set it at all if we detect integer until we have a proper test setup with oracle (btw. we've added mssql recently)

dlt/sources/sql_database/schema_types.py
anuunchin: I see how it's not the correct way to do this,

PS: the initial implementation of this PR was based on `._type_affinity`, but is it really okay to use private attributes like this one? 👀

jflipts: I'm not an oracle expert either. I'm just tasked with implementing extracts from an existing oracle db. `._type_affinity` uses the same implementation as is used here. 

If I understand it correctly, then arrow doesn't support 128 bit integers? I'm not sure how you would deal with that. Maybe interpret it as a decimal so it is coerced to https://arrow.apache.org/docs/python/generated/pyarrow.decimal128.html?

dlt/sources/sql_database/schema_types.py
jflipts: The precision is decimal, but the combination with integer doesn't make much sense to me.

dlt/sources/sql_database/schema_types.py
rudolfix: Oracle does not have 128bit either. What our current fix does is to interpret any NUMERIC with 0 scale as integer Python type while it should be simply DECIMAL with 0 scale.

~@anuunchin let's go for type affinity. And we'll add Oracle to our tests because more problems are popping up.~

rudolfix: OK I checked again what Oracle does and it looks like it always is using decimal arithmetic on NUMERIC types. So there are no binary integers in Oracle at all. 

binary arithmetic is done on float/double types only.

@jflipts it looks like our code works correctly. Numeric types must always be Python Decimals. otherwise you'll do wrong type of arithmetic on them ie. decimal division is very different from binary division.

I just wonder why are you getting floats in the first place? Are you coercing decimals to floats?
https://docs.sqlalchemy.org/en/20/dialects/oracle.html#cx-oracle-numeric

if so large NUMERIC must fail, otherwise you lose significant bits in your numbers.

jflipts: I'm getting floats, because that is the default behavior of the combination of sqlalchemy number with dlts type mapping for NUMBER types where number has a none or 0 scale.

sqlalchemy sets asdecimal to false `asdecimal = bool(scale and scale > 0)` https://github.com/sqlalchemy/sqlalchemy/blob/e833474db064df32a7c02e809893e71ac4df4996/lib/sqlalchemy/dialects/oracle/types.py#L48

I'm  not sure why they introduced it like this in sqlalchemy, over 15 years ago:
* https://github.com/sqlalchemy/sqlalchemy/commit/d564baf593311ca5bc62dbc401e57986bf9169ba
* https://github.com/sqlalchemy/sqlalchemy/issues/885


dlt sets the type to double if asdecimal is false
```
if sql_t.asdecimal is False:
            col["data_type"] = "double"
```
https://github.com/dlt-hub/dlt/blob/a3ec5bf4e06fc5764d6f25b81f5cc98cae23a555/dlt/sources/sql_database/schema_types.py#L119

My original solution was essentially to just interpret NUMBER the same in dlt as it was implemented in sqlalchemy, but now I'm not even sure it is correct there. In any case double/float is very wrong for NUMBER(p) with a 0 scale as the oracle documentation states it is an integer.
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-9401BC04-81C4-4CD5-99E7-C5E25C83F608:~:text=Specify%20an%20integer%20using%20the%20following%20form%3A

dlt/sources/sql_database/schema_types.py
anuunchin: Changed to map Oracle NUMBER(p, 0) and NUMBER(p) to `decimal` with scale=0 instead of `bigint` ( to preserve Oracle's decimal arithmetic while fixing the precision loss bug that occurred with double mapping). The scale check now takes precedence over sqlalchemy's asdecimal=False flag, but that should be fine imo

kobethuwis: https://github.com/dlt-hub/dlt/issues/3133#issuecomment-3532415761 might help out here.

ivasio:
I've investigated Oracle numeric types and how sqlalchemy handles them and arrived at the same conclusion - NUMBER should be always interpreted as `decimal` to preserve original DB semantics, it's not a IEEE754 floating point number. And I'm also not sure if sqlalchemy handles it correctly. @kobethuwis thanks for the suggestion, nice idea with _if it fits in 64bits, put it in a BIGINT_ - I decided to skip it to keep type handling unified/simplified here

I made it always use `decimal` for Oracle NUMBER, and added an integration test to ensure type, precision and scale are correct for all Oracle numeric types. In addition to unit tests by @anuunchin of course!

ivasio:
Seems like there's another layer to the problem. Destination types are indeed correct as of now in this PR, but as long as sqlalchemy initially yields these values as `float` (which it does pretty much always as mentioned above when scale is not specified explicitly), our manipulations on dlt schema don't protect us from floating point error in the data. I have a pretty hacky workaround that monkey-patches sqlalchemy, and it allowed me to achieve the desired effect also in the case of `DECFLOAT` on Snowflake destination (tested only on `sqlalchemy` backend though, likely to have corner-case problems on pyarrow with unset precision):

```py
def get_multi_columns(
    original_fn,
    connection,
    *,
    schema,
    filter_names,
    scope,
    kind,
    dblink=None,
    **kw,
):
    superclass_result = original_fn(
        connection,
        schema=schema,
        filter_names=filter_names,
        scope=scope,
        kind=kind,
        dblink=dblink,
        **kw,
    )
    result = {}
    for key, columns in superclass_result:
        for column in columns:
            if isinstance(column["type"], NUMBER):
                if column["type"].scale == 0:
                    column["type"] = INTEGER()
                else:
                    column["type"] = NUMBER(
                        precision=column["type"].precision,
                        scale=column["type"].scale,
                        asdecimal=True,
                    )
        result[key] = columns
    return result.items()


def engine_adapter_callback(engine: Engine) -> Engine:
    engine.dialect.get_multi_columns = partial(get_multi_columns, engine.dialect.get_multi_columns)
    return engine
```

Ideally, of course, I'd attack the root of the problem in sqlalchemy - make the logic from this patch default [in the dialect class](https://github.com/sqlalchemy/sqlalchemy/blob/d608f62ec435ab790eb2a7dcfa672708195dc406/lib/sqlalchemy/dialects/oracle/base.py#L2250) and [in the type system](https://github.com/sqlalchemy/sqlalchemy/blob/d608f62ec435ab790eb2a7dcfa672708195dc406/lib/sqlalchemy/dialects/oracle/types.py#L35), but I doubt it'd be merged soon. In general, rigorous type handling would remove the possibility to interpret Oracle `NUMBER` as `float` at all, which is backward-incompatible and which I doubt will be met with enthusiasm by general audience of sqlalchemy

@rudolfix what do you think should be the best course of action here?

ivasio:
OK, the deeper problem solved by patching the `OracleDialect` instance, closely to what was as suggested above, but without using `INTEGER` for `scale == 0`. Works well for `sqlalchemy` and `pyarrow` backends, for `pandas` auto-conversion to `int` happens for `scale == 0` and other decimals are converted to float as for other source DBs

@rudolfix please review, implementation as discussed offline

dlt/sources/sql_database/helpers.py
```diff
@@ -14,6 +15,9 @@
 )
 import operator
 
+from sqlalchemy.dialects.oracle import NUMBER
```
rudolfix: please import those via `dlt.common.libs.sql_alchemy` so friendly exception messages are triggered if dependency is not present

dlt/sources/sql_database/schema_types.py
```diff
@@ -12,6 +12,7 @@
 from typing_extensions import TypeAlias
 
 from sqlalchemy.exc import NoReferencedTableError
+from sqlalchemy.dialects.oracle import NUMBER
```
rudolfix: not used IMO. pls remove

dlt/sources/sql_database/helpers.py
rudolfix: or maybe import it only when we are patching the dialect (in the right function)

rudolfix:
> I think we discussed that we'll add DECFLOAT type in this PR? if not pls write a ticket for that (and obviously the test in (2) we'll need this DECFLOAT so we can add this in separate PR)

OK I see DECFLOAT PR! I'll take a look tomorrow

ivasio:
@rudolfix thanks for the review! Fixed the individual comments. On the general one:

1. Is it enough to add this point to https://github.com/dlt-hub/dlt/issues/3495 or should it be in this PR?
2. Oracle NUMBER supports precision only up to 39/40 decimal digits, which is basically 128 bits. The correctly failing scenario I told you about was NUMBER() a column with two numbers of effective scales 0 and e.g. 20, but with the same default max precision. Is that the test you'd like me to add, or some other one?
3. Moved the DECFLOAT PR to draft, will explain the requirements there in more detail
5. Oh thanks, will check!

rudolfix (approved):
LGTM!

Issue #3145, opened 2025-09-29
Slow parquet reading


mattiasthalen:
I'm having issues with this pipeline:
```python
# Define the tpch source
@dlt.source()
def tpch(source_path="/lakehouse/default/Files/tpch"):
    from dlt.sources.filesystem import filesystem, read_parquet

    tables = [
        "customer",
        "lineitem",
        "nation",
        "orders",
        "part",
        "partsupp",
        "region",
        "supplier",
    ]

    for table in tables:
        resource = filesystem(
            bucket_url=source_path,
            file_glob=f"{table}/{table}.*.parquet",
            files_per_page=1,
            incremental=dlt.sources.incremental("modification_date"),
        )

        resource_pipe = resource | read_parquet()

        yield resource_pipe.with_name(f"raw__tpch__{table}").apply_hints(write_disposition="append")
```

It takes more than 10min just to extract. But if I vibe code a parquet reader using duckdb, it's done within a minute:
```python
from typing import TYPE_CHECKING, Any, Dict, Iterator, Optional
from dlt.sources import TDataItems, DltResource, DltSource
from dlt.sources.filesystem import FileItemDict
from dlt.extract import decorators

def _read_parquet_duckdb(
    items: Iterator[FileItemDict],
    chunk_size: Optional[int] = None,      # set e.g. 200_000 to stream batches
    use_pyarrow: bool = False,             # unused here; kept for signature compatibility
    **duckdb_kwargs: Any,
) -> Iterator[TDataItems]:
    import duckdb

    # One connection is fine for local scans
    con = duckdb.connect()

    for item in items:
        # FileItemDict from dlt.filesystem has a stable string path in "file_url"
        # (works for local paths and URLs like s3://...)
        path = item["file_url"]

        # DuckDB expects a path or glob, not a file-like object
        rel = con.from_parquet(path, **duckdb_kwargs)  # returns a Relation

        # Convert to Arrow once; then optionally re-chunk for dlt
        tbl = rel.arrow()  # pyarrow.Table

        if chunk_size and chunk_size > 0:
            for batch in tbl.to_batches(max_chunksize=chunk_size):
                yield batch  # pyarrow.RecordBatch
        else:
            yield tbl       # pyarrow.Table

read_parquet_duckdb = decorators.transformer()(_read_parquet_duckdb)
```

I've tried several chunk sizes for the standard `read_parquet()`, but with no luck.
Am I missing something, or is duckdb just that much better?

[...]

PR #3149, opened 2025-09-29
fix: 3145 add `read_parquet(use_arrow: bool)`

zilto:
Attempt to fix https://github.com/dlt-hub/dlt/issues/3145

## Changes
- the main change is a single line: `yield batch.to_pylist()` -> `yield batch if use_pyarrow else batch.to_pylist()`
- add basic tests for all readers (previously untested)
- fixed typing for arg `items` from the incorrect `Iterator` to `Iterable`

## Future work
- unify reader signatures
- implement CSV reader with `pyarrow` instead of `pandas` given the former is a required dependency of `dlt`
- set default  `read_parquet(use_pyarrow=True)` instead of `False` because it should be faster and `pyarrow` is a required deps.
- add a `read_parquet_duckdb()` reader OR automatically use `duckdb` inside `read_parquet()` if `duckdb` is available

### DIFF
--- a/dlt/common/storages/fsspec_filesystem.py
+++ b/dlt/common/storages/fsspec_filesystem.py
@@ -298,6 +298,8 @@ class FileItemDict(DictStrAny):
                 bytes_io,
                 **text_kwargs,
             )
+        # `FileItemDict` kwarg `fsspec` is `Optional`. If `fsspec=None` this code branch
+        # will fail.
         else:
             if "file" in self.fsspec.protocol:
                 # use native local file path to open file:// uris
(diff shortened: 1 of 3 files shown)


rudolfix (approved):
LGTM!

zilto:
@rudolfix any objection to setting `use_pyarrow = True` as default? User reported a 20x performance improvement from 10min to <1min loading big parquet files

zilto:
> @rudolfix any objection to setting `use_pyarrow = True` as default? User reported a 20x performance improvement from 10min to <1min loading big parquet files

After discussion, we decided to keep `use_pyarrow=False` as default for full backwards compatibility.

There could be edge cases that differ between Python and Pyarrow normalization and type inference. This could evolve the schema or break pipelines.

## TODO
Once we evaluate some of these edge cases, we should set a deprecation warning akin to 
> In dlt==X.Y.Z, we're introduced `use_pyarrow: bool` with default `False`. Set to `True` for better performance. In `dlt==X.Y.Z`, default will be `use_pyarrow=True`. This should cause no error, but you can try `use_pyarrow=True` right now to validate it.

Issue #3139, opened 2025-09-27
'FATAL: password authentication failed for user' when using DuckLake destination with PostgreSQL catalog


karhendana:
### dlt version

1.17.0

### Describe the problem

DLT's DuckLake ATTACH statement implementation has a bug where database passwords are being replaced with asterisks (***) when configuring PostgreSQL (and MySQL) catalogs. This prevents successful catalog connections as the masked password is being used in the actual connection string.

When using DuckLake with a PostgreSQL catalog, the `build_attach_statement()` method in `ducklake/sql_client.py` is incorrectly handling the database URL by allowing SQLAlchemy's default password masking behavior to occur.

```
        elif catalog.drivername in ("postgres", "postgresql", "mysql"):
            attach_statement = (
                f"ATTACH IF NOT EXISTS 'ducklake:{catalog.drivername}:{catalog.to_url()}'"
            )
```

**Proposed Fix**

Update the code to explicitly disable password masking:

```
                f"ATTACH IF NOT EXISTS 'ducklake:{catalog.drivername}:{catalog.to_url().render_as_string(hide_password=False)}'"
```

### Steps to reproduce

```
import dlt
from dlt.destinations.impl.ducklake.configuration import DuckDbBaseCredentials

credentials = DuckLakeCredentials(
    "lake_catalog",
    catalog="postgresql://loader:pass@localhost:5432/dlt_data",
    storage="s3://dlt-ci-test-bucket/lake",
)
ducklake = dlt.destinations.ducklake(credentials=credentials)
```

resulting in

```
FATAL: password authentication failed for user "loader"
```

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12

### dlt destination

DuckLake

### Additional information

Related issue: https://github.com/sqlalchemy/sqlalchemy/discussions/10996#discussioncomment-8434129

sh-rp:
NOTE: Something with our test setup seems to be broken if this is not caught?

zilto:
@karhendana I can't reproduce your error. Can you provide a more complete repro of your environment? (python dependencies, postgres version of your instance)

## Reproduction
The following code works against the [postgres configuration we use in test + CI](https://github.com/dlt-hub/dlt/blob/devel/tests/load/postgres/docker-compose.yml).
```python
import dlt
from dlt.destinations.impl.ducklake.configuration import DuckLakeCredentials

credentials = DuckLakeCredentials(
    "lake_catalog",
[...]

PR #3150, opened 2025-09-29
fix: 3139 pass SQLAlchemy credentials to f-string

zilto:
Solves issue #3139 

## Problem
When creating the `ATTACH` statement for the ducklake client, password value from SQLAlchemy would render as `***` instead of its value. This was not caught in our CI / tests because we use `sqlalchemy 1.4` and this behavior seems to be only in `sqlalchemy 2.0`

## Changes
- change `db_url = catalog.to_url()` -> `db_url = catalog.to_url().render_as_string(hide_password=False)`
- misc:
  - replaced `posixpath` with `pathlib` (posixpath is an internal module as its docstring mentions)
  - remove unused imports

### DIFF
--- a/dlt/destinations/impl/ducklake/ducklake.py
+++ b/dlt/destinations/impl/ducklake/ducklake.py
@@ -46,9 +46,11 @@ ATTACH 'ducklake:{catalog_database}' (DATA_PATH '{data_storage}');
 
 adds the ducklake `Catalog` to your duckdb database
 """
+from __future__ import annotations
+
 import os
+import pathlib
 from packaging.version import Version
-import posixpath
 from typing import Iterable, List, Optional, Sequence
 
 import dlt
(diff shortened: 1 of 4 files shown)


dlt/destinations/impl/ducklake/ducklake.py
```diff
@@ -46,9 +46,11 @@
 
 adds the ducklake `Catalog` to your duckdb database
 """
+from __future__ import annotations
```
anuunchin, line 49: A question for understanding - this is the first time I'm seeing this  and looked it up -> Is the reason to use this instead of a double quoted forward reference the fact that, starting from python 3.11, all annotations are stored as strings by default and we can just remove this when python 3.10 is no longer supported (instead of going over double quoted references)? 👀

zilto: In short:
- When you import or execute a Python file, it is read from top to bottom.
- This created issues for the earliest Python versions of type annotations.
  ```python
  class A:
     def __init__(self): ...
     
     # can't annotate `B` because it is not defined yet when going top-to-bottom
     def get_b(self) -> B: ...
     
  class B:
    ...   
  ``` 
- The first solution was to define types as strings
  ```python
  # keeping everything else the same
  class A:
    # ...
    def get_b(self) -> "B": ...
  ```
- This is error-prone and didn't support IDE features (e.g., can't rename class, navigate to class definition)
- Python added the built-in `from __future__ import annotations`. This means the full file is read from top-to-bottom and all the import mechanism is resolved before annotations are added to Python objects
- This allows to write the previously problematic `def get_b(self) -> B: ...` even if `B` isn't defined yet because `B` will be defined by the time the file and imports are resolved.

TL;DR, we can and we should have `from __future__ import annotations` on all files. It's common to have this as a linting rule. It hasn't enforced as the default behavior either
> from __future__ import annotations was previously scheduled to become mandatory in Python 3.10, but the Python Steering Council twice decided to delay the change ([announcement for Python 3.10](https://mail.python.org/archives/list/python-dev@python.org/message/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/); [announcement for Python 3.11](https://mail.python.org/archives/list/python-dev@python.org/message/VIZEBX5EYMSYIJNDBF6DMUMZOCWHARSO/)). No final decision has been made yet.

PR #3152, opened 2025-09-30
fixes `test_track_anon_event` test

rudolfix:
### Description
`test_track_anon_event` was making connections to CI tracker which could result in timeout exception (chances increased after we lowered timeouts to 1.0 sec). this PR uses mock "post" request instead


zilto (approved):
Thanks for the fix!

Issue #3140, opened 2025-09-28
feat(ducklake): Disambiguate DuckLake configuration names


zilto:
The default naming in `ducklake` destination are confusing IMHO. To fully grasp it, you need a good understanding of `ducklake` format AND the catalog database.

<img width="1706" height="1797" alt="Image" src="https://github.com/user-attachments/assets/a4eab611-cd46-4287-baab-550371af4634" />

This screenshot shows loading data using the native duckdb connection vs dlt `ducklake` destination. The left panel shows the produced database structure for native (top) and dlt (bottom).

## The problem
The current `ducklake_catalog` is not a "ducklake catalog"; it's a **duckdb catalog for ducklake data**, i.e., a database-level abstraction. Here, `ducklake_catalog` is the namespace that points to the actual data found in storage. Although accurate, the default name is confusing IMO. 

Also, when attaching the ducklake client to the catalog (could be duckdb, sqlite, postgres, etc.), it creates "duckdb catalog for the ducklake catalog" with template `__ducklake_metadata_{ducklake_catalog.catalog}` which adds to the confusion.

Furthermore, [ducklake specs](https://ducklake.select/docs/stable/duckdb/usage/connecting) doesn't seem to name the entity level that is named `ducklake_catalog` here. 
 
## Suggested solution
- rename config `catalog_name` back to `ducklake_name`. The config should not have the word `catalog` because it doesn't relate to a "ducklake catalog" in any way.
- set default config `ducklake_name="ducklake" or  `ducklake_name="dlt-ducklake"`.   It should have the word `ducklake` because it shows that duckdb is using ducklake

The suggestion here is backwards incompatible and breaking. I think it's worth making given it significantly improves clarity and `ducklake` has been available only 2 working days.

Configuration should look like

```toml
[destination.ducklake.credentials]
ducklake_name="dlthub"

[destination.ducklake.credentials.catalog]
drivername="postgres"
user="loader"
password="pass"
host="localhost"
port=5432
database="dlt_data"

[destination.ducklake.credentials.storage]
bucket_url="s3://dlt-ci-bucket"

[destination.ducklake.credentials.storage.credentials]
aws_access_key_id = "<configure me>" # fill this in!
aws_secret_access_key = "<configure me>" # fill this in!
```

rudolfix:
[...]

PR #3153, opened 2025-09-30
fix(ducklake): 3140 disambiguate config key and default values

zilto:
- Follows #3140 
- Implements reviews from the end of PR #2709 

## Changes
The main change is changing configuration `catalog_name: ducklake_catalog` to `ducklake_name: ducklake`.

## Future todos
It would be nice to support a `:pipeline:` directive such that `ducklake_name` takes the value of the `dlt.Pipeline` by default. This would match the behavior for `duckdb` and `filesystem` destinations


zilto:
Merging since changed were approved by @rudolfix

Issue #3155, opened 2025-10-01
Feature: Add resource history support to access parent data


ArneDePeuter:
### Feature description

Currently, when building resource/transformer pipelines in `dlt`, referencing parent data requires explicitly passing it down through every child transformer. This often leads to parent fields being unnecessarily propagated into downstream records. This complicates data consistency, pipeline logic, etc...

The proposed feature introduces **`history`**, a special parameter available in resources/transformers, that provides access to any yielded data point from a parent resource/transformer, as long as that parent has been defined with `keep_history=True`.

This enables direct access to parent records without having to explicitly pass them down the pipeline.,

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Consider the following situation:

```python
@dlt.resource(keep_history=True)
def users():
    yield {"id": 1, "name": "Alice"}
    yield {"id": 2, "name": "Bob"}

@dlt.transformer(data_from=users)
def posts(user):
    yield {"post_id": 1, "text": "Hello"}
    yield {"post_id": 2, "text": "World"}

@dlt.transformer(data_from=posts)
def comments(post, history: dlt.History = None):
    user = history[users]
    """
    Maybe here for a request we all of the sudden need the user id because of poor api design
    Instead of propagating the user_id with the json and adding data complexity we can retreive it form the history
    """
    yield {"comment_id": 10, "post_id": post["post_id"], "user_id": user["id"]}
```

Without `history`, the `user` object would need to be passed through `posts` → `comments`, cluttering the downstream schema with redundant fields.

With `history`, transformers can reference upstream records directly, improving **data consistency** and **schema cleanliness**.

### Proposed solution

* Introduce a `history` object injected into transformers when declared as a parameter:

  ```python
  def transformer_fn(item, history: dlt.History = None): ...
  ```
  * The definition style of `history` is the same as the `meta` feature

* `history` allows access to parent resources/transformers by either:

  * **resource reference**: `history[users]`
  * **string key/name**: `history["users"]`
* Access is only available if the parent was defined with `keep_history=True`. Otherwise, an `InvalidHistoryAccess` exception is raised.
* `EMPTY_HISTORY` is provided when no history is available.
* Attempting to access a non-existent parent raises `InvalidHistoryAccess`.
* The feature is optional and does not break existing transformers.

zilto:
[...]

PR #3156, opened 2025-10-01
feat/3155 history

ArneDePeuter:
### Description

This PR introduces the `history` feature (see issue #3155), which allows transformers to access parent records directly when `keep_history=True` is set.  
This removes the need to explicitly propagate parent fields through every child transformer, simplifying pipelines and reducing schema clutter.

### Implementation

- When a resource is defined with `keep_history=True`, a `history` dictionary is made available to all subsequent nodes.  
- Only nodes marked with `keep_history=True` contribute their data to this dictionary, ensuring that only the required information is retained.  
- If no node (or any of its parents) uses the history feature, all nodes receive a shared `EMPTY_HISTORY` object. This avoids additional allocations and keeps the overhead negligible, since only a single immutable reference is passed around.  

The design follows a similar approach to the existing `meta` feature for consistency.  
Comprehensive tests have been added to validate the behavior of this new feature.

### Tiny Additional Proposal

As a follow-up improvement, we could simplify function signatures by removing the need for explicit `= None` defaults.  
For example, instead of writing:

```python
def transformer(item, meta=None, history=None):
    ...

```
we could allow:
```python
def transformer(item, meta, history):
    ...

```
This would make transformer definitions cleaner and reduce boilerplate in function signatures.

### DIFF
--- a/dlt/extract/history.py
+++ b/dlt/extract/history.py
@@ -1,23 +1,28 @@
 from __future__ import annotations
-
-from typing import TYPE_CHECKING, Any, Union, Iterable
+from typing import Any, Optional, Union, TYPE_CHECKING
 from dlt.extract.exceptions import InvalidHistoryAccess
 
 if TYPE_CHECKING:
     from dlt.extract.resource import DltResource
 
 
-class History(dict[str, Any]):
+class History:
     """
-    A dict-like wrapper that provides ergonomic access to historical resource outputs
-    by either a string key or a ``DltResource`` reference.
+    A simple tree node:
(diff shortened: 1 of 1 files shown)


zilto:
Hey @ArneDePeuter, we really appreciate your work but I will close this PR as `Not currently planned`. `dlt` is currently built around pipes, which we implicitly optimize as trees. For example, a `@dlt.resource` will be process items once, but can be dispatched to multiple `@dlt.transformer`.

image

This history feature would open the door to a DAG structure, which would have a lot of unexpected side-effects.

PR #3164, opened 2025-10-02
feat:3162 add resource name to incremental extract duplicate checks logging.

### LINKED ISSUE #3162: feat: add to logging resource name on  incremental extract duplicate thresholds
## Description

As a DLT technical user on multiple extract duplicate warnings would like to see what is the resource the thresholds are breached.

## Examples 

```{"written_at":"2024-01-15T14:18:11.004Z","written_ts":12323743747832468,"component_name":"salesforce","process":12345,"taskName":null,"msg":"Large number of records (201) sharing the same value of cursor field '<yourcursorfield>'. This can happen if the cursor field has a low resolution (e.g., only stores dates without times), causing many records to share the same cursor value. Consider using a cursor column with higher resolution to reduce the deduplication state size.","type":"log","logger":"dlt","thread":"MainThread","level":"WARNING","module":"__init__","line_no":600,"version":{"dlt_version":"1.15.1","pipeline_name":"<pipeline_name>"}}```

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

On Salesforce for example it is hard to tell which is the resource deduped/with source data issues

Image

[Slack reference ](https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1752750909719419?thread_ts=1752668575.790299&cid=C04DQA7JJN6)

### Proposed solution

Add to logger.warning `_check_duplicate_cursor_threshold`  the resource name in question.

### Related issues

_No response_

### PR DESCRIPTION
**Pull request by and2reak**
### Description

Improve logging on incremental extract duplicate checks.

Issue #3162

### Related Issues

- Resolves #3162 

### Additional Context
When the threshold is reached the logging does not provide information on which resource it refers to. Depending on the checks the existing warning could appear thousands of times and it does not provide the clarity to debug which resources are in question. 
 This feature will add the resource name to the warnings and allows easier debugging both in the framework and in destination data.

### DIFF
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -610,12 +610,11 @@ class Incremental(ItemTransform[TDataItem], BaseConfiguration, Generic[TCursorVa
     ) -> None:
         if initial_hash_count <= Incremental.duplicate_cursor_warning_threshold < final_hash_count:
             logger.warning(
-                f"Large number of records ({final_hash_count}) sharing the same value of "
-                f"cursor field '{self.cursor_path}'. This can happen if the cursor "
-                "field has a low resolution (e.g., only stores dates without times), "
-                "causing many records to share the same cursor value. "
-                "Consider using a cursor column with higher resolution to reduce "
-                "the deduplication state size."
+                f"Large number of records ({final_hash_count}) sharing the same value of cursor"
+                f" field '{self.cursor_path}' on resource {self.resource_name}. This can happen if"
+                " the cursor field has a low resolution (e.g., only stores dates without times),"
+                " causing many records to share the same cursor value. Consider using a cursor"
+                " column with higher resolution to reduce the deduplication state size."
             )
 
 

### CODE REVIEW

File: dlt/extract/incremental/__init__.py [hunk 1]
```diff
@@ -610,12 +610,11 @@ def _check_duplicate_cursor_threshold(
...
         if initial_hash_count <= Incremental.duplicate_cursor_warning_threshold < final_hash_count:
             logger.warning(
-                f"Large number of records ({final_hash_count}) sharing the same value of "
-                f"cursor field '{self.cursor_path}'. This can happen if the cursor "
-                "field has a low resolution (e.g., only stores dates without times), "
-                "causing many records to share the same cursor value. "
-                "Consider using a cursor column with higher resolution to reduce "
-                "the deduplication state size."
+                f"Large number of records ({final_hash_count}) sharing the same value of cursor"
+                f" field '{self.cursor_path}' on resource {self.resource_name}. This can happen if"
```
anuunchin:
```suggestion
                f" field '{self.cursor_path}' on resource '{self.resource_name}'. This can happen if"
```

anuunchin (changes requested):

dlt/extract/incremental/__init__.py
anuunchin: @and2reak  just a minor thing to align with the existing format 🙂

and2reak: @anuunchin solved. I did it with a new commit.

anuunchin (approved):

PR #3171, opened 2025-10-06
Feat/adds workspace

rudolfix:
### Description
This PR adds a new dlt run context that is called Workspace. Core features compatible with dlthub project:
* adds profiles that refer to config/secrets variant. there are 4 built in profiles and command to pin profile to workspace
* adds profile-aware config/secrets toml providers
* workspaces isolate pipelines (working dirs) and local files. it groups them in dedicated folders for easy cleanup
* adds cli to manage workspace and profiles
* adds workspace and pipeline mcp server (refactors code in `dlt-plus` package)

**new workspace must be enabled by creating empty `.dlt\.workspace` file (feature flag)**
lookup for workspace root **is not recursive**. you must be in the top workspace folder to activate it

OSS functionality won't be impacted in any way for now

new cli commands (preliminary)
`dlt profile [info|list|pin]` (profile pinning makes give profile a default only locally, pin is gitigniored)
`dlt workspace [info|clean|mcp`] (info is pretty cool, we can move `mcp` somewhere else. I wanted to hide it under feature flag for now)

TODO:
* code refactor: all code related to workspace (command line, marimo/streamlit etc.) will be moved to `_workspace` module
* related dependencies will be removed from the core deps and moved to `cli` and `workspace` extras
* add WorkspaceConfiguration
* a few less significant tests are missing. **do proper test setup**
* update dashboard app so it shows current profile name and allows to switch profiles (will also work with `projects` as interfaces got standardized)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,7 +180,6 @@ postgis = [
 workspace = [
     "duckdb>=0.9",
     "ibis-framework>=10.5.0 ; python_version >= '3.10'",
-    "duckdb",
     "pyarrow>=16.0.0",
     "marimo>=0.14.5",
     "mcp>=1.2.1 ; python_version >= '3.10'",
--- a/tests/workspace/cli/test_mcp_commands.py
+++ b/tests/workspace/cli/test_mcp_commands.py
@@ -1,3 +1,10 @@
+import sys
+import pytest
+
+# skip the whole module on python < 3.10
+if sys.version_info < (3, 10):
+    pytest.skip("requires Python 3.10+", allow_module_level=True)
+
 from mcp.server.fastmcp import FastMCP
 from pytest_console_scripts import ScriptRunner
 from pytest_mock import MockerFixture
--- a/tests/workspace/mcp/__init__.py
+++ b/tests/workspace/mcp/__init__.py
@@ -1,0 +1,6 @@
+import sys
+import pytest
+
+# skip the whole module on python < 3.10
+if sys.version_info < (3, 10):
+    pytest.skip("requires Python 3.10+", allow_module_level=True)
--- a/tests/workspace/test_providers.py
+++ b/tests/workspace/test_providers.py
@@ -3,7 +3,7 @@ from pathlib import Path
 
 from dlt._workspace.providers import ProfileSecretsTomlProvider
 
-TESTS_CASES_DIR = os.path.join("tests", "workspace/cases/provider")
+TESTS_CASES_DIR = os.path.join("tests", "workspace", "cases", "provider")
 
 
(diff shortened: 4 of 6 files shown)


dlt/_workspace/cli/_workspace_command.py
```diff
@@ -0,0 +1,109 @@
...
+        )
+        clean_local_parser.add_argument(
+            "--skip-data-dir",
+            action="store_true",
+            default=False,
+            help="Do not delete pipelines working dir.",
+        )
+
+        subparsers.add_parser(
+            "info",
```
sh-rp, line 48: I tested this out, very cool :). Maybe we should also have `dlt info` that shows the same info for non-workspace projects.

sh-rp: One note: if I pin a profile it will tell me at the top that <profile_name> is pinned. If nothing is pinned, it should still tell you what profile it will use by default. You can kind of guess that it is dev from the output, but mentioning it explicitly would be good imho.

dlt/_workspace/cli/_workspace_command.py
```diff
@@ -0,0 +1,109 @@
...
+    fmt.echo("dlt reads configuration from following locations:")
+    for provider in providers_context.providers:
+        fmt.echo("* %s" % fmt.bold(provider.name))
+        for location in provider.locations:
+            fmt.echo("    %s" % location)
+        if provider.is_empty:
+            fmt.echo("    provider is empty")
+
+
+def clean_workspace(run_context: RunContextBase, args: argparse.Namespace) -> None:
```
sh-rp, line 98: We should tell the user what is about to happen and ask wether they want to proceed, not just delete all local files.

dlt/_workspace/cli/utils.py
```diff
@@ -0,0 +1,118 @@
...
+from dlt.common.configuration.specs.pluggable_run_context import (
+    RunContextBase,
+    ProfilesRunContext,
+)
+from dlt.common.storages.file_storage import FileStorage
+
+from dlt.cli import echo as fmt
+
+
+def display_run_context_info() -> None:
```
sh-rp, line 17: Ok, this is where this profile output comes from. I think we need this to contain info for non-pinned default profiles too.

dlt/_workspace/mcp/prompts/pipeline.py
```diff
@@ -0,0 +1,63 @@
+# ruff: noqa: E501
+
+import textwrap
+
+
+def dlt_welcome() -> str:
```
sh-rp, line 6: These do not seem to be used anywhere?

PR #3178, opened 2025-10-06
add custom header generation to ArrowToCsvWriter

burnash:
PyArrow's CSVWriter doesn't support specifying the quoting style for headers as of current release 21.0.0 (issue [apache/arrow#47575](https://github.com/apache/arrow/issues/47575)). This caused headers to always be quoted regardless of the `quoting` parameter setting. 

This PR implements header quoting style support.


rudolfix (approved):
LGTM!

PR #3179, opened 2025-10-06
feat: unify `dlt.Relation` API and create bound Ibis tables

zilto:
This continues the work started in #2498 (see previous review round)

## Summary
- Add a generic Ibis backend called `_DltBackend` that forwards Ibis expressions to the native destination SQL client provided by dlt
- Allows to create bound Ibis table, chain expressions and call `.execute()` or `.to_pyarrow()` directly on the Ibis object
- Simplifies the `dlt.Relation` object by removing the `.table(..., table_type=...)` kwarg

### Red / blue code
Reusing the language from [this blog about sync/async](https://journal.stuffwithstuff.com/2015/02/01/what-color-is-your-function/), the `dlt.Dataset.table()` method introduces red and blue code. Based on `table_type=` kwarg, it returns a `dlt.Relation` or `ibis.Table` object, which are really different are currently hard to mix. 

Returning different types of objects based on a kwarg is generally something to avoid. This pattern also prevents Ibis users from all convenience methods available on `dlt.Relation` 

### Better ergonomics
I prefer using Ibis when interactively exploring datasets and writing transformations in a notebook. Currently, the ergonomics are poor because I need to:
1. get a `dlt.Dataset`
2. get an unbound ibis table via `.table(..., table_type="ibis")`
3. write queries on the Ibis table (here, it's native Ibis features)
4. pass the query to `dlt.Dataset` to create a new `dlt.Relation` 
5. execute the relation to see results

Step `4.` is the most awkward because you need to constantly add / delete `dataset(my_ibis_expr).df()` to view results. pattern

### Red / blue code
The method `dlt.Dataset.table(...)` should always return a `dlt.Relation` and `dlt.Relation.to_ibis()` returns a native Ibis object to process further. This allows to deprecate `.table(..., table_type=...)`.

The method `.to_ibis()` is in line with other methods like `.to_pandas()`, `.to_pyarrow()`, `.to_sql()`,  `.to_sqlglot()`, etc. that we could provide on `dlt.Relation` (future work)

### Better ergonomics
Having a `_DltBackend` allows us to get bound Ibis tables while avoiding connecting to the destination (the main reason why we we're using unbound tables). Bound tables allow to:
1. get a `dlt.Dataset`
2. get a `dlt.Relation`
3. (optional) use `dlt.Relation` convenience methods (e.g., joining on references)
4. call `.to_ibis()` to get a native object
5. manipulate the ibis object and call `.to_pyarrow()` or `.execute()` to view results directly (this uses dlt's SQL client)

This feels much more familiar to Ibis users and ensures compatibility with the rest of the Ibis ecosystem. 

## TODO
- Update docs & educational content
- Add deprecation warnings for `.table(..., table_type=...)`
- Review marimo dashboard and features

## Future work
- Currently, `dlt.Dataset.ibis()` returns a native Ibis connection based on the specific dlt destination. By returning the `_DltBackend` instance, this could support a broader set of destinations (e.g., if we implement an SQL interface for LanceDB with an in-memory DuckDB)

### DIFF
--- a/dlt/common/libs/ibis.py
+++ b/dlt/common/libs/ibis.py
@@ -2,42 +2,35 @@ from __future__ import annotations
 
 import contextlib
 from typing import TYPE_CHECKING, Any, Union
-
-# ibis imports follow the convention used in the ibis source code
-import ibis
-import ibis.expr.datatypes as dt
-import ibis.expr.operations as ops
-import ibis.expr.schema as sch
-import ibis.expr.types as ir
-import sqlglot as sg
-from ibis.backends import _get_backend_names, NoUrl, NoExampleLoader
-from ibis.backends.sql import SQLBackend
-from ibis.formats import TypeMapper
+from typing_extensions import override
 
 import dlt
-from dlt.helpers.ibis import _get_ibis_to_sqlglot_compiler
+from dlt.helpers.ibis import create_ibis_backend, _get_ibis_to_sqlglot_compiler
 from dlt.common.schema import Schema as DltSchema
 from dlt.common.destination import TDestinationReferenceArg
-from dlt.common.schema.typing import TDataType
-from dlt.destinations.dataset.factory import dataset as dataset_factory
+from dlt.common.schema.typing import TDataType, TTableSchema
+from dlt.common.exceptions import MissingDependencyException
 
 if TYPE_CHECKING:
     import pandas as pd
     import pyarrow as pa
+    from ibis import BaseBackend
 
+try:
+    # ibis imports follow the convention used in the ibis source code
+    import ibis
+    import ibis.expr.datatypes as dt
+    import ibis.expr.operations as ops
+    import ibis.expr.schema as sch
+    import ibis.expr.types as ir
+    import sqlglot as sg
+    from ibis.backends import _get_backend_names, NoUrl, NoExampleLoader
+    from ibis.backends.sql import SQLBackend
+    from ibis.formats import TypeMapper
+except ImportError:
+    raise MissingDependencyException("dlt ibis helpers", ["ibis-framework"])
 
-# TODO move this as a destination capability; the destination should declare if it's compatible
-SUPPORTED_DESTINATIONS = (
-    "postgres",
-    "snowflake",
-    "filesystem",  # uses duckdb
-    "mssql",
-    "bigquery",
-    "redshift",  # uses postgres
-    "motherduck",  # uses duckdb
-)
-
-ALL_IBIS_BACKENDS = set(_get_backend_names())
+# TODO move `dlt.helpers.ibis` content to this module
 
 
 # TODO complete implementation and register to singledispatch `sch.infer.register`
@@ -46,7 +39,6 @@ class DltType(TypeMapper):
     def from_ibis(cls):
         raise NotImplementedError
 
-
     @classmethod
     def to_ibis(cls, typ: TDataType, nullable: bool | None = None) -> dt.DataType:
         nullable = True if nullable is None else bool(nullable)
@@ -54,19 +46,55 @@ class DltType(TypeMapper):
 
 
 # TODO support `database` kwarg (equiv. `dataset_name`) to enable DltBackend to access multiple database
-# NOTE this should support all SQLGlot dialects https://sqlglot.com/sqlglot/dialects.html#Dialect
+# NOTE most of the code below was derived from the Ibis BaseBackend and SQLBackend implementation
+class _DltBackend(SQLBackend, NoUrl, NoExampleLoader):
+    """Ibis backend that delegates execution to dlt's native SQL engine.
 
-class DltBackend(SQLBackend, NoUrl, NoExampleLoader):
+    To make Ibis expressions executable, they need to be bound to a table on the
+    backend (i.e., data that exists somewhere). By default, Ibis backends work by
+    creating and maintaining a connection to the backend.
+
+    This `DltBackend` is "lazier" and doesn't make a connection to the backend where
+    the data lives ("backend" in Ibis, "destination" in dlt). Instead, it uses dlt
+    metadata to register what table exists and create bound expressions.
+
+    For example, if you use Ibis with the Snowflake backend, you will get use Ibis'
+    implementation. If you use the DltBackend with a Snowflake destination, you will
+    use the dlt SQL client for Snowflake. The SQL query created should be the same
+    because dlt internally uses the `ibis -> sqlglot` compiler, but the actual
+    execution might differ. This is especially likely when calling `.execute()` or
+    other methods that return data in memory.
+    """
+
     name = "dlt"
     supports_temporary_tables = False
     supports_python_udfs = False
 
+    @classmethod
+    def from_dataset(cls, dataset: dlt.Dataset) -> _DltBackend:
+        """Create an Ibis `DltBackend` from a `dlt.Dataset`
+
+        This enables `dlt.Relation.to_ibis()` to create bound tables.
+        """
+        # NOTE it's essential to keep this lazy, no round-trip to destination
+        # sync with destination should be made through the dataset.
+        new_backend = cls()
+        new_backend._dataset = dataset
+        new_backend.compiler = _get_ibis_to_sqlglot_compiler(new_backend._dataset._destination)
+        return new_backend
+
+    def to_native_ibis(self, *, read_only: bool = False) -> BaseBackend:
+        return get_native_ibis_backend(self._dataset, read_only=read_only)
+
     @property
+    @override
     def version(self) -> str:
         import importlib.metadata
+
         return importlib.metadata.version("dlt")
 
     # NOTE can change signature
+    @override
     def do_connect(
         self,
         destination: TDestinationReferenceArg,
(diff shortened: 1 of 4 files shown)


tests/libs/test_ibis.py
```diff
@@ -0,0 +1,153 @@
...
+    "populated_pipeline",
+    duckdb_conf,
+    indirect=True,
+    ids=lambda x: x.name,
+)
+def test_user_workflow(populated_pipeline: dlt.Pipeline):
+    expected_columns = ["_dlt_id", "id"]
+
+    dataset = populated_pipeline.dataset()
+    con = dataset.ibis()
```
rudolfix: you should close this at the end of tests. windows tests are failing because of that:
```
2025-10-13T19:51:46.5223332Z tests/libs/test_ibis.py::test_instantiate_backend PASSED                 [ 55%]
2025-10-13T19:51:48.9647088Z tests/libs/test_ibis.py::test_connect_to_backend[duckdb-no-staging] PASSED [ 55%]
2025-10-13T19:51:48.9701902Z tests/libs/test_ibis.py::test_list_tables[duckdb-no-staging] PASSED      [ 55%]
2025-10-13T19:51:48.9850111Z tests/libs/test_ibis.py::test_get_schema[duckdb-no-staging] PASSED       [ 55%]
2025-10-13T19:51:48.9965690Z tests/libs/test_ibis.py::test_get_bound_table[duckdb-no-staging] PASSED  [ 55%]
2025-10-13T19:51:49.0406419Z tests/libs/test_ibis.py::test_execute_expression[duckdb-no-staging] PASSED [ 55%]
2025-10-13T19:51:49.0975756Z tests/libs/test_ibis.py::test_user_workflow[duckdb-no-staging] PASSED    [ 55%]
2025-10-13T19:51:49.1642700Z tests/libs/test_parquet_writer.py::test_parquet_writer_schema_evolution_with_big_buffer ERROR [ 55%]
```
windows does not allow for the file to be deleted if it is open. linux allows to remove the open node without problems. a lot of the problems of this kind come out. also a test that makes sure that `duckdb` is really closed after you close ibis connection with dlt backend inside makes a lot of sense

tests/libs/test_ibis.py
```diff
@@ -0,0 +1,153 @@
...
+    expr = table.select("_dlt_id", "id")
+    table2 = backend.execute(expr)
+
+    assert isinstance(table2, pd.DataFrame)
+    assert expr.schema().equals(expected_schema)
+    assert set(table2.columns) == set(expected_schema.names)
+
+
+@pytest.mark.no_load
+@pytest.mark.essential
```
rudolfix: you can mark the whole module as essential

tests/dataset/test_relation.py
```diff
@@ -0,0 +1,52 @@
+import pytest
+from ibis import ir
+
+import dlt
+
+# TODO move destination-independent tests from `test_read_interfaces.py` to this module
```
rudolfix, line 7: pls skip those tests on python 3.9. I know it is EOL but I'd get rid of it separately

dlt/common/libs/ibis.py
```diff
@@ -0,0 +1,214 @@
...
+        )
+        self.compiler = _get_ibis_to_sqlglot_compiler(self._dataset._destination)  # type: ignore[arg-type]
+
+    @contextlib.contextmanager
+    # @override
+    def _safe_raw_sql(self, *args: Any, **kwargs: Any) -> Iterator[Any]:
+        yield self.raw_sql(*args, **kwargs)
+
+    # @override
+    def raw_sql(self, query: Union[str, sg.Expression], **kwargs: Any) -> Any:
```
rudolfix, line 134: TODO:
if we want to make reads efficient: `ibis` has methods to get arrow and panda frames natively. `dlt` cursor is able (in many cases) to get those without going through dbapi and list of python tuples at all. IMO this is a good followup ticket.

dlt/common/libs/ibis.py
```diff
@@ -0,0 +1,214 @@
...
+        return _to_ibis_schema(self._dataset(query).schema)
+
+    # required for marimo DataSources UI to work
+    # @override
+    def table(self, name: str, /, *, database: tuple[str, str] | str | None = None) -> ir.Table:
+        """Construct a table expression"""
+        # TODO maybe there's a more straighforward way to retrieve catalog and db
+        sql_client = self._dataset.sql_client
+        catalog = sql_client.catalog_name()
+        database = sql_client.capabilities.casefold_identifier(sql_client.dataset_name)
```
rudolfix, line 181: we should definitely add tests on all destinations ie. snowflake (upper case casefolding) and clickhouse (schema-less) and see what happens. some tests in `test_read_interfaces` may be IMO reused

dlt/common/libs/ibis.py
```diff
@@ -0,0 +1,214 @@
...
+    def raw_sql(self, query: Union[str, sg.Expression], **kwargs: Any) -> Any:
+        """Execute SQL string or SQLGlot expression using the dlt destination SQL client"""
+        with contextlib.suppress(AttributeError):
+            if isinstance(query, sg.Expression):
+                query = query.sql(dialect=self.compiler.dialect)
+            else:
+                query = sg.transpile(query, write=self.compiler.dialect)[0]
+
+        assert isinstance(query, str)
+        with self._dataset.sql_client as client:
```
rudolfix, line 142: this is cool because we close the connection after use. nice for duckdb. I do not think this will be used in any backend code that needs to optimize opening connections. so just a note

tests/libs/test_ibis.py
zilto: To be honest, I just copy-pasted these `.mark` from `tests/load/test_read_interfaces.py`; I don't know what they do. In the recent commit, I removed them all. 

What are they used for? when should I include them?

dlt/common/libs/ibis.py
zilto: > without going through dbapi and list of python tuples at all
I'm unsure about the dlt implementation, but I suggest against coupling more with Ibis. Here, `_DltBackend` is simply pass through. 

If we want to improve the dlt SQL engine, we should do that directly and the benefits will propagate everywhere.

Ibis does:
1. User ibis expression
2. Compile Ibis to SQLGlot using Ibis destination compiler
3. Transpile SQLGlot to destination SQL based on Ibis compiler
4. Execute SQL with Ibis destination SQL engine
5. Retrieve results with PyArrow / ADBC
6. Return dataframe (polars, pandas, pyarrow)

This `_DltBackend` does:
1. Get `dlt.Relation` from `dlt.Dataset`
2. Convert `dlt.Relation` to native Ibis expression
(passing back to dlt)
3. Passing back the Ibis expression to `dlt.Dataset` to get a new `dlt.Relation`
(if executing with ibis)
3. Compile Ibis to SQLGlot using Ibis destination compiler (that matches the dlt destination)
4. Transpile SQLGlot to destination SQL based on Ibis compiler
5. Execute SQL with dlt destination SQL engine
6. Retrieve results with dlt destination mechanism
7. Return dataframe (polars, pandas, pyarrow)

dlt/common/libs/ibis.py
zilto: I really want to write tests in a way that we don't have to parameterize per destination. Each destination-specific test is something I can't run locally and prevents me from developing efficiently.

### Testing `dlt.common.­libs.ibis`
The only coupling points I see between `_DltBackend` and dlt destinations are:

1. `_get_ibis_to_sqlglot_compiler(dlt.Dataset._destination)` does return a valid and adequate Ibis compile
2. `dlt.Dataset.dataset_name == _DltBackend.current_database`
3. `dlt.Schema.tables.keys() == tables in the current database schema (i.e., dataset)`
4. the code you highlighted: being able to retrieve `catalog` and `database` from `SqlClientBase` subclasses. I expect this to be well-tested elsewhere

Since we're just proxying Ibis -> SQGlot -> dlt, we shouldn't need tests that make connections to the dlt destination (those features should be tested elsewhere). 

### About `test_read_interfaces`
I think a lot of these tests are redundant. What they truly test if whether our `dlt.Relation -> SQLGlot` conversion is adequate. We don't need to test execution on backend. All the dataset creation is what clogs our CI environment too

dlt/common/libs/ibis.py
```diff
@@ -0,0 +1,233 @@
...
+        self.compiler = _get_ibis_to_sqlglot_compiler(self._dataset._destination)  # type: ignore[arg-type]
+
+    # derived from Ibis Snowflake implementation
+    @contextlib.contextmanager
+    # @override
+    def _safe_raw_sql(self, query: sge.ExpOrStr, **kwargs: Any) -> Any:
+        with contextlib.suppress(AttributeError):
+            query = _transpile(query, target_dialect=self.compiler.dialect)
+
+        with contextlib.closing(self.raw_sql(query, **kwargs)) as cur:
```
rudolfix: your implementation of `raw_sql` is yielding a list of tuples (rows), not a cursor. that's why IMO tests are failing now.

dlt/common/libs/ibis.py
rudolfix: yes, I know how it works. And we already implement native arrow and panda readers for many destinations (incl. snowflake and duckdb). there are tickets for postgres + ADBC. Whenever you write `arrow` or `iter_arrow` on the dataset and native reader is implemented in `DBApiCursor` - dlt will skip the `dbapi` calls.

dlt/common/libs/ibis.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # derived from Ibis Snowflake implementation
+    # @override
+    def raw_sql(self, query: Union[str, sg.Expression], **kwargs: Any) -> Any:
+        """Execute SQL string or SQLGlot expression using the dlt destination SQL client"""
+        with contextlib.suppress(AttributeError):
+            query = _transpile(query, target_dialect=self.compiler.dialect)
+
+        assert isinstance(query, str)
+        with self._dataset.sql_client as client:
```
rudolfix, line 142: why you cannot call `dataset.query` instead. this will also transpile the query for you (IMO).

dlt/common/libs/ibis.py
```diff
@@ -0,0 +1,233 @@
...
+            result = client.execute_sql(query)
```
rudolfix, line 143: note: this returns rows, not cursor

dlt/common/libs/ibis.py
zilto: `dataset.query()` doesn't execute anything; it produces a `dlt.Relation`.

I agree that we could improve coupling for transpiling queries

dlt/common/libs/ibis.py
zilto: I am not able to fix this in a reasonable time. Will keep for future work.

Ibis backends assume there exists a `.con` attribute that is [DB API 2.0 compliant](https://peps.python.org/pep-0249/). I think it would be generally good to respect it for `SqlClientBase`. A lot of Python library do (e.g., psycopg, marimo)

tests/dataset/test_relation.py
zilto: I can't figure out how to do this. Given I need to import `ibis` at the module level, the pytest test collector directly crash

tests/libs/test_ibis.py
zilto: What do I need to do to make things work with Windows? I understand the issue, but I have no clue what needs to be done to fix this.

tests/libs/test_ibis.py
```diff
@@ -0,0 +1,182 @@
...
+    duckdb_conf,
+    indirect=True,
+    ids=lambda x: x.name,
+)
+def test_user_workflow(populated_pipeline: dlt.Pipeline):
+    expected_columns = ["_dlt_id", "id"]
+
+    dataset = populated_pipeline.dataset()
+    # use context manager to close conn to data source
+    with dataset.ibis() as con:
```
zilto: didn't know you could use it as a context manager! Cool solution

Issue #3154, opened 2025-10-01
Switch preprocess_docs script from JS to python


alkaline-0:
### Documentation description

Convert preprocess_docs from JS to python. 
To test this conversion the following steps are recommended:

1. Run the javascript version
2. commit processed docs to git (temporarily)
3. Run the python version
4. See if there are any differences in the output, very easy to do as git will show you exactly what is not right

### Are you a dlt user?

None

zilto:
Closing since this was moved to `Done` on project board.

Issue #3046, opened 2025-09-01
Generate destination docs section from DestinationCapabilitiesContext object


sh-rp:
### TLDR
Every destination provides a DestinationCapabilitiesContext instance that tells dlt internally how to behave when normalizing data for and loading data to this destination. There is a lot of information in this object that we could use to automatically generate a config section for the website to explain to the user which features this destination supports. For example `preferred_loader_file_format` tells dlt, which file_format for loading data is used if none is explicitely selected, or `supports_tz_aware_datetime` tells dlt wether the destination supports datetime columns that store a timezone.

### Steps
Unfortunately all our docs scripts are written in js at this moment, so we can't integrate these changes into docs/`website/tools/preprocess_docs.js` but need to create an additional python script which is run after this preprocess step. Ideally preprocess docs would be written in python, maybe we can do this to soonish.
* Familiarize yourself with the docs build process 'docs/website/README.md' in the core repo. If you see any outdated info in there, please update the README file. You can check out the packages.json to see which scripts are run when the website is built and deployed. There are some tools that modify the markdown files to insert code snippets.
* Create a new script `website/tools/insert_destination_capabilities.py`
* When run, this script should inspect all files in the folder docs/website/docs_processed, these are the processed markdown files that are served as our docs at dlthub.com/docs or locally when you run `npm start`. If a marker named `<!--@@@DLT_DESTINATION_CAPABILITIES <destination_name>-->` is found, a new markdown table should be inserted here with information about this destination. See how something similar is done with `<!--@@@DLT_SNIPPET <snippet_name>-->` markers.
* You can get the capabilities object of each destination like this (duckdb example):
```py
from dlt.destinations import duckdb
caps = duckdb.capabilities()
```
This might only work if destination credentials are provided, so consider making the _raw_capabilities() method a public and static method and use that to get the destination capabilities of a destination.
* Add a `<!--@@@DLT_DESTINATION_CAPABILITIES <destination_name>-->` marker on each destination page such as docs/website/docs/dlt-ecosystem/destinations/duckdb.md.
* For a start render a table there that includes the following information for each destination:
  * preferred_loader_file_format
  * supported_loader_file_formats
  * preferred_staging_file_format
  * supported_staging_file_formats
  * has_case_sensitive_identifiers
  * supported_merge_strategies
  * supported_replace_strategies
  * supports_tz_aware_datetime
  * supports_naive_datetime
* Add this new script in package.json right after every time we call `tools/preprocess_docs.js`

[...]

PR #3188, opened 2025-10-09
Feat/3154 convert script preprocess docs to python and add destination capabilities section to destination pages

alkaline-0:
Description
Convert preprocess_docs from javascript to python as previously suggested in ticket https://github.com/dlt-hub/dlt/issues/3046 and https://github.com/dlt-hub/dlt/issues/3154.

Changes:

1. tools/preprocess_docs.py - Main preprocessing script with watch mode
2. tools/preprocess_destination_capabilities.py - Refactored to library module
3. package.json - Integrated Python preprocessing into npm workflow
4. requirements.txt - Added python-debouncer>=0.1.0
5. tools/preprocess_tuba.py - Added caching (40s → 0.2s performance improvement)
6. tools/preprocess_snippets.py - Fixed line reading for consistency
7. tools/constants.py - Added destination capabilities constants
8. tools/preprocess_docs.js - Removed

Related Issues
https://github.com/dlt-hub/dlt/issues/3154

Additional Context
The preprocessing needs to interact with dlt's Python codebase (e.g., dlt.common.destination.capabilities), which was being done via subprocess calls. Native Python allows direct imports.

### DIFF
--- a/docs/website/tools/constants.py
+++ b/docs/website/tools/constants.py
@@ -47,7 +47,10 @@ CAPABILITIES_DOC_LINK_PATTERNS = [
     ("table", "../table-formats/"),
     ("file_format", "../file-formats/"),
     ("merge", "../../general-usage/merge-loading#merge-strategies"),
-    ("replace", "../../general-usage/full-loading#the-truncate-and-insert-strategy"),
+    (
+        "replace",
+        "../../general-usage/full-loading#choosing-the-correct-replace-strategy-for-your-full-load",
+    ),
     ("time", "../../general-usage/schema#handling-of-timestamp-and-time-zones"),
     ("dialect", "../../general-usage/dataset-access/dataset"),
     (
--- a/docs/website/tools/preprocess_destination_capabilities.py
+++ b/docs/website/tools/preprocess_destination_capabilities.py
@@ -91,7 +91,7 @@ def _format_doc_link_for_key(key: str, link: str) -> str:
     if key == "merge":
         return f"[Merge strategy]({link})"
     if key == "replace":
-        return f"[Truncate-and-insert strategy]({link})"
+        return f"[Replace strategy]({link})"
     if key == "time":
         return f"[Timestamps and Timezones]({link})"
     if key == "dialect":


docs/website/tools/preprocess_change.py
```diff
@@ -0,0 +1,125 @@
...
+def _handle_change(file_path: str) -> None:
+    """Process a single file change."""
+    try:
+        from .preprocess_docs import (
+            preprocess_docs,
+            check_docs,
+            process_doc_file,
+            process_example_change,
+        )
+        from .constants import WATCH_EXTENSIONS
```
sh-rp: as a general rule, please move all imports to the top of the python file unless there is a good reason not to (for example if some things only work if certain dependencies installed). As far as I can see, there is no reason to do this anywhere here.

docs/website/tools/preprocess_change.py
```diff
@@ -0,0 +1,125 @@
...
+
+def handle_change(file_path: str) -> None:
+    """Queue a file change event for processing."""
+    with _processing_lock:
+        if file_path in _processing_files:
+            print(f"File {file_path} already queued/processing, skipping")
+            return
+        _processing_files.add(file_path)
+
+    event = FileChangeEvent(file_path=file_path, timestamp=datetime.now())
```
sh-rp: This queue is a smart solution, it does not capture all changes properly though. Consider for example the snippets that get injected into the .md files. If you change a snippet (which is in a python file) the associated .md file should be updated, this does not happen now. My take:
* Remove the queue
* On any change, rebuild all docs, it is fast enough imho
* If the user changes anything while the script is running, discard that iteration by checking if the lock is locked. Alternatively you could queue this, but then we only queue one item ever so we don't have multiple runs when they are not needed. Ideally try to keep the code as simple as possible.

docs/website/tools/constants.py
```diff
@@ -0,0 +1,61 @@
...
+SELECTED_CAPABILITIES_ATTRIBUTES = {
+    "preferred_loader_file_format",
+    "supported_loader_file_formats",
+    "preferred_staging_file_format",
+    "supported_staging_file_formats",
+    "has_case_sensitive_identifiers",
+    "supported_merge_strategies",
+    "supported_replace_strategies",
+    "supports_tz_aware_datetime",
+    "supports_naive_datetime",
```
sh-rp, line 40: It's a good selection! Let's also add:
* preferred and selected table format
* sqlglot_dialect

docs/website/package.json
```diff
@@ -4,24 +4,25 @@
...
+    "preprocess-docs": "PYTHONPATH=. uv run python -u -m tools.preprocess_docs",
+    "preprocess-docs:cloudflare": "PYTHONPATH=. python -u -m tools.preprocess_docs",
     "render-cli-docs": "PYTHONPATH=. uv run dlt render-docs > docs/reference/command-line-interface-generated.md",
-    "build:cloudflare": "npm run update-versions && npm run preprocess-docs && npm run gen-api-ref-netlify && docusaurus build --out-dir build/docs",
+    "build:cloudflare": "npm run update-versions  && npm run preprocess-docs:cloudflare && npm run gen-api-ref:cloudflare &&  docusaurus build --out-dir build/docs",
+    "build:netlify": "npm run update-versions && npm run preprocess-docs:cloudflare && npm run gen-api-ref:cloudflare &&  docusaurus build --out-dir build/docs",
     "serve:cloudflare": "wrangler dev --port 3001 --local",
     "preview:cloudflare": "npm run build:cloudflare && npm run serve:cloudflare",
-    "deploy:cloudflare": "npm run build:cloudflare && wrangler deploy"
+    "deploy:netlify": "npm run build:cloudflare && wrangler deploy"
```
sh-rp: this should be deploy:cloudflare

docs/website/package.json
```diff
@@ -4,24 +4,25 @@
...
-    "preprocess-docs": "node tools/preprocess_docs.js",
     "gen-api-ref": "PYTHONPATH=. uv run pydoc-markdown && uv run python clean_pydoc_sidebar.py",
-    "gen-api-ref-netlify": "PYTHONPATH=. pydoc-markdown && python clean_pydoc_sidebar.py",
+    "gen-api-ref:cloudflare": "PYTHONPATH=. pydoc-markdown && python clean_pydoc_sidebar.py",
+    "preprocess-docs": "PYTHONPATH=. uv run python -u -m tools.preprocess_docs",
+    "preprocess-docs:cloudflare": "PYTHONPATH=. python -u -m tools.preprocess_docs",
     "render-cli-docs": "PYTHONPATH=. uv run dlt render-docs > docs/reference/command-line-interface-generated.md",
-    "build:cloudflare": "npm run update-versions && npm run preprocess-docs && npm run gen-api-ref-netlify && docusaurus build --out-dir build/docs",
+    "build:cloudflare": "npm run update-versions  && npm run preprocess-docs:cloudflare && npm run gen-api-ref:cloudflare &&  docusaurus build --out-dir build/docs",
+    "build:netlify": "npm run update-versions && npm run preprocess-docs:cloudflare && npm run gen-api-ref:cloudflare &&  docusaurus build --out-dir build/docs",
```
sh-rp, line 22: we can remove this one after this change has been merged to devel and we switch to the build:cloudlfare command on cloudflare. this is just a detail of moving from netlify to cloudflare

docs/website/tools/preprocess_change.py
```diff
@@ -0,0 +1,125 @@
...
+            process_example_change(file_path)
+        elif "docs" in file_path and ext in WATCH_EXTENSIONS:
+            process_doc_file(file_path)
+    finally:
+        with _processing_lock:
+            _processing_files.discard(file_path)
+
+
+def watch() -> None:
+    """Start watching for file changes."""
```
sh-rp: this can be greatly simplified if we remove the queue. I would do something like this (pseudo code):

```py
thread_lock = Lock() # thread lock
pending_changes = False # can be set to "True" if there are any changes detected during a run.

def process_all_files():
   ... # process all files


def process_all_files_locked():
   with thread_lock():
       process_all_files()

def on_file_change():
   """gets called by watcher on any file change"""
   if thread_lock.locked:
      pending_changes = True # update is running, we remember that there was a change, could be called multiple times during a run
      return

    # run as long as there are consecutive pending changes
    pending_changes = True
    while pending_changes:
       pending_changes = False
       process_all_files_locked()
   

```

docs/website/docs/dlt-ecosystem/destinations/weaviate.md
```diff
@@ -9,6 +9,13 @@ keywords: [weaviate, vector database, destination, dlt]
 [Weaviate](https://weaviate.io/) is an open-source vector database. It allows you to store data objects and perform similarity searches over them.
 This destination helps you load data into Weaviate from [dlt resources](../../general-usage/resource.md).
 
+
+## Destination capabilities
```
sh-rp: Not super important, but the generator for the destination capabilities table could also create this title and the sentence below, otherwise you have the exact same sentence copied in each doc

PR #3191, opened 2025-10-10
databricks: removes cluster on create table test and allows only partition

rudolfix:
### Description
We support `cluster` and `partition` hints on databricks and one of the tests that was setting both on a single table started to fail.

`full ci` label added to run all databrick tests

### DIFF
--- a/tests/load/test_job_client.py
+++ b/tests/load/test_job_client.py
@@ -252,7 +252,7 @@ def test_complete_load(naming: str, client: SqlJobClientBase) -> None:
 
 @pytest.mark.parametrize(
     "client",
-    destinations_configs(default_sql_configs=True),
+    destinations_configs(default_sql_configs=True, exclude=["bigquery"]),
     indirect=True,
     ids=lambda x: x.name,
 )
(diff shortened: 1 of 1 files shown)


tests/load/test_job_client.py
```diff
@@ -252,7 +252,7 @@ def test_complete_load(naming: str, client: SqlJobClientBase) -> None:
 
 @pytest.mark.parametrize(
     "client",
-    destinations_configs(default_sql_configs=True),
+    destinations_configs(default_sql_configs=True, exclude=["bigquery"]),
```
ivasio, line 255: Why exclude bigquery? I though this test was failing only with destination_type = databricks?

rudolfix: we have a bigquery variant below so this one is just repeating one of the cases of `test_schema_update_create_table_bigquery_hidden_dataset`. why a separate test for bigquery: you can have hidden datasets in bigquery. in that case we use very different code path to get schema of the dataset. they are not available in INFORMATION SCHEMA.

Issue #3103, opened 2025-09-19
Ensure consistency in HexBytes coercion


sh-rp:
TLDR
We use the hexbytes library to convert binary datafields into HexStrings. You can see the relevant test changed and discussed in this PR: https://github.com/dlt-hub/dlt/pull/3100. 

It turns out that newer versions of the HexBytes library do not prepend the strings with "0x". We want consistent behavior, so we need to:

* Replace HexBytes with our own implementation
* The coercion test (modified in the linked PR) should be reverted to its old form
*  Remove HexBytes from our pyproject.toml and update uv.lock

PR #3200, opened 2025-10-13
feat/3103: Ensure consistency in HexBytes coercion 

alkaline-0:
**Description**
The newer versions of the HexBytes library do not prepend the strings with "0x".  To ensure the consistency of behavior, I removed the dependency on the library hexbytes and added our own implementation with unit tests as instructed in [ticket-3103](https://github.com/dlt-hub/dlt/issues/3103)

**Changes:**
1. Added custom implementation of `HexBytes` class to `dlt/common/libs/hexbytes.py`
2. Added test suite in `tests/common/libs/test_hexbytes.py`
3. Fixed import sorting in affected files.

**Related Ticket:** 
[ticket-3103](https://github.com/dlt-hub/dlt/issues/3103)

### DIFF
--- a/dlt/common/libs/hexbytes.py
+++ b/dlt/common/libs/hexbytes.py
@@ -28,9 +28,9 @@ class HexBytes(bytes):
         bytesval = HexBytes._to_bytes(val)
         return cast(HexBytes, super().__new__(cls, bytesval))  # type: ignore  # https://github.com/python/typeshed/issues/2630  # noqa: E501
 
-    def hex(
+    def hex(  # noqa: A003
         self, sep: Union[str, bytes] = None, bytes_per_sep: "SupportsIndex" = 1
-    ) -> str:  # noqa: A003
+    ) -> str:
         """
         Output hex-encoded bytes, with an "0x" prefix.
         Everything following the "0x" is output exactly like :meth:`bytes.hex`.


dlt/common/libs/hexbytes.py
```diff
@@ -0,0 +1,104 @@
...
+    def hex(  # noqa: A003
+        self, sep: Union[str, bytes] = None, bytes_per_sep: "SupportsIndex" = 1
+    ) -> str:
+        """
+        Output hex-encoded bytes, with an "0x" prefix.
+
+        Everything following the "0x" is output exactly like :meth:`bytes.hex`.
+        """
+        raw_hex = super().hex()
+        if raw_hex.startswith("0x"):
```
rudolfix: bytes will never return a `0x` prefix so you do not need this condition

dlt/common/libs/hexbytes.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @overload  # noqa: F811
+    def __getitem__(self, key: slice) -> "HexBytes":  # noqa: F811
+        ...
+
+    def __getitem__(  # noqa: F811
+        self, key: Union["SupportsIndex", slice]
+    ) -> Union[int, bytes, "HexBytes"]:
+        result = super().__getitem__(key)
+        if hasattr(result, "hex"):
```
rudolfix: this is so weird... why not just return:
`return result if isinstance(key, int) else self.__class__(result)`
?

dlt/common/libs/hexbytes.py
```diff
@@ -0,0 +1,104 @@
...
+            if val < 0:
+                raise ValueError(f"Cannot convert negative integer {val} to bytes")
+            else:
+                return HexBytes._to_bytes(hex(val))
+        elif isinstance(val, memoryview):
+            return bytes(val)
+
+    @staticmethod
+    def _hexstr_to_bytes(hexstr: str) -> bytes:
+        if hexstr.startswith(("0x", "0X")):
```
rudolfix: you can use `removeprefix`

dlt/common/libs/hexbytes.py
```diff
@@ -0,0 +1,104 @@
...
+            non_prefixed_hex = hexstr[2:]
+        else:
+            non_prefixed_hex = hexstr
+
+        if len(hexstr) % 2:
```
rudolfix: use `non_prefixed_hex` here

dlt/common/libs/hexbytes.py
```diff
@@ -0,0 +1,104 @@
...
+            padded_hex = "0" + non_prefixed_hex
+        else:
+            padded_hex = non_prefixed_hex
+
+        try:
+            ascii_hex = padded_hex.encode("ascii")
+        except UnicodeDecodeError:
```
rudolfix: wrong exception! `UnicodeEncodeError` 🤯 also there's `bytes.fromhex` that does everything

dlt/common/libs/hexbytes.py
```diff
@@ -0,0 +1,104 @@
...
+        Convert a hex string, integer, or bool, to a bytes representation.
+        Alternatively, pass through bytes or bytearray as a bytes value.
+        """
+        if isinstance(val, bytes):
+            return val
+        elif isinstance(val, str):
+            return HexBytes._hexstr_to_bytes(val)
+        elif isinstance(val, bytearray):
+            return bytes(val)
+        elif isinstance(val, bool):
```
rudolfix: please remove `bool` and `int` from bytes-like and do not support this conversion. converting `int` to `bytes` should always take into account endianness and there are methods on `int` type to do that already. what happens here is just misleading

dlt/common/libs/hexbytes.py
```diff
@@ -0,0 +1,104 @@
...
+        if hasattr(result, "hex"):
+            return type(self)(result)
+        else:
+            return result
+
+    def __repr__(self) -> str:
+        return f"HexBytes({self.hex()!r})"
+
+    @staticmethod
+    def _to_bytes(val: Union[bool, bytearray, bytes, int, str, memoryview]) -> bytes:
```
rudolfix: why `BytesLike` is not used here?

rudolfix: one more thing: we should just override `fromhex` and not adding a new method

PR #3204, opened 2025-10-14
Feat: dlt.destination as named destination factory

### LINKED ISSUE #3189: repurpose @dlt.destination decorator to also serve as named destination factory
**Background**
We'd like to use the same pattern we have in `dlt.pipeline` and `dlt.dataset` or `dlt.resource` (user friendly "constructors" for Pipeline, Dataset and DltResource classes) for creating destinations. We want to use `dlt.destination` for that.
`dlt.destination` already is used as decorator so we are extending existing construct. We already have `dlt.resource` that may be used as decorator (converts a function into a resource) or as a factory (converts any data into a resource).

With `dlt.destination` we support the following
1. `@dlt.destination` - converts a sink function into a custom destination - returns custom destination factory
2. overload that will be used as contstructor
```py
@overload
def destination(
    destination_name: str,
    /,
    *,
    destination_type: Optional[str] = None,
    credentials: Optional[Any] = None,
    kwargs: Any,
) -> AnyDestination:
```

example
```py
import dlt
d_ = dlt.destination("warehouse")
d_ = dlt.destination("warehouse", destination_type="ducklake")
```

**Requirements**
1. * [ ] is backward compatible in decorator mode: in terms of executed code, type checks (mypy) and docstrings available when coding 
2. * [ ] creates destination factory using `Destination.from_reference` is positional argument is instance of str.
3. * [ ] our docs and examples are updated: `Destination.from_reference` is removed in favor of a new constructs

**Implementation**
Here's result of my hacking (making sure such double use is actually possible).
- overloads sends data to **kwargs of implementation from which you should pass data to from_reference
- docstring not complete for new overloads
- other overloads will still use docstrings on implementation

```py
import functools

import inspect
from typing import Any, Type, Optional, Callable, Union, overload
from typing_extensions import Concatenate
from dlt.common.destination.reference import AnyDestination, Destination, DestinationReference
from dlt.common.reflection.spec import get_spec_name_from_f
from dlt.common.typing import AnyFun

from functools import wraps

from dlt.common import logger
from dlt.common.destination import TLoaderFileFormat
from dlt.common.typing import TDataItems
from dlt.common.schema import TTableSchema
from dlt.common.destination.capabilities import TLoaderParallelismStrategy

from dlt.common.utils import get_callable_name, get_full_callable_name, is_inner_callable
from dlt.destinations.impl.destination.factory import destination as _destination
from dlt.destinations.impl.destination.configuration import (
    TDestinationCallableParams,
    CustomDestinationClientConfiguration,
)


@overload
def destination(
    func: Callable[
        Concatenate[Union[TDataItems, str], TTableSchema, TDestinationCallableParams], Any
    ],
    /,
    *,
    loader_file_format: TLoaderFileFormat = None,
    batch_size: int = 10,
    name: str = None,
    naming_convention: str = "direct",
    skip_dlt_columns_and_tables: bool = True,
    max_table_nesting: int = 0,
    spec: Type[CustomDestinationClientConfiguration] = None,
    max_parallel_load_jobs: Optional[int] = None,
    loader_parallelism_strategy: Optional[TLoaderParallelismStrategy] = None,
) -> Callable[TDestinationCallableParams, _destination]: ...


@overload
def destination(
    func: None = ...,
    /,
    *,
    loader_file_format: TLoaderFileFormat = None,
    batch_size: int = 10,
    name: str = None,
    naming_convention: str = "direct",
    skip_dlt_columns_and_tables: bool = True,
    max_table_nesting: int = 0,
    spec: Type[CustomDestinationClientConfiguration] = None,
    max_parallel_load_jobs: Optional[int] = None,
    loader_parallelism_strategy: Optional[TLoaderParallelismStrategy] = None,
) -> Callable[
    [Callable[Concatenate[Union[TDataItems, str], TTableSchema, TDestinationCallableParams], Any]],
    Callable[TDestinationCallableParams, _destination],
]: ...


@overload
def destination(
    destination_name: str,
    /,
    *,
    destination_type: Optional[str] = None,
    credentials: Optional[Any] = None,
    kwargs: Any,
) -> AnyDestination:
    """A destination constructor"""


def destination(
    func: Union[Optional[AnyFun], str] = None,
    /,
    loader_file_format: TLoaderFileFormat = None,
    batch_size: int = 10,
    name: str = None,
    naming_convention: str = "direct",
    skip_dlt_columns_and_tables: bool = True,
    max_table_nesting: int = 0,
    spec: Type[CustomDestinationClientConfiguration] = None,
    max_parallel_load_jobs: Optional[int] = None,
    loader_parallelism_strategy: Optional[TLoaderParallelismStrategy] = None,
    **kwargs: Any,
) -> Any:
    """A decorator that transforms a function that takes two positional arguments "table" and "items" and any number of keyword arguments with defaults
    into a callable that will create a custom destination. The function does not return anything, the keyword arguments can be configuration and secrets values.

    #### Example Usage with Configuration and Secrets:

    >>> @dlt.destination(batch_size=100, loader_file_format="parquet")
    >>> def my_destination(items, table, api_url: str = dlt.config.value, api_secret = dlt.secrets.value):
    >>>     print(table["name"])
    >>>     print(items)
    >>>
    >>> p = dlt.pipeline("chess_pipeline", destination=my_destination)

    Here all incoming data will be sent to the destination function with the items in the requested format and the dlt table schema.
    The config and secret values will be resolved from the path destination.my_destination.api_url and destination.my_destination.api_secret.

    Args:
        func (Optional[AnyFun]): A function that takes two positional arguments "table" and "items" and any number of keyword arguments with defaults which will process the incoming data.

        loader_file_format (TLoaderFileFormat): defines in which format files are stored in the load package before being sent to the destination function, this can be puae-jsonl or parquet.

        batch_size (int): defines how many items per function call are batched together and sent as an array. If you set a batch-size of 0, instead of passing in actual dataitems, you will receive one call per load job with the path of the file as the items argument. You can then open and process that file in any way you like.

        name (str): defines the name of the destination that gets created by the destination decorator, defaults to the name of the function

        naming_convention (str): defines the name of the destination that gets created by the destination decorator. This controls how table and column names are normalized. The default is direct which will keep all names the same.

        skip_dlt_columns_and_tables (bool): defines wether internal tables and columns will be fed into the custom destination function. This is set to True by default.

        max_table_nesting (int): defines how deep the normalizer will go to normalize nested fields on your data to create subtables. This overwrites any settings on your source and is set to zero to not create any nested tables by default.

        spec (Type[CustomDestinationClientConfiguration]): defines a configuration spec that will be used to to inject arguments into the decorated functions. Argument not in spec will not be injected

        max_parallel_load_jobs (Optional[int]): how many load jobs at most will be running during the load

        loader_parallelism_strategy (Optional[TLoaderParallelismStrategy]): Can be "sequential" which equals max_parallel_load_jobs=1, "table-sequential" where each table will have at most one loadjob at any given time and "parallel"
    Returns:
        Any: A callable that can be used to create a dlt custom destination instance
    """

    def decorator(
        destination_callable: Callable[
            Concatenate[Union[TDataItems, str], TTableSchema, TDestinationCallableParams], Any
        ]
    ) -> Callable[TDestinationCallableParams, _destination]:
        # resolve destination name
        destination_name = name or get_callable_name(destination_callable)

        # synthesize new Destination factory
        class _ConcreteDestinationBase(_destination):
            def __init__(self, **kwargs: Any):
                default_args = dict(
                    spec=spec,
                    destination_callable=destination_callable,
                    loader_file_format=loader_file_format,
                    batch_size=batch_size,
                    destination_name=destination_name,
                    naming_convention=naming_convention,
                    skip_dlt_columns_and_tables=skip_dlt_columns_and_tables,
                    max_table_nesting=max_table_nesting,
                    max_parallel_load_jobs=max_parallel_load_jobs,
                    loader_parallelism_strategy=loader_parallelism_strategy,
                )
                super().__init__(**{**default_args, **kwargs})

        cls_name = get_spec_name_from_f(destination_callable, kind="Destination")
        module = inspect.getmodule(destination_callable)
        # synthesize type
        D: Type[_destination] = type(
            cls_name,
            (_ConcreteDestinationBase,),
            {"__module__": module.__name__, "__orig_base__": _destination},
        )
        # add to the module
        setattr(module, cls_name, D)
        # register only standalone destinations, no inner
        if not is_inner_callable(destination_callable):
            DestinationReference.register(D, get_full_callable_name(destination_callable))

        @wraps(destination_callable)
        def wrapper(
            *args: TDestinationCallableParams.args, **kwargs: TDestinationCallableParams.kwargs
        ) -> _destination:
            if args:
                logger.warning(
                    "Ignoring positional arguments for destination callable %s",
                    destination_callable,
                )
            return D(**kwargs)  # type: ignore[arg-type]

        setattr(wrapper, "_factory", D)  # noqa
        return wrapper

    if func is None:
        # we're called with parens.
        return decorator
    elif isinstance(func, str):
        # extract from kwargs ie by popping etc
        return Destination.from_reference(func)

    # we're called as @dlt.destination without parens.
    return decorator(func)

```

anuunchin:
This PR adjusts the existing `dlt.pipeline()` function to serve as a named destination factory.  For that end, a new overload is defined that passes passes relevant arguments to `Destination.from_reference()`. As a result, `Destination.from_reference()` is removed from docs, as `dlt.destination()` is deemed more idiomatic. Relevant tests are adjusted/added.

Resolves #3189

### DIFF
--- a/docs/website/docs/general-usage/destination.md
+++ b/docs/website/docs/general-usage/destination.md
@@ -21,7 +21,7 @@ Above, we want to use the **filesystem** built-in destination. You can use short
 * Use a [**named destination**](#use-named-destinations) with a configured type
 <!--@@@DLT_SNIPPET ./snippets/destination-snippets.py::custom_destination_name-->
 
-Above, we use a custom destination name and configure the destination type to **filesystem** using an environment variable. This approach is especially useful when switching between destinations without modifying the actual pipeline code. See details in the [section on using named destinations](#use-named-destinations-to-switch-destinations-without-changing-pipeline-code).
+Above, we use a custom destination name and configure the destination type to **filesystem** using an environment variable. This approach is especially useful when switching between destinations without modifying the actual pipeline code. See details in the [section on using named destinations](#use-named-destinations-to-switch-destinations-without-changing-code).
 
 * Use full **destination factory type**
 <!--@@@DLT_SNIPPET ./snippets/destination-snippets.py::class_type-->
--- a/docs/website/docs/general-usage/credentials/setup.md
+++ b/docs/website/docs/general-usage/credentials/setup.md
@@ -747,7 +747,7 @@ You have additional options for using multiple instances of the same source:
 
 1. Use the `clone()` method as explained in the [sql_database documentation](../../dlt-ecosystem/verified-sources/sql_database/advanced.md#configure-many-sources-side-by-side-with-custom-sections).
 
-2. Create [named destinations](../destination.md#configure-multiple-destinations-in-a-pipeline) to use the same destination type with different configurations.
+2. Create [named destinations](../destination.md#configure-multiple-destinations-of-the-same-type) to use the same destination type with different configurations.
 :::
 
 ## Troubleshoot configuration errors


dlt/destinations/decorators.py
```diff
@@ -23,28 +21,11 @@
...
-    naming_convention: str = "direct",
-    skip_dlt_columns_and_tables: bool = True,
-    max_table_nesting: int = 0,
-    spec: Type[CustomDestinationClientConfiguration] = None,
-    max_parallel_load_jobs: Optional[int] = None,
-    loader_parallelism_strategy: Optional[TLoaderParallelismStrategy] = None,
-) -> Callable[TDestinationCallableParams, _destination]: ...
-
-
 @overload
```
anuunchin, line 24: went with separate dosctrings for each of them so that the user can scroll down in the ide and only see the relevant docstring per overload 👀

dlt/destinations/decorators.py
```diff
@@ -165,9 +276,30 @@ def wrapper(
...
+    elif isinstance(func_or_name, str):
+        # Factory mode: create destination instance
+        destination_type = kwargs.pop("destination_type", None)
+        if destination_type:
+            return Destination.from_reference(
+                ref=destination_type,
+                destination_name=func_or_name,
+                **kwargs,
+            )
+        elif kwargs.get("destination_callable"):
```
anuunchin: thought special handling of custom destinations is needed because intuitively, the user would try to do something like
```py
dlt.destination("some_name", destination_callable=some_func) 
```
👀

tests/pipeline/test_pipeline_state.py
```diff
@@ -103,8 +102,8 @@ def test_state_repr() -> None:
 def test_restore_state_props() -> None:
     p = dlt.pipeline(
         pipeline_name="restore_state_props",
-        destination=Destination.from_reference("redshift", destination_name="redshift_name"),
-        staging=Destination.from_reference("filesystem", destination_name="filesystem_name"),
+        destination=dlt.destination("redshift_name", destination_type="redshift"),
+        staging=dlt.destination("filesystem_name", destination_type="filesystem"),
```
zilto: To promote backwards compatibility, it's better to not modify tests. Instead, add a parameterization to ensure we support both `Destination.from_reference(...)` and `dlt.destination(...)`.

OTOH, this code change tells me "`Destination.from_reference()` is being replaced / deprecated. It was an internal API and we don't consider this change breaking".

tests/load/pipeline/test_postgres.py
```diff
@@ -58,20 +58,30 @@ def test_postgres_encoded_binary(
 
 # do not remove - it allows us to filter tests by destination
 @pytest.mark.no_load
+@pytest.mark.parametrize(
+    "use_dest_decorator",
```
zilto: `use_dest_decorator` isn't the clearest name. I suggest `use_destination_factory` or `use_top_level_destination`.

You can also add details to the docstring of the test function to explain the `use_destination_factory` kwarg and its value

tests/load/pipeline/test_pipelines.py
```diff
@@ -1147,13 +1154,21 @@ def test_data():
...
     assert "no destination type was configured" in str(py_exc.value)
 
     # if destination contains dots, no fallbacks must happen
     with pytest.raises(UnknownDestinationModule) as py_exc:
-        dlt.pipeline(destination="dlt.destinations.unknown")
+        (
+            dlt.destination("dlt.destinations.unknown")
+            if use_dest_decorator
+            else dlt.pipeline(destination="dlt.destinations.unknown")
+        )
```
zilto: This works, but easier to reason about and maintain with the `if/else` outside the `with` clause IMO

```python
if use_destination_decorator:
  with pytest.raises(...) as py_exc:
    dlt.destination(...)
else:
  with pytest.raises(...) as py_exc:
    dlt.pipeline(...)
```

tests/load/duckdb/test_duckdb_client.py
```diff
@@ -538,14 +546,14 @@ def test_keeps_initial_db_path() -> None:
     os.environ["CREDENTIALS"] = db_path
 
     p = dlt.pipeline(pipeline_name="quack_pipeline", destination=dlt.destinations.duckdb())
-    assert p.state["_local"]["initial_cwd"] == os.path.abspath(os.path.curdir).lower()
+    assert p.state["_local"]["initial_cwd"] == os.path.abspath(os.path.curdir)
```
zilto: Why is this assumption changed? 

A good rule of thumb for backwards compatibility:
- `assert` are immutable.
- Add `assert` when adding features
- Add `assert` when fixing bugs (you refine your assumptions about specific behavior)
- Edit `assert` when fixing bugs (only if your test didn't actually match the intended behavior; this implies it could have been caught in the initial PR implementing the feature)
- Remove `assert` when removing a feature
- Edit `assert` when making a breaking change (the intended behavior / assumptions changed)   

Modifying a test means there's a change in behavior. Is this breaking for internal components or the user? One can be ok, the other we must avoid

dlt/destinations/decorators.py
```diff
@@ -165,9 +276,30 @@ def wrapper(
         setattr(wrapper, "_factory", D)  # noqa
         return wrapper
 
-    if func is None:
+    if func_or_name is None:
```
zilto, line 278: I try to have a single `return` statement per function especially when there are multiple nested `if/else` cases. The exception to this is when you have early exit condition at the beginning of a function (i.e., guard clauses). 

It typically has this shape:

```python
destination = None
# alternatively you can set the type with no assignment, but 
# make sure `destination` value is set by the end of function
# destination: Destination

if func_or_name is None:
  destination = decorator
elif isinstance(func_or_name, str):
  # ...
  if destination_type:
     destination = Destination.from_reference(...)
  elif kwargs.get("destination_callable"):
     destination = Destination.from_reference(...)
  else:
     destination = Destination.from_reference(...)
else:
   destination = decorator(func_or_name)
   
# mypy will check this assertion
assert isinstance(destination,Destination)

# extra safeguard (optional)
if destination is None:
  raise RuntimeError("Uh oh, this is a bug")
  
return destination
```

tests/pipeline/test_pipeline_state.py
anuunchin: Also just realized that this is wrong to be removing the usage of the "destination_name" argument in this test - fixing it now

tests/load/pipeline/test_postgres.py
anuunchin: I see how the usage of the word "decorator" is actually wrong - cause in this case it doesn't really work as a decorator

tests/load/pipeline/test_pipelines.py
anuunchin: Now that i take a look at it, i have no idea why I wrote it like that initially -> also fixed 🫡

dlt/destinations/decorators.py
anuunchin: oowwww, got it

anuunchin: I put the the stuff that existed before the new logic - it's definitely clearer now

tests/load/duckdb/test_duckdb_client.py
anuunchin: Looks like this test is not run without the full ci label at all.... I discovered it's failing (also in devel) because I happened to run the entire `test_duckdb_client.py` 👀

But I have an idea where it might be coming from - will fix in a separate PR

anuunchin:
@zilto  @rudolfix  considering the only usage pattern that may cause issues is

```py
def print_sink(items: TDataItems, table: TTableSchema) -> None:
    for item in items:
        print(1)

dest_ref = dlt.destination(print_sink, "parquet", 1)
```
which is not recommended anywhere in the docs and is not a conventional pattern, but still runs fine with the updates cause the main `def destination()` accepts positional args, I agree with @rudolfix  that it's not very problematic 👀

dlt/destinations/decorators.py
```diff
@@ -57,12 +38,71 @@ def destination(
...
+
+
+@overload
 def destination(
-    func: Optional[AnyFun] = None,
+    func: Callable[
+        Concatenate[Union[TDataItems, str], TTableSchema, TDestinationCallableParams], Any
+    ],
     /,
+    *,
```
zilto, line 105: about dlt.destination() signature

## Current
```python
def destination(func = None, /, loader_file_format = None, ..., **kwargs): ...

# allowed
destination()
destination(foo)
destination(loader_file_format="parquet")
destination(foo, "parquet")
destination(foo, loader_file_format="parquet")

# not allowed
# func is positional only; this is to ensure it's the decorated function
destination(func=foo)
destination(func=foo, loader_file_format="parquet")

```
## Change (current PR)
```python
# adding the `*`
def destination(func = None, /, *, loader_file_format = None, ..., **kwargs): ...
```

The main thing it breaks is `destination(foo, "parquet")`. Given it is used as a decorator, it would break the rare
```python
dlt.destination(print_sink, "parquet", 1)

# this can't happen because `"parquet"` would be assigned to `func` and error
@dlt.destination("parquet")
def foo(...): ...
```

## Backwards compatibility and helpful message
If it's breaking, it's good to have a "migration message" to help users. We want to explicitly catch users with `destination(func, "parquet")`. So will we use a catchall `*args`

```python
def destination(func, /, *args, loader_file_format=None, batch_size=10, ..., **kwargs):
   # we caught a user passing `destination(func, "parquet")`
   if args:
     warning.warn(DeprecationWarning("use positional arguments"))
```
But, we also know the order of the now keyword only args! So if we have `len(args)==2`, they **must** be `loader_file_format` and then `batch_size`. Also, it would be impossible for **old code** to pass `destination(func, "parquet", loader_file_format="parquet")`

```python
def destination(func, /, *args, loader_file_format = None, batch_size=None):
  # warning
  # zip will stop after the shortest iterable, i.e., `args`
  args_to_kwargs = dict(zip(kwarg_names, args))
``` 

Note that **new code** could write `destination(func, "parquet", loader_file_format="parquet")`. Therefore, you need to check / deduplicate kwargs

```python
def destination(func, /, *args, loader_file_format = None, batch_size=None):
  # locals() will only includes things received by the function call if called here
  _initial_args = locals()
  # warning
  args_to_kwargs = dict(zip(kwarg_names, args))
  
  for arg in args_to_kwargs:
     if arg in _initial_args:
        raise ValueError(f"you passed a value for `{arg}` both as positional and keyword argument. Positional args are there for backwards compatibility reasons. They shouldn't never be used and will be deprecated")
     
``` 

You can wrap all of that in a `_resolve_destination_args(initial_args, args)` function to not clutter `destination()`

anuunchin: While trying to implement the above, I came to the conclusion that we do not need to ensure backwards compatibility 👀 , for something like:
```py
dest_ref = dlt.destination(print_sink, "parquet", 10)
```
because we haven't changed the order of keywords in the main `def destination()` function, and it would still work

zilto: Interesting. I'm a bit confused about how things work when you have `@overload` defining `destination(func, /, *, format)` but the main function defines `destination(func, /, format)`. `mypy` doesn't complain?

The overloads are compatible but with more constraints. This makes the function callable, but the base function is incorrectly typed. In my short repro, I get

image

anuunchin: `Interesting. I'm a bit confused about how things work when you have @overload defining destination(func, /, *, format) but the main function defines destination(func, /, format). mypy doesn't complain?` Mypy doesn't complain because all actual calls in the codebase use keyword arguments 👀

anuunchin: @zilto  I have the following questions, could you help me resolve them? 

As suggested earlier, say, we have something like:
```py
def _resolve_args_to_kwargs(initial_args: Dict[str, Any]) -> Dict[str, Any]:
    sig = inspect.signature(destination)
    param_names = [k for k in sig.parameters.keys() if k not in {"func_or_name", "args", "kwargs"}]
    
    pos_args = initial_args["args"]
    args_to_kwargs = dict(zip(param_names, pos_args))

    for key in args_to_kwargs:
        if initial_args[key]:
            raise ValueError(f"Cannot specify `{key}` both positionally and by keyword. Positional arguments are deprecated - use keyword-only.")

    initial_args.update(args_to_kwargs)
    return initial_args
```
One of the issues I see is that at this point it would be impossible to know whether initial args include default values that the user didn't explicitly pass or default values that were explicitly passed. Meaning, in the following case:

```py
dest_ref = dlt.destination(print_sink, "parquet", 10)
```

`batch_size=10` will always be available in initial args and we would raise unnecessarily. This can be resolved with another wrapper, but that would defy the purpose of not cluttering the main function 👀. Should we just still raise in this case? If yes, shouldn't we just raise if args is not empty in all cases? (since destination initialization happens before anything happens, I'm assuming it's wouldn't be a problem for users)

The above method would also require reassigning the keyword args in `def destination()` with what has been resolved from args, because otherwise we would get unused argument errors, right? ie, something like:
```py
        initial_args = locals().copy()
        resolved_args = _resolve_args_to_kwargs(initial_args)
        loader_file_format = resolved_args["loader_file_format"]
        batch_size = resolved_args["batch_size"]
        name = resolved_args["name"]
        naming_convention = resolved_args["naming_convention"]
        skip_dlt_columns_and_tables = resolved_args["skip_dlt_columns_and_tables"]
        max_table_nesting = resolved_args["max_table_nesting"]
        spec = resolved_args["spec"]
        max_parallel_load_jobs = resolved_args["max_parallel_load_jobs"]
        loader_parallelism_strategy = resolved_args["loader_parallelism_strategy"]
```
Would this be still acceptable, even though it also defies the purpose of keeping `def destination()` clean? 👀

rudolfix: somehow mypy does not complain. I wrote this test
```py
def test_backward_signature_compat() -> None:

    @dlt.destination("parquet")
    def parquet_sink(items: TDataItems, table: TTableSchema):
        assert table["file_format"] == "parquet"

    p = dlt.pipeline("sink_test", destination=parquet_sink, dev_mode=True)
    p.run([1, 2, 3], table_name="digits")    
```
to see what happens and this obviously uses `parquet` as destination name and raises an exception. Adding `*` to signature to enforce positional arguments does not make any impact on behavior. Also the lint error is the same.

my take: we keep the current code (more or less). we could write tons of code to "parse" arguments ourselves and cover the case above but IMO it is not worth the effort. We should write good release notes and luckily problem is surfaced by typechecker and at runtime it will fail immediately

dlt/destinations/decorators.py
```diff
@@ -72,44 +112,113 @@ def destination(
...
-        Any: A callable that can be used to create a dlt custom destination instance
+        AnyDestination: An initialized destination.
+    """
+    ...
+
+
+def destination(
+    func_or_name: Union[Optional[AnyFun], str] = None,
+    /,
+    loader_file_format: TLoaderFileFormat = None,
```
rudolfix, line 205: let's enforce positional arguments: `*,` - it is right thing to do and skipping it does not help backward compat in any way

docs/website/docs/general-usage/destination.md
```diff
@@ -126,15 +127,11 @@ client_email = "please set me up!"
 You can then use this destination in your pipeline as follows:
 ```py
 import dlt
-from dlt.common.destination import Destination
```
rudolfix, line 129: This whole section is confusing. What it tries to explain: what to do if I want to define many destinations of the same type. The answer is: use named destination. The toml file should have two destinations (both of bigquery type) defined and then two pipelines in the example

tests/destinations/test_custom_destination.py
```diff
@@ -25,6 +25,8 @@
 from tests.load.utils import (
     assert_all_data_types_row,
 )
+from tests.utils import preserve_environ
```
rudolfix: it is already in conftest and will be auto included.

tests/load/duckdb/test_duckdb_client.py
```diff
@@ -307,7 +307,22 @@ def test_credentials_wrong_config() -> None:
 
 
 @pytest.mark.no_load
-def test_duckdb_in_memory_mode_via_factory():
+@pytest.mark.parametrize(
```
rudolfix: this is not the correct way to test it... leave all other test intact or switch to destination factory. you need to write separate tests for new functionality

docs/website/docs/general-usage/destination.md
anuunchin: indeed, i'll fix this

anuunchin: it makes sense to have a separate section that explains what the named destination is at all - I added it with the relevant info as subsections

docs/website/docs/general-usage/snippets/destination-snippets.py
```diff
@@ -103,6 +91,41 @@ def destination_instantiation_snippet() -> None:
...
+    pipeline = dlt.pipeline("pipeline", destination="my_destination")
+    # @@@DLT_SNIPPET_END named_destination_string_reference
+
+    # @@@DLT_SNIPPET_START avoid_example
+    import os
+    import dlt
+
+    os.environ["DESTINATION__FILESYSTEM__DESTINATION_TYPE"] = "duckdb"
+
+    pipeline = dlt.pipeline("pipeline", destination="bigquery")
```
rudolfix, line 123: example tester fails here and I think it is right. I think you want `destination="filesystem"` here? and comparisons below should match

anuunchin: oww, right 🫡

PR #3207, opened 2025-10-15
Feat: workspace file selector, package builder

### LINKED ISSUE #3214: create workspace deployments (tarballs, packages)
**Background**
**Deployment** is a representation (archive, tarball) of a local Workspace that is used by Runtime to run **scripts.**

1. deployment is **owned** by a single **Workspace**
2. deployment contains a tree of eligible Workspace **files**
3. deployment contains a platform agnostic list of all Python **dependencies**
4. deployment has many **versions**. a **version** is identified by a content hash of all the files (we can also have numeric versions and other metadata in Runtime database)
5. deployment has a single **active deployment** which is the most recent version
6. deployment **does not** contain configuration and secrets
7. (optional) deployment must be signed and encrypted in transport and at rest

To create a deployment we need
1. a file selector that will select all the files that are eligible to be in deployment. example: a .gitignore globber
2. a serializable deployment manifest that describes the content of the deployment
3. methods to create and unpack a binary deploy package
4. content hash of the final deployment package

ad 1.
a file selector is an iterator/generator that we can swap in the future. for example, instead of globbing files we could take files from git index of the local repo.

ad 2.
manifest is a part of deployment package and adds metadata on the content of it. it also serves as package format version (`engine_version`)

ad 3.
deployment package is self contained package that contains manifest and all the files. for example:
```text
/
|--- manifest.yaml
|--- files
        | -- content of the workspace produced by file selector
```
by placing manifest at the top we avoid conflicts with paths in `files`

NOTE: we assume that Python dependencies are part of the files that will be added to deployment. Initially we support what `uv` supports by default:
* uv lock
* requirements file
* anything else uv can use ie. https://packaging.python.org/en/latest/specifications/pylock-toml/
anyway: we are pushing dependency resolution to a host in data plane.

a gzipped tarball is OK.

ad 4.
content hash will be computed on final binary blob of deployment package and are also used for versioning in Runtime.
later we'll store packages signed and encrypted (ie. HMAC)

**Requirements**
1. * [ ] implement file selector. we want to follow `.gitignore` and skip `.dlt` folder. `.gitignore` will be there after `dlt init`. GPT-5 says there are nice .gitignore selectors for Python and we are OK to take dependency
2. * [ ] define manifest which may be `TypedDict` for now. We need only
- `engine_version`: manifest package format version (look like what other `engine_version` fields are doing ie. in `Schema`
- list of files in deployment
4. * [ ] implement package builder that will create a gzipped tarball from a given file selector.
- try to avoid copying of the files. add files to tarball directly (if possible) into `files` folder
- add manifest at the end
- gzip
- generate content hash using sha256
5. * [ ] we can implement unpack method in a separate PR

**Implementation**
Create new `deployment` module in `dlt/_workspace`

**Further Work**
We'll need deployment package checks to run client and server-side.
* is right set of dependencies present? is there requirements.txt in a right format? is there `uv.lock` and is it in sync?
* are there any Python scripts in it?

anuunchin:
This PR adds a file selector and a tarballed package builder in `dlt/_workspace` by implementing the following:

- File selector that respects `.gitignore` (or provided ignore file) and excludes the `.dlt/` folder
- Package builder that creates gzipped tarballs with selected files under the `files/` directory and `manifest.yaml` file
- Appropriate tests

Resolves #3214

### DIFF
--- a/dlt/common/utils.py
+++ b/dlt/common/utils.py
@@ -137,6 +137,10 @@ def digest256_tar_stream(stream: BinaryIO, chunk_size: int = 8192) -> str:
     Hashes only filenames and file contents, ignoring timestamps and other metadata.
     This ensures identical file contents produce identical hashes regardless of when
     the tar was created.
+
+    Note: This function operates entirely in-memory using tar.extractfile() which reads
+    from the archive stream. No files are written to disk, preventing leakage of sensitive
+    data that may be contained in the archive.
     """
     stream.seek(0)
     hash_obj = hashlib.sha3_256()
--- a/tests/workspace/deployment/test_package_builder.py
+++ b/tests/workspace/deployment/test_package_builder.py
@@ -13,14 +13,13 @@ from dlt._workspace.deployment.package_builder import (
 from dlt._workspace.deployment.file_selector import WorkspaceFileSelector
 from dlt._workspace.deployment.manifest import DEPLOYMENT_ENGINE_VERSION
 
-from tests.workspace.utils import isolated_workspace, WORKSPACE_CASES_DIR
+from tests.workspace.utils import isolated_workspace
 
 
 def test_write_package_to_stream() -> None:
     """Test building deployment package to a stream and verify structure."""
 
-    run_dir = os.path.join(WORKSPACE_CASES_DIR, "default")
-    with isolated_workspace(run_dir, "test_write_package_to_stream") as ctx:
+    with isolated_workspace("default") as ctx:
         builder = DeploymentPackageBuilder(ctx)
         selector = WorkspaceFileSelector(ctx, ignore_file=".ignorefile")
 
@@ -66,8 +65,7 @@ def test_write_package_to_stream() -> None:
 def test_build_package() -> None:
     """Test that deployment packages are content-addressable with reproducible hashes."""
 
-    run_dir = os.path.join(WORKSPACE_CASES_DIR, "default")
-    with isolated_workspace(run_dir, "test_package_builder") as ctx:
+    with isolated_workspace("default") as ctx:
         builder = DeploymentPackageBuilder(ctx)
         selector = WorkspaceFileSelector(ctx)
 
@@ -75,6 +73,8 @@ def test_build_package() -> None:
         assert str(package_path).startswith(f"{ctx.data_dir}{os.sep}deployment-")
         assert len(content_hash) == 44  # sha3_256 base64 string
 
+        # NOTE: Sleep ensures tarballs have different timestamps in their metadata, proving
+        # digest256_tar_stream produces identical hashes despite different creation times
         time.sleep(0.2)
 
         package_path_2, content_hash_2 = builder.build_package(selector)
--- a/tests/workspace/deployment/test_file_selector.py
+++ b/tests/workspace/deployment/test_file_selector.py
@@ -3,7 +3,7 @@ import pytest
 
 from dlt._workspace.deployment.file_selector import WorkspaceFileSelector
 
-from tests.workspace.utils import isolated_workspace, WORKSPACE_CASES_DIR
+from tests.workspace.utils import isolated_workspace
 
 
 @pytest.mark.parametrize(
@@ -23,8 +23,7 @@ def test_file_selector_respects_gitignore(with_additional_exclude: bool) -> None
     if with_additional_exclude:
         expected_files.remove("additional_exclude/empty_file.py")
 
-    run_dir = os.path.join(WORKSPACE_CASES_DIR, "default")
-    with isolated_workspace(run_dir, "test_file_selector") as ctx:
+    with isolated_workspace("default") as ctx:
         selector = WorkspaceFileSelector(
             ctx, additional_excludes=additional_excludes, ignore_file=".ignorefile"
         )


dlt/_workspace/deployment/file_selector.py
```diff
@@ -0,0 +1,45 @@
+from typing import Iterator, Protocol, Optional, List
+from pathlib import Path
+from pathspec import PathSpec
```
rudolfix, line 3: we'll need to add `pathspec` as dependency

dlt/_workspace/deployment/file_selector.py
```diff
@@ -0,0 +1,45 @@
...
+from pathspec.util import iter_tree_files
+
+from dlt.common.runtime.run_context import DOT_DLT
+
+
+class FileSelector(Protocol):
```
rudolfix: good idea but Python has `Iterarator` already. Just implement it.

dlt/_workspace/deployment/file_selector.py
```diff
@@ -0,0 +1,45 @@
...
+        if gitignore_path.exists():
+            with gitignore_path.open("r", encoding="utf-8") as f:
+                patterns.extend(f.read().splitlines())
+
+        # Add caller-provided excludes
+        patterns.extend(additional_excludes)
+
+        return PathSpec.from_lines("gitwildmatch", patterns)
+
+    def iter_files(self) -> Iterator[Path]:
```
rudolfix: this should be `__iter__` 

btw. you can also implement `Iterable` so you just return iterator from it

dlt/_workspace/deployment/file_selector.py
```diff
@@ -0,0 +1,45 @@
...
+
+class FileSelector(Protocol):
+    """Protocol for iterating over files eligible for deployment"""
+
+    def iter_files(self) -> Iterator[Path]:
+        """Yield relative paths of files eligible for deployment"""
+        ...
+
+
+class GitignoreFileSelector:
```
rudolfix: Name suggest a generic gititgnore selector. but your code is excluding `.dlt` so this is concrete implementation of `WorkspaceFileSelector`. Keep one class or split in two. I'd say one class is good for now

dlt/_workspace/deployment/file_selector.py
```diff
@@ -0,0 +1,45 @@
...
+    """File selector that respects .gitignore and excludes workspace internals"""
+
+    def __init__(
+        self, workspace_path: str, additional_excludes: Optional[List[str]] = None
```
rudolfix: you can just take `WorkspaceRunContext` and use `run_dir` as workspace path and `settings_dir` as `.dlt` path to exclude

dlt/_workspace/deployment/file_selector.py
```diff
@@ -0,0 +1,45 @@
...
+    ) -> None:
+        self.root_path: Path = Path(workspace_path)
+        self.spec: PathSpec = self._build_pathspec(additional_excludes or [])
+
+    def _build_pathspec(self, additional_excludes: List[str]) -> PathSpec:
+        """Build PathSpec from .gitignore + defaults + additional excludes"""
+        patterns: List[str] = [f"{DOT_DLT}/"]
```
rudolfix: do not hardcode path. See comment above. And make sure `settings_dir` is a subpath of `run_dir` before adding to excluded. Note: both can be relative/absolute. Makes sense to normalize them before any use

dlt/common/utils.py
```diff
@@ -119,6 +120,14 @@ def digest256(v: str) -> str:
     return base64.b64encode(digest).decode("ascii")
 
 
+def digest256_file_stream(stream: BinaryIO) -> str:
+    """Returns a base64 encoded sha3_256 hash of a binary stream"""
+    stream.seek(0)
```
rudolfix, line 126: there's a way to do that without materializing data. a hash can be built by adding chunks

dlt/_workspace/deployment/package_builder.py
```diff
@@ -0,0 +1,69 @@
...
+                "files": manifest_files,
+            }
+            manifest_yaml = yaml.dump(
+                manifest, allow_unicode=True, default_flow_style=False, sort_keys=False
+            ).encode("utf-8")
+            manifest_info = tarfile.TarInfo(name=DEFAULT_MANIFEST_FILE_NAME)
+            manifest_info.size = len(manifest_yaml)
+            tar.addfile(manifest_info, BytesIO(manifest_yaml))
+
+        return digest256_file_stream(output_stream)
```
rudolfix: OK! here @sh-rp reported that `tar` adds creation timestamp to the byte stream. so we are going to get a different digest each time. my take: ask @sh-rp what procedure he uses in runtime. my take would be to do a chunked digest and add original files to it. in that case we could add it as "DIGEST" to tarball at the end

also you do not do gzip here right?

dlt/_workspace/deployment/package_builder.py
```diff
@@ -0,0 +1,69 @@
...
+DEFAULT_DEPLOYMENT_FILES_FOLDER = "files"
+DEFAULT_MANIFEST_FILE_NAME = "manifest.yaml"
+DEFAULT_DEPLOYMENT_PACKAGE_LAYOUT = "deployment-{timestamp}.tar.gz"
+
+
+class DeploymentPackageBuilder:
+    """Builds gzipped deployment package from file selectors"""
+
+    def __init__(self, workspace_root: Path):
+        self.workspace_root = Path(workspace_root)
```
rudolfix: good idea. but we have `data_dir` of RunContext to store such transient things. Also see `get_data_entity()`

dlt/_workspace/deployment/file_selector.py
anuunchin: owww, 🫡

dlt/_workspace/deployment/package_builder.py
anuunchin: i do gzip with "w|gz" 👀

pyproject.toml
```diff
@@ -183,6 +183,7 @@ workspace = [
     "pyarrow>=16.0.0",
     "marimo>=0.14.5",
     "mcp>=1.2.1 ; python_version >= '3.10'",
+    "pathspec>=0.11.2",
```
rudolfix, line 186: make sure that min pathspec is at least 6 weeks old. we have CI job that will run tests on min versions

tests/workspace/deployment/test_package_builder.py
```diff
@@ -0,0 +1,83 @@
...
+    run_dir = os.path.join(WORKSPACE_CASES_DIR, "default")
+    with isolated_workspace(run_dir, "test_package_builder") as ctx:
+        builder = DeploymentPackageBuilder(ctx)
+        selector = WorkspaceFileSelector(ctx)
+
+        package_path, content_hash = builder.build_package(selector)
+        assert str(package_path).startswith(f"{ctx.data_dir}{os.sep}deployment-")
+        assert len(content_hash) == 44  # sha3_256 base64 string
+
+        time.sleep(0.2)
```
rudolfix, line 78: explain why it sleeps (timestamp in tar must change, right?)

dlt/common/utils.py
```diff
@@ -119,6 +121,41 @@ def digest256(v: str) -> str:
...
+    hash_obj = hashlib.sha3_256()
+
+    with tarfile.open(fileobj=stream, mode="r:*") as tar:
+        members = sorted(tar.getmembers(), key=lambda m: m.name)
+
+        for member in members:
+            hash_obj.update(member.name.encode())
+            if member.isfile():
+                f = tar.extractfile(member)
```
rudolfix, line 154: note: this is not writing any files, file object is open in-memory. no leakage of sensitive data

pyproject.toml
anuunchin: owww, got it

rudolfix (approved):
LGTM!

PR #3209, opened 2025-10-15
Feat: implement runtime CLI auth

### LINKED ISSUE #3208: Implement runtime CLI auth
### Feature description

Implement `dlt runtime login` and `dlt runtime logout` commands that would store JWT token in global secrets and ensure synchronisation of current workspace with user's default remote workspace

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

_No response_

### Proposed solution

_No response_

### Related issues

_No response_

ivasio:
Closes #3208 

- Added the `runtime` command with 2 subcommands: `login` and `logout`
- Implemented `AuthService` that implements auth-related business logic. This service is currently used in CLI representation-level handler functions that implement CLI interaction with the user. In the future, the service or `authorise()` handler function can be reused in other `dlt runtime` subcommands
- Secrets/configs reading and writing is implemented by constructing providers explicitly or even directly using `tomlkit` to ensure values are read from/written to the specific files
- Runtime API SDK generation is added together with `make` command. Generated SDK code is ignored during linting


dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,118 @@
...
+class RuntimeCommand(SupportsCliCommand):
+    command = "runtime"
+    help_string = "Connect to Runtime and manage your remote Workspaces"
+    description = """"""
+
+    def configure_parser(self, parser: argparse.ArgumentParser) -> None:
+        self.parser = parser
+
+        subparsers = parser.add_subparsers(
+            title="Available subcommands", dest="profile_command", required=False
```
sh-rp: this should be dest="runtime_command". I think this will show up in the help and also the docs when you have regenerated them.

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,118 @@
...
+from typing import Optional
+
+from dlt._workspace._workspace_context import active
+from dlt._workspace.auth import AuthService
+from dlt._workspace.exceptions import (
+    RuntimeNotAuthenticated,
+    WorkspaceIdMismatch,
+    LocalWorkspaceIdNotSet,
+)
+from dlt._workspace.runtime_clients import AUTH_BASE_URL
```
sh-rp: I think these URLs should be part of the runtime config sections in config.toml. The associated class is RuntimeConfiguration. This way you can point the runtime commands to various endpoints for development or staging environments. The defaults can be there localhost urls for now.

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,118 @@
...
+
+def authorise(auth_service: Optional[AuthService] = None) -> None:
+    if auth_service is None:
+        auth_service = AuthService(run_context=active())
+        auth_service.authenticate()
+
+    try:
+        auth_service.authorise()
+    except LocalWorkspaceIdNotSet:
+        fmt.echo("No workspace id found in local config, using default remote workspace")
```
sh-rp: I think in the future we want may want to ask the user here, there may be cases where the user accidentally connects a different workspace. I would also maybe not say anything about "default workspace". From the perspective of the user, there is just one workspace for now.

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,118 @@
...
+from dlt.cli import SupportsCliCommand, echo as fmt
+
+from dlt._workspace.runtime_clients.auth import Client as AuthClient
+from dlt._workspace.runtime_clients.auth.api.github import github_oauth_complete, github_oauth_start
+
+
+class RuntimeCommand(SupportsCliCommand):
+    command = "runtime"
+    help_string = "Connect to Runtime and manage your remote Workspaces"
+    description = """"""
```
sh-rp: I would fill this out, this will also go into the generated docs which you can find at https://dlthub.com/docs/reference/command-line-interface

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,118 @@
...
+            "Do you want to overwrite the local workspace id with the remote one?",
+            choices=["yes", "no"],
+            default="yes",
+        )
+        if should_overwrite == "yes":
+            auth_service.overwrite_local_workspace_id()
+            fmt.echo("Local workspace id overwritten with remote workspace id")
+        else:
+            raise RuntimeError("Unable to synchronise remote and local workspaces")
+    fmt.echo("Authorised to workspace %s" % fmt.bold(auth_service.workspace_id))
```
sh-rp: Small not important nit-pick. We use us english internally and externally, where authorise is spelt authorize. Ultimately we should probably extract all of those cli strings into a strings file at some point and can get AI to proof-read. That is for later though :)

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,118 @@
...
+        auth_service.authorise()
+    except LocalWorkspaceIdNotSet:
+        fmt.echo("No workspace id found in local config, using default remote workspace")
+        auth_service.overwrite_local_workspace_id()
+    except WorkspaceIdMismatch as e:
+        fmt.warning(
+            "Workspace id in local config (%s) is not the same as remote workspace id (%s)"
+            % (e.local_workspace_id, e.remote_workspace_id)
+        )
+        should_overwrite = fmt.prompt(
```
sh-rp: I would use `fmt.confirm` here. We have nice test helpers as well as a non-interactive mode that only works with the confirm.

dlt/_workspace/auth.py
```diff
@@ -0,0 +1,181 @@
...
+from dlt.common.configuration.specs.runtime_configuration import RuntimeConfiguration
+
+
+class AuthInfo(BaseModel):
+    user_id: str
+    email: str
+    jwt_token: str
+
+
+class AuthService:
```
sh-rp: FYI: Generally speaking we use a more functional programming style for all the cli helpers, you can see this very well in the deploy commands. I'm not sure if this ultimately is good or bad. I'd say you can choose to follow that pattern or choose to create these classes for all the runtime interaction stuff. Just keep in mind that these classes are very short lived and do not retain any state between cli commands, so they are not really that useful. If you want to introduce some kind of dependency injection between these services for better testability, that would be a reason to build it this way. Testing pure functions that only depend on the inputs is a quite nice pattern too though.

dlt/_workspace/cli/_runtime_command.py
ivasio: Yeah, for different workspace (another id already set locally) there's another exception with different handling. But makes sense, I'll be asking for confirmation and remove "default" for the time being

dlt/_workspace/cli/_runtime_command.py
ivasio: Thanks, I was looking for something like that!

dlt/_workspace/auth.py
ivasio: Yeah, I've noticed - I've implemented the entrypoints in functional style. Chose a class for this one to specifically encapsulate all the auth internals and auth state. It will be very useful in next commands (`dlt runtime deploy` and `dlt runtime run`) - there will be many calls to auth-protected API, so I think keeping this state even for one command is worth it. I could switch to functional style too, it would be a bit more verbose as to my taste, but I've also noticed some internals such as `Provider` and `RunContext` are implemented as classes, so would rather stick to class here if you don't mind?

ivasio:
Thanks @sh-rp ! I've fixed all problems you pointed out. Also made a little move, `auth.py` -> `runtime.py`, because the client helpers that should be now there if I consider your comments, don't fit anywhere else. And `AuthService` is a `ProfileAuthService` if I'm specific, so sorry for the move after the first review, but I think it makes more sense now 

Regarding your PR-level comments, 
- this code is under the `.workspace` feature flag, so it doesn't get rendered to Markdown docs by `update-cli-docs`. Not exactly sure why tests are failing, I'll rerun, but seems like general flakiness, cause I haven't added any new ones yet
- there's only one make command and one helper script, I thought it was more useful to have code generation commands in the same repo where the generated code will live, as we agreed on the meeting. Easier to have the openapi URLs "external" to this codebase, rather then the output directory, or do you think otherwise?

Otherwise, waiting for @rudolfix 's CLI test helpers to jump on the tests and finalize this PR

Makefile
```diff
@@ -185,3 +185,14 @@ start-dlt-dashboard-e2e:
 # creates the dashboard test pipelines globally for manual testing of the dashboard app and cli
 create-test-pipelines:
 	uv run python tests/workspace/helpers/dashboard/example_pipelines.py
+
+## Generates python clients for runtime api and auth services
+# Before running, adjust the URL or run `make up` in runtime first
+generate-python-clients:
```
rudolfix, line 186: please generate client from openapi specs that we keep in this repo and can always use to recreate old clients.
* we have tools module where we can keep open api specs for different services ie `runtime_client/auth.spec.yaml` or smth
* I think it makes sense to keep list of major versions there. obviously we have just one now
* does not make sense to distribute specs with OSS code (as it is now)

dlt/_workspace/configuration.py
```diff
@@ -22,8 +23,14 @@ class WorkspaceSettings(BaseConfiguration):
 class WorkspaceRuntimeConfiguration(RuntimeConfiguration):
     """Extends runtime configuration with dlthub runtime"""
 
-    # TODO: connect workspace to runtime here
-    # TODO: optionally define scripts and other runtime settings
+    __section__: ClassVar[str] = "workspace_runtime"
```
rudolfix: mhmmm maybe let's keep base class section. I think you can remove this line...

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+    SecretsTomlProvider,
+)
+from dlt.common.configuration.resolve import resolve_configuration
+from dlt.common.configuration.specs.pluggable_run_context import RunContextBase
+from dlt.common.runtime.run_context import active
+
+
+@dataclass
+class AuthInfo:
+    user_id: str
```
rudolfix, line 31: Is it random user id that our runtime generates? or user id from SSO ie. github? should be the first one

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+    email: str
```
rudolfix, line 32: is there any reason to send PII data to the client?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+    jwt_token: str
```
rudolfix: tbh. jwt should contain all auth data you need: ie `sub` should be user id. what data are you getting in jwt?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class RuntimeAuthService:
+    """
+    Implements login, logout and auth check internals
```
rudolfix, line 38: IMO authentication should be split from connecting the local workspace to remote one. We should have two public methods here

- authenticate: obtains valid jwt
- connect: connects local workspace to remote one which ends with a valid workspace_id

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+
+    _run_context: RunContextBase
+    _local_workspace_id: Optional[str] = None
+    _remote_workspace_id: Optional[str] = None
+
+    def __init__(self, run_context: RunContextBase):
+        self._run_context = run_context
+
+    @property
+    def workspace_run_context(self) -> WorkspaceRunContext:
```
rudolfix, line 55: most of the methods below shoudl be private. or not?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+            "workspace_id",
+            str(self._remote_workspace_id),
+            None,
+            WorkspaceRuntimeConfiguration.__section__,
+        )
+        local_toml_config.write_toml()
+        self._local_workspace_id = self._remote_workspace_id
+
+    def _read_token(self) -> AuthInfo:
+        config = resolve_configuration(WorkspaceRuntimeConfiguration())
```
rudolfix: resolved runtime configuration is already in `workspace_run_context`

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+        self._local_workspace_id = config.workspace_id
+
+        if not self._local_workspace_id:
+            raise LocalWorkspaceIdNotSet(self._remote_workspace_id)
+        elif self._local_workspace_id != self._remote_workspace_id:
+            raise WorkspaceIdMismatch(self._local_workspace_id, self._remote_workspace_id)
+
+        return self.workspace_id
+
+    def overwrite_local_workspace_id(self) -> None:
```
rudolfix, line 103: this is private method I assume?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+        local_toml_config = ConfigTomlProvider(self.workspace_run_context.settings_dir)
+        local_toml_config.set_value(
+            "workspace_id",
+            str(self._remote_workspace_id),
+            None,
+            WorkspaceRuntimeConfiguration.__section__,
```
rudolfix: I'd go for known_sections.WORKSPACE, RuntimeConfiguration.__section__ to have it nicely scoped

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+    def _delete_token(self) -> None:
+        # delete from global secrets directly, because in other cases config deletion is not supported
+        secrets_path = os.path.join(self.run_context.global_dir, SECRETS_TOML)
+        if not os.path.isfile(secrets_path):
+            return
+        toml = TOMLFile(secrets_path)
+        doc = toml.read()
+
+        # Safely check for structure and remove the key
+        runtime_section = doc.get(WorkspaceRuntimeConfiguration.__section__)
```
rudolfix: pls add a method to the provider itself along `set_value` and write proper tests for it

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def login() -> None:
+    auth_service = RuntimeAuthService(run_context=active())
+    try:
+        auth_info = auth_service.authenticate()
+        fmt.echo("Already logged in as %s" % fmt.bold(auth_info.email))
+        authorize(auth_service=auth_service)
+    except RuntimeNotAuthenticated:
+        fmt.echo("Logging in with Github OAuth")
```
rudolfix, line 59: makes sense to extract for testing. is it testable at all?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+                jwt_token=token.decode("utf-8"), email=payload["email"], user_id=payload["user_id"]
+            )
+        except (KeyError, TypeError) as e:
+            raise RuntimeNotAuthenticated("Failed to validate JWT payload") from e
+
+        return auth_info
+
+
+def get_auth_client() -> AuthClient:
+    config = resolve_configuration(WorkspaceRuntimeConfiguration())
```
rudolfix: you'll need workspace run context to get configuration

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+        if isinstance(token, str):
+            token = token.encode("utf-8")
+        try:
+            payload = jwt.decode(token, options={"verify_signature": False})
+        except jwt.PyJWTError as e:
+            raise RuntimeNotAuthenticated("Failed to decode JWT") from e
+
+        try:
+            auth_info = AuthInfo(
+                jwt_token=token.decode("utf-8"), email=payload["email"], user_id=payload["user_id"]
```
rudolfix, line 153: why `sub` is not used for `user_id`? what is the identity the jwt is issued against?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+
+    def login(self, token: str) -> AuthInfo:
+        self._save_token(token)
+        return self.auth_info
+
+    def logout(self) -> None:
+        self._delete_token()
+        self._remote_workspace_id = None
+
+    def authorize(self) -> str:
```
rudolfix: from my POV this is workspace `connect` operation, We do not do any workspace authorization here. It is probably OK for now. authorization should be a part of connect process in the future here ie. `view` or `connect` permission will be requested on worskpace with given id. now you do not even send existing workspace id if it is there

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
+import os
+from typing import Optional
+
+from dataclasses import dataclass
+import jwt
```
rudolfix: we do not use `jwt` - but other library (jose-jwt) please take a look at `dlthub` repo. reason: not to force users to use 100mb size of `cryptography` lib.

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,191 @@
...
+from git import Union
```
rudolfix: this should come from typing

dlt/_workspace/runtime.py
sh-rp: I think sub is also set with the same id, not sure, but it would be easy to change.

ivasio:
Thanks for the detailed review @rudolfix ! To answer the general comment, do you mean single file module, or actually a package? Happy to introduce a package, but wouldn’t put everything into one module, please see the rationale below
- I tried to follow the current approach overall used in CLI and pointed out by Dave, where most logic lives in handler functions. It is used in the further work, please see the WIP branch
- auth functionality is a bit separate though because it’s going to be used in a lot of other methods. so I separated only auth business logic into a separate class (business logic layer) to be able to use it directly in CLI handler functions (adapter layer), so ports/adapters-like approach. business logic layer methods that are called from adapter layer are public, all the other ones are protected
- because auth process is interactive, I e.g. separated authorize into separate methods, and yes, implementation detail of overwriting the workspace id is exposed to adapter layer
- This separation seems to make sense to me, please let’s discuss if you’d rather me inline this class into the handlers. It’s easy to do now as long as we reuse the interactive login as in the WIP branch, but if we are to invoke auth business logic separately going forward, the inlining would prohibit that

Makefile
ivasio: as far as I remember, we’ve discussed offline that we store specs and generated code in OSS _workspace for the time being to avoid generating a separate package, which is what I’m doing here. happy to change the path of saving the spec, or refactor otherwise - let’s discuss?

dlt/_workspace/configuration.py
ivasio: sure! just for me to understand better, what’s your heuristic, how many config fields deserve their own section, or how do you decide it otherwise?
and btw @rudolfix  after switching I experienced that `PluggableRunContext.reload_providers` didn’t fully do the job. I’ve patched the implementation a bit in a separate commit https://github.com/dlt-hub/dlt/pull/3209/commits/9a12719a1ac4fe8aaf6c11cd24537c44f34efa5f , please take a look if it’s the right direction - it’s the only thing that helped

dlt/_workspace/runtime.py
ivasio: it is indeed our user id generated by the backend

dlt/_workspace/runtime.py
ivasio: it’s coming from the server in jwt payload, as long as we store the token on the fs, I thought it was ok to use it in memory. I’m using it here to show the user the email they logged in with (printed to stdout with asterisks). happy to get rid of it here, or remove from jwt payload on the server side too, let me know what to do better

dlt/_workspace/runtime.py
ivasio: already split, renamed authorize to connect as you requested in another comment. for the rationale why more public methods are needed from my POV, please see main comment - happy to discuss

dlt/_workspace/runtime.py
ivasio: please see main comment

dlt/_workspace/runtime.py
ivasio: great, thanks! fixed

dlt/_workspace/runtime.py
ivasio: no, please see general comment, it’s invoked from adapter/cli layer

dlt/_workspace/runtime.py
ivasio: oh thanks for pointing that out, forgot to change to set “” as we discussed! done

dlt/_workspace/cli/_runtime_command.py
ivasio: in the current “integration” tests, stdout of the process is tested, not a bad approach as it seems to me, although a lot of mocking, I must admit. I can extract unit tests for the Auth class, would you rather me do it instead of integration tests?

dlt/_workspace/runtime.py
ivasio: I’d use user_id in sub and get rid of email, happy to refactor it on both client and server side

dlt/_workspace/runtime.py
ivasio: sure thing, changed!

dlt/_workspace/runtime.py
ivasio: thanks, great catch!

ivasio:
@rudolfix I was able to seriously reduce the number of generated models by post-processing the OpenAPI spec we're getting from the server (litestar spec generation is flawed in there, see https://github.com/litestar-org/litestar/issues/3021). We're now reusing the same models for exceptions everywhere, agreed with @sh-rp that it's going to be the case from the server side too. Awaiting your review(s)

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,127 @@
...
+                body=github_oauth_complete.GithubDeviceFlowLoginRequest(
+                    device_code=login_request.device_code
+                ),
+            )
+            if isinstance(token_response, github_oauth_complete.LoginResponse):
+                auth_info = auth_service.login(token_response.jwt)
+                fmt.echo("Logged in as %s" % fmt.bold(auth_info.email))
+                connect(auth_service=auth_service)
+                break
+            elif isinstance(token_response, github_oauth_complete.ErrorResponse400):
```
rudolfix, line 85: how other 40x errors are handled? is there a response that says we should wait? we should loop only on that response. for anything else we should raise

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,127 @@
...
+                raise RuntimeError("Failed to complete authentication with Github")
```
rudolfix, line 86: mhmmm is this some kind of pattern? why can't we just reraise HttpError or something similar?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
...
+    def connect(self) -> str:
+        # Currently, ensuring workspace id is the same as default workspace id of the user
+        if not self._remote_workspace_id:
+            client = get_api_client(self)
+            me_response = me.sync(client=client)
+
+            if isinstance(me_response, MeResponse):
+                self._remote_workspace_id = str(me_response.default_workspace.id)
+            else:
+                raise RuntimeError("Failed to get me response")
```
rudolfix, line 92: is this possible? maybe the client could raise on non 20x responses so you do not need to check it? maybe we should have something like `raise_on_error()` from `requests`? what you do is hiding information what went wrong

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
...
+
+        return self.workspace_id
+
+    def overwrite_local_workspace_id(self) -> None:
+        local_toml_config = ConfigTomlProvider(self.workspace_run_context.settings_dir)
+        local_toml_config.set_value(
+            "workspace_id",
+            str(self._remote_workspace_id),
+            None,
+            RuntimeConfiguration.__section__,
```
rudolfix, line 109: known_sections,WORKSPACE, RuntimeConfiguration.__section__ is a better (we scope this setting to workspace runtime)

dlt/_workspace/runtime.py
rudolfix: if this is some kind of assert just use assert... no one is running Python with asserts removed nowadays

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
...
+    def _read_token(self) -> AuthInfo:
+        config = self.workspace_run_context.runtime_config
+        if not config.auth_token:
+            raise RuntimeNotAuthenticated("No token found")
+        self.auth_info = self._validate_and_decode_jwt(config.auth_token)
+        return self.auth_info
+
+    def _save_token(self, token: str) -> AuthInfo:
+        self.auth_info = self._validate_and_decode_jwt(token)
+        value = [WritableConfigValue("auth_token", str, token, (RuntimeConfiguration.__section__,))]
```
rudolfix, line 123: same section here

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
...
+        os.makedirs(global_path, exist_ok=True)
+        secrets = SecretsTomlProvider(settings_dir=global_path)
+        write_values(secrets._config_toml, value, overwrite_existing=True)
+        secrets.write_toml()
+        return self.auth_info
+
+    def _delete_token(self) -> None:
+        # delete from global secrets directly, because in other cases config deletion is not supported
+        local_toml_config = SecretsTomlProvider(self.workspace_run_context.global_dir)
+        local_toml_config.set_value(
```
rudolfix, line 135: we should definitely implement "delete" operation but not now

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
...
+            "auth_token",
+            "",
+            None,
+            RuntimeConfiguration.__section__,
```
rudolfix, line 139: also sections here

dlt/_workspace/runtime.py
rudolfix: please remove user_id from jwt. you can keep email. we should stick to pre-defined scopes in jwt if possible

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
...
+        try:
+            payload = jose_jwt.decode(token, key="", options={"verify_signature": False})
+        except JOSEError as e:
+            raise RuntimeNotAuthenticated("Failed to decode JWT") from e
+
+        try:
+            auth_info = AuthInfo(
+                jwt_token=token.decode("utf-8"), email=payload["email"], user_id=payload["user_id"]
+            )
+        except (KeyError, TypeError) as e:
```
rudolfix, line 155: huh! I think when we start to really validating token, there's an exception for that. `ValueError` maybe?

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
...
+        local_toml_config.set_value(
+            "auth_token",
+            "",
+            None,
+            RuntimeConfiguration.__section__,
+        )
+        local_toml_config.write_toml()
+
+    def _validate_and_decode_jwt(self, token: Union[str, bytes]) -> AuthInfo:
+        if isinstance(token, str):
```
rudolfix, line 144: when is it bytes? you always pass string here

dlt/common/configuration/specs/pluggable_run_context.py
```diff
@@ -206,7 +210,10 @@ def reload(
 
     def reload_providers(self) -> None:
         self.providers = ConfigProvidersContainer(self.context.initial_providers())
-        self.providers.add_extras()
+        # Invalidate any cached configuration on the context so it re-resolves using new providers
+        self.context.reset_config()
```
rudolfix, line 214: I looked into the code and instead of adding new `reset_config` method you can just call
`self.initialize_runtime()` which will reset the config and also apply new settings to runtime (as define by contecxt provider)

pyproject.toml
```diff
@@ -186,6 +186,9 @@ workspace = [
     "marimo>=0.14.5",
     "mcp>=1.2.1 ; python_version >= '3.10'",
     "pathspec>=0.11.2",
+    "httpx>=0.27.0",
+    "python-jose>=3.5.0",
+    "types-python-jose>=3.5.0.20250531",
```
rudolfix, line 191: this is `dev` group dependency, please be careful when adding those to extras and main deps!

tools/clean_openapi_spec.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def construct_error_schema_name(status_code: str) -> str:
+    return f"ErrorResponse{status_code}"
+
+def construct_error_schema_ref(status_code: str) -> dict:
+    return {"$ref": f"#/components/schemas/{construct_error_schema_name(status_code)}"}
+
+
+def extract_postprocess_error_response_schemas(spec: dict) -> dict:
```
rudolfix, line 14: thanks for fixing this!

dlt/_workspace/cli/_runtime_command.py
```diff
@@ -0,0 +1,127 @@
...
+    LocalWorkspaceIdNotSet,
+    RuntimeNotAuthenticated,
+    WorkspaceIdMismatch,
+)
+from dlt._workspace.runtime import RuntimeAuthService, get_auth_client
+from dlt._workspace.runtime_clients.auth.api.github import github_oauth_complete, github_oauth_start
+from dlt._workspace.cli import echo as fmt
+
+
+class RuntimeCommand(SupportsCliCommand):
```
rudolfix, line 17: please follow other commands implementation. what you do here: you expose this module via dlt plugin. this will force import of this module when we are registering plugins which will also import those files from the `runtime_clients`. we do not want that

please look dlt/_workspace/cli/_plugins.py top level comments and follow everything there

dlt/_workspace/runtime.py
```diff
@@ -0,0 +1,174 @@
+import os
+from typing import Optional, Union
+from dataclasses import dataclass
+
+from jose.exceptions import JOSEError
```
rudolfix, line 5: because this is extra dependency it will be safer to import it only in the function that needs it. and there's just one. I'm still undecided how we'll deal with workspace extra. for sure we do not need it upfront

ivasio:
merging CLI into `runtime`

PR #3211, opened 2025-10-15
Fix: add support for yield_map in rest resource

### LINKED ISSUE #3194: Add support for `add_yield_map` in `ProcessingSteps`
### Feature description

`ProcessingSteps` currently only supports `filter` and `map`. `yield_map` which would correspond to an `add_yield_map` call logically would be a good fit here. 

### Are you a dlt user?

Yes, I use it for fun.

### Use case

I would just like to have parity between the declarative and imperative paradigms of the library.

### Proposed solution

`ProcessingSteps` would now accept a `yield_map` field. 

### Related issues

_No response_

ivasio:
Resolves #3194 

Added support for yield_map in rest resource

### DIFF
--- a/tests/sources/rest_api/integration/test_processing_steps.py
+++ b/tests/sources/rest_api/integration/test_processing_steps.py
@@ -5,7 +5,12 @@ import pytest
 
 import dlt
 from dlt.sources.rest_api import RESTAPIConfig, rest_api_source
-from ..conftest import DEFAULT_COMMENTS_COUNT, DEFAULT_PAGE_SIZE, DEFAULT_TOTAL_PAGES, DEFAULT_REACTIONS_COUNT
+from ..conftest import (
+    DEFAULT_COMMENTS_COUNT,
+    DEFAULT_PAGE_SIZE,
+    DEFAULT_TOTAL_PAGES,
+    DEFAULT_REACTIONS_COUNT,
+)
 
 
 def _make_pipeline(destination_name: str):


ivasio:
these changes led to some unforseen tests failing, on it

tests/sources/rest_api/integration/test_processing_steps.py
```diff
@@ -259,3 +260,141 @@ def extend_body(row):
...
+                    {"yield_map": flatten_reactions},
+                ],
+            },
+        ],
+    }
+    mock_source = rest_api_source(config)
+
+    data = list(mock_source.with_resources("posts"))
+
+    assert all("reaction" in record and "reactions" not in record for record in data)
```
rudolfix, line 298: this may be nitpicking but it makes sense to assert expected records count in all tests. here I think it is:
DEFAULT_PAGE_SIZE * DEFAULT_TOTAL_PAGES*DEFAULT_REACTIONS_COUNT right?

same for other tests. such assert is makes sure that records were correctly flattened and all expected data got extracted

rudolfix (changes requested):
code is excellent! I'd add a small improvement to tests

we still need docs update. https://dlthub.com/docs/dlt-ecosystem/verified-sources/rest_api/basic#processing-steps-filter-and-transform-data looks like a good place

rudolfix (approved):
LGTM! thanks for docs and tests!

PR #3225, opened 2025-10-21
docs: replace `.table(..., table_type="ibis")` with `.to_ibis()`

zilto:
follows #3179 

Documentation and tests were updated to remove `.table(..., table_type="ibis")` to `.table(...).to_ibis()`. 

The changes to the `dlt.Dataset` are breaking, but there's a clear and useful exception being raised.

Alternative: catch `**kwargs` and handling things behind the scene. We would raise a deprecation warning instead.

### DIFF
--- a/dlt/common/libs/ibis.py
+++ b/dlt/common/libs/ibis.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, Any, Generator, Mapping, Optional, Union
 
 import dlt
 from dlt.common.destination.dataset import SupportsDataAccess
-from dlt.helpers.ibis import DATA_TYPE_MAP, create_ibis_backend
+from dlt.helpers.ibis import DATA_TYPE_MAP, create_ibis_backend, _get_ibis_to_sqlglot_compiler
 from dlt.common.schema import Schema as DltSchema
 from dlt.common.destination import TDestinationReferenceArg
 from dlt.common.schema.typing import TDataType, TTableSchema
@@ -85,7 +85,6 @@ class _DltBackend(SQLBackend, NoUrl, NoExampleLoader):
     name: str = "dlt"
     supports_temporary_tables = False
     supports_python_udfs = False
-    compiler = sc.duckdb.compiler
 
     @classmethod
     def from_dataset(cls, dataset: dlt.Dataset) -> _DltBackend:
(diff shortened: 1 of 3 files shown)


rudolfix (changes requested):
I'm quite sure this is not going to work. Current implementation of `_DltBackend` skips the whole machinery that makes `to_sql` on `Relation` work for different dialect. Failures for snowflake and clickhouse are real and symptoms of some kind of subtle problem.

`create_unbound_ibis_table` allowed you to write destination-agnostic ibis expressions but then we used this expression to generate destination-dependent sqlglot and test sql query. `DltBackend` should do IMO the same.

I need a little bit of time to do a real review and do a few experiments.

zilto:
> Failures for snowflake and clickhouse are real and symptoms of some kind of subtle problem.
Yup, I need to investigate

> Current implementation of `_DltBackend` skips the whole machinery that makes `to_sql` on `Relation` work for different dialect. 
Maybe I missed a normalization step, but what `_DltBackend` is intended to do is just execute the raw SQL produced by `dlt.Relation` using the `dlt.Dataset` SQL client.

Currently, `.to_ibis()` does this
```python
class Relation:
    def to_ibis(self) -> ir.Table:
        from dlt.common.libs.ibis import _DltBackend

        backend = _DltBackend.from_dataset(self._dataset)
        if self._table_name:
            ibis_table = backend.table(self._table_name)
        else:
            ibis_table = backend.sql(self.to_sql())

        return ibis_table
        
# ...
class _DltBackend:
    def raw_sql(self, query: Union[str, sg.Expression], **kwargs: Any) -> Any:
        with contextlib.suppress(AttributeError):
             query = _transpile(query, target_dialect=self.compiler.dialect)

        assert isinstance(query, str)

        with self._dataset.sql_client as client:
             result = client.execute_sql(query)

        return result
```
Maybe I need to use `backend.raw_sql()` instead of `.sql()`


I'll write a test that checks that the content of `ibis.Expr` for `.to_ibis()` and `.table(..., table_type="ibis")` are equal

rudolfix:
@zilto I changed the backend to always go through our relation when compiling and executing queries. the basic idea was already there: you use relation to compute ibis schemas out of tables and queries. I added code to execute queries that they also generate
overall our cursor is dbapi compatible so it worked right away. also we modeled pandas/arrow method on duckdb/ibis so I was able to fit them in

the other change was to pass raw query to get an ibis expression, not the normalized one. that assures that all identifiers come from dlt schema and are not casefolded or flattened (clickhouse case)

PR #3229, opened 2025-10-21
avoids passing naming conventions as modules

### LINKED ISSUE #1622: Supplying Naming Convention as Module Breaks Parallelization
### dlt version

0.5.1

### Describe the problem

On another note, re: [custom naming conventions](https://dlthub.com/docs/examples/custom_naming). I just wanted to flag that normalization parallelization fails when supplying a custom naming convention module directly to a destination. This is because modules can't be pickled for multiprocessing. Using the fully qualified path to the naming convention's module fixes this. I'm not sure if y'all consider this a bug or not, but could be helpful to document as it can be a bit tricky debug

### Expected behavior

_No response_

### Steps to reproduce

Supply a naming convention module to a destination and run enl steps using multiprocess

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

### LINKED ISSUE #2998: Change normalization for Swedish characters
### Feature description

Normalize the Swedish chars `åäö` to the common used `aao`, instead of `___`

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

_No response_

### Proposed solution

_No response_

### Related issues

_No response_

rudolfix:
### Description
This PR adds $PWD to `sys.path` when running cli commands to reproduce Python behavior when it runs scripts. The example problem: custom naming modules were not importable and any pipeline inspection command was failing.

This also fixes problems reported:
fixes #1622 
fixes #2998 

where custom naming modules were directly used. Now the documentation and example uses module names which are pickable.


docs/website/docs/general-usage/naming-convention.md
```diff
@@ -141,6 +141,24 @@ naming="tests.common.cases.normalizers.sql_upper"
 ```
 `dlt` will import `tests.common.cases.normalizers.sql_upper` and use the `NamingConvention` class found in it as the naming convention.
 
+:::tip
+Do not pass custom naming convention as modules if you do it explicitly. We recommend pattern below:
+```py
+import dlt
+
+# you should be able to import sql_cs_latin2 here!
```
VioletM: Maybe remove this line? It's confusing, you say do not import it as modules, and then you say `you should be able to import sql_cs_latin2 here!`

docs/website/docs/general-usage/naming-convention.md
```diff
@@ -141,6 +141,24 @@ naming="tests.common.cases.normalizers.sql_upper"
 ```
 `dlt` will import `tests.common.cases.normalizers.sql_upper` and use the `NamingConvention` class found in it as the naming convention.
 
+:::tip
+Do not pass custom naming convention as modules if you do it explicitly. We recommend pattern below:
```
VioletM: So I'm creating my own module and then pass it as a string in the destination config, and dlt will attempt to import it? If I understood it correctly, I think it requires a bit more explanation what's happening. Right now users might think that `sql_cs_latin2` is one of the predefined available naming conventions. Especially if they not read it carefully. 

Also, should it be mentioned in `## Write your own naming convention` section? I would expect this information to be there

VioletM (changes requested):

PR #3233, opened 2025-10-21
ignores native config values if config spec does not implement those

### LINKED ISSUE #3216: Conflicting `RUNTIME` env var on databricks
### dlt version

1.17.1

### Describe the problem

I consistently got the following error while trying to run a very simple dlthhub pipeline on databricks in serverless mode. I tried running both

`InvalidNativeValue:RuntimeConfiguration cannot parse the configuration value provided. The value is of type str and comes from the sections () Value may be a secret and is not shown. Details:`

### Expected behavior

Data should be available in the duckdb in memory database

### Steps to reproduce

1. Open a databricks notebook in serverless mode 
2. Paste shit reduced example:

CELL 1
```
!pip install dlt[duckdb] duckdb
%restart_python
```
CELL 2 (The order of the imports is important)
```python
import sys
import types
sys.meta_path = [h for h in sys.meta_path if "PostImportHook" not in repr(h)]

for name, module in list(sys.modules.items()):
    if not isinstance(module, types.ModuleType):
        continue
    if getattr(module, "__file__", "") is not None and getattr(module, "__file__", "").startswith(
        "/databricks/spark/python/dlt"
    ):
        del sys.modules[name]

import dlt
import requests 
import duckdb
import dlt.destinations

db = duckdb.connect(":memory:")

url = "https://api.github.com/repos/dlt-hub/dlt/issues"
# Make a request and check if it was successful
response = requests.get(url)
response.raise_for_status()

pipeline = dlt.pipeline(
    pipeline_name="github_issues",
    destination=dlt.destinations.duckdb(":memory:"),
    dataset_name="github_data",
)
# The response contains a list of issues
load_info = pipeline.run(response.json(), table_name="issues")
```

### Operating system

Linux

### Runtime environment

Kubernetes

### Python version

3.12

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

delete the RUNTIME env var in databricks in the pipeline/notebook 
I haven't seen any side effects for the moment but there could be so be careful

rudolfix:
### Description
`dlt` allows to pass "native value" to the configuration. For example `ConnectionStringConfiguration` will accept sql alchemy like connection string as a single value from which it will parse hostname, user, password etc.

most of config spec does not implement this functionality. this PR makes those config spec to silently ignore such values if they appear. use case "RUNTIME" env variable on databricks triggered `RuntimeConfiguration` to parse it as native value.

fixes #3216 

also adds instructions to set up dlt on serverless data bricks

### DIFF
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -63,11 +63,11 @@ jobs:
           - os: windows-latest
             python-version: "3.11"
             shell: cmd
-            pytest_args: '-m "not forked"'
+            pytest_args: '-m "not forked and not rfam"'
           - os: windows-latest
             python-version: "3.13"
             shell: cmd
-            pytest_args: '-m "not forked"'
+            pytest_args: '-m "not forked and not rfam"'
 
     defaults:
       run:
--- a/docs/tools/prepare_examples_tests.py
+++ b/docs/tools/prepare_examples_tests.py
@@ -11,8 +11,10 @@ EXAMPLES_DIR = "../examples"
 
 # settings
 SKIP_FOLDERS = ["archive", ".", "_", "local_cache"]
-SKIP_EXAMPLES: List[str] = []
+# @pytest.mark.rfam
+SKIP_EXAMPLES: List[str] = ["backfill_in_chunks", "connector_x_arrow"]
 SKIP_FORK_EXAMPLES: List[str] = ["custom_destination_lancedb"]
+
 
 # the entry point for the script
 MAIN_CLAUSE = 'if __name__ == "__main__":'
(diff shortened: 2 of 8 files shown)


docs/website/docs/dlt-ecosystem/destinations/databricks.md
```diff
@@ -728,7 +728,36 @@ databricks_adapter(
 ## Troubleshooting
 Use the following steps to avoid conflicts with Databricks' built-in Delta Live Tables (DLT) module and enable dltHub integration.
 
-### 1. Add an `init` script
+### Enable dlt on serverless (16.x)
+Live Tables (DLT) are not available on serverless but the import machinery that is patching DLT is still there in form of import hooks. You
```
VioletM, line 732:
```suggestion
Delta Live Tables (DLT) are not available on serverless, but the import machinery that is patching DLT is still there in form of import hooks. You
```

VioletM (dismissed):
small edits

Issue #3023, opened 2025-08-22
Duplicate records for a source with primary key and merge write disposition (duplicate _dlt_id as well)


jukiewiczm:
### dlt version

1.11.0

### Describe the problem

Related Slack discussion:
https://dlthub-community.slack.com/archives/C04DQA7JJN6/p1755775557702299?thread_ts=1755443136.138759&cid=C04DQA7JJN6

After certain pipeline runs, I'm seeing duplicate rows for a resource that has a `merge` write disposition and contains a primary key.
For this to happen, the loading job needs to get unexpectedly terminated, i.e. a service OOM or an external cancellation signal.
As mentioned in the above Slack discussion, I managed to narrow down the issue to two potential scenarios:
* staging dataset not getting truncated before load
* the same file (or the duplicate of it) appearing twice in a load package (this I consider a more likely scenario, as I believe it also happened once or twice when my write disposition was a "replace" one)

I therefore have the following questions, besides yourselves obviously figuring out whether there's a problem in the code that can be fixed:
* are there any known circumstances that could cause the staging dataset to not get truncated before load?
* are the jobs files "copied" rather than moved, or do you have any other idea on what might cause the duplicates to appear?
* Do you have any ideas on how this can be worked around? I'm currently trying to work around it with dropping pending packages (including partially loaded ones), though this is far from ideal, as I believe this can lead to data being not only incomplete, but also forever corrupt (in a case where a row gets split into multiple files due to unnesting and only some of these files getting loaded).
    * is it perhaps easily possible to make the loading process transactional, at least with regard to the tables/rows that are related to each other due to unnesting? That would at the very least allow me to not have corrupt data if the `drop_pending_packages` turns out to help.
* Any chances there were any updates related to this between dlt 1.11.0 and 1.15.0? I could try upgrading

### Expected behavior

No duplicates in the destination, even if loading fails during the process.

### Steps to reproduce

Unfortunately, this is not easy to reproduce, as it's related to an unexpected failures during loading phase. Since I don't know at which point the failure has to appear, I'm unable to provide a reliable example that would reproduce this behavior.

### Operating system

Linux

### Runtime environment

Other

### Python version

3.12

### dlt destination

MotherDuck

jukiewiczm:
FYI: Dropping the pending packages including partially loaded packages did indeed help to mitigate this (dropping the pending ones was the key), but as mentioned in the description, that's still far from ideal.

rudolfix:
[...]

PR #3234, opened 2025-10-21
graceful signal handler

rudolfix:
### Description
Changes how signal handler in `dlt` works.
1. handles SIGTERM and SIGINT and delays executing signal handlers on first signal of given kind
2. this gives opportunity to drain job pools (ie. in load step) gracefully
3. another signal of given kind (ie CTRL-C twice) executes signal handler immediately (outcome is process termination)
4. change: `signals.sleep` wakes up on signal but does not raise (it was actually preventing proper drain on load pools)
5. QoL: `KeyboardInterrup` is not handled by Pipeline so traces are saved for further investigation
6. docs: enabling handlers for setups where pipelines run in a separate threads

NOTE: in main thread, signals are delayed only when necessary (possibly non-atomic operation take place ie. load phase)

may fix (partially) #3023


rudolfix:
@sh-rp good point about SIGTERM. we should somehow document this behavior vs. timeouts. I'll try to update the docs still in this release.
I think users should let `dlt` drain the pool for 2 minutes and then IMO SIGKILL is OK, we can also bring back delayed raise via `sleep` or some other check that will exit the process the normal way which also involves saving/sending the trace.

PR #3244, opened 2025-10-23
Fixes docs on schema file naming convention

willi-mueller:
### Description
Neither `{source name}_schema.yml` nor `{source name}.schema.yml` worked in my experiments. Only `{source name}.schema.yaml` worked.

### DIFF
--- a/docs/website/docs/general-usage/schema.md
+++ b/docs/website/docs/general-usage/schema.md
@@ -532,7 +532,7 @@ all the settings to default.
 
 ### Automatically load schema file stored with source python module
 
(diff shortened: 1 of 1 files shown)


AstrakhantsevaAA:
@rudolfix is it expected behaviour?

rudolfix:
@AstrakhantsevaAA yes this is correct. we look for `yaml` files

willi-mueller:
> @AstrakhantsevaAA yes this is correct. we look for `yaml` files

Thank you for reviewing it and merging it quickly.

PR #3247, opened 2025-10-24
Chore: Update docs npm dependencies and clean up docs build tooling

sh-rp:
### Description

This PR is a maintenance PR on the docs and the tooling there. What has happened:
* Update npm dependencies for building the docs website.
* Remove some files that were needed for the Netlify deployment, but now no longer are.
* Move all Python dependencies that are needed for docs to the new subproject in the docs folder. Removes now unneeded requirements files.
* Moved all Python scripts needed for docs work (linting, testing, and building docs) into a new Python project in the docs folder, added script entry points for all.
* Create a separate makefile just for docs work.
* Consolidated docs workflows for testing and linting snippets as well as examples into a single one.
* Made the build process on Cloudflare exactly the same as the local process (removes a bunch of duplicated npm commands; some are kept for a bit for version deployment backwards compatibility).
* Fixes a bunch of linter errors in snippets, examples, and notebooks.
* Updates main docs readme file
* Reformats all snippets, examples and education notebooks to be a bit narrower so they fit on small resolution screens on the website better.
* Small changes needed to make the tests folder available in the subproject until we have our own dlt pytest module

Review Notes:
* The code of snippets, notebooks and examples has not been changed (or just slightly, only typing comments and other typing related stuff), they are just reformatted, so no need to check deeply imho
* Please make sure that the new workflows and Makefile commands secure the same amount of linting and testing as we had before on each run. I double checked it myself but a second set of eyes would be good.


pyproject.toml
```diff
@@ -244,6 +244,9 @@ dev = [
     "types-paramiko>=3.5.0.20250708",
     "graphviz>=0.21",
     "nbqa>=1.9.1",
+    "pymongo>=4.15.3",
```
sh-rp: these are needed for tests, were in the docs section before

docs/pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+
+[project.scripts]
+lint-embedded-snippets = "docs_tools.snippets.lint_embedded_snippets:main"
+fix-grammar = "docs_tools.grammar.fix_grammar_gpt:main"
+preprocess-docs = "docs_tools.markdown.preprocess_docs:main"
+clean-pydoc-sidebar = "docs_tools.api_docs.clean_pydoc_sidebar:main"
+prepare-examples-tests = "docs_tools.examples.prepare_examples_tests:main"
+
+[tool.uv.sources]
+dlt = { path = "..", editable = true }
```
rudolfix, line 71: good!

docs/sitecustomize.py
```diff
@@ -0,0 +1,7 @@
+import os, sys
+
+# make tests folder available to the docs tests, not needed after we add a dlt pytest plugin
+tests_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "tests"))
```
rudolfix: this is not needed. you do it in pyproject, this is the main one. also
- carry some of the markers (rfam)
- clean pythonpath
- addopts

```
[tool.pytest.ini_options]
pythonpath = ["dlt", "docs/website/docs"]
norecursedirs = [".direnv", ".eggs", "build", "dist"]
addopts = "--showlocals --durations 10 -m 'not rfam'"
xfail_strict = true
log_cli_level = "INFO"
console_output_style = "count"
python_files = ["test_*.py", "*_test.py", "*snippets.py", "*snippet.pytest"]
python_functions = ["*_test", "test_*", "*_snippet"]
filterwarnings = ["ignore::DeprecationWarning"]
markers = [
    "essential: marks all essential tests",
    "no_load: marks tests that do not load anything",
    "rfam: marks tests that use rfam db"
]
```

docs/website/docs/conftest.py
```diff
@@ -1,5 +1,12 @@
 import os
 
+import sys
+
+# make tests folder available to the docs tests
```
rudolfix: this is not necessary with pythonpath setting

pyproject.toml
```diff
@@ -301,27 +300,6 @@ dbt = [
...
-    "lancedb>=0.8.2 ; python_version < '3.13'",
-    "openai>=1.45",
-    "connectorx>=0.4.2 ; python_version >= '3.10'",
-    "modal>=0.64.170",
-    # google secrets provider
-    "google-api-python-client>=1.7.11",
-    "google-auth-oauthlib>=1.0.0,<2",
-]
 
 [project.entry-points.dlt]
```
rudolfix, line 304: pls clean up pytest.ini entry

rudolfix (changes requested):
thx for fixing this!  IMO small pytest.ini cleanup is needed

rudolfix:
@sh-rp also main CONTRIBUTING. does it need to change?

sh-rp:
> @sh-rp also main CONTRIBUTING. does it need to change?

it looks like all the info is in the docs readme and nothing in the main contributing guide.

pyproject.toml
sh-rp: I removed the docs paths and test patterns, I think that should be sufficient, right?

docs/website/docs/conftest.py
sh-rp: Thanks for the tip, this works!

rudolfix (dismissed):
LGTM!

PR #3248, opened 2025-10-24
adds more signal options

rudolfix:
### Description
More signal handling options for typical production cases. See commit log and docs update


docs/website/docs/running-in-production/running.md
```diff
@@ -325,3 +327,68 @@ if __name__ == "__main__":
     load_info = load()
 ```
 
+### Allow for graceful shutdown
+`dlt` attempts a graceful shutdown of a running pipeline with custom signals handling. Currently SIGINT (CTRL-C) and SIGTERM
+are intercepted. Handlers are installed only when it is necessary:
+- `normalize` step: raises `SignalReceivedException` at certain checkpoints, typically immediately
+- `load` step: on a first received signal, it attempts to drain the job pool by not accepting new load jobs but waiting for executing jobs to complete.
+   On a second signal, default handler is called resulting in KeyboardInterrupt or immediate process termination (SIGTERM)
+- `extract` step does not intercept signals and uses default handlers. However it will raise `SignalReceivedException` is `dlt` custom signalling is activated.
```
sh-rp: To me it is not cleaer what "dlt custom signalling" means

docs/website/docs/running-in-production/running.md
```diff
@@ -325,3 +327,68 @@ if __name__ == "__main__":
...
+
+Signal interception works in orchestrators that run your code in a separate process and propagate SIGTERM/SIGINT:
+- Dagster: default multiprocess and Kubernetes executors start a process per op/run; Kubernetes will send SIGTERM, respect the Pod grace period. Avoid thread-based executors for the pipeline step or wrap with delayed_signals as shown above.
+- Airflow: task runners execute each task in its own process (Local/Celery/Kubernetes executors). On cancel/timeout, the task process receives SIGTERM then SIGKILL; if using KubernetesExecutor, rely on the Pod grace period.
+- Prefect: Subprocess flow/task runners and Kubernetes jobs deliver SIGTERM to your process; if you use thread-based concurrency inside a task, wrap the outermost entrypoint with delayed_signals.
+
+:::
+
+#### Write custom signal handler
+You can install your own signal handler. We are working on making the `signals.py` pluggable. Currently you can just disable signal interception globally
```
sh-rp: Here I would also explain better what "signal interception" means. I know what "intercept_signals" does when reading the code, but if I were only to read the docs, I would not know.

docs/website/docs/running-in-production/running.md
```diff
@@ -325,3 +327,68 @@ if __name__ == "__main__":
     load_info = load()
 ```
 
+### Allow for graceful shutdown
```
sh-rp: Maybe run the grammar checker on this whole block.

dlt/load/load.py
```diff
@@ -89,6 +89,9 @@ def __init__(
         self.load_storage: LoadStorage = self.create_storage(is_storage_owner)
         self._loaded_packages: List[LoadPackageInfo] = []
         self._job_metrics: Dict[str, LoadJobMetrics] = {}
+        # job completed signalling event (start blocking)
+        self._done_event = BoundedSemaphore()
+        self._done_event.acquire()
```
sh-rp, line 94: This I don't understand. You acquire this once on the main thread but every individual job releases it on the job thread? How does this work? I think I am missing some part of this, I would expect an exception if there are more than 2 jobs.

rudolfix: that's actually trivial: semaphore is created in release state so first `acquire` will just get through. I want first loop in `load` to sleep. if I do not acquire semaphore right away - it skips. btw. typically you can set if semaphore is released or not (initial value) in constructor. but not in Python apparently

docs/website/docs/running-in-production/running.md
rudolfix: yeah you are right I'll do that :/

PR #3250, opened 2025-10-24
Feat: last pipeline run section in dashboard

anuunchin:
This PR introduces a new pipeline run section to the dashboard app which uses a simple html snippet.

Screenshot 2025-11-03 at 12 04 30


rudolfix:
terminology (from https://www.notion.so/dlthub/dlt-Entities-their-Lifecycles-and-Relations-2679fb8e23cf80169921c4ea069bfff8?source=copy_link#2679fb8e23cf8109b9c3eee928a2a377). definitions are pretty clear and you can easily refer to code
- pipeline run succeeded or failed (correct your status)

Idea display status of load package(s) as boxes like the "COMPLETED"
- pipeline created load packages: collect packages from all steps to get load id and package
- display color coded states of all packages (with load ids on it): pending | loaded | aborted
- package can be "partially loaded" (I think in pending and aborted states). there's a method on package info. I hope we have it in trace. if you see it - it should be a red color ie: "load_id: partially loaded in aborted"

schema indicators
- look for "applied schema update" if it is found in the trace - display "migration" box

for later
- detect state sync (loaded job with state table name)
- detect `dev_mode` (I do not think we have it in trace but we could use dataset name)
- detect full `refresh` type (I do not think we hace it in trace)

feedback to current UI
experiment: if we have trace just put it on top, without "Pipeline Run" and switch. Also you do not need to repeat pipeline name, it is already on top

dlt/_workspace/helpers/dashboard/dlt_dashboard.py
```diff
@@ -72,20 +72,32 @@ def home(
...
-                    label="<small>Open local data location</small>",
-                    on_click=lambda _: utils.open_local_folder(local_dir),
+            if local_dir := utils.get_local_data_path(dlt_pipeline):
+                _buttons.append(
+                    mo.ui.button(
+                        label="<small>Open local data location</small>",
+                        on_click=lambda _: utils.open_local_folder(local_dir),
+                    )
+                )
+            if dlt_pipeline.last_trace:
```
rudolfix: last_trace loads trace and is pretty heavy. better use := to get trace object in `if`

dlt/_workspace/helpers/dashboard/utils.py
```diff
@@ -809,3 +824,287 @@ def _humanize_datetime(dt: Union[str, int]) -> str:
...
+}
+
+
+class PipelineStepData(NamedTuple):
+    step: TVisualPipelineStep
+    duration_ms: float
+    failed: bool
+
+
+def _format_duration(ms: float) -> str:
```
rudolfix, line 853: we have `humanize` library. it should be able to generate friendly strings. ie `humanize.precisedelta` pls take a look

dlt/_workspace/helpers/dashboard/utils.py
```diff
@@ -809,3 +824,287 @@ def _humanize_datetime(dt: Union[str, int]) -> str:
...
+    # Build the general info of the exection
+    general_info = f"""
+    <div>Last execution ID: <strong>{transaction_id[:8]}</strong></div>
+    <div>Total time: <strong>{_format_duration(total_ms)}</strong></div>
+    """
+
+    # Build the pipeline execution timeline bar and labels
+    segments, labels = [], []
+    for i, step in enumerate(steps_data):
+        percentage = step.duration_ms / total_ms * 100
```
rudolfix, line 901: not sure if step may have `0` duration. but makes sense to check it

dlt/_workspace/helpers/dashboard/utils.py
```diff
@@ -809,3 +824,287 @@ def _humanize_datetime(dt: Union[str, int]) -> str:
...
+        segments.append(
+            "<div"
+            f' style="width:{percentage}%;background-color:{color};border-radius:{radius};"></div>'
+        )
+        labels.append(
+            f'<span><span style="color:{color};">●</span> '
+            f"{step.step.capitalize()} {_format_duration(step.duration_ms)}</span>"
+        )
+    segments_bar = f"""
+    <div style="
```
rudolfix: OMG. I think it is good for now. but we could probably use css to keep those.

rudolfix (changes requested):
LGTM. all the comments are minor. visually is really good.

you could add example pipeline with partially loaded package:

```py
@dlt.destination
    def wait_forever(item, schema):
        # never exit
        from time import sleep

        # this one does not wake up on signal
        while not _done:
            sleep(1)

        # make the job fail if it gets here
        raise KeyboardInterrupt()

    pipeline = dlt.pipeline(
        "signal_waiter", destination=wait_forever(), dataset_name="_data", progress=DictCollector()
    )
```
you can modify this destination so it passes for one table but raises for another triggering "partially loaded" state.

dlt/_workspace/helpers/dashboard/utils.py
anuunchin: humanize doesn't seem to be generating friendly enough strings - I think it's better to leave it the way it is currently 👀

PR #3262, opened 2025-10-28
fixes git import and enables tests

### LINKED ISSUE #3263: dlt 1.18.0 requires git runtime
### dlt version

1.18.0

### Describe the problem

Using dagster_dlt an error was raised due to a missing git dependency: 

> `The git executable must be specified in one of the following ways:
>     - be included in your $PATH
>     - be set via $GIT_PYTHON_GIT_EXECUTABLE
>     - explicitly set via git.refresh(<full-path-to-git-executable>)
> 
> All git commands will error until this is rectified.
> 
> This initial message can be silenced or aggravated in the future by setting the
> $GIT_PYTHON_REFRESH environment variable. Use one of the following values:
>     - quiet|q|silence|s|silent|none|n|0: for no message or exception
>     - warn|w|warning|log|l|1: for a warning message (logging level CRITICAL, displayed by default)
>     - error|e|exception|raise|r|2: for a raised exception
> 
> Example:
>     export GIT_PYTHON_REFRESH=quiet
> ` while importing module qxdl.catalog.definitions. Local modules were resolved using the working directory `/project/dagster`. If another working directory should be used, please explicitly specify the appropriate path using the `-d` or `--working-directory` for CLI based targets or the `working_directory` configuration option for workspace targets. `

Installing git in the container resolves the issue. When pinning to dlt==1.17.1 there is no issue.

### Expected behavior

The dlt modules should load without errors even if the git runtime is not present.

### Steps to reproduce

pip install dlt latest (1.18.0)
import dlt 

### Operating system

Linux

### Runtime environment

Docker, Docker Compose

### Python version

3.12

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
* fixes #3263 
* enables minimal dlt import and run tests on alpine
* enable hub tests and transformation snippets
* bumps version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,8 +134,8 @@ clean-dist:
 
 publish-library: clean-dist build-library
 	ls -l dist/
-	@read -p "Enter PyPI API token: " PYPI_API_TOKEN; echo ; \
-	uv publish --token "$$PYPI_API_TOKEN"
+	@bash -c 'read -s -p "Enter PyPI API token: " PYPI_API_TOKEN; echo; \
+	uv publish --token "$$PYPI_API_TOKEN"'
 
 test-build-images: build-library
 	# NOTE: uv export does not work with our many different deps, we install a subset and freeze
--- a/dlt/_workspace/_plugins.py
+++ b/dlt/_workspace/_plugins.py
@@ -10,9 +10,7 @@ __all__ = ["plug_workspace_context_impl"]
 
 def is_workspace_dir(run_dir: str) -> bool:
     """Checks if `run_dir` contains dlt workspace, this is true if a config file is found"""
-    if os.path.isfile(os.path.join(run_dir, DOT_DLT, ".workspace")):
-        return True
-    return False
+    return os.path.isfile(os.path.join(run_dir, DOT_DLT, ".workspace"))
 
 
 @_plugins.hookimpl(specname="plug_run_context", trylast=False)
--- a/dlt/pipeline/current.py
+++ b/dlt/pipeline/current.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 """Easy access to active pipelines, state, sources and schemas"""
 
 from typing import TYPE_CHECKING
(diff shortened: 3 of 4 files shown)


dlt/_workspace/_plugins.py
```diff
@@ -1,21 +1,20 @@
...
+from dlt.common.runtime.run_context import DOT_DLT, RunContext
 
 __all__ = ["plug_workspace_context_impl"]
 
 
+def is_workspace_dir(run_dir: str) -> bool:
+    """Checks if `run_dir` contains dlt workspace, this is true if a config file is found"""
+    if os.path.isfile(os.path.join(run_dir, DOT_DLT, ".workspace")):
+        return True
+    return False
```
zilto: nit: could be `return os.path.isfile(os.path.join(run_dir, DOT_DLT, ".workspace"))`

dlt/pipeline/current.py
```diff
@@ -30,14 +33,19 @@ def pipeline() -> _Pipeline:
     return _pipeline()
 
 
+def workspace() -> "WorkspaceRunContext":
```
zilto: nit: no need to use `"type"` if you add `from __future__ import annotations`

Makefile
```diff
@@ -134,18 +134,17 @@ clean-dist:
 
 publish-library: clean-dist build-library
 	ls -l dist/
-	@read -sp "Enter PyPI API token: " PYPI_API_TOKEN; echo ; \
+	@read -p "Enter PyPI API token: " PYPI_API_TOKEN; echo ; \
```
zilto: If I understand, this is no longer passed as a secret on the user's terminal history. Shouldn't change anything in CI behavior?

zilto (commented):
Looks good. A few nit. Feel free to ignore if it slows down patch release

Makefile
rudolfix: this works only on mac. on linux `sh` is the default shell and it does not have this option. I enforced shell here. btw. this is not run on ci

PR #3271, opened 2025-10-30
fix/3165 Athena LakeFormation permissions are required even tho Lakeformation is not used

### LINKED ISSUE #3165: Athena LakeFormation permissions are required
### dlt version

1.17.0

### Describe the problem

When using the athena destination, it seems that lakeformation permission are required.

`An error occurred (AccessDeniedException) when calling the GetResourceLFTags operation: User: <redacted> is not authorized to perform: lakeformation:GetResourceLFTags on resource: arn:aws:lakeformation:eu-central-1:<redacted>:catalog:<redacted> because no identity-based policy allows the lakeformation:GetResourceLFTags action`.

The code correctly checks if the config block is none to call lake formation, but somehow the default config is not None. https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/athena/athena.py#L345
I wasn't able to figure out where why the config doesn't default to None https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/athena/configuration.py#L27

### Expected behavior

The athena destination should not require Lakeformation permission when lakeformation is not used.

### Steps to reproduce

- Setup an IAM user without any lakeformation permissions.
- Configure an Athena destination with default configurations:

  ```
  DESTINATION__FILESYSTEM__BUCKET_URL="s3://my-bucket/dlt"
  DESTINATION__QUERY_RESULT_BUCKET="s3://my-athena-bucket/dlt"
  DESTINATION__ATHENA_WORK_GROUP="default"
  ```

- Run a pipeline against the destination
- The pipeline should fail with an error that the user is not authorized to perform GetResourceLFTags 

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

When defining the destination in code, you can set the lakeformation_config explicitly to None, then it works as expected.
```
    destination = athena(
        query_result_bucket="...",
        athena_work_group="...",
        lakeformation_config=None,
    )
```

However, I'm not sure how you can configure an environment variable or toml variable to None?

alkaline-0:
### Description

1. Gate Lake Formation (LF) tag management strictly on configuration: tags should not be managed unless LF is explicitly set and configured.
2. Add tests to ensure no LF calls are made when LF config is not set, preventing unintended GetResourceLFTags requests.

### Related Issues
- [Fixes #3165](https://github.com/dlt-hub/dlt/issues/3165)

### Changes:
1. implemented a test in /tests/load/pipeline/test_athena.py
2. Changed the check in AthenaClient to verify that enabled flag is true before calling manage_lf_tags() since the resolver does not change that value even tho it initializes an object of lakeformation even tho lakeformation config obj is None.

### DIFF
--- a/tests/load/pipeline/test_athena.py
+++ b/tests/load/pipeline/test_athena.py
@@ -28,26 +28,37 @@ pytestmark = pytest.mark.essential
 
 @pytest.mark.parametrize(
     "destination_config",
-    destinations_configs(default_sql_configs=True, subset=["athena"]),
+    destinations_configs(
+        default_sql_configs=True,
+        subset=["athena"],
+        with_table_format="iceberg",
+    ),
     ids=lambda x: x.name,
 )
+@pytest.mark.parametrize("lf_enabled", [True, False], ids=["lf-on", "lf-off"])
 def test_athena_lakeformation_config_gating(
-    destination_config: DestinationTestConfiguration, mocker
+    destination_config: DestinationTestConfiguration, lf_enabled: bool, mocker, monkeypatch
 ) -> None:
-    pytest.importorskip("pyathena")
+    # Configure Lake Formation gating via env (read by client config)
+    monkeypatch.setenv("DESTINATION__LAKEFORMATION_CONFIG__ENABLED", str(lf_enabled))
+
     pipeline = destination_config.setup_pipeline("athena_" + uniq_id(), dev_mode=True)
 
     with pipeline.destination_client() as client:
         mocked_manage = mocker.patch(
             "dlt.destinations.impl.athena.athena.AthenaClient.manage_lf_tags"
         )
+        # avoid side effects in parent update_stored_schema; gating still runs in AthenaClient
         mocker.patch(
             "dlt.destinations.job_client_impl.SqlJobClientWithStagingDataset.update_stored_schema",
             return_value=None,
         )
 
         client.update_stored_schema()
-        mocked_manage.assert_not_called()
+        if lf_enabled:
+            mocked_manage.assert_called()
+        else:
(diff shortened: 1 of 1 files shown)


tests/load/pipeline/test_athena.py
```diff
@@ -26,6 +26,30 @@
...
 
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["athena"]),
+    ids=lambda x: x.name,
+)
+def test_athena_lakeformation_config_gating(
+    destination_config: DestinationTestConfiguration, mocker
+) -> None:
+    pytest.importorskip("pyathena")
```
sh-rp: This should not be needed. This test is only selected when the athena destination is active(see destination_config setup above) and in that case, the athena extra will also be present.

tests/load/pipeline/test_athena.py
```diff
@@ -26,6 +26,30 @@
 pytestmark = pytest.mark.essential
 
 
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["athena"]),
+    ids=lambda x: x.name,
+)
+def test_athena_lakeformation_config_gating(
```
sh-rp, line 39: I think these kind of tests are "more secure" if you parametrize them to have one run with lf enabled and check that the mocked function is called and then one run where it is not and thus the mocked function is not called

rudolfix (commented):
LGTM! this type of test is sufficient - clearly we deal with configuration problem, not edge behavior of athena

sh-rp (approved):
LGTM

PR #3273, opened 2025-10-31
resolves "default" limit via ibis options

rudolfix:
### Description
resolves "default" limit via ibis options, previously "default" str was cast to int


zilto (approved):
LGTM. Waiting for CI to complete

PR #3275, opened 2025-10-31
feat: updated scaffolding template

zilto:
This updates the template with several changes:

1. more descriptive module-level docstring; removes keyword `vibe coding` and specifics about dlt commands
2. fix import path `from dlt.sources.rest_api.typing`. The previous import was unintended and creates linting errors with pylance (this throws off IDE agents like Cursor)
3. Remove instructions to set API key in plaintext. This suggests that the LLM should read / write your API key. Instead, added a comment about how `dlt.secrets.value` works
4. Remove `get_data()` function. This scope is unecessary and adds complexity if we intend to run this as a script.
5. Remove `dataset_name` kwarg. This simplifies the work of the LLM. Anyways, the name is derived from the pipeline name.

~6. Use `dev_mode=True`, this avoids schema migrations across runs.~
~7. Use `refresh="drop_sources"` to avoid state conflicts as the agents does multiple iterations~

### DIFF
--- a/docs/education/dlt-fundamentals-course/lesson_3_pagination_and_authentication_and_dlt_configuration.ipynb
+++ b/docs/education/dlt-fundamentals-course/lesson_3_pagination_and_authentication_and_dlt_configuration.ipynb
@@ -560,6 +560,7 @@
     "from dlt.sources.helpers.rest_client.auth import BearerTokenAuth\n",
     "from dlt.common.typing import TDataItems\n",
     "\n",
+    "\n",
     "@dlt.source\n",
     "def github_source(\n",
     "    access_token=dlt.secrets.value,\n",


rudolfix (dismissed):
LGTM, thx for fixing this

rudolfix (approved):
Note: I've modified AST parser to allow optional dataset names and script import checker to allow pipelines to be instantiated on import (makes sense also for runtime)

pipeline running, iterating resources and source instantiation is not allowed

zilto:
Converted to draft to avoid merging. I realized that `dev_mode` has some counterintuitive behavior #3276 which may be confusing for humans / LLMs. 

I think the `workspace` feature and automated local `pipelines_dir` is generally helpful for LLMs since it gives the full "project context/state" under the IDE project root.

PR #3279, opened 2025-11-03
clones command repos in global_dir & bumps to 1.18.2

rudolfix:
### Description
clones command repos in global_dir & bumps to 1.18.2

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "dlt"
-version = "1.18.1"
+version = "1.18.2"
 description = "dlt is an open-source python-first scalable data loading library that does not require any backend to run."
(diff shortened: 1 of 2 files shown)


dlt/_workspace/cli/utils.py
```diff
@@ -16,7 +16,6 @@
 from dlt.common.runtime.telemetry import with_telemetry
 from dlt.common.storages.file_storage import FileStorage
 
-from dlt._workspace.profile import LOCAL_PROFILES
```
sh-rp, line 19: As far as I can see, this variable is never used anywhere in dlt, so you can also delete it from `profile.py`

rudolfix: let's keep them for now. I planned additional warnings when you delete data in profile that is not "local"

Issue #3042, opened 2025-08-29
'Connection' object has no attribute 'thrift_backend'


themattmorris:
### dlt version

1.15.0

### Describe the problem

When using databricks-sql-connector versions greater than 4.0.5, the following error is encountered:

```
dlt.load.exceptions.LoadClientJobRetry: Job with `job_id=attachments.ca2dcd246f.parquet` had 5 retries which is a multiple of `max_retry_count=5`. Exiting retry loop. You can still rerun the load package to retry this job. Last failure message was: 'Connection' object has no attribute 'thrift_backend'
```

It looks like the `thrift_backend` attribute has been removed from the `databricks-sql-connector` beginning with version `4.0.6`.

This line is where the problem is occurring:

https://github.com/dlt-hub/dlt/blob/edad825a593b0e901145b8ee8d77371982b33418/dlt/destinations/impl/databricks/databricks.py#L79

### Steps to reproduce

Run any dlt pipeline with a rest_api resource on Databricks with a version of `databricks-sql-connector > 4.0.6`. I can provide exact code snippets if necessary.

### Operating system

Linux

### Runtime environment

Kubernetes

### Python version

3.13

### dlt data source

Custom OneNote client

### Other deployment details

This is deploying to Databricks. I could not find that listed in one of the destination options.

sh-rp:
Hey @themattmorris, I can't seem to reproduce this. I am testing on python 3.13 with databricks-sql-connector version 4.1.2:

```
├── databricks-sql-connector v4.1.2 (extra: databricks)
│   ├── lz4 v4.4.4
│   ├── oauthlib v3.2.2
│   ├── openpyxl v3.1.5
│   │   └── et-xmlfile v2.0.0
│   ├── pandas v2.2.3 (*)
│   ├── pyjwt v2.10.1 (*)
│   ├── python-dateutil v2.9.0.post0 (*)
│   ├── requests v2.31.0 (*)
│   ├── thrift v0.20.0
│   │   └── six v1.17.0
│   └── urllib3 v1.26.20
```

Can you provide a self contained pipeline script that reproduces this error and a list of all your installed dependencies?

Thanks

themattmorris:
Hi, thanks for the response! I am seemingly using a very similar setup. This is running python 3.13 on Linux. Here are the python dependencies:

```bash
pip install dlt[databricks,parquet]
```

```
databricks-sdk==0.65.0
[...]

PR #3313, opened 2025-11-12
fixes athena refresh mode

### LINKED ISSUE #3296: Refreshing pipeline state & data with `refresh="drop_sources"` does not work for Iceberg parquet with destination Athena
### dlt version

dlt 1.18.1

### Describe the problem

I'm setting up a dlt pipeline that moves data over to an AWS Glue Catalog, made presentable in Athena as an Iceberg Parquet. During development, I wanted to be able to 'reset' my dlt pipeline cleanly, which is why I tried out the `dev_mode=true` argument as mentioned in the [doc ](https://dlthub.com/docs/general-usage/pipeline#do-experiments-with-dev-mode). However, this argument is not suitable for my case as it introduces a new AWS Glue Database on every run (for which I still need to do manual cleanup in Glue & S3 afterwards). For this reason, I looked into refreshing the pipeline state with other args (as mentioned in the [doc](https://dlthub.com/docs/general-usage/pipeline#refresh-pipeline-data-and-state)).

Upon using `pipeline.run(..., refresh="drop_resources")`, it became apparent to me that the arguments simply do not work as intended for an Athena Iceberg Parquet destination with dlt. The reason for this is that dlt first removes the data in S3, but tries to query the Iceberg table afterwards. Since the Iceberg metadata has already been deleted, this query fails and terminally fails the dlt pipeline.

### Expected behavior

The expected behaviour of the ´refresh="drop_resources"´ is not documented, but for ´drop_sources´ it is documented [here](https://dlthub.com/docs/general-usage/pipeline#drop-tables-and-pipeline-state-for-a-source-with-drop_sources).

Image

After doing an initial run, a parquet file lands in the staging directory and it is correctly loaded to my destination (as an Iceberg table in an AWS Glue Database). After doing a second run, I would expect the run to:

1 Extract the data
2 Normalize it according to my code
3 Remove the AWS S3 data & AWS Glue tables for both the staging & final datasets
4 Remove records in the _dlt tables related to this dataset
5 Upload the staging data
6 Load the new dataset as an Iceberg Parquet

### Steps to reproduce

The issue can be easily reproduced by loading any dataset to AWS Glue/Athena as destination while specifying the ´refresh="drop_resources"´ arg. The first run will succeed, but the second one will fail with the following error;

```bash
<class 'dlt.load.exceptions.LoadClientJobFailed'>
Job with `job_id=my_table.ab2811c91f.sql` and `load_id=1762440090.5874617` failed terminally with message: [ErrorCode: INTERNAL_ERROR_QUERY_ENGINE] Amazon Athena experienced an internal error while executing this query. Please contact AWS support for further assistance. You will not be charged for this query. We apologize for the inconvenience.. The package is aborted and cannot be retried.
```

This happened because the following query in Athena failed:

```sql
DELETE FROM "my-dataset"."my-table" WHERE 1=1
```

Upon further inspection, the underlying error in Athena was:
```bash
GENERIC_INTERNAL_ERROR: com.amazonaws.trino.exceptions.UnrecoverableS3OperationException: com.amazonaws.services.s3.model.AmazonS3Exception: The specified key does not exist. (Service: Amazon S3; Status Code: 404; Error Code: NoSuchKey; ... , Key: my-dir/private/dlt/my-dataset/my-table/metadata/00000-677a1769-e85c-4bde-a33d-7543bb2bcc15.metadata.json)
```

Which make sense, since the table still exists in Glue and shows up in Athena, but internally refers to the already removed Iceberg metadata.json file in S3.

Image

### Operating system

Linux

### Runtime environment

Virtual Machine

### Python version

3.13

### dlt data source

I'm using an Oracle database as a datasource, but this is unrelated to the issue.

### dlt destination

AWS Athena / Glue Catalog

### Other deployment details

_No response_

### Additional information

Here's my relevant code:

```python
    source = sql_database(
        engine_from_credentials(get_db_credentials()),
        schema="my_schema",
        defer_table_reflect=True,
        table_names=[table_name],
        backend="pyarrow",
        query_adapter_callback=query_adapter_callback,
    )

    source.my_table_name.apply_hints(
        write_disposition={"disposition": "replace"},
        primary_key="primary_id",
    )

    destination = athena(
        query_result_bucket=f"s3://{athena_bucket_name}/my-dir/dlt",
        aws_data_catalog="awsdatacatalog",
        athena_work_group=f"my-workgroup",
        force_iceberg=True,
        enable_dataset_name_normalization=False,
        lakeformation_config=None
    )    

pipeline_name = "extract_my_model"

    pipeline = dlt.pipeline(
        pipeline_name=pipeline_name,
        destination=destination,
        staging=filesystem(
            bucket_url=f"s3://{bucket_name}/{my-dir}/private/dlt",
            enable_dataset_name_normalization=False,
        ),
        dataset_name=f"my-dataset",
        progress="log",
    )

    load_info = pipeline.run(source, refresh="drop_resources")
    logger.info(f"Load info: {load_info}")
    #logger.info(pipeline.default_schema.to_pretty_yaml())

    logger.info(pipeline.last_trace)
```

rudolfix:
### Description

fixes #3296 - here we forgot a selector for dropped tables on Athena - which functions similarly to truncated tables selector.
fixes: #3042


sh-rp (approved):
LGTM

PR #3321, opened 2025-11-14
fix/3190 Fixed the persistence issue of boundary timestamp after removing it

### LINKED ISSUE #3190: The value of _dlt_valid_from and _dlt_valid_to keeps being set to a former `boundary_value`
### dlt version

1.17.1

### Describe the problem

I’d like to set a custom value for `_dlt_valid_from` the first time I run the pipeline with scd2 merge strategy, and pipeline run timestamp for the following runs. I tried setting `boundary_timestamp` for the first run and then removing it.

The first time, I’m pulling data from a local backup :

```
@dlt.resource(
    write_disposition={"disposition": "merge", "strategy": "scd2", "boundary_timestamp": "2024-12-12"},
)
def my_resource():
    # reading data from a local json backup
    yield {"a": 1}
```

For the second run, I’m deleting the "boundary_timestamp":

```
@dlt.resource(
)
def my_resource():
    # pulling data from an external API
    yield {"a": 2}
```

The table after the second run looks like this:

```
> select _dlt_valid_from, _dlt_valid_to from table;
┌──────────────────────────┬──────────────────────────┬───────┬────────────────────┬────────────────┐
│     _dlt_valid_from      │      _dlt_valid_to       │   a   │    _dlt_load_id    │    _dlt_id     │
│ timestamp with time zone │ timestamp with time zone │ int64 │      varchar       │    varchar     │
├──────────────────────────┼──────────────────────────┼───────┼────────────────────┼────────────────┤
│ 2024-12-12 01:00:00+01   │ 2024-12-12 01:00:00+01   │     1 │ 1760084979.8407435 │ Ec9HPMTWWD8wGA │
│ 2024-12-12 01:00:00+01   │ NULL                     │     2 │ 1760085096.16156   │ Wb8LIv72/V4uzw │
└──────────────────────────┴──────────────────────────┴───────┴────────────────────┴────────────────┘
```
The first row’s `_dlt_valid_to` column and the second row’s `_dlt_valid_from` column were both set to the initial value of `boundary_timestamp`.

### Expected behavior

I expect the first row’s `_dlt_valid_to` column and the second row’s `_dlt_valid_from` column to be set to the current timestamp at the moment when the pipeline is run.

### Steps to reproduce

First run:

```
@dlt.resource(
    write_disposition={"disposition": "merge", "strategy": "scd2", "boundary_timestamp": "2024-12-12"},
)
def my_resource():
    # reading data from a local json backup
    yield {"a": 1}
```

Second run, without `boundary_timestamp` and modified data:

```
@dlt.resource(
)
def my_resource():
    # pulling data from an external API
    yield {"a": 2}
```

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

_No response_

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

alkaline-0:
## Description

- Fix SCD2 sticky boundary behavior: when write_disposition.strategy="scd2" and boundary_timestamp is omitted, always set x-boundary-timestamp to the current load package’s created_at. This ensures each run uses the effective time of that run rather than persisting the previous boundary.
- Also updated the SCD2 tests to validate this case

## Related Issues
Fixes #3190

### DIFF
--- a/dlt/destinations/impl/sqlalchemy/merge_job.py
+++ b/dlt/destinations/impl/sqlalchemy/merge_job.py
@@ -375,9 +375,12 @@ class SqlalchemyMergeFollowupJob(SqlMergeFollowupJob):
             format_datetime_literal = (
                 DestinationCapabilitiesContext.generic_capabilities().format_datetime_literal
             )
-        created_at = current_load_package()["state"]["created_at"]
         _boundary_ts = cast(Optional[TAnyDateTime], root_table.get("x-boundary-timestamp"))
-        boundary_ts: TAnyDateTime = _boundary_ts if _boundary_ts is not None else created_at
+        boundary_ts: TAnyDateTime = (
+            _boundary_ts
+            if _boundary_ts is not None
+            else current_load_package()["state"]["created_at"]
+        )
         boundary_ts = ensure_pendulum_datetime_utc(boundary_ts)
 
         boundary_literal = format_datetime_literal(boundary_ts, caps.timestamp_precision)
--- a/dlt/extract/hints.py
+++ b/dlt/extract/hints.py
@@ -823,24 +823,21 @@ class DltResourceHints:
 
             if wd.get("strategy") == "scd2":
                 wd = cast(TScd2StrategyDict, wd)
-
-                art = wd.get("active_record_timestamp")
-                if art is not None:
-                    try:
-                        ensure_pendulum_datetime_utc(art)
-                    except (ValueError, TypeError) as exc:
-                        raise ValueError(
-                            f"could not parse `active_record_timestamp` value `{art}`"
-                        ) from exc
-
-                bt = wd.get("boundary_timestamp")
-                if bt is not None:
-                    try:
-                        ensure_pendulum_datetime_utc(bt)
-                    except (ValueError, TypeError) as exc:
-                        raise ValueError(
-                            f"could not parse `boundary_timestamp` value `{bt}`"
-                        ) from exc
+                for ts in ("active_record_timestamp", "boundary_timestamp"):
+                    if (
+                        ts == "active_record_timestamp"
+                        and wd.get("active_record_timestamp") is None
+                    ):
+                        continue  # None is allowed for active_record_timestamp
+                    if ts in wd:
+                        if wd[ts] is None:  # type: ignore[literal-required]
+                            continue
+                        try:
+                            ensure_pendulum_datetime_utc(wd[ts])  # type: ignore[literal-required]
+                        except Exception:
+                            raise ValueError(
+                                f"could not parse `{ts}` value `{wd[ts]}`"  # type: ignore[literal-required]
+                            )
 
     @staticmethod
     def validate_reference_hint(template: TResourceHints) -> None:


dlt/extract/hints.py
```diff
@@ -722,8 +725,17 @@ def _merge_merge_disposition_dict(dict_: Dict[str, Any]) -> None:
 
         if merge_strategy == "scd2":
             md_dict = cast(TScd2StrategyDict, md_dict)
-            if "boundary_timestamp" in md_dict:
+            boundary = md_dict.get("boundary_timestamp")
```
rudolfix: the root problem here IMO is that user is not correctly resetting the boundary value (or it does not work). we already do what you implemented below when we generate SQL:
```
boundary_ts = ensure_pendulum_datetime_utc(
            root_table.get("x-boundary-timestamp", current_load_package()["state"]["created_at"])  # type: ignore[arg-type]
        )
```
and IMO the problem was user was skipping this hint on a next run instead of setting boundary to `None`. Let's do an experiment first. see below

tests/load/pipeline/test_scd2.py
```diff
@@ -726,6 +728,37 @@ def r(data):
                 "boundary_timestamp": ts4,
             }
         )
+    with mock.patch(
+        "dlt.common.storages.load_package.precise_time",
```
rudolfix: let's do an experiment here. please stash changes in `hints.py` so we work with the original code and modify the test to pass `"boundary_timestamp": None` 

this test should pass (we drop the hint, code in `sql_jobs.py` sets the package timestamp). if the test passes we just need to amend our docs in scd2 on how to reset the boundary.

rudolfix (changes requested):
pls check the review comments. schema hints are additive: schema fragments may come from many places: import schema, explicit code, resource hints, apply hints etc. the fact that hint is not present is not enough to remove it from the schema.
That's why user needs to pass None here. If it does not work we have a real problem though...

dlt/extract/hints.py
```diff
@@ -823,19 +823,24 @@ def validate_write_disposition_hint(template: TResourceHints) -> None:
...
-                        continue  # None is allowed for active_record_timestamp
-                    if ts in wd:
-                        try:
-                            ensure_pendulum_datetime_utc(wd[ts])  # type: ignore[literal-required]
-                        except Exception:
-                            raise ValueError(
-                                f"could not parse `{ts}` value `{wd[ts]}`"  # type: ignore[literal-required]
-                            )
+
+                art = wd.get("active_record_timestamp")
```
rudolfix: previous loop was IMO more elegant. too much duplicated code here IMO

dlt/destinations/impl/sqlalchemy/merge_job.py
```diff
@@ -374,10 +375,10 @@ def gen_scd2_sql(
             format_datetime_literal = (
                 DestinationCapabilitiesContext.generic_capabilities().format_datetime_literal
             )
-
-        boundary_ts = ensure_pendulum_datetime_utc(
-            root_table.get("x-boundary-timestamp", current_load_package()["state"]["created_at"])  # type: ignore[arg-type]
-        )
+        created_at = current_load_package()["state"]["created_at"]
```
rudolfix: code is correct. but you do not need to take "created_at" from the package unconditionally. do it in ternary operator below `else current_load_package()["state"]["created_at"]`

same thing in sql_jobs

PR #3322, opened 2025-11-15
feat: Snowflake-provided OAuth token auth

### LINKED ISSUE #2976: Support Snowflake-provided OAuth token in Snowpark Container Services

jorritsandbrink:
### Description
This PR enables [Snowflake-provided OAuth token](https://docs.snowflake.com/en/developer-guide/snowpark-container-services/additional-considerations-services-jobs#connecting-with-a-snowflake-provided-oauth-token) authentication for the `snowflake` destination, which is relevant when running `dlt` inside Snowpark Container Services.

### Related Issues

- Closes #2976


dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -30,7 +37,17 @@ class SnowflakeCredentials(ConnectionStringCredentials):
     private_key_passphrase: Optional[TSecretStrValue] = None
     application: Optional[str] = SNOWFLAKE_APPLICATION_ID
 
-    __config_gen_annotations__: ClassVar[List[str]] = ["password", "warehouse", "role"]
+    _snowflake_host: Optional[str] = None
+    """Snowflake account URL, e.g. https://kgiotue-wn98412.snowflakecomputing.com"""
+    _use_snowflake_session_token: bool = False
+
+    __config_gen_annotations__: ClassVar[List[str]] = [
```
jorritsandbrink, line 42: I added `host` and `username` to `__config_gen_annotations__` because the `tests/workspace/cli/test_config_toml_writer.py::test_write_spec_without_defaults` was failing without them, but I don't actually know what these annotations are for.

jorritsandbrink:
@burnash Could you review this PR?

Perhaps it's better to split the code into `SnowflakeCredentialsWithoutDefaults` and `SnowflakeCredentials`, but I wasn't sure so for now it only has `SnowflakeCredentials`. Let me know what you think!

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -48,6 +65,28 @@ def parse_native_representation(self, native_value: Any) -> None:
...
+                "`username` is required in SnowflakeCredentials when `host` is provided."
+            )
+        elif self.username and not self.host:
+            raise ConfigurationValueError(
+                "`host` is required in SnowflakeCredentials when `username` is provided."
+            )
+        elif not self.host and not self.username:
+            if snowflake_session_token_available():
+                logger.info(
+                    "`host` and `username` missing in SnowflakeCredentials. Use Snowflake-provided"
```
burnash:
```suggestion
                    "`host` and `username` missing in SnowflakeCredentials. Using Snowflake-provided"
```
should it be "Using..."?

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -48,6 +65,28 @@ def parse_native_representation(self, native_value: Any) -> None:
...
+                    " OAuth token instead."
+                )
+                self._set_use_snowflake_session_token()
+            else:
+                raise ConfigurationValueError(
+                    "`host` and `username` not provided in SnowflakeCredentials, but there is no"
+                    " Snowflake-provided OAuth token available."
```
burnash:
```suggestion
                    "`host` and `username` not provided in SnowflakeCredentials. "
                    "Snowflake-provided OAuth token authentication (for Snowpark Container Services) "
                    "was attempted but failed. If running in Snowpark Container Services, ensure the token " 
                    "file exists."
```

burnash: This message appears when the token file is missing. It might confuse users who aren't running in Snowpark Container Services. I suggest mentioning Snowpark Container Services to clarify the environment. What do you think?

jorritsandbrink:
@burnash thanks! I committed your suggestions. Okay to merge now?

burnash (dismissed):
Looks good, thank you @jorritsandbrink

burnash:
@jorritsandbrink thank you very much! I'm working on fixing the flaky tests, I'll rebase this PR & merge it after I merge the tests fix.

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -18,9 +24,10 @@
 @configspec(init=False)
 class SnowflakeCredentials(ConnectionStringCredentials):
     drivername: Final[str] = dataclasses.field(default="snowflake", init=False, repr=False, compare=False)  # type: ignore[misc]
-    host: str = None
     database: str = None
-    username: str = None
+    host: Optional[str] = None
+    """Snowflake account identifier, e.g. `kgiotue-wn98412`"""
+    username: Optional[str] = None
```
rudolfix, line 29: yes, user should be optional. we should not require it when authenticator is set. actually this is IMO a bug and we can fix it here since we already do this for "default oauth" handling

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -48,6 +65,30 @@ def parse_native_representation(self, native_value: Any) -> None:
                 setattr(self, param, self.query.get(param))
 
     def on_resolved(self) -> None:
```
rudolfix, line 64: it will be way cleaner to use `on_partial` for this. this is how we handle default credentials on aws:
```py
def on_partial(self) -> None:
        # Try get default credentials
        session = self._to_botocore_session()
        if self._from_session(session) and not self.is_partial():
            self.resolve()
```
then we can still keep host required (it is for 99.9% of cases) and just set it here and resolve.

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -48,6 +65,30 @@ def parse_native_representation(self, native_value: Any) -> None:
...
+        # host and username should be both present or both missing
+        if self.host and not self.username:
```
rudolfix: If we make host (snowflake account) required we do not need those. The username depends on authenticator, not on host and what we IMO need to do is: if authenticator is not set or if it set to `snowflake` then we require user name. otherwise we let the code go. see:
https://docs.snowflake.com/en/developer-guide/python-connector/python-connector-connect

both oauth and WIF do not need user name (OKTA needs it but I hope they will just fail with a message and user will figure out they need to pass it)

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -80,17 +122,29 @@ def to_connector_params(self) -> Dict[str, Any]:
 
         conn_params: Dict[str, Any] = dict(
             query,
```
rudolfix, line 98: FYI: this will pass authenticator and token to the connection params!

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -80,17 +122,29 @@ def to_connector_params(self) -> Dict[str, Any]:
...
-            user=self.username,
-            password=self.password,
             account=self.host,
             database=self.database,
         )
 
+        if self._use_snowflake_session_token:
```
rudolfix: you do not need this flag at all. you just need to set all the correct values in `_set_use_snowflake_session_token`

rudolfix: you can keep the old code here

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -80,17 +122,29 @@ def to_connector_params(self) -> Dict[str, Any]:
...
+
         if self.application != "" and "application" not in conn_params:
             conn_params["application"] = self.application
 
         return conn_params
 
+    def _set_use_snowflake_session_token(self) -> None:
+        self.host = os.environ["SNOWFLAKE_ACCOUNT"]
+        self._snowflake_host = os.environ["SNOWFLAKE_HOST"]
+        self.authenticator = "oauth"
```
rudolfix: IMO we should ask user to set authenticator to `oauth` explicitly. and only then activate automated token discovery. We do that anyway: ie. if username and host are set we never discover this token. IMO this is confusing. We should tell user that if they set `oauth` and skip `token` OR `host` we'll take it from standard env variables

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -80,17 +122,29 @@ def to_connector_params(self) -> Dict[str, Any]:
...
             query,
-            user=self.username,
-            password=self.password,
             account=self.host,
             database=self.database,
         )
 
+        if self._use_snowflake_session_token:
+            assert conn_params["authenticator"] == "oauth"
+            conn_params["host"] = self._snowflake_host
```
rudolfix: this we can always set...

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -80,17 +122,29 @@ def to_connector_params(self) -> Dict[str, Any]:
...
         if self.application != "" and "application" not in conn_params:
             conn_params["application"] = self.application
 
         return conn_params
 
+    def _set_use_snowflake_session_token(self) -> None:
+        self.host = os.environ["SNOWFLAKE_ACCOUNT"]
+        self._snowflake_host = os.environ["SNOWFLAKE_HOST"]
+        self.authenticator = "oauth"
+        self._use_snowflake_session_token = True
```
rudolfix, line 147: just set `self.token` already

docs/website/docs/dlt-ecosystem/destinations/snowflake.md
```diff
@@ -132,6 +133,16 @@ or in the connection string as query parameters.
 
 In the case of external authentication, you need to find documentation for your OAuth provider. Refer to Snowflake [OAuth](https://docs.snowflake.com/en/user-guide/oauth-intro) for more details.
 
+**Snowflake-provided OAuth token authentication** is the recommended way to authenticate when running `dlt` in Snowpark Container Services. If both the host and username are **not** passed, `dlt` will look for the [Snowflake-provided OAuth token](https://docs.snowflake.com/en/developer-guide/snowpark-container-services/additional-considerations-services-jobs#connecting-with-a-snowflake-provided-oauth-token). Hence, to use this form of authentication, do **not** pass a host and username. For example:
+ ```toml
+[destination.snowflake.credentials]
+database = "dlt_data"
```
rudolfix, line 139: I'd ask for specific authenticator

docs/website/docs/dlt-ecosystem/destinations/snowflake.md
```diff
@@ -132,6 +133,16 @@ or in the connection string as query parameters.
...
+```
+or
+```toml
+destination.snowflake.credentials="snowflake:///dlt_data"
```
rudolfix: snowflake:///dlt_data?authenticator=oauth

rudolfix (changes requested):
I think we can simplify the code a lot and make using the "default credentials" less confusing for the user than it is now

dlt/destinations/impl/snowflake/configuration.py
```diff
@@ -10,17 +11,22 @@
...
+    snowflake_session_token_available,
+)
 
 
 SNOWFLAKE_APPLICATION_ID = "dltHub_dlt"
 
 
 @configspec(init=False)
-class SnowflakeCredentials(ConnectionStringCredentials):
+class SnowflakeCredentialsWithoutDefaults(ConnectionStringCredentials):
```
jorritsandbrink, line 24: I did not inherit from `CredentialsWithDefault` because its methods didn't seem useful here.

dlt/destinations/impl/snowflake/configuration.py
jorritsandbrink: I use the `_use_snowflake_session_token` flag to know if we should refresh the token. Is there a better way?

jorritsandbrink:
@rudolfix thanks for your review, I agree that asking the user to set `authenticator` to `oauth` is better.

I updated the code. Could you review again?

Issue #3316, opened 2025-11-13
feat: `dlt.Schema.to_mermaid()`


zilto:
Add a mechanism to export the `dlt.Schema` as a [Mermaid diagram](https://mermaid.js.org/syntax/entityRelationshipDiagram.html) string

<img width="1114" height="295" alt="Image" src="https://github.com/user-attachments/assets/1365a6ea-3f4f-48f1-b12b-5aee7f163012" />

## Motivation
Mermaid is a diagram format supported in many places:
- GitHub markdown (great for `README`, PR comments)
- marimo (dlt Dashboard, interactive notebook)
- IDE chat windows (e.g., Copilot, Cursor, Continue)

The key usecase is visualizing the `dlt.Schema` directly inside the AI chat window. This is not possible with the existing DBML and DOT exports.

## Implementation
The implementation should match the [.to_dot() (graphviz)](https://github.com/dlt-hub/dlt/pull/2959/files) and [.to_dbml()](https://github.com/dlt-hub/dlt/pull/2929) PRs.

- Create a module `dlt/helpers/mermaid.py`
- Build the diagram by manipulating strings without additional Python dependencies
- Write modular functions and tests for generating columns, tables, etc.
- Add `dlt.Schema.to_mermaid()` method
- Add to the CLI `dlt schema --format mermaid FILE` and `dlt pipeline NAME schema --format mermaid`
- Update CLI docs
- Add example export to the docs

Key consideration: Try to support converting the schema `dlt -> mermaid` AND `mermaid -> dlt` when possible. This makes testing easier. But it's likely impossible. For example, the DBML export doesn't support all possible column attributes

PR #3323, opened 2025-11-15
Feat/3316 add mermaid diagram export to schema

heaven00:
### Description
Adding capability to generate mermaid diagrams from schema, the representation in mermaid includes the following
- column names
- data type
- primary key or foreign key
- relationships ( first_entity, cardinality, second_entity, label)

### Related Issues
- https://github.com/dlt-hub/dlt/issues/3316

### Additional Context

Task List

- [x]  Create a module dlt/helpers/mermaid.py
- [x]  Build the diagram by manipulating strings without additional Python dependencies
- [x]  Write modular functions and tests for generating columns, tables, etc.
- [x]  Add dlt.Schema.to_mermaid() method
- [x]  Add to the CLI dlt schema --format mermaid FILE and dlt pipeline NAME schema --format mermaid
- [ ]  Update CLI docs
- [ ]  Add example export to the docs

### DIFF
--- a/tests/helpers/test_mermaid.py
+++ b/tests/helpers/test_mermaid.py
@@ -261,7 +261,9 @@ def test_schema_to_mermaid_generates_an_er_diagram(example_schema: dlt.Schema):
     assert mermaid_str.startswith("erDiagram")
 
 
-def test_schema_to_mermaid_generates_valid_mermaid_str_without_dlt_tables(example_schema: dlt.Schema):
+def test_schema_to_mermaid_generates_valid_mermaid_str_without_dlt_tables(
+    example_schema: dlt.Schema,
+):
     expected_mermaid_str = _expected_mermaid_str()
     mermaid_str = schema_to_mermaid(
         example_schema.to_dict(), example_schema.references, include_dlt_tables=False
@@ -402,7 +404,7 @@ def _expected_mermaid_str(with_dlt_tables: bool = False):
         purchases__items }|--|| purchases : _dlt_parent
         purchases__items }|--|| purchases : _dlt_root
     """
-    
+
 
 def _normalize_whitespace(text):
     """Normalize whitespace in a string by replacing all whitespace sequences with single spaces."""


heaven00:
@zilto I seem to have steered away from regular parameters a bit in the implementation.

- Do I need to work with `TStoredSchema` instead of the `Schema` object?
- Is it okay to include `table_names` as parameter or stick to show everything and just provide removing dlt private tables?

dlt/common/schema/schema.py
```diff
@@ -795,6 +795,24 @@ def to_dot(
...
+            table_names: Optional list of table names to include.  If an empty list
+                (the default) all tables are shown. This can be used to focus the
+                diagram on a subset of the schema.
+
+        Returns:
+            A string containing a Mermaid ERdiagram of the schema.
+        """
+        from dlt.helpers.mermaid import schema_to_mermaid
+
+        return schema_to_mermaid(self, table_names, include_dlt_tables)
```
zilto: This must be called on a `TTStoredSchema` object, i.e., a dictionary. You need to remove processing hints. See `.to_dbml()` and `.to_dot()` for reference.

Remove the `table_names()` filtering arg (detailed in other comment)

dlt/helpers/mermaid.py
```diff
@@ -0,0 +1,77 @@
+"""Build a mermaid graph representation using raw strings without additional dependencies"""
+from dlt.common.schema.schema import Schema
+from dlt.common.schema.typing import TTableReferenceStandalone, TTableSchema, TTableSchemaColumns
+
+
+def schema_to_mermaid(schema: Schema, table_names: list[str] = [], include_dlt_talbes: bool = True) -> str:
```
zilto: - Needs to be called on `TTStoredSchema` instead of `Schema`. (see comment on `dlt/common/schema/schema.py`

- typo in `include_dlt_talbes`

- I would arg `table_names` and the associated filtering logic. We want to keep these functions focused on rendering. Users can filter the content `TTStoredSchema` before passing the object to `schema_to_mermaid()`

dlt/helpers/mermaid.py
```diff
@@ -0,0 +1,77 @@
...
+        mermaid_str += table_schema
+
+    if references:
+        for ref in references:
+            ref_txt = _reference_to_text(ref)
+            mermaid_str += ref_txt
+    return mermaid_str
+
+
+def _relevant_references(table_names):
```
zilto: Add a docstring describing what a "relevant reference is"

dlt/helpers/mermaid.py
```diff
@@ -0,0 +1,77 @@
...
+    return lambda x: (x["table"] in table_names) & (x["referenced_table"] in table_names)
+
+
+def _table_to_text(table: TTableSchema) -> str:
+    items = [table.get("name", ""), "{", _columns_to_text(table.get("columns", {})), "}\n"]
+    return "".join(items)
+
+
+def _columns_to_text(columns: TTableSchemaColumns) -> str:
```
zilto: I believe it's worth having a `_to_mermaid_column()` function that generates a single column. Then we have:

- `_to_mermaid_column()`
- `_to_mermaid_table()`
- `_to_mermaid_reference()`
- `schema_to_mermaid()`

Then, add unit tests to this function to check if it produces the expected results for:
- no hint
- each individual hint
- 2+ hints

dlt/helpers/mermaid.py
```diff
@@ -0,0 +1,77 @@
...
+        if column_schema.get("primary_key"):
+            rows += f"{column_schema['data_type']} {column_name} PK \n"
+        elif column_schema.get("unique"):
+            rows += f"{column_schema['data_type']} {column_name} UK \n"
+        else:
+            rows += f"{column_schema['data_type']} {column_name} \n"
+    return rows
+
+
+_CARDINALITY_ARROW = {
```
zilto: Most usage of `TTableReference` don't have a cardinality set. You could add a `default` key (it's currently hardcoded in `_reference_to_text()`

Then, define a new `Literal` type `TMermaidArrows`

Then, add type hints to this constant 
```python
dict[Literal["default"] | TReferenceCardinality, TMermaidArrows]
```

zilto:
## Additional arg
I think it would be useful to have a top-level arg `hide_columns: bool` (available on `schema_to_mermaid()` and `dlt.Schema.to_mermaid()`. `if hide_columns is True` then we just have table names and references. This gives a much more compact visual when you have a lot of tables (this could be added in `.to_dot()` and `.to_dbml()`)

dlt/common/schema/schema.py
heaven00: the code is a lot simpler if we use `schema_to_mermaid(self.to_dict(), self.references...` by passing additional references because these are inferred and not always present in the `TTStoredSchema` ideally the fix would be to have it populated in the `TTStoredSchema` inside references.

It becomes easier to not replicate the references but it does go against the current interface. so against my own better judgement for now I am just going to redo the logic to find references within column names and references list (I may just be lacking some context)

zilto: - The key point is that `dlt.Schema` is a complex instantiated object and `TTStoredSchema` is a dictionary. `schema_to_mermaid` should use `TTstoredSchema` as input
- you shouldn't reimplement the logic of `dlt.Schema.references`. For context, this `.references` method was added after `.to_dot()` and `.to_dbml()` were created.

TL;DR feel free to write `schema_to_mermaid(..., references=self.references)`

heaven00:
hey @zilto almost done with the feedback you provided (thanks for that) 

one question, at the moment when we include all the dlt tables the _dlt_version renders separately

Screenshot 2025-11-21 at 08-10-00 Online FlowChart   Diagrams Editor - Mermaid Live Editor

though the version is populated by every table should it be any different?  things pending

- [ ] hide_columns feature
- [ ] cli integration and testing with the new parameters for pipeline and schema
- [ ] documentation update

zilto:
closing in favor of #3364

PR #3325, opened 2025-11-16
feat/3324-adding-iceberg-catalogs-compatibilities

### LINKED ISSUE #3324: Iceberg destination only supports in-memory SQLite catalog - blocks REST and production catalogs
### Feature description

I would like to add support for other Iceberg catalogs, rather than just the SQLite, to the OSS version of DLT. I believe this is not a complicated addition, as we can leverage the existing `pyiceberg` functionalities to load catalog configs from the `.pyiceberg.yaml` and `load_catalog`. This allows the rest of the implementation to continue as normal. 

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

Yes! So far DLT is awesome, but the integration with Iceberg catalogs is not full (understandably as this is offered in DLT+). But while I understand that in DLT+ we may have much better features on top of this, I think the option to allow to connect to more catalogs for the OSS is a key feature to allow DLT to be a top-tier ingestion framework (which already is, but this would be like top tier ++++++ 😄  )

### Proposed solution

I am preparing a PR leveraging the `load_catalog` function in pyiceberg. All the rest largely remains the same, we just leverage `load_catalog` to find the catalog, and return it to the pipeline as normal. I added other elements such as constructing the catalog config when required, but in essence, this is the focus. 

Right now I am using this approach by monkey-patching DLT before running the pipeline. I am currently running tests and making some corrections to the pr. 

### Related issues

_No response_

lfagliano:
### Description

This PR removes the hardcoded in-memory SQLite catalog limitation and enables DLT to work with all PyIceberg-supported catalog types including REST catalogs, through leveraging the `load_catalog` functionality. The actual execution and operation of Iceberg format remains the same, we just load the catalog and pass them along to the pipeline.  

Previously, the Iceberg destination hardcoded catalog creation to use `sqlite:///:memory:`, limiting Iceberg support:

```python
catalog = get_sql_catalog(
    catalog_name or "default", 
    "sqlite:///:memory:",  # No other options!
    self.config.credentials
)
```

Leverage PyIceberg's built-in `load_catalog()` function to support all catalog types through standard configuration methods. The implementation provides a priority chain that tries multiple configuration sources and falls back gracefully.

In essence, this allows the user to add a `.pyiceberg.yaml` file to reference to their catalogs (which is in line with pyiceberg), similar to this: 
```yaml
# ~/.pyiceberg.yaml
catalog:
  production:
    type: rest
    uri: https://localhost:8181/catalog
    warehouse: analytics
```

The changes I made do not require the user to do much more than that, and they can continue using Iceberg as they were. 

### Related Issues

- Fixes #3324 

### Additional Context

To test it I run this for some time, monkey patching the core package. I then also added my contributions, built the package and tried it again through two of my prod pipelines. 

As for the tests, I run them locally, but I came across some what I think are broken tests -- because they had absolutely nothing to do with my changes, but I am gonna be checking them more in detail tomorrow.

### DIFF
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -138,7 +138,7 @@ jobs:
         if: matrix.python-version != '3.14'
 
       - name: Install pipeline and sources dependencies
-        run: uv sync ${{ matrix.uv_sync_args }} --extra http --extra duckdb --extra cli --extra parquet --extra deltalake --extra sql_database --group sentry-sdk --group pipeline --group sources --group ibis
+        run: uv sync ${{ matrix.uv_sync_args }} --extra http --extra duckdb --extra cli --extra parquet --extra deltalake --extra pyiceberg --extra sql_database --group sentry-sdk --group pipeline --group sources --group ibis
         if: matrix.python-version != '3.14'
 
       - name: Run extract and pipeline tests


dlt/common/libs/pyiceberg.py
```diff
@@ -142,6 +144,273 @@ def get_sql_catalog(
...
+    from pyiceberg.catalog import load_catalog
+
+    search_paths = []
+    if config_path:
+        search_paths.append(Path(config_path))
+    else:
+        # Search in standard locations
+        search_paths.extend(
+            [
+                Path.cwd() / ".pyiceberg.yaml",
```
rudolfix: please take a look at `dlt.current.run_context()` and use
```
def run_dir(self) -> str:
        """Defines the context working directory, defaults to cwd()"""
```

dlt/common/libs/pyiceberg.py
```diff
@@ -142,6 +144,273 @@ def get_sql_catalog(
...
+                Path.cwd() / ".dlt" / ".pyiceberg.yaml",
```
rudolfix: `in run_context` use `def get_setting(self, setting_path: str) -> str:` with `.pyiceberg.yaml` as argument to get stuff from `.dlt`
```py
def get_setting(self, setting_path: str) -> str:
        """Gets path in settings_dir where setting (ie. `secrets.toml`) are stored"""
```

overall dlt workspace structure is abstracted from physical location via this class

dlt/common/libs/pyiceberg.py
```diff
@@ -142,6 +144,273 @@ def get_sql_catalog(
...
+                Path.home() / ".pyiceberg.yaml",
```
rudolfix: not every system has home dir and even users. check the documentation and maybe use exception handler

dlt/common/libs/pyiceberg.py
```diff
@@ -142,6 +144,273 @@ def get_sql_catalog(
...
+    config_dict = config_dict.copy()
+    if credentials:
+        fileio_config = _get_fileio_config(credentials)
+        config_dict.update(fileio_config)
+
+    return load_catalog(catalog_name, **config_dict)
+
+
+def load_catalog_from_env(
+    catalog_name: Optional[str] = None,
```
rudolfix: this looks like overkill. `pyiceberg` will take settings from env variables it knows AFAIK and dlt probes env variables automatically. no need to do that IMO

dlt/common/libs/pyiceberg.py
```diff
@@ -142,6 +144,273 @@ def get_sql_catalog(
...
+
+    # Merge filesystem credentials if provided
+    if credentials:
+        fileio_config = _get_fileio_config(credentials)
+        config.update(fileio_config)
+
+    return load_catalog(catalog_name, **config)
+
+
+def get_catalog(
```
rudolfix, line 327: here's the biggest change: do not modify `filesystem` configuration. decorate this function with `@with_config` and it will synthesize dlt configuration for you. you can also use specific `spec` argument and convert the code below into a configspec (https://dlthub.com/docs/devel/general-usage/credentials/advanced#write-custom-specs)
```py
    # Iceberg catalog configuration
    iceberg_catalog_name: Optional[str] = "default"
    """Name of the Iceberg catalog to use. Corresponds to catalog name in .pyiceberg.yaml"""

    iceberg_catalog_type: Optional[str] = "sql"
    """Type of Iceberg catalog: 'sql', 'rest', 'glue', 'hive', etc."""

    iceberg_catalog_uri: Optional[str] = None
    """
    URI for SQL catalog (e.g., 'postgresql://...') or REST catalog endpoint.
    If not provided, defaults to in-memory SQLite for backward compatibility.
    """
    iceberg_catalog_config: Optional[Dict[str, Any]] = None
```

for example:
```py
@with_config(section="iceberg_catalog")
def get_catalog(
   iceberg_catalog_name: str = "default"
   ...
```
will take configuration from
```toml
[iceberg_catalog]
iceberg_catalog_name="my_catalog"
...
```
or from corresponding env variables

dlt/common/libs/pyiceberg.py
```diff
@@ -142,6 +144,273 @@ def get_sql_catalog(
...
+
+    This function tries to load a catalog in the following priority order:
+    1. From explicit config dictionary (if catalog_config provided)
+    2. From .pyiceberg.yaml file
+    3. From environment variables
+    4. Fall back to in-memory SQLite catalog
+
+    Args:
+        catalog_name: Name of the catalog (default: "default")
+        catalog_type: Type of catalog ('sql' or 'rest') - used for fallback
```
rudolfix: this is not passed and IMO should

dlt/common/libs/pyiceberg.py
```diff
@@ -142,6 +144,273 @@ def get_sql_catalog(
...
+        config.update(fileio_config)
+
+    return load_catalog(catalog_name, **config)
+
+
+def get_catalog(
+    catalog_name: str = "default",
+    catalog_uri: Optional[str] = None,
+    catalog_config: Optional[Dict[str, Any]] = None,
+    credentials: Optional[FileSystemCredentials] = None,
```
rudolfix, line 331: do you intend to configure via toml or this is always passed explicitly?

dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -26,6 +26,38 @@ class FilesystemDestinationClientConfiguration(FilesystemConfigurationWithLocalF
     always_refresh_views: bool = False
     """Always refresh table scanner views by setting the newest table metadata or globbing table files"""
 
+    # Iceberg catalog configuration
+    iceberg_catalog_name: Optional[str] = "default"
```
rudolfix: please do not modify this configuration

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -1000,16 +1000,47 @@ def get_open_table_catalog(self, table_format: TTableFormat, catalog_name: str =
...
-        # create namespace
-        catalog.create_namespace(self.dataset_name)
+        # Use catalog name from config or parameter
+        catalog_name = catalog_name or self.config.iceberg_catalog_name or "default"
+
+        logger.info(f"Loading Iceberg catalog: {catalog_name}")
+
+        try:
+            # Try to load catalog using new function
+            catalog = self._catalog = get_catalog(
```
rudolfix: here's magic: just call get_catalog(credentials=self.config.credentials) and dlt will inject missing args from config

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -1000,16 +1000,47 @@ def get_open_table_catalog(self, table_format: TTableFormat, catalog_name: str =
...
+
+            catalog = self._catalog = get_sql_catalog(
+                catalog_name, "sqlite:///:memory:", self.config.credentials
+            )
+
+        # Create namespace
+        try:
+            catalog.create_namespace(self.dataset_name)
+            logger.info(f"Created Iceberg namespace: {self.dataset_name}")
+        except Exception as e:
```
rudolfix: please catch more specific exception (is there exception for entity to already exist?)

lfagliano:
Heyo @rudolfix! Thanks for the review! Yes I think I can do those! Not quite sure about timelines or ETAs, but I will give it a shot as fast as I can 💪🏻

dlt/common/libs/pyiceberg.py
lfagliano: Done!

dlt/common/libs/pyiceberg.py
lfagliano: Done! Very much loved this setup

dlt/common/libs/pyiceberg.py
lfagliano: I removed home for now

dlt/common/libs/pyiceberg.py
lfagliano: Big change here, I opted to remove _env and consolidate it under the _yaml which is now called _pyiceberg. 

Both yaml and env are automatically managed by pyiceberg so I removed the abstraction. I did made a thin wrapper around the load_catalog to determine if the envs or yaml where present, otherwise rise the CatalogNotFound exception.

dlt/common/libs/pyiceberg.py
lfagliano: Thanks! Applied this, such a breeze

dlt/common/libs/pyiceberg.py
lfagliano: Done!

dlt/common/libs/pyiceberg.py
lfagliano: So I opted to leave the credentials here to keep with backwards compatibility (the epehemerous catalog). However, I don't expect to use the `destination.filesystem` to feed the storage credentials. 

Instead, I am expecting to see them in the `iceberg_section` **when** needed. Normally you would get vended credentials, which don't require declaring the storage credentials, and in some setups you would use remote signing, in those I do expect to see it in the `secrets.toml` under the `[iceberg_catalog.iceberg_catalog_config]`

(All of this will be in the docs)

lfagliano: So I opted to leave the credentials here to keep with backwards compatibility (the epehemerous catalog). However, I don't expect to use the `destination.filesystem` to feed the storage credentials. 

Instead, I am expecting to see them in the `iceberg_section` **when** needed. Normally you would get vended credentials, which don't require declaring the storage credentials, and in some setups you would use remote signing, in those I do expect to see it in the `secrets.toml` under the `[iceberg_catalog.iceberg_catalog_config]`

(All of this will be in the docs)

dlt/destinations/impl/filesystem/filesystem.py
lfagliano: Done!

dlt/destinations/impl/filesystem/filesystem.py
lfagliano: Done!

dlt/destinations/impl/filesystem/configuration.py
lfagliano: Done!

lfagliano:
Hi @rudolfix ! I am on holidays this week, but I am almost done. Next week I will still be working on some tests and documentation.

But for now, I made a number of changes to the PR and addressed your comments. Let me know if you have more!

I do have a question. Right now, I did integration tests against SQLLite and the Rest Catalog. Taking inspiration from `pyiceberg`test [setup](https://github.com/apache/iceberg-python/blob/main/dev%2Fdocker-compose-integration.yml for the docker file), they basically spin a docker compose during CI to test the REST catalog. I am planning on doing the same for Rest and Postgres. Would you agree with this approach? (I am not sure if this doesn't need to be in a different PR to make the relevant CI changes)

lfagliano:
Added the documentation as well! I am now only waiting for continuing with the test (first I want to know if what I proposed is a correct approach for how the project wants to do testing)

dlt/common/libs/pyiceberg.py
```diff
@@ -1,12 +1,18 @@
+from pydantic import BaseModel, ConfigDict, Field
+from pyiceberg.catalog import load_catalog
```
rudolfix: please move this import to other pyiceberg imports below (we have a mechanism for friendly dependency exceptions)
pydantic is an optional dependency. please import it after iceberg (also down below)

dlt/common/libs/pyiceberg.py
```diff
@@ -1,12 +1,18 @@
...
+from dlt.common.configuration.inject import with_config
```
rudolfix: pls place it with other dlt imports. (we'll add isort soon)

docs/website/docs/dlt-ecosystem/destinations/iceberg.md
```diff
@@ -51,6 +40,61 @@ When using Iceberg with object stores like S3, additional permissions may be req
...
+In the back, dlt utilizes `load_catalog`, so we try to keep the config we will pass along, as close as possible to the `pyiceberg` one. Specifically, we want to provide a catalog name and catalog type (`rest` or `sql`), and we provide them under the `iceberg_catalog` section of the `secrets.toml` or the env vars. These two variables will be used either to detect the catalog we want to load (for example, you have a `.pyiceberg.yaml` with multiple catalogs), and to do some validations when required. 
+
+```toml
+[iceberg_catalog]
+iceberg_catalog_name = "default"
+iceberg_catalog_type = "rest"
+```
+or 
+```sh
+export DLT_ICEBERG_CATALOG_NAME=default
```
rudolfix: I think it is `ICEBERG_CATALOG_NAME` or `ICEBERG_CATALOG__ICEBERG_CATALOG_NAME`

dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -1,3 +1,4 @@
+from pyiceberg.exceptions import NamespaceAlreadyExistsError
```
rudolfix: best if you import it on demand - in the function that raises this exception (already imports other pyiceberg things)

tests/common/libs/pytest.ini
```diff
@@ -0,0 +1,27 @@
+[pytest]
```
rudolfix: is this really needed? if you need to change our pytest config then let's discuss it (but I do not think you need)

lfagliano: removed! this was a slip from my local dev setup

lfagliano:
@rudolfix  thanks for the review! I addressed your comments, let me know if you would like me to change anything else. 

About the tests, only today I saw your comment and I had already started over the weekend with the integration tests trying to do a similar setup to `pyiceberg`, and I was about to commit them. I tried to re-work them a bit. Basically I am adding a `docker-compose` file that can be used to set up the whole testing environment with the two catalogs (rest & postgres), as well as a `init-iceberg-catalogs.py` which sets up the catalogs for the tests. Originally I thought that would be what we would put in CI, but for now I added `skip` to the REST and Postgres fixtures, so that they are essentially optional. I have been using them for the last couple of days and found them useful. Let me know what you think!

rudolfix:
@lfagliano what I meant is that we are not ready to build persisted infra with Lakekeeper and s3 for testing. I see that you already implemented local test setup. my take would be:

* if you have your integration tests working we can merge them but skipped. we'll need to add a separate workflow after [#perallel](https://github.com/dlt-hub/dlt/pull/3479) is merged (or add them to common, not sure yet)
* if the integration tests are not ready we can just merge what we have right now

overall this looks pretty good!

lfagliano:
@rudolfix I was referring to the current integration tests in the PR, which they should already be skipped if they dont detect any catalog present (btw, I moved away from lakekeeper for the tests, instead I am using the apache testing [image](https://hub.docker.com/r/apache/iceberg-rest-fixture) )

But then, I think I'm happy with moving forward with what we have already :D 🚀 Is there anything else you need me to do? I just corrected the lint errors

rudolfix:
@lfagliano there were still a few linter errors. I fixed them so we can iterate quicker. Are you make to run `make format` and `make lint` in your env? those should catch problems before pushing

lfagliano:
@rudolfix damn, sorry that was my bad. I did `make lint` and `make format` when I opened the PR and forgot to do it now, i just did the mypy run from the actions. Apologies. ~I also see now that there are problems with the imports, I am a bit pressed on time, but I can do them on the morning.~ (Edit: I see there is no problem now.. thanks 🙏🏼 )

PR #3329, opened 2025-11-17
Docs: Lifecycle of a dlt transformation (a sql query model of a transformation relation)

### LINKED ISSUE #3206: feat(relation): clarify `dlt.Relation` lifecycle
### Summary
Currently, `dlt.Relation` is coupled to `pipeline.run()` in a way that those two are not equivalent

```python
@dlt.transformation
def foo(dataset: dlt.Dataset):
  yield dataset.table("bar")

# retrieve relation from transformation then execute and get data
result_rel: dlt.Relation = list(foo(dataset))[0]
result_rel.df()

# execute transformation via pipeline.run() then get data
pipeline.run([foo])
result2_rel = pipeline.dataset().table("foo")
result2_rel.df()
```
This behavior is inconsistent with `@dlt.resource` where doing `list(resource)` will gather records in-memory and add the `_dlt_id` and `_dlt_load_id` fields.

This happens because `dlt.Pipeline.run()` goes through Extract, Normalize, Load. The Normalize step adds the `_dlt_id` and `_dlt_load_id` columns.

### Motivation
We want to be able to define `@dlt.transformation` interactively without requiring the heavy `dlt.Pipeline.run()` constructs on each execution. 

The column `_dlt_id` exists on all tables produced by dlt by default and is useful to uniquely identify records. This is useful for transformation when:
- identifying rows for data quality checks
- incremental strategies
- smart joins

### Intended behavior
1.  `list(transformation)[0]` should include the `_dlt_id` and `_dlt_load_id` just like `list(resource)` does
2.  Currently, transformation outputs are all treated like root tables; they have their own `_dlt_id` and `_dlt_load_id` with no parent.
  - This could change in the future (e.g., we generate dimensions and facts), but is good for now
3. Executing `dlt.Relation` via `.df()`, `.fetchall()`, etc. shouldn't add `_dlt_id` nor `_dlt_load_id`. Relations are a more generic mechanism than transformations
4. Methods and features related to `dlt.Dataset` shouldn't mutate queries to include / generate `_dlt_id` nor `_dlt_load_id` 

### Potential solution
The internals of `@dlt.transformation` (e.g., `make_transformation_function`)  produce a `dlt.Relation`. The code should be changed such that `dlt.Relation` generates the `_dlt_id` and `_dlt_load_id` instead of this being done in the Normalization of `pipeline.run()`

question: can I have a list of things that happen a Normalization for `@dlt.transformation`?

anuunchin:
This PR explains what exactly happens in the extract, normalize and load stages when a transformation is run through a pipeline.

Resolves #3206

A lifecycle test is added to show what happens to the model in the pipeline.run().

### DIFF
--- a/docs/uv.lock
+++ b/docs/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
 requires-python = ">=3.9.2, <3.14"
 resolution-markers = [
     "python_full_version >= '3.13' and platform_python_implementation != 'PyPy' and sys_platform != 'emscripten'",
@@ -1169,7 +1169,7 @@ requires-dist = [
     { name = "cron-descriptor", marker = "extra == 'cli'", specifier = ">=1.2.32" },
     { name = "databricks-sdk", marker = "extra == 'databricks'", specifier = ">=0.38.0" },
     { name = "databricks-sql-connector", marker = "python_full_version >= '3.13' and extra == 'databricks'", specifier = ">=3.6.0" },
-    { name = "databricks-sql-connector", marker = "python_full_version < '3.13' and extra == 'databricks'", specifier = ">=2.9.3" },
+    { name = "databricks-sql-connector", marker = "python_full_version < '3.13' and extra == 'databricks'", specifier = ">=2.9.3,<4" },
     { name = "db-dtypes", marker = "extra == 'bigquery'", specifier = ">=1.2.0" },
     { name = "db-dtypes", marker = "extra == 'gcp'", specifier = ">=1.2.0" },
     { name = "deltalake", marker = "extra == 'deltalake'", specifier = ">=0.25.1" },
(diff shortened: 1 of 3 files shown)


zilto (commented):
The documentation improvements were helpful. OTOH, some tests are failing, but it seems unrelated to code changes?

docs/website/docs/hub/features/transformations/transformation-snippets.py
```diff
@@ -14,6 +14,7 @@ def fruitshop_pipeline() -> dlt.Pipeline:
 
     # @@@DLT_SNIPPET_START quick_start_example
```
anuunchin, line 16: edited these because the docs say "you can copy paste", but imports still needed to be added 👀

docs/uv.lock
zilto, line 1: should this be there? I don't see why we should modify the `databricks-sql-connector` dependency for this PR. If this was also introduced in another PR, please `git rebase`

anuunchin: agh, me always with this lock file... removed it 🫡

Issue #1234, opened 2024-04-17
support bulk copy for mssql and synapse


rudolfix:
**Background**
The way we insert rows into mssql is ineffective. We should switch to bulk copy. MS odbc drivers come with `bcp` command that we can use.
https://github.com/yehoshuadimarsky/bcpandas/blob/master/bcpandas/utils.py
is a project that does that quite well

**Tasks**
1. * [ ] allow `mssql` and `synapse` to handle `csv`
2. * [ ] use `bcp` in copy jobs

jorritsandbrink:
@rudolfix Do we also want to do this for `synapse`? For `synapse` we already implement `SynapseCopyFileLoadJob` based on `COPY INTO`, which is the recommended approach:

_"While dedicated SQL pools support many loading methods, including popular SQL Server options such as [bcp](https://learn.microsoft.com/en-us/sql/tools/bcp-utility?toc=/azure/synapse-analytics/sql-data-warehouse/toc.json&view=azure-sqldw-latest&preserve-view=true) and the [SqlBulkCopy API](https://learn.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlbulkcopy?toc=/azure/synapse-analytics/sql-data-warehouse/toc.json), the fastest and most scalable way to load data is through PolyBase external tables and the [COPY statement](https://learn.microsoft.com/en-us/sql/t-sql/statements/copy-into-transact-sql?toc=/azure/synapse-analytics/sql-data-warehouse/toc.json&view=azure-sqldw-latest&preserve-view=true)._

_With PolyBase and the COPY statement, you can access external data stored in Azure Blob storage or Azure Data Lake Store via the T-SQL language. For the most flexibility when loading, we recommend using the COPY statement._"

Source: https://learn.microsoft.com/en-us/azure/synapse-analytics/sql-data-warehouse/design-elt-data-loading

rudolfix:
@jorritsandbrink it just may work for synapse if we implement it ie. via csv files + bcp but it is not required. The reason for the ticket is abysmal insert performance of sql server. it is WAY slower than postgres. nevertheless priority of this ticket is quit low (at least for now)

cody-scott:
Chiming in here; one of the big hold back for shifting to dlt is the abysmal odbc performance over bcp. 

Some of our sources are tall and wide so bcp helps a ton with performance, so if this comes to be then I would be thrilled!

[...]

PR #3333, opened 2025-11-18
ingests parquet into mssql, mysql and sqlite via ADBC

rudolfix:
### Description
Allows to directly ingest parquet into mssql via ADBC. Requires `dbc` command from Columnar to install mssql driver. Ingestion is significantly faster compared to INSERT we use now (surely more than 10x, benchmarks not yet completed).

ADBC was also enabled for sqlite and mysql in sqlalchemy destination

improves https://github.com/dlt-hub/dlt/issues/1234

while implementing I discovered that `@with_config` decorator does not work on modules that store hints as strings. that got fixed and properly tested.

### DIFF
--- a/dlt/normalize/items_normalizers.py
+++ b/dlt/normalize/items_normalizers.py
@@ -30,6 +30,7 @@ from dlt.common.schema.utils import (
     dlt_id_column,
     dlt_load_id_column,
     has_table_seen_data,
+    is_complete_column,
     normalize_table_identifiers,
     is_nested_table,
     has_seen_null_first_hint,
@@ -555,7 +556,13 @@ class JsonLItemsNormalizer(ItemsNormalizer):
                     )
                     parent_col_schema = parent_col_schemas.get(last_ident_path)
 
-                    if parent_col_schema and has_seen_null_first_hint(parent_col_schema):
+                    # remove only incomplete columns: both None, simple and complex values may be received by a column
+                    # in any order
+                    if (
+                        parent_col_schema
+                        and has_seen_null_first_hint(parent_col_schema)
+                        and not is_complete_column(parent_col_schema)
+                    ):
                         parent_col_schemas.pop(last_ident_path)
                         parent_updates = schema_update.get(parent_name, [])
                         for j, parent_update in enumerate(parent_updates):
@@ -579,7 +586,6 @@ class JsonLItemsNormalizer(ItemsNormalizer):
         updated_table_partial: TPartialTableSchema = None
         table = self.schema._schema_tables.get(table_name)
         if not table:
-            # print("NEW TABLE", table_name)
             table = utils.new_table(table_name, parent_table)
         table_columns = table["columns"]
 
--- a/tests/normalize/test_json_item_inference.py
+++ b/tests/normalize/test_json_item_inference.py
@@ -705,8 +705,11 @@ def test_coerce_null_value_in_nested_table(
 @pytest.mark.parametrize(
     "use_very_long_col_name", [True, False], ids=["very_long_col_name", "short_col_name"]
 )
+@pytest.mark.parametrize(
+    "with_simple_value", [True, False], ids=["with_simple_value_after", "no_simple_value_after"]
+)
 def test_coerce_null_value_as_compound_columns(
-    item_normalizer: JsonLItemsNormalizer, use_very_long_col_name: bool
+    item_normalizer: JsonLItemsNormalizer, use_very_long_col_name: bool, with_simple_value: bool
 ) -> None:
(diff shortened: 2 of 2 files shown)


docs/website/docs/dlt-ecosystem/destinations/mssql.md
```diff
@@ -112,6 +112,48 @@ If you set the [`replace` strategy](../../general-usage/full-loading.md) to `sta
...
+uv tool run dbc search
+```
+`dlt` will make **parquet** the preferred file format once driver is detected at runtime. This method is 10x-70x faster than INSERT and
+we make it a default for all input data types.
+
+Not all arrow data types are supported by the driver, see driver docs for more details:
+* fixed length binary
+* time with precision different than microseconds
+
+We copy parquet files with batches of size of 1 row group. All groups are copied in a single transaction.
```
djudjuu, line 141: not clear what size you mean here

docs/website/docs/dlt-ecosystem/destinations/sqlalchemy.md
```diff
@@ -231,6 +231,36 @@ The following write dispositions are supported:
...
+with `uv` you can run `dbc` directly:
+```sh
+uv tool run dbc install sqlite
+```
+Note that **we do not detect sqllite** driver [installed via Python package](https://arrow.apache.org/adbc/current/driver/sqlite.html)
+
+You must set have correct driver installed and `loader_file_format` set to `parquet` in order to use ADBC. If driver is not found,
+`dlt` will convert parquet into INSERT statements.
+
+Not all arrow data types are supported by the **sqlite**:
```
djudjuu: Better:
```suggestion
Note: The following arrow data types are NOT supported by the **sqlite**:
```

dlt/destinations/impl/mssql/mssql.py
```diff
@@ -76,6 +82,69 @@ def _new_temp_table_name(cls, table_name: str, op: str, sql_client: SqlClientBas
...
+    if TYPE_CHECKING:
+        from adbc_driver_manager.dbapi import Connection
+
+    def _connect(self) -> "Connection":
+        from adbc_driver_manager import dbapi
+
+        self._config = self._job_client.config  # type: ignore[assignment]
+        conn_dsn = self.odbc_to_go_mssql_dsn(self._config.credentials.get_odbc_dsn_dict())
+        conn_str = ";".join([f"{k}={v}" for k, v in conn_dsn.items()])
+        logger.info(f"ADBC connect to {conn_str}")
```
djudjuu: question: are we logging credentials and/or secrets here?

dlt/destinations/impl/sqlalchemy/load_jobs.py
```diff
@@ -74,6 +78,93 @@ def run(self) -> None:
...
+                elif lk == "ssl_key":
+                    mapped["tls-key"] = v
+                elif lk == "ssl_mode":
+                    mapped["tls"] = v
+                else:
+                    mapped[k] = v
+
+            username = url.username or ""
+            password = url.password or ""
+            auth = f"{username}:{password}@" if username or password else ""
```
djudjuu: question: do we need to encode special chars differently if passes as uri?

djudjuu:
## on adbc

- only happy path tests for data formats (decimal, binary etc) but I confirmed manually that error messages are helpful
- the mechanism and docs to enable parquet loading are clear. install driver and set `loader_file_format`, docs are clear too :heavy_check_mark: 

ok, few questions mostly centered around the connection establishing (which is different for the three destinations)
- is is possible we log a secret when we log connection str?
- does the connection string conversion fail gracefully? (its not tested) 

## on with_config
:brain: :exploding_head: 

basically this makes sure that the resolver always has (the right) global context when its resolving type-hints. espeially type hints that are forwardRefs. So during module compilation ForwardRefs are created...; resolution is later, when code like dlt’s decorators or `resolve_configuration` are executed

dlt/destinations/impl/mssql/mssql.py
rudolfix: we absolutely are 🤯 thanks for spotting this

dlt/destinations/impl/sqlalchemy/load_jobs.py
djudjuu:
```
from urllib.parse import quote
...
-            username = url.username or ""
-            password = url.password or ""
-            auth = f"{username}:{password}@" if username or password else ""
+            username = quote(url.username or "", safe="")
+            password = quote(url.password or "", safe="")
+            auth = f"{username}:{password}@" if username or
```

PR #3338, opened 2025-11-19
Fix DocSearch v4 styles

burnash:


burnash:
@zilto I believe the upgrade of docusaurus in #3247 broke custom styling of docs.

PR #3354, opened 2025-11-20
(fix) 3346 fix trace loading

### LINKED ISSUE #3342: bug: backwards incompatible change `LoadJobMetrics`
dlt version: 1.18.2

##  problem

#3248 introduced a new field to the `NamedTuple` `LoadJobMetrics`. This breaks backwards compatibility with previous `trace` object. 

For instance, when using the dlt Dashboard, unpickling the stored object will crash the dashboard completely. (e.g., `LoadJobMetrics` pickled doesn't match the current dlt version)

https://github.com/user-attachments/assets/69eb132b-0a5f-4ec6-a150-8cb312bf540e

## Potential solution
Add a `__getstate__()` method to `LoadJobMetrics`. This method is called by the pickling module. It will be able to set a `None` value to `LoadJobMetrics` if it's not retrieved from the unpickled dictionary. Include a clear docstring

## Future work
Identify all places where `pickle` is used to create lasting artifacts. This should be avoided and replaced wherever possible.

rudolfix:
### Description
fixes #3342 

it also fixes a case when single step traces were saved instead of full merged trace. that produced inconsistent behavior in `pipeline.last_trace`

### DIFF
--- a/tests/pipeline/test_pipeline_trace.py
+++ b/tests/pipeline/test_pipeline_trace.py
@@ -9,11 +9,9 @@ import pytest
 import requests_mock
 import yaml
 
-import pendulum
-
 import dlt
 
-from dlt.common import json
+from dlt.common import json, pendulum
 from dlt.common.configuration.specs import CredentialsConfiguration, RuntimeConfiguration
 from dlt.common.configuration.specs.config_providers_context import ConfigProvidersContainer
 from dlt.common.configuration.utils import get_resolved_traces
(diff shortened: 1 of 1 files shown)


tests/pipeline/test_pipeline_trace.py
```diff
@@ -9,6 +9,8 @@
 import requests_mock
 import yaml
 
+import pendulum
```
zilto: I believe this should be imported from `dlt.common.pendulum`. This constraint is enforced in `dlt/` and not `tests/` because we sometimes need to test pendulum directly.

tests/pipeline/test_pipeline_trace.py
```diff
@@ -225,13 +231,18 @@ def data():
...
     assert_trace_serializable(trace)
-    assert isinstance(p.last_trace.last_load_info, LoadInfo)
-    p.activate()
+    assert isinstance(pipeline.last_trace.last_load_info, LoadInfo)
+    pipeline.activate()
+
+    # copy trace to well known location so it can be used as fixture
+    trace_dir = os.path.join(TEST_STORAGE_ROOT, f"dlt.{dlt.__version__}")
+    os.makedirs(trace_dir)
+    save_trace(trace_dir, pipeline._last_trace)
```
zilto: Is this temporary code or it's part of the test logic?

zilto (dismissed):
Minor questions / tweaks. Otherwise, looks good. The backwards compatibility test is a good idea

tests/pipeline/test_pipeline_trace.py
rudolfix: this is not a part of test logic. but I do not want to remove that. I use this test to generate a fixture for backward compat test. I'll add a clear comments here.

PR #3356, opened 2025-11-20
(fix) use sparse checkout for dlt init dlthub

### LINKED ISSUE #3349: use sparse checkout for llm context
**Background**
We distribute LLM Context (rules/specs for generating data sources) as repo - same thing we do with verified sources. Here the repo has grown to thousands of folders but in a normal use case we need just one. We can use sparse checkout before we fix this problem systematically.

**Tasks**
Fix `vibe_source_setup` (we'll rename this function later) so it is passing `source` to `get_fresh_repo_files`. Source is folder name that must be checked out.

example cli:
```sh
git clone --depth=1 --filter=blob:none --no-checkout https://github.com/dlt-hub/vibe-hub.git
git sparse-checkout init --cone
```

then for every source 
```sh
git sparse-checkout set abbyy
```
where abbyy is source name.

Tests:
1. `get_fresh_repo_files` test the clone (do we see expected files and nothing else) and then with other source name for already cloned repo
2. check for unknow source name
3. existing tests must pass

rudolfix:
### Description
fixes #3349

### DIFF
--- a/tests/libs/test_git.py
+++ b/tests/libs/test_git.py
@@ -130,13 +130,14 @@ def test_sparse_checkout(test_storage: FileStorage) -> None:
     assert repo_storage.has_folder("stripe")
     assert len(repo_storage.list_folder_dirs(".")) == 2  # .git stripe
 
-    # unknown path
+    # unknown path in case repo is already cloned and checkout was done
     repo_storage = get_fresh_repo_files(CONTEXT_REPO, test_storage.storage_path, path="__unknown")
     assert not repo_storage.has_folder("__unknown")
     assert len(repo_storage.list_folder_dirs(".")) == 1  # .git
 
 
-def test_sparse_checkout_path_not_exist(test_storage: FileStorage) -> None:
+def test_sparse_checkout_path_not_exist_on_clone(test_storage: FileStorage) -> None:
+    # unknown path before first checkout
     repo_storage = get_fresh_repo_files(CONTEXT_REPO, test_storage.storage_path, path="__unknown")
     assert not repo_storage.has_folder("__unknown")
     assert len(repo_storage.list_folder_dirs(".")) == 1  # .git


dlt/common/libs/git.py
```diff
@@ -70,14 +70,37 @@ def get_default_branch(repo: Repo) -> str:
...
     # update remotes and check if heads are same. ignores locally modified files
     with Repo(repo_path) as repo:
         # use custom environment if specified
         with repo.git.custom_environment(GIT_SSH_COMMAND=with_git_command):
+            # if path is set, use sparse checkout
+            if path is not None:
+                # assume that sparse checkout was enabled when cloning
+                repo.git.sparse_checkout("set", path)
             # checkout branch before fetching
             repo.git.checkout(branch or get_default_branch(repo))
```
zilto, line 105: I believe you need an `if / else` here? Otherwise, you're moving into sparse checkout than back to regular checkout?

tests/libs/test_git.py
```diff
@@ -116,6 +117,31 @@ def test_fresh_repo_files_branch_change(test_storage: FileStorage) -> None:
...
+
+    # checkout the other one
+    repo_storage = get_fresh_repo_files(CONTEXT_REPO, test_storage.storage_path, path="stripe")
+    assert repo_storage.has_folder("stripe")
+    assert len(repo_storage.list_folder_dirs(".")) == 2  # .git stripe
+
+    # unknown path
+    repo_storage = get_fresh_repo_files(CONTEXT_REPO, test_storage.storage_path, path="__unknown")
+    assert not repo_storage.has_folder("__unknown")
+    assert len(repo_storage.list_folder_dirs(".")) == 1  # .git
```
zilto, line 136: to remove, this is a duplicate of the next test `test_sparse_checkout_path_not_exist()`

dlt/common/libs/git.py
rudolfix: 1. repo.git.sparse_checkout("set", path) - this only sets the path for the checkout
2. then regular checkout happens

so this is actually OK

tests/libs/test_git.py
rudolfix: comment is not clear. here I'm testing what happens when I pull path that does not exist on already cloned repo. in the next path I do it on a fresh repo, where no previous checkout was done, the code path and git commands are different. I'll write clearer comment

zilto: aah ok! so this test checks specifically the sequence of two operations

zilto (approved):
Feature looks good! 

Some tests seem to fail because we spam GitHub with requests. This will pollute CI further. Feel free to fix now or in follow-up

Issue #3210, opened 2025-10-15
`primary_key` not removed with `apply_hints(primary_key="")` (same issue with `merge_key`)


jorritsandbrink:
### dlt version

1.17.1

### Describe the problem

Providing an empty value for `primary_key` to `apply_hints` does not remove a primary key that was set before:

1. `my_resource.apply_hints(primary_key="id")` sets the primary key correctly
2. `my_resource.apply_hints(primary_key="")` does not remove the primary key (neither does `my_resource.apply_hints(primary_key=[])`)

The documentation for `apply_hints` says: 

_"Pass empty value (for a particular type i.e. "" for a string) to remove a hint."_

The same behavior occurs for `merge_key`.

I found this related issue that may suggest this is intended behavior, but I'm not sure and I think this behavior is confusing: https://github.com/dlt-hub/dlt/issues/2440

### Expected behavior

`my_resource.apply_hints(primary_key="")` (or `my_resource.apply_hints(primary_key=[])` removes the primary key (if exists)

### Steps to reproduce

```py
import json

import dlt


@dlt.resource
def my_table():
    return [{"id": 1, "foo": 1}, {"id": 2, "foo": 1}]


pipe = dlt.pipeline(destination="duckdb")

# Set primary_key using apply_hints
my_table.apply_hints(primary_key="id")
pipe.run(my_table())
table_schema = pipe.default_schema.get_table("my_table")
print('Table schema after `my_table.apply_hints(primary_key="id")` (primary_key set: CORRECT):')
print(json.dumps(table_schema, indent=2))

# Attempt to unset primary_key through apply_hints: doesn't work
my_table.apply_hints(primary_key="")  # primary_key=[] has same behavior as primary_key=""
pipe.run(my_table())
table_schema = pipe.default_schema.get_table("my_table")
print()
print('Table schema after `my_table.apply_hints(primary_key="")` (primary_key still set: INCORRECT):')
print(json.dumps(table_schema, indent=2))

# Set primary_key to False using `columns` as workaround: works
my_table.apply_hints(columns={"id": {"primary_key": False}})
pipe.run(my_table())
table_schema = pipe.default_schema.get_table("my_table")
print()
print('Table schema after `my_table.apply_hints(columns={"id": {"primary_key": False}})` (primary_key set to False: CORRECT, but workaround):')
print(json.dumps(table_schema, indent=2))
```

Output:

```sh
[...]

PR #3280, opened 2025-11-03
Fix: primary_key not removed with apply_hints(primary_key="")

anuunchin:
This PR attempts to solve the issue where setting primary_key to an empty string or None in `apply_hints` does not actually remove it from schema. 

Resolves #3210

PR #3361, opened 2025-11-21
Fix: primary_key not removed with apply_hints(primary_key="")

anuunchin:
This PR attempts to solve the issue where setting `primary_key` to an empty string in `apply_hints` does not actually remove it from schema.

First attempted to be resolved in #3280.

> TLDR: This solution uses the `x-extractor` field to pass empty value hints around. 

Resolves https://github.com/dlt-hub/dlt/issues/3210

### DIFF
--- a/dlt/common/schema/typing.py
+++ b/dlt/common/schema/typing.py
@@ -101,10 +101,17 @@ TColumnHint = Literal[
 COLUMN_HINTS: Set[TColumnHint] = set(get_args(TColumnHint))
 
 
+TColumnPropMergeType = Literal[
+    "replacable",
+    "removable",
+]
+
+
 class TColumnPropInfo(NamedTuple):
     name: Union[TColumnProp, str]
     defaults: Tuple[Any, ...] = (None,)
     is_hint: bool = False
+    merge_type: TColumnPropMergeType = "replacable"
 
 
 _ColumnPropInfos = [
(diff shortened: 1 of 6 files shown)


dlt/common/schema/typing.py
```diff
@@ -101,10 +101,17 @@
 COLUMN_HINTS: Set[TColumnHint] = set(get_args(TColumnHint))
 
 
+TColumnPropMergeType = Literal[
+    "replacable",
```
rudolfix, line 105: just mark certain hints as compound

rudolfix (changes requested):
We can handle everything in `diff_table` if we mark certain hints as `compound` and implement missing behaviors. Here's how hints are evolved:

- we have resource definition (columns, primary keys), apply hints, and hints manipulation directly in code. **this always happens from scratch** in each pipeline run. user can do here whatever user wants. no changes are needed
- extract starts: here we must deal with schema from previous run or with imported schema
- this happens just before extract: `source.schema.update_schema(source_schema)`: we use newly computed source schema to update pipeline schema (table by table)
- tables diffs are calculated: during that process a diff is created which is a partial table with new and modified columns
- the diff is applied to table in pipeline schema and columns are merged with merge_columns.
- table updates happen a few times more (dynamic tables in the extract, schema evolution in the normalize) but the logic is the same

the root of the problem is in `diff_table`: it just blindly merges column hints additively. we must handle compound hints here. we assume that partial table has complete information on each hint. so for example:
- if there a column that sets a primary key, we must reset those keys on all columns of pipeline schema table (table_a) and add them to diff.
- if there a column without primary key but in the table_a the key is set we know that primary key hint is modified and we need to remove it from table_a fully
tldr ;> a block replaces existing block, even if new block is empty (but intersection of column of old blocks with columns in the diff cannot be empty)

a few notes:
1. preserve the column order (merge_columns does it, it will be sufficient)
2. `merge_columns` won't need `columns_partial` - this should be a flag when generating diff (see docstrings)
3. we are implementing `if `columns_partial` is False, hints like `primary_key` and `merge_key` are dropped from `columns_a` and replaced from `columns_b``
4. `merge_column` merge default is always True. you can remove that flag and remove all "false" usage (let's see what happens)

this is not that complicated overall. and the hints merge code is in better shape than I remembered...

anuunchin:
Continued in #3431

PR #3363, opened 2025-11-21
Fix: type annotation failure in compute_columns_schema

### LINKED ISSUE #3345: bug: missing column with `SqlModel`
## Summary
Running a `dlt.hub.transformation` that returns a `dlt.Relation` where the first row contains zeros incorrectly treats the columns as missing.

## Repro

```python
# repro.py
import dlt
from dlt.sources.rest_api import rest_api_resources
import dlthub.data_quality as dq

@dlt.source
def jaffleshop():
    jaffle_rest_resources: list = rest_api_resources({
        "client": {
            "base_url": "https://jaffle-shop.dlthub.com/api/v1",
            "paginator": {"type": "header_link"},
        },
        "resources": [  # individual resources
            "customers",
            "products",
            "orders",
        ],
        # set the time range for all resources
        "resource_defaults": {
            "endpoint": {
                "params": {
                    "start_date": "2017-01-01",
                    "end_date": "2017-01-15",
                },
            },
        },
    })
    
    return jaffle_rest_resources


@dlt.hub.transformation
def jaffle_checks(dataset: dlt.Dataset) -> dlt.Relation:
    checks = {
        "orders": [dq.checks.is_unique("id"), dq.checks.case("subtotal > 0")]
    }
    return dq.prepare_checks(dataset, checks=checks)


pipeline = dlt.pipeline("jaffle_dq", destination="duckdb")


if __name__ == "__main__":
    pipeline.run([jaffleshop()])
``` 
Output of running the code
```shell
❯ uv run python repro.py
DATE|[WARNING]|356876|137503479621440|dlt|validate.py|verify_normalized_table:100|In schema `jaffleshop`: The following columns in table `jaffle_checks` did not receive any data during this load:
  - success_count
  - success_rate

This can happen if you specify columns manually, for example, using the `merge_key`, `primary_key` or `columns` argument but they do not exist in the data.
```

If I change the `if __name__ == "__main__":` clause, I get a different result from what is materialized
```python
if __name__ == "__main__":
    query = dq.prepare_checks(
        pipeline.dataset(),
        checks={
            "orders": [dq.checks.is_unique("id"), dq.checks.case("subtotal > 0")],
        },
    )
    print(query.arrow())
```
Output
```shell
❯ uv run python option1.py
pyarrow.Table
table_name: string
check_qualified_name: string
row_count: int64
success_count: int32
success_rate: double
----
table_name: [["orders","orders"]]
check_qualified_name: [["id__is_unique","subtotal__case__GT"]]
row_count: [[1570,1570]]
success_count: [[1570,1557]]
success_rate: [[1,0.9917197452229299]]
``` 

## Suggested solution
I'm unsure about the origin of the bug. I wonder if we can simplify the lifecycle of `dlt.Relation` and `SqlModel`

anuunchin:
This PR is an attempt to resolve the issue where sqlglot is not able to annotate the type of a column.

Root cause: In some cases, the `selects` of the `expression` passed to the `compute_columns_schema` in `dlt.dataset.lineage.py` include bare Identifiers. `qualify()` fails to qualify these bare Identifier nodes. Then `annotate_types()` fails downstream because it's working with unqualified columns.

Resolves #3345

### DIFF
--- a/dlt/common/libs/sqlglot.py
+++ b/dlt/common/libs/sqlglot.py
@@ -1,4 +1,4 @@
-from typing import Optional, Union, Set, Any, Iterable, Literal
+from typing import Optional, Union, Set, Any, Iterable, Literal, cast
 
 from dlt.common.utils import without_none
 from dlt.common.exceptions import TerminalValueError
@@ -632,3 +632,14 @@ def build_typed_literal(
         return sge.Tuple(expressions=[_literal(v) for v in value])
     else:
         return _literal(value)
+
+
+def wrap_identifiers_in_columns(expression: sge.Query) -> sge.Query:
+    """Wrap bare Identifier nodes inside Alias with Column nodes"""
+
+    def transform(node: sge.Expression) -> sge.Expression:
+        if isinstance(node, sge.Alias) and isinstance(node.this, sge.Identifier):
+            return sge.Alias(this=sge.Column(this=node.this.copy()), alias=node.alias)
+        return node
+
+    return cast(sge.Query, expression.transform(transform))
--- a/dlt/dataset/lineage.py
+++ b/dlt/dataset/lineage.py
@@ -2,7 +2,6 @@ import logging
 from typing import Optional, Tuple, cast
 
 import sqlglot.expressions as sge
-
 from sqlglot.errors import OptimizeError
 from sqlglot.schema import Schema as SQLGlotSchema, ensure_schema
 from sqlglot.optimizer.annotate_types import annotate_types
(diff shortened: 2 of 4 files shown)


zilto (commented):
I'm not sure it's the right solution. An `Identifier` could be a table, schema, view, etc. in addition to an alias or column. 

Though your fix does work because having `Column` objects allows `annotate_types()` to properly resolve types. This logic is very opaque to us...

I investigated some more and will add notes in the initial issue

anuunchin:
Resolved in the dlthub repo

PR #3365, opened 2025-11-22
feat: `snowflake` clustering key modifications

### LINKED ISSUE #3306: Add CLUSTER support on the ALTER TABLE for the Snowflake destination
### Feature description

The clustering hints for the Snowflake destination are currently only respected when the new table is created
```
@dlt.resource(columns={"my_column": {"cluster": True}})
def my_resource():
```

I want dlt to add the clustering clause to the SQL query, even if the cluster hint is added to the existing table.

Code adding the cluster hint for the current implementation: https://github.com/dlt-hub/dlt/blob/229f05f42fdda868b1d30e38a89012ebcd23cbb7/dlt/destinations/impl/snowflake/snowflake.py#L212

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

I need to change the clustering hints for existing table

### Proposed solution

dlt should check if hints are changed and update the clustering statement accordingly

### Related issues

_No response_

jorritsandbrink:
### Description
Currently, the `snowflake` destination only interprets `cluster` hints on table creation. Any subsequent changes in `cluster` hints are ignored. This PR changes that, such that `cluster` hints are also interpreted on table alterations.

These three cases are supported:
(1) add clustering: without clustering -> with clustering
(2) change clustering: change set of cluster columns or column order
(3) drop clustering: with clustering -> without clustering

Note that there is no active hint change detection. Any changes in `cluster` hints are only processed when the current table alteration mechanism is triggered, which is only when a new column is added.

### Related Issues

- Closes #3306

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -278,11 +278,7 @@ class AthenaClient(SqlJobClientWithStagingDataset, SupportsStagingDestination):
             raise e
 
     def _get_table_update_sql(
-        self,
-        table_name: str,
-        new_columns: Sequence[TColumnSchema],
-        generate_alter: bool,
-        storage_columns: Optional[Sequence[TColumnSchema]] = None,
+        self, table_name: str, new_columns: Sequence[TColumnSchema], generate_alter: bool
     ) -> List[str]:
         bucket = self.config.staging_config.bucket_url
         dataset = self.sql_client.dataset_name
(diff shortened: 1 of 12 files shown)


jorritsandbrink:
Hey @rudolfix, could you review this PR?

rudolfix (changes requested):
this is excellent! end to end test is missing though: where we actually run the queries on snowflake. you can add a pipeline testing this in `test_snowflake_pipeline.py`.

jorritsandbrink:
@rudolfix Good call! I added a pipeline test and it exposed an issue with my implementation: it relied on `SqlJobClientBase.get_storage_tables`, which does not include `cluster` hints. More so, even when overriding with `SnowflakeClient.get_storage_tables` and adding the `cluster` hint, it doesn't work because the hints are only applied when a new column is added. Any `cluster` hint change without new column won't be applied, and thus will not show up in the return value of `get_storage_tables` in subsequent runs.

Long story short, I switch to reading the `cluster` hints from the table schema, which does work and actually simplifies the implementation.

So, could you review again?

Issue #3311, opened 2025-11-12
feat: Add 'Insert-Only' Merge Strategy to preserve ingestion metadata and prevent key updates/deletes


OnAzart:
### Feature description

Introduce a new dedicated merge strategy (or a configurable option within the existing `merge` disposition) that enables **idempotent, key-based appending** by implementing **"Insert Only on Key Mismatch"**.

This strategy's sole function is to insert records whose `primary_key` (e.g., a hash or unique ID) is not currently present in the destination table, while silently skipping/ignoring any record whose key already exists.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

**Title:** Preserving Ingestion Metadata (`_dlt_load_id`) During Full-History Deduplication

I am extracting **full history** from a source system in each load and need to ingest only new data, deduplicated by a **hash column** (the `primary_key`).

The current strategies cause issues with downstream data modeling (e.g., dbt incremental models):

1.  **Pure `append`** strategy fails due to duplicates.
2.  **`upsert`** strategy updates the row's ingestion timestamp (`_dlt_load_id`) each time the full history is loaded, even if the record content hasn't changed.
3.  **`delete-insert`** strategy also deletes the old record and inserts the new one, destroying the original ingestion timestamp (`_dlt_load_id`).
4.  **SCD Type 2** is not suitable as the goal is simple deduplication by content hash, not tracking history with `start_date`/`end_date`.

This feature is essential to ensure that the original ingestion timestamp is preserved, allowing for reliable and efficient downstream incremental modeling based on the `_dlt_load_id`.

### Proposed solution

Implement a new `write_disposition` strategy (e.g., **`append_unique`** or **`merge_insert_only`**) or a configuration flag for the existing `merge` disposition.

The resulting SQL execution at the destination should utilize the most efficient method available:

| Condition | Action |
| :--- | :--- |
| **Key NOT Found** | **INSERT** the new record. |
| **Key FOUND** | **SKIP** the record (`DO NOTHING`). |

This would offload the complex pre-deduplication logic (which currently requires querying the destination for existing keys in Python) to the optimized destination engine using native constructs like:
* SQL `MERGE INTO ... WHEN NOT MATCHED THEN INSERT ... WHEN MATCHED THEN DO NOTHING`
* PostgreSQL `INSERT INTO ... ON CONFLICT DO NOTHING`
* Appropriate constructs for other databases (e.g., `WHERE NOT EXISTS`).

PR #3372, opened 2025-11-24
feat: Add insert-only merge strategy for improved performance and data immutability

OnAzart:
### Description

This PR implements a new **`insert-only` merge strategy** for dlt that inserts new records based on primary key without updating existing records. This strategy provides better performance and data immutability guarantees for append-only data loading scenarios..

**Key features:**
- **Insert-only merging**: New records are inserted; existing records remain unchanged (no updates)
- **Hard delete support**: Records marked for deletion are filtered from staging before insertion
- **Nested table support**: Recursive insert-only behavior for parent-child relationships
- **16 destinations supported**: Snowflake, BigQuery, PostgreSQL, Databricks, DuckDB (v1.4.0+), and more
- **Performance optimized**: Eliminates UPDATE operations for better performance than upsert

**Implementation details:**
- Added `"insert-only"` to `TLoaderMergeStrategy` type definition
- Implemented `gen_insert_only_sql()` method in `SqlMergeFollowupJob` class
- Added schema validation requiring `primary_key` for insert-only strategy
- Updated destination capabilities for all 16 supported destinations
- Version-gated support for DuckDB (requires MERGE statement support from v1.4.0+)

### Related Issues
https://github.com/dlt-hub/dlt/issues/3311

#### Motivation

Many data pipelines work with immutable data where records should never change once created (events, logs, transactions, audit trails). The existing merge strategies either update records (upsert) or perform delete+insert operations (delete-insert), which is unnecessary overhead when data immutability is guaranteed.

The `insert-only` strategy fills this gap by providing:
1. **Data immutability guarantees**: Existing records are never modified
2. **Idempotent pipeline runs**: Re-running a pipeline only adds missing records
3. **Better performance**: Eliminates UPDATE operations entirely

#### Comparison with Existing Strategies

| Strategy | Updates | Deletes | Requires PK | Performance | Use Case |
|----------|---------|---------|-------------|-------------|----------|
| **insert-only** (new) | ❌ No | ❌ No | ✅ Yes | ⭐⭐⭐ Best | Immutable data |
| **upsert** | ✅ Yes | ✅ Yes | ✅ Yes | ⭐⭐ Good | Mutable records |
| **delete-insert** | ✅ Yes | ✅ Yes | ⚠️ Optional | ⭐ Slower | Batch updates |

#### SQL Example

The insert-only strategy generates MERGE statements without the UPDATE clause:

```sql
-- Root table with hard delete filtering
MERGE INTO target_table d 
USING (SELECT * FROM staging_table WHERE deleted_at IS NULL) s
ON d.id = s.id
WHEN NOT MATCHED
    THEN INSERT (id, name, value) VALUES (s.id, s.name, s.value);
-- Notice: No "WHEN MATCHED THEN UPDATE" clause

### DIFF
--- a/dlt/destinations/impl/duckdb/factory.py
+++ b/dlt/destinations/impl/duckdb/factory.py
@@ -136,8 +136,13 @@ def duckdb_merge_strategies_selector(
         import duckdb as _duckdb
 
         if Version(_duckdb.__version__) < Version("1.4.0"):
+            # MERGE statement was added in DuckDB 1.4.0
+            # Remove strategies that require MERGE for older versions
             legacy_strategies = list(supported_merge_strategies)
-            legacy_strategies.remove("upsert")
+            if "upsert" in legacy_strategies:
+                legacy_strategies.remove("upsert")
+            if "insert-only" in legacy_strategies:
+                legacy_strategies.remove("insert-only")
             supported_merge_strategies = legacy_strategies
     except ImportError:
         # return default if duckdb not installed


OnAzart:
@rudolfix that's an awesome comment, thanks! 
I was also considering the option to add this to the existing strategy, but had some hesitations because of the definition of upsert (update + insert), and corresponding to this definition as being without update. 
Would it be logical for the user to look for that logic there?

It will definitely be easier to implement and not overload strategies, just want to make sure you are okay with that too.

rudolfix:
@OnAzart OK then let's try a separate merge-strategy: `insert-only` but 
1. still please reuse the same code as `upsert` ie. call it with a flag that will skip updates
2. as mentioned, we need deterministic row_key (_dlt_id) like for upsert. see get_root_row_id_type
3. we need `root_key` propagation: see requires_root_key
4. since it is trivial to update `deltalake` and `iceberg` to handle new merge strategy: see `merge_delta_table` and `merge_iceberg_table`
5. you marked lancedb as implementing this strategy. it is also trivial to update it (`write_records`)
6. `verify_schema_merge_disposition` must verify new merge strategy for primary key

changes are not hard nor complicated. for better undrestaning look for "upsert" string in code base. your real challenge are tests. btw. your current tests are 100% not passing. you must run your tests on duckdb and local filesystem (iceberg, delta) before submitting.

adding new merge strategy is a big change and we we'll need to test it properly. so you should expect some before we merged it...

rudolfix:
@OnAzart this PR continues here #3741 thanks for contributing this!

Issue #3276, opened 2025-11-01
bug: toggling `dev_mode` doesn't reset `dataset_name`


zilto:
Run a pipeline without `dev_mode` (defaults to False); You get a dataset named `{dataset_name}`

Repro
1. Set `dev_mode=True`. For each run, it produces a namespaced dataset `{dataset_name}_{timestamp}`. 
2. Set `dev_mode=False`, data is written to the latest namespaced dataset. 

Intuitively, Step 2 should create or write to non-namespaced dataset named `{dataset_name}`.

I don't see good a reason why the sequence `dev_mode=True -> dev_mode=False` should matter. The behavior of `dev_mode=False` should be consistent (not stateful)

I'm filing this as a bug given behavior isn't intuitive.

rudolfix:
note: this only happens when `dataset_name` is not set explicitly in the pipeline. in that case `dataset_name` comes from existing state. to fix this edge case, `dev_mode` flag must be saved in state (local part) and when state is loaded and this flag is set a new empty state should be returned to reset pipeline

PR #3381, opened 2025-11-25
Fix/3276: ensure consistency of dataset_name when dev_mode is set to false

alkaline-0:
## Description

Persist previous dev_mode in local state and reset pipeline state on dev_mode toggle from True→False. Guarantees that non-dev runs always use a non-suffixed dataset name (explicit dataset_name if provided; otherwise default), independent of prior dev runs.

## Details:

Persist _local["_last_dev_mode"] on state writes.

## Files touched:

1. dlt/pipeline/pipeline.py
2. tests/pipeline/test_pipeline.py

Related Issues:
Fixes https://github.com/dlt-hub/dlt/issues/3276

### DIFF
--- a/dlt/common/pipeline.py
+++ b/dlt/common/pipeline.py
@@ -479,7 +479,8 @@ class TPipelineLocalState(TypedDict, total=False):
     """Run dir when pipeline was instantiated for a first time, defaults to cwd on OSS run context"""
     last_run_context: Optional[TLastRunContext]
     """Context from the last successful pipeline run or sync"""
-    _last_dev_mode: bool  # indicates whether previous run used dev_mode
+    dev_mode: bool
+    destination_name: Optional[str]
     """Indicates whether previous run used dev_mode; used to reset state on dev->non-dev toggle"""
 
 
--- a/dlt/pipeline/pipeline.py
+++ b/dlt/pipeline/pipeline.py
@@ -1249,15 +1249,15 @@ class Pipeline(SupportsPipeline):
         # attach to pipeline if folder exists and contains state
         reset_state: bool = False
         if has_state:
-            _local_state = json_decode_state(self._pipeline_storage.load(Pipeline.STATE_FILE))
-            reset_state = (
-                bool(_local_state.get("_local", {}).get("_last_dev_mode")) and not self.dev_mode
-            )
+            _last_dev_move = self.get_local_state_val("dev_mode")
+            reset_state = _last_dev_move and not self.dev_mode
         if reset_state:
             self._create_pipeline()
         else:
             self._attach_pipeline()
 
+        self.set_local_state_val("dev_mode", self.dev_mode)
+        self.set_local_state_val("dataset_name", self.dataset_name)
         # create schema storage
         self._schema_storage = LiveSchemaStorage(self._schema_storage_config, makedirs=True)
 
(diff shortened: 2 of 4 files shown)


dlt/common/pipeline.py
```diff
@@ -479,6 +479,8 @@ class TPipelineLocalState(TypedDict, total=False):
     """Run dir when pipeline was instantiated for a first time, defaults to cwd on OSS run context"""
     last_run_context: Optional[TLastRunContext]
     """Context from the last successful pipeline run or sync"""
+    _last_dev_mode: bool  # indicates whether previous run used dev_mode
```
rudolfix: if you name this `dev_mode` pipeline props will save and restore this prop automatically, `_` skips the prop

dlt/pipeline/pipeline.py
```diff
@@ -1246,11 +1247,16 @@ def _configure(
 
         self.must_attach_to_local_pipeline = must_attach_to_local_pipeline
         # attach to pipeline if folder exists and contains state
+        reset_state: bool = False
         if has_state:
-            self._attach_pipeline()
-        else:
-            # this will erase the existing working folder
+            _local_state = json_decode_state(self._pipeline_storage.load(Pipeline.STATE_FILE))
```
rudolfix: use `get_local_state_val` instead of this

dlt/pipeline/pipeline.py
```diff
@@ -1686,12 +1692,17 @@ def managed_state(self, *, extract_state: bool = False) -> Iterator[TPipelineSta
 
     def _state_to_props(self, state: TPipelineState) -> None:
         """Write `state` to pipeline props."""
+        if state.get("_local", {}).get("_last_dev_mode") and not self.dev_mode:
```
rudolfix: I do not think you need to modify this function if you name local prop `dev_mode`. also why you drop `dataset_name` here? we remove pipeline state fully in init.

dlt/pipeline/pipeline.py
```diff
@@ -1732,6 +1743,7 @@ def _props_to_state(self, state: TPipelineState) -> TPipelineState:
         if self._staging:
             state["staging_type"] = self._staging.destination_type
             state["staging_name"] = self._staging.configured_name
+        state["_local"]["_last_dev_mode"] = self.dev_mode
```
rudolfix: same here

dlt/pipeline/pipeline.py
```diff
@@ -1771,7 +1783,9 @@ def _bump_version_and_extract_state(
 
         Storage will be created on demand. In that case the extracted package will be immediately committed.
         """
-        _, hash_, _ = bump_pipeline_state_version_if_modified(self._props_to_state(state))
+        # ensure state reflects current props and toggle info before hashing/bumping
+        self._props_to_state(state)
```
rudolfix: this looks like invasive change... why you need to write state here?

tests/load/pipeline/test_drop.py
```diff
@@ -25,7 +25,17 @@
 
 
 def _attach(pipeline: Pipeline) -> Pipeline:
-    return dlt.attach(pipeline.pipeline_name, pipelines_dir=pipeline.pipelines_dir)
+    p = dlt.attach(
```
rudolfix, line 28: I'm not sure any test should be modified until we write the test for the feature itself. there may be a few cases where we remove dev_mode during the test. then we'll see

rudolfix (changes requested):

dlt/pipeline/pipeline.py
alkaline-0: This PR was an attempt to figure out the right approach and show how far I got to get feedback :D The reason I called this here is that by removing the dataset name from _state_to_props I'd force the make_dataset_name to produce a new name.

dlt/common/pipeline.py
alkaline-0: That also meant that it's public, My intention was to make it a private attribute for our state. I was aware of the props being populated when you remove the "_" but made.a judgement call to make it encapsulated

dlt/pipeline/pipeline.py
alkaline-0: since I was using _last_dev and the props didn't fill it automatically I wanted to ensure the values persisted

PR #3386, opened 2025-11-26
Fix: pipeline_drop.__init__() got an unexpected keyword argument 'no_pwd'

### LINKED ISSUE #3379: Unable to drop pipeline resources using CLI
### dlt version

1.18.2

### Describe the problem

In DLT 1.18, an error is thrown when trying to drop a pipeline resource

```
pipeline_drop.__init__() got an unexpected keyword argument 'no_pwd'
```

### Expected behavior

_No response_

### Steps to reproduce

1. Trying to drop a pipeline using the CLI
2. The error is thrown
```
pipeline_drop.__init__() got an unexpected keyword argument 'no_pwd'
```

### Operating system

Linux

### Runtime environment

Local

### Python version

3.13

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
This PR just explicitly passes the required keyword arguments to the `drop_pipeline`. Previously, global arguments that were being passed upstream were causing an unexpected argument error.

Resolved #3379

> On that note, a better solution may be to filter out global arguments before they reach the `def execute()` of `PipelineCommand()`, because they are not being used anyway.

### DIFF
--- a/dlt/_workspace/cli/_pipeline_command.py
+++ b/dlt/_workspace/cli/_pipeline_command.py
@@ -1,6 +1,7 @@
 import os
 import yaml
 from typing import Any, Sequence, Tuple
+from inspect import signature
 import dlt
 
 from dlt.common.json import json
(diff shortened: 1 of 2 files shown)


dlt/_workspace/cli/_pipeline_command.py
```diff
@@ -386,7 +387,14 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
...
     if operation == "drop":
-        drop = pipeline_drop(p, **command_kwargs)
+        drop = pipeline_drop(
+            p,
+            resources=command_kwargs.get("resources", ()),
+            schema_name=command_kwargs.get("schema_name"),
+            state_paths=command_kwargs.get("state_paths", ()),
+            drop_all=command_kwargs.get("drop_all", False),
+            state_only=command_kwargs.get("state_only", False),
+        )
```
anuunchin, line 397: We could also inspect the signature of pipeline_drop and make this more dynamic 👀

PR #3392, opened 2025-11-26
Implement advanced data masking for SQL sources

### LINKED ISSUE #2116: Add a more realistic data masking example
### Documentation description

Currently, the data masking example at https://dlthub.com/docs/general-usage/customising-pipelines/pseudonymizing_columns is not very realistic, as it's unlikely anyone would want to use a function with hardcoded column names. Instead, a function which takes `columns` as a parameter could be shown. Such function is more complicated that the dummy example, but it would be helpful to people wanting to implement masking by showing a real-world example.

Here's the function I wrote (note that it's only tested with `sql_database` and `sql_table` sources).

```python
def mask_sql_db_columns(columns: list[str]) -> pa.Table | pd.DataFrame | dict:
    """Mask specified columns in a SQL Database table.

    All backends supported by the sql_database source, as of version 1.4.1, are
    supported. See https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/sql_database/helpers.py#L50


    Args:
        columns (list[str]): The list of columns to mask.

    Returns:
        pa.Table | pd.DataFrame | dict[str, Any]: The table or row with the specified columns
            masked.

    """

    def mask_columns(
        table_or_row: pa.Table | pd.DataFrame | dict[str, Any],
    ) -> pa.Table | pd.DataFrame | dict[str, Any]:
        mask = "******"

        # Handle `pyarrow` and `connectorx` backends.
        if isinstance(table_or_row, pa.Table):
            table = table_or_row
            for col in table.schema.names:
                if col in columns:
                    table = table.set_column(
                        table.schema.get_field_index(col),
                        col,
                        pa.array([mask] * len(table[col])),
                    )
            return table

        # Handle `pandas` backend.
        if isinstance(table_or_row, pd.DataFrame):
            table = table_or_row

            for col in table.columns:
                if col in columns:
                    table[col] = mask
            return table

        # Handle `sqlalchemy` backend.
        if isinstance(table_or_row, dict):
            row = table_or_row
            for col in row:
                if col in columns:
                    row[col] = mask
            return row

        # Handle unsupported table types.
        msg = f"Unsupported table type: {type(table_or_row)}. Supported backends: (pyarrow, connectorx, pandas, sqlalchemy)."
        raise NotImplementedError(msg)

    return mask_columns
 ```
 
 Then, it can be used like so:
 
```python
table = sql_table_source(table="test_table")
table.add_map(mask_sql_db_columns(columns=["value"]))

pipeline = dlt.pipeline()
load_info = pipeline.run(table)
```

### Are you a dlt user?

Yes, I'm already a dlt user.

shanto12:
Added advanced data masking functionality for SQL databases, allowing flexible column masking with multiple methods and support for various data backends.

### Related Issues
- Resolves #2116- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/docs/website/docs/general-usage/customising-pipelines/pseudonymizing_columns.md
+++ b/docs/website/docs/general-usage/customising-pipelines/pseudonymizing_columns.md
@@ -57,3 +57,107 @@ pipeline = dlt.pipeline(pipeline_name='example', destination='bigquery', dataset
 load_info = pipeline.run(source_instance)
 ```
 
+
+
+## Advanced Data Masking for SQL Database Sources
+
+The example above works well for simple cases with hardcoded column names. For production use cases, you might want a more flexible function that:
+
+- Takes column names as parameters
+- Supports multiple masking methods (mask with asterisks or nullify)
+- Works with different data backends (PyArrow, Pandas, SQL)
+
+Here's a more realistic data masking example:
+
+```py
+from enum import StrEnum
+import pyarrow as pa
+import pandas as pd
+import dlt
+
+class MaskingMethod(StrEnum):
+    MASK = "mask"
+    NULLIFY = "nullify"
+
+def mask_sql_db_columns(
+    columns: list[str],
+    method: MaskingMethod | None = None,
+    mask: str = "*******",
+) -> pa.Table | pd.DataFrame | dict:
+    """Mask specified columns in a SQL Database table.
+    
+    Args:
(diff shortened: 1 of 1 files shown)


rudolfix:
@shanto12 several implementations of this were merged into one and added to docs here: https://dlthub.com/docs/general-usage/customising-pipelines/pseudonymizing_columns. thanks for contributing this!

PR #3404, opened 2025-11-29
fix: 3276 - toggling dev_mode doesn't reset dataset_name causing inconsistencies of dataset naming behavior when creating a pipeline without dev_mode

alkaline-0:
## Problem

When a pipeline runs with dev_mode=True, its dataset_name is suffixed with a unique pipeline_instance_id. On subsequent runs with dev_mode=False, the suffix could persist via restored state, causing:

1. Confusing dataset naming (dev suffix remains in non-dev runs)
2. Inconsistent behavior across attaches/initializations
3. State/schema mismatches after local wipes

## Suggested solution

1. We persist a _local["dev_mode"] flag to detect the toggle and decide the reset early.
2. Dev→non-dev is now a “hard reset”: we wipe the pipeline working directory and start from a clean state.
3. After reset, we recompute dataset_name as if it were the first initialization, without the dev suffix.

## Implementation details
Add dev_mode to TPipelineLocalState. 
Hard reset on dev→non-dev
In _configure, If _local["dev_mode"] was True and current self.dev_mode is False:

- Wipe the working dir (_create_pipeline).
- Replace the injected in-memory state (if available) with default_pipeline_state() to reflect a true fresh start.
- recompute dataset_name (non-dev), then save.

## Issues
solves #3276

### DIFF
--- a/dlt/pipeline/state_sync.py
+++ b/dlt/pipeline/state_sync.py
@@ -57,10 +57,7 @@ def migrate_pipeline_state(
     if from_engine == to_engine:
         return cast(TPipelineState, state)
     if from_engine == 1 and to_engine > 1:
-        state["_local"] = {
-            "dev_mode": state.get("dev_mode", False),
-            "initial_cwd": os.path.abspath(dlt.current.run_context().local_dir),
-        }
+        state["_local"] = {}
         from_engine = 2
     if from_engine == 2 and to_engine > 2:
         # you may want to recompute hash
--- a/tests/load/pipeline/test_pipelines.py
+++ b/tests/load/pipeline/test_pipelines.py
@@ -395,7 +395,11 @@ def test_run_dev_mode(destination_config: DestinationTestConfiguration) -> None:
     # print(info)
 
     # restore the pipeline
-    p = dlt.attach()
+    p = dlt.attach(
+        destination=destination_config.destination_factory(),
+        staging=destination_config.staging,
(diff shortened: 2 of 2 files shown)


dlt/pipeline/pipeline.py
```diff
@@ -1247,13 +1271,15 @@ def _configure(
             )
 
         self.must_attach_to_local_pipeline = must_attach_to_local_pipeline
-        # attach to pipeline if folder exists and contains state
-        if has_state:
```
rudolfix: `_get_prev_dev_mode_from_state` reproduces what `get_local_state_val` already does (it will load and migrate state). I think we can reuse it. We also need to "migrate" the state:
1. in `def _get_state(self)` please set `dev_mode` in local state key does not exist (see what we do with initial_cwd)
2. get previous dev mode with `get_local_state_val`
3. then we recreate pipeline or not - as you do below.

tests/load/pipeline/test_drop.py
```diff
@@ -167,7 +167,7 @@ def test_drop_command_resources_and_state(
     if not in_source:
         source = list(source.selected_resources.values())
 
-    pipeline = destination_config.setup_pipeline("droppable", dev_mode=True)
+    pipeline = destination_config.setup_pipeline("droppable")
```
rudolfix: would be good to know why you remove this flag here. our change should be not visible in most of the cases

PR #3409, opened 2025-12-01
Fix: reset config in PluggableRunContext.reload_providers

ivasio:
Current implementation wasn't resetting the config fully, e.g. in workspace contexts `Container()[PluggableRunContext].reload_providers()` wasn't fully reloading the providers due to `WorkspaceRunContext` keeping the cached value in `_config` property

This PR implements an abstract method that Context classes (i.e. children of `RunContextBase`) should reimplement in order to reset their cached config properties. This caching is implemented on the child level, so I have seen no way to work around adding another abstract method into the class hierarchy of `RunContextBase`

### DIFF
--- a/.github/workflows/test_hub.yml
+++ b/.github/workflows/test_hub.yml
@@ -81,7 +81,7 @@ jobs:
         run: make dev
 
       - name: Install dlthub
-        run: uv run pip install --upgrade --force-reinstall --no-cache-dir ${{ matrix.dlthub_dep }}
+        run: uv run pip install --upgrade --force-reinstall --pre --no-cache-dir ${{ matrix.dlthub_dep }}
 
       - name: Run tests
         run: pytest tests/hub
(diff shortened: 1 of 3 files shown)


dlt/common/configuration/specs/pluggable_run_context.py
```diff
@@ -206,6 +210,9 @@ def reload(
 
     def reload_providers(self) -> None:
         self.providers = ConfigProvidersContainer(self.context.initial_providers())
+        # Invalidate any cached configuration on the context so it re-resolves using new providers
```
rudolfix: I'd do this after extra providers are added

dlt/common/runtime/run_context.py
```diff
@@ -114,6 +114,10 @@ def plug(self) -> None:
     def unplug(self) -> None:
         pass
 
+    def reset_config(self) -> None:
+        self._runtime_config = None
```
rudolfix, line 120: I'd follow the same pattern that workspace config does:
- here we just reset _runtime_config
- move
```
if runtime_config is None:
            self._runtime_config = resolve_configuration(RuntimeConfiguration())
        else:
            self._runtime_config = runtime_config
```
to `runtime_config` so it creates it on demand

and
.initialize_runtime(self.name, self._runtime_config) -> initialize_runtime(self.name, self.runtime_config)
so it creates runtime_config

@abstractmethod
    def reset_config(self) -> None:
        """Hook for contexts that store resolved configuration to reset it"""

ivasio: calling `self.initialize_runtime` in the getter now

this suggestion `.initialize_runtime(self.name, self._runtime_config) -> initialize_runtime(self.name, self.runtime_config)` would lead to infinite recursion, so kept it as it was

PR #3391, opened 2025-11-26
Add Hugging Face filesystem destination

lhoestq:
### Description

Add support for Hugging Face storage (filesystem)

```python
from dlt.destinations import filesystem

destination = filesystem(f"hf://datasets/{username}/{dataset_name}")
```

### Related Issues

- close the issue https://github.com/dlt-hub/dlt/issues/1227
- this requires https://github.com/huggingface/huggingface_hub/pull/3575 to be merged and released (then happy to add the "hf" extra in pyproject.toml if it makes sense)

### Additional Context

this allows HF users to make datasets on HF from anywhere, which they can use to train/eval AI models

I was also thinking of defining a dedicated `huggingface` destination which would structure the files in "the HF way" (train/test separation, dataset viewer configuration, parquet optimizations), lmk if you think this could be a welcome contribution to push this further.

```python
# idea
from dlt.destinations import huggingface

destination = huggingface(f"{username}/{dataset_name}")
```

Issue #1227, opened 2024-04-16
Hugging Face Hub integration


jorritsandbrink:
### Desired functionality
- I want to read files from a `dataset` hosted on the Hugging Face Hub as a `dlt` source
- I want to write files to a `dataset` hosted on the Hugging Face Hub as a `dlt` destination

### Supported file types
While the HF Hub can host any filetype (it's a Git repo under the hood), it's probably wise to limit initial support to `dlt`'s currently supported tabular data file formats: `csv`, `jsonl`, and `parquet`. We could extend later if users show interest.

### Approach
In line with Hugging Face' [direction](https://www.linkedin.com/feed/update/urn:li:activity:7056363700126920704/), it makes sense to treat the HF Hub as a cloud storage bucket similar to Amazon S3, Azure Blob Storage, and Google Cloud Storage. This should also make it relatively straightforward to implement the desired functionality, as the `huggingface_hub` Python library [implements](https://github.com/huggingface/huggingface_hub/blob/main/docs/source/en/guides/hf_file_system.md) `fsspec`, which `dlt` can already handle.

### Supported write dispositions
Using this approach, HF Hub will fall under the `filesystem` source/destination category. The `filesystem` destination supports the `append` and `replace` write dispositions, but not `merge`. We can implement `merge` later if users show interest (that would be for the `filtesystem` destination as a whole, not just HF Hub).

### Missing functionality
- Hugging Face credentials abstraction

### Are you a dlt user?

Yes, I'm already a dlt user.

rudolfix:
@jorritsandbrink looks like a good first step.

> I want to read files from a dataset hosted on the Hugging Face Hub as a dlt source

I'm however curious how the HF datasets library would behave with the data we produce. if you you create a few parquet files where schema evolves (we append columns at the end) and push them to the same dataset, then request with `load_dataset` what is going to happen? Can I upload csvs and see them as single parquet on the client side? also streaming
```
dataset = load_dataset('oscar-corpus/OSCAR-2201', 'en', split='train', streaming=True)
```

so:

> I want to access my dataset, train and stream like any other HF datasets

rudolfix:
@jorritsandbrink seems that we'd need to bring the files locally and merge them using arrow or duckdb then emit unified dataset. this seems both very useful and doable. However we need to get more info on what the HF users need because this is IMO quite and investment.

PR #3410, opened 2025-12-01
Add `huggingface()` destination

lhoestq:
### Description

Add `huggingface()` destination

```python
from dlt.destinations import huggingface

destination = huggingface(dataset_name="lhoestq/dlt-experiment")
```

It writes the files as `"data/{split}-xxxxx-of-nnnnn.parquet"` to match the usual/default HF structure, but we could use whatever structure and add the associated file pattern in YAML (needed for the HF Dataset Viewer) instead if it makes sense.

It also adds the default Parquet value:
- `use_content_defined_chunking=True` -> for [Parquet CDC](https://huggingface.co/blog/parquet-cdc)
- `write_page_index=True` -> this allows better filtering and random access, esp. in the HF Dataset Viewer

**Question 1**: `use_content_defined_chunking` is only available in recent versions of `pyarrow`, how would you like this to be handled ?

**Question 2**: is there a way for the JobClient to retrieve information from the jobs ? I would like each job to pre-upload the files and the JobClient to `commit()` once the jobs are all done, and for this I need each job to return some information about the pre-upload step (more precisely, the id of each remote file after pre-upload)

### Related Issues

- related to https://github.com/dlt-hub/dlt/pull/3391 which adds HF filesystem support
- continuation of https://github.com/dlt-hub/dlt/issues/1227

### DIFF
--- a/dlt/destinations/impl/huggingface/huggingface.py
+++ b/dlt/destinations/impl/huggingface/huggingface.py
@@ -73,13 +73,8 @@ class HfClient(JobClientBase):
     def complete_load(self, load_id: str) -> None:
         """Marks the load package with `load_id` as completed in the destination. Before such commit is done, the data with `load_id` is invalid."""
 
-        file_paths_in_repo = [
-            file.path
-            for file in self.hf_api.list_repo_tree(self.dataset_name, repo_type="dataset", path_in_repo=f".load_id={load_id}")
-        ]
-
         def format_path(i):
-            return f"data/{self.split}-{i:05d}-of-{len(file_paths_in_repo):05d}.parquet"
+            return f"data/{self.split}-{i:05d}-of-{len(files_to_rename):05d}.parquet"
 
         def rename(old_path, new_path):
             if old_path != new_path:
@@ -87,13 +82,23 @@ class HfClient(JobClientBase):
                     src_path_in_repo=old_path, path_in_repo=new_path
                 )
                 yield CommitOperationDelete(path_in_repo=old_path)
-        
+
+        files_to_delete = list(self.hf_api.list_repo_tree(self.dataset_name, repo_type="dataset", path_in_repo="data"))
+        operations = [
+            CommitOperationDelete(file.path)
+            for file in files_to_delete
+            if file.path.startswith(f"data/{self.split}-")
+        ]
+
+        files_to_rename = list(self.hf_api.list_repo_tree(self.dataset_name, repo_type="dataset", path_in_repo=f".load_id={load_id}"))
+        operations += [
+            operation
+            for i, file in enumerate(files_to_rename)
+            for operation in rename(file.path, format_path(i))
+        ]
+
         self._create_commits(
-            operations=[
-                operation
-                for i, file_path_in_repo in enumerate(file_paths_in_repo)
-                for operation in rename(file_path_in_repo, format_path(i))
-            ],
+            operations=operations,
             message="Upload using dlt",
         )
 


lhoestq:
Hi @rudolfix let me know what you think :)

dlt/common/data_writers/writers.py
```diff
@@ -351,6 +353,8 @@ def _create_writer(self, schema: "pa.Schema") -> "pa.parquet.ParquetWriter":
             coerce_timestamps=self.parquet_format.coerce_timestamps,
             allow_truncated_timestamps=self.parquet_format.allow_truncated_timestamps,
             use_compliant_nested_type=self.parquet_format.use_compliant_nested_type,
+            use_content_defined_chunking=self.parquet_format.use_content_defined_chunking,
```
rudolfix, line 356: we'd like to support wide range of arrow versions. please set those only if arrow version allows that. ie. we do that below
```py
# detect non-null columns receiving nulls. above v.19 it is checked in `write_table`
        if Version(pyarrow.__version__).major < 19:
            table = table.cast(self.schema)
        # Write
        self.writer.write_table(table, row_group_size=self.parquet_format.row_group_size)
```
we also
1. document that in docs
2. we will create `huggingface` extra (as we do for each destination) in PyPI package where we will include all dependencies including pyarrow version that has those options

dlt/destinations/impl/huggingface/huggingface.py
```diff
@@ -0,0 +1,135 @@
...
+class HfLoadJob(RunnableLoadJob):
+
+    def __init__(self, file_path: str) -> None:
+        super().__init__(file_path)
+        self._job_client: "HfClient" = None
+
+    def run(self):
+        self._job_client.hf_api.upload_file(
+            path_or_fileobj=self._file_path,
+            path_in_repo=f".load_id={self._load_id}/{self.file_name()}",
```
rudolfix, line 25: I think this is good for a first version! FYI: we have a decent support for file layouts that can be reused ie. look at `FilesystemDestinationClientConfiguration` and https://docs-data-quality-docs.services4758.workers.dev/docs/devel/dlt-ecosystem/destinations/filesystem#files-layout

dlt/destinations/impl/huggingface/huggingface.py
```diff
@@ -0,0 +1,135 @@
...
+    def create_load_job(
+        self, table: PreparedTableSchema, file_path: str, load_id: str, restore: bool = False
+    ) -> RunnableLoadJob:
+        """Creates a load job for a particular `table` with content in `file_path`. Table is already prepared to be loaded."""
+        return HfLoadJob(file_path)
+
+    def complete_load(self, load_id: str) -> None:
+        """Marks the load package with `load_id` as completed in the destination. Before such commit is done, the data with `load_id` is invalid."""
+
+        def format_path(i):
```
rudolfix, line 76: if you want to commit all the files together this is the correct place. All jobs share "load_package_state" where they can leave additional info that will be persisted ie. across retries.
* destination_state() - will get a shared state dics
* commit_load_package_state() - will commit dict after it got written (we do not do it automatically right now)
you can do that at the end of copy job
NOTE: this is multithreaded code. you can use state dict as lock. commit_load_package_state is thread safe

However there are a few other ways to do that:
- we have a way for a job to create a followup job. ie. a copy job copies the file to a bucket and creates an sql follow up job with COPY command. (basically a file with url to copy but payload can be anything). follow up jobs are scheduled as fast as possible and they do not wait for any other jobs
- table chain followup job: here loader executes all jobs loading to particular root table and all child tables and then a followup job is created for all those tables. example: a MERGE sql job that needs all the data in staging dataset. still those jobs do not wait for other root tables
we can use any of those as well. let me know. I can explain more

dlt/destinations/impl/huggingface/huggingface.py
```diff
@@ -0,0 +1,135 @@
...
+            return f"data/{self.split}-{i:05d}-of-{len(files_to_rename):05d}.parquet"
+
+        def rename(old_path, new_path):
```
rudolfix, line 79: so code below renames a set of files? I'm not 100% sure why there's a separate delete operation?

anyway: this code must be idempotent.. dlt will re-try `complete_load` if it fails. and it should be able to restart where it stopped. if you use package state you'll probably want to delete processed files from it and commit after each rename

ofc a single "batch" rename on huggingface side would be the safest. but with the above we are also fine

dlt/destinations/impl/huggingface/huggingface.py
rudolfix: one more details: `dlt` may load data to a single "table" in many chunks in many threads at once. This can be controlled (ie. by table chain jobs or by sequential loading strategy). this is FYI: you know better if HF can handle such thing

rudolfix (commented):
hey @lhoestq 
this looks really good and I think is pretty close to a first version we can merge. I answered your questions in review comments.

I'm excited for this to be merged. We can add more advanced features later (like state and schema restoration to support incremental loading)

zilto:
Closing since #3669 was merged

PR #3414, opened 2025-12-02
feat/3187 add engine args sqlalchemy source and destination

### LINKED ISSUE #3187: Allow passing oracle arraysize to sqlalchemy when using sql_database source
### Feature description

When using the sql_database source, I want to be able to pass the oracle arraysize to sql alchemy as documented here: https://docs.sqlalchemy.org/en/20/dialects/oracle.html#engine-options-consumed-by-the-sqlalchemy-oracledb-dialect-outside-of-the-driver

Sql alchemy then passes the setting to the underlying driver.

Being able to specify the arraysize is essential when extracting large tables since it can greatly reduce the number of round trips to the database.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

_No response_

### Proposed solution

_No response_

### Related issues

_No response_

tetelio:
### Description
dlt users want to pass SQLAlchemy engine arguments for:
1. SQL sources (sql_database and sql_table)
2. SQL destination

### Summary of Changes

This PR introduces support for configuring SQLAlchemy engine arguments across both sources and destinations, addressing a long-standing usability gap:

- **Sources (`sql_database`, `sql_table`)**
  - Added a new parameter `engine_kwargs`, forwarded directly to `sqlalchemy.create_engine()`.
  - Supports configuring reflection behavior and extraction for the SQLAlchemy backend.
  - Includes comprehensive tests:
    - invalid kwargs propagation  
    - reflection failure  
    - SQLite timeout enforcement  
    - interaction with backend_kwargs when using non-SQLAlchemy backends  
    - SQLAlchemy echo logging behavior

- **Destination (SQLAlchemy)**
  - Adds missing documentation for existing `engine_args`.
  - Adds tests to verify correct behavior for both in-memory and file-based SQLite destinations using `engine_args`.

- **Documentation**
  - Adds new docs for source-side `engine_kwargs`.
  - Adds new docs for destination-side `engine_args`.
  - Clarifies that source and destination engines are separate and configured independently.

**Question:** To reduce user confusion, should the source parameter be named `engine_kwargs` (current) or unified as `engine_args` (matching the destination)?

### Related Issues

- Resolves #3187

### DIFF
--- a/dlt/destinations/impl/sqlalchemy/configuration.py
+++ b/dlt/destinations/impl/sqlalchemy/configuration.py
@@ -30,6 +30,7 @@ class SqlalchemyCredentials(ConnectionStringCredentials):
         self, connection_string: Optional[Union[str, Dict[str, Any], "Engine"]] = None
     ) -> None:
         super().__init__(connection_string)  # type: ignore[arg-type]
+        self._conn_lock = threading.RLock()
 
     def _resolve_engine_kwargs(self) -> Dict[str, Any]:
         if self.engine_kwargs and self.engine_args:
@@ -61,11 +62,6 @@ class SqlalchemyCredentials(ConnectionStringCredentials):
             super().parse_native_representation(native_value)
 
     def borrow_conn(self) -> "Connection":
-        if getattr(self, "_conn_owner", None) is False:
-            return self.engine.connect()
-
-        if not hasattr(self, "_conn_lock"):
-            self._conn_lock = threading.Lock()
 
         # obtain a lock because we have refcount concurrency
         with self._conn_lock:
@@ -83,9 +79,6 @@ class SqlalchemyCredentials(ConnectionStringCredentials):
             raise
 
     def return_conn(self, borrowed_conn: "Connection") -> None:
-        if getattr(self, "_conn_owner", None) is False:
-            borrowed_conn.close()
-            return
         # close the borrowed conn
         with self._conn_lock:
             borrowed_conn.close()
@@ -112,13 +105,13 @@ class SqlalchemyCredentials(ConnectionStringCredentials):
         import sqlalchemy as sa
 
         # get existing or open and set new engine
-        engine_kwargs = self._resolve_engine_kwargs()
         self._engine = getattr(
             self,
             "_engine",
             None,
         )
         if self._engine is None:
+            engine_kwargs = self._resolve_engine_kwargs()
             self._engine = sa.create_engine(
                 self.to_url().render_as_string(hide_password=False), **engine_kwargs
             )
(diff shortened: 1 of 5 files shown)


dlt/destinations/impl/sqlalchemy/configuration.py
```diff
@@ -53,6 +53,9 @@ def borrow_conn(self) -> "Connection":
         return engine_.connect()
 
     def return_conn(self, borrowed_conn: "Connection") -> None:
+        if getattr(self, "_conn_owner", None) is False:
```
rudolfix: it looks like passing external engine here never worked before... we need to test it well

dlt/sources/sql_database/__init__.py
```diff
@@ -108,8 +112,12 @@ def sql_database(
     else:
         reflection_level = reflection_level or "minimal"
 
-    # set up alchemy engine
-    engine = engine_from_credentials(credentials)
+    engine_kwargs = engine_kwargs or {}
+    engine = engine_from_credentials(
+        credentials,
+        False,
```
rudolfix: would be good to pass `may_dispose_after_use` here

tests/load/sqlalchemy/test_sqlite_destination.py
```diff
@@ -0,0 +1,70 @@
...
+        destination=sqlalchemy(engine),
+        dataset_name='main'
+    )
+
+    info = pipeline.run(
+        some_data(),
+        table_name="inmemory_table"
+    )
+
+    assert_load_info(info)
```
rudolfix: I'd check if engine is not disposed after use in the pipeline

tests/sources/sql_database/test_engine_kwargs.py
```diff
@@ -0,0 +1,141 @@
...
+        sql_table(
+            table="dummy",
+            credentials=credentials,
+            engine_kwargs={"invalid_argument_name": True},
+        )
+
+
+def test_engine_kwargs_timeout_is_honored_sqlite():
+    """SQLite timeout in engine_kwargs must be honored."""
+    fd, db_path = tempfile.mkstemp(suffix=".db")
```
rudolfix: we have a test storage which is automatically cleared.

dlt/destinations/impl/sqlalchemy/configuration.py
tetelio: i added that passsing built engine is not disposed after use

rudolfix (dismissed):
LGTM

rudolfix (approved):
LGTM!

rudolfix (commented):
LGTM!

PR #3421, opened 2025-12-02
Fix/3159 pydantic model incorrect serialization

### LINKED ISSUE #3159: Interaction between values returned by a resource and configuring a pydantic schema
### dlt version

1.16.0

### Describe the problem

I have defined a resource that yields instances of a pydantic model. That was working perfectly until I configured the model as the schema using `dlt.resource(..., columns=Model)`. When doing that, the instances yielded from that resource are automatically transformed into a `dict`. Is that expected? I suppose, it could be a side-effect of how your validation code is defined.

This is somewhat annoying as I then need to `model_validate` again on transformers using that resource. That was not the case before.

### Expected behavior

How about skipping any validation steps if the value returned is already an instance of the pydantic model?

### Steps to reproduce

I currently don't have a minimal, reproducible example, but I can create one if truly desired.

### Operating system

Linux

### Runtime environment

Local

### Python version

3.12

### dlt data source

not relevant

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

tetelio:
### Description

This PR adds an explicit `return_validated_models` option to the Pydantic `DltConfig` to control whether validated Pydantic model instances are returned during extraction instead of being converted to dictionaries.

**Problem:**
When using a Pydantic model as a validator with `@dlt.resource(columns=Model)`:
1. dlt validates the data using the Pydantic model
2. Always converts validated objects to dictionaries before passing them downstream
3. Transformers receive dicts instead of model instances
4. Users must manually reconstruct Pydantic objects in transformers

**Solution:**
This PR adds a new `DltConfig` option: `return_validated_models`. When set to `True`, the Pydantic validator returns validated model instances instead of converting them to dictionaries. This preserves backward compatibility (defaulting to `False`) while giving users explicit control over the output format.

**Usage:**
```python
from typing import ClassVar
from pydantic import BaseModel
from dlt.common.libs.pydantic import DltConfig
import dlt

class User(BaseModel):
    id: int
    name: str
    dlt_config: ClassVar[DltConfig] = {"return_validated_models": True}

@dlt.resource(columns=User)
def users():
    yield {"id": 1, "name": "Alice"}
    yield User(id=2, name="Bob")

@dlt.transformer(data_from=users)
def process_users(user: User):
    # user is a Pydantic model instance, not a dict
    print(f"Processing user: {user.name}")
    yield user
```

**Changes:**
- Added `return_validated_models` option to `DltConfig` in `dlt/common/libs/pydantic.py`
- Updated `PydanticValidator` in `dlt/extract/validation.py` to check the flag and return model instances when enabled
- Added unit tests for the extract validator and pipeline integration tests
- Updated documentation in `docs/website/docs/general-usage/resource.md`

### Related Issues

- Fixes #3159

---------------------


dlt/extract/validation.py
```diff
@@ -40,18 +40,29 @@ def __call__(self, item: TDataItems, meta: Any = None) -> TDataItems:
...
-                    self.table_name, self.list_model, item, self.column_mode, self.data_mode
-                )
-            ]
-        item = validate_and_filter_item(
+            input_is_model = bool(item) and isinstance(item[0], PydanticBaseModel)
+            validated_list = validate_and_filter_items(
+                self.table_name, self.list_model, item, self.column_mode, self.data_mode
+            )
+            if input_is_model:
+                input_fields = set(item[0].__class__.model_fields.keys())
```
rudolfix: this is a good catch. the model that is used to validate may be different from model instances passed to pipeline. even if the model is the same there are some weird cases or revalidation was requested. so we cannot just skip validation here.

still - doing this check and going back to dict is IMO not intuitive and expensive. we also probe just a first element and assume that the list is of uniform item types.

overall I think we need something simpler here

rudolfix (changes requested):
trying to keep the same type of input and output automatically is pretty hard. we have many weird edge cases here. even if model instance is validated with the same model, full revalidation is possible and some items may got filtered out from the list. and no matter what we do, we will not be backward compatible (always dict after validation).

my take is to simplify it and extend `DltConfig` to include a flag like `return_validated_models` if set to True, we just return the validator output, without going to dict.

this config is applied to a model, see other options. this also needs to be documented

PR #3426, opened 2025-12-03
Fix: digest256_tar_stream fixed

anuunchin:
This PR adjusts the digest256_tar_stream so that it accepts a filtering function. The related `PackageBuilder.write_package_to_stream` now excludes the manifest.yaml.

### DIFF
--- a/dlt/common/utils.py
+++ b/dlt/common/utils.py
@@ -1,5 +1,5 @@
 import re
-from typing import TYPE_CHECKING, Any, BinaryIO, Literal
+from typing import TYPE_CHECKING, Any, BinaryIO, IO
 import os
 from pathlib import Path
 import sys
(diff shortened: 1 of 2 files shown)


dlt/common/utils.py
```diff
@@ -131,33 +131,50 @@ def digest256_file_stream(stream: BinaryIO, chunk_size: int = 4096) -> str:
...
                 f = tar.extractfile(member)
                 if f:
                     while chunk := f.read(chunk_size):
                         hash_obj.update(chunk)
 
     digest = hash_obj.digest()
-    return base64.b64encode(digest).decode("ascii")
+    content_hash = base64.b64encode(digest).decode("ascii")
+
+    return content_hash, file_names
```
anuunchin, line 177: We can also just keep it the way it was, but I don't like the fact that we'd have to iterate over tar members again, just to get the file names in the runtime 👀

PR #3430, opened 2025-12-03
refactors platform artifacts

rudolfix:
### Description
see commit log

### DIFF
--- a/dlt/_workspace/_workspace_context.py
+++ b/dlt/_workspace/_workspace_context.py
@@ -98,6 +98,14 @@ class WorkspaceRunContext(ProfilesRunContext):
         # this also resolves workspace config if necessary
         initialize_runtime(self.name, self.config.runtime)
 
+        # if on runtime, add additional tracker
+        if self.runtime_config.run_id:
+            from dlt._workspace.helpers.runtime import runtime_artifacts
+            from dlt.pipeline import trace
+
+            if runtime_artifacts not in trace.TRACKING_MODULES:
+                trace.TRACKING_MODULES.append(runtime_artifacts)
+
     @property
     def runtime_config(self) -> WorkspaceRuntimeConfiguration:
         return self.config.runtime
--- a/dlt/pipeline/__init__.py
+++ b/dlt/pipeline/__init__.py
@@ -1,7 +1,6 @@
 from typing import Sequence, Type, cast, overload, Optional
 from typing_extensions import TypeVar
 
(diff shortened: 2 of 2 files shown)


dlt/_workspace/helpers/runtime/runtime_artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+from dlt._workspace.run_context import DEFAULT_WORKSPACE_WORKING_FOLDER
+from dlt._workspace._workspace_context import WorkspaceRunContext
+
+
+@configspec
+class RuntimeArtifactsConfiguration(BaseConfiguration):
+    artifacts: FilesystemConfiguration = None
+
+
+def sync_from_runtime() -> None:
```
sh-rp, line 26: Two notes here:
* In the old implementation, we sometimes got some kind of "e-tag expired" exception. I think what happens is, that you do a ls of existing blobs and e-tags are created, but if in the meantime a pipeline uploads new stuff, they expire and the e-tags invalidate, although the blob-path still exists. Afaik there is a setting to supress this warning.
* This sync happens everytime you open the dashboard. Ideally we only sync files that have changed since the last sync, often there will be no changes at all if a user switches tabs a lot in the runtime web-ui. I did not have time to implement anything like this and it is not top prio, but we should have it at some point.

dlt/_workspace/helpers/runtime/runtime_artifacts.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    Send the full trace pickled to the runtime bucket
+    """
+    pickled_trace = pickle.dumps(trace)
+    _write_to_bucket(
+        fs,
+        bucket_url,
+        pipeline.pipeline_name,
+        [
+            "trace.pickle",
```
sh-rp, line 111: As discussed, we need to transition this to some kind of datalake or at least a serialized dictionary format sooner than later.

dlt/_workspace/helpers/dashboard/dlt_dashboard.py
```diff
@@ -1002,8 +1002,10 @@ def utils_discover_pipelines(
     # sync from runtime if enabled
     _tmp_config = utils.resolve_dashboard_config(None)
     if _tmp_config.sync_from_runtime:
```
sh-rp, line 1004: I don't think we really need this var. If the credentials are available, we should sync. But this is not important rn.

PR #3431, opened 2025-12-04
Fix: merge_columns not removing compound properties that should be removed

### LINKED ISSUE #3040: Delete-Insert does add extra-field as condition
### dlt version

1.21.1

### Describe the problem

I have some weird behaviour with `{"disposition": "merge",  "strategy": "delete-insert"}`

```
@dlt.resource(
        name="time_off",
        write_disposition={
            "disposition": "merge",
            "strategy": "delete-insert",
            "deduplicated": False,
        },
        merge_key=["month_key"],
        primary_key=(),
    )
```
For some reason the DELETE statement (BigQuery) still adds another field into the condition and not just the merge key
How can that be?

Image

I also dropped the pipeline state several times

### Expected behavior

I would expect that only the month_key is added to the DELETE-statement

### Steps to reproduce

use 
```
        write_disposition={
            "disposition": "merge",
            "strategy": "delete-insert",
            "deduplicated": False,
        },
```
And check the DELETE statement

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

Custom API

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

_No response_

anuunchin:
This PR adjusts the `def merge_columns()` function that previously handled things only additively. This caused an issue where compound properties, such as the `merge_key`, were being replaced with new values. 

Additionally, a test is added. 

> This PR implements some of the things mentioned in [this related PR](https://github.com/dlt-hub/dlt/pull/3361). 

Resolved #3040

### DIFF
--- a/docs/website/docs/general-usage/snippets/schema-snippets.py
+++ b/docs/website/docs/general-usage/snippets/schema-snippets.py
@@ -65,6 +65,7 @@ def test_nested_hints_primary_key() -> None:
     assert row_count == [("customers", 3), ("customers__purchases", 3)]
     # @@@DLT_SNIPPET_END nested_hints_primary_key
 
+
 def test_compound_hints_direct_key_precedence() -> None:
     """Demonstrates that direct key hints take precedence over column level hints."""
     # @@@DLT_SNIPPET_START compound_hints_direct_key_precedence
@@ -84,10 +85,18 @@ def test_compound_hints_direct_key_precedence() -> None:
 
     pipeline.run(my_resource)
     # @@@DLT_SNIPPET_END compound_hints_direct_key_precedence
-   
+
     # Verify: col_1 should have primary_key (direct key takes precedence)
-    assert pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get("primary_key") is True
-    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get("primary_key")
+    assert (
+        pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get(
+            "primary_key"
+        )
+        is True
+    )
+    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get(
+        "primary_key"
+    )
+
 
 def test_compound_hints_empty_direct_key_precedence() -> None:
     """Demonstrates that direct key hints take precedence over column level hints."""
@@ -96,6 +105,7 @@ def test_compound_hints_empty_direct_key_precedence() -> None:
         destination="duckdb",
         dataset_name="my_data",
     )
+
     # @@@DLT_SNIPPET_START test_compound_hints_empty_direct_key_precedence
     @dlt.resource(
         name="my_table",
@@ -107,10 +117,15 @@ def test_compound_hints_empty_direct_key_precedence() -> None:
 
     pipeline.run(my_resource)
     # @@@DLT_SNIPPET_END test_compound_hints_empty_direct_key_precedence
-   
+
     # Verify: both should not have primary_key (direct key takes precedence)
-    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get("primary_key")
-    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get("primary_key")
+    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get(
+        "primary_key"
+    )
+    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get(
+        "primary_key"
+    )
+
 
 def test_compound_hints_direct_key_precedence_apply_hints() -> None:
     """Demonstrates that direct key hints take precedence over column level hints in apply_hints."""
@@ -119,6 +134,7 @@ def test_compound_hints_direct_key_precedence_apply_hints() -> None:
         destination="duckdb",
         dataset_name="my_data",
     )
+
     # @@@DLT_SNIPPET_START test_compound_hints_direct_key_precedence_apply_hints
     @dlt.resource(
         name="my_table",
@@ -130,10 +146,16 @@ def test_compound_hints_direct_key_precedence_apply_hints() -> None:
 
     pipeline.run(my_resource)
     # @@@DLT_SNIPPET_END test_compound_hints_direct_key_precedence_apply_hints
-   
+
     # Verify: both should not have primary_key (direct key takes precedence)
-    assert pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get("merge_key") is True
-    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get("merge_key")
+    assert (
+        pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get("merge_key")
+        is True
+    )
+    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get(
+        "merge_key"
+    )
+
 
 def test_compound_hints_replace_previous_compound_props() -> None:
     """Demonstrate what new compound prop hints on existing resource replace previous ones."""
@@ -142,6 +164,7 @@ def test_compound_hints_replace_previous_compound_props() -> None:
         destination="duckdb",
         dataset_name="my_data",
     )
+
     # @@@DLT_SNIPPET_START test_compound_hints_replace_previous_compound_props
     @dlt.resource(
         name="my_table",
@@ -149,22 +172,28 @@ def test_compound_hints_replace_previous_compound_props() -> None:
     )
     def my_resource():
         yield {"col_1": 1, "col_2": 2}
-    
+
     pipeline.run(my_resource)
 
-    @dlt.resource(
+    @dlt.resource(  # type: ignore[no-redef]
         name="my_table",
         columns={"col_1": {"partition": True}},
     )
     def my_resource():
         yield {"col_1": 1, "col_2": 2}
-    
+
     pipeline.run(my_resource)
     # @@@DLT_SNIPPET_END test_compound_hints_replace_previous_compound_props
 
     # Verify: only col_1 should have partition property
-    assert pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get("partition") is True
-    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get("partition")
+    assert (
+        pipeline.default_schema.tables["my_table"]["columns"]["col_1"].get("partition")
+        is True
+    )
+    assert not pipeline.default_schema.tables["my_table"]["columns"]["col_2"].get(
+        "partition"
+    )
+
 
 def test_empty_value_key_hints_do_not_replace_previous_hints() -> None:
     """Demonstrate that new empty value key hint hints on existing resource do nothing."""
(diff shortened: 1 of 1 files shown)


dlt/common/schema/typing.py
```diff
@@ -120,7 +121,7 @@ class TColumnPropInfo(NamedTuple):
     TColumnPropInfo("primary_key", (False, None)),
     TColumnPropInfo("sort", (False, None)),
     TColumnPropInfo("unique", (False, None)),
-    TColumnPropInfo("merge_key", (False, None)),
+    TColumnPropInfo("merge_key", (False, None), False, True),
```
rudolfix, line 128: partition, primary_key and cluster are also compound

dlt/common/schema/utils.py
```diff
@@ -154,6 +154,37 @@ def has_default_column_prop_value(prop: str, value: Any) -> bool:
...
+        return ColumnPropInfos[prop].compound is True
+    # for any unknown property
+    return False
+
+
+def remove_compound_props(
+    columns: TTableSchemaColumns, compound_props: List[str]
+) -> TTableSchemaColumns:
+    """Removes compound properties from all columns in place."""
+    non_compound = [prop for prop in compound_props if not is_compound_prop(prop)]
```
rudolfix: we are too defensive. assume only valid compound prop is passed

dlt/common/schema/utils.py
```diff
@@ -154,6 +154,37 @@ def has_default_column_prop_value(prop: str, value: Any) -> bool:
...
+    if non_compound:
+        raise ValueError(
+            f"The following properties are not compound properties: {', '.join(non_compound)}"
+        )
+
+    for column in columns.values():
+        is_nullable = is_nullable_column(column)
+        for prop in compound_props:
+            if not is_nullable:
```
rudolfix: if someone sets `nullable`: then it is applied after this check so s/he will always see this warning (you remove compound hints, then merge new hints...) my take is to remove it

dlt/common/schema/utils.py
```diff
@@ -436,28 +467,30 @@ def merge_column(
 def merge_columns(
     columns_a: TTableSchemaColumns,
     columns_b: TTableSchemaColumns,
-    merge_columns: bool = False,
-    columns_partial: bool = True,
+    keep_compound_props: bool = False,
```
rudolfix: we do not use this flag, we can add it when it is needed

dlt/common/schema/utils.py
```diff
@@ -436,28 +467,30 @@ def merge_column(
...
+    * New columns from `columns_b` are added to `columns_a`
+    * Existing columns that appear in both sets are merged using `merge_column`
+    * If `preserve_compound_props` is False, compound properties (like `primary_key`, `merge_key`) found in `columns_b` are first removed from all columns in `columns_a`
+    * If `preserve_compound_props` is True, compound properties in `columns_a` are kept, allowing incremental merging
+    * Incomplete columns in `columns_a` that are complete in `columns_b` are removed and re-added to preserve column order
     """
-    if columns_partial is False:
-        raise NotImplementedError("Using `merge_columns()` requires `columns_partial=False`")
+    if keep_compound_props is False:
+        compound_props: List[str] = []
```
rudolfix: it is better to use `Set` here

dlt/common/schema/utils.py
```diff
@@ -154,6 +154,37 @@ def has_default_column_prop_value(prop: str, value: Any) -> bool:
...
+
+    for column in columns.values():
+        is_nullable = is_nullable_column(column)
+        for prop in compound_props:
+            if not is_nullable:
+                logger.warning(
+                    f"Removing compound property '{prop}' from column '{column['name']}', but"
+                    " 'unique' constraint remains set to True."
+                )
+            column.pop(prop, None)  # type: ignore[misc]
```
rudolfix, line 185: note: this removes False values as well

tests/pipeline/test_schema_updates.py
```diff
@@ -71,3 +71,29 @@ def resource():
         },
         "max_nesting": 50,
     }
+
+
+def test_changing_merge_key_between_runs() -> None:
```
rudolfix: we also need to test `apply_hints`. what happens if:
1. use `columns` with `merge_key` and `merge_key` argument together (AFAIK they'll get merged, not replaced, I think explicit merge_key should replace what is in columns)
2. use apply_hints twice: do we replace previous settings for `merge_key`? or additive merge?
3. what about merge_key=()?

btw:
```py
@staticmethod
    def _merge_keys(dict_: TResourceHints) -> None:
        """Merges primary and merge keys into columns in place."""

        if "primary_key" in dict_:
            DltResourceHints._merge_key("primary_key", dict_.pop("primary_key"), dict_)  # type: ignore
        if "merge_key" in dict_:
            DltResourceHints._merge_key("merge_key", dict_.pop("merge_key"), dict_)  # type: ignore
```
we could replace it by merge_columns: first generate columns for both keys (mind the preserved nullability in ``) and then we use `merge_columns`. in case of `len(merge_key)==0` we remove compund hints explicitly (because we do not implement this in `merge_columns` (note: this will only fix behavior withing `apply_hints`. we still are not able to remove `merge_key` in the final resource hints)

rudolfix (dismissed):
this will fix a lot of nasty cases! we need more test and IMO some edge cases in `apply_hints` will come out

tests/pipeline/test_schema_updates.py
anuunchin: I've opted out from using `merge_columns` here for now to keep things simpler

dlt/normalize/items_normalizers.py
```diff
@@ -545,7 +545,9 @@ def _normalize_chunk(
                             continue
                         # theres a new table or new columns in existing table
                         # update schema and save the change
-                        schema.update_table(partial_table, normalize_identifiers=False)
+                        schema.update_table(
+                            partial_table, normalize_identifiers=False, merge_compound_props=True
+                        )
```
anuunchin: We still need to keep the possibility of merging compound keys here. Example scenario: when nested hints with referenced keys are passed, referenced keys are overwritten due to how  the underlying `diff_table` works

TLDR: merge_columns should remove compound columns before it reaches the normalization stage, everything that happens after should work as before imo 👀

anuunchin: Although it's not ideal to be passing this flag 3 levels down...

tests/extract/test_sources.py
```diff
@@ -1973,7 +1976,7 @@ def empty_gen():
     table_b = empty.compute_table_schema(meta=TableNameMeta("table_b"))
     assert table_b["name"] == "table_b"
     assert table_b["write_disposition"] == "merge"
-    assert len(table_b["columns"]) == 0
+    assert table_b["columns"] == {"": {"name": "", "nullable": False, "primary_key": True}}
```
anuunchin: Empty value key should survive compute_table_schema because it's called before we merge columns - there's no other way to use this empty value key column otherwise (unless we take a completely different approach)

tests/pipeline/test_schema_updates.py
```diff
@@ -71,3 +73,298 @@ def resource():
         },
         "max_nesting": 50,
     }
+
```
anuunchin, line 83: have kept some repeatedly used code snippets as they are for demonstration purposes

dlt/normalize/items_normalizers.py
anuunchin: PS: this case can be proven with the `test_nested_hints_primary_key` test - i.e. it won't pass if we don't merge

anuunchin:
This will need more tests for all related places, as merge_columns affects many more things than I've anticipated - but the  current version is worth a look 

PS: I scrapped the authoritative empty column logic that should remove all keys in the last commit

@rudolfix

dlt/common/schema/utils.py
```diff
@@ -436,28 +458,31 @@ def merge_column(
...
+
+    Args:
+        merge_compound_props: If set to True, compound properties from columns_b are merged to columns_a,
+            instead of replacing them.
+
+    * Incomplete columns in `columns_a` that are complete in `columns_b` are removed and re-added to preserve column order
     """
-    if columns_partial is False:
-        raise NotImplementedError("Using `merge_columns()` requires `columns_partial=False`")
+    compound_props: set[str] = set()
```
rudolfix: you need this only if you remove compound props, no need to always compute

dlt/common/schema/utils.py
```diff
@@ -436,28 +458,31 @@ def merge_column(
...
+    for column_b in columns_b.values():
+        compound_props.update(prop for prop in column_b if is_compound_prop(prop))
+    if compound_props and not merge_compound_props:
+        remove_compound_props(columns=columns_a, compound_props=compound_props)
 
     # remove incomplete columns in table that are complete in diff table
     for col_name, column_b in columns_b.items():
+        if not col_name:
```
rudolfix: does it ever happen? should not be possible: without a name column cannot be a part of column dict (key is name)

dlt/extract/hints.py
```diff
@@ -488,20 +490,14 @@ def apply_hints(
                     columns = ensure_table_schema_columns(columns)
                     # this updates all columns with defaults
                     assert isinstance(t["columns"], dict)
-                    t["columns"] = merge_columns(t["columns"], columns, merge_columns=True)
+                    t["columns"] = merge_columns(t["columns"], columns)
```
rudolfix, line 491: here we just update columns via `apply_hints`. those changes are accumulated and then applied all together to existing schema. my take: keep merging here. to replace people can use primary or merge keys or adapters (btw. I'm pretty sure we have adapters that do that incorrectly without replace)

dlt/extract/extractors.py
```diff
@@ -510,9 +510,7 @@ def _compute_tables(
                         " schema and data were unmodified. It is up to destination to coerce the"
                         " differences when loading. Change log level to INFO for more details."
                     )
-                utils.merge_columns(
-                    arrow_table["columns"], computed_table["columns"], merge_columns=True
-                )
+                utils.merge_columns(arrow_table["columns"], computed_table["columns"])
```
rudolfix, line 520: good!

dlt/normalize/items_normalizers.py
rudolfix: yeah - this is pretty clear why

dlt/common/schema/utils.py
anuunchin: It's a remnant of the version where I hade empty values removing already existing keys from schema

anuunchin: removed ✅

docs/website/docs/general-usage/snippets/schema-snippets.py
```diff
@@ -2,7 +2,7 @@
 from dlt.common import Decimal
 
 
-def nested_hints_primary_key() -> None:
```
anuunchin, line 5: This was not tested

tests/load/lancedb/test_pipeline.py
```diff
@@ -70,6 +70,51 @@ def some_data() -> Generator[DictStrStr, Any, None]:
     assert some_data.compute_table_schema()["columns"]["content"]["merge_key"] is True
 
 
+def test_changing_merge_key() -> None:
```
anuunchin, line 73: lancedb is the only adapter that explicitly deals with a compound key property and so I just wanted to test the logic here

anuunchin: similarly, bigquery adapter already had replacement logic for cluster and partition

anuunchin: The only adapter that will merge is databricks and I'd like to move that to a separate PR 👀

PR #3434, opened 2025-12-04
feat: S3 Tables support for `athena` destination

### LINKED ISSUE #3268: Missing AWS S3 Tables support in Athena destination
### Feature description

Hi there, 

I'm trying to use the Athena destination with S3 tables but I ran into many difficulties. Want to report back some of the issues I know.

Setup: 

I want to load some data into AWS Iceberg-backed data lake, and specifically, using S3 table buckets. 

1. go to AWS console and create a S3 table bucket, and select "enable AWS analytics integration". This option will automatically create Iceberg catalogs in Glue and LakeFormation with the right permissions and all that. 
3. Go to Athena on the AWS console, make sure you can see the new catalog. It should be named as `s3tablescatalog/<your bucket>`.

In DLT, I tried 3 ways of writing to Athena, and I ran into problems with both.

1. Use the Athena destination
2. Use the SqlAlchemy destination but with a credential like "awsathena://...". The PyAthena library adds support of this dialect to SqlAlchemy.
3. Use the DuckDB destination where I create my own DuckDB connections, prepare the connection with CREATE SECRET, ATTACH, and other stuff before passing it to DLT.

For approach one, 

1. `dlt.destinations.imple.athena.athena.AthenaClient._get_table_update_sql` always inserts the "LOCATION" clause. However, for S3 tables, one should NOT specify a location. The S3 tables bucket is managed in the sense that location is automatically determined.
2. `dlt.destinations.impl.athena.configuration.AthenaClientConfiguration.query_result_bucket` should be optional. This is because AWS Athena allows you to specify a Workgroup that has a managed query result location. In PyAthena, if you specify such a workgroup, you no longer need to specify a query result bucket. So I think this should be marked optional. 
3. Not a bug but need to point out, to use S3 tables, one needs to explicitly set the `aws_data_catalog=s3tablscatalog/<your bucket>`. 
4. After patching these, the parquet files does appear in the S3 table bucket if you do ListObjects. But it is not registered as tables for some reason. I can't see these tables in Athena or in S3 Table bucket webpage. 

For approach two, 

1. DLT tries to create tables with a table name that has "_" as the prefix, like "_dlt_load". It seems that Athena is not happy about these kinds of names. 

For approach three, 

1. Same problem, DLT tries to create tables with "_" prefix that Athena don't like. 
2. I also ran into this issue https://github.com/duckdb/duckdb-httpfs/issues/154 because DLT tries to use the default database and schema and I need to do `USE ...` to prepare the connections.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Use DLT to load data into AWS S3 Tables

### Proposed solution

1. strengthen the support of S3 tables for the Athena destination 
2. allow more customizations for the DuckDB destination so that it works better with DuckDB's ATTACH capabilities
3. allow easy customizations of DLT's state table names to be more compatible. 

### Related issues

_No response_

jorritsandbrink:
### Description
This PR adds support for S3 Tables with the `athena` destination. 

### Related Issues

- Closes #3268


tests/load/test_sql_client.py
```diff
@@ -161,16 +158,16 @@ def test_create_drop_dataset(naming: str, client: SqlJobClientBase) -> None:
     assert client.is_storage_initialized() is True
     client.sql_client.drop_dataset()
     assert client.is_storage_initialized() is False
-    with pytest.raises(DatabaseUndefinedRelation):
```
jorritsandbrink, line 164: Removed because this seems to be redundant with `test_database_exceptions`.

jorritsandbrink:
@rudolfix Could you review this PR?

Some notes:
- most changes are in the test code, not in the library code
- all `athena` queries are now fully qualified including the catalog name
- user chooses S3 Tables by setting `aws_data_catalog` to an S3 Tables Catalog
- when user sets an S3 Tables Catalog:
  - `naming_convention` is automatically set to new `s3_tables` naming convention
  - production (non-staging) and staging tables are registered in separate catalogs
  - all production tables are Iceberg tables (including the pipeline state table, as opposed to when using Iceberg without S3 Tables)

rudolfix (approved):
@jorritsandbrink this is really good and I do not see any problems in this PR! all the extensions and generalizations, even if they go pretty deep, make sense. also thanks for fixing all the tests.

I'll add full ci label to run all athena tests.

there's one just thing that I do not understand though: why we need `staging_aws_data_catalog`? is there some kind of limitation of s3 tables - can't do COPY from s3 tables to s3 tables?

rudolfix:
> there's one just thing that I do not understand though: why we need staging_aws_data_catalog? is there some kind of limitation of s3 tables - can't do COPY from s3 tables to s3 tables?

OK! staging tables are regular athena tables over parquet files... so indeed s3 tables catalog can't work here!

PR #3436, opened 2025-12-05
pyarrow: respect resource hints before extract 

### LINKED ISSUE #2788: Schema hints not taken into account by PyArrow?
### dlt version

1.12.0

### Describe the problem

Hello,

I've been struggling with the following error while using the PyArrow engine. The SQLAlchemy works fine but is just painfully slow compared to the PyArrow one, when it works.

With PyArrow, the run starts out with the following INFO-level logs:

> extractors.py|_compute_tables:477|In resource: picks, when merging arrow schema on column short_term_booking_price_with_service_fee. The hint scale value 18 defined in resource will overwrite arrow hint with value 9.

Then this (note that it's telling me to change to log level INFO, which it already is set at):

> 2025-06-18 19:34:09,767|[WARNING]|76065|8591068288|dlt|extractors.py|_compute_tables:484|In resource: picks, when merging arrow schema with dlt schema, several column hints were different. dlt schema hints were kept and arrow schema and data were unmodified. It is up to destination to coerce the differences when loading. Change log level to INFO for more details.

To finally fail like so:

> In processing pipe `picks`: extraction of resource `picks` in `generator` `table_rows` caused an exception: Conversion to arrow failed for field `short_term_booking_price_with_service_fee` with dlt hint `data_type=decimal` and `inferred_arrow_type=decimal128(38, 9)` Insufficient decimal precision. Consider setting `precision` and `scale` hints: https://dlthub.com/docs/general-usage/schema/#tables-and-columns

Using a breakpoint right before `pipeline.run`, I double checked the config:
```python
(Pdb) sql_source.picks.columns['short_term_booking_price_with_service_fee']
{'name': 'short_term_booking_price_with_service_fee', 'nullable': True, 'data_type': 'decimal', 'precision': 38, 'scale': 18}
```

Other tables work fine but somehow my source database has some columns that have up to 12 or 13 digits, so these tables all fail.

I've tried with a duckdb and Snowflake destination, both with a clean new pipeline (deleted the ~/.dlt files etc) to no avail.

Any help would be much appreciated, thank you!!

### Expected behavior

The hints should be taken into account and prevent a casting failure from arrow

### Steps to reproduce

Hopefully the bits in the description are enough… If not I'll try to export a limited version of my database/table

### Operating system

macOS

### Runtime environment

Local

### Python version

3.10

### dlt data source

Postgres

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

djudjuu:
Here's a PR summary:

---

### Problem

When using `sql_table()` with `backend="pyarrow"`, user-provided column hints via `apply_hints()` were not being applied before data extraction. PyArrow would use the reflected database types, ignoring user overrides like changing `decimal` to `double`.

```python
table = sql_table(..., backend="pyarrow")
table.apply_hints(columns={"numeric_col": {"data_type": "double"}})  # Was ignored!
```

### Solution

Merge user-provided column hints with SQL reflection hints **before** passing them to the `TableLoader`. This ensures PyArrow uses the user-specified types when extracting data.
- warn if backend is pyarrow and user provides dynamic hints (callable) as those can't be used. in that case fallback to table reflection 

### Changes

**`dlt/sources/sql_database/helpers.py`:**
- Added logic to merge `resource.columns` hints with reflection hints before extraction, *both for defer_reflect=True/False*
- Added warning when callable (dynamic) hints are provided with pyarrow backend
-

**`tests/load/sources/sql_database/test_sql_database_source.py`:**
- Added `test_pyarrow_applies_hints_before_extract` test
- Parameterized for `defer_reflect`
- verify if pipeline-schema and pyarrow-schema are in sync (and loader file format is applied)

### follow up issues ??
-**primary_key** hint no merged. in case of deferred-reflect the current fix only merges `column` hints from resource and there is code that 

```
# ... case deferred-reflect
... 
        # set the primary_key in the incremental
        if incremental and incremental.primary_key is None:
            primary_key = hints["primary_key"]
            if primary_key is not None:
                incremental.primary_key = primary_key
... # << if user supplied primary key hint its not being used
```

- **Callable hints**: Dynamic column hints (callables) cannot be merged statically and will be applied later during extraction. A warning is logged.

- **Pydantic models**: When used as columns, Pydantic creates a validator that has compatibility issues with PyArrow's data format (separate from this fix). <<< ai commment, i need to investigate still

---


dlt/sources/sql_database/helpers.py
```diff
@@ -348,7 +350,17 @@ def table_rows(
...
     try:
-        limit = dlt.current.resource().limit
+        resource = dlt.current.resource()
+        resource_columns = resource.columns
+        if resource_columns and hints["columns"]:
+            print("hints from reflection: ", hints["columns"])
+            print("hints from resource: ", resource_columns)
+            resource_columns_as_hints = ensure_table_schema_columns(resource_columns)
+            # print("resource_columns_as_hints: ", resource_columns_as_hints)
+            hints["columns"] = merge_columns(hints["columns"], resource_columns_as_hints)
```
djudjuu: even without this line my tests pass...so somehow the schema is reflected...

djudjuu: what happened: because the table wasnt there yet, duckdb created it using the schema which contained the hint and then coerced the pyarrow into it (or coerced it on read?)
but when checking the actual datatype of the arrow loadpackage then there is a difference made by these lines

dlt/sources/sql_database/helpers.py
```diff
@@ -348,7 +350,19 @@ def table_rows(
 
     limit = None
     try:
-        limit = dlt.current.resource().limit
+        resource = dlt.current.resource()
+        resource_columns = resource.columns
+        # Only merge if both resource_columns and hints["columns"] are dicts (not callables)
```
djudjuu: :thinking: investigating...

dlt/sources/sql_database/helpers.py
```diff
@@ -348,7 +350,20 @@ def table_rows(
...
+        resource_columns = resource.columns
+        if resource_columns and hints["columns"]:
+            if callable(resource_columns):
+                if backend == "pyarrow":
+                    logger.warning(
+                        f"Dynamic column hints for '{table.name}' cannot be applied with pyarrow "
+                        "backend. Use static hints (dict/list) to override reflected types."
+                    )
+            else:
+                resource_columns_as_hints = ensure_table_schema_columns(resource_columns)
```
rudolfix: resource columns are always `TTableSchemaColumns`. is there a reason you use `ensure_table_schema_columns`?

tests/load/sources/sql_database/test_sql_database_source.py
```diff
@@ -84,6 +87,79 @@ def convert_time_to_us(table):
...
+        }
+    elif columns_hint_type == "sequence":
+        columns_hint = [
+            {"name": "numeric_col", "data_type": "double"},
+        ]
+    else:
+        raise ValueError(f"Unknown columns_hint_type: {columns_hint_type}")
+
+    table = sql_table(
+        credentials=postgres_db.credentials,
```
rudolfix, line 104: please parametrize test by `defer_table_reflect`. when it is enabled, your test should fail (if not we need to find the cause).

this option is used ie. on Airflow to reflect tables during extraction, not when resource instance is created.

dlt/sources/sql_database/helpers.py
```diff
@@ -348,7 +350,20 @@ def table_rows(
 
     limit = None
     try:
-        limit = dlt.current.resource().limit
+        resource = dlt.current.resource()
```
rudolfix: see my `defer_table_reflect` comment in test. hints must be applied before
```py
yield dlt.mark.with_hints(
            [],
            dlt.mark.make_hints(
                **hints,
            ),
        )
```
which updates hints dynamically. pls find a nice code structure so we do not duplicate column merge

tests/load/sources/sql_database/test_sql_database_source.py
```diff
@@ -84,6 +87,79 @@ def convert_time_to_us(table):
     return new_table
 
 
+@pytest.mark.parametrize(
+    "columns_hint_type",
```
rudolfix: I'm not sure parametrizing by this makes sense: internally you'll get normalized resource.columns. if not lmk.

tests/load/sources/sql_database/test_sql_database_source.py
djudjuu: actually, now that we merge the hints also on deferred reflection, I actually do not want the test to fail, do you agree?

djudjuu:
ok, @rudolfix now I am actually confident I understood what happened and what is happenin now :see_no_evil: 

if defer:
- reflect from table + merge hints (unless they are of type callable)
if already reflected:
- try use hints from resource (unless they are callable or resource not there) (fallback: reflect from table)

expected (asserted) result: `apply_hints()` is always respected, schema in pipeline and parquet file are aligned

tests/load/sources/sql_database/test_sql_database_source.py
rudolfix: this is very good now

PR #3437, opened 2025-12-05
Fix: extensive .gitignore for dlt init

anuunchin:
This PR simply adjusts the .gitignore so it's more comprehensive. The base reference is https://github.com/github/gitignore/blob/main/Python.gitignore + dlt specific ignores.

### DIFF
--- a/dlt/_workspace/_templates/_single_file_templates/.gitignore
+++ b/dlt/_workspace/_templates/_single_file_templates/.gitignore
@@ -1,4 +1,5 @@
-# ignore secrets, virtual environments and typical python compilation artifacts
+# dlt-specific ignores
+# secrets and credentials
 secrets.toml
 *.secrets.toml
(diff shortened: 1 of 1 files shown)


dlt/_workspace/_templates/_single_file_templates/.gitignore
```diff
@@ -7,11 +8,199 @@ secrets.toml
...
+# pixi
+pixi.lock
+.pixi
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
+__pypackages__/
+
+# Celery stuff
+celerybeat-schedule
+celerybeat.pid
```
anuunchin, line 130: This one (for example) probably irrelevant, but should we keep these kinds of ignores as well? 👀  @sh-rp  imo yes

PR #3438, opened 2025-12-05
Skip doc examples requiring secrets on fork PRs

burnash:
When a PR comes from a fork, GitHub Actions doesn't expose repository secrets. This causes the docs examples tests to fail because many examples require external credentials (Postgres, BigQuery, Slack webhooks, etc.).

This PR adds an `EXAMPLES_REQUIRING_SECRETS` list to `prepare_examples_tests.py` that skips these examples when `IS_FORK=true`.

### DIFF
--- a/docs/docs_tools/examples/prepare_examples_tests.py
+++ b/docs/docs_tools/examples/prepare_examples_tests.py
@@ -13,8 +13,29 @@ EXAMPLES_DIR = "./examples"
 # settings
 SKIP_FOLDERS = ["archive", ".", "_", "local_cache"]
 # @pytest.mark.rfam
-SKIP_EXAMPLES: List[str] = ["backfill_in_chunks", "connector_x_arrow", "transformers"]
+SKIP_EXAMPLES: List[str] = [
+    "backfill_in_chunks",
+    "connector_x_arrow",
+    "transformers",
+]
+# Examples will be skipped from forked subprocesses
 SKIP_FORK_EXAMPLES: List[str] = ["custom_destination_lancedb"]
+
+# Examples that require external secrets (cloud credentials, API keys, etc.)
+# These will be skipped when running on fork PRs where secrets are not available
+EXAMPLES_REQUIRING_SECRETS: List[str] = [
+    "chess",
+    "chess_production",
+    "custom_destination_bigquery",
+    "custom_destination_lancedb",
+    "custom_naming",
+    "google_sheets",
+    "incremental_loading",
+    "nested_data",
+    "pdf_to_weaviate",
+    "postgres_to_postgres",
+    "qdrant_zendesk",
+]
 
 
 # the entry point for the script
@@ -40,13 +61,23 @@ def main() -> None:
     # get args
     args = parser.parse_args()
 
+    # Check if CI is running on a fork pull request
+    is_fork = os.environ.get("IS_FORK") == "true"
+
     count = 0
+    skipped_for_fork = 0
     for example in next(os.walk(EXAMPLES_DIR))[1]:
         # skip some
         if any(map(lambda skip: example.startswith(skip), SKIP_FOLDERS)):
             continue
 
         if example in SKIP_EXAMPLES:
+            continue
+
+        # Skip examples requiring secrets when running on fork PRs
+        if is_fork and example in EXAMPLES_REQUIRING_SECRETS:
+            skipped_for_fork += 1
+            fmt.note(f"Skipping {example} (requires secrets, running on fork PR)")
             continue
 
         count += 1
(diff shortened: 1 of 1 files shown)


docs/docs_tools/examples/prepare_examples_tests.py
```diff
@@ -40,7 +61,11 @@ def main() -> None:
     # get args
     args = parser.parse_args()
 
+    # Check if CI is running on a fork pull request
+    is_fork = os.environ.get("IS_FORK") == "true"
```
djudjuu, line 65: we have a helper:
https://github.com/dlt-hub/dlt/blob/d9e15a765dd22983698af7bba6124b712c7fefea/tests/utils.py#L551
```


def is_running_in_github_fork() -> bool:
    """Check if executed by GitHub Actions, in a repo fork."""
    is_github_actions = is_running_in_github_ci()
    is_fork = os.environ.get("IS_FORK") == "true"  # custom var set by us in the workflow's YAML
    return is_github_actions and is_fork
    ```

File: docs/docs_tools/examples/prepare_examples_tests.py [hunk 2]
```diff
@@ -13,9 +13,30 @@
...
 SKIP_FORK_EXAMPLES: List[str] = ["custom_destination_lancedb"]
 
+# Examples that require external secrets (cloud credentials, API keys, etc.)
+# These will be skipped when running on fork PRs where secrets are not available
+EXAMPLES_REQUIRING_SECRETS: List[str] = [
+    "chess",
+    "chess_production",
+    "custom_destination_bigquery",
+    "custom_destination_lancedb",
+    "custom_naming",
```
djudjuu, line 31: learning question: why this example?

docs/docs_tools/examples/prepare_examples_tests.py
```diff
@@ -13,9 +13,30 @@
...
+    "google_sheets",
+    "incremental_loading",
+    "nested_data",
```
djudjuu, line 34: question:partial_loading also uses aws creds so should be in there?

docs/docs_tools/examples/prepare_examples_tests.py
burnash: Good point! However, the `prepare_examples_tests.py` script is in `docs/docs_tools/examples/` and isn't a pytest test file, rather a standalone script that generates test files. Importing from `tests/utils` would create an awkward dependency from the docs tooling on the test utilities module. Additionally, the existing helper `is_running_in_github_fork()` does an extra check for `GITHUB_ACTIONS == "true"` via `is_running_in_github_ci`, but here we only need to check IS_FORK since this script only runs in CI anyway (and having IS_FORK=true already implies ci context)

If we want to keep things consistent, we could either:
1. Keep it as-is (simple inline check, no dependency)
2. Extract a shared utility for fork detection that both modules can use

I'd lean toward keeping it simple here, but happy to refactor if you prefer.

docs/docs_tools/examples/prepare_examples_tests.py
burnash: custom_naming example uses dlt.destinations.postgres() which requires Postgres credentials 

https://github.com/dlt-hub/dlt/blob/289e00dece9f396cf9fd5208f3f707fbbbb61375/docs/examples/custom_naming/custom_naming.py#L36

docs/docs_tools/examples/prepare_examples_tests.py
burnash: partial_loading example, it uses local filesystem storage, not aws/s3 creds:

https://github.com/dlt-hub/dlt/blob/289e00dece9f396cf9fd5208f3f707fbbbb61375/docs/examples/partial_loading/partial_loading.py#L143

burnash:
Thanks for the review @djudjuu, afaik the tests won't rerun on merge to devel.

docs/docs_tools/examples/prepare_examples_tests.py
djudjuu: ah right, and in filesystem the bucket_url (assigned above) is used to determine the kind of credential needed (and local filesystem doesnt require any so nothing ever gets resolved)

docs/docs_tools/examples/prepare_examples_tests.py
djudjuu: makes sense...so its just a bit inconsistent that this example doesnt include a `example.secrets.toml`

docs/docs_tools/examples/prepare_examples_tests.py
djudjuu: ah true... lets definitely not import that. option 2 is fine!

djudjuu:
> Thanks for the review @djudjuu, afaik the tests won't rerun on merge to devel.
ah true...i clicked on the action and its just the deploy job

PR #3439, opened 2025-12-05
Fix: Invisible sections are receiving border and background color in dashboard

anuunchin:
This PR fixes the below:
Screenshot 2025-12-05 at 16 17 43
So that it looks:
Screenshot 2025-12-05 at 16 18 16


rudolfix (approved):
LGTM!

PR #3442, opened 2025-12-05
Fix race condition in LimitItem 

burnash:
This PR fixes race condition in LimitItem.

Say it's parallelized  and `max_items=1` and:

1. Item A arrives:
- `self.count` becomes 1.
- `(self.count >= self.max_items)` is True.
- `self.exhausted` set to True.
- `self.gen.close()` is called.
- Returns Item A.
2. Item B arrives:
- The generator didn't stop fast enough, so `__call__` runs again.
- self.count becomes 2.
- (self.count >= self.max_items) is True.
- self.exhausted set to True.
- `return item` returns Item B.

Can be tested with:

```bash
for i in {1..100}; do pytest tests/extract/test_sources.py::test_limit_edge_cases[10] || break; done
```

### DIFF
--- a/dlt/extract/items_transform.py
+++ b/dlt/extract/items_transform.py
@@ -199,6 +199,10 @@ class LimitItem(ItemTransform[TDataItem, Dict[str, Any]]):
         if item is None:
             return None
 
+        # do not return any late arriving items
+        if self.exhausted:
+            return None
+
         if self.count_rows:
             self.count += count_rows_in_items(item)
         else:
@@ -219,9 +223,6 @@ class LimitItem(ItemTransform[TDataItem, Dict[str, Any]]):
             # otherwise never return anything
             if self.max_items != 0:
                 return item
-
-        # do not return any late arriving items
-        if self.exhausted:
             return None
 
         return item


djudjuu:
leaving it just because its so nice:

I'll explore the codebase to understand how `ItemTransform` classes relate to generators and thread pool processing. Let me trace through the extraction pipeline.

[3 tools called]

Let me explore the `Pipe` class and the concurrency module in more detail to understand the full picture:

[3 tools called]

Now let me look at how resources add the `LimitItem` transform and check the PR fix:

[5 tools called]

Now I have a complete picture! Let me explain the architecture and the race condition:

## The Big Picture: ItemTransform, Generators, and Thread Pools

### 1. **Architecture Overview**

```
┌──────────────────────────────────────────────────────────────────────────────┐
│                              DltResource                                      │
│                                                                               │
│   ┌────────────────────────────────────────────────────────────────────────┐ │
│   │                               Pipe                                      │ │
│   │                                                                         │ │
│   │  ┌─────────────┐    ┌─────────────┐    ┌─────────────┐    ┌──────────┐ │ │
│   │  │ Generator   │───▶│ MapItem     │───▶│ FilterItem  │───▶│ LimitItem│ │ │
│   │  │ (step 0)    │    │ (step 1)    │    │ (step 2)    │    │ (step 3) │ │ │
│   │  └─────────────┘    └─────────────┘    └─────────────┘    └──────────┘ │ │
│   │       ▲                                                                 │ │
│   │       │ gen                                                             │ │
│   └───────┼─────────────────────────────────────────────────────────────────┘ │
└───────────┼───────────────────────────────────────────────────────────────────┘
            │
            ▼
┌──────────────────────────────────────────────────────────────────────────────┐
│                           PipeIterator                                        │
│                                                                               │
│   Iterates through items, applies steps sequentially.                         │
│   ┌─────────────────────────────────────────────────────────────────────────┐ │
│   │                        FuturesPool                                       │ │
│   │   - ThreadPoolExecutor for callables                                     │ │
│   │   - AsyncIO event loop for awaitables                                    │ │
│   │   - max_parallel_items: 20 (default)                                     │ │
│   │   - workers: 5 (default)                                                 │ │
│   └─────────────────────────────────────────────────────────────────────────┘ │
└──────────────────────────────────────────────────────────────────────────────┘
```

### 2. **The Pipe Structure**

A `Pipe` is a chain of steps stored in `_steps`:

```36:36:dlt/extract/pipe.py
# steps in a pipe
self._steps: List[TPipeStep] = []
```

- **Step 0** is typically the **generator** (data source)
- **Steps 1+** are **transforms** (`ItemTransform` subclasses: `MapItem`, `FilterItem`, `LimitItem`, etc.)

When a pipe is prepared for iteration, `evaluate_gen()` binds each `ItemTransform` to the pipe:

```301:305:dlt/extract/pipe.py
# evaluate transforms
for step_no, step in enumerate(self._steps):
    if isinstance(step, ItemTransform):
        self._steps[step_no] = step.bind(self)
```

### 3. **How Items Flow Through the Pipeline**

The `PipeIterator.__next__()` method orchestrates the flow:

```220:250:dlt/extract/pipe_iterator.py
# advance to next step
step = pipe_item.pipe[pipe_item.step + 1]
try:
    next_meta = pipe_item.meta
    next_item = step(item, meta=pipe_item.meta)  # <-- CALLS ItemTransform.__call__()
    ...

# create next pipe item if a value was returned. A None means that item was consumed/filtered out
if next_item is not None:
    pipe_item = ResolvablePipeItem(
        next_item, pipe_item.step + 1, pipe_item.pipe, next_meta
    )
else:
    pipe_item = None  # Item filtered out
```

### 4. **Parallelization & Thread Pool**

When you call `resource.parallelize()`, the generator is wrapped with `wrap_parallel_iterator()`:

```238:270:dlt/extract/utils.py
def wrap_parallel_iterator(f: TAnyFunOrGenerator) -> TAnyFunOrGenerator:
    """Wraps a generator for parallel extraction"""

    def _gen_wrapper(*args: Any, **kwargs: Any) -> Iterator[TDataItems]:
        ...
        def _parallel_gen() -> TDataItems:
            # This callable will be submitted to the thread pool
            nonlocal busy
            nonlocal exhausted
            try:
                return next(gen)  # Calls next() on the underlying generator
            except StopIteration:
                exhausted = True
                return None
            finally:
                busy = False

        while not exhausted:
            ...
            yield _parallel_gen  # <-- Yields a CALLABLE, not data!
```

**Key insight**: Instead of yielding data items, `wrap_parallel_iterator` yields **callables** (`_parallel_gen`).

In `PipeIterator.__next__()`, when a callable is encountered:

```199:205:dlt/extract/pipe_iterator.py
with pipe_context(pipe_item.pipe):
    if isinstance(item, Awaitable) or callable(item):
        # Callables are submitted to the pool to be executed in the background
        self._futures_pool.submit(pipe_item)
        pipe_item = None
        # Future will be resolved later, move on to the next item
        continue
```

The callable is **submitted to the thread pool** where it executes in parallel:

```117:120:dlt/extract/concurrency.py
elif callable(item):
    # pass pipe context to thread pool, happens automatically for coroutines
    ctx = contextvars.copy_context()
    future = self._ensure_thread_pool().submit(ctx.run, item)
```

### 5. **The Race Condition in `LimitItem`**

Here's the problematic flow with `max_items=1` and parallelization:

```
Main Thread                         Thread Pool Worker(s)
─────────────────────────────────────────────────────────────────────
1. Generator yields callable A
2. Submit A to thread pool
3. Generator yields callable B      
4. Submit B to thread pool
   (max_parallel_items allows this)
                                    5. Worker executes A, returns item
                                    6. Worker executes B, returns item
7. Resolve future A → item A
8. Apply LimitItem to item A:
   - count = 1
   - exhausted = True
   - gen.close()
   - return item A ✓
9. Resolve future B → item B
10. Apply LimitItem to item B:
    - count = 2
    - (exhausted is True, but...)
    - Check: if self.exhausted: return None
    
    BUT WAIT! The old code had the check AFTER returning:
```

Looking at the current `LimitItem.__call__()`:

```197:227:dlt/extract/items_transform.py
def __call__(self, item: TDataItems, meta: Any = None) -> Optional[TDataItems]:
    # do not count None
    if item is None:
        return None

    if self.count_rows:
        self.count += count_rows_in_items(item)
    else:
        self.count += 1  # <-- INCREMENT HAPPENS FIRST

    # detect when the limit is reached
    if (
        (self.count >= self.max_items and self.max_items >= 0)
        ...
    ):
        self.exhausted = True
        if inspect.isgenerator(self.gen):
            self.gen.close()
        if self.max_items != 0:
            return item  # <-- RETURNS ITEM BEFORE CHECKING exhausted AGAIN

    # do not return any late arriving items
    if self.exhausted:  # <-- THIS CHECK COMES TOO LATE!
        return None

    return item
```

**The bug**: When item B arrives:
1. It increments `count` to 2
2. The condition `(self.count >= self.max_items)` is `True`
3. `self.exhausted` is set to `True` (already was)
4. **It returns item B** because `max_items != 0`

The check `if self.exhausted: return None` at line 224-225 **only runs if the limit condition wasn't met**. But by that point, the item has already been returned!

### 6. **The Fix (PR #3442)**

The fix should check `self.exhausted` **at the very beginning** before processing:

```python
def __call__(self, item: TDataItems, meta: Any = None) -> Optional[TDataItems]:
    # EARLY EXIT: don't process any late-arriving items
    if self.exhausted:
        return None
        
    if item is None:
        return None
    # ... rest of the logic
```

This ensures that once `exhausted` is set by the first item reaching the limit, all subsequent items are immediately filtered out, regardless of the order in which futures complete.

### Summary Diagram

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                    Extraction with Parallelization                           │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                              │
│  Generator ─┬─▶ callable() ─┬─▶ [Thread Pool] ─┬─▶ resolved item            │
│             │               │                   │                            │
│             ├─▶ callable() ─┼─▶ [Thread Pool] ─┼─▶ resolved item            │
│             │               │                   │                            │
│             └─▶ callable() ─┴─▶ [Thread Pool] ─┴─▶ resolved item            │
│                                                         │                    │
│                                                         ▼                    │
│                                                 ┌──────────────┐             │
│                                                 │  LimitItem   │             │
│                                                 │  (SHARED!)   │◀── Race!    │
│                                                 └──────────────┘             │
│                                                         │                    │
│                                                         ▼                    │
│                                                   Final Items                │
└─────────────────────────────────────────────────────────────────────────────┘
```

The `LimitItem` instance is **shared** across all items flowing through the pipe. When futures resolve non-deterministically (due to threading), multiple items can slip through before `exhausted` is properly checked.

rudolfix:
Anton's fix is correct. What is important (not sure the diagram shows it) is that all the counting and filtering happens in the main thread. only `gen` element may be parallelized to a pool but it is collected into main thread before any next step (like limit item) is applied.

the root problem in case of async generators is that we do `self.gen.close()` which closes a wrapper generator over async function... and this indeed may allow additional elements to be spit out from it. my intuition was that bug is somewhere there (why only async generators have problems? that does not happen in regular parallel generators). I hope this fix is sufficient :)

PR #3445, opened 2025-12-08
docs: add runtime docs to CLI reference

ivasio:
- generating the reference docs for the entire hub extra, including the runtime
- added the check to the CI

note, following commands were listed as available, but were not actually rendered by the generator. nor are they now, but now they are also [not listed](https://github.com/dlt-hub/dlt/pull/3445/files#diff-24c28d4566e8ee8d0f53350a663e3d4dfa099e728c176d7d688a0df2ea633b6aL34) in the full list of available commands: `transformation,source,destination,dbt,dataset,cache`

### DIFF
--- a/docs/uv.lock
+++ b/docs/uv.lock
@@ -834,11 +834,7 @@ wheels = [
 
 [[package]]
 name = "dlt"
-<<<<<<< HEAD
 version = "1.20.0"
-=======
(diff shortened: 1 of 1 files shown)


rudolfix (dismissed):
LGTM! I disabled the automated docs check in lint due to #3454

PR #3448, opened 2025-12-08
Fix state file being ignored when pipline_name includes FILENAME_SEPA…

### LINKED ISSUE #3427: State restoration fails silently when pipeline name contains double underscores (`__`)
### dlt version

1.19.1

### Describe the problem

When a pipeline name contains double underscores (`__`), state files are written successfully but cannot be restored on subsequent runs. This causes incremental loading to silently fail, re-processing all data from the beginning.

This issue only manifests when state sync from destination is required (e.g., running on job runners like Databricks, Airflow, or CI systems that start with a clean filesystem). When running locally with a persistent pipeline working directory, the local state file is used and everything works fine - which made this bug quite hard to find.

## Root Cause

The filesystem destination uses `__` as `FILENAME_SEPARATOR` when naming state files:

https://github.com/dlt-hub/dlt/blob/10cd9089d9da9c449a4195370d1547d9dfd2801e/dlt/destinations/impl/filesystem/filesystem.py#L87
https://github.com/dlt-hub/dlt/blob/10cd9089d9da9c449a4195370d1547d9dfd2801e/dlt/destinations/impl/filesystem/filesystem.py#L769-L774

When parsing state files, dlt splits by `__` and expects exactly 3 parts:

https://github.com/dlt-hub/dlt/blob/10cd9089d9da9c449a4195370d1547d9dfd2801e/dlt/destinations/impl/filesystem/filesystem.py#L727-L741

(notice this bit:

```python
        if len(fileparts) != 3:
            continue  # <-- Silently skips files that don't have exactly 3 parts
```

If the pipeline name itself contains `__`, the split produces more than 3 parts, and the state file is silently skipped.

### Example

With pipeline name `my_pipeline__resource`:
- State file written: `my_pipeline__resource__1234567890.123__abc123hash.jsonl`
- When split by `__`: `["my_pipeline", "resource", "1234567890.123", "abc123hash"]` → 4 parts
- Expected: 3 parts → file is skipped
- Result: "The state was not found in the destination" even though it exists

## Impact

- Incremental loading breaks silently - users may not notice until they see duplicate data or performance issues
- The warning message "state was not found" is misleading since the state file does exist
- This is especially problematic when pipeline names are constructed dynamically (e.g., `f"{source}_{table}"` where components may contain underscores that get normalized to `__`)
- Debugging is difficult because the issue doesn't reproduce locally

## Environment

- dlt version: 1.19.1 (likely affects all versions using filesystem destination)
- Destination: filesystem
- State sync: enabled (restore_from_destination=True)
- Runtime: Databricks job runner (clean filesystem between runs)

### Expected behavior

We need to fail the run. And add validation in `dlt.pipeline()` to reject pipeline names containing `__`. 

### Steps to reproduce

This bash script:

```bash
# Clean up from previous runs
rm -rf /tmp/test_dlt_destination
rm -rf ~/.dlt/pipelines/my_pipeline__with__underscores

# Create the Python reproduction script
cat > /tmp/test_dlt_double_underscore.py << 'EOF'
# /// script
# requires-python = ">=3.10"
# dependencies = [
#     "dlt[filesystem]==1.19.1",
# ]
# ///
import os

# Enable dlt logging before importing dlt
os.environ["RUNTIME__LOG_LEVEL"] = "INFO"

import dlt

# Pipeline name with double underscores (common when normalizing dots or other characters)
pipeline = dlt.pipeline(
    pipeline_name="my_pipeline__with__underscores",
    destination=filesystem(bucket_url="/tmp/test_dlt_destination"),
    dataset_name="test_data"
)

@dlt.resource
def my_data():
    yield {"id": 1, "value": "test"}

# Run the pipeline
pipeline.run(my_data())
print("Pipeline run completed")
EOF

echo "=== First run (writes state) ==="
uv run /tmp/test_dlt_double_underscore.py 2>&1 | grep -i "state" || true

echo ""
echo "=== Checking state file exists ==="
find /tmp/test_dlt_destination -name "*.jsonl" -path "*_dlt_pipeline_state*" | head -5

echo ""
echo "=== Simulating clean job runner (deleting local pipeline state) ==="
rm -rf ~/.dlt/pipelines/my_pipeline__with__underscores

echo ""
echo "=== Second run (should restore state but fails) ==="
uv run /tmp/test_dlt_double_underscore.py 2>&1 | grep -i "state"

echo ""
echo "=== Notice: 'state was not found' even though the state file exists! ==="
```

This gives:

```
=== First run (writes state) ===
2025-12-03 15:53:02,844|[INFO]|43342|8393318720|dlt|pipeline.py|_restore_state_from_destination:1606|The state was not found in the destination filesystem (dlt.destinations.filesystem):test_data

=== Checking state file exists ===
/tmp/test_dlt_destination/test_data/_dlt_pipeline_state/my_pipeline__with__underscores__1764773497.998475__bb32ee1f4580d6fbb5aa3f8696c18dfd99e04f12f1507eb21b483bee81661dea.jsonl
/tmp/test_dlt_destination/test_data/_dlt_pipeline_state/my_pipeline__with__underscores__1764773520.456136__bb32ee1f4580d6fbb5aa3f8696c18dfd99e04f12f1507eb21b483bee81661dea.jsonl

=== Simulating clean job runner (deleting local pipeline state) ===

=== Second run (should restore state but fails) ===
2025-12-03 15:53:03,270|[INFO]|43348|8393318720|dlt|pipeline.py|_restore_state_from_destination:1606|The state was not found in the destination filesystem (dlt.destinations.filesystem):test_data
2025-12-03 15:53:03,314|[INFO]|43348|8393318720|dlt|worker.py|_get_items_normalizer:132|A file format for table _dlt_pipeline_state was specified to preferred in the resource so jsonl format being used.
2025-12-03 15:53:03,315|[INFO]|43348|8393318720|dlt|validate.py|validate_and_update_schema:26|Updating schema for table _dlt_pipeline_state with 1 deltas
2025-12-03 15:53:03,315|[INFO]|43348|8393318720|dlt|normalize.py|clean_x_normalizer:174|Table _dlt_pipeline_state has seen data for the first time with load id 1764773583.275845
2025-12-03 15:53:03,324|[INFO]|43348|8393318720|dlt|load.py|submit_job:172|Will load file 1764773583.275845/new_jobs/_dlt_pipeline_state.fdfa61a7f3.0.jsonl.gz with table name _dlt_pipeline_state
2025-12-03 15:53:03,324|[INFO]|43348|8393318720|dlt|load.py|complete_jobs:471|Job for _dlt_pipeline_state.fdfa61a7f3.jsonl.gz completed in load 1764773583.275845

=== Notice: 'state was not found' even though the state file exists! ===
```

### Operating system

Linux, macOS

### Runtime environment

Other

### Python version

3.12

### dlt data source

doesn't matter.

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

Travior:
This PR fixes a bug where an existing `_dlt_pipeline_state` tables would be ignored if the `pipeline_name` contained the `FILE_SEPARATOR` (`__`).

Additionally a test for this is added

Resolves #3427

### DIFF
--- a/dlt/destinations/impl/filesystem/filesystem.py
+++ b/dlt/destinations/impl/filesystem/filesystem.py
@@ -784,7 +784,7 @@ class FilesystemClient(
             filename = os.path.splitext(os.path.basename(filepath))[0]
             fileparts = filename.rsplit(
                 FILENAME_SEPARATOR, maxsplit=2
-            )  # pipeline_name, load_id, version_hash
+            )  # name, load_id, version_hash
             if len(fileparts) != 3:
                 continue
             # Filters only if pipeline_name provided
--- a/tests/load/filesystem/test_filesystem_client.py
+++ b/tests/load/filesystem/test_filesystem_client.py
@@ -423,17 +423,26 @@ def test_get_storage_version_invalid(invalid_version_info: Union[str, Dict[str, 
             client.get_storage_versions()
 
 
-def test_list_dlt_table_files_with_separator_in_pipeline_name() -> None:
+@pytest.mark.parametrize("pipeline_name",
+    [
+        "my__pipeline",
+        "my_pipeline__long_name__with__underscores",
+        "__",
+        "____",
+        "__long_name__with__underscores__",
+    ],
+)
+def test_list_dlt_table_files_with_separator_in_pipeline_name(pipeline_name: str) -> None:
     filesystem_ = filesystem("random_location")
     client = _client_factory(filesystem_)
     client.initialize_storage()
 
     state_table_dir = client.get_table_dir(client.schema.state_table_name)
     client.fs_client.mkdirs(state_table_dir)
-    test_file = client.pathlib.join(state_table_dir, "my__pipeline__load123__hash123.jsonl")
+    test_file = client.pathlib.join(state_table_dir, f"{pipeline_name}__load123__hash123.jsonl")
     client.fs_client.touch(test_file)
 
     results = list(client._list_dlt_table_files(client.schema.state_table_name))
     assert len(results) == 1
     assert results[0][0] == test_file
-    assert results[0][1] == ["my__pipeline", "load123", "hash123"]
+    assert results[0][1] == [pipeline_name, "load123", "hash123"]


dlt/destinations/impl/filesystem/filesystem.py
```diff
@@ -734,11 +734,14 @@ def _list_dlt_table_files(
         for filepath in all_files:
             filename = os.path.splitext(os.path.basename(filepath))[0]
             fileparts = filename.split(FILENAME_SEPARATOR)
```
rudolfix: you could use `rsplit` with max splits = 2 so the first part will always be pipeline name then you do not need to change any code below

Travior: Makes a lot of sense, force pushing so that I leave the blame intact

tests/load/filesystem/test_filesystem_client.py
```diff
@@ -421,3 +421,19 @@ def test_get_storage_version_invalid(invalid_version_info: Union[str, Dict[str,
     else:
         with pytest.raises(UnsupportedStorageVersionException):
             client.get_storage_versions()
+
+
+def test_list_dlt_table_files_with_separator_in_pipeline_name() -> None:
```
burnash: Very good, let's cover more test cases with `pytest.mark.parametrize`, see the example test above: e.g. single `__`, multiple `__`, consider other edge cases.

burnash (changes requested):
@Travior the fix is correct as @rudolfix said above. Let's make this fix even better. The test covers the basic case, but could you add parametrization to cover edge cases? Think about variations in how __ can appear in the pipeline name. See the test above for the syntax.

PR #3460, opened 2025-12-10
add support for gcp adc with iceberg

benjibuiltit:
### Description
Add support for GCP ADC with Iceberg table format

### Additional Context
Pyiceberg gcs settings being set in this PR are documented here:
https://py.iceberg.apache.org/configuration/#google-cloud-storage

### DIFF
--- a/dlt/common/configuration/specs/gcp_credentials.py
+++ b/dlt/common/configuration/specs/gcp_credentials.py
@@ -24,28 +24,28 @@ from dlt.common.configuration.specs.base_configuration import (
 from dlt.common.utils import is_interactive
 
 
-def _get_pyiceberg_fileio_config_from_adc(
-    credentials: Any, project_id: Optional[str]
-) -> Dict[str, Any]:
-    """Get pyiceberg FileIO config from Google Application Default Credentials.
+# GCS scope required for OAuth2 token requests
+GCS_SCOPE = "https://www.googleapis.com/auth/devstorage.read_write"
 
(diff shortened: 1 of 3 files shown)


dlt/common/configuration/specs/gcp_credentials.py
```diff
@@ -24,6 +24,35 @@
...
+) -> Dict[str, Any]:
+    """Get pyiceberg FileIO config from Google Application Default Credentials.
+
+    This refreshes the ADC credentials to get a valid OAuth2 token and returns
+    the configuration dict that pyiceberg expects for GCS access.
+    """
+    import google.auth
+    from google.auth.transport.requests import Request
+
+    native_credentials, project = google.auth.default()
```
rudolfix: I do not think you need to do it here: when you are calling this function you will get `credentials` and `project_id` that are already taken from default credentials.

so this function will just set fileio from any credentials + project_id passed in the arguments.

btw. now I see that we could implement `to_pyiceberg_fileio_config` on `GcpServiceAccountCredentialsWithoutDefaults`:

```
def to_pyiceberg_fileio_config(self) -> Dict[str, Any]:
        return _get_pyiceberg_fileio_config_from_adc(self.to_native_credentials(), self.project_id)
```
and that actually could be tested

rudolfix (changes requested):
@benjibuiltit thanks for this contribution. code looks good - I think it could be still simplified a bit. Default credentials are hard to test... are you running this code somewhere in production?

I also left a suggestion to apply the same trick to service credentials. this is actually testable and here we could help to set it up

PR #3463, opened 2025-12-10
3353 normalize start method spawn seems to ignore environment variables

djudjuu:
- **test for bad behavior**
- **fix**
- **another test for the new mechanism**


tests/common/runtime/test_config_section_context_restore.py
```diff
@@ -0,0 +1,143 @@
+"""Test that ConfigSectionContext is properly restored in spawned worker processes."""
```
djudjuu: do we want this test? I wanted to have something that is closer to the actual change, whereas the other test is more of a regression test that fixes the exact issue

tests/load/pipeline/test_pipelines.py
```diff
@@ -568,6 +568,60 @@ def test_pipeline_data_writer_compression(
...
 
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(
+        default_sql_configs=True,
+        all_buckets_filesystem_configs=True,
+    ),
+    ids=lambda x: x.name,
+)
+def test_normalize_compression_with_spawn_workers(
```
rudolfix: cool test! but running for all destinations is an overkill. pls. move it to `tests/pipeline`, it does need load step.

tests/load/pipeline/test_pipelines.py
```diff
@@ -568,6 +568,60 @@ def test_pipeline_data_writer_compression(
...
+        table_format=destination_config.table_format,
+        loader_file_format=destination_config.file_format,
+    )
+
+    # Normalize with multiple workers and spawn method
+    p.normalize(workers=workers)
+
+    # Check that normalized files are not compressed
+    load_storage = p._get_load_storage()
+    normalized_packages = load_storage.list_normalized_packages()
```
rudolfix: this method is already on the pipeline

tests/common/runtime/test_config_section_context_restore.py
```diff
@@ -0,0 +1,143 @@
...
+from dlt.common.configuration.container import Container
+from dlt.common.configuration.specs import PluggableRunContext, ConfigSectionContext
+from dlt.common.configuration.specs.base_configuration import BaseConfiguration, configspec
+from dlt.common.configuration.resolve import resolve_configuration
+from dlt.common.runners.configuration import PoolRunnerConfiguration
+from dlt.common.runners.pool_runner import create_pool
+
+
+@configspec
+class SectionedTestConfig(BaseConfiguration):
```
rudolfix: makes sense to have those tests. pls. move them to `test_runners.py`. (tbh. worker init tests were missing altogether)

tests/common/runtime/test_config_section_context_restore.py
```diff
@@ -0,0 +1,143 @@
...
+    Returns:
+        Tuple of (resolved_value, sections_from_context)
+    """
+    section_ctx = Container()[ConfigSectionContext]
+    config = resolve_configuration(SectionedTestConfig())
+
+    return config.test_value, section_ctx.sections
+
+
+def test_config_section_context_restored_in_spawn_worker() -> None:
```
rudolfix: actually sections should be visible both in case of spawn and fork. so test both (parametrize)

tests/common/runtime/test_config_section_context_restore.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+
+    # Verify that config resolution used the correct sections
+    assert result_value == "sectioned_value", (
+        f"Expected 'sectioned_value' but got '{result_value}'. "
+        "Config resolution did not use the restored ConfigSectionContext sections."
+    )
+
+
+def test_config_section_context_with_pipeline_name() -> None:
```
rudolfix: does not make sense to do those tests. we are testing if context was passed. the test above does that. here you only test config resolver behavior IMO. this is done elsewhere

tests/common/runtime/test_config_section_context_restore.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Set up ConfigSectionContext in main process
+    section_context = ConfigSectionContext(
+        pipeline_name=None,
+        sections=("my_section",),
+    )
+
+    # Store it in container
+    container = Container()
+    container[ConfigSectionContext] = section_context
```
rudolfix: we are also missing the case when there's not section context in container (do `del` on container[ConfigSectionContext]`)
you can parametrize this test to chekc that

djudjuu:
all failing tests are :
image
but they work locally. is github ci somehow different?

djudjuu:
> In GitHub CI, the Python process typically has:

- more active threads (pytest, coverage, logging, executor internals),
- different import timing,
- and lazier worker creation by ProcessPoolExecutor.

> Forking in the presence of threads and lazy initialization is explicitly unsafe in Python. As a result, workers may be forked before the section context is set, or the container may be re-initialized in the worker process, causing the inherited ConfigSectionContext to be missing. This leads workers to fall back to the default (empty) sections, which is exactly what the failing test shows.

chatgpt-5

so maybe we set the pytest expectation for fork-method to fail and do  a follow up issue?

are people running dlt-pipelines in github-actions? those would be affected...

rudolfix (commented):
looking at failed tests:
- `fork` is not supported on windows. what's weird that creation of process pool is not failing. instead something else happens that skips passing the section context to your test. maybe somehow `spawn` is used? then obviously no context is passed (you think we do fork but system does spawn)
```
if start_method != "fork":
            ctx = Container()[PluggableRunContext]
```
you can see what happens if `multiprocessing.get_context(method=start_method),` force=True is passed

2. another thing for fork:
if you start the pool BEFORE adding section to container then obviously you get empty container in the fork - because the process forks when the pool starts

3, on mac the default method is spawn. so I think problem is same as windows. for some reason the start type is not really set

you need to debug carefully how pool is created

rudolfix:
@djudjuu this looks good now. you can remove explicit start method from failing tests.

PR #3470, opened 2025-12-10
Add GitHub Actions workflow for dependency submission of uv.lock files

### LINKED ISSUE #3461: Set up GitHub-based security alerts for dependencies
- Ensure GitHub dependency graph + Dependabot alerts are enabled for the repo.
- Add a GitHub Action that submits deps from pyproject.toml / uv.lock (rmuir/uv-dependency-submission?)

Related: https://github.com/dlt-hub/dlt/issues/957

burnash:
Adds a workflow which submits uv.lock dependencies to GitHub's dependency graph for security scanning.
See: https://github.com/rmuir/uv-dependency-submission

Closes #3461

### DIFF
--- a/.github/workflows/dependency-submission.yml
+++ b/.github/workflows/dependency-submission.yml
@@ -1,0 +1,33 @@
+# Submits uv.lock dependencies to GitHub's dependency graph for security scanning.
+# Requires: Settings → Security → Dependency graph enabled.
+# See: https://github.com/rmuir/uv-dependency-submission
+name: Dependency Submission
+
+on:
+  workflow_dispatch:
+  push:
+    branches: ['master', 'devel']
+    paths:
+      - '**/uv.lock'
+
+permissions: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  dependency-submission:
+    name: Submit uv dependencies
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: write   # required to submit dependency graph data
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+        with:
+          persist-credentials: false
+
+      - name: Submit dependency snapshot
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-submission.yml
```diff
@@ -0,0 +1,33 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  dependency-submission:
+    name: Submit uv dependencies
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: write   # required to submit dependency graph data
```
rudolfix, line 25: this allows this workflow to modify content of our repository and gives write access to tons of other api. ie. pushing commits to branches becomes possible.

we do not have any workflows that require write access and it is better if that stays this way. the alternative is to export lockfiles ourselves which can be fully automated (but may generate one more conflict when merging).

rudolfix (changes requested):
tbh. I'm against having a workflow with write access to the repo

.github/workflows/dependency-submission.yml
rmuir: Hey, i saw this PR, I think you might have some confusion about how actions permissions work. All your other workflows are running with blanket privilege. You can see this at the beginning of your logs.

Screen_Shot_2025-12-11_at_07 07 12

I recommend use of the [zizmor](https://github.com/zizmorcore/zizmor) tool to enforce that workflows run with least-privilege and only use the permissions they need.

rudolfix: @rmuir thanks for pointing that out! it is actually default org level setting to give full write access to github tokens... pretty unexpected tbh.!

rmuir: github actions is trappy in many ways. zizmor and actionlint are tools I use to detect and prevent the traps, they are really good. I use them as a ci "self-check" in all repos i deal with as a best-practice. Example workflow (assumes you have advanced security/code-scanning enabled: https://github.com/apache/lucene/blob/main/.github/workflows/actions.yml)

`uv-dependency-submission` action declares only one permission because that is all it needs to upload the dependency graph data. 

of course it will still work fine if you remove all the `permission` clauses from the workflow, to look like other workflows that don't define permissions, but then it will run with more permissions than are needed.

Since the action is important for supply-chain security, I wanted to have good defaults.

PR #3479, opened 2025-12-15
Feat/faster testing

tetelio:
# PR Summary

This PR improves CI and local test execution time by parallelizing test execution. This is achieved thanks to pytest xdist and fully isolating test workers at the filesystem and service level to avoid collision between workers. 

## Changes

- Introduced worker-aware helpers in `tests.utils` to provide a worker-specific test storage root and stable worker ID.  
- Updated tests to consistently use worker-specific storage roots so parallel workers never write to the same filesystem paths.  
- Used the worker ID to isolate external or service-backed resources when filesystem isolation alone is insufficient.  
- Updated load test destinations that rely on global catalogs (e.g. MotherDuck or ducklake) to use worker-specific catalogs to prevent cross-worker conflicts.  
- Increased the number of parallel workers where safe, particularly for remote destinations that scale elastically (e.g. Athena), which were previously the slowest tests.  
- Kept parallelism conservative for services backed by shared or limited execution engines (e.g. Synapse, Databricks), where increasing workers can overwhelm the backend rather than improve performance.  
- Common tests are CPU bound and GitHub-hosted runner constraints have 4 cores fully used. Further improvements  would require increasing the number of CI machines generating more OS/Python/test-suite combinations, reducing testing-matrix readability. Local development machines typically have more than 4 cores (e.g. 12-core setups), allowing the common tests to run in ~5 minutes using the provided fast local test Make targets (`make test-common-p`). Tests can be executed serially with `make test-common`
- Preserved explicit separation between parallel-safe tests and serial-only tests, ensuring that tests with known concurrency constraints are always executed sequentially.


tests/common/schema/test_import_normalizers.py
```diff
@@ -124,7 +124,6 @@ def test_import_normalizers() -> None:
     config, naming, json_normalizer = import_normalizers(configured_normalizers())
     assert isinstance(naming, snake_case.NamingConvention)
     # no maximum length: we do not know the destination capabilities
-    assert naming.max_length is None
```
tetelio, line 127: This depends on test execution order and fails sometimes

tests/common/runtime/test_telemetry.py
```diff
@@ -207,6 +210,11 @@ def test_execution_context_with_plugin() -> None:
         assert context["dlthub"] == {"name": "dlthub", "version": "1.7.1"}
     finally:
         sys.path.remove(plus_path)
+        sys.modules.pop("dlthub", None)
+        sys.modules.pop("dlthub.version", None)
+
+    after = get_execution_context()
```
tetelio, line 216: Avoid test leakage

tests/common/configuration/test_credentials.py
```diff
@@ -351,13 +372,30 @@ def test_needs_scopes_for_refresh_token() -> None:
...
+    # This keeps the test deterministic and consistent across pytest and xdist.
+
     c = GcpOAuthCredentialsWithoutDefaults()
     # without refresh token
     c.parse_native_representation(OAUTH_USER_INFO % "")
     assert c.refresh_token is None
     assert c.token is None
-    with pytest.raises(AssertionError):
-        c.auth()
+    with patch.object(
```
tetelio, line 347: Oauth browser window opens with xdist

tests/common/storages/test_local_filesystem.py
```diff
@@ -93,11 +93,15 @@ def test_file_win_configuration() -> None:
 @pytest.mark.parametrize(
     "bucket_url,file_url",
     (
-        (r"/src/local/app", "file:///src/local/app"),
-        (r"_storage", "file://" + pathlib.Path("_storage").resolve().as_posix()),
+        ("/src/local/app", "file:///src/local/app"),
```
tetelio, line 96: This uses "TEST_STORAGE" as a sentinel because @pytest.mark.parametrize values are evaluated at collection time and can’t safely call functions. We replace it at test runtime with get_test_storage_root() to get a per-test, xdist-safe path and avoid shared filesystem state.

tests/common/storages/test_local_filesystem.py
```diff
@@ -130,12 +134,19 @@ def test_local_user_posix_path_configuration(bucket_url: str) -> None:
 @pytest.mark.parametrize(
     "bucket_url,local_path",
     (
-        ("file://_storage", "_storage"),
-        ("file://_storage/a/b/c.txt", "_storage/a/b/c.txt"),
+        ("FILE_TEST_STORAGE", None),
```
tetelio, line 137: This uses "TEST_STORAGE" as a sentinel because @pytest.mark.parametrize values are evaluated at collection time and can’t safely call functions. We replace it at test runtime with get_test_storage_root() to get a per-test, xdist-safe path and avoid shared filesystem state.

tests/load/lancedb/fixtures.py
```diff
@@ -0,0 +1,10 @@
+import pytest
+import dlt
+
+from tests.utils import get_test_worker_id
+
+
+@pytest.fixture
```
tetelio: The issue is concurrency: LanceDB uses shared on-disk state per directory, so dataset/collection names do not provide isolation. Under pytest-xdist, each worker must use a separate directory to avoid file-level collisions.

tests/load/utils.py
```diff
@@ -382,9 +384,15 @@ def destinations_configs(
         ]
         destination_configs += [
             DestinationTestConfiguration(destination_type="duckdb", file_format="parquet"),
-            DestinationTestConfiguration(destination_type="ducklake", supports_dbt=False),
             DestinationTestConfiguration(
```
tetelio, line 456: DuckLake uses shared on-disk metadata and multi-file state; dataset names don’t provide isolation. Under xdist, each worker must use a separate database/namespace to avoid metadata collisions.

tests/load/utils.py
```diff
@@ -382,9 +384,15 @@ def destinations_configs(
...
-                destination_type="motherduck", file_format="insert_values"
+                destination_type="ducklake",
+                supports_dbt=False,
+                credentials={"ducklake_name": f"ducklake_{worker}"},
+            ),
+            DestinationTestConfiguration(
```
tetelio, line 461: MotherDuck databases are global per account and persist across runs. Parallel workers must use worker-specific database names to avoid schema and table collisions.

tests/load/utils.py
```diff
@@ -458,10 +466,13 @@ def destinations_configs(
     if default_vector_configs:
         destination_configs += [
             DestinationTestConfiguration(destination_type="weaviate"),
-            DestinationTestConfiguration(destination_type="lancedb"),
+            DestinationTestConfiguration(
```
tetelio, line 539: LanceDB and local Qdrant both persist shared on-disk state per directory and assume a single writer. Under pytest-xdist, each worker must use a worker-specific path/URI to avoid file-level collisions and corrupted state.

tests/pipeline/test_dlt_versions.py
```diff
@@ -33,7 +34,30 @@
 from dlt.destinations.impl.filesystem.sql_client import FilesystemSqlClient
 
 from tests.pipeline.utils import airtable_emojis, load_table_counts, assert_load_info
-from tests.utils import TEST_STORAGE_ROOT
```
tetelio, line 36: This refactor runs the version-migration tests inside a real workspace context instead of relying on a shared cwd and TEST_STORAGE_ROOT. It switches to per-test storage via get_test_storage_root() and a small test_workspace helper to mirror actual CLI execution. Test behavior and assertions are unchanged; this only improves isolation, determinism, and xdist safety.

tests/workspace/cli/conftest.py
```diff
@@ -1,3 +1,23 @@
+# MUST be before importing dlt
+import os
```
tetelio: Question: is it ok changing the home dir during test execution? I havent seen more side effects but i dont know if its used for inspecting results of any type later

tetelio: A per-worker fake home and XDG data directory is created and exported before importing dlt. This ensures that CLI and workspace code never touches the real user home directory and that each pytest worker operates in a fully isolated environment. The change guarantees deterministic behavior across local runs, CI, and pytest-xdist.

tests/workspace/cli/test_init_command.py
```diff
@@ -297,14 +302,26 @@ def test_init_all_sources_together(repo_dir: str, workspace_files: FileStorage)
...
+        set(get_source_candidates(initial_repo_dir)).union(set(CORE_SOURCES)).union(set(TEMPLATES))
     )
+
+    # Clean up the initial repo dir as we don't need it anymore
+    shutil.rmtree(initial_repo_dir, ignore_errors=True)
+
     for candidate in source_candidates:
-        # this is not really changing chdir - we are setting the same folder on a new inode
-        os.chdir(restore_clean_workspace("empty"))
+        # Clean workspace by removing all files except .dlt and .global_dir
```
tetelio, line 317: manual file cleaning is safer for parallel pytest xdist workers than changing directories

tests/workspace/cli/utils.py
```diff
@@ -12,16 +12,11 @@
 
 from dlt._workspace.cli import echo, DEFAULT_VERIFIED_SOURCES_REPO, DEFAULT_VIBE_SOURCES_REPO
```
tetelio, line 12: Repository directories are now created relative to the current working directory instead of a global test storage root. Workspace file access also consistently uses the current directory. This aligns the tests with how the CLI actually behaves, since the CLI is inherently cwd-driven.

tests/workspace/helpers/dashboard/conftest.py
```diff
@@ -20,13 +20,13 @@ def pipeline(request):
     return request.getfixturevalue(request.param)
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture
```
tetelio, line 23: Most fixtures were changed from session-scoped to function-scoped. This prevents dashboard tests from sharing pipeline state across tests. A dedicated temporary pipelines directory fixture was added to support ordering and timestamp-based tests without relying on persistent filesystem state.

tests/workspace/helpers/dashboard/example_pipelines.py
```diff
@@ -119,9 +121,10 @@ def create_success_pipeline_duckdb(pipelines_dir: str = None, db_conn: Any = Non
 
 
 def create_success_pipeline_filesystem(
-    pipelines_dir: str = None, bucket_url: str = "_storage/data"
+    pipelines_dir: str = None,
 ):
     """Create a test pipeline with filesystem destination, properties see `run_success_pipeline`"""
+    bucket_url: str = str(Path(get_test_storage_root()) / "data")
```
tetelio, line 127: Filesystem-based pipelines now store data under the test storage root instead of _storage/data. This ensures that filesystem destinations are isolated per test worker and do not collide when tests run concurrently.

tests/workspace/helpers/dashboard/test_cli_utils.py
```diff
@@ -0,0 +1,93 @@
+import tempfile
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_dashboard_config.py
```diff
@@ -0,0 +1,46 @@
+import os
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_data_formatting.py
```diff
@@ -0,0 +1,147 @@
+from datetime import datetime
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_loads.py
```diff
@@ -0,0 +1,140 @@
+import pytest
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_pipeline_operations.py
```diff
@@ -0,0 +1,140 @@
+import pytest
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_pipeline_workflow.py
```diff
@@ -0,0 +1,64 @@
+import marimo as mo
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_query_operations.py
```diff
@@ -0,0 +1,60 @@
+import pytest
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_table_operations.py
```diff
@@ -0,0 +1,132 @@
+import pytest
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_trace_operations.py
```diff
@@ -0,0 +1,106 @@
+import pytest
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tests/workspace/helpers/dashboard/test_visualization.py
```diff
@@ -0,0 +1,85 @@
+import pytest
```
tetelio, line 1: This file contains tests that existed previously in tests_utils.py without modifications. Only for better organisation

tetelio:
Motherduck destination will not pass because i need to change the credentials structure but I didnt want to break the CI pipeline for others. Motherduck tests pass locally.

.github/workflows/test_common.yml
```diff
@@ -63,11 +62,9 @@ jobs:
           - os: windows-latest
             python-version: "3.11"
             shell: cmd
-            pytest_args: '-m "not forked and not rfam"'
```
tetelio, line 66: offloaded to Makefile since this is an OS concern and not a CI concern

tests/common/configuration/test_credentials.py
```diff
@@ -351,13 +334,30 @@ def test_needs_scopes_for_refresh_token() -> None:
 
 
 def test_requires_refresh_token_no_tty():
+    # NOTE:
+    # This test must NOT rely on the presence or absence of a real TTY.
+    #
+    # Under pytest-xdist, worker processes do not have a TTY, which can cause
```
rudolfix: AFAIK no interactive flow happens if process is not at tty so this sentence is not true or I do not understand it. overall patching function for consistent behavior makes sense

tests/common/configuration/test_credentials.py
rudolfix: probably console is attached. or maybe there's no entry module:
```
assert (
                sys.stdin.isatty() or is_interactive()
            ), "Must have a tty or interactive mode for web flow"
```

tests/common/configuration/utils.py
```diff
@@ -109,7 +114,13 @@ def on_resolved(self) -> None:
 @pytest.fixture(scope="function")
 def environment() -> Any:
     saved_environ = environ.copy()
```
rudolfix, line 116: you  could just use `custom_environ` ctx manager `custom_environ` instead

tests/common/schema/test_import_normalizers.py
rudolfix: something must be leaking destination context... OK to remove that

tests/conftest.py
```diff
@@ -56,11 +59,19 @@ def pytest_configure(config):
     # patch the configurations to use test storage by default, we modify the types (classes) fields
     # the dataclass implementation will use those patched values when creating instances (the values present
     # in the declaration are not frozen allowing patching). this is needed by common storage tests
+    from tests.utils import get_test_worker_id
+
+    worker = get_test_worker_id()
```
rudolfix: why this is here and not in fixture that creates test storage. was not needed previously and IMO still should not be needed. did you try to move this code to `utils.py`?

tests/load/lancedb/fixtures.py
rudolfix: why this is needed? lance databases are created in `local_dir` which points to test storage that varies per worker. make sure `auto_test_run_context` applies. it may be also connected to 
```
def get_test_storage_root() -> str:
    return os.environ.get(DLT_TEST_STORAGE_ROOT, "_storage")
```
which uses env variable set by pytest_init which I pointed out in other commenet. maybe if you compute the path each time it is needed the problem goes away

tests/load/utils.py
rudolfix: this again should not be needed. something is wrong somewhere else. under xdist you should automatically create databases in test storage, if that does not happen we must look why

tests/load/utils.py
rudolfix: we have one shared motherduck database for all tests and it was already working in parallel

tests/pipeline/test_dlt_versions.py
```diff
@@ -33,7 +34,30 @@
...
+
+
+@contextmanager
+def test_workspace(extra_env: Optional[Dict[str, str]] = None):
+    """
+    Sets up a real DLT workspace exactly like CLI execution.
+
+    This mirrors create_test_run_context behavior:
+    - runs inside a workspace directory
+    - activates WorkspaceRunContext
```
rudolfix: 1. where is this activation? we need to create .workspace file somewhere
2. note that this runs very old versions of dlt in Venv, they do not know what workspace is
3. this function sets `DLT_DATA_DIR` which moves data dir to test storage which is fine but that's not workspace...

tests/workspace/cli/conftest.py
rudolfix: why we need this? `workspace` tests should actually run in parallel without any modifications because the workspace is created in random dir per test. tldr;> it should work with hardcoded `_storage` folder. OFC we have a fixture that deletes the whole folder so we still need dynamic name but in principle very little modifications should be needed.

tests/workspace/utils.py
```diff
@@ -30,11 +30,16 @@ def isolated_workspace(
     """
     new_run_dir = restore_clean_workspace(name)
     with set_working_dir(new_run_dir):
-        ctx = switch_context(new_run_dir, profile=profile, required=required)
-        assert ctx.run_dir == new_run_dir
+        ctx = switch_context(".", profile=profile, required=required)
```
rudolfix: 1. why we change this function at all? it should depend on test storage root

tests/workspace/utils.py
```diff
@@ -30,11 +30,16 @@ def isolated_workspace(
...
         # also mock global dir so it does not point to default user ~
         if isinstance(ctx, WorkspaceRunContext):
-            ctx._global_dir = os.path.abspath(".global_dir")
+            ctx._global_dir = os.path.join(
```
rudolfix: what was wrong with the old `_global_dir`? we already change the working dir to isolated folder so old code whould be fine

tests/common/configuration/utils.py
tetelio: no, this `environment` fixture does `environ.clear()`, which fully isolates the environment from other tests, its a clean slate. i dont want to temporary set some env values and then remove them, while preserving all the previous environment variables

tests/conftest.py
tetelio: I will remove the cleaning of the folder here, that should be managed by fixtures, as you said

However, due to the current test setup, we still need to eagerly set the test storage root environment variable here. Some tests resolve values that depend on get_test_storage_root() at import time, and without initializing the env var early, those imports fail.

Long term, we should avoid resolving test configuration at import time and instead compute it lazily (eg. inside fixtures or runtime helpers). That would require refactoring a significant number of tests, so for now this keeps the behavior correct without a large disruptive change. For now import time was ok because there was only one test storage root, but not anymore

example of test that evaluates at import time
https://github.com/dlt-hub/dlt/blob/9f2917b49cc2e4d83d058e76f0dfccc8b581b2cb/tests/workspace/utils.py#L19

tests/pipeline/test_dlt_versions.py
tetelio: true, the docstring was just wrong, rest stays. probably i was switching back and forth when touching the real workspace tests. also the previous context manager was csalled `with test_workspace`, my bad

tetelio:
@rudolfix regarding https://github.com/dlt-hub/dlt/pull/3479#discussion_r2678836746 (somehow i cant respond on the thread)

LanceDB uses shared on-disk state per database directory, and dataset names don’t provide real isolation—they’re only table-name prefixes. dlt’s dataset concept assumes schema-level isolation (as in SQL engines), but LanceDB’s local backend has a flat catalog, so all tables for all datasets in the same directory share metadata. Under pytest-xdist, workers race during storage initialization, specifically on creating the sentinel table (_dltSentinelTable), which is created unconditionally on init and thus collides across workers even with different dataset names. That’s why tests pass serially but fail in parallel. You’re right that we don’t need per-test temp dirs—scoping the LanceDB directory by worker id is enough to avoid this race.

Update: i actually. know why there are collisions
The secrets.toml
[destination.lancedb.credentials]
uri = "/tmp/.lancedb"

So this is the fallback URL for lancedb if a uri is not specified.
We can
1. remove the uri in the secrets.toml and the lancedb files are correctly created in the testing storage (but lets see what happens with other tests, and we need to notify devs that are running these tests with credentials to modify the secrets.toml)

2. Or we keep the default and we override the URI when needed.

I think more correct doing t, do you see any problems?

tetelio:
@rudolfix regarding ducklake
https://github.com/duckdb/ducklake/issues/459
people have observed the asme problems i am observing with same ducklake name (catalog) and concurrent writers

```
dlt/pipeline/pipeline.py:612: PipelineStepFailed
________________________________________________________________ test_drop_command_resources_and_state[False-ducklake-no-staging] ________________________________________________________________
[gw11] darwin -- Python 3.11.14 /Users/alberto/code/official_repos/dlt/.venv/bin/python3

self = <dlt.destinations.impl.ducklake.sql_client.DuckLakeSqlClient object at 0x13a793310>, args = (), kwargs = {}
db_ex = DatabaseTransientException(TransactionException('TransactionContext Error: Failed to commit: Failed to commit DuckLake...lake_data_file_pkey"\nDETAIL:  Key (data_file_id)=(668) already exists.\nCONTEXT:  COPY ducklake_data_file, line 1\n'))

    @wraps(f)
    def _wrap(self: SqlClientBase[Any], *args: Any, **kwargs: Any) -> Any:
        try:
>           return f(self, *args, **kwargs)

args       = ()
db_ex      = DatabaseTransientException(TransactionException('TransactionContext Error: Failed to commit: Failed to commit DuckLake...lake_data_file_pkey"\nDETAIL:  Key (data_file_id)=(668) already exists.\nCONTEXT:  COPY ducklake_data_file, line 1\n'))
f          = <function DuckDbSqlClient.commit_transaction at 0x13a40d940>
kwargs     = {}
self       = <dlt.destinations.impl.ducklake.sql_client.DuckLakeSqlClient object at 0x13a793310>

dlt/destinations/sql_client.py:462: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <dlt.destinations.impl.ducklake.sql_client.DuckLakeSqlClient object at 0x13a793310>

    @raise_database_error
    def commit_transaction(self) -> None:
>       self._conn.commit()
E       duckdb.duckdb.TransactionException: TransactionContext Error: Failed to commit: Failed to commit DuckLake transaction: Failed to write data file information to DuckLake: Failed to copy data: ERROR:  duplicate key value violates unique constraint "ducklake_data_file_pkey"
E       DETAIL:  Key (data_file_id)=(668) already exists.
E       CONTEXT:  COPY ducklake_data_file, line 1

self       = <dlt.destinations.impl.ducklake.sql_client.DuckLakeSqlClient object at 0x13a793310>

dlt/destinations/impl/duckdb/sql_client.py:158: TransactionException
```

tetelio:
@rudolfix i checked motherduck and there are race problems with tests/load/test_job_client.py

i think there is only one schema and one staging schema for the whole test file and only table names change with uniq_id(). When initializing storage/updating schema with many parallel workers, race conditions happen. I have repeated tests many times and they fail deterministically depending on if i use a database per worker or not


```
The above exception was the direct cause of the following exception:

client = <dlt.destinations.impl.motherduck.motherduck.MotherDuckClient object at 0x13ec6f6d0>, write_disposition = 'merge', replace_strategy = ''
file_storage = <dlt.common.storages.file_storage.FileStorage object at 0x13ec5e050>

    @pytest.mark.parametrize(
        "write_disposition,replace_strategy",
        [
            ("append", ""),
            ("merge", ""),
            ("replace", "truncate-and-insert"),
            ("replace", "insert-from-staging"),
            ("replace", "staging-optimized"),
        ],
    )
    @pytest.mark.parametrize(
        "client",
        destinations_configs(default_sql_configs=True, table_format_filesystem_configs=True),
        indirect=True,
        ids=lambda x: x.name,
    )
    def test_write_dispositions(
        client: SqlJobClientBaseWithDestinationTestConfiguration,
        write_disposition: TWriteDisposition,
        replace_strategy: TLoaderReplaceStrategy,
        file_storage: FileStorage,
    ) -> None:
        if not client.capabilities.preferred_loader_file_format:
            pytest.skip("preferred loader file format not set, destination will only work with staging")
        # set filesystem views to autorefresh
        set_always_refresh_views(client.config)
    
        table_name = "event_test_table" + uniq_id()
        column_schemas, data_row = get_columns_and_row_all_types(client.config)
        root_table = new_table(
            table_name, write_disposition=write_disposition, columns=column_schemas.values()
        )
        root_table["table_format"] = client.destination_config.table_format
        root_table["file_format"] = client.destination_config.file_format
        child_table_name = client.schema.naming.make_path(table_name, "child")
        # add child table without write disposition so it will be inferred from the parent
        child_table = new_table(
            child_table_name, columns=column_schemas.values(), parent_table_name=table_name
        )
        client.schema.update_table(root_table)
        client.schema.update_table(child_table)
    
        prepared_root_table = client.prepare_load_table(table_name)
    
        if write_disposition == "replace":
            available_replace_strategy = resolve_replace_strategy(
                prepared_root_table, replace_strategy, client.capabilities
            )
            if not available_replace_strategy:
                pytest.skip(f"replace strategy {replace_strategy} not available on this destination")
            client.config.replace_strategy = replace_strategy
    
        if write_disposition == "merge":
            available_merge_strategy = resolve_merge_strategy(
                client.schema.tables, prepared_root_table, client.capabilities
            )
            if not available_merge_strategy:
                pytest.skip("destination does not support merge")
            # add root key
            # client.schema.tables[table_name]["columns"]["col1"]["root_key"] = True
            # add primary key
            client.schema.tables[table_name]["columns"]["col1"]["primary_key"] = True
            # client.schema.tables[table_name]["table_format"] =
    
        client.schema._bump_version()
        client.update_stored_schema()
    
        truncate_tables = [table_name]
    
        # create staging for merge/replace dataset
        if isinstance(client, WithStagingDataset):
            with client.with_staging_dataset():
                client.initialize_storage()
                client.schema._bump_version()
>               client.update_stored_schema()

available_merge_strategy = 'delete-insert'
child_table = {'columns': {'col1': {'data_type': 'bigint', 'name': 'col1', 'nullable': False}, 'col10': {'data_type': 'date', 'name'...event_test_table22e2174009434a0ef2f88de986a171e3__child', 'parent': 'event_test_table22e2174009434a0ef2f88de986a171e3'}
child_table_name = 'event_test_table22e2174009434a0ef2f88de986a171e3__child'
client     = <dlt.destinations.impl.motherduck.motherduck.MotherDuckClient object at 0x13ec6f6d0>
column_schemas = {'col1': {'data_type': 'bigint', 'name': 'col1', 'nullable': False}, 'col10': {'data_type': 'date', 'name': 'col10', '...l', 'nullable': True}, 'col12': {'data_type': 'timestamp', 'name': 'col12', 'nullable': False, 'timezone': False}, ...}
data_row   = {'col1': 989127831, 'col10': Date(2023, 2, 27), 'col10_null': None, 'col12': DateTime(2299, 12, 31, 0, 0, 1), ...}
file_storage = <dlt.common.storages.file_storage.FileStorage object at 0x13ec5e050>
prepared_root_table = {'columns': {'col1': {'data_type': 'bigint', 'name': 'col1', 'nullable': False}, 'col10': {'data_type': 'date', 'name'...event_test_table22e2174009434a0ef2f88de986a171e3', 'resource': 'event_test_table22e2174009434a0ef2f88de986a171e3', ...}
replace_strategy = ''
root_table = {'columns': {'col1': {'data_type': 'bigint', 'name': 'col1', 'nullable': False}, 'col10': {'data_type': 'date', 'name'...event_test_table22e2174009434a0ef2f88de986a171e3', 'resource': 'event_test_table22e2174009434a0ef2f88de986a171e3', ...}
table_name = 'event_test_table22e2174009434a0ef2f88de986a171e3'
truncate_tables = ['event_test_table22e2174009434a0ef2f88de986a171e3']
write_disposition = 'merge'

tests/load/test_job_client.py:801: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
dlt/destinations/job_client_impl.py:314: in update_stored_schema
    applied_update = self._execute_schema_update_sql(only_tables)
        __class__  = <class 'dlt.destinations.job_client_impl.SqlJobClientBase'>
        applied_update = {}
        expected_update = None
        only_tables = None
        schema_info = None
        self       = <dlt.destinations.impl.motherduck.motherduck.MotherDuckClient object at 0x13ec6f6d0>
dlt/destinations/job_client_impl.py:637: in _execute_schema_update_sql
    self.sql_client.execute_many(sql_scripts)
        only_tables = None
        schema_update = {'_dlt_loads': {'columns': {'inserted_at': {'data_type': 'timestamp', 'name': 'inserted_at', 'nullable': False}, 'load...test_table22e2174009434a0ef2f88de986a171e3', 'resource': 'event_test_table22e2174009434a0ef2f88de986a171e3', ...}, ...}
        self       = <dlt.destinations.impl.motherduck.motherduck.MotherDuckClient object at 0x13ec6f6d0>
        sql_scripts = ['CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BI...  NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL)']
dlt/destinations/sql_client.py:192: in execute_many
    ret.append(self.execute_sql(sql_fragment + ";", *args, **kwargs))
        args       = ()
        kwargs     = {}
        ret        = []
        self       = <dlt.destinations.impl.motherduck.sql_client.MotherDuckSqlClient object at 0x13ec6cf50>
        sql_fragment = 'CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BIG...R  NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL)'
        statements = ['CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BI...  NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL)']
dlt/destinations/impl/duckdb/sql_client.py:171: in execute_sql
    with self.execute_query(sql, *args, **kwargs) as curr:
        args       = ()
        kwargs     = {}
        self       = <dlt.destinations.impl.motherduck.sql_client.MotherDuckSqlClient object at 0x13ec6cf50>
        sql        = 'CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BIG...  NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL);'
/opt/homebrew/Cellar/python@3.11/3.11.14_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
        self       = <contextlib._GeneratorContextManager object at 0x13ec7dd10>
dlt/destinations/sql_client.py:457: in _wrap_gen
    raise db_ex.with_traceback(ex.__traceback__) from ex
        args       = ('CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BI...NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL);',)
        db_ex      = DatabaseUndefinedRelation(CatalogException('Catalog Error: Schema with name test_7c1947fd767d0ba7de0f5b01e60d60a1_staging does not exist!'))
        f          = <function DuckDbSqlClient.execute_query at 0x12f2d6de0>
        kwargs     = {}
        self       = <dlt.destinations.impl.motherduck.sql_client.MotherDuckSqlClient object at 0x13ec6cf50>
dlt/destinations/sql_client.py:451: in _wrap_gen
        args       = ('CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BI...NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL);',)
        db_ex      = DatabaseUndefinedRelation(CatalogException('Catalog Error: Schema with name test_7c1947fd767d0ba7de0f5b01e60d60a1_staging does not exist!'))
        f          = <function DuckDbSqlClient.execute_query at 0x12f2d6de0>
        kwargs     = {}
        self       = <dlt.destinations.impl.motherduck.sql_client.MotherDuckSqlClient object at 0x13ec6cf50>
dlt/destinations/impl/duckdb/sql_client.py:195: in execute_query
    raise outer
        args       = ()
        cur        = <duckdb.duckdb.DuckDBPyConnection object at 0x12f0ad170>
        db_args    = None
        kwargs     = {}
        query      = 'CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BIG...  NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL);'
        self       = <dlt.destinations.impl.motherduck.sql_client.MotherDuckSqlClient object at 0x13ec6cf50>
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <dlt.destinations.impl.motherduck.sql_client.MotherDuckSqlClient object at 0x13ec6cf50>
query = 'CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BIG...  NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL);'
args = (), kwargs = {}, db_args = None, cur = <duckdb.duckdb.DuckDBPyConnection object at 0x12f0ad170>

    @contextmanager
    @raise_database_error
    def execute_query(self, query: AnyStr, *args: Any, **kwargs: Any) -> Iterator[DBApiCursor]:
        assert isinstance(query, str)
        db_args = args if args else kwargs if kwargs else None
        if db_args:
            # TODO: must provide much better refactoring of params
            query = query.replace("%s", "?")
    
        # we can't duplicate connection because local settings (ie. search path) are lost
        cur = self._conn
        try:
>           yield self.cursor_impl(cur.execute(query, db_args))  # type: ignore
E           dlt.destinations.exceptions.DatabaseUndefinedRelation: Catalog Error: Schema with name test_7c1947fd767d0ba7de0f5b01e60d60a1_staging does not exist!

args       = ()
cur        = <duckdb.duckdb.DuckDBPyConnection object at 0x12f0ad170>
db_args    = None
kwargs     = {}
query      = 'CREATE TABLE IF NOT EXISTS "dlt_data"."test_7c1947fd767d0ba7de0f5b01e60d60a1_staging"."_dlt_version" (\n"version" BIG...  NOT NULL,\n"col6_precision" DECIMAL(6,2)  NOT NULL,\n"col7_precision" BLOB  NOT NULL,\n"col12" TIMESTAMP  NOT NULL);'
self       = <dlt.destinations.impl.motherduck.sql_client.MotherDuckSqlClient object at 0x13ec6cf50>

dlt/destinations/impl/duckdb/sql_client.py:189: DatabaseUndefinedRelation
```

PR #3481, opened 2025-12-15
3480 chore update weaviate destination to use v4 client

djudjuu:
internally move to v4 client which has better batching and error handling. 
additional config value connection_type = "local" | "cloud" | "custom" is created to facilitate connecting to different weaviate instances 
`destination=dlt.destinations.weaviate(connection_type="local")`
type deduced from api-url and then used to set the ports to the default values

follow up: 
- more options for vectorization exist in the v4 client (quantization, multi-vector-embedding, bring your own vector) which are really useful. These configuration options need yet to be made available to the user, probably in a follow up PR.

### DIFF
--- a/docs/uv.lock
+++ b/docs/uv.lock
@@ -951,6 +951,7 @@ requires-dist = [
     { name = "lancedb", marker = "python_full_version < '3.13' and extra == 'lancedb'", specifier = ">=0.22.0" },
     { name = "marimo", marker = "extra == 'workspace'", specifier = ">=0.14.5" },
     { name = "mcp", marker = "python_full_version >= '3.10' and extra == 'workspace'", specifier = ">=1.2.1" },
+    { name = "oracledb", marker = "extra == 'oracle'", specifier = ">=3.4.1" },
     { name = "orjson", marker = "python_full_version >= '3.14'", specifier = ">=3.11.0" },
     { name = "orjson", marker = "platform_python_implementation != 'PyPy' and sys_platform != 'emscripten'", specifier = ">=3.6.7,!=3.9.11,!=3.9.12,!=3.9.13,!=3.9.14,!=3.10.1,<4" },
     { name = "orjson", marker = "platform_python_implementation != 'PyPy' and sys_platform != 'emscripten'", specifier = ">=3.10.1" },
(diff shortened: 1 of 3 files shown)


djudjuu:
### 1. New gRPC Port Requirement (50051)

**What changed:** The Weaviate Python client v4 uses gRPC for improved performance on batch operations and queries. This requires port **50051** to be accessible.

**Why it matters:** If you're running Weaviate locally with Docker, you must expose the gRPC port in addition to the HTTP port:

**What changed:** Weaviate v4 renamed "classes" to "collections" throughout the API.

**Impact on dlt:** The internal implementation now uses collection-based APIs, but **backwards compatibility is maintained** - you can still use the same dlt pipeline code without changes.

### DataType and Tokenization

## Tokenization Support

 The v4 implementation properly exposes tokenization configuration through the Property class.
(typed!)

**Available tokenization methods:**
- `Tokenization.WORD` - Split on whitespace and punctuation (default)
- `Tokenization.LOWERCASE` - Like WORD, but lowercased
- `Tokenization.WHITESPACE` - Split only on whitespace
- `Tokenization.FIELD` - Index entire field as one token

**Usage with weaviate_adapter (unchanged):**
```python
weaviate_adapter(
    data,
    vectorize=["title"],
    tokenization={"title": "word", "description": "whitespace"},
)
```

## Batching Differences

### v3 Batch API
```python
with client.batch(
    batch_size=100,
    timeout_retries=5,
    callback=check_results,
) as batch:
    batch.add_data_object(data, class_name, uuid=uuid)
```

### v4 Batch API
```python
collection = client.collections.get(collection_name)
with collection.batch.dynamic() as batch:
    batch.add_object(properties=data, uuid=uuid)

# Check for failures after batch completes
failed_objects = collection.batch.failed_objects
```

**Key differences:**
1. **Collection-scoped:** Batches are now tied to a specific collection
2. **Dynamic batching:** `batch.dynamic()` automatically optimizes batch sizes
3. **Failure handling:** Failed objects are collected in `batch.failed_objects` instead of a callback
4. **Simpler API:** Properties are passed directly, no need to specify class name per object

### Performance Improvements
- **gRPC support:** Faster batch imports and queries through binary protocol
- **Dynamic batching:** Automatic optimization of batch sizes based on object size
  -> not configurable via dlt `config.toml` but could be done
- **Connection pooling:** Better resource management for concurrent operations

### Improved Error Handling
- More descriptive error messages
- Structured error objects for batch failures
- Better retry handling built into the client

rudolfix (approved):
LGTM! I won't be able to fully review this because we mostly do some tricky weaviate client stuff. I do not see anything suspicious and tests are passing

PR #3483, opened 2025-12-15
Add column description support for BigQuery destination

### LINKED ISSUE #2179: feat: Add column description support for BigQuery adapter
### Feature description

I'd like to suggest adding support for column descriptions in the BigQuery adapter. This would allow users to define and manage column-level metadata when loading data into BigQuery tables. This feature would enhance the data documentation and governance capabilities within the dlt BigQuery integration.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

When loading data into BigQuery, having clear and accurate column descriptions is critical for data governance and documentation purposes. Currently, DLT doesn't provide a way to define column descriptions during data loading, making it difficult to maintain metadata and properly document data structures.

**Problem solved:**

- Data teams must manually add column descriptions after data is loaded.
- Risk of inconsistent or missing column documentation
- Additional overhead of maintaining separate data dictionaries
- Difficulty in tracking and versioning changes to column metadata

### Proposed solution

Implement column description support in the BigQuery adapter by

Extending the BigQuery adapter interface to accept column descriptions during schema definition
Modify dlt/destinations/impl/bigquery/bigquery_adapter.py and dlt/destinations/impl/bigquery/bigquery.py to add column description parameters:

```python:dlt/destinations/impl/bigquery/bigquery.py
class BigQueryClient(SqlJobClientWithStagingDataset, SupportsStagingDestination):
    ...
    def _get_column_def_sql(self, column: TColumnSchema, table: PreparedTableSchema = None) -> str:
        column_def_sql = super()._get_column_def_sql(column, table)
        if column.get(ROUND_HALF_EVEN_HINT, False):
            column_def_sql += " OPTIONS (rounding_mode='ROUND_HALF_EVEN')"
        if column.get(ROUND_HALF_AWAY_FROM_ZERO_HINT, False):
            column_def_sql += " OPTIONS (rounding_mode='ROUND_HALF_AWAY_FROM_ZERO')"
        
        # adding this
        if "description" in column:
            column_def_sql += f" OPTIONS (description='{column['description']}')"
        
        return column_def_sql
    ...
```

```python:dlt/destinations/impl/bigquery/bigquery_adapter.py
def bigquery_adapter(
    data: Any,
    partition: TColumnNames = None,
    cluster: TColumnNames = None,
    round_half_away_from_zero: TColumnNames = None,
    round_half_even: TColumnNames = None,
    table_description: Optional[str] = None,
    table_expiration_datetime: Optional[str] = None,
    insert_api: Optional[Literal["streaming", "default"]] = None,
    autodetect_schema: Optional[bool] = None,
    partition_expiration_days: Optional[int] = None,
    column_descriptions: Optional[Dict[str, str]] = None,  # Added parameter
) -> DltResource:
    ...
    # adding column description
    if column_descriptions:
        for column_name, description in column_descriptions.items():
            if column_name in column_hints:
                column_hints[column_name]["description"] = description
            else:
                column_hints[column_name] = {"name": column_name, "description": description}
    
    if partition:
        ...
```

The implementation would match BigQuery's native support for column descriptions, while maintaining dlt's current functionality and performance.
An example usage could look like this

```python
data = [{"id": 1, "name": "Alice"}, {"id": 2, "name": "Bob"}]
column_descriptions = {"id": "User ID", "name": "User Name"}

pipeline = dlt.pipeline(
    pipeline_name="description_test",
    destination="bigquery",
    dataset_name="description_test",
    export_schema_path="dlt_schemas/export",
)
load_info = pipeline.run(
    bigquery_adapter(data, column_descriptions=column_descriptions),
    table_name="users",
    write_disposition="replace",
    refresh="drop_data",
)

print(load_info)
```

### Related issues

No related issues found.

Travior:
This PR adds support for the `description` column hint in BigQuery. Additionally explicit string escaping for BigQuery is added

Closes #2179

### DIFF
--- a/tests/load/bigquery/test_bigquery_table_builder.py
+++ b/tests/load/bigquery/test_bigquery_table_builder.py
@@ -1,6 +1,7 @@
 import os
+import re
 from copy import deepcopy
-from typing import Iterator, Dict, Any, List
+from typing import Iterator, Dict, Any, List, Union
 
 import google
 import pytest
@@ -551,18 +552,28 @@ def test_bigquery_no_partition_by_integer(
             assert not has_partitions
 
 
-def test_bigquery_column_description(gcp_client: BigQueryClient) -> None:
+@pytest.mark.parametrize(
+    "description,expect_column_options",
+    [
+        ("This is a test description for column col1", True),
+        (None, False),
+        ("", False),
+    ],
+)
+def test_bigquery_column_description(
+    gcp_client: BigQueryClient, description: Union[str, None], expect_column_options: bool
+) -> None:
     columns = deepcopy(TABLE_UPDATE)
-    for column in columns:
-        column["description"] = f"This is a test description for column {column['name']}"
+    columns[0]["description"] = description
 
     sql = gcp_client._get_table_update_sql("event_test_table", columns, False)[0]
     sqlfluff.parse(sql, dialect="bigquery")
     assert "event_test_table" in sql
-    for column in columns:
-        assert (
-            f"OPTIONS (description='This is a test description for column {column['name']}')" in sql
-        )
+
+    if expect_column_options:
+        assert (re.search(r"OPTIONS \(description='.*'\)", sql)) is not None
+    else:
+        assert "OPTIONS" not in sql
 
 
 def test_bigquery_column_description_character_escaping(gcp_client: BigQueryClient) -> None:


dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -465,10 +466,22 @@ def _get_info_schema_columns_query(
 
     def _get_column_def_sql(self, column: TColumnSchema, table: PreparedTableSchema = None) -> str:
         column_def_sql = super()._get_column_def_sql(column, table)
+
+        # generate additional column options clause
```
burnash, line 470: Thanks for fixing this!

tests/common/data_writers/test_data_writers.py
```diff
@@ -175,6 +177,32 @@ def test_string_literal_escape_unicode() -> None:
...
+    # test backslash escaping
+    assert escape_bigquery_literal("path\\to\\file") == "'path\\\\to\\\\file'"
+    # test unicode passthrough
+    assert (
+        escape_bigquery_literal("イロハニホヘト チリヌルヲ ワカヨタレソ ツネナラム")
+        == "'イロハニホヘト チリヌルヲ ワカヨタレソ ツネナラム'"
+    )
+
+
+def test_identifier_escape_bigquery_literal() -> None:
```
burnash: Is the function name a misnomer?

```suggestion
def test_escape_bigquery_identifier() -> None:
```

tests/common/data_writers/test_data_writers.py
```diff
@@ -175,6 +177,32 @@ def test_string_literal_escape_unicode() -> None:
     )
 
 
+def test_string_literal_escape_bigquery() -> None:
```
burnash: This is good, however only string values are tested here. Please consider adding other types handled by `escape_bigquery_literal` e.g. datetime, collection types, bytes, None, etc.

tests/load/bigquery/test_bigquery_table_builder.py
```diff
@@ -534,6 +534,66 @@ def demo_source() -> DltResource:
             assert not has_partitions
 
 
+def test_bigquery_column_description(gcp_client: BigQueryClient) -> None:
```
burnash: Please test for edge cases for column description: empty string, None, etc. You can parametrize this test function, for example.

tests/load/bigquery/test_bigquery_table_builder.py
```diff
@@ -534,6 +534,66 @@ def demo_source() -> DltResource:
...
+    assert "event_test_table" in sql
+    assert r"OPTIONS (description='\'\';) DROP TABLE --')" in sql
+
+
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["bigquery"]),
+    ids=lambda x: x.name,
+)
+def test_bigquery_with_column_description_and_rounding_mode_hints(
```
burnash, line 595: Very good!

tests/load/bigquery/test_bigquery_table_builder.py
Travior: What are we expecting to happen? I reckon that bigquery would likely throw an error for the generated query if we try to set `description=NULL` (which would be the output when escaping `None`), because it expects a string literal. Should this test just assert that this "incorrect" SQL is generated properly (to test the non-happy path) or should we have some string conversion before?

burnash: What I'm suggesting is a negative test: if I understood the code correctly, setting `column["description"] = None`, or empty string in line 540 below, should not produce correct SQL withot any OPTIONS clause.

burnash:
@Travior thanks for the fixes, I've added a comment about the negative test above https://github.com/dlt-hub/dlt/pull/3483#discussion_r2680771602

tests/load/bigquery/test_bigquery_table_builder.py
Travior: I see what you mean, I have refactored the test to test those edge cases

burnash (approved):
Looks good, thank you, @Travior

PR #3496, opened 2025-12-18
Feat/3362 - add oracle testing setup

ivasio:
# Changes in the testing setup
- Added oracle, now oracle source tests can be added. Very basic set of tests is currently there, to be improved in other PRs
- There's now 3 groups of "local source" tests running in a matrix in CI: those that use oracle, mssql or postgres/other sources. When adding new tests for the first 2 categories, `mssql` and `oracle` pytest marks should be used, e.g. like [here](https://github.com/dlt-hub/dlt/pull/3496/files#diff-347261338567cd1395ed8ab6d7f5a4c1b4a0d6fc2f5c48c57b7ead2d3b1e837cR34)

# Oracle - current changes and plans
- Currently, we're [claiming](https://dlthub.com/docs/dlt-ecosystem/verified-sources/sql_database/troubleshooting#oracle) we don't fully support `oracledb` driver in thin mode. The direction me and @rudolfix have chosen is to actually support it as much as we can and test/document the caveats. The older `cx_Oracle` has been deprecated for 3 years now in favour of `oracledb`, so we don't want to promote it's usage and don't test it
- In this PR, an `oracle` extra has been added which installs only the `oracledb` driver. In fact, it's only usable with `sqlalchemy>2`, but I couldn't add this requirement to avoid the conflict with requirement groups defined in pyproject.toml
- Documentation etc. to be added in separate PRs. This one is aimed to unlock the other ones in terms of testing setup

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -404,5 +404,7 @@ filterwarnings = ["ignore::DeprecationWarning"]
 markers = [
     "essential: marks all essential tests",
     "no_load: marks tests that do not load anything",
-    "rfam: marks tests that use rfam db"
+    "rfam: marks tests that use rfam db",
+    "oracle: marks tests that use local oracle db",
+    "mssql: marks tests that use local mssql db"
 ]


tests/load/sources/sql_database/conftest.py
```diff
@@ -4,6 +4,7 @@
 
 import dlt
 from dlt.sources.credentials import ConnectionStringCredentials
+from tests.load.sources.sql_database.oracle_source import OracleSourceDB
```
burnash, line 7: This import is outside of try/except below. Is this intentional? Not sure if it errors on tests collection if the oracle extra is not installed (I don't see any driver usage in OracleSourceDB)

ivasio: it actually wasn't intentional, but test collection works fine as can be seen in local sources tests for [postgres](https://github.com/dlt-hub/dlt/actions/runs/20461028938/job/59077266890) and [mssql](https://github.com/dlt-hub/dlt/actions/runs/20461028938/job/59077267198) where oracle extra is not installed. just tried removing the try/except below and feel it's there for a good reason, but oracle doesn't need this try/except I'd say

burnash (approved):

PR #3497, opened 2025-12-18
Fix cluster hint overriding partition hint on bigquery

### LINKED ISSUE #3471: bigquery_adapter() silently drops partition hint when same column is used for both partition and cluster
### dlt version

1.14.1 (but the bug is present in the latest devel branch code as well)

### Describe the problem

When using `bigquery_adapter()` with the same column for both partition and cluster parameters, the partition hint is silently overwritten by the cluster hint. This results in tables being created with clustering but no partitioning, even though partitioning was explicitly requested.
This is because column_hints is a dictionary keyed by column name, and the cluster hint assignment overwrites the partition hint assignment.

### Expected behavior

When the same column is specified for both partition and cluster:

Option 1: dlt should raise a clear ValueError explaining that a column cannot be both the partition and cluster column, since BigQuery's partition column already provides ordering benefits.

Option 2: dlt should merge the hints, resulting in column_hints[column] = {"name": column, PARTITION_HINT: True, CLUSTER_HINT: True} (this may not be useful since partitioning on a column already provides similar benefits to clustering on it).

### Steps to reproduce

```python
import dlt
from dlt.destinations.impl.bigquery.bigquery_adapter import bigquery_adapter, PARTITION_HINT, CLUSTER_HINT

@dlt.resource(columns=[{"name": "timestamp_col", "data_type": "timestamp"}])
def test_data():
    yield {"timestamp_col": "2024-01-01T00:00:00Z"}

# Apply adapter with same column for partition and cluster
bigquery_adapter(
    test_data,
    partition="timestamp_col",
    cluster=["timestamp_col", "other_col"]
)

# Check the column hints - PARTITION_HINT is missing!
print(test_data.columns)
# Output: {'timestamp_col': {'name': 'timestamp_col', 'x-bigquery-cluster': True}, ...}
# Expected: {'timestamp_col': {'name': 'timestamp_col', 'x-bigquery-partition': True, 'x-bigquery-cluster': True}, ...}

# Verify partition hint is lost
assert PARTITION_HINT in test_data.columns.get("timestamp_col", {}), "PARTITION_HINT was overwritten!"
# AssertionError: PARTITION_HINT was overwritten!
```

Minimal reproducer script:

```python
from dlt.destinations.impl.bigquery.bigquery_adapter import (
    bigquery_adapter, 
    PARTITION_HINT, 
    CLUSTER_HINT
)
import dlt

@dlt.resource
def data():
    yield {"ts": "2024-01-01"}

bigquery_adapter(data, partition="ts", cluster=["ts"])

# This will fail - proving the bug
assert PARTITION_HINT in data.columns["ts"], f"Bug! Hints: {data.columns['ts']}"
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.11

### dlt data source

sql_database (PostgreSQL via PyArrow backend)

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

Suggested fix

```python
# Option 1: Raise error
if partition and cluster:
    partition_col = partition if isinstance(partition, str) else partition.column_name
    if partition_col in cluster:
        raise ValueError(
            f"Column '{partition_col}' cannot be both partition and cluster column. "
            "The partition column already provides query optimization benefits similar to clustering."
        )

# Option 2: Merge hints (update instead of replace)
for column_name in cluster:
    if column_name in column_hints:
        column_hints[column_name][CLUSTER_HINT] = True
    else:
        column_hints[column_name] = {"name": column_name, CLUSTER_HINT: True}
```

Travior:
Fixes a bug where applying both a partition hint and a cluster hint on the same column would result in the partition hint being silently dropped.

Closes #3471

### DIFF
--- a/tests/load/bigquery/test_bigquery_table_builder.py
+++ b/tests/load/bigquery/test_bigquery_table_builder.py
@@ -303,7 +303,6 @@ def test_adapter_hints_comprehensive_single_column() -> None:
             "score": 100.0,
         }
 
-    # hints should merge even across multiple adapter calls
     bigquery_adapter(partitioned_table, partition="user_id", cluster="user_id")
     assert partitioned_table.columns == {
         "user_id": {"name": "user_id", PARTITION_HINT: True, CLUSTER_HINT: True},


tests/load/bigquery/test_bigquery_table_builder.py
```diff
@@ -292,6 +292,25 @@ def partitioned_table():
     assert expected_clause in sql_partitioned
 
 
+def test_adapter_hints_comprehensive_single_column() -> None:
```
burnash, line 295: A unit test is good. But this particular scenario likely already worked before the fix because `bigquery_adapter()` below creates a new `column_hints = {}` each time. 
It's best to add also a unit test for the actual bug: a single call with both partition and cluster on the same column.

tests/load/bigquery/test_bigquery_table_builder.py
```diff
@@ -1001,6 +1020,36 @@ def sources() -> List[DltResource]:
         assert ["col1"] == hints_cluster_fields, "`hints` table IS NOT clustered by `col1`."
 
 
+@pytest.mark.parametrize(
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["bigquery"]),
+    ids=lambda x: x.name,
+)
+def test_adapter_hints_clustering_and_partitioning(
```
burnash, line 1026: Good!

dlt/destinations/impl/bigquery/bigquery_adapter.py
```diff
@@ -141,7 +141,7 @@ def bigquery_adapter(
             column.pop(PARTITION_HINT, None)  # type: ignore[typeddict-item]
 
         if isinstance(partition, str):
-            column_hints[partition] = {"name": partition, PARTITION_HINT: True}  # type: ignore[typeddict-unknown-key]
+            column_hints.setdefault(partition, {"name": partition})[PARTITION_HINT] = True  # type: ignore[typeddict-unknown-key]
```
burnash: This fix is correct. 
Here's also a chance to improve code quality: this pattern of setting column hints repeats 4 times. Let's extract this pattern into a helper function (e.g. `_set_column_hint()`) for maintainability. Most likely it could be used in other adapters, but this is a for a follow-up PR. For now I'd keep `_set_column_hint` in this module.

tests/load/bigquery/test_bigquery_table_builder.py
```diff
@@ -292,6 +292,24 @@ def partitioned_table():
...
+    def partitioned_table():
+        yield {
+            "user_id": 10000,
+            "name": "user 1",
+            "created_at": "2021-01-01T00:00:00Z",
+            "category": "category 1",
+            "score": 100.0,
+        }
+
+    # hints should merge even across multiple adapter calls
```
burnash:
```suggestion
```
Let's remove the stale comment

PR #3501, opened 2025-12-22
chore/Remove private key and replace with double b64 encoded key to avoid github ban in tests

tetelio:
Github does checks on public repos to find private keys and deactivate them. We have a private key in our OSS repo for some git tests. This PR double encodes the deploy key in base64 twice so that the content is not flagged as a key and hence banned.


tests/libs/test_git.py
```diff
@@ -25,7 +26,7 @@
 
 
 def test_ssh_key_context() -> None:
-    secret = load_secret("deploy_key")
+    secret = decode_b64b64(load_secret("deploy-key-b64-b64"))
```
tetelio, line 29: kebab or snake case are exchangeable because we normalise the string to kebab case from snake

ivasio:
@tetelio lgtm, although the tests are still failing, pls check

PR #3503, opened 2025-12-23
chore: upgrade the minimum MotherDuck duckdb version

elefeint:
### Description
Per [MotherDuck version lifecycle](https://motherduck.com/docs/troubleshooting/version-lifecycle-schedules/), DuckDB 1.2.2 and below are about to be EoL. 
This PR updates the minimum version to DuckDB 1.3.2 and limits the previously applied Windows exclusion to only the crashing versions.

### Related Issues

n/a

### Additional Context

The uv version went backwards, likely because the dependencies became simpler again.


rudolfix (approved):
LGTM!

elefeint:
Odd that the windows builds crash -- I tested on Windows 11 with python 3.13, and did not reproduce it.

rudolfix:
> Odd that the windows builds crash -- I tested on Windows 11 with python 3.13, and did not reproduce it.

1.3.2 does not contain builds for "older" windows version (apparently crashes on W2025 server). You could try to bump duckdb to 1.4.3 (only in lock file that we use in CI) that Motherduck supports. we can run it and see what happens

rudolfix:
@elefeint today we merged https://github.com/dlt-hub/dlt/pull/3527 (lancedb sql client via duckb extension) which requires the newest duckdb client. so change that you did is already on devel. btw. update was a little bit more elaborate ie. we also had to deal with new ducklake and delta lake extensions

PR #3504, opened 2025-12-24
Fix: Connectorx arrow_stream timestamp conversion issue

### LINKED ISSUE #3491: cast_date64_columns_to_timestamp uses .view() instead of .cast(), causing 1000x timestamp error with ConnectorX arrow_stream
### dlt version

1.20.0

### Describe the problem

When using ConnectorX with return_type="arrow_stream", timestamp columns are returned as `date64[ms]` instead of `timestamp[us]`. 

[Timestamp columns incorrectly converted to Date64 with arrow_stream return_type #866](https://github.com/sfu-db/connector-x/issues/866)

The `cast_date64_columns_to_timestamp` function in pyarrow.py uses .view() to reinterpret the bits, but this doesn't rescale the units, causing `milliseconds` to be interpreted as `microseconds` (1000x diff).

#3218

This results in dates like 2025-12-14 being stored as 1970-01-21.

### Expected behavior

_No response_

### Steps to reproduce

```python
source = sql_database(
    connection_string,
    table_names=["my_table"],
    backend="connectorx",
    backend_kwargs={"return_type": "arrow_stream"},  # ← triggers bug
)
source.my_table.apply_hints(incremental=dlt.sources.incremental("updated_at"))
pipeline.run(source)
# Watermark: 1970-01-21 <- (should be 2025-12-14)
```

## Root Cause

ConnectorX `arrow_stream` returns `date64[ms]`, but dlt [`cast_date64_columns_to_timestamp`](https://github.com/dlt-hub/dlt/blob/devel/dlt/common/libs/pyarrow.py) uses `.view(timestamp[us])` which reinterprets bits without unit conversion:

```python
new_col = col.view(pyarrow.timestamp("us"))  # treats ms as μs → 1000x
```

**Example:**
- Raw value: `1765747379262` (milliseconds)
- `.view(timestamp[us])` → interprets as `1765747379262 μs` → **1970-01-21** 
- Correct: `1765747379262 ms` = `1765747379262000 μs` → **2025-12-14** 

## Fix

Probably Use `.view()` to reinterpret type, then `.cast()` to convert units?

```python
# date64[ms] → view as timestamp[ms] → cast to timestamp[us]
chunk_ts_us = pyarrow.compute.cast(chunk.view(pyarrow.timestamp("ms")), pyarrow.timestamp("us"))
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

SQL

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

louiewhw:
## Summary
- rescale `date64[ms]` columns to `timestamp[us]` via `pyarrow.compute.cast`
- update date64 timestamp tests to expect ms→us scaling (including chunked arrays)

## Context
Fix #3491.
The previous implementation used `.view()` to reinterpret `date64` as `timestamp[us]`, preserving raw bits. `date64` values are defined in **milliseconds**, so this produced timestamps ~1000× too small. The new logic
rescales ms→us

## Testing
- Check unittest

### DIFF
--- a/tests/libs/pyarrow/test_pyarrow.py
+++ b/tests/libs/pyarrow/test_pyarrow.py
@@ -496,26 +496,24 @@ def test_fill_empty_source_column_values_with_placeholder() -> None:
     assert new_table.equals(expected_table)
 
 
-def test_cast_date64_columns_to_timestamp_preserves_ms_bits() -> None:
-    # Prepare timestamp[us] values with non-ms microseconds to detect precision loss
-    us_values = [0, 1001, 1609459200123123, 1609459200456789, None]
-    ts_us_arr = pa.array(us_values, type=pa.timestamp("us"))
-    # Mimic connectorx mis-typed output by reinterpreting as date64[ms]
-    date64_arr = ts_us_arr.view(pa.date64())
+def test_cast_date64_columns_to_timestamp_rescales_ms_to_us() -> None:
+    # Prepare date64[ms] values and verify ms -> us rescaling.
+    ms_values = [0, 1001, 1609459200123, 1609459200456, None]
+    date64_arr = pa.array(ms_values, type=pa.date64())
     tbl = pa.table({"ts_like": date64_arr})
 
-    # Reinterpret date64 -> timestamp[us] (naive)
+    # Rescale date64[ms] -> timestamp[us] (naive)
     out = cast_date64_columns_to_timestamp(tbl)
 
-    # Type is timestamp[us] and values are preserved exactly (no precision loss)
     assert pa.types.is_timestamp(out["ts_like"].type)
     assert out["ts_like"].type == pa.timestamp("us")
-    expected_us = ts_us_arr
-    # Table columns are ChunkedArray; compare against a chunked view of the expected array
+    expected_us = pa.array(
+        [0, 1001000, 1609459200123000, 1609459200456000, None],
+        type=pa.timestamp("us"),
+    )
     assert out["ts_like"].equals(pa.chunked_array([expected_us]))
-    # Additionally ensure ms integer 1609459200456 is present when converting back to ms
     micros = pa.compute.cast(out["ts_like"], pa.int64()).combine_chunks()
-    assert micros[3].as_py() == 1609459200456789
+    assert micros[3].as_py() == 1609459200456000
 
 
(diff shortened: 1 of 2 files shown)


tests/libs/pyarrow/test_pyarrow.py
```diff
@@ -496,26 +496,24 @@ def test_fill_empty_source_column_values_with_placeholder() -> None:
     assert new_table.equals(expected_table)
 
 
-def test_cast_date64_columns_to_timestamp_preserves_ms_bits() -> None:
-    # Prepare timestamp[us] values with non-ms microseconds to detect precision loss
-    us_values = [0, 1001, 1609459200123123, 1609459200456789, None]
-    ts_us_arr = pa.array(us_values, type=pa.timestamp("us"))
-    # Mimic connectorx mis-typed output by reinterpreting as date64[ms]
```
anuunchin, line 503: @ivasio this test explicitly assumes that connectorx with the `arrow_stream` return type produces a date64[ms] array whose underlying int values are actually microseconds - this doesn't seem to be the case. Wondering if we're missing something important here that was previously accounted for 🤔

ivasio:
I see a more deep/serious problem: it seems like in streaming mode `connectorx.read_sql` just drops the whole time part and leaves only the dates, while still showing `date64[ms]` schema type. So we can of course work on correct types being returned in the schema, but I'm not sure this behaviour is acceptable in general. Default non-streaming mode works fine

I've reproduced in on postgres with latest connectorx 0.4.4 . @louiewhw are you getting the time part on [this line](https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/sql_database/helpers.py#L268), or is it also getting cut?

Edit: never mind, it's a date64 view, underlying value is indeed in milliseconds

tests/libs/pyarrow/test_pyarrow.py
ivasio: Yes, seems we mistakenly assumed underlying values were int microseconds, but date64 actually stores milliseconds. IMO, Louie correctly proposes to use milliseconds here

anuunchin (changes requested):
@louiewhw Would be great if you fix the failing tests and the PR is good to go :)

anuunchin:
@louiewhw  I'll go ahead and fix the tests :)

anuunchin:
Continued in #3528

PR #3512, opened 2026-01-05
 use new scaffold api for vibe sources

djudjuu:
- fetch from new scaffold api (merge when exists)
- configurable via config-class

integration tests will fail until its deployed
NOTE: scaffold-api should have caching to now slow down the tests (or reduce the amount of sources that we use in the integration tests

### DIFF
--- a/dlt/_workspace/cli/_ai_command.py
+++ b/dlt/_workspace/cli/_ai_command.py
@@ -3,7 +3,6 @@ import shutil
 from pathlib import Path
 from typing import List, Tuple, get_args, Literal, Union, cast
 
-from dlt.common import logger
 from dlt.common.libs import git
 from dlt.common.pipeline import get_dlt_repos_dir
 from dlt.common.runtime import run_context
@@ -134,8 +133,6 @@ def vibe_source_setup(
     dest_dir = Path(run_context.active().run_dir)
     copied_files, count_files = _copy_repo_files(src_dir, dest_dir)
     if count_files == 0:
-        # NOTE: this should never happen, but we can't be sure
-        logger.warning(f"Source `{source}` was found but has no files to copy.")
         fmt.warning("We have nothing for %s at dltHub yet." % fmt.bold(source))
     else:
         fmt.echo(
--- a/dlt/_workspace/cli/_scaffold_api_client.py
+++ b/dlt/_workspace/cli/_scaffold_api_client.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict
+from typing import Optional
 import io
 import tempfile
 import zipfile
--- a/tests/workspace/cli/test_init_command.py
+++ b/tests/workspace/cli/test_init_command.py
@@ -58,7 +58,6 @@ from tests.workspace.cli.utils import (
     get_repo_dir,
     get_workspace_files,
     _cached_init_repo,
-    _cached_init_vibe_repo,
 )
 from tests.common.utils import modify_and_commit_file
 from tests.utils import IMPLEMENTED_DESTINATIONS, get_test_storage_root
@@ -698,7 +697,7 @@ def test_init_all_vibe_sources_together(workspace_files: FileStorage) -> None:
     # we test 5 hardcoded sources, use this to get all sources instead
     # vibe_source_candidates = [*get_source_candidates(vibe_repo_dir, source_type="vibe")]
     # Note: if we want to point test to non-production URL, we can set the environment variable
-    # os.environ["RUNTIME__SCAFFOLD__DOCS_API_URL"] = "http://localhost:8000"
+    # os.environ["RUNTIME__WORKSPACE__SCAFFOLD_DOCS_API_URL"] = "http://localhost:8000"
     random_vibez = [
         "news_api",
         "alpaca",
--- a/tests/workspace/cli/test_pipeline_command.py
+++ b/tests/workspace/cli/test_pipeline_command.py
@@ -16,7 +16,6 @@ from tests.workspace.cli.utils import (
     repo_dir,
     cloned_init_repo,
     _cached_init_repo,
-    _cached_init_vibe_repo,
 )
 
 
--- a/tests/workspace/cli/utils.py
+++ b/tests/workspace/cli/utils.py
@@ -38,6 +38,16 @@ def _cached_init_repo(tmp_path_factory) -> FileStorage:
 
 
 @pytest.fixture
+def cloned_init_repo(_cached_init_repo: FileStorage) -> FileStorage:
+    target = os.path.join(
+        get_dlt_repos_dir(),
+        f"verified_sources_repo_{uniq_id()}",
+    )
+    shutil.copytree(_cached_init_repo.storage_path, target)
+    return FileStorage(target)
+
+
+@pytest.fixture
 def repo_dir(cloned_init_repo: FileStorage) -> str:
     return get_repo_dir(cloned_init_repo, f"verified_sources_repo_{uniq_id()}")
 


dlt/_workspace/cli/_scaffold_api_client.py
```diff
@@ -0,0 +1,81 @@
...
+import requests
+from pydantic import BaseModel
+
+import os
+from dlt.common.storages.file_storage import FileStorage
+
+from dlt._workspace.cli.exceptions import ScaffoldSourceNotFound, ScaffoldApiError
+
+# TODO point to deployed scaffold-api
+SCAFFOLD_API_BASE_URL = "http://localhost:8000"
```
ivasio: I think it belongs somewhere [here](https://github.com/dlt-hub/dlt/blob/devel/dlt/_workspace/configuration.py#L29). You can always change the default URL when the API is deployed, but also we have to be careful not to release this code before it is deployed (or not merge it before then as an alternative)

dlt/_workspace/cli/_scaffold_api_client.py
```diff
@@ -0,0 +1,81 @@
...
+        raise ScaffoldApiError(f"Invalid JSON response from scaffold-api: {str(e)}", source_name)
+    except requests.RequestException as e:
+        raise ScaffoldApiError(
+            f"There was an error connecting to the scaffold-api: {str(e)}", source_name
+        )
+
+
+def get_scaffold_files_storage(source_name: str) -> Optional[FileStorage]:
+    """
+    Fetche scaffold content content for a given source and return it as a FileStorage object in a
```
ivasio: typo in "Fetche"

dlt/_workspace/cli/_scaffold_api_client.py
```diff
@@ -0,0 +1,81 @@
...
+    Returns:
+        FileStorage containing the source files, or None if source doesn't exist
+
+    Raises:
+        ScaffoldSourceNotFound: If the source doesn't exist (404)
+        ScaffoldApiError: If there's an error connecting to the API
+    """
+    scaffold_files_dict = _get_scaffold_files(source_name)
+
+    temp_dir = tempfile.mkdtemp(prefix=f"scaffold_{source_name}_")
```
ivasio, line 59: Would there be any benefit to store these scaffolds in global .dlt folder?

tests/workspace/cli/test_scaffold_api_client.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from dlt._workspace.cli._scaffold_api_client import _get_scaffold_files
+from dlt._workspace.cli.exceptions import ScaffoldSourceNotFound
+
+
+def test_get_scaffold_files_success():
+    """Test successful retrieval of scaffold files for existing source"""
+    result = _get_scaffold_files("github")
```
ivasio: CI hasn't got here because of the other failing tests, but this test will fail because it relies on local webserver

dlt/_workspace/cli/_scaffold_api_client.py
djudjuu: hmm, but the dlt init dlthub:source command is not a workspace command. but having it configurable via env is nice, ill do a separate configuraton
 runtime__scaffold__docs_api_base_url

dlt/_workspace/cli/_scaffold_api_client.py
djudjuu: i dont think so. only when we also look it up again there and or handle conflict when they exist already...when doing it from git repo, there was extra logic to see if the content hashes are still the same as in the original. I prefer to keep it simple for now..

dlt/_workspace/cli/_scaffold_api_client.py
djudjuu: [runtime.scaffold]
docs_api_url = "http://localhost:8000"

tests/workspace/cli/test_scaffold_api_client.py
djudjuu: the integration test will fail until we have the scaffold api up and running

djudjuu:
> belongs outside of _workspace/configuration

writing this reply i realized i had a wrong chain of thought. We can actually place the config there even if the users never init the workspace and just run the `dlt init dlthub:source destination` 

but placing the config under runtime.workspace (or just runtime) will work regardless, so I rewrote to place it according to your original comment

dlt/_workspace/configuration.py
```diff
@@ -30,6 +30,8 @@ class WorkspaceRuntimeConfiguration(RuntimeConfiguration):
     """Base URL for the dltHub Runtime authentication API"""
     api_base_url: Optional[str] = "https://dlthub.app/api/api"
     """Base URL for the dltHub Runtime API"""
+    scaffold_docs_api_url: Optional[str] = "https://scaffold.private.apps.dlthub.com"
```
djudjuu: should be public?

dlt/_workspace/cli/_ai_command.py
```diff
@@ -2,12 +2,15 @@
 import shutil
 from pathlib import Path
 from typing import List, Tuple, get_args, Literal, Union, cast
+from dlt.common import logger
 
+from dlt._workspace.cli.exceptions import ScaffoldApiError, ScaffoldSourceNotFound
```
rudolfix: pls group with other `_workpace` imports. also logger above

dlt/_workspace/cli/_ai_command.py
```diff
@@ -113,24 +116,26 @@ def ai_setup_command(
...
+        fmt.warning("There was an error connecting to the scaffold-api: %s" % str(e))
+        return
+    src_dir = Path(src_storage.storage_path)
 
     # where the command is ran, i.e., project root
     dest_dir = Path(run_context.active().run_dir)
     copied_files, count_files = _copy_repo_files(src_dir, dest_dir)
     if count_files == 0:
+        # NOTE: this should never happen, but we can't be sure
+        logger.warning(f"Source `{source}` was found but has no files to copy.")
```
rudolfix: actually this is a valid use case - when given scaffold is not found we were returning 0 files, how do you handle this case? are we raising 404?

dlt/_workspace/cli/_scaffold_api_client.py
```diff
@@ -0,0 +1,66 @@
...
+    headers = {"Accept": "application/zip"}
+    try:
+        response = requests.get(url, headers=headers, timeout=SCAFFOLD_API_TIMEOUT)
+    except requests.RequestException as e:
+        raise ScaffoldApiError(
+            f"There was an error connecting to the scaffold-api: {str(e)}", source_name
+        )
+
+    # Handle non-200 responses
+    if response.status_code == 404:
```
rudolfix, line 50: we were not raising but just returning 0 files and we were continuing init process. we were generating a simple pipeline template the usual way and you also get ie. cursor rules. my take is to keep it that way

tests/workspace/cli/test_init_command.py
```diff
@@ -650,48 +648,54 @@ def test_incompatible_dlt_version_warning(repo_dir: str, workspace_files: FileSt
...
-        "jobnimbus",
-        "piwik_pro",
-        "perplexity_ai",
-        "maileon",
-        "wrike_project_management",
-        "rocketreach",
-        "wordpress_site",
-        "deepinfra",
-        "no_crm_io",
+        # "fulcrum_data_management",
```
rudolfix: are those gone?

tests/workspace/cli/test_init_command.py
```diff
@@ -650,48 +648,54 @@ def test_incompatible_dlt_version_warning(repo_dir: str, workspace_files: FileSt
...
+    assert (
+        "file(s) supporting github were copied:" in _out
+        and "github.md" in _out
+        and "github-docs.yaml" in _out
+    )
 
 
-def test_init_all_vibe_sources_together(vibe_repo_dir: str, workspace_files: FileStorage) -> None:
-    # we test 20 hardcoded sources, use this to get all sources instead
+def test_init_all_vibe_sources_together(workspace_files: FileStorage) -> None:
```
rudolfix, line 696: we do not have test when scaffold is not found... that's why the old behavior was not enforced. let's add such test

tests/workspace/cli/test_init_command.py
djudjuu: no, but since the repo is not cloned as a fixture anymore but we hit the actual api we can not speed it up and using more sources adds little benefit, but slows tests down, so I figured maybe 5 instead of 20 is enough

dlt/_workspace/cli/_ai_command.py
djudjuu: done

dlt/_workspace/cli/_ai_command.py
djudjuu: so if the source doesnt exist -> 404
if there are files, but they are empty: 500 (since this is invalid state)
so right now, arriving at this line should be impossible. But to futureproof it, if somehow something gets messed up, I left it in.

in both error cases warning is logged, and the regular template file and cursor rules are created.

dlt/_workspace/cli/_scaffold_api_client.py
djudjuu: yes, I am handling the errors in the cli flow (_ai_command.py)

tests/workspace/cli/test_init_command.py
djudjuu: lets make it a few more then

tests/workspace/cli/test_init_command.py
djudjuu: added. tests checks nonexisting vibe source emits warning and creates default template

djudjuu:
actually, lets wait with merging this
I want to run the tests again against the 0.1.4. scaffold-api backend
(will trigger re-run once its deployed

UPDATE: the api is deployed, comiit 14b... is running against it

dlt/_workspace/cli/_ai_command.py
```diff
@@ -3,11 +3,14 @@
 from pathlib import Path
 from typing import List, Tuple, get_args, Literal, Union, cast
 
+from dlt.common import logger
```
rudolfix: please do not use logger in `cli` modules. we issue warnings via `fmt`

rudolfix (approved):
fixed that line.... LGTM!

Issue #3447, opened 2025-12-08
Support DECFLOAT type for the Snowflake destination


VioletM:
### Feature description

Snowflake recently introduced a new data type, DECFLOAT, which stores numbers with variable precision. This is analogous to Oracle's NUMBER data type when precision and scale is not defined.

https://docs.snowflake.com/en/release-notes/2025/9_37#new-decfloat-data-type
https://docs.snowflake.com/en/sql-reference/data-types-numeric#label-data-type-decfloat

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

In dlt, the default is to cast these values to `FLOAT`. You can cast them to `NUMBER` using `type_adapter_callbacks`, but that would require determining the effective precision and scale for each column individually. As a result, there is no single precision/scale pair that fits all columns, and computing per-column stats over all rows is not feasible.

So we effectively have two realistic options:

* `FLOAT` – arithmetic works out of the box, but there is a risk of precision loss 
* `STRING` – no precision loss, but you can’t do arithmetic in the destination without an extra conversion step.

A third alternative is `DECFLOAT`, which may offer precise representation while still supporting arithmetic in the destination.

rudolfix:
@VioletM DECFLOATs are FLOATs with decimal arithmetic and should not be used for any calculations involving money amounts and similar where precision loss is unacceptable. If the intention is just to store values and the input data is within DECFLOAT precisions - this should be OK.
btw. decimal arithmetic in Oracle is very different from DECFLOAT arithmetic's. I do not have enough knowledge but if the scope here is going beyond just storing values (even SUM() may produce rounding errors on DECFLOAT) they caution is needed and maybe additional step in dlt pipeline is needed that will validate incoming values ie. they are withing a range that will not produce precision loss.

there are two options to carry those values
str
Decimal (can store arbitrary numbers)
double should never be used!

the latter one seems more natural to me. We just need to make sure that Oracle returns right values - we have several other tickets to do that.

ivasio:
Agree with @rudolfix the latter is more preferable! I'll add support for `DECFLOAT` on Snowflake destination and ensure proper support for unbounded decimals in Oracle. In the meanwhile, a workaround that enables to run it already. Tested on default `reflection_level="full"`, please lmk if other ones need to be tested as well. Also, I'm using `oracledb` and `sqlalchemy>2`, should work for `cx_Oracle` if they are using this driver, but I can test separately. `...` below denote other arguments

```py
[...]

PR #3513, opened 2026-01-05
Feat/3447 - add DECFLOAT support for Snowflake

ivasio:
Implements #3447 . Properly works only on sqlalchemy backend (pyarrow doesn't support floating-point decimals and pandas has even bigger decimal problems)

To be done: 
- implementation: currently, this PR modifies the default behaviour when dlt type is decimal with both precision and scale being None. This is backwards incompatible, so instead of making this new logic the default, it should only work for some opt-in flag/config/...
- tests: add a test for oracle source and snowflake destination to ensure that a column with otherwise unsupported values will work. Oracle type should be `NUMBER()`, we should ensure that a `DECFLOAT` column on Snowflake gets created and that 2 values (one with 30 signs before the comma and another with 30 signs after) are transferred exactly, without loss of data

### DIFF
--- a/dlt/destinations/impl/snowflake/configuration.py
+++ b/dlt/destinations/impl/snowflake/configuration.py
@@ -173,8 +173,8 @@ class SnowflakeClientConfiguration(DestinationClientDwhWithStagingConfiguration)
     use_vectorized_scanner: bool = False
     """Whether to use or not use the vectorized scanner in COPY INTO"""
 
-    unbound_decimal_to_decfloat: bool = False
-    """Whether to convert unbound DECIMALs to DECFLOAT type"""
+    use_decfloat: bool = False
+    """Whether to use DECFLOAT type for unbound decimals instead of DECIMAL"""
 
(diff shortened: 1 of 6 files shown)


Travior:
The basic functionality works and is configurable, proper tests are missing, but I can contribute those depending on how/who continues with this ticket.

What works:
- Numbers that are stored as python `Decimals` get correctly loaded as `DECFLOAT` if configured

Caveats:
- non text based staging format (like `parquet`) don't work. I believe the only supported ones are `jsonl` and `csv`
- When interacting with destination datasets (or via `pipeline.sql_client()`), pythons default precision must be increased to avoid potential data loss (pythons: `decimal.getcontext().prec` is 28, snowflakes `DECFLOAT` supports up to 36)
- The arrow path (e.g. `Relation.df()`) doesn't work with `DECFLOAT`. When using the `DB-API`  paths (like  `cursor.fetchall()` the snowflake connector converts back to `Decimal` with all the already mentioned precision caveats)

PR #3521, opened 2026-01-07
Fix/3376 state per schema

### LINKED ISSUE #3376: append pipeline state to the relevant load package when extracting
**Background**
Pipeline state is extracted and loaded together with the data. The idea is the load packages are self contained and loading them fully keeps the dataset and pipeline in consistent state. Packages should not depend on other packages to be consistent.
Currently extract method will store the state with the package that contains data for the default schema and this state is extracted only once at the end after extraction of data ended. For multi schema pipelines state is not attached to packages for the other schemas. This needs to be fixed.

**Tasks**

1. * [ ] pipeline extract source by source using `for source in data_to_sources(` we must order sources returned `data_to_sources` by schema so if there are many instances of the same source they will be extracted one after another.
2. * [ ] call `_bump_version_and_extract_state` in a loop when you detect that schema name changed. this makes sure we write state only once per package. also pass the schema from the source to direct extracted state to the right package.
3. * [ ] if `pipeline.config.use_single_dataset` is False we must keep old behavior (state extracted only once to the default schema

**Tests**
There are very detailed tests in `test_pipeline_state` and test for many schemas may fail. Those failing tests must be carefully analyzed and fixed. On top of that:

1. we must make sure that `data_to_sources` orders returned sources by schema name
2. we must make sure that if we extract multiple sources with the same schema, state is extracted only once
3. if we extract multiple source with different names (schemas) each package contains state (if source extraction modified it!). this also means that both schemas will contain `_dlt_pipeline_state` which actually makes sense
4. if sources do not modify state (ie. no inremental loading) only the first package contain state (the initial version)

and probably more

anuunchin:
This PR adjusts the `pipeline.py` so that if there are multiple schemas, the state is extracted to the corresponding packages if the configuration allows. Relevant tests are added/ modified.

Resolves #3376

### DIFF
--- a/dlt/pipeline/pipeline.py
+++ b/dlt/pipeline/pipeline.py
@@ -452,7 +452,9 @@ class Pipeline(SupportsPipeline):
         )
         try:
             with self._maybe_destination_capabilities():
-                # unique schemas as we extract all sources
+                # extract all sources ordered by schema name
+                # while tracking schema transitions to extract
+                # state to each schema's package in multi-dataset mode
                 last_schema = None
                 for source in data_to_sources(
                     data,


dlt/pipeline/pipeline.py
```diff
@@ -477,11 +489,13 @@ def extract(
                         refresh=refresh or self.refresh,
                     )
 
-                # this will update state version hash so it will not be extracted again by with_state_sync
```
anuunchin, line 480: I find this comment unnecessary to keep

tests/pipeline/test_pipeline.py
```diff
@@ -3198,9 +3512,18 @@ def get_subpages(page_item):
...
     )
-    assert_load_info(info, 2)
-    # ten subpages because only 1 page is extracted in the second source (see gen_pages exit condition)
-    assert load_table_counts(pipeline) == {"subpages": 10, "pages": 10}
+    if use_single_dataset:
+        assert_load_info(info, 2)
+        # ten subpages because only 1 page is extracted in the second source (see gen_pages exit condition)
+        assert load_table_counts(pipeline) == {"subpages": 10, "pages": 10}
+    else:
+        assert_load_info(info, 1)
```
anuunchin, line 3520: pipeline.dataset("test") which is used by `load_table_counts` doesn't seem to be working - this should be a different ticket, so went with directly using the sql client

rudolfix (approved):
LGTM! all the test cases I could think of are there... It seems that we stopped to generate additional load package just to load the state if source schema was different from default pipeline schema which is great.

(Opus also agrees)

PR #3523, opened 2026-01-08
Fix/Fixing extraction of csv to dlt on rest_api

lfpll:
### Description
Currently when extracting `csv` apis the api fails, because it uses the method `response.json()` to all outputs, this adds support for extracting data from CSV responses in the REST API client.

**Changes:**
- Added `response_format` parameter to `RESTClient` and `Endpoint` configuration since some http responses can end with `/text` but be csv
- Implemented automatic CSV detection via `Content-Type` header (ending in `/csv`)
- Added `_extract_response_csv()` method that parses CSV content into a list of dictionaries using Python's `csv.DictReader`
- CSV responses automatically use `SinglePagePaginator` since CSV endpoints typically don't support pagination
- Users can also explicitly set `response_format: "csv"` in endpoint configuration

**Usage example:**
```python
config = {
    "client": {"base_url": "https://api.example.com"},
    "resources": [
        {
            "name": "data",
            "endpoint": {
                "path": "/export.csv",
                "response_format": "csv",  # Optional if Content-Type header is set
            },
        }
    ],
}
```

### Additional Context

This feature enables the REST API source to consume APIs that return CSV data instead of JSON, which is common for data export endpoints on reporting APIs and currently fails because it tries to extract using `response.json()` method

### DIFF
--- a/tests/sources/rest_api/conftest.py
+++ b/tests/sources/rest_api/conftest.py
@@ -410,6 +410,36 @@ def mock_api_server():
         def posts_with_reactions(request, context):
             return paginate_by_page_number(request, generate_posts_with_reactions())
 
+        @router.get(r"/posts_csv$")
+        def posts_csv(request, context):
+            """Return posts as CSV with proper Content-Type header"""
+            posts = generate_posts()[:5]
+            csv_content = "id,title\n"
+            for post in posts:
+                csv_content += f"{post['id']},{post['title']}\n"
+            context.headers["Content-Type"] = "text/csv"
+            return csv_content
+
+        @router.get(r"/posts_csv_no_content_type$")
+        def posts_csv_no_content_type(request, context):
+            """Return posts as CSV without Content-Type header (for explicit format testing)"""
+            posts = generate_posts()[:5]
+            csv_content = "id,title\n"
+            for post in posts:
+                csv_content += f"{post['id']},{post['title']}\n"
+            return csv_content
+
+        @router.get(r"/posts/(\d+)/comments_csv$")
+        def post_comments_csv(request, context):
+            """Return comments for a post as CSV (for mixed JSON parent + CSV child testing)"""
+            post_id = int(request.url.split("/")[-2])
+            comments = generate_comments(post_id, count=3)
+            csv_content = "id,post_id,body\n"
+            for comment in comments:
+                csv_content += f"{comment['id']},{comment['post_id']},{comment['body']}\n"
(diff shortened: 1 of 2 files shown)


dlt/sources/helpers/rest_client/client.py
```diff
@@ -353,6 +369,10 @@ def detect_paginator(self, response: Response, data: Any) -> BasePaginator:
         Returns:
             BasePaginator: The paginator instance that was detected.
         """
+
+        if self.response_format == 'csv':
+            return SinglePagePaginator()
```
lfpll, line 374: I think this is way easier and cleaner them having to tweak the paginator itself.

anuunchin:
Thanks for the effort here! For future changes like this, I’d strongly recommend opening an issue first and discussing the approach with the maintainers before starting a PR. 

Since the requested functionality is already supported via [response actions](https://dlthub.com/docs/dlt-ecosystem/verified-sources/rest_api/advanced#response-actions)  (see example below), I'll go ahead and close the PR

```python
import dlt
import csv
import io
import json
from requests import Response
from typing import Any


def csv_to_json_hook(response: Response, *args: Any, **kwargs: Any) -> Response:
    csv_content = response.text
    reader = csv.DictReader(io.StringIO(csv_content))
    data = list(reader)
    json_content = json.dumps(data).encode()
    response._content = json_content
    return response


def main():
    from dlt.sources.rest_api import rest_api_source
    from dlt.sources.rest_api.typing import RESTAPIConfig
    from dlt.sources.helpers.rest_client.paginators import SinglePagePaginator
    
    config: RESTAPIConfig = {
        "client": {
            "base_url": "https://gist.githubusercontent.com",
        },
        "resources": [
            {
                "name": "pokemon",
                "endpoint": {
                    "path": "/armgilles/194bcff35001e7eb53a2a8b441e8b2c6/raw/92200bc0a673d5ce2110aaad4544ed6c4010f687/pokemon.csv",
                    "paginator": SinglePagePaginator(),
                    "response_actions": [csv_to_json_hook], 
                },
            }
        ],
    }
    
    pipeline = dlt.pipeline(
        pipeline_name="csv_test",
        destination="duckdb",
        dataset_name="pokemon_csv",
        dev_mode=True,
    )
    
    load_info = pipeline.run(rest_api_source(config))
    print(load_info)

    row_counts = pipeline.dataset().row_counts().df()
    print(row_counts)


if __name__ == "__main__":
    main()
```

PR #3527, opened 2026-01-09
feat(lancedb): implement `SqlClientBase`

zilto:
# Summary
Add SQL capabilities (`dlt.Dataset`,  `dlt.Relation`, `@dlt.hub.transformation`, data quality) to LanceDB destination via the official [lance-duckdb](https://github.com/lance-format/lance-duckdb/tree/main)

## Implementation
The implementation reuses code from the SQL clients based on files (filesystem, iceberg) and clients that use DuckDB extensions (ducklake, motherduck) 
- Make `LanceDBClient` inherit from `WithSqlClient` and add required methods. This is required to enable compatibility with other parts of `dlt`
- Add a `LanceDBSQLClient` that inherits from DuckDB's client. It automatically enables the `lance` extension and creates views to enable consistent querying
    ```sql
    // for example, user can write
    SELECT * FROM foo
    // instead of 
    SELECT * FROM "/path/to/instance/foo.lance"
    ```
    
## Known TODOs
- [x] ensure the file path builder handles dataset names and schema's normalization naming convention
- [ ] add this client and `lancedb` to the broader data access test suite (we should test for local duckdb and maybe duckdb on S3-compatible?)

## Decisions to make
This is a simple working prototype that cuts a lot of fat. Despite having similarities with other clients, I worry that too much coupling hurts maintainability. 

Things to consider building:
- add `.create_secret()` method to store object store credentials (e.g., AWS) in DuckDB to support `SELECT * FROM "s3://..."`
- add a `.storage` attribute for the filesystem / object store that allows us to crawl through tables (similar to `WithTableScanner` or `ducklake`)
- how much coupling do we need with `WithTableScanner`?
- Add `.lance` as a `table_format` for our typing

### DIFF
--- a/tests/load/pipeline/test_duckdb.py
+++ b/tests/load/pipeline/test_duckdb.py
@@ -158,7 +158,7 @@ def test_duck_precision_types(destination_config: DestinationTestConfiguration) 
     )
 
     with pipeline.sql_client() as client:
-        table = client.native_connection.sql("SELECT * FROM row").arrow()
+        table = client.native_connection.sql("SELECT * FROM row").fetch_arrow_table()
 
     # only us has TZ aware timestamp in duckdb, also we have UTC here
     _verify_schema(table.schema, pa.decimal128(38, 0))


dlt/destinations/impl/lancedb/sql_client.py
```diff
@@ -0,0 +1,113 @@
...
+            query = query.replace("%s", "?")
+
+        expression = sqlglot.parse_one(query, read="duckdb")
+        load_tables: dict[str, str] = {}
+        for table in expression.find_all(exp.Table):
+            if not table.this:
+                continue
+
+            schema = table.db
+            if schema or schema.lower() != self.lancedb_client.dataset_name:
```
rudolfix: this skips sentinel tables, right?

rudolfix (commented):
LGTM!
- deriving from `DuckDbSqlClient` (and not from `WithTableScanners`) is OK with this case. I moved all the code that deals with SECRET creation to `DuckDbSqlClient` so we can easily support `.lance` tables on bucket
- namespaces work only on lancedb cloud. so existing mechanism to prefix tables must stay - at least for local databases.
- I'm curious if our dlt ibis backend and ibis query work with it :)
- does it make sense to enable lance in `create_ibis_backend`? we'd need to create all views upfront like we do with filesystem though

overall this is great!

zilto:
> namespaces work only on lancedb cloud. so existing mechanism to prefix tables must stay - at least for local databases.

Namespaces work locally if you install `lance-namespaces`. We should definitely adopt it. Should be behind a feature flag to ensure backwards compatibility by default

> deriving from DuckDbSqlClient (and not from WithTableScanners) is OK with this case. I moved all the code that deals with SECRET creation to DuckDbSqlClient so we can easily support .lance tables on bucket

Technically, we could get table URI reliably 100% of the time by using `lancedb_con.open_table("foo").to_lance().uri`. That would require talking to the destination. It's probably much more reliable to implement `dlt (1) to lance (1)` instead of `dlt (1) to object store supported by lance (n)`

> I'm curious if our dlt ibis backend and ibis query work with it :)
does it make sense to enable lance in create_ibis_backend? we'd need to create all views upfront like we do with filesystem though

Currently, the `dlt.Dataset`,  `dlt.Relation`, `@dlt.hub.transformation`, etc. work out of the box. It uses the `_DltBackend` which doesn't go through `create_ibis_backend`. Though, adding `create_ibis_backend()` would enable seeing data in `marimo`. (we could also plug the `_DltBackend` in marimo directly actually)

zilto:
@rudolfix I need your help:
1. figure out Python dependencies (update `pyproject.toml` and `uv.lock`). The DuckDB extension is only available for `duckdb>=1.4.3`, which is conflicting with the `pipeline` dependency group (`<1.4.0`).
2. it's reasonable to have `duckdb` a mandatory dependency inside `dlt[lancedb]`, but we could separate it further 

we could also remove the `tantivy` dependency inside `lance` group



I think the PR is in a good state for a first release. Having users should help us find bugs more efficiently. We'll be able to decide between "fixing the current implementation" vs. "upgrading to using namespaces"

rudolfix (approved):
LGTM!

1. I see lockfile with the 1.4.3 duckdb. that's good. we should upgrade anyway (see #3503 ) if there are any broken tests related to iceberg or azure I can investigate - those duckdb extensions were causing problems before
2. I verified  that lance tests are passing without duckdb installed. this is good
3. OK to add new dependencies to `lancedb` group

zilto:
> I see lockfile with the 1.4.3 duckdb. that's good. we should upgrade anyway

Ok, so there were a few more changes required to accomodate `duckdb>=1.4.3`. In particular, there was a breaking change in `.arrow()` return value. This was particularly annoying for linting because the main dependency group used for linting is on `duckdb==1.3.2` and **some** of the tests (where `lancedb` extra is installed) are on `duckdb==1.4.3.`

zilto:
@rudolfix update on CI results:
- docs linting is failing because of zendesk
- 2 failures: ducklake hard delete is failing with `duckdb==1.4.3`
- 36 failures: delta error, don't know if related to duckdb
- 91 errors: iceberg

rudolfix:
> @rudolfix update on CI results:
> 
> * docs linting is failing because of zendesk
> * 2 failures: ducklake hard delete is failing with `duckdb==1.4.3`
> * 36 failures: delta error, don't know if related to duckdb
> * 91 errors: iceberg

ducklake problem is clear (MERGE statement got enabled and it does not work in a few tests). I'll investigate others and try to fix them

Issue #2673, opened 2025-05-23
[SNOWFLAKE] Avoid table downtime in staging-optimized replace strategy


kang8:
### Feature description

Implement Snowflake's `ALTER TABLE ... SWAP` syntax for the `staging-optimized` replace strategy to eliminate table downtime during data replacement.

Currently, when using `write_disposition="replace"` with `replace_strategy="staging-optimized"` on Snowflake destination, the destination table is dropped and recreated from the staging table. This causes a brief period where the table doesn't exist, leading to "table does not exist" errors for queries running during the replacement process. Since I run dlt scripts hourly, this issue occurs frequently and disrupts downstream processes.

Snowflake provides an [ALTER TABLE ... SWAP WITH](https://docs.snowflake.com/en/sql-reference/sql/alter-table#swapping-tables) syntax that can atomically swap two tables by renaming them. This would allow the staging table to instantly become the destination table without any downtime.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

I have a dlt pipeline that runs hourly with the following resource configuration:

```py
@dlt.resource(
    primary_key="id",
    write_disposition="replace",
)
def users():
   ...
```

My destination is Snowflake, and I'm using the `staging-optimized` replace strategy for performance reasons. However, the current implementation drops the destination table before recreating it from staging, which causes:

1. Brief periods where the table doesn't exist, resulting in query failures
2. Disruption to downstream processes that depend on continuous table availability

The alternative strategies have their own issues:

* `truncate-and-insert`: Causes data to be temporarily absent (even worse than missing table)
* `insert-from-staging`: Solves the availability issue but significantly increases load time

### Proposed solution

Modify the `staging-optimized` strategy for Snowflake to use `ALTER TABLE ... SWAP WITH` instead of dropping and recreating tables. The implementation would:

1. Load data into staging table (current behavior) 
2. Use ALTER TABLE destination_table SWAP WITH staging_table to atomically swap the tables
3. (Optional)Drop the old destination table (now named as staging)

This approach would:

* Maintain the performance benefits of staging-optimized
* Ensure zero downtime for the destination table
* Provide atomic table replacement

Note: The SWAP syntax has some restrictions (e.g., cannot swap permanent/transient tables with temporary tables, requires OWNERSHIP privileges), but these shouldn't be issues for typical dlt usage patterns.

### Related issues

None.

Travior:
There is a possible breaking change with how permissions and future grants work with this.
[...]

PR #3540, opened 2026-01-15
feat/2673 snowflake swap replace

Travior:
Closes: #2673


rudolfix (approved):
LGTM!
I added a more elaborate test (nested tables, multiple loads), the code itself was OK.
thanks!

Issue #3402, opened 2025-11-28
feat: internal `filter_loads(dataset, table_name, load_ids)`


zilto:
# Goal
When interacting with a `dlt.Dataset`, it's useful to select data associated with specific specific `_dlt_load_id` value(s). This includes: incremental transformations, unit tests, metrics, data quality checks.

We should implement an internal function for this specific purpose.

# Implementation
The challenge is that only **root tables** have a `_dlt_load_id`. But all tables can be joined back to a root table.

The main function should look like:
```python
def filter_loads(
  dataset: dlt.Dataset,
  table_name: str,
  load_ids: str | Collection[str],
) -> dlt.Relation: ...
```

This needs to handle 3 main distinct cases, which should have their own function:
1. `table_name` is a root table:  It includes a `_dlt_load_id` column. Filter directly
2. `table_name` is a child table with a `root_key`: It can be joined to its root table directly. Filter the root table and inner join with child table
3. `table_name` is a child table without a `root_key` but has a `parent_key`: Need to traverse child-parent relationships up to the root table. Filter the root table and apply inner joins in a loop

Assertions / tests:
- before the main logic, we should check that provided `load_ids` exist in `_dlt_loads.load_id`
- raise if `table_name` doesn't exist on `dataset.tables`. Use `ValueErrorWithKnownValues` and return the list of table names
- properly handle a list of load ids or a single string input; remember a `str` is a `Sequence`
- In case 3), it's possible that the chain of joins "breaks" because there's no matching `_dlt_load_id`. You can exist early and return an empty table

Others:
- the final table should have the same columns as `dataset.table()`, but also include a `_dlt_load_id` column
- use `sqlglot` for writing the queries
- the bulk of the code can be added to `dlt/dataset/relation.py` or `dlt/destinations/queries.py` for lower level sqlglot code

# Relevant code / references
- A good amount of the necessary work exists in #2960;  But contrary to this PR, we don't want to return joined tables.
- `dlt.Schema.references` returns the implicit references (parent-child, root-child)
- `dlt/common/schema/utils.py` has several functions to create references, identify root/parent/child key column, identify if table is root, retrieve children of table, etc. 

# Follow-up
If we are confident, we could also introduce a new public API that uses this internal method

```python
class Relation:
  def filter_loads(self, load_ids: str | list[str]) -> dlt.Relation:
    return filter_loads(
     dataset=self._dataset,
     table_name=self._table_name,  # this only exists for tables created via `.table()`
[...]

PR #3547, opened 2026-01-16
feat: filter `dlt.Relation` by `load_id`

zilto:
- implements #3402 (see for details)

## To review
- does the user facing APIs make sense?

## Comments
- could have more testing
- need to update docs, but let's keep it experimental while we try it out

### DIFF
--- a/dlt/dataset/relation.py
+++ b/dlt/dataset/relation.py
@@ -430,7 +430,6 @@ class Relation(WithSqlClient):
 
         assert isinstance(column_or_expr, str)
         column_name = column_or_expr
-        normalized_column_name = self._dataset.schema.naming.normalize_identifier(column_name)
 
         if isinstance(operator, str):
             try:
@@ -443,15 +442,15 @@ class Relation(WithSqlClient):
                 )
 
         sqlgot_type = to_sqlglot_type(
-            dlt_type=self.columns_schema[normalized_column_name].get("data_type"),
-            precision=self.columns_schema[normalized_column_name].get("precision"),
-            timezone=self.columns_schema[normalized_column_name].get("timezone"),
-            nullable=self.columns_schema[normalized_column_name].get("nullable"),
+            dlt_type=self.columns_schema[column_name].get("data_type"),
+            precision=self.columns_schema[column_name].get("precision"),
+            timezone=self.columns_schema[column_name].get("timezone"),
+            nullable=self.columns_schema[column_name].get("nullable"),
         )
 
         value_expr = build_typed_literal(value, sqlgot_type)
 
-        column = sge.Column(this=sge.to_identifier(normalized_column_name, quoted=True))
+        column = sge.Column(this=sge.to_identifier(column_name, quoted=True))
 
         condition: sge.Expression = None
         if operator == "in":


dlt/dataset/dataset.py
```diff
@@ -438,3 +460,33 @@ def _get_dataset_schema_from_destination_using_dataset_name(
...
+
+
+def _get_load_ids(dataset: dlt.Dataset) -> list[str]:
+    """Get a list of load ids associated with the dataset."""
+    loads_table = dataset.loads_table()
+    query = (
+        loads_table
+        # need to filter out rare case where a dataset includes multiple `dlt.Schema`
+        # this mechanism is currently used by data quality metrics and checks
+        .where("schema_name", "eq", dataset.schema.name)
```
rudolfix: we should not use raw identifiers - all of them should be normalized via dataset.schema.naming. this applies here and to `C_DLT_LOADS_TABLE_LOAD_ID` and other column names.

the `_dlt_loads` table you get correctly from the Schema instance - so it is normalized.

this may sound like nitpicking but for example `s3 tables` on `athena` does not allow for `_` to start any names and we use naming convention to generate compatible names. some vector destinations (weaviate) capitalize identifiers etc.

dlt/dataset/relation.py
```diff
@@ -306,13 +316,16 @@ def head(self, limit: int = 5) -> Self:
...
 
-    def select(self, *columns: str) -> Self:
+    def select(self, *columns: str, _merge_subqueries: bool = True) -> Self:
         """CReate a `Relation` with the selected columns using a `SELECT` clause."""
         proj = [sge.Column(this=sge.to_identifier(col, quoted=True)) for col in columns]
         subquery = self.sqlglot_expression.subquery()
         new_expr = sge.select(*proj).from_(subquery)
         rel = self.__copy__()
-        rel._sqlglot_expression = merge_subqueries(new_expr)
+        if _merge_subqueries:
```
rudolfix: mhmm why we optimize queries here at all? it should rather happen when generating final query ie. optionally in `to_sql` and do that just before execution (ie. in `load` stage)

dlt/dataset/relation.py
```diff
@@ -486,6 +499,55 @@ def filter(  # noqa: A003
...
+        This should only raise if the `dlt.Schema` was tempered, breaking the
+        dlt-generated root and parent relationships.
+        """
+        table_schema = self._dataset.schema.tables[self._table_name]
+
+        if C_DLT_LOAD_ID in self.columns:
+            return self
+        elif schema_utils.has_column_with_prop(table_schema, "root_key"):
+            return _add_load_id_via_root_key(self)
+        elif schema_utils.has_column_with_prop(table_schema, "parent_key"):
```
rudolfix, line 521: cool to have this fallback!

dlt/dataset/dataset.py
```diff
@@ -206,11 +207,10 @@ def __call__(
         """Convenience method to proxy `Dataset.query()`. See this method for details."""
         return self.query(query, query_dialect, _execute_raw_query=_execute_raw_query)
 
-    def table(self, table_name: str, **kwargs: Any) -> dlt.Relation:
+    def table(
+        self, table_name: str, *, load_ids: Optional[Collection[str]] = None, **kwargs: Any
```
rudolfix, line 211: yeah I think this is pretty useful to end user that wants to analyze particular loads. I already shared a bigger perspective where we also accept `Incremental` instance here which gets translated into subquery.

dlt/dataset/dataset.py
```diff
@@ -438,3 +460,33 @@ def _get_dataset_schema_from_destination_using_dataset_name(
...
+
+def _get_latest_load_id(dataset: dlt.Dataset) -> Optional[str]:
+    """Get the latest load id associated with the dataset."""
+    loads_table = dataset.loads_table()
+    query = (
+        loads_table
+        # need to filter out rare case where a dataset includes multiple `dlt.Schema`
+        # this mechanism is currently used by data quality metrics and checks
+        .where("schema_name", "eq", dataset.schema.name)
+        .select(C_DLT_LOADS_TABLE_LOAD_ID)
```
rudolfix: this is OK for all practical purposes, note that table has `inserted_at` column that you can use to get the most recently processed load_id. a matter of definition what `latest` means

tests/dataset/test_relation.py
```diff
@@ -1,13 +1,137 @@
...
+}
+
+
+TLoadsFixture = tuple[dlt.Dataset, tuple[str, str], tuple[dict[str, Any], dict[str, Any]]]
+
+
+@pytest.fixture(scope="module")
+def loads_with_root_key() -> TLoadsFixture:
+    """Create a pipeline with nested data across multiple loads."""
+    pipeline = dlt.pipeline("test_from_loads", destination="duckdb", full_refresh=True)
```
rudolfix: use `dev_mode`

tests/dataset/test_relation.py
```diff
@@ -72,3 +196,115 @@ def test_transformed_relation_to_ibis_(purchases: dlt.Relation) -> None:
...
+@pytest.mark.parametrize("add_load_id_column", [True, False])
+@pytest.mark.parametrize(
+    "dataset_with_loads",
+    [
+        pytest.param("with_root_key", id="root_key-True"),
+        pytest.param("without_root_key", id="root_key-False"),
+    ],
+    indirect=True,
+)
+def test_relation_from_loads(
```
rudolfix, line 297: would make sense to also parametrize by naming convention like in reference tests:
```
@pytest.mark.parametrize(
    "name_normalizer_ref",
    (
        "tests.common.cases.normalizers.title_case",
        "tests.common.cases.normalizers.sql_upper",
        "tests.common.cases.normalizers.snake_no_x",
    ),
)
def test_references_parameterized_by_naming(schema: dlt.Schema, name_normalizer_ref: str) -> None:
    schema._normalizers_config["names"] = name_normalizer_ref
    schema.update_normalizers()
```

rudolfix (changes requested):
this looks pretty good! please look at my comments on naming conventions

dlt/dataset/dataset.py
zilto: We had this discussion before. Though, I realized that all of this is normalized by `dlt.Relation.to_sql()`, which is the only exit point (used by `.df()`, `.arrow()`, etc.)

In other words, the `dlt.Relation` should be destination-agnostic. It's execution is destination-specific (e.g., normalization, transpilation, SQL client optimization)

This query normalization responsibility should be delegated to a single place (e.g., no user-query rewrite until execution). Otherwise it's very hard to debug for users and during development

dlt/dataset/relation.py
zilto: The `merge_subqueries()` call was already in the code. Removing it broke some tests. I decided to use an `if/else` that made old and new functionality work

I agree that `merge_subqueries()` shouldn't be here. Though, I would need to investigate further to make that change (for a follow-up PR)

dlt/dataset/dataset.py
zilto: I wondered if there was a practical difference between `inserted_at` and `load_id`. My guess:
- `load_id` the value is set by `dlt` at the beginning of the load phase
- `inserted_at` the value is set by the destination when the data is written on the destination

tests/dataset/test_relation.py
zilto: I can parameterize tests evaluating the query, but I can't alter the normalizer config set on the duckdb destination / dataset AFAIK (it wouldn't match stored data)

A few notes:
- If we have the right architecture for transpiling queries, we shouldn't need to parameterize all `dlt.Relation` tests. We should only parameterize the transpiling code ([related](https://github.com/dlt-hub/dlt/pull/3547#discussion_r2708356653))
- This 3 value parameterization adds 72 cases

zilto: Locally, I changed the `load_with_root_key` and `loads_without_root_key` fixture to run against the Clickhouse container and everything succeeded

dlt/dataset/dataset.py
rudolfix: User facing interface accepts identifiers as is - they must match the schema of the dataset which obviously has identifiers that are already normalized.

The problem that you describe is in our internal code where we need to deal with any naming convention and you are trying to use a constant as an identifier without normalization. I get that it is hard to debug. The way to solve would be to create an opaque `NormalizedIdentifier = NewType(str)` to enforce normalized identifier  on certain methods (but not on public interface where user does not care). I can look into that. I already have opaque type for `PreparedTableSchema` so no one using "raw" table schemas during loading.

zilto: Ok, I understand better looking back at the code. My understanding for documentation purpose:

Write flow
1. extract source data (`products`)
2. normalize data table and column names (`products -> Products`)
3. write data to storage and finalize dlt schema (`Products`)

Read flow
1. dataset is instantiated from dlt schema (`Products`)
2. user writes query (`SELECT * FROM Products`)
3. query is validated against schema and destination (e.g., `Products` exists)

Therefore, any internal `dlt.Dataset` or `dlt.Relation` component needs to match the `dlt.Schema` normalization, even for constants

dlt/dataset/relation.py
```diff
@@ -417,6 +430,7 @@ def where(
 
         assert isinstance(column_or_expr, str)
         column_name = column_or_expr
+        normalized_column_name = self._dataset.schema.naming.normalize_identifier(column_name)
```
rudolfix: here you do not need to normalize column. `column_or_expr` (if `str`) should be already normalized. you do that already in the code that you fixed above. did you have any failing tests?

PR #3557, opened 2026-01-20
Feat: 3375 - abort packages manually

### LINKED ISSUE #3375: abort packages manually by default
**Background**
Currently any job that fails with terminal error will abort the package automatically. This is too opinionated and we should give users a choice what to do. The same mechanism can be also used to abort pending packages (ie. with retries) properly. Current mechanism does not ie. restore pipeline state.

I consider this change between fixing a bug and changing behavior. We must correct how packages are aborted but we leave a flag that will do auto abort on terminal exceptions.

**Tasks**

1. * [ ] Instead of failing packages, job with terminal exception should be moved to new jobs with retry count increased. We should start draining the package immediately.
2. * [ ] We need a new exception (not `LoadClientJobFailed`) to raise after drain to indicate that there was a terminal exception in one of the jobs.
3. * [ ] We need to store `.exception` messages in some kind of separate folder (previously we stored them only in `failed_jobs`). And we should do it for each retry (transient and terminal). exception message should indicate the state of the job in the first line: "retry" or "failed"
4. * [ ] We need to implement new `abort_package` helper which will be similar to `drop` helper (see pipeline top level module).

On job with terminal exception ("failed" state):
1. We do not move it to "failed_jobs" automatically, we retry it and issue pending LoadClientJobFailed. 
2. Pool drains, load ends with new exception and package is always pending. We do not abort automatically. (we do not complete with aborted flag, we raise)
3. User may chose to fail any job (except completed and failed already). Failing jobs moves those jobs to “failed_jobs” and copies exception message (see above) so we are backward compatible.
4. User should be able to easily list jobs that were retried and identify those that were retried due to terminal error
5. User may always chose to retry (even if some jobs are moved to failed)
6. User may chose to abort the package.
Abort must be a managed operation similar to `drop` helper. Action must happen via `pipeline.load` and generate run trace.

1. only pending package may be aborted (the one being currently loaded)
2. you mark package to be aborted ie. via writing a flag to package state and run `load`
3. if flag is present, load.py must abort the package

Aborting means:
- all jobs that were retried and are not in terminal state are moved to “failed_jobs”
- package is aborted and moved to “loaded” state (this is a terminal state). Existing code should be used to that
- when package is aborted: pipeline must be completely wiped out and re-synced. That include any other load package

Other tasks
- add new load option to auto abort on terminal error (current behavior) but do it as above. this will provide backward compatibility with current behavior
- we need a command line that will replace `drop-pending-packages` and that will add `fail-job` command on package command
- add new `abort_package` method
- replace `def drop_pending_packages(self, with_partial_loads: bool = True) -> None:
` with abort_package 

Docs:
- we need to update docs that explains when package is aborted
- we need to update entity diagram

anuunchin:
Resolves #3375 

#### Problem

When a job fails with a terminal error during load, dlt immediately moves it to `failed_jobs` and aborts the package — the user has no chance to inspect, retry, or decide what to do. The existing `drop_pending_packages` command silently deletes packages without recording what happened, and the pipeline state is left inconsistent with the destination.

#### What changed

**New load behavior: `auto_abort_on_terminal_error` flag (default `False`)**

**breaking**
Previously: packages with failed jobs were auto aborted, leaving user with no choice and (for non-ephemeral storage) with possibly inconsistent local state.
Now: on failed jobs package does not abort and the failed job may still be retried if you want to do it. Several options are added that help clean up inconsistent state if package is partially loaded. See https://dlthub.com/docs/running-in-production/running#partially-loaded-packages.

Note: on ephemeral storages where pipeline working dir is wiped this change does not have effect - they effectively auto abort by wiping the local storage. you can still benefit from new reconcillation tools described in docs.

Set auto_abort_on_terminal_error to True to go back to old behavior.

image

A new `LoaderConfiguration` option controls what happens when a job fails terminally:

| `auto_abort` | `raise_on_failed` | job action  | package action | Exception |
|---|---|---|---|---|
| `True` | `True` | fail | abort | LoadClientJobFailed raised (old default) |
| `True` | `False` | fail | abort | none |
| `False` | `True` | retry | keep pending | LoadClientJobTerminalRetry raised |
| `False` | `False` | fail | complete as loaded | none |

With `auto_abort_on_terminal_error=False` and `raise_on_failed_jobs=True`, the job is retried (moved back to `new_jobs` with incremented retry count) and the package stays pending. The user gets a `LoadClientJobTerminalRetry` exception and can decide to retry, fail individual jobs, or abort the whole package.

**Exception storage**

Every retry (transient or terminal) now saves the exception message to a `.exceptions` folder inside the load package. The first line indicates the type (`retry: terminal` or `retry: transient`).

**Abort mechanism**

Aborting a package is now a managed operation that runs through `pipeline.load()`:
1. An `abort_requested` flag is written to the package state
2. On next `load()`, the loader sees the flag and calls `abort_package` — moves retried pending jobs to `failed_jobs`, marks the package as aborted
3. Other pending packages are deleted and local pipeline state and schemas are restored from the `.restore` snapshot each package carries (rewinding to the point at which the aborted package started, without destination access)

`Pipeline.abort_packages()` handles the full flow: set the flag → `load()` → delete other packages and restore state. The CLI shows a dry-run plan and asks for confirmation before running it.

**Pipeline-level API for job management**

Three new methods on `Pipeline`:
- `list_pending_retry_jobs_in_package(load_id)` — list jobs pending a retry as (job file, folder) tuples
- `fail_pending_job(load_id, job_file_name)` — move a retried job to `failed_jobs`
- `retry_failed_job(load_id, job_file_name)` — move a failed job back for retry

**CLI commands**

- `dlt pipeline <name> abort-packages` — replaces `drop-pending-packages`. Shows what will be aborted (jobs, packages), confirms, then aborts and restores state.
- `dlt pipeline <name> load-package <load-id> fail-job <job>` — fail a specific pending retry job. Accepts either a `job_id` (e.g. `table_name.abc123.jsonl`) or full file name (e.g. `table_name.abc123.2.jsonl`). Shows exception info and retry count before confirming.
- `dlt pipeline <name> drop-pending-packages` — now shows a deprecation warning pointing to `abort-packages`.

**Deprecations**

- `Pipeline.drop_pending_packages()` emits `DltDeprecationWarning(since="1.30.0")` → use `abort_packages` instead
- `drop-pending-packages` CLI command prints deprecation warning → use `abort-packages`

#### Breaking changes

- `PackageStorage.retry_job` signature change: `retry_job(load_id, file_name)` now requires two additional positional arguments: `failed_message: str` and `exception_type: TExceptionType`.


dlt/pipeline/helpers.py
```diff
@@ -191,3 +192,34 @@ def prepare_refresh_source(
...
+    @property
+    def is_empty(self) -> bool:
+        return not self.load_ids
+
+    def __call__(self) -> LoadInfo:
+        self.pipeline.abort_packages(self.load_ids)
+        load_info = self.pipeline.load()
+        self.pipeline.drop()
+        self.pipeline.sync_destination()
+        return load_info
```
anuunchin, line 224: This should have proper exception wrapping, but I want to make sure this is going in the right direction at all (I tried to replicate pipeline abort)

dlt/_workspace/cli/_pipeline_command.py
```diff
@@ -297,14 +297,109 @@ def _display_pending_packages() -> Tuple[Sequence[str], Sequence[str]]:
...
+            " resync state. Use `dlt pipeline %s abort-packages` instead." % pipeline_name
+        )
         extracted_packages, norm_packages = _display_pending_packages()
         if len(extracted_packages) == 0 and len(norm_packages) == 0:
             fmt.echo("No pending packages found")
         if fmt.confirm("Delete the above packages?", default=False):
             p.drop_pending_packages(with_partial_loads=True)
             fmt.echo("Pending packages deleted")
 
+    if operation == "fail-job":
```
anuunchin, line 351: or  is `dlt pipeline <name> load-package <load-id> fail-job <job>` better? 👀

anuunchin: Maybe there should be `list-pending-jobs` as well, so the user experiences:

- Pendingterminal jobs exception
- list-pending-jobs cli
- fail-job

dlt/common/storages/load_package.py
```diff
@@ -465,7 +500,44 @@ def fail_job(self, load_id: str, file_name: str, failed_message: Optional[str])
             file_name,
         )
 
-    def retry_job(self, load_id: str, file_name: str) -> str:
+    def fail_pending_job(self, load_id: str, file_name: str) -> str:
```
rudolfix: same here

rudolfix (approved):
LGTM!

I fixed a few things to bring this up to date with changes made over last few months. Also both auto abort and manual abort go through exactly the same code path. Some code from `abort.py` got moved to pipeline (which must be able to auto abort now)

PR #3570, opened 2026-01-24
Redshift fix: Use correct view to determine schema existence

timH6502:
## Description

Currently, the RedshiftSqlClient relies on `INFORMATION_SCHEMA.SCHEMATA` to determine if a schema exists. Unfortunately, this is not listing all the schemas (that's why there is this workaround, where the `public` schema is handled explicitly). The current way can cause errors, if the schema already exists, but is not listed in `INFORMATION_SCHEMA.SCHEMATA` when attempting to create it. 

To fix this, I've added added a condition that checks if the schema exists using `svv_redshift_schemas`, which lists schemas accessible to the current user (see https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_REDSHIFT_SCHEMAS.html). I'm not aware if there are older Redshift versions, that do not have `svv_redshift_schemas`. Therefore, the old behaviour was preserved if the view does not exists. 

### Result
Redshift pipelines are less prone to intermittent “schema already exists” failures while keeping the previous behavior as a fallback when `svv_redshift_schemas` is not available/visible.

Code has been formatted using the black formatter. Please let me know if there's something else to be done.

### DIFF
--- a/dlt/destinations/impl/redshift/redshift.py
+++ b/dlt/destinations/impl/redshift/redshift.py
@@ -28,7 +28,7 @@ from dlt.common.configuration.specs import AwsCredentialsWithoutDefaults
 
 from dlt.destinations.insert_job_client import InsertValuesJobClient
 from dlt.destinations.sql_jobs import SqlMergeFollowupJob
-from dlt.destinations.exceptions import DatabaseTerminalException
+from dlt.destinations.exceptions import DatabaseTerminalException, DatabaseException
 from dlt.destinations.job_client_impl import CopyRemoteFileLoadJob
 from dlt.destinations.impl.postgres.sql_client import Psycopg2SqlClient
 from dlt.destinations.impl.redshift.configuration import RedshiftClientConfiguration
@@ -59,7 +59,8 @@ class RedshiftSqlClient(Psycopg2SqlClient):
         try:
             rows = self.execute_sql(query, *db_params)
             return len(rows) > 0
-        except psycopg2.Error:
+        except DatabaseException:
+            # fallback to INFORMATION_SCHEMA.SCHEMATA if svv_redshift_schemas is not available
             return super().has_dataset()
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


dlt/destinations/impl/redshift/redshift.py
```diff
@@ -48,11 +47,31 @@ class RedshiftSqlClient(Psycopg2SqlClient):
...
+            db_params = []
+            _, schema_name, _ = self._get_information_schema_components()
+            query = f"SELECT 1 FROM {all_schemas_view_name} WHERE schema_name = %s"
+            db_params.append(schema_name)
+            rows = self.execute_sql(query, *db_params)
+            return len(rows) > 0
+
         return super().has_dataset()
 
+    def create_dataset(self) -> None:
```
rudolfix: all destination fail when they create already existing dataset. this is enforced by standard test suite. let's not change that here. fix above should be sufficient to not re-create existing schemas

dlt/destinations/impl/redshift/redshift.py
```diff
@@ -48,11 +47,31 @@ class RedshiftSqlClient(Psycopg2SqlClient):
...
         # In Redshift, the 'public' schema always exists but may not be
         # returned by INFORMATION_SCHEMA.SCHEMATA query, so we handle it as a special case
-        if self.dataset_name == "public":
+        if self.dataset_name.lower() == "public":
             return True
 
+        all_schemas_view_name = "svv_redshift_schemas"
+        # Check for svv_redshift_schemas
+        check_svv_redshift_schemas_exists_query = (
+            "SELECT 1 FROM information_schema.tables WHERE table_name = %s"
```
rudolfix: `svv_redshift_schemas` is a restricted name? we do not need to select the schema where the system view resides?

rudolfix: also: maybe it makes sense to just use `svv_redshift_schemas` and fallback only if the view is not found? then we have less roundtrips to the server and less chances for a race condition. still makes sense to do a fallback. I can't find any info when `svv_redshift_schemas` is added but I expect it to be present on almost all Redshift installations

dlt/destinations/impl/redshift/redshift.py
timH6502: I removed that and adjusted the PR title.

dlt/destinations/impl/redshift/redshift.py
timH6502: Specifying a schema name is not necessary. 

We could use try-except logic and not check whether the schema is available first. Then, in the except block, we can use the old behavior.

> I expect it to be present on almost all Redshift

I'd expect that as well. However, since I'm not 100% sure, I think it would be better to preserve the old behavior.

timH6502:
> this is still missing. makes sense to skip the check

Done

rudolfix (approved):
LGTM! small correction in exception handing was needed

PR #3574, opened 2026-01-26
feat: add parallelized flag for rest_api dependent resources (#1883)

### LINKED ISSUE #1883: rest_api: use @defer decorator on transformer loop for parallel execution
### Source name

rest_api

### Describe the data you'd like to see

Dependent resources loop over all data items in a batch sent from the connected parent. We can easily parallelize this loop.
1. detect if `parallelize` flag is enabled on a transformer when generating loop over items
2. if so use decorated version of the function (via dlt.defer)
in essence such decorator wraps transformer function in callables and many of those can be executed at the same time

### Are you a dlt user?

Yes, I'm already a dlt user.

### Do you ready to contribute this extension?

Yes, I'm ready.

### dlt destination

_No response_

### Additional information

_No response_

Shadesfear:
Adds parallelized flag support for dependent resources (transformers) in rest_api_source. When enabled, child resource fetches run concurrently in dlt's thread pool instead of sequentially.
{
    "name": "post_comments",
    "parallelized": True,
    "endpoint": {"path": "posts/{resources.posts.id}/comments"},
}
Related Issues
- Closes #1883

### DIFF
--- a/dlt/sources/rest_api/config_setup.py
+++ b/dlt/sources/rest_api/config_setup.py
@@ -400,9 +400,9 @@ def expand_and_index_resources(
         _bind_path_params(endpoint_resource)
 
         resource_name = endpoint_resource["name"]
-        assert isinstance(resource_name, str), (
-            f"Resource name must be a string, got {type(resource_name)}"
-        )
+        assert isinstance(
+            resource_name, str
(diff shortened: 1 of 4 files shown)


rudolfix (approved):
LGTM! I added 2 missing tests for related cases (parallel + incremental + include from parent)

PR #3577, opened 2026-01-26
Add docs about parallel testing

tetelio:
This PR supplements #3479 to document the parallelization strategy in the contributing guide. A minimal overview of the parallelization strategy is given + some basic instructions on how to use tests helpers to keep tests parallel safe

The goal of the reviewer is:

1. Confirm documentation is clear
2. Confirm that documentation correctly refers to makefile where more implementation details of the parallelization can be found (but they dont need to be fully described in the contributing file to not overload it)
3. Confirm that reviewer would be able to create new parallel safe tests using the mentioned python test helpers in the contributing guide

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,9 +160,13 @@ Our goal is to maintain stability and compatibility across all environments. Ple
 
 Parallel testing has been introduced to allow for more time-efficient testing via `pytest-xdist`. The harnessing of parallel testing has already been done in the root Makefile of the project. You can consult the Makefile directly to see the specifics of the implementation. Summarized, the different testing suites that are described below will have an equivalent Make invocation command with a suffix `-p`, which will enable parallelized testing for the selected tests. This parallelizing strategy helps surfacing test leakages an execution-order dependencies.
 
-How does the parallelization strategy work? `pytest-xdist` spawns python processes where all tests are collected, then the `pytest-xdist` workers will pick up any test available to be executed (in any order, from any module). By default, when running tests locally, the number of python parallel processes is set to auto. This will spawn as many processes as cpu cores the machine has. You can override this value by passing PYTEST_XDIST_N to the make invocation command with the number of desired processes.
+How does the parallelization strategy work? `pytest-xdist` spawns python processes where all tests are collected, then the `pytest-xdist` workers will pick up any test available to be executed (in any order, from any module). By default, when running tests locally, the number of python parallel processes is set to auto. This will spawn as many processes as cpu cores the machine has. You can override this value by passing `PYTEST_XDIST_N` to the `make` invocation command with the number of desired processes. Example:
 
-Not all tests are able to be run in parallel, an example are tests that require all cpu cores saturating for performance reasons. Those tests have been marked with the pytest `serial` marker. Any parallel test execution will do a second pass to run tests that have the serial marker to be run without parallelism.
+```sh
+PYTEST_XDIST_N=4 make test-common-p
+```
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -154,11 +154,27 @@ Our goal is to maintain stability and compatibility across all environments. Ple
...
 ## Testing
 
-`dlt` uses `pytest` for testing. 
+`dlt` uses `pytest` for testing.
+
+### Parallel testing
+
+Parallel testing has been introduced to allow for more time-efficient testing via `pytest-xdist`. The harnessing of parallel testing has already been done in the root Makefile of the project. You can consult the Makefile directly to see the specifics of the implementation. Summarized, the different testing suites that are described below will have an equivalent Make invocation command with a suffix `-p`, which will enable parallelized testing for the selected tests. This parallelizing strategy helps surfacing test leakages an execution-order dependencies.
+
+How does the parallelization strategy work? `pytest-xdist` spawns python processes where all tests are collected, then the `pytest-xdist` workers will pick up any test available to be executed (in any order, from any module). By default, when running tests locally, the number of python parallel processes is set to auto. This will spawn as many processes as cpu cores the machine has. You can override this value by passing PYTEST_XDIST_N to the make invocation command with the number of desired processes.
```
zilto: Suggested edits:
- use \`  to wrap `PYTEST_XDIST_N` and `make`
- below paragraph, add example invocation

```shell
PYTEST_XDIST_N=4 make test
```

In my experience, using all CPUs can be very aggressive and slowdown other things you're doing on your machine. Would there be an option for like "75% of CPU threads"? WDYT

CONTRIBUTING.md
```diff
@@ -154,11 +154,27 @@ Our goal is to maintain stability and compatibility across all environments. Ple
...
+
+Not all tests are able to be run in parallel, an example are tests that require all cpu cores saturating for performance reasons. Those tests have been marked with the pytest `serial` marker. Any parallel test execution will do a second pass to run tests that have the serial marker to be run without parallelism.
```
zilto: Writing tip: the most important word should be at the very beginning of the paragraph. This helps people can long text. I recommend rephrasing:

> The `serial` pytest marker is set for tests that shouldn't run in paralle. This is useful when ...

CONTRIBUTING.md
tetelio: i will adapt the docs but @zilto would you be so kind to run `make test-common-p `with your machine on the devel branch? My macbook makes the fans spin faster but i dont get any delay/freezing behaviour.

In general machines are not using extensively all the cpu cores (another story is memory) and before adding any 75% cpu cores rule (which i dont know if i can easily achieve), i would like to know how your machine performs.

tetelio:
@zilto regarding your two global optional suggestions: i dont want to include fixture explanation because i dont even know myself why we have some fixtures in autodiscovery mode (in conftest.py files) and some others in other files where the fixtures need to be imported to be used. that alone throws me out of understanding the testing setup. i would be up for doing this after our other priorities as a standalone task, because i would need to dig deeper into why the testing setup looks like it does now

tetelio:
@zilto if you meant which fixtures are needed for parallel testing ,there is none, only the helpers to be used as pure python functions

CONTRIBUTING.md
zilto: Ok, ran it twice! It didn't affect performance for web browing, YouTube, music, IDE usage, etc. while it ran. It was able to saturate ~95% of CPU according to `htop`. RAM hovers around 11Gb which might be a bigger issue for some smaller devices

image

Once again, congrats!

tetelio: nice, as i expected. thanks for checking!

PR #3578, opened 2026-01-26
flatten logic; improve typing

zilto:
Remove the dynamic methods `._wrap_iter()` and `._wrap()` from `dlt.Relation`. While convenient, they added confusing steps to the stack trace and muddied the typing. 

Now, the flattened methods `.df()`, `.arrow()`, etc. have explicit typing instead of `Any`.

There should be exactly no behavior change, hence no test update

edit: test failures seem unrelated


tetelio:
@zilto i can chip in that the test failures is something that happens sometimes with the new parallel testing. I just replayed the two failed test and they passed with no problem. I will write down these two tests to see if they can be fixed

PR #3584, opened 2026-01-28
(fix) do not write `_dlt_load_id` as dict on mssql + adbc

### LINKED ISSUE #3551: Adding _dlt_load_id to Arrow fails when using adbc
### dlt version

1.20.0

### Describe the problem

I have a source which yields a PyArrow Table. I'm trying to load this data to an MSSQL destination using the adbc driver. I want to add the `_dlt_load_id` column to the data I'm loading. That's why I've added:
```
[normalize.parquet_normalizer]
add_dlt_load_id = true
```
to `config.toml`.
When I run the pipeline I get the following error:
```
adbc_driver_manager.NotSupportedError: NOT_IMPLEMENTED: [mssql] Unsupported bind parameter or bulk ingest type dictionary<values=utf8, indices=int8, ordered=false> for field _dlt_load_id
```

### Expected behavior

I expect that the data gets loaded to a table in the destination, containing the _dlt_load_id column.

### Steps to reproduce

You need a SQL Server Database or an Azure SQL Database as destination.
Add the a`dbc-driver-manager` and `dbc` packages to the project
```
pip install adbc-driver-manager dbc
dbc install mssql
```
The following example demonstrates the issue:
```
import dlt
import polars as pl

CONN_STR = "mssql://username:password@host/database?trust_server_certificate=true"


def data():
    df = pl.from_dict({"a": [1, 2], "b": [3, 4]})
    yield df.to_arrow()


test_pipeline = dlt.pipeline(
    pipeline_name="test",
    destination=dlt.destinations.mssql(CONN_STR),
    dataset_name="landing",
    progress="log",
    dev_mode=True,
)

# Run
info = test_pipeline.run(data=data(), loader_file_format="parquet")

print(info)
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

Azure SQL Database

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

rudolfix:
### Description
 fixes #3551

### DIFF
--- a/tests/load/test_dummy_client.py
+++ b/tests/load/test_dummy_client.py
@@ -427,6 +427,10 @@ def test_try_retrieve_job() -> None:
     load.pool = ThreadPoolExecutor()
     jobs = load.start_new_jobs(load_id, schema, [])  # type: ignore
     assert len(jobs) == 2
+    # wait for jobs submitted to the thread pool to complete before resuming
+    for j in jobs:
+        while j.state() not in ("completed", "failed", "retry"):
+            sleep(0.01)
     # now jobs are known
     jobs = load.resume_started_jobs(load_id, schema)
     assert len(jobs) == 2


tests/load/pipeline/test_adbc_loading.py
```diff
@@ -75,20 +78,79 @@ def test_adbc_parquet_loading(destination_config: DestinationTestConfiguration)
...
+
+    This test verifies the fix for https://github.com/dlt-hub/dlt/issues/3551
+    where ADBC drivers (especially MSSQL) fail on dictionary-encoded Arrow arrays.
+    """
+    if destination_config.destination_name == "sqlalchemy_sqlite":
+        pytest.skip(
+            "ADBC disabled for SQLite due to WAL mmap conflicts between sqlite3 and"
+            " adbc_driver_sqlite"
+        )
+    import pyarrow as pa
```
djudjuu: unused import

tests/load/pipeline/test_adbc_loading.py
```diff
@@ -75,20 +78,79 @@ def test_adbc_parquet_loading(destination_config: DestinationTestConfiguration)
...
+    assert rows[0][0] == data_["col1"]  # col1 (bigint)
+    assert rows[0][4] == data_["col5"]  # col5 (text)
+    # verify child table values
+    child_rows = pipeline.dataset().complex_resource__child.fetchall()
+    assert len(child_rows) == 3
+    assert set(row[0] for row in child_rows) == {1, 2, 3}  # value column
+    # verify pipeline state table values
+    state_rows = pipeline.dataset()._dlt_pipeline_state.fetchall()
+    assert len(state_rows) == 1
+    assert state_rows[0][2] == "pipeline_adbc"  # pipeline_name column
```
djudjuu, line 107: I don't see why we need this here.

tests/load/pipeline/test_adbc_loading.py
```diff
@@ -75,20 +78,79 @@ def test_adbc_parquet_loading(destination_config: DestinationTestConfiguration)
     def complex_resource():
         # add child table
         data_["child"] = [1, 2, 3]
-        yield data_
+        assert len([data_] * 10) == 10
+        yield [data_] * 10
```
djudjuu, line 83: learning question: So will this data be dictionary-encoded now? I am not sure I can see why that would happen. and are you testing merge behavior too?
so you're testing merge behavior too? nice, but could be mentioned with a comment since we test a few things at the same time now
would be nice to mention explicitly the purpose of the *10 (as these generated tests usually do spell it out)
```suggestion
        # yield the same data 10 times (same primary key to test batch-loading and merge-disposition)
        yield [data_] * 10
```

tests/load/pipeline/test_adbc_loading.py
```diff
@@ -75,20 +78,79 @@ def test_adbc_parquet_loading(destination_config: DestinationTestConfiguration)
...
 
     info = pipeline.run(complex_resource())
     jobs = get_load_package_jobs(
         info.load_packages[0], "completed_jobs", "complex_resource", ".parquet"
     )
     # there must be a parquet job or adbc is not installed so we fall back to other job type
```
djudjuu, line 89:
```suggestion
    # there must be a parquet job or adbc is not installed so we fall back to other job type
    # it should be just one job due to batching
```

tests/load/pipeline/test_adbc_loading.py
```diff
@@ -75,20 +78,79 @@ def test_adbc_parquet_loading(destination_config: DestinationTestConfiguration)
...
     assert len(jobs) == 1
-    # make sure we can read data back. TODO: verify data types
-    rows = pipeline.dataset().table("complex_resource").fetchall()
+    # verify row count and selected column values (int and string)
+    rows = pipeline.dataset().complex_resource.fetchall()
     assert len(rows) == 1
```
djudjuu, line 97:
```suggestion
    # merge disposition should deduplicate to one single row:
    assert len(rows) == 1
```

tests/conftest.py
```diff
@@ -53,6 +55,40 @@ def _list_vault(self):
...
+        "iceberg",
+        "lance",
+        "motherduck",
+        "parquet",
+        "postgres_scanner",
+        "spatial",
+        "sqlite_scanner",
+    ]
+
+    conn = duckdb.connect()
```
djudjuu: nitpick:
could be more defensive about the connection with a context manager and also don't swallow silent so we know there was someething unexpected with setup. Or do we expect this to happen a lot? still we use a learning signal
```
with duckdb.connect() as conn:
    for ext in extensions:
        try:
            conn.execute(f"INSTALL {ext}")
        except Exception as e:
            logger.debug(f"Could not pre-install DuckDB extension '{ext}': {e}")
```

tests/load/sqlalchemy/test_sqlalchemy_configuration.py
```diff
@@ -21,3 +28,65 @@ def test_sqlalchemy_credentials_from_engine() -> None:
...
+    # explicit relative location: relocated to local_dir
+    c = resolve_configuration(
+        SqlalchemyClientConfiguration(
+            credentials=SqlalchemyCredentials("sqlite:///./local.db"),
+        )._bind_dataset_name(dataset_name="test_dataset")
+    )
+    db_path = os.path.join(local_dir, "local.db")
+    assert c.credentials.database.endswith(db_path)
+
+    # pipeline context: uses pipeline_name for the filename
```
djudjuu: just a nitpick: this tests for using ~/.dlt/<pipeline_name.db>

we could add one for :pipeline: arg which shuold go to ~/.dlt/pipelines/<pipeline_name.db> I believe...

tests/load/pipeline/test_adbc_loading.py
rudolfix: because ADBC driver for sqlite was corrupting data and I was getting values from a different table. So I added a smoke test. And at the end - sqllite ADBC support got removed

tests/load/pipeline/test_adbc_loading.py
rudolfix: valid point. this was mostly because of sqllite again corrupting data - testing what happens when I load more than 1 row. I will add a comment

PR #3585, opened 2026-01-28
feat/3525-update-dataset-schema-api

### LINKED ISSUE #3525: feat: API to update `dlt.Dataset.schema`
## Summary
When a `dlt.Dataset` object is instantiated, the `.schema` property can't be updated. This means the `dlt.Dataset` object is not aware of new tables created after init (by this process or other processes).

For example, this throws "table not found"
```python
import dlt

@dlt.hub.transformation
def bar(dataset: dlt.Dataset):
   yield [{"id": 1}]


pipeline = dlt.pipeline(...)
# dataset is instantiated without `bar`
dataset = pipeline.dataset()
pipeline.run(bar(dataset))

# `dataset` object has outdated schema that doesn't
# include `bar`
dataset.table("bar") 
```
 
## Problem
Instantiating a `dlt.Dataset` requires a `dlt.Schema`. It can be provided in 3 ways:
1. explicitly at `dlt.Dataset.__init__()`
2. automatically retrieved from local `.dlt/`
3. automatically retrieved from destination

Though, the schema is never refreshed once set to `._schema`. This behavior is in-place to prevent making multiple calls to the destination.

## Solution
Add an explicit method to refresh the dataset's schema. This method can be private while we experiment. I believe similar logic exists on the `dlt.Pipeline` object.

Mock up
```python
pipeline = dlt.pipeline(...)
# dataset is instantiated without `bar`
dataset = pipeline.dataset()
pipeline.run(bar(dataset))

# `dataset` object has outdated schema that doesn't
# include `bar`
dataset.sync_schema()
dataset.table("bar") 
```

This `.sync_schema()` could be called implicitly by `dlt.Pipeline.run()` when it processes a transformation.

### `.sync_schema()` signature
The `sync_schema` method could simply delete the cached `._schema`, forcing the object to reevaluate the schema. It could also provide an explicit interface to select behavior 1, 2, or 3
```python
class Dataset:
   # ...
   def sync_schema(
      self,
      schema: dlt.Schema | None = None,  # for behavior 1
      local_only: bool = False,  # True for behavior 2; False for 3
   )
```
```

### PR DESCRIPTION
**Pull request by munish7771**
### Description
This PR adds a `sync_schema` method to `dlt.Dataset`, allowing us to refresh a dataset’s schema when it becomes stale.
It supports:
1. Fetching the latest schema by dataset name (default behavior)
2. Updating from an explicit `dlt.Schema` or schema name
3. Resetting to a local-only schema state

### Related Issues
- Fixes #3525

### Additional Context
- Added a new function `sync_schema` to `dlt\dataset\dataset.py`
- Added unit tests to `tests\dataset\test_dataset.py`

### CODE REVIEW

**Comment by munish7771** [comment]
@zilto looks like the actions failed due to an internal server error. Could you please re trigger the workflow and review this PR?

**Comment by munish7771** [comment]
@rudolfix The refresh breaks when called via `pipeline.dataset().refresh()`
Looks like the pipeline.dataset() resolves the default schema to a full schema object before passing it to dataset. In refresh() it can't tell if the schema came from the user or the pipeline default. What do you suggest?

**Comment by rudolfix** [comment]
@munish7771 feel free to store origin schema argument to be able to pass to new dataset instance in refresh(). ie. `self._schema_origin = schema` in `__init__`

**Comment by munish7771** [comment]
@rudolfix what i meant was if we create the dataset using the pipeline api using dataset(), the function always provides the dataset constructor with a schema even if we explicitly didn't add any. in this case it will always be a caught exception. 
for-eg, with the current code changes if i run the following - 
```
import dlt

pipeline = dlt.pipeline(
    pipeline_name="test_pipeline",
    destination="duckdb",
    dataset_name="test_dataset"
)
pipeline.run([{"id": 1, "name": "Sharma"}], table_name="users")
dataset = pipeline.dataset()
dataset.refresh()
```
it gives error - **TypeError: refresh() is not supported when the Dataset was created with a Schema instance.**

is this intended? One way I can think of is to change the behaviour of the pipeline.dataset() function, but I am not sure if its that way for a reason.

zilto:
@munish7771 Can you rebase your branch on `devel`? 

Some tests are failing on CI, but are unrelated to your changes. They were recently fixed and merged to devel

munish7771:
Replaced by a clean PR due to rebase issues → #3802
FYI: @zilto @rudolfix

PR #3590, opened 2026-01-29
feat: add `dlt.Relation.join(...)` based on normalizer and references

Travior:
WIP for `join()` on `dlt.Relation`

Still a lot of development artifacts (like unused join methods etc., missing tests etc.).

### DIFF
--- a/.github/workflows/test_destinations_remote.yml
+++ b/.github/workflows/test_destinations_remote.yml
@@ -106,6 +106,22 @@ jobs:
               post_install_commands: "uv run pip install openai"
               always_run_all_tests: true
 
+            # Lance — file and s3 protocols only, no need to test az and gs as all cloud storages go
+            # through same code paths thanks to `object_store` Rust crate
+            - name: lance_local
+              destinations: "[\"lance\"]"
+              filesystem_drivers: "[\"file\"]"
+              extras: "--extra lance"
+              post_install_commands: "uv run pip install openai"
+              always_run_all_tests: true
+
+            - name: lance_s3
+              destinations: "[\"lance\"]"
+              filesystem_drivers: "[\"s3\"]"
+              extras: "--extra lance --extra s3"
+              post_install_commands: "uv run pip install openai"
+              always_run_all_tests: true
+
             # Motherduck
             - name: motherduck
               destinations: "[\"motherduck\"]"
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -748,6 +748,28 @@ def get_normalized_arrow_fields_mapping(schema: pyarrow.Schema, naming: NamingCo
     return name_mapping
 
 
+def dlt_column_to_arrow_field(
+    column: TColumnSchema,
+    caps: DestinationCapabilitiesContext,
+    timestamp_timezone: str = "UTC",
+) -> pyarrow.Field:
+    """Convert a single dlt column schema to a PyArrow field.
+
+    Args:
+        column (TColumnSchema): dlt column schema with at least `name` and `data_type`.
+        caps (DestinationCapabilitiesContext): Destination capabilities for type mapping.
+        timestamp_timezone (str): Timezone for timestamp columns.
+
+    Returns:
+        pyarrow.Field: Corresponding PyArrow field.
+    """
+    return pyarrow.field(
+        column["name"],
+        get_py_arrow_datatype(column, caps, timestamp_timezone),
+        nullable=column.get("nullable", True),
+    )
+
+
 def columns_to_arrow(
     columns: TTableSchemaColumns,
     caps: DestinationCapabilitiesContext,
@@ -765,17 +787,9 @@ def columns_to_arrow(
     caps = caps or DestinationCapabilitiesContext.generic_capabilities()
     return pyarrow.schema(
         [
-            pyarrow.field(
-                name,
-                get_py_arrow_datatype(
-                    schema_item,
-                    caps,
-                    timestamp_timezone,
-                ),
-                nullable=schema_item.get("nullable", True),
-            )
-            for name, schema_item in columns.items()
-            if schema_item.get("data_type") is not None
+            dlt_column_to_arrow_field(column, caps, timestamp_timezone)
+            for column in columns.values()
+            if column.get("data_type") is not None
         ]
     )
 
--- a/dlt/common/libs/pydantic.py
+++ b/dlt/common/libs/pydantic.py
@@ -44,7 +44,13 @@ try:
 
     if not PYDANTIC_VERSION.startswith("2."):
         raise ImportError(f"Found pydantic {PYDANTIC_VERSION} but dlt requires pydantic>=2.0")
-    from pydantic import BaseModel, ValidationError, Json, create_model
+    from pydantic import (
+        BaseModel,
+        ValidationError,
+        Json,
+        create_model,
+        RootModel as PydanticRootModel,
+    )
     from pydantic.fields import FieldInfo
     from pydantic.warnings import PydanticDeprecationWarning
 
@@ -122,6 +128,7 @@ def _build_discriminator_map(
 
     ann = root_field.annotation
     discriminator: Optional[str] = None
+    union_args: Tuple[Any, ...] = ()
 
     if is_annotated(ann):
         args = get_args(ann)
@@ -135,8 +142,13 @@ def _build_discriminator_map(
             if discriminator:
                 break
         union_args = get_args(args[0])
-    else:
-        return None
+
+    # pydantic 2.13+: annotation is plain Union (not Annotated), discriminator
+    # is on root_field directly
+    if not discriminator and isinstance(root_field.discriminator, str):
+        discriminator = root_field.discriminator
+    if not union_args:
+        union_args = get_args(ann)
 
     if not discriminator or not union_args:
         return None
@@ -326,8 +338,6 @@ def apply_schema_contract_to_model(
     if data_mode == "evolve":
         # create a lenient model that accepts any data
         if getattr(model, "__pydantic_root_model__", False):
-            from pydantic import RootModel as PydanticRootModel
-
             model = type(
                 model.__name__ + "Any",
                 (PydanticRootModel[Any],),
@@ -405,11 +415,12 @@ def apply_schema_contract_to_model(
             return f.annotation  # type: ignore[no-any-return]
 
     if getattr(model, "__pydantic_root_model__", False):
-        from pydantic import RootModel as PydanticRootModel
-
         root_field = model.model_fields.get("root")
         if root_field:
             processed_ann = _process_annotation(_rebuild_annotated(root_field))
+            # preserve discriminator that pydantic 2.13+ strips from Annotated metadata
+            if isinstance(root_field.discriminator, str) and not is_annotated(processed_ann):
+                processed_ann = Annotated[processed_ann, FieldInfo(discriminator=root_field.discriminator)]  # type: ignore[assignment]
             new_rm = type(
                 model.__name__ + "Extra" + extra.title(),
                 (PydanticRootModel[processed_ann],),  # type: ignore[valid-type]
--- a/dlt/destinations/__init__.py
+++ b/dlt/destinations/__init__.py
@@ -11,6 +11,7 @@ from dlt.destinations.impl.athena.factory import athena
 from dlt.destinations.impl.redshift.factory import redshift
 from dlt.destinations.impl.qdrant.factory import qdrant
 from dlt.destinations.impl.lancedb.factory import lancedb
+from dlt.destinations.impl.lance.factory import lance
 from dlt.destinations.impl.motherduck.factory import motherduck
 from dlt.destinations.impl.weaviate.factory import weaviate
 from dlt.destinations.impl.destination.factory import destination
@@ -35,6 +36,7 @@ __all__ = [
     "redshift",
     "qdrant",
     "lancedb",
+    "lance",
     "motherduck",
     "weaviate",
     "synapse",
--- a/dlt/destinations/adapters.py
+++ b/dlt/destinations/adapters.py
@@ -2,7 +2,8 @@
 
 from dlt.destinations.impl.weaviate.weaviate_adapter import weaviate_adapter
 from dlt.destinations.impl.qdrant.qdrant_adapter import qdrant_adapter
-from dlt.destinations.impl.lancedb import lancedb_adapter
+from dlt.destinations.impl.lance.lance_adapter import lance_adapter
+from dlt.destinations.impl.lancedb.lancedb_adapter import lancedb_adapter
 from dlt.destinations.impl.bigquery.bigquery_adapter import bigquery_adapter, bigquery_partition
 from dlt.destinations.impl.synapse.synapse_adapter import synapse_adapter
 from dlt.destinations.impl.clickhouse.clickhouse_adapter import clickhouse_adapter
@@ -14,6 +15,7 @@ from dlt.destinations.impl.filesystem.iceberg_adapter import iceberg_adapter, ic
 __all__ = [
     "weaviate_adapter",
     "qdrant_adapter",
+    "lance_adapter",
     "lancedb_adapter",
     "bigquery_adapter",
     "bigquery_partition",
--- a/dlt/destinations/impl/lance/configuration.py
+++ b/dlt/destinations/impl/lance/configuration.py
@@ -1,0 +1,194 @@
+from __future__ import annotations
+
+import dataclasses
+import os
+from typing import TYPE_CHECKING, Any, Dict, Literal, Optional, Final, ClassVar, List, Type
+
+from dlt.common import logger
+from dlt.common.configuration import configspec
+from dlt.common.configuration.specs.base_configuration import (
+    BaseConfiguration,
+    CredentialsConfiguration,
+    resolve_type,
+)
+from dlt.common.configuration.specs.mixins import WithObjectStoreRsCredentials
+from dlt.common.destination.client import DestinationClientDwhConfiguration
+from dlt.common.storages.configuration import (
+    FileSystemCredentials,
+    FilesystemConfigurationWithLocalFiles,
+    WithLocalFiles,
+)
+
+if TYPE_CHECKING:
+    from lance.namespace import DirectoryNamespace
+    from lancedb.embeddings import EmbeddingFunction, EmbeddingFunctionRegistry
+
+DEFAULT_LANCE_BUCKET_URL = "."  # active run dir
+DEFAULT_LANCE_NAMESPACE_NAME = "dlt_lance_root"
+DEFAULT_LANCE_VECTOR_COLUMN_NAME = "vector"
+
+# NOTE: list providers with `EmbeddingFunctionRegistry.get_instance()._functions.keys()`
+TEmbeddingProvider = Literal[
+    "bedrock-text",
+    "cohere",
+    "colbert",
+    "colpali",
+    "gemini-text",
+    "gte-text",
+    "huggingface",
+    "imagebind",
+    "instructor",
+    "jina",
+    "ollama",
+    "open-clip",
+    "openai",
+    "sentence-transformers",
+    "siglip",
+    "voyageai",
+    "watsonx",
+]
+
+
+@configspec(init=False)
+class LanceStorageConfiguration(FilesystemConfigurationWithLocalFiles):
+    namespace_name: Optional[str] = DEFAULT_LANCE_NAMESPACE_NAME
+    """Name of subdirectory in `bucket_url` to use as namespace root. Leave empty to use `bucket_url` as namespace root."""
+    branch_name: Optional[str] = None
+    """Name of branch to use for read/write table operations. Uses `main` branch if not set."""
+    options: Optional[Dict[str, str]] = None
+    """Options to pass to storage client. See available options at https://lance.org/guide/object_store/#general-configuration.
+
+    Will be merged with `credentials`-derived options (if present), with `options` taking precedence in case of conflicts.
+    """
+
+    __config_gen_annotations__: ClassVar[List[str]] = ["bucket_url"]
+
+    def __init__(
+        self,
+        bucket_url: str = DEFAULT_LANCE_BUCKET_URL,
+        namespace_name: Optional[str] = DEFAULT_LANCE_NAMESPACE_NAME,
+        branch_name: Optional[str] = None,
+        credentials: Optional[FileSystemCredentials] = None,
+        options: Optional[Dict[str, str]] = None,
+    ) -> None:
+        super().__init__(bucket_url=bucket_url, credentials=credentials)
+        self.namespace_name = namespace_name
+        self.branch_name = branch_name
+        self.options = options
+
+    @property
+    def namespace_uri(self) -> str:
+        namespace_uri = self.bucket_url.rstrip("/")
+        if self.namespace_name:
+            namespace_uri += "/" + self.namespace_name
+        return namespace_uri
+
+    @resolve_type("credentials")
+    def resolve_credentials_type(self) -> Type[CredentialsConfiguration]:
+        return super().resolve_credentials_type()
+
+    def on_resolved(self) -> None:
+        super().on_resolved()
+        credentials = (
+            self.credentials.to_object_store_rs_credentials()
+            if isinstance(self.credentials, WithObjectStoreRsCredentials)
+            else {}
+        )
+        # set default timeouts for cloud storage to prevent infinite hangs in the Rust object_store
+        if not self.is_local_filesystem:
+            defaults = {"connect_timeout": "30s", "timeout": "120s"}
+            self.options = defaults | (self.options or {})
+        if credentials or self.options:
+            self.options = credentials | (self.options or {})
+
+    def make_directory_namespace(self) -> DirectoryNamespace:
+        from lance.namespace import DirectoryNamespace
+
+        storage_props = {f"storage.{k}": v for k, v in (self.options or {}).items()}
+        return DirectoryNamespace(root=self.namespace_uri, **storage_props)
+
+
+@configspec
+class LanceEmbeddingsCredentials(CredentialsConfiguration):
+    api_key: str = None
+    """API key for embedding model provider."""
+
+    __config_gen_annotations__: ClassVar[List[str]] = ["api_key"]
+
+
+@configspec
+class LanceEmbeddingsConfiguration(BaseConfiguration):
+    credentials: Optional[LanceEmbeddingsCredentials] = None
+    """Credentials for embedding model provider. Leave empty if authentication is not required (e.g. local providers)."""
+
+    vector_column: str = DEFAULT_LANCE_VECTOR_COLUMN_NAME
+    """Name of column to store vector embeddings in."""
+    provider: TEmbeddingProvider = None
+    """Provider of model used to generate embeddings, e.g. `cohere` or `openai`.
+
+    Find all providers at https://github.com/lancedb/lancedb/tree/main/python/python/lancedb/embeddings.
+    """
+    name: str = None
+    """Name of model used by provider to generate embeddings, e.g. `embed-english-v3.0`."""
+    max_retries: Optional[int] = 3
+    """Number of retries for embedding requests. Set to 0 to disable retries."""
+    kwargs: Optional[Dict[str, Any]] = None
+    """Additional provider-specific keyword arguments passed to `EmbeddingFunction.create()`."""
+
+    __config_gen_annotations__: ClassVar[List[str]] = [
+        "provider",
+        "name",
+    ]
+
+    # NOTE: env var names derived from `lancedb==0.26.1` source code.
+    # Providers with `api_key` field (openai, jina, watsonx) also accept
+    # the key via kwargs, but we set the env var regardless for consistency.
+    _PROVIDER_ENV_VAR_NAMES: ClassVar[Dict[TEmbeddingProvider, str]] = {
+        "cohere": "COHERE_API_KEY",
+        "gemini-text": "GOOGLE_API_KEY",
+        "jina": "JINA_API_KEY",
+        "openai": "OPENAI_API_KEY",
+        "voyageai": "VOYAGE_API_KEY",
+        "watsonx": "WATSONX_API_KEY",
+    }
+
+    def create_embedding_function(self) -> EmbeddingFunction:
+        from lancedb.embeddings import EmbeddingFunctionRegistry
+
+        if self.credentials:
+            self._set_provider_api_key_env_var(self.credentials.api_key)
+        kwargs = {"name": self.name, "max_retries": self.max_retries} | (self.kwargs or {})
+        return EmbeddingFunctionRegistry.get_instance().get(self.provider).create(**kwargs)
+
+    def _set_provider_api_key_env_var(self, api_key: str) -> None:
+        if env_var := self._PROVIDER_ENV_VAR_NAMES.get(self.provider):
+            os.environ[env_var] = api_key
+        else:
+            logger.warning(
+                "Provider %r is not in _PROVIDER_ENV_VAR_NAMES. The provided API key"
+                " will be ignored.",
+                self.provider,
+            )
+
+
+@configspec
+class LanceClientConfiguration(WithLocalFiles, DestinationClientDwhConfiguration):
+    destination_type: Final[str] = dataclasses.field(  # type: ignore
+        default="lance", init=False, repr=False, compare=False
+    )
+    storage: LanceStorageConfiguration = None
+    """Storage configuration including URI and cloud credentials."""
+    embeddings: Optional[LanceEmbeddingsConfiguration] = None
+    """Optional embeddings configuration to add a vector embedding column. Leave empty to skip embedding generation."""
+
+    def on_resolved(self) -> None:
+        # propagate pipeline context (local_dir, pipeline_name, etc.) to nested storage
+        if isinstance(self.storage, WithLocalFiles):
+            self.storage.attach_from(self)
+            if not self.storage.is_resolved():
+                self.storage.resolve()
+            else:
+                self.storage.normalize_bucket_url()
+
+    def fingerprint(self) -> str:
+        return self.storage.fingerprint()
--- a/dlt/destinations/impl/lance/exceptions.py
+++ b/dlt/destinations/impl/lance/exceptions.py
@@ -1,0 +1,56 @@
+from functools import wraps
+import re
+from typing import Any, List
+
+from dlt.common.destination.exceptions import (
+    DestinationUndefinedEntity,
+    DestinationTerminalException,
+    DestinationTransientException,
+)
+from dlt.common.destination.client import JobClientBase
+from dlt.common.typing import TFun
+
+
+LANCE_NOT_FOUND = r"not\s+found"
+LANCE_DOES_NOT_EXIST = r"does\s+not\s+exist"
+LANCE_MANIFEST_MODE = r"manifest\s+mode\s+is\s+enabled"
+
+LANCE_UNDEFINED_ENTITY_PATTERN = re.compile(
+    rf"(?i){LANCE_NOT_FOUND}|{LANCE_DOES_NOT_EXIST}|{LANCE_MANIFEST_MODE}"
+)
+
+
+class LanceEmbeddingsConfigurationMissing(DestinationTerminalException):
+    def __init__(self, table_name: str, columns: List[str]) -> None:
+        columns_str = ", ".join(f"'{col}'" for col in columns)
+        super().__init__(
+            f"Table `{table_name}` has columns marked for embedding ({columns_str}) but is"
+            " missing embeddings configuration. Either configure `embeddings` on the lance"
+            " destination or remove the `embed` argument from `lance_adapter()`."
+        )
+
+
+def is_lance_undefined_entity_exception(e: Exception) -> bool:
+    """Returns True if exception indicates an undefined entity (e.g. missing namespace, table, or branch).
+
+    Used to work around:
+    - bug: https://github.com/lance-format/lance/issues/6240
+    - fact that `LanceDataset.checkout_version()` raises `ValueError` with "not found" message if
+    version does not exist
+    """
+    return isinstance(e, (RuntimeError, ValueError)) and bool(
+        LANCE_UNDEFINED_ENTITY_PATTERN.search(str(e))
+    )
+
+
+def raise_destination_error(f: TFun) -> TFun:
+    @wraps(f)
+    def _wrap(self: JobClientBase, *args: Any, **kwargs: Any) -> Any:
+        try:
+            return f(self, *args, **kwargs)
+        except Exception as e:
+            if is_lance_undefined_entity_exception(e):
+                raise DestinationUndefinedEntity(e) from e
+            raise DestinationTransientException(e) from e
+
+    return _wrap  # type: ignore[return-value]
--- a/dlt/destinations/impl/lance/factory.py
+++ b/dlt/destinations/impl/lance/factory.py
@@ -1,0 +1,96 @@
+from __future__ import annotations
+
+from typing import Any, Type, TYPE_CHECKING
+
+from dlt.common.destination.configuration import ParquetFormatConfiguration
+from dlt.common.destination import Destination, DestinationCapabilitiesContext
+from dlt.common.destination.capabilities import DataTypeMapper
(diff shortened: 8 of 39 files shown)


dlt/dataset/relation.py
```diff
@@ -345,6 +447,27 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
         rel._sqlglot_expression = rel.sqlglot_expression.order_by(order_expr)
         return rel
 
+    def join(self, other: str | Self, how: Literal["left", "inner"]) -> Self:
+        if not self._table_name:
```
Travior: Can we be this defensive here? If we don't have `self._table_name` we have modified the relation (through filter or different transformations) and can't infer the correct joins anymore right?

dlt/dataset/relation.py
```diff
@@ -54,6 +55,107 @@
...
 
 
+def _parent_reference_chain_to_root(
+    schema: dlt.Schema, table: str
+) -> list[TTableReferenceStandalone]:
+    chain: list[TTableReferenceStandalone] = []
+    while parent := schema.tables[table].get("parent"):
+        chain.append(schema_utils.create_parent_child_reference(schema.tables, table))
+        table = parent
+    return chain
```
Travior: This seems very similar to what we do in `_add_load_via_parent_key_query` (which uses
```python
    for reference in schema_utils.get_all_parent_child_references_from_root(
        table_schemas, root_table_name
    ):
        reference_chain.append(reference)
        if reference["table"] == table_name:
            break
```
).
Is there a potential problem if we have multiple "parallel branches from the root table"? The utils function gives all children depth first in a flat list.
Also I suppose the correct type is just `TTableReference` or would it be fine to create a `TTableReferenceStandalone` here as we know the table name?

dlt/dataset/relation.py
```diff
@@ -54,6 +55,107 @@
...
+        raise ValueError(f"Reference {ref} has no base table")
+    return TTableReferenceStandalone(
+        {
+            **ref,
+            "table": ref["referenced_table"],
+            "referenced_table": ref["table"],
+            "columns": ref["referenced_columns"],
+            "referenced_columns": ref["columns"],
+        }
+    )
```
Travior: Is it fine to break the invariant "a reference always points from cold to parent" to make joins like `product` -> `product__items` easier if specified like this? Or should we keep the direction and handle this once we generate join conditions?

dlt/dataset/relation.py
```diff
@@ -54,6 +55,107 @@
...
+    # direct reference
+    for ref in schema.references:
+        if ref["table"] == left and ref["referenced_table"] == right:
+            return [ref]
+        if ref["table"] == right and ref["referenced_table"] == left:
+            return [_reverse_ref(ref)]
+
+    # through n-level parent-child references
+    chain = _resolve_parent_reference_chain(schema, left, right)
+    return chain
```
Travior: Is there another case of references that we need to handle (e.g. is there a case where we have maybe a dlt_root_id that is not part of `schema.references` or would be handled by walking up/down the parent-child chain?

dlt/dataset/relation.py
zilto: - we should be defensive
- the `._table_name` has been implicit so far. We could add a `.base_table_name` or `.origin_table_name` using a `@property`. It would return the `._table_name` (str or None)
- you're right that `.join()` probably doesn't work if chained after `.where()`. We should update `.where()` to preserve the `._table_name` if present
- we should have a test like this
```python
@pytest.mark.parametrize(
  ("method_name", "kwargs"),
  [
    ("where", {...}),
    ("select", {...}),
    ...
  ]
)
@pytest.mark.parametrize(
  "relation",   # parametrize for rels that do and do not have `._table_name`
)
def test_origin_table_name_is_preserved(
  relation: dlt.Relation,
  method_name: str,
  kwargs: dict[str, Any],
) -> None:
  expected_table_name = ...
  method = getattr(relation, method_name)

  output_rel = method(**kwargs)

  assert output_rel.origin_table_name = expected_table_name
```

dlt/dataset/relation.py
```diff
@@ -345,6 +447,27 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
...
         return rel
 
+    def join(self, other: str | Self, how: Literal["left", "inner"]) -> Self:
+        if not self._table_name:
+            raise ValueError("Reference-based join requires a base table relation")
+
+        left = self._table_name
+        right = other._table_name if isinstance(other, Relation) else other
+        if not isinstance(right, str):
+            raise ValueError("Unable to fetch table name for relation")
```
zilto: Could have clearer error message

```python
ValueError("Relation `other` doesn't have a `origin_table_name`. Join can't be resolved.")
```

dlt/dataset/relation.py
zilto: If we want to support `parent -> child` and  `child -> parent`

```python
dataset.table("orders").join("orders__items")

dataset.table("orders__items").join("orders")
```

We need to consider a few other things:
- the `cardinality` and `label` or `TTableReference` are not reversible (as done here via `**ref`). e.g., A user has a single password (user, one-to-one, password), but multiple users can have the same password (password, one-to-many, user) 
- In parent -> child, the default should be `left join`. In child -> parent, the default should be `right join` (for the `Relation.join()` API)
- The `origin_table_name` will remain the first one defined by `dataset.table(...)`. This means
```python
# valid: parent -> child1 -> child 2
dataset.table("orders").join("orders__items").join("orders__items__details")
# valid: child1 -> parent -> child2
dataset.table("orders__items").join("orders").join("orders__items__details")
# valid: parent -> child1 -> child1 from other branch
dataset.table("orders").join("orders__items").join("orders__status_updates")

# invalid: child1 -> parent -> child1 from other branch
dataset.table("orders__items").join("orders").join("orders__status_updates")
```
  - we __could__ support this invalid case, but then we open the door for unexpected errors. At this point, it seems clearer if users write the join directly with SQL / Ibis

dlt/dataset/relation.py
Travior: I believe this is an issue. Look at #3610

rudolfix:
my 2 cents in here. please allow for overloaded signature where `other` may be also a `Relation`: this will allow to join relations from other datasets/schemas. Internally I'll track those in parent relation to produce complete multi dataset lineage. Actually this change is pretty trivial. Here's the signature:

```py
def join(
      self,
      other: Relation,
      on: Union[str, sge.Expression],
      how: Literal["inner", "left", "right", "full"] = "inner",
  ) -> Relation:
      """Join this relation with another relation.

      Supports cross-schema and cross-dataset joins if both relations are on the
      same physical destination.

      Args:
          other: Relation to join with (can be from different schema/dataset)
          on: Join condition as SQL string or SQLGlot expression
              Examples: "left.id = right.id" or sge.EQ(...)
          how: Join type - "inner", "left", "right", or "full"

      Returns:
          New Relation representing the joined query

      Example:
          >>> customers = dataset1.table("customers")
          >>> orders = dataset2.table("orders")
          >>> joined = customers.join(
          ...     orders,
          ...     on="customers.customer_id = orders.customer_id",
          ...     how="left"
          ... )
      """
 ```

File: dlt/common/schema/utils.py [hunk 6]
```diff
@@ -1232,6 +1232,22 @@ def get_all_parent_child_references_from_root(
     return children_refs
 
 
+def get_all_parent_references_to_root(
+    tables: TSchemaTables, table_name: str
+) -> list[TTableReference]:
```
zilto, line 1216: Good pattern to always return a `list` instead of `None | list`. Simplifies logic for callers of this function

dlt/dataset/relation.py
```diff
@@ -350,6 +529,151 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
...
+
+        Args:
+            other: Table name or relation to join.
+        Returns:
+            A new relation with the join(s) applied.
+        """
+        join_plan = self._discover_join_params(other)
+        return self._apply_join(join_plan)
+
+    def _resolve_magic_join_target(self, other: str | Self) -> str:
```
zilto: ### Avoid private methods on public objects when possible
This method along with `_discover_join_params()`, `_get_or_create_join_base_query()`, `_get_or_create_join_base_query()`, and `_apply_joins()` are properly implemented a "private" methods with the `_` prefix.

Even though end-users shouldn't use them, making private methods available on a public object will inevitably result in users depending on private behavior. 

#### Methods as functions
To minimize this risk, it's safer to implement as utility functions that are detached from `dlt.Relation`. It's typically a simple rewrite if the method is stateless. For example `_resolve_magic_join_target` becomes

```python
def _resolve_magic_join_target(left: dlt.Relation, other: str | dlt.Relation) -> str:
  # every `self` is now `left` 
``` 

If we expect other parts of the codebase to use them (i.e., some sort of "robust internal method"), we can remove the `_` prefix to communicate to other developers / end-users "hey this one is reliable and tested enough"

```python
def resolve_magic_join_target(left: dlt.Relation, other: str | dlt.Relation) -> str: ... 
```

dlt/dataset/relation.py
```diff
@@ -55,6 +62,164 @@
...
+            return list(zip(self.referenced_columns, self.columns))
+        return list(zip(self.columns, self.referenced_columns))
+
+    @property
+    def join_type(self) -> TJoinType:
+        """The SQL join type: 'left' or 'right'."""
+        return "left" if self.direction == "LEFT" else "right"
+
+
+def _to_join_ref(ref: TTableReference, direction: Literal["LEFT", "RIGHT"]) -> _JoinRef:
```
zilto: given `direction` here needs to have the same `Literal` as `_JoinRef.direction`, it should be extracted out

tests/dataset/test_relation_join.py
```diff
@@ -0,0 +1,537 @@
...
+            "missing 'table' or 'referenced_table'",
+        ),
+        (
+            TTableReference(
+                table="users__orders",
+                referenced_table="users",
+                columns=[],
+                referenced_columns=["id"],
+            ),
+            "RIGHT",
```
zilto: hardcoded values are a good indicator that there should be a `TJoinDirection` literal or enum.

tests/dataset/test_relation_join.py
```diff
@@ -0,0 +1,537 @@
...
+        lambda r: r.where("order_id", "gt", 0),
+        lambda r: r.order_by("order_id"),
+        lambda r: r.limit(1),
+        lambda r: r.select("order_id").max(),
+    ],
+    ids=["select", "where", "order_by", "limit", "agg-max"],
+)
+def test_origin_table_name_preserved_across_transformations(
+    dataset_with_loads: TLoadsFixture,
+    transform: Any,
```
zilto: `Any` is the equivalent of no type annotations, so it's a bit of a "code smell". Here we can annotate `Callable[[dlt.Dataset], dlt.Relation]` which means "singe input of dataset type, outputs a relation".

tests/dataset/test_relation_join.py
```diff
@@ -0,0 +1,537 @@
...
+) -> None:
+    dataset, _, _ = dataset_with_loads
+    rel = build_rel(dataset)
+    target = _resolve_other(dataset, other)
+
+    with pytest.raises(ValueError, match=match):
+        rel.join(target)
+
+
+@pytest.mark.parametrize(
```
zilto, line 449: I really like this parameterization! It has named cases, it's easy to read what's supported, and adding new cases is simple.

dlt/dataset/relation.py
```diff
@@ -55,6 +62,164 @@
...
+    def on_pairs(self) -> list[tuple[str, str]]:
+        """(existing_side_col, new_side_col) pairs for the ON clause."""
+        if self.direction == "LEFT":
+            return list(zip(self.referenced_columns, self.columns))
+        return list(zip(self.columns, self.referenced_columns))
+
+    @property
+    def join_type(self) -> TJoinType:
+        """The SQL join type: 'left' or 'right'."""
+        return "left" if self.direction == "LEFT" else "right"
```
zilto: This approach totally makes sense, but it's less "stylistically aligned" with `dlt`. Dataclasses are heavier than `TypedDict` (which are just dicts at runtime) and have their own quirks. 

Adding methods on data objects is convenient but can reduce clarity: is the object responsible to define its utilities and transformations? or are we expected to apply functions on it to process it? `TypedDict` avoid this ambiguity because you can't extend it with methods

Here, `_JoinRef` could be a typed dict and properties could be utility functions

zilto (commented):
Great job! I left detailed comments and already made light changes. I tried to leave a clean commit-by-commit history.

This looks ready for release. If we have more sophisticated needs, we should implement a full-on graph interface (it will be useful in other places in the code base) instead of adding logic directly on `dlt.Relation`

zilto:
@rudolfix I extensively reviewed the PR and the joining logic supports many cases. One missing part is determining the "column name aliasing" convention. What happens when you join tables `customers` with `customers__orders` and `id` column exists in both tables?

This is an important contract with end-users if they start writing queries with it

### Ideas / notes
- always prefix columns from the joined table, e.g., `customers__orders__id`, `customers__orders__ordered_at`, etc. This gives the assumptions: 
    ```python
    table = dataset.table("customers")
    other_table_name = "customers__orders"
    joined = table.join(other_table)
    
    # original columns are preserved
    assert all(col in joined.columns for col in table.columns)
    # joined columns are aliased
    assert all(col not in joined.columns for col in dataset.table(other_table_name).columns)
    ```
- note that "other" could be the parent or child table depending on how the join is defined
    ```python
    parent_origin = dataset.table("customers").join("customers__orders")
    assert "id" in parent_origin.columns  # customers.id
    assert "customers__orders__id" in parent_origin.columns
    
    child_origin = dataset.table("customers__orders").join("customers")
    assert "id" in child_origin.columns  # customers__orders.id
    assert "customers__id" in child_origin.columns
    ```
- this behavior is deterministic and reliable for users; not too magical
- for more complex cases, people should use ibis, SQL or dataframe library
- we could add a `alias` (or `as` or `prefix`) keyword to `join` to avoid extremely long names e.g., 
    ```python
    table = dataset.table("customers").join("customers__orders__items", alias="items") 
    # columns will be prefixed with `items__` instead of `customers__orders__items__`
    ```
- alternative: use the number suffix approach currently implemented unless an `alias` keyword is passed explicitly

### Future work
if we have two origin tables that have valid join conditions, we should be able to pass a `dlt.Relation` instance of a table name as string. 

This allows to filter rows and columns before joining, which is simpler than doing it post-join
```python
table = dataset.table("customers")
# apply simple transformation on an origin table
other_table =  (
   dataset.table("customers__orders__items")
   .where("created_at", "gt", "2025-01-01")
   .select("created_at", "id", "price")
)
# resolve join based on the origin table; join columns need to exist
joined = table.join(other_table) 
```
This might be out-of-scope. Maybe people should use ibis instead. Let's see if people use `.join()`, want it and vote on the issue

dlt/dataset/relation.py
```diff
@@ -96,6 +107,17 @@ def __init__(
         self._table_name = table_name
         self._execute_raw_query: bool = _execute_raw_query
 
+        # Track the original base table for chained join validation
+        self._origin_table_name: Optional[str] = table_name
+        # NOTE for future complex graph and lineage resolution, this information should be tracked
+        # in a centralized object instead of on directly on the `dlt.Relation`
+        # necessary to allow for chained joins while keeping correct cardinality
+        self._joined_table_aliases: Optional[dict[str, str]] = (
```
rudolfix: `_joined_table_aliases`, `_next_join_alias_index`, `_is_joinable_graph` are leaking from `_discover_join_params` - you should be able to get this information from the sqlglot expression on relation

dlt/dataset/relation.py
```diff
@@ -96,6 +107,17 @@ def __init__(
         self._table_name = table_name
         self._execute_raw_query: bool = _execute_raw_query
 
+        # Track the original base table for chained join validation
+        self._origin_table_name: Optional[str] = table_name
```
rudolfix: I do not think we need this. IMO the base class has a bug in `__copy__` that does not preserve `_table_name`. I'd try to fix it and then remove this prop

dlt/dataset/relation.py
```diff
@@ -607,7 +649,71 @@ def __repr__(self) -> str:
         return simple_repr("dlt.Relation", **without_none(kwargs))
 
     def __copy__(self) -> Self:
```
rudolfix, line 706: IMO if we cleanup the code we won't need this

dlt/dataset/relation.py
```diff
@@ -607,7 +649,71 @@ def __repr__(self) -> str:
...
+the logic to utility functions that return `_TRelation`.
+Returning the concrete type `Relation` would fail type checking.
+"""
+
+
+@dataclass(frozen=True)
+class _JoinRef:
+    """A resolved table reference annotated with join direction."""
+
+    table: str
```
rudolfix: we already have `TTableReference`? why not to use it here? you can embed it in this dataclass as a field or convert it to TypedDict (but you'll lose your fancy helper methods :/)

dlt/dataset/relation.py
```diff
@@ -607,7 +649,71 @@ def __repr__(self) -> str:
...
+
+    @property
+    def on_pairs(self) -> list[tuple[str, str]]:
+        """(existing_side_col, new_side_col) pairs for the ON clause."""
+        if self.direction == "left":
+            return list(zip(self.referenced_columns, self.columns))
+        return list(zip(self.columns, self.referenced_columns))
+
+    @property
+    def join_type(self) -> TJoinType:
```
rudolfix: is the value returned ever different from `direction`?

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+        query=query,
+        joins=joins,
+        joined_table_aliases=alias_map,
+        next_alias_index=start_index if pending else next_index,
+        origin_table_name=left._origin_table_name,
+        is_joinable_graph=True,
+    )
+
+
+def _get_or_create_join_base_query(relation: Relation) -> sge.Select:
```
rudolfix: this should not be needed. and IMO will cause problems in the code that uses it:
- when you join leave "this" (left) table intact
- do not alias, do not expand the select list. SELECT list and FROM must be identical to what they were before applying join
- just use the query that is already on relation (relation.sqlglot_expression). it will build a right SELECT for you if needed

rudolfix: in joins refer to "this" (the base table that you join to) just by its base name. if that was `users` do `users.id` . alias all the other tables

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+        raise ValueError("This relation has been transformed and is not a join-graph relation.")
+
+    other_table = _resolve_magic_join_target(left, right)
+    # refs is the entire reference chain from origin to target(`other`) table
+    refs = _resolve_reference_chain(left._dataset.schema, left._origin_table_name, other_table)
+
+    alias_map = (
+        left._joined_table_aliases.copy()
+        if left._joined_table_aliases
+        else {left._origin_table_name: "t0"}
```
rudolfix: do not alias `left`

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+    if not left._origin_table_name:
+        raise ValueError("This relation has no base table to resolve references.")
+    if not left._is_joinable_graph:
+        raise ValueError("This relation has been transformed and is not a join-graph relation.")
+
+    other_table = _resolve_magic_join_target(left, right)
+    # refs is the entire reference chain from origin to target(`other`) table
+    refs = _resolve_reference_chain(left._dataset.schema, left._origin_table_name, other_table)
+
+    alias_map = (
```
rudolfix: instead of using a field on relation compute aliases from the existing query
```
query = left.sqlglot_expression.copy()
alias_map = _extract_table_aliases(query)
```
```
def _extract_table_aliases(expr: sge.Select) -> dict[str, str]:
      """Extract {logical_table_name: alias_or_name} from FROM and JOIN clauses."""
      tables: dict[str, str] = {}

      from_clause = expr.find(sge.From)
      if from_clause:
          table = from_clause.find(sge.Table)
          if table and isinstance(table.this, sge.Identifier):
              name = table.this.name
              alias_node = table.args.get("alias")
              tables[name] = alias_node.this.name if alias_node else name

      for join in expr.args.get("joins", []):
          table = join.find(sge.Table)
          if table and isinstance(table.this, sge.Identifier):
              name = table.this.name
              alias_node = table.args.get("alias")
              tables[name] = alias_node.this.name if alias_node else name

      return tables
```

dlt/dataset/relation.py
rudolfix: also * changes select everything from left and all right tables. this will generate clashes on columns ie. after qualify
```
SELECT "users"."_dlt_id", "users"."name", t1."_dlt_id", t1."order_id"
```
and our `compute_columns_schema` has a bug:
```
dlt_table_schema[col.output_name] = {
      "name": col.output_name,                                                                                                                                                                                                                     ...
  }  
``` will just overwrite stuff. we have options
1. write a test that confirms the problem - fix later
2. actually fix `compute_columns_schema` to auto alias columns with duplicates
3. remove this function: `build_select_expr` does not use *

File: dlt/dataset/relation.py [hunk 21]
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+        left._joined_table_aliases.copy()
+        if left._joined_table_aliases
+        else {left._origin_table_name: "t0"}
+    )
+    next_index = (
```
rudolfix: why can't you just use `len(alias_map)`? you just need this to generate next alias so counting tables and joins is enough. also if you remove `_get_or_create_join_base_query` things will get even faster.

btw. I propose to change aliasing scheme. to counting to get "t0" is not enough

dlt/dataset/relation.py
```diff
@@ -607,7 +649,71 @@ def __repr__(self) -> str:
...
+
+
+@dataclass
+class _JoinPlan:
+    query: sge.Select
+    joins: list[_JoinParams]
+    joined_table_aliases: dict[str, str]
+    next_alias_index: int
+    origin_table_name: str
+    is_joinable_graph: bool
```
rudolfix: this suggest that you can have JoinPlan on something that is not joinable? and indeed this is always True... so maybe remove?

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+
+def _discover_join_params(left: Relation, right: Union[str, Relation]) -> _JoinPlan:
+    """Discover join plan from schema reference chain.
+
+    Validates preconditions, resolves the reference chain, filters out
+    already-joined tables, and builds join parameters for the remaining refs.
+    """
+    if not left._origin_table_name:
+        raise ValueError("This relation has no base table to resolve references.")
+    if not left._is_joinable_graph:
```
rudolfix: I do not think we need it - if you can find a Table in a relation.sqlglot_query then it is joinable... so IMO you can remove _is_joinable_graph from relation

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+    assert relation._origin_table_name is not None
+    return sge.Select(expressions=[sge.Star()]).from_(
+        sge.Table(
+            this=sge.to_identifier(relation._origin_table_name, quoted=True),
+            alias=sge.TableAlias(this=sge.to_identifier("t0", quoted=False)),
+        )
+    )
+
+
+def _apply_join(relation: _TRelation, plan: _JoinPlan) -> _TRelation:
```
rudolfix: why this is a separate function? it asks to be _apply_join(self) on Relation! or I'm missing something?

dlt/dataset/relation.py
```diff
@@ -96,6 +107,17 @@ def __init__(
         self._table_name = table_name
         self._execute_raw_query: bool = _execute_raw_query
 
+        # Track the original base table for chained join validation
+        self._origin_table_name: Optional[str] = table_name
+        # NOTE for future complex graph and lineage resolution, this information should be tracked
+        # in a centralized object instead of on directly on the `dlt.Relation`
+        # necessary to allow for chained joins while keeping correct cardinality
+        self._joined_table_aliases: Optional[dict[str, str]] = (
+            {table_name: "t0"} if table_name else None
```
rudolfix: I commented on this already. but the "this" / "left" table query should stay intact - no aliasing, no changes to the SELECT

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
     rel = relation.__copy__()
     rel._sqlglot_expression = query
     return rel
+
+
+def _to_join_ref(ref: TTableReference, direction: TJoinDirection) -> _JoinRef:
+    if "table" not in ref or ref["table"] is None or "referenced_table" not in ref:
```
rudolfix: we have a really good validator for `TTableReference` in dlthub repo... let's bring it here in another ticket!

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+    """
+    upward_chain_from_left = schema_utils.get_all_parent_references_to_root(schema.tables, left)
+    upward_chain_from_right = schema_utils.get_all_parent_references_to_root(schema.tables, right)
+
+    for idx, left_ref in enumerate(upward_chain_from_left):
+        if "referenced_table" not in left_ref or "table" not in left_ref:
+            break
+        if left_ref["referenced_table"] == right:
+            # right is a parent of left: natural direction (for references), use RIGHT JOIN
+            return [_to_join_ref(ref, "right") for ref in upward_chain_from_left[: idx + 1]]
```
rudolfix: why `RIGHT`? if you join orders with users you will get a lot of NULLs for users that do not have orders. this is not intuitive behavior - typically you do INNER join so you'll get orders with users attached.

my take would be to propagate direction from "def join()` or use INNER

dlt/dataset/relation.py
```diff
@@ -350,6 +377,21 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
         rel._sqlglot_expression = rel.sqlglot_expression.order_by(order_expr)
         return rel
 
+    def join(
+        self,
+        other: str | Self,
```
rudolfix, line 363: join direction should be passed here, by default it is INNER join - as usual for databases. see my other comments

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+            # right is a parent of left: natural direction (for references), use RIGHT JOIN
+            return [_to_join_ref(ref, "right") for ref in upward_chain_from_left[: idx + 1]]
+
+    for idx, right_ref in enumerate(upward_chain_from_right):
+        if "referenced_table" not in right_ref or "table" not in right_ref:
+            break
+        if right_ref["referenced_table"] == left:
+            # left is a parent of right: reverse chain, use LEFT JOIN
+            return [
+                _to_join_ref(ref, "left") for ref in reversed(upward_chain_from_right[: idx + 1])
```
rudolfix: similarly to above: you'll get parents without orders - but default direction in SQL is INNER. not intuitive

dlt/dataset/relation.py
```diff
@@ -778,3 +884,244 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+    Returns:
+        List of _JoinRef where directions is:
+        - "RIGHT" when joining from child to parent (natural ref direction)
+        - "LEFT" when joining from parent to child
+    """
+    if left == right:
+        raise ValueError(f"Cannot join a table to itself: {left}")
+    # Check direct references first
+    for ref in schema.references:
+        if ref.get("table") == left and ref.get("referenced_table") == right:
```
rudolfix: same thing as in `_resolve_parent_reference_chain`

dlt/dataset/relation.py
zilto: My motivation was that writing `dataset.table("customers").join("customers__orders")`  shouldn't drop rows from `"customers"`. 

Here, `right`  means that `dataset.table("customers__orders").join("customers")`  does the same operation as above. I'm not a big fan of this because it's harder to reason about. I think it was added after your previous comment / it was not part of initial scope. 

IMO, we should only allow to walk down the tree from a given node. e.g., doing `dataset.table("customers__orders").join("customers")`  would raise "customers is not a child of customers__orders". These restriction wouldn't apply for explicit TTableReference

Travior: I think after the refactor that includes manual join kind specification we can keep the functionality of joining from child to parent, now we just use the `kind` the user provides

dlt/dataset/relation.py
Travior: This was one of @zilto's feedback to reduce private methods on public objects and treat them as internal functions for now

Travior:
The main changes made:
- we parse most of the previous internal state from the sqlglot expression now, only exception is `_is_join_eligible`. Right now there is no support for joins on arbitrary relations (like once modified via aggregations or created from raw sql). I think trying to infer this from the sqlglot ast would be a bit tedious and run on a lot of heuristics (are certain keywords present, did the user unselect a column we need for a join etc.).
- we now only project columns the user explicitly joined. Columns from the left table are never aliased, while ones from the right are either prefixed by the table name or a user provided alias.
- we allow the user the specify the join kind (default inner). This eliminates a lot of the confusion around intermediate join kinds as we can now re-use the user provided one (e.g. when the user wants a left join, we shouldn't drop rows with an INNER join in the chain). There is an optimization opportunity here where we can collapse some intermediary joins to use `INNER` within a paranthesized join expression, but that makes future attachment of new joins messy

dlt/dataset/relation.py
```diff
@@ -607,7 +671,34 @@ def __repr__(self) -> str:
         return simple_repr("dlt.Relation", **without_none(kwargs))
 
     def __copy__(self) -> Self:
-        return self.__class__(dataset=self._dataset, query=self.sqlglot_expression)
+        rel = self.__class__(dataset=self._dataset, query=self.sqlglot_expression)
+        rel._table_name = self._table_name
+        rel._is_join_eligible = False
```
rudolfix: This prevents a lot of legit cases where your join works perfectly. I include a few tests to prove that. IMO sqlglot is able to push join into right place in your query. it works pretty well. and in cases it doe snot work your code does the right checks and returns meaningful exception (even if not intended)

tldr;> remove this flag... let's not outsmart out users

dlt/dataset/relation.py
```diff
@@ -607,7 +671,34 @@ def __repr__(self) -> str:
...
+        return rel
+
+
+_TRelation = TypeVar("_TRelation", bound=Relation)
```
rudolfix: OK now I get the intention behind this, but we are still just half way there. Right we eliminated all leaks from join construction into Relation and you are able to convert `_apply_join` into pure utility function that does not depend on Relation at all. see comments below. also Opus will do a refactor for you... you have good test coverage so that is safe

rudolfix: this will also allow you to move this code to `_join.py` module - no more circular deps on Relation if you wish. We'll have a nice code cleanup

dlt/dataset/relation.py
```diff
@@ -778,3 +869,384 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+                    this=sge.to_identifier(column_name, quoted=True),
+                ),
+                alias=sge.to_identifier(output_name, quoted=True),
+            )
+        )
+
+    query.set("expressions", [*normalized_left_expressions, *appended_target_columns])
+
+
+def _apply_join(
```
rudolfix: if you eliminate Relation and use this signature (or similar)
```py
def _apply_join(
    expression: sge.Query,
    *,
    schema: dlt.Schema,
    left_table: str,
    right_table: str,
    projection_prefix: str,
    kind: TJoinType = "inner",
) -> sge.Select:
```
so instead of taking Relation, you take expression from relation, schema, left and right table names and return the joined query

dlt/dataset/relation.py
```diff
@@ -778,3 +869,384 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+    _apply_join_projection(
+        relation,
+        query,
+        target_table=target_table,
+        target_qualifier=target_qualifier,
+        projection_prefix=projection_prefix,
+        allow_existing_target_projection=not join_params,
+    )
+
+    rel = relation.__copy__()
```
rudolfix: this should be in join() method - with the proposed refactor or not. all other methods do that

tests/dataset/conftest.py
```diff
@@ -18,3 +19,152 @@
 def module_tmp_path(tmp_path_factory: pytest.TempPathFactory) -> pathlib.Path:
     """Temporary directory that persist for the lifetime of test `.py` file."""
     return tmp_path_factory.mktemp("pytest-test_relation")
+
+
+USERS_DATA_0 = [
```
rudolfix: conftest is injected into every test module of tests/dataset. so typically you put here things that you want all other modules to use ie. auto fixtures or common fixtures. `USERS_DATA_0` seems to be used only to test joins. what I do in that case: I move such things to `utils.py` and import selectively when I need it. OFC our code is not completely clean but here IMO those test cases should not be automatically merged with other test modules

dlt/dataset/relation.py
```diff
@@ -607,7 +670,33 @@ def __repr__(self) -> str:
         return simple_repr("dlt.Relation", **without_none(kwargs))
 
     def __copy__(self) -> Self:
-        return self.__class__(dataset=self._dataset, query=self.sqlglot_expression)
+        rel = self.__class__(dataset=self._dataset, query=self.sqlglot_expression)
+        rel._table_name = self._table_name
```
burnash, line 708: This looks like a regression: I think preserving `_table_name` here breaks `to_ibis()` API - so `to_ibis()` now silently drops transforms:

```py
  if self._table_name:
      ibis_table = backend.table(self._table_name)
  else:
      ibis_table = backend.sql(...) 
```

quick check with the test data from test_relation_join():

```py
(Pdb) base = dataset.table("users")
(Pdb) filtered = base.where("name", "eq", "Alice")
(Pdb) base.to_ibis().count().execute()
3
(Pdb) filtered.to_ibis().count().execute()
3
```

Similar, in `from_loads()` now allows filtered relations silently.

dlt/dataset/relation.py
```diff
@@ -778,3 +867,378 @@ def _add_load_id_via_parent_key(relation: dlt.Relation) -> dlt.Relation:
...
+            if alias_index.isdigit():
+                next_index = max(next_index, int(alias_index) + 1)
+    return next_index
+
+
+def _resolve_magic_join_target(left: Relation, right: Union[str, Relation]) -> str:
+    """Resolve magic-join target table name from input."""
+    if isinstance(right, Relation):
+        # TODO: remove once we allow cross-dataset joins
+        if right._dataset is not left._dataset:
```
burnash: Looks like false-rejection: uses object identity so:

```
ds1 = pipeline.dataset()
ds2 = pipeline.dataset()
ds1 is ds2  # False
```

Travior:
@rudolfix made the structural changes you asked for. One note:
Right now you can (somewhat arbitrarily) construct a relation/query that produces non-valid sql after a join:
```python
rel = pipeline.dataset().table("customers").select("name").max() \
        .join("orders")
```
This produces the following sql:
```
SELECT MAX("name"),
       orders."order_id" AS "orders__order_id",
       orders."customer_id" AS "orders__customer_id",
       orders."order_date" AS "orders__order_date",
       orders."total_amount" AS "orders__total_amount",
       orders."_dlt_load_id" AS "orders___dlt_load_id",
       orders."_dlt_id" AS "orders___dlt_id"
FROM "customers"
INNER JOIN "orders" ON customers."customer_id" = orders."customer_id"
```

If we keep an internal variable (like `_is_join_eligible`), we could "disable" joins after aggregations and throw a nice error. We could also inspect the query for aggregations but I fear that we will start to add a lot of checks for edge cases if we extend the relation interface in the future (more operations).

---

Other notes:
- fixed the regression that @burnash noted: I think additionally checking `self._query` is fine in `to_ibis`. Technically there is a behavior change if someone manually initializes a Relation with `table_name` and `query` (now `query` will take priority instead of the table). Not sure how valid that case is but we could also guard with an XOR in `__init__` against that (although also technically breaking)
- added tests for user defined references in join generation

dlt/dataset/relation.py
Travior: yep, you're right, thanks for the catch 🙂

rudolfix:
@Travior IMO this is 100% ok - if someone joins on a column that got aggregated then the query is wrong. I know that this is "magic" join but here magic ends... this is really good now! let's merge and we continue further ie. I need magic joins for `_dlt_load_id` and `_dlt_loads` table :)

dlt/dataset/relation.py
burnash: I think Relation.from_loads still may have this issue with `rel._table_name = self._table_name` below in the __copy__(), correct me if I'm wrong.
e.g. dataset.table("orders").where("id", "eq", 101).from_loads(...)  will pass, when it should raise value error here.

burnash, line 610: Sorry I think I missed this last round, but it looks similar to other `__copy__` _table_name cases. Most likely we need the same guard for query here as well. Before the table name in `__copy__` preservation it should have probably errored on None -> keyerror. Now it'll go into `_add_load_id_via_root_key` and discard prior transformation. E.g. `ds.table("users").where("id", "eq", 1).with_load_id_col()`

dlt/dataset/relation.py
```diff
@@ -350,6 +358,86 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
...
+        self,
+        other: str | Self,
+        kind: TJoinType = "inner",
+        alias: Optional[str] = None,
+    ) -> Self:
+        """Join this relation to another table using dlt schema references.
+
+        Join conditions are discovered automatically from the schema's reference
+        chain (parent/child/root relationships created by dlt during loading).
+        Both the current relation and ``other`` must be base-table relations
```
burnash, line 372: > Both the current relation and ``other`` must be base-table relations

Should we enforce this in code? Since __copy__ now preserves _table_name it survive transforms.

tests/dataset/test_relation_join.py
```diff
@@ -0,0 +1,923 @@
...
+            id="query-relation-not-joinable",
+        ),
+        pytest.param(
+            lambda ds: ds.table("users__orders").limit(5).select("order_id"),
+            "users",
+            "no base table to resolve references",
+            id="subquery-hides-base-table",
+        ),
+    ],
+    ids=lambda v: v if isinstance(v, str) else None,
```
burnash: nitpick: the `pytest.param` already have explicit `id`, not sure lambda is needed here.

dlt/dataset/relation.py
burnash, line 610: Sorry I think I missed this last round, but it looks similar to other `__copy__` _table_name cases. Most likely we need the same guard for query here as well. Before the table name in `__copy__` preservation it should have probably errored on None -> keyerror. Now it'll go into `_add_load_id_via_root_key` and discard prior transformation. E.g. `ds.table("users").where("id", "eq", 1).with_load_id_col()`

burnash (changes requested):
Looks good, I have some comments: I think I missed another _table_name related issue (should be the last one), and two smaller things.

dlt/dataset/relation.py
```diff
@@ -350,6 +358,86 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
         rel._sqlglot_expression = rel.sqlglot_expression.order_by(order_expr)
         return rel
 
+    def join(
+        self,
+        other: str | Self,
+        kind: TJoinType = "inner",
```
burnash, line 365:
```suggestion
        *,
        kind: TJoinType = "inner",
```

@Travior let's align this with the extended `join` api from https://github.com/dlt-hub/dlt/issues/3747 and make `kind` and below keyword-only args

Travior: @burnash 👍 sounds good, will change this

PR #3593, opened 2026-01-31
Fix failing testsdue parallelism

tetelio:
After the parallel testing changes, some tests are flaky. This PR fixes a race condition test in clickhouse by adding a uniq_id() to a dataset name and the qdrant test by running qdrant exclusively serially.


tests/load/qdrant/test_pipeline.py
```diff
@@ -376,7 +376,7 @@ def test_qdrant_local_parallelism_disabled(preserve_environ) -> None:
     os.environ["DATA_WRITER__FILE_MAX_ITEMS"] = "20"
 
     with TemporaryDirectory() as tmpdir:
-        p = dlt.pipeline(destination=dlt.destinations.qdrant(path=tmpdir))
+        p = dlt.pipeline(destination=dlt.destinations.qdrant(qd_path=tmpdir))
```
tetelio, line 379: `path` argument is not used by [qdrant](https://dlthub.com/docs/dlt-ecosystem/destinations/qdrant#setup-guide) when passed outside of the credentials object. better using `qd_path` directly

tests/load/test_job_client.py
```diff
@@ -1143,7 +1143,7 @@ def _load_something(_client: SqlJobClientBase, expected_rows: int) -> None:
     ids=lambda x: x.name,
 )
 def test_schema_retrieval(destination_config: DestinationTestConfiguration) -> None:
-    p = destination_config.setup_pipeline("schema_test", dev_mode=True)
+    p = destination_config.setup_pipeline("schema_test" + uniq_id(), dev_mode=True)
```
tetelio, line 1146: This is a parametric test that, without `uniq_id()`, will have collisions between tests (same tests with parameters taken different values but same pipeline name)

tests/load/utils.py
```diff
@@ -554,12 +555,18 @@ def destinations_configs(
             ),
             DestinationTestConfiguration(
                 destination_type="qdrant",
-                credentials=dict(path=str(Path(FILE_BUCKET) / f"qdrant_data_{worker}")),
+                credentials=dict(path="qdrant_data"),
```
tetelio, line 557: @rudolfix
question: why did we even use a the FILE_BUCKET before?
i can put it back, but we should be fine with a relative path (tests pass)

tests/load/utils.py
```diff
@@ -554,12 +555,18 @@ def destinations_configs(
...
+                env_vars={
+                    "HF_TOKEN": hf_token,
```
tetelio: having a huggingface token (still to be added to the credentials in google secrets) allows for bigger amounts of calls to huggingface (qdrant client needs to get a [model](https://github.com/dlt-hub/dlt/blob/51946ed1bc052089114e11f1821b7e6c4e37eaf8/dlt/destinations/impl/qdrant/configuration.py#L108) from huggingface and without hf token the IP gets rate limited very easily -> which makes the tests fail)

tetelio: we dont need the token any more by running the tests serially

tests/load/utils.py
rudolfix: no idea! very old code!

rudolfix (approved):
LGTM!!

Issue #3554, opened 2026-01-19
Oracle Sqlalchemy destination varchar is created as clob


kullanici0606:
### dlt version

1.20.0

### Describe the problem

When I try to load data from Postgresql to Oracle using Sqlalchemy destination, all varchar fields are converted to CLOB type. When I try same operation with Sling Data, it correctly creates the varchar fields with lengths.

This is my code:
```
def load_psql_to_oracle() -> None:
    oracledb.init_oracle_client()

    source = sql_database(schema="public",
                          backend='pyarrow',
                          chunk_size=100000).with_resources("table1")
    pipeline = dlt.pipeline(
        pipeline_name="load_psql_to_oracle",
        destination="sqlalchemy",
        dataset_name="dest_schema",
        progress="log"
    )
    load_info = pipeline.run(source,
                             write_disposition="replace"
                             )
    print(load_info)
```

The table DDL for Postgrsql:
```sql
 create table table1(id int, name varchar(10), surname varchar(20));
```

The table created by DLT on Oracle:
```sql
CREATE TABLE DEST_SCHEMA.TABLE1
(
  ID              NUMBER(19),
  NAME            CLOB,
  SURNAME         CLOB,
  "_dlt_load_id"  CLOB                          NOT NULL,
  "_dlt_id"       VARCHAR2(128 CHAR)            NOT NULL
)
```

### Expected behavior

Column `name` should be `VARCHAR2(10)` and `surname` should be `VARCHAR2(20)`

### Steps to reproduce

1. Create a Postgreql table in Postgres: `create table table1(id int, name varchar(10), surname varchar(20));`
2. Insert dummy data: `insert into table1 values(5, "foo", "bar");`
3. Run the provided code above

### Operating system

Linux

### Runtime environment

Local

### Python version

3.10

### dlt data source

Postgresql (sql_database)

### Other deployment details

Destination: Oracle DB using sqlalchemy destination

sebfest:
Hi.

I faced the same problem. 

What you can do first is to set the reflection_level="full_with_precision" on your source/ressource. The SqlalchemyTypeMapper will convert text to to sa.Text() if it does not find precision for the column, which leads to all the CLOBs in your oracledb. 

If that does not work, you can still try to define your own typemapper:

```python

import sqlalchemy as sa
from sqlalchemy.dialects import oracle

from dlt.destinations.impl.sqlalchemy.type_mapper import SqlalchemyTypeMapper


[...]

Issue #2497, opened 2025-04-08
Allow to register SqlAlchemy custom dialect capabilities for SqlAlchemy destination


rudolfix:
**Background**
There are huge differences across the SqlAlchemy dialects and there are so many dialect that we must rely on our users to provide us info on how sql alchemy destination behaves for different backends. What we want to do:
- allow users to register custom functions that generate destination capabiltis based on sql alchemy dialect.
- allow users to easily plug their own `TypeMapper` by converting existing SqlAlchemy TypeMapper to visitor pattern where behavior for certian data types can be overriden.
- some dialect require custom order of columns (#2449 ) so we'll allow users to define TableAdapter callback which receives table definition before it is materialized (this is a new cap)
- custom database error handler so undefined relations can be detected (we hardcode those now) (this is a new cap)

The end goal is to add those dialect definitions to code base and let users to easily submit more

rudolfix:
solved by #3600

Issue #2451, opened 2025-03-28
Incompatible column type for _dlt_load_id in metadata tables causes errors during state loading


Smriti675:
### dlt version

dlt 1.8.1

### Describe the problem

When using DB2 as a destination with `dlt.destinations.sqlalchemy`, the internal _dlt_load_id column in the metadata tables is created as a CLOB. This leads to errors during joins and comparisons (e.g., SQL0420N) because DB2 cannot compare CLOBs in the query:

```
error I encountered
``` <class 'dlt.destinations.exceptions.DatabaseTransientException'>
(ibm_db_dbi.ProgrammingError) ibm_db_dbi::ProgrammingError: Binding Error: [IBM][CLI Driver][DB2/LINUXX8664] SQL0134N  Improper use of a string column, host variable, constant, or function "LOAD_ID".  SQLSTATE=42907 SQLCODE=-134
[SQL: SELECT testingestr."_dlt_pipeline_state"."version", testingestr."_dlt_pipeline_state".engine_version, testingestr."_dlt_pipeline_state".pipeline_name, testingestr."_dlt_pipeline_state".state, testingestr."_dlt_pipeline_state".created_at, testingestr."_dlt_pipeline_state".version_hash, testingestr."_dlt_pipeline_state"."_dlt_load_id" 
FROM testingestr."_dlt_pipeline_state" INNER JOIN testingestr."_dlt_loads" ON testingestr."_dlt_loads".load_id = testingestr."_dlt_pipeline_state"."_dlt_load_id" 
WHERE testingestr."_dlt_pipeline_state".pipeline_name = ? AND testingestr."_dlt_loads".status = ? ORDER BY testingestr."_dlt_loads".load_id DESC]
[parameters: ('db2_test_pipeline', 0)]
(Background on this error at: https://sqlalche.me/e/20/f405) ```


Is there a supported way to override the column type for _dlt_load_id in both _dlt_pipeline_state and _dlt_loads tables using something or way to fix this?

### Expected behavior

_No response_

### Steps to reproduce

- Set up a DB2 instance with the appropriate IBM DB driver.
- Run this minimal dlt script:

```
import dlt
import sqlalchemy

engine = sqlalchemy.create_engine("db2+ibm_db://db2inst1:my_password@127.0.0.1:50000/TESTDB")

pipeline = dlt.pipeline(
    pipeline_name="db2_test_pipeline",
    destination=dlt.destinations.sqlalchemy(engine),
[...]

Issue #2341, opened 2025-02-21
DLT fails to create required tables in target schema


JZ-poc:
### dlt version

1.7.0

### Describe the problem

DLT (version 1.7.0) is failing to create the required tables in the target schema, resulting in a `pyexasol.exceptions.ExaQueryError` exception. The error message indicates that the object `META."_dlt_pipeline_state"` is not found.

Here the trace: 
`root@cff28a6a366f:/app/test_csv_to_exasol# python test_pipeline.py 
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/dlt/pipeline/pipeline.py", line 786, in _sync_destination
    remote_state = self._restore_state_from_destination()
  File "/usr/local/lib/python3.10/site-packages/dlt/pipeline/pipeline.py", line 1541, in _restore_state_from_destination
    state = load_pipeline_state_from_destination(self.pipeline_name, job_client)
  File "/usr/local/lib/python3.10/site-packages/dlt/pipeline/state_sync.py", line 130, in load_pipeline_state_from_destination
    state = client.get_stored_state(pipeline_name)
  File "/usr/local/lib/python3.10/site-packages/dlt/destinations/impl/sqlalchemy/sqlalchemy_job_client.py", line 316, in get_stored_state
    with self.sql_client.execute_query(query) as cur:
  File "/usr/local/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.10/site-packages/dlt/destinations/impl/sqlalchemy/db_api_client.py", line 341, in execute_query
    yield SqlaDbApiCursor(self._current_connection.execute(query, *args))  # type: ignore[call-overload, abstract]
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1385, in execute
    return meth(self, multiparams, params, _EMPTY_EXECUTION_OPTS)
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/sql/elements.py", line 334, in _execute_on_connection
    return connection._execute_clauseelement(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1577, in _execute_clauseelement
    ret = self._execute_context(
  File "/usr/local/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1953, in _execute_context
    self._handle_dbapi_exception(
[...]

PR #3600, opened 2026-02-02
(feat) 2497 customize sqlalchemy destination with dialect capabilities

rudolfix:
### Description
implements #2497 

will allow to easily customize type mapping, adjusting SQLAlchemy table schemas before creation or any adjustment of destination capabilities. may help fixing:

#3554 
#2451 
#2341 
#991

### DIFF
--- a/dlt/destinations/impl/sqlalchemy/db_api_client.py
+++ b/dlt/destinations/impl/sqlalchemy/db_api_client.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import (
     Optional,
     Iterator,
@@ -97,7 +99,7 @@ class SqlalchemyClient(SqlClientBase[Connection]):
     ) -> None:
         super().__init__(credentials.database, dataset_name, staging_dataset_name, capabilities)
         self.credentials = credentials
-        self._dialect_caps: "DialectCapabilities" = capabilities.dialect_capabilities
+        self._dialect_caps: DialectCapabilities = capabilities.dialect_capabilities
         self._current_connection: Optional[Connection] = None
         self._current_transaction: Optional[SqlaTransactionWrapper] = None
         self.metadata = sa.MetaData()
@@ -423,18 +425,15 @@ class SqlalchemyClient(SqlClientBase[Connection]):
         return existing, missing_columns, reflected is not None
 
     def _make_database_exception(self, e: Exception) -> Exception:  # type: ignore[override]
+        # NOTE: override uses self (not static) because it needs _dialect_caps
         from dlt.destinations.impl.sqlalchemy.dialect import GENERIC_TERMINAL_PATTERNS
 
-        # NoSuchTableError is always an undefined relation
         if isinstance(e, sa.exc.NoSuchTableError):
             return DatabaseUndefinedRelation(e)
 
         if isinstance(e, (sa.exc.ProgrammingError, sa.exc.OperationalError)):
-            # delegate undefined-relation detection to dialect capabilities
-            undef = self._dialect_caps.is_undefined_relation(e)
-            if undef is True:
+            if self._dialect_caps.is_undefined_relation(e) is True:
                 return DatabaseUndefinedRelation(e)
-            # if the dialect didn't claim this as undefined, check generic terminal patterns
             msg = str(e).lower()
             for pat in GENERIC_TERMINAL_PATTERNS:
                 if pat in msg:
@@ -443,7 +442,6 @@ class SqlalchemyClient(SqlClientBase[Connection]):
         elif isinstance(e, sa.exc.IntegrityError):
             return DatabaseTerminalException(e)
         elif isinstance(e, sa.exc.DatabaseError):
-            # check dialect-specific undefined relation (e.g. Oracle ORA-00942 as DatabaseError)
             if self._dialect_caps.is_undefined_relation(e) is True:
                 return DatabaseUndefinedRelation(e)
             return DatabaseTransientException(e)
--- a/dlt/destinations/impl/sqlalchemy/dialect.py
+++ b/dlt/destinations/impl/sqlalchemy/dialect.py
@@ -12,8 +12,6 @@ from dlt.common.destination.capabilities import DataTypeMapper, DestinationCapab
 from dlt.common.destination.typing import PreparedTableSchema
 
 
-# generic undefined-relation patterns shared across dialects
-# matched against lowercased str(exception)
 _GENERIC_UNDEFINED_RELATION_PATTERNS = [
     # MySQL / MariaDB
     "unknown database",
@@ -37,15 +35,15 @@ _GENERIC_UNDEFINED_RELATION_PATTERNS = [
     # Exasol (broad)
     " not found",
 ]
+"""Generic undefined-relation patterns shared across dialects, matched against lowercased str(exception)."""
 
-# when an exception matches these but not the undefined-relation patterns above,
-# it is still terminal (e.g. wrong column name, syntax referring to non-existing object)
 GENERIC_TERMINAL_PATTERNS = [
     "no such",
     "not found",
     "not exist",
     "unknown",
 ]
+"""Terminal patterns for exceptions that don't match undefined-relation but are still non-transient."""
 
 
 class DialectCapabilities:
@@ -65,9 +63,6 @@ class DialectCapabilities:
     it in subclasses or add entries to SQLGLOT_DIALECTS for non-obvious mappings.
     """
 
-    # backend name → sqlglot dialect name
-    # only entries where the two names differ need to be listed; for all others
-    # the property falls back to the backend name itself
     SQLGLOT_DIALECTS: Dict[str, str] = {
         "postgresql": "postgres",
         "mssql": "tsql",
@@ -75,6 +70,7 @@ class DialectCapabilities:
         "awsathena": "athena",
         "teradatasql": "teradata",
     }
+    """Backend name to sqlglot dialect name. Only entries where the two differ; others fall back to the backend name."""
 
     def __init__(self, backend_name: str = "") -> None:
         self._backend_name = backend_name
@@ -85,7 +81,6 @@ class DialectCapabilities:
 
         Looks up SQLGLOT_DIALECTS first, falls back to the backend name itself.
         """
-        # use backend name as fallback if mapping not found
         return self.SQLGLOT_DIALECTS.get(self._backend_name, self._backend_name)
 
     def adjust_capabilities(
@@ -97,7 +92,6 @@ class DialectCapabilities:
 
         Called during adjust_capabilities on the factory. Modify caps in-place.
         """
-        pass
 
     def type_mapper_class(self) -> Type[DataTypeMapper]:
         """Return the type mapper class for this dialect"""
@@ -133,8 +127,6 @@ class DialectCapabilities:
                 return True
         return None
 
-
-# registry
 
 DIALECT_CAPS_REGISTRY: Dict[str, Type[DialectCapabilities]] = {}
 """Maps dialect / backend name to the DialectCapabilities class that handles it."""
@@ -173,9 +165,6 @@ def get_dialect_capabilities(dialect_name: str) -> Optional[DialectCapabilities]
     if caps_cls is not None:
         return caps_cls(dialect_name)
     return None
-
-
-# built-in dialect capabilities
 
 
 class MysqlDialectCapabilities(DialectCapabilities):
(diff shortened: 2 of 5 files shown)


dlt/destinations/impl/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,261 @@
...
+    def adjust_capabilities(
+        self,
+        caps: DestinationCapabilitiesContext,
+        dialect: sa.engine.interfaces.Dialect,
+    ) -> None:
+        """Adjust destination capabilities for this dialect.
+
+        Called during adjust_capabilities on the factory. Modify caps in-place.
+        """
+        pass
```
zilto: `pass` is not required given docstring is present

dlt/destinations/impl/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,261 @@
...
+    def __init__(self, backend_name: str = "") -> None:
+        self._backend_name = backend_name
+
+    @property
+    def sqlglot_dialect(self) -> str:
+        """The sqlglot dialect name for this backend.
+
+        Looks up SQLGLOT_DIALECTS first, falls back to the backend name itself.
+        """
+        # use backend name as fallback if mapping not found
```
zilto: minimally: remove comment
optionally: docstring

dlt/destinations/impl/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,261 @@
...
+      (e.g. reorder columns for StarRocks)
+    * is_undefined_relation -- detect "table/schema not found" errors for the dialect
+
+    The sqlglot_dialect property maps backend names to sqlglot dialect names. Override
+    it in subclasses or add entries to SQLGLOT_DIALECTS for non-obvious mappings.
+    """
+
+    # backend name → sqlglot dialect name
+    # only entries where the two names differ need to be listed; for all others
+    # the property falls back to the backend name itself
```
zilto: remove comment or move to docstring

dlt/destinations/impl/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,261 @@
...
+from typing import Any, Dict, Optional, Type
+
+import sqlalchemy as sa
+
+from dlt.common.destination.capabilities import DataTypeMapper, DestinationCapabilitiesContext
+from dlt.common.destination.typing import PreparedTableSchema
+
+
+# generic undefined-relation patterns shared across dialects
+# matched against lowercased str(exception)
```
zilto: remove or write as top-level docstring under the constant (i.e., match Python convention)

dlt/destinations/impl/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,261 @@
...
+    "is an undefined name",
+    # Apache Hive
+    "table not found",
+    "database does not exist",
+    # Exasol (broad)
+    " not found",
+]
+
+# when an exception matches these but not the undefined-relation patterns above,
+# it is still terminal (e.g. wrong column name, syntax referring to non-existing object)
```
zilto: remove comment or move as docstring under constant

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -81,6 +97,7 @@ def __init__(
     ) -> None:
         super().__init__(credentials.database, dataset_name, staging_dataset_name, capabilities)
         self.credentials = credentials
+        self._dialect_caps: "DialectCapabilities" = capabilities.dialect_capabilities
```
zilto: Replace quoted `"DialectCapabilities"` with `DialectCapabilities` (already in `TYPE_CHECKING` block). Add `from __future__ import annotations` at the beginning of the file

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -405,58 +422,30 @@ def compare_storage_table(self, table_name: str) -> Tuple[sa.Table, List[sa.Colu
             missing_columns = [c for c in all_columns if c.name not in reflected.columns]
         return existing, missing_columns, reflected is not None
 
-    @staticmethod
-    def _make_database_exception(e: Exception) -> Exception:
+    def _make_database_exception(self, e: Exception) -> Exception:  # type: ignore[override]
+        from dlt.destinations.impl.sqlalchemy.dialect import GENERIC_TERMINAL_PATTERNS
+
+        # NoSuchTableError is always an undefined relation
```
zilto: remove comment

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -405,58 +422,30 @@ def compare_storage_table(self, table_name: str) -> Tuple[sa.Table, List[sa.Colu
             missing_columns = [c for c in all_columns if c.name not in reflected.columns]
         return existing, missing_columns, reflected is not None
 
-    @staticmethod
-    def _make_database_exception(e: Exception) -> Exception:
+    def _make_database_exception(self, e: Exception) -> Exception:  # type: ignore[override]
```
zilto, line 427: why is there a `type: ignore` ?

dlt/destinations/impl/sqlalchemy/db_api_client.py
```diff
@@ -405,58 +422,30 @@ def compare_storage_table(self, table_name: str) -> Tuple[sa.Table, List[sa.Colu
...
-                "not found",
-                "not exist",
-                "unknown",
-            ]
-            for pat_ in terminal_patterns:
-                if pat_ in msg:
+            # delegate undefined-relation detection to dialect capabilities
+            undef = self._dialect_caps.is_undefined_relation(e)
+            if undef is True:
+                return DatabaseUndefinedRelation(e)
```
zilto: Reading the diff, `under` seems to be an unused variable. Can write

```python
if self._dialect_caps.is_undefined_relation(e):
  return DatabaseUndefinedRelation(e)
```

OTOH, the same function call is on line 446. Could reuse `undef` there

dlt/destinations/impl/sqlalchemy/dialect.py
```diff
@@ -0,0 +1,261 @@
...
+        table: sa.Table,
+        table_schema: PreparedTableSchema,
+    ) -> sa.Table:
+        """Modify an sa.Table object before it is created or used for loading.
+
+        Return the (possibly modified) table. The default implementation is a no-op.
+        """
+        return table
+
+    def is_undefined_relation(self, e: Exception) -> Optional[bool]:
```
zilto, line 113: Note. Relying on the exception messages adds dependencies to both the DB backend version and the DB Python SDK. Typically, exception messages are "weak public API" and prone to change.

Could be added either as a comment, note in the docstring, or near the tests saying "tests may break if X changes; implementation and tests will need to be updated"

There's also a risk for false positive where an unrelated exception message matches one of the patterns

dlt/destinations/impl/sqlalchemy/factory.py
```diff
@@ -75,76 +65,35 @@ def adjust_capabilities(
...
-                caps.sqlglot_dialect = "postgres"
-            elif backend_name == "awsathena":
-                caps.sqlglot_dialect = "athena"
-            elif backend_name == "mssql":
-                caps.sqlglot_dialect = "tsql"
-                caps.type_mapper = MssqlVariantTypeMapper
-            elif backend_name == "teradatasql":
-                caps.sqlglot_dialect = "teradata"
+            # look up registered dialect capabilities
+            from dlt.destinations.impl.sqlalchemy.dialect import (
```
zilto: why the Python import here?

I'm guessing that the module-level registry in `dlt/destinations/impl/sqlalchemy/dialect.py` could be causing import cycles and this inner import is required to avoid it. If it's the case, maybe we want to fix the registry approach instead

dlt/destinations/impl/sqlalchemy/factory.py
```diff
@@ -75,76 +65,35 @@ def adjust_capabilities(
...
-                UnsupportedTypeMapper as MysqlVariantTypeMapper,
-                UnsupportedTypeMapper as TrinoVariantTypeMapper,
-            )
-
-        dialect = config.get_dialect()
-        if dialect is not None:
+        dialect_type = config.get_dialect()
+        if dialect_type is not None:
+            # instantiate dialect to access properties and pass to capabilities
+            dialect = dialect_type()
```
zilto: `dialect_type()` is an ambiguous name IMO. I understand that it's a `type` object (i.e., Python `class`). 

What's probably misleading is that `SqlalchemyClientConfiguration.get_dialect()` doesn't return a `dialect`, it returns a `dialect_type`. Consider renaming the method to `get_dialect_class()`

dlt/destinations/impl/sqlalchemy/factory.py
```diff
@@ -75,76 +65,35 @@ def adjust_capabilities(
...
-            from dlt.common.destination.capabilities import (
-                UnsupportedTypeMapper as MssqlVariantTypeMapper,
-                UnsupportedTypeMapper as MysqlVariantTypeMapper,
-                UnsupportedTypeMapper as TrinoVariantTypeMapper,
-            )
-
-        dialect = config.get_dialect()
-        if dialect is not None:
+        dialect_type = config.get_dialect()
+        if dialect_type is not None:
```
zilto: Consider using a guard clause? Most of this method seems stuck under the `if dialect_type is not None`

```python
dialect_class = config.get_dialect()
if dialect_class is None:
  return super(sqlalchemy, cls).adjust_capabilities(caps, config, naming)

dialect = dialect_class()
# ... the rest
return super(sqlalchemy, cls).adjust_capabilities(caps, config, naming)
```

tests/load/sqlalchemy/test_sqlalchemy_dialect.py
```diff
@@ -0,0 +1,711 @@
...
+
+    # ORA-00942 detection
+    assert dc.is_undefined_relation(Exception("ORA-00942: table or view does not exist")) is True
+    # generic patterns also match through super()
+    assert dc.is_undefined_relation(Exception("relation does not exist")) is True
+    # unrelated error returns None
+    assert dc.is_undefined_relation(Exception("syntax error")) is None
+
+
+# -- sqlglot_dialect mapping --
```
zilto: remove all of these section header comments

tests/load/sqlalchemy/test_sqlalchemy_dialect.py
```diff
@@ -0,0 +1,711 @@
...
+    "destination_config",
+    destinations_configs(default_sql_configs=True, subset=["sqlalchemy"]),
+    ids=lambda x: x.name,
+)
+def test_type_mapper_param_overrides_dialect_caps(
+    destination_config: DestinationTestConfiguration,
+) -> None:
+    """Backward compat: passing type_mapper= directly still overrides dialect capabilities."""
+    from dlt.common import json
+    from dlt.destinations.impl.sqlalchemy.configuration import SqlalchemyCredentials
```
zilto: why are there nested import in a test?

tests/load/sqlalchemy/test_sqlalchemy_dialect.py
```diff
@@ -0,0 +1,711 @@
...
+        def process_result_value(self, value, dialect):
+            return json.loads(value) if value is not None else None
+
+    class DirectMapper(SqlalchemyTypeMapper):
+        def _db_type_from_json_type(self, column, table=None):
+            return CustomJsonStr(length=256)
+
+    # pass type_mapper directly -- this should override whatever dialect capabilities set
+    dest_ = dlt.destinations.sqlalchemy(
+        credentials=destination_config.credentials, type_mapper=DirectMapper  # type: ignore[arg-type]
```
zilto, line 605: why type:ignore

dlt/destinations/impl/sqlalchemy/type_mapper.py
```diff
@@ -72,41 +81,94 @@ def _to_db_decimal_type(self, column: TColumnSchema) -> sa.types.TypeEngine:
...
-            return self._to_db_decimal_type(column)
-        elif sc_t == "wei":
-            wei_precision, wei_scale = self.capabilities.wei_precision
-            return sa.Numeric(precision=wei_precision, scale=wei_scale)
-        elif sc_t == "date":
-            return sa.Date()
-        elif sc_t == "time":
-            return self._create_date_time_type(sc_t, precision, column.get("timezone"))
+        # dispatch to per-type visitor method
+        method = getattr(self, f"_db_type_from_{sc_t}_type", None)
```
zilto: This inheritance and dynamism makes it very hard to know what's available statically. Should minimally have a parameterized test that checks that all the `SqlalchemyTypeMapper` implementations have the required type conversions

rudolfix:
> _lots_ of code comments to delete and some docstring IMO. Didn't comment on every single one of them, but left some directions on different "cases" (e.g., comments that should be constant docstring, comments that should be method docstring, comments that serve as section headers, suspicious `type: ignore` instead of fixing the typing issue)

thanks for the review!! code style rules need improvement for sure. they are based on our CONTRIBUTING... I already amended them since this PR. btw. those comments that separate blocks of code are super annoying but somehow opus is hardwired to use them. I already had a rule against them. MAYBE I'LL WRITE IT IN CAPS :)

dlt/destinations/impl/sqlalchemy/type_mapper.py
rudolfix: the getattr dispatch is the standard Python visitor pattern. ast.NodeVisitor.visit() does exactly the same thing:                                                            
```py
  method = 'visit_' + node.__class__.__name__                                                                                                                                                                                                  visitor = getattr(self, method, self.generic_visit)
  return visitor(node)
```

you are right with the tests

rudolfix:
@zilto not sure you want to review this again? I fixed all suggestions and rewrote rules for commenting and importing Python modules.

PR #3608, opened 2026-02-05
feat: `clickhouse_cluster` destination

### LINKED ISSUE #2200: Cluster support for clickhouse
### Feature description

ALTER, CREATE, DROP queries in multiple nodes Clickhouse setup should include "on cluster <cluster_name>" clause to run query on all nodes in cluster.
Now [it](https://github.com/dlt-hub/dlt/blob/devel/dlt/destinations/impl/clickhouse/sql_client.py#L138) seems that it has not been implemented. 

### Are you a dlt user?

### Use case

Correct work with multiple nodes setup

### Proposed solution

on_cluster_mode and cluster_name added to clickhouse.credentials
all ddl parameterised with "on cluster"

### Related issues

https://github.com/dlt-hub/dlt/issues/1183

jorritsandbrink:
### Description
This PR adds a new destination `clickhouse_cluster` that:
- builds on top of `clickhouse` destination
- adapts generated SQL to work well in cluster setups
- extends functionality to support creation of additional distributed tables

### Related Issues

- Closes #2200 

### Additional Context
Related PR: https://github.com/dlt-hub/dlt/pull/2573

### DIFF
--- a/.github/workflows/test_destinations_remote.yml
+++ b/.github/workflows/test_destinations_remote.yml
@@ -245,6 +245,10 @@ jobs:
       - name: create secrets.toml
         run: pwd && echo "$DLT_SECRETS_TOML" > tests/.dlt/secrets.toml
 
+      - name: append clickhouse_cluster secrets to secrets.toml
+        run: awk '/^\[destination\.clickhouse_cluster\]/{p=1} /^\[/ && !/^\[destination\.clickhouse_cluster/ && p{exit} p' tests/.dlt/dev.secrets.toml >> tests/.dlt/secrets.toml
+        if: ${{ matrix.needs_clickhouse_cluster }}
+
       # NOTE: essential tests are always run
       - name: Run essential tests Linux
         run: make test-dest-remote-essential


tests/load/test_job_client.py
```diff
@@ -839,22 +853,16 @@ def test_write_dispositions(
                 # we overwrite with the same row. merge falls back to replace when no keys specified
                 assert len(db_rows) == 1
             else:
-                # NOTE: on second load, number of records in table "t" is zero in case of merge on
```
jorritsandbrink, line 842: This bug got solved by adding an `insert_deduplication_token` to `INSERT INTO` statements.

jorritsandbrink:
@rudolfix Could you you review this PR?

I struggled a lot trying to get consistent results, especially with the `delete-insert` merge strategy, because it relies on subqueries which ClickHouse doesn't seem to like when using replicated/distributed tables. Spent too much time debugging flaky tests, but it seems they're all fixed now.

Notes:
- when working with distributed tables, `SELECT` and `INSERT` operations should be done on the distributed table, while other operations (e.g. `TRUNCATE`, `DELETE FROM`) should be done on the base table
- when querying a table trough `pipeline.dataset()`, it will resolve to the distributed table (querying the base table would only return data from one shard)

PR #3613, opened 2026-02-05
feat: composable marimo widgets

zilto:
## Summary
This adds composable widgets to `dlt` built with `marimo`. A widget is **composable** because you can give it inputs and read its outputs. For `marimo < 0.19`, it wasn't possible to pass inputs.

This follows-up #3021 (read-only widgets) and #3080 (closed in favor of this one)

## Content
- instructions on how to create widgets
- updated previous read-only widgets (schema viewer and load package viewer)
- add simple pipeline selector widget
- add a dependency to `mowidgets`, which implements the widget logic. It's MIT-licensed library and hosted on GitHub. I created it on my personal time.

## Open question
I don't think creating a new `dlt[widgets]` group is the best idea. This should probably be in `dlt[workspace]`. Though, we would need to bump the `marimo` notebook significantly from 0.14 to 0.19. IMHO, we shouldn't support old dependencies in dev-tooling (e.g., drop old marimo versions). This is different than pipelines, sources, and destinations where backwards compatibility is essential

## Next steps
- Build more "single-purpose" widgets
- Rebuild the dlt Dashboard to use composable widgets. 
  - There is no immediate gain, but it will improve maintainability and ensure parity between "dashboard" and "interactive widgets" moving forward
- Add public documentation
- Add tests to individual widgets, especially those used in the dashboard. We can focus on critical paths.

---

**Continue Tasks:** ✅ 2 no changes — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fdlt-hub%2Fdlt%2Fpull%2F3613&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,8 @@ dependencies = [
     "orjson>=3.10.1 ; platform_python_implementation != 'PyPy' and sys_platform != 'emscripten'",
     "orjson>=3.11.0 ; python_version > '3.13'",
     "tenacity>=8.0.2",
-    "jsonpath-ng>=1.5.3",
+    # 1.8.0 wraps Child/Descendants str() in parens, breaking resolve_paths round-trip in drop.py
+    "jsonpath-ng>=1.5.3,<1.8",
     "fsspec>=2022.4.0",
     "packaging>=21.1",
     "pluggy>=1.3.0",
--- a/uv.lock
+++ b/uv.lock
@@ -2377,7 +2377,7 @@ requires-dist = [
     { name = "humanize", specifier = ">=4.4.0" },
     { name = "ibis-framework", marker = "python_full_version >= '3.10' and extra == 'lancedb'", specifier = ">=12.0.0" },
     { name = "ibis-framework", marker = "python_full_version >= '3.10' and extra == 'workspace'", specifier = ">=12.0.0" },
-    { name = "jsonpath-ng", specifier = ">=1.5.3" },
+    { name = "jsonpath-ng", specifier = ">=1.5.3,<1.8" },
     { name = "lancedb", marker = "extra == 'lancedb'", specifier = ">=0.22.0" },
     { name = "marimo", marker = "extra == 'workspace'", specifier = ">=0.14.5" },
     { name = "mcp", marker = "python_full_version >= '3.10' and extra == 'workspace'", specifier = ">=1.13.0" },


zilto:
I rebased and forced pushed. Last CI run, the dlt Dashboard end-to-end tests were the only failures. The same tests succeeded locally

dlt/helpers/marimo/__init__.py
```diff
@@ -1,8 +1,61 @@
-from dlt.helpers.marimo._widgets import render, load_package_viewer, schema_viewer
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import marimo
```
rudolfix: marimo is an optional dependency, please import it with out standard pattern

dlt/helpers/marimo/_pipeline_selector.py
```diff
@@ -0,0 +1,45 @@
...
+
+with app.setup:
+    import marimo as mo
+    from dlt.common.storages import FileStorage
+    from dlt.common.pipeline import get_dlt_pipelines_dir
+
+
+@app.cell
+def _():
+    _storage = FileStorage(get_dlt_pipelines_dir())
```
rudolfix: we already have this widget. `list_local_pipelines()` is shared with cli and workspace dashboard - produces ordered pipeline list also with "last run" info. you just need to reuse it all here

dlt/helpers/marimo/_pipeline_selector.py
```diff
@@ -0,0 +1,45 @@
...
+    except Exception:
+        _pipelines = []
+
+    pipelines_locations = {p: _storage.storage_path + "/" + p for p in sorted(_pipelines)}
+    return (pipelines_locations,)
+
+
+@app.cell
+def _(pipelines_locations):
+    _first_pipeline_name = next(iter(pipelines_locations.keys()))
```
rudolfix: this will crash on empty pipeline list

dlt/helpers/marimo/_pipeline_selector.py
```diff
@@ -0,0 +1,45 @@
...
+    )
+    pipeline_selector
+    return (pipeline_selector,)
+
+
+@app.cell
+def _(pipeline_selector, pipelines_locations):
+    pipeline_name = pipeline_selector.value
+    pipeline_path = pipelines_locations[pipeline_name]  # noqa: F841
+    return
```
rudolfix, line 38: what this cell is doing? do we need to return pipeline name + path here? to be used downstream?

rudolfix (changes requested):
This is great work. Except review comments:
1. docs/website/docs/general-usage/dataset-access/marimo.md already documents this feature so should be updated or removed
2. there's no test coverage - it should be trivial to LLM-add it based on the scope of the ticket
3. since we change the public interface, let's move widgets to `_workspace/helpers` module (this is workspace thing clearly)

There are possible inconsistecies:
 1. Bug: render() missing app argument for load_package_widget (__init__.py:21). The load_package_viewer branch calls load_package_widget(*args, **kwargs) without passing app, while the other two branches pass app. Either
  load_package_widget signature needs to drop app (line 38) or render() needs to pass it.                                                                                                                                                      2. Bug: Stale error message (__init__.py:27). Says "app must be either load_package_viewer or schema_viewer" but should include pipeline_selector.

pls ping me how you want to proceed. I have an implementation plan that will fix all of this they I've created during this review

dlt/helpers/marimo/_pipeline_selector.py
zilto: Assigned variables are not used in the notebook directly. They're meant to be available for widget users.

Including `pipeline_selector.value` forces the cell to run on value change of the UI component. On value change, update `pipeline_path`

rudolfix:
@zilto  regarding `widgets` extra - I'd remove it. we can place marimo + mowidgets into `workspace` indeed. tbh. this is still not 100% to me what we want to keep there. right now it is just convenience for the users.

zilto:
> @zilto regarding `widgets` extra - I'd remove it. we can place marimo + mowidgets into `workspace` indeed. tbh. this is still not 100% to me what we want to keep there. right now it is just convenience for the users.

Note that `mowidgets` requires `marimo >= 1.19.0`

zilto:
>since we change the public interface, let's move widgets to _workspace/helpers module (this is workspace thing clearly)

To me, it doesn't make sense because we want people to import these widgets in their code and `_workspace` is explicitly a private module

dlt/helpers/marimo/_pipeline_selector.py
zilto: `list_local_pipelines()` is prone to breaking because it wants to read traces, looks at file stats, etc. 

This implementation is simpler with what's strictly necessary

rudolfix (approved):
OK, let's go with the current location. I fixed a few linting errors and allowed marimo on Python 3.10. let's see if that works
LGTM now!

PR #3615, opened 2026-02-06
(fix) suppress spurious psutil warning during dlt init

rudolfix:
### Description
Suppress spurious "psutil dependency is not installed" warning printed to stdout during `dlt init` when importing pipeline template scripts.

---

**Continue Tasks:** ✅ 1 no changes — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fdlt-hub%2Fdlt%2Fpull%2F3615&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/tests/pipeline/test_dlt_versions.py
+++ b/tests/pipeline/test_dlt_versions.py
@@ -105,7 +105,7 @@ def test_pipeline_with_dlt_update(test_storage: FileStorage) -> None:
     ):
         # create virtual env with (0.3.0) before the current schema upgrade
         with Venv.create(
-            tempfile.mkdtemp(), ["dlt[duckdb]==0.3.0", "json-logging==1.4.1rc0"]
+            tempfile.mkdtemp(), ["dlt[duckdb]==0.3.0", "json-logging==1.4.1rc0", "setuptools<80"]
         ) as venv:
             # NOTE: we force a newer duckdb into the 0.3.0 dlt version to get compatible duckdb storage
             venv.install_deps(venv.context, ["duckdb" + "==" + pkg_version("duckdb")])
@@ -334,7 +334,7 @@ def test_filesystem_pipeline_with_dlt_update(test_storage: FileStorage) -> None:
 
     # execute in test storage
     with test_workspace():
-        with Venv.create(tempfile.mkdtemp(), ["dlt==0.4.9"]) as venv:
+        with Venv.create(tempfile.mkdtemp(), ["dlt==0.4.9", "setuptools<80"]) as venv:
             try:
                 print(venv.run_script("github_pipeline.py", "filesystem", "20"))
             except CalledProcessError as cpe:
@@ -407,7 +407,7 @@ def test_load_package_with_dlt_update(test_storage: FileStorage) -> None:
     with test_workspace({"DESTINATION__DUCKDB__CREDENTIALS": "duckdb:///test_github_3.duckdb"}):
         # create virtual env with (0.3.0) before the current schema upgrade
         with Venv.create(
-            tempfile.mkdtemp(), ["dlt[duckdb]==0.3.0", "json-logging==1.4.1rc0"]
+            tempfile.mkdtemp(), ["dlt[duckdb]==0.3.0", "json-logging==1.4.1rc0", "setuptools<80"]
         ) as venv:
             venv.install_deps(venv.context, ["duckdb" + "==" + pkg_version("duckdb")])
             # extract and normalize on old version but DO NOT LOAD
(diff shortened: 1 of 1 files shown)


dlt/_workspace/cli/_init_command.py
```diff
@@ -343,19 +345,28 @@ def init_pipeline_at_destination(
...
+        else:
+            # core sources are imported directly from the pipeline script
+            # which is in the _workspace module
+            import_pipeline_script(
+                os.path.dirname(source_configuration.src_pipeline_script),
+                os.path.basename(source_configuration.src_pipeline_script),
+                ignore_missing_imports=True,
+            )
+    finally:
+        sys.stdout = _saved_stdout
```
djudjuu: did you tell claude to try it out? for me it doesnt work!
here is tldr:
claude:

> he default value sys.stdout is evaluated once at class definition time and frozen. Every subsequent LogCollector() call without an explicit logger argument gets that frozen reference, regardless of what sys.stdout currently points to.

because the template does progress="log" it gets the default parameter which was evaluated and before your redirect and stored a reference to the original stdout.

only if we do assingn a new instance in the vibe-template code will it work:
```
    progress=LogCollector( dump_system_stats=True, logger=sys.stdout)
```

djudjuu: overall i dont really like the solution here.  what if we introduce other output in the future that will also get lost? (i mean warning and errors will still go through, unless they are written to stdout)

claude:
> The proper fix is at the source -- making LogCollector emit that psutil message through a mechanism that's designed to be suppressible (like warnings.warn()).

personally: I think the best is to just initiate LogCollector explicitly in the vibe-template. it exposes the paramters more directly (hints how to debug if things dont work). the progress="log" is nice for humans, but not that navigatable from a code perspective.

rudolfix: you (or claude) are right - i tested it with psutil still present so nothing got displayed. redirecting stdout is a normal procedure though and `LogCollector` has a pretty nasty bug by accepting `stdout` as default. I think we had users complaining that log output behaves weirdly...

I'll have to fix that

tests/pipeline/test_pipeline.py
```diff
@@ -1846,16 +1848,26 @@ def items2():
 
 
 def test_pipeline_log_progress() -> None:
+    from dlt.common.warnings import DltDeprecationWarning
```
djudjuu: imports to the top (is it in your rules/skills?)

tests/pipeline/test_pipeline.py
```diff
@@ -1864,6 +1876,29 @@ def test_pipeline_log_progress() -> None:
...
+        assert "redirected message" in buf.getvalue()
+
+    # explicit TextIO stream is used directly
+    with io.StringIO() as buf:
+        collector2 = LogCollector(logger=buf, dump_system_stats=False)
+        collector2._log(logging.WARNING, "stream message")
+        assert "stream message" in buf.getvalue()
+
+    # logger=None is deprecated and maps to "dlt_logger"
+    from dlt.common.warnings import DltDeprecationWarning
```
djudjuu: to the top

tests/pipeline/test_pipeline.py
```diff
@@ -1846,16 +1848,26 @@ def items2():
...
-        destination="dummy", progress=dlt.progress.log(0.5, logger=None, log_level=logging.WARNING)
+        destination="dummy",
+        progress=dlt.progress.log(0.5, logger="dlt_logger", log_level=logging.WARNING),
     )
-    # collector was created before pipeline so logger is not attached
-    assert cast(LogCollector, p.collector).logger is None
+    assert cast(LogCollector, p.collector).logger == "dlt_logger"
+    p.extract(many_delayed(2, 10))
+    # dlt logger attached after first extract
+    assert cast(LogCollector, p.collector).logger is not None
```
djudjuu: well. it was not None before too, so that's not really a test for anything.
either we check if it is no longer "dlt_logger" or whether its an instance of logging.Logger

```
    assert isinstance(cast(LogCollector, p.collector).logger, (logging.Logger, logging.LoggerAdapter))
```

dlt/_workspace/cli/_init_command.py
```diff
@@ -343,19 +341,25 @@ def init_pipeline_at_destination(
...
-        )
-    else:
-        # core sources are imported directly from the pipeline script which is in the _workspace module
-        import_pipeline_script(
-            os.path.dirname(source_configuration.src_pipeline_script),
-            os.path.basename(source_configuration.src_pipeline_script),
-            ignore_missing_imports=True,
-        )
+    # suppress warnings emitted during import (e.g. psutil missing from LogCollector)
+    with warnings.catch_warnings():
```
djudjuu, line 345: i like it. 
nitpick: technically the ignored warnings could be just around the != core import, but I guess all other warnings that could potentially come up (maybe related to dbmysql on linux when importing sql_database_pipeline.py) will then just pop up a bit later, which is still fine ux

djudjuu (changes requested):
just a few minor things (now, imports to the top) and one assertion in the test

PR #3617, opened 2026-02-08
docs: add realistic closure-based data masking example

### LINKED ISSUE #2116: Add a more realistic data masking example

veeceey:
## Summary

Replaces the hardcoded column name example in the pseudonymizing columns documentation with a more realistic closure-based approach that works with all backends supported by the `sql_database` source.

## Changes

- Updated `/docs/website/docs/general-usage/customising-pipelines/pseudonymizing_columns.md`
- Replaced the dummy source example with a real-world `mask_columns()` function that uses closures
- Added support for all sql_database backends (PyArrow, ConnectorX, Pandas, SQLAlchemy)
- Added two masking methods: `MASK` (replace with a mask string) and `NULLIFY` (replace with NULL)
- Included usage examples with `sql_table` source

## Details

The new example demonstrates:
- Using closures to capture parameters (columns, method, mask) rather than hardcoding values
- A more realistic function that can be reused across different resources
- Proper handling of different data formats based on the backend
- Both MASK and NULLIFY methods for different use cases

This addresses issue #2116 by providing a practical example that users can adapt for their own data masking needs.

Closes #2116

---

**Continue Tasks:** ✅ 1 no changes — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fdlt-hub%2Fdlt%2Fpull%2F3617&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/docs/examples/data_masking/data_masking.py
+++ b/docs/examples/data_masking/data_masking.py
@@ -1,0 +1,185 @@
+"""
+---
+title: Data masking for SQL database sources
+description: Learn how to mask sensitive columns using closures that work with all sql_database backends
+keywords: [data masking, pseudonymize, anonymize, sql_database, example]
+---
+
+This example shows how to build a reusable column-masking function for the
+`sql_database` source. The function uses a closure to capture masking
+configuration, so you can apply it to any resource via `add_map()`.
+
+Based on an implementation by [Michał Zawadzki](https://github.com/trymzet)
+and [Varun Chawla](https://github.com/veeceey).
+
+We'll learn how to:
+
+- Use a [closure](https://docs.python.org/3/faq/programming.html#why-am-i-getting-an-unboundlocalerror-when-the-variable-has-a-value) to create a configurable masking callback
+- Handle all backends supported by [sql_database](../dlt-ecosystem/verified-sources/sql_database/index.md): PyArrow, ConnectorX, Pandas, and SQLAlchemy
+- Support two masking strategies: replace with a mask string, or nullify
+
+## Usage with `sql_table`
+
+```py
+from dlt.sources.sql_database import sql_table
+
+table = sql_table(table="users")
+table.add_map(mask_columns(columns=["email", "ssn"]))
+
+pipeline = dlt.pipeline(
+    pipeline_name="masked_data",
+    destination="duckdb",
+    dataset_name="mydata",
+)
+load_info = pipeline.run(table)
+```
+
+"""
+
+from enum import Enum
+from typing import Any, Callable, Dict, List, Optional, Union
+
+import pyarrow as pa
+import pandas as pd
+
+
+class MaskingMethod(str, Enum):
+    MASK = "mask"
+    NULLIFY = "nullify"
+
+
+def mask_columns(
+    columns: List[str],
+    method: Optional[MaskingMethod] = None,
+    mask: str = "******",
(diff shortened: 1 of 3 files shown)


veeceey:
The new closure-based example is more realistic and production-ready because:

1. **Reusable**: The function accepts column names as parameters instead of hardcoding them, making it usable across different tables and resources
2. **Complete backend support**: Works with all backends supported by sql_database source (PyArrow, ConnectorX, Pandas, SQLAlchemy)
3. **Flexible masking**: Supports both MASK (replace with string) and NULLIFY (replace with NULL) methods
4. **Software engineering best practice**: Demonstrates closures, which is a concept many non-professional software engineers may be unfamiliar with but is essential for this use case

The example code is based on real-world usage and has been tested with sql_database and sql_table sources.

veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

PR #3626, opened 2026-02-10
optimizes relational normalizer and injectable context locking 

rudolfix:
### Description
This PR adds several optimizations to relational normalizer (mostly json data normalization and schema evolution)
Final benchmark (this branch vs. main)
- 5x on processing flat database like data (but you should use arrow for that anyway)
- ~2x on nested (github issues) REST API like data
- ~ 1.8x on nested very wide data (ethereum blocks)
optimizations were benchmarked and results were consistent

Additionally ISO timestamp parsing increased by 2-3x by removing timezone conversions. further reduction is possible if we remove pendulum

this branch also handles additional edge cases of locks on `Container` and injectable context

---

**Continue Tasks:** ✅ 1 no changes — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fdlt-hub%2Fdlt%2Fpull%2F3626&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/dlt/common/normalizers/json/__init__.py
+++ b/dlt/common/normalizers/json/__init__.py
@@ -2,9 +2,9 @@ import abc
 from typing import Any, Dict, Generic, Type, Generator, Protocol, Tuple, TYPE_CHECKING, TypeVar
 
 from dlt.common.typing import DictStrAny, TDataItem, StrAny
+from dlt.common.data_types.typing import TDataType
 
 if TYPE_CHECKING:
-    from dlt.common.data_types.typing import TDataType
     from dlt.common.schema import Schema
 else:
     Schema = Any
@@ -44,22 +44,22 @@ class DataItemNormalizer(abc.ABC, Generic[TNormalizerConfig]):
         pass
 
     @property
-    def py_type_to_sc_type_map(self) -> Dict[Type[Any], "TDataType"]:
+    def py_type_to_sc_type_map(self) -> Dict[Type[Any], TDataType]:
         """Lookup from Python type to schema type. Override in subclasses."""
         return {}
 
     @abc.abstractmethod
-    def py_type_to_sc_type(self, t: Type[Any]) -> "TDataType":
+    def py_type_to_sc_type(self, t: Type[Any]) -> TDataType:
         """Resolve Python type to schema type including subclass/Enum handling."""
         ...
 
     @abc.abstractmethod
-    def can_coerce_type(self, to_type: "TDataType", from_type: "TDataType") -> bool:
+    def can_coerce_type(self, to_type: TDataType, from_type: TDataType) -> bool:
         """Return True if `from_type` can be coerced to `to_type`."""
         ...
 
     @abc.abstractmethod
-    def coerce_type(self, to_type: "TDataType", from_type: "TDataType", value: Any) -> Any:
+    def coerce_type(self, to_type: TDataType, from_type: TDataType, value: Any) -> Any:
         """Coerce `value` from `from_type` to `to_type`. Raises ValueError on failure."""
         ...
 
--- a/tests/common/schema/test_schema_migrations.py
+++ b/tests/common/schema/test_schema_migrations.py
@@ -2,7 +2,7 @@ import os
 import pytest
 
 from dlt.common.schema.exceptions import SchemaEngineNoUpgradePathException
-from dlt.common.schema.migrations import migrate_schema
+from dlt.common.schema.migrations import migrate_schema, _remove_seen_null_first
 from dlt.common.schema.normalizers import DEFAULT_NAMING_MODULE
 from dlt.common.schema.schema import Schema
 from dlt.common.schema.typing import TStoredSchema
@@ -186,8 +186,6 @@ def test_row_and_parent_key_migration_upper_case() -> None:
 
 def test_seen_null_first_cleanup() -> None:
     """seen-null-first columns are cleaned up when schema is loaded."""
-    from dlt.common.schema.migrations import _remove_seen_null_first
-
     schema = Schema("test")
 
     # incomplete column with seen-null-first and no data_type -> should be removed


rudolfix:
## Scope

### Container locking
- `Container._LOCK` upgraded from `Lock` to `RLock`. Per-thread contexts restructured from `Dict[int, Dict[...]]` to `Dict[int, Tuple[Dict[...], RLock]]` — every thread gets its own lock at creation time.
- `__getitem__` uses double-checked locking for safe default context creation under contention.
- `injectable_context` always locks setup/teardown (previously no locking at all when `lock_context=False`). Parameter renamed to `lock_context_on_yield` to reflect new semantics — the lock is only held across yield when explicitly requested.
- Removed `main_context` alias and `_context_container_locks` lazy dict. Added `get_worker_contexts()` for passing worker-affinity contexts to process pools.
- New `worker_affinity: ClassVar[bool]` on `ContainerInjectableContext` — when `True`, the context is included in `get_worker_contexts()` and passed to process pool workers via `restore_run_context`. `ConfigSectionContext` sets it to `True` so pipeline name/sections are available in spawned workers.

### Replacing `seen-null-first` with side-channel tracking
- The old mechanism inserted incomplete columns with `x-normalizer: {seen-null-first: True}` into the schema during normalization, then required complex cleanup in `clean_x_normalizer` (table existence checks, long-name edge cases, lru_cache hacks).
- Replaced with `_null_only_columns: Dict[str, Set[str]]` on both `JsonLItemsNormalizer` and `ArrowItemsNormalizer`. Null-only column names are tracked outside the schema entirely. After normalization completes, `reconcile_null_only_columns()` checks which columns still have no type and generates appropriate warnings (`UnboundColumnWithoutTypeException` vs `UnboundColumnException`).
- Arrow path: `pa.null()` fields are dropped from inferred table schemas in `ArrowExtractor._compute_tables`. `remove_null_columns` now stores removed column names in arrow schema metadata (`dlt.null_columns` key), and `normalize_py_arrow_item` preserves this metadata through the schema rebuild. `ArrowItemsNormalizer._collect_null_columns_from_arrow_metadata` reads them back into the side-channel tracker.
- Schema migration: `_remove_seen_null_first()` runs on schema load to clean legacy hints — incomplete columns with the hint are removed, typed columns get the hint stripped. Tested with e2e upgrade from dlt 1.14.0.

### Relational normalizer optimizations
- `_flatten()` rewritten from recursive to iterative stack. Method lookups (`_normalize_identifier`, `_shorten_fragments`, `_is_nested_type`) cached as locals.
- `_link_row()` / `_extend_row()` inlined into call sites. `if extend:` guards skip no-op `.update()` calls.
- `_get_propagation_mapping` wrapped in `lru_cache`. `_get_propagated_values` returns `None` instead of empty dict when no propagations apply, `_normalize_row` skips propagation entirely when `propagation_config` is falsy.
- Propagation `extend` dict is now copied (`{**extend, **propagated}`) instead of mutated in-place — fixes a bug where table-level propagated values leaked across sibling rows in a list.
- `_clear_schema_caches()` invalidates all schema-dependent lru_caches (`_is_nested_type`, `_should_be_nested`, `_get_table_nesting_level`, `_get_primary_key`, `_get_propagation_mapping`) on `extend_table` / `remove_table`.

### Item normalizer optimizations (jsonl + arrow)
- `items_normalizers.py` split into `items_normalizers/` package (base, jsonl, arrow, model, file_import). `__init__.py` re-exports all public names.
- `JsonLItemsNormalizer._coerce_row` fast path: for columns already in the schema with a known `data_type`, if the Python type maps directly via `_type_map` the value is accepted with zero overhead. If coercion is needed and `can_coerce_type` confirms it, tries the fast coercion path before falling through to `_coerce_non_null_value`.
- Coercion interface delegated to data item normalizer (`_type_map`, `_can_coerce_type`, `_coerce_type`, `_py_type_to_sc_type`) instead of importing `coerce_value`/`py_type_to_sc_type` directly — enables pluggable coercion for the no-coercion variant.
- `PY_TYPE_TO_SC_TYPE` dict for O(1) type lookup. `_COERCE_DISPATCH` dict replaces the if/elif chain in `coerce_value()`.

### `relational_no_coercion` normalizer
- New module `dlt.common.normalizers.json.relational_no_coercion` — subclass of the relational normalizer that overrides `can_coerce_type` (returns `True` only for same-type) and `coerce_type` (raises `ValueError` on cross-type). Every type mismatch produces a variant column.
- Enabled by new abstract methods on `DataItemNormalizer` base class: `py_type_to_sc_type()`, `can_coerce_type()`, `coerce_type()` plus `py_type_to_sc_type_map` property.
- Configured via `json_normalizer = {module = "relational_no_coercion"}` in `config.toml`. Documented in `docs/website/docs/general-usage/schema.md`.

### Pendulum optimizations and fixes
- `ensure_pendulum_dt()` — fast paths for naive (no tz) and UTC datetimes, avoiding `_safe_timezone()` / `convert()` overhead entirely.
- `to_pendulum_tz()` — short-circuits for already-pendulum timezones and stdlib `timezone` fixed offsets (uses `pendulum.fixed_timezone()`), only falls back to `_safe_timezone()` for named timezones.
- `create_dt()` — three-way branch: naive → direct construct, fixed offset → direct construct without `convert()`, named tz → full `convert()`.
- `ensure_pendulum_datetime_non_utc()` — string inputs try `parse_iso8601()` first for a fast path before falling through to the full parser.
- `_text_to_date_fast()` — uses C-level `datetime.date.fromisoformat()` before falling back to pendulum.
- Bug fix in `ensure_pendulum_date()`: simplified to delegate to `ensure_pendulum_datetime_non_utc().in_tz(UTC).date()` — same behavior, less code.
- `parse_iso_like_datetime()` replaced `contextlib.suppress(ValueError)` with direct `try/except` (measurable context manager overhead on hot path).

PR #3627, opened 2026-02-12
(bugfix) (chores) cleans up query lifecycle

### LINKED ISSUE #3625: bug: query normalization doesn't handle `variant` data type
Using the utility `from dlt.destinations.queries import _normalize_query`, it will incorrectly normalize columns with variant types. 

For example
```sql
SELECT value__v_double FROM foo
```
will be normalized to
```sql
SELECT value_v_double FROM foo
```

rudolfix:
### Description
* fixes #3625 
* `normalize_query_identifiers` for (re) normalizing identifiers according to name convention ie. when your query contains raw idents
* renames, moves and documents functions that govern query lifecycle. a +- industry standard terminology is used

details on query lifecycle in the comment

---

**Continue Tasks:** ✅ 1 no changes — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fdlt-hub%2Fdlt%2Fpull%2F3627&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/dlt/dataset/relation.py
+++ b/dlt/dataset/relation.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 from collections.abc import Collection
 
+from functools import partial
 from typing import overload, Union, Any, Generator, Optional, Sequence, Type, TYPE_CHECKING
 from textwrap import indent
 from contextlib import contextmanager
@@ -248,8 +249,10 @@ class Relation(WithSqlClient):
             query = bind_query(
                 qualified_query=_qualified_query,
                 sqlglot_schema=self._dataset.sqlglot_schema,
-                expand_table_name=lambda name: self.sql_client.make_qualified_table_name_path(
-                    name, quote=False, casefold=False
+                expand_table_name=partial(
+                    self.sql_client.make_qualified_table_name_path,
+                    quote=False,
+                    casefold=False,
                 ),
                 casefold_identifier=self.sql_client.capabilities.casefold_identifier,
             )
--- a/dlt/destinations/queries.py
+++ b/dlt/destinations/queries.py
@@ -1,3 +1,4 @@
+from functools import partial
 from typing import Any, List
 
 import sqlglot.expressions as sge
@@ -22,8 +23,8 @@ def _normalize_query(
     return bind_query(
         qualified_query,
         sqlglot_schema,
-        expand_table_name=lambda name: sql_client.make_qualified_table_name_path(
-            name, quote=False, casefold=False
+        expand_table_name=partial(
+            sql_client.make_qualified_table_name_path, quote=False, casefold=False
         ),
         casefold_identifier=casefold_identifier,
     )
--- a/dlt/normalize/items_normalizers.py
+++ b/dlt/normalize/items_normalizers.py
@@ -1,5 +1,5 @@
 from copy import copy
-from typing import Callable, List, Dict, Sequence, Set, Any, Optional, Tuple
+from typing import List, Dict, Sequence, Set, Any, Optional, Tuple
 from abc import abstractmethod
 from functools import lru_cache
 
@@ -92,14 +92,10 @@ class ItemsNormalizer:
 
 
 class ModelItemsNormalizer(ItemsNormalizer):
-    @property
-    def _normalize_casefold(self) -> Callable[[str], str]:
-        from dlt.common.libs.sqlglot import create_outer_select_identifier_normalizer
-
-        return create_outer_select_identifier_normalizer(
-            self.schema.naming,
-            self.config.destination_capabilities.casefold_identifier,
-        )
+    def _normalize_casefold(self, identifier: str) -> str:
+        # use normalize_path() to preserve __ separators in already-normalized names
+        normalized = self.schema.naming.normalize_path(identifier)
+        return self.config.destination_capabilities.casefold_identifier(normalized)
 
     def _adjust_outer_select_with_dlt_columns(
         self,
--- a/tests/destinations/test_queries.py
+++ b/tests/destinations/test_queries.py
@@ -1,3 +1,4 @@
+from functools import partial
 from typing import cast
 
 import duckdb
@@ -132,8 +133,8 @@ ORDER BY i.id ASC
         normalized_query_expr = bind_query(
             qualified_query=cast(sge.Query, qualified_query_expr),
             sqlglot_schema=sqlglot_schema,
-            expand_table_name=lambda name: sql_client.make_qualified_table_name_path(
-                name, quote=False, casefold=False
+            expand_table_name=partial(
+                sql_client.make_qualified_table_name_path, quote=False, casefold=False
             ),
             casefold_identifier=sql_client.capabilities.casefold_identifier,
         )
--- a/tests/normalize/test_model_item_normalizer.py
+++ b/tests/normalize/test_model_item_normalizer.py
@@ -701,9 +701,7 @@ def test_variant_column_names_preserved(
     # Get the aliases from the outer SELECT
     aliases = [select.alias for select in parsed_query.selects]
 
-    # Verify that variant column names are preserved (double underscores intact)
-    # BUG: Currently this will FAIL because _normalize_casefold uses normalize_identifier
-    # which collapses __ to _, turning value__v_double into value_v_double
+    # verify that variant column names are preserved (double underscores intact)
     assert (
         "value__v_double" in aliases
     ), f"Variant column 'value__v_double' was not preserved. Got aliases: {aliases}"


rudolfix:
## Query Compilation Pipeline in `Relation.to_sql()`

The `to_sql()` method implements a **multi-stage query compilation pipeline** that transforms a logical SQL query into an executable physical query for the target database. The pipeline follows standard database query processing stages:

---

### **Stage 1: Semantic Analysis & Name Resolution**

```python
_, _qualified_query = _get_relation_output_columns_schema(self)
```

**Performed by:** `sqlglot.optimizer.qualify()`

**Professional terms:**
- **Name resolution** - Resolving unqualified identifiers to their fully-qualified forms
- **Star expansion** - Expanding `SELECT *` to explicit column lists
- **Scope analysis** - Determining which columns belong to which tables in the query's namespace
- **Column disambiguation** - Adding table prefixes to resolve ambiguous column references

**Output:** A **qualified query** where all identifiers have been resolved against the catalog schema.

---

### **Stage 2: Name Binding & Physical Rewriting**

```python
query = bind_query(
    qualified_query=_qualified_query,
    sqlglot_schema=self._dataset.sqlglot_schema,
    expand_table_name=lambda name: self.sql_client.make_qualified_table_name_path(...),
    casefold_identifier=self.sql_client.capabilities.casefold_identifier,
)
```

**Professional terms:**
- **Name binding** - Mapping logical schema objects to physical database objects
- **Identifier normalization** - Applying destination-specific identifier casing rules
- **Identifier quoting** - Delimited identifier handling for case preservation
- **Table name expansion** - Resolving logical table names to fully-qualified physical paths (catalog.schema.table)
- **Case folding** - Applying destination-specific case transformation (UPPER for Snowflake, lower for Postgres)
- **Alias injection** - Adding column aliases to preserve semantic equivalence after case folding

**Output:** A **bound query** with all identifiers mapped to physical database objects and properly escaped.

---

### **Stage 3: Code Generation (SQL Emission)**

```python
return query.sql(dialect=self.destination_dialect, pretty=pretty)
```

**Professional terms:**
- **SQL generation** - Emitting SQL text from the abstract syntax tree
- **Dialect transpilation** - Translating generic SQL to target database dialect
- **Syntax rewriting** - Applying dialect-specific syntax transformations
- **Pretty printing** - Optional formatting for human readability

**Output:** An **executable SQL string** in the target database's SQL dialect.

---

## Complete Pipeline Summary

```
Logical Query (user input)
    ↓
┌─────────────────────────────────────────┐
│  SEMANTIC ANALYSIS                      │
│  - Name resolution                      │
│  - Star expansion                       │
│  - Scope analysis                       │
└─────────────────────────────────────────┘
    ↓
Qualified Query (all names resolved)
    ↓
┌─────────────────────────────────────────┐
│  NAME BINDING & REWRITING               │
│  - Table name expansion                 │
│  - Identifier case folding              │
│  - Identifier quoting                   │
│  - Alias injection                      │
└─────────────────────────────────────────┘
    ↓
Bound Query (physical objects mapped)
    ↓
┌─────────────────────────────────────────┐
│  CODE GENERATION                        │
│  - SQL emission                         │
│  - Dialect transpilation                │
└─────────────────────────────────────────┘
    ↓
Executable SQL String
```

---

## Terminology Notes

- **Qualified query** = Query where all identifiers have been resolved within their scope
- **Bound query** = Query where logical names have been mapped to physical database objects
- **Semantic analysis** = Understanding the meaning and structure of the query
- **Name binding** = Connecting symbolic names to actual storage locations
- **Transpilation** = Translating between SQL dialects (source-to-source compilation)

This pipeline mirrors the classic database query compilation stages: **Parsing → Semantic Analysis → Optimization → Code Generation**, though optimization is handled separately by SQLGlot's optimizer.

rudolfix:
clickhouse error is unrelated

dlt/normalize/items_normalizers.py
```diff
@@ -92,9 +92,13 @@ def __call__(self, extracted_items_file: str, root_table_name: str) -> List[TSch
 
 
 class ModelItemsNormalizer(ItemsNormalizer):
-    def _normalize_casefold(self, ident: str) -> str:
-        return self.config.destination_capabilities.casefold_identifier(
-            self.schema.naming.normalize_identifier(ident)
+    @property
```
zilto: I don't think this should be a property. If it needs to be a property, the method name shouldn't be a verb.

This changes the meaning of line 131 and 132.

> NORM_C_DLT_LOAD_ID = self._normalize_casefold(C_DLT_LOAD_ID)
> NORM_C_DLT_ID = self._normalize_casefold(C_DLT_ID)

Instead of a calling functions, we're accessing a property which is a callable,  which is called. Seems convoluted.

tests/libs/test_sqlglot.py
```diff
@@ -556,3 +557,75 @@ def test_reorder_or_adjust_outer_select(dialect: TSqlGlotDialect) -> None:
...
+    # import the naming convention module
+    if naming_convention == "snake_case":
+        naming_module_path = "dlt.common.normalizers.naming.snake_case"
+    else:
+        naming_module_path = naming_convention
+    naming_module = importlib.import_module(naming_module_path)
+    naming = naming_module.NamingConvention()
+
+    # test query with mixed case, paths, and aliases (including __ in table/alias names)
+    query = cast(
```
zilto, line 581: why do we need the `cast()` here?

dlt/dataset/relation.py
```diff
@@ -245,10 +245,12 @@ def to_sql(self, pretty: bool = False, *, _raw_query: bool = False) -> str:
             query = self.sqlglot_expression
         else:
             _, _qualified_query = _get_relation_output_columns_schema(self)
-            query = _normalize_query(
+            query = bind_query(
                 qualified_query=_qualified_query,
                 sqlglot_schema=self._dataset.sqlglot_schema,
-                sql_client=self.sql_client,
+                expand_table_name=lambda name: self.sql_client.make_qualified_table_name_path(
```
zilto: can we avoid using anonymous function via `lambda` ? 

The same pattern would be cleaner with

```python
import functools

query = bind_query(
  qualified_query=_qualified_query,
  sqlglot_schema=self._dataset.sqlglot_schema,
  expand_table_name=functools.partial(
    self.sql_client.make_qualified_table_name_path,
    quote=False,
    casefold=False,
  ),
)
```

dlt/destinations/queries.py
```diff
@@ -1,68 +1,32 @@
...
-
-    return qualified_query
+    """Backward compatibility wrapper for bind_query.
+
+    TODO: remove after next dlthub release
+    """
+    return bind_query(
+        qualified_query,
+        sqlglot_schema,
+        expand_table_name=lambda name: sql_client.make_qualified_table_name_path(
```
zilto: suggestion: avoid `lambda` and use `functools.partial`

tests/destinations/test_queries.py
```diff
@@ -129,10 +129,12 @@ def test_normalize_query():
     )
 
     with duckdb_destination_client.sql_client as sql_client:
-        normalized_query_expr = _normalize_query(
+        normalized_query_expr = bind_query(
             qualified_query=cast(sge.Query, qualified_query_expr),
             sqlglot_schema=sqlglot_schema,
-            sql_client=sql_client,
+            expand_table_name=lambda name: sql_client.make_qualified_table_name_path(
```
zilto: avoid `lambda`,  use `functools.partial()`

dlt/common/libs/sqlglot.py
```diff
@@ -848,6 +848,20 @@ def build_outer_select_statement(
     return outer_select, needs_reordering
 
 
+def create_outer_select_identifier_normalizer(
```
zilto: This functions seems to be used in a [single place](https://github.com/dlt-hub/dlt/pull/3627/changes#diff-91126b068ef86213d8e95fcc59fa504fc708a9a20bc74c36a165d080a7e26050R99) and seems to be unecessary complexity.

The entire logic of this code path is confusing and seems unkind for the next person that will need to touch the code... Should at least comment **why**.

zilto (commented):
Lots of comments to remove

dlt/normalize/items_normalizers.py
rudolfix: OK - I'll merge `create_outer_select_identifier_normalizer` back here

tests/libs/test_sqlglot.py
rudolfix: `normalize_query_identifiers` takes `Query` as input. I could typeignore instead but I think this is fine

rudolfix:
@zilto addressed all the above. tests are failing because scaffold api fails at this moment

PR #3628, opened 2026-02-12
detect when running via agent

rudolfix:
### Description
This updates exec info detection with cursor, claude and codex agents. verified via auth documentation and manual attempts

---

**Continue Tasks:** ✅ 1 no changes — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fdlt-hub%2Fdlt%2Fpull%2F3628&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/tests/common/configuration/utils.py
+++ b/tests/common/configuration/utils.py
@@ -115,17 +115,6 @@ class InstrumentedConfiguration(BaseConfiguration):
 def environment(preserve_environ) -> Any:
     environ.clear()
     return environ
-    # saved_environ = environ.copy()
-    # # preserve these explicitly
-    # preserved = {
-    #     DLT_TEST_STORAGE_ROOT: saved_environ.get(DLT_TEST_STORAGE_ROOT),
-    #     PYTEST_XDIST_WORKER: saved_environ.get(PYTEST_XDIST_WORKER),
-    # }
-    # environ.clear()
-    # environ.update({k: v for k, v in preserved.items() if v is not None})
-    # yield environ
-    # environ.clear()
-    # environ.update(saved_environ)
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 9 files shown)


dlt/common/runtime/exec_info.py
```diff
@@ -61,6 +61,14 @@ def exec_info_names() -> List[TExecInfoNames]:
         names.append("gcp_cloud_function")
     if is_streamlit():
         names.append("streamlit")
+    if is_claude_code():
```
martinibach, line 64: Please add focused tests for the new agent detection branches (`is_claude_code`, `is_claude_code_cli`, `is_cursor`, `is_codex`) and `exec_info_names()` composition. Current telemetry assertions only check type/subset, so regressions here would be silent.

dlt/common/runtime/exec_info.py
```diff
@@ -76,6 +84,47 @@ def is_streamlit() -> bool:
...
+    return os.environ.get("CLAUDECODE") == "1" and os.environ.get("CLAUDE_CODE_ENTRYPOINT") == "cli"
+
+
+def is_cursor() -> bool:
+    """Return True if running in Cursor IDE Agent.
+
+    Detects CURSOR_AGENT environment variable set when Cursor Agent is running.
+    https://cursor.com/docs/agent/terminal
+    """
+    return os.environ.get("CURSOR_AGENT") is not None
```
martinibach: Presence-only check can false-positive when `CURSOR_AGENT` is set to an empty or disabled value by inherited env. Suggest making it value-aware for consistency with Claude/Codex checks.

```suggestion
    return os.environ.get("CURSOR_AGENT") == "1"
```

PR #3648, opened 2026-02-18
fix/ autouse test storage dir not empty

tetelio:
This PR attemps to fix https://github.com/dlt-hub/dlt/actions/runs/22117124000/job/63928366759?pr=3647 

IMPORTANT NOTE: this error is not raised always in CI or locally. it happens _only_ sometimes. Since it is not always raised, tests passing are not a signal of a perfect solution. The theoretical correctness of this solution should be evaluated carefully.

```py
E   OSError: [Errno 39] Directory not empty: '/home/runner/work/dlt/dlt/_storage_gw0/'

/opt/.../shutil.py:763: in rmtree
    onerror(os.rmdir, path, sys.exc_info())

...

FileStorage.rmtree_del_ro
    if os.path.isdir(name):
>       os.rmdir(name)
E       OSError: [Errno 39] Directory not empty: '/home/runner/work/dlt/dlt/_storage_gw0/'

```

I think the problem is in the [delete read only](https://github.com/dlt-hub/dlt/blob/d48b237951915079e300cd6434e86444105abbc8/dlt/common/storages/file_storage.py#L322) implementation of file storage:
```py
 @staticmethod
    def rmtree_del_ro(action: AnyFun, name: str, exc: Any) -> Any:
        if action is os.unlink or action is os.remove or action is os.rmdir:
            os.chmod(name, stat.S_IWRITE)
            if os.path.isdir(name):
                os.rmdir(name)
            else:
                os.remove(name)
```

The shutil documentation mentions how to do this exact action of dealing with read-only stuff https://docs.python.org/3/library/shutil.html#rmtree-example

```py
def remove_readonly(func, path, _):
    "Clear the readonly bit and reattempt the removal"
    os.chmod(path, stat.S_IWRITE)
    func(path)
```

I will keep the ` if action is os.unlink or action is os.remove or action is os.rmdir` to ensure only destructive operations are being handled here. For the rest of the implementation, we can just keep action(path), because shutil already knows if dealing with dir or file, hence propagates the right operation. That function respects all shutil internals while changing the bit (file or dir) to being able to write to it, then attempting the same operation.


rudolfix:
@tetelio another root cause could be a race condition where another tests writes to the folder while stuff is being deleted. this would be a sign of a deeper problem in test setup (or dlt code). it get that use of fixed place for test storage with auto clean could we replaced with a random tmp folder but then we'll never catch problems like this (not the first one to be sure)

tetelio:
@rudolfix understood. Do you still want to proceed with this PR or we decline it? I think the code is _more correct_ but i dont know if the code is a solution, as you mentioned

PR #3659, opened 2026-02-20
Fix: Mermaid doesn't handle incomplete columns

anuunchin:
When a column is in the source but doesn't have data, it is set in the schema without data_type. Mermaid is not able to handle this becuase it directly tries to access `column["data_type"]`

PS: The mcp used in skills is consistently failing to use mermaid because of this.

### DIFF
--- a/dlt/helpers/mermaid.py
+++ b/dlt/helpers/mermaid.py
@@ -62,7 +62,10 @@ def _to_mermaid_table(
 
 # TODO add scale & precision to `data_type`
 def _to_mermaid_column(column: TColumnSchema, hide_descriptions: bool = False) -> str:
-    mermaid_col = column.get("data_type", "unknown") + " " + column["name"]
+    # NOTE: incomplete columns (no data_type) currently arise from "seen-null-first" propagation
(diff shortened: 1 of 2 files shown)


dlt/helpers/mermaid.py
```diff
@@ -62,7 +62,7 @@ def _to_mermaid_table(
 
 # TODO add scale & precision to `data_type`
 def _to_mermaid_column(column: TColumnSchema, hide_descriptions: bool = False) -> str:
-    mermaid_col = column["data_type"] + " " + column["name"]
+    mermaid_col = column.get("data_type", "unknown") + " " + column["name"]
```
zilto: `unknown` is confusing to users IMO. It doesn't communicate why / how. 

I suggest `missing data_type` or `no data seen`. Pick the expression (or other) that best describes the cases covered by "unknown"

zilto (changes requested):

dlt/helpers/mermaid.py
anuunchin: 🫡

PR #3660, opened 2026-02-20
Dashboard cleanup

sh-rp:
## Testing needed

- **Manual testing with various pipeline result states** (successful runs, failed extracts/loads, never-run pipelines, no-destination pipelines, broken traces, multi-schema pipelines) is needed to verify the simplified UI works correctly across all sections
- **Data quality integration** (`dlthub.data_quality`) needs to be tested with the refactored code
- Looking for feedback on the simplified design and code organization

### Code organization
- **Broke up monolithic `utils.py` (1028 lines) and `dlt_dashboard.py`** into focused modules under `utils/`: `queries.py`, `schema.py`, `trace.py`, `pipeline.py`, `data_quality.py`, `visualization.py`, `home.py`, `ui.py`, `formatters.py`
- **Removed `ui_elements.py`**, consolidated UI helpers into `utils/ui.py`
- **Added `const.py`** for magic strings/numbers previously scattered across files
- **Added `types.py`** with `TypedDict` definitions (`TNameValueItem`, `TTableListItem`, `TLoadItem`, etc.)

### Boilerplate reduction
- **`TSectionStrings` NamedTuple** groups section name/title/subtitle into one object — simplified 8 section call sites from ~7 lines to ~2 lines each
- **`BoundedDict`** deduplicates cache eviction logic across query and schema caches
- **Walrus operators** applied throughout to reduce assign-then-check patterns
- **Removed redundant `traceback_string=traceback.format_exc()` kwargs** at 6 call sites where `error_callout` already defaults to it
- **`build_load_details` now accepts `TLoadItem`** directly instead of 3 extracted fields
- **Unified empty `_result` handling** — all dashboard cells return `None` instead of an empty `mo.vstack([])` when there is nothing to render

### Cleanup
- Extracted all user-facing strings into `strings.py`
- Added dedicated `app_pipeline_no_trace` message for pipelines without trace data (previously reused the "pipeline not found" text)
- Fixed several obvious bugs and typography issues
- Cleaned up CSS
- Added missing unit tests for dashboard sections
- Added XML-based HTML validation tests for generated visualization HTML (using stdlib `xml.etree.ElementTree`)
- Added proper type annotations throughout

### E2E test fixes
- Fixed code editor text assertions to use `.cm-line` CSS locator instead of `get_by_text` — works across marimo versions where syntax highlighting splits text into `<span>` elements
- Scoped short-string assertions (e.g. `"id"`) to `get_by_role("table")` to avoid ambiguity with code editor content

**Net: ~2700 lines added (new modules + tests), ~2100 lines removed → code is more modular with better test coverage**

### DIFF
--- a/dlt/_workspace/helpers/dashboard/dlt_dashboard_styles.css
+++ b/dlt/_workspace/helpers/dashboard/dlt_dashboard_styles.css
@@ -16,8 +16,8 @@
   --yellow-text: #AD7F37;
   --green-bg: #a8e4ac;
   --green-text: #468A47;
-  --red-bg: #D47672;
-  --red-text: #FBE8E8;
+  --red-text: #D47672;
+  --red-bg: #FBE8E8;
 }
 
 /* Reduce gap between cells */
--- a/dlt/_workspace/helpers/dashboard/strings.py
+++ b/dlt/_workspace/helpers/dashboard/strings.py
@@ -104,7 +104,7 @@ If the dashboard cannot connect to the destination, you will receive a warning a
 * [Marimo docs](https://docs.marimo.io/) - Learn more about Marimo, the framework that powers the dltHub workspace dashboard
 
 <small>
-2025 [dltHub](https://dlthub.com)
+2026 [dltHub](https://dlthub.com)
 </small>
 
 """
--- a/dlt/_workspace/helpers/dashboard/utils/pipeline.py
+++ b/dlt/_workspace/helpers/dashboard/utils/pipeline.py
@@ -186,12 +186,14 @@ def exception_section(p: dlt.Pipeline) -> List[mo.Html]:
         return []
 
     last_exception = exception_step.exception_traces[-1]
-    title = f"{last_exception['exception_type']}: {last_exception['message']}"
+    title = f"Exception encountered during last run in step **{exception_step.step}**"
+    subtitle = f"{last_exception['exception_type']}: **{last_exception['message']}**"
 
     _result = []
     _result.append(
         ui.title_and_subtitle(
             title,
+            subtitle,
             title_level=2,
         )
     )


dlt/_workspace/cli/utils.py
```diff
@@ -1,7 +1,9 @@
 import ast
 import os
 import shutil
-from typing import Any, Callable, Dict, List, Tuple, cast
+from typing import Any, Callable, Dict, List, Tuple
+
+from dlt._workspace.helpers.dashboard.types import TPipelineListItem
```
zilto: The name `.types` will collide with importing the built-in `types`.

rudolfix:
@sh-rp do you have still this claude session? can you convert it into `cleanup-dashboard-code` skill or smth? this code tends to get noisy. so a good cleanup from time to time is necessary. also pls. merge this ASAP... may generate a lot of conflicts

dlt/_workspace/cli/utils.py
sh-rp: Thanks for checking this! I will rename it to typing bc we have it like this in the rest of the codebase. I think it generally should not be a problem having module paths that include a section that is named like some other top level import.

dlt/_workspace/helpers/dashboard/strings.py
anuunchin: Should we update the year to 2026?

dlt/_workspace/helpers/dashboard/dlt_dashboard_styles.css
```diff
@@ -25,6 +20,11 @@
   --red-text: #FBE8E8;
```
anuunchin: wondering whether it's a bit too light (the `failed` and `aborted` badges)

Image

dlt/_workspace/helpers/dashboard/dlt_dashboard.py
anuunchin, line 1: if the load step fails, we see an error like this, but this looks like a bug/logic issue in dlt itself in `dlt/common/pipeline.py::LoadInfo` on line 341 where we assert that metrics list is always 1 

Image

dlt/_workspace/helpers/dashboard/dlt_dashboard_styles.css
sh-rp: You are right!

dlt/_workspace/helpers/dashboard/dlt_dashboard.py
sh-rp: I can't see this. I have a pipeline that breaks at the load step but this is not displayed. How do I reproduce?

PR #3661, opened 2026-02-20
feat(workspace): add default exclude patterns for file selector

canassa:
### Description

Add a `DEFAULT_EXCLUDES` list to `WorkspaceFileSelector` to ensure well known non deployable directories and files are always excluded, even when no `.gitignore` is present.

The list should include:

* `.git/`
* `.venv/`
* `__pycache__/`
* `node_modules/`
* `.mypy_cache/`
* `.pytest_cache/`
* `.ruff_cache/`
* `.tox/`
* `*.egg-info/`
* `*.pyc`

### DIFF
--- a/dlt/_workspace/deployment/file_selector.py
+++ b/dlt/_workspace/deployment/file_selector.py
@@ -5,18 +5,23 @@ from pathspec.util import iter_tree_files
 
 from dlt._workspace._workspace_context import WorkspaceRunContext
 
-# well-known directories and files that should never be deployed
-DEFAULT_EXCLUDES: List[str] = [
-    ".git/",
-    ".venv/",
+# fallback ignore patterns used when no ignore file is found in the workspace
+DEFAULT_IGNORES: List[str] = [
     "__pycache__/",
-    "node_modules/",
-    "*.pyc",
-    ".mypy_cache/",
-    ".pytest_cache/",
-    ".ruff_cache/",
-    ".tox/",
+    "*.py[cod]",
+    ".venv/",
+    "venv/",
+    "dist/",
+    "build/",
     "*.egg-info/",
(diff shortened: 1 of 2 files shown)


dlt/_workspace/deployment/file_selector.py
```diff
@@ -5,6 +5,20 @@
 
 from dlt._workspace._workspace_context import WorkspaceRunContext
 
+# well-known directories and files that should never be deployed
+DEFAULT_EXCLUDES: List[str] = [
```
sh-rp: Cursor suggests this list (I asked for bare essentials, so this could be longer but it is a good fallback)

```
  *.py[cod]
  .venv/
  venv/
  dist/
  build/
  *.egg-info/
  .mypy_cache/
  .ruff_cache/
  .pytest_cache/
  .coverage
  htmlcov/
  *.so
  .DS_Store
  .env
```

dlt/_workspace/deployment/file_selector.py
```diff
@@ -36,6 +50,7 @@ def __init__(
     def _build_pathspec(self, additional_excludes: List[str]) -> PathSpec:
         """Build PathSpec from ignore file + defaults + additional excludes"""
         patterns: List[str] = [f"{self.settings_dir.relative_to(self.root_path)}/"]
+        patterns.extend(DEFAULT_EXCLUDES)
```
sh-rp: Please only use the fallback if there is no gitignore. This way the user can control this. If they want to commit the build folder for example, they can do it. With your implementation it will always be stripped out.

dlt/_workspace/deployment/file_selector.py
sh-rp: node_modules you can remove for sure here.

PR #3667, opened 2026-02-23
Fetch Databricks Compute Credentials

### LINKED ISSUE #3536: Databricks notebook default credentials are not derived as expected
### dlt version

1.20

### Describe the problem

The [docs](https://dlthub.com/docs/dlt-ecosystem/destinations/databricks#direct-load-databricks-managed-volumes) state that:
>When executed in a Databricks Notebook, dlt uses the notebook context for configuration if not explicitly provided.

My interpretation of this is that it will use the hostname and HTTP path of the compute that is *currently* running the dlt process. After all, that is part of the notebook's context.

In reality, it gets the details of *a* SQL warehouse that the user has access to. We only became aware of this after deleting the only warehouse that our source load service principal had access to (the 'Starter Endpoint' that all users have access to by default) resulting in our process failing.

See the snippet below of the current code from `DatabricksCredentials.on_resolved` in `dlt/destinations/impl/databricks/configuration.py`:

```py
        if not self.server_hostname or not self.http_path:
            try:
                # attempt to fetch warehouse details
                from databricks.sdk import WorkspaceClient

                w = WorkspaceClient()
                # warehouse ID may be present in an env variable
                if w.config.warehouse_id:
                    warehouse = w.warehouses.get(w.config.warehouse_id)
                else:
                    # for some reason list of warehouses has different type than a single one 🤯
                    warehouse = list(w.warehouses.list())[0]  # type: ignore[assignment]
                logger.info(
                    f"Will attempt to use warehouse {warehouse.id} to get sql connection params"
                )
                self.server_hostname = self.server_hostname or warehouse.odbc_params.hostname
                self.http_path = self.http_path or warehouse.odbc_params.path
            except Exception:
                pass
```

### Expected behavior

As mentioned above, I would expect dlt to use the hostname and HTTP path of the compute that is *currently* running the dlt process, when running in a Databricks Notebook.

### Steps to reproduce

Run a dlt process with a Databricks destination (with `server_hostname` or `http_path` omitted from the credentials) in a Databricks notebook as a user (or service principal) who has no access to any SQL warehouses on the workspace.

### Operating system

Linux

### Runtime environment

Other

### Python version

3.12

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

The **hostname** is available from the Databricks SDK (as used in the current code to get SQL warehouse details). For example:

```py

from databricks.sdk import WorkspaceClient
from urllib.parse import urlparse

w = WorkspaceClient()

# Full workspace base URL, e.g. "https://adb-1234567890123456.12.azuredatabricks.net"
workspace_url = w.config.host

# Just the hostname part, e.g. "adb-1234567890123456.12.azuredatabricks.net"
server_hostname = urlparse(workspace_url).hostname
```

The **HTTP path** can be derived using `WorkspaceClient.dbutils` (which is already used in the `DatabricksCredentials` class):

```py
notebook_context = w.dbutils.notebook.entry_point.getDbutils().notebook().getContext()
cluster_id = notebook_context.clusterId().get()
workspace_id= notebook_context.workspaceId().get()

http_path = f"sql/protocolv1/o/{workspace_id}/{cluster_id}"
```

aditypan:
Added support for fetching credentials from databricks compute (shared / job) when a notebook using dlt is running on top of it. This will fix the issue of trying to connect to databricks sql warehouse by default.

### Related Issues

- Fixes #...
- Closes #...
- Resolves #3536

### DIFF
--- a/dlt/destinations/impl/databricks/configuration.py
+++ b/dlt/destinations/impl/databricks/configuration.py
@@ -49,12 +49,18 @@ class DatabricksCredentials(CredentialsConfiguration):
             if w is not None:
                 try:
                     # attempt context authentication via notebook token
-                    self.access_token = w.dbutils.notebook.entry_point.getDbutils().notebook().getContext().apiToken().getOrElse(None)  # type: ignore[union-attr]
+                    self.access_token = (
+                        w.dbutils.notebook.entry_point.getDbutils()
+                        .notebook()
+                        .getContext()
+                        .apiToken()
+                        .getOrElse(None)
+                    )
                 except Exception:
                     self.access_token = None
 
                 try:
-                    self.access_token = w.config.authenticate  # type: ignore[assignment]
+                    self.access_token = w.config.authenticate
                     logger.info(f"Will attempt to use default auth of type {w.config.auth_type}")
                 except Exception:
                     pass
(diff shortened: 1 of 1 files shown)


aditypan:
@rudolfix Thanks for the help!

PR #3669, opened 2026-02-23
feat: `hf` protocol (Hugging Face) support for `filesystem` destination

### LINKED ISSUE #1227: Hugging Face Hub integration

jorritsandbrink:
### Description
This PR adds support for the `hf` (Hugging Face) protocol to the `filesystem` destination.

### Related Issues

- Closes #1227 

### Additional Context
This PR builds upon https://github.com/dlt-hub/dlt/pull/3391 and takes inspiration from https://github.com/dlt-hub/dlt/pull/3410.

### DIFF
--- a/dlt/destinations/impl/filesystem/filesystem.py
+++ b/dlt/destinations/impl/filesystem/filesystem.py
@@ -410,22 +410,46 @@ class HfFilesystemCommitJob(ReferenceFollowupJob):
     def run(self) -> None:
         from huggingface_hub import CommitOperationAdd
 
+        operations: List[Any] = [
+            CommitOperationAdd(path_in_repo=self.make_path_in_repo(path), path_or_fileobj=path)
+            for path in self.file_paths
+        ]
+
+        # for replace disposition, delete existing table files in the same atomic commit
+        delete_ops = self._get_delete_operations()
+        if delete_ops:
+            operations = delete_ops + operations
+
         start = _time.monotonic()
         self._job_client.hf_api.create_commit(
             repo_id=self._job_client.repo_id,
-            operations=[
-                CommitOperationAdd(path_in_repo=self.make_path_in_repo(path), path_or_fileobj=path)
-                for path in self.file_paths
-            ],
+            operations=operations,
             commit_message=f"Add files to {self.load_table_name}",
             repo_type="dataset",
         )
         elapsed = _time.monotonic() - start
+        n_adds = len(self.file_paths)
+        n_deletes = len(delete_ops) if delete_ops else 0
         logger.info(
             f"HF commit to {self._job_client.repo_id} for {self.load_table_name}"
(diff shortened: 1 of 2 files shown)


jorritsandbrink:
@rudolfix Could you review this PR?

Notes:
1. [Failing test](https://github.com/dlt-hub/dlt/actions/runs/22351285851/job/64682071343?pr=3669) is a flake.
2. Commit conflicts may occur when multiple threads commit concurrently to the same dataset repo. These errors are considered transient, so the retry mechanism kicks in and the commits go through in second instance. I've batched all data files for a table chain in a single commit to minimize this, but it still happens when loading multiple tables. Try e.g. `uv run pytest tests/load/pipeline/test_filesystem_pipeline.py::test_pipeline_merge_write_disposition[hf://datasets/dlthub] -s` to see commit conflicts in action.

rudolfix (approved):
this is really good! I do not have many comments - code is clean and sufficient amount of tests is active.

I pushed one change to separate upload from commit. My understanding is that we can prepare blobs fully in parallel and the conflicts are only on the commit itself - so I split this two phases. not sure it will reduce number of conflicts but will speed up upload (each file gets a thread). let's see. we can always reduce the number of threads in destinations caps.

pls take a look!

thanks!

jorritsandbrink:
@rudolfix Great improvements, thanks! Looks good to merge from my end.

PR #3676, opened 2026-02-24
Fix/3637 cli info, show fail on custom destinations 

### LINKED ISSUE #3637: `dlt pipeline xxx info` does not work when the destination is using  `@dlt.destination` decorator
### dlt version

1.21.0

### Describe the problem

Using the example code from https://dlthub.com/docs/dlt-ecosystem/destinations/destination#set-up-a-destination-function-for-your-pipeline, give as the `Destination module xxx is not registered` when I try the `dlt pipeline xxxx info`.

### Expected behavior

I would expect the `dlt pipeline xxx info` (and other pipeline subcommands) to work even when the destination is created via the `@dlt.destination` decorator. 

Or at least some documentation in the https://dlthub.com/docs/dlt-ecosystem/destinations/destination#set-up-a-destination-function-for-your-pipeline , explaining why I won't work. 

### Steps to reproduce

The project setup : 
```
mkdir dlttest
cd dlttest
uv init --bare --python '>=3.13,<3.14'
uv add 'dlt[cli,workspace]'
source .venv/bin/activate
```

The contents of `custom_destination_pipeline.py`, taken verbatim from [dltHub docs](https://dlthub.com/docs/dlt-ecosystem/destinations/destination#set-up-a-destination-function-for-your-pipeline):
```
import dlt
from dlt.common.typing import TDataItems
from dlt.common.schema import TTableSchema


@dlt.destination(batch_size=10)
def my_destination(items: TDataItems, table: TTableSchema) -> None:
    print(table["name"])
    print(items)


pipeline = dlt.pipeline("custom_destination_pipeline", destination=my_destination)
pipeline.run([1, 2, 3], table_name="items")

```

I can run the pipeline fine, **no errors**: 
```
python custom_destination_pipeline.py
items
[{'value': 1}, {'value': 2}, {'value': 3}]
```

But the `dlt pipeline custom_destination_pipeline info` does **NOT work**:
```
dlt pipeline custom_destination_pipeline info
Destination module `__main__.MyDestinationDestination` is not registered.Modules and attributes were tried in the following order and failed to import:
        mod:`__main__` attr: `MyDestinationDestination` failed due to: ImportSpecError and causing exception: __main__.__spec__ is None

NOTE: Please refer to our docs at 'https://dlthub.com/docs/reference/command-line-interface#dlt-pipeline' for further assistance.
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

python array for test

### dlt destination

_No response_

### Other deployment details

The destination is a  python function decorated with `@dlt.destination` 

### Additional information

_No response_

anuunchin:
Fixes UnknownDestinationModule errors when running dlt pipeline <name> info/show on pipelines using `@dlt.destination`.          
                                                                                                                              
### Underlying problem                                                                                                                        
                                                                                                                               
  `@dlt.destination` synthesizes a class in the caller's module (e.g. __main__.MyDestinationDestination). This path is
  non-importable from other processes (CLI, dashboard), causing dlt.attach() to fail.

### Changes

  - factory.py: Override destination_type to fall back to dlt.destinations.destination when the class lives in __main__
  - configuration.py: Clearer error message explaining the cross-process limitation
  - dashboard/utils/pipeline.py: Catch ConfigurationValueError so dashboard doesn't crash
  - destination.md: Added "CLI and dashboard limitations" section
  - Tests: Cover __main__ fallback, CLI info command, and dashboard operations for all three custom destination patterns
  (decorator, callable, string ref)

Resolves #3637


ecerulm:
I just tried this 

```
uv add 'git+https://github.com/dlt-hub/dlt.git@fix/3637-custom-dest-type' # to use dlt from this pr branch

uv run custom_destination_pipeline.py
items
[{'value': 1}, {'value': 2}, {'value': 3}]

uv run dlt pipeline custom_destination_pipeline show
2026-02-24 14:14:04,071|[WARNING]|63987|6195720192|dlt|factory.py|__init__:119|No destination callable provided, providing dummy callable which will fail on load.
2026-02-24 14:14:05,099|[WARNING]|63987|6195720192|dlt|factory.py|__init__:119|No destination callable provided, providing dummy callable which will fail on load.
```

When the marimo notebook opens it no longer shows the traceback/stracktrace 

image

So it looks good

ecerulm:
It still shows the warning `No destination callable provided, providing dummy callable which will fail on load` . I don't have too much experience in dlt so I don't  know if warnings like this can be safely ignored. But in principle I find this warning confusing as I actually provide a `destination_callable` so if it's possible it would be desirable to get rid of this warning if it's actually superfluous. Otherwise, new user like me will waste time trying to find why the warning appears.

ecerulm:
I just want to mention I tried this branch using https://github.com/ecerulm/dlt-destination-callable-package where the callable is from an installed package and I still  get the warning 


>Custom destination callable was not provided, using a dummy placeholder. Pipeline inspection will work but loading will not. This typically happens when a pipeline using @dlt.destination defined in a local script is accessed from another process (CLI, dashboard) because the callable cannot be imported **if it is not part of an installed package.**

As I said before I'm fairly new to dlt, and  I stitll  don't understand what I need to do to get rid of this warning, I think I have provided a callable that is defined in a package [`destination_callable="dlt_destination_callable_package.my_destination_alt"`](https://github.com/ecerulm/dlt-destination-callable-package/blob/6ca9a7c822582a492c86ce8597bfc994c180178e/src/dlt_destination_callable_package/pipeline.py#L18C13-L18C88), the pipeline runs but the `dlt pipeline pipeline20260225 show` still emits the warning.

anuunchin:
@ecerulm, did you rerun the pipeline with the dlt version in this branch? What exact steps are you taking? Also, please note that the implementation in this PR may be subject to change

Out of curiosity, what destination are you using though? 


EDIT: I just confirmed the string reference's callable is not persisted, so it's not available from another process.  The current solution only applies to @dlt.destination decorated functions

ecerulm:
For reference : 


> @ecerulm, did you rerun the pipeline with the dlt version in this branch? What exact steps are you taking? 

Yes, I did rerun 

```
uv run pipeline # which runs src/dlt_destination_callable_package/pipeline.py:run() method
items
[{'value': 1}, {'value': 2}, {'value': 3}]

uv run dlt pipeline pipeline20260225 info
2026-02-26 08:46:46,094|[WARNING]|51708|8502947968|dlt|factory.py|__init__:117|Custom destination callable was not provided, using a dummy placeholder. Pipeline inspection will work but loading will not. This typically happens when a pipeline using @dlt.destination defined in a local script is accessed from another process (CLI, dashboard) because the callable cannot be imported if it is not part of an installed package.
Found pipeline pipeline20260225 in /Users/rubenlaguna/.dlt/pipelines
....
```

This pipeline is just copied from the example from the official documentation [Custom destination: Reverse ET](https://dlthub.com/docs/dlt-ecosystem/destinations/destination) , using the third alternative called `Via a fully qualified string to function location` in that page. 

See https://github.com/ecerulm/dlt-destination-callable-package/blob/6ca9a7c822582a492c86ce8597bfc994c180178e/README.md



> Also, please note that the implementation in this PR may be subject to change

Yes, I understand I'm just trying it out with a [dummy example](https://github.com/ecerulm/dlt-destination-callable-package/blob/6ca9a7c822582a492c86ce8597bfc994c180178e/src/dlt_destination_callable_package/pipeline.py#L18) copied from [Custom destination: Reverse ET](https://dlthub.com/docs/dlt-ecosystem/destinations/destination). 

I'm just trying to provided more context for you. 


> Out of curiosity, what destination are you using though?

https://github.com/ecerulm/dlt-destination-callable-package/blob/6ca9a7c822582a492c86ce8597bfc994c180178e/src/dlt_destination_callable_package/pipeline.py#L15-L19
```
        destination=dlt.destination(
            "my_destination",
            # destination_callable=my_destination_alt,
            destination_callable="dlt_destination_callable_package.my_destination_alt",
        ),
```

> EDIT: I just confirmed the string reference's callable is not persisted, so it's not available from another process. The current solution only applies to @dlt.destination decorated functions


Ok if I use the `@dlt.decorator` the warning is gone
```
 rm -rf ~/.dlt/pipelines/pipeline20260225 && uv run pipeline && uv run dlt pipeline pipeline20260225 info
items
[{'value': 1}, {'value': 2}, {'value': 3}]
Found pipeline pipeline20260225 in /Users/rubenlaguna/.dlt/pipelines
Synchronized state:
```

ecerulm:
Would it be possible to support the three forms of `dlt.destination` described in [Use the custom destination in `dlt` pipeline][1]
  * **Directly reference the destination function**: (using the decorator) This works now **without warning** using this PR
  * **Via the `dlt.destination()` function**: (not using decorator) This shows the warning. 
  * **Via a fully qualified string to function location**: (not using decorator) This still shows the warning, you mention the  `string reference's callable is not persisted,`, is it out of the question to persist it? 
   
And if not, would it possible to document the problem in [Use the custom destination in dlt pipeline][1], and explain clearly what stops working when you see that warning? 

I still don't understand the warning meaning. When I see   `Pipeline inspection will work but loading will not.` part, I wonder if I run `dlt pipeline xxx info` is it "loading anything"  or not.  I f I run `dlt pipeline xxx show` is there any part of `dlt pipeline xx show` that will not work due to the "loading"?. I'm assuming that none of those commands "load" anything but since I'm a newbie at dlt...

In  any case, it's **great** that now works for decorated function, looking forward to see this merged in next release. 
   
 [1]: https://dlthub.com/docs/dlt-ecosystem/destinations/destination#use-the-custom-destination-in-dlt-pipeline "Use the custom destination in `dlt` pipeline"

anuunchin:
@ecerulm  thanks for the additional context, I see how the behaviour/warning is confusing - will try to make it clearer both in the warning and the docs

By `Out of curiosity, what destination are you using though?` I meant what's the end destination you want to load data to using the custom destination functionality

ecerulm:
> By `Out of curiosity, what destination are you using though?` I meant what's the end destination you want to load data to using the custom destination functionality

Ah, I originally wanted to "load" from GCP BigQuery and  write to Airtable HyperDB ([upsertRecords](https://airtable.com/developers/web/api/hyperdb-upsert-records-by-primary-keys) and [deleteRecords](https://airtable.com/developers/web/api/hyperdb-delete-records-by-primary-keys))

Now I'm doing it via a custom python script.

dlt/_workspace/helpers/dashboard/utils/pipeline.py
```diff
@@ -132,6 +133,9 @@ def get_local_data_path(pipeline: dlt.Pipeline) -> str:
     """Get the local data path of a pipeline"""
     if not pipeline.destination:
         return None
+    # custom destination config does not implement WithLocalFiles
```
rudolfix: do we need this? code below should handle this. if we need to extend `except` list then so be it. IMO you can just remove it...

dlt/common/destination/reference.py
```diff
@@ -146,7 +146,13 @@ def configured_name(self) -> str:
 
     @property
     def destination_type(self) -> str:
-        full_path = self.__class__.__module__ + "." + self.__class__.__qualname__
+        cls = self.__class__
```
rudolfix: this is custom destination code leaking into the base class. why? IMO not needed

dlt/common/destination/reference.py
```diff
@@ -401,15 +407,28 @@ def from_reference(
...
+        try:
+            factory = cls.find(ref)
+            return factory(**kwargs)
+        except UnknownDestinationModule:
+            # plain functions from installed packages used as destination_callable (without
+            # @dlt.destination decorator) store their importable path as destination_type
+            # in pipeline state (e.g. "my_pkg.my_func"). find() fails here because the ref
+            # resolves to a function, not a Destination subclass. import it and wrap in the
+            # custom destination factory.
+            if "." in ref:
```
rudolfix: I think I know what happens here but this is too complicated for what it gives us. if we want to encode callable name in `destination_type` we need something better...

dlt/common/reflection/ref.py
```diff
@@ -8,6 +10,27 @@
 from dlt.common.typing import TAny
 
 
+@functools.lru_cache(maxsize=64)
+def is_installed_module(module_name: str) -> bool:
```
rudolfix: AFAIK this is used to decide if we keep callable type in `destination_type` or `dlt.destinations.destination`. I'm not sure what problem it solved. looks like a real edge case... and this is still IMO fragile. I think there's a better way

docs/website/docs/release-notes/1.22.md
```diff
@@ -0,0 +1,27 @@
+---
+title: "Release highlights: 1.22"
```
rudolfix: did you write this as well :) ?

dlt/destinations/impl/destination/factory.py
```diff
@@ -153,6 +161,24 @@ def __init__(
             **kwargs,
         )
 
+    @property
+    def destination_type(self) -> str:
```
rudolfix: this is the right place to fix the problem. isn't it enough to just return `__orig_base__` here? or `dlt.destinations.destination` if empty. then it will be recreated with dummy callable but that's ok if you do not plan to write into it.

a lot of effort went here for storing and restoring callable type. IMO this will never fully work - code is always needed to create writable custom destination. I'll skip it for now.

(if you want to try: put the callable type in destination_name, this is stored in pipeline state and passed to destination factory when restoring. then if destination_callable is empty and destination name contains dots - you can use it as destination_callable - still I think this is overkill for this ticket)

PR #3677, opened 2026-02-24
feat(pipeline): context manager for pre/post operation on `pipeline.run()`

rudolfix:
### Description
This PR enables and demonstrates customization of the run method: it injects additional data to be loaded in the same trace transaction in an idempotent (can be retried) way.

See the included test for PoC

### DIFF
--- a/tests/pipeline/test_pipeline.py
+++ b/tests/pipeline/test_pipeline.py
@@ -71,7 +71,7 @@ from dlt.pipeline.exceptions import (
 from dlt.pipeline.helpers import retry_load
 
 from dlt.pipeline.pipeline import Pipeline
-from dlt.pipeline.trace import PipelineTrace, PipelineStepTrace
+from dlt.pipeline.trace import PipelineTrace, PipelineStepTrace, load_trace
 from dlt.pipeline.typing import TPipelineStep
 
 from tests.common.utils import TEST_SENTRY_DSN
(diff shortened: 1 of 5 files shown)


tests/pipeline/test_pipeline.py
```diff
@@ -4856,6 +4856,180 @@ def fail_extract():
...
+            # guard idempotency via local state
+            try:
+                self.get_local_state_val(SIDECAR_LOADED_KEY)
+            except KeyError:
+                self.set_local_state_val(SIDECAR_LOADED_KEY, True)
+                Pipeline._run_once(self, sidecar_source(), **kwargs)
+            return load_info
+
+        def __reduce__(self):
+            return (Pipeline.__new__, (Pipeline,), Pipeline.__getstate__(self))
```
zilto, line 4942: This code really deserve some comments and explanation. 

## Class vs instance
AFAIU, the key trick is to use the class object instead of the instance

```python
# via class, passing `self` explicitly
Pipeline._run_once(self, data, **kwargs)

# instead of via instance
self._run_once(data, **kwargs) 
```

I understand that `SidecarPipeline` overrides `_run_once` which is called by the user's code `pipeline.run()`. So `SidecarPipeline._run_once()` references `Pipeline._run_once()` to avoid recursion

## State key
How / why is the local state key set? It seems to not be used anywhere. I imagine that on a retry, you can tell if you need to retry main pipeline, sidecar pipeline, or both

zilto (commented):
My summary of the design constraints:
- `Pipeline.run()` is a public interface; it is what people call in their code.
- `Pipeline.run()` decorated with a bunch of context manager and has a lot of built-in assumptions for transactions, which makes patching it hard.
- `Pipeline._run_once()` is a new method that is private and include the whole logic of what `Pipeline.run()` does. It can be patched while preserving all of the decorators, context manager, public interface

## Suggestion
- rename `Pipeline._run_once()` to `Pipeline._run()` 
- move the logic of `._run_once()` to an internal function called `_run_pipeline()`
    ```python
    def _run_pipeline(
       pipeline: Pipeline,
        data: Any,
        *,
        ...,
    ) -> LoadInfo:
        pipeline.extract(...)
        pipeline.normalize()
        return pipeline.load(...)

    class Pipeline:
       def _run(self, ...):
           return _run_pipeline(...)

        def run(self, ...):
            return self._run(...)
    ```
    This reduces clunkiness of patching by replacing base class invocation `Pipeline._run(...)`  with functional pattern `_run_pipeline()`

    ```python
    class SidecarPipeline(Pipeline):
        def _run(self, data: Any, **kwargs: Any) -> LoadInfo:
            load_info = Pipeline._run(self, data, **kwargs)
            return load_info
    ```
    Becomes
    ```python
    class SidecarPipeline(Pipeline):
        def _run(self, data: Any, **kwargs: Any) -> LoadInfo:
            load_info = _run_pipeline(self, data, **kwargs)
            return load_info
    ```

zilto:
related #4320

Issue #3657, opened 2026-02-20
Incremental primary_key is overridden by the resource's primary_key


burnbay:
### dlt version

1.19.1

### Describe the problem

I'm setting primary_key on a table resource like this:
```python
resource.apply_hints(primary_key=["key1", "key2"])
```
I'm also setting incremental where i define the `primary_key = ()` to avoid flooding the state. This is stated in the documentation: `primary_key: Optional primary key used to deduplicate data. If not provided, a primary key defined by the resource will be used. Pass a tuple to define a compound key. Pass empty tuple to disable unique checks` 
We have a couple of tables with lots of rows with the same incremental value. 

When I run my code the primary key from the table resource is applied to the incremental.primary_key here:
https://github.com/dlt-hub/dlt/blob/4caedcad4921d4f8c81489bc662ad182420bf17a/dlt/extract/resource.py#L526

This will break what is stated in the docs. 

### Expected behavior

I expect the `incremental.primary_key = ()` to be respected to diable unique checks.

### Steps to reproduce

```python
resource = sql_table(...)

resource.apply_hints(incremental = dlt.sources.incremental(..., primary_key=())

resource.apply_hints(primary_key = ["key1", "key2"])

```

### Operating system

Linux

### Runtime environment

Docker, Docker Compose

### Python version

3.11

### dlt data source

database

### dlt destination

Snowflake

shnhdan:
Hi! I’d like to work on this issue if it’s still open and unassgined.

Could you confirm the expected behaviour for precedence between incremental and resource `primary_key`, and whether there are constraints or existing tests I should follow?

burnbay:
A little more context. If I set the `incremental` parameter in the `sql_table` resource and not applying it as a hint it seems to work. 

```python
_table_resource = sql_table(
            credentials=_source_engine,
            incremental=self.get_incremental_info(),  # Need to specify incremental here if we want to set primary_key = () to avoid flooding the state
...
)
```
But if I apply it as hints only I get the reported behaviour.

anuunchin:
Hi @shnhdan @burnbay, thanks for raising this! The resource's primary_key should be inherited by the incremental only when the incremental's primary_key is None (not explicitly set)  imo - this should be a pretty easy fix in `dlt/extract/resource.py` - ping me when you open a PR :)

Tests to be aware of:
- test_incremental_explicit_disable_unique_check
- test_override_primary_key_in_pipeline

[...]

PR #3679, opened 2026-02-25
Fix primary key assignment for incremental resources

shnhdan:
### Description
When you explicitly pass primary_key=() to Incremental to disable deduplication, it gets silently overwritten by the resource's own primary_key. This fixes that by only inheriting the resource's key when the incremental one was never set.

### Related Issues
https://github.com/dlt-hub/dlt/issues/3657

- Fixes #...
- Closes #...
- Resolves #...

### DIFF
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -218,7 +218,7 @@ class Incremental(
             "initial_unique_hashes_count": 0,
             "final_unique_hashes_count": 0,
         }
-        self._dedup_key_from_hints: bool = False if primary_key is not None else None
+        self._dedup_key_from_hints: Optional[bool] = False if primary_key is not None else None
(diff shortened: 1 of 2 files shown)


shnhdan:
@rudolfix @anuunchin could you take a look at this?

rudolfix:
@shnhdan the root cause of a problem is that current code does not delineate `primary_key` hints passed via resource and passed directly to incremental. I'm working on systematic fix and proper test case - what we had before was too brittle. This will take a few more days. I plan a patch release next week - and this fix will get in

shnhdan:
@rudolfix Thanks for working on a proper fix! Would it be possible to reference this PR in the commit or changelog?

rudolfix:
@shnhdan yes, we plan a release on Monday or Tuesday

PR #3686, opened 2026-02-26
control section opening from url

djudjuu:
when using claude skills I thought it would be nice if claude would send the user to a specific section of the dashboard in order to tell them about the pipeline state  or to answer questions the user had (instead of writing custom scripts). 
so this feature adds simple way to pre-open a section. 
I also tried scrolling it into view (e.g. by appending #section-name to the url, but because the elements are rendered dynamically based on the pipeline-arg from query it didnt work.


Travior (approved):
Works perfectly well. I also experimented with scrolling to a target, theoretically we can add a custom `<head>` with `js` that parses the hash and scrolls to the target, but because of how dynamically different widgets get loaded and potentially mutated this is a bit ugly. I think for now this works perfectly fine as is

PR #3688, opened 2026-02-27
docs: redirects for removed pages

djudjuu:
some pages were still searched for and gave users (aka crawlers) 404s. 
there is now a redirects array which the worker uses to send visitors to alternative pages that serve their intent.
also, regression test added as post-built check that warns if future builds break the redirect 
i decided to run it on the actually built files to be as close to the truth as possible (vs checking existence of .md/x files) and modelled it on the verify-llm-text.js script for consistency

### DIFF
--- a/docs/website/redirects.js
+++ b/docs/website/redirects.js
@@ -68,12 +68,6 @@ const REDIRECTS = [
         to: "/docs/general-usage/dataset-access/"
     },
 
-    // dlt+ redirect
-    {
-        from: "/docs/hub",
-        to: "/docs/hub/intro"
-    },
-
     // top-404 redirects
     {
         from: "/docs/dlt-ecosystem/llm-tooling/cursor-restapi",


VioletM:
My suggestion:
/docs/hub/reference — /docs/hub/intro 
/docs/general-usage/connectors — /docs/dlt-ecosystem/verified-sources/ 
/docs/api_reference/pipeline/configuration —  /docs/general-usage/credentials/
/docs/walkthroughs/grouping-resources — https://dlthub.com/docs/general-usage/source
/docs/general-usage/dlt — /docs/intro

docs/website/worker.ts
```diff
@@ -74,6 +74,72 @@ const REDIRECTS = [
...
+    {
+        from: "/docs/dlt-ecosystem/llm-tooling/cursor-restapi",
+        to: "/docs/dlt-ecosystem/llm-tooling/llm-native-workflow"
+    },
+    {
+        from: "/docs/general-usage/filesystem",
+        to: "/docs/dlt-ecosystem/verified-sources/filesystem/"
+    },
+    {
+        from: "/docs/hub/features/projects",
```
VioletM: I don't think we need it, you can remove this one

PR #3701, opened 2026-03-03
feat: make promote_options configurable for Arrow table concatenation…

### LINKED ISSUE #3697: feat: Make promote_options configurable for Arrow table concatenation
### Feature description

When a dlt resource yields multiple Arrow tables, `ArrowToParquetWriter.write_data()` concatenates them with `pyarrow.concat_tables(tables, promote_options="none")`. This requires identical schemas across all buffered tables. Any column type difference causes an ArrowInvalid crash.

dlt does not use its schema to cast Arrow types — `should_normalize_py_arrow_item_column()` returns False for all non-timestamp types. So even if the dlt schema knows a column should be double, an `int64` Arrow column passes through unchanged. The crash happens during extract, before any destination sees the data.

Example
```python
import dlt
import pandas as pd

@dlt.resource
def multi_file_resource():
    # File A: amount has decimals → pandas infers float64
    yield pd.DataFrame({"id": [1, 2], "amount": [967702.03, 123.45]})

    # File B: amount is all round numbers → pandas infers int64
    yield pd.DataFrame({"id": [3, 4], "amount": [18111, 25000]})

pipeline = dlt.pipeline(
    pipeline_name="reproduce_arrow_invalid",
    destination="duckdb",
    dataset_name="test",
)

pipeline.run(multi_file_resource())
# ArrowInvalid: Schema at index 1 was different:
#   amount: double
# vs
#   amount: int64
```

Per this [doc](https://arrow.apache.org/docs/python/generated/pyarrow.concat_tables.html#pyarrow-concat-tables), it seems to make sense to allow the `promote_options` to be configurable, in order to use  `promote_options=default` for example

I'm willing to submit a PR for this

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

Reading in multiple dfs within the same pipeline run, where same col might have different but compatible schema. 

### Proposed solution

Proposal:

Make promote_options configurable on BufferedDataWriterConfiguration.

```
# config.toml — the concat happens during extract
[extract.data_writer]
arrow_concat_promote_options = "default"

# Or scoped to a specific pipeline
[my_pipeline.extract.data_writer]
arrow_concat_promote_options = "default"
```

### Related issues

_No response_

AyushPatel101:
### Description

Add arrow_concat_promote_options to BufferedDataWriterConfiguration so pyarrow.concat_tables() can use "default" or "permissive" instead of the hardcoded "none". This lets pipeline authors opt in to automatic type promotion when yielding multiple Arrow tables / Pandas DataFrames with slightly different inferred types.

### Related Issues
- Closes #3697

### DIFF
--- a/tests/libs/test_parquet_writer.py
+++ b/tests/libs/test_parquet_writer.py
@@ -391,7 +391,7 @@ def test_arrow_parquet_row_group_size() -> None:
 
 def _caps_with_promote(promote_options: str) -> DestinationCapabilitiesContext:
     caps = DestinationCapabilitiesContext.generic_capabilities()
-    caps.parquet_format = ParquetFormatConfiguration(arrow_concat_promote_options=promote_options)
+    caps.parquet_format = ParquetFormatConfiguration(arrow_concat_promote_options=promote_options)  # type: ignore[arg-type]
     return caps
 
 
--- a/dlt/common/destination/configuration.py
+++ b/dlt/common/destination/configuration.py
@@ -37,7 +37,7 @@ class ParquetFormatConfiguration(BaseConfiguration):
     """When False, constant columns (like _dlt_load_id) will use regular arrays instead of
     dictionary-encoded arrays. Set to False for destinations using ADBC drivers that don't
     support dictionary types (e.g., MSSQL)."""
-    arrow_concat_promote_options: str = "none"
+    arrow_concat_promote_options: Literal["none", "default", "permissive"] = "none"
     """PyArrow concat_tables promote_options for Arrow table concatenation. "none" (default)
     requires identical schemas and enables zero-copy concat. "default" promotes within type
     families (e.g. int32 to int64). "permissive" promotes across families (e.g. int64 to double)."""


dlt/common/data_writers/buffered.py
```diff
@@ -31,6 +31,7 @@ class BufferedDataWriterConfiguration(BaseConfiguration):
         file_max_items: Optional[int] = None
         file_max_bytes: Optional[int] = None
         disable_compression: bool = False
+        arrow_concat_promote_options: str = "none"
```
rudolfix: please look at `ParquetFormatConfiguration` - we have it exactly for cases like this

docs/website/docs/dlt-ecosystem/file-formats/parquet.md
```diff
@@ -75,6 +76,7 @@ DATA_WRITER__FLAVOR
 DATA_WRITER__VERSION
 DATA_WRITER__DATA_PAGE_SIZE
 DATA_WRITER__TIMESTAMP_TIMEZONE
+DATA_WRITER__ARROW_CONCAT_PROMOTE_OPTIONS
```
rudolfix, line 79: `ParquetFormatConfiguration` should be compatible with this

tests/libs/test_parquet_writer.py
```diff
@@ -388,6 +388,107 @@ def get_id_() -> int:
...
+    table_float = pa.Table.from_pydict({"col1": pa.array([2.0], type=pa.float64())})
+    with pytest.raises((pa.lib.ArrowInvalid, pa.lib.ArrowTypeError)):
+        with get_writer(
+            ArrowToParquetWriter, buffer_max_items=10, arrow_concat_promote_options="default"
+        ) as writer:
+            writer.write_data_item(table_int, columns=c1)
+            writer.write_data_item(table_float, columns=c1)
+
+
+def test_arrow_concat_promote_options_permissive_null_column() -> None:
```
rudolfix: `test_arrow_concat_promote_options_permissive_null_column` and the null→string portion of `test_arrow_concat_promote_options_default` are very similar. Same for the int+float
   setup across `permissive` and `default` tests. These could be combined with `@pytest.mark.parametrize`

AyushPatel101:
Ah thank you, I didn't catch the `ParquetFormatConfiguration` code path. Let me know if there's anything else I can fix

AyushPatel101:
@rudolfix in case my updated changes were missed, lmk if theres anything else I can do here

Issue #944, opened 2024-02-07
Allow to backup and restore pipeline working directory


rudolfix:
**Background**
Backup and restore functions allow to work with snapshots of the pipeline working directory. Working directory contains full pipeline state and load packages and having such functions allows for example:
1. take a snapshot after extract step and use it to load the same extracted data into several destinations (we can call it "forking" pipeline)
2. take a snapshot after normalize step and use it to load data to several destinations of the same type (ie. two postgres databases) or different type (if file format is compatible: you can load same parquet files to duckdb, bigquery and snowflake)
3. take snapshot after load/run to save pipeline state (and schemas) for destinations that do not support state sync (ie. sinks)
4.. take snapshot after failed load to finalize it later
5. take snapshots after extract or normalize and keep them to be able to "replay" loading
6. use snapshot to pass packages between independently working extract, normalize and load steps ie. to scale things horizontally.

Note: snapshot must be restored in local file system, dlt does not support fsspec to keep its working folder (but it works on fuse)

**Requirements**
Implementation idea is to have helper functions that take pipeline instance, filesystem destination and snapshot name as an input, zip the pipeline state and load packages and pass it to a bucket.
Restore works in opposite way
Would be cool to have a context manager in which we can wrap pipeline run that would make sure that state is backed up and restored on run (sink support)
Would be cool to somehow chunk zipped files if too large
User should be able to tell helper function what to backup (only state, also data, also completed packages)
We should support pipeline fork on restore by allowing to change the pipeline name and destination

PoC: https://gist.github.com/rudolfix/ee6e16d8671f26ac4b9ffc915ad24b6e

PR #3706, opened 2026-03-04
Add dlt.sources.raw_export for exporting extracted data to object storage

### LINKED ISSUE #2947: Ability to load raw data into object storage
### Feature description

Add an option to export raw extracted data packages to a filesystem or object store (e.g., S3, GCS) before the normalization step processes them. This would enable users to preserve the original extracted data for reprocessing, auditing, or backup purposes.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

_No response_

### Proposed solution

Add a configuration option to export raw extracted packages before normalization. This would leverage dlt's existing filesystem destination, allowing users to configure a separate filesystem destination specifically for raw data export.

```
pipeline = dlt.pipeline(
    "my_pipeline",
    destination="redshift",
    staging="filesystem",           # Normalized data destination
    raw_data_export="filesystem",   # Raw data destination
    dataset_name="processed"
)
```

### Related issues

_No response_

AyushPatel101:
### Description

Add RawDataExporter, a new ItemTransform step that exports data to object storage (S3, GCS, Azure, local filesystem) during extraction. Sits after Incremental and LimitItem in the pipe chain (affinity 1.2), capturing exactly what gets loaded to the destination.

Supports all item types:
- FileItemDict: byte-perfect copy with deterministic paths
- dict: buffered as JSONL with load_id-scoped paths for run isolation
- Arrow/DataFrame: written as Parquet with load_id-scoped paths

Key features:
- Resource function argument pattern (like dlt.sources.incremental)
- export_only mode turns dlt into an incremental file copier
- dev_mode disabled by default, opt-in with _dev path isolation
- Credentials resolved from secrets.toml/env via @configspec
- teardown() lifecycle hook on ItemTransform for buffer flushing

### Related Issues

Closes #2947

### Additional Context

The original issue proposed raw_data_export="filesystem" as a pipeline-level staging destination. I didn't do that because: 

1. Wrong pipeline stage. Staging sits in Load, after normalization. Data is already column-renamed, type-cast, and flattened. Raw export needs pre-normalization data — it sits in Extract.

2. Wrong granularity. Staging applies to all resources. Raw export is per-resource, declared in the function signature like `dlt.sources.incremental`.

3. Wrong purpose. Staging exists for warehouse COPY INTO commands. It writes schema files, _dlt_loads tables, load packages. Raw export writes just data files — no metadata.

4. Already exists for that use case. If you want normalized data in S3, use destination="filesystem". Raw export solves a different problem: capturing exact extracted items for audit, backup, or reprocessing.

5. No existing hook point. Extract and Normalize have zero knowledge of staging. Adding raw export via staging would require forking the entire pipeline at the Extract/Normalize boundary. An ItemTransform step is a surgical, pattern-consistent addition.


rudolfix (changes requested):
thanks for this! however this is pretty complicated way to do that. IMO the pattern described here:
https://github.com/dlt-hub/dlt/issues/944
will be simpler and possibly more aligned with what users want

the idea here is that after each step transition we do a snapshot of the pipeline working directory and send it to a bucket. so this is purely file based. there are a few ideas of how user interface could look here in the ticket (ie. context manager with ie. configured filesystem destination that we can use to backup data). pls. let me know if you want to do this. first we'd need to define user interface and then agree on implementation plan

AyushPatel101:
Thanks for the feedback @rudolfix. I've been looking at the #944 pattern you described: purely file-based snapshots after each step transition. Here's a proposed user interface:

**Context manager (primary):**

```python
p = dlt.pipeline("my_pipeline", destination="bigquery")

# snapshot extracted packages to S3 after each step
with p.snapshot_to("s3://my-bucket/snapshots"):
    p.run(my_source())

# select which steps trigger a snapshot
with p.snapshot_to("s3://bucket/", steps=("extract", "normalize")):
    p.run(my_source())
```

**Explicit functions (for users calling extract/normalize/load separately):**

```python
from dlt.pipeline.snapshot import snapshot_packages, restore_packages

# snapshot
p.extract(my_source())
snapshot_packages(p, "s3://bucket/snapshots", step="extract")

# restore into a different pipeline (forking)
p2 = dlt.pipeline("fork", destination="snowflake")
restore_packages(p2, "s3://bucket/snapshots", step="extract", pipeline_name="my_pipeline")
p2.normalize()
p2.load()

# restore just state (for destinations without state sync)
from dlt.pipeline.snapshot import restore_state
restore_state(p, "s3://bucket/snapshots", load_id="latest")
```

**Remote layout** mirrors the load package structure:

```
s3://bucket/
  {pipeline_name}/
    {load_id}/
      extracted/          ← schema.json, load_package_state.json, new_jobs/...
      normalized/         ← if requested
      state.json          ← pipeline state at snapshot time
```

The `destination` param accepts a bucket URL string or a `Destination` instance, using the same resolution as the filesystem destination. No changes to the extract layer. This would cover both #2947 (raw data preservation) and #944 (state persistence, forking, horizontal scale).

I'm interested in taking this on. Let me know if this direction works or if you'd like any changes to the interface.

klaassens:
Thanks for working on this, raw_export would enable the use of dlthub for our use case as well. 

For filesystem sources this works well. For REST API sources the dict to JSONL doesn't preserve the raw data. RESTClient calls response.json() before the export step sees anything, and the exporter re-serializes the dicts into JSONL. Our policy requires data "stored exactly as it arrives, in its original format, without transformations".

Would be great to have a path where the exporter can write one file per HTTP response byte-for-byte.

AyushPatel101:
> Would be great to have a path where the exporter can write one file per HTTP response byte-for-byte.

CC @rudolfix to correct me if I'm wrong. Also agree with the above sentiment that is a very useful feature that I would love to use as soon as possible, so hoping to get a review here on proposed architecture above when possible.

I've been thinking about @klaassens' concern around preserving raw HTTP response data for REST API sources, and I wanted to share where I landed.

TLDR - I think this might be out of scope for this PR, and I don't think the snapshot mechanism would need to change to support it. Happy to discuss if others see it differently though.

## Why I think neither approach can solve it today

From what I can tell, the raw HTTP response bytes get discarded in `dlt/sources/helpers/rest_client/client.py` when `extract_response()` calls `response.json()`. Once that happens, the original wire format is gone. Both the `raw_export` ItemTransform approach and the `snapshot` approach only ever see parsed Python dicts, so it seems like neither one could preserve something that no longer exists by the time they run.

## How this could potentially work with the snapshot approach

I noticed that dlt already has `ImportFileMeta`, which allows sources to save raw bytes to a temp file and tell the extractor to import it directly. If the REST API source were enhanced to optionally capture `response.content` before parsing and yield it via `ImportFileMeta`, those raw bytes would land in the extraction working directory as-is (in `new_jobs/`). The snapshot mechanism would then pick them up and copy them to remote storage automatically, without needing any changes on the pipeline side.

## What I'm thinking for architecture

It seems like sources would be the right place to own the responsibility of preserving raw data, since they're the ones that actually have access to the raw bytes. Snapshot would then just copy whatever extraction produced to remote storage, including any raw files.

If that's right, this would keep it to one pipeline-level mechanism (snapshot) and the REST API raw response preservation could be tackled in a separate PR targeting the source layer. But I'd love to hear thoughts on whether that separation makes sense.

PR #3716, opened 2026-03-06
feat: add retrieve_file_metadata option to file reading functions

### LINKED ISSUE #3632: Add argument to include metadata from `FileItemDict.FileItem` in output from `filesystem.readers`
### Feature description

I would like to be able to do:
```python
filesystem | read_csv(include_file_metadata=True)
filesystem | read_csv_duckdb(include_file_metadata=True)
filesystem | read_parquet(include_file_metadata=True)
filesystem | read_jsonl(include_file_metadata=True)
```
And then include fields from `FileItemDict.FileItem` in the output. Not really any opinion if it should be possible to include only a subset of the fields. Assume that would make the solution more complex but maybe more useful?

Ie, these fields should be in the output from any of the readers:
```
class FileItem(TypedDict):
    """A DataItem representing a file"""

    file_url: str
    file_name: str
    relative_path: str
    mime_type: str
    encoding: NotRequired[str]
    modification_date: pendulum.DateTime
    size_in_bytes: int
    file_content: NotRequired[bytes]
```

### Are you a dlt user?

None

### Use case

This is very helpful for tracing/logging and can be used for deduplication etc. downstream.

### Proposed solution

_No response_

### Related issues

_No response_

pierre-monnet:
### Description
This pull request adds the ability to include file metadata (file name and modification date) in the output of all filesystem reader functions. The change is implemented by introducing a new `retrieve_file_metadata` parameter to each function. Comprehensive tests are also added to ensure this feature works for all supported file formats and reading modes.

### Related Issues

- Resolves #3632


dlt/sources/filesystem/readers.py
```diff
@@ -13,12 +13,16 @@
...
+    items: Iterable[FileItemDict],
+    chunksize: int = 10000,
+    retrieve_file_metadata: bool = False,
+    **pandas_kwargs: Any,
 ) -> Iterator[TDataItems]:
     """Reads csv file with Pandas chunk by chunk.
 
     Args:
         chunksize (int): Number of records to read in one chunk
+        retrieve_file_metadata (bool): Whether to include file metadata (file name and modification date) in the output.
```
anuunchin: Why only file name and modification date? I'm assuming because these are the ones that are most likely needed, but not including the other ones creates unnecessary fragmentation imo.

dlt/sources/filesystem/readers.py
```diff
@@ -120,6 +148,10 @@ def _read_csv_duckdb(
     for item in items:
         with item.open() as f:
             file_data = duckdb.from_csv_auto(f, **duckdb_kwargs)  # type: ignore
+            if retrieve_file_metadata:
+                file_data = file_data.select(
+                    f"*, '{item['file_name']}' AS file_name, CAST('{item['modification_date']}' AS TIMESTAMP) AS file_modification_date"
+                )
```
anuunchin: The `add_filename` should be incorporated in the new implementation, otherwise it's duplicate logic

Issue #3346, opened 2025-11-20
chore: move all tooling to `pyproject.toml`


zilto:
We can move `tox.ini`, `mypy.ini`, `pytest.ini`, etc. to `pyproject.toml`. The configuration should map one-to-one, but needs to adapt from `.ini` to `.toml`.

elviskahoro:
pytest.ini has already been moved! We only need to move the others: tox.ini and mypy.

zilto:
reopening because we still need to merge the flake8 config found in `tox.ini`. Flake8 doesn't support `pyproject.toml`. I suggest moving the config to Ruff (not an exact mapping) and deprecate `tox.ini`

Noone9029:
Hi! I’d like to pick up the remaining work on this issue.

I understand that pytest and mypy have already been migrated, and that the
remaining scope is the Flake8 configuration in tox.ini. I also saw that #3828
was closed with the work explicitly available for another contributor.

My plan is to:

- map every remaining Flake8 rule and plugin setting to its Ruff equivalent
- preserve the banned-import and per-file-ignore behavior
- identify any setting that does not have exact Ruff parity before removing it
- remove or deprecate tox.ini once lint behavior is covered
- update dependencies and the lockfile if required
- run the repository lint and common test gates

I’ll keep the PR limited to this tooling migration.

PR #3725, opened 2026-03-09
fix tox issue #3346

Afilimok:
Co-Authored-By: Oz <oz-agent@warp.dev>


Afilimok:
@zilto can you review

PR #3726, opened 2026-03-10
Fix versioning instructions in readme

Travior:
Poetry’s caret (^) requirement ensures that the leftmost non-zero version number remains fixed. The only exception is when the major version is 0, where it allows breaking changes in new minor versions.
When dlt was version 0.X, this meant only patch releases were accepted. Now—at version 1.0 or higher—it would allow all minor releases, which we do not recommend.
Similarly, the compatible release operator (~=) allows updates to the rightmost part of the version specifier. You must specify the patch version when installing; otherwise, all minor version bumps will be accepted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,9 +96,7 @@ You can find examples for various use cases in the [examples](docs/examples) fol
 * `minor` new features, sometimes automatic migrations
 * `patch` bug fixes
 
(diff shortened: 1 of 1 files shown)


Travior:
Merged in #3729

PR #3727, opened 2026-03-10
New dlthub high level structure

VioletM:


lis365b (dismissed):
I suggest we move https://docs-new-structure-dlthub-docs.services4758.workers.dev/docs/devel/general-usage/dashboard to Ingestion because the dashboard has more relevant info for inspecting the pipeline, not so much about data quality (or we split it into "inspecting the pipeline" vs. "inspecting data quality"

Otherwise looking good thanks for implementing it!

elviskahoro (approved):

PR #3732, opened 2026-03-11
Fix for infinite loop for bigquery job with internal error

### LINKED ISSUE #3722: Bigquery destination stuck on internalError in infinite loop
### dlt version

1.17.1

### Describe the problem

When a BigQuery load job fails with `internalError`, the `BigQueryLoadJob.run()` polling loop treats it as "job still running" and continues polling indefinitely. Since the job is already in a terminal state on BigQuery's side (`DONE` with error), every subsequent poll returns the same error, creating an **infinite loop** with no escape.

The problematic code in `dlt/destinations/impl/bigquery/bigquery.py`:

```python
elif reason in ["internalError"]:
    logger.warning(
        f"Got reason `{reason}` for job `{self._file_name}`, job considered still"
        f" running. ({self._bq_load_job.error_result})"
    )
    continue  # BUG: loops forever since BQ job is already DONE
```

Because BigQuery has already marked the job as `DONE` (with error), `done()` returns `True` with the same `internalError` on every iteration, resulting in an infinite loop. There is no timeout, max iteration count, or other escape mechanism.

In our case, this caused **20+ hours of data staleness** before it was detected. The pipeline process appeared alive (still writing `_dlt_pipeline_state`), but no data was being loaded for the affected table (`track`). BigQuery `INFORMATION_SCHEMA.JOBS` confirmed the load job was in `DONE` state with `internalError` at `2026-03-08T17:17:11Z`, while the dlt log at `17:17:27` still showed "job considered still running."

The `internalError` occurred during a period of high slot usage from concurrent CDC merge jobs on the same project, which is consistent with BigQuery's documentation that `internalError` is a transient, retryable server-side issue.

### Expected behavior

`internalError` should be treated as a **transient, retryable failure**, The `internalError` branch should raise `DatabaseTransientException` instead of `continue`, so `run_managed()` can catch it and set the job state to `"retry"`

### Steps to reproduce

1. Run a dlt pipeline loading data into BigQuery
2. Have concurrent heavy jobs running on the same BQ project to create resource pressure
3. A load job encounters a transient `internalError`
4. The job completes on BigQuery's side with state `DONE` and `error_result.reason = "internalError"`
5. dlt logs: `Got reason 'internalError' for job '<file>', job considered still running`
6. The pipeline hangs indefinitely — no retry, no failure, no timeout

### Operating system

Linux

### Runtime environment

Airflow

### Python version

3.11

### dlt data source

custom source built in house

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

_No response_

aditypan:
### Description
Fix for infinite loop for bigquery job with internal error

### Related Issues

- Closes #3722


dlt/destinations/impl/bigquery/bigquery.py
```diff
@@ -116,11 +116,12 @@ def run(self) -> None:
                     )
                 )
             elif reason in ["internalError"]:
-                logger.warning(
-                    f"Got reason `{reason}` for job `{self._file_name}`, job considered still"
-                    f" running. ({self._bq_load_job.error_result})"
+                raise DatabaseTransientException(
+                    Exception(
+                        f"Bigquery Job failed with internal error, reason: `{reason}`. "
```
anuunchin: The job here refers to a dlt job file, let's include the self._fila_name reference

anuunchin: On that note, though, why don't we just remove this elif branch, since it's handled in the else?

tests/load/bigquery/test_bigquery_load_job.py
anuunchin, line 1: We should parametrize this test and put it in `test_bigquery_client.py`

anuunchin (changes requested):
Thanks for the PR!

aditypan:
@anuunchin Have updated the PR as per the comments. Kindly take a look.

rudolfix (approved):
I did a small code clean up - we are good to go

PR #3735, opened 2026-03-12
Fixed the mapping of timezone in sql timestamp datatype

### LINKED ISSUE #3730: Oracle DATE columns receive incorrect timezone from `sqla_col_to_column_schema`
### dlt version

1.23.0

### Describe the problem

In the `sqla_col_to_column_schema` function in schema_types.py, it performs the following check when `reflection_level` is either "full" or "full_with_precision".

```python
[...]
elif isinstance(sql_t, sqltypes.DateTime):
        col["data_type"] = "timestamp"
        # special handling for MSSQL
        col["timezone"] = sql_t.timezone or sql_t.__visit_name__ in ("DATETIMEOFFSET")
[..]
```

However, the `("DATETIMEOFFSET")` is not a tuple, since it's missing a `,`. Python will instead perform a substring match on DATETIMEOFFSET. This means when an Oracle DATE column with a `sql_t.timezone` of False and a `sql_t.__visit_name__`  of "DATE" is passed through this logic, it sets the `col["timezone"]` to be true because DATE is a substring of DATETIMEOFFSET.

The effect of this is that timezone naive Oracle DATE columns will have a timezone incorrectly attached to them in the dlt table schema. 

### Expected behavior

When reading an Oracle DATE column with no timezone information and having full or full_with_precision reflection set in the call to `sql_table` , it should use the value of sql_t.timezone (which is False), set the col["timezone"] to be False and have the final column end up being a datetime with no timezone. 

### Steps to reproduce

I have made a script to show off the behaviour.

```python
from sqlalchemy import Column
from sqlalchemy.dialects import oracle

from dlt.sources.sql_database.schema_types import sqla_col_to_column_schema

col = Column("d", oracle.DATE()) 
schema = sqla_col_to_column_schema(col, reflection_level="full")

print("visit_name:", col.type.__visit_name__)      # DATE
print("timezone attr:", col.type.timezone)         # False
print("schema timezone:", schema.get("timezone"))  # currently True (bug)
```

### Operating system

Linux

### Runtime environment

Local

### Python version

3.13

### dlt data source

Oracle

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

I don't have the full understanding of why this workaround was first implemented, but from looking at the docs for MSSQL types for date and time here https://learn.microsoft.com/en-us/sql/t-sql/data-types/data-types-transact-sql?view=sql-server-ver17 , it would appear that `datetimeoffset` is the only data type for date/time with timezone information. This leads me to believe that the substring matching was not intended.

Perhaps sql_t.timezone was incorrect for this field, and thus a workaround was needed?

As for fixes, I would image something where it gets set to a Tuple, list or a set.

```python
# Set DATETIMEOFFSET as a Tuple
col["timezone"] = sql_t.timezone or sql_t.__visit_name__ in ("DATETIMEOFFSET",)
```

or

```python
# Set DATETIMEOFFSET as a list
col["timezone"] = sql_t.timezone or sql_t.__visit_name__ in ["DATETIMEOFFSET"]
```

aditypan:
### Description
The changes fixes the timezone to be set correctly incase of timestamp / datetime column datatype.

### Related Issues

- Fixes #3730

### DIFF
--- a/dlt/sources/sql_database/schema_types.py
+++ b/dlt/sources/sql_database/schema_types.py
@@ -164,7 +164,7 @@ def sqla_col_to_column_schema(
     elif isinstance(sql_t, sqltypes.DateTime):
         col["data_type"] = "timestamp"
         # special handling for MSSQL
-        col["timezone"] = sql_t.timezone or sql_t.__visit_name__ in ("DATETIMEOFFSET")
+        col["timezone"] = sql_t.timezone or sql_t.__visit_name__ in ("DATETIMEOFFSET",)
     elif isinstance(sql_t, sqltypes.Date):
         col["data_type"] = "date"
     elif isinstance(sql_t, sqltypes.Time):
--- a/tests/sources/sql_database/test_schema_types.py
+++ b/tests/sources/sql_database/test_schema_types.py
@@ -1,6 +1,7 @@
 import pytest
 import sqlalchemy as sa
-from sqlalchemy.dialects.mssql import UNIQUEIDENTIFIER
+from sqlalchemy.dialects.mssql import DATETIMEOFFSET, UNIQUEIDENTIFIER
+from sqlalchemy.dialects.oracle import DATE as OracleDATE
 from sqlalchemy.dialects.postgresql import UUID
 
 from dlt.sources.sql_database.schema_types import (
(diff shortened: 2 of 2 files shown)


rudolfix (approved):
LGTM! thx for finding this

kien-truong-pguru:
This is actually a breaking change if a `datetime` column is used as the incremental cursor:

* The cursor will store a datetime-aware value, due to previous loads returned datetime-aware values
* The new data will have a datetime-naive value after this fix

PR #3736, opened 2026-03-12
chore: switch lancedb example from spotify to podcastindex

Travior:
Switch API provider from Spotify to podcast index to resolve failing example


rudolfix (approved):
thanks for fixing this!

PR #3743, opened 2026-03-14
fix: support explicit non-utf8 encoding in filesystem read_csv

### LINKED ISSUE #2945: Add support for csv files using different encoding than utf-8 in read_csv
### Feature description

I was working on a pipeline that reads csv files from SFTP server. I followed [dlt's guide](https://dlthub.com/docs/dlt-ecosystem/verified-sources/filesystem/basic#1-initialize-a-filesystem-resource) for using filesystem resource, ran pipeline and got a UnicodeDecodeError (`'utf-8' codec can't decode byte 0xff in position 0: invalid start byte`).

After downloading some files manually to check what's going on it turned out that they're using utf-16 encoding. However, passing `encoding='utf-16'` to `read_csv` didn't help, this error kept happening. This encoding would be passed to pandas' `read_csv` function, but it wouldn't have any effect.

[Some](https://stackoverflow.com/questions/54078205/python-paramiko-utf-8-error-when-trying-to-stream-file-from-sftp-server) [people](https://stackoverflow.com/questions/68530700/reading-csv-file-into-pandas-from-sftp-server-via-paramiko-fails-with-utf-8-c) had very similar problem and paramiko seems to be behind this, because it assumes that everything uses utf-8 encoding.

In the end I bypassed this by writing function very similar to dlt's [_read_csv](https://github.com/dlt-hub/dlt/blob/master/dlt/sources/filesystem/readers.py#L13), but:
- I initialized my own SFTP client from paramiko that I used to open csv files
- opened csv file was wrapped using `TextIOWrapper(csv_file, encoding=encoding)` (encoding was passed to this function)

This works, but unlike other cases (where I'm downloading zip files from SFTP server, for example) I can't use `filesystem` resource directly (and instead use general `dlt.resource`) which I'd like to do.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

I'm trying to read csv files from SFTP server that use different encoding than utf-8. I'm receiving following error: `'utf-8' codec can't decode byte 0xff in position 0: invalid start byte`.

### Proposed solution

1. `read_csv` could have configurable encoding.
2. inside `read_csv` function encoding would be taken into account before calling `pd.read_csv` - encoding passed there seems to have no effect.

### Related issues

_No response_

biefan:
## Summary
- update `_read_csv` to honor explicit `encoding` at file-open level
- keep existing behavior for default reads (`rb`) when no encoding is provided
- map `encoding_errors` to file-open `errors` when decoding at open-time
- add regression test for reading UTF-16 CSV via `_read_csv(..., encoding="utf-16")`

## Why
For some remote file systems (notably SFTP/paramiko stacks), decoding can happen before pandas applies `encoding=...`. Opening with explicit text encoding avoids premature UTF-8 decoding failures.

## Validation
- `uv run --group dev --group pipeline python -m pytest -q tests/sources/filesystem/test_readers.py -k "read_csv and not duckdb"`

Fixes #2945

### DIFF
--- a/dlt/sources/filesystem/readers.py
+++ b/dlt/sources/filesystem/readers.py
@@ -27,11 +27,21 @@ def _read_csv(
 
     # apply defaults to pandas kwargs
     kwargs = {**{"header": "infer", "chunksize": chunksize}, **pandas_kwargs}
+    # For some remote file systems (for example, sftp/paramiko), decoding may happen before
+    # pandas gets a chance to apply `encoding=...`. If encoding is explicitly provided,
+    # decode at file-open level and let pandas consume text directly.
+    open_mode = "rb"
+    open_kwargs = {}
+    if "encoding" in kwargs:
+        open_mode = "rt"
+        open_kwargs["encoding"] = kwargs.pop("encoding")
+        if "encoding_errors" in kwargs:
+            open_kwargs["errors"] = kwargs.pop("encoding_errors")
 
     for file_obj in items:
         # Here we use pandas chunksize to read the file in chunks and avoid loading the whole file
         # in memory.
-        with file_obj.open() as file:
+        with file_obj.open(mode=open_mode, **open_kwargs) as file:
             for df in pd.read_csv(file, **kwargs):
                 yield df.to_dict(orient="records")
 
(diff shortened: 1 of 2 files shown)


rudolfix (approved):
LGTM! thanks for fixing this

PR #3759, opened 2026-03-19
feat(snowflake): Extend Snowflake query tags beyond load jobs

### LINKED ISSUE #3717: Query_tag not applying to all SQL pipeline statements
### dlt version

dlt 1.21.0

### Describe the problem

Hello,

We are have Snowflake as a destination in our data pipelines and use the "query_tag" parameter to identify which pipeline executed which SQL instructions. However, not all statements sent by the DLT pipeline receive it. In the first screenshot it can be seen that line 417, the truncate statement that clears dataset on "replace" write dispositions, comes before the query_tag change (query tag is shown in the right-most column). It uses a different session ID than the put+copy into statements, so another "alter session" should be implemented here.

Image

Also, the "insert into _dlt_loads" statement is also excluded from receiving the query tag change, as it uses a different session ID than the previous statements too.

Image

### Expected behavior

The same query tag should be shown in all statements executed during a pipeline run.

### Steps to reproduce

Define a pipeline with Snowflake destination and query_tag parameter 

Image

Then run the pipeline. "insert into _dlt_loads" statement on snowflake will not be tagged. If write_disposition is set to replace, the truncate statement will not be tagged either.

### Operating system

Linux

### Runtime environment

Docker, Docker Compose

### Python version

3.13

### dlt data source

Anything works. It can be tested with

@dlt.resource(name="contacts_0", write_disposition="replace", primary_key="id")
def hubspot_resource():
    for j in range(15):
        yield {
            "id": j,
            "email": f"user{j}@company.com",
        }

### dlt destination

Snowflake

### Other deployment details

_No response_

### Additional information

_No response_

Travior:
- Expand Snowflake query tagging from load jobs to broader dlt operations, including storage setup, schema/state reads, schema updates, load execution, load completion, and table drops
- Generalize query tag payloads by renaming TJobQueryTags to TQueryTags, adding an operation field, and keeping a compatibility export for existing imports
- Update docs
- Update tests

Closes #3717

### DIFF
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -1,4 +1,3 @@
-from dlt.common.storages.load_package import load_package_state, CurrentLoadPackageStateNotAvailable
 import os
 from abc import abstractmethod
 import base64
@@ -45,7 +44,7 @@ from dlt.common.schema.utils import (
 )
 from dlt.common.utils import read_dialect_and_sql
 from dlt.common.storages import FileStorage
-from dlt.common.storages.load_package import LoadJobInfo, ParsedLoadJobFileName
+from dlt.common.storages.load_package import LoadJobInfo, ParsedLoadJobFileName, load_package_state, CurrentLoadPackageStateNotAvailable
 from dlt.common.schema import TColumnSchema, Schema, TTableSchemaColumns, TSchemaTables
 from dlt.common.schema import TColumnHint
 from dlt.common.destination.client import (
--- a/dlt/destinations/job_client_impl.py
+++ b/dlt/destinations/job_client_impl.py
@@ -1,3 +1,4 @@
+from dlt.common.storages.load_package import load_package_state, CurrentLoadPackageStateNotAvailable
 import os
 from abc import abstractmethod
 import base64
@@ -920,19 +921,31 @@ WHERE """
     ) -> None:
         from dlt.common.pipeline import current_pipeline
 
+        if not load_id:
+            try:
+                load_id = load_package_state()["load_id"]
+            except CurrentLoadPackageStateNotAvailable:
+                load_id = ""
+
         pipeline = current_pipeline()
         pipeline_name = pipeline.pipeline_name if pipeline else ""
-        table_name = table["name"] if table else ""
+        if table:
+            table_name = table["name"] if table else ""
+            resource = (
+                get_inherited_table_hint(
+                    self.schema.tables, table_name, "resource", allow_none=True
+                )
+                or ""
+            )
+        else:
+            table_name = ""
+            resource = ""
+
         self.sql_client.set_query_tags(
             {
                 "operation": operation,
                 "source": self.schema.name,
-                "resource": (
-                    get_inherited_table_hint(
-                        self.schema.tables, table_name, "resource", allow_none=True
-                    )
-                    or ""
-                ),
+                "resource": resource,
                 "table": table_name,
                 "load_id": load_id,
                 "pipeline_name": pipeline_name,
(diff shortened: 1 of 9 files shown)


dlt/destinations/sql_client.py
```diff
@@ -41,14 +41,18 @@
...
 
     source: str
     resource: str
     table: str
     load_id: str
     pipeline_name: str
+    operation: str
+
+# compatibility export, query tags are no longer job scoped
+TJobQueryTags = TQueryTags
```
burnash, line 58: Aliasing it directly to TQueryTags now makes operation required as well, so external code that may already import TJobQueryTags and construct the old 5-field payload most likely starts failing mypy/pyright even though the comment above describes it as compatible. If we want to preserve that guarantee, make operation optional on the new type.

dlt/destinations/job_client_impl.py
```diff
@@ -290,6 +290,7 @@ def drop_storage(self) -> None:
                 self.sql_client.drop_dataset()
 
     def initialize_storage(self, truncate_tables: Iterable[str] = None) -> None:
+        self._set_query_tags(operation="prepare_storage")
```
burnash, line 293: Good, but there's another potential improvement: here it'll carry load_id as an empty string, so TRUNCATE running below defaults load_id to "". The initial issue asks for pipeline name propagation, however having load_id gives true per-run traceability. Same for `update_stored_schema()`, and `drop_tables()` below.

I think one option to get load_id here is to thread it from initialize_package, but there might be other options.

tests/load/pipeline/test_snowflake_pipeline.py
```diff
@@ -89,12 +89,26 @@ def test_snowflake_query_tagging(
...
 
+    operations = {call.args[1]["operation"] for call in set_query_tags_spy.call_args_list}
+    assert {"prepare_storage", "update_stored_schema", "load", "complete_load"} <= operations
+
+    set_query_tags_spy.reset_mock()
+    pipeline._schema_storage.clear_storage()
+    pipeline.sync_destination()
+
+    operations = {call.args[1]["operation"] for call in set_query_tags_spy.call_args_list}
+    assert {"get_stored_state", "get_stored_schema"} <= operations
```
burnash: It's better to do exact check here to prevent silent regressions.

burnash (changes requested):
Good direction, please see my comments.

dlt/destinations/job_client_impl.py
```diff
@@ -907,24 +914,39 @@ def verify_schema(
...
+    def _set_query_tags(
+        self, operation: str, *, load_id: str = "", table: Optional[PreparedTableSchema] = None
+    ) -> None:
         from dlt.common.pipeline import current_pipeline
 
+        if not load_id:
+            try:
+                load_id = load_package_state()["load_id"]
+            except CurrentLoadPackageStateNotAvailable:
+                load_id = ""
```
Travior, line 927: @burnash do you think its fine to try to get the load_id from the load_package like this or should we thread it more explicitly in the call sites?

dlt/destinations/job_client_impl.py
```diff
@@ -1,3 +1,4 @@
+from dlt.common.storages.load_package import load_package_state, CurrentLoadPackageStateNotAvailable
```
burnash: minor: let's move the local imports below (next to other dlt.common.*)

dlt/destinations/job_client_impl.py
burnash: I like this idea. I was thinking of threading load id explicitly - but this is more elegant imo.

burnash (changes requested):
Looks good, let's fix the import issue and this is good to go.

edit: also pease merge devel, that would fix the lint failure

burnash (approved):
Looks very good!

rudolfix (approved):
LGTM!

PR #3760, opened 2026-03-19
(fix) avoid leaking PUA markers in nested fields

### LINKED ISSUE #3755: Bug in nested models when using pydantic
### dlt version

1.23.0

### Describe the problem

Since 1.23.0 we find extraneous data in nested models, on fields of "special" types. That is: `date`, `datetime`, `Decimal`. The extraneous data is an UTF character prefixing the real data (clearly from [here](https://github.com/dlt-hub/dlt/blob/f348b12a3af2c3e293d3093822a592ab89788b40/dlt/common/json/__init__.py#L80-L89)).
While trying to reproduce, seems like it's only when using Pydantic.

See the **Steps to reproduce** for more information. We rolled back to 1.22.2 and that works as expected.

### Expected behavior

No UTF markers appearing in exported data 😁 

### Steps to reproduce

Minimal script to reproduce:

```python
# /// script
# requires-python = ">=3.14"
# dependencies = [
#     "dlt[duckdb]",
#     "pyarrow",
#     "pydantic",
# ]
# ///

from datetime import date
import dlt
from pydantic import BaseModel

print(f"dlt version: {dlt.__version__}")


class NestedModel(BaseModel):
    date_field: date


class UserModel(BaseModel):
    id: int
    date_field: date
    nested: NestedModel


@dlt.resource(
    table_name="users",
    max_table_nesting=0,
    write_disposition="replace",
    primary_key="id",
    columns=UserModel,
)
def get_users():
    return [
        {
            "id": 1,
            "date_field": date(2026, 1, 1),
            "nested": {
                "date_field": date(2026, 1, 1),
            },
        },
    ]


pipeline = dlt.pipeline(
    pipeline_name="quick_start",
    destination="duckdb",
    dataset_name="mydata",
)
load_info = pipeline.run(get_users)


print(pipeline.dataset().users.limit(10).arrow())
```

The result is:

```
$ uv run dlt_test.py 
dlt version: 1.23.0
pyarrow.Table
id: int64
date_field: date32[day]
nested: string
_dlt_load_id: string
_dlt_id: string
----
id: [[1]]
date_field: [[2026-01-01]]
nested: [["{"date_field":"2026-01-01"}"]]
_dlt_load_id: [["1773763966.361013"]]
_dlt_id: [["wTSCwuNFTXeyQQ"]]
```

In particular: `nested: [["{"date_field":"2026-01-01"}"]]` contains ``.

### Operating system

Linux

### Runtime environment

Local

### Python version

3.13

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

serl:
### Description

Some PUA markers leak in nested models when using Pydantic.
This should fix that.

Note that I'm not familiar with the codebase, Claude Opus 4.6 did most of the work.
That said, the test looks solid to me and it doesn't pass on `devel`, while it passes with the proposed fix.

### Related Issues

- Fixes #3755

### DIFF
--- a/dlt/common/json/__init__.py
+++ b/dlt/common/json/__init__.py
@@ -16,7 +16,7 @@ from dlt.common.exceptions import TypeErrorWithKnownTypes
 from dlt.common.pendulum import pendulum
 from dlt.common.arithmetics import Decimal
 from dlt.common.wei import Wei
-from dlt.common.utils import map_nested_values_in_place
+from dlt.common.utils import map_nested_values_in_place  # noqa: F401
 from dlt.common.libs.hexbytes import HexBytes
 
 TPuaDecoders = List[Callable[[Any], Any]]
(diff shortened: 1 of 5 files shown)


rudolfix (approved):
LGTM! I added optimized pua decoder which was long in the plans

PR #3763, opened 2026-03-20
feat(ducklake): add metadata_schema option to ATTACH statement

### LINKED ISSUE #3764: feat(ducklake): allow configuring metadata_schema independently from ducklake_name
### Description

When using DuckLake with a SQL catalog such as PostgreSQL, the metadata schema may already exist under a schema name that does not match `ducklake_name`.

Today, `dlt` uses `ducklake_name` when generating the DuckLake `ATTACH` statement for SQL-based catalogs, which means `METADATA_SCHEMA` is effectively tied to the attach/database alias.

This works well when the DuckLake catalog is created by `dlt`, but it becomes limiting when a user wants to connect `dlt` to an existing DuckLake catalog whose metadata schema is different, for example `public` or another custom schema.

### Problem

A user may already have a DuckLake catalog in PostgreSQL with metadata stored in a schema name that is not equal to `ducklake_name`.

In that case, they may want to keep using:

- `ducklake_name` as the DuckLake attach alias / logical database name in `dlt`
- a different metadata schema name for the existing catalog

At the moment, those two concepts cannot be configured independently.

### Proposed solution

Keep the current `ducklake_name` option and behavior as-is, but add a new optional `metadata_schema` option that overrides the `METADATA_SCHEMA` value in the generated DuckLake `ATTACH` statement.

Behavior:

- if `metadata_schema` is not set, keep the current behavior
- if `metadata_schema` is set, use it for `METADATA_SCHEMA`
- continue using `ducklake_name` as the attach alias / logical DuckLake name

### Example

Current behavior:

```sql
ATTACH 'ducklake:postgres:...' AS my_ducklake (
  DATA_PATH 's3://bucket/path',
  METADATA_SCHEMA 'my_ducklake'
)
```

Proposed behavior:

```sql
ATTACH 'ducklake:postgres:...' AS my_ducklake (
  DATA_PATH 's3://bucket/path',
  METADATA_SCHEMA 'public'
)
```

### Why this is useful

This would make it easier to use `dlt` with pre-existing DuckLake catalogs without requiring users to rename or recreate metadata schemas.

### Additional context

Related issue in DuckLake: [duckdb/ducklake#817](https://github.com/duckdb/ducklake/issues/817)

DuckLake currently stores view definitions with the attached catalog alias embedded in the persisted SQL. As described in duckdb/ducklake#817, this suggests that the alias supplied in ATTACH ... AS <alias> can, in some cases, have implications beyond being a purely session-local convenience, particularly for portability and object resolution across sessions.

This provides additional context for the proposal here. In dlt, METADATA_SCHEMA is currently derived from ducklake_name for SQL-backed DuckLake catalogs. That behavior is a sensible default, especially when the catalog is created and managed entirely by dlt, because the logical DuckLake name and the metadata schema can be chosen together from the beginning.

At the same time, these refer to different layers of the system:

ducklake_name: the logical DuckLake alias used by dlt in the generated ATTACH statement
metadata_schema: the physical schema in the SQL catalog where DuckLake metadata is stored
When connecting dlt to an existing DuckLake catalog, the metadata schema may already be fixed, for example public or another custom schema. In those cases, it can be useful to configure metadata_schema independently, while still preserving the desired ducklake_name for the DuckLake attachment.

This becomes more relevant if catalog alias choice also matters for compatibility with persisted objects such as views. In that case, keeping ducklake_name and metadata_schema independently configurable allows dlt to adapt more easily to pre-existing environments without changing the default behavior for new setups.

The intent of this proposal is therefore not to change the current default model, but to add a small amount of flexibility for integrations where the DuckLake alias and the SQL metadata schema are determined separately.

### Related PR

- #3763

sangwookWoo:
### Description

Add `metadata_schema: Optional[str]` option to `DuckLakeCredentials`. When set, the generated DuckLake `ATTACH` statement uses the provided value for `METADATA_SCHEMA`, allowing the metadata schema to be configured independently from `ducklake_name`.

If omitted, `ducklake_name` is used as the default metadata schema to preserve the current behavior.

### Related Issues

- Closes #3764

### Additional Context

Configurable via Python API or environment variable `DESTINATION__DUCKLAKE__CREDENTIALS__METADATA_SCHEMA`.

```py
import dlt
from dlt.destinations.impl.ducklake.configuration import DuckLakeCredentials

destination = dlt.destinations.ducklake(
    credentials=DuckLakeCredentials(
        catalog="postgresql://loader:loader@localhost:5432/dlt_data",
        storage="s3://bucket/data",
        metadata_schema="public",
    ),
)
```

Test: `uv run --extra ducklake pytest tests/load/ducklake/test_ducklake_client.py -v -k metadata_schema`

### DIFF
--- a/dlt/destinations/impl/ducklake/configuration.py
+++ b/dlt/destinations/impl/ducklake/configuration.py
@@ -35,15 +35,22 @@ def _get_ducklake_capabilities() -> DestinationCapabilitiesContext:
 @configspec(init=False)
 class DuckLakeCredentials(DuckDbBaseCredentials):
     ducklake_name: str = DEFAULT_DUCKLAKE_NAME
+    metadata_schema: Optional[str] = None
     catalog: ConnectionStringCredentials = None
     # NOTE: consider moving to DuckLakeClientConfiguration so bucket_url is not a secret
     storage: FilesystemConfiguration = None
 
-    __config_gen_annotations__: ClassVar[list[str]] = ["ducklake_name", "catalog", "storage"]
+    __config_gen_annotations__: ClassVar[list[str]] = [
+        "ducklake_name",
+        "metadata_schema",
+        "catalog",
+        "storage",
+    ]
 
     def __init__(
         self,
         ducklake_name: str = DEFAULT_DUCKLAKE_NAME,
+        metadata_schema: Optional[str] = None,
         catalog: Union[str, ConnectionStringCredentials] = None,
         storage: Union[str, FilesystemConfiguration] = None,
     ) -> None:
@@ -55,6 +62,9 @@ class DuckLakeCredentials(DuckDbBaseCredentials):
                 This value is mainly used as ATTACH name for the ducklake database and
                 as names for catalog and storage files if not configured explicitly.
                 If omitted, ducklake name is derived from destination name or pipeline name.
+            metadata_schema: str, optional
+                Metadata schema to use for SQL-based catalogs. If omitted, defaults to
+                `ducklake_name`.
             catalog: Either a connection string (for example,
                 "sqlite:///catalog.sqlite", "duckdb:///catalog.duckdb",
                 or "postgres://loader:loader@localhost:5432/dlt_data") or a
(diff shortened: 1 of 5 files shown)


zilto (approved):
LGTM! Thanks for your contribution and adding tests + docs :)

PR #3767, opened 2026-03-22
(fix) replays state transitions after crashed to avoid inconsistent load package state

### LINKED ISSUE #2837: duplicate records with same _dlt_id
### dlt version

1.12.3

### Describe the problem

I'm replicating some tables from postgres to another postgres using DLT. I'm using incremental loading and I noticed that I've duplicate records in the destination. I noticed duplicate records where the records have same `_dlt_load_id` and `_dlt_id`. I also noticed other cases where, only one of the duplicate records has a _dlt_load_id in _dlt_loads table. 

I had to rerun the dlt loads sometimes due to db connection error failures. But I was expecting dlt to handle connection errors and not load partial data. There are no records in `_dlt_loads` table with non-zero status even though I had connection errors from source. Since the destination connection wasn't dropped, shouldn't it have records with non-zero status.
Can someone help me understand how to avoid the issue and handle failure cases? Here is code.

Image

### Expected behavior

If I use incremental loading with merge strategy, there shouldn't be duplicate records.

### Steps to reproduce

try setting up postgres to postgres incremental loading and drop the source connection.

### Operating system

Linux

### Runtime environment

Docker, Docker Compose

### Python version

3.11

### dlt data source

Postgres

### dlt destination

Postgres

### Other deployment details

_No response_

### Additional information

_No response_

### LINKED ISSUE #3023: Duplicate records for a source with primary key and merge write disposition (duplicate _dlt_id as well)

rudolfix:
### Description
As discussed in issues below: there's a gap between a commit to database done by particular job and a state transition of that job. This PR writes a pending state transition marker to load package "immediately" after the commit minimizing such gaps.
"immediately" - some kind of gap will always exist. right now we close transaction scope and immediately write this file

the other part of this PR will deduplicate the child tables when merging from staging dataset. here we add some cost of the query to avoid very rare cases. @niyas-apron @jukiewiczm do you have on opinion if this is worthy to keep?

fixes #3023 
fixes #2837


rudolfix:
**How pending transitions work:**

When a load job commits its transaction in a worker thread, there's a gap before the main thread moves the job file from `started_jobs/` to `completed_jobs/`. If the process crashes in this window, the job is re-executed on resume, loading the same data to staging twice.

This PR closes that gap with a `.pending_transitions/` folder in each load package:

1. **On commit**: After `run()` succeeds (or fails terminally), an `_on_completed` callback writes a small JSON marker file to `.pending_transitions/{job_file_name}` containing the target state and error message. This happens in the worker thread, immediately after the DB transaction closes, before the semaphore release.

2. **On resume**: `resume_started_jobs` checks for pending transition markers. If found, it creates a real job instance via the destination client (preserving the correct job type and `create_followup_jobs` implementation), then calls `set_final_state()` to place it in the recorded terminal state without re-executing. The job flows through the normal `complete_jobs` path — followup jobs, metrics, file moves all work as usual.

3. **On completion**: The marker is consumed (deleted) only after the actual file move to `completed_jobs/` or `failed_jobs/` succeeds.

4. **Consistency check**: If a pending transition exists without a corresponding started job, an assertion fails — this catches code errors.

The second part adds `ROW_NUMBER() OVER (PARTITION BY _dlt_id)` deduplication for nested/child tables in the merge SQL, as a safety net against duplicate staging rows. This is gated by the existing `deduplicated` flag (`x-stage-data-deduplicated`) so users who opt out of dedup are not affected.

jukiewiczm:
> ### Description
> 
> As discussed in issues below: there's a gap between a commit to database done by particular job and a state transition of that job. This PR writes a pending state transition marker to load package "immediately" after the commit minimizing such gaps. "immediately" - some kind of gap will always exist. right now we close transaction scope and immediately write this file
> 
> the other part of this PR will deduplicate the child tables when merging from staging dataset. here we add some cost of the query to avoid very rare cases. @niyas-apron @jukiewiczm do you have on opinion if this is worthy to keep?
> 
> fixes #3023 fixes #2837

Thanks for taking care of those @rudolfix! I'm not knowledgeable enough within dlt internals to give a proper review, but yes, I would absolutely love to have those introduced to the codebase. It solves some of the problems I've been experiencing and having to circumvent in far from optimal ways. Additionally, it increases reliability greatly, which is IMO one of the top priorities for libraries like dlt. At the same time, I consider the additional overhead negligible, as dlt is clearly not after any kind of real-time speed, especially not speed at the cost of reliability. So nothing but goodies here IMO.

PR #3770, opened 2026-03-23
Multischema datasets

### LINKED ISSUE #3746: (feat) multischema datasets with local and foreign schemas
followup: https://github.com/dlt-hub/dlt/issues/3640

## Introduction

We've decided that users interact with **datasets**, not schemas, when querying or transforming data. If a pipeline loads a "hubspot" source with a "users" endpoint into a "warehouse_data" dataset, users query that "users" table via `SELECT * FROM warehouse_data.users`, not `SELECT * FROM hubspot.users` (where "hubspot" is the source/schema name).

This is technically incorrect: "warehouse_data" should be substituted for "hubspot" during the query binding phase. "hubspot" is a logical schema materialized in "warehouse_data". But we decided to hide that from the user.

Another example: we have a "hubspot" schema and a "_dlt_data_quality" sidecar schema where we keep quarantined data. Both live in the same dataset. If we want to select "bad records", we write:
```sql
SELECT * FROM warehouse_data.users AS u
JOIN warehouse_data._dq_data_quarantine AS q ON q._dlt_id = u._dlt_id
WHERE ...
```

The correct way would be:
```sql
SELECT * FROM hubspot.users AS u
JOIN _dlt_data_quality._dq_data_quarantine AS q ON q._dlt_id = u._dlt_id
WHERE ...
```

which is **resolved** to the upper query during the bind stage (see query lifecycle below).

This implementation assumes we want to keep this behavior.

## Query Lifecycle

See also: [query lifecycle comment on PR #3627](https://github.com/dlt-hub/dlt/pull/3627#issuecomment-3890128350)

A query goes through four stages before execution:

1. **Parse** (`Relation.sqlglot_expression`): user SQL string / SQLGlot AST / Ibis expression is parsed into a SQLGlot expression tree using logical (dlt schema) identifiers.

2. **Qualify** (`lineage.compute_columns_schema` → SQLGlot `qualify()`): resolves bare table and column references against the `SQLGlotSchema` (built from dlt schemas scoped under `dataset_name`). Expands `SELECT *`, attaches each column to its source table, and infers types via `annotate_types()`.

3. **Bind** (`bind_query()` in `dlt/common/libs/sqlglot.py`): maps logical identifiers to physical database objects:
   - **Table name expansion**: resolves to fully qualified paths via `sql_client.make_qualified_table_name_path` (e.g., ClickHouse `dataset.table` → `dataset___table`)
   - **Identifier case-folding**: applies destination-specific casing (`str.upper` for Snowflake, `str.lower` for most, `str` for case-sensitive)
   - **Identifier quoting**: quotes all identifiers for safe execution
   - **Alias preservation**: for case-folding destinations, adds aliases to preserve dlt schema column names (e.g., `SELECT "VALUE" AS "value"`)

4. **Render** (`query.sql(dialect=...)`): generates executable SQL in the destination's dialect.

The key insight is that the **qualify** step uses a `SQLGlotSchema` object built by `create_sqlglot_schema()`, which currently maps tables from a **single** dlt schema under `dataset_name`. To support multiple schemas, we only need to merge multiple dlt schemas into that same `SQLGlotSchema` — the qualify and bind steps work unchanged.

### Current signatures

```python
# dlt/dataset/lineage.py — builds the SQLGlotSchema used by qualify
def create_sqlglot_schema(
    schema: dlt.Schema,
    dataset_name: str,
    dialect: TSqlGlotDialect,
) -> SQLGlotSchema:

# dlt/dataset/lineage.py — runs qualify + annotate, returns column schema
def compute_columns_schema(
    expression: sge.Expression,
    dialect: TSqlGlotDialect,
    ...
) -> Tuple[TTableSchemaColumns, Optional[sge.Query]]:

# dlt/common/libs/sqlglot.py — binds logical names to physical destination objects
def bind_query(
    qualified_query: sge.Query,
    *,
    expand_table_name: Callable[[str], List[str]],
    casefold_identifier: Callable[[str], str],
) -> sge.Query:
```

```python
# dlt/dataset/dataset.py — current Dataset constructor (single schema)
class Dataset:
    def __init__(
        self,
        destination: TDestinationReferenceArg,
        dataset_name: str,
        schema: Union[dlt.Schema, str, None] = None,
    ) -> None:

# dlt/pipeline/pipeline.py — current pipeline.dataset() (passes single schema)
def dataset(self, schema: Union[Schema, str, None] = None) -> dlt.Dataset:
```

## Requirements

Allow `Dataset` to store multiple schemas — both local (within the same dataset) and foreign (from other datasets) — and enhance the qualify and bind steps to resolve multi-schema queries.

### Local schema refactor
* Allow many internal schemas to be passed to `dlt.Dataset`, reproducing the "pipeline" behavior where multiple sources load into one dataset.
* The first schema in the list is the "default schema", available via the `schema` property.
* Other schemas are available via the `schemas` property (a list of schemas).
* All local schemas are mapped to `self.dataset_name`.
* `create_sqlglot_schema` should accept multiple `dataset_name → Schema` mappings and merge them into a single `SQLGlotSchema`. Since all local schemas share the same `dataset_name`, tables from all schemas become visible during qualification.

### `pipeline.dataset()` behavior — **breaking change**
* Pipeline passes **all** its schemas when calling `pipeline.dataset()` (not just the default).
* Add a `Dataset.from_pipeline()` classmethod that creates a Dataset with all pipeline schemas.

### Foreign schemas
Foreign schemas will be added by explicit joins (separate ticket) that join across datasets.
* Foreign schemas should be exposed via a `foreign_schemas` property.
* Each foreign schema is mapped to its own `dataset_name`.
* Allow adding explicit foreign schemas via `_add_foreign_schema(dataset_name, schema)`.
* Note: currently `bind_query` binds to a single destination, so all datasets may share the same catalog prefix. Some destinations allow joins across catalogs. We should consider keeping a catalog mapped to each foreign schema as well.

### Other requirements
* Dataset remains a lazily populated object — schemas must be loaded on demand (as they are now).

## Suggested implementation

1. Implement multiple local schemas
2. Implement foreign schemas together with explicit join (separate ticket)

burnash:
Closes #3746

### DIFF
--- a/dlt/dataset/dataset.py
+++ b/dlt/dataset/dataset.py
@@ -434,26 +434,23 @@ class Dataset:
         return simple_repr("dlt.dataset", **without_none(kwargs))
 
     def __str__(self) -> str:
-        # obtain schema. this eventually loads it from destination (if only name was provided)
         _client = self.destination_client
-        # str(config) provides displayable physical location
         destination_info = f"{self._destination.destination_name}[{str(_client.config)}]"
-        # new schema is created if dataset is not found or schema.name is not materialized yet
-        if self.schema.is_new:
-            schema_info = "\nDataset is not available at the destination.\n"
-        else:
-            schema_names_str = "`, `".join(s.name for s in self.schemas)
-            schema_info = f"with tables in dlt schema `{schema_names_str}`:\n"
+        header = f"Dataset `{self.dataset_name}` at `{destination_info}` "
 
-        msg = f"Dataset `{self.dataset_name}` at `{destination_info}` {schema_info}"
-        tables: list[str] = []
-        for schema in self.schemas:
-            extend_list_deduplicated(tables, schema.data_table_names())
-        if tables:
-            msg += ", ".join(tables)
+        if self.schema.is_new:
+            return header + "\nDataset is not available at the destination.\n"
+
+        schemas = self.schemas
+        if len(schemas) == 1:
+            lines = [header + f"with schema `{schemas[0].name}`:"]
         else:
-            msg += "No data tables found in schema."
-        return msg
+            lines = [header + "with schemas:"]
+        for s in schemas:
+            tables = ", ".join(s.data_table_names()) or "No data tables found."
+            prefix = f"  {s.name}: " if len(schemas) > 1 else "  "
+            lines.append(prefix + tables)
+        return "\n".join(lines)
 
 
 def dataset(
--- a/dlt/dataset/dataset.py
+++ b/dlt/dataset/dataset.py
@@ -1,7 +1,20 @@
 from __future__ import annotations
 
 from types import TracebackType
-from typing import Any, Optional, Type, Union, TYPE_CHECKING, Literal, overload, Collection
+from typing import (
+    Any,
+    Collection,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Type,
+    Tuple,
+    Union,
+    TYPE_CHECKING,
+    Literal,
+    overload,
+)
 
 from sqlglot.schema import Schema as SQLGlotSchema
 import sqlglot.expressions as sge
@@ -14,8 +27,14 @@ from dlt.common.versioned_state import decompress_state
 from dlt.common.destination.reference import AnyDestination, TDestinationReferenceArg, Destination
 from dlt.common.destination.client import JobClientBase, SupportsOpenTables, WithStateSync
 from dlt.common.schema import Schema
+from dlt.common.schema import utils as schema_utils
 from dlt.common.typing import Self
-from dlt.common.schema.typing import C_DLT_LOAD_ID, C_DLT_LOADS_TABLE_LOAD_ID, LOADS_TABLE_NAME
+from dlt.common.schema.typing import (
+    C_DLT_LOAD_ID,
+    C_DLT_LOADS_TABLE_LOAD_ID,
+    LOADS_TABLE_NAME,
+    TTableSchema,
+)
 from dlt.common.utils import simple_repr, without_none
 from dlt.common.exceptions import ValueErrorWithKnownValues
 from dlt.destinations.sql_client import SqlClientBase, WithSqlClient
@@ -39,15 +58,24 @@ class Dataset:
         self,
         destination: TDestinationReferenceArg,
         dataset_name: str,
-        schema: Union[dlt.Schema, str, None] = None,
+        schema: Union[dlt.Schema, str, Sequence[dlt.Schema], None] = None,
     ) -> None:
         self._destination_reference = destination
         self._destination: AnyDestination = Destination.from_reference(destination)
         self._dataset_name = dataset_name
-        self._schema: Union[dlt.Schema, str, None] = schema
         self._pipeline_name: Optional[str] = None
-        """If _schema not provided, used to get default schema from state"""
-        # self._sqlglot_schema: SQLGlotSchema = None
+        """If _default_schema not provided, used to get default schema from state"""
+
+        if isinstance(schema, Sequence) and not isinstance(schema, str):
+            if not schema:
+                raise ValueError("schema sequence must not be empty")
+            self._default_schema: Union[dlt.Schema, str, None] = schema[0]
+            self._additional_schemas: Tuple[dlt.Schema, ...] = tuple(schema[1:])
+        else:
+            self._default_schema = schema
+            self._additional_schemas = ()
+
+        self._unified_schema: Optional[dlt.Schema] = None
         self._sql_client: SqlClientBase[Any] = None
         self._opened_sql_client: SqlClientBase[Any] = None
         self._table_client: SupportsOpenTables = None
@@ -71,17 +99,17 @@ class Dataset:
     def schema(self) -> dlt.Schema:
         """dlt schema associated with the dataset.
 
-        If no provided at dataset initialization, it is fetched from the destination. Fallbacks
+        If not provided at dataset initialization, it is fetched from the destination. Fallbacks
         to local dlt pipeline metadata.
         """
-        if isinstance(self._schema, dlt.Schema):
-            return self._schema
+        if isinstance(self._default_schema, dlt.Schema):
+            return self._default_schema
 
         maybe_schema: Optional[dlt.Schema] = None
 
-        if isinstance(self._schema, str):
+        if isinstance(self._default_schema, str):
             maybe_schema = _get_dataset_schema_from_destination_using_schema_name(
-                self, self._schema
+                self, self._default_schema
             )
 
         if not maybe_schema:
@@ -92,8 +120,47 @@ class Dataset:
             maybe_schema = dlt.Schema(self.dataset_name)
 
         assert isinstance(maybe_schema, dlt.Schema)
-        self._schema = maybe_schema
-        return self._schema
+        self._default_schema = maybe_schema
+        return self._default_schema
+
+    @property
+    def schemas(self) -> Tuple[dlt.Schema, ...]:
+        """All local schemas in this dataset, default schema first."""
+        return (self.schema, *self._additional_schemas)
+
+    def _get_unified_schema(self) -> dlt.Schema:
+        """Merged schema for query-time table resolution.
+
+        Single-schema case returns the default schema directly.
+        Multi-schema case clones the default schema and merges tables from
+        additional schemas (root tables first, then nested children).
+        """
+        if self._unified_schema is not None:
+            return self._unified_schema
+        if not self._additional_schemas:
+            self._unified_schema = self.schema
+        else:
+            unified = self.schema.clone()
+            for schema in self._additional_schemas:
+                for table in schema.tables.values():
+                    if not schema_utils.is_nested_table(table):
+                        for chain_table in schema_utils.get_nested_tables(
+                            schema._schema_tables, table["name"]
+                        ):
+                            unified.update_table(chain_table, normalize_identifiers=False)
+            self._unified_schema = unified
+        return self._unified_schema
+
+    def get_tables(self) -> Dict[str, TTableSchema]:
+        """Merged table mapping across all local schemas."""
+        return self._get_unified_schema().tables
+
+    def get_table_schema(self, table_name: str) -> TTableSchema:
+        """Look up a single table across all local schemas."""
+        tables = self.get_tables()
+        if table_name not in tables:
+            raise KeyError(f"Table '{table_name}' not found in dataset schemas")
+        return tables[table_name]
 
     @property
     def tables(self) -> list[str]:
@@ -103,8 +170,9 @@ class Dataset:
         execution, tables may exist on the destination, but will only appear on the dataset once
         `pipeline.run()` is done.
         """
-        # return only completed tables
-        return self.schema.data_table_names() + self.schema.dlt_table_names()
+        # return only completed tables from all schemas
+        unified = self._get_unified_schema()
+        return unified.data_table_names() + unified.dlt_table_names()
 
     def _ipython_key_completions_(self) -> list[str]:
         """Provide table names as completion suggestion in interactive environments."""
@@ -112,11 +180,11 @@ class Dataset:
 
     @property
     def sqlglot_schema(self) -> SQLGlotSchema:
-        """SQLGlot schema of the dataset derived from the dlt schema."""
+        """SQLGlot schema of the dataset derived from all dlt schemas."""
         # NOTE: no cache for now, it is probably more expensive to compute the current schema hash
         # to see wether this is stale than to compute a new sqlglot schema
         return lineage.create_sqlglot_schema(
-            self.schema, self.dataset_name, dialect=self.destination_dialect
+            self._get_unified_schema(), self.dataset_name, dialect=self.destination_dialect
         )
 
     @property
@@ -273,12 +341,14 @@ class Dataset:
             dlt.Relation: Relation for the query that computes the requested row count.
         """
 
+        unified = self._get_unified_schema()
+
         selected_tables = table_names or []
         if not selected_tables:
             if data_tables:
-                selected_tables += self.schema.data_table_names(seen_data_only=True)
+                selected_tables += unified.data_table_names(seen_data_only=True)
             if dlt_tables:
-                selected_tables += self.schema.dlt_table_names()
+                selected_tables += unified.dlt_table_names()
 
         # filter tables so only ones with dlt_load_id column are included
         dlt_load_id_col = None
@@ -287,7 +357,7 @@ class Dataset:
             selected_tables = [
                 table
                 for table in selected_tables
-                if dlt_load_id_col in self.schema.tables[table]["columns"].keys()
+                if dlt_load_id_col in unified.tables[table]["columns"].keys()
             ]
 
         union_all_expr: Optional[sge.Query] = None
@@ -390,22 +460,23 @@ class Dataset:
         if self.schema.is_new:
             schema_info = "\nDataset is not available at the destination.\n"
         else:
-            schema_info = f"with tables in dlt schema `{self.schema.name}`:\n"
+            schema_names_str = "`, `".join(s.name for s in self.schemas)
+            schema_info = f"with tables in dlt schema `{schema_names_str}`:\n"
 
         msg = f"Dataset `{self.dataset_name}` at `{destination_info}` {schema_info}"
-        if self.schema:
-            tables = [name for name in self.schema.data_table_names()]
-            if tables:
-                msg += ", ".join(tables)
-            else:
-                msg += "No data tables found in schema."
+        unified = self._get_unified_schema()
+        tables = unified.data_table_names()
+        if tables:
+            msg += ", ".join(tables)
+        else:
+            msg += "No data tables found in schema."
         return msg
 
 
 def dataset(
     destination: TDestinationReferenceArg,
     dataset_name: str,
-    schema: Union[Schema, str, None] = None,
+    schema: Union[Schema, str, Sequence[Schema], None] = None,
 ) -> Dataset:
     return Dataset(destination, dataset_name, schema)
 
@@ -490,13 +561,13 @@ def _get_dataset_schema_from_dataset_dlt_tables(
 def _get_load_ids(dataset: dlt.Dataset) -> list[str]:
     """Get a list of load ids associated with the dataset."""
     loads_table = dataset.loads_table()
+    schema_names = [s.name for s in dataset.schemas]
+    normalized_schema_col = dataset.schema.naming.normalize_identifier("schema_name")
+    normalized_load_id_col = dataset.schema.naming.normalize_identifier(C_DLT_LOADS_TABLE_LOAD_ID)
     query = (
-        loads_table
-        # need to filter out rare case where a dataset includes multiple `dlt.Schema`
-        # this mechanism is currently used by data quality metrics and checks
-        .where(dataset.schema.naming.normalize_identifier("schema_name"), "eq", dataset.schema.name)
-        .select(dataset.schema.naming.normalize_identifier(C_DLT_LOADS_TABLE_LOAD_ID))
-        .order_by(dataset.schema.naming.normalize_identifier(C_DLT_LOADS_TABLE_LOAD_ID), "asc")
+        loads_table.where(normalized_schema_col, "in", schema_names)
+        .select(normalized_load_id_col)
+        .order_by(normalized_load_id_col, "asc")
     )
     load_ids: list[str] = [load_id[0] for load_id in query.fetchall()]
     return load_ids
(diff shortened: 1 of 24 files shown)


dlt/dataset/dataset.py
```diff
@@ -92,8 +120,47 @@ def schema(self) -> dlt.Schema:
...
-        return self._schema
+        self._default_schema = maybe_schema
+        return self._default_schema
+
+    @property
+    def schemas(self) -> Tuple[dlt.Schema, ...]:
+        """All local schemas in this dataset, default schema first."""
+        return (self.schema, *self._additional_schemas)
+
+    def _get_unified_schema(self) -> dlt.Schema:
```
rudolfix: there's already a method to unify schemas on Schema object but the goal of this PR is to **not** unify schemas but to improve out lineage/query qualification methods to accept many schemas to generate lineage information. see comment here:
https://github.com/dlt-hub/dlt/issues/3746#issuecomment-4067254001

tldr;> I already tried schema unification for mcp/workbench but this is a dead end and I want to replace it with this: schemas are separate but we create an unified SQLGlot schema for query qualification. it will also be IMO easier

rudolfix (commented):
pls take a look at the original issue: we want to unify SQLGlot schema, not dlt schemas - many reasons for that, one of those being able to support foreign schemas with the same code. ping me for details

dlt/dataset/dataset.py
burnash: yes that was my question

dlt/dataset/lineage.py
```diff
@@ -26,40 +26,51 @@
 
 
 def create_sqlglot_schema(
-    schema: dlt.Schema,
+    schemas: Union[dlt.Schema, Sequence[dlt.Schema]],
```
rudolfix: if you so mapping dataset_name: List[Schema] - we'll cover foreign schemas as well. (and I think we discussed to do it now)

dlt/pipeline/pipeline.py
```diff
@@ -1891,7 +1891,9 @@ def dataset(self, schema: Union[Schema, str, None] = None) -> dlt.Dataset:
         """Returns a dataset object for querying the destination data.
 
         Args:
-            schema (Union[Schema, str, None]): Schema name or Schema object to use. If None, uses the default schema if set.
+            schema (Union[Schema, str, None]): Schema name or Schema object to use. If None,
```
rudolfix: here we should accept single schema or a list. now the question: do we do that via `schema` (backward compat) or we add `schemas` and deprecate `schema`?

dlt/dataset/dataset.py
```diff
@@ -39,19 +57,59 @@ def __init__(
...
-        """If _schema not provided, used to get default schema from state"""
-        # self._sqlglot_schema: SQLGlotSchema = None
+        """If set, used to resolve default schema from pipeline state"""
+
+        self._schemas: Dict[str, dlt.Schema] = {}
+        self._default_schema_name: Optional[str] = None
+        self._schema_arg: Optional[str] = None
+        """Deferred schema name for lazy resolution from destination"""
+
+        if isinstance(schema, Sequence) and not isinstance(schema, str):
```
rudolfix: Here I'd just validate and store _schema_arg and resolve later - on first access of any schema related property

dlt/dataset/dataset.py
```diff
@@ -39,19 +57,59 @@ def __init__(
...
+        elif isinstance(schema, str):
+            self._schema_arg = schema
+
         self._sql_client: SqlClientBase[Any] = None
         self._opened_sql_client: SqlClientBase[Any] = None
         self._table_client: SupportsOpenTables = None
 
+    @classmethod
+    def from_pipeline(
+        cls, pipeline: SupportsPipeline, dataset_name: Optional[str] = None
```
rudolfix: my idea for this function was - if I know pipeline name then I do not have to pass schema names. Now I see there's little value in it.
1. if I have instance I can do pipeline.dataset()
2. if I have name I can do dlt.pipeline(name, dataset, destination).dataset()
having this class method does not add anything, it is better to improve pipeline.dataset() below

dlt/dataset/dataset.py
```diff
@@ -69,31 +127,37 @@ def ibis(self, read_only: bool = False) -> IbisBackend:
 
     @property
     def schema(self) -> dlt.Schema:
-        """dlt schema associated with the dataset.
+        """Default dlt schema associated with the dataset.
```
rudolfix: we can simplify code here if any access to `schema` and `schemas` property fully resolves `self._schema_arg` to Schema objects. then we can keep use `self._default_schema` as marker that all schemas were resolved

dlt/dataset/dataset.py
```diff
@@ -69,31 +127,37 @@ def ibis(self, read_only: bool = False) -> IbisBackend:
...
-        if isinstance(self._schema, dlt.Schema):
-            return self._schema
+        if self._default_schema_name and self._default_schema_name in self._schemas:
+            return self._schemas[self._default_schema_name]
 
         maybe_schema: Optional[dlt.Schema] = None
 
-        if isinstance(self._schema, str):
+        if self._schema_arg:
             maybe_schema = _get_dataset_schema_from_destination_using_schema_name(
```
rudolfix: please review `_get_dataset_schema_from_dataset_dlt_tables` - it can use _pipeline_name to resolve all schemas. thanks to this we can have this sequence fully lazy:
```py
p = dlt.pipeline(name, dataset_name, destination)
ds = p.dataset()  # dataset with all schemas
ds.schema  # only now we access destination to get the schema
```
`_get_dataset_schema_from_dataset_dlt_tables` takes just the default schema but it is trivial to make it to take all

dlt/dataset/dataset.py
```diff
@@ -69,31 +127,37 @@ def ibis(self, read_only: bool = False) -> IbisBackend:
...
-        self._schema = maybe_schema
-        return self._schema
+        self._schemas[maybe_schema.name] = maybe_schema
+        self._default_schema_name = maybe_schema.name
+        return maybe_schema
+
+    @property
+    def schemas(self) -> Tuple[dlt.Schema, ...]:
+        """All local schemas in this dataset, default schema first."""
+        self.schema  # ensure default schema is resolved
```
rudolfix: this also works :)

dlt/dataset/dataset.py
```diff
@@ -103,20 +167,26 @@ def tables(self) -> list[str]:
         execution, tables may exist on the destination, but will only appear on the dataset once
         `pipeline.run()` is done.
         """
-        # return only completed tables
-        return self.schema.data_table_names() + self.schema.dlt_table_names()
+        seen: set[str] = set()
+        result: list[str] = []
+        for schema in self.schemas:
+            for name in schema.data_table_names() + schema.dlt_table_names():
```
rudolfix: you could just do set union with set1 | set2 etc instead of looping

dlt/dataset/dataset.py
```diff
@@ -275,20 +345,30 @@ def row_counts(
 
         selected_tables = table_names or []
         if not selected_tables:
+            seen: set[str] = set()
             if data_tables:
-                selected_tables += self.schema.data_table_names(seen_data_only=True)
+                for schema in self.schemas:
+                    for t in schema.data_table_names(seen_data_only=True):
```
rudolfix: as above: you could just do an union with `seen` - no reason to loop

dlt/dataset/dataset.py
```diff
@@ -275,20 +345,30 @@ def row_counts(
...
+                for schema in self.schemas:
+                    for t in schema.dlt_table_names():
+                        if t not in seen:
+                            seen.add(t)
+                            selected_tables.append(t)
 
         # filter tables so only ones with dlt_load_id column are included
         dlt_load_id_col = None
         if load_id:
             dlt_load_id_col = self.schema.naming.normalize_identifier(C_DLT_LOAD_ID)
```
rudolfix: this must be done for each schema separately

dlt/dataset/dataset.py
rudolfix, line 349: note: this reimplements filtering by `_dlt_load_id` available on the relation

dlt/dataset/dataset.py
```diff
@@ -490,13 +576,13 @@ def _get_dataset_schema_from_dataset_dlt_tables(
 def _get_load_ids(dataset: dlt.Dataset) -> list[str]:
```
rudolfix: if we have multiple schemas then those load_id getters are misleading (they always were but probably no one noticed)
we can discuss what to do here. my take:
- select load_ids for default schema
- take schema_name argument to select load_ids from other schema. IMO it does not make sense to select from multiple schemas - we do not even know to which schema given load id belongs
btw. schema corresponds to a single dlt source. and each source gets its separate load package. so this is fundamental mechanism not a trick

dlt/dataset/dataset.py
```diff
@@ -505,12 +591,12 @@ def _get_load_ids(dataset: dlt.Dataset) -> list[str]:
 def _get_latest_load_id(dataset: dlt.Dataset) -> Optional[str]:
```
rudolfix: same here

dlt/pipeline/pipeline.py
```diff
@@ -1906,6 +1908,7 @@ def dataset(self, schema: Union[Schema, str, None] = None) -> dlt.Dataset:
             )
 
         schema_name = None
+        _use_from_pipeline = False
```
rudolfix: just pass pipeline name as before... overall this function should not change much. we pass schema argument to dataset with similar resolution as below

dlt/pipeline/pipeline.py
```diff
@@ -1924,17 +1927,21 @@ def dataset(self, schema: Union[Schema, str, None] = None) -> dlt.Dataset:
...
+                schema = self.default_schema
 
         try:
-            dataset = dlt.dataset(
-                self._destination,
-                self.dataset_name,
-                schema=schema,
-            )
-            # allow dataset to lazily resolve schema from pipeline state
-            dataset._pipeline_name = self.pipeline_name
```
rudolfix, line 1937: currently pipeline name is used only if no schema was provided in any form. we must still support this path but load all schemas from the pipeline, not just one

rudolfix (changes requested):

dlt/dataset/dataset.py
burnash: Not sure I understand this. Could you elaborate? If the user passes actual Schema instances, there's nothing to "resolve": they're already materialized. The lazy path only matters for string/pipeline-name resolution. As far as I understand that's the case for schemas already loaded from local pipeline storage.

dlt/dataset/lineage.py
burnash: Done in 7e16d0b

dlt/pipeline/pipeline.py
burnash: I'd to keep `schema` and widen it:  backward compatible, Dataset.__init__ already accepts the union type. 

Done in 7e16d0b

dlt/dataset/dataset.py
burnash: Removed in 7e16d0b

dlt/dataset/dataset.py
burnash: Done in 7e16d0b

dlt/dataset/dataset.py
burnash: I was unsure if we need to order these with default schema tables first. Do we need to order tables?

dlt/dataset/dataset.py
burnash: Done in 7e16d0b

dlt/dataset/dataset.py
burnash: implemented

dlt/dataset/dataset.py
rudolfix: I refer to code organization. Here you are trying to partially "resolve" the input argument and I suggest that  you move all of that to `_resolve_schemas` and here just store 
```py
self._schema_arg = schema
```
all the fields that you store below, you store sometimes and you still need to preserve schema_arg

dlt/dataset/dataset.py
```diff
@@ -67,33 +100,47 @@ def ibis(self, read_only: bool = False) -> IbisBackend:
...
+        """Resolve all schemas on first access.
 
-        If no provided at dataset initialization, it is fetched from the destination. Fallbacks
-        to local dlt pipeline metadata.
+        Uses ``_default_schema_name`` presence in ``_schemas`` as a marker
+        that resolution already happened.
         """
-        if isinstance(self._schema, dlt.Schema):
-            return self._schema
+        if self._default_schema_name and self._default_schema_name in self._schemas:
```
rudolfix: can I have `_default_schema_name` set and not be present in `self._schemas`? does it happen in `__init__`?

dlt/dataset/dataset.py
```diff
@@ -275,27 +328,37 @@ def row_counts(
 
         selected_tables = table_names or []
         if not selected_tables:
+            seen: set[str] = set()
```
rudolfix: looks like we need util function that extends list without duplicates: maybe `extend_list_deduplicated` in utils.py would do?

dlt/dataset/dataset.py
```diff
@@ -275,27 +328,37 @@ def row_counts(
...
-                selected_tables += self.schema.dlt_table_names()
+                for schema in self.schemas:
+                    for t in schema.dlt_table_names():
+                        if t not in seen:
+                            seen.add(t)
+                            selected_tables.append(t)
 
         # filter tables so only ones with dlt_load_id column are included
-        dlt_load_id_col = None
+        # normalize per-schema since naming conventions may differ
```
rudolfix, line 336: true!

dlt/dataset/lineage.py
```diff
@@ -26,43 +26,52 @@
...
-    nested_schema = {dataset_name: sqlglot_schema}
+    Args:
+        schema_map: Mapping of dataset_name to dlt schemas. Local schemas
+            all map to the same dataset_name. Foreign schemas (future) map to
+            their own dataset_names.
+        dialect: SQLGlot dialect for the target destination.
+    """
+    nested_schema: Dict[str, Dict[str, Any]] = {}
+
+    for dataset_name, schemas in schema_map.items():
```
rudolfix, line 47: here I'd use `unify_schemas`: that will take care of merging tables and will reuse well tested code. there's still a limitation on not accepting schemas with naming convention that differ but that we can lift in this or separate PR.

having such schema - you can use the original code of this function with the unified schema

dlt/dataset/lineage.py
```diff
@@ -26,43 +26,52 @@
...
 ) -> SQLGlotSchema:
-    """Create an SQLGlot schema using a dlt Schema and the destination capabilities.
+    """Create an SQLGlot schema from dlt schemas mapped to dataset names.
 
-    The SQLGlot schema automatically scopes the tables to the `dataset_name`.
-    This can allow cross-dataset transformations on the same physical location.
-    No name translation nor case folding is performing. All identifiers correspond
-    to identifiers in dlt schema.
-    """
+    Multiple dlt schemas are kept separate at the data-model level but merged
```
rudolfix: please write better docstring. what "schema_map" is? what is the function of the key in its dict (it qualifies tables, pretty fundamental!)

rudolfix (changes requested):
looks pretty good! but code could be simpler. we could also use `unify_schemas` when generating sqlglot schema

dlt/dataset/dataset.py
burnash: Done in c7a0126

dlt/dataset/dataset.py
burnash: Fixed in c7a0126

dlt/dataset/lineage.py
burnash: Done in c7a0126
there's an issue with the diff naming convention indeed, stumbled upon this pr's tests.

dlt/dataset/lineage.py
burnash: updated the docstring

dlt/dataset/dataset.py
burnash: true. I think extend_list_deduplicated has a bug: it dedupes against the old list, but not within the incoming list itself. I've fixed it in 19ecc1ce78f4a675983712523b87085d86de1c3e 
lmk if the original behaviour is intended

dlt/destinations/sql_client.py
```diff
@@ -350,6 +357,13 @@ def sql_client_class(self) -> Type[SqlClientBase[TNativeConn]]:
         pass
 
 
+@runtime_checkable
```
rudolfix: this (and the above `set_schemas`) is not the best way to add many schemas to `WithTableScanner`. also: runtime `Protocol` are verified by duck typing check (Python checks if instance has methods of the protocol - it is not a class) so
1. it is very slow
2. it is always true in your case because base client has this method

I'd suggest to create abstract class `WithSchemas` here and add it to `WithTableScanners`. then you can use `WithSchemas` to check if given client supports `set_schema`.

dlt/destinations/impl/duckdb/sql_client.py
```diff
@@ -495,6 +497,7 @@ def __init__(
         )
         self.remote_client = remote_client
         self.schema = remote_client.schema
+        self._schemas: Dict[str, Schema] = {remote_client.schema.name: remote_client.schema}
```
rudolfix: let's just name this `schemas` - like in dataset

dlt/destinations/impl/duckdb/sql_client.py
```diff
@@ -537,36 +540,58 @@ def should_replace_view(self, view_name: str, table_schema: PreparedTableSchema)
...
+        self, view_name: str, table_schema: PreparedTableSchema, schema: Schema = None
+    ) -> None:
         pass
 
     @abstractmethod
     def can_create_view(self, table_schema: PreparedTableSchema) -> bool:
         """Tells if a view for a table `table_schema` can be created"""
         pass
 
+    def set_schemas(self, schemas: Sequence[Schema]) -> None:
```
rudolfix, line 566: good find, I forgot about table scanners. I wanted to make it a mixin class (not to derive from duckdb) then adding `set_schemas` to it would be trivial

dlt/dataset/dataset.py
```diff
@@ -421,7 +472,10 @@ def get_dataset_destination_client(dataset: dlt.Dataset) -> JobClientBase:
 def get_dataset_sql_client(dataset: dlt.Dataset) -> SqlClientBase[Any]:
     client = get_dataset_destination_client(dataset)
     if isinstance(client, WithSqlClient):
-        return client.sql_client
+        sql_client = client.sql_client
+        if isinstance(sql_client, SupportsMultiSchema):
```
rudolfix: this is always True

rudolfix (changes requested):
good find with `WithTableScanners` supporting just one schema! I suggest some improvements with instance check. I just looked at the test:
1. no test with schemas that have overlapping tables which are not identical (_dlt_tables are) but have non conflicting columns (different names or same types)
2. a test that shows column conflict that actually prevents schema unification (ie. same column name, different data type - that should fail)
3. schemas with different naming convention ie. sql naming convention case sensitive and insensitive. I'd like to let `unify_schema` to work with different naming conventions. for now test should expect `unify_schema` to fail

dlt/dataset/dataset.py
burnash: yes, forgot to remove `set_schemas` from SqlClientBase when moving to protocol

dlt/destinations/impl/duckdb/sql_client.py
burnash: Could you clarify if mixing refactor of WithTableScanners be part of this PR or a separate one? The coupling to duckdb is deep now.

dlt/destinations/sql_client.py
burnash: Implemented in 3b1c7b3

dlt/destinations/impl/duckdb/sql_client.py
burnash: changed in 3b1c7b3

dlt/dataset/dataset.py
burnash: fixed in 3b1c7b3

dlt/destinations/impl/duckdb/sql_client.py
rudolfix: not this PR, this is old tech debt not to be fixed now

rudolfix (changes requested):
there was still a problem with filesystem sql_client that I discovered by running test with conflicting user tables on it
- we need to UNION views that point to separate data locations
- we need to merge columns in views on the same data location
this is now done. create_view just returns sql - this allows manipulation of SQL to handle cases above.
I promoted a few tests so they run on all destinations incl. lance and delta/iceberg. this surfaced ibis() creation problem (schemas were not passed there). that I fixed

I think ticket is complete. docs still remain:
docs/website/docs/general-usage/dataset-access/dataset.md
should explain how we deal with multiple schemas in dataset. it could be NOTE admonition - we do not recommend having such datasets.

rudolfix (dismissed):
LGTM!

rudolfix (approved):
LGTM!

PR #3773, opened 2026-03-23
fix: check duckdb version to install lance extension

zilto:
On `duckdb` release 1.5, the `lance`  extension was promoted from community to built-in. This changed the required SQL command to install the extension. This broke the `dlt` features when using the latest `duckdb` dependencies. This wasn't caught by CI because it uses the locked `duckdb` version or lowest deps, not latest deps.

I didn't add tests because checking for different `duckdb` version is a bit annoying on CI. We would ideally need it, but the logic is rather simple here


anuunchin (approved):
🧠

PR #3775, opened 2026-03-23
Fix/3592 bypass non interactive flag prompts

### LINKED ISSUE #3592: The --non-interactive flag appears to default to “No” for prompt confirmations
### dlt version

1.20.0

### Describe the problem

We’re attempting to run dlt commands in a non-interactive environment (Airflow) and have run into unexpected behavior with confirmation prompts when the `--non-interactive` is used.

When running a command that includes a confirmation prompt, e.g. `dlt pipeline drop` with `--non-interactive`, dlt prints the planned operation but exits without performing the operation. This makes it difficult to automate operations (e.g. state refreshes, dropping tables) where interactive input is not possible.

### Expected behavior

Passing `--non-interactive` should allow the command to proceed and complete the requested operation.

When running a pipeline command with the `--non-interactive` option, 

the command:
* prints the tables/state that would be dropped
* assumes confirmation (`Y`) for the action
* proceeds with dropping the pipeline tables/state
* exits successfully

### Actual behavior

the command:
* prints the tables/state that would be dropped
* exits successfully
* does not perform the drop operation

### Steps to reproduce

Run `drop` pipeline command to drop all resources with the `non-interactive` option

```bash
uv run dlt --non-interactive pipeline {pipeline_name} drop --drop-all
```

### Operating system

Linux

### Runtime environment

Airflow

### Python version

3.13

### dlt data source

Issue is source-agnostic. Observed with verified sources (Zendesk, Chess) and REST API source

### dlt destination

Google BigQuery

### Other deployment details

_No response_

#### Questions

1. Is the current behavior of `--non-interactive` intentionally designed to default confirmations to “No”?

2. Is there a supported way to pass confirmation intent via flags (e.g. --yes, --assume-yes)?

kperkins96:
### Description

Fixes the `--non-interactive` flag to consistently accept all confirmations instead of using default values. Previously, when `--non-interactive` was used, CLI commands would still abort if a confirmation's default was `False`, making the flag unreliable for automation and CI/CD workflows.

**Changes:**
- Modified `NonInteractiveAction` in [dlt/_workspace/cli/_dlt.py](dlt/_workspace/cli/_dlt.py) to set `ALWAYS_CHOOSE_VALUE = True` instead of `ALWAYS_CHOOSE_DEFAULT = True`
- Updated user-facing message to clarify behavior: "Confirmations are automatically accepted"
- Added comprehensive test coverage in [tests/workspace/cli/test_non_interactive.py](tests/workspace/cli/test_non_interactive.py) for `drop`, `sync`, and `drop-pending-packages` commands
- Updated [tests/workspace/cli/conftest.py](tests/workspace/cli/conftest.py) to properly import test fixtures

**Behavior:**
- **Before**: `--non-interactive` would use the default answer for each prompt (could be `True` or `False`)
- **After**: `--non-interactive` always accepts confirmations (`True`), allowing commands to proceed without user interaction

This makes the flag reliable for CI/CD pipelines and automated scripts where prompts cannot be answered interactively.

### Related Issues

- Fixes #3592 

### Additional Context

**Testing:**
- Added [test_non_interactive.py](tests/workspace/cli/test_non_interactive.py) 
- New tests use `echo.always_choose(False, True)` context manager to simulate `--non-interactive` behavior
- Ran `make format && make lint` - all checks pass
- Ran `make test-workspace` - all tests pass

**Implementation notes:**
- The fix leverages the existing `ALWAYS_CHOOSE_VALUE` in [echo.py](dlt/_workspace/cli/echo.py)
- When set to `True`, `confirm()` and `prompt()` return `True` instead of prompting the user
- The `--non-interactive` flag only affects behavior when explicitly used - no changes to default CLI behavior

### DIFF
--- a/dlt/_workspace/cli/_dlt.py
+++ b/dlt/_workspace/cli/_dlt.py
@@ -68,11 +68,8 @@ class NonInteractiveAction(argparse.Action):
         values: Any,
         option_string: str = None,
     ) -> None:
-        fmt.ALWAYS_CHOOSE_DEFAULT = True
-        fmt.note(
-            "Non interactive mode. Default choices are automatically made for confirmations and"
-            " prompts."
-        )
+        fmt.ALWAYS_CHOOSE_VALUE = True
+        fmt.note("Non interactive mode. Confirmations are automatically accepted.")
 
 
 class DebugAction(argparse.Action):
--- a/dlt/_workspace/cli/_dlt.py
+++ b/dlt/_workspace/cli/_dlt.py
@@ -68,8 +68,34 @@ class NonInteractiveAction(argparse.Action):
         values: Any,
         option_string: str = None,
     ) -> None:
+        fmt.ALWAYS_CHOOSE_DEFAULT = True
+        fmt.note(
+            "Non interactive mode. Default choices are automatically made for confirmations and"
+            " prompts."
+        )
+
+
+class YesAction(argparse.Action):
+    def __init__(
+        self,
+        option_strings: Sequence[str],
+        dest: Any = argparse.SUPPRESS,
+        default: Any = argparse.SUPPRESS,
+        help: str = None,  # noqa
+    ) -> None:
+        super(YesAction, self).__init__(
+            option_strings=option_strings, dest=dest, default=default, nargs=0, help=help
+        )
+
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Any,
+        option_string: str = None,
+    ) -> None:
         fmt.ALWAYS_CHOOSE_VALUE = True
-        fmt.note("Non interactive mode. Confirmations are automatically accepted.")
+        fmt.note("Automatically accepting all confirmations.")
 
 
 class DebugAction(argparse.Action):
@@ -122,6 +148,12 @@ def _create_parser() -> Tuple[argparse.ArgumentParser, Dict[str, SupportsCliComm
             "Non interactive mode. Default choices are automatically made for confirmations and"
             " prompts."
         ),
+    )
+    parser.add_argument(
+        "--yes",
+        "-y",
+        action=YesAction,
+        help="Automatically accept all confirmations without prompting.",
     )
     parser.add_argument(
         "--debug",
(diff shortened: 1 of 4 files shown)


anuunchin (changes requested):
Thanks for the PR! 

Changing `--non-interactive` from "use defaults" to "always accept" is a breaking change. Anyone currently using --non-interactive in CI scripts may be relying on destructive commands being safely skipped - which makes sense imo

How about adding a separate --yes / -y flag that sets ALWAYS_CHOOSE_VALUE = True, while leaving  --non-interactive unchanged? That would address the reported issue and just be an additive change

kperkins96:
Added a new `--yes` / `-y` flag that automatically accepts all confirmations without prompting and preserved existing `--non-interactive` behavior.

**Changes:**
- Added new `YesAction` class in [dlt/_workspace/cli/_dlt.py](dlt/_workspace/cli/_dlt.py) that sets `ALWAYS_CHOOSE_VALUE = True`
- Added `--yes` / `-y` argument to the CLI parser
- `--non-interactive` remains unchanged (uses `ALWAYS_CHOOSE_DEFAULT = True`)
- Added test coverage in [tests/workspace/cli/test_yes_flag.py](tests/workspace/cli/test_yes_flag.py)

**Behavior:**
- **`--non-interactive`** (unchanged): Uses each prompt's default value (safe for existing scripts)
- **`--yes` / `-y`** (new): Always accepts confirmations, allowing commands to proceed without interaction

kperkins96:
@anuunchin do you mind taking another look when you get a chance? Thanks!

msmahliss:
Hey @anuunchin. Would be great to see this change merged. Our team is currently blocked from running useful pipeline commands in Airflow and this PR would resolve that. Does Kate's change cover your concerns?

anuunchin:
Hi @kperkins96, @msmahliss, sorry for the wait. The PR is currently awaiting review from another engineer on our side, and I’ll get back to you as soon as possible. Would it be okay if we take over the PR if needed?

EDIT: You will be mentioned as co-authors of course :)

kperkins96:
Sounds good! If needed, feel free to take over the PR.

anuunchin:
Closing in favor of #3910 (includes the commits here)

PR #3779, opened 2026-03-24
fix: deepcopy paginator in child resource to prevent state corruption

### LINKED ISSUE #3772: rest_api: Pages for child resources are skipped when parent resource has multiple pages
When using `rest_api_resources` with a parent-child dependency chain where the parent has multiple pages, the child resource silently skips pages. In the reproduction below, only 2 out of 3 expected items are loaded because the second page is never requested.

This appears to be the same root cause as #3685, which was closed without a reproducible example. This issue provides one.

## How to reproduce

Run the script below. It sets up a mock API with a parent-child dependency (products -> reviews) and uses `rest_api_resources` to load the data.

- 3 products across 2 pages (page_size=2)
- prod_A and prod_C have 0 reviews
- prod_B has 3 reviews (page_size=2, so 2 pages are needed)

```python
"""Tested with dlt 1.24.0, Python 3.13.5."""

import socket
import threading
import time

import dlt
import duckdb
import uvicorn
from dlt.sources.rest_api.typing import ClientConfig, Endpoint, EndpointResource
from fastapi import FastAPI

PAGE_SIZE = 2

app = FastAPI()
request_log: list[str] = []


@app.get("/products")
def get_products(offset: int = 0, limit: int = PAGE_SIZE):
    request_log.append(f"GET /products?offset={offset}")
    if offset == 0:
        return {
            "products": [{"id": "prod_A"}, {"id": "prod_B"}],
            "count": 2,
            "totalCount": 3,
        }
    elif offset == 2:
        return {
            "products": [{"id": "prod_C"}],
            "count": 1,
            "totalCount": 3,
        }
    return {"products": [], "count": 0, "totalCount": 3}


@app.get("/products/{product_id}/reviews")
def get_reviews(product_id: str, offset: int = 0, limit: int = PAGE_SIZE):
    request_log.append(f"GET /products/{product_id}/reviews?offset={offset}")
    if product_id == "prod_B":
        if offset == 0:
            return {
                "reviews": [{"id": "r1"}, {"id": "r2"}],
                "count": 2,
                "totalCount": 3,
            }
        elif offset == 2:
            return {
                "reviews": [{"id": "r3"}],
                "count": 1,
                "totalCount": 3,
            }
    return {"reviews": [], "count": 0, "totalCount": 0}


def get_free_port() -> int:
    with socket.socket() as s:
        s.bind(("127.0.0.1", 0))
        return s.getsockname()[1]


port = get_free_port()
base_url = f"http://127.0.0.1:{port}"
server = uvicorn.Server(uvicorn.Config(app, host="127.0.0.1", port=port, log_level="error"))
threading.Thread(target=server.run, daemon=True).start()
time.sleep(0.5)

config = RESTAPIConfig(
    client=ClientConfig(base_url=base_url),
    resources=[
        EndpointResource(
            name="products",
            selected=False,
            primary_key="id",
            endpoint=Endpoint(
                path="/products",
                paginator=OffsetPaginator(limit=PAGE_SIZE, total_path="totalCount"),
                data_selector="products",
            ),
        ),
        EndpointResource(
            name="reviews",
            primary_key="id",
            write_disposition="merge",
            endpoint=Endpoint(
                path="/products/{resources.products.id}/reviews",
                paginator=OffsetPaginator(limit=PAGE_SIZE, total_path="totalCount"),
                data_selector="reviews",
            ),
        ),
    ],
)


@dlt.source
def source():
    yield from rest_api_resources(config)


db_name = ":memory:repro"
query_conn = duckdb.connect(db_name)

pipeline = dlt.pipeline(
    pipeline_name="repro",
    destination=dlt.destinations.duckdb(duckdb.connect(db_name)),
    dataset_name="data",
)
pipeline.run(source())

print("Request order:")
for r in request_log:
    print(f"  {r}")

rows = query_conn.sql("SELECT id FROM data.reviews ORDER BY id").fetchall()
loaded_ids = [r[0] for r in rows]

print(f"\nLoaded reviews: {loaded_ids}")
print(f"Expected:       ['r1', 'r2', 'r3']")

if loaded_ids == ["r1", "r2", "r3"]:
    print("\nOK - all reviews loaded")
else:
    missing = set(["r1", "r2", "r3"]) - set(loaded_ids)
    print(f"\nBUG - missing reviews: {missing}")

server.should_exit = True
```

## Expected behavior

All 3 reviews are loaded. The request log includes `GET /products/prod_B/reviews?offset=2` to fetch the second page:

```
Loaded reviews: ['r1', 'r2', 'r3']
```

## Actual behavior

Only 2 of 3 reviews are loaded. The second page of prod_B's reviews (`offset=2`) is never requested. Instead, prod_A's reviews are fetched a second time:

```
Request order:
  GET /products?offset=0
  GET /products?offset=2
  GET /products/prod_A/reviews?offset=0
  GET /products/prod_C/reviews?offset=0
  GET /products/prod_B/reviews?offset=0
  GET /products/prod_C/reviews?offset=0    <-- spurious duplicate

Loaded reviews: ['r1', 'r2']
Expected:       ['r1', 'r2', 'r3']

BUG - missing reviews: {'r3'}
```

## Hints

The `OffsetPaginator` for the reviews resource appears to be created once in `create_resources` and passed by reference to `paginate_dependent_resource`. When the reviews resource receives different batches of parent items (products), each batch seems to create a separate generator that shares the same paginator instance. If `PipeIterator` interleaves these generators (via `next_item_mode = "round_robin"`), one generator's `update_state` call could affect another generator's `has_next_page` check, causing it to skip pages.

### dlt version

1.24.0

### Operating system

Linux

### Runtime environment

Local

### Python version

3.13

anuunchin:
One liner that deepcopies paginator in child resource to prevent state corruption. Appropriate test is added.

Resolves #3772

### DIFF
--- a/dlt/sources/rest_api/config_setup.py
+++ b/dlt/sources/rest_api/config_setup.py
@@ -1,5 +1,5 @@
 import warnings
-from copy import copy
+from copy import copy, deepcopy
 from typing import (
     Generator,
     Type,
@@ -1068,7 +1068,7 @@ def paginate_dependent_resource(
             params=processed_data.params,
             json=processed_data.json,
             data=processed_data.data,
-            paginator=paginator,
+            paginator=deepcopy(paginator),
             data_selector=data_selector,
             hooks=hooks,
         ):
--- a/tests/sources/rest_api/configurations/test_configuration.py
+++ b/tests/sources/rest_api/configurations/test_configuration.py
@@ -1045,6 +1045,6 @@ def test_paginator_overrides_single_entity_paths(
         assert isinstance(captured["endpoint_paginator"], SinglePagePaginator)
     elif isinstance(expected_endpoint_paginator_type, str):
         expected_endpoint = paginators.get(expected_endpoint_paginator_type)
-        assert captured["endpoint_paginator"] is expected_endpoint
+        assert isinstance(captured["endpoint_paginator"], type(expected_endpoint))
     else:
         assert captured["endpoint_paginator"] is None
--- a/tests/sources/rest_api/integration/test_offline.py
+++ b/tests/sources/rest_api/integration/test_offline.py
@@ -1880,6 +1880,50 @@ def test_post_data_param_with_list_of_tuples(mock_api_server):
         }
 
 
+def test_dependent_resource_paginated_child(mock_api_server):
+    """Test that child resources load all pages when the parent has multiple pages.
+
+    Reproduces #3772: when PipeIterator round-robins between generators for
+    different parent pages, the shared paginator state was corrupted, causing
+    child pagination to stop early and lose records.
+    """
+    pipeline = dlt.pipeline(
+        pipeline_name="rest_api_mock",
+        destination="duckdb",
+        dataset_name="rest_api_mock",
+        dev_mode=True,
+    )
+
+    mock_source = rest_api_source(
+        {
+            "client": {"base_url": "https://api.example.com"},
+            "resources": [
+                "posts",
+                {
+                    "name": "post_comments",
+                    "endpoint": {
+                        "path": "posts/{resources.posts.id}/comments",
+                        "paginator": {
+                            "type": "page_number",
+                            "base_page": 1,
+                            "total_path": "total_pages",
+                        },
+                    },
+                },
+            ],
+        }
+    )
+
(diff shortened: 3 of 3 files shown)


rudolfix (approved):
LGTM! (nasty bug indeed)

PR #3810, opened 2026-03-30
feat: `lance` destination

### LINKED ISSUE #3546: feat: support `table_format="lance"` on filesystem destination
The [pylance](https://github.com/lance-format/lance) library is now at version 1.0. It makes it trivial to write `pyarrow` data to Lance tables.

Example from their README
```python
import lance

import pandas as pd
import pyarrow as pa
import pyarrow.dataset

df = pd.DataFrame({"a": [5], "b": [10]})
uri = "/tmp/test.parquet"
tbl = pa.Table.from_pandas(df)
pa.dataset.write_dataset(tbl, uri, format='parquet')

parquet = pa.dataset.dataset(uri, format='parquet')
lance.write_dataset(parquet, "/tmp/test.lance")
```
This would allow dlt users to write `.lance` files to all supported file system and object store destinations (e.g., S3, local files, Azure, GCP)

## TODOs
- Implement feature in `dlt.destinations.impl.filesystem`
- Add a `file_format` in `dlt.common.schema.typing`
- Add a docs page for the destinations table format (like [Iceberg](https://dlthub.com/docs/dlt-ecosystem/destinations/iceberg))
- Add to docs page for the API Reference of table formats ([reference](https://dlthub.com/docs/dlt-ecosystem/table-formats))

jorritsandbrink:
### Description
- adds destination for Lance table format (powered by `lance` lib) with optional vector embedding generation (powered by `lancedb` lib)
- supports local storage and cloud storage (`s3`, `az`, `gs`)
- uses [Lance Directory Namespace](https://lance.org/format/namespace/dir/catalog-spec/#lance-directory-namespace-catalog-spec) (V2 spec) for table management
- supports [branching](https://lance.org/format/table/branch_tag/)

### Related Issues

Closes #3546 

### Additional context

We already have `lancedb` destination:
- current `lancedb` destination: fully powered by `lancedb` lib, supports local database + LanceDB Cloud
- new `lance` destination: powered by `lance` lib for table management and `lancedb` lib for embedding generation, supports local storage + cloud storage, uses namespace and supports branching

Starting point for `lance` destination code was copy from `lancedb` destination code.


dlt/destinations/impl/lance/configuration.py
```diff
@@ -0,0 +1,180 @@
...
+    "watsonx",
+]
+
+
+@configspec(init=False)
+class LanceStorageConfiguration(FilesystemConfigurationWithLocalFiles):
+    namespace_name: Optional[str] = DEFAULT_LANCE_NAMESPACE_NAME
+    """Name of subdirectory in `bucket_url` to use as namespace root. Leave empty to use `bucket_url` as namespace root."""
+    branch_name: Optional[str] = None
+    """Name of branch to use for read/write table operations. Uses `main` branch if not set."""
```
zilto, line 57: Suggestion: rename config fields
- `namespace_name` -> `namespace`
- `branch_name` -> `branch` 

This would match the convention of Python API of the `lance` library e.g., `dataset.create_branch(branch: str)`.

dlt/destinations/impl/lance/configuration.py
```diff
@@ -0,0 +1,180 @@
...
+        self.namespace_name = namespace_name
+        self.branch_name = branch_name
+        self.options = options
+
+    @property
+    def namespace_url(self) -> str:
+        namespace_url = self.bucket_url.rstrip("/")
+        if self.namespace_name:
+            namespace_url += "/" + self.namespace_name
+        return namespace_url
```
zilto: In Lance, this argument is named `uri` 

> lance.dataset()
> uri: str, Path, LanceDataset, or None
> Where to write the dataset to (directory). If a LanceDataset is passed, the session will be reused. Either uri or (namespace + table_id) must be provided, but not both.

Suggestion:
- rename `namespace_url`  to `uri` (there doesn't to be any "uri that is not a namespace uri"
- or minimally rename `namespace_url` to `namespace_uri`; `dlt` uses `bucket_url`, but it seems better to align with `lance`'s `uri`  convention

dlt/destinations/impl/lance/configuration.py
```diff
@@ -0,0 +1,180 @@
...
+        super().on_resolved()
+        credentials = (
+            self.credentials.to_object_store_rs_credentials()
+            if isinstance(self.credentials, WithObjectStoreRsCredentials)
+            else {}
+        )
+        if credentials or self.options:
+            self.options = credentials | (self.options or {})
+
+    def make_directory_namespace(self) -> DirectoryNamespace:
```
zilto, line 104: AFAIU, this approach won't support Lance lakehouse using a REST catalog.

Via the LanceDB entrypoint ([docs](https://docs.lancedb.com/namespaces#directory-namespaces))
```python
import lancedb

db = lancedb.connect("./local_lancedb")
db = lancedb.connect_namespace("dir", {"root": "./local_lancedb"})
db = lancedb.connect_namespace(
    "rest",
    {
        "uri": "https://<your_catalog>.internal.<your_org>.com",
        "headers.x-api-key": os.environ["API_KEY"],
    },
)
```

reading the docstring examples in the [source code](https://github.com/lance-format/lance/blob/main/python/python/lance/namespace.py), there's a dynamic approach possible

```python
import lance_namespace

# "dir" approach supports local and object stores
ns = lance_namespace.connect("dir", {"root": "memory://test"})

# "rest" approach supports catalogs and hosted lance
ns = lance_namespace.connect("rest", {"uri": "http://localhost:4099"})
``` 

this can be kept as future work

dlt/destinations/impl/lance/exceptions.py
```diff
@@ -0,0 +1,56 @@
...
+LANCE_UNDEFINED_ENTITY_PATTERN = re.compile(
+    rf"(?i){LANCE_NOT_FOUND}|{LANCE_DOES_NOT_EXIST}|{LANCE_MANIFEST_MODE}"
+)
+
+
+class LanceEmbeddingsConfigurationMissing(DestinationTerminalException):
+    def __init__(self, table_name: str, columns: List[str]) -> None:
+        columns_str = ", ".join(f"'{col}'" for col in columns)
+        super().__init__(
+            f"Table '{table_name}' has columns marked for embedding ({columns_str}) but is"
```
zilto: \`  rather than '  around `table_name`

dlt/destinations/impl/lance/factory.py
```diff
@@ -0,0 +1,94 @@
...
+        # enable creation of nested types to support own vectors
+        caps.supports_nested_types = True
+
+        # must store arrow-compatible nested types, not parquet default - otherwise schema checker in lance fails
+        caps.parquet_format = ParquetFormatConfiguration(use_compliant_nested_type=False)
+
+        return caps
+
+    @property
+    def client_class(self) -> Type["LanceClient"]:
```
zilto: can use lowercase `type[LanceCLient]` (it's already imported in the `TYPE_CHECKING` block

dlt/destinations/impl/lance/jobs.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(
+        self,
+        file_path: str,
+        table_schema: TTableSchema,
+    ) -> None:
+        from dlt.destinations.impl.lance.lance_client import LanceClient
+
+        super().__init__(file_path)
+        self._job_client: "LanceClient" = None
+        self._table_schema: TTableSchema = table_schema
```
zilto: Unclear why this needs to be imported here. Is it to trigger missing dependencies eagerly?

Otherwise, the type annotation can be defined using a `TYPE_CHECKING` block instead of string ref `"LanceClient"`

dlt/destinations/impl/lance/jobs.py
```diff
@@ -0,0 +1,82 @@
+import pyarrow as pa
+import pyarrow.parquet as pq
```
zilto: import from `dlt.common.libs.pyarrow`  instead of directly. This will trigger missing dependency exceptions

NOTE: our linter should catch this

dlt/destinations/impl/lance/lance_client.py
```diff
@@ -0,0 +1,577 @@
...
+        return LanceSQLClient
+
+    @property
+    def sql_client(self) -> SqlClientBase[Any]:
+        # inner import because `LanceSQLClient` depends on `duckdb` and is optional
+        from dlt.destinations.impl.lance.sql_client import LanceSQLClient
+
+        if not self._sql_client:
+            self._sql_client = LanceSQLClient(self)
+        return self._sql_client
```
zilto: Can we use the `LanceClient.sql_client`  property to ensure consistency? Also centralizes the import

dlt/destinations/impl/lance/lance_client.py
zilto, line 1: Rename `branch_name`  to `branch`  in several locations

dlt/destinations/impl/lance/sql_client.py
zilto, line 1: future work: this could be unified with `LanceDBSQLClient`

dlt/destinations/impl/lance/configuration.py
```diff
@@ -0,0 +1,180 @@
...
+    FileSystemCredentials,
+    FilesystemConfigurationWithLocalFiles,
+)
+
+if TYPE_CHECKING:
+    from lance.namespace import DirectoryNamespace
+    from lancedb.embeddings import EmbeddingFunction, EmbeddingFunctionRegistry
+
+DEFAULT_LANCE_BUCKET_URL = "."  # active run dir
+DEFAULT_LANCE_NAMESPACE_NAME = "dlt_lance_namespace"
```
zilto: Default namespace should resolve to the `dataset_name` by default, similar to `duckdb`  or `filesystem`  destinations. It makes the config lifecyle a bit more complex, but better UX and consistent with `dlt` destinations.

Currently, running this code with same `table_name` and different `dataset_name` doesn't produce collisions because `lance_namespace` produces a unique hash

```python
import dlt

pipe = dlt.pipeline("p1", destination="lance")
pipe.run(
    [{"id": 1}],
    dataset_name="dataset1",
    table_name="wont_collide"
)

pipe.run(
    [{"id": 2}],
    dataset_name="dataset2",
    table_name="wont_collide"
)

```

Output tree
```shell
dlt_lance_namespace
├── 1c50ad49_dataset1$_dlt_pipeline_state
├── 36de8119_dataset2$wont_collide
├── 71d4d882_dataset2$_dlt_pipeline_state
├── 73f3dde1_dataset1$_dlt_version
├── 884fdd37_dataset1$_dlt_loads
├── __manifest
├── b196fe11_dataset1$wont_collide
├── f7d66103_dataset2$_dlt_loads
└── f8337515_dataset2$_dlt_version
```

vs. switching the destination to `filesystem`

```python
pipe = dlt.pipeline("p1", destination=dlt.destinations.filesystem("."))
```


```shell
dataset1
  ├── _dlt_loads
  ├── _dlt_pipeline_state
  ├── _dlt_version
  ├── init
  └── wont_collide
dataset2
  ├── _dlt_loads
  ├── _dlt_pipeline_state
  ├── _dlt_version
  ├── init
  └── wont_collide
```

docs/website/docs/dlt-ecosystem/destinations/lance.md
```diff
@@ -0,0 +1,290 @@
...
+
+[Lance](https://lance.org) is an open-source columnar data format designed for AI/ML workloads, with native support for versioning, zero-copy access, and fast vector search. The `lance` destination lets you load data into Lance datasets stored on local disk or cloud object storage (S3, Azure, GCS).
+
+Optionally, the destination can generate **vector embeddings** using the [LanceDB](https://lancedb.com/) embedding functions library.
+
+
+<!--@@@DLT_DESTINATION_CAPABILITIES lance-->
+
+:::note Lance vs. LanceDB destination
+dlt ships two Lance-related destinations:
```
zilto, line 17: After discussion with the LanceDB team, a simple explanation of `lance` vs. `lancedb` destinations would be:
- both `lance` and `lancedb` destinations leverage Lance tables under the hood
- the `lancedb` destination makes it simple to get the benefits of `.lance` for simple architecture that want catalog over S3 or local embeddable databse (like `duckdb`, `sqlite`)
- the `lance` destination allows to integrate with custom managed LanceDB Enterprise inside your own external catalog (Unity, Polars, etc.) and compatible compute engine.

For dlt, another concern is that upgrading `lancedb` destination to use the newer `lance-namespace` would required migrating all data or provide backwards-compatible migration.

As maintainers our goal is to not phase out `lancedb` but rather to:
1. provide modern Lance/LanceDB features via `lance`  destination
2. add deprecation / migration warning to `lancedb` destination that eventually `lancedb` will be upgraded
3. To upgrade `lancedb` destination, we point it to `lance` with some specific configs. At the same time, we can enable a `lancedb-legacy` destination for users who don't want to upgrade. This destination wouldn't be tested

zilto (commented):
Read all the code and ran the test suite locally. The namespace and branch features are well supported! Left suggestions for renaming. 

I have a bigger questions / topics for future work.

## Directory vs. REST namespace
This PR supports directory namespaces, which work with file layout. This requires direct access to local filesystem, S3 bucket, etc.

Another pattern is the REST namespace using an external catalog (similar to iceberg). It should be simple enough to enable via an additional config flag. Would require a bit more effort to add in CI.

This will be required for LanceDB enterprise support

## Embedding adapter
In LanceDB, users can define embedding settings on the table and it's stored on the database instance. In Lance, there's no such features.

IMO `lancedb_adapter` makes for a nice UX and we might want to implement a more generic `@dlt.transformer` based approach that works with all destination. The orphan removal mechanism could be a generic SQL query based `_dlt_load_id` joins for cleanups

dlt/destinations/impl/lance/exceptions.py
jorritsandbrink: done: https://github.com/dlt-hub/dlt/pull/3810/changes/bec3f4a95279b080ad8654f2472148739e0837cb

dlt/destinations/impl/lance/factory.py
jorritsandbrink: Done: https://github.com/dlt-hub/dlt/pull/3810/changes/83bef11e5280c333d9f3b48569f68ccbc050c47b. Was sticking with uppercase variants for consistency with existing codebase but happy to use lowercase going forward.

dlt/destinations/impl/lance/jobs.py
jorritsandbrink: Done: https://github.com/dlt-hub/dlt/pull/3810/changes/cf3631abec6ac89d5eddd951758229ab49267a20. Remnant from `lancedb` destination.

dlt/destinations/impl/lance/jobs.py
jorritsandbrink: done: https://github.com/dlt-hub/dlt/pull/3810/changes/5d540c624c5f38166c0721fe3d0afedfb31db149

dlt/destinations/impl/lance/lance_client.py
jorritsandbrink: done: https://github.com/dlt-hub/dlt/pull/3810/changes/ede950eab15afc7ae2f3003beab15e364c35d1a3

dlt/destinations/impl/lance/configuration.py
jorritsandbrink: `<object>_name` names are a well-established convention in `dlt` code base (e.g. `pipeline_name`, `schema_name`, `table_name`) and clearly signal it's just the name, not a more complex object. Renaming `namespace_name` to `namespace` could also lead to confusion with the `LanceClient.namespace` that holds the `DirectoryNamespace`.

dlt/destinations/impl/lance/configuration.py
jorritsandbrink: Not a fan of `uri` because it's not descriptive. Okay with switching to `namespace_uri`, though would prefer `namespace_url` for consistency with `dlt` (i.e. `bucket_url`) over consistency with `lance`.

dlt/destinations/impl/lance/configuration.py
jorritsandbrink: Yes! Only supports Directory Namespace for now. Had the same idea to keep REST namespace as follow up.

dlt/destinations/impl/lance/configuration.py
jorritsandbrink: Reason for current setup is equivalency to SQL databases:
- database server -> bucket
- database -> root namespace
- schema -> child namespace (dataset namespace)

docs/website/docs/dlt-ecosystem/destinations/lance.md
jorritsandbrink: I think suggested explanations are not completely accurate:

> the lancedb destination makes it simple to get the benefits of .lance for simple architecture that want catalog over S3 or local embeddable databse (like duckdb, sqlite)

`lancedb` destination supports local database + LanceDB Cloud, not S3, right? Or at least S3 is not tested.

> the lance destination allows to integrate with custom managed LanceDB Enterprise inside your own external catalog (Unity, Polars, etc.) and compatible compute engine.

`lance` destination currently doesn't currently support those things.

Don't have a strong opinion on `lancedb` destination deprecation approach — fine with whatever is agreed upon.

jorritsandbrink:
@zilto thanks for thorough review! I pushed some commits to address some of your comments, and pushed back against some of your other comments where I didn't immediately agree.

dlt/destinations/impl/lance/configuration.py
zilto: Makes sense!

dlt/destinations/impl/lance/configuration.py
zilto: I suggest we go with `namespace_uri`. Since it's the most accurate and explicit.

My gripe is that `bucket_url` in `dlt` is improper and it should be `bucket_uri`. Notably, examples in the dlt docs `s3://dlt-ci-test-bucket` are not URLs.

dlt/destinations/impl/lance/configuration.py
zilto: Ok great! REST namespace would be a high priority feature. We can do it in a follow-up ticket

docs/website/docs/dlt-ecosystem/destinations/lance.md
zilto: >I think suggested explanations are not completely accurate

At least, this mapping `name: description` should be our goal. "lancedb library is directory-based lance lakehouse + convenience methods".

> lancedb destination supports local database + LanceDB Cloud, not S3, right? Or at least S3 is not tested.

LanceDB Cloud product is not the flagship / has lower adoption. We should push people towards `lance` destination for LanceDB Enterprise setup with REST catalog

I just tried S3 and you're right. The `dlt` lancedb destination doesn't support S3. The `lancedb` library can natively write to S3 and other buckets but `dlt` is not leveraging it. Could be a follow-up ticket, but probably not worth adding and supporting if we currently have 0 users for it

> lance destination currently doesn't currently support those things.

I believe all of this will be supported with the simple addition of REST namespaces, which is a priority ticket. We can update the docs now or then


### Thoughts
- maybe in the future we have a single `lance` destination and we set default `catalog="lancedb"`

dlt/destinations/impl/lance/configuration.py
zilto: `lance` should be conceptually closer to file-based destinations (S3, Iceberg, etc.) where directories are used for namespacing. [LanceDB code examples](https://lancedb.com/blog/branching-and-shallow-clone/#creating-and-using-branches) typically include the dataset as part of the namespace URI e.g., `lance.dataset("s3://bucket/my-dataset")` 

I imagine the user workflow to be
1. set `bucket_url` in `config.toml`. Let's say it's per project. `bucket_url="s3://analytics/support-bot"`
2. user creates a `dlt.Pipeline` with a `dataset_name`, e.g. `dataset_name="raw_logs"`
3. `dlt` resolves the uri `s3://analytics/support-bot/raw_logs`

This probably makes it easier to manage AWS permissions per "dlt dataset" because they would each have their own subdirectory

zilto:
Reviewed the latest changes and replied to remaining comments. Looks good!

dlt/destinations/impl/lance/configuration.py
jorritsandbrink: Fair :) done: https://github.com/dlt-hub/dlt/pull/3810/commits/dafd43f10b3b80c803f21e2e1ef840a3f418ae93

dlt/destinations/impl/lance/configuration.py
jorritsandbrink: I'm thinking the current child-namespace-per-dataset approach will work better for the REST namespace support we want to add next. Maybe AI is more convincing than me:

image


The Lance [docs](https://lance.org/format/namespace/client/object-relationship/#root-namespace) also say: _"The root namespace is assumed to always exist and is ready to be connected to by a tool to explore objects in the namespace."_. That assumption wouldn't really hold if we switch to root-namespace-per-dataset.

docs/website/docs/dlt-ecosystem/destinations/lance.md
jorritsandbrink: Agreed that the suggested explanations should be our goal! I'd say we first implement REST namespace, and then update the docs to reflect those capabilities.

dlt/destinations/impl/lance/configuration.py
zilto: Ok, makes sense! Maybe we can still improve the default value for clarity instead of `dlt_lance_namespace`?

IMO "namespace" is more of an internal term

Ideas:
1. `dlt_lance_destination`
2. `dlt_lance_root` (my preferred)
3. `lance_root`
4. `lance_destination`

jorritsandbrink: changed it to `dlt_lance_root` : https://github.com/dlt-hub/dlt/pull/3810/commits/3daacd872cec2e50443415c1dd96f48954fa32d7

zilto:
@jorritsandbrink can you investigate this CI failure related to Lance? https://github.com/dlt-hub/dlt/actions/runs/23918365661/job/70255384762

Maybe it's due to parallel test execution with poor isolation (i.e., the bug is related to the test suite, not the feature)

Also, can you rebase and see if it fixes the other CI failures?

jorritsandbrink:
@zilto indeed a test code failure! Nasty env var leak at C level. Fixed now.

rudolfix (dismissed):
LGTM! I pushed to smaller fixes @jorritsandbrink if you want to take a look

rudolfix (approved):
LGTM!

PR #3812, opened 2026-03-31
fix: `scd2` unmapped insert for nested tables

### LINKED ISSUE #3811: `scd2` nested table inserts are nondeterministic
### dlt version

1.20.0, verified not fixed on devel

### Describe the problem

When loading data from a REST API with a weak but nested schema, using the `scd2` merge strategy to load the Databricks destination (in some Databricks configurations) may cause errors if the nested object returns fewer columns after its initial run. While this is not common when working with mature APIs, it does cause obscure error messages and prevent the use of `scd2` with some APIs.

This complex scenario will produce an error such as:
```
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/dlt/common/destination/client.py", line 424, in run_managed
    self.run()
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/dlt/destinations/job_client_impl.py", line 102, in run
    self._sql_client.execute_many(self._split_fragments(sql))
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/dlt/destinations/sql_client.py", line 195, in execute_many
    result = self.execute_sql(statement, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/dlt/destinations/impl/databricks/sql_client.py", line 112, in execute_sql
    with self.execute_query(sql, *args, **kwargs) as curr:
  File "/usr/lib/python3.12/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/dlt/destinations/sql_client.py", line 454, in _wrap_gen
    raise db_ex.with_traceback(ex.__traceback__) from ex
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/dlt/destinations/sql_client.py", line 448, in _wrap_gen
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/dlt/destinations/impl/databricks/sql_client.py", line 143, in execute_query
    curr.execute(query, db_args)
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/databricks/sql/telemetry/latency_logger.py", line 182, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/databricks/sql/client.py", line 1324, in execute
    self.active_result_set = self.backend.execute_command(
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/databricks/sql/backend/thrift_backend.py", line 1054, in execute_command
    execute_response, has_more_rows = self._handle_execute_response(
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/databricks/sql/backend/thrift_backend.py", line 1261, in _handle_execute_response
    final_operation_state = self._wait_until_command_done(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/databricks/sql/backend/thrift_backend.py", line 940, in _wait_until_command_done
    self._check_command_not_in_error_or_closed_state(
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.12/site-packages/databricks/sql/backend/thrift_backend.py", line 633, in _check_command_not_in_error_or_closed_state
    raise ServerOperationError(
dlt.destinations.exceptions.DatabaseTerminalException: [DATATYPE_MISMATCH.CAST_WITH_CONF_SUGGESTION] Cannot resolve "promise_date" due to data type mismatch: cannot cast "TIMESTAMP" to "BOOLEAN" with ANSI mode on.
If you have to cast "TIMESTAMP" to "BOOLEAN", you can set "spark.sql.storeAssignmentPolicy" as 'LEGACY'. SQLSTATE: 42K09; line 1 pos 0
2026-03-30 20:20:23,654|[ERROR]|3838|134401015455872|dlt|load.py|complete_jobs:442|Job for teamwork_sales_order_full.6d2a4a0ef0.sql failed terminally in load 1774901514.3399007 with message [DATATYPE_MISMATCH.CAST_WITH_CONF_SUGGESTION] Cannot resolve "promise_date" due to data type mismatch: cannot cast "TIMESTAMP" to "BOOLEAN" with ANSI mode on.
If you have to cast "TIMESTAMP" to "BOOLEAN", you can set "spark.sql.storeAssignmentPolicy" as 'LEGACY'. SQLSTATE: 42K09; line 1 pos 0
```

### Expected behavior

While the error is obscure, I would expect the load of this table to work successfully. Specifically, it appears that [this](https://github.com/dlt-hub/dlt/blob/163ac5ed60e441947108b4675f47ea2fc644c87b/dlt/destinations/sql_jobs.py#L1010) insert statement on line 1010 of dlt/dlt/destinations/sql_jobs.py could better handle changes in the staging table schema resulting from the implementation of the destination bulk write process.

### Steps to reproduce

```
#!/usr/bin/env python3
"""Minimal scenario: `scd2` merge + nested JSON + evolving child shape.

Run twice from the same directory so pipeline state and the DuckDB file persist::

    uv run python example/repro.py 1
    uv run python example/repro.py 2

Setup:

- Root natural key: ``entity_id`` (``merge_key``).
- Two sibling list fields under the same parent row: ``tags`` and ``items``.
- First load: ``items`` rows expose three fields (``a``, ``b``, ``c``).
- Second load: the same nested list only exposes two fields (``a``, ``b``)—typical
  for weakly typed JSON APIs where keys appear and disappear between extracts.

Why DuckDB often still succeeds here: the loader uses ``INSERT INTO ... BY NAME``
when reading Parquet into tables, so column alignment is name-based.

What actually failed on Databricks (real pipeline): nested ``scd2`` follow-up SQL
(see ``gen_scd2_sql`` in ``sql_jobs.py``) uses::

    INSERT INTO <bronze_child> SELECT * FROM <staging_child> WHERE ...

Under Spark/Databricks with ANSI mode, that requires each target column to accept
the staged value's type. A weak-schema field (e.g. ``promise_date``) can end up
as ``TIMESTAMP`` in one Delta table and ``BOOLEAN`` in the other (inference drift,
evolving JSON, or staging vs destination DDL out of sync). The engine then errors
with something like: cannot cast ``TIMESTAMP`` to ``BOOLEAN`` for ``promise_date``
(DATATYPE_MISMATCH with ``storeAssignmentPolicy`` / LEGACY mentioned as a bypass—
fixing types or the generated SQL is preferable to relaxing ANSI).

If this script completes without error on your setup, the scenario is still
useful for a bug report; reproduce on Databricks with nested ``scd2`` and
inconsistent child column types across loads.
"""

from __future__ import annotations

import argparse
import os
import sys
from pathlib import Path

import dlt


def _dim_source(rows: list[dict]):
    @dlt.source(root_key=True)
    def src():
        @dlt.resource(
            merge_key="entity_id",
        )
        def dim_entity():
            yield rows

        return dim_entity

    return src()


def load_one() -> None:
    rows = [
        {
            "entity_id": 1,
            "tags": [{"tag": "alpha"}],
            "items": [
                {"a": 1, "b": 2, "c": 3},
            ],
        }
    ]
    p = dlt.pipeline(
        pipeline_name="example_scd2_nested_weak_schema",
        destination="duckdb",
        dev_mode=False,
    )
    info = p.run(_dim_source(rows))
    if info.has_failed_jobs:
        sys.exit(f"Load 1 failed: {info}")


def load_two() -> None:
    rows = [
        {
            "entity_id": 1,
            "tags": [{"tag": "alpha"}],
            "items": [
                {"a": 10, "b": 20},
            ],
        }
    ]
    p = dlt.pipeline(
        pipeline_name="example_scd2_nested_weak_schema",
        destination="duckdb",
        dev_mode=False,
    )
    info = p.run(_dim_source(rows))
    if info.has_failed_jobs:
        sys.exit(f"Load 2 failed: {info}")


def main() -> None:
    parser = argparse.ArgumentParser(description=__doc__.split("\n\n")[0])
    parser.add_argument(
        "step",
        type=int,
        choices=(1, 2),
        help="1 = first load (three fields on items), 2 = second load (two fields)",
    )
    args = parser.parse_args()

    root = Path(__file__).resolve().parent
    # Keep .dlt/ and default duckdb path stable across both steps.
    os.chdir(root)

    if args.step == 1:
        load_one()
    else:
        load_two()


if __name__ == "__main__":
    main()
```

### Operating system

Linux

### Runtime environment

Other

### Python version

3.12

### dlt data source

API

### dlt destination

_No response_

### Other deployment details

I'm using the Databricks destination which doesn't seem to be an option in the dropdown. I'm running this code on a Databricks cluster that works in most other pipelines we have created.

### Additional information

I plan to create a PR for this issue.

deschman:
### Description

For nested tables loaded with the `scd2` merge strategy, the follow-up SQL that inserts new rows from the staging table into the destination used `INSERT INTO … SELECT *`. That ties column order to whatever the bulk writer produced in staging. On Spark/Databricks (ANSI mode), if staging and destination disagree on a column’s type—e.g. schema drift or evolving JSON—`SELECT *` can produce values that do not assign cleanly to the target table and surfaces as obscure `DATATYPE_MISMATCH` errors.

This change generates an explicit column list from the destination table schema (excluding the SCD2 `from`/`to` metadata columns) for both the `INSERT` target list and the `SELECT`, so values are aligned by column name with the destination DDL instead of by staging column order.

### Related Issues

- Fixes #3811

### Additional Context

- Repro context and discussion: weak/nested schemas where child fields appear or disappear between loads; DuckDB often tolerates this with name-based parquet loads, while Databricks failed on the previous `SELECT *` pattern.
- I have read [CONTRIBUTING.md](../CONTRIBUTING.md).
- I ran `make test-common` locally and the suite was green.


rudolfix (approved):
LGTM! thanks for fixing this

PR #3814, opened 2026-03-31
fix: correctly materialize empty tables for multi-table resources

### LINKED ISSUE #3840: Empty table materialization does not work for multi-table resources
### dlt version

1.24.0

### Describe the problem

Using `dlt.mark.materialize_table_schema` does not work as expected for resources that generate multiple tables with dynamic names. If at least one of the tables contain data, the tables are simply not materialized, although they do appear in the schema stored in `_dlt_version`. Otherwise, if none of the tables contain data, an exception is raised as a table with a name matching the resource is not found.

### Expected behavior

It is expected that if a resource yields the following item, an empty table is materialized in the destination based on the hints associated.

```python
yield dlt.mark.with_hints(
    dlt.mark.make_hints(
        table_name=...,
        ...,
    ),
    create_table_variant=True,
)
```

However, the concrete table does not appear in the destination.

### Steps to reproduce

The issue can be reproduced by extracting the following resource:

```python
@dlt.resource
def data(yield_one, yield_two):
    yield dlt.mark.with_hints(
        dlt.mark.make_hints(
            table_name="table_one",
            write_disposition="replace",
            columns={"column_one": {"data_type": "text"}},
        ),
        create_table_variant=True,
    )
    yield dlt.mark.with_hints(
        dlt.mark.make_hints(
            table_name="table_two",
            write_disposition="replace",
            columns={"column_two": {"data_type": "bigint"}},
        ),
        create_table_variant=True,
    )
    if yield_one:
        yield dlt.mark.with_table_name(
            {"column_one": "something"},
            table_name="table_one",
        )
    if yield_two:
        yield dlt.mark.with_table_name(
            {"column_two": 5},
            table_name="table_two",
        )
```

Using `data(yield_one=False, yield_two=False)` should cause an error, while e.g. `data(yield_one=True, yield_two=False)` causes only the `table_one` to materialize.

### Operating system

Linux, macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

Custom standalone resource with multiple tables.

### dlt destination

Google BigQuery, DuckDB

### Other deployment details

Our use case is the following. We are extracting data from a source that streams records for multiple tables as a single stream. There is no guarantee that the records come in a specific order (i.e. one table at a time), and due to technical limitations we cannot split it into multiple streams to have one per table. We managed to make this work using a single dlt resource that annotates the records using `dlt.mark.with_table_name`. In addition, we yielded statically known (potentially incomplete) schemas using `dlt.mark.with_hints`. This has worked well until now.

Recently, we observed that those statically known tables were not getting materialized when they were empty. Due the the nature of the data, some tables might be empty at some points in time (e.g. event data from a specific time window). To ensure that the tables are never missing and to ease downstream usage, we then tried to materialize the empty tables by yielding `dlt.mark.materialize_table_schema` along with the statically known schema, but using the current version of dlt, it was not functioning correctly.

### Additional information

Proposed fix can be found in PR https://github.com/dlt-hub/dlt/pull/3814.

rooperuu:
### Description

Currently, empty table materialization does not work correctly for resources that generate multiple tables. This PR changes it so that empty tables are tracked in addition to resources to correctly write empty table files during extraction.

Fixes #3840

### Additional Context

- I have run `make test-common` locally. A few tests failed due to my local venv setup, but they seemed unrelated.
- Here is an example that fails with dlt version 1.24.0, but works with these changes:

```python
import dlt
import duckdb


@dlt.resource
def data(yield_one, yield_two):
    yield dlt.mark.with_hints(
        dlt.mark.make_hints(
            table_name="table_one",
            write_disposition="replace",
            columns={"column_one": {"data_type": "text"}},
        ),
        create_table_variant=True,
    )
    yield dlt.mark.with_hints(
        dlt.mark.make_hints(
            table_name="table_two",
            write_disposition="replace",
            columns={"column_two": {"data_type": "bigint"}},
        ),
        create_table_variant=True,
    )
    if yield_one:
        yield dlt.mark.with_table_name(
            {"column_one": "something"},
            table_name="table_one",
        )
    if yield_two:
        yield dlt.mark.with_table_name(
            {"column_two": 5},
            table_name="table_two",
        )


con = duckdb.connect(":memory:")

dlt.run(
    data(yield_one=True, yield_two=False),
    destination=dlt.destinations.duckdb(con),
)
assert len(con.sql("SELECT * FROM table_one")) == 1
assert len(con.sql("SELECT * FROM table_two")) == 0

dlt.run(
    data(yield_one=False, yield_two=False),
    destination=dlt.destinations.duckdb(con),
)
assert len(con.sql("SELECT * FROM table_one")) == 0
assert len(con.sql("SELECT * FROM table_two")) == 0

dlt.run(
    data(yield_one=True, yield_two=True),
    destination=dlt.destinations.duckdb(con),
)
assert len(con.sql("SELECT * FROM table_one")) == 1
assert len(con.sql("SELECT * FROM table_two")) == 1
```

### DIFF
--- a/dlt/extract/extract.py
+++ b/dlt/extract/extract.py
@@ -359,7 +359,7 @@ class Extract(WithStepInfo[ExtractMetrics, ExtractInfo]):
         data_tables = {t["name"]: t for t in schema.data_tables()}
         tables_by_resources = utils.group_tables_by_resource(data_tables)
 
-        resources_and_tables = {}
+        resources_and_tables: Dict[str, List[str]] = {}
         for resource_name, table_name in tables_with_empty:
             resources_and_tables.setdefault(resource_name, []).append(table_name)
 
--- a/dlt/extract/extract.py
+++ b/dlt/extract/extract.py
@@ -359,7 +359,7 @@ class Extract(WithStepInfo[ExtractMetrics, ExtractInfo]):
         tables_by_resources = utils.group_tables_by_resource(data_tables)
         for resource_name in resources_with_empty:
             if resource := source.resources.selected.get(resource_name):
-                if tables := tables_by_resources.get("resource_name"):
+                if tables := tables_by_resources.get(resource_name):
                     # write empty tables
                     for table in tables:
                         # we only need to write empty files for the root tables
--- a/dlt/extract/extract.py
+++ b/dlt/extract/extract.py
@@ -333,41 +333,49 @@ class Extract(WithStepInfo[ExtractMetrics, ExtractInfo]):
     ) -> None:
         schema = source.schema
         json_extractor = extractors["object"]
-        resources_with_items = set().union(*[e.resources_with_items for e in extractors.values()])
+        tables_with_items = set().union(*[e.tables_with_items for e in extractors.values()])
         # find REPLACE resources that did not yield any pipe items and create empty jobs for them
         # NOTE: do not include tables that have never seen data
         data_tables = {t["name"]: t for t in schema.data_tables(seen_data_only=True)}
         tables_by_resources = utils.group_tables_by_resource(data_tables)
         for resource in source.resources.selected.values():
-            if resource.write_disposition != "replace" or resource.name in resources_with_items:
-                continue
             if resource.name not in tables_by_resources:
                 continue
             for table in tables_by_resources[resource.name]:
+                write_disposition = table.get("write_disposition") or resource.write_disposition
+                if write_disposition != "replace" or (resource.name, table["name"]) in tables_with_items:
+                    continue
                 # we only need to write empty files for the root tables
                 if not utils.is_nested_table(table):
                     json_extractor.write_empty_items_file(table["name"])
 
         # collect resources that received empty materialized lists and had no items
-        resources_with_empty = (
+        tables_with_empty = (
             set()
-            .union(*[e.resources_with_empty for e in extractors.values()])
-            .difference(resources_with_items)
+            .union(*[e.tables_with_empty for e in extractors.values()])
+            .difference(tables_with_items)
         )
         # get all possible tables
         data_tables = {t["name"]: t for t in schema.data_tables()}
         tables_by_resources = utils.group_tables_by_resource(data_tables)
-        for resource_name in resources_with_empty:
+
+        resources_and_tables = {}
+        for resource_name, table_name in tables_with_empty:
+            resources_and_tables.setdefault(resource_name, []).append(table_name)
+
+        for resource_name, table_names in resources_and_tables.items():
             if resource := source.resources.selected.get(resource_name):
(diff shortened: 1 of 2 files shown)


anuunchin:
Hi @rooperuu, could you first open an issue and describe what the use case is -> I'm curious why not just use two separate resources 

Edit: The tracking granularity in the extractor does seem to be a bug though, it would just be great to have more context on the use case and whether we should take an alternative path (or even implement a bigger change) to resolve this 


A proper test will be needed as well :) (But ping me when you open the issue)

rooperuu:
@anuunchin Thanks for the comment! I created an issue and tried to explain our use case in more detail.

dlt/extract/extractors.py
```diff
@@ -120,10 +120,10 @@ def __init__(
...
         self.naming = schema.naming
         self.collector = collector
-        self.resources_with_items: Set[str] = set()
-        """Tracks resources that received items"""
-        self.resources_with_empty: Set[str] = set()
-        """Track resources that received empty materialized list"""
+        self.tables_with_items: Set[Tuple[str, str]] = set()
+        """Tracks (resource, table) pairs that received items"""
+        self.tables_with_empty: Set[Tuple[str, str]] = set()
+        """Tracks (resource, table) pairs that received empty materialized list"""
```
anuunchin, line 126: instead of tracking a set of tuples, we can just track table names directly imo, ie:

```python
self.tables_with_items: Set[str] = set()
self.tables_with_empty: Set[str] = set()
```

dlt/extract/extract.py
```diff
@@ -333,41 +333,49 @@ def _write_empty_files(
...
+                        if table["name"] not in table_names:
+                            continue
                         # we only need to write empty files for the root tables
                         if not utils.is_nested_table(table):
                             json_extractor.write_empty_items_file(table["name"])
                 else:
                     table_name = json_extractor._get_static_table_name(resource, None)
-                    if table_name:
+                    if table_name and table_name in table_names:
                         json_extractor.write_empty_items_file(table_name)
```
anuunchin, line 379: Then, if we just track the table names, this part simplifies to:

```python
tables_with_empty = (                                                                                                                
    set()                                                                                                                          
    .union(*[e.tables_with_empty for e in extractors.values()])                                                                      
    .difference(tables_with_items)                             
)                                                                                                                                    
for table_name in tables_with_empty:                                                                                               
    json_extractor.write_empty_items_file(table_name)
```

anuunchin (changes requested):
Thanks for the PR!

I think there might be a simpler approach here - see the comments. @rudolfix would be great to have your input on the direction in my comments

PS: The PR will need relevant tests to be merged :)

anuunchin:
Closing in favor of #3901 (will mention the co-authors :))

PR #3837, opened 2026-04-07
feat: add native Polars DataFrame and LazyFrame support in resources …

### LINKED ISSUE #3836: Native Polars DataFrame support in resources
### Feature description

dlt should natively accept polars.DataFrame yielded directly from a @dlt.resource, the same way pandas.DataFrame is accepted today.

```python
import polars as pl
import dlt

@dlt.resource
def events():
    yield pl.DataFrame({"id": [1, 2, 3], "ts": ["2024-01-01", "2024-01-02", "2024-01-03"]})

pipeline = dlt.pipeline(pipeline_name="polars_demo", destination="duckdb")
pipeline.run(events())
```

Today this fails because Polars DataFrames are not recognized by the extract pipeline. Users must manually call .to_arrow() or .to_pandas() before yielding, which is a papercut that shouldn't exist given how straightforward the conversion is.

The implementation is minimal because dlt already has a complete Arrow extraction path. Polars DataFrames convert to pyarrow.Table via the zero-copy df.to_arrow() method, so the only work is detecting the type and inserting the conversion at the same points where pandas is already handled.

I'm happy to submit a PR for this.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

_No response_

### Proposed solution

Add native Polars DataFrame detection and auto-conversion to the existing Arrow extraction pipeline. The approach mirrors exactly how pandas is handled today:

1. Create `dlt/common/libs/polars.py` with lazy import and `polars_to_arrow()` - auto-collects LazyFrames, then calls `.to_arrow()`
2. Add `polars.DataFrame` and `polars.LazyFrame` to type checks in `wrappers.py`, `utils.py`, `extractors.py`, and `incremental/__init__.py`
3. Add early conversion in `ArrowIncremental` (incremental runs before the extractor in the pipe)
4. Add `polars` as an optional dependency in `pyproject.toml`

All changes follow the existing `try/except MissingDependencyException` pattern - when polars is not installed, the code paths are completely unchanged.

### Related issues

_No response_

AyushPatel101:
…(#3836)

Polars DataFrames and LazyFrames can now be yielded directly from @dlt.resource without manual conversion. They are auto-detected and converted to Arrow tables via the existing Arrow extraction pipeline. LazyFrames are auto-collected before conversion.

### Related Issues

- Fixes #3836 
- Closes #...
- Resolves #...


rudolfix (dismissed):
LGTM! @AyushPatel101 thanks for adding this. I pushed several improvements to this PR: 
we were importing a lot of optional dependencies unconditionally. with `deltars` `import dlt` and `dlt` command line got even more slower.

now imports are lazy and instance checks are done only when modules are already imported. this is makes dlt startup significantly faster

PR #3839, opened 2026-04-08
(fix) preserve insert_at step ordering across resource clone

### LINKED ISSUE #3838: insert_at step ordering silently destroyed by resource clone during pipeline.run()
### dlt version

1.24.0

### Describe the problem

When a user positions a pipe step (via `add_map`, `add_filter`, `add_metrics`, or `add_step` with `insert_at`) **after** the `Incremental` step, `pipeline.run()` silently reverses the ordering. The step ends up running **before** Incremental, producing incorrect results with no error or warning.

The step reordering happens in both cases, but the observable impact differs:

| Incremental type | Steps reordered? | Data corruption? | Why? |
|---|---|---|---|
| **Hints-based** (`apply_hints`) | Yes | **Yes** | `Incremental` is a pure pipe step; step order = execution order |
| **Signature-based** (`cursor` param) | Yes | No | `IncrementalResourceWrapper.wrap()` filters at the generator level; pipe step is redundant |

### When the bug does NOT trigger

Calling the resource directly (`r = my_resource()`) sets `_args_bound = True`, causing `_eject_config` to skip the eject/inject cycle. The pipe ordering is preserved. This masks the bug in simple test cases.

I am willing to submit a PR about this request

### Expected behavior

`insert_at` positioning should be preserved across `pipeline.run()`. MetricsItem placed after Incremental should only see rows that survived filtering:

```
Steps: [gen, Incremental, MetricsItem]  — before and after clone
Output: row_count == 2  (only rows with updated_at >= 2)
```

### Steps to reproduce

```python
import dlt

@dlt.source
def my_source():
    @dlt.resource
    def events():
        yield [
            {"id": 1, "updated_at": 1},
            {"id": 2, "updated_at": 2},
            {"id": 3, "updated_at": 3},
        ]
    return events

def count_rows(items, meta, metrics):
    if isinstance(items, list):
        metrics["row_count"] = metrics.get("row_count", 0) + len(items)

source = my_source()
r = source.events  # _args_bound is False

r.apply_hints(incremental=dlt.sources.incremental("updated_at", initial_value=2))
r.add_metrics(count_rows, insert_at=len(r._pipe))
# Steps: [gen, Incremental(affinity=1), MetricsItem(affinity=0)]

pipeline = dlt.pipeline(pipeline_name="test", destination="duckdb", dev_mode=True)
pipeline.run(r)

    if step.step == "extract":
        for load_id, metrics_list in step.step_info.metrics.items():
            for m in metrics_list:
                for resource_name, resource_m in m["resource_metrics"].items():
                    if resource_m.custom_metrics:
                        print(f"{resource_name}: {resource_m.custom_metrics}")

# Expected: row_count == 2  (2 rows pass incremental: updated_at >= 2)
# Actual:   row_count == 3  (all rows — MetricsItem ran BEFORE incremental)
```

Output

```
events: {'row_count': 3, 'unfiltered_items_count': 3, 'unfiltered_batches_count': 1, ...}
```

### Operating system

Linux

### Runtime environment

Local

### Python version

3.13

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

AyushPatel101:
The eject/inject cycle during pipeline.run() re-added the Incremental step via append_step(), which sorts by placement_affinity and destroys any explicit insert_at positioning. Save the incremental step index on eject and restore it on inject so user-defined step order is preserved.

### Related Issues

- Fixes #3838 
- Closes #...
- Resolves #...

### DIFF
--- a/dlt/extract/resource.py
+++ b/dlt/extract/resource.py
@@ -470,22 +470,16 @@ class DltResource(Iterable[TDataItem], DltResourceHints):
             self._pipe.insert_step(item_transform, insert_at)
         return self
 
-    def _remove_incremental_step(self) -> int:
-        """Removes incremental step from pipe, returns its previous index (-1 if not found)"""
-        return self._pipe.remove_by_type(Incremental, IncrementalResourceWrapper)
+    def _remove_incremental_step(self) -> Optional[int]:
+        """Removes incremental step from pipe, returns its previous index (None if not found)"""
+        idx = self._pipe.remove_by_type(Incremental, IncrementalResourceWrapper)
+        return idx if idx >= 0 else None
 
     def set_incremental(
         self,
         new_incremental: Union[Incremental[Any], IncrementalResourceWrapper],
-        insert_at: Optional[int] = None,
     ) -> Optional[Union[Incremental[Any], IncrementalResourceWrapper]]:
-        """Set/replace the incremental transform for the resource.
-
-        Args:
-            new_incremental: The Incremental instance/hint to set or replace.
-            insert_at: Explicit pipe index for the incremental step. When provided,
-                bypasses `placement_affinity` ordering to preserve user-defined step order.
-        """
+        """Set/replace the incremental transform for the resource, preserving step position."""
         if new_incremental is Incremental.EMPTY:
             new_incremental = None
         if new_incremental:
@@ -493,27 +487,23 @@ class DltResource(Iterable[TDataItem], DltResourceHints):
             if resource_primary_key := self._hints.get("primary_key"):
                 new_incremental.set_deduplication_key(resource_primary_key, from_hints=True)
         incremental = self.incremental
+        existing_idx: Optional[int] = None
         if incremental is not None:
             if isinstance(new_incremental, IncrementalResourceWrapper):
-                # completely replace the wrapper, preserving position if not given
-                prev_idx = self._remove_incremental_step()
-                if insert_at is None and prev_idx >= 0:
-                    insert_at = prev_idx
-                self.add_step(new_incremental, insert_at=insert_at)
+                # completely replace the wrapper at the same position
+                existing_idx = self._remove_incremental_step()
+                self.add_step(new_incremental, insert_at=existing_idx)
             elif isinstance(incremental, IncrementalResourceWrapper):
                 incremental.set_incremental(new_incremental, from_hints=True)
             else:
-                prev_idx = self._remove_incremental_step()
-                if insert_at is None and prev_idx >= 0:
-                    insert_at = prev_idx
-                # re-add the step
+                existing_idx = self._remove_incremental_step()
                 incremental = None
         if incremental is None:
             # if there's no wrapper add incremental as a transform
             if new_incremental:
                 if not isinstance(new_incremental, IncrementalResourceWrapper):
                     new_incremental = Incremental.ensure_instance(new_incremental)
-                self.add_step(new_incremental, insert_at=insert_at)
+                self.add_step(new_incremental, insert_at=existing_idx)
         return new_incremental
 
     def _set_hints(
@@ -690,28 +680,23 @@ class DltResource(Iterable[TDataItem], DltResourceHints):
         except Exception:
             pass
 
-    def _eject_config(self) -> Optional[int]:
+    def _eject_config(self) -> bool:
         """Unwraps the pipe generator step from config injection and incremental wrappers by restoring the original step.
 
-        Removes the step with incremental wrapper. Should be used before a subsequent _inject_config is called on the
-        same pipe to successfully wrap it with new incremental and config injection.
+        Should be used before a subsequent _inject_config is called on the same pipe to successfully
+        wrap it with new incremental and config injection.
         Note that resources with bound arguments cannot be ejected.
-
-        Returns:
-            Index where incremental step was located, or `None` if nothing was ejected.
         """
         if not self._pipe.is_empty and not self._args_bound:
             orig_gen = getattr(self._pipe.gen, "__GEN__", None)
             if orig_gen:
-                incr_idx = self._remove_incremental_step()
                 self._pipe.replace_gen(orig_gen)
-                return incr_idx if incr_idx >= 0 else 0
-        return None
+                return True
+        return False
 
     def _inject_config(
         self,
         incremental_from_hints_override: Optional[bool] = None,
-        incremental_insert_at: Optional[int] = None,
     ) -> Self:
         """Wraps the pipe generation step in incremental and config injection wrappers and adds pipe step with
         Incremental transform.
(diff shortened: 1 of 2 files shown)


dlt/extract/resource.py
```diff
@@ -470,17 +470,21 @@ def add_step(
             self._pipe.insert_step(item_transform, insert_at)
         return self
 
-    def _remove_incremental_step(self) -> None:
-        self._pipe.remove_by_type(Incremental, IncrementalResourceWrapper)
+    def _remove_incremental_step(self) -> int:
+        """Removes incremental step from pipe, returns its previous index (-1 if not found)"""
+        return self._pipe.remove_by_type(Incremental, IncrementalResourceWrapper)
```
anuunchin: How about 

```python
        idx = self._pipe.remove_by_type(Incremental, IncrementalResourceWrapper)
        return idx if idx >= 0 else None
```
here?

dlt/extract/resource.py
```diff
@@ -685,23 +690,29 @@ def _set_explicit_args(
         except Exception:
             pass
 
-    def _eject_config(self) -> bool:
+    def _eject_config(self) -> Optional[int]:
```
anuunchin: I'd push back on the signature change here, we could delegate the entire incremental remove/add logic to the `_set_incremental` method instead. So here we'd have:

```python
      if not self._pipe.is_empty and not self._args_bound:                          
          orig_gen = getattr(self._pipe.gen, "__GEN__", None)                         
          if orig_gen:                                                                
              self._pipe.replace_gen(orig_gen)                                        
              return True                                                             
      return False   
```

(with proper dosctring adjustments)

dlt/extract/resource.py
```diff
@@ -490,25 +494,26 @@ def set_incremental(
...
+                    insert_at = prev_idx
                 # re-add the step
                 incremental = None
         if incremental is None:
             # if there's no wrapper add incremental as a transform
             if new_incremental:
                 if not isinstance(new_incremental, IncrementalResourceWrapper):
                     new_incremental = Incremental.ensure_instance(new_incremental)
-                self.add_step(new_incremental)
+                self.add_step(new_incremental, insert_at=insert_at)
```
anuunchin: We should preferably keep the removal/addition of the incremental step with a position constraint in one place imo:
(next comment is related)

```python
        existing_idx: Optional[int] = None
        if incremental is not None:
            if isinstance(new_incremental, IncrementalResourceWrapper):
                # completely replace the wrapper at the same position
                existing_idx = self._remove_incremental_step()
                self.add_step(new_incremental, insert_at=existing_idx)
            elif isinstance(incremental, IncrementalResourceWrapper):
                incremental.set_incremental(new_incremental, from_hints=True)
            else:
                existing_idx = self._remove_incremental_step()
                # re-add the step
                incremental = None
        if incremental is None:
            # if there's no wrapper add incremental as a transform
            if new_incremental:
                if not isinstance(new_incremental, IncrementalResourceWrapper):
                    new_incremental = Incremental.ensure_instance(new_incremental)
                self.add_step(new_incremental, insert_at=existing_idx)
```

tests/extract/test_incremental.py
```diff
@@ -1240,6 +1240,87 @@ def set_default_updated_at_arrow(records):
...
+            for load_id, metrics_list in step.step_info.metrics.items():
+                for m in metrics_list:
+                    for resource_name, resource_m in m["resource_metrics"].items():
+                        if resource_m.custom_metrics:
+                            # metrics step ran after incremental, so only 2 rows pass
+                            assert resource_m.custom_metrics["row_count"] == 2, (
+                                f"Expected 2 rows after incremental filter, "
+                                f"got {resource_m.custom_metrics['row_count']}"
+                            )
+
```
anuunchin: Let's do 
```python
    extract_info = p.last_trace.last_extract_info
    resource_metrics = extract_info.metrics[extract_info.loads_ids[0]][0]["resource_metrics"]["events"]
    assert resource_metrics.custom_metrics["row_count"] == 2
```
(this is an existing pattern)

tests/extract/test_incremental.py
```diff
@@ -1240,6 +1240,87 @@ def set_default_updated_at_arrow(records):
...
+                {"id": 3, "updated_at": 3},
+            ]
+
+        return events
+
+    source = my_source()
+    r = source.events
+
+    # verify step ordering before clone: incremental should be the last step
+    from dlt.extract.incremental import IncrementalResourceWrapper
```
anuunchin: This is already imported in this file

tests/extract/test_incremental.py
```diff
@@ -1240,6 +1240,87 @@ def set_default_updated_at_arrow(records):
...
+                    for resource_name, resource_m in m["resource_metrics"].items():
+                        if resource_m.custom_metrics:
+                            # metrics step ran after incremental, so only 2 rows pass
+                            assert resource_m.custom_metrics["row_count"] == 2, (
+                                f"Expected 2 rows after incremental filter, "
+                                f"got {resource_m.custom_metrics['row_count']}"
+                            )
+
+
+def test_insert_at_ordering_preserved_signature_incremental() -> None:
```
anuunchin: This test is unnecessary imo

tests/extract/test_incremental.py
anuunchin, line 1: Would be great if we compare before and after states around `r._clone()`, something like:

```python
    layout_before = [type(s).__name__ for s in r._pipe._steps]
    incr_idx_before = r._pipe.find(Incremental, IncrementalResourceWrapper)

    cloned = r._clone()

    layout_after = [type(s).__name__ for s in cloned._pipe._steps]
    incr_idx_after = cloned._pipe.find(Incremental, IncrementalResourceWrapper)
    assert incr_idx_after == incr_idx_before
    assert layout_after == layout_before
```

dlt/extract/resource.py
anuunchin: This will need tests for both arg based incremental

```python
def events(updated_at=dlt.sources.incremental("updated_at", initial_value=2))
```

and hint based incremental via `apply_hints`

anuunchin (changes requested):
Thanks for the PR and the repro code in the ticket!

Left a few comments with suggestions :)

dlt/extract/resource.py
AyushPatel101: Should be fixed now

dlt/extract/resource.py
AyushPatel101: Should be fixed

dlt/extract/resource.py
AyushPatel101: Should be fixed

tests/extract/test_incremental.py
AyushPatel101: Should be fixed

tests/extract/test_incremental.py
AyushPatel101: Should be fixed

tests/extract/test_incremental.py
AyushPatel101: Should be fixed

AyushPatel101 (commented):
Thanks for the feedback! All of this should be fixed now, but let me know if I missed anything

AyushPatel101:
Hi @anuunchin, bumping this in case my fix was missed here, let me know if theres anything else I need to adjust here

PR #3843, opened 2026-04-09
Don't filter out steps with exceptions in trace.asdict()

anuunchin:
This PR adjusts trace.asdict() in the way that it does NOT filter out pipelines that fail in the sync step during run before extract.

An appropriate test is added.


rudolfix (approved):
LGTM!

PR #3848, opened 2026-04-09
Increase timeout for schema assertion in e2e dashboard test

burnash:
The default 5s Playwright timeout for to_be_attached() is insufficient on Windows CI resulting in random CI failures. Match the 15s timeout used by other assertions in the same test.


rudolfix (approved):
LGTM! thanks!

PR #3854, opened 2026-04-11
Fix/3849: defer state publish in RunnableLoadJob until _on_completed marke…

### LINKED ISSUE #3849: package can be finalized/renamed before job `_on_completed(...)` finishes writing transition marker
### dlt version

1.24.0

### Describe the problem

A `RunnableLoadJob` can publish `state == "completed"` before its `_on_completed(...)` callback has finished persisting the pending-transition marker in the package directory.

Because the loader uses a single shared `_done_event` for the whole package, completion of any worker wakes the main loop, which then polls **all** running jobs. If another job already reports `"completed"` but is still inside `_on_completed(...)`, the loader may:
- remove that job from `started_jobs`
- clear its pending transition
- finalize the package
- rename `normalized/<load_id>` to `loaded/<load_id>`

while the worker thread is still writing files inside that same package directory.

### Expected behavior

_No response_

### Steps to reproduce

/

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

ugbotueferhire:
# [Fix] – Eliminate race condition between worker state publish and package finalization in `RunnableLoadJob`

## Description

**Context:**
A race condition existed in `RunnableLoadJob.run_managed()` where `self._state` was published to a terminal value (`"completed"` or `"failed"`) *before* the `_on_completed` callback had finished persisting the pending-transition marker to disk. This created a window where the main loader thread could observe the terminal state via `job.state()`, proceed to finalize the load package (renaming its directory), and collide with the worker thread still writing the transition marker inside the now-renamed directory. The result was intermittent filesystem errors or silently lost transition markers, potentially compromising data integrity on resume. See [#3849](https://github.com/dlt-hub/dlt/issues/3849).

**Approach:**
The fix introduces a **local `terminal_state` variable** in `run_managed()` that tracks the intended terminal state throughout execution and the `finally` block. The key behavioral change is:

- **Deferred state publication:** `self._state` is only assigned *after* `_on_completed()` has finished persisting the marker and `_finished_at` has been set. This ensures the main thread cannot observe a terminal state until all side effects are complete.
- **Semaphore release moved outside the retry guard:** The `_done_event.release()` call is now unconditionally placed after `self._state` is published, but still gated on non-retry terminal states. This preserves the existing wake-up semantics while guaranteeing correct ordering.
- **No new locks or synchronization primitives:** The fix is purely a reordering of existing operations, keeping the change minimal and non-invasive.

**Modified file:**
- [`dlt/common/destination/client.py`](file:///c:/Users/USER/Desktop/dlt/dlt/common/destination/client.py#L448-L494) — `RunnableLoadJob.run_managed()` method

**Execution ordering (before → after):**

```diff
  # BEFORE (racy)                        # AFTER (safe)
  try:                                    try:
-     self._state = "completed"           +     terminal_state = "completed"
  except ...:                             except ...:
-     self._state = "failed"              +     terminal_state = "failed"
  finally:                                finally:
-     if self._state != "retry":          +     if terminal_state != "retry":
-         self._on_completed(...)          +         self._on_completed(...)
-         self._finished_at = now()        +         self._finished_at = now()
-         self._done_event.release()       +     self._state = terminal_state    # ← publish AFTER marker is safe
+                                          +     if terminal_state != "retry":
+                                          +         self._done_event.release()
```

## Impact

- **Functionality:** The main loader thread can no longer finalize a load package while a worker is still writing the transition marker. Resume-after-crash is now fully reliable.
- **Data Integrity:** Eliminates the possibility of lost or corrupt transition markers caused by directory renames racing against writes.
- **Performance:** Zero overhead — the fix is a pure reordering of assignments, no new locks, no new allocations.
- **Backwards Compatibility:** No public API changes. The observable behavior (state transitions, callback ordering, semaphore semantics) is identical; only the internal timing is tightened.

## Tests

- [x] All existing tests pass
- [x] Dedicated race-condition-aware tests validate the fix

### Unit Tests (`tests/load/test_jobs.py`) — **24/24 PASSED** ✅

| Test | What it validates |
|------|-------------------|
| `test_on_completed_called_on_success` | `_on_completed` fires with `("completed", None)` on success |
| `test_on_completed_called_on_failure` | `_on_completed` fires with `("failed", traceback)` on terminal exception |
| `test_on_completed_not_called_on_retry` | `_on_completed` is **never** called on transient exceptions |
| `test_on_completed_exception_propagates` | Exceptions raised inside `_on_completed` are not swallowed |
| `test_on_completed_fires_before_semaphore_release` | `_on_completed` executes before `_finished_at` is set and before the semaphore is released — **directly validates the ordering guarantee this fix provides** |
| `test_runnable_job_results` | End-to-end state transitions for success, retry, and failure paths |
| `test_set_final_state_*` | `set_final_state()` correctly forces terminal states |

### Integration Tests (`tests/load/test_dummy_client.py`) — **51/52 PASSED** ✅

All functional loader-loop tests pass, including:

| Test | What it validates |
|------|-------------------|
| `test_completed_loop` | Full load lifecycle completes correctly |
| `test_resume_with_pending_completed_transition` | Resume correctly skips re-execution when marker exists |
| `test_resume_with_pending_failed_transition` | Resume handles failed-state markers |
| `test_load_single_thread` | Single-threaded loader path |
| `test_load_multiple_packages` | Multi-package loading |
| `test_spool_job_retry_*` | Retry mechanics remain correct |

## VISUALS
image
image

> [!NOTE]
> `test_big_load_packages` (1/52 failure) is a pre-existing **performance benchmark** that expects 1000 empty jobs to complete in under 15 seconds. It timed out at 37s due to local machine load — this is environment-specific and **unrelated to this change**.


ugbotueferhire:
Hi @Travior , could you review these changes when you have some time? Happy to make any adjustments needed. Thanks for your time!

Travior (changes requested):
Hey, @ugbotueferhire thanks for opening a PR for this.
The main changes look very good, the comments are a bit verbose but not the end of the world.

There are two main concerns that I have:
1. the changes look to correctly solve the race condition, but we should probably introduce a regression test that reproduces this ordering deterministically. I would look at `tests/load/test_jobs.py::test_on_completed_fires_before_semaphore_release` and `tests/common/configuration/test_container.py::test_container_thread_affinity` on how to do tests around load_jobs and how to use `threading.Barrier`
2. right now if `_on_completed` raises (for example when the `.dlt` folder becomes unavailable) the job stays in a non-terminal state. We need to set a terminal state without swallowing the exception, otherwise execution will halt indefinitely. There is no test that checks this right now, which would be a great addition too

If there is anything that is unclear or you need further help feel free to ping me, thanks again for your work!

ugbotueferhire:
Hey @Travior, thanks for the feedback 🤙🏾 I've addressed both concerns:

1. **Regression test**: Added `test_state_not_published_before_on_completed_finishes` — uses `threading.Barrier` (following the `test_container_thread_affinity` pattern) to deterministically verify that `state()` stays `"running"` while `_on_completed` is executing.

2. **`_on_completed` exception safety**: Wrapped the callback in `try/except` so if it raises (e.g. `.dlt` folder unavailable), the job is forced to `"failed"`, `_finished_at` is set, and the semaphore is released before re-raising. Added two tests for this: `test_on_completed_exception_sets_terminal_state` and `test_on_completed_exception_releases_semaphore`. Also updated `test_on_completed_exception_propagates` to assert terminal state.

 All 27 tests in `test_jobs.py` pass.

Travior:
Thanks for coming back to this. I'll have a look at the tests tomorrow and will come back to you

ugbotueferhire:
Hey @Travior, reminder incase you forgot, Please Whenever you have a moment, would love to get this over the line. Thanks!

Travior:
This looks quite good now.
Just one note: When `_on_completed` raises we override `self._exception` (even though this might already be populated if the load has failed). We should probably keep the "more valuable" exception from the Destination in this double failure case.
Otherwise this looks really good, tests are good

ugbotueferhire:
Thanks for flagging this.  just updated _on_completed so we only store the callback error if the job hadn’t already failed, which means we now keep the original destination exception in the double-failure case. I also added a regression test for that path.

Travior:
Thanks a lot, I'll let CI run and check in on this tomorrow morning!

Travior:
Hey @ugbotueferhire I finally got back to this. The previous test error was because a legacy test provided us a completed load package without the `.pending_transitions` folder. This would lead to our `_on_completed` to fail and since this counts as a load failure now the test wouldn't pass. I have pushed a fix for this and let the tests run again. if this fixes the issue I'll get a second opinion on this PR but it looks very good from my perspective now

ugbotueferhire:
Hey @Travior, thanks for pushing the follow-up fixes here. Just checking in: did the rerun clear the last issue, or is there anything else you’d like me to address on this PR? Happy to help with any final cleanup.

Travior:
Hi, sorry that it took me so long to come back to this. This issue has been mainly resolved by an unrelated PR that is already merged

PR #3856, opened 2026-04-11
Fix/3788 always register profile workspace cli

### LINKED ISSUE #3788: `dlt profile` command should always be available with guidance when workspace is not initialized
## Problem

`dlt profile` is a plugin-injected CLI command that only appears when a `WorkspaceRunContext` or `ProjectRunContext` is active. Without `.dlt/.workspace` (or `dlt.yml`), the command is completely absent from `dlt --help` and returns "invalid choice" when invoked directly.

This means users and AI agents that read the docs or rules referencing `dlt profile <name> pin` have no way to discover why the command doesn't exist or how to enable it. The command simply vanishes.

### Current behavior

```
$ dlt --help
# profile not listed in Available Subcommands

$ dlt profile list
dlt: error: argument command: invalid choice: 'profile'
```

### How the plugin chain works

1. `dlt._workspace.cli._plugins.plug_cli_profile` returns `ProfileCommand` only when `is_workspace_active()` is True
2. `is_workspace_active()` checks `run_context.__class__.__name__ == "WorkspaceRunContext"`
3. `WorkspaceRunContext` is only created when `.dlt/.workspace` exists (checked by `is_workspace_dir()` in `dlt/_workspace/_plugins.py`)
4. Without that file, the context falls back to plain `RunContext`, both guards return `None`, and argparse never learns about `profile`

The dlthub package has a parallel path via `ProjectRunContext` (requires `dlt.yml`), but the result is the same: no project/workspace marker = no command.

## Expected behavior

`dlt profile` should **always** be registered as a CLI command. When workspace context is not active, it should print a helpful message instead of not existing:

```
$ dlt profile list

Profiles require an initialized workspace.
Create .dlt/.workspace to enable profile support:

  touch .dlt/.workspace

Then run `dlt profile list` again to see available profiles.
See https://dlthub.com/docs/hub/core-concepts/profiles-dlthub.md
```

This is the same pattern used for other gated features (e.g., `_PondCommand` returns "Please install dlthub[cache] to enable transformations" rather than hiding the command entirely).

## Why this matters

- **Discoverability**: Docs and AI agent rules reference `dlt profile <name> pin`. When the command doesn't exist, there's no breadcrumb to follow.
- **AI agents**: Agents that run `dlt --help` to discover available commands will never find `profile` and cannot self-correct without this guidance.
- **Silent config failures**: Profile-scoped files like `dev.config.toml` are silently ignored without workspace context, and without `dlt profile` existing, there's no command to even diagnose the problem.

## Suggested fix

In `dlt._workspace.cli._plugins.plug_cli_profile`, always return a command class. When `is_workspace_active()` is False, return a placeholder that prints setup instructions (similar to the existing `_PondCommand` / `_CacheCommand` pattern already used in `dlthub.project.cli._plugins`):

```python
@plugins.hookimpl(specname="plug_cli")
def plug_cli_profile() -> Type[plugins.SupportsCliCommand]:
    if is_workspace_active():
        from dlt._workspace.cli.commands import ProfileCommand
        return ProfileCommand
    else:
        class _ProfilePlaceholder(SupportsCliCommand):
            command = "profile"
            help_string = "Manage workspace profiles (requires .dlt/.workspace)"
            def execute(self, args):
                echo.echo("Profiles require an initialized workspace.")
                echo.echo("Run: touch .dlt/.workspace")
        return _ProfilePlaceholder
```

## Environment

- dlt version: 1.24.0
- dlthub version: 0.22.1
- Python: 3.13
- OS: macOS

## Related

- #3789 -- Profile config files silently ignored without .dlt/.workspace marker (silent failure side of the same problem)

ugbotueferhire:
# [Fix] – Always register `dlt profile` and `dlt workspace` CLI commands

Closes #3788

## Description

**Context:**
The `dlt profile` and `dlt workspace` CLI commands were **completely invisible** when no `.dlt/.workspace` marker file existed. The plugin hooks `plug_cli_profile()` and `plug_cli_workspace()` in `dlt._workspace.cli._plugins` returned `None` when `is_workspace_active()` was `False`. Since pluggy silently filters `None` from hook results, argparse never learned these commands existed — they were absent from `dlt --help` and returned `"invalid choice"` when invoked directly. This broke discoverability for users, AI agents, and documentation references to `dlt profile <name> pin`.

**Approach:**
Added lightweight placeholder command classes that are registered unconditionally, following the same graceful-degradation pattern used by `DeployCommand.execute()` (which prints guidance when `pipdeptree`/`cron_descriptor` are missing rather than hiding the command):

- **`_ProfilePlaceholder`** added to `dlt/_workspace/cli/commands.py` — stands in for `ProfileCommand` when workspace context is inactive. Accepts any arguments via `nargs="*"` and prints actionable setup instructions on execute.
- **`_WorkspacePlaceholder`** added to `dlt/_workspace/cli/commands.py` — identical pattern for `WorkspaceCommand`.
- **`plug_cli_profile()`** in `dlt/_workspace/cli/_plugins.py` — returns `_ProfilePlaceholder` instead of `None` when workspace is inactive.
- **`plug_cli_workspace()`** in `dlt/_workspace/cli/_plugins.py` — returns `_WorkspacePlaceholder` instead of `None` when workspace is inactive.

No changes to the `SupportsCliCommand` protocol, the argparse wiring in `_dlt.py`, or any other command. When `.dlt/.workspace` exists, the real commands are returned exactly as before.

**Impact:**
- **Functionality:** `dlt profile` and `dlt workspace` are now always visible in `dlt --help` and print clear setup instructions when invoked outside a workspace context, instead of silently vanishing.
- **Developer Experience:** Users and AI agents that read docs referencing `dlt profile <name> pin` now get an actionable error with exact remediation steps (`mkdir -p .dlt && touch .dlt/.workspace`) instead of a cryptic argparse rejection.
- **Robustness:** Zero regression risk — when workspace is active, the original `ProfileCommand` and `WorkspaceCommand` are still returned by the plugin hooks. The placeholder classes are only used in the fallback path.

## Visual Proof

image

## Tests

- [ ] New tests added
- [x] Existing tests unaffected (no regressions)

**Manual verification performed:**
- `dlt --help` lists both `profile` and `workspace` without `.dlt/.workspace` ✅
- `dlt profile list` prints setup instructions and exits non-zero ✅
- `dlt workspace info` prints setup instructions and exits non-zero ✅
- With `.dlt/.workspace` present, real `ProfileCommand` / `WorkspaceCommand` execute normally ✅

### DIFF
--- a/dlt/_workspace/cli/_plugins.py
+++ b/dlt/_workspace/cli/_plugins.py
@@ -89,7 +89,9 @@ def plug_cli_profile() -> Type[plugins.SupportsCliCommand]:
 
         return ProfileCommand
     else:
-        return None
+        from dlt._workspace.cli.commands import _ProfilePlaceholder
+
+        return _ProfilePlaceholder
 
 
 @plugins.hookimpl(specname="plug_cli")
@@ -99,4 +101,6 @@ def plug_cli_workspace() -> Type[plugins.SupportsCliCommand]:
 
         return WorkspaceCommand
     else:
-        return None
+        from dlt._workspace.cli.commands import _WorkspacePlaceholder
+
+        return _WorkspacePlaceholder
--- a/dlt/_workspace/cli/commands.py
+++ b/dlt/_workspace/cli/commands.py
@@ -1114,3 +1114,44 @@ Pin a profile to the Workspace, this will be the new default profile while it is
             pin_profile(workspace_context, args.profile_name)
         else:
             self.parser.print_usage()
+
+
+class _ProfilePlaceholder(SupportsCliCommand):
+    """Registered when workspace context is not active so `dlt profile` is always discoverable."""
+
+    command = "profile"
+    help_string = "Manage Workspace built-in profiles (requires .dlt/.workspace)"
+
+    def configure_parser(self, parser: argparse.ArgumentParser) -> None:
+        self.parser = parser
+        # Accept but ignore any subcommands/args so the command never errors on parse
+        parser.add_argument("remainder", nargs="*", help=argparse.SUPPRESS)
+
+    def execute(self, args: argparse.Namespace) -> None:
+        fmt.warning(
+            "Profiles require an initialized workspace.\n"
+            "Create the workspace marker to enable profile support:\n\n"
+            "  mkdir -p .dlt && touch .dlt/.workspace\n\n"
+            "Then run 'dlt profile' again."
+        )
+        raise CliCommandException(error_code=-1)
+
+
+class _WorkspacePlaceholder(SupportsCliCommand):
+    """Registered when workspace context is not active so `dlt workspace` is always discoverable."""
+
+    command = "workspace"
+    help_string = "Manage current Workspace (requires .dlt/.workspace)"
+
+    def configure_parser(self, parser: argparse.ArgumentParser) -> None:
+        self.parser = parser
+        parser.add_argument("remainder", nargs="*", help=argparse.SUPPRESS)
+
+    def execute(self, args: argparse.Namespace) -> None:
+        fmt.warning(
+            "Workspace commands require an initialized workspace.\n"
+            "Create the workspace marker to enable workspace support:\n\n"
+            "  mkdir -p .dlt && touch .dlt/.workspace\n\n"
+            "Then run 'dlt workspace' again."
+        )
+        raise CliCommandException(error_code=-1)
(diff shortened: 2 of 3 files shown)


ugbotueferhire:
Hi @anuunchin , whenever you have a moment, could you please take a look at this PR for a review? Let me know if you need any changes. Thanks!

ugbotueferhire:
Hey @anuunchin, following up on this PR. Pretty low-risk change, would appreciate a review when you have bandwidth or if you could assign someone to this, thanks :)

zilto:
Issue was closed

PR #3863, opened 2026-04-14
Fix Pydantic annotation metadata handling across 2.10-2.13

### LINKED ISSUE #3861: dlt fails to infer schema for tagged unions (Pydantic RootModel) on Pydantic 2.13
### dlt version

1.24

### Describe the problem

Upgrading to pydantic 2.13 breaks dlt's handling of RootModel[Annotated[Union[..., ...], Field(discriminator=...)]]. dlt stops recognizing the discriminated union, infers a single root column instead of variant fields, and schema-contract pipeline tests fail downstream in DuckDB with NOT NULL errors on root

### Expected behavior

_No response_

### Steps to reproduce

common tests fail

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

### LINKED ISSUE #3862: dlt loses Pydantic Annotated field behavior when rebuilding column schemas for schema contracts
When a Pydantic model is used as a columns= schema, dlt may rebuild that model internally to apply schema contract behavior such as freeze, evolve, or discard_row.<br>During that rebuild, dlt does not preserve Annotated\[...\] metadata correctly. As a result, Pydantic behaviors attached through annotations can stop working on the rebuilt model.<br>From a user perspective, this means a Pydantic schema can validate one way on its own, but behave differently once passed to dlt as a column schema.

## Observed behavior

* root-level or field-level Pydantic annotation behavior may be lost after dlt rebuilds the model
* this can affect discriminated unions, validators, and other semantics attached through Annotated\[...\]
* the resulting dlt pipeline may infer the wrong schema, validate rows incorrectly, or fail at load time

Note that this is different from [OSS-38](https://linear.app/dlthub/issue/OSS-38/dlt-fails-to-infer-schema-for-tagged-unions-pydantic-rootmodel-on) as this is broader than tagged unions and has nothing to do with how Pydantic stores tagged union metadata in 2.13+ (e.g. this issue exists on 2.12 already)

E.g. this breaks:

```python
from typing import Any, ClassVar, Literal, Union
import dlt
from pydantic import AfterValidator, Field, RootModel
from typing_extensions import Annotated
from dlt.common.libs.pydantic import BaseModel, DltConfig

def reject_forbidden_clicks(value: Any) -> Any:
    if isinstance(value, ClickEvent) and value.element_id == "forbidden":
        raise ValueError("forbidden click target")
    return value

class ClickEvent(BaseModel):
    kind: Literal["click"]
    id: int
    element_id: str

class PurchaseEvent(BaseModel):
    kind: Literal["purchase"]
    id: int
    amount: float

# having more than one metadata object in an annotated type breaks right now
EventUnion = Annotated[
    Union[ClickEvent, PurchaseEvent],
    Field(discriminator="kind"),
    AfterValidator(reject_forbidden_clicks),
]

class Event(RootModel[EventUnion]):
    dlt_config: ClassVar[DltConfig] = {"is_authoritative_model": True}


@dlt.resource(
    name="events",
    columns=Event,
    schema_contract={"data_type": "discard_row"},
)
def event_stream():
    yield {"kind": "click", "id": 1, "element_id": "hero_button"}
    yield {"kind": "click", "id": 2, "element_id": "forbidden"}
```

Travior:
Fixes #3862 #3861

## Summary

- Fix discriminated `RootModel` handling across Pydantic 2.10 → 2.13
- Preserve root-level `Annotated[...]` metadata when dlt rebuilds Pydantic models for schema contracts
- Add regressions for discriminated-union schema inference and root-metadata semantics

## Problem 1 — Pydantic 2.13 moved where the discriminator lives

For tagged unions wrapped in a `RootModel`, Pydantic 2.13 moved the discriminator off the `Annotated` metadata of the root field and onto `root_field.discriminator` directly:

| | `root_field.annotation` | `root_field.discriminator` |
|---|---|---|
| **≤ 2.12** | `Annotated[Union[...], FieldInfo(discriminator="kind")]` | `None` |
| **≥ 2.13** | `Union[...]` (bare) | `"kind"` |

dlt assumed the discriminator was always recoverable from `root_field.annotation`, so on 2.13 discriminated root models were treated like a plain `RootModel` with a single `root` column, producing e.g. `NOT NULL constraint failed: click_events.root`.

This is partially fixed already, but only for the case where 2.13 strips the annotation down to a bare `Union`. It misses the case where `root_field.annotation` is **still** an `Annotated[...]` on 2.13 because additional metadata (e.g. an `AfterValidator`) is present — the discriminator lives on `root_field.discriminator` but the annotation is not bare, so the rescue path was skipped.

## Problem 2 — dlt was corrupting `Annotated[...]` metadata

While rebuilding models for schema contracts, dlt did:

```python
return Annotated[_process_annotation(a_t), tuple(a_m)]
```

This collapses *N* metadata entries into one tuple entry, changing Pydantic's semantics. For example:

```python
Annotated[Union[ClickEvent, PurchaseEvent],
          Field(discriminator="kind"),
          AfterValidator(reject_forbidden_clicks)]
```

was effectively rebuilt as:

```python
Annotated[Union[...], (FieldInfo(...), AfterValidator(...))]
```

For regular `BaseModel` this went unnoticed because `create_model(__base__=model, ...)` re-inherits the original, correctly-spread metadata. For `RootModel` there is no `__base__` rescue (it's built with `type(name, (PydanticRootModel[ann],), ...)`), so validators and other root-level metadata were silently dropped.

## Fix

- Support both discriminator representations:
  - older Pydantic: inside `Annotated[..., FieldInfo(...)]`
  - newer Pydantic: on `root_field.discriminator`, including when `root_field.annotation` is still `Annotated` due to other metadata
- Preserve root-model discriminator semantics when rebuilding contract-adjusted `RootModel[...]`
- Rebuild `Annotated[...]` metadata by preserving entries individually instead of packing them into a tuple


dlt/common/libs/pydantic.py
```diff
@@ -369,7 +381,7 @@ def _process_annotation(t_: Type[Any]) -> Type[Any]:
         """Recursively recreates models with applied schema contract"""
         if is_annotated(t_):
             a_t, *a_m = get_args(t_)
-            return Annotated[_process_annotation(a_t), tuple(a_m)]  # type: ignore[return-value]
+            return Annotated[(_process_annotation(a_t), *a_m)]  # type: ignore[return-value]
```
Travior, line 384: To me it's unclear why we previously wrapped other annotation metadata into tuples, losing their semantics:
`Annotated[str, FieldInfo(...), AfterValidator(...)]`
would be turned into:
`Annotated[str, (FieldInfo(...), AfterValidator(...)]`.

Pydantic then ignores this tuple and we effectively lose the validator (or other annotation metadata)

rudolfix: this is certainly a bug. I probably overlooked that in claude-generated code.

rudolfix (approved):
LGTM! it is very hard to verify Pydantic quirks - I added two more tests to have (hopefully) complete coverage - `test_apply_contract_preserves_nested_annotated_metadata_entries` checks tuple bug directly

Issue #3758, opened 2026-03-18
feat: implement join compatibility check for destinations


rudolfix:
## Background

We need to determine whether two `dlt.Dataset` instances can be joined in a single SQL query. The current `is_same_physical_destination` (`dlt/dataset/dataset.py:426`) compares `str(config)` — which calls `str(credentials)`, rendering the full connection URL including username, port, and database:

```python
def is_same_physical_destination(dataset1, dataset2):
    return str(dataset1.destination_client.config) == str(dataset2.destination_client.config)
```

This is **too strict** for destinations that support cross-database joins (Snowflake, BigQuery, MSSQL, ClickHouse, Databricks, Dremio, Athena) — they return `False` even when tables can be joined across databases on the same server/account.

It is **semantically wrong** for the filesystem destination where joinability comes from the in-process DuckDB query engine, not data location.

A `fingerprint()` method exists on `DestinationClientConfiguration` that hashes a location identifier (e.g. host), but it is **not used** by `is_same_physical_destination` and is missing on several destinations (DuckDB, Athena, SQLAlchemy, Fabric).

### 1. Add `location()` to `DestinationClientConfiguration`

Returns the unhashed physical location string — the identity of where this destination lives. Replaces per-destination `fingerprint()` overrides. `fingerprint()` becomes derived on the base class and is no longer overridden per-destination:

```python
def location(self) -> str:
    """Returns the physical location of this destination (e.g. host, bucket URL, project id)"""
    return ""

def fingerprint(self) -> str:
    """Returns a hash of location()"""
    loc = self.location()
    return digest128(loc) if loc else ""
```

### 2. Add `can_join_with(other)` to `DestinationClientConfiguration`

Determines if tables from this destination can be joined with tables from `other` in a single query. This reflects the **dlt destination's current capability**, not the theoretical capability of the underlying database engine (e.g. DuckDB supports ATTACH for cross-database joins, but dlt doesn't implement it yet).

Default implementation compares locations. Destinations override for special semantics (e.g. filesystem, Postgres).

```python
def can_join_with(self, other: "DestinationClientConfiguration") -> bool:
    """Check if tables from this destination can be joined with tables from other in a single query"""
    self_loc = self.location()
    other_loc = other.location()
    if not self_loc or not other_loc:
        return False
    return self_loc == other_loc
```

### 3. Update `is_same_physical_destination` in `dlt/dataset/dataset.py`

Delegate to `can_join_with` on the config.

## Join compatibility per destination

[...]

PR #3868, opened 2026-04-15
Explicit joins in Relation.join()

### LINKED ISSUE #3747: (feat) implement explicit joins with local and foreign datasets
followup: #3403
requires: #3746 (multischema datasets) and PR #3590 (magic joins)

## Introduction

PR #3590 introduced `Relation.join()` that auto-discovers join conditions from dlt schema references (parent/child relationships). This ticket extends it with an explicit `on` parameter, enabling:
- Explicit join conditions between same-dataset tables (no schema reference chain needed)
- Cross-dataset joins where `other` is a Relation from a different Dataset

A cross-dataset join introduces a foreign schema into the query. We use the local/foreign schema mechanism from #3746 to register it, making the foreign dataset's tables visible to the qualify and bind steps.

## Signature

```python
TJoinType = Literal["left", "right", "inner", "full"]


# Overload 1: Magic join — auto-discovers ON from schema references
@overload
def join(
    self,
    other: Union[str, Self],
    *,
    kind: TJoinType = "inner",
    alias: Optional[str] = None,
) -> Self: ...

# Overload 2: Explicit ON condition
@overload
def join(
    self,
    other: Union[str, Self],
    on: Union[str, sge.Expression],
    *,
    kind: TJoinType = "inner",
    alias: Optional[str] = None,
) -> Self: ...

# Implementation
def join(
    self,
    other: Union[str, Self],
    on: Union[str, sge.Expression, None] = None,
    *,
    kind: TJoinType = "inner",
    alias: Optional[str] = None,
) -> Self:
    """Join this relation with another table or relation.

    When ``on`` is omitted, join conditions are discovered automatically
    from the dlt schema's reference chain (parent/child relationships
    created during loading). Both sides must be base-table relations
    from the same dataset.

    When ``on`` is provided, any two relations can be joined — including
    relations from different datasets on the same physical destination
    (cross-dataset join). The ``on`` condition uses logical column names
    from the dlt schema.

    Args:
        other: Table name (str) or Relation to join with. When a Relation
            from a different Dataset is passed, ``on`` is required.
        on: Join condition as SQL string or SQLGlot expression. Disables
            auto-discovery from schema references when provided.
            Examples: ``"users.id = orders.user_id"``,
            ``sge.EQ(this=sge.column("id", "users"),
            expression=sge.column("user_id", "orders"))``
        kind: SQL join type — ``"inner"``, ``"left"``, ``"right"``,
            or ``"full"``. Defaults to ``"inner"``.
        alias: Column projection prefix for the joined table's columns.
            Columns from ``other`` appear as ``{alias}__{column}``.
            Defaults to the target table name.

    Returns:
        New Relation with the join applied and the joined table's
        columns appended to the projection.

    Raises:
        ValueError: If ``on`` is omitted and schema references between
            the two tables cannot be resolved, or if ``other`` is from
            a different dataset and ``on`` is not provided.

    Example:
        >>> # Magic join via schema references (same dataset)
        >>> users = dataset.table("users")
        >>> joined = users.join("users__orders", kind="left")

        >>> # Explicit join (same or cross-dataset)
        >>> customers = dataset1.table("customers")
        >>> orders = dataset2.table("orders")
        >>> joined = customers.join(
        ...     orders,
        ...     on="customers.customer_id = orders.customer_id",
        ...     kind="left",
        ... )
    """
```

### Explicit join behavior
- When `on` is provided: parse the string into a SQLGlot expression (if needed), build the JOIN node with the explicit ON condition. No schema reference lookup.
- When `on` is omitted: existing magic join behavior from PR #3590 (unchanged).
- Column projection uses the same `{alias}__{column}` convention as magic joins.

### Cross-dataset join via Relation
- When `other` is a `Relation` from a different `Dataset`, `on` is required.
- Automatically register the foreign schema: `self._dataset._add_foreign_schema(other._dataset.dataset_name, other._dataset.schema)`. This makes foreign tables visible to the qualify step.

### Dataset-qualified string in `other`
- Accept `"warehouse_data.users"` as `other` — parse into dataset_name + table_name.
- Validate that the dataset is known (local or already registered as foreign).
- If not known, raise `ValueError` with guidance to pass a `Relation` instead (so the schema can be auto-registered).
- We recommend users pass a `Relation` rather than a qualified string — it carries the schema needed for automatic registration.

## Query binder limitations for cross-dataset joins

Cross-dataset joins expose three limitations in the current qualify/bind pipeline that must be addressed:

### 1. `create_sqlglot_schema` is single-dataset

**File:** `dlt/dataset/lineage.py` — `create_sqlglot_schema()` (line 28)

Currently wraps all tables under a single `{dataset_name: {table: columns}}`. Tables from foreign datasets are invisible to `qualify()`, so column references to foreign tables cannot be resolved.

### 2. `bind_query` uses a single `expand_table_name` callback with no dataset context

**File:** `dlt/common/libs/sqlglot.py` — `bind_query()` (line 947)

The `expand_table_name` callback has signature `Callable[[str], List[str]]` — it receives only the table name. It is bound (via `functools.partial`) to the sql_client's own `dataset_name`. When `bind_query` encounters a foreign table node whose `node.db` differs from the local dataset, it still expands using the local dataset name — or skips expansion entirely if the table isn't in the SQLGlot schema.

### 3. `make_qualified_table_name_path` is hardcoded to `self.dataset_name`

**File:** `dlt/destinations/sql_client.py` — `make_qualified_table_name_path()` (line 226)

Always uses `self.dataset_name` (line 235). There is no way to override the dataset for a foreign table. Similarly, `catalog_name()` is per-sql_client, so if two datasets live in different catalogs the path cannot be constructed correctly.

### Unit tests (`tests/dataset/`)
- Explicit `on` join between two tables in the same dataset
- Explicit `on` join with `other` as a Relation from a different Dataset (cross-dataset)
- `"dataset.table"` string parsing in `other`
- Foreign schema auto-registered when cross-dataset Relation is joined
- Column projection with explicit `alias`
- Error cases: `on` omitted with cross-dataset Relation, unknown dataset string

### Integration tests with duckdb
- Two pipelines loading to two datasets on same duckdb
- Cross-dataset join produces correct SQL and returns correct data

### Smoke tests on all destinations (CI matrix)
- Same-dataset explicit join works on each destination

### Cross-catalog tests (ducklake or similar)
These are expected to initially fail — the goal is to document binder limitations:
- Join across catalogs where dataset name on each catalog is different
- Join across catalogs where dataset name is the same on both catalogs (tests that the binder distinguishes by catalog, not just dataset_name)

burnash:
Closes #3747


dlt/destinations/sql_client.py
```diff
@@ -225,20 +225,27 @@ def make_qualified_table_name(
 
     # TODO make it a staticmethod to avoid passing SQLClient instances all around
     def make_qualified_table_name_path(
-        self, table_name: Optional[str], quote: bool = True, casefold: bool = True
+        self,
+        table_name: Optional[str],
+        quote: bool = True,
+        casefold: bool = True,
+        dataset_name: Optional[str] = None,
```
rudolfix, line 232: I noticed you allow to pass catalogs into sqlglot schema qualifications. so maybe also allow catalog overwrite here as well?

dlt/dataset/relation.py
```diff
@@ -350,6 +359,199 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
...
+
+            # physical destination check
+            if target_dataset is not self._dataset:
+                if not self._dataset.is_same_physical_destination(target_dataset):
+                    raise ValueError(
+                        "Cannot join relations from different physical destinations: "
+                        f"'{target_dataset.dataset_name}' vs '{self._dataset.dataset_name}'"
+                    )
+                # cross-dataset filesystem not supported
+                if isinstance(self.sql_client, WithSchemas):
```
rudolfix, line 506: this is a good followup ticket. if we use duckdb ATTACH we will be able to join dataset on separate physical locations. ie. joining lance and HF tables will be possible

dlt/dataset/relation.py
```diff
@@ -350,6 +359,199 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
...
+
+        Returns:
+            Tuple of (target_dataset, target_table_name, target_columns).
+        """
+        if isinstance(other, dlt.Relation):
+            target_dataset = other._dataset
+
+            # physical destination check
+            if target_dataset is not self._dataset:
+                if not self._dataset.is_same_physical_destination(target_dataset):
```
rudolfix, line 500: our physical destination check is currently half implemented:
https://github.com/dlt-hub/dlt/issues/3758

currently FYI but we can add this to our estimations

dlt/dataset/relation.py
```diff
@@ -350,6 +359,199 @@ def order_by(self, column_name: str, direction: TSortOrder = "asc") -> Self:
...
+            query = _apply_join(
+                self.sqlglot_expression,
+                schema=self._dataset.schema,
+                left_table=self._table_name,
+                right_table=target_table,
+                projection_prefix=projection_prefix,
+                kind=kind,
+            )
+        else:
+            if target_dataset is not self._dataset:
```
rudolfix, line 456: hmmm this line basically answers if dataset is foreign or not. I'd probably create `is_foreign_dataset` in `Dataset` that answers that given the "other" dataset. 
1. if foreign - we do what you do here
2. if not - we add schemas from "other" to local schemas in "self" - those that are not present in "self"

now when dataset is foreign:
- different dataset name
- same name but different physical location - but those can't be joined :) so I think we are lucky with name comparison
(we could also compare catalogs if destination client supports them - but that IMO can be shifted to the moment we deal with filesystem foreign joins)

dlt/destinations/sql_client.py
burnash: could you elaborate on the use of catalog here? correct me if I'm wrong: from how I see it `create_sqlglot_schema` does {dataset_name: {table: columns}} so catalog is always empty.

dlt/dataset/relation.py
burnash: Yes, the identity check is wrong here. dataset name is better but if won't it produce false negative for case insensitive destinations?

burnash: maybe something like `(destination_fingerprint, effective_dataset_name)` in the spirit of #3758

rudolfix: I'd just compare dataset names for now. `_resolve_join_target` already checks if physical locations are same (#3758 addresses joinability explicitly so this check will be even more sound). this covers IMO all practical cases.

once #3758 is implemented we will be able to:
- can_join_with to check if we can join at all
- location() + dataset_name for the identity check

dlt/destinations/sql_client.py
rudolfix: This function was intended to generate fully qualified table names using dataset name and catalog name on `self`. now we can also use it to use arbitrary dataset name. so why not catalog name as well? `bind_query` may need it but you are right - there's no such case now. I see this only for filesystem case where we have many duckdb attached to represent foreign datasets. and those are visible as catalogs that needs to be added during bind_query.

rudolfix: tldr;> ignore this comment for now

rudolfix (approved):
LGTM!

* I merged Benjamins join compat into the branch and swapped old checks
* I fixed one non essential Snowflake test (dataset normalization)

PR #3887, opened 2026-04-23
fix: add pandas to filesystem source dependencies

### LINKED ISSUE #3876: Pipeline example dependencies don't include an install of pandas to support read_csv
### dlt version

1.25.0

### Describe the problem

When initialising a dlt project, the dependencies.txt doesn't ask for pandas; however, when actually trying to use read_csv() from filesystem, there is a requirement on pandas being installed

### Expected behavior

Pandas be included in dependencies or be bundled within filesystem import

### Steps to reproduce

# create a new directory
mkdir filesystem_local_pipeline
cd filesystem_local_pipeline
# initialize a new pipeline with your source and destination
dlt init filesystem postgres
# install the required dependencies
pip install -r requirements.txt

#No pandas is present in the requirements.txt
#Attempting to run examples here produces an error than pandas isn't installed: https://dlthub.com/docs/pipelines/filesystem-local/load-data-with-python-from-filesystem-local-to-postgres

### Operating system

Windows

### Runtime environment

Local

### Python version

3.11

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

Bhagwat-45:
## Description
Running `dlt init filesystem <destination>` followed by `pip install -r requirements.txt` 
does not install pandas. However, the generated pipeline uses `read_csv()` which requires 
pandas, causing a confusing `ModuleNotFoundError` at runtime.

**Changes:**
- Add `pandas>=1.3.0` to the `filesystem` extra in `pyproject.toml` so it is included 
  when running `pip install -r requirements.txt` after `dlt init filesystem`
- Add `MissingDependencyException` in `_read_csv` for a clear, actionable error message 
  if pandas is not installed
- Update `uv.lock` to reflect new pandas dependency

## Related Issues
- Closes #3876

## Additional Context
The fix follows the same `MissingDependencyException` pattern already used in 
`dlt/sources/sql_database/helpers.py` for optional dependencies like `connectorx`.

### DIFF
--- a/dlt/sources/filesystem/readers.py
+++ b/dlt/sources/filesystem/readers.py
@@ -3,7 +3,7 @@ from dlt.common import json
 from dlt.common.typing import copy_sig_any
 from dlt.sources import TDataItems, DltResource, DltSource
 from dlt.sources.filesystem import FileItemDict
-from dlt.common.libs.pandas import pandas as pd
+from dlt.common.libs.pandas import pandas
 
 from .helpers import fetch_arrow, fetch_json
 
@@ -23,8 +23,6 @@ def _read_csv(
     Returns:
         TDataItem: The file content
     """
-    
-
     # apply defaults to pandas kwargs
     kwargs = {**{"header": "infer", "chunksize": chunksize}, **pandas_kwargs}
     # For some remote file systems (for example, sftp/paramiko), decoding may happen before
@@ -42,9 +40,8 @@ def _read_csv(
         # Here we use pandas chunksize to read the file in chunks and avoid loading the whole file
         # in memory.
         with file_obj.open(mode=open_mode, **open_kwargs) as file:
-            for df in pd.read_csv(file, **kwargs): # type: ignore[attr-defined]
+            for df in pandas.read_csv(file, **kwargs):
                 yield df.to_dict(orient="records")
-
 
 # NOTE inconsistent kwarg convention across readers `chunk_size` vs. `chunksize`
 # snakecased `chunk_size` is the more appropriate Python convention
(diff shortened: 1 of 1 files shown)


dlt/sources/filesystem/readers.py
```diff
@@ -1,5 +1,5 @@
 from typing import TYPE_CHECKING, Any, Iterable, Iterator, Optional
-
+from dlt.common.exceptions import MissingDependencyException
```
zilto: to remove

dlt/sources/filesystem/readers.py
```diff
@@ -23,7 +23,10 @@ def _read_csv(
     Returns:
         TDataItem: The file content
     """
-    import pandas as pd
+    try:
+        import pandas as pd
+    except ImportError:
+        raise MissingDependencyException("filesystem reader (read_csv)", ["pandas"])
```
zilto: remove `try/except`. Instead do `from dlt.common.libs import pandas`

pyproject.toml
```diff
@@ -99,6 +99,7 @@ ducklake = [
 filesystem = [
     "s3fs>=2022.4.0",
     "botocore>=1.28",
+    "pandas>=1.3.0",
```
zilto: remove dependency and revert changes to `uv.lock` accordingly.

zilto (changes requested):
Thanks for the contribution. `pandas`  should be an optional dependency for the `filesystem` source because not all users of `filesystem` are using `read_csv()`.

Your approach to import `pandas` inside `_read_csv()` inside a `try/except` block is correct. 

Though, we have a convention for this. Instead, you can simply do `from dlt.common.libs.pandas import pandas`. This file at `dlt/common/libs/pandas.py` already includes the `try/except` block

Bhagwat-45:
Hi @zilto, I've addressed all the requested changes:

Replaced the try/except block with from dlt.common.libs import pandas as pd
Removed the MissingDependencyException import
Reverted pyproject.toml and uv.lock

Please let me know if anything else needs to be changed!

PR #3889, opened 2026-04-23
Implement incremental filtering for `dlt.Relation`

### LINKED ISSUE #3750: (feat) use Incremental instances as Relation filters
**Background**

`Incremental` instances are used to filter source data in `sql_database` (via SQLAlchemy WHERE), `rest_api` (via URL parameter substitution), and other sources. This ticket extends the same concept to `Relation` — generating SQLGlot WHERE clauses from `Incremental` state.

This ticket does **not** cover `ModelIncremental` (pluggable transform for automatic incremental on models) or common time sources for `end_value` — those are separate follow-ups.

requires: PR #3590 (magic joins)

### `Dataset.table()` extension

```python
def table(
    self,
    table_name: str,
    *,
    load_ids: Optional[Collection[str]] = None,
    incremental: Optional[Incremental] = None,  # NEW
    **kwargs: Any,
) -> dlt.Relation:
```

When `incremental` is provided, equivalent to `dataset.table(name).incremental(incremental)`.

### `Relation.incremental()`

```python
def incremental(self, incremental: Incremental) -> Self:
    """Apply incremental filtering as a WHERE clause on this relation.

    Generates a SQL WHERE condition from the Incremental's cursor state
    (last_value, end_value, range boundaries). Only supports `max` and
    `min` as `last_value_func` — custom functions raise ValueError.

    The cursor_path is interpreted as:
    - Simple name ("created_at") — WHERE on this relation's column
    - Dotted name ("_dlt_loads.inserted_at") — auto-join the table
      (via schema references from PR #3590), then WHERE on the
      joined column

    Args:
        incremental: Incremental instance with cursor configuration.

    Returns:
        New Relation with the WHERE clause applied.
    """
```

### `Relation.is_incremental`

```python
@property
def is_incremental(self) -> bool:
    """Returns True if this relation has an incremental WHERE clause."""
```

Walks the SQLGlot AST looking for nodes with `meta.get("dlt_incremental")` — no flag needed on the Relation instance, since SQLGlot `.meta` survives `.copy()` and all AST transformations.

## Cursor on another table

The user can define the incremental column as `"table.column"`. This triggers an auto-join (PR #3590) on `"table"` and builds the WHERE clause on `"column"`.

Example — incremental on `_dlt_loads.inserted_at`:
```python
dataset.table("orders", incremental=dlt.sources.incremental("_dlt_loads.inserted_at"))
```

This joins `orders._dlt_load_id == _dlt_loads._dlt_load_id` (discovered from schema references) and filters on `_dlt_loads."inserted_at"`.

Requirements:
- Auto-join must succeed: the schema must contain an explicit table reference or nested table relationship between the base table and the join target.
- If the target table is already joined (from a prior `.join()` call), reuse the existing join — don't re-join.
- The auto-join must **not** modify the projection — no columns from the joined table are added to the SELECT list. Only the JOIN and WHERE are added.

### `Relation.incremental()` method

**File:** `dlt/dataset/relation.py`

```python
def incremental(self, inc: Incremental) -> Self:
    table_name, cursor_column = _parse_incremental_cursor_path(inc.cursor_path)

    rel = self
    if table_name:
        # check if already joined
        qualifier_map = _extract_joined_table_aliases(rel.sqlglot_expression)
        if table_name not in qualifier_map:
            # auto-join without projection
            join_params, target_table, target_qualifier = _discover_join_params(rel, table_name)
            rel = _apply_join(
                rel, join_params,
                target_table=target_table,
                target_qualifier=target_qualifier,
                projection_prefix=table_name,
                project=False,  # NEW flag — skip _apply_join_projection
            )
            table_qualifier = target_qualifier
        else:
            table_qualifier = qualifier_map[table_name]

        column_ref = sge.Column(
            this=sge.to_identifier(cursor_column, quoted=True),
            table=sge.to_identifier(table_qualifier, quoted=False),
        )
    else:
        column_ref = sge.Column(this=sge.to_identifier(cursor_column, quoted=True))

    condition = _build_incremental_condition(inc, column_ref)
    condition.meta["dlt_incremental"] = True
    return rel.where(condition)
```

### `_apply_join` extension — `project` flag

**File:** `dlt/dataset/relation.py` (in PR #3590's `_apply_join` function)

Add `project: bool = True` parameter. When `False`, skip the `_apply_join_projection()` call — the JOIN is added to the query but no columns from the joined table appear in the SELECT list.

```python
def _apply_join(
    relation: _TRelation,
    join_params: list[_JoinParams],
    *,
    target_table: str,
    target_qualifier: str,
    projection_prefix: str,
    kind: TJoinType = "inner",
    project: bool = True,  # NEW
) -> _TRelation:
    query = relation.sqlglot_expression.copy()
    # ... build JOIN expressions (unchanged) ...

    if project:
        _apply_join_projection(...)

    rel = relation.__copy__()
    rel._sqlglot_expression = query
    return rel
```

### WHERE generation — `_build_incremental_condition`

**File:** `dlt/dataset/relation.py`

Port of `sql_database/helpers.py:172-196` logic from SQLAlchemy to SQLGlot:

```python
def _build_incremental_condition(
    inc: Incremental,
    column_ref: sge.Column,
) -> sge.Expression:
    """Build SQLGlot WHERE condition from Incremental state.

    Only supports max and min as last_value_func.
    """
    last_value_func = inc.last_value_func
    if last_value_func not in (min, max):
        raise ValueError(
            "Incremental WHERE generation only supports min/max last_value_func,"
            f" got {last_value_func}"
        )

    start_value = inc.last_value  # reads from state, applies lag
    end_value = inc.end_value

    if last_value_func is max:
        start_op = sge.GTE if inc.range_start == "closed" else sge.GT
        end_op = sge.LT if inc.range_end == "open" else sge.LTE
    else:  # min
        start_op = sge.LTE if inc.range_start == "closed" else sge.LT
        end_op = sge.GT if inc.range_end == "open" else sge.GTE

    conditions = []
    if start_value is not None:
        conditions.append(start_op(this=column_ref, expression=_to_typed_literal(start_value)))
        if end_value is not None:
            conditions.append(end_op(this=column_ref, expression=_to_typed_literal(end_value)))
        if inc.on_cursor_value_missing == "include":
            # OR cursor IS NULL — wraps the existing conditions
            pass  # applied after AND combination below

    if inc.on_cursor_value_missing == "exclude":
        conditions.append(sge.Not(this=sge.Is(this=column_ref, expression=sge.Null())))

    if not conditions:
        raise ValueError("Incremental has no last_value and no on_cursor_value_missing='exclude'")

    result = conditions[0]
    for c in conditions[1:]:
        result = sge.And(this=result, expression=c)

    # wrap with OR IS NULL for "include" mode
    if inc.on_cursor_value_missing == "include" and start_value is not None:
        result = sge.Or(this=result, expression=sge.Is(this=column_ref, expression=sge.Null()))

    return result
```

Value conversion uses `build_typed_literal()` (already in `dlt/dataset/relation.py`) for type-safe SQLGlot literals.

### Cursor path parsing — `_parse_incremental_cursor_path`

```python
def _parse_incremental_cursor_path(cursor_path: str) -> tuple[Optional[str], str]:
    """Parse 'table.column' or 'column' from incremental cursor_path."""
    if "." in cursor_path:
        table_name, column_name = cursor_path.rsplit(".", 1)
        return table_name, column_name
    return None, cursor_path
```

### `is_incremental` property

```python
@property
def is_incremental(self) -> bool:
    for node in self.sqlglot_expression.walk():
        if node.meta.get("dlt_incremental"):
            return True
    return False
```

This works because SQLGlot `.meta` dict survives `.copy()` and embedding in larger ASTs.

### Column incremental with various data types
- `datetime` cursor: `incremental("created_at", initial_value=pendulum.parse("2024-01-01"))`
- `bigint` cursor: `incremental("id", initial_value=0)`
- Verify generated WHERE clause matches expected SQL for each type

### Incremental parameters
- `range_start="open"` vs `"closed"` → `>` vs `>=`
- `range_end="open"` vs `"closed"` → `<` vs `<=`
- `last_value_func=min` → reversed operators
- `end_value` set → adds upper bound condition
- `on_cursor_value_missing="include"` → `OR col IS NULL`
- `on_cursor_value_missing="exclude"` → `AND col IS NOT NULL`
- `lag` applied → verify `last_value` is adjusted
- Custom `last_value_func` → raises `ValueError`

### Failure modes
- Column not found in schema → error from `Relation.where()` / column type lookup
- `last_value_func` is not `min` or `max` → `ValueError`
- Dotted cursor path with unknown table → auto-join fails with schema reference error
- Dotted cursor path with no reference chain → auto-join fails

### Incremental on table level vs query
- `dataset.table("orders", incremental=...)` → same as `dataset.table("orders").incremental(...)`
- `dataset.query("SELECT ...").incremental(...)` → works on arbitrary queries (cursor column must be in projection)

### Auto-join tests
- Cursor on un-joined table → auto-join added, no projection pollution, WHERE on join qualifier
- Cursor on already-joined table → no re-join, WHERE uses existing join qualifier
- Auto-join wrapped in subsequent `.select()` / `.where()` → verify qualification still works
- `_dlt_loads.inserted_at` example end-to-end with duckdb

### `is_incremental` detection
- `relation.is_incremental` is `False` before `.incremental()`
- `relation.is_incremental` is `True` after `.incremental()`
- Survives chaining: `relation.incremental(...).select("id").is_incremental` is `True`
- Regular `.where()` does not set `is_incremental`

## Note: `project=False` does not prevent access to joined columns

When the auto-join uses `project=False`, the joined table's columns are not in the SELECT list but **are** accessible in the FROM/JOIN clause. If the user later calls `.select("id", "amount", "inserted_at")`, the `inserted_at` column resolves to `_dlt_loads.inserted_at` at the database level.

This works both with and without `merge_subqueries`:
- With merge: `.select()` collapses the subquery, and the SELECT sits directly on the JOIN — columns from both tables are visible.
- Without merge: the inner `SELECT *` passes through all columns from both sides of the JOIN, so the outer SELECT can reference them.

The `project=False` flag keeps the **default** projection clean (only the base table's columns), but does not restrict what the user can explicitly select afterwards.

burnash:
Closes #3750


dlt/dataset/relation.py
```diff
@@ -439,6 +450,123 @@ def join(
...
+        rel._incremental_ctx = _RelationIncrementalContext(
+            incremental=incremental,
+            cursor_column=column_ref.copy(),
+        )
+        return rel
+
+    @property
+    def is_incremental(self) -> bool:
+        """True if any clause on this relation was produced by `.incremental()`."""
+        # TODO: leaks True on aggregate relations because the inner subquery still
```
burnash: The issue (https://github.com/dlt-hub/dlt/issues/3750) recommends storing this as a flag in sqlglot meta, but this imo has downsides: now there's no single source of truth and _incremental_ctx also should survive transformations. So I'd drop the flag.

tests/dataset/test_relation_incremental.py
```diff
@@ -0,0 +1,622 @@
...
+    # is filter-only (project=False path).
+    selects = rel.sqlglot_expression.selects
+    output_names = [expr.output_name for expr in selects]
+    assert not any(name.startswith("_dlt_loads__") for name in output_names)
+
+
+def test_incremental_dotted_cursor_runtime_columns_base_only(
+    incremental_dataset: dlt.Dataset,
+) -> None:
+    inc: dlt.sources.incremental[Any] = dlt.sources.incremental(
```
rudolfix: very cool! this will require end to end test (processing subsequent package ids)

rudolfix (commented):
this looks pretty good! 99% of use cases will be `datetime` cursors (this include `_dlt_load_id` which we'll join to `_dlt_loads` to use `inserted_at` as cursor - that's why we need this autojoin)

1. it makes sense to reuse parts of `to_sqlglot_filter` - it got tested across many destinations, which surfaced sqlglot problems with datetime literal
- taking sqlglot Column and type is a good idea - better then `to_sqlglot_filter`
- `lower, upper = incremental.resolve_bounds(apply_lag=apply_lag)` - this gets correct range from incremental - also from ad hoc instances (not only inside resource) ie. without end_value set. (I added this exactly due to this problem)
- code that formats datetime literals depending on destination caps should be ported
- there are unit test: you'll probably have your own
- end to end tests in `test_read_interfaces` are worth preserving - they test datetime literals on all destinations - was not easy to make it work on all of them
2. OK to drop this meta tag for now... let's see when we need it. 
3. we may still want to expose something similar to `to_sqlglot_filter` from `_incremental` module - for users that want to generate expression themselves and pass them to `where`)
4. `_parse_incremental_cursor_path` this will be pretty useful. let's sync on those max/min values - datetime values will come mostly from external schedulers but we should still support "standalone" incrementals

dlt/extract/incremental/transform.py
```diff
@@ -610,3 +610,49 @@ def _process_null_at_cursor_path(self, tbl: "pa.Table") -> Tuple["pa.Table", "pa
...
+      advanced from observed data.
+    - `range_start="open"`, no `end_value`: stateful open-range: aggregate runs
+      and `last_value` advances. Open range on the next run excludes the
+      boundary, so no deduplication is required.
+    - Otherwise (closed-range stateful): rejected as boundary deduplication via
+      `unique_hashes` cannot be reproduced from a single aggregate.
+    """
+
+    def __call__(self, relation: TDataItem) -> Tuple[Optional[TDataItem], bool, bool]:
+        ctx = getattr(relation, "_incremental_ctx", None)
```
rudolfix, line 628: please check dlthub review - here we should apply incremental automatically if not yet applied

dlt/dataset/_incremental.py
```diff
@@ -0,0 +1,233 @@
...
+) -> sge.Select:
+    """Build `SELECT <func>(alias) FROM (SELECT cursor AS alias FROM <filtered>)`.
+
+    Bare cursor: wraps the base query as a subquery so projections, GROUP BY,
+    HAVING and aliased computed cursors are preserved. Qualified cursor
+    (`table.column`, from an auto-join): replaces the base query's projection
+    list inline so the join qualifier resolves.
+    """
+    if ctx.incremental.end_value is None and (
+        base_query.args.get("limit") is not None or base_query.args.get("order") is not None
```
rudolfix: what is wrong with "ORDER"? order alone returns all rows and is acceptable

dlt/dataset/relation.py
```diff
@@ -438,6 +448,134 @@ def join(
...
+
+        Raises:
+            ValueError: If the cursor path is a JSONPath expression, if a dotted
+                cursor is used on a relation whose FROM is not a bare base table
+                (e.g. after `.from_loads()` or on a `.query(...)` relation), if
+                the referenced table is not in the dataset schema, or if
+                `last_value_func` is not `min` or `max`.
+
+        Notes:
+            Aggregate (GROUP BY) cursors with `range_start="open"`: late
```
rudolfix: 1. This is a rare edge case - worth describing - but maybe in our docs. Relation docs are not yet updated - it belongs there
2. the case described here is intentional `range_start="open"` is explicit setting and it means: ignore late arriving data at the boundary. so this is intentional correct behavior

dlt/dataset/relation.py
```diff
@@ -438,6 +448,134 @@ def join(
...
+            rel = self.__copy__()
+            if condition is not None:
+                rel._sqlglot_expression = rel.sqlglot_expression.where(condition)
+            rel._incremental_ctx = _RelationIncrementalContext(
+                incremental=incremental,
+                cursor_column=column_ref.copy(),
+            )
+            return rel
+
+        if not self._table_name:
```
rudolfix, line 487: you could easily extract part below into a helper function and reuse incremental application logic in 
```
if table_name is None:
```
path, incremental building like:
```
sqlglot_type = _sqlglot_type_for_column(columns, column_name)
      _maybe_warn_on_cursor_missing_raise(incremental, columns, column_name)
      condition    = _build_incremental_condition(incremental, column_ref, sqlglot_type)

      rel = self.__copy__()
      rel._sqlglot_expression = query.where(condition) if condition is not None else query
      rel._incremental_ctx = _RelationIncrementalContext(
          incremental=incremental,
          cursor_column=column_ref.copy(),
      )
      return rel
```
is identical for both

dlt/dataset/relation.py
```diff
@@ -438,6 +448,134 @@ def join(
...
+            project=False,
+        )
+        qualifier_map = _extract_joined_table_aliases(query)
+        target_qualifier = qualifier_map[table_name]
+
+        column_ref = sge.Column(
+            this=sge.to_identifier(column_name, quoted=True),
+            table=sge.to_identifier(target_qualifier, quoted=False),
+        )
+        target_columns = self._dataset.schema.tables[table_name].get("columns", {})
```
rudolfix: 1. materialized tables have columns
2. this expression also returns incomplete/not materialized columns. 
3. use `get_table_columns` 

having such columns is a byproduct of letting our users to define columns partially (ie. to say column with this name should be null, without setting the data type). maybe we should cleanup schemas when they are used in Relation class... but that's not something we can do now

dlt/extract/incremental/__init__.py
```diff
@@ -642,10 +643,17 @@ def _make_or_get_transformer(self, cls: Type[IncrementalTransform]) -> Increment
 
     def _get_transform(self, items: TDataItems) -> IncrementalTransform:
         """Gets transform implementation that handles particular data item type"""
+        # Lazy import to avoid failure with a partially-initialised
+        # `dlt.extract` during dlt startup.
+        # TODO: we should consider creating a registry for transforms
```
rudolfix, line 652: I agree. we are reworking item detection in other pull requests. there's should be universal dispatched function (well - it is there but not used) that will return item type from which right incremental type could be derived

rudolfix (changes requested):
`ModalIncremental` looks good - see the other PR for more details on that.
I've found two places where incomplete columns are allowed (raw read of columns from table schema). the other is old one on autojoin (dlt/dataset/_join.py:316). I included two tests that surface those

at this point we IMO need at least one smoke test for Incremental working on all destinations - there's still no end to end test that runs pipeline a few times and observes right data being selected via incremental

dlt/extract/incremental/transform.py
burnash: Done in b87ca5d8b18ce934fb370d62af4aac8b35b47678

dlt/dataset/relation.py
burnash: Fixed in 199c63956889e3e0bcafed48528e5d0c6f69b5ea

dlt/dataset/relation.py
burnash: Fixed in e77b73e

dlt/dataset/_incremental.py
burnash: True, fixed in 643c72a

dlt/dataset/relation.py
burnash: Removed in 6697aa4b255cf93797ccae8e045bb5aed6487866

dlt/common/destination/capabilities.py
```diff
@@ -157,6 +163,12 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     "Escapes string literal"
     casefold_identifier: TCasefoldIdentifier = str
     """Casing function applied by destination to represent case insensitive identifiers."""
+    null_safe_aggregate: Optional[TNullSafeAggregate] = None
+    """Wraps an aggregate so that an empty input yields `NULL` (standard SQL).
+
+    Leave unset on destinations that already follow standard SQL. Set on
```
rudolfix: IMO this should be removed. it is opus talking to itself. this information is relevant but not here - place where we write or test actual clickhouse case.

btw. no matter how many rules against this are present opus still generates this slop. so we must self review

dlt/common/libs/sqlglot.py
```diff
@@ -637,6 +654,76 @@ def _literal(v: Any) -> sge.Expression:
...
+            generic tz-aware cast is emitted.
+
+    Returns:
+        Tuple[Optional[sge.DataType], Any, Any]: `(sqlglot_type, lower, upper)`,
+            where `sqlglot_type` is `None` when no cast should be emitted
+            (sqlite's numeric-affinity TIMESTAMP).
+    """
+    # lazy import — `dlt.common.data_writers` reaches back into
+    # `dlt.common.destination`, which imports this module at load time
+    from dlt.common.data_writers.escape import format_datetime_value
```
rudolfix, line 683: inline import, another of opus problems

dlt/dataset/_incremental.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+_AGG_CURSOR_ALIAS = "__dlt_inc_cursor"
+
+
+@dataclass(frozen=True)
+class _RelationIncrementalContext:
+    """Private per-relation marker tying a `Relation` back to its `Incremental`.
+
+    Set by `Relation.incremental()` and consumed by downstream lifecycle
```
rudolfix: please remove or move the place where it is actually relevant (dlthub module?)

dlt/dataset/_incremental.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+def _build_incremental_aggregate(
+    base_query: sge.Query,
+    ctx: _RelationIncrementalContext,
+    destination_capabilities: Optional[DestinationCapabilitiesContext] = None,
+) -> sge.Select:
+    """Build `SELECT <func>(alias) FROM (SELECT cursor AS alias FROM <filtered>)`.
+
+    Bare cursor: wraps the base query as a subquery so projections, GROUP BY,
```
rudolfix: another opus problem: instead of writing code comments where info below actually matters it talks to itself in docstrings

rudolfix: code below is tricky. comments are required

dlt/dataset/_incremental.py
```diff
@@ -0,0 +1,259 @@
...
+    agg: sge.Expression = agg_func
+    if destination_capabilities is not None and destination_capabilities.null_safe_aggregate:
+        agg = destination_capabilities.null_safe_aggregate(agg_func)
+    return sge.Select(expressions=[agg]).from_(inner.subquery())
+
+
+def _parse_incremental_cursor_path(cursor_path: str) -> Tuple[Optional[str], str]:
+    """Split `table.column` into parts, or return `(None, column)` for a bare field.
+
+    Rejects JSONPath expressions (wildcards, array indices, `$` root markers) that
```
rudolfix: same - code comment

dlt/dataset/_join.py
```diff
@@ -343,8 +352,14 @@ def _apply_join(
     right_table: str,
     projection_prefix: str,
     kind: TJoinType = "inner",
+    project: bool = True,
 ) -> sge.Select:
-    """Apply schema-driven join(s) to ``expression`` and return the new query."""
+    """Apply schema-driven join(s) to `expression` and return the new query.
+
+    When `project` is `False` the JOIN is added to the query but the SELECT
```
rudolfix: or we explain all args or we comment on `project` in code

dlt/dataset/_join.py
```diff
@@ -374,13 +389,18 @@ def _apply_join(
...
+            query,
+            schema=schema,
+            left_table=left_table,
+            target_table=right_table,
+            target_qualifier=target_qualifier,
+            projection_prefix=projection_prefix,
+            allow_existing_target_projection=not join_params,
+        )
+    else:
+        # Filter-only join: qualify the left projection so a bare `*` does not
```
rudolfix: are you using claude code? do rules load? comments should start with small letter. this is a nitpick but maybe your agent loop does not work

dlt/dataset/relation.py
```diff
@@ -438,6 +449,153 @@ def join(
...
+
+        Raises:
+            ValueError: If the cursor path is a JSONPath expression, if a dotted
+                cursor is used on a relation whose FROM is not a bare base table
+                (e.g. after `.from_loads()` or on a `.query(...)` relation), if
+                the referenced table is not in the dataset schema, or if
+                `last_value_func` is not `min` or `max`.
+
+        Notes:
+            Scheduler mode (`end_value` set): align window bounds to
```
rudolfix: why this is relevant here? there's a lot of slang that we never use in our docs. if there's end_value we use it as "upper" bound in WHERE - that's it. lag is not applied (this is how incremental works). if you want to keep this note - it is just one short sentence

dlt/dataset/relation.py
```diff
@@ -438,6 +449,153 @@ def join(
...
+
+        Args:
+            incremental (Incremental[Any]): The incremental whose cursor path and
+                range define the filter. `last_value_func` must be `min` or `max`.
+
+        Returns:
+            Self: A new relation with the incremental filter applied.
+
+        Raises:
+            ValueError: If the cursor path is a JSONPath expression, if a dotted
```
rudolfix: I'd skip it. exception messages are self explanatory

dlt/dataset/relation.py
```diff
@@ -438,6 +449,153 @@ def join(
...
+
+    @property
+    def is_incremental(self) -> bool:
+        """True if any clause on this relation was produced by `.incremental()`."""
+        return self._incremental_ctx is not None
+
+    def _incremental_aggregate_relation(self) -> Optional[Self]:
+        """Return a relation computing `<last_value_func>(cursor)` over this relation.
+
+        Used by downstream lifecycle code (e.g. dlthub transformations) to advance
```
rudolfix: irrelevant (opus slop)

dlt/extract/incremental/sql.py
```diff
@@ -1,63 +1,19 @@
...
-
-    # sqlite's CAST AS TIMESTAMP/TIMESTAMPTZ goes through NUMERIC affinity and
-    # parses only leading digits — drop the cast so text comparison works
-    if dialect == "sqlite":
-        sqlglot_type = None
-
-    return sqlglot_type, lower, upper
-
-
 def to_sqlglot_filter(
```
rudolfix: is this used anywhere? this should produce exactly the same result as your new code. so it should be safe to remove

dlt/extract/incremental/transform.py
```diff
@@ -610,3 +610,52 @@ def _process_null_at_cursor_path(self, tbl: "pa.Table") -> Tuple["pa.Table", "pa
...
+        if ctx is None:
+            # Bare relation, no `.incremental()`. Auto-apply using the parent `Incremental`
+            relation = relation.incremental(self._incremental)
+
+        if self.end_value is not None:
+            # External scheduler/ephemeral mode: state not advanced from observed data.
+            self.seen_data = True
+            return relation, False, False
+
+        if self.range_start != "open":
```
rudolfix: I'd let users do whatever they want: if they want <) or (> range or even () - let them have it. also when "merge" is used - we disable dedup. I'd remove any restrictions but add end to end test with different ranges and make sure correct data is coming

tests/dataset/test_relation_incremental.py
```diff
@@ -0,0 +1,940 @@
...
+        pytest.param("", "non-empty string", id="empty"),
+        pytest.param("col.", "not a plain column identifier", id="trailing-dot"),
+        pytest.param(".col", "not a plain column identifier", id="leading-dot"),
+        pytest.param('"col with.dot"', "not a plain column identifier", id="quoted-with-dot"),
+        pytest.param("$.name", "JSONPath expression", id="jsonpath-root"),
+        pytest.param("items[0]", "JSONPath expression", id="array-index"),
+    ],
+)
+def test_parse_incremental_cursor_path_rejects_malformed(cursor_path: str, match: str) -> None:
+    from dlt.dataset._incremental import _parse_incremental_cursor_path
```
rudolfix: inline import

tests/dataset/test_relation_incremental.py
```diff
@@ -0,0 +1,940 @@
...
+    )
+    with warnings.catch_warnings(record=True) as captured:
+        warnings.simplefilter("always", UserWarning)
+        incremental_dataset.table("events").incremental(incremental)
+    assert (
+        captured == []
+    ), f"unexpected warning for policy={policy!r}: {[str(w.message) for w in captured]}"
+
+
+def _model_transformer(
```
rudolfix: this should IMO go to tests/pipeline in a separate module. we could also keep it in "extract" tests. we are doing end to end tests here, no?

tests/dataset/test_relation_incremental.py
```diff
@@ -0,0 +1,940 @@
...
+        "last_value": 0,
+        "start_value": 0,
+    }
+    relation = incremental_dataset.table("events")
+    incremental_transform = incremental._get_transform(relation)
+    assert isinstance(incremental_transform, ModelIncremental)
+    assert incremental_transform.cursor_path == "id"
+
+
+def test_model_incremental_advances_last_value_for_open_range(
```
rudolfix: incremental is tested in "extract". there are some tests in "pipeline" module (if they are end to end ie. a few subsequent pipeline runs)

tests/dataset/test_relation_incremental.py
```diff
@@ -0,0 +1,940 @@
...
+    )
+    transformer = _model_transformer(start_value=2)
+    out, start_out_of_range, end_out_of_range = transformer(relation)
+
+    assert out is relation
+    assert (start_out_of_range, end_out_of_range) == (False, False)
+    assert transformer.last_value == 5
+
+
+def test_model_incremental_no_advance_in_scheduler_mode(
```
rudolfix: I'd rather use "when_end_value_is_set" - scheduler mode I think leaks from "dlthub" specs

tests/load/test_read_interfaces.py
```diff
@@ -929,6 +929,78 @@ def _ts(d: Any) -> int:
     assert [_ts(d) for d in actual_dts] == [_ts(d) for d in expected_dts]
 
 
+@pytest.mark.no_load
+@pytest.mark.essential
+def test_relation_incremental_datetime_on_dataset(populated_pipeline: Pipeline) -> None:
+    """End-to-end: dataset.table('items').incremental(<datetime cursor>) on every destination.
+
+    Mirrors `test_to_sqlglot_filter_on_dataset` but exercises the `_build_incremental_condition`
```
rudolfix: I think `test_to_sqlglot_filter_on_dataset` is removed now?

tests/dataset/test_relation_incremental_sql.py
```diff
@@ -0,0 +1,344 @@
...
+from dlt.common.pendulum import pendulum
+from dlt.dataset._incremental import (
+    _build_incremental_aggregate,
+    _build_incremental_condition,
+    _RelationIncrementalContext,
+)
+from dlt.extract.incremental import Incremental
+
+
+def _caps(
```
rudolfix: cases in this module should be compared to cases in `test_relation_incremental`, new cases should be moved and this module - deleted

tests/load/test_read_interfaces.py
rudolfix: ahh OK - it is still here: test_to_sqlglot_filter_on_dataset should be deleted.

rudolfix (changes requested):
code is LGTM. tests are good. the problem is in docstrings and comments.

also we do not need `to_sqlglot_filter` - you can expose something like `to_sqlglot_filter` from your code then drop `sql` module. we do not need two places that generates SQL from incremental. maybe I was not clear enough in my previous review

docs are still missing.

dlt/common/destination/capabilities.py
burnash: yes, missed this one

dlt/extract/incremental/sql.py
burnash: I think you've added it in a parallel PR before, was not clear for me if you have other usage for this method tbh.

burnash:
@rudolfix 
>  you can expose something like to_sqlglot_filter 

could you elaborate on that, the use case, or API?

dlt/common/libs/sqlglot.py
burnash: it's not opus, please see the comment above. if you hoist `dlt.common.data_writers.escape`, there's a circular import:

dlt.common.data_writers.escape -> dlt.common.data_writers ->  dlt.common.data_writers.writers -> dlt.common.data_writers.exceptions -> dlt.common.destination -> dlt.common.destination.capabilities -> dlt.common.libs.sqlglot

I can move `resolve_timestamp_cast` somewhere else, but I think it belongs here. Let me know if you have a better place for it in mind.

dlt/common/destination/capabilities.py
burnash: Fixed in 410f76b

dlt/dataset/_incremental.py
burnash: Removed in 410f76b

dlt/dataset/_incremental.py
burnash: Fixed in 410f76b

dlt/dataset/_incremental.py
burnash: Fixed in 410f76b

dlt/dataset/_join.py
burnash: Done in 410f76b

dlt/dataset/relation.py
burnash: Removed in 410f76b

dlt/dataset/relation.py
burnash: Removed in 410f76b

dlt/dataset/relation.py
burnash: Removed in 410f76b

dlt/extract/incremental/sql.py
burnash: Removed the module & function in 410f76b

dlt/extract/incremental/transform.py
burnash: Removed the condition in 410f76b

tests/dataset/test_relation_incremental.py
burnash: Fixed in 410f76b

tests/dataset/test_relation_incremental.py
burnash: Moved to tests/pipeline/test_model_incremental.py in 410f76b

tests/dataset/test_relation_incremental.py
burnash: Fixed in 410f76b

tests/load/test_read_interfaces.py
burnash: Deleted in 410f76b

tests/dataset/test_relation_incremental_sql.py
burnash: Moved new cases to tests/dataset/test_relation_incremental.py in 410f76b

tests/dataset/test_relation_incremental.py
burnash: Renamed

dlt/common/libs/sqlglot.py
rudolfix: OK! I'd say the problem is in `data_writers` - we can easily move escape function to a separate module... not in this ticket though

rudolfix (dismissed):
LGTM!

dlt/extract/incremental/transform.py
```diff
@@ -610,3 +610,35 @@ def _process_null_at_cursor_path(self, tbl: "pa.Table") -> Tuple["pa.Table", "pa
...
+class ModelIncremental(IncrementalTransform):
+    """Incremental transform for `Relation` items.
+
+    Filtering happens via SQL pushdown when `Relation.incremental(cursor)` is applied.
+    When `end_value` is set, state is not advanced from observed data; otherwise the
+    aggregate over the filtered relation advances `last_value`.
+    """
+
+    # parent `Incremental` so we can auto-apply below
+    _incremental: Optional["Incremental[Any]"] = None  # type: ignore[name-defined]  # noqa: F821
```
rudolfix: FYI: when you do this you are creating a class, not instance attribute, you can IMO remove this None assignment. At this point I do not see any better way to pass Incremental instance here without refactoring a lot of code.

burnash: yes, Python gotcha. I'll drop None.

burnash: Fixed in 08a026f

PR #3890, opened 2026-04-23
fix: clearer error from `dlt.attach()` when pipeline cannot be restored

### LINKED ISSUE #3394: improve `pipeline.attach` error message
**Background**
A common problem with `pipeline.attach` is the lack of local pipeline state from which properties like destination and dataset names could be extracted. There's a fallback that will attempt to sync from destination if destination is provided (explicitly or configured).

Current exception message does not explain what happened and how to remedy this
Pipeline with `pipeline_name=eth_transform_pipeline` and `pipelines_dir=/tmp/.tmpIjHOe4/app/.dlt/.var/prod/pipelines/eth_transform_pipeline` could not be restored: the pipeline was not found in /tmp/.tmpIjHOe4/app/.dlt/.var/prod/pipelines/eth_transform_pipeline found and no destination was provided to restore from.

**Tasks**
There a few cases in the `pipeline.attach` that need improvement (see all the exception raised). We need to explain each case better.
1. no pipeline working dir and no destination: explain that user tried to attach to an existing pipeline and it seems that the pipeline was never run because its state and working dir is missing. Explain that we can obtain the state from destination but we need destination name/type and dataset name (in args to attach or via config)
2. in case when sync_destination happened: explain the above and that we tried destination + dataset but we could find pipeline state there (add names to exception message)

In both cases suggest that if user wants just a dataset, user should use `dlt.dataset()` which does not require pipeline state to access data

bjoaquinc:
### Description
This PR rewrites the `CannotRestorePipelineException` raised by `dlt.attach()` when pipeline state cannot be restored. The previous error told users only that the pipeline was not found at a path, without explaining what was missing or how to recover. The new messages name the requirements, show a concrete example interpolated with the user's own pipeline name, and offer `dlt.dataset()` as a lighter-weight alternative when pipeline state is not actually needed. Also suppresses a redundant inner exception that was cluttering tracebacks.

### Related Issues

- Resolves #3394 

### Changes
1. **Rewrote the "no destination" error message** — names the required inputs, shows a concrete `dlt.attach(...)` example with the user's pipeline name, and offers `dlt.dataset()` as a lighter alternative.

2. **Rewrote the "remote state not found" error message** — reports both locations checked and offers a `dlt.dataset()` example interpolated with the user's actual destination and dataset.

3. **Suppressed the inner exception chain with `raise ... from None`** — hides a redundant duplicate exception from tracebacks while preserving `__context__` for debuggers and observability tools.

4. **Updated test assertions** in `tests/load/pipeline/test_pipelines.py` and `tests/workspace/cli/common/test_cli_invoke.py` to match the new wording.

5. **Corrected the inline comment** above the raise (`env` → `config`).

### DIFF
--- a/dlt/pipeline/__init__.py
+++ b/dlt/pipeline/__init__.py
@@ -244,15 +244,24 @@ def attach(
         )
 
         # we cannot restore the pipeline if the destination is not provided
-        # or found in the env, wipe the working folder and re-raise
+        # or found in the config, wipe the working folder and re-raise
         if not p.destination:
             p._wipe_working_folder()
             raise CannotRestorePipelineException(
                 pipeline_name,
                 p.working_dir,
-                f"the pipeline was not found in {p.working_dir} found and no destination was"
-                " provided to restore from.",
-            )
+                f"No local pipeline state found in '{p.working_dir}'.\n\n"
+                "dlt can restore pipeline state from a destination, but requires a destination."
+                " Pass it as an argument to dlt.attach() or configure it via dlt's"
+                " config system.\n\n"
+                f'`dlt.attach(pipeline_name="{pipeline_name}", destination="<your-destination>",'
+                ' dataset_name="<your-dataset>")`\n\n'
+                f"Note: if no dataset_name is provided, dlt looks in '{pipeline_name}_dataset'"
+                " by default.\n\n"
+                "If you only need to read or query data without pipeline state, use"
+                " `dlt.dataset()` instead:\n\n"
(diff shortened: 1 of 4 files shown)


bjoaquinc:
None of the errors seem related to the PR

zilto (approved):
Changes LGTM. You're right that CI is leaky.

Though the pyarrow exceptions are a bit worrisome: https://github.com/dlt-hub/dlt/actions/runs/24842341762/job/72720205651?pr=3890

PR #3894, opened 2026-04-27
Preserve string encoding for UUID columns under pyarrow 24+

### LINKED ISSUE #3893: `row_tuples_to_arrow` switches UUID columns from `string` to `extension<arrow.uuid>` on pyarrow 24+
On pyarrow 24.0.0, UUID columns with no explicit dlt `data_type` are inferred as the native `extension<arrow.uuid>` type instead of the `string` type dlt has produced in prior version.

CI: https://github.com/dlt-hub/dlt/actions/runs/24839586689/job/73159319854?pr=3889#step:18:439

Repro:

```python
from uuid import uuid4
from dlt.common.libs.pyarrow import row_tuples_to_arrow
from dlt.common.destination import DestinationCapabilitiesContext

rows = [(uuid4(),), (uuid4(),)]
result = row_tuples_to_arrow(
    rows,
    DestinationCapabilitiesContext().generic_capabilities(),
    columns={"uuid_col": {"name": "uuid_col"}},
    tz="UTC",
)
print(result[0].type)  # pyarrow 23: string,  pyarrow 24: extension<arrow.uuid>
```

Likely introduced by https://github.com/apache/arrow/pull/48727

burnash:
Fixes #3893

### DIFF
--- a/dlt/common/libs/pyarrow.py
+++ b/dlt/common/libs/pyarrow.py
@@ -49,6 +49,7 @@ import ctypes
 TAnyArrowItem = Union[pyarrow.Table, pyarrow.RecordBatch]
 
 ARROW_DECIMAL_MAX_PRECISION = 76
+ARROW_UUID_EXTENSION_NAME = "arrow.uuid"
 
 
 class UnsupportedArrowTypeException(DltException):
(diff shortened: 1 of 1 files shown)


zilto (approved):
LGTM. Ideally, we could catch regressions where `pyarrow` is updated and the output type changes.

This would be possible with parameterized type conversion tests `@pytest.mark.paramerize` similar to what we do with `sqlglot`

PR #3897, opened 2026-04-28
Chore: ci from fork

### LINKED ISSUE #3850: Chore: `ci from fork` label triggers remote tests but secrets are unavailable
When a maintainer adds the ci from fork label to a fork PR, the remote destination tests (test_destinations_remote, test_tools_dbt_runner) run but fail because secrets are empty. 

Root cause: PR #2728 consolidated all workflows into main.yml using the pull_request trigger. The previous per-destination workflows used pull_request_target, which exposes secrets to fork PRs by running the base branch's workflow definition. With pull_request, GitHub blocks secret access for forks regardless of labels or secrets: inherit. 

The `authorize_run_from_fork` job gates whether secret-requiring jobs run, but can't make secrets available, which is determined by the trigger type.  

### Proposed fix

Extract the secret-requiring jobs (test_destinations_remote, test_tools_dbt_runner) into a separate workflow triggered by pull_request_target. Keep main.yml on pull_request for everything else (lint, common tests, local destinations).

#### Details

- Create `fork_tests_with_secrets.yml`, which is `a pull_request_target` workflow that calls `test_destinations_remote.yml and `test_tools_dbt_runner.yml` with secrets. Gate job checks: PR is from a fork and has ci from fork label. Internal PRs skip   entirely (no duplicate runs).                                                                                                        
- In `main.yml`, we rename `authorize_run_from_fork` → `authorize_secrets`, simplify condition to internal PRs only. Remove ci from fork label check. Fork PRs still get lint, common tests, local destinations.

> It would be nice to keep the previous UX where a maintainer would add the `ci-from-fork` label once and then approve workflows with each new commit. The above proposal may require removing and adding the label each time, unless we adjust things in the repo settings.

PS: workflow YAML changes in PRs won't take effect for the remote test workflow until merged

anuunchin:
This PR adjusts the main workflow in the way that:
- It runs tests that require secrets only on internal PRs (where the head repo matches the target repo)
- PRs from forks use the new `fork_tests_with_secrets.yml` workflow that runs secret-requiring tests, triggered when:
    - The `ci from fork` label is added (needs to be added once)
    - New pushes to the fork branch automatically re-trigger the workflow (no need to re-label) 
    - Each triggered run requires a maintainer to manually approve it via the fork-ci environment before secrets are exposed  

> This will need a new environment `fork-ci` with `Required reviewers` enabled under `Environment protection rules`

Resolves #3850

### DIFF
--- a/.github/workflows/fork_tests_with_secrets.yml
+++ b/.github/workflows/fork_tests_with_secrets.yml
@@ -2,8 +2,11 @@ name: dlt | fork tests with secrets
 
 on:
   pull_request_target:
-    branches: [master, devel, runtime]
+    branches: [devel]
     types: [labeled, synchronize]
+
+permissions:
+  contents: read
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
--- a/.github/workflows/fork_tests_with_secrets.yml
+++ b/.github/workflows/fork_tests_with_secrets.yml
@@ -1,0 +1,36 @@
+name: dlt | fork tests with secrets
+
+on:
+  pull_request_target:
+    branches: [master, devel, runtime]
+    types: [labeled, synchronize]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+
+  authorize:
+    name: authorize fork run (requires reviewer approval)
+    if: >-
+      github.event.pull_request.head.repo.full_name != github.repository
+      && contains(github.event.pull_request.labels.*.name, 'ci from fork')
+    runs-on: ubuntu-latest
+    environment: fork-ci
+    steps:
+      - run: echo "Fork PR authorized — running secret-requiring tests against ${{ github.event.pull_request.head.sha }}"
+
+  test_destinations_remote:
+    name: test remote destinations with secrets (fork)
+    needs: [authorize]
+    uses: ./.github/workflows/test_destinations_remote.yml
+    secrets: inherit
+    with:
+      run_full_test_suite: ${{ contains(github.event.pull_request.labels.*.name, 'ci full') }}
+
+  test_tools_dbt_runner:
+    name: test dbt runner (fork)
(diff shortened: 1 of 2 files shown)


.github/workflows/fork_tests_with_secrets.yml
```diff
@@ -0,0 +1,36 @@
+name: dlt | fork tests with secrets
+
+on:
+  pull_request_target:
+    branches: [master, devel, runtime]
```
anuunchin: We could keep only devel as well

rudolfix: only `devel`

.github/workflows/fork_tests_with_secrets.yml
```diff
@@ -0,0 +1,36 @@
...
+    types: [labeled, synchronize]
+
```
rudolfix, line 7: I'd add `permissions: contents: read` just in case. right now we rely on org setting.

.github/workflows/main.yml
```diff
@@ -16,11 +16,11 @@ jobs:
     name: docs changes
     uses: ./.github/workflows/get_docs_changes.yml
 
-  # if the PR is from a fork, we need to authorize the secrets access and remote destinations run with a lable
-  authorize_run_from_fork:
-    name: check if fork and if so wether secrets are available
-    # run when label is assigned OR when we are not a fork
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ci from fork') || (github.event.pull_request.head.repo.full_name == github.repository && (github.event.action == 'opened' || github.event.action == 'synchronize'))}}
+  # gate for secret-requiring jobs: secrets are only available for internal (same-repo) PRs under pull_request.
```
rudolfix: we must ban the word "gate" not sure where Opus gets it. use "check" or "constraint"

rudolfix (changes requested):
requiring approval from reviewers is good. this needs to get to master before any fork will see those workflows so I assume we can't test it really

rudolfix (approved):
LGTM! thanks for fixing this!

PR #3898, opened 2026-04-28
Make the naming-convention check in `Schema.unify_schemas()` opt-in

burnash:
- Make the naming-convention check in `Schema.unify_schemas()` opt-in
- Drop max_length tests workaround


rudolfix (changes requested):
code is good but we need a test that will unify two schemas with very different naming conventions. I'll take
- snake_case
- direct
- upper case (available in test fixtures)

add some tables, including nested tables (those have different table separators). add some emojis to the names

then unify schemas and make sure you get the expected (verbatim) names. bonus for end to end test where we use this schema to query tables from duckdb (full unicode support in names - should work)

rudolfix (dismissed):
LGTM!

rudolfix (approved):
LGTM!

PR #3901, opened 2026-04-28
fix: correctly materialize empty tables for multi-table resources

### LINKED ISSUE #3840: Empty table materialization does not work for multi-table resources

anuunchin:
Continuation of #3814

This PR adjusts the extractor in the way that it tracks empty tables, not resources to allow multi empty table materialization.

A relevant tests is added.

Resolves https://github.com/dlt-hub/dlt/issues/3840

### DIFF
--- a/dlt/extract/extract.py
+++ b/dlt/extract/extract.py
@@ -345,41 +345,49 @@ class Extract(WithStepInfo[ExtractMetrics, ExtractInfo]):
     ) -> None:
         schema = source.schema
         json_extractor = extractors["object"]
-        resources_with_items = set().union(*[e.resources_with_items for e in extractors.values()])
+        tables_with_items = set().union(*[e.tables_with_items for e in extractors.values()])
         # find REPLACE resources that did not yield any pipe items and create empty jobs for them
         # NOTE: do not include tables that have never seen data
         data_tables = {t["name"]: t for t in schema.data_tables(seen_data_only=True)}
         tables_by_resources = utils.group_tables_by_resource(data_tables)
         for resource in source.resources.selected.values():
-            if resource.write_disposition != "replace" or resource.name in resources_with_items:
-                continue
             if resource.name not in tables_by_resources:
                 continue
             for table in tables_by_resources[resource.name]:
+                write_disposition = table.get("write_disposition") or resource.write_disposition
+                if write_disposition != "replace" or (resource.name, table["name"]) in tables_with_items:
+                    continue
                 # we only need to write empty files for the root tables
                 if not utils.is_nested_table(table):
                     json_extractor.write_empty_items_file(table["name"])
 
         # collect resources that received empty materialized lists and had no items
-        resources_with_empty = (
+        tables_with_empty = (
             set()
-            .union(*[e.resources_with_empty for e in extractors.values()])
-            .difference(resources_with_items)
+            .union(*[e.tables_with_empty for e in extractors.values()])
+            .difference(tables_with_items)
         )
         # get all possible tables
         data_tables = {t["name"]: t for t in schema.data_tables()}
         tables_by_resources = utils.group_tables_by_resource(data_tables)
-        for resource_name in resources_with_empty:
+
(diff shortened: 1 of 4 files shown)


dlt/extract/extract.py
```diff
@@ -345,42 +345,30 @@ def _write_empty_files(
...
-            .difference(resources_with_items)
+            .union(*[e.tables_with_empty for e in extractors.values()])
+            .difference(tables_with_items)
         )
-        # get all possible tables
-        data_tables = {t["name"]: t for t in schema.data_tables()}
-        tables_by_resources = utils.group_tables_by_resource(data_tables)
-        for resource_name in resources_with_empty:
-            if resource := source.resources.selected.get(resource_name):
-                if tables := tables_by_resources.get("resource_name"):
```
rudolfix, line 374: probably `"resource_name"` -> `resource_name` would also fix this problem. what a stupid bug. also it is clear that it was never tested

rudolfix: OK that was also wrong for other reasons

rudolfix (approved):
LGTM! there was trying to materialize nested tables but it was wrong in many different ways. your is more elegant. I added two end to end tests

burnash:
Reverting this fix due to data truncation issue reported in #3998

PR #3477, opened 2025-12-13
Enhancement Feature: Add Arrow IPC writer

rory-data:
### Description
This pull request adds support for the [Apache Arrow IPC (Feather v2) file format](https://arrow.apache.org/docs/format/Columnar.html#ipc-file-format) as a new data writer, allowing both Python object data and Arrow tables to be written in IPC format i.e. `.arrow` file extension. It introduces new configuration options, updates destination capabilities, and makes necessary typing and import adjustments to support the new format. 

A user can select writing to Arrow IPC via the `file_format="arrow"` hint on a resource. This hint will switch the Arrow writer from the default Parquet file format to use Arrow IPC and write accordingly

### Related Issues
- Closes #3452 

---

### Change Summary
* Added `ArrowIPCDataWriter` for writing Python object data to Arrow IPC files, and `ArrowToArrowIPCWriter` for writing Arrow tables/batches directly to IPC files. Both various IPC-specific options, including internal IPC compression. 
* Introduced `ArrowIPCFormatConfiguration` for configuring IPC writer options such as format type, compression, and threading. 
* Updated `DestinationCapabilitiesContext` to include `arrow` and added "arrow" to supported loader file formats, enabling destinations to declare IPC support. 
* Added `arrow` to the `TLoaderFileFormat` type alias to recognise Arrow IPC as a valid file format i.e. `.arrow` file extension. 
* Registered Arrow IPC with extract and storage processes using `arrow_ipc` as an alias for the `arrow` data file format so to not impact global structures. This includes a switcher for writing Arrow formats based on the resource `file_format` hint
* Refactored imports and typing throughout to support new writers and configuration. 
* Registered new writers as native writers and adjusted code to ensure proper handling of Arrow and IPC formats. 
* Minor import order and dependency cleanups for clarity and maintainability. 

---

### Additional Context
The Apache Arrow IPC Feather (v2)  file format is a lightweight, language-agnostic, and high-performance columnar file format for storing Apache Arrow Tables or data frames. 

Key features:
- Adopts the Arrow Interprocess Communication (IPC) File Format, ensuring compatibility and stability with all Arrow implementations.
- Supports all Apache Arrow data types, including nested and complex types.
- Natively supports fast compression codecs like LZ4 (default) and ZSTD.
- Allows for efficient random access and slicing of data without needing to load the entire file into memory.

PR #3904, opened 2026-04-29
feat: Databricks Zerobus loading

### LINKED ISSUE #3874: feat(databricks): Support ingestion via Zerobus
# Summary
Extend the Databricks destination with Zerobus load job. This is similar to BigQuery streaming load ([code](https://github.com/dlt-hub/dlt/blob/d68ab5f5f014c75f1f4e8680feaaa0be029374cc/dlt/destinations/impl/bigquery/bigquery.py#L584), [docs](https://github.com/dlt-hub/dlt/blob/d68ab5f5f014c75f1f4e8680feaaa0be029374cc/dlt/destinations/impl/bigquery/bigquery.py#L584))

# Context
[Zerobus](https://docs.databricks.com/aws/en/ingestion/zerobus-overview) is a "no-infra" product for streaming / near-real time ingestion. When toggled on, it opens an HTTP endpoint where data can pushed to and Databricks ingest it in the associated table.

see [Python SDK](https://github.com/databricks/zerobus-sdk/tree/main/python)
see [guide](https://docs.databricks.com/aws/en/ingestion/zerobus-ingest)

# Scope
- extend Databricks config to allow Zerobus settings (try to align with BigQuery streaming)
- support passing an acknowledgement callback to `dlt.destinations.databricks()` (see guide)
- proper error handling: Zerobus will throw errors on invalid schemas; dlt is responsible for evolving the schema via DDL (see guide)
- Add zerobus logic unit tests
- Add to databricks destination with zerobus to parameterized destination tests
- allow the `destination_client`  to grant permission for Zerobus for the table ([see here](https://docs.databricks.com/aws/en/ingestion/zerobus-ingest#create-a-service-principal-and-grant-permissions))
  - this would be a great quality of life improvement; it's optional for v1 if it becomes very complicated (could encounter issues because toggling Zerobus requires more elevated priviledges than using it)

## out of scope (for v1)
- asynchronous Zerobus SDK support
- Protobuf Zerobus SDK support

jorritsandbrink:
### Description
Adds support for using [Zerobus](https://github.com/databricks/zerobus-sdk) to load data into Databricks Delta tables.

API:
- use Zerobus loading: `databricks_adapter(my_resource, insert_api="zerobus")`
- use "standard" loading: do not set `insert_api` or set it to `copy_into`

Notes:
- the API is consistent with `bigquery`'s API:  `bigquery_adapter(my_resource, insert_api="streaming")`
- the `zerobus` insert API is only supported for the `append` write disposition

### Related Issues

Closes #3874


jorritsandbrink:
This PR is not ready to merge (it's missing docs, tests, and config options), but before polishing it I want to get feedback on a performance concern: the new `zerobus` insert API is significantly slower than the existing `copy_into` insert API.

I've done perf tests using NYC taxi data, which identified three main bottlenecks for `zerobus`:
1. record acknowledgement (server side)
2. load job file reading (client side)
3. load job record encoding (client side)

### 1. Record acknowledgment
I think this one is out of our control, though we can skip waiting for acknowledgement at the cost of correctness guarantees (I've tested that and it makes things much faster, though still not faster than `copy_into`).

### 2. load job file reading (client side)
We have to read the JSON/Parquet load job file back into Python memory, so we can encode the records and feed it into the Zerobus SDK.

### 3. load job record encoding (client side)
We have to do row-by-row processing to encode the values into a format that the Zerobus SDK accepts.

### Squeezing Zerobus into the wrong mold?
Current `zerobus` pipeline:

_resource emits dicts -> `extract` writes to disk -> `normalize` reads from disk, does per record processing and writes to disk -> `load` reads from disk, does per record processing, and emits into Zerobus SDK_

I think this doesn't fit the Zerobus streaming paradigm well. Something like _resource emits dicts into Zerobus SDK_ would be a more natural fit.

### Zerobus Arrow support
Databricks is working on [Arrow support](https://github.com/databricks/zerobus-sdk/issues/19) which looks like it might land soon. I think this will speed up or completely remove the need for (2) and (3). Together with skipping record acknowledgment, this might push `zerobus` perf ahead of `copy_into` perf.

@zilto what are your thoughts?

zilto:
Thanks for the performance analysis! Code looks good. 

## My understanding
To double-check my understanding, when comparing "datatbricks" vs "databricks zerobus"
1. extract and normalize steps are identical; only load step differs
2. we're using Zerobus with: JSON encoding,  record acknowledgement (`ingest_record_offset()`), synchronously

# Performance improvement avenues
## Arrow support
Zerobus Arrow support would be a quick win. We have a solid but incomplete PR (#3477) for Arrow IPC support that would skip a lot of the serialization costs in extract and normalize phases.

## Parallelize load phase
`dlt` comes from batch world and `zerobus` from stream world. Typically, `dlt` scales by batching records and making a big INSERT with transactional guarantees. OTOH, `zerobus` scales by having more connections push data (what the docs indicate). 

AFAIK, `dlt` will parallelize load step per load package file ([docs](https://dlthub.com/docs/reference/performance#load)). We can try setting the `normalize.data_writer.file_max_bytes` and see how it compares to `copy_into` insert?

## Disabling normalize and load steps
I agree with you that those features would be desirable more generally for `dlt` (trade-off in features, guarantees, and performance). Though, I think we should explore them in separate PRs

jorritsandbrink:
> To double-check my understanding, when comparing "datatbricks" vs "databricks zerobus"
> 
> extract and normalize steps are identical; only load step differs
> we're using Zerobus with: JSON encoding, record acknowledgement (ingest_record_offset()), synchronously

Correct!

Furthermore, I have tried:
- `ingest_records_nowait()` instead of  `ingest_records_offset()` — this shifts waiting for acknowledgment to `stream.close()`, and thus doesn't help much (skipping `stream.close()` does actually save a lot of time, but comes at the expense of correctness guarantees)
- parallellizing work accross load jobs by setting `normalize.data_writer.file_max_bytes` as you suggested — this definitely reduces load time, but still loses to `copy_into`
- tuning different parameters such as number of parallel loads, size of the batch yielded by the `dlt` resource, size of the batch yielded into the Zerobus stream

None of it led to `zerobus` being faster than `copy_into`. Even on workloads that should favor Zerobus, such as a resource yielding Python dictionaries and using the `jsonl` file format, `copy_into` was faster. With a resource that yields Arrow data and the `parquet` file format, `copy_into` was much faster.

@zilto I'm happy to polish the PR to finalize it, but the current Zerobus performance makes me doubt whether there is any merit to merging this right now. Let me know what you think.

jorritsandbrink:
Switched Zerobus serialization over from JSON to Arrow, which became available recently in `databricks-zerobus-ingest-sdk==1.2.0`. This made the code both simpler and faster. I've been able to get the `zerobus` path faster than the `copy_into` path in a 3M row NYC taxi benchmark by tuning the batch size and using `ZSTD` as `ipc_compression`.

Important to note that Arrow isn't officially supported yet in the Zerobus Python SDK. It's not documented, but there is an [example](https://github.com/databricks/zerobus-sdk/blob/16b5fd4db2c416ad3671fe3411c5e8dd4743ba57/python/examples/sync_example_arrow.py) in the repo.

@zilto Can you review this PR?

dlt/common/typing.py
```diff
@@ -125,6 +125,10 @@ class SecretSentinel:
 """A single data item as extracted from data source"""
 TDataItems: TypeAlias = Union[TDataItem, List[TDataItem]]
 "A single data item or a list as extracted from the data source"
+TDataRecord = dict[str, Any]
+"""Table row dictionary. Not guaranteed to be JSON serializable without custom encoding."""
+TDataRecordBatch = list[TDataRecord]
+"""List of table row dictionaries. Not guaranteed to be JSON serializable without custom encoding."""
```
zilto, line 131: I like the centralization of batching logic. Though, how does `TDataRecord` and `TDataRecordBatch` differ from `TDataItem` and `TDataItems`?

(it's a bit annoyting that `TDataItems` is not guaranteed to be a list...)

dlt/destinations/impl/databricks/databricks.py
```diff
@@ -1,56 +1,105 @@
...
+    Dict,
+    Generic,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Union,
+    cast,
+)
+from urllib.parse import urlparse
```
zilto, line 25: duplicate imports with lines 3-6. Please fix. Otherwise, can do in follow-up

tests/load/utils.py
```diff
@@ -458,6 +458,21 @@ def destinations_configs(
...
+    ]
+    cid_configs_by_cid = {config.cid: config for config in cid_configs if config.cid is not None}
+
+    # default sql configs that are also default staging configs
+    default_sql_configs_with_staging = [
+        # Athena needs filesystem staging, which will be automatically set; we have to supply a bucket url though.
+        cid_configs_by_cid["athena"],
+        cid_configs_by_cid["athena-iceberg"],
+        cid_configs_by_cid["athena-s3-tables"],
     ]
```
zilto, line 476: CI workflows related to these configurations seem to be failing now. I retried them and they failed again.

zilto (changes requested):
We need to investigate:
- why `athena` tests are failing. Some of their config were edited in this PR
- why `hf` tests are failing. We're hitting rate limits. Do you expect any behavior change from this PR (e.g., batching logic)? 

If those are unrelated, we can merge.

Otherwise, two minor questions / nits.

dlt/destinations/impl/databricks/databricks.py
jorritsandbrink: fixed in https://github.com/dlt-hub/dlt/pull/3904/commits/0faec9939ffba743c4192b567cad012c6bdddcd1

dlt/common/typing.py
jorritsandbrink: I think they roughly represent the same concepts. My issue is that `TDataItem` is just `Any`, which makes dev hard.

tests/load/utils.py
jorritsandbrink: It's not related to changes in this PR. The failing tests also fail on `origin/devel`.

zilto (approved):
The failing CI is also failing on `devel` branch, so seems unrelated.

Let's merge. Good job!

PR #3905, opened 2026-04-29
add join compatibility based on `physical_location` to destination configurations

### LINKED ISSUE #3758: feat: implement join compatibility check for destinations

Travior:
These are the currently enforced rules, together with the rules that diverge from the original ticket:
Another note: the accessor is called `physical_location()`, as big query already defines a config property called `location`.

Here are the tables formatted in Markdown:

### Compatibility Overview vs Spec

| Destination | Current Location / Identity | Current `can_join_with` Rule | Divergence From Spec |
| :--- | :--- | :--- | :--- |
| Postgres | host:port | Same host:port and same database | Matches |
| Redshift | host:port | Same host:port and same database | Matches |
| Snowflake | account host | Same account host | Matches |
| BigQuery | project_id | Same project ID | Matches |
| MSSQL | host:port | Same host:port; database may differ | Minor: spec says host, implementation includes port |
| Synapse | host:port | Same host:port; database may differ | Minor: spec says inherited host, implementation includes port |
| ClickHouse | host:port | Same host:port; database may differ | Minor: spec says host, implementation includes port |
| Databricks | server_hostname | Same server hostname | Matches |
| Athena | region/catalog | Same AWS region and same data catalog | Diverges: spec says location is only aws_data_catalog; implementation also requires same AWS region |
| Dremio | host:port | Same host:port | Minor: spec says host, implementation includes port |
| DuckDB | database file path or :memory: | Same database path | Matches |
| MotherDuck | physical_destination() is empty; fingerprint/token used separately | Same non-empty access token | Diverges structurally: spec says location is access token and default same-location rule; implementation intentionally does not expose token as location and overrides can_join_with |
| Filesystem | remote: scheme://netloc; local: "" | Always True with another filesystem config | Matches |
| DuckLake | credential-free catalog identity + ducklake_name | Same catalog identity and same DuckLake name | Matches |

### SQLAlchemy Overview vs Spec

| Dialect | Current Rule | Divergence From Spec |
| :--- | :--- | :--- |
| postgresql | Same host:port and same database | Matches |
| mysql | Same host:port; database may differ | Matches |
| mssql | Same host:port; database may differ | Matches |
| oracle | Same host:port; database may differ | Matches |
| db2 | Same host:port; database may differ | Matches |
| sqlite | Same database file path | Matches |
| unknown/other | Same host:port and same database | Matches |

### Additional Current Definitions Not In Spec

| Destination | Current Rule |
| :--- | :--- |
| LanceDB | Same LanceDB URI and same dataset_separator; dataset name may differ |
| Lance | Same catalog root and same bound dataset name |
| Weaviate | Never joinable |
| Qdrant | Never joinable |

Still TODO/unclear:
- CI secrets.toml (or config?) needs `destination.snowflake.join_compatibility_database`
- integration tests intentionally don't execute explicit joins yet as query APIs don't properly support cross destination/database data access (e.g. snowflake qualified name doesn't include the database / filesystem client doesn't expose option to register "foreign" file as duckdb view)
- Should we explicitly add more databases to the tests (when destination supports the join like snowflake), if so we need to manually set up different more databases / catalogs for those providers


dlt/dataset/dataset.py
```diff
@@ -498,12 +498,11 @@ def get_dataset_sql_client(dataset: dlt.Dataset) -> SqlClientBase[Any]:
 
 
 def is_same_physical_destination(dataset1: dlt.Dataset, dataset2: dlt.Dataset) -> bool:
```
rudolfix, line 506: let's deprecate this (we have decorator somewhere) and then remove when `dlthub` is fixed

dlt/destinations/impl/duckdb/configuration.py
```diff
@@ -318,5 +318,11 @@ def __init__(
         )
         self.create_indexes = create_indexes
 
+    def physical_destination(self) -> str:
+        """Returns the database file path or ':memory:'."""
+        if self.credentials and self.credentials.database:
```
rudolfix, line 323: `duckdb` is an interesting case. we could easily join any database with another (including this new "hosted" duckdb available via host:port) if we do ATTACH command. that is something to do in the future ie. by extending #3747 implementation - to do ATTACH on demand when foreign dataset is detected.

dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -44,6 +45,37 @@ class FilesystemDestinationClientConfiguration(FilesystemConfigurationWithLocalF
...
+
+        url = urlparse(self.bucket_url)
+        return f"{url.scheme}://{url.netloc}"
+
+    def fingerprint(self) -> str:
+        # Explicit override to resolve MRO ambiguity: without it, Python picks
+        # FilesystemConfiguration.fingerprint() (which hashes the raw bucket URL)
+        # over DestinationClientConfiguration.fingerprint() (which hashes
+        # physical_destination()).  Do not remove.
+        return DestinationClientStagingConfiguration.fingerprint(self)
```
rudolfix: please tests this. MRO can change ie. someone may add another base class to it

dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -44,6 +45,37 @@ class FilesystemDestinationClientConfiguration(FilesystemConfigurationWithLocalF
     def resolve_credentials_type(self) -> Type[CredentialsConfiguration]:
         return super().resolve_credentials_type()
 
+    def physical_destination(self) -> str:
+        """Returns scheme://netloc for remote filesystems, or "" for local."""
+        if not self.bucket_url:
+            return ""
+
+        if self.is_local_path(self.bucket_url):
+            return ""
```
rudolfix: I think we should return normalized (abs) path. AFAIK fingerprint didn't do it - it is by design forcing all local paths into a single fingerprint not to inflate telemetry destinations. here it is not a concern

dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -44,6 +45,37 @@ class FilesystemDestinationClientConfiguration(FilesystemConfigurationWithLocalF
...
+        return DestinationClientStagingConfiguration.fingerprint(self)
+
+    def can_join_with(self, other: DestinationClientConfiguration) -> bool:
+        """Returns True for any other filesystem destination.
+
+        Filesystem tables are queried through a local engine (e.g. DuckDB) that
+        can access multiple storage backends in a single query, so join
+        compatibility is determined by the engine, not by the storage location.
+        """
+        if isinstance(other, FilesystemDestinationClientConfiguration):
```
rudolfix: NOTE: this will require extension of #3747 - ATTACH foreign duckdb with all its views and tables to current duckdb in the dataset and using 3 part qualification when binding queries. should be "easily" doable. not this ticket though!

dlt/destinations/impl/lance/configuration.py
```diff
@@ -353,5 +356,23 @@ def make_namespace(self) -> "LanceNamespace":
             props.update(self.credentials.to_namespace_properties())
         return connect(self.catalog_type, props)
 
-    def fingerprint(self) -> str:
-        return self.storage.fingerprint() if self.storage else ""
+    def physical_destination(self) -> str:
+        """Returns the resolved Lance catalog root."""
+        if (
```
rudolfix, line 465: Another interesting case. We could just ATTACH duckdb in lance sql_client to existing sql_client (in a dataset). in that case lance is joinable with any filesystem and duckdb-based sql_client. we can keep that for now.

dlt/destinations/impl/filesystem/configuration.py
```diff
@@ -44,6 +45,37 @@ class FilesystemDestinationClientConfiguration(FilesystemConfigurationWithLocalF
...
 
+    def physical_destination(self) -> str:
+        """Returns scheme://netloc for remote filesystems, or "" for local."""
+        if not self.bucket_url:
+            return ""
+
+        if self.is_local_path(self.bucket_url):
+            return ""
+
+        from urllib.parse import urlparse
```
rudolfix: pls import at the top

dlt/destinations/impl/motherduck/configuration.py
```diff
@@ -123,12 +126,29 @@ class MotherDuckClientConfiguration(DestinationClientDwhWithStagingConfiguration
...
     def fingerprint(self) -> str:
-        """Returns a fingerprint of user access token"""
+        """Returns a fingerprint of user access token."""
         if self.credentials and self.credentials.password:
             return digest128(self.credentials.password)
         return ""
 
+    def can_join_with(self, other: DestinationClientConfiguration) -> bool:
+        """Returns True for MotherDuck configs with the same token."""
+        if not isinstance(other, MotherDuckClientConfiguration):
```
rudolfix, line 141: good for now. but I'm pretty sure we'll be able to join databases from different accounts via ATTACH

dlt/destinations/impl/sqlalchemy/configuration.py
```diff
@@ -262,3 +265,66 @@ def on_resolved(self) -> None:
...
+            return database or ""
+
+        if host:
+            # Default-vs-explicit port mismatches may reject otherwise valid joins.
+            if port:
+                return f"{host}:{port}"
+            return host
+        return ""
+
+    def can_join_with(self, other: DestinationClientConfiguration) -> bool:
```
rudolfix: you could eliminate some code below (IMO) via 2 helper classes:
* we can join two physical destinations
* as above but also database name must mach

helper 1 can call helper 2. also helper 1 looks like super().can_join_with() where we just compare physical locations.

dlt/common/destination/client.py
```diff
@@ -163,10 +164,29 @@ class DestinationClientConfiguration(BaseConfiguration):
 
     __recommended_sections__: ClassVar[Sequence[str]] = (known_sections.DESTINATION, "")
 
+    def physical_destination(self) -> str:
```
rudolfix: `physical_location`? the original called for `location`

dlt/common/destination/client.py
```diff
@@ -163,10 +164,29 @@ class DestinationClientConfiguration(BaseConfiguration):
...
+        """Returns a non-secret destination identity, or "" when unavailable."""
+        return ""
+
     def fingerprint(self) -> str:
```
rudolfix, line 173: I know that it would be really good to share some code with `fingerprint()` but this one is used by telemetry and if we do that - several reports will be affected. let's talk about it

tests/load/pipeline/test_join_compatibility.py
```diff
@@ -0,0 +1,208 @@
...
+    DestinationTestConfiguration,
+    destinations_configs,
+)
+from tests.pipeline.utils import assert_load_info
+
+
+pytestmark = pytest.mark.essential
+
+
+SAME_DATABASE_JOIN_COMPATIBILITY_CONFIGS = destinations_configs(
```
rudolfix, line 23: this is good. can be reused for for external/foreign join test!

rudolfix (changes requested):
1. I think you wanted to name `physical_destination` -> `physical_location`?
2. `fingerprint` contract was changed. let's discuss that. impact is potentially pretty big
3. why sqlalchemy fingerprint test is passing? was it compatible or we do not test it at all?
4. we need another ticket to extend #3747 - to allow to ATTACH one duckdb client to another. I can take care of that
5. Test setup in `load/pipeline` is great. let's reuse it for end to end tests for #3747

rudolfix:
Automated review brought this (I didn't verify):

> Duplicate parametrize cases (tests/destinations/test_join_compatibility.py:407-420): fabric_port and fabric_default_port are byte-identical — same factory, same expected value. One
  should set an explicit port (e.g., credentials.port = 1433) to actually distinguish "default port" from "explicit port".

dlt/dataset/dataset.py
```diff
@@ -497,13 +498,14 @@ def get_dataset_sql_client(dataset: dlt.Dataset) -> SqlClientBase[Any]:
...
+)
 def is_same_physical_destination(dataset1: dlt.Dataset, dataset2: dlt.Dataset) -> bool:
-    """Check if both datasets are at the same physical destination.
-
-    This is done by comparing the fingerprint of both destination configs. There
-    are potential false positive if two different config give access to the same destination.
-    """
-    return str(dataset1.destination_client.config) == str(dataset2.destination_client.config)
+    """Check if both datasets are at the same physical destination."""
+    return dataset1.destination_client.config.can_join_with(dataset2.destination_client.config)
```
rudolfix: make sure that it is never used in our code. `Relations` is using it and we may have release before affected part is removed - then users' we'll see deprecation all the time

dlt/destinations/impl/databricks/configuration.py
```diff
@@ -237,7 +236,9 @@ def _coerce_ipc_compression(ipc_compression: Union[str, IPCCompression]) -> IPCC
 
 @configspec
 class DatabricksClientConfiguration(DestinationClientDwhWithStagingConfiguration):
-    destination_type: Final[str] = dataclasses.field(default="databricks", init=False, repr=False, compare=False)  # type: ignore[misc]
```
rudolfix, line 240: this "Final" makes sense even with mypy complaining. you removed it in several other places.

dlt/destinations/impl/bigquery/configuration.py
```diff
@@ -39,7 +40,14 @@ def get_location(self) -> str:
...
     def fingerprint(self) -> str:
-        """Returns a fingerprint of project_id"""
+        """Returns a fingerprint of the credentials project id."""
         if self.credentials and self.credentials.project_id:
             return digest128(self.credentials.project_id)
         return ""
+
+    def physical_location(self) -> str:
+        """Returns configured project id, falling back to credentials."""
+        project_id = self.project_id
```
rudolfix: I checked again and bigquery can join based on location. so here just return `self.location`.
https://docs.cloud.google.com/bigquery/docs/locations#specify_locations

so now it looks like `self.location()` is just right! and we can adopt this name everywhere like in the original ticket.

dlt/destinations/impl/databricks/configuration.py
```diff
@@ -287,7 +288,13 @@ def on_resolved(self) -> None:
             )
 
     def fingerprint(self) -> str:
-        """Returns a fingerprint of host part of a connection string"""
+        """Returns a fingerprint of the server hostname."""
         if self.credentials and self.credentials.server_hostname:
```
rudolfix: this could reuse physical_location

tests/destinations/test_join_compatibility.py
```diff
@@ -0,0 +1,1195 @@
...
+    WeaviateClientConfiguration,
+    WeaviateCredentials,
+)
+
+
+ConfigFactory: TypeAlias = Callable[[], DestinationClientConfiguration]
+ExpectedLocation: TypeAlias = Union[str, Callable[[], str]]
+
+
+class _PhysicalDestinationConfig(DestinationClientConfiguration):
```
rudolfix, line 100: duplicate code

tests/destinations/test_join_compatibility.py
```diff
@@ -0,0 +1,1195 @@
...
+
+
+@pytest.mark.parametrize("factory,expected", PHYSICAL_DEST_CASES)
+def test_physical_location(factory: ConfigFactory, expected: ExpectedLocation) -> None:
+    if callable(expected):
+        expected = expected()
+    assert factory().physical_location() == expected
+
+
+# can_join_with() matrices (symmetric)
```
rudolfix: pls remove separator comments. just keep single file or split into many files.

tests/destinations/test_join_compatibility.py
```diff
@@ -0,0 +1,1195 @@
...
+def test_cross_type_different_physical_locations() -> None:
+    sf = SnowflakeClientConfiguration(
+        credentials=SnowflakeCredentials("snowflake://u:p@a1.snowflake.com/db")
+    )
+    bq = BigQueryClientConfiguration(credentials=GcpServiceAccountCredentials(project_id="p2"))
+    assert sf.physical_location() != bq.physical_location()
+    assert_not_joinable(sf, bq)
+
+
+# Filesystem special cases
```
rudolfix: another idea to reduce size of tis file is to move specific tests to respective configuration test which every destination has.

rudolfix (changes requested):
this is really good and close to be merged. I pushed a commit here:
* it splits ability to read data from (select ... join) from ability to write data from (INSERT SELECT) - this will be needed for transformations engine. also not all engines will be SQL based in the future
* I fixed a few obscure data location issues
* filesystem - I decided to use physical location to check joinability. we'll replace when auto attach will be implemented

what remains:
* Final was removed in several places
* bigquery - so it seems it can join based on location. and we already have `location()` on the configuration. so we could just rename `physical_location` to it. please confirm
* tests are very good but if you want to split this large file I'd move it to where finerprint tests are... not needed now though!

dlt/destinations/impl/bigquery/configuration.py
Travior: You seem to be right from what I read. The question is do we want to mix the derived `location()` that will then be overridden with the `location` attribute only for bigquery? In my mind keeping `physical_location()` as the universal accessor (even though the wiring for big query is just dead boilerplate) saves a couple headaches.

rudolfix (approved):
LGTM! great work. we'll merge this after 1.28.1 is released

PR #3906, opened 2026-04-29
fix: relax JSON normalizer type validation

### LINKED ISSUE #3892: Data item normalizer type validation is overly strict
### dlt version

1.25.0

### Describe the problem

Some operations on `DltSource` only work if the normalizer is configured as the default `dlt.common.normalizers.json.relational`. For example, accessing max_table_nesting or root_key will fail if using any other normalizer, such as `dlt.common.normalizers.json.relational_no_coercion`. We observed this while trying to switch to a custom normalizer that extends the default relational normalizer by adding side effects without changing the resulting tables or rows.

### Expected behavior

We expected that the normalizer would just work as long as it is compatible.

### Steps to reproduce

The error can be reproduced like this:

```python
import os
import dlt

os.environ["SCHEMA__JSON_NORMALIZER"] = '{"module": "dlt.common.normalizers.json.relational_no_coercion"}'

@dlt.source
def data():
    yield from []

print(data().max_table_nesting)
```

This example raises `InvalidJsonNormalizer`. With the `SCHEMA__JSON_NORMALIZER` definition removed, this just prints `None`.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

Empty `DltSource` works, see example.

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

rooperuu:
### Description

Support various operations that previously only worked on the `dlt.common.normalizers.json.relational` normalizer. This includes `dlt.common.normalizers.json.relational_no_coercion` or any derived custom normalizers such as:

```python
from dlt.common.normalizers.json import relational

class DataItemNormalizer(relational.DataItemNormalizer):
    pass
```
This also fixes `max_nesting` configuration on these normalizer subtypes.

```python
import os
import dlt

os.environ["SCHEMA__JSON_NORMALIZER"] = (
    '{"module": "dlt.common.normalizers.json.relational_no_coercion", "config": {"max_nesting": 3}}'
)

@dlt.source
def nothing():
    yield from []

source = nothing()
print(source.max_table_nesting)  # 3
```

### Related Issues

- Fixes #3892

### DIFF
--- a/dlt/common/normalizers/json/relational.py
+++ b/dlt/common/normalizers/json/relational.py
@@ -332,7 +332,7 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
             RelationalNormalizerConfig,
             self.schema._normalizers_config["json"].get("config") or {},
         )
-        DataItemNormalizer._validate_normalizer_config(self.schema, config)
+        self._validate_normalizer_config(self.schema, config)
 
         # add hints, do not compile.
         self.schema._merge_hints(
--- a/dlt/common/normalizers/json/relational.py
+++ b/dlt/common/normalizers/json/relational.py
@@ -1,4 +1,5 @@
 from functools import lru_cache, partial
+from importlib import import_module
 from typing import (
     ClassVar,
     Dict,
@@ -37,6 +38,7 @@ from dlt.common.schema.utils import (
 )
 from dlt.common.utils import update_dict_nested
 from dlt.common.normalizers.json import (
+    SupportsDataItemNormalizer,
     TNormalizedRowIterator,
     wrap_in_dict,
     DataItemNormalizer as DataItemNormalizerBase,
@@ -433,7 +435,8 @@ class DataItemNormalizer(DataItemNormalizerBase[RelationalNormalizerConfig]):
     def ensure_this_normalizer(cls, norm_config: TJSONNormalizer) -> None:
         # make sure schema has right normalizer
         present_normalizer = norm_config["module"]
-        if present_normalizer != cls.__module__:
+        module = cast(SupportsDataItemNormalizer, import_module(present_normalizer))
+        if not issubclass(module.DataItemNormalizer, cls):
             raise InvalidJsonNormalizer(cls.__module__, present_normalizer)
 
     @classmethod
--- a/dlt/common/schema/normalizers.py
+++ b/dlt/common/schema/normalizers.py
@@ -9,7 +9,6 @@ from dlt.common import known_env
 from dlt.common.configuration.inject import with_config
 from dlt.common.configuration.specs import known_sections
 from dlt.common.schema.configuration import SchemaConfiguration
-from dlt.common.normalizers.exceptions import InvalidJsonNormalizer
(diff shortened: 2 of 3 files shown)


Travior:
@rooperuu thanks a lot for posting your fix. There's a couple of things I would want to do differently when incorporating this into `dlt`. I forgot to ask: Are you willing (and have time 😁) to work on this then I can give you a proper review, otherwise I can take this over

rooperuu:
@Travior Thanks for taking a look!

I think we need this for our use case, so I can spend some time on this, or whichever is easiest. Let me know what you would like to change and I can do changes accordingly.

dlt/common/normalizers/json/relational.py
Travior, line 440: In my mind we should check the "shape" of the normalizer here instead of rigidly checking the module for equality

dlt/common/schema/normalizers.py
```diff
@@ -107,20 +106,18 @@ def import_normalizers(
...
         item_normalizer["module"] = json_module_name
+    json_module = cast(SupportsDataItemNormalizer, import_module(json_module_name))
     # if max_table_nesting is set, we need to set the max_table_nesting in the json_normalizer
     if destination_capabilities and destination_capabilities.max_table_nesting is not None:
         # TODO: this is a hack, we need a better method to do this
         from dlt.common.normalizers.json.relational import DataItemNormalizer
 
-        try:
-            DataItemNormalizer.ensure_this_normalizer(item_normalizer)
+        if issubclass(json_module.DataItemNormalizer, DataItemNormalizer):
```
Travior: This should keep the `ensure_this_normalizer` check.

dlt/extract/source.py
```diff
@@ -424,12 +430,14 @@ def root_key(self) -> Optional[bool]:
 
     @root_key.setter
     def root_key(self, value: bool) -> None:
+        data_normalizer = self._schema.data_item_normalizer
+        assert isinstance(data_normalizer, RelationalNormalizer)
```
Travior: we don't need this assertion if we correctly validate in `dlt.common.normalizers.json.relational`

Travior (changes requested):
In my mind we can have a bit cleaner separation of concerns.
I think `ensure_this_normalizer` should be the place where we check normalizer compatibility (we should probably treat subclass == compatible for now). Then we need minimal to no changes in extraction or schema handling

dlt/common/normalizers/json/relational.py
rooperuu: I switched to an import and subclass check rather than direct string comparison. Now `Superclass.ensure_this_normalizer(subclass_config)` should always pass.

rooperuu:
@Travior Thanks for the review. I think what you're suggesting makes sense. I changed the implementation so that `ensure_this_normalizer` is relaxed from equality to a subclass comparison, and code paths elsewhere are kept essentially unchanged.

Edit: It seems a similar fix has already been applied in https://github.com/dlt-hub/dlt/pull/3808. Maybe this PR can be scrapped.

Travior:
Yeah, thanks for flagging this. The reproduction snippet works (at least for me) on the latest `dlt` release. I'll close this PR and corresponding issue for now. If you encounter any other issues with your specific use case feel free to open a new ticket

PR #3908, opened 2026-04-30
feat: `lance` destination REST Namespace support (experimental)

jorritsandbrink:
### Description
Adds experimental support for REST Namespaces to `lance` destination.

This is considered experimental because it's only tested against an ephemeral in-memory Lance REST Namespace server proxy (based on `RestAdapter`), not against a real production-grade Lance REST Namespace server.


tests/libs/test_ibis.py
```diff
@@ -14,39 +16,44 @@
...
+    scope="module",
+    params=destinations_configs(
+        default_sql_configs=True,
+        subset=["duckdb"],
+        file_format=None,
+    ),
+    ids=lambda x: x.name,
+)
+def destination_config(request: pytest.FixtureRequest) -> DestinationTestConfiguration:
+    return cast(DestinationTestConfiguration, request.param)
```
zilto, line 45: nice cleanup

tests/load/test_read_interfaces.py
```diff
@@ -147,13 +147,38 @@ def orderable_in_chain():
...
+
+
 @pytest.fixture(scope="module")
 def populated_pipeline(
-    request, auto_module_test_storage, preserve_module_environ, auto_module_test_run_context
+    destination_config: DestinationTestConfiguration,
+    module_lance_rest_server: None,
+    auto_module_test_storage,
+    preserve_module_environ_per_destination_config,
+    auto_module_test_run_context,
```
zilto, line 178: If I understand, before we had: `X configs * Y destinations * Z tests`.  Now, this centralizes fixtures at the module level to have `X configs * Y destinations`. 

I'm concerned that:
- some test mutate the destination
- some destination lock on concurrent read (e.g., `duckdb`)

tests/load/test_read_interfaces.py
```diff
@@ -26,18 +25,19 @@
...
-    SFTP_BUCKET,
     MEMORY_BUCKET,
+    SFTP_BUCKET,
+    destinations_configs,
+    drop_pipeline_data,
 )
 from tests.utils import (
-    preserve_module_environ,
-    auto_module_test_storage,
+    _preserve_environ,
```
zilto, line 37: what's the difference between `_preserve_environ` (which seems private) and the previous `preserve_module_environ` ?

tests/load/test_read_interfaces.py
jorritsandbrink: > some test mutate the destination

I think this is a valid concern, but not something introduced in this PR. `populated_pipeline` was already module-scoped, and each `destination_config` was coming from a single centralized `configs` list. This PR doesn't change behavior, it only makes the code DRYer: instead of parametrizing each test with the same `configs` list, we create a single module-scoped fixture.

> some destination lock on concurrent read (e.g., duckdb)

If I understand correctly, each worker creates it's own version of the module-scoped fixture, and tests are executed in sequence on a worker, so concurrent reads against the same duckdb instance do not happen.

tests/load/test_read_interfaces.py
jorritsandbrink: `preserve_module_environ` is a fixture, and `_preserve_environ` is a helper to create fixtures. I'm using `_preserve_environ` in the new `preserve_module_environ_per_destination_config`, which is why I import it in `test_read_interfaces.py`. Perhaps we make `_preserve_environ` public?

PR #3910, opened 2026-05-04
Fix/3592 bypass non interactive flag prompts

### LINKED ISSUE #3592: The --non-interactive flag appears to default to “No” for prompt confirmations

anuunchin:
Continuation of #3775 due to the `ci from fork` label currently not working. 

In a nutshell, adds a new -y/--yes flag that auto-accepts all `confirm()` prompts. This is implemented via a dedicated ALWAYS_CONFIRM flag in `echo.py` that only affects `confirm()`, so it does not leak into `prompt()` or `text_input()`.

Resolves #3592


rudolfix (dismissed):
LGTM! I slightly cleaned code in this PR

Travior (approved):

PR #3925, opened 2026-05-08
test: add SPN Fabric token refresh automation

jorritsandbrink:
### Description
The SPN owning the Fabric warehouse needs to refresh its Fabric token at least every thirty days: https://learn.microsoft.com/en-us/fabric/data-warehouse/service-principals#token-renewal-and-initialization-requirements. This PR automates that process, so that our CI tests for the `fabric` destination are always able to run.


rudolfix (approved):
LGTM!

PR #3927, opened 2026-05-08
feat: clickhouse staging-optimized

### LINKED ISSUE #3926: staging-optimized strategy for Clickhouse
### Feature description

I'd like to have the `staging-optimized` replace strategy for the Clickhouse destination.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

We've been using DLT in production to move data from Google Drive to Clickhouse. Sometimes we get empty tables for a bit of time. I think it happens due to how `insert-from-staging` truncates the destination table, then inserts from the staging table. This leaves a period of time without data, because Clickhouse doesn't have transactions.

### Proposed solution

I think `staging-optimized` should eliminate this downtime. The Postgres destination does this with a transaction. Clickhouse doesn't have transactions, but has the [`EXCHANGE TABLES`](https://clickhouse.com/docs/en/sql-reference/statements/exchange) statement for atomic swaps.

### Related issues

_No response_

filipesilva:
### Description

Adds the `staging-optimized` replace strategy to the Clickhouse destination, using the [EXCHANGE TABLES](https://clickhouse.com/docs/en/sql-reference/statements/exchange) statement for atomic swaps.

### Related Issues

- Fixes https://github.com/dlt-hub/dlt/issues/3926

### Additional Context

Python is not my primary language, and I have used LLM agent assistance to produce this PR. 

I have tested it locally with a local clickhouse, but wasn't able to test the `*-staging-s3-*` and `*-staging-az-*` tests because those seem to need CI credentials.


filipesilva:
Heya @zilto is there anything I can do to help move this forward?

dlt/destinations/impl/clickhouse/clickhouse.py
```diff
@@ -296,6 +298,30 @@ def requires_temp_table_for_delete(cls) -> bool:
...
+        table_chain: Sequence[PreparedTableSchema],
+        sql_client: SqlClientBase[Any],
+    ) -> List[str]:
+        sql: List[str] = []
+        for table in table_chain:
+            with sql_client.with_staging_dataset():
+                staging_table_name = sql_client.make_qualified_table_name(table["name"])
+            table_name = sql_client.make_qualified_table_name(table["name"])
+            sql.append(f"EXCHANGE TABLES {staging_table_name} AND {table_name}")
+            sql.append(f"TRUNCATE TABLE {staging_table_name}")
```
rudolfix: do not truncate tables here. this makes that job non idempotent. if truncation fails the job will be retried and you'll exchange again and truncate the table with data. - `dlt` truncates staging dataset before the load. also user can do that with the sql client

tests/load/pipeline/test_replace_disposition.py
```diff
@@ -525,3 +530,104 @@ def load_items_empty():
...
+    ),
+    ids=lambda x: x.name,
+)
+def test_clickhouse_atomic_swap_replace(
+    destination_config: DestinationTestConfiguration,
+    mocker: MockerFixture,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Test ClickHouse atomic swap via EXCHANGE TABLES with sequential loads, nested tables, and empty resource."""
+    from dlt.destinations.sql_jobs import SqlStagingFollowupJob, SqlStagingReplaceFollowupJob
```
rudolfix: import at the top

tests/load/pipeline/test_replace_disposition.py
```diff
@@ -525,3 +530,104 @@ def load_items_empty():
...
+    assert swap_spy.call_count == 1
+    assert load_table_counts(pipeline, "items", "items__sub_items") == {
+        "items": 3,
+        "items__sub_items": 6,
+    }
+    table_dicts = load_tables_to_dicts(pipeline, "items")
+    assert {int(r["id"]) for r in table_dicts["items"]} == {100, 101, 102}
+
+    # third load: schema evolution adds a new column, EXCHANGE must work after ALTER
+    @dlt.resource(name="items", write_disposition="replace", primary_key="id")
```
rudolfix: is this the core thing that this test is checking? otherwise - we have test that will check optimized replace for all destinations that enable it.

if so this is clickhouse specific test (we check if EXCHANGE works after ALTER so this tests clickhouse engine not dlt) - you can still keep it but plese move to load/pipeline/test_clickhouse.py

rudolfix (changes requested):
@filipesilva this is pretty cool, I didn't notice you can swap tables in clickhouse. I think we should not truncate table in replace job. see my review. thanks!

tests/load/pipeline/test_replace_disposition.py
filipesilva: Yes it's not testing anything clickhouse specific. Will remove. Thanks for the pointer!

dlt/destinations/impl/clickhouse/clickhouse.py
filipesilva: Roger, sounds good to me.

filipesilva:
@rudolfix thanks for taking the time to review, the comments should be addressed now.

rudolfix (approved):
LGTM! apologies for a long wait

filipesilva:
No worries, I hear you folks are quite busy with your new offering lately :D
Please let me know if there's trouble with these changes!

PR #3928, opened 2026-05-09
fix: coerce double to decimal via str() to avoid IEEE 754 expansion

### LINKED ISSUE #3917: Coercion from double to decimal can produce unexpected values
### dlt version

1.26.0

### Describe the problem

The coercion logic converts Python `float` to `decimal.Decimal` directly using the constructor. This can convert seemingly acceptable values to ones that do not adhere to the precision and scale constraints:

```python
from decimal import Decimal
Decimal(34.7)  # Decimal('34.7000000000000028421709430404007434844970703125')
```
In some destinations, such as DuckDB, this does not cause any issue as the values are rounded during loading. It seems that DuckDB loads only fail if a value has too large magnitude to fit in the given precision, so in the above case, the number would be rounded back to the original.

In BigQuery, however, the loads are strict, and the above case is rejected causing an error. The only way to prevent the error is rounding the values before loading.

### Expected behavior

_No response_

### Steps to reproduce

The error can be reproduced by trying to load the resource defined below to any destination that does not automatically round the values during a load, such as BigQuery.

### Operating system

macOS, Linux

### Runtime environment

Local

### Python version

3.12

### dlt data source

```python
import dlt

@dlt.resource(columns={"some_number": {"data_type": "decimal"}})
def example():
    yield {"some_number": 34.7}
```

### dlt destination

Google BigQuery

### Other deployment details

_No response_

### Additional information

This could be fixed in multiple ways.

One is to do the conversion via `Decimal(str(float_value))`, which fixes the example with `34.7` by giving `Decimal('34.7')`. This works because `str(float_value)` gives the shortest decimal representation that corresponds to the binary representation of the float value rather than the closest decimal value. This would still never round the values, so inputs like `0.1234567899` would still cause an error rather than using `Decimal('0.123456790')`. In some sense an error for invalid values is good, but these currently depend on the destination.

More advanced solutions could utilize the precision and scale hints. These could be enforced after the basic coercion by rounding to at most `scale` decimal places. Values that do not fit in the `precision` at all should maybe continue to fail.

AnnasMazhar:
Closes dlt-hub/dlt#3917

When coercing a double (float) to Decimal, passing the float directly
to Decimal() triggers the full IEEE 754 double-precision representation,
producing values like Decimal('34.70000000000000284...') for float 34.7.

Fix: coerce via Decimal(str(value)) instead, which preserves the
human-readable representation. Same fix applied to double→Wei coercion.

Added test: test_coerce_double_to_decimal_no_float_expansion


rudolfix (changes requested):
any situation when binary numbers are converted to decimal is a little bit hopeless. your fix assumes that binary floats are used to store decimals (`str` algo for float finds first stable decimals, dumps the "unstable" binary part).
this is probably a good default that follows user intent (forcing decimal over float)

my take would be to document this behavior somewhere in `schema.md` docs page. @AnnasMazhar  pls lmk. if you are able to do it. users should know that such conversion is not precise and should be avoided if those decimals represent currency.

dlt/common/data_types/type_helpers.py
```diff
@@ -206,12 +206,12 @@ def _text_to_date_fast(value: str) -> PendulumDate:
     # to decimal
     ("decimal", "text"): _text_to_decimal,
     ("decimal", "bigint"): Decimal,
-    ("decimal", "double"): Decimal,
+    ("decimal", "double"): lambda v: Decimal(str(v)),
```
rudolfix: makes sense to create `_double_to_decimal` to follow pattern above

tests/common/schema/test_coercion.py
rudolfix, line 138: it also makes sense to drop quantization here. this workaround is no longer needed

tests/common/schema/test_coercion.py
```diff
@@ -153,6 +153,20 @@ def test_coerce_to_numeric(dec_cls: Type[Any], data_type: TDataType) -> None:
...
+    # 34.7 as float would expand to 34.70000000000000284... with Decimal(34.7)
+    # Using Decimal(str(34.7)) preserves "34.7"
+    result = coerce_value("decimal", "double", 34.7)
+    assert result == Decimal("34.7")
+    assert str(result) == "34.7"
+
+    # Same issue applies to Wei which extends Decimal
+    result = coerce_value("wei", "double", 34.7)
+    assert result == Wei("34.7")
+    assert str(result) == "34.7"
```
rudolfix, line 167: it also makes sense to show that `Decimal(34.70000000000000284)` is now `34.7` and demonstrate how to get the exact binary value via quantization.

AnnasMazhar:
@rudolfix Thanks for the review and the suggestion that's a great point. Users should definitely be aware that this conversion follows intent but isn't mathematically precise. I've added a note to docs/website/docs/general-usage/schema.md in the coercion section explaining the str() behavior and warning against using it for currency. Let me know if you'd like any changes to the wording.

rudolfix (approved):
LGTM!

PR #3947, opened 2026-05-18
fix(mssql): ingest parquet row-groups individually to bound ADBC driver   memory

### LINKED ISSUE #3915: loading parquet files into mssql via adbc results in OOM
### dlt version

1.26.0

### Describe the problem

We have a simple pipeline from sql server -> azure sql. We are using pyArrow to generate parquet files from the source and using the parquet file with ADBC to load to Azure. This process is running a container on azure container apps w/ 8GB of memory.

The above process works well for most tables, but larger tables result in the OOM Killer terminating the container with a 137 exit code. We used max file size to create multiple, smaller parquet files but that didn't work.

Upon further investigation of the DLT source code. It seems that a combination of iterating over the row groups and the way the mssql ADBC driver works will attempt to load the entire data set into memory..

### Expected behavior

I would expect the load to succeed with a limited amount of memory, regardless of the size of a table.

### Steps to reproduce

use the mssql destination with ADBC and parquet to load a file that is larger than available system memory

### Operating system

Linux

### Runtime environment

Kubernetes

### Python version

3.12

### dlt data source

Sql Server via `sql_database` with pyArrow backend

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

We worked around this issue by creating a monkey-patch that loops over the rowgroups and passes each one into `adbc_ingest` (attached).

[injest_patch.py](https://github.com/user-attachments/files/27441156/injest_patch.py)

wtfashwin:
Closes #3915.

  ## What

  `AdbcParquetCopyJob.run()` passes a single generator over every row-group to
  `cur.adbc_ingest(...)`. The mssql ADBC driver buffers the full input stream
  before flushing, so a parquet file larger than available memory triggers the
  OOM killer (reporter saw exit 137 on an 8 GB container). Postgres ADBC streams
   incrementally and doesn't hit this.

  ## Why this approach

  Two minimal options:

  - override `run()` entirely in `MssqlParquetCopyJob` (the reporter's
  monkey-patch shape)
  - gate the per-row-group path on the shared base class via a class attribute
  
  The second keeps the ingest/commit/logging flow in one place and makes the
  design choice explicit on the subclass that needs it. Postgres / sqlalchemy
  stay on the default streaming form, which preserves their existing
  transactional behaviour (one commit per file, not per row-group).
  
  Iterating `pq_stream_with_new_columns` per row-group and calling `adbc_ingest`
   once per resulting `pyarrow.Table` is what bounds the driver memory —
  `commit()` after each call lets the driver release its internal buffer.
  
  ## How
  
  - `AdbcParquetCopyJob` gains `_ingest_per_rowgroup: bool = False` and a branch
   in `run()` that either streams all batches in one call (default) or iterates
  row-groups (when set).
  - `MssqlParquetCopyJob` sets `_ingest_per_rowgroup = True`.
  - Hoisted `import time` to module level (was inline inside `run`) to match the
   project's import rule.
  - `without_none` is computed once outside the loop instead of per-call.
  
  ## Scope
  
  - Not touching `pq_stream_with_new_columns(row_groups_per_read=1)` default — 
  that helper has callers outside the mssql path and the row-group cardinality 
  choice is the writer's, not ours.
  - No new user-facing config knob. The mssql ADBC driver's buffering is a 
  driver fact; the right place to encode it is the destination class, not user 
  config.
  - Not touching the postgres / sqlalchemy code paths.
  
  ## Tests
  
  `tests/destinations/test_adbc_jobs.py` (new, 152 LOC, mock-based — no real 
  ADBC driver needed):
  
  - `test_default_ingest_streams_all_batches_in_single_call` — base class, one 
  ingest + one commit
  - `test_per_rowgroup_ingest_calls_adbc_ingest_per_rowgroup` — 3-row-group 
  parquet → 3 ingests + 3 commits, asserts `pyarrow.Table` arg type, `append` 
  mode, catalog/schema kwargs
  - `test_per_rowgroup_ingest_handles_single_rowgroup` — single-row-group edge
  - `test_per_rowgroup_ingest_drops_none_catalog_and_schema` — `without_none` 
  strips both kwargs when the destination disables them
  - `test_mssql_copy_job_opts_into_per_rowgroup_ingest` — guarded with 
  `importorskip("pyodbc")`
  - `test_postgres_copy_job_keeps_streaming_ingest` — guarded with 
  `importorskip("psycopg2")` — regression guard
  
  4/6 pass locally (others skip without driver libs); CI will hit the remaining 
  two.
  
  ## Backwards compat
  
  No public API change. `_ingest_per_rowgroup` is an internal class attribute. 
  Postgres and sqlalchemy ADBC paths are byte-for-byte unchanged.
  PRBODY


prowe (approved):
Thank you for the quick fix to this one

wtfashwin:
> Thank you for the quick fix to this one

@prowe thanks for the review and approval! The outstanding checks are fork-gated jobs showing as skipping pending reviewer approval (mergeStateStatus is BLOCKED). Could you approve the workflow run so the secret-dependent CI can execute? Happy to rebase if needed.

prowe:
> > Thank you for the quick fix to this one
> 
> @prowe thanks for the review and approval! The outstanding checks are fork-gated jobs showing as skipping pending reviewer approval (mergeStateStatus is BLOCKED). Could you approve the workflow run so the secret-dependent CI can execute? Happy to rebase if needed.

I'm not a maintainer on the project (I'm just a reporter and interested party) so I don't have those permissions

rudolfix:
@wtfashwin I reviewed this PR. this is really nice upgrade! I have a question that will impact full review: is it really necessary to commit each row group separately? if so: we trade atomicity for being able to stream data and this is a significant change. in that case I'd look for making this setting configurable and most probably set it to `False` by default.

do you still have your test setup? if so please lmk. if we can move the commit out of the loop.
I also pushed the newest devel to unblock ci. thanks!

rudolfix:
@wtfashwin a quick research with opus tells that underlying go driver will flush buffers after every ingest so commit can indeed be moved out of the loop

wtfashwin:
@rudolfix thanks for the careful review and for pushing devel to unblock CI.

You're right , committing per row-group isn't necessary. The driver flushes its buffer on every `adbc_ingest` call, so the per-row-group ingest alone is what bounds the memory; the commit doesn't need to be inside the loop. I've moved `conn.commit()` out so we keep one transaction per file and don't trade away atomicity. No new config knob needed then -- both paths commit once per file, the mssql path just splits the ingest calls.

Updated the test to assert 3 ingests + 1 commit, and refreshed the docstring to explain the buffer-flush reasoning. I don't have the original 8 GB repro container anymore, but the memory bound holds the same way since each `adbc_ingest` is still capped at a single row-group. Pushed.

dlt/destinations/_adbc_jobs.py
```diff
@@ -47,6 +48,18 @@ def has_adbc_driver(driver: str, disable_adbc_detection: bool = False) -> Tuple[
 
 
 class AdbcParquetCopyJob(RunnableLoadJob, HasFollowupJobs, ABC):
+    _ingest_per_rowgroup: bool = False
+    """When True, call `adbc_ingest` once per parquet row-group instead of once per file.
+
+    Some ADBC drivers (notably mssql) buffer the entire input stream in memory before
```
rudolfix: this is opus talking to itself... please remove that or move as comments to relevant parts of the code (ie. mssql.py)

dlt/destinations/impl/mssql/mssql.py
```diff
@@ -92,6 +92,9 @@ def _new_temp_table_name(cls, table_name: str, op: str, sql_client: SqlClientBas
 
 class MssqlParquetCopyJob(AdbcParquetCopyJob):
     _config: MsSqlClientConfiguration
+    # mssql ADBC driver buffers the full input stream in memory; flush per row-group
+    # to bound peak memory and avoid OOM on large parquet files. refs #3915
```
rudolfix: pls. remove `refs`

tests/destinations/test_adbc_jobs.py
```diff
@@ -0,0 +1,153 @@
...
+                {"id": list(range(g * rows_per_group, (g + 1) * rows_per_group))},
+                schema=schema,
+            )
+            writer.write_table(tbl, row_group_size=rows_per_group)
+
+
+class _StubAdbcParquetCopyJob(AdbcParquetCopyJob):
+    """Minimal concrete subclass to exercise the base `run()` contract.
+
+    Skips the heavy `RunnableLoadJob.__init__` (which parses filenames and
```
rudolfix: pls remove or move as comment to where it is relevant

tests/destinations/test_adbc_jobs.py
```diff
@@ -0,0 +1,153 @@
...
+        return self._conn_cm
+
+    def _set_catalog_and_schema(self) -> Tuple[str, str]:
+        return "cat", "sch"
+
+
+def _make_mock_conn(rows_per_ingest: int) -> Tuple[MagicMock, MagicMock, MagicMock]:
+    """Build a context-manager-shaped mock connection.
+
+    Returns the outer `with`-target, the inner connection mock (for asserting
```
rudolfix: this should be a comment

tests/destinations/test_adbc_jobs.py
```diff
@@ -0,0 +1,153 @@
...
+
+    job.run()
+
+    # `without_none` must strip both kwargs so the driver doesn't receive Nones
+    for call in cur.adbc_ingest.call_args_list:
+        assert "catalog_name" not in call.kwargs
+        assert "db_schema_name" not in call.kwargs
+
+
+def test_mssql_copy_job_opts_into_per_rowgroup_ingest() -> None:
```
rudolfix: please move to ie. `test_mssql_extras.py` then you do not even need to skip the test

tests/destinations/test_adbc_jobs.py
```diff
@@ -0,0 +1,153 @@
...
+    """`MssqlParquetCopyJob` must set the per-rowgroup flag so the bug fix sticks."""
+    pytest.importorskip("pyodbc")
+    from dlt.destinations.impl.mssql.mssql import MssqlParquetCopyJob
+
+    assert MssqlParquetCopyJob._ingest_per_rowgroup is True
+
+
+def test_postgres_copy_job_keeps_streaming_ingest() -> None:
```
rudolfix: pls move ie to `test_postgres_client.py`

tests/destinations/test_adbc_jobs.py
```diff
@@ -0,0 +1,153 @@
...
+    job = _StubAdbcParquetCopyJob(file_path, conn_cm)
+
+    job.run()
+
+    # default behaviour: one `adbc_ingest` call covering every batch
+    assert cur.adbc_ingest.call_count == 1
+    assert conn.commit.call_count == 1
+
+
+def test_per_rowgroup_ingest_calls_adbc_ingest_per_rowgroup(tmp_path) -> None:
```
rudolfix: this test and `test_per_rowgroup_ingest_handles_single_rowgroup` can be parametrized into one. we are also missing a test with empty parquet (0 groups)

tests/destinations/test_adbc_jobs.py
```diff
@@ -0,0 +1,153 @@
...
+    job._ingest_per_rowgroup = True
+
+    job.run()
+
+    # single row-group still goes through the per-rowgroup path: one ingest, one commit
+    assert cur.adbc_ingest.call_count == 1
+    assert conn.commit.call_count == 1
+
+
+def test_per_rowgroup_ingest_drops_none_catalog_and_schema(tmp_path) -> None:
```
rudolfix: particular implementations are testing this so you can remove

rudolfix (changes requested):
now code looks good! the remaining problems are
* there are LLM artifacts that must be cleaned up
* tests sprawl - test can be parametrized, removed and moved around

recent LLMs overdo both things and we are trying to keep it under control. thx!

rudolfix (approved):
LGTM!

PR #3955, opened 2026-05-20
fix(filesystem): default bucket_url to '.' for bare destination string

### LINKED ISSUE #3886: feat(filesystem): allow bare `dlt.pipeline(..., destination="filesystem")`
## Problem
Currently, the following will throw an error complaining that `bucket_url` is not set in `config.toml` 

```python
dlt.pipeline(
    ...,
    destination="filesystem",
)
```

This is the required valid code

```python
dlt.pipeline(
    ...,
    # first parameter is `bucket_url`
    destination=dlt.destinations.filesystem("."),
)
```

## Desired solution
Passing `destination="filesystem"` should default to `dlt.destinations.filesystem(".")` when no config is found. 

This would match the behavior of the `duckdb` destination which can be called via `dlt.pipeline(..., destination="duckdb")` without any config

Bhagwat-45:
### Description

Allows bare `dlt.pipeline(..., destination="filesystem")` without requiring `bucket_url` to be set in `config.toml`, matching the existing behavior of the `duckdb` destination.

Previously this would throw a `ConfigFieldMissingException`. The fix sets `bucket_url = "."` (current directory) as the default in `FilesystemConfiguration`. The existing `normalize_bucket_url()` method already handles `"."` correctly, converting it to an absolute `file://` URL at resolve time — so this is a minimal, safe one-line change.

**Files changed:**
- `dlt/common/storages/configuration.py` — set `bucket_url: str = "."` as the default
- `tests/sources/filesystem/test_config_sections.py` — rewrote test to reflect new default; added test verifying config section path for cloud destinations (which still require credentials)
- `tests/load/filesystem/test_filesystem_factory.py` — updated stale comment
- `tests/workspace/cli/test_config_toml_writer.py` — updated assertion and comment

### Related Issues
- Fixes #3886

### Additional Context

Tests were run locally and passed:
```
pytest tests/common/storages/test_local_filesystem.py \
       tests/sources/filesystem/test_config_sections.py \
       tests/load/filesystem/test_filesystem_factory.py \
       tests/workspace/cli/test_config_toml_writer.py -v
```
Result: 55 passed, 21 skipped, 0 failed.

### DIFF
--- a/dlt/common/storages/configuration.py
+++ b/dlt/common/storages/configuration.py
@@ -201,7 +201,7 @@ class FilesystemConfiguration(BaseConfiguration):
         "sftp": SFTPCredentials,
     }
 
-    bucket_url: str = None
+    bucket_url: str = "."
 
     # should be a union of all possible credentials as found in PROTOCOL_CREDENTIALS
     credentials: FileSystemCredentials = None
--- a/tests/load/filesystem/test_filesystem_factory.py
+++ b/tests/load/filesystem/test_filesystem_factory.py
@@ -81,7 +81,7 @@ def test_resolve_bucket_url_ignores_foreign_credentials():
         fs = filesystem()
         assert fs.is_hf
 
-    # without bucket_url, _resolve_bucket_url returns None and is_hf is False
+    # without explicit bucket_url, defaults to "." (local filesystem), so is_hf is False
     with custom_environ({"CREDENTIALS": "postgres://loader:password@localhost:5432/dlt_data"}):
         fs = filesystem()
         assert not fs.is_hf
--- a/tests/sources/filesystem/test_config_sections.py
+++ b/tests/sources/filesystem/test_config_sections.py
@@ -1,17 +1,37 @@
 import pytest
+import os
 
 from dlt.sources.filesystem import filesystem, read_parquet
 from dlt.common.configuration.exceptions import ConfigFieldMissingException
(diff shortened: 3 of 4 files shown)


rudolfix:
@Bhagwat-45 by setting a default you are making `bucket_url` optional. this will impact a lot of users on filesystems different than local. same configuration is shared with ducklake, lance and iceberg catalogs. tbh. it will be very confusing if instead of failing with missing configuration, files are created in locally. very hard to diagnose and also leaks data into place where it should not be. my take is that you add the following to your global config toml (so all your projects use it)
```toml
[credentials]
bucket_url="file:///."
```
(I didnt check it, pwd may be slightly different)
this will act as an default that any more specific configuration can overwrite

zilto:
@rudolfix the initial issue was for quality of life and supporting, similar to `destination="duckdb"` 

```python
dlt.pipeline(..., destination="filesystem")
```

Setting up `config.toml` or `dlt.destinations.filesystem(bucket_url=...)` can trip up new users and agents

## Potential solutions
- IMHO, it makes sense to have `filesystem` default to local. It doesn't make sense to expect your pipeline will write to S3 if you didn't set any credentials
- The tricky part is how `filesystem` credentials are used by other destinations. In these cases, can the "parent config" set default value to `None` / missing? For example, `ducklake` could remove `bucket_url="."`

If this is too complicated, we should avoid changes.

PR #3970, opened 2026-05-22
docs CI: smarter verify-redirects and dedicated CI step

### LINKED ISSUE #3968: CI improvement for redirects in docs: verify-redirects fails on devel-only targets during restructures
### Feature description

# Problem
scripts/verify-redirects.js runs at the end of npm run build and checks every redirect target against the master snapshot (frozen at the last dlt release). Qhen we restructure docs on devel and add legacy redirects pointing at the new paths, these paths don't exist in master yet and the check fails and blocks CI, even though the redirects are correct and users see no broken pages (false positive).

Relevance: I have this problem with [#3960](https://github.com/dlt-hub/dlt/pull/3960) (hub folder restructure): 27 redirects flagged red, all false positives; and recently we missed a real redirect issue and merged. 

# Proposed solution
Adjust the CI script with a devel-fallback for redirects: if a target is missing in master, check build/docs/devel/<path> too
*  if devel has it → WARN ("devel-only, will resolve at next release")
* if neither has it → ERROR (today's behavior). exit code stays gated on errors only.

Additional QoL improvement: Split verify-redirects into its own CI step in .github/workflows/build_docs.yml
* the standalone npm run verify-redirects script already exists
* move it out of npm run build so it has its own pass/fail visibility in CI.

**Desired outcomes:**. 
* CI keeps blocking on real broken redirects
* CI stops blocking on the false positives we work around manually today
* self-healing: once master refreshes after the next dlt release, warnings auto-clear.

files to be touched:
docs/website/scripts/verify-redirects.js
docs/website/package.json
.github/workflows/build_docs.yml

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

_No response_

### Proposed solution

_No response_

### Related issues

_No response_

lis365b:
### Description
Improve CI for redirects by reducing false positives in CI when restructuring docs

### Related Issues
- Resolves #3968  https://github.com/dlt-hub/dlt/issues/3968

### Additional Context
- adds a "Verify redirects" step inside the existing build docs job (in .github/workflows/build_docs.yml). no renames, only a new step in an existing job.

### Open questions: 
 1. do devel / master have required status checks configured?
2.  if yes, and if the list has individual step-level checks (not just the umbrella build docs job), do i need to add "Verify redirects" once this PR merges?


docs/website/README.md
```diff
@@ -64,7 +64,34 @@ For most authoring purposes, once you are happy with your changes running locall
...
 ```
 
-That command generates static content into the `build` directory, which can be served using any static contents hosting service, for example, `npm run serve`
+That command generates static content into the `build` directory, which can be served using `npm run serve`.
+
+### What `npm run build` does
+
+The full build runs these steps in order:
+
+1. **`npm run update-versions`** — clones `dlt`, checks out the latest release tag, freezes the content into `versioned_docs/version-master/`. This is the **master snapshot** (served at `/docs/`); your branch is served at `/docs/devel/`. See [Docs versions](#docs-versions) below.
```
burnash:
```suggestion
1. **`npm run update-versions`** — clones `dlt`, checks out the `master` branch, freezes the content into `versioned_docs/version-master/`. This is the **master snapshot** (served at `/docs/`); your branch is served at `/docs/devel/`. See [Docs versions](#docs-versions) below.
```

make it agree with what update_versions.js does

burnash (approved):
Looks good! Thank you @lis365b

PR #3980, opened 2026-05-26
fix(3979): allow for legacy ducklake migration

TomBurdge:
- **feat(ducklake): add AUTOMATIC_MIGRATION flag to ducklake**
  # Context
  When attaching a v0.4 ducklake catalog with a v1.x version of the duckb
  ducklake extension with dlt, there is the following error:
  `_duckdb.InvalidInputException: Invalid Input Error: DuckLake catalog
  version mismatch: catalog version is 0.4, but the extension requires
  version 1.0. To automatically migrate, set AUTOMATIC_MIGRATION to TRUE
  when attaching.`
  
  Fortunately [the duckdb docs are quite useful on
  this](https://ducklake.select/docs/stable/duckdb/guides/troubleshooting#connecting-to-an-older-ducklake).
  you just need to attach with the AUTOMATIC_MIGRATION flag as true:
  `ATTACH '...' (AUTOMATIC_MIGRATION);`
  
  # The Change
  Add a (by default off) `AUTOMATIC_MIGRATION` flag to the ducklake
  destination so that users can upgrade legacy catalogs in their pipelines
  with DLT. Off by default means no breaking changes.
  
  # The Benefit
  Users can run their pipelines again even when they upgrade the duckdb
  version/ducklake extension and use a legacy catalog.
  

- **test(ducklake): update test for AUTOMATIC_MIGRATION flag**
  

- **docs(ducklake): add docs to cover AUTOMATIC_MIGRATION flag for ducklake**

### DIFF
--- a/dlt/destinations/impl/ducklake/configuration.py
+++ b/dlt/destinations/impl/ducklake/configuration.py
@@ -140,6 +140,8 @@ class DuckLakeClientConfiguration(WithLocalFiles, DestinationClientDwhWithStagin
     credentials: DuckLakeCredentials = None
     create_indexes: bool = False  # does nothing but required
     override_data_path: bool = False
+    automatic_migration: bool = False
+    """When true, attaches with `AUTOMATIC_MIGRATION true` so DuckDB migrates an older DuckLake catalog schema on attach."""
 
     def fingerprint(self) -> str:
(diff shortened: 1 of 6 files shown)


rudolfix (approved):
LGTM!

PR #3988, opened 2026-05-28
Added escape characters to correct deployment file name

nuetu:
CLI reference page on docs: 
double underscore results in bold on markdown - replaced each  __deployment__.py & __deployment__
with
\_\_deployment__.py & \_\_deployment__

### Related Issues

- Fixes #...
- Closes #...
- Resolves #...


burnash (approved):
Thanks for the fix @nuetu

PR #3996, opened 2026-05-29
Feat/even more faster testing

tetelio:
## Summary

Two categories of changes:
1. **Cosmetic only** — renames `name:` fields for jobs/workflows (improves GitHub UI readability, no functional effect)
2. **Structural refactors** — splits monolithic jobs into explicit matrix entries with a `suite` or `display_name` dimension; changes runners from generic to `blacksmith-*` variants

### Structural refactors detail

**`test_common.yml` — biggest change:** the previous 10 fat sequential jobs (each running all suites back-to-back) are replaced by 27 parallel focused jobs. Each OS/Python combination now spawns 3 independent jobs:

| Suite | What runs | Before | After |
|---|---|---|---|
| `common/pipeline-mindeps-workspace` | `test-common-core`, `test-tools`, `test-common-source`, `test-pipeline-min`, `test-pipeline-arrow`, `test-workspace` | combined into every job | own job |
| `pipeline-full` | `test-pipeline-full` | combined into every job | own job |
| `sqlalchemy-2` | `test-with-sqlalchemy-2` | combined into every job | own job |

Same 10 OS/Python combinations, same total test work — but now the three suites run in parallel instead of sequentially. Wall-clock time is roughly `max(suite)` instead of `sum(suites)`.

**Other structural refactors:**

| Workflow | Before | After |
|---|---|---|
| `test_tools_dbt_runner.yml` | 1 job: no-venv phase then venv phase sequentially | 2 parallel jobs, one per phase |
| `test_sources_local.yml` | 3 jobs with implicit `allows_sqlalchemy_1_4` flag controlling which steps run | 5 explicit jobs with `suite` dimension (`sources-load` vs `sql-database`) |
| `test_hub.yml` | cross-product matrix + excludes | equivalent explicit `include` list |

### Same test coverage, only one functional reduction ⚠️

**`lance_s3` in `test_destinations_remote.yml`:** `always_run_all_tests: true` was removed. Non-essential tests for `lance_s3` now only run when the full test suite is triggered (`ci full` label, scheduled run), not on every PR. Essential tests still always run.


rudolfix (approved):
LGTM!
I added Python 3.14 runs to it - that required bumping a few deps. all works now and tests run fast

Issue #3998, opened 2026-05-29
Incremental load on BigQuery truncates destination table in case of no changed source data


willi-mueller:
### dlt version

1.27

### Describe the problem

Loading an incremental resource with `write_disposition="replace"` ⇒ full data extracted and loaded.

Incremental load with no changes in the source data and `write_disposition=None` ⇒ incremental cursor bound to last value ⇒ no data extracted ⇒ data in BigQuery staging table is preserved, but BigQuery destination table is truncated ⇒ destination data lost. Every time.

Data loss did not happen up to dlt version 1.26. It starts from version 1.27.

### Expected behavior

dlt only truncates the destination when I refresh and not on incremental load with `write_disposition="merge"` set on the resource level. 

### Steps to reproduce

Unfortunately, the following code does not display the data deletion. Only when I retrieve records from the actual Salesforce data source, the problem occurs. 

```py
import dlt
from typing import Iterable
from dlt.sources import DltResource
from dlt.sources import incremental
from dlt.common.typing import TDataItem


@dlt.source(
    name="salesforce",
    schema_contract={"tables": "evolve", "columns": "freeze", "data_type": "freeze"},
)
def salesforce_source() -> Iterable[DltResource]:
    @dlt.resource(write_disposition="merge", primary_key="Id")
    def record_type_test(
        last_timestamp: incremental[str] = dlt.sources.incremental(
            "SystemModstamp", initial_value=None
        ),
    ) -> Iterable[TDataItem]:
        yield _get_records()

    return (record_type_test,)


def _get_records() -> Iterable[TDataItem]:
    """This mimics the implementation of get_records() from the verified Salesforce dlt source."""
    yield from [
        {
            "Id": "0120Y000000lCpiQAE",
            "Name": "FAQ",
            "SystemModstamp": "2026-05-19T08:23:22.000000Z",
        },
        {
            "Id": "0122X000000AsLOQA0",
            "Name": "Option",
            "SystemModstamp": "2020-03-27T13:08:38.000000Z",
        },
    ]


pipeline = dlt.pipeline(
    pipeline_name="salesforce",
    destination="bigquery",
    dataset_name="dev_experiment",
    staging="filesystem",
    # refresh="drop_resources",
)
load_info = pipeline.run(
    salesforce_source().with_resources("record_type_test"),
    # write_disposition="replace",
)
print(load_info)
```

Logs from running it with actual Salesforce data:

## First run with `write_disposition="replace"`

<details>
[...]

PR #4000, opened 2026-05-29
fix: incremental merge truncates destination on no-data runs

burnash:
Port of the 1.27.2 hotfix (#3998) to devel + regression test


rudolfix (approved):
LGTM!

Issue #3141, opened 2025-09-29
Pipeline state tables are not created when using Oracle via SQLAlchemy as a destination


Agfarid:
### dlt version

1.17.0

### Describe the problem

When loading data into an Oracle database, on the first run (when the dlt internal tables do not exist yet), dlt tries to retrieve the pipeline state by executing a SELECT on tables such as _dlt_pipeline_state and _dlt_loads before attempting to create them.. This results in the following error:

```
<class 'dlt.destinations.exceptions.DatabaseTransientException'>
(oracledb.exceptions.DatabaseError) ORA-00942: table or view does not exist
Help: https://docs.oracle.com/error-help/db/ora-00942/
[SQL: SELECT [SCHEMA_NAME]."_dlt_pipeline_state".version, [SCHEMA_NAME]."_dlt_pipeline_state".engine_version, [SCHEMA_NAME]."_dlt_pipeline_state".pipeline_name, [SCHEMA_NAME]."_dlt_pipeline_state".state, [SCHEMA_NAME]."_dlt_pipeline_state".created_at, [SCHEMA_NAME]."_dlt_pipeline_state".version_hash, [SCHEMA_NAME]."_dlt_pipeline_state"."_dlt_load_id"
FROM [SCHEMA_NAME]."_dlt_pipeline_state" JOIN [SCHEMA_NAME]."_dlt_loads" ON [SCHEMA_NAME]."_dlt_loads".load_id = [SCHEMA_NAME]."_dlt_pipeline_state"."_dlt_load_id"
WHERE [SCHEMA_NAME]."_dlt_pipeline_state".pipeline_name = :pipeline_name_1 AND [SCHEMA_NAME]."_dlt_loads".status = :status_1 ORDER BY [SCHEMA_NAME]."_dlt_loads".load_id DESC]
```

If I manually create these tables in the target schema, this error no longer occurs and DLT proceeds.

This issue is similar to what is happening on the SAP HANA SQLAlchemy destination (dlt-hub/dlt#2110) and on SQL Server SQLAlchemy destination (dlt-hub/dlt#2193).

### Expected behavior

The 3 tables _dlt_pipeline_state, _dlt_version and _dlt_loads should be created by dlt if they're missing on the first run.

### Steps to reproduce

Run a pipeline where the destination is sqlalchemy and the destination server is Oracle

### Operating system

Linux

### Runtime environment

Local

### Python version

3.11

### dlt data source

sql_database

### dlt destination

[...]

PR #4001, opened 2026-05-29
fix(sqlalchemy): do not emit CREATE SCHEMA on Oracle (#3939)

### LINKED ISSUE #3939: Oracle 12c ORA-02420: missing schema authorization clause
### dlt version

1.26.0

### Describe the problem

Trying 
a) to fetch a small table from Oracle ( dates2 ) and then 
b) load it back to Oracle under a different table name ( dates3 ). 

The first operation finishes smooth, and without errors. The second one raises : 

(oracledb.exceptions.DatabaseError) ORA-02420: missing schema authorization clause
Help: https://docs.oracle.com/error-help/db/ora-02420/
[SQL: CREATE SCHEMA "ORACLE_USER_SCHEMA"]
(Background on this error at: https://sqlalche.me/e/20/4xp6)

### Expected behavior

To create metadata tables ( _dlt_loads, _dlt_pipeline_state, _dlt_version ) and materialize the table "dates3" in Oracle .

### Steps to reproduce

```
import dlt
from dlt.sources.sql_database import sql_table
from dlt.destinations import sqlalchemy

CREDS = "DRIVER + CONNECTION_STRING_GOES_HERE"

def load_tables() -> None:
    pipeline = dlt.pipeline(
        pipeline_name="oracle",
        destination=dlt.destinations.sqlalchemy(
            credentials=CREDS,
            enable_dataset_name_normalization=False
        ),
        dataset_name='ORACLE_USER_SCHEMA',
        dev_mode=False
    )
    src = sql_table(credentials=CREDS, table='dates2')  # a)
    print ( list(src) ) # Runs fine, shows table results 
    info = pipeline.run(src, write_disposition='replace', table_name='dates3') # b)

load_tables()
```

### Operating system

Windows

### Runtime environment

Local

### Python version

3.13

### dlt data source

Oracle through sqlalchemy

### dlt destination

Oracle through sqlalchemy

### Other deployment details

Targets Oracle through sqlalchemy

### Additional information

Related with :
[https://github.com/dlt-hub/dlt/issues/3141#issuecomment-3606265725](url)

python lib versions: 

dlt -> 1.26.0
oracledb -> 4.0.0
sqlalchemy -> 2.0.49
python -> 3.13.2

backend version:
Oracle -> 12c

DRACULA1729:
### Description

Loading to Oracle through the `sqlalchemy` destination fails on the first run with `ORA-02420: missing schema authorization clause`. dlt runs `CREATE SCHEMA "<dataset>"` when it cannot find the target dataset, and Oracle rejects that. In Oracle a schema is a database user, so there is no bare `CREATE SCHEMA` (it needs `CREATE SCHEMA AUTHORIZATION <user> ...`).

This adds dataset-lifecycle extension points to `DialectCapabilities` and routes the SQL client through them. The base implementations keep the old behavior for every other dialect, so nothing changes for Postgres/MySQL/etc.

`OracleDialectCapabilities` overrides them:

- `dataset_exists` matches case-insensitively, since Oracle folds unquoted identifiers to upper case. Loading into an existing schema is now detected and no creation is attempted.
- `create_dataset` no longer emits `CREATE SCHEMA`. If the schema is missing it raises a terminal error explaining that the schema (and the `<dataset>_staging` schema, for merge/replace) has to be created up front as a user with grants, instead of surfacing the cryptic ORA-02420.
- `drop_dataset` drops the tables inside the schema instead of `DROP SCHEMA`, which on Oracle would require `DROP USER` (a DBA privilege).

### Related Issues

- Fixes #3939

### Additional Context

- Follows the maintainer guidance in #3141: staging features are not hard-restricted; if the user pre-creates the `<dataset>_staging` schema, merge/replace still work.
- Oracle is not in the sqlalchemy-destination CI matrix, so the new tests are unit-level (mock-based) in `tests/load/sqlalchemy/test_sqlalchemy_dialect.py` and run without a live Oracle.
- Docs updated: the Oracle limitations section and the dialect-capabilities extension-points table in `sqlalchemy.md`.
- Ran locally: the new dialect tests pass; ruff, mypy, flake8 and black are clean on the changed files.

### DIFF
--- a/dlt/destinations/impl/sqlalchemy/db_api_client.py
+++ b/dlt/destinations/impl/sqlalchemy/db_api_client.py
@@ -209,7 +209,7 @@ class SqlalchemyClient(SqlClientBase[Connection]):
                 # has_schema resolves the dataset within the current catalog
                 return self.dialect.has_schema(self._current_connection, self.dataset_name)  # type: ignore[attr-defined,no-any-return]
             schema_names = self.engine.dialect.get_schema_names(self._current_connection)  # type: ignore[attr-defined]
-        return self._dialect_caps.dataset_exists(schema_names, self.dataset_name)
+        return self.dataset_name in schema_names
 
     def _sqlite_dataset_filename(self, dataset_name: str) -> str:
         current_file_path = Path(self.database_name)
--- a/dlt/destinations/impl/sqlalchemy/db_api_client.py
+++ b/dlt/destinations/impl/sqlalchemy/db_api_client.py
@@ -29,7 +29,6 @@ from dlt.destinations.exceptions import (
     DatabaseTerminalException,
     DatabaseTransientException,
     LoadClientNotConnected,
-    DatabaseException,
 )
 from dlt.common.destination.dataset import DBApiCursor
 from dlt.common.typing import TFun
@@ -210,7 +209,7 @@ class SqlalchemyClient(SqlClientBase[Connection]):
                 # has_schema resolves the dataset within the current catalog
                 return self.dialect.has_schema(self._current_connection, self.dataset_name)  # type: ignore[attr-defined,no-any-return]
             schema_names = self.engine.dialect.get_schema_names(self._current_connection)  # type: ignore[attr-defined]
-        return self.dataset_name in schema_names
+        return self._dialect_caps.dataset_exists(schema_names, self.dataset_name)
 
     def _sqlite_dataset_filename(self, dataset_name: str) -> str:
         current_file_path = Path(self.database_name)
@@ -287,15 +286,12 @@ class SqlalchemyClient(SqlClientBase[Connection]):
     def create_dataset(self) -> None:
         if self.dialect_name == "sqlite":
             return self._sqlite_create_dataset(self.dataset_name)
-        self.execute_sql(sa.schema.CreateSchema(self.dataset_name))
+        self._dialect_caps.create_dataset(self)
 
     def drop_dataset(self) -> None:
         if self.dialect_name == "sqlite":
             return self._sqlite_drop_dataset(self.dataset_name)
-        try:
-            self.execute_sql(sa.schema.DropSchema(self.dataset_name, cascade=True))
-        except DatabaseException:  # Try again in case cascade is not supported
-            self.execute_sql(sa.schema.DropSchema(self.dataset_name))
+        self._dialect_caps.drop_dataset(self)
 
     def truncate_tables(self, *tables: str) -> None:
         # TODO: alchemy doesn't have a construct for TRUNCATE TABLE
--- a/dlt/destinations/impl/sqlalchemy/dialect.py
+++ b/dlt/destinations/impl/sqlalchemy/dialect.py
@@ -4,12 +4,16 @@ Users can register custom DialectCapabilities subclasses to adapt the destinatio
 for dialects that are not built-in. See register_dialect_capabilities.
 """
 
-from typing import Any, Dict, Optional, Type
+from typing import Any, Dict, List, Optional, Type, TYPE_CHECKING
 
 import sqlalchemy as sa  # noqa
 
 from dlt.common.destination.capabilities import DataTypeMapper, DestinationCapabilitiesContext
 from dlt.common.destination.typing import PreparedTableSchema
+from dlt.destinations.exceptions import DatabaseException, DatabaseTerminalException
+
+if TYPE_CHECKING:
+    from dlt.destinations.impl.sqlalchemy.db_api_client import SqlalchemyClient
 
 
 _GENERIC_UNDEFINED_RELATION_PATTERNS = [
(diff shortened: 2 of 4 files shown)


DRACULA1729:
This is ready for review. The fork-gated workflows (the ones that need repo secrets) are stuck in the "requires reviewer approval" state, so they need a maintainer to approve the run before they execute.

@ivasio you dug into this exact Oracle case in #3141, so it's probably familiar territory. @rudolfix it extends the `DialectCapabilities` system from #3600 with dataset-lifecycle hooks. Happy to rebase or adjust anything.

rudolfix:
@DRACULA1729 were you able to run data loading correctly with this fix? oracle CI setup is still not ready on our side but my quick tests revealed a few further problems with type mapper, case folding etc. I'd also allow dlt to create and drop users, otherwise our test suites will never pass. that could be disabled in production tough

DRACULA1729:
Spun up Oracle 23 free-lite locally and dug into it. Here's what I found.

Loading into an existing schema (the user's own) works with the current fix: data loads and reads back fine. The breakage shows up on the second run, with `ORA-22848: cannot use CLOB type as comparison key`. dlt's `text` columns (`pipeline_name`, `version_hash`) map to CLOB, and `get_stored_state` compares them in WHERE/JOIN. That's the type-mapper problem you hit.

The tricky bit: the short key columns want `VARCHAR2`, but `schema` is already ~3.7KB for a single-table pipeline and `state` can grow large, so those have to stay CLOB. dlt's `text` type doesn't distinguish a short key from a large blob, so I wanted your call on the direction:
(a) give the internal key columns an explicit precision in the common schema (cleanest, but it touches every destination and changes the schema hash),
(b) keep an Oracle-local mapper that sends unbounded text to `VARCHAR2(4000)` and only the known large internal columns to CLOB, or
(c) override the state/schema lookups on Oracle so the CLOB comparison is done safely.
Which way would you go?

On case folding: the factory sets `casefold_identifier = str.lower` for Oracle, but Oracle folds unquoted identifiers to upper, so it should be `str.upper` for interop with existing tables. Easy fix.

On create/drop users: agreed it's needed for the suite to pass. To make it line up with your CI, will it connect to a PDB (so dataset-named users are valid without the `C##` prefix) and grant the loader `CREATE USER`/`DROP USER`? And what password and tablespace should created users get? Happy to implement once I know the shape you're aiming for so it doesn't clash with the CI you're setting up.

graham-macleod-nhsnss:
I'm not sure if this helps or hinders, but I am coming up against this bug in dlt 1.29.1, but the behaviour I'm seeing is slightly more worrying. Not only is it trying to create the schema for staging, it's trying to create the schema with the name {dataset_name}_{YYYYMMDDHH24MISS}. 

> [SQL: CREATE SCHEMA my_dataset_20260803123221]

Including `staging_dataset_name_layout` also doesn't force a consistent name to be used when trying to create, so even if a schema is created, it seems as though the existing schema would be disregarded and it would always attempt to create a new one.

```
    dest_ = dlt.destinations.sqlalchemy(staging_dataset_name_layout="%s_STAGING")
    pipeline = dlt.pipeline(
            pipeline_name="pipeline_name_here",
            destination=dest_,
            dataset_name="MY_DATASET",
            progress="log",
    )
```

I'll add this info to the issue as well.

DRACULA1729:
@rudolfix rebased on devel and cut this down.

To your question: yes, loading into an existing schema works. I ran it against Oracle 23 free-lite locally, data loads and reads back on the first run. The second run is where the CLOB comparison blows up, which is the type mapper problem and separate from what this PR does.

Two things I had wrong, both fixed here.

The case folding. I said `casefold_identifier` should be `str.upper` for Oracle. I don't think that's right. Oracle's `get_schema_names()` already runs names through `normalize_name()`, which lowercases anything all uppercase, so dlt's lowercase dataset name matches an existing uppercase schema with no help from us. dlt quotes its own identifiers after that, so it stays internally consistent. Moving to upper is a real interop decision and this bug doesn't need it.

Which makes the `dataset_exists` hook I added pointless, and a bit worse than pointless: on a quoted mixed case schema it would report "exists" and the table DDL would fail anyway. Dropped it, so `has_dataset` is back to what devel has.

I also dropped the `has_dataset()` re-check inside Oracle's `create_dataset`. `initialize_storage` only calls `create_dataset` when `has_dataset` returned False, so that branch could never run, and neither could the test I wrote for it.

What's left is two hooks (`create_dataset`, `drop_dataset`) plus the Oracle overrides: raise a readable error instead of ORA-02420, and drop the tables instead of DROP SCHEMA.

On the rest: for the CLOB keys I'll go with the Oracle local mapper unless you'd rather put explicit precision on the internal columns in the common schema. Happy to write create/drop user too, but I'd be guessing at PDB vs CDB, password and tablespace, and I'd rather line up with your CI than have you rip it out later. Both as follow up PRs so this one can go in on its own.

@ivasio you looked at this in #3141, so the shape of it may be familiar if you have a minute.

Also, the report from @graham-macleod-nhsnss on #3939 about `CREATE SCHEMA my_dataset_20260803123221` isn't a second bug, that's `dev_mode` adding a timestamp. I've answered on the issue.

Issue #4005, opened 2026-05-30
athena: iceberg append silently accumulates staging across loads; truncate flag is ineffective


tderk:
## Summary

`AthenaIcebergJobClient.should_truncate_table_before_load_on_staging_destination` hardcodes `False` for any iceberg table (`dlt/destinations/impl/athena/athena.py:467-472`). The general `truncate_tables_on_staging_destination_before_load` config flag added in #1717 is therefore ignored for iceberg targets.

Together with `SqlStagingFollowupJob._generate_insert_sql` doing an unbounded `INSERT INTO target SELECT cols FROM staging` (`dlt/destinations/sql_jobs.py:88-110`), this means:

- Append load 1: N rows land in staging, N rows in iceberg target. OK.
- Append load 2: M new rows land in staging — but staging now also still contains load 1's N rows. INSERT scans N+M rows. Iceberg target receives load 1's rows a second time.
- Load K: iceberg target has duplicated all prior loads K-1 times. Storage and Athena scan cost grow quadratically.

There is currently no opt-in to truncate the iceberg staging table per load.

## Environment

- dlt 1.x on `devel`
- Athena destination + filesystem staging
- Glue catalog
- `table_format=\"iceberg\"`, `write_disposition=\"append\"`

## Reproducer

```python
import dlt

@dlt.resource(write_disposition=\"append\", table_format=\"iceberg\")
def items():
    yield [{\"id\": 1}]

pipeline = dlt.pipeline(
    pipeline_name=\"athena_iceberg_append_bug\",
    destination=dlt.destinations.athena(query_result_bucket=\"s3://.../athena-results\"),
    staging=dlt.destinations.filesystem(bucket_url=\"s3://.../staging\"),
    dataset_name=\"bug_repro\",
)
pipeline.run(items())   # iceberg target: 1 row
pipeline.run(items())   # iceberg target: 3 rows (1 + 1 duplicate + 1 new)
pipeline.run(items())   # iceberg target: 6 rows
```

Expected: 1, 2, 3. Observed: 1, 3, 6.

## Root cause

`should_truncate_table_before_load_on_staging_destination` returns False unconditionally for iceberg, ignoring `truncate_tables_on_staging_destination_before_load` from the base config. The staging external table therefore accumulates every load's parquets, and the followup INSERT scans the whole accumulated staging.

## Proposed fix

[...]

PR #4006, opened 2026-05-30
fix(athena): add truncate_iceberg_staging opt-in to clear staging across loads

tderk:
## Summary

The Athena destination unconditionally returns `False` from `should_truncate_table_before_load_on_staging_destination` for iceberg tables (`dlt/destinations/impl/athena/athena.py:467-472`), so the generic `truncate_tables_on_staging_destination_before_load` flag introduced in #1717 has no effect on iceberg targets.

Combined with the unbounded `INSERT INTO target SELECT * FROM staging` in `SqlStagingFollowupJob` (`dlt/destinations/sql_jobs.py:88-110`), appending into an iceberg target re-copies every previous load's rows on each subsequent load: storage and Athena-scan cost grow quadratically and the iceberg table accumulates duplicates.

This PR adds an explicit opt-in config flag on `AthenaClientConfiguration` so operators can request a per-load truncate of the iceberg staging external table. Default value preserves existing behaviour.

## Closes

- #4005

## Behaviour matrix

| table_format | write_disposition | `truncate_iceberg_staging` | truncate? |
|---|---|---|---|
| non-iceberg | replace | any | True (unchanged) |
| non-iceberg | append | any | False (unchanged) |
| iceberg | any | None / False | False (unchanged) |
| iceberg | any | True | **True (new)** |

## Tests

- `tests/load/athena_iceberg/test_athena_iceberg.py::test_iceberg_staging_truncate_flag` — two appends of disjoint payloads with `truncate_iceberg_staging=True`; asserts target row count == sum of payloads (without the flag the second load would also re-insert the first load's rows).

## Lint

```
$ uv run ruff check dlt/destinations/impl/athena/configuration.py dlt/destinations/impl/athena/athena.py tests/load/athena_iceberg/test_athena_iceberg.py
All checks passed!
$ uv run mypy dlt/destinations/impl/athena/configuration.py dlt/destinations/impl/athena/athena.py
(no output)
```

Test requires the Athena CI credentials label.

## Related

- #1717 — original truncate flag, which this PR extends for iceberg
- Follow-up: a separate PR will scope the staging→iceberg `INSERT INTO` to the current load via `_dlt_load_id`, as defence-in-depth for users who keep accumulated staging on purpose (e.g. audit). Filed as a sibling issue.

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -465,9 +465,15 @@ class AthenaClient(SqlJobClientWithStagingDataset, SupportsStagingDestination):
         return super().should_load_data_to_staging_dataset(table_name)
 
     def should_truncate_table_before_load_on_staging_destination(self, table_name: str) -> bool:
-        # on athena we only truncate replace tables that are not iceberg
         table = self.prepare_load_table(table_name)
-        if table["write_disposition"] == "replace" and not self._is_iceberg_table(table):
+        if self._is_iceberg_table(table):
+            # Iceberg targets get their data copied out of the staging external
+            # table with ``INSERT INTO target SELECT * FROM staging``. Without
+            # an opt-in truncate, parquet files accumulate in the staging
+            # location across loads and every subsequent ``INSERT`` re-copies
+            # the older loads' rows into the iceberg target.
+            return bool(self.config.truncate_iceberg_staging)
+        if table["write_disposition"] == "replace":
             return True
         return False
 
--- a/dlt/destinations/impl/athena/configuration.py
+++ b/dlt/destinations/impl/athena/configuration.py
@@ -29,6 +29,11 @@ class AthenaClientConfiguration(DestinationClientDwhWithStagingConfiguration):
     force_iceberg: Optional[bool] = None
     table_location_layout: Optional[str] = DEFAULT_TABLE_LOCATION_LAYOUT
     table_properties: Optional[Dict[str, str]] = None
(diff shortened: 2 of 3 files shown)


tderk:
Closing as redundant. Empirical verification in a live Athena + filesystem-staging setup (Salesforce Product2, 2 consecutive append loads, no flag set) shows the staging-dataset truncate ALREADY happens by default for iceberg tables via the existing `should_load_data_to_staging_dataset_on_staging_destination=True` path: `init_client` includes iceberg tables in `staging_tables_with_jobs`, which is passed as `truncate_tables` to the staging-dataset `_init_dataset_and_update_schema` call (load/utils.py:165). After iter 2, the staging path holds only iter 2's parquet file; iter 1's was truncated. Target row count = sum of per-iter rows (no duplication).

Additionally, the patch in this PR introduces a regression in shared-bucket setups: returning True from `should_truncate_table_before_load_on_staging_destination` for iceberg makes `init_client`'s FIRST `_init_dataset_and_update_schema` (filesystem main-role) call `filesystem.truncate_tables([table])` against the main dataset path. When the Athena iceberg target location shares an S3 prefix with the staging filesystem's main dataset (the default when `staging=filesystem(bucket_url=...)` inherits the destination's dataset_name), the truncate wipes the iceberg metadata/data dirs and the next INSERT fails with `ICEBERG_MISSING_METADATA`.

PR #4008 (issue #4007) already addresses the actual append-duplication risk via load-scoped `WHERE _dlt_load_id = '...'` in the staging→iceberg copy, which is a smaller surface change.

Issue #4007, opened 2026-05-30
athena: iceberg staging→target INSERT is not scoped to the current load


tderk:
## Summary

`SqlStagingFollowupJob._generate_insert_sql` (`dlt/destinations/sql_jobs.py:88-110`) emits an unbounded `INSERT INTO target(cols) SELECT cols FROM staging` whenever a staging copy runs. For the Athena destination targeting iceberg tables, this means: any parquet file that happens to be in the staging external table's S3 prefix is copied into the iceberg target, regardless of which load wrote it.

Combined with #4005 (iceberg staging is never truncated unless the operator opts in), the second and subsequent loads re-copy older loads' rows into the iceberg target. Even with the truncate flag from #4005 enabled, scoped INSERT is still valuable as defence-in-depth: orphan parquets from failed loads, manual uploads, or eventual-consistency races during S3 lifecycle expiry can still leak rows into the next iceberg INSERT.

## Reproducer

```python
import dlt

@dlt.resource(write_disposition="append", table_format="iceberg")
def items():
    yield [{"id": 1}]

pipeline = dlt.pipeline(
    pipeline_name="athena_iceberg_scope_bug",
    destination=dlt.destinations.athena(query_result_bucket="s3://.../athena-results"),
    staging=dlt.destinations.filesystem(bucket_url="s3://.../staging"),
    dataset_name="bug_repro",
)
pipeline.run(items())   # iceberg target: 1 row
pipeline.run(items())   # iceberg target: 3 rows (1 + 1 duplicate + 1 new)
```

Expected: 1, then 2. Observed: 1, then 3.

## Proposed fix (Athena iceberg only)

Override the staging copy followup job for iceberg targets so the SELECT is scoped to the current load via the always-present `_dlt_load_id` column:

```sql
INSERT INTO target(cols)
SELECT cols FROM staging
WHERE _dlt_load_id = '<current load_id>'
```

The current load id is available inside the followup-job generator via `dlt.common.storages.load_package.load_package_state` (already used by the merge job at `dlt/destinations/sql_jobs.py:22,950`). The `_dlt_load_id` column is present on every dlt-managed table by default (`dlt/common/schema/typing.py:47`).

### Query cost

[...]

PR #4008, opened 2026-05-30
fix(athena): scope iceberg staging->target INSERT to current load

tderk:
## Summary

For Athena iceberg targets the staging external table accumulates parquet files across loads (`should_truncate_table_before_load_on_staging_destination` hardcodes False, see #4005). Each subsequent `INSERT INTO target SELECT * FROM staging` in `SqlStagingFollowupJob` therefore re-copies older loads' rows into the iceberg target — duplicates accumulate and Athena scan cost grows quadratically.

This PR overrides the staging copy followup job for iceberg targets with a `WHERE _dlt_load_id = '<current load_id>'` clause. The current load id is resolved via `dlt.common.storages.load_package.load_package_state` (the same mechanism the merge followup job already uses at `dlt/destinations/sql_jobs.py:22,950`).

## Closes

- #4007

## Query cost

Athena's parquet reader applies predicate pushdown using row-group min/max statistics. Every staging parquet written by a single load has a constant `_dlt_load_id`, so the WHERE clause prunes non-matching files at the row-group reader stage — effectively per-load scoping without a Hive partition setup. No filesystem-side layout changes, no per-load `ALTER ADD PARTITION`, no cross-destination behaviour change.

## Opt-out

`AthenaClientConfiguration.scope_iceberg_staging_insert_to_load` (Optional[bool], default None / safe path). Set to `False` to restore the legacy unbounded `SELECT *` behaviour for operators who deliberately rely on the staging accumulation semantic.

Defensive fallback to the unbounded `SELECT` when `_dlt_load_id` is not on the table or the load package state is unavailable.

## Behaviour matrix

| iceberg | `scope_iceberg_staging_insert_to_load` | INSERT shape |
|---|---|---|
| no | any | unchanged — non-iceberg path |
| yes | None / True (default) | `INSERT ... SELECT cols FROM staging WHERE _dlt_load_id = '<load>'` |
| yes | False | `INSERT ... SELECT cols FROM staging` (legacy) |

## Scope

Strictly Athena iceberg followup. Other destinations consuming `SqlStagingFollowupJob` (BigQuery, ClickHouse, Databricks, Dremio, Fabric, Synapse, MS SQL, Postgres, Redshift, Snowflake) keep their current behaviour. No changes to merge or replace followups.

## Tests

- `tests/load/athena_iceberg/test_athena_iceberg.py::test_iceberg_staging_insert_scoped_to_load` — two consecutive appends of disjoint payloads; asserts target row count == sum of payloads (without the fix the second load would yield 8 rows because of duplicate copy).

## Lint

```
$ uv run ruff check dlt/destinations/impl/athena/configuration.py dlt/destinations/impl/athena/athena.py tests/load/athena_iceberg/test_athena_iceberg.py
All checks passed!
$ uv run mypy dlt/destinations/impl/athena/athena.py
(no errors)
```

Test requires the Athena CI credentials label.

## Related

- #4005 / PR #4006 — sibling truncate flag. The two PRs are complementary: this one is correct on its own, and works as defence-in-depth even when the operator also enables the truncate flag (orphan parquets, eventual-consistency races during lifecycle expiry, manual uploads).

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -3,6 +3,7 @@ from typing import (
     Any,
     Sequence,
     Tuple,
+    Type,
     List,
     Dict,
     Iterable,
@@ -42,6 +43,9 @@ from dlt.common.schema.typing import (
 from dlt.common.destination import DestinationCapabilitiesContext, PreparedTableSchema
 from dlt.common.destination.utils import resolve_merge_strategy
 from dlt.common.destination.client import FollowupJobRequest, SupportsStagingDestination, LoadJob
+from dlt.common.schema.typing import C_DLT_LOAD_ID
+from dlt.common.schema.utils import get_columns_names_with_prop
+from dlt.common.storages.load_package import load_package_state as current_load_package
 from dlt.destinations.sql_jobs import (
     SqlStagingCopyFollowupJob,
     SqlStagingReplaceFollowupJob,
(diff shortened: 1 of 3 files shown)


tderk:
Closing as redundant. Same reasoning as PR #4006 (issue #4005): empirical verification shows the staging external table for iceberg targets is already truncated by default before each load via the existing `should_load_data_to_staging_dataset_on_staging_destination` path. `init_client` (load/utils.py:165) includes iceberg tables in `staging_tables_with_jobs`, which is passed as `truncate_tables` to the staging-dataset `_init_dataset_and_update_schema` call. After each load completes, only the current load's parquet remains in the staging path — verified by listing S3 after 2 consecutive append loads of Salesforce Product2 into an Athena iceberg target: only the second load's parquet file present, first load's file truncated. Target row count = sum of per-iter rows (6800 = 2 × 3400, no duplication).

Since staging only ever contains the current load's data at copy time, the unscoped `INSERT INTO target SELECT * FROM staging` in `SqlStagingCopyFollowupJob` already inserts the correct set of rows. The load-scoped `WHERE _dlt_load_id = '...'` filter this PR introduces is a no-op in practice.

Sorry for the noise — the underlying premise ("staging accumulates across loads → unbounded INSERT re-copies older loads") turned out not to match observed runtime behaviour.

Issue #3823, opened 2026-04-02
Athena: configurable S3 root for Iceberg table location independent of staging bucket


jflipts:
### Feature description

The Athena destination derives the final Iceberg table LOCATION from staging_config.bucket_url (hardcoded in athena._get_table_update_sql). There is no way to configure a separate S3 root for the final table data. The staging Parquet files and final Iceberg table data always share the same S3 bucket/prefix root while there is no need for this.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

Our environment has strict S3 prefix-based access control where we want to keep staging data internal under a prefix like s3://datalake/source_system/private/... while the final tables should be stored on a sibling prefix like s3://datalake/source_system/output/...

### Proposed solution

Add an optional table_bucket_url field to AthenaClientConfiguration. When set, _get_table_update_sql uses it as the S3 root for the Iceberg table LOCATION while continuing to use staging_config.bucket_url for the staging EXTERNAL TABLE LOCATION.

PR #4009, opened 2026-05-30
feat(athena): allow separate bucket for iceberg target via iceberg_bucket_url

tderk:
## Summary

The Athena destination derives the iceberg target table `LOCATION` from `self.config.staging_config.bucket_url` (`dlt/destinations/impl/athena/athena.py:393, :437`). Hive external staging tables and final iceberg data therefore have to share an S3 root. Deployments with prefix-scoped IAM policies, per-prefix lifecycle rules, or bucket-level KMS keys cannot separate the short-lived staging zone from the durable iceberg storage today.

This PR adds an optional `iceberg_bucket_url` config field. When set, the iceberg table `LOCATION` uses it as the root; staging external tables continue to live under `staging_config.bucket_url`. Default value preserves existing behaviour.

## Closes

- #3823

## Behaviour

| `iceberg_bucket_url` | Iceberg `LOCATION` root | Hive external (staging) | S3 Tables Catalog |
|---|---|---|---|
| None (default) | `staging_config.bucket_url` (unchanged) | `staging_config.bucket_url` | catalog-managed |
| explicit S3 URL | configured value | `staging_config.bucket_url` (unchanged) | catalog-managed (unchanged) |

S3 Tables Catalog is intentionally unaffected — the catalog manages its own locations, same as `table_location_layout` is ignored on that path today.

## Tests

- `tests/load/athena_iceberg/test_athena_configuration.py::test_iceberg_bucket_url_optional`
  - `iceberg_bucket_url=None` resolves to `None` (default, no AWS call needed).
  - Explicit `s3://my-iceberg-target` round-trips through resolved config.

Local run:
```
$ uv run pytest tests/load/athena_iceberg/test_athena_configuration.py::test_iceberg_bucket_url_optional -xvs
... 1 passed in 0.65s
```

## Lint

```
$ uv run ruff check dlt/destinations/impl/athena/ tests/load/athena_iceberg/
All checks passed!
$ uv run mypy dlt/destinations/impl/athena/configuration.py dlt/destinations/impl/athena/athena.py
(no errors)
```

## Scope

Strictly Athena destination. Only the iceberg `LOCATION` is decoupled — no changes to staging path, query result bucket, merge/replace jobs, or other destinations. No data migration required; existing tables continue to resolve to their stored `LOCATION` so this PR is purely additive for new tables.

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -332,7 +332,11 @@ class AthenaClient(SqlJobClientWithStagingDataset, SupportsStagingDestination):
     def _get_table_update_sql(
         self, table_name: str, new_columns: Sequence[TColumnSchema], generate_alter: bool
     ) -> List[str]:
-        bucket = self.config.staging_config.bucket_url
+        staging_bucket = self.config.staging_config.bucket_url
+        # Iceberg targets may live in a dedicated bucket independent of the
+        # staging external tables; fall back to staging bucket when not set
+        # (preserves prior behaviour). See https://github.com/dlt-hub/dlt/issues/3823
+        iceberg_bucket = self.config.iceberg_bucket_url or staging_bucket
         dataset = self.sql_client.dataset_name
 
         sql: List[str] = []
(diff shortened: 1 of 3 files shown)


tderk:
Closing on the author's side. The original issue #3823 is still valid (decoupling iceberg target storage from staging is a useful capability), but for our own deployment we ended up not needing it — the shared-bucket layout works correctly without this PR once the assumptions in PR #4006 and PR #4008 are dropped (see the close comments on those PRs). Happy to revisit if a maintainer thinks this is the right shape for the feature.

PR #4011, opened 2026-06-01
fix: emit foreign key only if create_indexes is enabled in databricks

burnash:
On the Databricks destination, a references hint emitted `ADD FOREIGN KEY` unconditionally even with the default `create_indexes=False`. Since the matching `PRIMARY KEY/UNIQUE` is only created when `create_indexes=True`, Unity Catalog rejected the foreign key and failed with (UC_REFERENTIAL_CONSTRAINT_DOES_NOT_EXIST):

```
ErrorClass=NOT_FOUND.UC_REFERENTIAL_CONSTRAINT_DOES_NOT_EXIST] Table `dlt_ci.join_relational_202605291019544316.users` does not have either a primary key constraint or a UNIQUE constraint.
```

CI: https://github.com/dlt-hub/dlt/actions/runs/26630632112/job/78480942833


rudolfix (approved):
LGTM!

PR #4016, opened 2026-06-02
Preserve timezone=False for naive Arrow timestamps

### LINKED ISSUE #4015: Naive Arrow timestamps lose timezone=False and are loaded as UTC-aware instead of naive
### dlt version

1.27.2

### Describe the problem

When a source returns Arrow data with naive timestamps, DLT infers `timezone=False`
 for those columns. During capabilities adjustment, that explicit hint is dropped, so the downstream loader treats the column as timezone-aware by default and writes it as UTC-aware instead of keeping it naive.

This breaks [the documented pass-through behavior](https://dlthub.com/docs/general-usage/schema#handling-of-timestamp-and-time-zones) for explicit `timezone` hints and affects Arrow-based extraction paths, including ConnectorX in arrow and arrow_stream modes.

### Expected behavior

If a timestamp column is naive in Arrow and/or has `timezone=False` explicitly set, DLT should preserve that hint end-to-end and load the column as a naive timestamp in the destination.

### Steps to reproduce

1. Use a SQL source that returns both:
  * `timestamp without time zone`
  * `timestamp with time zone`
2. Extract via ConnectorX or another Arrow-based path.
3. Run the pipeline to Parquet/filesystem.
4. Inspect the resulting schema and Parquet types.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

PostgreSQL

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

dbascoules:
### Description

This PR fixes a regression where DLT dropped the explicit timezone=False hint inferred from naive Arrow timestamps during capabilities adjustment.

Previously, Arrow input with naive timestamps was normalized as if the timezone hint were absent, which caused downstream loading to convert values to UTC-aware timestamps instead of keeping them naive. This affected [pass-through behavior](https://dlthub.com/docs/general-usage/schema#handling-of-timestamp-and-time-zones) for sources that already expose naive Arrow timestamps, including ConnectorX-based extraction.

The fix preserves timezone=False when it is explicitly present on a column, while continuing to strip other timezone values so the default behavior remains unchanged where no explicit hint is set.

### Related Issues

- Fixes #4015

### Additional Context

- Added or updated tests covering naive Arrow timestamps and timezone-aware timestamps.
- Verified the fix with ConnectorX in both arrow and arrow_stream modes.
- Confirmed naive timestamps stay naive in Parquet output and timestamptz values remain UTC-aware.

### DIFF
--- a/dlt/common/destination/capabilities.py
+++ b/dlt/common/destination/capabilities.py
@@ -330,7 +330,8 @@ def adjust_column_schema_to_capabilities(
 ) -> TColumnSchema:
     """Modify column schema in place according to destination capabilities.
     * timestamp and time precision are limited by min(destination max, parquet writer max)
-    * timezone is always removed (enables default behavior)
+    * timezone=False inferred from naive Arrow timestamps is preserved, other timezone hints fall back
+      to default behavior
     * default timestamp precision for destination is removed
     """
     data_type = column.get("data_type")
@@ -347,8 +348,9 @@ def adjust_column_schema_to_capabilities(
             ),
         )
 
-        # remove timezone flag to fallback to default tz-aware and let normalizer to correct the data
-        column.pop("timezone", None)
+        # preserve explicit timezone=False inferred from naive Arrow timestamps
+        if column.get("timezone") is not False:
+            column.pop("timezone", None)
 
         # do not write default precision
         if precision == caps.timestamp_precision:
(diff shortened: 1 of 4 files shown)


rudolfix:
what you say looks right! this behavior is thoroughly tested - I need to see why some of the tests had failed.

PR #4021, opened 2026-06-03
fix(duckdb): refresh credential_chain secrets to survive temp-token expiry (#3987)

### LINKED ISSUE #3987: filesystem SQL client S3 secret uses credential_chain without REFRESH, so temp creds aren't refreshed on long reads
### dlt version

1.27.0

### Describe the problem

TLDR: On longer jobs, S3's cred can timeout on duckdb's `sql_client` in ~8h without refresh.

When reading a filesystem dataset via `dlt.dataset(...).sql_client` (DuckDB), the S3 secret is created once on `open_connection`. For default-chain AWS credentials, `create_secret` in `dlt/destinations/impl/duckdb/sql_client.py` emits:

```sql
CREATE OR REPLACE SECRET ... (
    TYPE S3,
    PROVIDER credential_chain,
    REGION '...', ENDPOINT '...', SCOPE '...', URL_STYLE '...', USE_SSL ...
)
```

This resolves the credential chain once and never refreshes. On a long-held connection (we stream Arrow batches for hours while a slow downstream consumes them), the temporary ECS task-role token expires and the next lazy `GET` fails:

```
OSError: HTTP Error: HTTP GET error on 's3://.../part-00010-....parquet' (HTTP 400)
```

DuckDB supports `REFRESH auto` on `credential_chain` secrets, which re-runs the chain when credentials expire. dlt never sets it, so there's no way to keep a long-lived `sql_client` reading past the credential TTL without manually re-creating the secret.

### Expected behavior

Secrets created from `credential_chain` should include `REFRESH auto` (or expose an option to enable it), so long-running reads on temporary credentials don't die with HTTP 400 on expiry.

### Steps to reproduce

1. Run on ECS/Lambda (temporary role creds) or any setup where the default chain returns a short-lived token.
2. Open a filesystem dataset and stream slowly:

```python
ds = dlt.dataset(
    destination=dlt.destinations.filesystem(bucket_url="s3://my-bucket/path"),
    dataset_name="my_dataset",
)
with ds.sql_client as client:
    with client.execute_query("SELECT * FROM my_table") as cursor:
        for batch in cursor.iter_arrow(chunk_size=5000):
            time.sleep(...)  # simulate slow consumer; hold the connection past creds TTL
```

3. Once the token expires, the next S3 GET fails with HTTP 400.

Workaround: after opening the connection, `CREATE OR REPLACE SECRET <dlt_secret_name> (TYPE S3, PROVIDER credential_chain, REFRESH auto, ...)` to override dlt's secret.

### Operating system

Linux

### Runtime environment

AWS ECS (Fargate)

### Python version

3.13

### dlt data source

`filesystem` (S3, DuckDB SQL client / `dlt.dataset`)

0ywfe:
## Summary

Closes #3987 — DuckDB S3 secrets created via `PROVIDER credential_chain` never refresh, so long-held `sql_client` connections die with HTTP 400 `ExpiredToken` once the temporary AWS token rotates (typical lifetime ~8h on ECS / EKS IRSA / EC2 instance profile).

DuckDB 1.1.0 added `REFRESH auto` on credential_chain secrets, which re-runs the provider chain on expiry. This is the DuckDB-documented best practice for any long-running consumer of credential_chain — without it the whole point of using a refreshable chain is defeated by the snapshot-once behaviour.

### Behavioral changes
- **DuckDB S3 + default credentials**: secret is created with `PROVIDER credential_chain` + `REFRESH auto` (re-resolves the AWS chain on expiry) instead of frozen static keys (#3987). Static keys and external boto sessions stay frozen.
- **AWS handover**: `to_s3fs_credentials` / `to_object_store_rs_credentials` / `to_pyiceberg_fileio_config` omit static key/secret/token for *refreshable* default credentials (IMDS/ECS) and hand over to the consumer's own chain so it refreshes. Deferred (SSO / IRSA / assume-role), static, and external sessions remain frozen.
- **AWS `to_sts_credentials()`** now sets STS `DurationSeconds` from `aws_sts_token_expiration_hours` (default 24h, previously STS's implicit 12h) — affects e.g. Databricks S3 staging.
- **Azure default credentials** are handed over to each consumer's own chain (object_store resolves env service principal / workload identity / managed identity itself; managed identity is no longer frozen to a ~1h token).
- **Azure + iceberg (pyiceberg/adlfs)**: default credentials require `AZURE_STORAGE_ANON=false` in the environment (adlfs defaults to anonymous and pyiceberg sets no `anon` flag); external (pinned) azure sessions raise `UnsupportedAuthenticationMethodException` for iceberg.

### New
- **External azure sessions**: `AzureCredentials.from_credential(...)` / `AzureServicePrincipalCredentials.from_credential(...)` accept a user-passed azure-identity credential (mirrors AWS `from_session`). Always frozen — bearer token for object_store/DuckDB, live credential object for adlfs.
- **DuckDB azure default credentials** now produce a secret (previously none for the default/managed-identity case): `PROVIDER credential_chain` for ambient default, `PROVIDER access_token` (frozen bearer) for external sessions.
- **Configurable minted-token lifetimes**: `azure_sas_token_expiration_hours` (account-key SAS) and `aws_sts_token_expiration_hours` (STS), both default 24h.
- New `InvalidAzureCredential` exception.


rudolfix:
@0ywfe besides missing REFRESH, we were missing a proper handover of azure credentials. we also did a bigger overhaul of the credentials handover. I've added a few commits to your PR that should fix most of the edge cases in duckdb. thx for the fix!

rudolfix (commented):
LGTM!

PR #4026, opened 2026-06-05
fix: resolve GCP service-account vs oauth credentials correctly (#3328)

### LINKED ISSUE #3328: dlt tries OAuth in GCP Instead of Service Account Credentials when given Service Account Credentials via credential file
### dlt version

dlt 1.17.1

### Describe the problem

When using GCP service account credentials, referenced either in environment variables or in `secrets.toml` I get this following error.

```
`GcpOAuthCredentials` cannot parse the configuration value provided. The value is of type `str` and comes from the sections `('credentials',)` Value may be a secret and is not shown. Details:  The expected representation for `GcpOAuthCredentials` is a string with serialized oauth2 user info and may be wrapped in 'install'/'web' node - depending of oauth2 app type.
```

### Expected behavior

dlt should be using `GcpServiceAccountCredentials`, which currently can be used when explicitly provided within  the pipeline as in

```
    with open("creds.json", "r") as f:
        creds =  json.load(f)

    service_dict = json.loads(json.dumps(creds))

    gcp_credentials = GcpServiceAccountCredentials()
    gcp_credentials.parse_native_representation(json.dumps(creds))
```

### Steps to reproduce

set up GCP service account credentials and attempt to use them either in secrets.toml as in 

```
[destination.filesystem]
bucket_url = "gs://yourbucket...

[destination.filesystem]
credentials="path/to/your/json"
```

when running a dlt pipeline for filesystem destination

### Operating system

macOS

### Runtime environment

Local

### Python version

3.10

### dlt data source

API

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

Pawansingh3889:
Closes #3328.

### What's going on

GCS / filesystem credentials are typed as `Union[GcpServiceAccountCredentials, GcpOAuthCredentials]`. When dlt resolves a union it only falls through to the next member if a spec raises `NativeValueError`. The problem: neither GCP spec checked that the JSON actually matched its own type.

- `GcpOAuthCredentials.parse_native_representation` happily `json.loads`-es a *service-account* JSON (a service-account file even has a `client_id`), so resolution can land on oauth.
- When things do fail, the union re-raises the **last** member's error, which is why people see `GcpOAuthCredentials cannot parse the configuration value provided...` even though they supplied a service account. That's both the error in #3328 and the "redirects to the OAuth consent screen instead of using the service account" symptom from the google_sheets source.

### Changes

- `GcpOAuthCredentials` rejects a JSON that looks like a service account (`type == "service_account"` or a `private_key` present) by raising `InvalidGoogleOauth2Json`.
- `GcpServiceAccountCredentials` rejects oauth client-secret / user info (`client_secret` / `installed` / `web` / `type == "authorized_user"`) by raising `InvalidGoogleServicesJson`.
- Both are `NativeValueError` subclasses, so a credentials union now falls through to the correct spec instead of mis-parsing.
- `credentials` may now also be a **path to a json file** — if the value is an existing file it's read from disk, so `credentials = "/path/to/service_account.json"` works (the literal case in the issue).

### Tests

Added to `tests/common/configuration/test_credentials.py`:
- each spec rejects the other's JSON,
- a `Union[...]` resolves a service-account JSON to the service-account spec and oauth info to the oauth spec,
- credentials are read from a file path.

The GCP/oauth credential tests pass locally (9 passed, including the existing ones), `black` and `flake8` are clean, and no existing credential test changed behaviour.

### DIFF
--- a/dlt/common/configuration/specs/gcp_credentials.py
+++ b/dlt/common/configuration/specs/gcp_credentials.py
@@ -1,4 +1,5 @@
 import dataclasses
+import os
 import sys
 from typing import Any, ClassVar, Final, List, Tuple, Union, Dict, Optional
 
@@ -23,9 +24,25 @@ from dlt.common.configuration.specs.base_configuration import (
 )
 from dlt.common.utils import is_interactive
 
-
 # GCS scope required for OAuth2 token requests
 GCS_SCOPE = "https://www.googleapis.com/auth/devstorage.read_write"
+
+
+def _maybe_read_credentials_file(native_value: str) -> str:
+    """Return file contents if ``native_value`` is a path to an existing file.
+
+    Lets users point ``credentials`` at a service-account / oauth json *file*
+    instead of pasting the json itself. Anything that is not an existing file
+    path (json content included) is returned unchanged.
+    """
+    try:
+        if os.path.isfile(native_value):
+            with open(native_value, encoding="utf-8") as f:
+                return f.read()
+    except (OSError, ValueError):
+        # value too long for a path, contains null bytes, etc. -> not a file
+        pass
+    return native_value
 
 
 def _get_pyiceberg_fileio_config(credentials: Any, project_id: Optional[str]) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


Pawansingh3889:
Rebased onto devel (the branch had drifted a long way behind) and pushed; the fix and its tests are unchanged and apply cleanly. #3328 is still open and this resolves the service-account vs oauth detection it describes. CI will need a maintainer to approve the workflow run for a fork PR. Happy to adjust anything on review.

PR #3871, opened 2026-04-16
fix: DuckLake DuckDB-backed catalog attach incorrectly applies META_TYPE 'sqlite'

Analect:
drivername='duckdb' was handled identically to 'sqlite' in build_attach_statement, adding META_TYPE 'sqlite' / META_JOURNAL_MODE 'WAL' parameters that cause DuckDB-format catalog files to fail with:

`Failed to execute query "PRAGMA journal_mode=WAL": file is not a database`

Split the combined elif branch so `drivername='duckdb'` generates a clean attach:

    ATTACH IF NOT EXISTS 'ducklake:{path}' AS {name} (DATA_PATH '...')

The `DuckLakeCredentials` docstring explicitly lists `duckdb:///catalog.duckdb` as a valid catalog URI, confirming this is a supported use case.

#### Problem
When using a DuckDB-backed DuckLake catalog (`catalog="duckdb:///catalog.duckdb"`),
`pipeline.run()` fails with:

    dlt.destinations.exceptions.DestinationConnectionError:
    Failed to execute query "PRAGMA journal_mode=WAL": file is not a database

#### Root cause

`build_attach_statement` in `sql_client.py` handled both `drivername='sqlite'`
and `drivername='duckdb'` in a single branch, always appending
`META_TYPE 'sqlite', META_JOURNAL_MODE 'WAL', META_BUSY_TIMEOUT 1000`.
`META_TYPE 'sqlite'` causes DuckLake to use SQLite operations on the catalog
file. For a DuckDB-format file this hits the SQLite `PRAGMA journal_mode=WAL`
and raises "file is not a database".

#### Fix

Split the combined `elif` into two separate branches. `drivername='duckdb'`
now generates a clean attach without `META_TYPE`:

    ATTACH IF NOT EXISTS 'ducklake:{path}' AS {name} (DATA_PATH '...')

#### Tests

Added `tests/load/pipeline/test_ducklake_attach.py` with unit tests for
`build_attach_statement` covering both the sqlite and duckdb cases. Tests
can be run without external credentials.

### Related Issues

- Fixes #3870

### Additional Context                                                                                                                                                 
   
Tests pass without external credentials or a live DuckLake instance: 

\```
pytest tests/load/pipeline/test_ducklake_attach.py -v

tests/load/pipeline/test_ducklake_attach.py::test_sqlite_catalog_includes_meta_type_sqlite PASSED
[...]

PR #4034, opened 2026-06-09
(fix) propagate load worker setup failures as retries

### LINKED ISSUE #3951: Load jobs that fail on setup end up orphaned
### dlt version

1.26.0

### Describe the problem

After upgrading from 1.24.x -> 1.26.0 we noticed that our dev job would sometimes hang after running for several minutes. The last log lines printed were the ones at the end of the `AdbcParquetCopyJob.run` method that show that a table was completed. The database showed no queries in flight.

### Expected behavior

I would expect the job to continue or throw an exception.

### Steps to reproduce

This is intermittent. But we were loading 6 tables into Azure SQL via Parquet with a couple of them having several million rows.

### Operating system

Linux

### Runtime environment

Other

### Python version

3.11

### dlt data source

Sql Server

### dlt destination

_No response_

### Other deployment details

We are loading to Azure Sql using the parquet ADBC loader.

### Additional information

In order to troubleshoot this issue, I checked out the DLT repo and had co-pilot analyze all the callers of the run method to look at what happens right after the call returns. It generated the attached theory that it was deadlocking due to exceptions being thrown in the finally block. `RunnableLoadJob.run_managed` has changed across these versions. There is low a call to `self._on_completed(self._state, failed_message)` that was not there in earlier versions. The theory is that this method can sometimes throw an exception and cause the job to deadlock.

[troubleshooting_load_job_freezes.md](https://github.com/user-attachments/files/28011821/troubleshooting_load_job_freezes.md)

Travior:
Closes #3951

### DIFF
--- a/dlt/load/load.py
+++ b/dlt/load/load.py
@@ -278,19 +278,18 @@ class Load(Runnable[Executor], WithStepInfo[LoadMetrics, LoadInfo]):
         """
         Start a load job in a separate thread
         """
-        active_job_client = (
-            self.get_staging_destination_client(schema)
-            if use_staging_client
-            else self.get_destination_client(schema)
-        )
         try:
+            active_job_client = (
+                self.get_staging_destination_client(schema)
+                if use_staging_client
+                else self.get_destination_client(schema)
+            )
             with self.maybe_with_staging_dataset(active_job_client, use_staging_dataset):
                 job.run_managed(active_job_client, self._done_event)
         except Exception as e:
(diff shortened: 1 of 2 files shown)


Travior:
I believe this is actually mainly solved by https://github.com/dlt-hub/dlt/commit/2cb31eb22bd8203b45455982b67d136b3acf5ab8 in #3871 . The setup calls are now side effect free, opening a connection now only happens when the job state is "running" already.

PR #4039, opened 2026-06-09
fix(extract): start resource progress timer at pipe loop start (#3518)

### LINKED ISSUE #3518: RestAPI incorrectly reports resource runtime for non-paginated requests
### dlt version

1.18.2

### Describe the problem

When using the RestAPI source to pull API resources that are singular requests that take a long time for TTFB, time to first byte, the timing report does not accurately reflect the response. In the following logs the request takes ~6 minutes to get the response and returns a large ~100mb json array with 34k records in it. 

```
------------------------------- Extract rest_api -------------------------------
Resources: 0/1 (0.0%) | Time: 0.00s | Rate: 0.00/s

------------------------------- Extract rest_api -------------------------------
Resources: 0/1 (0.0%) | Time: 423.05s | Rate: 0.00/s
my_slow_api_resource: 34727  | Time: 0.00s | Rate: 2141994044.24/s

------------------------------- Extract rest_api -------------------------------
Resources: 1/1 (100.0%) | Time: 423.06s | Rate: 0.00/s
my_slow_api_resource: 34727  | Time: 0.01s | Rate: 2406218.01/s

```

This log is from a configured info level, and every 20s. Given this is an isolated run it's easy to see, but generally this resource is in a batch of 12 and the fact it takes so long is not obvious in the group as all of this vendors endpoints are not paginated. So each resource says it takes <1s. 

### Expected behavior

The resource should record the start of it's timing before it makes a single request, not from the end of the first request. 

### Steps to reproduce

Create a flask endpoint that sleeps 30s before returning a response but maintains the connection. Observe DLT log timing recorded <1s for the resource. 

### Operating system

macOS

### Runtime environment

Virtual Machine

### Python version

3.11

### dlt data source

RestAPI Source

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

ernestprovo23:
### the bug
for a rest_api (or any) resource whose request is slow or non paginated, the per
resource line in LogCollector reports a rate of millions/s. the total extract time is
correct, only the per resource rate is wrong.

cause: LogCollector creates a counter lazily on the first update for a given key and
sets its start_time at that moment. the per resource counter is keyed on the table name
and is only updated when the first rows arrive (extractors.py _write_item / _import_item).
for a slow single response that first update happens after the whole wait, so start_time
is set late, elapsed time is ~0, and count/elapsed blows up. the aggregate "Resources"
counter is registered at pipe loop start so its timer is correct.

### the fix
register each selected resource's progress counter with inc=0 at pipe loop start, right
where the "Resources" counter is registered. this sets start_time before the request goes
out, so elapsed time includes the wait and the rate is sane.

### the table vs resource keying
the per resource line is keyed on the normalized table name, not the resource name. three
cases:

- static single table resource (the reported case): the static table name is known up
  front and equals where rows land, so pre-registration uses the same key as the row
  update. fixed.
- dynamic table name (table_name callable): table is unknown until items arrive, so it is
  not pre-registered. nothing changes for these.
- per item redirect via dlt.mark.with_table_name: these have no dynamic table name flag,
  so they resolve to their static default name when pre-registered, but rows actually go
  elsewhere. to avoid a phantom "name: 0" line that never advances, the stale
  pre-registered counter is dropped on the first write that targets a different table.

### tests
- collector unit tests assert start_time is fixed at registration and elapsed reflects a
  later clock jump (and a control showing the buggy late start_time without registration).
- extract integration tests with an injected clock assert the static resource counter
  start_time is set before a simulated wait, and that neither the with_table_name redirect
  case nor the table_name callable case leaves a phantom counter line.

closes #3518

### DIFF
--- a/dlt/common/runtime/collector.py
+++ b/dlt/common/runtime/collector.py
@@ -174,6 +174,14 @@ class LogCollector(Collector):
             self.messages[counter_key] = message
         self.maybe_log()
 
+    def discard(self, name: str, label: str = None) -> None:
+        """Removes counter `name` (and its info and message) if it exists."""
+        counter_key = f"{name}_{label}" if label else name
+        if counter_key in self.counters:
+            del self.counters[counter_key]
+            del self.counter_info[counter_key]
+            del self.messages[counter_key]
+
     def maybe_log(self) -> None:
         """Check if should report and if so, call self.on_log"""
         current_time = self._clock()
--- a/dlt/common/runtime/collector_base.py
+++ b/dlt/common/runtime/collector_base.py
@@ -60,6 +60,10 @@ class Collector(ABC, SupportsTracking):
         """Stops counting. Should close all counters and release resources ie. screen or push the results to a server."""
         pass
 
+    def discard(self, name: str) -> None:
+        """Removes a counter `name` if it exists. No-op by default."""
+        pass
+
     def __call__(self: TCollector, step: str) -> TCollector:
         """Syntactic sugar for nicer context managers"""
         self.step = step
--- a/dlt/extract/extract.py
+++ b/dlt/extract/extract.py
@@ -522,6 +522,14 @@ class Extract(WithStepInfo[ExtractMetrics, ExtractInfo]):
                 ) as pipes:
                     left_gens = total_gens = len(pipes._sources)
                     collector.update("Resources", 0, total_gens)
+                    # register per resource progress counters before iterating so their
+                    # start_time includes the wait for the first item. without this a slow
+                    # or non paginated request reports an astronomical rate because the
+                    # counter is only created when the first rows arrive (#3518). the
+                    # collector is shared by all extractors and table name normalization is
+                    # identical across them, so any extractor registers the same counter key
+                    for resource in source.resources.selected.values():
+                        extractors["object"].start_resource(resource)
                     for pipe_item in pipes:
                         curr_gens = len(pipes._sources)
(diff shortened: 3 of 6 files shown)


rudolfix:
@ernestprovo23 there are 200+ files in your PR. did you resolve merge conflict? best if you start from fresh devel and cherry pick your work

ernestprovo23:
ah my bad, that's a base-branch artifact not the actual change. i opened it against `master` instead of `devel`, so it was diffing ~15 merges of drift. just retargeted base to `devel` and it collapses to the real diff: 1 commit, 6 files, the #3518 fix. no fresh branch or cherry-pick needed, it's already clean against devel (shows mergeable now). lmk if CI needs a workflow approval on your end to kick it off.

ernestprovo23:
@rudolfix did what you suggested, went back to fresh devel and re-applied the work. the extractor refactor to `tables_with_items` landed in between so a plain rebase wasn't going to be clean anyway.

6 files / 205 added, 0 deleted, mergeable again. the old branch had also quietly dropped an assertion in `test_extract.py`, that's gone.

ran it locally, `tests/extract` + `tests/common/runtime` green at 1200 pass, black/ruff/flake8 clean, mypy clean on the touched files. the two failures i hit (`test_source_reference_import_core` and the telemetry module) reproduce on stock devel without my changes.

btw there are two other open PRs on #3518 now, #4106 and #4121, if you'd rather take one of those.

PR #4043, opened 2026-06-10
Fix retry after zero

### LINKED ISSUE #4036: Rate Limits: "Retry-After: 0" causes immediate retry loop, exhausting all attempts without backoff
### dlt version

1.27.2

### Describe the problem

When an API returns **429 Too Many Requests** with `Retry-After: 0`, dlt retries immediately with 0s wait on every subsequent attempt. This exhausts `request_max_attempts` in rapid succession without ever giving the server's rate limit window time to clear.

**Root cause**
  
`wait_exponential_retry_after._parse_retry_after` in `dlt/sources/helpers/requests/retry.py`:
  
```
return max(self.min, min(self.max, seconds))
```

`wait_exponential` defaults `min=0`, so Retry-After: 0 → max(0, 0) → 0. Since `_get_retry_after` returns 0 (not None), `__call__` returns it directly and bypasses the exponential backoff entirely.

### Expected behavior

Retry-After: 0 means "you can retry now" per RFC, but in practice many APIs send it while still enforcing a limit window. When the header value is 0 (or negative for a past date), dlt should fall through to exponential backoff rather than retrying instantly.
  
**Fix:**

Treat seconds <= 0 as "no actionable hint" and return None, letting tenacity's exponential backoff take over:

```  
def _parse_retry_after(self, retry_after: str) -> Optional[float]:
      ...
      if seconds <= 0:
          return None  # fall through to exponential backoff
      return max(self.min, min(self.max, seconds))
```

### Steps to reproduce

Reproduced with CoinGecko's free public API.

```
import dlt
from dlt.sources.rest_api import rest_api_source


def print_headers(response, *args, **kwargs):
    print(
        f"[page={response.request.url.split('page=')[-1].split('&')[0]}] "
        f"status={response.status_code} "
        f"retry_after={response.headers.get('Retry-After')}"
    )
    return response


def load_coins() -> None:
    """Load coin market data from CoinGecko into DuckDB.

    Free public API — no auth needed.
    Rate limit: ~30 req/min → per_page=1 exhausts it quickly to trigger 429.
    """
    pipeline = dlt.pipeline(
        pipeline_name="coingecko",
        destination="duckdb",
        dataset_name="coingecko_data",
        dev_mode=True,
    )

    source = rest_api_source(
        {
            "client": {
                "base_url": "https://api.coingecko.com/api/v3/",
            },
            "resources": [
                {
                    "name": "coins_markets",
                    "endpoint": {
                        "path": "coins/markets",
                        "params": {
                            "vs_currency": "usd",
                            "per_page": 1,  # one coin per request → many pages → hits rate limit fast
                            "order": "market_cap_desc",
                        },
                        "paginator": {
                            "type": "page_number",
                            "page_param": "page",
                            "base_page": 1,
                            "total_path": None,
                            "maximum_page": 15000,
                            "stop_after_empty_page": True,
                        },
                        "response_actions": [print_headers],
                    },
                    "primary_key": "id",
                    "write_disposition": "replace",
                }
            ],
        }
    )

    print(pipeline.run(source))


if __name__ == "__main__":
    load_coins()
```
  

**Observed behavior**

With `request_max_attempts=10`, all 10 attempts fire back-to-back after a single 60s wait:
```
  [page=6] status=429 retry_after=60   ← attempt 1: waits 60s ✓
  [page=6] status=429 retry_after=0    ← attempt 2: waits 0s  ✗
  [page=6] status=429 retry_after=0    ← attempts 3–10: waits 0s ✗
  PipelineStepFailed
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

CoinGecko's free public API

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

This investigation was triggered by recurring user complaints about pipelines failing with 429 despite this implemented retry after logic. The pattern kept coming up but the root cause wasn't obvious which is what led to tracing it down to this specific edge case.

AstrakhantsevaAA:
### Description

When an API returns 429 Too Many Requests with Retry-After: 0, dlt retries immediately with 0s wait on every subsequent attempt. 

Fix:

Treat seconds <= 0 as "no actionable hint" and return None, letting tenacity's exponential backoff take over

### Related Issues

- Fixes #4036

### DIFF
--- a/tests/sources/helpers/test_requests.py
+++ b/tests/sources/helpers/test_requests.py
@@ -200,7 +200,7 @@ def test_wait_retry_after_int(mock_sleep: mock.MagicMock) -> None:
 
 @pytest.mark.parametrize(
     "retry_after_value",
-    ["0", "-1", "  0  "],
+    ["0", "Tue, 10 Jun 2025 00:00:00 GMT", "  0  "],
 )
 def test_wait_retry_after_zero_falls_through_to_backoff(
     mock_sleep: mock.MagicMock, retry_after_value: str
--- a/tests/sources/helpers/test_requests.py
+++ b/tests/sources/helpers/test_requests.py
@@ -198,6 +198,49 @@ def test_wait_retry_after_int(mock_sleep: mock.MagicMock) -> None:
     assert 4 <= mock_sleep.call_args[0][0] <= 5  # Adds jitter up to 1s
 
 
+@pytest.mark.parametrize(
+    "retry_after_value",
+    ["0", "-1", "  0  "],
+)
+def test_wait_retry_after_zero_falls_through_to_backoff(
+    mock_sleep: mock.MagicMock, retry_after_value: str
+) -> None:
+    """Retry-After: 0 (or negative) must fall through to exponential backoff, not retry instantly."""
+    session = Client(request_backoff_factor=1, request_max_attempts=3).session
+    url = "https://example.com/data"
+    m = requests_mock.Adapter()
+    session.mount("https://", m)
+    responses: List[Dict[str, Any]] = [
+        dict(text="error", headers={"retry-after": retry_after_value}, status_code=429),
+        dict(text="success"),
+    ]
+    m.register_uri("GET", url, responses)
+    session.get(url)
+
+    # exponential backoff must have fired — sleep called with a positive value
+    mock_sleep.assert_called_once()
+    assert mock_sleep.call_args[0][0] > 0
+
+
+def test_wait_retry_after_zero_then_nonzero_uses_header(mock_sleep: mock.MagicMock) -> None:
+    """After falling back to backoff on Retry-After: 0, a subsequent non-zero header is respected."""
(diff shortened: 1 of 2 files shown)


tests/sources/helpers/test_requests.py
```diff
@@ -198,6 +198,49 @@ def test_wait_retry_after_int(mock_sleep: mock.MagicMock) -> None:
     assert 4 <= mock_sleep.call_args[0][0] <= 5  # Adds jitter up to 1s
 
 
+@pytest.mark.parametrize(
+    "retry_after_value",
+    ["0", "-1", "  0  "],
```
burnash:
```suggestion
    ["0", "Tue, 10 Jun 2025 00:00:00 GMT", "  0  "],
```

I think "-1" above never reaches the seconds - the regex on line 84 excludes it: `re.match(r'^\s*[0-9]+\s*$', '-1')` will produce None. It's best to replace it with just a past date - then it'll enter the negative seconds branch

burnash (approved):
Looks good!

PR #4044, opened 2026-06-10
(feat) prunes deps from launcher groups if present in user deps

rudolfix:
### Description
* duplicate deps are eliminated
* numpy and pandas got removed from rows to arrow conversion and from dashboard deps
* row conversion above uses fast-path (pure arrow) that could be to 2x faster or Python zip that is as good as pandas


tetelio (dismissed):
Very good job @rudolfix , especially trimming down deps in dashboard

PR #4046, opened 2026-06-10
retry flaky databricks and motherduck remote tests

burnash:
These two destinations fail every now and then transient and unrelated to the PR being tested. databricks throws ConnectionResetError/RequestError on connect, motherduck crashes the xdist worker (segfault in the duckdb extension).

Adds pytest-rerunfailures and opts in just those two (for now) via rerun_args in the matrix, scoped with --only-rerun so genuine failures still fail fast (motherduck uses a "sentinel" since only the crash needs retrying).

We can also extend this to other remote tests - but I'd keep it scoped


tetelio (approved):
LGTM, thanks!

rudolfix (approved):
this is great!

PR #4049, opened 2026-06-11
Feat/add cdn marimo launcher

tetelio:
## Summary

- Serve marimo frontend assets from jsDelivr CDN via `--asset-url` (configurable, empty string disables).
- Fix launcher path resolution: use `find_spec` instead of `import_module` so marimo/streamlit notebooks aren't executed before `marimo run` / `streamlit run` (~1–1.5s faster port readiness).


rudolfix (approved):
LGTM!

PR #4054, opened 2026-06-12
fix: propagate _dlt_root_id to child tables for scd2 merge strategy

### LINKED ISSUE #4047: `scd2` merge strategy missing from `requires_root_key()` strategy whitelist — child tables silently lose `_dlt_root_id`
### dlt version

1.27.2

### Describe the problem

Switching a resource's `write_disposition` from `"merge"` (default `delete-insert` strategy) to `{"disposition": "merge", "strategy": "scd2"}` silently disables `_dlt_root_id` propagation to nested child tables.

On a fresh dataset the load **succeeds**, but the child tables are created **without** `_dlt_root_id` at all — so nested rows cannot be linked back to a specific parent version, which defeats the point of SCD2 for nested data. On a dataset that previously loaded with plain `merge` (where the child schema already has `_dlt_root_id REQUIRED`), the scd2 load **fails**, because the new child JSONL no longer contains the column.

### Root cause

`requires_root_key()` in `dlt/common/normalizers/json/helpers.py` auto-enables root-key propagation only for these merge strategies:

```python
merge_requires = (
    merge_strategy in ["delete-insert", "upsert", "insert-only"]
    if root_key_propagation is None
    else root_key_propagation
)
```

`"scd2"` is missing from the list, so the auto-detection that fires for plain `merge` does not fire for scd2 — even though scd2 semantically needs the same lineage column on child tables.

### Expected behavior

`"scd2"` should be included in the strategy list (or the check should key off the `merge` disposition rather than the strategy), so that switching a nested resource to scd2 keeps `_dlt_root_id` on child tables without the user having to know to also set `root_key=True` on the source.

### Steps to reproduce

Verified on dlt 1.27.2 against BigQuery:

```python
import dlt

@dlt.source(name="repro")  # note: no root_key=True
def repro():
    @dlt.resource(
        name="parent",
        primary_key="id",
    )
    def parent():
        yield {"id": 1, "name": "foo", "children": [{"k": "v1"}]}
        yield {"id": 2, "name": "bar", "children": [{"k": "v2"}]}
    return parent

p = dlt.pipeline(pipeline_name="repro_scd2", destination="bigquery", dataset_name="dlt_repro")
p.run(repro())
```

Observed: load succeeds; `parent__children` schema is `{k, _dlt_parent_id, _dlt_list_idx, _dlt_id}` — **no** `_dlt_root_id`.

Expected: `parent__children` includes `_dlt_root_id`, as it does for `delete-insert`/`upsert` merges.

### Workaround (and why it is hazardous)

`@dlt.source(root_key=True)` enables propagation, but:

1. It is source-wide, so **every** resource in the source — including `replace`/`append` ones — gets `_dlt_root_id` required on its child tables. There is no per-resource control (`root_key_propagation` in the internal signatures is plumbing for the same source-wide setting, not a per-resource override).
2. If any scd2 data was already loaded while this bug was active, applying the workaround afterwards fails hard on BigQuery: dlt emits `ALTER TABLE ... ADD COLUMN _dlt_root_id ... NOT NULL` against the now-populated child table, and BigQuery rejects it (`Cannot add required fields to an existing schema`). Filed separately as dlt-hub/dlt#4048 (same DDL failure class as dlt-hub/dlt#4041).

### Suggested fix

Add `"scd2"` to the strategy list in `requires_root_key()`.

### Operating system

macOS

### Python version

3.14

### Destination

Google BigQuery

### Additional information

Related: dlt-hub/dlt#3811 (its reproducer used `@dlt.source(root_key=True)` without explaining why — that is exactly the workaround for this bug).

AnnasMazhar:
requires_root_key() had a whitelist of strategies that auto-enable root key propagation: delete-insert, upsert, insert-only. scd2 was missing, so nested tables silently lost the _dlt_root_id column — breaking parent-version linkage.

Added "scd2" to the whitelist. Updated the corresponding test expectation.

Fixes #4047

### DIFF
--- a/dlt/common/normalizers/json/helpers.py
+++ b/dlt/common/normalizers/json/helpers.py
@@ -138,7 +138,7 @@ def requires_root_key(
     """Checks if table chain containing `table` requires root_key to propagate.
 
     1. if there's any table with `root_key` in table chain already - we always propagate
-    2. if write disposition is merge and merge strategy is "delete-insert", "upsert" and root_key_propagation is not False
+    2. if write disposition is merge and merge strategy is "delete-insert", "upsert", "scd2" and root_key_propagation is not False
     3. root_key_propagation is True
     """
     table_name = root_table["name"]
@@ -148,7 +148,7 @@ def requires_root_key(
     else:
         merge_strategy = resolve_merge_strategy(schema.tables, root_table)
         merge_requires = (
-            merge_strategy in ["delete-insert", "upsert", "insert-only"]
+            merge_strategy in ["delete-insert", "upsert", "insert-only", "scd2"]
             if root_key_propagation is None
             else root_key_propagation
         )
--- a/tests/common/normalizers/test_json_relational.py
+++ b/tests/common/normalizers/test_json_relational.py
@@ -980,7 +980,7 @@ def test_caching_perf(norm: RelationalNormalizer) -> None:
         ("delete-insert", None, False, True),
         ("upsert", None, False, True),
         ("insert-only", None, False, True),
-        ("scd2", None, False, False),
+        ("scd2", None, False, True),
         ("append", None, False, False),
         ("replace", None, False, False),
         # Test with root_key_propagation explicitly True (should always be True)
@@ -1016,7 +1016,7 @@ def test_caching_perf(norm: RelationalNormalizer) -> None:
         "delete-insert_default_no-nested_requires-key",
         "upsert_default_no-nested_requires-key",
         "insert-only_default_no-nested_requires-key",
-        "scd2_default_no-nested_no-key-required",
+        "scd2_default_no-nested_requires-key",
         "append_default_no-nested_no-key-required",
         "replace_default_no-nested_no-key-required",
         # Names for explicit True tests


AnnasMazhar:
👋 friendly ping — this is a small fix (+4/-4) that adds scd2 to the requires_root_key() strategy whitelist, fixing issue #4047. Would appreciate a review when anyone has a moment. Thanks!

AnnasMazhar:
👋 gentle bump — this is a small +4/-4 fix that adds scd2 to the `requires_root_key()` whitelist. Would appreciate a review when someone has a moment. Thanks!

AnnasMazhar:
👋 **Third bump** — this 4-line fix (root `_dlt_root_id` propagation for SCD2 merge) has been open since 23 Jun with no reviewer engagement. The fix is minimal (+4/-4) and the related issue has been open even longer. Would appreciate a review when someone has a moment. Happy to address any feedback.

AnnasMazhar:
Closing due to maintainer inactivity after 32 days and 3 bump comments with no reviewer engagement. This small +4/-4 SCD2 fix is available for anyone who needs it. Happy to reopen if there's interest.

AnnasMazhar:
Closing due to maintainer inactivity. The fix is minimal (+4/-4) and available for anyone who needs it. Happy to reopen if there's interest.

PR #4057, opened 2026-06-14
fix(sql_database): warn when type_adapter_callback returns None and clarify docs

### LINKED ISSUE #4032: Consecutive batches become incompatible and silently break the pipeline when `type_adapter_callback` returns `None`
### dlt version

1.26.0 (also reproduces on 1.27.2 and `devel`).

### Describe the problem

The `sql_database` source's `type_adapter_callback` is designed for
selective type overrides. The natural pattern is:

```python
def adapt(sql_type):
    if needs_override(sql_type):
        return SomeNewType()
    return None   # keep the reflected type
```

But in
[`dlt/sources/sql_database/schema_types.py`](https://github.com/dlt-hub/dlt/blob/devel/dlt/sources/sql_database/schema_types.py)
in `sqla_col_to_column_schema`:

```python
if type_adapter_callback:
    sql_t = type_adapter_callback(sql_t)
    ...
if sql_t is None:
    # Column ignored by callback
    return col
```

Returning `None` makes the function exit before setting `data_type` on
the column. So every column that the callback "passes through" ends
up with a dlt column schema of just `{"name": ..., "nullable": ...}`
— no `data_type`, no precision, no scale.

Downstream, `convert_numpy_to_arrow` sees `data_type=None`, falls back
to `pa.array(values, type=None)`, and **pyarrow infers the decimal
precision from each batch's values**. For a `NUMERIC(7,2)` column where
batch 1 happens to contain only small values and batch 2 contains
larger values, the two batches get incompatible `decimal128(P, S)`
types. `ArrowToParquetWriter` opens the file with batch 1's schema and
the next batch fails with:

```
ValueError: Table schema does not match schema used to create file
```

The extract terminates mid-run.

### Expected behavior

`type_adapter_callback` returning `None` should mean "no override, use
the reflected type". Either:

1. Read `None` as "keep `sql_t` unchanged" and continue the reflection
   path that fills in `data_type`, precision and scale.
2. Or update the docstring to make absolutely explicit that `None`
   discards the type information and users must `return sql_type` for
   the "no change" case.

Today the API silently turns a documented "selective override" pattern
into "drop type info on every column the callback didn't touch", which
is a footgun.

### Steps to reproduce

Empty Postgres on `localhost:55432`, e.g.:

```bash
docker run --rm -d -p 55432:5432 \
  -e POSTGRES_DB=repro -e POSTGRES_USER=repro -e POSTGRES_PASSWORD=repro \
  postgres:17-alpine
```

`repro.py`:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "dlt[duckdb,postgres,parquet,sql_database]==1.26.0",
#   "psycopg2-binary", "numpy", "pandas",
# ]
# ///
import os
os.environ["DATA_WRITER__BUFFER_MAX_ITEMS"] = "1"

from sqlalchemy import Text
from sqlalchemy.sql.sqltypes import Numeric
import dlt, psycopg2
from dlt.sources.sql_database import sql_database

PG_URL = "postgresql://repro:repro@localhost:55432/repro"

def adapt(sql_type):
    # Override one specific case, pass through everything else.
    if isinstance(sql_type, Numeric) and sql_type.precision is None:
        return Text()
    return None  # "no change" — but dlt reads this as "drop the type"

c = psycopg2.connect(PG_URL); c.autocommit = True; cur = c.cursor()
cur.execute("DROP TABLE IF EXISTS t")
cur.execute("CREATE TABLE t (id INT PRIMARY KEY, price NUMERIC(7,2) NOT NULL)")
cur.executemany("INSERT INTO t VALUES (%s, %s)", [(i, i/100) for i in range(1, 100)])
cur.executemany("INSERT INTO t VALUES (%s, %s)", [(i, i)     for i in range(100, 200)])
c.close()

pipe = dlt.pipeline(pipeline_name="mwe",
                    destination=dlt.destinations.duckdb("/tmp/mwe.duckdb"))
pipe.run(sql_database(
    credentials=PG_URL, table_names=["t"],
    backend="pyarrow", chunk_size=99,
    type_adapter_callback=adapt,
))
```

Output:

```
ValueError: Table schema does not match schema used to create file: 
table:
id: int64 not null
price: decimal128(5, 2) not null vs. 
file:
id: int64 not null
price: decimal128(2, 2) not null
```

Returning `sql_type` from the callback instead of `None` (i.e. the
"no override" path explicitly returns the original type) makes the
extract succeed and the destination column is `DECIMAL(7, 2)` as
declared by Postgres.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

`dlt.sources.sql_database` against Postgres 17 with `backend="pyarrow"`
and a `type_adapter_callback`.

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

Hit in production on `NUMERIC(7,2)` and `INT` columns where the
callback was designed to coerce unbounded `NUMERIC` and `JSONB` to
`TEXT` and pass everything else through with `return None`. The pattern
matches the dlt docs example for `type_adapter_callback`. Workaround
in our codebase was a one-line change:

```python
- return None
+ return sql_type
```

Either docs change or behaviour change would prevent the next user
from rediscovering this.

DRACULA1729:
Closes #4032.

When `type_adapter_callback` returns `None`, dlt drops the reflected `data_type` for that column. That's documented behaviour ("infer the type from data"), but it bites under `backend="pyarrow"`/`"pandas"`: pyarrow then infers the type per batch, so a `NUMERIC(7,2)` column comes out as `decimal128(5,2)` in one batch and `decimal128(2,2)` in another. `ArrowToParquetWriter` opens the file with the first batch's schema, the next batch doesn't match, and the extract dies mid-run with:

```
ValueError: Table schema does not match schema used to create file
```

The catch is that the natural "no change" pattern looks like `return None`, when it actually has to be `return sql_type`.

### What I did

I kept the `None` behaviour rather than flipping it to mean "keep the reflected type". Returning `None` to force data inference is intentional and documented (it's the escape hatch for custom dialect types that don't map onto standard SQLAlchemy types), so changing it would be a breaking change. Instead this makes the footgun loud:

- Log a warning when the callback returns `None` and the reflected type is dropped, naming the column and explaining the per-batch inference risk.
- Clarify the `sql_database`/`sql_table` docstrings and the `advanced.md` docs: return the input unchanged for the "no change" case, `None` discards the type.
- Tests in `tests/sources/sql_database/test_schema_types.py` for both the warning path and the pass-through (`return sql_type`) path.

### Out of scope

This doesn't touch the deeper behaviour where any `data_type=None` column infers per batch (that also hits unknown reflected types, not just the callback). Happy to follow up on that separately if you'd like it addressed.

### DIFF
--- a/dlt/sources/sql_database/__init__.py
+++ b/dlt/sources/sql_database/__init__.py
@@ -94,7 +94,12 @@ def sql_database(
         include_views (bool): Reflect views as well as tables. Note view names included in `table_names` are always included regardless of this setting.
 
         type_adapter_callback (Optional[TTypeAdapter]): Callable to override type inference when reflecting columns.
-            Argument is a single sqlalchemy data type (`TypeEngine` instance) and it should return another sqlalchemy data type, or `None` (type will be inferred from data)
+            Argument is a single sqlalchemy data type (`TypeEngine` instance). Return another sqlalchemy
+            data type to override it, return the argument unchanged to keep the reflected type, or return
+            `None` to drop the reflected type and infer the data type from the data.
+            Warning: for a selective override, return the unchanged argument for types you do not adapt.
+            Returning `None` discards the reflected type, and with the `pyarrow`/`pandas` backends the
+            inferred type may differ across batches (e.g. decimal precision), which can fail the load.
 
         query_adapter_callback (Optional[TQueryAdapter]): Callable to override the SELECT query used to fetch data from the table.
             The callback receives the sqlalchemy `Select` and corresponding `Table`, 'Incremental` and `Engine` objects and should return the modified `Select` or `Text`.
@@ -248,7 +253,12 @@ def sql_table(
         backend_kwargs (Dict[str, Any], optional): kwargs passed to table backend ie. "conn" is used to pass specialized connection string to connectorx.
 
         type_adapter_callback (Optional[TTypeAdapter]): Callable to override type inference when reflecting columns.
-            Argument is a single sqlalchemy data type (`TypeEngine` instance) and it should return another sqlalchemy data type, or `None` (type will be inferred from data)
+            Argument is a single sqlalchemy data type (`TypeEngine` instance). Return another sqlalchemy
+            data type to override it, return the argument unchanged to keep the reflected type, or return
+            `None` to drop the reflected type and infer the data type from the data.
+            Warning: for a selective override, return the unchanged argument for types you do not adapt.
+            Returning `None` discards the reflected type, and with the `pyarrow`/`pandas` backends the
+            inferred type may differ across batches (e.g. decimal precision), which can fail the load.
 
         included_columns (Optional[List[str]]): List of column names to select from the table. If not provided, all columns are loaded.
 
--- a/dlt/sources/sql_database/schema_types.py
+++ b/dlt/sources/sql_database/schema_types.py
@@ -123,7 +123,18 @@ def sqla_col_to_column_schema(
             sql_t = sql_t()
 
     if sql_t is None:
-        # Column ignored by callback
+        # `type_adapter_callback` returned None: drop the reflected type and let the
+        # backend infer the data type from the data (documented behavior). Warn because
+        # inference is per-batch for the `pyarrow`/`pandas` backends, so a column like
+        # NUMERIC(7, 2) can yield incompatible decimal types across batches and fail the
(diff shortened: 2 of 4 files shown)


DRACULA1729:
@rudolfix @sh-rp when you get a chance, could one of you take a look? You know this corner of `sql_database` best. No rush.

PR #4059, opened 2026-06-15
fix(cli): route dlthub handoff NOTE to stderr so piping isn't broken

### LINKED ISSUE #4035: bug(cli): `dlthub` message breaks piping
The following command outputs the schema in the specified format

```shell
dlt local schema --format FORMAT SCHEMA_FILE_PATH
```

However, the latest changes added a message that is printed whenever `dlt` is called instead of `dlthub` 

```shell
NOTE: Please use dlthub as top level command. Check `dlthub local --help` for former dlt commands. Falling back to dlthub command set.
```

This breaks the ability to pipe commands and push values to a file. For example, this will produce invalid JSON

Input: user command
```shell
dlt local schema --format json SCHEMA_FILE_PATH > my_schema.json
```

Output: content of `my_schema.json` is invalid json because it captures `NOTE: ` message
```shell
NOTE: Please use dlthub as top level command. Check `dlthub local --help` for former dlt commands. Falling back to dlthub command set.

{
  "version": 2,
  "version_hash": "bsxTAxmbRGV29fefv3NIfsTHr8meNMN4j8wAcwFmCVQ=",
  "engine_version": 11,
  "name": "sql_database",
  ...
}
```

yashs33244:
Closes #4035.

## Summary

When a dlt workspace is active, every `dlt ...` command printed a green `NOTE: Please use dlthub ...` banner to **stdout** before delegating to the dlthub command set — corrupting piped output (e.g. `dlt local schema --format json SCHEMA > out.json` wrote the NOTE into the JSON, breaking it).

This routes the banner to **stderr**: `fmt.note()` gains an opt-in `err: bool = False` (default preserves every existing caller, including the deploy-NOTE test that asserts on stdout), and the single handoff call site in `_main()` passes `err=True`.

## Tests

- Regression test red-green proven: fails without the fix (NOTE on stdout, stderr empty), passes with it
- `black --check`, `ruff check`, `flake8`, `mypy` green on the 3 touched files

### DIFF
--- a/dlt/_workspace/cli/_dlt.py
+++ b/dlt/_workspace/cli/_dlt.py
@@ -375,12 +375,15 @@ def _main() -> None:
     # when workspace is active, dlt commands mirrors dlthub
     if is_workspace_active():
         host = "dlthub"
+        # emit the handoff banner on stderr so it never corrupts piped stdout (e.g. `dlt local
+        # schema --format json ... > file.json`)
         fmt.note(
             "Please use %s as top level command. Check `%s` for former dlt commands. "
             "Falling back to dlthub command set."
-            % (fmt.bold("dlthub"), fmt.bold("dlthub local --help"))
+            % (fmt.bold("dlthub"), fmt.bold("dlthub local --help")),
+            err=True,
         )
-        fmt.echo()
+        fmt.echo(err=True)
     else:
         host = "dlt"
     exit(main(host))
--- a/dlt/_workspace/cli/echo.py
+++ b/dlt/_workspace/cli/echo.py
@@ -131,8 +131,8 @@ def warning(msg: str) -> None:
     click.secho("WARNING: " + msg, fg="yellow")
 
 
-def note(msg: str) -> None:
-    click.secho("NOTE: " + msg, fg="green")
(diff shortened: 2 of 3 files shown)


rudolfix:
@yashs33244 this definitely makes sense! however we are considering deactivating `dlt` command entirely when `dlthub` command is active - just providing relevant help. details: https://github.com/dlt-hub/dlt/pull/4037

do you see any value in keeping `dlt` as an alternative command host for `dlthub`? they are doing the same thing when workspace is activated

yashs33244:
Good question. I'd lean toward keeping `dlt` as a thin alias for backward-compat - existing scripts, CI jobs, and docs call `dlt ...` directly, so a hard deactivation could break people mid-workflow - but that's your call and the deprecation path in #4037 makes sense. Either way this fix is low-risk and helps today: until `dlt` is fully retired, the handoff NOTE on stdout corrupts piped output (e.g. `dlt ... schema --format json > schema.json` writes the banner into the file). Happy to close this if you're deactivating `dlt` imminently, or keep it as a safety net in the meantime - whichever you prefer.

PR #4061, opened 2026-06-15
fix(common): use ISO week/year in detect_datetime_format so week-date cursors round-trip

yashs33244:
`detect_datetime_format` (`dlt/common/time.py`) mapped ISO 8601 week dates (`YYYY-Www`) to `"%Y-W%W"` / `"%Y-W%W-%u"`. `%W` (locale week-of-year, tied to the calendar year `%Y`) disagrees with the ISO week `%V` + ISO week-numbering year `%G` at year boundaries.

The detected format is used by `dlt/extract/incremental/lag.py` to re-serialize a parsed-and-lagged string cursor, so it must round-trip — and it did not: `"2026-W01"` parsed to 2025-12-29 and rendered back as `"2025-W52"`, silently corrupting the saved incremental cursor at year boundaries.

Fix: emit `"%G-W%V"` / `"%G-W%V-%u"`.

Red-green proven: 9 tests fail without the fix. An exhaustive 2015–2030 round-trip scan shows 0 regressions and 303 previously-broken week values fixed; out-of-range weeks (e.g. W53 in a 52-week year) are still rejected at parse time. `tests/common/test_time.py` 361 passed, `tests/extract/test_lag.py` 426 passed. `black --check`, `ruff check`, `flake8`, and `mypy` are clean on the touched files.

### DIFF
--- a/dlt/common/time.py
+++ b/dlt/common/time.py
@@ -383,9 +383,9 @@ def detect_datetime_format(value: str) -> Optional[str]:
         re.compile(r"^\d{4}-\d{2}-\d{2}$"): "%Y-%m-%d",  # Date only
         re.compile(r"^\d{4}-\d{2}$"): "%Y-%m",  # Year and month
         re.compile(r"^\d{4}$"): "%Y",  # Year only
-        # Week-based date formats
-        re.compile(r"^\d{4}-W\d{2}$"): "%Y-W%W",  # Week-based date
-        re.compile(r"^\d{4}-W\d{2}-\d{1}$"): "%Y-W%W-%u",  # Week-based date with day
+        # Week-based date formats (ISO 8601: week-numbering year %G + ISO week %V)
+        re.compile(r"^\d{4}-W\d{2}$"): "%G-W%V",  # Week-based date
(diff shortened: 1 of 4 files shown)


rudolfix (dismissed):
LGTM!

yashs33244:
The failing checks look unrelated to this change, which only touches `dlt/common/time.py` (the ISO-week format strings) plus `tests/common/test_time.py` and `tests/extract/test_lag.py`:

- **Windows `common tests` (3.11 / 3.13)** — the only failures are in `tests/common/storages/test_local_filesystem.py::test_local_path_win_configuration` (Windows `file://` URL encoding), e.g. `assert 'file://local…' == 'file://local…'` on visually-identical strings. That suite has nothing to do with datetime-format detection, and the time/lag tests added here pass on Windows (the job reports `3 failed, 1668 passed`, all 3 being those filesystem tests). The `common | common tests` check is just the matrix-rollup gate failing because of that Windows leg.
- **`airflow 3 tests`** — a `dlt_smoke` DAG run (`RuntimeError: DAG run failed`); the smoke source doesn't exercise ISO-week cursors, so it's unaffected by this patch.
- **Linux (3.10–3.14) and macOS** pass fully, including the new tests.

I've rebased onto the latest `devel` for a clean run. Happy to dig in further if any of these actually trace back to the change.

PR #4068, opened 2026-06-16
feat(dashboard): show dataset browser by default in the dltHub dashboard

lis365b:
### Description
Removes dashboard entry page and opens the dashboard directly on the dataset browser without toggle. 

* On open it auto-selects the most recently used pipeline
* * moves the dataset browser to the top, and removes its toggle so it shows by default
* makes the empty-state copy more generic to read correctly both locally and in the dltHub runtime 

### Related Issues

- Resolves [dlt-hub/product_backlog#153](https://github.com/dlt-hub/product_backlog/issues/153)

### Additional Context

The runtime's default dashboard is this same module, so the change ships there on the next dlt bump. No new dependencies; docs updated.


dlt/_workspace/helpers/dashboard/utils/pipeline.py
anuunchin, line 1: Considering `pipeline_link_list` was removed, we should probably remove it from line 18 of `tests/workspace/helpers/dashboard/test_all_cells.py`

dlt/_workspace/helpers/dashboard/utils/home.py
```diff
@@ -73,33 +71,18 @@ def home_header_row(
     )
 
 
-def render_workspace_home(
+def render_no_pipelines_home(
```
anuunchin, line 72: I'm thinking whether this should have a refresh button (or some smarter solution), because when a user opens the dashboard, then runs a pipeline, the only way to get it in the dashboard is to refresh the browser -> on the platform, it may take a bit to refresh the entire thing

anuunchin: Image

This is a nit: maybe the Pipeline selection option should be hidden here, since it has nothing anyway

dlt/_workspace/helpers/dashboard/strings.py
```diff
@@ -44,70 +38,33 @@ class TSectionStrings(NamedTuple):
...
-3. Ensure you have set the correct pipelines directory (using the `pipelines_dir` CLI argument).
-4. Restore a pipeline with this name from a destination using [`dlthub local pipeline sync <pipeline_name>`]({_sync_help_url}).
+2. Run a pipeline with this name, then refresh.
+3. Ensure the correct pipelines directory is configured.
+4. Restore a pipeline with this name from its destination (see the [sync docs]({_sync_help_url})).
 
-This page will automatically refresh with your pipeline data once you have run a pipeline with this name on this machine.
+The page updates with your pipeline data once a pipeline with this name has run.
 
 """
```
anuunchin: This text seems to be unreachable: When i try to open the dashboard with a non-existent pipeline, it gives me: 

Image

anuunchin: This appears to be a pre-existing issue, though

tests/e2e/helpers/dashboard/test_e2e.py
anuunchin, line 1: A test for render_no_pipelines_home would be great since it's a new function (it should probably start the dashboard against an empty pipelines dir)

anuunchin: A test for auto-selecting most-recent pipeline would be useful as well, along with a test that checks whether the dataset browser is open by default without any user interaction

anuunchin (changes requested):
Requesting changes because I think we need more tests, the other comments are primarily suggestions

lis365b:
thanks Anuun, addressed your concerns (replied inline)

dlt/_workspace/helpers/dashboard/utils/pipeline.py
lis365b: Done — removed the stale `dlt_pipeline_link_list` entry from `test_all_cells.py`.

dlt/_workspace/helpers/dashboard/utils/home.py
lis365b: I tried it, but re-running pipeline discovery on refresh reset the selected pipeline back to the most recent one, so I dropped it for now. To pick up a brand-new pipeline you reload the dashboard (the pre-existing behavior). Happy to revisit a smarter version separately.

lis365b: about the nit: Done — hid the empty pipeline dropdown on the no-pipelines page.

dlt/_workspace/helpers/dashboard/strings.py
lis365b: Agreed — removed the unreachable `app_pipeline_not_found` message and its dead render branch (a non-existent pipeline hits `home_error_attach_pipeline` instead).

tests/e2e/helpers/dashboard/test_e2e.py
lis365b: Added `test_no_pipelines_home`, which starts the dashboard against an empty pipelines dir.

lis365b: Added `test_auto_select_most_recent_pipeline` and `test_dataset_browser_shown_by_default`.

anuunchin (approved):
✨

PR #4069, opened 2026-06-16
(fix) install latest airflow dependencies with airflow constraints file declared

Travior:
Adds a small script that fetches the latest airflow version and builds the correct constraint file for it. Airflow suggests using the constrained install when running it as an application (e.g. to run DAGs) which we do in the smoke tests


rudolfix (approved):
LGTM! but really - airflow just should install itself correctly in the first place...

burnash (approved):

PR #4072, opened 2026-06-16
fix(cli): cap codex skill descriptions at 1024 chars on ai init

lis365b:
### Description

* Codex enforces a **1024-character limit on a skill's `description` frontmatter** (`codex_core` `validate_len` uses `value.chars().count() > 1024`) and silently drops any skill that exceeds it at session load — emitting only a stderr `ERROR`. 
* The (always-loaded) `dlthub-router` skill from the dlthub AI workbench has a description >1024 chars, so `dlthub ai init --agent codex` installed a `SKILL.md` that Codex then drops 
* dlt copies workbench skills verbatim,

Changes:
The Codex install path now truncates the installed`SKILL.md` description to the cap (and warns when it does)

### Related Issues
Fixes dlt-hub/dlthub-ai-workbench-internal#75

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,13 +93,13 @@ PYTEST_XDIST_DIST  ?= worksteal
 PYTEST_TARGET_ARGS :=
 
 # Internal marker model
-PARALLEL_MARKER_EXPR = (not serial and not forked)
-SERIAL_MARKER_EXPR   = (serial or forked)
+PARALLEL_MARKER_EXPR = (not serial and not forked and not rfam)
+SERIAL_MARKER_EXPR   = (serial or forked) and not rfam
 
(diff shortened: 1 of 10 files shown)


lis365b:
end to end validations run (AI workbench and the `codex` CLI):
- `dlthub ai init --agent codex` installs `dlthub-router` with description capped to exactly 1024 chars
- `codex exec ... --skip-git-repo-check` no longer logs `invalid description: exceeds maximum length of 1024 characters` and the skill loads
- Counterfactual confirmed: an un-truncated (1124-char) description reproduces the exact error (shows that the cap prevents it)

rudolfix (approved):
LGTM!

PR #4074, opened 2026-06-17
Remove support for python 3.9

Travior:
This PR removes support for Python 3.9 as it is EOL since 31.10.2025.

We have since stopped testing against that version in CI and should not advertise it as a supported version anymore. This will also allow us to further bump dependencies that have dropped python 3.9 without needing guards against the python version.

This PR also adjusts the target version for `ruff` and `mypy`, in turn I have fixed newly marked lint errors around `NewType` assignments in the typing tests. This PR does not aim to upgrade away from older typing patterns that weren't supported on Python 3.9


rudolfix (approved):
LGTM! thx for the cleanup

Issue #4052, opened 2026-06-12
Duplicates in incremental load with `write_disposition='append'` after failing to update `_dlt_pipeline_state` table


willi-mueller:
### dlt version

1.27.2

### Describe the problem

When a pipeline fails to update the table ` _dlt_pipeline_state`, it raises `dlt.load.exceptions.LoadClientJobFailed`. The next run introduces duplicate data when the resource is incremental and configured with `write_disposition='append'`. 

I wonder whether this is a regression introduced by https://github.com/dlt-hub/dlt/pull/3877 because in dlt v1.18 I've never seen this defect.

The resource loaded in such a pipeline is configured as follows:

```py
 @dlt.resource(
        write_disposition="append",
        primary_key="Id",
        columns={...},
    )
    def case_history(
        last_timestamp: incremental[str] = dlt.sources.incremental(
            "CreatedDate", initial_value=None
        ),
    ) -> Iterable[TDataItem]:
        yield get_records(...)
```

The pipeline is as follows:
```py

    dlt_tasks = PipelineTasksGroup("salesforce_pipeline")

    pipeline = dlt.pipeline(
        pipeline_name="salesforce",
        destination="bigquery",
        dataset_name="landing_salesforce",
        staging="filesystem",
    )

    source = my_salesforce.my_salesforce_source()
    dlt_tasks.add_run(
        pipeline,
        source.with_resources("case_history"),
        decompose="parallel-isolated",
        trigger_rule="all_done",
        provide_context=True,
    )
```

### Expected behavior

An incremental resource should not append previously appended data.

### Steps to reproduce

```py
import dlt
from typing import Iterable
from dlt.sources import DltResource
from dlt.sources import incremental
from dlt.common.typing import TDataItem


@dlt.source(
    schema_contract={"tables": "evolve", "columns": "freeze", "data_type": "freeze"},
)
def a_source() -> Iterable[DltResource]:
    @dlt.resource(write_disposition="append", primary_key="Id")
    def a_resource(
        last_timestamp: incremental[str] = dlt.sources.incremental(
            "SystemModstamp", initial_value=None
        ),
    ) -> Iterable[TDataItem]:
        yield _get_records()

    return (a_resource,)


def _get_records() -> Iterable[TDataItem]:
    """This mimics the implementation of get_records() from the verified Salesforce dlt source."""
    yield from [
        {
            "Id": "0120Y000000lCpiQAE",
            "Name": "FAQ",
[...]

PR #4079, opened 2026-06-17
fix(airflow): auto-disable staging truncation for parallel runs (#4052)

DRACULA1729:
### Description

When an Airflow DAG runs a dlt source with `decompose="parallel"` or `decompose="parallel-isolated"`, each resource gets its own pipeline but they all share one staging area (the staging destination and dataset name). One task can truncate staging files that another task is still loading. That aborts the package, and with `write_disposition="append"` it surfaces as duplicate rows on the next run once the stale state is restored from the destination. Full root cause is in #4052.

Following @rudolfix's suggestion in the issue, `PipelineTasksGroup` now disables staging-destination truncation automatically for parallel and parallel-isolated runs:

- New `truncate_staging_destination` argument on `PipelineTasksGroup`.
- `None` (default): auto. Disables truncation for `parallel` / `parallel-isolated`, leaves it untouched for single or serialized runs.
- `True` / `False`: forces truncation on or off for every task in the group.

Under the hood this sets `destination.truncate_tables_on_staging_destination_before_load` for the group's tasks. Athena is unaffected since it does not truncate the staging area.

### Related Issues

- Relates to #4052

### Additional Context

- Tests in `tests/helpers/airflow_tests/test_airflow_wrapper.py` cover the auto behavior across decompositions and the forced override. They check the resolved setting bound on each task, so they run without a live destination.
- Verified locally against Airflow 2.11: the new tests pass, the existing decompose/parallel tests still pass, and ruff/black/flake8/mypy are clean.
- Added a note to the Airflow Composer deploy page describing the behavior.

Note: this branch is based on a slightly older `devel` (my fork was behind and the gh token lacks the `workflow` scope to carry the upstream workflow-file commits). Happy to rebase onto the latest `devel` if you prefer.


DRACULA1729:
@rudolfix this is ready for your review whenever you get a chance. Tests pass locally against Airflow 2.11 and the existing decompose tests are still green.

DRACULA1729:
A couple of things I would like your read on before this is final, @rudolfix:

1. Process-global config leak. `_run` writes `dlt.config["destination.truncate_tables_on_staging_destination_before_load"]` only when the resolved value is not `None`. That write is process-global, so on a worker reused across task instances (Celery with process reuse) a prior `parallel` task that set `False` can leak into a later `none`/`serialize` task, which then skips truncation it would otherwise do. LocalExecutor forks a child per task, so it is not affected. The existing `raise_on_failed_jobs` write has the same global-state shape but always writes, so it never goes stale; my conditional skip is the new wrinkle. Do you prefer I capture-and-restore the value around the run, or always write a definitive value? The latter could override a user's config.toml setting for non-parallel runs, so I lean towards restore.

2. Test depth. The test asserts the value bound to each task's partial, not that `_run` applies it at runtime, so a regression inside `_run` would still pass. I kept it light to avoid running a full pipeline per case, but I can add one parallel-isolated case that runs through `dag.test()` and checks the config is actually applied. Want that?

Happy to do whichever you prefer.

willi-mueller (commented):
Thank you for creating this PR! Can you help me see where the truncation is triggered or avoided? I can only see how the flag is set and passed on but I wonder whether it has any effect.

Regarding the test: I wonder whether the truncation or not can be tested explicitly like in `test_truncate_staging_dataset`?

DRACULA1729:
Thanks for digging in, @willi-mueller.

Where it takes effect: the flag sets the destination config `truncate_tables_on_staging_destination_before_load`. Each staging-capable destination reads it in `should_truncate_table_before_load_on_staging_destination()` (e.g. `dlt/destinations/impl/bigquery/bigquery.py`), and the loader calls that in `dlt/load/load.py` (the `init_client(...)` for the staging destination) to decide whether to truncate the staging tables/files before the copy. Setting it to `False` makes the loader skip truncating the shared staging area, which is what stops parallel tasks from deleting each other's files.

On the test: the config-to-truncation link itself is already covered end to end by `test_truncate_staging_dataset` (it counts staging files with the flag on and off). What was missing was proof that the group applies that config at run time, so I added `test_truncate_staging_destination_applied_at_runtime`: it runs a parallel-isolated group via `dag.test()` and asserts the config the loader reads ends up `False`. The parametrized test still covers that each task resolves the right value across decompose modes. Just pushed.

willi-mueller:
@DRACULA1729 thank you for clarifying, I see how the options is set. Thanks also for expanding the test suite!

DRACULA1729:
@rudolfix Mind taking a look here!

rudolfix (approved):
LGTM!

it's a good point that we leave modified config values after runner ends. I added a context manager so we can have those changes only in scope of the run.

also I think that a leak ie. across celery workers is not possible. process should be forked before any user supplied code runs on it... so each process should be clean. but maybe there's something I do not know :) anyway - with the fix above - this is under control

I also combined two tests cases into one.

thanks for fixing this

PR #4081, opened 2026-06-18
Docs/hub notifications

ShreyasGS:
### Description
added docs for Notifications for Slack and Email
New entry in sidebar


zilto (approved):
Approved. IMO, content of this kind should be under "Cookbook" when it requires the user to "build something". Here, I wouldn't call this a native dltHub feature (yet).

### Proposed solution
In these scenarios:
1. Add a page under `Cookbook > dltHub` with the Python code and context
2. Add a page under the main docs `dltHub > Notifications` that explains the "business problem" and links to the cookbook.

PR #4088, opened 2026-06-19
Additional copy options

timH6502:
### Description
Add Redshift destination configuration option, that appends extra options to the generated Redshift COPY command for staged loads. This allows users to configure Reshift-specific COPY behavior.
For more information see: https://docs.aws.amazon.com/redshift/latest/dg/copy-parameters-data-conversion.html

### Result
Provides more flexibility and especially enables options like FILLRECORD.
Tested for S3 -> Redshift Serverless

### DIFF
--- a/dlt/destinations/impl/redshift/configuration.py
+++ b/dlt/destinations/impl/redshift/configuration.py
@@ -27,6 +27,7 @@ class RedshiftClientConfiguration(PostgresClientConfiguration):
 
     staging_iam_role: Optional[str] = None
     has_case_sensitive_identifiers: bool = False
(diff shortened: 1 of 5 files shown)


rudolfix (changes requested):
LGTM! but please add a note to redshift.md docs page - preferably with an example of additional option

timH6502:
> LGTM! but please add a note to redshift.md docs page - preferably with an example of additional option

@rudolfix I've added a short section to redshift.md

rudolfix (dismissed):
LGTM!

rudolfix (approved):
LGTM! tests passed now

Issue #4095, opened 2026-06-22
Dashboard "Refresh" button appears to do nothing (stale query cache; clarify its purpose)


lis365b:
## Summary

The global **Refresh** button on the pipeline dashboard appears to do nothing in the common case. It is *not* actually dead - clicking it re-runs the home cell and re-renders every section - but the result is usually visually identical, so users perceive it as broken. The one case where it really fails to update is **Browse Data**, due to a query-result cache that Refresh never invalidates.

## How Refresh currently works

- The home cell references `dlt_refresh_button` (`dlt_dashboard.py:50-51`) so a click re-runs the cell and its downstream sections (overview, schema, trace, loads, state, browse-data).
- Separately, a file watcher (`watch_changes` cell, `mo.watch.file(get_trace_file_path(...))`) auto-refreshes the view whenever the pipeline's trace file changes — i.e. when the pipeline is re-run elsewhere. So trace/metadata already updates automatically.

Evidence that Refresh *does* re-render: with a pipeline open, the "Executed: … ago" line stayed `"a few seconds ago"` after 67s of no interaction, then changed to `"1 minute ago"` immediately after clicking Refresh. So the click does trigger a re-run; it just produces an identical-looking page when nothing changed.

## Root cause for Browse Data

Query results are stored in a process-global cache (`_query_result_cache = BoundedDict(64)`, `utils/queries.py:104`), keyed by `(pipeline_name, dataset_name, query)` in `_execute_query_cached` (`utils/queries.py:89`). Refresh re-runs the query cell, but for a previously-seen query the cache returns the **old** saved result. The cache is only cleared by the dedicated "clear cache" button (`clear_query_cache()`, `utils/queries.py:28`), never by Refresh. So after re-running a pipeline with new data, Refresh shows stale rows in Browse Data.

Important: The dashboard is **read-only** — it `dlt.attach()`es and reads on-disk/destination state; it never runs the pipeline. So neither Refresh nor a cache bust produces new data. New rows only appear if the user re-ran the pipeline elsewhere. A cache bust only matters in the "I re-ran the pipeline, now show me the new rows in Browse Data" scenario.

## Options

1. **Bust the query cache on Refresh** — call `clear_query_cache()` when `dlt_refresh_button.value` is set, so Refresh becomes a true "hard reload" (trace + data). Conventional behavior; only changes anything in the re-ran-pipeline case.
[...]

Issue #4093, opened 2026-06-22
Dashboard: deselecting the pipeline shows the "no pipelines found" landing with no way to re-select


lis365b:
### dlt version
`1.28.1`

### Describe the problem
In the dlthub dashboard (marimo notebook), the pipeline picker is a `multiselect` with `max_selections=1`. When a user **clears/deselects** the currently selected pipeline (so the dropdown is empty), the dashboard falls back to the *empty-workspace* landing page, even though pipelines exist.

This has two consequences:

1. **Misleading copy** — the page renders the "no pipelines found yet" callout (`strings.home_no_pipelines`), implying the workspace has no pipelines, when in fact one was simply deselected.
2. **Dead end** — `render_no_pipelines_home()` deliberately omits the pipeline dropdown ("there is nothing to select"), so once deselected the user has **no control to re-select a pipeline**. They must reload the page or hand-edit the `?pipeline=` URL param to recover.

Root cause: the empty-selection case and the empty-workspace case both collapse to `dlt_pipeline_name = None`, and the `home` cell routes both into the same `else` branch.

### Expected behavior
The dashboard should distinguish "no pipelines exist in the workspace" from "pipelines exist but none is currently selected":

- **Empty workspace** (`dlt_all_pipelines` is empty) → keep the current `render_no_pipelines_home()` landing.
- **Deselected, pipelines exist** → keep rendering the pipeline dropdown so the user can re-select, with a neutral "Select a pipeline to inspect" hint instead of the "no pipelines found" message.

### Steps to reproduce
1. Run `dlt pipeline <name> show` (or `dlt dashboard`) in a workspace that has at least one local pipeline.
2. In the pipeline dropdown at the top, click the selected pipeline to **deselect** it so the dropdown is empty.
3. Observe: the page shows the "No pipelines found yet…" callout and the pipeline dropdown disappears, leaving no way to pick a pipeline again.

### Operating system
macOS 

### Runtime environment
Local

### Python version
3.11

PR #4096, opened 2026-06-22
fix(dashboard): keep the pipeline selector when no pipeline is selected

lis365b:
## What

Improves the dltHub dashboard's "no pipelines" / "no pipeline selected" experience and makes the empty-state copy context-neutral.

fix (issue #4093): when pipelines exist but the user clears the selection in the pipeline dropdown, the dashboard now keeps the pipeline dropdown visible and shows  "No pipeline selected. Choose a pipeline from the dropdown above to inspect it."

## Changes
- **`dlt_dashboard.py`** — the `home` cell now branches on `dlt_all_pipelines`: pipelines exist but none selected → keep the selector (`render_no_pipeline_selected_home`); for a really empty workspace: `render_no_pipelines_home`
- On the **attach-error path** (e.g. a bad `?pipeline=ghost` URL) I removed the dead/unnecesssary Refresh button 

## Tests
- added additional unit tests

Local run: dashboard suite 246 passed; e2e 18 passed. (Pre-existing `test_trace` count assertion fails only with the `hub` extra installed locally; passes in CI, its unrelated, see #4095 context.)

### DIFF
--- a/dlt/_workspace/cli/utils.py
+++ b/dlt/_workspace/cli/utils.py
@@ -10,11 +10,8 @@ from typing import (
     Dict,
     FrozenSet,
     List,
-    Literal,
     Optional,
     Tuple,
-    Union,
-    overload,
 )
 
 import dlt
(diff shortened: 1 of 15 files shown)


dlt/_workspace/helpers/dashboard/utils/pipeline.py
```diff
@@ -65,6 +67,20 @@ def get_pipeline(pipeline_name: str, pipelines_dir: str) -> dlt.Pipeline:
     return p
 
 
+def has_local_pipelines(pipelines: List[TPipelineListItem], pipelines_dir: str) -> bool:
```
tetelio: @lis365b this helper `has_local_pipelines` could be removed if we thread the data from `list_local_pipelines`. 
list_local_pipelines already looks into the disk so it could return `dlt_local_pipeline_names` in the return tuple. Its not a performance requirement but rather code re-use benefit. I think this instruction should be enough for any coding agent but happy to guide through the change if required.

dlt/_workspace/cli/utils.py
```diff
@@ -86,27 +97,55 @@ def _get_pipeline_initial_cwd(pipelines_dir: str, pipeline_name: str) -> Optiona
         return None
 
 
+@overload
```
tetelio: @lis365b  you can drop the @overload / include_local_pipeline_names pattern: its unnecessary typing ceremony for this change.

Always return a third value: pipeline directory names as discovered on disk (before URL/CLI extras are merged in). Callers that dont need it can unpack with _, e.g. pipelines_dir, pipelines, _ = list_local_pipelines(...).

That keeps a single read, a single return shape, and avoids conditional API surface (overload is more commonly used for return union types for example, where static checker cannot know if you return a type or another at runtime)

tetelio (changes requested):

tetelio (approved):
LGTM!

PR #4103, opened 2026-06-23
fix: detect duckdb_engine schemas in SQLAlchemy destination

### LINKED ISSUE #3907: SQLAlchemy destination + duckdb_engine fails with "Schema with name "{name}" already exists!"
### dlt version

1.26.0

### Describe the problem

When using the SQLAlchemy destination to write to DuckDB (via duckdb_engine), pipeline runs fail with the error:

```
(_duckdb.CatalogException) Catalog Error: Schema with name "myschema" already exists!
[SQL: CREATE SCHEMA myschema]
```

This appears to be due to [SqlalchemyClient.has_dataset](https://github.com/dlt-hub/dlt/blob/89d59d4470716e26b2cf114d61a8e4ada5dbe0ad/dlt/destinations/impl/sqlalchemy/db_api_client.py#L201-L204) not realizing that duckdb_engine returns the list of schemas namespaced by the corresponding database holding the schema ([duckdb_engine.Dialect.get_schema_names](https://github.com/Mause/duckdb_engine/blob/414afdf19c06c3bfa4c23e9892440f9294fbc35b/duckdb_engine/__init__.py#L377-L398)). i.e. if my pipeline name is `mydata` and my dataset name is `myschema`, duckdb_engine returns `mydata.myschema` in the list of returned schemas. dlt, however, is looking only for `myschema`. As a result, dlt believes the schema is missing and it attempts to create a schema that already exists, resulting in the above error.

### Expected behavior

dlt should detect that the schema already exists and not attempt to create it again, allowing the pipeline to complete successfully.

### Steps to reproduce

The following Python program reproduces the problem:

```python
import os
import shutil

import dlt
from sqlalchemy import create_engine

# ENGINE = "sqlite"
ENGINE = "duckdb"
PIPELINE = "mydata"
DATASET = "myschema"

if __name__ == '__main__':
    dbpath = f"{PIPELINE}.db"
    if os.path.exists(dbpath):
        os.remove(dbpath)
    if os.path.isdir(PIPELINE):
        shutil.rmtree(PIPELINE)

    engine = create_engine(f"{ENGINE}:///{dbpath}")

    pipeline = dlt.pipeline(
        destination=dlt.destinations.sqlalchemy(engine),
        pipeline_name=PIPELINE,
        dataset_name=DATASET,
        pipelines_dir=".",
    )

    info = pipeline.run([
        {"id": 1},
        {"id": 2},
        {"id": 3},
    ], table_name="numbers")

    print(info)
```

Here's the contents of my pyproject.toml:

```toml
[project]
name = "minimal_dlt_schema_problem_repro"
version = "0.1.0"
description = "Add your description here"
requires-python = ">=3.14"
dependencies = [
    "dlt[duckdb,sqlalchemy]>=1.26.0",
    "duckdb-engine>=0.17.0",
]
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

It doesn't matter. I can reproduce this with a custom data source as well as with a static list of Python maps.

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

While we could sidestep this problem by using the [DuckDB destination](https://dlthub.com/docs/dlt-ecosystem/destinations/duckdb), we're using SQLAlchemy because the DuckDB destination does not support [encrypted DuckDB files](https://duckdb.org/docs/lts/sql/statements/attach#database-encryption). In order to support encrypted DuckDB files, our best option is to use SQLAlchemy, which will allow us to execute the necessary SQL for all new connections in order to `ATTACH` an encrypted database (this functionality isn't available when utilizing the DuckDB destination).

fengjikui:
## Summary
- detect duckdb_engine's database-qualified schema names in the SQLAlchemy destination
- add a regression test that runs the same DuckDB-backed SQLAlchemy pipeline twice against one database file

## Root Cause
`duckdb_engine` returns schema names as `database.schema` from `get_schema_names()`. `SqlalchemyClient.has_dataset()` only checked for the bare dlt dataset name, so a second run against the same DuckDB file believed the dataset was missing and tried `CREATE SCHEMA` again, triggering `Catalog Error: Schema with name "myschema" already exists!`.

## Validation
- `uv run --with pytest --with sqlalchemy --with duckdb-engine --with duckdb --with pandas --with pyarrow pytest tests/load/sqlalchemy/test_duckdb_engine_destination.py -q`
- `uv run --with pytest --with sqlalchemy pytest tests/load/sqlalchemy/test_sqlite_destination.py::test_file_based_database_with_engine_kwargs tests/load/sqlalchemy/test_sqlite_destination.py::test_merge_with_main_dataset -q`

Closes #3907

### DIFF
--- a/.github/workflows/test_destinations_local.yml
+++ b/.github/workflows/test_destinations_local.yml
@@ -72,7 +72,7 @@ jobs:
             filesystem_drivers: "[\"memory\", \"file\"]"
             extras: "--extra sqlalchemy --extra filesystem --extra parquet --extra postgres --extra oracle --group adbc"
             needs_mysql: true
-            post_install_commands: "uv run pip install pymysql && uv run pip install sqlalchemy==1.4 && uv run dbc install mysql && uv run dbc install sqlite"
+            post_install_commands: "uv run pip install pymysql duckdb-engine && uv run pip install sqlalchemy==1.4 && uv run dbc install mysql && uv run dbc install sqlite"
 
           # SQLAlchemy 2.0 (same as above but with sqlalchemy 2.0)
           - display_name: sqlalchemy-2.0
@@ -80,7 +80,7 @@ jobs:
             filesystem_drivers: "[\"memory\", \"file\"]"
             extras: "--extra sqlalchemy --extra filesystem --extra parquet --extra postgres --extra oracle --group adbc"
             needs_mysql: true
-            post_install_commands: "uv run pip install pymysql && uv run pip install sqlalchemy==2.0 && uv run dbc install mysql && uv run dbc install sqlite"
+            post_install_commands: "uv run pip install pymysql duckdb-engine && uv run pip install sqlalchemy==2.0 && uv run dbc install mysql && uv run dbc install sqlite"
 
     env:
       ACTIVE_DESTINATIONS: ${{ matrix.destinations }}
--- a/dlt/destinations/impl/sqlalchemy/configuration.py
+++ b/dlt/destinations/impl/sqlalchemy/configuration.py
@@ -214,6 +214,7 @@ class SqlalchemyCredentials(ConnectionStringCredentials):
 
 
 SQLITE_DB_NAME_PAT = "%s.db"
+DUCKDB_DB_NAME_PAT = "%s.duckdb"
 
 
 @configspec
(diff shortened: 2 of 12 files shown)


rudolfix:
@fengjikui just out of curiosity: what is your use case for encrypted duckdb databases? are you also generating encrypted parquets? we could add direct support to our duckdb destination

fengjikui:
Thanks for asking. The encrypted DuckDB part comes from the original #3907 use case: the reporter is using the SQLAlchemy destination because they need to run connection setup SQL for encrypted DuckDB files, which the native DuckDB destination does not currently expose.

I do not have a separate encrypted-DuckDB production setup to represent beyond that issue. The bug this PR scopes in on is narrower: with `duckdb_engine`, schema introspection returns names like `<database>.<schema>`, while `SqlalchemyClient.has_dataset()` was checking only `<schema>`. That makes a second run think the dataset is missing and issue `CREATE SCHEMA`, which then fails because the schema already exists.

If you decide to add direct encrypted DuckDB support to the native destination, I would be happy to help test or shape it from the #3907 repro path. For this PR I kept the change limited to making the SQLAlchemy destination correctly recognize the existing DuckDB schema.

rudolfix:
we just merged #4161 enabling full test workflow for duckdb + sqlalchemy. this case got fixed as well

PR #4105, opened 2026-06-23
Bump jsonpath-ng and improve test coverage

### LINKED ISSUE #3869: Sub-dependency `ply` is now archived and has an associated CVE, included because jsonpath-ng dep is pinned to <1.8.0 version
### dlt version

1.25.0

### Describe the problem

Dlt includes a package called jsonpath-ng, which in turn includes a sub-dependency called ply.

Ply was archived at the end of last year, and also includes a [critical CVE](https://github.com/advisories/GHSA-qc6m-pwr3-g72p).

Jsonpath-ng [recently vendored the ply package](https://github.com/h2non/jsonpath-ng/pull/208) into its package and removed the insecure parts of the code. This change made it into their recent 1.8.0 release.

Currently, [dlt pins the jsonpath-ng](https://github.com/dlt-hub/dlt/blob/devel/pyproject.toml#L54-L55) to versions before 1.8.0, due to required code changes in the new version.

The ask here is to support jsonpath-ng version 1.8.0 to get rid of the dependence on an archived package with a critical CVE

### Expected behavior

_No response_

### Steps to reproduce

Run poetry install and inspect the dependency tree

### Operating system

Linux, macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

_No response_

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

Travior:
Bumps `jsonpath-ng` to `1.8.0`. This removes a transitive dependency on `ply` which was archived and contained an active CVE (not exploitable through `dlt` but still showing up on security scanners). Due to changes in how the string conversion works from `1.8.0` onwards we have to vendor in our own logic to convert a `JSONPath` to a `str` value that can be re-parsed again.

Closes #3869

### DIFF
--- a/dlt/common/jsonpath.py
+++ b/dlt/common/jsonpath.py
@@ -3,7 +3,7 @@ from itertools import chain
 
 from dlt.common.typing import DictStrAny
 
-from jsonpath_ng import JSONPath, Fields as JSONPathFields
+from jsonpath_ng import JSONPath, Fields as JSONPathFields, DatumInContext, Root, This
 from jsonpath_ng.ext import parse as _parse
 
 TJsonPath = Union[str, JSONPath]  # Jsonpath compiled or str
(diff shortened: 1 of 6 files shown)


burnash (approved):
Very good, thanks for the additional tests, @Travior

PR #4117, opened 2026-06-24
fix: make Snowflake schema creation idempotent

### LINKED ISSUE #3891: fix(sql_client) SnowflakeSqlClient.create_dataset() missing IF NOT EXISTS causes race condition with concurrent pipeline runs
### dlt version

1.24

### Describe the problem

SnowflakeSqlClient.create_dataset() issues a plain CREATE SCHEMA without IF NOT EXISTS. When two pipeline workers or processes attempt to initialize the same schema concurrently, one of them fails with a "Schema already exists" error from Snowflake.

One of the concurrent workers throws an error because the schema was created by the other worker between the existence check and the CREATE SCHEMA execution.

### Expected behavior

Schema creation is idempotent. Concurrent initializations should both succeed, just as they would with CREATE SCHEMA IF NOT EXISTS.

### Steps to reproduce

Rare race condition, so it can be hard to reproduce, but:
Run two dlt pipelines targeting the same Snowflake schema simultaneously (e.g. in a parallelized task runner or Kubernetes job), where neither has previously created the schema. 

### Operating system

Linux

### Runtime environment

Docker, Docker Compose

### Python version

3.13

### dlt data source

_No response_

### dlt destination

Snowflake

### Other deployment details

_No response_

### Additional information

Suggested fix:
python
```
# dlt/destinations/sql_client.py
def create_dataset(self) -> None:
    self.execute_sql(
        "CREATE SCHEMA IF NOT EXISTS %s" % self.fully_qualified_dataset_name()
    )
```

fengjikui:
### Description

Snowflake dataset creation now uses `CREATE SCHEMA IF NOT EXISTS` instead of plain `CREATE SCHEMA`. This removes the race where two concurrent workers both pass the existence check and one fails because the other created the schema first.

The change is scoped to `SnowflakeSqlClient`; the generic SQL client behavior remains unchanged for other destinations.

### Related Issues

- Fixes #3891

### Additional Context

Validation run locally:

```text
env ACTIVE_DESTINATIONS='["snowflake"]' uv run --extra snowflake pytest tests/load/snowflake/test_snowflake_table_builder.py::test_create_dataset_uses_idempotent_schema_sql -q
# 1 passed in 0.16s

uv run ruff check dlt/destinations/impl/snowflake/sql_client.py tests/load/snowflake/test_snowflake_table_builder.py
# All checks passed!

uv run ruff format --check dlt/destinations/impl/snowflake/sql_client.py tests/load/snowflake/test_snowflake_table_builder.py
# 2 files already formatted

python3 -m py_compile dlt/destinations/impl/snowflake/sql_client.py tests/load/snowflake/test_snowflake_table_builder.py
```

### DIFF
--- a/dlt/common/destination/capabilities.py
+++ b/dlt/common/destination/capabilities.py
@@ -180,6 +180,7 @@ class DestinationCapabilitiesContext(ContainerInjectableContext):
     # use naming convention in the schema
     naming_convention: TNamingConventionReferenceArg = None
     alter_add_multi_column: bool = True
+    supports_create_schema_if_not_exists: bool = False
     supports_create_table_if_not_exists: bool = True
     supports_truncate_command: bool = True
     schema_supports_numeric_precision: bool = True
--- a/dlt/destinations/impl/snowflake/factory.py
+++ b/dlt/destinations/impl/snowflake/factory.py
@@ -145,6 +145,7 @@ class snowflake(Destination[SnowflakeClientConfiguration, "SnowflakeClient"]):
         caps.is_max_text_data_type_length_in_bytes = True
         caps.supports_ddl_transactions = True
         caps.alter_add_multi_column = True
+        caps.supports_create_schema_if_not_exists = True
         caps.supports_clone_table = True
         caps.supported_merge_strategies = ["delete-insert", "upsert", "scd2", "insert-only"]
         caps.supported_replace_strategies = [
--- a/dlt/destinations/impl/snowflake/sql_client.py
+++ b/dlt/destinations/impl/snowflake/sql_client.py
@@ -117,6 +117,9 @@ class SnowflakeSqlClient(SqlClientBase[snowflake_lib.SnowflakeConnection], DBTra
     def native_connection(self) -> "snowflake_lib.SnowflakeConnection":
         return self._conn
 
+    def create_dataset(self) -> None:
+        self.execute_sql("CREATE SCHEMA IF NOT EXISTS %s" % self.fully_qualified_dataset_name())
+
     def drop_tables(self, *tables: str) -> None:
         # Tables are drop with `IF EXISTS`, but snowflake raises when the schema doesn't exist.
         # Multi statement exec is safe and the error can be ignored since all tables are in the same schema.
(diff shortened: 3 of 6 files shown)


Travior (changes requested):
Thanks for the contribution, the general direction is correct, however if we start to introduce idempotent schema creation we probably shouldn't just override this for snowflake. In my mind we should add this to the destination capabilities (like we do with `supports_create_table_if_not_exists`) and conditionally emit this sql in the base class

Travior:
Hi @fengjikui, thanks for the PR. I've left a comment in my review. If you are willing to further work on this feel free to commit your changes, otherwise I can take this over

fengjikui:
Thanks, addressed in `17801a347`.

I moved the idempotent schema creation behavior into destination capabilities instead of keeping it as a Snowflake-only override:

- added `supports_create_schema_if_not_exists` on `DestinationCapabilitiesContext`, defaulting to the existing non-idempotent behavior;
- made `SqlClientBase.create_dataset()` emit `CREATE SCHEMA IF NOT EXISTS` only when that capability is enabled;
- opted Snowflake into the capability and removed the Snowflake `create_dataset()` override.

Validation:
- `uv run pytest tests/load/test_sql_client.py::test_create_dataset_uses_schema_idempotency_capability tests/load/snowflake/test_snowflake_table_builder.py::test_create_dataset_uses_idempotent_schema_sql -q` -> 3 passed
- `uv run ruff check dlt/common/destination/capabilities.py dlt/destinations/sql_client.py dlt/destinations/impl/snowflake/factory.py dlt/destinations/impl/snowflake/sql_client.py tests/load/test_sql_client.py tests/load/snowflake/test_snowflake_table_builder.py` -> passed
- `uv run ruff format --check dlt/common/destination/capabilities.py dlt/destinations/sql_client.py dlt/destinations/impl/snowflake/factory.py dlt/destinations/impl/snowflake/sql_client.py tests/load/test_sql_client.py tests/load/snowflake/test_snowflake_table_builder.py` -> passed
- `git diff --check` -> passed

Travior:
We've decided to go into a different direction with this and would rather fail fast and loud on schema conflicts like this. Thank you still for your contribution!

PR #4118, opened 2026-06-24
fix: preserve incremental step order on clone

fengjikui:
## Summary

Refs #3838.

This preserves the explicit pipe position of an existing incremental step when a resource is cloned and its hints/config wrappers are ejected and injected again. The issue was that hint-based incremental replacement during clone removed the existing incremental transform and re-added it through the default affinity-based append path, which could move user-inserted steps such as `MetricsItem` before incremental filtering.

The fix records the current incremental step index before replacement/ejection and passes that index back into `set_incremental` when the incremental is reinserted. Normal first-time incremental insertion still uses the existing affinity behavior.

## Validation

- `uv run --with duckdb --with pyarrow --with pydantic pytest tests/extract/test_incremental.py::test_clone_preserves_inserted_step_after_hints_incremental tests/extract/test_incremental.py::test_pydantic_columns_validator tests/extract/test_extract.py::test_add_metrics tests/extract/test_extract.py::test_custom_metrics_preserved_when_all_items_filtered -q` -> `6 passed in 1.37s`
- `uv run ruff check dlt/extract/resource.py tests/extract/test_incremental.py` -> `All checks passed!`
- `python3 -m py_compile dlt/extract/resource.py tests/extract/test_incremental.py` -> passed

I also ran the issue-style `pipeline.run` reproduction locally after the fix and observed `row_count: 2`, so the metrics step now sees only rows that survive incremental filtering.

### DIFF
--- a/dlt/extract/resource.py
+++ b/dlt/extract/resource.py
@@ -470,11 +470,13 @@ class DltResource(Iterable[TDataItem], DltResourceHints):
             self._pipe.insert_step(item_transform, insert_at)
         return self
 
-    def _remove_incremental_step(self) -> None:
-        self._pipe.remove_by_type(Incremental, IncrementalResourceWrapper)
+    def _remove_incremental_step(self) -> int:
+        return self._pipe.remove_by_type(Incremental, IncrementalResourceWrapper)
 
     def set_incremental(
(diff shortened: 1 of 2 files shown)


zilto:
Closing as duplicate #3839

PR #4120, opened 2026-06-24
fix(mssql): map varchar to text type

### LINKED ISSUE #4100: Allow support of using VARCHAR or NVARCHAR in MSSQL Destination
### Feature description

I would like an option to use either VARCHAR or NVARCHAR in the MSSQL destination. Due to the smaller size limitation our system currently uses VARCHAR for majority of the indexes and large text fields for search optimization.

Currently the data type mapping of the MSSQL destination is set to NVARCHAR which is causing truncation when trying to push data across from our source system into destination using DLT. It would be fantastic if there was an option to select which type of text field to use in the destination.

Not sure if you accept PR's from the public but I am open to submitting a PR to implement this feature.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

VARCHAR has a higher max character limit (8000 vs 4000) and currently need to store information that is higher then 4000 characters.

### Proposed solution

Add an optional configuration override similar to `has_case_sensitive_identifiers` that changes which type of string to use such as `string_data_type`.

Then create a new function in the `TypeMapperImpl` called `to_db_string_type` that allows subclass to override the string configuration on a case by case basis.

I can understand that most other DB engines do not have mulitple string data types hence this would not be a priority.

### Related issues

N/A

Mady356:
### Description

This PR adds MSSQL `varchar` to the destination-to-schema type mapping so that `varchar` columns are recognized as dlt `text` columns, matching the existing `nvarchar` behavior.

Changes:
- Added `varchar -> text` to `MsSqlTypeMapper.dbt_to_sct`.
- Added a focused regression test for MSSQL text type mapping from destination types.

### Related Issues

- Closes #4100

### Additional Context

Validated with:

```bash
pytest tests/load/mssql/test_type_mapper.py

### DIFF
--- a/dlt/destinations/impl/mssql/factory.py
+++ b/dlt/destinations/impl/mssql/factory.py
@@ -42,6 +42,7 @@ class MsSqlTypeMapper(TypeMapperImpl):
 
     dbt_to_sct = {
         "nvarchar": "text",
+        "varchar": "text",
         "float": "double",
         "bit": "bool",
         "datetimeoffset": "timestamp",
--- a/tests/load/mssql/test_type_mapper.py
+++ b/tests/load/mssql/test_type_mapper.py
@@ -1,0 +1,9 @@
+from dlt.destinations import mssql
+
+
+def test_from_destination_type_mssql_text_types() -> None:
+    caps = mssql().capabilities()
+    mapper = caps.get_type_mapper()
+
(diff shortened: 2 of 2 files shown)


Travior:
As mentioned in the issue this is expected behavior in my opinion

PR #3867, opened 2026-04-15
fix: strip trailing slash from FilesystemClient.dataset_path and get_table_dir

mattiasthalen:
### Description

`FilesystemClient.dataset_path` and `FilesystemClient.get_table_dir` both return paths ending with a trailing separator. This is benign on most backends (404 from `BlobClient.exists`, normalized to `False`), but OneLake (Microsoft Fabric) returns `403 ClientAuthenticationError` for the same request, which fatally kills the load at `initialize_storage` and `truncate_tables`.

This PR strips the trailing separator from both methods and adds regression tests that assert neither returned path ends with `/`.

### Related Issues

- Fixes #3866

### Additional Context

- Draft until the trailing-slash regression tests land and the filesystem test suite is green locally.
- Surfaced as part of #3865 (Fabric notebook user-identity auth). That feature PR will rebase on this bugfix once merged.
- Two-line change, no API surface impact, no new dependencies.

PR #4122, opened 2026-06-24
fix: avoid trailing filesystem directory probes

### LINKED ISSUE #3866: `FilesystemClient` trailing slash causes OneLake 403 at `initialize_storage` and `truncate_tables`
### dlt version

`1.25.0` (current `devel`, verified against commit `9636de5e`)

### Describe the problem

`FilesystemClient.dataset_path` and `FilesystemClient.get_table_dir` both return paths with a trailing separator. This is benign on most backends — `BlobClient.exists` responds with `404 ResourceNotFoundError` for a blob name ending in `/`, and `adlfs._exists` normalizes that to `False`. On OneLake (Microsoft Fabric), the same request returns `403 ClientAuthenticationError`, which bubbles all the way up and kills the load twice:

1. `FilesystemClient.initialize_storage` calls `self.fs_client.isdir(self.dataset_path)` as the first step, before any data is written — fatal 403 on the trailing-slash `dataset_path`.
2. `FilesystemClient.truncate_tables` iterates `get_table_dirs(...)` and calls `self.fs_client.exists(table_dir)` per table — fatal 403 on each trailing-separator `table_dir`.

Source lines on `devel` @ `9636de5e`:

* `dlt/destinations/impl/filesystem/filesystem.py:591` — `return self.pathlib.join(self.bucket_path, self.dataset_name, "")` forces a trailing slash.
* `dlt/destinations/impl/filesystem/filesystem.py:882` — `table_dir: str = self.pathlib.dirname(table_prefix) + self.pathlib.sep` forces a trailing separator.

Both are latent defects in generic `FilesystemClient`. OneLake just makes them visibly fatal by responding `403` instead of `404`.

### Expected behavior

`dataset_path` and `get_table_dir` return paths without a trailing separator. All downstream consumers that need to append further segments already do so via `pathlib.join(...)`, where whether the parent ends in `/` is irrelevant.

Proposed fix, two lines:

```python
# FilesystemClient.dataset_path
return self.pathlib.join(self.bucket_path, self.dataset_name)

# FilesystemClient.get_table_dir
table_dir: str = self.pathlib.dirname(table_prefix)
```

Regression tests should assert neither returned path ends with `/`, using an in-memory filesystem fixture so no real Azure/OneLake credentials are needed.

### Steps to reproduce

Minimal shape-only reproduction using the same `adlfs.AzureBlobFileSystem` instance against OneLake:

```python
# Against OneLake, same fs instance:
fs.isdir("<ws-guid>/<lh-guid>/Files/_dlt_stage/demo/")  # → raises 403 ClientAuthenticationError
fs.isdir("<ws-guid>/<lh-guid>/Files/_dlt_stage/demo")   # → True
```

End-to-end repro using a Fabric notebook kernel (requires Fabric tenant with Warehouse + Lakehouse grants as the interactive user):

1. Configure `DESTINATION__FILESYSTEM__BUCKET_URL` pointing to `abfss://<ws-guid>@onelake.dfs.fabric.microsoft.com/<lh-guid>/Files/_dlt_stage`.
2. Configure a dlt pipeline with any destination that uses `staging=\"filesystem\"` (e.g. `fabric`).
3. `pipeline.run(...)` — fails at `initialize_storage` with `403 ClientAuthenticationError` on the `fs.isdir(self.dataset_path)` call, before any data is uploaded.

### Operating system

Linux

### Runtime environment

Other

### Python version

3.11

### dlt data source

N/A — this is a `FilesystemClient` path-construction bug triggered by any resource. Repros with a minimal static resource.

### dlt destination

Filesystem & buckets

### Other deployment details

Microsoft Fabric Python notebook kernel, OneLake-backed filesystem staging, Fabric Warehouse as the primary destination. The filesystem destination ships a custom `OnelakeFileSystem` (in `fsspec_wrapper.trident.core`) registered as the `abfss://` handler, which is the code path that surfaces the 403 instead of a 404. Non-OneLake Azure Blob users still hit the latent bug but observe it as silent `False` from `isdir`/`exists`.

### Additional information

Surfaced as part of [OSS-41](https://linear.app/dlthub/issue/OSS-41/support-interactive-notebook-user-auth-for-the-fabric-destination) (Fabric notebook user-identity auth). Splitting this bugfix out as its own ticket per CONTRIBUTING — `fix/` branches require a ticket, and this is a pure bug with zero API surface that can land independently of the Fabric feature work. The Fabric feature PR will rebase on this once merged.

I have a validated reproduction and fix in a personal sandbox — happy to open the PR against `devel` with regression tests.

fengjikui:
## Description

Fixes #3866.

`FilesystemClient.dataset_path` and `FilesystemClient.get_table_dir()` are passed directly to fsspec directory probes such as `isdir()` and `exists()`. They currently force a trailing separator, which is benign on some backends but can fail on stricter implementations such as OneLake/Fabric.

This keeps the existing table-prefix behavior for layouts that need folder-style prefixes, but returns dataset/table directory paths without a trailing separator before those paths are probed or converted to remote URLs.

## Validation

```text
env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy uv run --with pyparsing pytest \
  'tests/load/filesystem/test_filesystem_client.py::test_filesystem_dirs_do_not_have_trailing_separators[data-{table_name}/{load_id}.{file_id}.{ext}]' \
  'tests/load/filesystem/test_filesystem_client.py::test_filesystem_dirs_do_not_have_trailing_separators[data-{table_name}.{load_id}.{file_id}.{ext}]' -q
# 2 passed

env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy uv run --with pyparsing pytest \
  tests/load/filesystem/test_filesystem_client.py::test_list_dlt_table_files_with_separator_in_pipeline_name -q
# 5 passed

env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy uv run --with pyparsing pytest \
  'tests/load/filesystem/test_filesystem_client.py::test_successful_load[data-{table_name}/{load_id}.{file_id}.{ext}-replace]' \
  'tests/load/filesystem/test_filesystem_client.py::test_successful_load[memory:///m-{table_name}/{load_id}.{file_id}.{ext}-replace]' -q
# 2 passed

env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy uv run ruff check \
  dlt/destinations/impl/filesystem/filesystem.py \
  tests/load/filesystem/test_filesystem_client.py
# All checks passed!

env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY -u http_proxy -u https_proxy -u all_proxy uv run python -m py_compile \
  dlt/destinations/impl/filesystem/filesystem.py \
  tests/load/filesystem/test_filesystem_client.py
# passed

git diff --check
# passed
```

I also tried the full parametrized `test_filesystem_dirs_do_not_have_trailing_separators` locally. The local `data` cases passed; remote bucket cases require credentials or optional filesystem extras that are not present in my local environment.

### DIFF
--- a/dlt/destinations/impl/filesystem/filesystem.py
+++ b/dlt/destinations/impl/filesystem/filesystem.py
@@ -591,7 +591,7 @@ class FilesystemClient(
         """A path within a bucket to tables in a dataset
         NOTE: dataset_name changes if with_staging_dataset is active
         """
-        return self.pathlib.join(self.bucket_path, self.dataset_name, "")  # type: ignore[no-any-return]
+        return self.pathlib.join(self.bucket_path, self.dataset_name)  # type: ignore[no-any-return]
 
     @contextmanager
     def with_staging_dataset(self) -> Iterator["FilesystemClient"]:
@@ -972,12 +972,12 @@ class FilesystemClient(
     def get_table_dir(
         self, table_name: str, remote: bool = False, schema_name: Optional[str] = None
     ) -> str:
-        """Returns a directory containing table files, ending with separator.
+        """Returns a directory containing table files.
         Note that many tables can share the same table dir
         """
         # dlt tables do not respect layout (for now)
         table_prefix = self.get_table_prefix(table_name, schema_name=schema_name)
-        table_dir: str = self.pathlib.dirname(table_prefix) + self.pathlib.sep
+        table_dir: str = self.pathlib.dirname(table_prefix)
         if remote:
             table_dir = self.make_remote_url(table_dir)
         return table_dir
--- a/tests/load/filesystem/test_filesystem_client.py
+++ b/tests/load/filesystem/test_filesystem_client.py
@@ -137,22 +137,26 @@ def test_filesystem_follows_local_dir(location: str) -> None:
 @pytest.mark.parametrize(
     "layout", ("{table_name}/{load_id}.{file_id}.{ext}", "{table_name}.{load_id}.{file_id}.{ext}")
 )
-def test_trailing_separators(layout: str, with_gdrive_buckets_env: str) -> None:
+def test_filesystem_dirs_do_not_have_trailing_separators(
+    layout: str, with_gdrive_buckets_env: str
+) -> None:
     os.environ["DESTINATION__FILESYSTEM__LAYOUT"] = layout
     load = setup_loader("_data")
     client: FilesystemClient = load.get_destination_client(Schema("empty"))  # type: ignore[assignment]
-    # assert separators
-    assert client.dataset_path.endswith("_data/")
-    assert client.get_table_dir("_dlt_versions").endswith("_dlt_versions/")
-    assert client.get_table_dir("_dlt_versions", remote=True).endswith("_dlt_versions/")
+    # dataset and table dirs are queried directly with fsspec and must not end
+    # in a separator: some backends reject trailing-separator directory probes.
+    assert client.dataset_path.endswith("_data")
+    assert not client.dataset_path.endswith("/")
+    assert client.get_table_dir("_dlt_versions").endswith("_dlt_versions")
+    assert client.get_table_dir("_dlt_versions", remote=True).endswith("_dlt_versions")
     is_folder = layout.startswith("{table_name}/")
     if is_folder:
-        assert client.get_table_dir("letters").endswith("_data/letters/")
-        assert client.get_table_dir("letters", remote=True).endswith("_data/letters/")
+        assert client.get_table_dir("letters").endswith("_data/letters")
+        assert client.get_table_dir("letters", remote=True).endswith("_data/letters")
     else:
         # strip prefix
-        assert client.get_table_dir("letters").endswith("_data/")
-        assert client.get_table_dir("letters", remote=True).endswith("_data/")
+        assert client.get_table_dir("letters").endswith("_data")
+        assert client.get_table_dir("letters", remote=True).endswith("_data")
     if is_folder:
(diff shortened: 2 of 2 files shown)


mattfaltyn:
@fengjikui Thanks for addressing the OneLake path-probe failure.

One concern before merge: this changes the documented [`FSClientBase` trailing-separator contract](https://github.com/dlt-hub/dlt/blob/5f600c0ea82fb9a484c2c49fee109c77a26b944c/dlt/destinations/fs_client.py#L14-L25) and leaves [`test_destination_config_in_name`](https://github.com/dlt-hub/dlt/blob/5f600c0ea82fb9a484c2c49fee109c77a26b944c/tests/destinations/test_destination_name_and_config.py#L214-L218) failing. I reproduced this against GitHub's current merge ref: **87 passed, 2 skipped, 1 failed**.

Could we either preserve the return contract and strip separators only at the affected `isdir`/`exists` calls, or update `FSClientBase` and the remaining contract test if the new return shape is intentional?

Also, [#3867](https://github.com/dlt-hub/dlt/pull/3867) appears to implement the same fix and already updates that test, so the two PRs should be reconciled.

fengjikui:
Thanks for catching this. I confirmed #3867 makes the same contract change and also updates `test_destination_config_in_name`, so this PR duplicates it. Closing in favor of #3867 rather than creating competing review work.

PR #4123, opened 2026-06-24
fix: include pandas in filesystem extra

### LINKED ISSUE #3876: Pipeline example dependencies don't include an install of pandas to support read_csv

fengjikui:
## Summary
- add pandas to the `filesystem` optional extra so `dlt[filesystem]` installs the dependency used by `dlt.sources.filesystem.read_csv`
- add a CLI init regression test that keeps the filesystem extra aligned with the generated filesystem pipeline examples

## Why
`dlt init filesystem postgres` generates requirements that rely on the `filesystem` extra, but the filesystem example calls `read_csv()`, whose implementation imports pandas directly. Without pandas in the extra, users can install the generated requirements and still hit a missing dependency when running the example.

Fixes #3876.

## Validation
- `uv run --extra hub pytest tests/workspace/cli/test_init_command.py::test_filesystem_extra_installs_read_csv_dependency tests/workspace/cli/test_init_command.py::test_init_command_core_source_requirements_with_extras -q` -> 3 passed
- `uv run --extra filesystem python - <<'PY' ...` confirmed editable metadata exposes `pandas<3,>=2.3.0; extra == 'filesystem'`
- `uv run python tools/check_dependency_changes.py origin/devel HEAD` reports the intended optional dependency addition: `[filesystem]: + pandas<3,>=2.3.0`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -102,6 +102,7 @@ ducklake = [
 filesystem = [
     "s3fs>=2022.4.0",
     "botocore>=1.28",
+    "pandas>=2.3.0,<3",
 ]
 s3 = [
     "s3fs>=2022.4.0",
--- a/tests/workspace/cli/test_init_command.py
+++ b/tests/workspace/cli/test_init_command.py
@@ -4,9 +4,11 @@ from copy import deepcopy
 import hashlib
 import os
 import contextlib
+from pathlib import Path
 from subprocess import CalledProcessError
 from typing import List, Tuple, Optional
 import pytest
+import tomlkit
 from pytest_console_scripts import ScriptRunner
 from unittest import mock
 from pytest import MonkeyPatch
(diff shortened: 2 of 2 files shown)


zilto:
Not currently planned as this should be an optional dependency (i.e., not all `filesystem` users need `pandas`)

PR #4124, opened 2026-06-25
docs(hub): add secrets management page

ShreyasGS:
### Description
Adds a Secrets management page to the dltHub docs under Pipeline operations, between Workspace setup and Deployments. Documents:

- How dltHub handles secrets you deploy (encryption at rest, VPC-isolated vault identity, short-lived runtime access)
- The bring-your-own-vault path via Google Secret Manager (provider config, deploy, runtime resolution)
- Static-egress-IPs hardening for vaults or warehouses that require IP allowlisting

The Google Secret Manager walkthrough delegates GCP-side steps (service-account creation, secret naming) to the existing OSS vaults reference, keeping the hub page focused on what's specific to deployed workspaces.

OSS vaults docs (general-usage/credentials/vaults.

Three supporting improvements that the hub page li

- Restructured the "Naming convention" examples inplest first: Store your whole secrets.toml(simplest) / Store per-section fragments / Store single values.
- Added the missing dlt_secrets_toml worked examplhe simplest and most common setup but wasn't shownexplicitly.
- Removed a duplicate normalization block (and its
- Flipped list_secrets = true → false in the "Alternative" config example so it actually matches the case when the service account
only has secretAccessor.

### DIFF
--- a/docs/website/docs/general-usage/credentials/vaults.md
+++ b/docs/website/docs/general-usage/credentials/vaults.md
@@ -130,7 +130,7 @@ enable_google_secrets = true  # google secrets provider is disabled by default
 [providers.google_secrets]
 only_secrets = false
 only_toml_fragments = false
-list_secrets = true  # we recommend pre-listing secrets to minimize calls to google backend
+list_secrets = false  # listing not available: dlt does direct lookups, needs only secretAccessor
 ```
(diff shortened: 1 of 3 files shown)


zilto (approved):
LGTM

PR #4130, opened 2026-06-28
feat(bigquery): implement atomic replace via GCS-staged WRITE_TRUNCATE_DATA

### LINKED ISSUE #4012: BigQuery: atomic metadata-preserving replace via `WRITE_TRUNCATE_DATA` load job
### Feature description

I may be missing an existing way to do this, so I wanted to ask before working on a PR.

I currently maintain a downstream patch around BigQuery replace behavior, and I would like to upstream a cleaner version if the direction makes sense.

I would like BigQuery to have an opt-in replace mode that replaces the data in an existing destination table without dropping or recreating the table.

BigQuery has a `WRITE_TRUNCATE_DATA` write disposition for Load and Query jobs. For existing tables, it is meant to replace the table data while preserving the table schema, constraints, and table metadata.

For dlt pipelines that already use filesystem staging, this seems useful because the final replace step could be a BigQuery Load Job from the staged files directly into the destination table. That would avoid both a BigQuery staging-table scan and clone/copy-style replacement of the destination table.

The default BigQuery replace behavior would stay unchanged. This would be an explicitly selected mode.

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

In our production pipelines, some full-refresh BigQuery tables are read continuously by downstream queries, views, and materialized views. Those destination tables can also have constraints, policy tags, row access policies, labels, descriptions, and table-level access controls that should survive a refresh.

The tradeoff I see with the current BigQuery replace strategies is:

- `truncate-and-insert` keeps the destination table object, but there can be a period where the table has already been truncated and the replacement data has not been fully loaded yet.
- `insert-from-staging` also keeps the destination table object, but it needs an `INSERT ... SELECT` from the BigQuery staging table, so the staged data is scanned again.
- `staging-optimized` avoids the empty-table window, but the clone/copy-style replacement can lose destination-side schema, constraints, or access/security metadata.

I also looked at BigQuery Copy Jobs. A Copy Job with `WRITE_TRUNCATE` avoids the empty-table window, but it uses replacement semantics that can remove destination-side constraints or use source-side schema/metadata.

I tested whether Copy Jobs can use `WRITE_TRUNCATE_DATA`, because the Python client exposes the enum value, but BigQuery rejected the job with:

```text
WriteDisposition WRITE_TRUNCATE_DATA is not yet supported for table copy jobs
```

So I do not currently see a BigQuery Copy Job path that gives all of these at once:

- no table-not-found or empty-table window;
- preserving the existing destination table object and its metadata;
- no query scan of a BigQuery staging table.

That is why I am looking at a Load Job with `WRITE_TRUNCATE_DATA`.

### Proposed solution

Use filesystem staging as the input to a final BigQuery Load Job.

Suggested initial scope:

- BigQuery destination only.
- Filesystem staging with remote URIs readable by BigQuery, such as GCS.
- One aggregated BigQuery Load Job per destination table.
- No silent fallback to a strategy with weaker semantics.
- Schema behavior similar to `truncate-and-insert`, not full table replacement.

The rough flow would be:

1. dlt writes the usual load files.
2. filesystem staging uploads them and creates `.reference` follow-up jobs, as it does today.
3. In this mode, BigQuery would not load each `.reference` job directly into the final table.
4. After all jobs for a table or table chain are complete, BigQuery would collect the staged file URIs for that destination table.
5. BigQuery would submit one Load Job into the final table:

```python
source_uris = [...]
destination = final_table
write_disposition = "WRITE_TRUNCATE_DATA"
create_disposition = "CREATE_NEVER"  # assuming dlt has already created/updated the table
```

I would expect the schema semantics to be close to `truncate-and-insert`:

- compatible full reloads should work;
- column additions applied by dlt before the Load Job should work;
- column drops, incompatible type changes, and full schema replacement would not be the goal of this mode.

For local files, I would probably leave support out initially. BigQuery can load multiple remote URIs in one Load Job, but local `load_table_from_file(...)` cannot combine multiple local files into one atomic Load Job. If this mode is selected for local multi-file loads, I would rather raise an actionable error than silently fall back to different replace semantics.

Before working on a PR, I would mainly like to confirm the API shape:

1. Would you prefer a new BigQuery `replace_strategy`, for example `atomic-load` or `truncate-data-load`?
2. Or should this be an opt-in mode for an existing strategy such as `truncate-and-insert` when filesystem staging is available?

My current preference is a new explicit strategy, because the implementation and failure modes are different from both `truncate-and-insert` and `staging-optimized`. But semantically it is closer to an atomic `truncate-and-insert` than to `staging-optimized`, so I am happy to follow the maintainers' preferred API.

### Related issues

I found these related items, although I am not aware of an existing dlt issue for this exact BigQuery behavior:

- Similar availability problem for Snowflake `staging-optimized` replace: #2673
- Snowflake atomic table swap implementation: #3540
- BigQuery `writeDisposition` REST documentation: https://cloud.google.com/bigquery/docs/reference/rest/v2/Job
- BigQuery audit metadata `WriteDisposition` documentation: https://cloud.google.com/bigquery/docs/reference/auditlogs/rest/Shared.Types/BigQueryAuditMetadata.WriteDisposition

oystersuki:
### Description

Adds an opt-in BigQuery config flag, `enable_atomic_replace`, that makes the `truncate-and-insert` replace strategy perform an **atomic, metadata-preserving** full replace using a single BigQuery Load Job with `write_disposition="WRITE_TRUNCATE_DATA"`, when a filesystem (GCS) staging destination is used.

This follows the design agreed with @rudolfix in the linked issue: a config flag on `BigQueryClientConfiguration` that piggybacks on `truncate-and-insert` (active when the strategy is `truncate-and-insert`, the flag is on, and a staging destination is configured), rather than a new `replace_strategy`. Defaulting it on for `truncate-and-insert` is intentionally left for review.

**Why `WRITE_TRUNCATE_DATA`.** I verified the four candidate mechanisms on BigQuery (US), against a table carrying a description, labels, column descriptions, a NOT ENFORCED primary key, a column policy tag, a row access policy, and a dependent materialized view:

| operation | table object | table desc / labels | column desc / policy tag / PK | row access policy | materialized view |
|---|---|---|---|---|---|
| `CREATE OR REPLACE` | replaced | lost | lost | lost | **broken** ("references a table that was deleted and recreated") |
| `WRITE_TRUNCATE` (load) | kept | preserved | lost (schema overwritten) | lost | survives* |
| `WRITE_TRUNCATE_DATA` (load) | kept | preserved | **preserved** | **preserved** | survives |
| Copy job `WRITE_TRUNCATE` | kept | preserved | lost | lost | survives* |

`WRITE_TRUNCATE_DATA` is the only operation that preserves column-level metadata (descriptions, policy tags, PK constraints) and row access policies while keeping the table object and dependent materialized views intact. It is also atomic (no empty-table window), does not scan a staging table, and stays a Load Job (load reservations, not query slots). Copy Jobs reject it (`WRITE_TRUNCATE_DATA is not yet supported for table copy jobs`), so a Load Job is the path. Only *materialized* views are ever affected (and only by `CREATE OR REPLACE`); logical views resolve by name and survive all four operations.

(* MV survival under schema-overwriting ops is for schema-compatible reloads; `WRITE_TRUNCATE_DATA` preserves the schema, so it is unconditional.)

**How it works.** With GCS staging, per-file `.reference` jobs become no-op `FinalizedLoadJobWithFollowupJobs` so they still trigger the table-chain hook. The hook (`create_table_chain_completed_followup_jobs`) emits one aggregated reference job per table, which runs as a dedicated `BigQueryAtomicReplaceLoadJob` that loads all of the table's staged GCS URIs into the destination in a single `WRITE_TRUNCATE_DATA` Load Job (`create_disposition=CREATE_NEVER`, `autodetect=False`). The change is scoped entirely to the BigQuery destination — no changes to filesystem staging, the load orchestration, capabilities, or other destinations.

The diff is larger than the core feature because correctness in the load pipeline required a few specific guards (each covered by tests):

- **No infinite loop:** the aggregated job is a `RunnableLoadJob` *without* `HasFollowupJobs`, so completing it does not re-trigger the table-chain hook (the same pattern `SqlMergeFollowupJob` relies on).
- **No silent data loss:** per-file jobs use `FinalizedLoadJobWithFollowupJobs` (not `FinalizedLoadJob`) so the table-chain hook actually fires; failure of the aggregated job propagates and fails the load.
- **Retry-safe job ids:** the BQ job id includes `load_id` and retry count, so a retried run is not deduplicated to a previous failed job's cached result.
- **URI collection timing:** URIs are scanned from `completed_jobs` at execution time rather than read from `LoadJobInfo` paths at followup-creation time, because the just-completed per-file job is still in `started_jobs` when the hook fires (its `job_to_job_info` path points at `completed_jobs` and would `FileNotFoundError`).
- **Skip the explicit TRUNCATE** for atomic tables (`initialize_storage` / `should_truncate_table_before_load`), since `WRITE_TRUNCATE_DATA` truncates atomically.
- **Local-file fallback:** when the flag is on but files are local (cannot be aggregated into one atomic Load Job), it falls back to regular `truncate-and-insert` with a warning. Non-GCS staging URIs raise a terminal error.

**Notes for review:**

- The client-version requirement is handled by passing the literal string `"WRITE_TRUNCATE_DATA"`. `WriteDisposition` is a plain class of string constants and `write_disposition` is sent to the REST API without client-side validation, so this works without raising the `google-cloud-bigquery` floor (`>=2.26.0`); the backend enforces support.

**Out of scope (possible follow-ups):**

- Making atomic replace the default for `truncate-and-insert`.
- Local-file atomic support (BigQuery cannot aggregate local files into one atomic Load Job).

### Related Issues

- Closes #4012

### Additional Context

The new behavior is **opt-in** (`enable_atomic_replace` defaults to `False`), so existing pipelines are unaffected.

Tests are in `tests/load/bigquery/test_bigquery_atomic_replace.py` and cover the guarantees above (loop prevention, failure propagation, data-loss prevention, load_id scope, retry idempotency, empty-load skip, multi-file aggregation, staging-config gating) without requiring live BigQuery. The metadata-preservation matrix above was verified against live BigQuery separately.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - name: Check out
-        uses: actions/checkout@master
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - uses: pnpm/action-setup@v2
         with:
--- a/.github/workflows/get_docs_changes.yml
+++ b/.github/workflows/get_docs_changes.yml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
           fetch-depth: 0
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,7 +37,7 @@ jobs:
     steps:
 
       - name: Check out
-        uses: actions/checkout@master
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Setup Python
         uses: actions/setup-python@v5
--- a/.github/workflows/test_common.yml
+++ b/.github/workflows/test_common.yml
@@ -85,7 +85,7 @@ jobs:
 
     steps:
       - name: Check out
-        uses: actions/checkout@master
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Setup Python
         uses: actions/setup-python@v5
--- a/.github/workflows/test_destinations_local.yml
+++ b/.github/workflows/test_destinations_local.yml
@@ -134,7 +134,7 @@ jobs:
 
     steps:
       - name: Check out
-        uses: actions/checkout@master
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       #
       # Start required services
--- a/.github/workflows/test_destinations_remote.yml
+++ b/.github/workflows/test_destinations_remote.yml
@@ -223,9 +223,12 @@ jobs:
     steps:
 
       - name: Check out
-        uses: actions/checkout@master
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
+          persist-credentials: false
+          # fork code runs by design, gated by per-run `fork-ci` environment approval
+          allow-unsafe-pr-checkout: true
 
       - name: Setup Python
         uses: actions/setup-python@v5
@@ -237,7 +240,8 @@ jobs:
         with:
           python-version: "3.11"
           activate-environment: true
-          enable-cache: true
+          # fork runs (pull_request_target) must not write the default-branch cache scope
+          enable-cache: ${{ github.event_name != 'pull_request_target' }}
 
       - name: Run pre install commands
         run: ${{ matrix.pre_install_commands }}
@@ -281,9 +285,12 @@ jobs:
     steps:
 
       - name: Check out
-        uses: actions/checkout@master
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
+          persist-credentials: false
+          # fork code runs by design, gated by per-run `fork-ci` environment approval
+          allow-unsafe-pr-checkout: true
 
       - name: Setup Python
         uses: actions/setup-python@v5
@@ -295,7 +302,8 @@ jobs:
         with:
           python-version: "3.11"
           activate-environment: true
-          enable-cache: true
+          # fork runs (pull_request_target) must not write the default-branch cache scope
+          enable-cache: ${{ github.event_name != 'pull_request_target' }}
 
       - name: Install dependencies
         # google-api-python-client for google secrets, botocore (s3 extra) for aws secrets manager
--- a/.github/workflows/test_docs.yml
+++ b/.github/workflows/test_docs.yml
@@ -51,7 +51,7 @@ jobs:
     steps:
 
       - name: Check out
-        uses: actions/checkout@master
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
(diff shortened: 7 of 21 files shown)


oystersuki:
Sorry, I force-pushed only to fix the commit message. No code changes.

rudolfix (dismissed):
@oystersuki thanks for the PR! I identified a few problems in it (for example handling of empty tables that should be just truncated was missing). those got fixed and pushed here.

now this is LGTM

oystersuki:
Thanks for review!
I'm sorry for making you revise it. I should have learned more about the dlt codebase.

PR #4138, opened 2026-06-30
feat(databricks): create tables with IF NOT EXISTS and reconcile columns for concurrent loads

### LINKED ISSUE #4135: feat(databricks): create tables with IF NOT EXISTS and reconcile columns under concurrent loads
## Feature description

When several dlt runs load into the same Databricks table concurrently and the table doesn't exist yet, each run independently decides the table is new and emits a plain `CREATE TABLE`. The first run wins; the others fail with *"table already exists"*. dlt should create user tables with `CREATE TABLE IF NOT EXISTS` on Databricks and reconcile columns so concurrent runs converge.

## Are you a dlt user?

Yes, I run dlt in production.

## Use case

Parallel orchestration (multiple Airflow tasks, parallel pipeline processes, partitioned backfills) regularly lands several runs on the same not-yet-existing table at once. The DDL phase races:

1. Every run calls `get_storage_tables()` → table missing → `generate_alter = False` (`dlt/destinations/job_client_impl.py:688`).
2. Every run emits `CREATE TABLE foo (...)`.
3. First commit wins; the rest raise *"table already exists"* and the load fails.

A subtler second bug hides behind the crash: a run carrying an extra column `c` would see its create become a no-op against the table another run already made, then fail when its load job writes `c` — a column that was never added.

## Proposed solution

Scoped to the Databricks destination:

1. Emit `CREATE TABLE IF NOT EXISTS` for user tables. Today `IF NOT EXISTS` is applied only to dlt internal tables in `_make_create_table` (`dlt/destinations/job_client_impl.py:715`), gated by the existing `supports_create_table_if_not_exists` capability.
2. **Reconcile columns:** after the (possibly no-op) create, re-read the actual table columns and emit `ALTER TABLE … ADD COLUMN` for any missing ones. Databricks has **no** `ADD COLUMN IF NOT EXISTS`, so reconciliation must re-read storage and tolerate the *"column already exists"* error two runs can hit when adding the same column concurrently.

This keeps the common identical-schema case race-free and lets divergent-schema runs converge.

## Related issues

None known.

francescomucio:
Closes #4135

## Problem

When several dlt runs load into the same Databricks table concurrently and the table doesn't exist yet, every run reads storage, sees the table missing (`generate_alter = False`), and emits a plain `CREATE TABLE`. The first commit wins; the rest fail with *"table already exists"*. Hiding behind that crash is a second bug: a run carrying an extra column would see its create become a no-op and then fail when its load writes a column that was never added.

## Change (Databricks only)

- **`CREATE TABLE IF NOT EXISTS` for all tables.** `_make_create_table` is overridden to drop the "dlt system tables only" restriction (still gated by the existing `supports_create_table_if_not_exists` capability). Covers both the base CREATE path and the custom cluster/partition/tblproperties path.
- **Column reconciliation after the create.** Databricks has **no** `ADD COLUMN IF NOT EXISTS`, so `_execute_schema_update_sql` re-reads the destination after the (possibly no-op) create and issues `ALTER TABLE … ADD COLUMN` for any columns still missing, tolerating the `FIELDS_ALREADY_EXISTS` error raised when a concurrent run adds the same column first.

Net effect: identical-schema parallel loads stop racing, and divergent-schema loads converge.

## Notes

- The reconciliation read runs only inside `update_stored_schema` (i.e. when the schema hash actually changes), not on every load.
- Reconciled columns may not appear in the per-load-package `applied_update` diff; this is informational only — the data write maps columns by name against the full stored schema, so the load is unaffected.

## Tests

Added to `tests/load/databricks/test_databricks_table_builder.py` (no live connection needed):
- `CREATE TABLE IF NOT EXISTS` emitted on both create paths.
- Reconciliation adds a column a concurrent create missed.
- `FIELDS_ALREADY_EXISTS` is tolerated; other errors re-raise.

### DIFF
--- a/dlt/destinations/impl/databricks/databricks.py
+++ b/dlt/destinations/impl/databricks/databricks.py
@@ -52,7 +52,7 @@ from dlt.common.destination.exceptions import (
 from dlt.common.exceptions import TerminalValueError
 from dlt.common.typing import TDataRecordBatch
 from dlt.common.schema import Schema, TColumnSchema, TTableSchema
-from dlt.common.schema.typing import TColumnType
+from dlt.common.schema.typing import TColumnType, TSchemaTables
 from dlt.common.schema.utils import get_columns_names_with_prop
 from dlt.common.storages import FilesystemConfiguration, fsspec_from_config
 from dlt.common.storages.configuration import ensure_canonical_az_url
@@ -68,7 +68,11 @@ from dlt.common.storages.load_package import (
     group_jobs_by_table_name,
 )
 from dlt.common.utils import uniq_id
-from dlt.destinations.exceptions import LoadJobTerminalException, LoadJobTransientException
+from dlt.destinations.exceptions import (
+    DatabaseTerminalException,
+    LoadJobTerminalException,
+    LoadJobTransientException,
+)
 from dlt.destinations.file_batching import (
     JsonlFileBatchIterator,
     ParquetFileBatchIterator,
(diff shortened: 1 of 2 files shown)


rudolfix:
@francescomucio the other PR just got merged. closing

PR #4140, opened 2026-06-30
feat(mssql): add Entra ID authentication methods for SQL Server, Synapse, and Fabric

### LINKED ISSUE #2059: Support passwordless authentication to Azure hosted MsSQL
### Feature description

Allow MsSQL destination to support passwordless authentication via a token.

This token is typically obtained using `azure.identity.DefaultAzureCredential()` but it can also be obtained via other methods like OIDC.

### Are you a dlt user?

Yes, I'm already a dlt user.

### Use case

I'm looking to get DLT setup reading/writing to MsSQL running in Azure and am hoping to use a [passwordless connection](https://learn.microsoft.com/en-us/azure/azure-sql/database/azure-sql-passwordless-migration-python?view=azuresql&tabs=sign-in-azure-cli%2Cazure-portal-create%2Cazure-portal-assign%2Capp-service-identity#update-the-local-connection-configuration).

### Proposed solution

Looking at the current implementation, this might be as simple as modifying `MsSqlCredentials` to accept either a username/password or a token. If present, this token would then be included in `pyodbc.connect` via the `attrs_before` argument. If this sounds like an OK approach I am happy to draft a PR.

I experimented with doing a slightly underhanded runtime patch of `pyodbc.connect` with the added token and this seemed to work OK with DLT loading a table quite happily.

### Related issues

_No response_

### LINKED ISSUE #4145: Add Microsoft Entra ID authentication methods to the MS SQL family destinations (SQL Server, Synapse, Fabric)
### Feature description

Add Microsoft Entra ID authentication to the `mssql`, `synapse`, and `fabric` destinations, covering both driver-native methods and token methods where dlt acquires and injects an access token.

### Use case

The MS SQL family destinations authenticate with SQL Server logins (username and password). Many Azure users, and all Fabric Warehouse users, authenticate with Microsoft Entra ID instead. Fabric Warehouse does not accept SQL logins at all, so without Entra ID support the destination cannot connect. Service principal, Azure CLI, environment, and managed-identity authentication are also needed for unattended/CI runs and for setups where no interactive login is available.

### Proposed solution

Add an `authentication` setting that selects one of:

- driver-native methods handled by the ODBC driver: `ActiveDirectoryServicePrincipal`, `ActiveDirectoryPassword`, `ActiveDirectoryIntegrated`, `ActiveDirectoryInteractive`;
- token methods where dlt uses `azure-identity` to acquire an access token and injects it through `attrs_before`, so they work cross-platform: `cli`, `environment`, `interactive`, `devicecode`, `managedidentity`, and `auto`/`default` (DefaultAzureCredential).

For Fabric, default to `ActiveDirectoryServicePrincipal` and fall back to DefaultAzureCredential when no service principal secret is configured.

### Related issues

N/A

sdebruyn:
### Description

Add an `authentication` credential option to `MsSqlCredentials` so the mssql, synapse and fabric destinations can authenticate to Azure-hosted SQL with Microsoft Entra ID instead of a SQL login.

Methods the ODBC driver supports natively are passed straight through as `Authentication=` in the connection string:

- `ActiveDirectoryServicePrincipal` (UID = client id, PWD = client secret)
- `ActiveDirectoryPassword` (UID/PWD)
- `ActiveDirectoryIntegrated`
- `ActiveDirectoryInteractive`
- `ActiveDirectoryMsi`

The two methods the ODBC driver does not support natively are handled by dlt with azure-identity: it acquires the access token itself and injects it via `attrs_before` (`SQL_COPT_SS_ACCESS_TOKEN`), so they work cross-platform (including macOS, where the driver's built-in Entra ID modes are unreliable):

- `ActiveDirectoryDefault` (with the thin alias `default`) — uses `DefaultAzureCredential`, which already covers environment variables, Azure CLI, managed identity, and so on.
- `ActiveDirectoryDeviceCode` — uses `DeviceCodeCredential`.

Empty `authentication` keeps the existing SQL login (backwards compatible). `ActiveDirectoryServicePrincipal` without a secret falls back to a `DefaultAzureCredential` token.

The shared auth helpers live in the mssql destination; synapse inherits them and fabric reuses them, dropping its previous fabric-only token override. azure-identity is imported lazily (only for the token-injection methods).

Adds unit tests for every method on mssql and fabric, a Docker-based integration test for the SQL-login connection path, and documentation.

### Related Issues

- Closes #2059
- Closes #4145

### Additional Context

- All tests ran locally and succeeded.
- I created dbt-fabric where I do the same for dbt Core. I also contributed the same feature to Soda Core.

### DIFF
--- a/dlt/destinations/impl/fabric/configuration.py
+++ b/dlt/destinations/impl/fabric/configuration.py
@@ -26,13 +26,13 @@ class FabricCredentials(AzureServicePrincipalCredentials):
 
     * **Driver-native** (the ODBC driver authenticates): `ActiveDirectoryServicePrincipal`
       (default), `ActiveDirectoryPassword`, `ActiveDirectoryIntegrated`,
-      `ActiveDirectoryInteractive`.
+      `ActiveDirectoryInteractive`, `ActiveDirectoryMsi`.
     * **azure-identity** (dlt acquires an access token and injects it, works cross-platform):
-      `auto`/`default` (DefaultAzureCredential), `cli`, `environment`, `interactive`,
-      `devicecode`, `msi`/`managedidentity`.
+      `ActiveDirectoryDefault` (alias `default`, uses `DefaultAzureCredential`),
+      `ActiveDirectoryDeviceCode` (uses `DeviceCodeCredential`).
 
     When `authentication` is left at its default but no Service Principal secret is configured,
-    dlt falls back to `DefaultAzureCredential` and injects its token.
+    dlt falls back to `ActiveDirectoryDefault` and injects its token.
 
     Inherits from AzureServicePrincipalCredentials for the Service Principal fields.
     """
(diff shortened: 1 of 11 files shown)


sdebruyn:
Closing in favor of #4147, which includes all commits from this PR plus `access_token`/`azure_credential` injection and `NotebookUtilsTokenCredential` for Microsoft Fabric runtime authentication.

PR #4150, opened 2026-07-01
Docs: page for playground workspace

AstrakhantsevaAA:
Explain what is the playground workspace, when and how to use it, in which cases its better to create a new workspace


docs/website/docs/hub/getting-started/playground-workspace.md
```diff
@@ -0,0 +1,51 @@
+---
+title: The Playground workspace
+description: What the Playground workspace is, when to use it, and when to create a dedicated workspace instead.
+keywords: [playground, workspace, onboarding, dlthub platform, profiles]
+---
+
+# The Playground workspace
+
+:::note
+Use of the dltHub platform and toolkits is subject to a commercial dltHub License.
```
martinibach: Without signup, the user can't use the playground workspace on dlthub.com platform so I would remove it here because it's more confusing.

martinibach (dismissed):

martinibach (approved):

PR #4154, opened 2026-07-02
fix: use default format for uv exports

### LINKED ISSUE #4133: dlthub deploy fails: uv export uses removed '--format requirements.txt' value (now 'requirements-txt')
[generated with claude]

## Summary

`dlthub deploy` fails when the workspace has a `uv.lock`, because the dependency export passes an outdated `--format` value to `uv`.

`dlt/_workspace/deployment/requirements.py` calls:

```
uv export --no-hashes --no-emit-project --no-default-groups --format requirements.txt
```

`uv` renamed the `requirements.txt` format value to `requirements-txt`. Current `uv` versions reject the old spelling:

```
error: invalid value 'requirements.txt' for '--format <FORMAT>'
  [possible values: requirements-txt]

  tip: a similar value exists: 'requirements-txt'
```

So the deploy aborts before syncing:

```
Failed to export workspace requirements: `uv export --no-hashes --no-emit-project --no-default-groups --format requirements.txt` failed with exit code 2
```

## Reproduction

1. A uv-managed dlt workspace that has a committed `uv.lock`.
2. `uv` ≥ 0.4.30 on PATH (I tested 0.4.30, 0.5.0, 0.5.31, 0.6.0, 0.6.10 — all reject `requirements.txt`).
3. Run `dlthub deploy` (or `dlthub run <script>.py`, which deploys first).

Deploy fails with the error above.

## Why it isn't always seen

`requirements.py` only calls `uv export` when a `uv.lock` is present. With no lockfile it parses `pyproject.toml` dependencies directly and skips the broken path — so lock-less workspaces deploy fine, which likely masks this in environments that don't commit a lock.

## Environment

- `dlt` 1.28.0 (also present in 1.28.1 — the old spelling is still hardcoded)
- `uv` 0.6.10
- Python 3.12, Linux

## Fix

Rename the two `--format requirements.txt` arguments in `dlt/_workspace/deployment/requirements.py` to `requirements-txt`. (Only the `--format` *values* — not the `REQUIREMENTS_TXT` filename constant.) A targeted local patch of those two lines resolves it; deploy then works with the system `uv` and no shim.

Optionally, detect the installed `uv` version and choose the format value accordingly if older `uv` releases need backward compatibility.

Sam-Razavi:
## Description

Fixes #4133.

`uv export` has used different explicit format values across uv versions. This PR removes the explicit `--format` argument so uv uses its default requirements output format, avoiding the version-specific invalid-format error.

Added regression coverage for the `uv.lock` export path. It verifies that both generated `uv export` commands omit `--format`.

## Related Issues

- Fixes #4133

## Additional Context

Focused validation completed:

- `uv run pytest tests/workspace/deployment/test_requirements.py -q` — 99 passed
- Real `uv.lock` export-path test passed
- Black, Ruff, targeted mypy, `git diff --check`, and `uv lock --check` passed

A broader `tests/workspace` run completed with 1,669 passed and 24 unrelated local Windows/environment failures: missing Streamlit, unavailable symlink privilege, and local file-URL/path handling. The requirements suite passed.

### DIFF
--- a/dlt/_workspace/deployment/requirements.py
+++ b/dlt/_workspace/deployment/requirements.py
@@ -409,8 +409,6 @@ def _export_from_pyproject(
                     "--no-hashes",
                     "--no-emit-project",
                     "--no-default-groups",
-                    "--format",
-                    "requirements.txt",
                 ],
                 cwd=workspace_root,
             )
@@ -425,8 +423,6 @@ def _export_from_pyproject(
                         "--no-default-groups",
                         "--only-group",
                         name,
-                        "--format",
-                        "requirements.txt",
                     ],
                     cwd=workspace_root,
                 )
--- a/tests/workspace/deployment/test_requirements.py
+++ b/tests/workspace/deployment/test_requirements.py
@@ -123,6 +123,43 @@ def test_pyproject_with_lock_resolves_all_groups() -> None:
     assert "numpy" in gpu_names
 
 
+def test_pyproject_with_lock_export_uses_default_uv_format(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    (tmp_path / "pyproject.toml").write_text(
+        "[project]\n"
+        'name = "deps-locked"\n'
+        'version = "0.0.1"\n'
+        'dependencies = ["requests>=2.0"]\n'
+        "\n"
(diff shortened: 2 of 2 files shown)


zilto:
Th relevant issue has already been resolved. Thanks for the contribution though!

PR #4155, opened 2026-07-02
fix(extract,normalize): correct table/resource metric aggregation

tetelio:
## fix(extract,normalize): correct table/resource metric aggregation

`itertools.groupby` was used to aggregate writer metrics by table/resource name,
but it only merges *consecutive* equal keys. With multiple normalize workers or
file rotation, jobs for the same table interleave in `job_metrics`, so
`table_metrics`/`resource_metrics` silently undercounted `items_count` and
other fields for any table split across non-adjacent jobs.

Replaces the groupby-based aggregation with `aggregate_job_metrics`
(`dlt/common/metrics.py`), a small order-independent dict-based grouping
helper, used in both `dlt/extract/extract.py` and `dlt/normalize/normalize.py`.

### Testing
- New unit tests for `aggregate_job_metrics` (`tests/common/test_metrics.py`)
- New extract/normalize/pipeline tests exercising file rotation and multiple
  workers, asserting `table_metrics`/`resource_metrics` match summed job metrics


rudolfix (approved):
LGTM! I pushed a test update

thanks for fixing this, can't believe that survived that long

PR #4179, opened 2026-07-06
fix(filesystem): compute schema update from stored schema for arrow data

### LINKED ISSUE #4156: Load package `schema_update` omits arrow tables on filesystem destination (trace under-reports schema migrations)
## Observation

On a **filesystem** destination, a table loaded from **arrow** data does not appear in the load package's `schema_update` (the per-package schema migration recorded in the trace / load info). The same load on a SQL destination (e.g. duckdb) reports it correctly. This makes the trace under-report which tables/columns were created or altered by a load.

This surfaced for us via `dlthub` materialized transformations (materialization emits arrow, and delta forces materialization), but it reproduces with a plain `pipeline.run(<pyarrow table>)` — no transformations, and independent of `table_format` (plain parquet and delta both hit it).

## Repro

```python
import os, shutil, dlt, pyarrow as pa

def schema_update_tables(pipeline):
        if step.step != "load":
            continue
        for pkg in step.step_info.load_packages:
            return list(pkg.schema_update.keys())

data = pa.table({"id": [1, 2], "name": ["a", "b"]})

# filesystem — user table 'things' is MISSING
W = "/tmp/tkt_fs"; shutil.rmtree(W, ignore_errors=True); os.makedirs(W); os.chdir(W)
fs = dlt.pipeline("p", destination=dlt.destinations.filesystem(bucket_url=f"file://{W}/lake"), dataset_name="d")
fs.run(data, table_name="things")
print("filesystem:", schema_update_tables(fs))

# duckdb — user table 'things' is present
W2 = "/tmp/tkt_duck"; shutil.rmtree(W2, ignore_errors=True); os.makedirs(W2); os.chdir(W2)
dk = dlt.pipeline("p", destination="duckdb", dataset_name="d")
dk.run(data, table_name="things")
print("duckdb    :", schema_update_tables(dk))
```

Output:

```
filesystem: ['_dlt_pipeline_state']
duckdb    : ['_dlt_pipeline_state', '_dlt_loads', '_dlt_version', 'things']
```

The behavior is deterministic (not intermittent) and also affects schema **evolution**: a subsequent arrow load that adds a column to `things` produces an empty `schema_update` (`tables=[]`) on filesystem, while duckdb records the added column. Loading the same data as **dict/JSON rows** (instead of arrow) reports correctly on filesystem too.

Versions: `dlt==1.27.0`, `deltalake==0.25.4`, Python 3.12.11.

## What is expected in the trace

Each load package carries a `schema_update` (serialized as `tables` in the load-package info / trace step info). It is meant to describe the schema migration that this load applied at the destination — the set of tables and columns that were newly created or added by this specific load. Consumers rely on it to answer "what changed at the destination in this run" (e.g. surfacing per-run schema/DDL migrations in observability tooling). For a freshly created data table with a load job, that table (with its columns) should be present in `schema_update`; for an evolved table, the added columns should be present — regardless of whether the data arrived as arrow or JSON, and regardless of destination type.

## Why it isn't working (brief)

The load-package `schema_update` is derived from the diff the **normalize** step produces. For arrow input, the column schema is resolved at **extract** time (from the arrow schema), so normalize sees no change and records an empty diff. SQL destinations recompute the actually-applied update at load time in `update_stored_schema` (via `_execute_schema_update_sql`, diffing the schema against the destination's stored state), so they report the table anyway. The **filesystem** client's `update_stored_schema` instead returns the (empty) `expected_update` verbatim, so the table is never reported.

AnnasMazhar:
## Problem

When loading arrow data to filesystem destinations, the trace shows no tables created even though data is written successfully.

Root cause: arrow data resolves columns at **extract** time. By the time normalize runs, the schema is already up-to-date, so it produces an empty `expected_update`. The filesystem client's `update_stored_schema` returns this verbatim — no schema changes are reported upstream.

SQL destinations (e.g. Postgres, DuckDB) handle this correctly: they ignore `expected_update` and compute the real diff by comparing the current schema against storage (`_execute_schema_update_sql`).

## Solution

When `expected_update` is explicitly empty (not `None`), compute the real `applied_update` by diffing the current schema's tables against the previously stored schema:

- **New tables** → include all columns
- **Existing tables with new columns** → include only the new columns
- **No changes** → return empty dict (no-op)

## Testing

- Existing filesystem client tests pass (9/9, excluding external-bucket tests)
- Arrow source pipeline tests all pass (58/58)
- `_compute_schema_update_from_stored` uses the same `get_new_table_columns` mechanism as the SQL client's `_create_table_update`

Fixes #4156

### DIFF
--- a/dlt/destinations/impl/filesystem/filesystem.py
+++ b/dlt/destinations/impl/filesystem/filesystem.py
@@ -4,6 +4,7 @@ import time as _time
 import orjson
 import base64
 from contextlib import contextmanager
+from copy import copy
 from types import TracebackType
 from typing import (
     ClassVar,
@@ -937,6 +938,14 @@ class FilesystemClient(
                 dirs_to_create = self.get_table_dirs(table_names)
                 for _, directory in zip(table_names, dirs_to_create):
                     self.fs_client.makedirs(directory, exist_ok=True)
+                # When expected_update is empty then normalize did not detect
+                # schema changes (arrow data resolves columns at extract time).
+                # Compute the real update by diffing against previously stored schema.
+                if expected_update is not None and not expected_update:
+                    applied_update = self._compute_schema_update_from_stored(
+                        only_tables
+                    )
+
                 # do not write a duplicate schema file when the hash is already stored
                 if not stored:
                     self._update_schema_in_storage(self.schema)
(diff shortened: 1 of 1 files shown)


AnnasMazhar:
Friendly bump — any updates on this review?

AnnasMazhar:
Friendly bump — any updates on this review? The fix handles the arrow-data empty expected_update edge case in filesystem normalization. No conflicts, ready for review.

AnnasMazhar:
👋 **Third bump** — this one-file fix has been open since July 6 with no reviewer engagement. The fix handles the arrow-data empty expected_update edge case in filesystem normalization (6 lines changed, +57/-0). The related issue #4156 affects users loading arrow data to filesystem destinations. Would appreciate a review when someone has a moment — happy to address any feedback.

AnnasMazhar:
👋 **Fourth bump** — this one-file fix for the arrow-data empty expected_update edge case in filesystem normalization has been open since July 6 (24 days) with no maintainer engagement. No conflicts, tests pass, would appreciate a review when someone has bandwidth.

AnnasMazhar:
Closing due to maintainer inactivity — this one-file fix for the arrow-data empty `expected_update` edge case (#4156) has been open since July 6 with four review bumps and no maintainer engagement. Happy to reopen if there's interest. No hard feelings — the fix is small, conflict-free, and the issue remains open.

PR #4187, opened 2026-07-07
fix: Normalize crashes with ArrowInvalid when adding _dlt_id to an empty Arrow table

### LINKED ISSUE #4175: Normalize crashes with ArrowInvalid when adding _dlt_id to an empty Arrow table (pa.array([]) infers null, not string)
### dlt version

1.28.1 (also reproduced on 1.12.1; still present on `devel` @ `1fe9c29`)

### Describe the problem

When a pipeline loads an **empty** Arrow table or DataFrame (zero rows, valid schema) and the parquet normalizer adds the `_dlt_id` column, the normalize step fails terminally with:

```
pyarrow.lib.ArrowInvalid: Field type did not match data type
```

wrapped in `NormalizeJobFailed` / `PipelineStepFailed`.

**Root cause** (confirmed against `devel` @ `1fe9c29`):

* `dlt/normalize/items_normalizers/arrow.py:101` builds the `_dlt_id` values with `lambda batch: pa.array(generate_dlt_ids(batch.num_rows))` — no explicit `type=`.
* For a zero-row batch, `generate_dlt_ids(0)` returns `[]`, and `pa.array([])` infers Arrow type `null`, not `string`.
* The declared field is `pa.field(c_dlt_id, pa.string(), nullable=False)` (`arrow.py:100`), so `tbl.append_column(field, gen_(tbl))` in `pq_stream_with_new_columns` (`dlt/common/libs/pyarrow.py:911`) raises `ArrowInvalid`.

Notably, the codebase already uses the correct pattern elsewhere: `dlt/common/libs/pyarrow.py:1071` builds a typed empty array (`pa.array([], type=pa.string())`) in the row-tuples inference path.

**Real-world impact:** this isn't only triggered by the `NORMALIZE__PARQUET_NORMALIZER__ADD_DLT_ID` toggle. On destinations that stage parquet, `write_disposition="merge"` adds `_dlt_id` in this same path — so any pipeline whose source can legitimately return zero rows (per-day/per-entity API pulls, incremental windows with no new data) crashes at normalize. We hit this in production loading to Snowflake with merge: a source that returns an empty CSV for one entity every day took the whole asset down. The failure is engine-independent (empty pandas and empty Polars→Arrow both fail identically) and version-independent (reproduced on 1.12.1, 1.28.1, and current `devel`).

### Expected behavior

An empty table with a valid schema should pass through normalize without crashing — either loading as an empty parquet with a typed (empty) `_dlt_id` column, or flowing into dlt's existing empty-table semantics. A zero-row source response is a normal condition for incremental/scheduled pipelines, not an error.

### Steps to reproduce

Self-contained, no credentials needed (duckdb; the env var forces the same `_write_with_dlt_columns` path that Snowflake + merge takes in production):

```python
# repro.py
import os
os.environ["NORMALIZE__PARQUET_NORMALIZER__ADD_DLT_ID"] = "true"

import dlt
import pandas as pd
import polars as pl

print(f"dlt version: {dlt.__version__}")

def run_case(name, data):
    pipeline = dlt.pipeline(
        pipeline_name=f"repro_{name}", destination="duckdb",
        dataset_name=f"repro_{name}", dev_mode=True,
    )
    try:
        pipeline.run(data, table_name="events", write_disposition="append")
        print(f"[{name}] OK")
    except Exception as e:
        print(f"[{name}] FAILED: {type(e).__name__}")
        import traceback; traceback.print_exc()

empty_pl = pl.DataFrame(schema={"id": pl.Int64, "value": pl.Utf8}).to_arrow()
run_case("empty_polars_arrow", empty_pl)

empty_pd = pd.DataFrame({"id": pd.Series(dtype="int64"), "value": pd.Series(dtype="object")})
run_case("empty_pandas", empty_pd)

ok_pd = pd.DataFrame({"id": [1], "value": ["a"]})
run_case("control_nonempty", ok_pd)
```

```
uv run --no-project --with "dlt[duckdb]" --with polars --with pandas --with pyarrow python repro.py
```

Result: both empty cases fail with the traceback below; the non-empty control succeeds.

```
  File ".../dlt/normalize/items_normalizers/arrow.py", line 114, in _write_with_dlt_columns
    for batch in pyarrow.pq_stream_with_new_columns(
  File ".../dlt/common/libs/pyarrow.py", line 911, in pq_stream_with_new_columns
    tbl = tbl.append_column(field, gen_(tbl))
  File "pyarrow/table.pxi", line 2481, in pyarrow.lib._Tabular.append_column
  File "pyarrow/table.pxi", line 5365, in pyarrow.lib.Table.add_column
  File "pyarrow/error.pxi", line 155, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 92, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Field type did not match data type

dlt.normalize.exceptions.NormalizeJobFailed: Job for `job_id=events.61b6fe5373.parquet`
failed terminally in load with `load_id=...` with message: Field type did not match data type.
```

### Operating system

macOS (mechanism is platform-independent; production incident was on Linux)

### Runtime environment

Local (also hit in production under Dagster)

### Python version

3.11

### dlt data source

Any in-memory tabular data: pandas DataFrame or Arrow table (from Polars in our case). Production source was a REST API returning a legitimately empty CSV for some entities.

### dlt destination

DuckDB (repro), Snowflake (production — merge disposition stages parquet and adds `_dlt_id`). The crash happens at normalize, before any destination I/O, so it's destination-agnostic.

### Additional information

* **Frequency:** deterministic — every run where the source yields zero rows on this path.
* **Workaround:** guard before the load (`if len(df) == 0: skip`), which we've rolled out team-wide, but it has to be replicated in every pipeline.
* **Suggested fix:** type the array explicitly in the `_dlt_id` generator at `dlt/normalize/items_normalizers/arrow.py:101` — `pa.array(generate_dlt_ids(batch.num_rows), type=pa.string())` — mirroring the existing typed-empty-array pattern at `dlt/common/libs/pyarrow.py:1071`. Happy to submit a PR with this plus an empty-table test if you agree with the direction (or would prefer skipping the append on zero-row batches instead).
* Related (but distinct) empty-data issues: [#263](<https://github.com/dlt-hub/dlt/issues/263>), dlt-hub/dlt#3998, [#2901](<https://github.com/dlt-hub/dlt/issues/2901>) concern empty-load *semantics* (truncation/materialization); this one is a hard crash before those semantics apply.

nizar-zerrad:
### Related Issues

- Fixes #4175 

### Description
As described in [Issue #4175](https://github.com/dlt-hub/dlt/issues/4175), loading an empty Arrow table or DataFrame with `add_dlt_id=True` causes the normalization step to fail with the following error:
`pyarrow.lib.ArrowInvalid: Field type did not match data type`.

### Root Cause
The error occurs because, when a table is empty, PyArrow automatically infers the data type of the generated ID column as `null` (the `DataType: null` in Arrow) instead of the expected `string`. 

### Solution
This fix ensures that the correct data type is explicitly assigned during the ID generation process, even for empty datasets, preventing the type mismatch.

### DIFF
--- a/dlt/normalize/items_normalizers/arrow.py
+++ b/dlt/normalize/items_normalizers/arrow.py
@@ -98,7 +98,7 @@ class ArrowItemsNormalizer(ItemsNormalizer):
                 (
                     -1,
                     pa.field(data_normalizer.c_dlt_id, pyarrow.pyarrow.string(), nullable=False),
-                    lambda batch: pa.array(generate_dlt_ids(batch.num_rows)),
+                    lambda batch: pa.array(generate_dlt_ids(batch.num_rows),type=pa.string()),
                 )
             )
 
--- a/dlt/normalize/items_normalizers/arrow.py
+++ b/dlt/normalize/items_normalizers/arrow.py
@@ -98,7 +98,7 @@ class ArrowItemsNormalizer(ItemsNormalizer):
                 (
                     -1,
                     pa.field(data_normalizer.c_dlt_id, pyarrow.pyarrow.string(), nullable=False),
-                    lambda batch: pa.array(generate_dlt_ids(batch.num_rows),type=pa.string()),
+                    lambda batch: pa.array(generate_dlt_ids(batch.num_rows), type=pa.string()),
                 )
             )
 
--- a/tests/normalize/test_normalize_arrow.py
+++ b/tests/normalize/test_normalize_arrow.py
@@ -88,13 +88,14 @@ def test_normalize_empty_arrow_input_writes_empty_job(
     step_info = raw_normalize.get_step_info(MockPipeline("arrow_empty_pipeline", True))  # type: ignore[abstract]
     assert step_info.metrics[load_id][0]["table_metrics"]["items"].items_count == 0
 
+
 @pytest.mark.parametrize("add_dlt_id", [True, False])
 def test_normalize_empty_arrow_input_parquet_output(
     parquet_caps: DestinationCapabilitiesContext, raw_normalize: Normalize, add_dlt_id: bool
 ) -> None:
     """An empty arrow input file yields an empty root-table job with a physically present empty
     parquet file (num_rows == 0)."""
-    raw_normalize.config.parquet_normalizer.add_dlt_id= add_dlt_id
+    raw_normalize.config.parquet_normalizer.add_dlt_id = add_dlt_id
     schema = _items_schema()
     empty = pyarrow.table({"id": pyarrow.array([], type=pyarrow.int64())})
     load_id = extract_arrow_items(raw_normalize.normalize_storage, empty, schema, "items")
--- a/tests/normalize/test_normalize_arrow.py
+++ b/tests/normalize/test_normalize_arrow.py
@@ -88,12 +88,13 @@ def test_normalize_empty_arrow_input_writes_empty_job(
     step_info = raw_normalize.get_step_info(MockPipeline("arrow_empty_pipeline", True))  # type: ignore[abstract]
     assert step_info.metrics[load_id][0]["table_metrics"]["items"].items_count == 0
 
-
+@pytest.mark.parametrize("add_dlt_id", [True, False])
 def test_normalize_empty_arrow_input_parquet_output(
-    parquet_caps: DestinationCapabilitiesContext, raw_normalize: Normalize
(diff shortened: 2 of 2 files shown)


rudolfix (approved):
LGTM!

PR #4210, opened 2026-07-13
feat: add client_secret_basic support to OAuth2ClientCredentials

### LINKED ISSUE #4206: Feature Request: Support client_secret_basic authentication method in OAuth2ClientCredentials
#### The Problem

dlt already provides an extensible OAuth authentication framework through `OAuth2AuthBase` and a built-in `OAuth2ClientCredentials` implementation for the OAuth 2.0 Client Credentials flow.

However, `OAuth2ClientCredentials` appears to support only the **`client_secret_post`** client authentication method, where the client credentials are included in the request body. Many OAuth 2.0 authorization servers—including those used by APIs such as eBay, Reddit, and Spotify—require the **`client_secret_basic`** authentication method defined by RFC 6749, where the client credentials are Base64-encoded and sent in the HTTP `Authorization` header.

As a result, integrations with these providers require developers to implement custom authenticators even though the remainder of the OAuth 2.0 Client Credentials flow is already supported by dlt.

#### The Solution

Extend `OAuth2ClientCredentials` with a `client_auth_method` parameter that allows developers to choose how client credentials are transmitted to the authorization server.

#### Proposed Implementation Details

1. **Add Parameter**

   Introduce a `client_auth_method` parameter to the `__init__` method:

   ```python
   client_auth_method: Literal[
       "client_secret_post",
       "client_secret_basic"
   ] = "client_secret_post"
   ```

   Using the OAuth 2.0 specification names keeps the API consistent with RFC 6749 while preserving backwards compatibility.

2. **Basic Authentication Logic**

   When initialized with:

   ```python
   client_auth_method="client_secret_basic"
   ```

   the token request should:

   - Concatenate `client_id` and `client_secret` as `client_id:client_secret`
   - Base64-encode the resulting string
   - Send the credentials in the HTTP header:

   ```text
   Authorization: Basic <base64_encoded_credentials>
   ```

   rather than transmitting them in the request body.

---

### Are you a dlt user?

Yes, I'm already a dlt user.

---

### Use case

I am building a data ingestion pipeline for the eBay Browse Search API using dlt. While configuring authentication, I consistently received `401 Unauthorized` responses despite using valid credentials.

After investigating, I found that the issue is not specific to eBay but rather to the supported OAuth client authentication method.

RFC 6749 defines multiple standard client authentication methods for the Client Credentials flow, including:

1. **`client_secret_post`** — client credentials are sent in the request body.
2. **`client_secret_basic`** — client credentials are sent in the HTTP `Authorization` header using HTTP Basic authentication.

Many OAuth providers—including eBay, Reddit, and Spotify—require `client_secret_basic`. Supporting this authentication method within `OAuth2ClientCredentials` would improve interoperability with standards-compliant OAuth 2.0 authorization servers while eliminating the need for custom implementations for these providers.

---

### Proposed solution

I propose extending `OAuth2ClientCredentials` with a configurable client authentication method.

Example API:

```python
def __init__(
    self,
    access_token_url: str,
    client_id: str,
    client_secret: str,
    access_token_request_data: Optional[Dict[str, Any]] = None,
    default_token_expiration: int = 3600,
    client_auth_method: Literal[
        "client_secret_post",
        "client_secret_basic"
    ] = "client_secret_post",
    session: Optional[requests.Session] = None,
):
```

When:

```python
client_auth_method="client_secret_basic"
```

the class would:

- Concatenate `client_id:client_secret`
- Base64-encode the credentials
- Send them in the HTTP header:

```text
Authorization: Basic <base64_encoded_credentials>
```

instead of transmitting them in the request body.

This approach is fully backwards compatible because the default behavior remains unchanged while enabling compatibility with OAuth providers that require `client_secret_basic`.

---

### Alternatives I have implemented

To unblock my current pipeline, I implemented a custom authenticator by extending `dlt.sources.helpers.rest_client.auth.Auth`. The custom authenticator performs the OAuth token request using `client_secret_basic`, caches the access token, refreshes it when expired, and injects the Bearer token into subsequent API requests.

This confirms that the only missing capability is support for the alternative OAuth client authentication method. Once the token request is performed using `client_secret_basic`, authentication succeeds without requiring any other changes to the OAuth flow.

My current implementation is shown below:

```python

@configspec
class EbayAuth(AuthConfigBase):
    """
    Custom authenticator for eBay OAuth2 Client Credentials flow.
    """
        # OAuth configuration
    client_id: TSecretValue = dlt.secrets.value
    client_secret: TSecretValue = dlt.secrets.value

    token_url: str = dlt.config.value
    scope: str = dlt.config.value
    grant_type: str = dlt.config.value

    marketplace_id: str = dlt.config.value

    token_expiration: int = 7200

        # Runtime state
    _access_token: str = ""
    _token_created_at: float = 0.0

    def _fetch_token(self) -> None:
        """
        Request a new OAuth access token from eBay.
        """

        # Build Basic Authentication credentials
        credentials = f"{self.client_id}:{self.client_secret}"

        encoded_credentials = base64.b64encode(
            credentials.encode("utf-8")
        ).decode("utf-8")

        # OAuth request headers
        headers = {
            "Authorization": f"Basic {encoded_credentials}",
            "Content-Type": "application/x-www-form-urlencoded",
        }

        # OAuth request payload
        payload = {
            "grant_type": self.grant_type,
            "scope": self.scope,
        }

        print("Fetching new eBay OAuth token...")

        response = requests.post(
            url=self.token_url,
            headers=headers,
            data=payload,
        )

        try:
            response.raise_for_status()

        except requests.HTTPError:
            print("OAuth Error:")
            print(response.text)
            raise

        token = response.json()

        self._access_token = token["access_token"]
        self._token_created_at = time.time()

        print("OAuth token acquired successfully.")

    def _is_token_expired(self) -> bool:
        """
        Check whether the current access token has expired.
        """

        # No token has been requested yet
        if self._access_token is None:
            return True

        # Calculate token age
        token_age = time.time() - self._token_created_at

        return token_age >= self.token_expiration
    

    def __call__(self, request: PreparedRequest) -> PreparedRequest:
        """
        Attach authentication headers to the outgoing request.
        """

        # Request a new token if required
        if self._is_token_expired():
            self._fetch_token()

        # Attach OAuth access token
        request.headers["Authorization"] = (
            f"Bearer {self._access_token}"
        )

        # Attach eBay marketplace header
        request.headers["X-EBAY-C-MARKETPLACE-ID"] = (
            self.marketplace_id
        )

        return request

---

### Related issues

_No response_

Mansoor-Ali-Data:
### Description
This PR introduces a client_auth_method parameter to the OAuth2ClientCredentials class, enabling support for the client_secret_basic authentication method defined in RFC 6749.

Currently, the class defaults to client_secret_post (sending credentials in the payload body). However, many major OAuth 2.0 authorization servers (e.g., eBay, Reddit, Spotify) strictly require credentials to be Base64-encoded and sent in the HTTP Authorization header as a Basic token.

By adding this configuration, developers can natively authenticate with these APIs without needing to write custom authenticators. The new parameter defaults to "client_secret_post", ensuring 100% backwards compatibility for existing dlt pipelines

### Related Issues

- Resolves #4206 

### Additional Context

- Implemented explicit string casting (str(self.client_id)) during the Base64 encoding step to ensure that dlt configuration objects (like TSecretStrValue) are properly evaluated before text transformation.
- Tested locally with an eBay Browse Search API pipeline, successfully resolving previous 401 Unauthorized errors without the need for an external custom Auth class.
- I have read the [Contributing to dlt](https://github.com/dlt-hub/dlt/compare/CONTRIBUTING.md) guide and verified that the changes run cleanly.

**Test Output:**
Successfully tested locally against the eBay Browse API:
Pipeline ebay_browse_ingestion load step completed in 6.34 seconds
1 load package(s) were loaded to destination filesystem and into dataset ebay
Load package is LOADED and contains no failed jobs

### DIFF
--- a/dlt/sources/helpers/rest_client/auth.py
+++ b/dlt/sources/helpers/rest_client/auth.py
@@ -161,6 +161,7 @@ class OAuth2ClientCredentials(OAuth2AuthBase):
     client_secret: TSecretStrValue = None
     access_token_request_data: Dict[str, Any] = None
     default_token_expiration: int = 3600
+    client_auth_method: Literal["client_secret_post", "client_secret_basic"] = "client_secret_post"
     session: Annotated[BaseSession, NotResolved()] = None
 
     def __post_init__(self) -> None:
@@ -191,18 +192,29 @@ class OAuth2ClientCredentials(OAuth2AuthBase):
         self.token_expiry = pendulum.now().add(seconds=expires_in_seconds)
 
     def build_access_token_request(self) -> Dict[str, Any]:
-        return {
-            "headers": {
-                "Content-Type": "application/x-www-form-urlencoded",
-            },
-            "data": {
-                "client_id": self.client_id,
-                "client_secret": self.client_secret,
-                "grant_type": "client_credentials",
-                **self.access_token_request_data,
-            },
+        data = {
+            "grant_type": "client_credentials",
+            **self.access_token_request_data,
+        }
+        headers = {
+            "Content-Type": "application/x-www-form-urlencoded",
         }
 
+        client_id_str = str(self.client_id)
+        client_secret_str = str(self.client_secret)
+
+        if self.client_auth_method == "client_secret_basic":
+            credentials = f"{client_id_str}:{client_secret_str}"
+            encoded_credentials = b64encode(credentials.encode("utf-8")).decode("utf-8")
+            headers["Authorization"] = f"Basic {encoded_credentials}"
+        else:
+            data["client_id"] = client_id_str
+            data["client_secret"] = client_secret_str
+        return{
+            "headers": headers,
+            "data": data,
+        }    
+    
     def parse_expiration_in_seconds(self, response_json: Any) -> int:
         return int(response_json.get("expires_in", self.default_token_expiration))
 
--- a/dlt/sources/helpers/rest_client/auth.py
+++ b/dlt/sources/helpers/rest_client/auth.py
@@ -210,11 +210,11 @@ class OAuth2ClientCredentials(OAuth2AuthBase):
         else:
             data["client_id"] = client_id_str
             data["client_secret"] = client_secret_str
-        return{
+        return {
             "headers": headers,
             "data": data,
-        }    
-    
+        }
+
     def parse_expiration_in_seconds(self, response_json: Any) -> int:
         return int(response_json.get("expires_in", self.default_token_expiration))
 


Mansoor-Ali-Data:
Hi @burnash! , Whenever you have some time, I'd really appreciate it if you could review this PR. Thank you!

PR #4211, opened 2026-07-13
Add emphemeral storage caveat

tetelio:
This PR extends the error message from a load package so that its warned that machines that fail with an ephemeral storage will not retry the load package on next run. This is because files will not exist on disk (this includes how dlthub platform currently works).


tetelio:
rendered message example 
image

rudolfix (approved):
LGTM!

PR #4215, opened 2026-07-13
fix: remove usage of `ModuleType.__dict__` infringing on PEP 562

zilto:
resolves issues: #4214 

After these few lines changed, code for my use case works. Let's see if tests fail

PR #4216, opened 2026-07-13
fix(workspace): discover __all__ jobs exposed via module __getattr__ (PEP 562)

### LINKED ISSUE #4214: bug: workspace manifest uses `__dict__.get()` instead of `getattr()`
The function `generate_manifest()` in `dlt/_workspace/deployment/manifest.py` doesn't respect PEP 562 by using `module.__dict__.get()`. It should be using `module.__getattr__()`, which respects the Python import logic and allows downstream users to override a module's `__getattr__()` function.

See official Python docs: https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access

This allows to create proxy deployment modules to dynamically override job attributes (e.g., adding a `timeout` or editing the `trigger`) when doing `dlthub deploy` 

## Use case
I have a GitOps project where I deploy to dltHub via a `__deployment__.py` for prod and a `__staging__.py` to staging.

vakudo:
Fixes #4214.

`generate_manifest()` in `dlt/_workspace/deployment/manifest.py` resolved each `__all__` entry with `deployment_module.__dict__.get(name)`. When a deployment module exposes its jobs lazily through a module-level `__getattr__` (PEP 562), those names aren't in `__dict__`, so they came back as `None`, produced a spurious "listed in `__all__` but not found in module" warning, and were dropped from the manifest.

Switched to `getattr(deployment_module, name, None)`, which respects the import protocol. The `__dict__`-scan path is unaffected since those names are always real attributes.

Added a regression test (`test_lazy_all_via_module_getattr`) and a `deployment_lazy` case module that exports its jobs via `__getattr__`. The test fails on the old code and passes with the fix.

Ran locally on the changed files: ruff, flake8 and mypy are clean, and the manifest-generator tests pass (the framework-detection tests need the marimo/streamlit/fastmcp extras and are unrelated to this change).

### DIFF
--- a/dlt/_workspace/deployment/manifest.py
+++ b/dlt/_workspace/deployment/manifest.py
@@ -517,7 +517,9 @@ def generate_manifest(
             names = list(deployment_module.__dict__.keys())
 
         for name in names:
-            obj = deployment_module.__dict__.get(name)
+            # use getattr so names exposed via module-level __getattr__ (PEP 562)
+            # are resolved, not just those present in __dict__
+            obj = getattr(deployment_module, name, None)
             if obj is None and iterate_all:
                 warnings.append(f"name {name!r} listed in __all__ but not found in module")
                 continue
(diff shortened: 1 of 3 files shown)


rudolfix:
@vakudo this issue was already solved by the original submitter in #4215

PR #4220, opened 2026-07-14
fix(filesystem): append extension when layout omits ext

### LINKED ISSUE #4219: Filesystem layouts without `{ext}` create unreadable extensionless files
### dlt version

1.29.0 (reproduced from `devel` at `43d4159b23bdd47c82c919f89daf295911b26dd6`)

### Describe the problem

When the filesystem destination uses a custom `layout` that omits the `{ext}` placeholder, the loaded data file is written without its format extension.

For example, this layout:

```python
layout="{table_name}/{load_id}.{file_id}"
```

produces a path such as:

```text
dataset/events/1783932226.077965.a04e24b9b5
```

The file contains JSONL data but has no `.jsonl` or `.jsonl.gz` suffix. A subsequent read through `pipeline.dataset()` cannot materialize the table and raises `dlt.destinations.exceptions.DatabaseUndefinedRelation`.

### Expected behavior

`create_path` explicitly has a fallback for layouts without `{ext}`: its comment says the extension is appended automatically. The generated path should therefore retain the custom layout and end in the full loader extension, for example:

```text
dataset/events/1783932226.077965.a04e24b9b5.jsonl
```

The resulting filesystem table should remain readable through `pipeline.dataset()`.

### Steps to reproduce

```python
import pathlib
import tempfile

import dlt

with tempfile.TemporaryDirectory() as tmp:
    root = pathlib.Path(tmp)
    pipeline = dlt.pipeline(
        pipeline_name="missing_ext",
        destination=dlt.destinations.filesystem(
            bucket_url=str(root / "bucket"),
            layout="{table_name}/{load_id}.{file_id}",
        ),
        dataset_name="dataset",
    )

    pipeline.run([{"id": 1}], table_name="events")

    data_files = sorted((root / "bucket" / "dataset" / "events").iterdir())
    print(data_files)  # the data file has no .jsonl/.jsonl.gz extension

    print(pipeline.dataset()["events"].fetchall())
    # dlt.destinations.exceptions.DatabaseUndefinedRelation:
    # Catalog Error: Table with name events does not exist
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

Inline Python list containing one dictionary.

### dlt destination

Filesystem & buckets

### Other deployment details

Reproduced with a local temporary filesystem bucket. No credentials or external services are required.

### Additional information

The likely root cause is in `dlt/destinations/path_utils.py::create_path`:

- `check_layout` returns `(all_valid_placeholders, placeholders_present_in_layout)`.
- `create_path` assigns the first value to `placeholders`, so `"ext" not in placeholders` is always false because `ext` is a valid standard placeholder.
- The fallback body currently assigns `path = job_info.full_extension()`; once reachable, it should append the full extension to the formatted path instead of replacing the path.

A focused regression test can extend `tests/destinations/test_path_utils.py::test_create_path` with a layout that actually omits `{ext}`. The current “extension gets added automatically” case still includes `{ext}`, so it does not exercise this path.

I searched the open issue tracker using the filesystem layout, `{ext}`, `create_path`, missing-extension, JSONL-extension, and observed-error terms. I also checked all currently open pull requests and none changes `dlt/destinations/path_utils.py` or its path utility tests. I would be happy to contribute the focused fix and regression test.

mattfaltyn:
### Description

Fix filesystem path generation when a layout omits `{ext}`. `create_path` was checking all allowed placeholders instead of the placeholders present in the layout, making the documented extension fallback unreachable. It now preserves the formatted path and appends the job's full extension.

Update the existing layout expectations and cover both uncompressed and compressed job names so `.jsonl.gz` remains intact.

### Related Issues

- Fixes #4219

### Additional Context

Validation:
- `make format-check` (1,269 files unchanged)
- `make lint` (mypy, Ruff, Flake8, Bandit, lockfile, dependency, extras-version, and API compatibility checks passed)
- Full `make test-common` scope with no exclusions on Python 3.11.12 (7,050 parallel-safe tests passed with 57 expected skips; all 213 serial/fork-sensitive tests passed)
- `pytest tests/destinations/test_path_utils.py -q` (70 passed)
- End-to-end filesystem pipeline write and dataset readback with a layout omitting `{ext}`

### DIFF
--- a/dlt/destinations/path_utils.py
+++ b/dlt/destinations/path_utils.py
@@ -235,12 +235,12 @@ def create_path(
     datetime_params = prepare_datetime_params(current_datetime, load_package_timestamp)
     params.update(datetime_params)
 
-    placeholders, _ = check_layout(layout, params)
+    _, placeholders = check_layout(layout, params)
     path = layout.format(**params)
 
     # if extension is not defined, we append it at the end
     if "ext" not in placeholders:
-        path = job_info.full_extension()
+        path += f".{job_info.full_extension()}"
 
     return path
 
--- a/tests/destinations/test_path_utils.py
+++ b/tests/destinations/test_path_utils.py
@@ -250,29 +250,34 @@ ALL_LAYOUTS = (  # type: ignore
         True,
         [],
     ),
-    ("{timestamp}{table_name}", f"{int(frozen_datetime.timestamp())}mocked-table", True, []),
-    ("{ddd}/{MMM}/{table_name}", "sun/apr/mocked-table", True, []),
+    ("{timestamp}{table_name}", f"{int(frozen_datetime.timestamp())}mocked-table.jsonl", True, []),
+    ("{ddd}/{MMM}/{table_name}", "sun/apr/mocked-table.jsonl", True, []),
     (
         "{Y}/{timestamp}/{table_name}",
-        f"{frozen_datetime.format('YYYY')}/{int(frozen_datetime.timestamp())}/mocked-table",
+        f"{frozen_datetime.format('YYYY')}/{int(frozen_datetime.timestamp())}/mocked-table.jsonl",
         True,
         [],
     ),
     (
         "{Y}/{timestamp_ms}/{table_name}",
-        f"{frozen_datetime.year}/{str(int(frozen_datetime.timestamp()*1000))}/mocked-table",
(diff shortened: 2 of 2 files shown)


Travior:
Thanks for the contribution. The fix looks sound. I'll let CI run on the PR so we can get this merged after

Travior (approved):
LGTM

PR #4226, opened 2026-07-15
fix(filesystem): apply extra_placeholders in get_table_prefix

### LINKED ISSUE #4224: Filesystem: `replace` silently deletes nothing when `extra_placeholders` overrides `{schema_name}`
### dlt version

1.29.0

### Describe the problem

When `extra_placeholders` overrides `{schema_name}`, the write path uses the overridden value but the truncate/list path still uses the real schema name. With `write_disposition="replace"` this means nothing is ever deleted. Every run appends another set of files, readers see duplicated rows, and there's no error or warning at any point.

The filesystem destination accepts `extra_placeholders={"schema_name": ...}`, and the write path honours it - [`prepare_params`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/path_utils.py#L156-L165) applies extra placeholders after the standard ones ([`schema_name`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/path_utils.py#L151) is set first), so the override wins and files land under the substituted value. This is the only way to customise the pre-table path segment while staying replace-compatible, since replace layouts only permit `{schema_name}` before `{table_name}`.

The write path picked up extra placeholder support in #1182 (feature request #930) but [`get_table_prefix`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/impl/filesystem/filesystem.py#L985) never did. It formats the layout with the schema name directly:

```python
table_prefix = self.table_prefix_layout.format(
    schema_name=schema_name or self.schema.name, table_name=table_name
)
```

So the write path and the table prefix disagree on the directory. Everything keyed off the prefix (replace truncation, `list_table_files`, dataset reads) points at a directory that never receives files.

It fails silently: [`check_layout`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/path_utils.py#L176) accepts the config, the write succeeds, and [`truncate_tables`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/impl/filesystem/filesystem.py#L771) matches zero files and returns.

### Expected behavior

Two possible fixes - either honour the override consistently or reject it loudly:

1. [`get_table_prefix`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/impl/filesystem/filesystem.py#L985) resolves `extra_placeholders` the same way [`create_path`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/path_utils.py#L212) does. The prefix can only contain `{schema_name}` ([`get_table_prefix_layout`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/path_utils.py#L248) enforces that), and the [callback contract](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/path_utils.py#L156-L165) is `(schema_name, table_name, load_id, file_id, ext)`, so it can run at prefix time with the last three as `None`. Constants just substitute.
2. If shadowing a standard placeholder is out of contract, [`check_layout`](https://github.com/dlt-hub/dlt/blob/a2af6432b687c649d0a944e568b58cb37899c4c9/dlt/destinations/path_utils.py#L176) should raise when an `extra_placeholders` key collides with a standard placeholder, rather than accepting a config where the write path and the delete path disagree.

Accepting the config silently and then diverging is the worst of both as the failure mode is data duplication with no signal.

### Steps to reproduce

Two replace runs of the same data, once without the override and once with it. Without → 1 file, truncation works. With → 2 files under the overridden path, truncation matches nothing.

```python
import glob, os, tempfile
import dlt
def run_twice(extra_placeholders, label):
    root = tempfile.mkdtemp(prefix=f"repro-{label}-")
    os.environ["DLT_DATA_DIR"] = tempfile.mkdtemp(prefix=f"dltdir-{label}-")
    dest = filesystem(
        bucket_url=f"file://{root}",
        layout="{schema_name}/{table_name}/{load_id}.{file_id}.{ext}",
        extra_placeholders=extra_placeholders,
    )
    for _ in range(2):
        p = dlt.pipeline(pipeline_name=f"repro_{label}", destination=dest, dataset_name="ds")
        p.run([{"id": 1}, {"id": 2}], table_name="items",
              write_disposition="replace", loader_file_format="parquet")
    files = glob.glob(f"{root}/ds/**/items/*.parquet", recursive=True)
    print(f"{label}: {len(files)} file(s) after two replace runs")
run_twice(None, "no_override")                  # -> 1 file
run_twice({"schema_name": "prod"}, "override")  # -> 2 files, both under ds/prod/items/
```

Expected output:

```
no_override: 1 file(s) after two replace runs
override: 2 file(s) after two replace runs
```

### Operating system

Linux, macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

_No response_

### dlt destination

Filesystem & buckets

### Other deployment details

_No response_

### Additional information

_No response_

axelray-dev:
## Summary

When `extra_placeholders` overrides `{schema_name}`, filesystem writes land under the overridden segment, but `get_table_prefix` still used the real schema name. Replace truncation and table listing therefore matched zero files and silently left old data in place.

## Changes

- Resolve `schema_name` for `get_table_prefix` through `path_utils.prepare_params` so constant and callable `extra_placeholders` overrides match the write path.
- Pass only resolved `schema_name` and the method `table_name` into the prefix layout format, so unrelated extra placeholders do not break prefix formatting.
- Allow optional `table_name` on `prepare_params` so prefix-time callables receive the table name (with load/file fields as `None`).
- Add a regression test for constant and callable schema overrides with an unrelated extra placeholder present.

## Verification

- `pytest tests/load/filesystem/test_filesystem_client.py::test_table_prefix_resolves_schema_name_extra_placeholder -q`
- `pytest tests/destinations/test_path_utils.py -q` (path utils suite)

Fixes #4224

### DIFF
--- a/dlt/destinations/path_utils.py
+++ b/dlt/destinations/path_utils.py
@@ -130,9 +130,9 @@ def prepare_params(
     job_info: Optional[ParsedLoadJobFileName] = None,
     schema_name: Optional[str] = None,
     load_id: Optional[str] = None,
+    table_name: Optional[str] = None,
 ) -> Dict[str, Any]:
     params: Dict[str, Any] = {}
-    table_name = None
     file_id = None
     ext = None
     if job_info:
@@ -146,6 +146,8 @@ def prepare_params(
                 "ext": ext,
             }
         )
+    elif table_name:
+        params["table_name"] = table_name
 
     if schema_name:
         params["schema_name"] = schema_name
--- a/dlt/destinations/impl/filesystem/filesystem.py
+++ b/dlt/destinations/impl/filesystem/filesystem.py
@@ -990,14 +990,20 @@ class FilesystemClient(
             # it is crucial to append and keep "/" at the end
             table_prefix = self.pathlib.join(table_name, "")
         else:
+            prefix_placeholders = set(path_utils.get_placeholders(self.table_prefix_layout))
+            prefix_extra_placeholders = {
+                key: value
+                for key, value in (self.config.extra_placeholders or {}).items()
+                if key in prefix_placeholders
+            }
             params = path_utils.prepare_params(
-                extra_placeholders=self.config.extra_placeholders,
+                extra_placeholders=prefix_extra_placeholders,
                 schema_name=schema_name or self.schema.name,
                 table_name=table_name,
             )
             table_prefix = self.table_prefix_layout.format(
                 schema_name=params.get("schema_name", schema_name or self.schema.name),
-                table_name=table_name,
+                table_name=params.get("table_name", table_name),
             )
         return self.pathlib.join(  # type: ignore[no-any-return]
             self.dataset_path, path_utils.normalize_path_sep(self.pathlib, table_prefix)
--- a/dlt/destinations/impl/filesystem/filesystem.py
+++ b/dlt/destinations/impl/filesystem/filesystem.py
@@ -990,8 +990,14 @@ class FilesystemClient(
             # it is crucial to append and keep "/" at the end
             table_prefix = self.pathlib.join(table_name, "")
         else:
+            params = path_utils.prepare_params(
+                extra_placeholders=self.config.extra_placeholders,
+                schema_name=schema_name or self.schema.name,
+                table_name=table_name,
+            )
             table_prefix = self.table_prefix_layout.format(
-                schema_name=schema_name or self.schema.name, table_name=table_name
+                schema_name=params.get("schema_name", schema_name or self.schema.name),
+                table_name=table_name,
             )
         return self.pathlib.join(  # type: ignore[no-any-return]
             self.dataset_path, path_utils.normalize_path_sep(self.pathlib, table_prefix)
--- a/tests/load/filesystem/test_filesystem_client.py
+++ b/tests/load/filesystem/test_filesystem_client.py
@@ -159,6 +159,37 @@ def test_trailing_separators(layout: str, with_gdrive_buckets_env: str) -> None:
         assert client.get_table_prefix("letters").endswith("_data/letters.")
 
 
+def test_table_prefix_resolves_schema_name_extra_placeholder() -> None:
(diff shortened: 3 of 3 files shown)


dlt/destinations/path_utils.py
```diff
@@ -130,9 +130,9 @@ def prepare_params(
     job_info: Optional[ParsedLoadJobFileName] = None,
     schema_name: Optional[str] = None,
     load_id: Optional[str] = None,
+    table_name: Optional[str] = None,
```
rudolfix, line 133: why this is needed? do you also plan to fix `table_name` override?

rudolfix (changes requested):
LGTM just one question about `table_name` being passed. thanks!

dlt/destinations/path_utils.py
axelray-dev: Good catch. Passing `table_name` is needed so callable overrides receive the same callback context as the write path, but the previous version did not apply a resolved `table_name` override. I updated `get_table_prefix()` to honor both `schema_name` and `table_name` overrides. It now resolves only placeholders used by the table prefix, so unrelated callbacks are not invoked without load/file metadata. I also added regression coverage for constant and callable overrides.

rudolfix (approved):
LGTM!

PR #4227, opened 2026-07-15
fix(rest_client): preserve paginator stop conditions

### LINKED ISSUE #4225: RangePaginator: has_more=true overrides maximum pagination limits
## dlt version

1.29.0 (`devel` at `a2af6432b`)

## Describe the problem

`RangePaginator.update_state()` correctly sets `has_next_page=False` when `maximum_offset`, `maximum_page`, or the response total has been reached. However, when `has_more_path` is also configured and the API returns `has_more=true`, the method then unconditionally overwrites that result with `True`.

This affects both `OffsetPaginator` and `PageNumberPaginator`. It causes an additional request beyond an explicitly configured hard limit and can lead to unintended over-fetching or extra API usage.

## Expected behavior

An explicit maximum should remain a hard upper bound, as documented. `has_more=false` should be able to stop pagination early, but `has_more=true` should not re-enable pagination after a maximum or total-based stop condition has already been reached.

## Steps to reproduce

No external API is required:

```python
from unittest.mock import Mock
from requests import Response

from dlt.sources.helpers.rest_client.paginators import (
    OffsetPaginator,
    PageNumberPaginator,
)

response = Mock(Response, json=lambda: {"has_more": True})

offset = OffsetPaginator(
    limit=10,
    offset=0,
    maximum_offset=10,
    total_path=None,
    has_more_path="has_more",
)
offset.update_state(response, data=[{"id": 1}])
print(offset.current_value, offset.has_next_page)

page = PageNumberPaginator(
    base_page=1,
    page=1,
    maximum_page=2,
    total_path=None,
    has_more_path="has_more",
)
page.update_state(response, data=[{"id": 1}])
print(page.current_value, page.has_next_page)
```

Actual output:

```text
10 True
2 True
```

Expected output:

```text
10 False
2 False
```

## Operating system

macOS

## Runtime environment

Local

## Python version

3.13

## dlt data source

REST API / `dlt.sources.helpers.rest_client`

## dlt destination

Not destination-specific.

## Other deployment details

The reproduction exercises paginator state directly and is platform-independent.

## Additional information

The conflicting logic is in `RangePaginator.update_state()`: the maximum/total check sets `_has_next_page = False`, after which the `has_more_path` branch assigns `self._has_next_page = has_more`.

Related work reviewed:

- #2816 and #2817 introduced `has_more_path`, but do not discuss its precedence relative to explicit maximum values.
- Open PR #4186 modifies the same method only to handle non-scalar total values; it does not address this interaction.
- Existing tests cover maximum values and `has_more_path` independently, but not together.

A minimal fix would preserve any existing stop decision and only allow `has_more=false` to stop pagination, rather than allowing `true` to re-enable it.

mattfaltyn:
### Description

Preserve existing stop decisions when an API returns `has_more=true`.

For numeric range pagination, `RangePaginator.update_state()` first stops at `maximum_offset`, `maximum_page`, or the response total. For cursor pagination, assigning a missing cursor stops `JSONResponseCursorPaginator`. Both methods previously replaced those decisions with the API's `has_more` flag, which could request another page without a valid cursor or beyond a hard limit.

`has_more=false` can still stop pagination early; `true` can no longer re-enable a paginator that has already reached another stop condition. Regression tests cover maximum limits, total-based limits, and missing cursors.

### Related Issues

- Fixes #4225

### Additional Context

Validation:

- `make lint`
- common parallel-safe tests: 7,051 passed, 57 skipped
- non-forked serial tests: 192 passed, 1 skipped
- forked tests: 21 passed, 1 skipped
- `.venv/bin/pytest tests/sources/helpers/rest_client/test_paginators.py -q` (100 passed)
- Black, Ruff, and mypy on the changed files

On macOS, forked tests were run in a fresh pytest process with `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES`.

### DIFF
--- a/tests/sources/helpers/rest_client/test_paginators.py
+++ b/tests/sources/helpers/rest_client/test_paginators.py
@@ -975,6 +975,12 @@ class TestJSONResponseCursorPaginator:
             excinfo.value
         )
 
+    def test_update_state_when_cursor_path_is_empty_string(self):
+        paginator = JSONResponseCursorPaginator(cursor_path="next_cursor")
+        response = Mock(Response, json=lambda: {"next_cursor": "", "results": []})
+        paginator.update_state(response)
+        assert paginator.has_next_page is False
+
     def test_has_more_does_not_override_missing_cursor(self):
         paginator = JSONResponseCursorPaginator(cursor_path="next_cursor", has_more_path="has_more")
         response = Mock(Response, json=lambda: {"next_cursor": "", "results": [], "has_more": True})
--- a/tests/sources/helpers/rest_client/test_paginators.py
+++ b/tests/sources/helpers/rest_client/test_paginators.py
@@ -349,6 +349,20 @@ class TestRangePaginator:
 
         assert paginator.has_next_page is False
 
+    def test_has_more_does_not_override_total(self) -> None:
+        paginator = RangePaginator(
+            param_name="page",
+            initial_value=0,
+            value_step=1,
+            total_path="total",
+            has_more_path="has_more",
+        )
+        response = Mock(Response, json=lambda: {"total": 1, "has_more": True})
+
+        paginator.update_state(response, data=NON_EMPTY_PAGE)
+
+        assert paginator.has_next_page is False
+
 
 @pytest.mark.usefixtures("mock_api_server")
 class TestOffsetPaginator:
--- a/tests/sources/helpers/rest_client/test_paginators.py
+++ b/tests/sources/helpers/rest_client/test_paginators.py
@@ -335,6 +335,20 @@ class TestRangePaginator:
         with pytest.raises(ValueError, match="body_path.*must not be empty"):
             RangePaginator._update_request_with_body_path(request, "", 0)
 
+    def test_has_more_does_not_override_maximum_value(self) -> None:
+        paginator = RangePaginator(
+            param_name="page",
+            initial_value=0,
+            value_step=1,
+            maximum_value=1,
+            has_more_path="has_more",
+        )
+        response = Mock(Response, json=lambda: {"has_more": True})
+
+        paginator.update_state(response, data=NON_EMPTY_PAGE)
+
+        assert paginator.has_next_page is False
+
 
 @pytest.mark.usefixtures("mock_api_server")
 class TestOffsetPaginator:
--- a/tests/sources/helpers/rest_client/test_paginators.py
+++ b/tests/sources/helpers/rest_client/test_paginators.py
@@ -975,9 +975,9 @@ class TestJSONResponseCursorPaginator:
             excinfo.value
         )
 
-    def test_update_state_when_cursor_path_is_empty_string(self):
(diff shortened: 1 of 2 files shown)


burnash (dismissed):
@mattfaltyn thanks for spotting this and for the fix.

burnash:
@mattfaltyn there's a similar issue with JSONResponseCursorPaginator.update_state. Would you be willing to apply the fix there as well in this PR? (all good if not).

mattfaltyn:
Thanks @burnash - applied the same precedence fix to `JSONResponseCursorPaginator.update_state()` and updated the regression test so `has_more=true` cannot re-enable pagination when no next cursor is returned.

tests/sources/helpers/rest_client/test_paginators.py
```diff
@@ -947,9 +975,9 @@ def test_both_cursor_param_and_cursor_body_path_raises_error(self):
...
         )
 
-    def test_update_state_when_cursor_path_is_empty_string(self):
-        paginator = JSONResponseCursorPaginator(cursor_path="next_cursor")
-        response = Mock(Response, json=lambda: {"next_cursor": "", "results": []})
+    def test_has_more_does_not_override_missing_cursor(self):
+        paginator = JSONResponseCursorPaginator(cursor_path="next_cursor", has_more_path="has_more")
+        response = Mock(Response, json=lambda: {"next_cursor": "", "results": [], "has_more": True})
         paginator.update_state(response)
         assert paginator.has_next_page is False
```
burnash, line 988:
```suggestion
    def test_update_state_when_cursor_path_is_empty_string(self):
        paginator = JSONResponseCursorPaginator(cursor_path="next_cursor")
        response = Mock(Response, json=lambda: {"next_cursor": "", "results": []})
        paginator.update_state(response)
        assert paginator.has_next_page is False

    def test_has_more_does_not_override_missing_cursor(self):
        paginator = JSONResponseCursorPaginator(cursor_path="next_cursor", has_more_path="has_more")
        response = Mock(Response, json=lambda: {"next_cursor": "", "results": [], "has_more": True})
        paginator.update_state(response)
        assert paginator.has_next_page is False
```

let's preserve the original test.

burnash (commented):
@mattfaltyn lgtm, i have a small comment

burnash (approved):
Looks good, thanks so much for fixing this @mattfaltyn

mattfaltyn:
> Looks good, thanks so much for fixing this @mattfaltyn

Happy to help 🙏

mattfaltyn:
@burnash I don’t have permission to rerun CI. The failures appear unrelated to this PR: Windows 3.13 hit microsecond-level timing assertions, and Windows 3.11 hit a transient `WinError 5` during `os.replace`. Both jobs passed in an earlier run of this PR. Could you please rerun the failed/cancelled Windows jobs when you have time?

burnash:
@mattfaltyn yeah, it's a flaky test I'm taking care of it. I'll rerun the failed jobs

PR #4231, opened 2026-07-15
fix(common): build boolean sqlglot literals as sge.Boolean, not numeric

### LINKED ISSUE #4230: build_typed_literal produces malformed numeric literal for bool values, crashing sqlglot's type annotation pass
### dlt version

1.29.0

### Describe the problem

`build_typed_literal()`'s inner `_literal()` helper (`dlt/common/libs/sqlglot.py`) branches on the Python type of a value to build the matching `sqlglot` expression:

```python
if v is None:
    lit = sge.Null()
elif isinstance(v, str):
    lit = sge.Literal.string(v)
elif isinstance(v, (int, float)):
    lit = sge.Literal.number(v)
elif isinstance(v, (bytes, bytearray)):
    ...
```

There is no dedicated branch for `bool`. Since `bool` is a subclass of `int` in Python, a `True`/`False` value falls into the `isinstance(v, (int, float))` branch and gets built as `sge.Literal.number(v)` — i.e. a numeric literal whose text is `str(True)` / `str(False)` (`"True"` / `"False"`), which is not valid numeric literal text.

This literal round-trips fine through plain SQL rendering, but it crashes as soon as it goes through `sqlglot`'s type-annotation optimizer pass (`sqlglot.optimizer.annotate_types.annotate_types`), which `Relation.where()`/`.filter()` rely on for type-aware SQL generation:

```
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
```

### Expected behavior

A `bool` value passed to `build_typed_literal()` should produce `sge.Boolean(this=v)`, matching how `sqlglot` represents boolean literals natively, and should survive `annotate_types()` without error.

### Steps to reproduce

```python
import sqlglot
import sqlglot.expressions as sge
from sqlglot.optimizer.annotate_types import annotate_types
from dlt.common.libs.sqlglot import build_typed_literal

condition = sge.EQ(this=sge.column("is_active"), expression=build_typed_literal(True))
select = sqlglot.select("*").from_("my_table").where(condition)

annotate_types(select)  # raises decimal.InvalidOperation
```

I have a fix ready (dedicated `isinstance(v, bool)` branch ahead of the `int`/`float` branch, producing `sge.Boolean(this=v)`) and will open a PR shortly.

---
Disclosure: this issue was drafted with AI assistance (Claude) after independently reproducing the bug end-to-end; I reviewed and take responsibility for the report.

chuenchen309:
### Description

`build_typed_literal()`'s inner `_literal()` helper had no dedicated branch for `bool`. Since `bool` is an `int` subclass, `True`/`False` fell into the `isinstance(v, (int, float))` branch and were built as `sge.Literal.number(v)` — a malformed numeric literal (text `"True"`/`"False"`) that crashes with `decimal.InvalidOperation` as soon as it goes through sqlglot's `annotate_types()` optimizer pass (used by `Relation.where()`/`.filter()`).

Fix: dedicated `isinstance(v, bool)` branch ahead of the `int`/`float` branch, producing `sge.Boolean(this=v)`.

Added two tests in `tests/libs/test_sqlglot.py`:
- `build_typed_literal(True/False)` produces `sge.Boolean` with the correct `.this`.
- A `WHERE` clause built from a bool literal survives `annotate_types()` (previously raised `decimal.InvalidOperation`).

Verified red (crash) before the fix, green after; full `tests/libs/test_sqlglot.py` suite passes (269 passed); `uv run ruff check` and `black --check` clean.

---
Disclosure: this PR was drafted with AI assistance (Claude) after I independently reproduced and fixed the bug; I reviewed the change and take responsibility for it.

### Related Issues

- Fixes #4230

### Additional Context

See the Description above for the repro, the fix and the test evidence.

### DIFF
--- a/dlt/common/libs/sqlglot.py
+++ b/dlt/common/libs/sqlglot.py
@@ -640,6 +640,8 @@ def build_typed_literal(
             lit = sge.Null()
         elif isinstance(v, str):
             lit = sge.Literal.string(v)
+        elif isinstance(v, bool):
+            lit = sge.Boolean(this=v)
         elif isinstance(v, (int, float)):
             lit = sge.Literal.number(v)
         elif isinstance(v, (bytes, bytearray)):
--- a/tests/libs/test_sqlglot.py
+++ b/tests/libs/test_sqlglot.py
@@ -3,10 +3,12 @@ import importlib
 import pytest
 import sqlglot
 import sqlglot.expressions as sge
+from sqlglot.optimizer.annotate_types import annotate_types
 
 from dlt.common.libs.sqlglot import (
     DLT_SUBQUERY_NAME,
     TSqlGlotDialect,
+    build_typed_literal,
     filter_select_column_names,
     from_sqlglot_type,
     get_select_column_names,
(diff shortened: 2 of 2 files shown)


zilto (approved):
LGTM. Thanks for submitting the fix!

PR #4235, opened 2026-07-16
fix(rest-client): allow JWT auth without scopes

### LINKED ISSUE #4234: OAuthJWTAuth crashes when optional scopes are omitted
### dlt version

1.29.0

Also reproduced against the current `devel` branch at `5c393c2ecc602f15bd9f2a0dea79b3c73d594248`.

### Describe the problem

Constructing `OAuthJWTAuth` without explicitly providing `scopes` raises a `TypeError` before any HTTP, cryptographic, or session work occurs.

The public constructor declares the field as optional and defaults it to `None`:

```python
scopes: Optional[Union[str, List[str]]] = None
```

However, `__post_init__` sends every non-string value to `str.join`:

```python
self.scopes = self.scopes if isinstance(self.scopes, str) else " ".join(self.scopes)
```

When the declared default is used, that becomes `" ".join(None)` and fails with:

```text
TypeError: can only join an iterable
```

This makes JWT bearer authentication unusable for authorization servers/configurations where no scope needs to be requested. The failure happens during object construction, so users cannot work around it later in the request lifecycle.

I reproduced the failure twice. As controls, both `scopes=""` and `scopes=["read", "write"]` construct successfully, which isolates the problem to the `None` branch.

### Expected behavior

Omitting the explicitly optional `scopes` argument should construct `OAuthJWTAuth` successfully.

Absent scopes should be represented intentionally in the JWT payload—such as by omitting the scope claim—rather than causing an internal Python `TypeError`. If scopes are intended to be mandatory, the constructor annotation/default should reflect that and validation should raise a clear configuration error.

OAuth's JWT bearer profile permits a scope to be requested but does not require one:

- https://www.rfc-editor.org/info/rfc7523/
- https://www.rfc-editor.org/info/rfc6749/

### Steps to reproduce

Run:

```python
from dlt.sources.helpers.rest_client.auth import OAuthJWTAuth

OAuthJWTAuth(
    client_id="id",
    private_key="unused",
    auth_endpoint="https://example.com/token",
)
```

Actual traceback:

```text
  ...
  File "dlt/sources/helpers/rest_client/auth.py", line 230, in __post_init__
    self.scopes = self.scopes if isinstance(self.scopes, str) else " ".join(self.scopes)
TypeError: can only join an iterable
```

Passing controls:

```python
OAuthJWTAuth(
    client_id="id",
    private_key="unused",
    auth_endpoint="https://example.com/token",
    scopes="",
)

auth = OAuthJWTAuth(
    client_id="id",
    private_key="unused",
    auth_endpoint="https://example.com/token",
    scopes=["read", "write"],
)
assert auth.scopes == "read write"
```

The existing `test_oauth_jwt_auth_success` test always supplies `scopes=["read", "write"]`, so the default-input path is not covered.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

`dlt.sources.helpers.rest_client.auth.OAuthJWTAuth`

### Destination

Not destination-specific; construction fails before a destination or remote API is used.

### Other deployment details

Reproduced with:

- macOS 26.5.2, ARM64
- Python 3.13.5
- dlt 1.29.0
- pytest 7.4.4

The focused REST-client suite still passes (`43 passed`), confirming this is a missing regression case rather than a currently failing test.

### Additional information

A focused regression test could be added alongside `test_oauth_jwt_auth_success` in `tests/sources/helpers/rest_client/test_client.py` to verify that construction without scopes succeeds and that the resulting payload handles absent scope intentionally.

The likely fix boundary is `OAuthJWTAuth.__post_init__` and, depending on the desired absent-scope representation, `create_jwt_payload`.

Repository-wide issue and PR searches across open, closed, merged, and draft states found no existing report or active fix for `OAuthJWTAuth`, the exact exception, or the affected fields.

mattfaltyn:
### Description

- Preserve `None` when normalizing optional JWT scopes instead of passing it to `str.join`.
- Omit the `scope` claim when no scopes were configured while preserving existing string and list behavior.
- Add a regression test for construction and payload generation without scopes.

`OAuthJWTAuth.scopes` is declared optional and defaults to `None`, but `__post_init__` treated every non-string value as an iterable. Using the declared default therefore raised `TypeError: can only join an iterable` before authentication could begin.

This restores the documented type contract for scope-less JWT bearer configurations without changing existing behavior for configured scopes.

### Related Issues

- Fixes #4234

### Additional Context

Validation completed:

- `uv run black dlt/sources/helpers/rest_client/auth.py tests/sources/helpers/rest_client/test_client.py`
- `python -m pytest tests/sources/helpers/rest_client -q` — 277 passed, 1 skipped
- Direct behavior matrix for omitted, empty-string, and list scopes
- `make lint` — mypy, Ruff, Flake8, Bandit, lockfile, dependency, and API-breaking checks passed
- `make test-common-p` — 7,008 passed and 64 skipped; 36 unrelated environment/upstream-facing tests failed because optional `streamlit` was unavailable, worktree/venv subprocess tests assume a different environment, and a live GitHub example hit the unauthenticated API rate limit

No documentation currently references `OAuthJWTAuth`; this change aligns runtime behavior with the existing optional annotation and default.

### DIFF
--- a/dlt/sources/helpers/rest_client/auth.py
+++ b/dlt/sources/helpers/rest_client/auth.py
@@ -227,7 +227,8 @@ class OAuthJWTAuth(BearerTokenAuth):
     session: Annotated[BaseSession, NotResolved()] = None
 
     def __post_init__(self) -> None:
-        self.scopes = self.scopes if isinstance(self.scopes, str) else " ".join(self.scopes)
+        if self.scopes is not None and not isinstance(self.scopes, str):
+            self.scopes = " ".join(self.scopes)
         self.token = None
         self.token_expiry: Optional[pendulum.DateTime] = None
         # use default system session unless specified otherwise
@@ -270,14 +271,16 @@ class OAuthJWTAuth(BearerTokenAuth):
 
     def create_jwt_payload(self) -> Dict[str, Union[str, int]]:
         now = pendulum.now()
-        return {
+        payload: Dict[str, Union[str, int]] = {
             "iss": self.client_id,
             "sub": self.client_id,
             "aud": self.auth_endpoint,
             "exp": math.floor((now.add(hours=1)).timestamp()),
             "iat": math.floor(now.timestamp()),
-            "scope": cast(str, self.scopes),
         }
+        if self.scopes is not None:
+            payload["scope"] = cast(str, self.scopes)
+        return payload
 
     def load_private_key(self) -> "PrivateKeyTypes":
         try:
--- a/tests/sources/helpers/rest_client/test_client.py
+++ b/tests/sources/helpers/rest_client/test_client.py
@@ -322,6 +322,17 @@ class TestRESTClient:
 
         assert_pagination(list(pages_iter))
 
+    def test_oauth_jwt_auth_without_scopes(self):
+        auth = OAuthJWTAuth(
+            client_id="test-client-id",
+            private_key=TEST_PRIVATE_KEY,
+            auth_endpoint="https://api.example.com/oauth/token",
+            session=Client().session,
+        )
+
+        assert auth.scopes is None
+        assert "scope" not in auth.create_jwt_payload()
+
     def test_oauth_jwt_auth_success(self, rest_client: RESTClient):
         auth = OAuthJWTAuth(
             client_id="test-client-id",


burnash (approved):
Great find, thanks so much @mattfaltyn

PR #4236, opened 2026-07-16
fix(test) stop flaky gather_metrics on early sleep return

burnash:
Fix CI failure: https://github.com/dlt-hub/dlt/actions/runs/29436958012/job/87549917933?pr=4227

### DIFF
--- a/tests/extract/data_writers/test_buffered_writer.py
+++ b/tests/extract/data_writers/test_buffered_writer.py
@@ -343,20 +343,22 @@ def test_gather_metrics(disable_compression: bool, writer_type: Type[DataWriter]
     now = time.time()
     c1 = new_column("col1", "bigint")
     t1 = {"col1": c1}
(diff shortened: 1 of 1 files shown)


tests/extract/data_writers/test_buffered_writer.py
tetelio, line 342: @burnash `tests/common/test_time.py` already covers monotonicity, wall-clock tracking, backward jumps, strictly-increasing mode etc

maybe it would fit here to use the same pattern as `test_log_collector_respects_log_period`, where a fake clock with a lambda is used to patch the writer._clock and then the timing assertions wont depend on brittle hopeful thresholds. What do you think?

burnash: Thanks for the idea @tetelio I've ported this fix to use fake clock.

tetelio (approved):
LGTM!

PR #4239, opened 2026-07-16
fix(dataset): migrate stored schemas from destination

bartcode:
Four callers that build a `Schema` from raw destination JSON used `Schema.from_stored_schema`, which skips `migrate_schema` and `apply_defaults`. As a result, `dlt.dataset()` (and the dashboard schema-version loader) could not read schemas written by older dlt versions — in the canonical stored format the column name lives only as the dict key, so any naming normalization on the resulting schema raised `KeyError: 'name'`.

## Fix

The four raw-JSON loaders now use `Schema.from_dict(..., validate_schema=False)`, consistent with every other stored-schema load (`schema_storage.py`, `pipeline.py`, `load_package.py`). `migrate_schema` and `apply_defaults` run; only the strict validator that could reject legacy stored schemas is skipped.

- `dlt/dataset/dataset.py` (3 sites)
- `dlt/_workspace/helpers/dashboard/utils/schema.py`

## Test

`test_dataset_on_schema_stored_by_old_dlt` stores a schema with `dlt[duckdb]==0.3.0` (engine 5), then opens it via `dlt.dataset()` and asserts the schema is migrated (data-type rename, row/parent-key derivation, path-separator pinning) and the data is queryable. Fails without the loader fix.

### DIFF
--- a/dlt/common/schema/utils.py
+++ b/dlt/common/schema/utils.py
@@ -678,10 +678,12 @@ def normalize_table_identifiers(table: TTableSchema, naming: NamingConvention) -
     columns = table.get("columns")
     if columns:
         new_columns: TTableSchemaColumns = {}
-        for c in columns.values():
+        for col_name, c in columns.items():
             c = copy(c)
-            origin_c_name = c["name"]
-            new_col_name = c["name"] = naming.normalize_path(c["name"])
+            # the inner "name" is stripped from stored schemas by remove_defaults;
+            # fall back to the dict key, which is the authoritative column name
+            origin_c_name = c.get("name", col_name)
+            new_col_name = c["name"] = naming.normalize_path(origin_c_name)
             # re-index columns as the name changed, if name space was reduced then
             # some columns now collide with each other. so make sure that we merge columns that are already there
             if new_col_name in new_columns:
--- a/dlt/common/schema/utils.py
+++ b/dlt/common/schema/utils.py
@@ -678,12 +678,10 @@ def normalize_table_identifiers(table: TTableSchema, naming: NamingConvention) -
     columns = table.get("columns")
     if columns:
         new_columns: TTableSchemaColumns = {}
-        for col_name, c in columns.items():
+        for c in columns.values():
             c = copy(c)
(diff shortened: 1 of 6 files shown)


dlt/common/schema/utils.py
```diff
@@ -678,10 +678,12 @@ def normalize_table_identifiers(table: TTableSchema, naming: NamingConvention) -
     columns = table.get("columns")
     if columns:
         new_columns: TTableSchemaColumns = {}
-        for c in columns.values():
+        for col_name, c in columns.items():
```
rudolfix: Please do not change that. I prefer code like that to fail than to cover up bugs in other places.

tests/common/schema/test_normalize_identifiers.py
```diff
@@ -237,6 +237,20 @@ def test_normalize_table_identifiers_merge_columns() -> None:
     }
 
 
+def test_normalize_table_identifiers_column_without_inner_name() -> None:
```
rudolfix: are you able to add a repro of your actual problem instead? it should fail on code before your changes

rudolfix (changes requested):
thanks for finding this! unfortunately mypy didn't catch this - from_stored_schema requires migrated and preprocessed type but json.loads returns Any.

please see my review

dlt/common/schema/utils.py
bartcode: Alright! I changed it back.

tests/common/schema/test_normalize_identifiers.py
bartcode: Do you agree with this new test?

rudolfix (approved):
root cause of the original problem is in dagster-dlt but the bugs that got fixed were real. I changed the repro test into migration test that fails without those fixes. we can merge it like now. 

I'd disconnect it from original issue as the bug fixed is different and change the PR subject. lmk if you are OK with that

bartcode:
@rudolfix , sounds good to me! I've edited the title and PR body.

PR #4245, opened 2026-07-16
fix(destinations): match compound SCD2 merge keys exactly

### LINKED ISSUE #4244: SCD2 compound merge-key collisions retire unrelated active records
### dlt version

`1.29.0`

Also reproduced on current `devel` at `449896006ab07b9042ef84eb1a8b66c864253de0`.

### Describe the problem

When SCD2 uses a compound `merge_key`, dlt decides which absent active records to retire by concatenating all key columns and comparing the resulting strings.

That mapping is not one-to-one. For example:

```text
("a", "bc") -> "abc"
("ab", "c") -> "abc"

(1, 23) -> "123"
(12, 3) -> "123"
```

As a result, an incremental load containing `("ab", "c")` incorrectly retires an existing active record keyed by `("a", "bc")`. The load succeeds without a warning, but the unrelated record receives a non-null `_dlt_valid_to` and disappears from the active SCD2 view.

This contradicts the documented behavior that compound natural keys are allowed and that records whose corresponding keys are absent from an incremental extract are not retired:

https://github.com/dlt-hub/dlt/blob/449896006ab07b9042ef84eb1a8b66c864253de0/docs/website/docs/general-usage/merge-loading.md#L378-L383

The issue was reproduced with:

- native DuckDB
- the SQLAlchemy destination backed by DuckDB
- text compound keys
- integer compound keys
- a separate PyPI installation of dlt 1.29.0

### Expected behavior

Compound merge keys should be compared structurally, column by column. Since `("a", "bc") != ("ab", "c")`, loading the latter key must not retire the former record.

After the second load, both records should remain active:

```text
("a",  "bc", "A", _dlt_valid_to=None)
("ab", "c",  "B", _dlt_valid_to=None)
```

### Steps to reproduce

Install the DuckDB extra and run:

```python
import tempfile
from pathlib import Path

import dlt


with tempfile.TemporaryDirectory() as temp_dir:
    root = Path(temp_dir)
    pipeline = dlt.pipeline(
        pipeline_name="scd2_compound_collision",
        destination=dlt.destinations.duckdb(str(root / "data.duckdb")),
        dataset_name="repro",
    )

    @dlt.resource(
        name="dim",
        merge_key=["key_1", "key_2"],
    )
    def dim(rows):
        yield rows

    pipeline.run(
        dim([{"key_1": "a", "key_2": "bc", "value": "A"}])
    )
    pipeline.run(
        dim([{"key_1": "ab", "key_2": "c", "value": "B"}])
    )

    rows = pipeline.dataset()(
        """
        SELECT key_1, key_2, value, _dlt_valid_to
        FROM dim
        ORDER BY value
        """
    ).fetchall()
    print(rows)

    assert rows[0] == ("a", "bc", "A", None)
```

Actual output:

```text
[
  ('a', 'bc', 'A', <non-null timestamp>),
  ('ab', 'c', 'B', None)
]
AssertionError
```

Control: change the existing key from `("a", "bc")` to `("a", "bd")`. Both records then remain active as expected.

The failure was repeated. A focused matrix produced four deterministic failures and two passing controls:

```text
text collision / native DuckDB       FAILED
text collision / SQLAlchemy DuckDB   FAILED
integer collision / native DuckDB    FAILED
integer collision / SQLAlchemy       FAILED
text non-collision / native DuckDB   PASSED
text non-collision / SQLAlchemy      PASSED
```

The repository's four existing non-colliding text/integer compound-key tests for native and SQLAlchemy DuckDB also pass.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

Python resource yielding dictionaries

### dlt destination

DuckDB; also reproduced through the SQLAlchemy destination using DuckDB

### Other deployment details

No external services or credentials are required.

### Additional information

The generic SQL implementation creates a collision-prone expression here:

https://github.com/dlt-hub/dlt/blob/449896006ab07b9042ef84eb1a8b66c864253de0/dlt/destinations/sql_jobs.py#L377-L378

and uses it in the SCD2 retirement predicate here:

https://github.com/dlt-hub/dlt/blob/449896006ab07b9042ef84eb1a8b66c864253de0/dlt/destinations/sql_jobs.py#L978-L989

Generated SQL includes:

```sql
AND CONCAT("key_1", "key_2") IN (
    SELECT CONCAT("key_1", "key_2")
    FROM "repro_staging"."dim"
)
```

The SQLAlchemy implementation independently repeats the same serialization:

https://github.com/dlt-hub/dlt/blob/449896006ab07b9042ef84eb1a8b66c864253de0/dlt/destinations/impl/sqlalchemy/merge_job.py#L338-L349

https://github.com/dlt-hub/dlt/blob/449896006ab07b9042ef84eb1a8b66c864253de0/dlt/destinations/impl/sqlalchemy/merge_job.py#L401-L410

A robust fix would compare each root/staging key column directly, for example with a correlated `EXISTS` containing:

```sql
root.key_1 = staging.key_1
AND root.key_2 = staging.key_2
```

This avoids delimiter and casting collisions and preserves exact compound-key semantics. A focused regression can extend the existing compound-key SCD2 test:

https://github.com/dlt-hub/dlt/blob/449896006ab07b9042ef84eb1a8b66c864253de0/tests/load/pipeline/test_scd2.py#L902-L956

mattfaltyn:
### Description

- Compare compound SCD2 merge keys by exact per-column equality instead of concatenating their values.
- Use correlated `EXISTS` predicates for standard SQL and SQLAlchemy, with dialect-safe tuple membership for Athena and ClickHouse.
- Avoid staging-alias shadowing when the destination table is named `s`; the regression test now covers that case explicitly.
- Extend the existing destination matrix with ambiguous text and numeric keys that previously retired an unrelated active record.

This fixes collisions such as `("a", "bc")` and `("ab", "c")`, which both became `"abc"` under the previous concatenation logic.

### Related Issues

- Closes #4244

### Additional Context

Validation:

- `make lint`
- Common parallel selection: 7,054 passed / 57 skipped; its only two failures were checkout-path URL assertions caused by the space in `/Volumes/Mac SSD`, and both passed from the identical no-space checkout
- Common serial/forked selection: 213 passed / 1 skipped; the three signal-launcher cases passed in fresh processes to avoid a macOS `pytest-forked` deadlock
- Local DuckDB/filesystem selection from the no-space checkout: 1,432 passed / 205 skipped
- 8 focused compound-key SCD2 cases passed across native DuckDB, DuckDB with Parquet staging, SQLAlchemy SQLite, and SQLAlchemy DuckDB under SQLAlchemy 2 (text and bigint keys)
- `.venv/bin/pytest tests/load/pipeline/test_scd2.py -k duckdb -q`: 45 passed / 6 skipped

The local-load serial selector has no active DuckDB/filesystem cases. On macOS, two inactive Databricks modules are not collectable because the Zerobus SDK is published only for Linux and Windows; those modules and dependency declarations are unchanged by this PR.

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -132,10 +132,14 @@ class AthenaMergeJob(SqlMergeFollowupJob):
         return sql, temp_table_name
 
     @classmethod
-    def gen_concat_sql(cls, columns: Sequence[str]) -> str:
-        # Athena requires explicit casting
-        columns = [f"CAST({c} AS VARCHAR)" for c in columns]
-        return f"CONCAT({', '.join(columns)})"
+    def gen_scd2_key_present_sql(
+        cls,
+        root_table_name: str,
+        staging_root_table_name: str,
+        merge_keys: Sequence[str],
+    ) -> str:
+        key_tuple = ", ".join(merge_keys)
+        return f"({key_tuple}) IN (SELECT {key_tuple} FROM {staging_root_table_name})"
 
     @classmethod
     def requires_temp_table_for_delete(cls) -> bool:
--- a/dlt/destinations/impl/clickhouse/clickhouse.py
+++ b/dlt/destinations/impl/clickhouse/clickhouse.py
@@ -294,6 +294,16 @@ class ClickHouseMergeJob(SqlMergeFollowupJob):
         return f"ALTER TABLE {table_name} UPDATE"
 
     @classmethod
+    def gen_scd2_key_present_sql(
+        cls,
+        root_table_name: str,
+        staging_root_table_name: str,
+        merge_keys: Sequence[str],
+    ) -> str:
+        key_tuple = ", ".join(merge_keys)
+        return f"({key_tuple}) IN (SELECT {key_tuple} FROM {staging_root_table_name})"
+
+    @classmethod
     def requires_temp_table_for_delete(cls) -> bool:
         return False
 
--- a/dlt/destinations/impl/sqlalchemy/merge_job.py
+++ b/dlt/destinations/impl/sqlalchemy/merge_job.py
@@ -1,5 +1,4 @@
-from typing import Sequence, Tuple, Optional, List, Union, cast
-import operator
+from typing import Sequence, Tuple, Optional, List, cast
 from dlt.common.libs.sql_alchemy import sa
 
 from dlt.common.typing import TAnyDateTime
@@ -335,20 +334,6 @@ class SqlalchemyMergeFollowupJob(SqlMergeFollowupJob):
             return sa.true()  # type: ignore[no-any-return]
 
     @classmethod
-    def _gen_concat_sqla(
-        cls, columns: Sequence[sa.Column]
-    ) -> Union[sa.sql.elements.BinaryExpression, sa.Column]:
-        # Use col1 + col2 + col3 ... to generate a dialect specific concat expression
-        result = columns[0]
-        if len(columns) == 1:
-            return result
-        # Cast because CONCAT is only generated for string columns
-        result = sa.cast(result, sa.String)
-        for col in columns[1:]:
-            result = operator.add(result, sa.cast(col, sa.String))
-        return result  # type: ignore[no-any-return]
-
-    @classmethod
     def gen_scd2_sql(
         cls,
         table_chain: Sequence[PreparedTableSchema],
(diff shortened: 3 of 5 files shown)


themattmorris:
This bug caused records to get closed out for us today as well. While we can monkeypatch to work around it for now, would love if this PR could get attention so it could be incorporated, especially considering others may be affected unknowingly or taking some time to get to the bottom of why their active records are lower than expected!

PR #4246, opened 2026-07-16
fix(cli): remove click dependency

### LINKED ISSUE #1658: remove click
We have click dependency but we do not use it. We use a few functions to display color messages and process yes/no choices. investigate if that can be easily removed

infisble:
### Description

Removes the direct `click` dependency from the CLI helpers while keeping the existing output and prompting behavior.

The output helpers now use Rich for terminal-safe rendering, and the small interactive helpers use standard input for confirmations, choices, and text prompts. The related type dependency and lockfile entries are removed as well.

Closes #1658

### Tests

- Added coverage for styled output, stderr, confirmations, choices, retries, and defaults
- Ran the CLI echo and invocation tests (`74 passed`)
- Ran Black, isort, mypy, Flake8, and Ruff on the changed files
- Built the wheel and source distribution

### DIFF
--- a/dlt/_workspace/cli/echo.py
+++ b/dlt/_workspace/cli/echo.py
@@ -1,9 +1,30 @@
 """CLI prompting and output helpers."""
 
-import sys
 import contextlib
-from typing import Any, Iterable, Iterator, Optional, ContextManager
-import click
+import sys
+from typing import IO, Any, ContextManager, Iterable, Iterator, Optional
+
+from rich.console import Console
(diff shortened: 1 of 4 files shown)


infisble:
Hi! I opened this before getting explicit confirmation on #1658, and the branch has since fallen behind `devel`. Before I rebase and resolve the conflict, could you confirm that replacing `click` with Rich plus small prompt helpers is still the direction you want? If so, I’m happy to update it and address any feedback.

PR #4251, opened 2026-07-17
fix(pipeline): preserve import schema after restore

### LINKED ISSUE #4242: save_schema poisons _imported_version_hash when sync_destination restores stale schema
## Bug

When `sync_destination()` restores a stale schema from the destination and `import_schema_path` is configured, `save_schema()` unconditionally stamps the restored schema with the current import file hash (`_imported_version_hash`). This makes `_maybe_import_schema()` think the import YAML is already applied and skip re-import — even though the restored schema predates the latest YAML change.

With `columns: freeze` this causes a **permanent deadlock** on ephemeral storage (e.g. Databricks `/local_disk0`): every run restores the stale schema, skip re-import, fails on the new column, and never updates the destination — so the next run hits the exact same failure.

**Affected versions:** 1.26.0 – 1.29.0 (`schema_storage.py` is identical across all).

## Root cause

`schema_storage.py:73` — `save_schema()` stamps any schema with the current import hash, regardless of whether it came from the import file or was just restored from the destination:

```python
schema._imported_version_hash = imported_schema.stored_version_hash  # fires unconditionally
```

`pipeline.py:868` calls `save_schema` on every schema restored from the destination:

```python
for schema in restored_schemas:
    self._schema_storage.save_schema(schema)  # poisons _imported_version_hash
```

`schema_storage.py:145` then short-circuits:

```python
if i_s.version_hash != rv_schema._imported_version_hash:  # always False → import skipped
    rv_schema.replace_schema_content(i_s, ...)
```

## Reproduction

```bash
# requires uv (https://docs.astral.sh/uv/)
uv run python repro.py   # Run 1: establish destination schema
uv run python repro.py   # Run 2: PipelineStepFailed (columns frozen)
uv run python repro.py --clean  # reset
```

<details>
<summary>pyproject.toml</summary>

```toml
[project]
name = "dlt-repro"
version = "0.1.0"
requires-python = ">=3.11"
dependencies = ["dlt[filesystem]==1.29.0", "PyYAML"]
```
</details>

<details>
<summary>repro.py</summary>

```python
import os, sys, shutil, yaml, dlt
from dlt.common.schema import Schema
from dlt.common.storages.schema_storage import SchemaStorage

BASE_DIR      = os.path.dirname(os.path.abspath(__file__))
SCHEMAS_DIR   = os.path.join(BASE_DIR, "schemas")
PIPELINES_DIR = os.path.join(BASE_DIR, "pipelines")
DEST_DIR      = os.path.join(BASE_DIR, "dest")
RUN_FLAG      = os.path.join(BASE_DIR, "_run2")

if "--clean" in sys.argv:
    for p in [SCHEMAS_DIR, PIPELINES_DIR, DEST_DIR, RUN_FLAG]:
        shutil.rmtree(p, ignore_errors=True) if os.path.isdir(p) else (os.remove(p) if os.path.isfile(p) else None)
    os.makedirs(DEST_DIR, exist_ok=True); print("Cleaned."); sys.exit(0)

def make_pipeline(with_import=False):
    return dlt.pipeline(
        pipeline_name="my_pipeline",
        destination=dlt.destinations.filesystem(DEST_DIR),
        dataset_name="my_dataset",
        import_schema_path=SCHEMAS_DIR if with_import else None,
        pipelines_dir=PIPELINES_DIR,
    )

if not os.path.exists(RUN_FLAG):
    print("=== RUN 1: establish destination schema (id only, columns:freeze) ===")
    @dlt.resource(name="items", write_disposition="append")
    def source_v1():
        yield {"id": 1}
    shutil.rmtree(SCHEMAS_DIR, ignore_errors=True)
    p = make_pipeline()
    p.run(source_v1(), schema_contract={"columns": "freeze"})
    os.makedirs(SCHEMAS_DIR)
    with open(os.path.join(SCHEMAS_DIR, "my.schema.yaml"), "w") as f:
        f.write(p.default_schema.to_pretty_yaml())
    open(RUN_FLAG, "w").close()
    shutil.rmtree(PIPELINES_DIR)
    print("Done. Run again to trigger the bug.")

else:
    print("=== RUN 2: add 'name' column, wipe local state, run ===")
    schema_path = os.path.join(SCHEMAS_DIR, "my.schema.yaml")
    with open(schema_path) as f:
        raw = yaml.safe_load(f.read())
    raw["tables"]["items"]["columns"]["name"] = {"data_type": "text", "nullable": True}
    raw["previous_hashes"] = raw.get("previous_hashes", []) + [raw["version_hash"]]
    raw["version"] += 1; raw["version_hash"] = "placeholder"
    updated = Schema.from_dict(raw)
    with open(schema_path, "w") as f:
        f.write(updated.to_pretty_yaml())
    shutil.rmtree(PIPELINES_DIR, ignore_errors=True)

    # instrument to show the bug
    _orig_save = SchemaStorage.save_schema
    def _save(self, schema):
        before = schema._imported_version_hash
        result = _orig_save(self, schema)
        if schema._imported_version_hash and before != schema._imported_version_hash:
            print(f"  [save_schema] stamped stale destination schema v{schema.stored_version} with new import hash")
            print(f"    columns: {list(schema.tables.get('items', {}).get('columns', {}).keys())}")
        return result
    SchemaStorage.save_schema = _save

    _orig_maybe = SchemaStorage._maybe_import_schema
    def _maybe(self, name, storage_schema=None):
        if storage_schema and name == "my":
            sv = Schema.from_dict(storage_schema, validate_schema=False)
            try:
                iv = Schema.from_dict(self._load_import_schema(name))
                if iv.version_hash == sv._imported_version_hash:
                    print(f"  [_maybe_import_schema] SKIP re-import (hashes match — BUG)")
                    print(f"    destination: {list(sv.tables.get('items', {}).get('columns', {}).keys())}")
                    print(f"    import:      {list(iv.tables.get('items', {}).get('columns', {}).keys())}")
            except FileNotFoundError:
                pass
        return _orig_maybe(self, name, storage_schema)
    SchemaStorage._maybe_import_schema = _maybe

    @dlt.resource(name="items", write_disposition="append")
    def source_v2():
        yield {"id": 2, "name": "Alice"}

    p = make_pipeline(with_import=True)
    try:
        p.run(source_v2(), schema_contract={"columns": "freeze"})
        print("Run SUCCEEDED — bug did not reproduce")
    except Exception as e:
        print(f"BUG REPRODUCED: {type(e).__name__}")
    os.remove(RUN_FLAG)
```
</details>

## Suggested fix

Only stamp `_imported_version_hash` when the schema being saved is not older than the import file:

```python
# schema_storage.py save_schema()
if self.config.import_schema_path:
    try:
        imported_schema = Schema.from_dict(self._load_import_schema(schema.name))
        if imported_schema.stored_version <= schema.stored_version:  # <-- guard added
            schema._imported_version_hash = imported_schema.stored_version_hash
    except FileNotFoundError:
        pass
```

AkhilTrivediX:
## Summary

Prevent schemas restored from a destination from being marked as if the current import schema had already been applied.

## Root cause

During destination synchronization, restored schemas were saved through the normal persistence path. That path links every saved schema to the current import schema. A stale restored schema could therefore receive the latest import hash without containing the imported changes; the next load then skipped the import.

## Changes

- Add an opt-out for import linking when a schema is persisted.
- Use it only for schemas restored from a destination.
- Keep the existing import-link behavior for ordinary local saves.
- Cover the stale restore case in storage and pipeline restore tests.

## Validation

- `uv run pytest tests/common/storages/test_schema_storage.py -q`
- `uv run pytest tests/pipeline/test_import_export_schema.py -q`
- `uv run pytest tests/load/pipeline/test_restore_state.py -k 'restore_schemas_while_import_schemas_exist and (duckdb-no-staging or duckdb-parquet-no-staging) and not sqlalchemy' -q`
- `uv run ruff check dlt/common/storages/schema_storage.py dlt/common/storages/live_schema_storage.py dlt/pipeline/pipeline.py tests/common/storages/test_schema_storage.py tests/load/pipeline/test_restore_state.py`
- `uv run black --check dlt/common/storages/schema_storage.py dlt/common/storages/live_schema_storage.py dlt/pipeline/pipeline.py tests/common/storages/test_schema_storage.py tests/load/pipeline/test_restore_state.py`
- `uv run mypy dlt/common/storages/schema_storage.py dlt/common/storages/live_schema_storage.py dlt/pipeline/pipeline.py`

Fixes #4242


tests/load/pipeline/test_restore_state.py
```diff
@@ -602,9 +607,10 @@ def test_restore_schemas_while_import_schemas_exist(
     schema = p.schemas["ethereum"]
     assert normalized_labels in schema.tables
     assert normalized_annotations in schema.tables
+    assert imported_after_restore in schema.tables
 
     # check if attached to import schema
-    assert schema._imported_version_hash == IMPORTED_VERSION_HASH_ETH_V10()
+    assert schema._imported_version_hash == imported_version_hash
```
burnash, line 668: the rewritten test now only covers the changed import case for the bug and replaces the old scenario instead of adding to it. the original version also pinned the unchanged import case: a restored schema whose import link matches the current import file is not overwritten. Consider asserting that after a restore before the new import is published, so both directions stay covered

tests/common/storages/test_schema_storage.py
```diff
@@ -105,6 +105,20 @@ def test_import_overwrites_existing_if_modified(
     assert_schema_imported(synced_storage, storage)
 
 
+def test_save_does_not_link_stale_schema_to_current_import(
+    ie_storage: SchemaStorage, storage: SchemaStorage
+) -> None:
+    stale_schema = Schema("ethereum")
+    storage.save_schema(stale_schema)
```
burnash: imo this initial save is redundant

burnash (changes requested):
Thanks you for the fix @AkhilTrivediX, please see my comments related to the tests.

burnash:
Thanks for the update @AkhilTrivediX, could you please merge devel branch to resolve the merge conflicts?

AkhilTrivediX:
Merged the `devel` branch and resolved the merge conflicts in `dlt/pipeline/pipeline.py`.

The `link_import_schema=False` behavior has been preserved and applied within the upstream `_install_state_and_schemas` helper method to ensure import schemas remain intact during restore.

The update is in `785a8ea`. The PR checks triggered by the merge commit should be running now.

tests/common/storages/test_schema_storage.py
burnash: @AkhilTrivediX please drop the redundant line

burnash (changes requested):

burnash:
@AkhilTrivediX please let me know if you're interested in addressing the requested changes and finishing this PR

burnash (dismissed):
looks good, thank you @AkhilTrivediX

rudolfix (approved):
LGTM!

PR #4259, opened 2026-07-19
fix(fabric): scale nvarchar precision for utf-8 varchar byte semantics

### LINKED ISSUE #4255: Fabric: scale NVARCHAR character precision for UTF-8 VARCHAR byte semantics
### Problem

When a SQL Server source column is `nvarchar(N)`, the dlt schema stores `precision=N` in character units. The Fabric destination converts `nvarchar` to `varchar` via string replacement, yielding `varchar(N)` -- but Fabric's UTF-8 `varchar(N)` uses byte semantics.

A value that fits in `nvarchar(100)` (100 Unicode characters) can require up to 400 UTF-8 bytes. The direct `varchar(100)` mapping silently truncates multi-byte data.

### Current workaround

Use `reflection_level="full"` instead of `full_with_precision` so Fabric creates `varchar(max)`, losing bounded-length information.

### Proposed fix

In `FabricTypeMapper.to_destination_type`, handle `text` explicitly: multiply the character precision by 4 (worst-case UTF-8 bytes per code point). Fall back to `varchar(max)` when the result exceeds Fabric's 8000-byte `varchar` limit or when no precision is set.

No schema-level changes needed -- the fix is self-contained in the Fabric type mapper.

sdebruyn:
## Summary

- Handle `text` type explicitly in `FabricTypeMapper.to_destination_type` before the parent `nvarchar` -> `varchar` string replacement
- Multiply character precision by 4 to safely cover UTF-8 encoding (worst-case 4 bytes per code point)
- Fall back to `varchar(max)` when the scaled length exceeds Fabric's 8000-byte limit or when no precision is set

## Test plan

- [x] `varchar(400)` for `nvarchar(100)` (4x scaling)
- [x] `varchar(max)` when scaled length > 8000
- [x] `varchar(max)` when no precision set
- [x] Small precision stays bounded (`varchar(40)` for precision 10)

Closes dlt-hub/dlt#4255

### DIFF
--- a/dlt/destinations/impl/fabric/factory.py
+++ b/dlt/destinations/impl/fabric/factory.py
@@ -53,6 +53,16 @@ class FabricTypeMapper(SynapseTypeMapper):
             # Fabric doesn't have native JSON type, use varchar instead of nvarchar
             return "varchar(%s)" % column.get("precision", "max")
 
+        if sc_t == "text":
+            # varchar uses byte semantics; scale nvarchar char precision by 4 (worst-case utf-8)
+            precision = column.get("precision")
+            if precision is not None:
+                safe_length = precision * 4
+                if safe_length > 8000:
+                    return "varchar(max)"
+                return "varchar(%i)" % safe_length
+            return "varchar(max)"
+
         if sc_t == "time":
             # Fabric requires explicit precision for TIME (0-6)
             precision = min(column.get("precision", 6), 6)
--- a/dlt/destinations/impl/fabric/factory.py
+++ b/dlt/destinations/impl/fabric/factory.py
@@ -54,11 +54,7 @@ class FabricTypeMapper(SynapseTypeMapper):
     def to_destination_type(self, column: TColumnSchema, table: PreparedTableSchema) -> str:
         """Override to use varchar instead of nvarchar and datetime2 instead of datetimeoffset"""
         sc_t = column["data_type"]
-        if sc_t == "json":
-            # Fabric doesn't have native JSON type, use varchar instead of nvarchar
-            return "varchar(%s)" % column.get("precision", "max")
-
-        if sc_t == "text":
+        if sc_t in ("json", "text"):
             precision = column.get("precision")
             if precision is None:
                 return "varchar(max)"
--- a/dlt/destinations/impl/fabric/factory.py
+++ b/dlt/destinations/impl/fabric/factory.py
@@ -22,6 +22,11 @@ if TYPE_CHECKING:
     from .fabric import FabricClient
 
 
+UTF8_MAX_BYTES_PER_CODE_POINT = 4
+FABRIC_VARCHAR_MAX_N = 8000
+"""Largest explicit length for `varchar(n)`, anything longer must be `varchar(max)`"""
+
+
 class FabricTypeMapper(SynapseTypeMapper):
     """Custom type mapper for Fabric Warehouse - replaces nvarchar with varchar and datetimeoffset with datetime2"""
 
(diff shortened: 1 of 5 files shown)


dlt/destinations/impl/fabric/factory.py
Travior: As far as I see it this should be dead code now and can be removed

tests/load/fabric/test_fabric_configuration.py
```diff
@@ -221,3 +221,33 @@ def test_fabric_credentials_authentication_method() -> None:
...
     assert dsn_dict["AUTHENTICATION"] == "ActiveDirectoryServicePrincipal"
+
+
+def test_fabric_type_mapper_nvarchar_to_varchar_length() -> None:
+    """Fabric varchar uses UTF-8 byte semantics; nvarchar precision (characters) must be scaled"""
+    from dlt.destinations.impl.fabric.factory import FabricTypeMapper
+    from dlt.common.destination import DestinationCapabilitiesContext
+    from dlt.common.schema.typing import TColumnSchema
+    from dlt.common.destination.typing import PreparedTableSchema
+    from typing import cast
```
Travior: I believe we can move these imports top level (and while we're at it do the same for `test_fabric_type_mapper`).

dlt/destinations/impl/fabric/factory.py
```diff
@@ -53,6 +53,16 @@ def to_destination_type(self, column: TColumnSchema, table: PreparedTableSchema)
             # Fabric doesn't have native JSON type, use varchar instead of nvarchar
             return "varchar(%s)" % column.get("precision", "max")
 
+        if sc_t == "text":
+            # varchar uses byte semantics; scale nvarchar char precision by 4 (worst-case utf-8)
+            precision = column.get("precision")
+            if precision is not None:
+                safe_length = precision * 4
```
Travior: This should probably be a named constant like: `UTF8_MAX_BYTES_PER_CODE_POINT`

dlt/destinations/impl/fabric/factory.py
```diff
@@ -53,6 +53,16 @@ def to_destination_type(self, column: TColumnSchema, table: PreparedTableSchema)
...
+                if safe_length > 8000:
```
Travior: and let's call this `FABRIC_VARCHAR_MAX_N`

tests/load/fabric/test_fabric_configuration.py
```diff
@@ -221,3 +221,33 @@ def test_fabric_credentials_authentication_method() -> None:
     # Verify ActiveDirectoryServicePrincipal is set
     dsn_dict = creds.get_odbc_dsn_dict()
     assert dsn_dict["AUTHENTICATION"] == "ActiveDirectoryServicePrincipal"
+
+
+def test_fabric_type_mapper_nvarchar_to_varchar_length() -> None:
```
Travior:
```suggestion
def test_fabric_type_mapper_scales_text_precision() -> None:
```
nit :)

Travior (changes requested):
Thanks for the contribution!

I added some small notes to the PR.
One thing I found is that there is a special branch when creating unique columns in `dlt/destinations/impl/fabric/fabric.py` (`FabricClient._get_column_def_sql`). I couldn't find any technical reasons for fabric on why we would need the 900 byte limit here, seems to be copied from MSSQL. In my mind we can probably remove that branch and just delegate to `self.type_mapper.to_destination_type(c, table)`, but this would need some testing.

sdebruyn:
> Thanks for the contribution!
> 
> I added some small notes to the PR. One thing I found is that there is a special branch when creating unique columns in `dlt/destinations/impl/fabric/fabric.py` (`FabricClient._get_column_def_sql`). I couldn't find any technical reasons for fabric on why we would need the 900 byte limit here, seems to be copied from MSSQL. In my mind we can probably remove that branch and just delegate to `self.type_mapper.to_destination_type(c, table)`, but this would need some testing.

This indeed seems to come from SQL Server, no need for that limit in Fabric.

Travior (approved):

Travior:
This looks good to me now. @burnash in case you want to sanity check

burnash (approved):
looks good, thanks @sdebruyn and @Travior

PR #4260, opened 2026-07-19
fix(fabric): allow time columns through the parquet load path

### LINKED ISSUE #4256: Fabric: allow TIME columns through the Parquet load path
### Problem

`SynapseTypeMapper.ensure_supported_type` raises `TerminalValueError` when the loader format is Parquet and the column type is `time`. `FabricTypeMapper` inherits this rejection unchanged, even though:

1. `FabricTypeMapper` already maps `time` to `time(p)` DDL correctly
2. Fabric Warehouse uses a different engine from Synapse Dedicated SQL Pools

This forces every source using `sql_database` against SQL Server to add a `type_adapter_callback` converting `Time` to `String`, discarding temporal type information.

### Proposed fix

Override `ensure_supported_type` in `FabricTypeMapper` to skip the time rejection for Parquet, then delegate to `super()` for everything else. The Synapse restriction remains in place for the Synapse destination.

### Open question

Whether Fabric Warehouse COPY INTO actually accepts the Parquet logical TIME type needs validation against a live Fabric endpoint. If it does not, the override should include a destination-owned conversion rather than pushing the workaround to every source.

sdebruyn:
## Summary

- Override `ensure_supported_type` in `FabricTypeMapper` to skip the inherited Synapse rejection for `time` columns when the loader format is Parquet
- The Synapse rejection remains in place for the Synapse destination
- `FabricTypeMapper` already maps `time` to `time(p)` DDL correctly; this change unblocks the load path

## Test plan

- [x] Synapse mapper still raises for `parquet` + `time`
- [x] Fabric mapper allows `parquet` + `time` without raising
- [x] Fabric mapper produces correct `time(6)` DDL
- [ ] Integration test against live Fabric Warehouse with Parquet staging

Closes dlt-hub/dlt#4256

### DIFF
--- a/dlt/destinations/impl/fabric/factory.py
+++ b/dlt/destinations/impl/fabric/factory.py
@@ -15,6 +15,7 @@ from dlt.common.normalizers.naming import NamingConvention
 from dlt.destinations.impl.synapse.factory import synapse, SynapseTypeMapper
 from dlt.common.destination.typing import PreparedTableSchema
 from dlt.common.schema.typing import TColumnSchema
+from dlt.common.typing import TLoaderFileFormat
 
 from .configuration import FabricClientConfiguration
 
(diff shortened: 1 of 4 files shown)


Travior (approved):
This also seems correct to me now, the test failures are unrelated. Fabric loads pass

Travior:
@burnash also looping you in here in case you want to double check

burnash (approved):
@Travior agree, I'll rerun the ci, once green it's good to go.

burnash:
@Travior lance failing is unrelated, we'll address that in a follow up pr.

PR #4261, opened 2026-07-19
fix(sql_database): map mssql money and smallmoney to decimal

### LINKED ISSUE #4257: sql_database: map MSSQL MONEY and SMALLMONEY to decimal with correct precision
### Problem

SQLAlchemy's MSSQL `MONEY` and `SMALLMONEY` types inherit from `TypeEngine`, not from `Numeric`. They expose no `precision`, `scale`, or `asdecimal` attributes. The `sqla_col_to_column_schema` function does not handle them, so they fall through with no `data_type` set.

This causes downstream issues: Arrow batches may infer inconsistent decimal precision across batches, and the known fixed semantics of these types are lost.

SQL Server defines:
- `money`: `decimal(19,4)` (range +/-922,337,203,685,477.5807)
- `smallmoney`: `decimal(10,4)` (range +/-214,748.3647)

### Current workaround

Read `sys.columns`/`sys.types` separately and apply static resource hints with `apply_hints()`. Source applications should not need their own SQL Server catalog implementation for standard types.

### Proposed fix

Add a `__visit_name__` check for `MONEY` and `SMALLMONEY` in `sqla_col_to_column_schema` (before the `isinstance(sql_t, sqltypes.Numeric)` branch), mapping them to `decimal` with their known fixed precision and scale. This follows the existing pattern used for `DATETIMEOFFSET`.

sdebruyn:
## Summary

- Add `__visit_name__` check for `MONEY` and `SMALLMONEY` in `sqla_col_to_column_schema` before the `isinstance(sql_t, sqltypes.Numeric)` branch
- Maps `money` to `decimal(19,4)` and `smallmoney` to `decimal(10,4)`, matching SQL Server's fixed semantics
- Follows the existing pattern used for `DATETIMEOFFSET`

## Test plan

- [x] `MONEY()` produces `decimal` with precision=19, scale=4
- [x] `SMALLMONEY()` produces `decimal` with precision=10, scale=4
- [x] All existing `test_schema_types.py` tests pass (16 passed, 3 skipped)

Closes dlt-hub/dlt#4257

### DIFF
--- a/dlt/sources/sql_database/schema_types.py
+++ b/dlt/sources/sql_database/schema_types.py
@@ -131,6 +131,10 @@ def sqla_col_to_column_schema(
     if _is_uuid_type(sql_t):
         # we represent UUID as text by default, see default_table_adapter
         col["data_type"] = "text"
+    elif sql_t.__visit_name__ in ("MONEY", "SMALLMONEY"):
+        col["data_type"] = "decimal"
+        col["precision"] = 19 if sql_t.__visit_name__ == "MONEY" else 10
+        col["scale"] = 4
     elif isinstance(sql_t, sqltypes.Numeric):
         # check for Numeric type first and integer later, some numeric types (ie. Oracle)
         # derive from both
--- a/dlt/sources/sql_database/schema_types.py
+++ b/dlt/sources/sql_database/schema_types.py
@@ -11,6 +11,7 @@ from typing import (
 )
 from typing_extensions import TypeAlias
 
+from sqlalchemy.dialects import mssql
 from sqlalchemy.exc import NoReferencedTableError
 
 from dlt.common.typing import TypedDict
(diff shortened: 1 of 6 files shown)


Travior:
Thanks for the contribution, looks appropriate, no notes from my end. I'll approve CI

Travior:
@burnash can you have a final look at this? looks good to me

burnash (changes requested):
Thanks for the fix @sdebruyn, the mapping is right for mssql, but `__visit_name__` check also matches postgresql.MONEY, and now Postgres money columns reflect as decimal. psycopg2 returns those as strings, and loads either fail or the values end up in a `__v_text` variant column.
I pushed a test to your branch that adds a money column to the Postgres so now CI catches that. 

I think the direction should be dialect specific. e.g. follow the schema_types.py::_is_uuid_type pattern and check isinstance(sql_t, (mssql.MONEY, mssql.SMALLMONEY)) rather than __visit_name__.
It would also be good to cover money in the load tests (tests/load/sources/sql_database/mssql_source.py, tests/load/sources/sql_database/test_mssql_database_source.py)

PR #4262, opened 2026-07-20
feat/add-require-instance-job-definition

tetelio:
## What

Adds an `instance` key to the job `require` spec so a job can declare its runner
instance requirements. `instance` is an arbitrary `Dict[str, Any]` — today the
runtime reads `instance.size` (e.g. `{"size": "medium"}`), and the dict is left
open for further configuration keys documented online.

The legacy `machine` key is kept and now marked **deprecated**: setting it emits a
`DltDeprecationWarning` pointing users at `instance`, but it is still accepted and
serialized. This is intentionally **additive** — removing/renaming `machine` would
break already-deployed manifests and needs the manifest schema-version + migration
mechanism we don't have yet.

## Why

The runtime reads instance sizing from `require.instance.size`; dlt OSS needs to
emit that key. Keeping `instance` an opaque dict avoids coupling dlt to a specific
set of instance options — the runtime reads `size` and ignores the rest, so new
options can ship without a dlt release.

## Changes

- `typing.py`: add `instance: Dict[str, Any]` to `TRequireSpec`; mark `machine`
  deprecated in its docstring.
- `decorators.py`: add `warn_deprecated_require()`, call it from `_job()`; update
  the `require` docstrings on `@job`, `@interactive`, and `@pipeline_run` to
  document `instance` and drop `machine`.
- `detectors.py`: emit the same deprecation warning when `machine` is set via a
  module-level `__require__`.
- No serialization change needed — `require` is passed through verbatim into the
  `TJobDefinition`, so `instance` lands in the serialized manifest the backend reads.

## Backward compatibility

Fully additive. `machine` keeps working (with a warning); existing manifests and
deployments are unaffected.

## Tests

- `instance` is stored on the job definition.
- `machine` still stored, and emits a deprecation warning.
- `instance` alone emits no warning.
- Manifest round-trip (`save_manifest` → `load_manifest`) preserves
  `require.instance.size` — the check Marcin can run against the runtime tests
  before/after any later `machine` cleanup.

## Out of scope (dlt-owned, later)

Removing/renaming `machine` once a manifest schema-version + migration mechanism
exists.


rudolfix (approved):
LGTM! I hope we'll be able to type "instance" properly, right now it is a bare dict

PR #4263, opened 2026-07-21
support IF NOT EXISTS for schema creation

Travior:
Adds a new capability for `CREATE SCHEMA IF NOT EXISTS`


rudolfix:
@Travior after discussing various options how to prevent races on creating schema we decided to do that differently: we fail as loud and as early as possible then we add a mechanism and helpers to retry just the schema migration. schema migration is idempotent so retrying it (with backoff and large jitter) looks like the only way to go.
this is all described here:
https://dlthub.com/docs/devel/running-in-production/running#retrying-only-the-schema-update-evolution

so it is better if we fail right away on a race to create a schema, than later at some random moment when tables are not consistent.

so let's close this PR and all other similar PRs (snowflake). thanks for spending your time on this (but I think we've found a better way out - this happens sometimes...)

Travior:
Makes sense, thanks for the heads up

PR #4264, opened 2026-07-21
fix(ci): preserve dependency resolution across test jobs

Travior:
Declares resolution and no-sync policies at the CI job level:
- Propagates matrix resolution settings through `UV_SYNC_ARGS` (for jobs that use `lowest-direct` or `upgrade`, this gets expanded via `make`)
- Prevents `uv run` from resyncing explicitly prepared CI environments via `UV_NO_SYNC`

Additionally:
- Updates the GitPython minimum version
- Keeps the SQLGlot compatibility test working across supported versions


tetelio (approved):
LGTM but please write a better PR message before merging :D

PR #4266, opened 2026-07-21
(feat) 4135-retry tenacity helper schema migrations

rudolfix:
### Description
alternative fix to #4135 (see #4138 )

1. it retries a failing schema migrations (with jitter and backoffs) via tenacity retry helper. 
```py
from tenacity import retry, stop_after_attempt, retry_if_exception, wait_random_exponential
from dlt.pipeline.helpers import retry_schema_update

if __name__ == "__main__":
    pipeline = dlt.pipeline(pipeline_name="chess_pipeline", destination="duckdb", dataset_name="games_data")

    @retry(
        stop=stop_after_attempt(5),
        # random jitter spreads retries so colliding parallel runs do not wake up and retry in lockstep
        wait=wait_random_exponential(multiplier=1, max=30),
        retry=retry_if_exception(retry_schema_update()),
        reraise=True,
    )
    def load():
        data = chess_source(["magnuscarlsen", "rpragchess"], start_month="2022/11", end_month="2022/12")
        return pipeline.run(data)

    load_info = load()
```
2. does not create comments on `_dlt` tables on databricks to prevent non-atomic CREATE TABLE
3. as a bonus - allows to disable comments creation (via TBLPROPS) altogether.


dlt/load/utils.py
```diff
@@ -219,12 +224,23 @@ def _init_dataset_and_update_schema(
...
+            only_tables=update_tables,
+            expected_update=expected_update,
+            # force schema update if tables dropped or truncated via refresh
+            force=bool(drop_tables or initial_truncate_tables),
+        )
+    except (DestinationSchemaTampered, DestinationUndefinedEntity):
+        # a tampered schema or a genuinely missing entity are terminal and must surface as-is
+        raise
+    except DestinationException as schema_ex:
+        raise SchemaUpdateError.from_cause(
```
burnash, line 238: i think the new wrapping breaks test_clickhouse_adapter_expr_fails here:
https://github.com/dlt-hub/dlt/actions/runs/30016811386/job/89251990451 also in remote https://github.com/dlt-hub/dlt/actions/runs/30016811386/job/89251990588

dlt/common/destination/exceptions.py
```diff
@@ -227,6 +227,46 @@ def __init__(self, schema_name: str, version_hash: str, stored_version_hash: str
         )
 
 
+class SchemaUpdateError(DestinationException):
+    """Raised when applying a schema update (DDL migration) to the destination fails."""
+
+    def __init__(
+        self,
+        schema_name: str,
+        table_names: Iterable[str],
```
burnash, line 236: It's Iterable[str] but the body does `list(table_names) if table_names is not None else None` below

burnash (changes requested):

rudolfix:
@burnash I fixed both things. athena errors are unrelated

PR #4269, opened 2026-07-22
(fix) uses case sensitive identifiers in sqlglot schema

rudolfix:
### Description
`sqlglot` 30.13.0 started to normalize identifiers in a few more cases. here we prevent case folding explicitly when we create sqlglot schema. this was already made when computing column schemas.

### DIFF
--- a/dlt/dataset/lineage.py
+++ b/dlt/dataset/lineage.py
@@ -74,7 +74,12 @@ def create_sqlglot_schema(
 
         nested_schema[dataset_name] = sqlglot_tables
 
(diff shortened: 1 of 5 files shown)


Travior (changes requested):
We should relax the upper bound on `sqlglot` to test this. Right now the `--upgrade` tests install:
` + sqlglot==30.12.0`

rudolfix:
> We should relax the upper bound on `sqlglot` to test this. Right now the `--upgrade` tests install: ` + sqlglot==30.12.0`

you are right! I was sure --upgrade ignores all constraints

PR #4276, opened 2026-07-23
(feat) snowflake nested types

rudolfix:
### Description
Adds support for snowflake nested types. At this moment user interface (piggy-backing on "json" data type) is experimental and will change.

* automatic support for parquet with nested fields
* support for jsonl/dicts with upfront `columns` definition (using pyarrow schema)
* schema and data type evolution (snowflake supports it)
* schema contracts on "data_type" do not see nested definitions now (so they will evolve)

see docs update for details


dlt/destinations/impl/snowflake/factory.py
```diff
@@ -119,6 +124,42 @@ def to_db_decimal_type(self, column: TColumnSchema) -> str:
...
+
+        if (
+            pyarrow.types.is_list(dtype)
+            or pyarrow.types.is_large_list(dtype)
+            or pyarrow.types.is_fixed_size_list(dtype)
+        ):
+            return f"ARRAY({self._to_nested_db_type(dtype.value_type, table)})"
+        if pyarrow.types.is_struct(dtype):
+            fields = ", ".join(
+                f"{dtype.field(i).name} {self._to_nested_db_type(dtype.field(i).type, table)}"
```
burnash: struct field names are interpolated into ddl unescaped, field names come from json keys and afaik don't go through the naming convention, so "-" or " " will brake create table ddl. (tested with e.g. pa.struct([("a-b", pa.string()), ("a b", pa.int64())])

dlt/destinations/impl/snowflake/snowflake.py
```diff
@@ -115,6 +117,11 @@ def run(self) -> None:
         if self._config.staging_config and self._config.staging_config.bucket_url:
             stage_bucket_url = self._config.staging_config.bucket_url
 
+        # null fields in structured columns only load with the vectorized scanner
+        use_vectorized_scanner = self._config.use_vectorized_scanner or (
```
burnash, line 123: this forces the vectorized scanner for every parquet load once use_nested_types is on, including tables with no nested columns, is this intentional? alternative is to scope it to tables which actually have x-nested-type - but not sure maybe something preventing that.

burnash (changes requested):
works for the case but there's an issue with ddl escaping, please see the comments

dlt/destinations/impl/snowflake/factory.py
rudolfix: @burnash good point! I'll investigate how identifiers in structs are handled

dlt/helpers/dbt/profiles.yml
```diff
@@ -140,7 +140,10 @@ athena:
       schema: "{{ var('destination_dataset_name', var('source_dataset_name')) }}"
       database: "{{ env_var('DLT__AWS_DATA_CATALOG') }}"
       # aws_profile_name: "{{ env_var('DLT__CREDENTIALS__PROFILE_NAME', '') }}"
-      work_group: "{{ env_var('DLT__ATHENA_WORK_GROUP', '') }}"
+      # only emit work_group when set: athena rejects an empty workGroup
```
burnash: i think the correct way to fix is

work_group: "{{ env_var('DLT__ATHENA_WORK_GROUP', 'primary') }}"

that worked for me. a newer dbt-athena release started passing the empty string through instead of dropping falsy values.

burnash: tried it here, let's see how it works on CI: https://github.com/dlt-hub/dlt/pull/4282

burnash (dismissed):

burnash (approved):

PR #4280, opened 2026-07-23
fix: avoid collisions in compound SCD2 merge keys

### LINKED ISSUE #4274: scd2/merge compound key comparison uses separator-less CONCAT, causing key collisions that retire wrong records
### dlt version

1.23.0 (still present on master: [`dlt/destinations/sql_jobs.py:377-378`](https://github.com/dlt-hub/dlt/blob/master/dlt/destinations/sql_jobs.py#L377))

### Describe the problem

`SqlMergeFollowupJob.gen_concat_sql` joins compound key columns with no separator:

```python
@classmethod
def gen_concat_sql(cls, columns: Sequence[str]) -> str:
    return f"CONCAT({', '.join(columns)})"
```

In `gen_scd2_sql` this is used to scope the retire statement by merge key:

```sql
UPDATE dataset.table SET valid_to = ...
WHERE valid_to IS NULL
  AND _dlt_load_id NOT IN (SELECT _dlt_load_id FROM staging.table)
  AND CONCAT(key_a, key_b) IN (SELECT CONCAT(key_a, key_b) FROM staging.table)
```

Distinct composite keys can collide after concatenation: `('doc-7', '12')` and `('doc-71', '2')` both concat to `'doc-712'`. When a record with key `('doc-71', '2')` arrives in an incremental batch, the *unrelated* active record `('doc-7', '12')` matches the `CONCAT ... IN` clause, fails the row-hash check, and is wrongly retired (`valid_to` set) even though it still exists in the source.

We hit this in production (BigQuery destination) on a table keyed by `(document_number, line_id)`: whenever a newer document whose number is a string-prefix extension of an older one appeared in an incremental batch, the older document's lines were falsely retired. The records then reappear on the next full load, so downstream sees rows flickering in and out of the active set.

### Expected behavior

Composite key comparison should be collision-free, e.g. insert a separator that is unlikely to appear in data:

```python
return f"""CONCAT({", '\\x1f', ".join(columns)})"""
```

or compare as a tuple / `EXISTS` join instead of a concatenated string. Related: a NULL in any key column currently makes the whole `CONCAT` NULL, silently excluding that row from the comparison — worth handling in the same fix (e.g. `COALESCE` with a null sentinel).

### Steps to reproduce

Create an scd2 table with compound `merge_key` and two rows keyed `('12', '3')` and `('1', '23')`; run an incremental load containing only one of them with changed content — the other row gets retired despite not being in the batch.

ANSHUL-REAL:
### Description

Avoid ambiguous compound-key comparisons when SCD2 identifies records to retire. Without a separator, `("doc-7", "12")` and `("doc-71", "2")` both become `doc-712`, which can retire an unrelated active record.

This adds a separator between key components in the shared SQL implementation and its Athena and SQLAlchemy variants, with a regression test for the collision.

### Related Issues

Closes #4274

### Validation

- `uv run ruff check dlt/destinations/sql_jobs.py dlt/destinations/impl/athena/athena.py dlt/destinations/impl/sqlalchemy/merge_job.py tests/load/pipeline/test_scd2.py`
- 6 focused native-DuckDB SCD2 tests, including the new collision regression

### DIFF
--- a/dlt/destinations/impl/athena/athena.py
+++ b/dlt/destinations/impl/athena/athena.py
@@ -135,7 +135,7 @@ class AthenaMergeJob(SqlMergeFollowupJob):
     def gen_concat_sql(cls, columns: Sequence[str]) -> str:
         # Athena requires explicit casting
         columns = [f"CAST({c} AS VARCHAR)" for c in columns]
-        return f"CONCAT({', '.join(columns)})"
+        return "CONCAT(" + ", '\\x1f', ".join(columns) + ")"
 
     @classmethod
     def requires_temp_table_for_delete(cls) -> bool:
--- a/dlt/destinations/impl/sqlalchemy/merge_job.py
+++ b/dlt/destinations/impl/sqlalchemy/merge_job.py
@@ -345,6 +345,7 @@ class SqlalchemyMergeFollowupJob(SqlMergeFollowupJob):
         # Cast because CONCAT is only generated for string columns
         result = sa.cast(result, sa.String)
         for col in columns[1:]:
+            result = operator.add(result, sa.literal("\\x1f"))
             result = operator.add(result, sa.cast(col, sa.String))
         return result  # type: ignore[no-any-return]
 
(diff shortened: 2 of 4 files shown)


burnash:
Thanks for the contribution. We're going with #4245 instead, as it looks like a better approach to solve this and has been opened earlier.

ANSHUL-REAL:
Thanks for the clear explanation and for pointing me to #4245. I appreciate you taking the time to review this. I will check for existing work and confirm the preferred approach before starting similar fixes in the future.

PR #4281, opened 2026-07-23
fix(snowflake): quote stage references and file paths in local file

### LINKED ISSUE #4278: Snowflake: local file loads fail when table name requires quoting
dlt version: 1.29.0

With a naming convention that preserves spaces in table names (e.g. duck_case), loading a local file into Snowflake fails with a SQL syntax error. Job files are named after the table, and the PUT statement interpolates both the file path and the table-stage reference unquoted:

PUT file:///.../my table.abc123.0.jsonl @dataset."%my table"/"<load_id>" OVERWRITE = TRUE, AUTO_COMPRESS = FALSE
-- 001003 (42000): SQL compilation error: syntax error ... unexpected 'table'.

Repro:

```python
# SCHEMA__NAMING=duck_case python repro.py
import dlt

pipe = dlt.pipeline(
    pipeline_name="space_table_repro",
    destination="snowflake",
    dataset_name="space_table_repro",
)
pipe.run([{"value": 1}], table_name="my table", loader_file_format="jsonl")
```

burnash:
Fixes #4278

### DIFF
--- a/dlt/destinations/impl/snowflake/snowflake.py
+++ b/dlt/destinations/impl/snowflake/snowflake.py
@@ -102,6 +102,7 @@ class SnowflakeLoadJob(RunnableLoadJob, HasFollowupJobs):
 
         qualified_table_name = self._sql_client.make_qualified_table_name(self.load_table_name)
         # this means we have a local file
+        stage_folder: str = ""
         stage_file_path: str = ""
         if is_local_file:
             if not self._stage_name:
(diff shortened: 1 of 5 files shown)


dlt/destinations/impl/snowflake/snowflake.py
```diff
@@ -131,13 +133,17 @@ def run(self) -> None:
         with self._sql_client.begin_transaction():
             # PUT and COPY in one tx if local file, otherwise only copy
             if is_local_file:
+                # backslash is an escape character in snowflake string literals, forward
+                # slashes are valid in PUT file URIs also on Windows
+                file_uri = "file://" + self._file_path.replace("\\", "/")
```
rudolfix, line 138: we have `make_file_url` function that does it and also more normalization but actually snowflake does not understand escaped file:// ... so this is the correct things to to

rudolfix (approved):
LGTM, thx for fixing this!

PR #4283, opened 2026-07-24
fix: add dot separator to get_table_prefix_layout when {table_name} terminates layout

### LINKED ISSUE #4265: Filesystem `layout="{table_name}"` makes `replace` delete prefix-named sibling tables
### dlt version

1.29.0 source checkout from current `devel` at `91403f1a39058230361820de414b718a22d7a423`.

### Describe the problem

When the filesystem destination uses `layout="{table_name}"`, replacing one table can silently delete an unrelated table whose name starts with the same prefix.

For example, after loading tables named `event` and `events`, replacing only `event` deletes the physical data file for `events`. The load completes successfully and rewrites `event`, so there is no failure or warning that unrelated data was removed.

The current write path automatically appends `.<full_extension>` when `{ext}` is omitted, producing files such as:

```text
event.jsonl.gz
events.jsonl.gz
```

However, table-prefix extraction returns the bare prefix `event`. Replace truncation selects files with `filepath.startswith(prefix)`, so both files are selected and deleted.

I reproduced this twice from a clean detached checkout of current upstream `devel`; both runs exited successfully with identical data loss.

### Expected behavior

Replacing `event` should update only `event.jsonl.gz`; `events.jsonl.gz` should remain present and unchanged.

If an end-of-layout `{table_name}` cannot be handled safely, the destination should reject the layout during validation instead of accepting it and performing an overbroad delete.

### Steps to reproduce

```python
import tempfile
from pathlib import Path

import dlt

root = Path(tempfile.mkdtemp(prefix="dlt-fs-prefix-repro-"))
bucket = root / "bucket"

pipeline = dlt.pipeline(
    pipeline_name="prefix_repro",
    destination=dlt.destinations.filesystem(
        bucket_url=str(bucket),
        layout="{table_name}",
    ),
    dataset_name="ds",
)

pipeline.run(
    dlt.resource([{"id": 1}], name="event"),
    loader_file_format="jsonl",
)
pipeline.run(
    dlt.resource([{"id": 2}], name="events"),
    loader_file_format="jsonl",
)

dataset = bucket / "ds"

def table_files():
    return sorted(path.name for path in dataset.glob("event*"))

print("before:", table_files())

pipeline.run(
    dlt.resource(
        [{"id": 3}],
        name="event",
        write_disposition="replace",
    ),
    loader_file_format="jsonl",
)

print("after:", table_files())
```

Actual output on current `devel`:

```text
before: ['event.jsonl.gz', 'events.jsonl.gz']
after: ['event.jsonl.gz']
```

A control run changing only the layout to `"{table_name}.{ext}"` preserves both files:

```text
before: ['event.jsonl.gz', 'events.jsonl.gz']
after: ['event.jsonl.gz', 'events.jsonl.gz']
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.13

### dlt data source

Inline Python resources containing dictionaries.

### dlt destination

Filesystem & buckets

### Other deployment details

The reproduction uses a local temporary filesystem and requires no credentials or external services. The affected prefix/deletion logic is shared by filesystem backends.

### Additional information

The root cause appears to span these paths:

- [`create_path`](https://github.com/dlt-hub/dlt/blob/91403f1a39058230361820de414b718a22d7a423/dlt/destinations/path_utils.py#L240-L245) appends a dot and the full extension when `{ext}` is omitted.
- [`get_table_prefix_layout`](https://github.com/dlt-hub/dlt/blob/91403f1a39058230361820de414b718a22d7a423/dlt/destinations/path_utils.py#L279-L289) intends to retain the separator after `{table_name}`, specifically to avoid selecting similarly named tables, but returns bare `{table_name}` when the placeholder terminates the layout.
- [`_delete_table_files`](https://github.com/dlt-hub/dlt/blob/91403f1a39058230361820de414b718a22d7a423/dlt/destinations/impl/filesystem/filesystem.py#L808-L819) deletes the files returned by prefix listing.
- [`list_files_with_prefixes`](https://github.com/dlt-hub/dlt/blob/91403f1a39058230361820de414b718a22d7a423/dlt/destinations/impl/filesystem/filesystem.py#L1029-L1045) uses raw `filepath.startswith(prefix)`.

The [filesystem layout documentation](https://github.com/dlt-hub/dlt/blob/91403f1a39058230361820de414b718a22d7a423/docs/website/docs/dlt-ecosystem/destinations/filesystem.md#L562-L569) says omitted extensions are automatically appended with a dot and that replace-compatible layouts need a separator after `{table_name}`.

A focused fix could make `get_table_prefix_layout("{table_name}")` derive `"{table_name}."`, matching the path produced by `create_path`. A stricter alternative is to reject the layout before loading.

Suggested regression coverage:

- Unit assertion for the exact end-of-layout case in `tests/destinations/test_path_utils.py`.
- Filesystem replace test with sibling tables `event` and `events`.

Focused validation on the clean upstream checkout:

```text
tests/destinations/test_path_utils.py: 70 passed
local-only tests/load/filesystem/test_filesystem_client.py: 53 passed, 1 skipped
```

AnnasMazhar:
## Breaking change

(this is possible but highly improbable case to happen in practice for delta or iceberg)

`get_table_prefix_layout` now keeps the separator that `create_path` appends when `{table_name}` terminates the layout. With `layout="{table_name}"` the table prefix becomes `event.` instead of `event`, which is what stops a `replace` of `event` from deleting `events`.

The table location moves with the prefix: a table that was at `<dataset>/event` is now at `<dataset>/event.`. Data files are unaffected because they always carry the extension, but delta and iceberg tables move — dlt creates a new empty table at the new location and the old data stays behind.

**If you used `layout="{table_name}"`, switch to `layout="{table_name}/"`.**

## Summary

Fixes a bug where filesystem layout `{table_name}` (without separator) makes `replace` truncation delete sibling tables whose names share the same prefix.

### Root Cause

When `{table_name}` terminates the layout and `{ext}` is omitted, `create_path` automatically appends `.<ext>` (e.g. `event.jsonl.gz`). However, `get_table_prefix_layout` returned only the bare `{table_name}` placeholder, producing an `event` prefix. The `replace` code path uses `startswith(prefix)` to select files for deletion, so `event` also matches `events.jsonl.gz`.

### Fix

When `{table_name}` is the last placeholder in the layout and `{ext}` is absent, `get_table_prefix_layout` now appends `.` to its return value. This matches the separator that `create_path` inserts, producing `event.` as the prefix — which correctly matches only `event.jsonl.gz`.

### Edge Cases Verified

| Layout | Prefix | Notes |
|--------|--------|-------|
| `{table_name}` | `event.` | Dot appended for implicit extension |
| `{schema_name}/{table_name}` | `my_schema/event.` | Same with path prefix |
| `{table_name}.{ext}` | `event.` | Dot from `{ext}` suffices, no extra dot |
| `{table_name}/` | `event/` | Slash separator suffices |
| `{schema_name}/{table_name}/{load_id}.{ext}` | `my_schema/event/` | Slash before `{load_id}` suffices |

### Tests

- **`test_get_table_prefix_layout_table_name_at_end`** — 5 assertions covering end-of-layout, with-extension-without-dot, slash-separator, and load_id-following edge cases
- All 71 existing `test_path_utils.py` tests continue to pass


AnnasMazhar:
👋 **Gentle bump** — this small fix (one function, dot-separator for terminating `{table_name}` layouts) has been open for a week with no activity. It addresses #4265 where `replace` deletes prefix-named sibling tables. Would appreciate a quick review.

AnnasMazhar:
Verified this fix against the issue's repro and current `devel` — evidence below.

**What I ran:**

1. **Issue repro on this branch** (two loads `event` + `events`, then `replace` on `event` with `layout="{table_name}"`):
   ```
   before: ['event.jsonl.gz', 'events.jsonl.gz']
   after:  ['event.jsonl.gz', 'events.jsonl.gz']   # both preserved ✅
   ```

2. **Control on clean upstream `devel`** (same script, no fix):
   ```
   before: ['event.jsonl.gz', 'events.jsonl.gz']
   after:  ['event.jsonl.gz']                       # events.jsonl.gz deleted — bug reproduces
   ```
   The regression test catches the real data-loss path.

3. **Unit suite:** `tests/destinations/test_path_utils.py` — 71 passed (includes the new `test_get_table_prefix_layout_table_name_at_end`).

4. **Branch state:** rebased onto current `devel` (was 7 commits behind; zero overlap with the touched files), all tests re-run green after rebase.

**On the CI gate:** the `dlt` workflow run for this branch is stuck in `action_required` — that's the fork-PR approval gate (no jobs failed; the run never started). If a maintainer can click **Approve and run** on the workflow, the checks will populate and the PR can proceed to review. No CI failure exists on this branch.

The fix itself is minimal and matches the issue's suggested direction: `get_table_prefix_layout` appends `.` when `{table_name}` terminates the layout and `{ext}` is absent, so prefix matching can't select sibling tables (`event` vs `events`).

dlt/destinations/path_utils.py
```diff
@@ -286,6 +286,12 @@ def get_table_prefix_layout(
             layout, "Table requires it's own folder, please add a '/' after your {table_name}. "
         )
 
+    # when {table_name} terminates the layout and {ext} is not present,
+    # create_path automatically prepends .<ext>, so the prefix must include
+    # the dot to avoid matching sibling tables (e.g. "event" matching "events")
+    if prefix[-1] == "}" and "ext" not in get_placeholders(layout):
+        prefix += "."
+
```
Travior: I think this is a bit brittle. We're basically trying to reconstruct the extension building logic we use in `create_path`:
```python
    # if extension is not defined, we append it at the end
    if "ext" not in placeholders:
        path += f".{job_info.full_extension()}"


```

Maybe we can create a shared helper that we use in `create_path` and here. If we create the invariant at the beginning that the passed layout will always contain a `{ext}` placeholder, then we wouldn't need any extra checks because the current path of "include one character after {table_name}" includes the `.` from the extension.

Travior: nit: I think the comment is a bit too verbose :)

tests/destinations/test_path_utils.py
```diff
@@ -419,6 +419,32 @@ def test_get_table_prefix_layout() -> None:
...
+
+    # {table_name}/ → slash separator suffices
+    prefix = get_table_prefix_layout("{table_name}/")
+    assert prefix == "{table_name}/"
+
+    # {schema_name}/{table_name}/{load_id}.{ext} → slash before {load_id} suffices
+    prefix = get_table_prefix_layout("{schema_name}/{table_name}/{load_id}.{ext}")
+    assert prefix == "{schema_name}/{table_name}/"
+
+
```
Travior: I think we can make this a bit more `pytest` idiomatic by parameterizing this unit test

Travior (changes requested):
Thanks for your contribution. I've left some comments.
In my opinion the most important thing that is still missing is a proper regression test around the actual file behavior (e.g. create `event` and `events`, show that we don't truncate `events` anymore)

AnnasMazhar:
Thanks for the review, @Travior. Pushed `8551cb3`:
- `get_table_prefix_layout` now normalizes the layout by appending `.{ext}` when missing (mirroring `create_path`), so the separator after `{table_name}` is always captured instead of reconstructing the extension logic.
- Added `test_replace_does_not_truncate_sibling_tables`: creates `event` and `events` files, then asserts truncating `event` no longer deletes `events` (fails on the old prefix logic).
- Parameterized the prefix unit test and trimmed the comment.

Unit suite: 76 passed. Regression test passes locally (file/memory buckets); cloud buckets run in CI.

AnnasMazhar:
Thanks for the detailed review, @Travior. All points are addressed now:

**Shared helper (main point):** pushed `f3d75ec` — extracted `_ensure_ext_in_layout(layout, placeholders)` which appends `.{ext}` to the layout when missing, and both `create_path` and `get_table_prefix_layout` now use it. This enforces the invariant you suggested (the layout always contains `{ext}`), so `get_table_prefix_layout` no longer reconstructs the extension logic, and the `.` after `{table_name}` is captured by the existing "include one char after table_name" slicing. The verbose comment is gone.

**Regression test (from `8551cb3`):** `test_replace_does_not_truncate_sibling_tables` creates `event` and `events` files and asserts truncating `event` no longer deletes `events` — fails on the old prefix logic, passes now (verified locally on file/memory buckets; cloud buckets run in CI).

**Parameterized unit test (from `8551cb3`):** `test_get_table_prefix_layout_table_name_at_end` covers dot/slash separators across 6 layouts.

Verified locally after the refactor: `tests/destinations/test_path_utils.py` — 76 passed, regression test passes, `ruff check` clean. The cloud-bucket tests need credentials so they run in CI.

AnnasMazhar:
Hi @Travior — all three review points are addressed in the commits pushed Aug 5 (8551cb306, f3d75ecab):

1. **Brittleness**: extracted shared `_ensure_ext_in_layout()` helper, now used by both `create_path` and `get_table_prefix_layout` — single source of truth for the `.{ext}` normalization.
2. **Parameterized unit test**: `test_get_table_prefix_layout_table_name_at_end` covers the table_name-terminates-layout case directly.
3. **Real file-behavior regression**: `test_replace_does_not_truncate_sibling_tables` creates `event.jsonl` + `events.jsonl`, truncates `event`, and asserts `events` survives — the actual collision case from #4265.

Verified locally before this comment: `tests/destinations/test_path_utils.py` → 76 passed; `tests/load/filesystem/test_filesystem_client.py -k 'truncate_sibling or prefix'` → 2 passed. Branch is 3 commits ahead of devel and mergeable.

Could you take another look when you have a moment? 🙏

rudolfix:
@AnnasMazhar @Travior thanks for the fix! this exposed a while class of similar problems where separators could collide with allowed chars in table names - leading to similar prefixing/delete problems. I've fixed them here (and pushed to fork).

rudolfix (approved):
LGTM!

PR #4287, opened 2026-07-26
fix(workspace): do not follow symlinks when selecting deployment files

### LINKED ISSUE #4273: bug(dlthub,cli): recursion error while running dlthub pipeline commands
### dlt version

1.29.0

### Describe the problem

I am currently hitting a bug while running `dlthub pipeline run pokemon_pipeline` seems like this command runs [this piece of code](https://github.com/dlt-hub/dlt/blob/devel/dlt/_workspace/deployment/file_selector.py#L81-L88) and currently I have a few symlinks in my system that create a recursion error, I managed to fix it locally by adding `follow_links=False` to the `iter_tree_files` function, but still this is not a proper solution. It seems that whatever file that is added to the .gitignore is ignored afterwards in the next line

### Expected behavior

I would expect for any folder that is in the `.gitignore` to be ignored by the dlthub CLI, instead of first traversing them.

### Steps to reproduce

Create a circular recursion with symlinks using `ln -s <source> <destination>` and add the folder into `.gitignore`

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

_No response_

### dlt destination

DuckDB

### Other deployment details

_No response_

### Additional information

_No response_

Sanjays2402:
### Description

`WorkspaceFileSelector.__iter__` walked the workspace with pathspec's `iter_tree_files`, which follows symlinks by default. A symlink pointing at one of its own parent directories makes the walk recurse until it raises `RecursionError`, so `dlthub pipeline run` failed outright — and adding the directory to `.gitignore` did not help, because the tree is walked before the ignore patterns are applied.

Passing `follow_links=False` makes the walk yield the link itself without descending through it.

### Related Issues

- Fixes #4273

### Additional Context

Added `test_file_selector_does_not_follow_symlink_loops` next to the existing selector tests; it fails with `RecursionError` without the change and passes with it. `tests/workspace/deployment/test_file_selector.py` is green (8 passed).

### DIFF
--- a/dlt/_workspace/deployment/file_selector.py
+++ b/dlt/_workspace/deployment/file_selector.py
@@ -83,7 +83,9 @@ class WorkspaceFileSelector(BaseFileSelector):
         from pathspec.util import iter_tree_files
 
         root_path = Path(self.root_path)
-        for file_path in iter_tree_files(self.root_path):
+        # do not follow symlinks: a symlink pointing at one of its own parents makes
+        # `iter_tree_files` recurse until it raises RecursionError
+        for file_path in iter_tree_files(self.root_path, follow_links=False):
             if not self.ignore_spec.match_file(file_path):
                 yield root_path / file_path, Path(file_path)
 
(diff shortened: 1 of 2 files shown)


zilto (approved):
Thanks for the fix. LGTM

PR #4289, opened 2026-07-26
(feat) emit input/output lineage in traces

rudolfix:
### Description
Lineage info is emitted in traces with data more or less similar to open lineage but in relational form that can be ingested with dlt (like the rest of traces).

A set of inputs (per resource + data location) is added to "extract" step trace. from that inputs dlt computes outputs (resource + data location) added to load step trace.

data location is similar to open lineage namespace and corresponds to ie. a dlt dataset, rest API or database with tables etc.

here's base for all inputs and outputs. `kind`, `resource_name` and `location` are required: without them a row adds nothing to what the trace already says. `location` is a **human readable scope**, not an identity - `localhost`, a private address or an in-memory database are reported as configured and must not be compared for joinability (a fingerprint may be added later for that).

```py
class TDataLocation(TypedDict):
    kind: str
    """Location type: `dataset`, `sql_database`, `filesystem`, `rest_api`, or a custom value."""
    resource_name: str
    """Resource that read from or wrote to this location."""
    location: str
    """Non-secret scope of the location, e.g. `postgresql://example.com:5432`, `s3://bucket`.
    Empty when the location has no public address, ie. a reverse ETL sink."""
    version: NotRequired[str]
    """Version of the location's contents as a whole."""
```

here's definition of dataset input / output. everything except the two dataset names is required, so a row keeps the same shape whatever destination it describes. `physical_dataset_name` is the name after normalization (`My_DataSet` -> `my_data_set`) and is what both sides agree on; `dataset_name` keeps what was configured, for display. `casefold` and `case_sensitive` come from the **runtime adjusted** capabilities (naming convention and destination config both change them), so a consumer can derive the destination-bound form of any identifier without dlt.

```py
class TSchemaReference(TypedDict):
    name: str
    version_hash: str


class TDatasetDataLocation(TDataLocation):
    schemas: List[TSchemaReference]
    """Schemas grouping the tables, each carrying its own version hash."""
    tables: List[str]
    """Tables touched, as plain names - dlt qualifies tables by dataset, not by schema."""
    destination_type: str
    destination_name: str
    destination_fingerprint: str
    """Identifies destinations whose `location` is not public, ie. motherduck. May be empty."""
    casefold: str
    """Name of the casefolding function the destination applies: `upper`, `lower` or `str`."""
    case_sensitive: bool
    """Whether the destination generates case sensitive identifiers, as adjusted at runtime."""
    dataset_name: NotRequired[str]
    """Logical dataset name as configured by the user, e.g. `My_DataSet`. Absent for sinks."""
    physical_dataset_name: NotRequired[str]
    """Normalized name as it exists in the store, e.g. `my_data_set`. Absent for sinks."""
```

`version` is a hash over the `version_hash` of all the schemas in the location, computed order independently, so a single scalar answers "did anything I read change".

for transformations inputs (per resource) are of "dataset" kind and several of them will be present if resource takes many datasets as arguments.

core sources were also instrumented:

- **sql_database** / **sql_table** - `kind="sql_database"`, the server scope (`postgresql://example.com:5432`), `database`, `db_schema` and the one table the resource reads. recorded when the resource is created, from the credentials that were passed in. the scope comes from a new `ConnectionStringCredentials.physical_location()` so credentials, dbapi driver and connect options are never part of it.
- **filesystem** - `kind="filesystem"`, the bucket scope and the `glob`. one entry for the glob, not per file: per file entries are unbounded and per file detail already lands in the loaded data as `FileItem.file_url`. recorded while the resource runs, because `bucket_url` may only be known after config resolution.
- **rest_api** - `kind="rest_api"`, `client.base_url` sanitized with `sanitize_url`, plus the endpoint **template** of the resource (never a resolved request url: unbounded cardinality and a secret leak surface). dependent resources are transformers and record their own endpoint, so they appear as well.

each of those subclasses `TDataLocation` with its own fields, ie.

```py
class TRestApiDataLocation(TDataLocation):
    endpoints: NotRequired[List[str]]
```

on top of that users can define their own inputs on resource. call it on the resource instance when you create one, or from inside a running resource via `dlt.current.resource()` when the location is only known once config is resolved. `kind` defaults to the name of the source the resource belongs to.

```py
class TOrdersApiLocation(TDataLocation):
    endpoints: NotRequired[List[str]]

@dlt.resource
def get_orders():
    yield [{"id": 1}, {"id": 2}]

@dlt.source
def orders_api():
    orders = get_orders()
    orders.add_input(TOrdersApiLocation(
        kind="rest_api",
        resource_name=orders.name,
        location="https://api.example.com",
        endpoints=["/orders"],
    ))
    return orders
```

recording is additive: no input means no entry, so traces of existing pipelines are unchanged. a transformer has no input of its own unless it reads one - being fed by another resource is not an input, that edge is already in `ExtractMetrics.dag`.

here's how inputs and outputs look in trace, for `sql_database` (sqlite) loaded into duckdb with `dataset_name="Marts"`:

```json
// trace__steps__extract_info__inputs
{
  "load_id": "1785098343.0313416",
  "extract_idx": 0,
  "kind": "sql_database",
  "resource_name": "customers",
  "location": "sqlite:///tmp/crm.db",
  "database": "/tmp/crm.db",
  "tables": ["customers"]
}

// trace__steps__load_info__outputs
{
  "load_id": "1785098343.0313416",
  "kind": "dataset",
  "resource_name": "customers",
  "location": "/tmp/marts.duckdb",
  "version": "W7OX1I+Eylroebrh6id4",
  "schemas": [{"name": "sql_database", "version_hash": "547e65vd7poMKv1Q9e5uk2Ul9dD1MBb0/XN01E0Ywi0="}],
  "tables": ["customers"],
  "destination_type": "dlt.destinations.duckdb",
  "destination_name": "duckdb",
  "destination_fingerprint": "",
  "casefold": "str",
  "case_sensitive": false,
  "dataset_name": "Marts_20260726083902",
  "physical_dataset_name": "marts_20260726083902"
}
```

loaded as `trace__steps__extract_info__inputs` and `trace__steps__load_info__outputs`, each with `__tables` / `__schemas` (and ie. `__endpoints`) child tables. the dataflow graph of your pipelines is then a SQL query:

```sql
select i.kind, i.location as read_from, o.physical_dataset_name as written_to, t.value as table_name
from trace__steps__load_info__outputs o
join trace__steps__load_info__outputs__tables t on t._dlt_parent_id = o._dlt_id
join trace__steps__extract_info__inputs i on i.resource_name = o.resource_name
```

a run writes one entry per (resource, location): a table spread over many load jobs is listed once, nested tables are attributed to the resource owning their root table, tables created by a dynamic `table_name` callable are picked up, and with `use_single_dataset=False` a single load step reports several datasets (`LoadInfo.dataset_name` can only name one of them).

`TRACE_ENGINE_VERSION` is bumped to `2` and the frozen trace contract in `tests/pipeline/cases/contracts/trace.schema.yaml` was regenerated - purely additive.

## todo
1. allow to define custom outputs for custom destinations (at least locations and any free data)
2. a big overhaul of how traces are generated

right now traces are generated in a relational form that is to store and query in the database. we plan upgrade trace emitters with other forms:

* oTel with a custom semantic convention for dlt pipelines (but mapping as many things into standard otel objects)
* open lineage both standalone or as a part of oTel traces

### DIFF
--- a/dlt/common/configuration/specs/connection_string_credentials.py
+++ b/dlt/common/configuration/specs/connection_string_credentials.py
@@ -46,6 +46,16 @@ class ConnectionStringCredentials(CredentialsConfiguration):
         except Exception:
             raise InvalidConnectionString(self.__class__, native_value, self.drivername)
 
+    def physical_location(self) -> str:
+        """Returns a human readable location of the data, ie. `postgresql://example.com:5432`."""
+        if self.host:
+            address = f"{self.host}:{self.port}" if self.port else self.host
+        else:
+            # file and memory backends have no server, the database locates them
+            address = self.database or ""
+        backend = self.drivername.split("+")[0] if self.drivername else ""
+        return f"{backend}://{address}" if backend else address
+
     def on_resolved(self) -> None:
         if self.password:
             self.password = self.password.strip()
--- a/dlt/common/destination/reference.py
+++ b/dlt/common/destination/reference.py
@@ -28,7 +28,12 @@ from dlt.common.destination.exceptions import (
     InvalidDestinationReference,
     UnknownDestinationModule,
 )
-from dlt.common.destination.client import DestinationClientConfiguration, JobClientBase
+from dlt.common.destination.client import (
+    DestinationClientConfiguration,
+    DestinationClientDwhConfiguration,
+    JobClientBase,
+)
+from dlt.common.metrics import TDatasetDataLocation, TSchemaReference, data_location_version
 from dlt.common.runtime.run_context import get_plugin_modules
 from dlt.common.schema.schema import Schema
 from dlt.common.typing import is_subclass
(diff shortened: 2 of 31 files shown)


sh-rp:
Follow up on transformation inputs:

- input `tables` should be the origin tables the transformation actually reads per resource, not all tables of the upstream dataset (available from `Relation.sqlglot_expression`)
- we may need to tell apart same-named tables coming from different datasets or schemas: separate datasets are already separate input rows, but two schemas in one dataset share a row and `tables` is a flat list of plain names, so the schema a table came from is lost

rudolfix:
> Follow up on transformation inputs:
> 
> * input `tables` should be the origin tables the transformation actually reads per resource, not all tables of the upstream dataset (available from `Relation.sqlglot_expression`)
> * we may need to tell apart same-named tables coming from different datasets or schemas: separate datasets are already separate input rows, but two schemas in one dataset share a row and `tables` is a flat list of plain names, so the schema a table came from is lost

1. `tables` work like you say. if `dataset` is an input - I take them from relation. if `output` I take them from actual completed jobs in load step
2. we do not qualify tables with schema names but with dataset names (we should to keep query purely "logical" but there were reasons not to do it). so when query is qualified we create an union schema where overlapping tables are union of the columns of tables from all schemas. if union is not possible (conflicting types) - query will not compile. This will almost never happen - before that data is materialized and incompatible types will fail at load. tldr;> you'll need to attribute names to schemas yourself. in dlt it is a few lines of code and Schema object has a function to union many schemas.

PR #4302, opened 2026-07-31
fix(filesystem): normalize OneLake directory probes

### LINKED ISSUE #4301: OneLake trailing-slash directory probes fail with 403 AuthenticationFailed in filesystem destination and Fabric staging
## dlt version

The affected directory-probe behavior is present in the generic filesystem client on current `devel` and applies independently of the credential implementation used to access OneLake.

## Describe the problem

A dlt filesystem load to Microsoft Fabric OneLake fails during storage initialization with:

```text
azure.core.exceptions.ClientAuthenticationError: Operation returned an invalid status 'Forbidden'
ErrorCode:AuthenticationFailed
```

The error looks like an identity or token problem, but the same Workspace Identity token can read and list the same path successfully. The actual trigger is a trailing slash on a directory path passed to an adlfs `exists` or `isdir` probe.

The generic filesystem client preserves a trailing separator in both places involved:

```python
@property
def dataset_path(self) -> str:
    return self.pathlib.join(self.bucket_path, self.dataset_name, "")

def get_table_dir(...):
    ...
    table_dir = self.pathlib.dirname(table_prefix) + self.pathlib.sep
    return table_dir
```

Those paths are then probed directly:

```python
if truncate_tables and self.fs_client.isdir(self.dataset_path):
    ...

if self.fs_client.exists(table_dir):
    ...
```

That behavior is normally harmless for object storage, but the OneLake Blob endpoint responds differently to the same existing directory depending only on the trailing slash.

This affects two use cases:

1. A direct `filesystem` destination writing under a Lakehouse `Files` path.
2. A Fabric Warehouse destination using OneLake through `staging=filesystem(...)`, which uses the same generic `FilesystemClient`.

## Environment

- Microsoft Fabric Spark Runtime 2.0
- Python 3.13
- Spark Job Definition invoked through a Data Pipeline
- Submitter: Fabric Workspace Identity / service principal
- `adlfs==2025.8.0`
- `fsspec==2026.6.0`
- `azure-core==1.39.0`
- `azure-storage-blob==12.30.0`
- OneLake URL form:
  `abfss://<workspace-id>@onelake.dfs.fabric.microsoft.com/<lakehouse-id>/Files/<bucket>`
- Azure account host: `onelake.blob.fabric.microsoft.com`
- Credential: a valid Azure SDK `TokenCredential` yielding Workspace Identity storage tokens

The storage token had the expected `https://storage.azure.com` audience and represented the Workspace Identity that submitted the Spark job.

## Evidence that this is path handling, not authorization

Using the exact same Workspace Identity token and target directory:

| Operation | Path | Result |
|---|---|---|
| Blob `HEAD` | path without trailing `/` | 200 |
| Blob `HEAD` | same path ending in `/` | 403 `AuthenticationFailed` |
| Blob container list | prefix ending in `/` | 200, children returned |
| DFS `getStatus` | same directory | 200 |
| `notebookutils.fs.ls` | same directory | succeeds |
| `AzureBlobFileSystem.isdir` | path without trailing `/` | `True` |
| `AzureBlobFileSystem.isdir` | path with trailing `/` | raises 403 |
| dlt/fsspec OneLake `isdir` | normalized path | `True` |
| dlt/fsspec OneLake `isdir` | path with trailing `/` | raises 403 |

Pre-acquiring and caching a storage token on the Spark driver did not change the failure.

## Steps to reproduce

This read-only reproducer can run inside a Fabric Spark Job Definition with a Lakehouse attached:

```python
from adlfs import AzureBlobFileSystem

credential = ...  # valid Azure SDK TokenCredential for the Workspace Identity
workspace_id = "<workspace-id>"
lakehouse_id = "<lakehouse-id>"
directory = (
    f"abfss://{workspace_id}@onelake.dfs.fabric.microsoft.com/"
    f"{lakehouse_id}/Files/<bucket>/<dataset>"
)

fs = AzureBlobFileSystem(
    account_name="onelake",
    account_host="onelake.blob.fabric.microsoft.com",
    credential=credential,  # valid Azure SDK TokenCredential for the Workspace Identity
)

assert fs.isdir(directory) is True
fs.isdir(directory + "/")  # ClientAuthenticationError: 403 AuthenticationFailed
```

The dlt-level reproducer is a filesystem pipeline with a replace resource and an already initialized dataset:

```python
import dlt
from dlt.common.configuration.specs import AzureCredentials

credential = ...  # valid Azure SDK TokenCredential for the Workspace Identity
credentials = AzureCredentials.from_credential(credential)
credentials.azure_storage_account_name = "onelake"
credentials.azure_account_host = "onelake.blob.fabric.microsoft.com"

destination = dlt.destinations.filesystem(
    bucket_url=(
        "abfss://<workspace-id>@onelake.dfs.fabric.microsoft.com/"
        "<lakehouse-id>/Files/<bucket>"
    ),
    credentials=credentials,
)

pipeline = dlt.pipeline(
    pipeline_name="onelake_repro",
    destination=destination,
    dataset_name="<dataset>",
)

@dlt.resource(write_disposition="replace")
def rows():
    yield {"id": 1}

pipeline.run(rows)
```

The first failing path is:

```text
FilesystemClient.initialize_storage
  -> self.fs_client.isdir(self.dataset_path)
  -> adlfs.spec.AzureBlobFileSystem._isdir
  -> _exists
  -> BlobClient.exists
  -> 403 AuthenticationFailed
```

Removing the trailing slash from `dataset_path` exposes the second occurrence:

```text
FilesystemClient.initialize_storage
  -> truncate_tables
  -> _delete_table_files
  -> self.fs_client.exists(table_dir)
  -> adlfs.spec.AzureBlobFileSystem._exists
  -> 403 AuthenticationFailed
```

Normalizing only `dataset_path` is therefore too narrow.

## Confirmed workaround

Normalizing only paths passed to existence probes fixes the load:

```python
def probe_path(path):
    return path.rstrip("/") or path
```

With `exists` and `isdir` using that normalized path while list, read, write, and delete paths remain unchanged:

- storage initialization and table truncation succeed;
- the filesystem load completes;
- post-run OneLake listings show the new load-state and dataset files.

## Expected behavior

OneLake filesystem destinations should initialize, truncate, and load under a valid Fabric identity without applications having to proxy dlt's filesystem client. Fabric Warehouse loads using OneLake staging should receive the same behavior.

## Suggested fix scope

The most targeted behavior is:

- detect OneLake from the bucket or Azure account host;
- normalize non-root paths with `path.rstrip("/") or path` before `exists` and `isdir` probes;
- preserve trailing separators for prefix calculation, listing, writes, and deletes;
- keep all non-OneLake filesystem behavior unchanged.

The filesystem client already contains a OneLake-specific delete workaround for the service returning HTTP 200 where the Azure SDK expects 202. Centralizing this directory-probe quirk in the same client avoids application-specific wrappers.

## Additional information

The OneLake service or adlfs may ultimately be the lower-level source of the inconsistent response. Handling the probe path in dlt provides stable behavior for both direct Lakehouse filesystem destinations and Fabric Warehouse staging.

sdebruyn:
## Summary

- Route OneLake filesystem configurations to a dedicated `OneLakeFilesystemClient` subclass
- Keep the generic `FilesystemClient` provider-agnostic with neutral `exists` and `isdir` hooks
- Strip trailing separators only in the OneLake subclass before directory probes
- Preserve existing paths for listing, writing, deletion, and prefix calculation

## Root cause

OneLake's Blob endpoint returns `403 AuthenticationFailed` when an existing directory is probed with a trailing slash. The same path succeeds without that separator, so valid Fabric identities could fail during filesystem storage initialization and table truncation.

This affects both direct filesystem destinations and Fabric Warehouse loads using OneLake through `staging=filesystem(...)`, because staging uses the same filesystem client probes before `COPY INTO` runs.

## Test plan

- Added factory coverage for OneLake selection through DFS bucket URLs and Blob account hosts, including environment configuration
- Added regression coverage for storage initialization and table-truncation probes
- Verified regular Azure filesystem configurations still use the generic client and retain existing trailing separators
- Local filesystem client and factory suites: 90 passed, 2 skipped
- Ruff, mypy, Black, and diff checks passed

Closes #4301

### DIFF
--- a/dlt/destinations/impl/filesystem/factory.py
+++ b/dlt/destinations/impl/filesystem/factory.py
@@ -1,4 +1,5 @@
 from typing import Any, Optional, Type, Union, Dict, TYPE_CHECKING, Sequence, Tuple
+from urllib.parse import urlparse
 
 from dlt.common.configuration import configspec
 from dlt.common.configuration.resolve import resolve_configuration
@@ -13,8 +14,23 @@ from dlt.destinations.impl.filesystem.configuration import (
     HfFilesystemDestinationClientConfiguration,
 )
 from dlt.destinations.impl.filesystem.filesystem import FilesystemClient, HfFilesystemClient
+from dlt.destinations.impl.filesystem.onelake import OneLakeFilesystemClient
 from dlt.destinations.impl.filesystem.typing import TCurrentDateTime, TExtraPlaceholders
 from dlt.common.normalizers.naming import NamingConvention
+
+_ONELAKE_HOSTS = {
+    "onelake.blob.fabric.microsoft.com",
+    "onelake.dfs.fabric.microsoft.com",
+}
+_AZURE_PROTOCOLS = {"az", "abfs", "adl", "abfss", "azure"}
+
+
+def _is_onelake_host(value: Optional[str]) -> bool:
+    if not value:
+        return False
+    parsed = urlparse(value if "://" in value else f"//{value}")
+    return (parsed.hostname or "").lower().rstrip(".") in _ONELAKE_HOSTS
+
 
 if TYPE_CHECKING:
     from dlt.common.libs.ibis import BaseBackend
@@ -71,7 +87,11 @@ class filesystem(Destination[FilesystemDestinationClientConfiguration, Filesyste
 
     @property
     def client_class(self) -> Type[FilesystemClient]:
-        return HfFilesystemClient if self.is_hf else FilesystemClient
+        if self.is_hf:
+            return HfFilesystemClient
+        if self.is_onelake:
+            return OneLakeFilesystemClient
+        return FilesystemClient
 
     def create_ibis_backend(
         self, client: "FilesystemClient", read_only: bool = False, schemas: "Sequence[Schema]" = ()
@@ -153,6 +173,19 @@ class filesystem(Destination[FilesystemDestinationClientConfiguration, Filesyste
         resolved = resolve_configuration(_BucketUrlConfig(), sections=sections)
         return resolved.bucket_url
 
+    def _resolve_azure_account_host(self, destination_name: Optional[str]) -> Optional[str]:
+        @configspec
+        class _AzureAccountHostConfig(BaseConfiguration):
+            azure_account_host: Optional[str] = None
+
+        sections = (
+            *FilesystemDestinationClientConfiguration.__recommended_sections__,
+            self.resolve_destination_name(destination_name),
+            "credentials",
+        )
+        resolved = resolve_configuration(_AzureAccountHostConfig(), sections=sections)
+        return resolved.azure_account_host
+
     def __init__(
         self,
         bucket_url: str = None,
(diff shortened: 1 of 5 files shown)


sdebruyn:
fixes #3866

PR #4303, opened 2026-08-02
(feat) add profile and location of resolved config value to trace

rudolfix:
### Description
* adds profile to `run_context` element in trace (shape change)
* adds location of resolved config/secret in resolver trace

(also fixes skills that create worktrees - new branches origins got borked)

@sh-rp will those be stored in runtime and somehow visible in UI? if not we need another ticket

Where are new values in trace:

* `trace.execution_context["run_context"]` — was the context name as `str`, now a dict with `name` and (when the context supports profiles) `profile`. 
* `trace.resolved_config_values[*].provider_location` — new trailing field on `SerializableResolvedValueTrace`

Examples:

Pipeline run in a workspace on the `dev` profile, serialized with `trace.asdict()` (what gets loaded into the destination / sent to the platform).

`execution_context`:

```json
{
  "ci_run": false,
  "python": "3.14.4",
  "cpu": 24,
  "exec_info": ["claude_code", "claude_code_cli"],
  "os": {
    "name": "Linux",
    "version": "6.6.87.2-microsoft-standard-WSL2"
  },
  "library": {
    "name": "dlt",
    "version": "1.30.0a0"
  },
  "run_context": {
    "name": "default",
    "profile": "dev"
  }
}
```

`resolved_config_values` (`value` and `default_value` are dropped by `asdict`). `config_val` is only in `config.toml`, `config_val_ovr` is in both files and the profile one won, `secrets_val_dev` is only in `dev.secrets.toml`:

```json
[
  {
    "key": "config_val",
    "is_secret_hint": false,
    "sections": [],
    "provider_name": "config.toml",
    "config_type_name": "DataConfiguration",
    "provider_location": "config.toml"
  },
  {
    "key": "config_val_ovr",
    "is_secret_hint": false,
    "sections": [],
    "provider_name": "config.toml",
    "config_type_name": "DataConfiguration",
    "provider_location": "dev.config.toml"
  },
  {
    "key": "secrets_val_dev",
    "is_secret_hint": true,
    "sections": [],
    "provider_name": "secrets.toml",
    "config_type_name": "DataConfiguration",
    "provider_location": "dev.secrets.toml"
  }
]
```

How locations is set for different providers:

| Provider | location of a resolved value |
|---|---|
| `ConfigTomlProvider` / `SecretsTomlProvider` | `config.toml`, `secrets.toml` |
| `ProfileConfigTomlProvider` / `ProfileSecretsTomlProvider` | exact file: `dev.config.toml` when the profile file held the value, `config.toml` when it fell through to the base file |
| `GoogleSecretsProvider` | vault identity, ie. `sa@x.com@chat-analytics` |
| `AwsSecretsManagerProvider` | account, region and prefix, ie. `fake_key@eu-central-1:dlt/` |
| `AirflowSecretsTomlProvider` | `""` |
| `EnvironProvider`, `ContextProvider`, `DictionaryProvider`, `StringTomlProvider` | `""` |
| `CustomLoaderDocProvider` | whatever `locations` the caller registered, else `""` |

Values written at runtime with `set_value` are not tagged and fall back to the provider location.

### DIFF
--- a/.claude/skills/create-worktree/SKILL.md
+++ b/.claude/skills/create-worktree/SKILL.md
@@ -47,6 +47,7 @@ git worktree list
   - If no conflict, create the worktree:
     - If `--branch` was given: `git worktree add .worktrees/<name> <ref>`
     - Otherwise: `git worktree add .worktrees/<name> --detach`
+    - Callers needing a new branch check it out afterwards: `git checkout --no-track -b <new-branch>`
     - If `--pr` was given, then also: `cd <worktree-path> && gh pr checkout <number>`
 - `gh pr checkout` handles fetching, fork tracking, and branch setup automatically.
 
--- a/.claude/skills/implement-issue/SKILL.md
+++ b/.claude/skills/implement-issue/SKILL.md
@@ -94,14 +94,17 @@ Use the category from step 2 and the issue number:
 
 Example: `fix/3529-dedup-sort-escape`
 
+If `ISSUE` was already a branch name rather than a number (no ticket exists), use it as given.
+
 #### 4b. Create worktree and dev environment
 
-Invoke the `/worktree-issue` skill with the branch name:
+Invoke the `/worktree-from-issue` skill with the branch name:
 ```
-/worktree-issue {branch-name}
+/worktree-from-issue {branch-name}
 ```
 
 This creates the worktree, checks out a new branch from `origin/devel`, and runs `/worktree-make-dev`.
+Use it also when there is no issue number.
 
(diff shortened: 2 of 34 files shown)


sh-rp:
@rudolfix this information currently is not consumed in the runtime. We will need to make a small adjustment to the telemetry to surface this (in the raw trace it will be visible of course). Nothing should break though.

sh-rp:
PS: Let's do this fix to the runtime after these new traces are released, then I can directly add them into the e2e tests.

PR #4305, opened 2026-08-04
feat(destinations): add cdc merge strategy for duckdb and snowflake

### LINKED ISSUE #3655: Enhanced Merge Logic with Delete Support and Change Data Capture for Snowflake Streams Integration
### Feature description

We need an enhanced merge operation in DLT that goes beyond the current upsert functionality to support a complete Change Data Capture (CDC) workflow compatible with Snowflake Streams. Specifically, we need a merge strategy that:
- Detects and applies inserts, updates, and deletes dynamically within a single merge operation
- Supports comparison beyond primary key matching (e.g., detecting actual data changes similar to SCD2 logic)
- Integrates seamlessly with Snowflake Streams to capture before/after images of changes
- Allows for conditional filtering on source data to optimize performance on large tables
- Performs change detection at ingestion time rather than deferring it to downstream processes

### Are you a dlt user?

Yes, I run dlt in production.

### Use case

The customer is implementing a Kappa architecture that combines batch ELT via DLT with streaming ingestion via HVR. 

Problem Statement:
- We ingest data from heterogeneous sources (APIs, databases, CRM systems like Microsoft Dynamics, and consumption data)
- Some source systems (like our CRM) perform hard deletes without any deletion flags or markers
- Our target is Snowflake with a data vault model architecture

We need our ingestion target tables in Snowflake to reflect the source data accurately, including capturing when records are deleted

Currently, DLT's upsert operation only handles inserts and updates on primary keys, not deletes

Without delete detection at ingestion time, we must perform full table comparisons downstream, which is inefficient and computationally expensive in Snowflake (creates unnecessary micro-partitions)

Current Workaround Limitations:

-We're forced to read full datasets for accurate CDC, which is not scalable
-We cannot use persistent markers for deletes since they don't exist in the source
-Our only option is full target CDC (comparing entire source and target), or filtered CDC (applying filters to both source and target for tables where we know change windows)
-Using dbt extensions to implement custom merge logic moves the responsibility outside DLT, complicating maintenance and pipeline orchestration

Desired End-State: After DLT ingestion, our target table should:

-Contain exactly the same data as the source
-Reflect all changes: inserts (new records), updates (changed records), and deletes (removed records)
-Work seamlessly with Snowflake Streams to capture change history for downstream data vault processing
-Minimize computational overhead by detecting changes at ingestion time

### Proposed solution

We propose implementing an enhanced merge operation in DLT specifically for Snowflake that supports the following:

**A) Enhanced Merge Strategy (e.g., merge_with_deletes or similar)**

Implement a new merge mode that:
- Performs a full outer join between source and target data
 - Detects three types of changes:
1. Inserts: Records present in source but not in target
2. Updates: Records in both source and target where at least one non-key column has changed
3. Deletes: Records present in target but not in source (indicating hard deletes)
- Applies all three operations within a single merge statement to Snowflake
- Supports both primary key matching AND column-level change detection (not just key existence)

**B) Snowflake Streams Integration**
- Native integration with Snowflake Streams for change capture
- Ensure merge operations work with Snowflake's stream technology to capture before/after images
- Allow downstream processes to consume changes from streams without requiring additional full table comparisons

**C) Conditional Filtering Support**
- Allow users to specify filter logic on both source and target data
- For large tables where changes are known to occur within specific time windows (e.g., "only the last 3 months can change"), apply the same filters to both source and target
- Reduce computational overhead by limiting comparisons to relevant data ranges

**D) Configuration Example (Pseudo-code)**

dlt.resource( name="target_table", write_disposition="merge", merge_key=["primary_key"], merge_type="full_cdc", # New merge type detect_deletes=True, # Enable delete detection detect_updates=True, # Detect actual changes, not just key matches filter_condition="last_modified >= date_sub(current_date(), 3)" # Optional filtering ) def my_resource(): return source_data

**E) Implementation Considerations**
- This feature should work within DLT's ingestion pipeline, not require downstream dbt extensions
- The merge logic should be optimized to avoid unnecessary micro-partition creation in Snowflake
- Support both full dataset comparisons and filtered comparisons based on user configuration
- Maintain backward compatibility with existing upsert operations

**F) Related Technology Context**
- Snowflake Streams: Tracks inserts, updates, and deletes with before/after images based on micro-partition changes
- Snowflake Tasks: Can be scheduled to trigger pipelines when stream data is available
- Data Vault Modeling: Requires change history tracking for proper historization in raw and hub/link layers

### Related issues

_No response_

adrienModeo:
Closes #3655

## What

Adds a new `cdc` merge strategy. Unlike `upsert`, it treats the loaded data as a
**complete snapshot of the source** and derives inserts, updates, and deletes
from it in a single `MERGE`:

- **insert** rows present in the snapshot but not in the target
- **update** rows whose non-key columns changed (`IS DISTINCT FROM`)
- **delete** rows present in the target but absent from the snapshot

Unchanged rows are never rewritten, so downstream change consumers (e.g.
Snowflake Streams) only see rows that were actually inserted/updated/deleted.

## Usage

```python
@dlt.resource(
    primary_key="id",
    write_disposition={"disposition": "merge", "strategy": "cdc"},
)
def items(): ...
```

- **`merge_filter`** — optional SQL condition applied to both the snapshot and
  the target so only a window (e.g. last 3 months) is compared and deleted.
- **`hard_delete`** hint — flagged rows are deleted even when present in the
  snapshot, and never inserted.
- **nested tables** — nested rows follow their parents on delete; removed list
  elements are deleted.
- **composite primary keys** — supported.

## Supported destinations

- `duckdb` (requires `duckdb` >= 1.4.0 — needs `DELETE` in `MERGE`)
- `snowflake`

Gated via `supported_merge_strategies`; other destinations reject `cdc`.

## Limitations

The snapshot's `primary_key` must be unique — dlt does **not** deduplicate for
this strategy (documented in `merge-loading.md`). A non-unique key produces a
nondeterministic `MERGE` on Snowflake.

## Tests

New tests in `test_merge_disposition.py`: basic strategy, `merge_filter`,
`hard_delete`, nested tables, and composite primary key. Verified on **DuckDB**
and on a **live Snowflake** account.

The change is purely additive — existing strategies (delete-insert / upsert /
scd2 / insert-only) are untouched and their tests still pass.


rudolfix:
@adrienModeo 
This is great and tbh. I'd go much further and support this on all destination and integrate with existing mechanisms that do similar/same things (scd2). What I'd change here:
- it looks like almost no new SQL code is required. What we need to do is to modify `upsert` SQL to support records deletion, merge filters etc.
- I'd support "merge_key" with exactly the same semantics as scd2. both `cdc` and `scd2` assume that your input data is a complete snapshot. and until now "merge_key` was defining how "wide" snapshot is. no merge key means it is all the data
- x-merge-filter is a refinement of "merge_key" and should be available everywhere merge key is available
- it should be pretty easy to support that on all destinations that support "upsert" (also filesystem based)

having your code and few other assumptions I can come with improvements to your PR pretty quickly. tell me if it is OK to push to your fork.

adrienModeo:
Hello @rudolfix, amazing ! Sure go ahead you can add commits

albertotonon:
Hey @rudolfix,

thanks for your motivation to extend this work!

My team, which @adrienModeo is consulting for, is planning to use this functionality in one of our products. To help us plan our work, do you have a rough idea of when you expect the additional changes you mentioned to land?

Alternatively, if those changes are likely to take longer, would it make sense to first merge the current (more limited) version of the PR, after addressing any necessary review feedback, and extend it afterwards?

Knowing roughly when this functionality may become available in dlt core would help us plan accordingly.

PR #4324, opened 2026-08-07
docs(hub): document refresh a single resource in triggers

ShreyasGS:
### Description

Adds a "Refresh a single resource" subsection under `hub/pipeline-operations/triggers.md#refresh-cascade`, plus a short "three sources of refresh" bullet list to make signal origins explicit.

## Why
Users following the `refresh="always"` pattern in `triggers.md` had no on-page guide for the common case of "rebuild only these named tables". They'd hit `KeyError`-shaped confusion or reach for `drop_sources` (which wipes the whole source). The addition covers the missing patterns and links out to the canonical refresh-mode reference.

## What's new in `triggers.md#refresh-cascade`
- **Three sources of `run_context["refresh"]`**: first run (never completed), explicit `--refresh` on CLI, or cascade from an upstream `refresh="always"` (with `refresh="block"` severing).
- **Scoped rebuild pattern**: `@run.pipeline(..., refresh="always")` combined with `pipeline.run(source.with_resources(*resources), ...)` and `pipeline.refresh = "drop_resources"`.
- **Config-driven resource list**: `[jobs.<module>.<job>]` in `.dlt/<profile>.config.toml`, so `dev` and `prod` can name different tables without a code change.
- **CLI override**: `dlthub local run refresh_tables -c 'resources=["..."]'` for one-off runs.
- **Bulk trigger by tag**: `dlthub job trigger tag:refresh`.
- **Cross-link** to [Doing a full or partial refresh](../../general-usage/incremental-loading.md#doing-a-full-or-partial-refresh) for `drop_data` vs `drop_resources` vs `drop_sources` semantics.

### DIFF
--- a/docs/website/docs/hub/pipeline-operations/triggers.md
+++ b/docs/website/docs/hub/pipeline-operations/triggers.md
@@ -132,9 +132,11 @@ Refresh policies:
 
 | Policy | Behaviour |
 |--------|-----------|
-| `"always"` | Originate a refresh signal on every run |
+| `"always"` | Clear the freshness of every transitively downstream job when this job starts, so each of them refreshes on its next run |
 | `"auto"` | Pass through any refresh signal received from upstream (default) |
-| `"block"` | Stop refresh propagation here |
+| `"block"` | Never refresh this job: an inbound cascade stops here, and an explicit `--refresh` on the job is ignored and logged |
+
+The cascade fires when the run starts, not when it completes, so a run that later fails has already cleared the freshness of everything downstream. The walk is transitive: it follows the freshness graph past the direct neighbours, stopping at `refresh="block"` jobs and at jobs that manage their own interval watermark.
 
 ```py
 @run.job(expose={"tags": ["backfill"]}, refresh="always")
@@ -149,11 +151,9 @@ dlthub job trigger "tag:backfill"
 dlthub run backfill --refresh    # explicit refresh on a single job
 ```
 
-`run_context["refresh"]` is `True` in three cases:
+On the platform, `run_context["refresh"]` is also `True` on a job's first run, since the job has never completed successfully and has no previous run to compare against. That makes the first run a natural backfill.
 
-- **First run:** the job has never completed successfully, which is the natural moment for a backfill.
-- **Explicit request:** you pass `--refresh` to `dlthub run`, `dlthub local run`, or `dlthub job trigger`.
-- **Cascade:** an upstream job with `refresh="always"` (or a `--refresh` triggered on an upstream) propagates the signal through the freshness graph. `refresh="block"` on a job severs the cascade at that node.
+Locally there is no completion history, so `dlthub local run` resolves the refresh signal from `--refresh` and the declared policy alone.
 
 Note that the refresh signal will not drop your data automatically, you should use one of the [refresh](../../general-usage/pipeline.md#refresh-pipeline-data-and-state) options available.
 ```py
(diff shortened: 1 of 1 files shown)


lis365b (commented):
checked this against the runtime code, comments inline. the snippet itself works: the pipeline.refresh assignment is honored, both toml section forms resolve, and tag:refresh is synthesized from expose.tags so the job needs no declared trigger.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -165,6 +171,56 @@ def build_report(run_context: TJobRunContext):
 ```
 Above we tell `dlt` to truncate all tables belonging to resources in `data_source()` if the refresh signal got passed in the `refresh` flag.
 
+### Refresh a single resource
+
+For a targeted rebuild, combine `refresh="always"` with `with_resources(*resources)` and drive the resource list from configuration so one job covers every table. The job below drops and reloads only the tables named at runtime:
```
lis365b: refresh="always" does not affect the scoping here. the body sets pipeline.refresh = "drop_resources" and never reads run_context["refresh"].

what it does affect: start_run sees policy == "always" and clears prev_completed_run on this job and everything transitively downstream in the freshness graph (block nodes excluded). those jobs get run_context["refresh"] = True on their next run and their interval window restarts, so the effect reaches past the resources named here.

either remove refresh="always", or read run_context["refresh"] in the body and document the downstream effect.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -149,6 +149,12 @@ dlthub job trigger "tag:backfill"
 dlthub run backfill --refresh    # explicit refresh on a single job
 ```
 
+`run_context["refresh"]` is `True` in three cases:
+
+- **First run:** the job has never completed successfully, which is the natural moment for a backfill.
+- **Explicit request:** you pass `--refresh` to `dlthub run`, `dlthub local run`, or `dlthub job trigger`.
+- **Cascade:** an upstream job with `refresh="always"` (or a `--refresh` triggered on an upstream) propagates the signal through the freshness graph. `refresh="block"` on a job severs the cascade at that node.
```
lis365b: block also ignores an explicit --refresh on the job itself, not just the cascade. start_run skips the clear and logs the request as ignored. add that to the bullet above.

also: the policy table says always originates on every run, but the cascade fires at run start, not on completion. a failed refresh run still clears prev_completed_run downstream.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -165,6 +171,56 @@ def build_report(run_context: TJobRunContext):
...
+dlthub local run refresh_tables -c 'resources=["purchases"]'
+```
+
+Then trigger every refresh job on the platform with a tag selector:
+
+```sh
+dlthub job trigger tag:refresh
+```
+
+See [doing a full or partial refresh](../../general-usage/incremental-loading.md#doing-a-full-or-partial-refresh) for `drop_data` vs `drop_resources` vs `drop_sources` semantics. Use `drop_data` when the schema is still correct and only the rows should be replaced. Use `drop_resources` when the schema changed and the table should be recreated.
```
lis365b: add the drop_sources caveat here instead of only linking it: it ignores with_resources and drops the whole source.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -165,6 +171,56 @@ def build_report(run_context: TJobRunContext):
...
+```
+
+Configure the resource list per profile in `.dlt/<profile>.config.toml`:
+
+```toml
+[jobs.refresh_jobs.refresh_tables]
+resources = ["inventory"]
+```
+
+The section key is `[jobs.<module>.<job>]`. Use `[jobs.<module>]` to configure every job in the module. Since configuration is per profile, `dev.config.toml` can point at a small table for fast local runs while `prod.config.toml` names the real one.
```
lis365b: the middle segment is the module name, and the snippet above has no filename. in jobs.py the key would be [jobs.jobs.refresh_tables]. name the file in the snippet, or state that the segment is the module.

lis365b (commented):
four smaller ones.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -165,6 +171,56 @@ def build_report(run_context: TJobRunContext):
...
+
+Override on the CLI for a one-off:
+
+```sh
+dlthub local run refresh_tables -c 'resources=["purchases"]'
```
lis365b, line 234: -c only exists on dlthub local run, there is no equivalent on dlthub run. state that on the platform the resource list comes from the profile config.toml uploaded at deploy, and that this line is the local check before deploying.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -165,6 +171,56 @@ def build_report(run_context: TJobRunContext):
...
+```
+
+Then trigger every refresh job on the platform with a tag selector:
+
+```sh
+dlthub job trigger tag:refresh
```
lis365b, line 243: job trigger only matches deployed jobs and does not sync. add dlthub deploy before this.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -165,6 +171,56 @@ def build_report(run_context: TJobRunContext):
 ```
 Above we tell `dlt` to truncate all tables belonging to resources in `data_source()` if the refresh signal got passed in the `refresh` flag.
 
+### Refresh a single resource
```
lis365b: this uses expose={"tags": [...]} and job trigger tag:... before the section that introduces them. the cascade and explicit-request bullets also repeat the policy table and the dlthub run backfill --refresh line above. suggest keeping only the first-run bullet and moving this subsection below tags and bulk triggering.

docs/website/docs/hub/pipeline-operations/triggers.md
```diff
@@ -149,6 +149,12 @@ dlthub job trigger "tag:backfill"
 dlthub run backfill --refresh    # explicit refresh on a single job
 ```
 
+`run_context["refresh"]` is `True` in three cases:
+
+- **First run:** the job has never completed successfully, which is the natural moment for a backfill.
+- **Explicit request:** you pass `--refresh` to `dlthub run`, `dlthub local run`, or `dlthub job trigger`.
```
lis365b: the first-run rule above is platform state. locally there is no completion history, refresh is resolved from --refresh and the declared policy only. say the list describes platform runs, and keep dlthub local run to the explicit case.

lis365b (changes requested):

PR #4336, opened 2026-08-10
full-loading page does not say which replace strategies each destination supports (test, ignore)

ShreyasGS:
Documentation for #4334: full-loading page does not say which replace strategies each destination supports (test, ignore)

Closes #4334

## What changed

Edited the existing page `general-usage/full-loading.md`. The sidebar is unchanged.

Its runnable code is in `full-loading-snippets.py`, which pytest executed during verification.

Sections: *Performing a full load*, *Choosing the correct replace strategy for your full load*.

## How this was checked

All claims checked against the checkout (dlt installed from source in docs/.venv).

Source read:
- `dlt/common/destination/utils.py::resolve_replace_strategy` — applies `capabilities.replace_strategies_selector` (if any) to `capabilities.supported_replace_strategies`, returns `None` when the list is empty or when `required_strategy` is not in it, otherwise `required_strategy or supported_replace_strategies[0]`. So the default is the first entry of the (possibly narrowed) list, not a hardcoded `truncate-and-insert`.
- `dlt/destinations/utils.py::verify_schema_replace_disposition` raises `SchemaCorruptedException("Requested replace strategy ... not available for table ...")`; `dlt/destinations/job_client_impl.py::prepare_load_table` asserts on the same condition. Either way the load step fails — there is no fallback.
- Per-table narrowing exists only in `dlt/destinations/impl/filesystem/factory.py::filesystem_replace_strategies_selector` and `dlt/destinations/impl/athena/factory.py::athena_replace_strategies_selector`: `delta`/`iceberg` tables → `["insert-from-staging"]`, all other tables → `["truncate-and-insert"]`.
- Runtime-only restriction quoted for ClickHouse: `dlt/destinations/impl/clickhouse/clickhouse.py::_verify_database_supports_exchange` raises `DestinationTerminalException` unless the database engine is Atomic or Shared.

Capabilities dumped from the branch (`<factory>().capabilities().supported_replace_strategies`), showing `truncate-and-insert` is always first where present: postgres/bigquery/snowflake/mssql/databricks/clickhouse have all three; duckdb, motherduck, redshift, synapse, dremio, sqlalchemy, athena, filesystem have `truncate-and-insert` + `insert-from-staging`; qdrant, weaviate, lancedb only `truncate-and-insert`.

Snippet run under the docs pytest setup → 1 passed, printing
```
['truncate-and-insert', 'insert-from-staging']
truncate-and-insert
insert-from-staging
None
```
[...]

PR #4338, opened 2026-08-11
fix(agentic-docs): arm the run on the agent label, not the label set

ShreyasGS:
## What changed

The `issues` trigger now arms on the **label that was just added** rather than on the label set the issue happens to carry:

```diff
-      contains(github.event.issue.labels.*.name, 'documentation') &&
-      contains(github.event.issue.labels.*.name, 'agent')
+      github.event.label.name == 'agent' &&
+      contains(github.event.issue.labels.*.name, 'documentation')
```

And the `/position` arm is removed, so a comment no longer starts a run.

## Why

Applying `documentation` and `agent` together sends two `labeled` webhooks. Both are evaluated against an issue that, by the time either is processed, already carries both labels — so both matched. On dlt-hub/dlt#4334 that started two runs in the same second, and they opened two pull requests for one issue (#4335 and #4336; #4335 has been closed as the duplicate).

`github.event.label.name` is the label that actually arrived, so exactly one of the two events matches.

`/position` was doing two jobs: naming where a page goes, and starting a run. The second meant a comment on an unlabelled issue could open a pull request, which made the labels that are supposed to authorise the run redundant. It still sets the location — it is read from the comments when a run starts — it just no longer triggers one.

## Behaviour change for maintainers

- **Apply `documentation` first, then `agent`.** `agent` is what arms it. Adding them in the other order does nothing until `agent` is re-added.
- **Re-adding `agent` is the retry.** Remove it and add it again to re-run, for example after correcting the page location.
- `/position` and `/instructions` comments no longer start anything; write them, then label.

This is the same scope-then-arm split as the UI agent loop in dlt-hub/runtime, where `frontend` alone never triggers and `agent:triage` arms.

## Also in this diff

`ACTOR` is now passed on the review and PR-comment arms too. It was already passed on the issue arm; the two copies of this file had drifted. It is inert for those paths — the job only uses `ACTOR` to decide whose comments are trusted for *placement*, which happens on the generation path — and it is never used to decide whether a run is allowed. Authorisation is settled inside the job by asking GitHub who wrote the comment or review that triggered it.

The rest of the diff is explanatory comments.

## Risk

Trigger-only. No change to what the job does once it starts, and the job re-checks the allowlist itself. Worst case if the expression were wrong is that labelling stops starting runs, which is visible immediately and reversible by revert.

### DIFF
--- a/.github/workflows/agentic-docs.yml
+++ b/.github/workflows/agentic-docs.yml
@@ -88,14 +88,26 @@ jobs:
     # Filtering by author cannot work, because the PAT belongs to a person.
     # `github.event.comment.body` is the raw body here, markers included, unlike
     # what the driver's fetch returns.
+    # `github.event.label.name` is the label that was *just added*, not the set the
+    # issue now carries. Testing the set fires once per label: two labels applied
+    # together produce two `labeled` events, both evaluated against an issue that
+    # already has both, so both match. That happened on issue #4334 and opened two
+    # pull requests. Only `agent` arms a run, and `documentation` scopes it — the
+    # same split as dlt-hub/runtime's UI loop, where `frontend` alone never
+    # triggers.
+    #
+    # The cost is that order matters: add `documentation` first, then `agent`. And
+    # `agent` is the retry — remove it and add it again to re-run. Do not "fix" this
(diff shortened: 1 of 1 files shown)


anuunchin (approved):
LGTM

PR #4343, opened 2026-08-11
fix(sql_database): map Oracle TIMESTAMP WITH LOCAL TIME ZONE to timezone=true

### LINKED ISSUE #4315: sql_database source: Oracle TIMESTAMP WITH LOCAL TIME ZONE reflects as timezone=False even though SQLAlchemy correctly sets local_timezone=True
## Title
sql_database source: Oracle `TIMESTAMP WITH LOCAL TIME ZONE` reflects as `timezone=False` even though SQLAlchemy correctly sets `local_timezone=True`

## Environment
- dlt 1.29.1
- oracledb 4.0.2 (thin mode)
- SQLAlchemy 2.0.51
- Oracle Database 19.28.0.0.0

## Description

`sqla_col_to_column_schema` (in `dlt/sources/sql_database/schema_types.py`) infers a
column's timezone-awareness with:

```python
elif isinstance(sql_t, sqltypes.DateTime):
    col["data_type"] = "timestamp"
    # special handling for MSSQL
    col["timezone"] = sql_t.timezone or sql_t.__visit_name__ in ("DATETIMEOFFSET",)
```

This only reads `sql_t.timezone`. For Oracle's `TIMESTAMP(n) WITH LOCAL TIME ZONE`
columns, SQLAlchemy's Oracle dialect reflects the type with `timezone=False` **and**
`local_timezone=True` -- a second, distinct attribute this code never checks:

```python
>>> from sqlalchemy import inspect, create_engine
>>> insp = inspect(create_engine(ORACLE_URL))
>>> insp.get_columns("some_table")[0]["type"].__dict__
{'timezone': False, 'local_timezone': True}
```

As a result, every Oracle `TIMESTAMP WITH LOCAL TIME ZONE` column is silently treated
as timezone-naive by dlt's schema inference, and loads into a timezone-aware
destination (e.g. Snowflake) as a naive/NTZ column, discarding the column's actual
semantics.

## Minimal repro

```python
from dlt.sources.sql_database import sql_database

# against any Oracle table with a TIMESTAMP(n) WITH LOCAL TIME ZONE column
source = sql_database(
    "oracle+oracledb://user:pass@host:1521/?service_name=XXX",
    schema="SOME_SCHEMA",
    table_names=["some_table"],
)
pipeline = dlt.pipeline(pipeline_name="repro", destination="duckdb", dataset_name="repro")
pipeline.run(source)
print(pipeline.default_schema.get_table("some_table")["columns"])
# -> timezone: False for the LTZ column -- expected True
```

## Expected
`col["timezone"]` should be `True` for a column where SQLAlchemy's Oracle dialect
reports `local_timezone=True`, since that's semantically a timezone-aware column
(Oracle stores it as an absolute instant and re-localizes it to the querying
session's timezone) -- it should not be conflated with a plain naive `TIMESTAMP`.

## Suggested fix
```python
col["timezone"] = (
    sql_t.timezone
    or getattr(sql_t, "local_timezone", False)
    or sql_t.__visit_name__ in ("DATETIMEOFFSET",)
)
```

## Workaround we're using in the meantime
A `type_adapter_callback` that detects `local_timezone=True, timezone=False` and
remaps to `sqlalchemy.dialects.oracle.TIMESTAMP(timezone=True)` before dlt's
inference runs.

Aryan-Pardeshi:
Fixes #4315

## Problem
`sqla_col_to_column_schema` infers timezone-awareness from `sql_t.timezone` only. Oracle's `TIMESTAMP WITH LOCAL TIME ZONE` columns are reflected by SQLAlchemy with `timezone=False` but `local_timezone=True`, so they were mapped to `timezone: false` in the dlt schema — forcing downstream timestamp normalization and breaking round-trips for LTZ data.

## Change
Also honor the `local_timezone` attribute (Oracle-only), via `getattr` so other dialects are unaffected.

## Tests
- `test_oracle_local_time_zone_mapped_to_timezone_true` — `TIMESTAMP(local_timezone=True)` → `timezone: true`
- `test_oracle_timestamp_plain_maps_to_timezone_false` — plain `TIMESTAMP` stays `timezone: false`

Verified: `tests/sources/sql_database/test_schema_types.py` — 19 passed under SQLAlchemy **2.0.51** and **2.1.0b3**.

### DIFF
--- a/dlt/sources/sql_database/schema_types.py
+++ b/dlt/sources/sql_database/schema_types.py
@@ -163,8 +163,13 @@ def sqla_col_to_column_schema(
             col["precision"] = sql_t.length
     elif isinstance(sql_t, sqltypes.DateTime):
         col["data_type"] = "timestamp"
-        # special handling for MSSQL
-        col["timezone"] = sql_t.timezone or sql_t.__visit_name__ in ("DATETIMEOFFSET",)
+        # special handling for MSSQL and Oracle TIMESTAMP WITH LOCAL TIME ZONE
+        # (Oracle reflects the latter with timezone=False but local_timezone=True)
+        col["timezone"] = (
+            sql_t.timezone
+            or getattr(sql_t, "local_timezone", False)
+            or sql_t.__visit_name__ in ("DATETIMEOFFSET",)
+        )
     elif isinstance(sql_t, sqltypes.Date):
         col["data_type"] = "date"
     elif isinstance(sql_t, sqltypes.Time):
--- a/tests/sources/sql_database/test_schema_types.py
+++ b/tests/sources/sql_database/test_schema_types.py
@@ -218,3 +218,26 @@ def test_get_table_references() -> None:
 
     refs = get_table_references(child)
     assert refs == []
+
+
+def test_oracle_local_time_zone_mapped_to_timezone_true() -> None:
+    """Oracle TIMESTAMP WITH LOCAL TIME ZONE (local_timezone=True) maps to timezone=True (issue #4315)."""
+    from sqlalchemy.dialects.oracle import TIMESTAMP as OracleTIMESTAMP
+
+    metadata = sa.MetaData()
+    table = sa.Table("t", metadata, sa.Column("col", OracleTIMESTAMP(local_timezone=True)))
+    col_schema = sqla_col_to_column_schema(table.c.col, "full")
+    assert col_schema is not None
+    assert col_schema["data_type"] == "timestamp"
+    assert col_schema["timezone"] is True
+
+
+def test_oracle_timestamp_plain_maps_to_timezone_false() -> None:
+    """Plain Oracle TIMESTAMP stays timezone=False; local_timezone defaults to False."""
+    from sqlalchemy.dialects.oracle import TIMESTAMP as OracleTIMESTAMP
+
+    metadata = sa.MetaData()
+    table = sa.Table("t", metadata, sa.Column("col", OracleTIMESTAMP()))
+    col_schema = sqla_col_to_column_schema(table.c.col, "full")
+    assert col_schema is not None
+    assert col_schema["timezone"] is False


rudolfix:
@Aryan-Pardeshi @geforce123 @ignatNakhai-rc
 fix here will generate a correct hint in `dlt` schema but oracle driver will still emit naive datetime for both TZ and LTZ and this naive datetime depends on database timezone / session timezone. `dlt` assumes all naive datetimes when converted to tz-aware (the hint is there so they will be converted) to happen in UTC.

my take: on top of this fix I'll add:
- TZ/LTZ timestamps will be fetched via `SYS_EXTRACT_UTC(column)` (via sqlalchemy `TypeDecorator`) - so all timestamps are UTC
- session will be always set to UTC

in that case all datetimes will be sent as UTC naive datetimes and correctly converted by `dlt`. does it sound good?

geforce123:
@rudolfix Sounds good, and matches something we hit independently in production, so glad to confirm it from the field rather than just in theory.

We ran into exactly this gap building an Oracle → Snowflake ingestion framework on dlt 1.29.1. Our workaround today is a `type_adapter_callback` + `query_adapter_callback` pair that does roughly the same thing by hand:

- Pin the session to UTC on every connection (`ALTER SESSION SET TIME_ZONE = 'UTC'` via a SQLAlchemy `connect` event) — same as your point 2.
- For native `WITH LOCAL TIME ZONE` columns: flip the schema hint (this PR).
- For `WITH TIME ZONE` columns carrying a *named region* (not a fixed offset): we hit `DPY-3022: named time zones are not supported in thin mode` at fetch time — this is a separate, driver-level decode failure, not a reflection-time issue, and it's what pushed us toward manually `CAST`-ing those columns to `TIMESTAMP(9) WITH LOCAL TIME ZONE` server-side before fetch, so the driver never has to decode a named region on the wire.

Two questions on `SYS_EXTRACT_UTC`, since the answer determines how much of our own workaround we could retire in favor of the upstream fix:

1. **Does it sidestep `DPY-3022` for named-region `WITH TIME ZONE` columns too?** If `SYS_EXTRACT_UTC` returns a plain `TIMESTAMP` with no zone/region encoded in the wire format (which is how we read the Oracle docs), the thin driver should never see a named-region value to choke on in the first place — a cleaner fix than our CAST-to-LTZ workaround. Worth confirming explicitly, since it'd let us drop that path entirely.
2. **Is `SYS_EXTRACT_UTC` actually session-independent for `WITH LOCAL TIME ZONE` columns specifically?** Our understanding is LTZ resolves relative to the *session's* timezone at read time (that's the whole point of "local"), so `SYS_EXTRACT_UTC` on an LTZ column would still depend on session state even though it doesn't for `WITH TIME ZONE` columns carrying an explicit offset/region. If that's right, the session-UTC pin isn't just defense-in-depth for LTZ — it's load-bearing, same as it is in our own implementation. Worth being explicit about which of the two mechanisms (session pin vs. `SYS_EXTRACT_UTC`) is actually doing the correctness work for each column type, so it's clear neither can be dropped independently.

One more thing worth flagging for sequencing: shipping the schema-hint fix (this PR) *without* the follow-up changes it a "loudly naive, obviously incomplete" bug into a "silently wrong whenever the session isn't UTC" one — dlt will now confidently attach UTC tzinfo to a naive value that may not actually be UTC. Might be worth landing both together, or calling out clearly in the changelog that a UTC session is now a hard prerequisite for correctness on Oracle TZ/LTZ columns.

Happy to test against a real Oracle instance with both native LTZ and named-region TZ columns if that's useful — we have both in our current LAB environment.

PR #4344, opened 2026-08-11
fix: annotate PipelineTrace last step accessors as Optional

### LINKED ISSUE #4321: Bad type hinting for PipelineTrace causes integration failures
### dlt version

1.17.0

### Describe the problem

The following property accessors have bad return type signatures
```python
    @property
    def last_extract_info(self) -> ExtractInfo:
        step_trace = self.last_pipeline_step_trace("extract")
        if step_trace and isinstance(step_trace.step_info, ExtractInfo):
            return step_trace.step_info
        return None

    @property
    def last_normalize_info(self) -> NormalizeInfo:
        step_trace = self.last_pipeline_step_trace("normalize")
        if step_trace and isinstance(step_trace.step_info, NormalizeInfo):
            return step_trace.step_info
        return None

    @property
    def last_load_info(self) -> LoadInfo:
        step_trace = self.last_pipeline_step_trace("load")
        if step_trace and isinstance(step_trace.step_info, LoadInfo):
            return step_trace.step_info
        return None
```
As a result, some integration modules like [dagster-dlt](https://pypi.org/project/dagster-dlt/) fail when loading failed load packages, resulting in 0 row counts, giving this error for ex.
```
The above exception was caused by the following exception:
AttributeError: 'NoneType' object has no attribute 'row_counts'

  File "/opt/dagster/app/.venv/lib/python3.12/site-packages/dagster/_core/execution/plan/utils.py", line 57, in op_execution_error_boundary
    yield
  File "/opt/dagster/app/.venv/lib/python3.12/site-packages/dagster/_utils/__init__.py", line 394, in iterate_with_context
    next_output = next(iterator)
                  ^^^^^^^^^^^^^^
  File "/opt/dagster/app/nxt/dlt/assets/posthog.py", line 45, in raw_posthog
    yield from dlt.run(context, dlt_source=posthog_src(start=start, end=end))
  File "/opt/dagster/app/.venv/lib/python3.12/site-packages/dagster_dlt/dlt_event_iterator.py", line 77, in __next__
    return next(self._inner_iterator)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/dagster/app/.venv/lib/python3.12/site-packages/dagster_dlt/resource.py", line 294, in _run
    metadata = self.extract_resource_metadata(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/dagster/app/.venv/lib/python3.12/site-packages/dagster_dlt/resource.py", line 134, in extract_resource_metadata
    rows_loaded = dlt_pipeline.last_trace.last_normalize_info.row_counts.get(
```
The bug might be on the dagster side but there is no way for the maintainers to detect such bugs if the type hinting is false.

### Expected behavior

Before submitting a PR I would like to know if this is an accepted coding convention in the dlt team. I've seen these missing Optional (x | None) errors pop up here and there especially in function signatures like
```python
@decorators.source
def sql_database(
    credentials: Union[ConnectionStringCredentials, Engine, str] = dlt.secrets.value,
    schema: Optional[str] = dlt.config.value,
    metadata: Optional[MetaData] = None,
    table_names: Optional[List[str]] = dlt.config.value,
    chunk_size: int = 50000,
    backend: TableBackend = "sqlalchemy",
    detect_precision_hints: Optional[bool] = False,
    reflection_level: Optional[ReflectionLevel] = "full",
    defer_table_reflect: Optional[bool] = None,
    table_adapter_callback: Optional[TTableAdapter] = None,
    backend_kwargs: Dict[str, Any] = None,
...
```
`backend_kwargs` should be type hinted as `Dict[str, Any] | None`.

### Steps to reproduce

Not the easiest thing to do, you would need to have a failing pipeline within dagster, save the load package by having persistent storage for the DLT_DATA_DIR, and letting dlt attempt to load the package again which is successful **however** prints out the error and refuses to return any metadata

### Operating system

Linux

### Runtime environment

Kubernetes

### Python version

3.12

### dlt data source

sql_database

### dlt destination

Google BigQuery

### Other deployment details

I run my dlt pipelines using a durable storage for my load packages in case they might fail since I work with sources where the incremental cursor state cannot be read from the destination storage, or applied to the source again.

### Additional information

The fix is simple, the type fixing in QA might not be so trival. Here is one example fix
```python
@property
def last_normalize_info(self) -> NormalizeInfo | None:
    step_trace = self.last_pipeline_step_trace("normalize")
    if step_trace and isinstance(step_trace.step_info, NormalizeInfo):
        return step_trace.step_info
    return None
```

bunnysayzz:
### Description

`PipelineTrace`'s last-step accessors were typed as returning their info type directly, but they return `None` whenever the step never ran (or its step info is missing). The false non-optional hints leak into downstream integrations: dagster-dlt dereferences `last_normalize_info.row_counts` and crashes with `AttributeError: 'NoneType' object has no attribute 'row_counts'` instead of being able to do an optional check.

This changes the four accessors to `Optional[...]`, matching their actual contract:

- `last_pipeline_step_trace` -> `Optional[PipelineStepTrace]`
- `last_extract_info` -> `Optional[ExtractInfo]`
- `last_normalize_info` -> `Optional[NormalizeInfo]`
- `last_load_info` -> `Optional[LoadInfo]`

Added a regression test asserting all accessors return `None` (and do not raise) on an empty trace.

### Related Issues

- Closes #4321

### Additional Context

Tested locally: the new regression test passes, and `tests/pipeline/test_pipeline_trace.py` shows the same pass/fail set as on `main` (the only delta is the new passing test). The remaining failures in that file are pre-existing `ModuleNotFoundError` for optional deps (`duckdb`, `pandas`) not installed in the local test environment.

Per the discussion with @zilto on #4321, only the type annotations in `dlt` are addressed here; the deeper dagster-side handling is tracked separately by the dagster team.

### DIFF
--- a/dlt/pipeline/trace.py
+++ b/dlt/pipeline/trace.py
@@ -170,7 +170,7 @@ class PipelineTrace(SupportsHumanize, _PipelineTrace):
             msg += "\n" + "\n\n".join([s.asstr(verbosity) for s in self.steps])
         return msg
 
-    def last_pipeline_step_trace(self, step_name: TPipelineStep) -> PipelineStepTrace:
+    def last_pipeline_step_trace(self, step_name: TPipelineStep) -> Optional[PipelineStepTrace]:
         matching_steps = [step for step in self.steps if step.step == step_name]
         if not matching_steps:
             return None
@@ -188,21 +188,21 @@ class PipelineTrace(SupportsHumanize, _PipelineTrace):
         return d
 
     @property
-    def last_extract_info(self) -> ExtractInfo:
+    def last_extract_info(self) -> Optional[ExtractInfo]:
         step_trace = self.last_pipeline_step_trace("extract")
         if step_trace and isinstance(step_trace.step_info, ExtractInfo):
             return step_trace.step_info
         return None
 
     @property
-    def last_normalize_info(self) -> NormalizeInfo:
+    def last_normalize_info(self) -> Optional[NormalizeInfo]:
         step_trace = self.last_pipeline_step_trace("normalize")
         if step_trace and isinstance(step_trace.step_info, NormalizeInfo):
             return step_trace.step_info
         return None
 
     @property
-    def last_load_info(self) -> LoadInfo:
+    def last_load_info(self) -> Optional[LoadInfo]:
         step_trace = self.last_pipeline_step_trace("load")
         if step_trace and isinstance(step_trace.step_info, LoadInfo):
             return step_trace.step_info
--- a/tests/pipeline/test_pipeline_trace.py
+++ b/tests/pipeline/test_pipeline_trace.py
@@ -18,6 +18,7 @@ from dlt.common.configuration.utils import get_resolved_traces
 from dlt.common.metrics import TDataLocation, TDatasetDataLocation
 from dlt.common.pipeline import ExtractInfo, NormalizeInfo, LoadInfo
 from dlt.common.schema import Schema
+from dlt.common.runtime.exec_info import get_execution_context
 from dlt.common.runtime.telemetry import stop_telemetry
 from dlt.common.typing import DictStrAny, DictStrStr, TRefreshMode, TSecretValue
 from dlt.common.utils import digest128, uniq_id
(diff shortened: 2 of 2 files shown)


zilto (approved):
LGTM

Issue #4345, opened 2026-08-11
feat: allow `config.yaml` and `secrets.yaml`


zilto:
## Summary
`dlt` has an elaborate configuration system based on `.toml` files and environment variables. It could easily support YAML too.

## Problem
When you have large configurations, `toml` and environment variables appear "flat". They make it hard to see the hierarchical structure. Editing them can also create silent failures via name collision

For example, in TOML: writing to `sources` would overwrite `sources.foo` and `sources.bar`, `destination.warehouse` and `destination.warehouse.credentials` are not co-located.
```toml
[sources.foo]
api_key = "abc"
url = "https://..."

[destination.warehouse]
destination_type = "snowflake"

[sources.bar]
api_key = "xyz"
start_date = "..."

[destination.warehouse.credentials]
password = "..."
```

In YAML, you would get: clear hierarchy, easier to avoid collisions, forces co-location of related config.

```yaml
sources:
  foo:
    api_key: "abc"
    url: "https://..."
  bar:
    api_key: "xyz"
    start_date: "..."

destination:
  warehouse:
    destination_type: "snowflake"
    credentials:
      password: "..." 
```

YAML is also probably easier to parse for LLMs when the config is large.

## Potential solution
- Implement a `YamlConfigProvider` and `YamlSecretProvider` similar to the TOML provider: https://github.com/dlt-hub/dlt/blob/af62b98e0ed892b1a01e9b78417170619093f327/dlt/common/configuration/providers/toml.py
- Ensure that all tests for `toml` provider succeed for YAML provider
- Make it an optional dependency group `yaml-config` (or other name). I strongly recommend using the Python package `py-yaml12` for the reasones explained here: https://opensource.posit.co/blog/2026-05-21_in-defense-of-yaml/#introducing-py-yaml12
- Set a deterministic resolution order for multiple config sources. For example `env var > toml > yaml`
- Log a clear warning message when both a `config.toml` and a `config.yaml` are found on disk (this should reuse existing logic when env var collide with toml)

zilto:
After more research, I found #1642 adding custom providers for YAML support requested in #1596. There is also this documentation page: https://dlthub.com/docs/examples/custom_config_provider

YAML support is not prominent in the current documentation and requires boilerplate code in all Python files relying on it. I think it's still valuable to have native `yaml` support considering `yaml` is already a Python dependency of the project

PR #4346, opened 2026-08-11
feat: `yaml` config provider

zilto:
Solves #4345  (and #1596). It expands on previous work that allowed custom config provider.

## Changes
There are a lot of files and lines of code changes, but it's inflated by `.yaml` test cases.

The main changes:
- move logic of the `toml` config provider to a base class `SettingsDocProvider`
- keep read/write logic in `toml` and `yaml` specific classes

## Future work
- update the docs to make more mentions of `config.yaml` and `secrets.yaml`
- ensure this is supported by `dlthub`
- ensure the dltHub AI Harness handles `.yaml` files properly (e.g., MCP tool to read values, mentions in skills)

### DIFF
--- a/dlt/_workspace/_workspace_context.py
+++ b/dlt/_workspace/_workspace_context.py
@@ -25,7 +25,13 @@ from dlt.common.typing import copy_sig_ret
 from dlt._workspace.configuration import WorkspaceConfiguration, WorkspaceRuntimeConfiguration
 from dlt._workspace.exceptions import WorkspaceRunContextNotAvailable
 from dlt._workspace.profile import BUILT_IN_PROFILES, DEFAULT_PROFILE, read_profile_pin
-from dlt._workspace.providers import ProfileConfigTomlProvider, ProfileSecretsTomlProvider
+from dlt._workspace.providers import (
+    ProfileConfigTomlProvider,
+    ProfileSecretsTomlProvider,
+    ProfileConfigYamlProvider,
+    ProfileSecretsYamlProvider,
+)
+from dlt.common.configuration.providers.utils import warn_on_toml_yaml_collision
 from dlt._workspace.run_context import (
     DEFAULT_LOCAL_FOLDER,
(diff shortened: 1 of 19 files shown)


dlt/_workspace/_workspace_context.py
```diff
@@ -225,10 +231,18 @@ def configured_profiles(self) -> List[str]:
         return list(configured)
 
     def _initial_providers(self, profile_name: str) -> List[ConfigProvider]:
+        secrets_toml = ProfileSecretsTomlProvider(self.settings_dir, profile_name, self.global_dir)
+        config_toml = ProfileConfigTomlProvider(self.settings_dir, profile_name, self.global_dir)
+        secrets_yaml = ProfileSecretsYamlProvider(self.settings_dir, profile_name, self.global_dir)
+        config_yaml = ProfileConfigYamlProvider(self.settings_dir, profile_name, self.global_dir)
+        warn_on_toml_yaml_collision(secrets_toml, secrets_yaml)
```
rudolfix, line 238: here I'd remove yaml providers that do not have any locations present from the list of providers below. otherwise they will be reported in configuration errors as empty providers without files. we can be smarter and also prune tomls if there are no locations defined (for both config and secrets)

rudolfix (commented):
A few things before full review:
- I'd prune providers that get into initial provider list to declutter exception messages (see comments)
- I'd whitelist `yaml` extensions in `ConfigurationFileSelector` to start sending this to runtime. enabling runtime to pass through yaml profiles is separate (right now yaml files will be pruned), but not hard
- tools to redact and update config/secret files that agents are using will need to be rewritten. right now they handle toml only. also rewriting a yaml path will need some kind of convention. we'll need to maybe switch to yaml parser that preserves comments and whitespaces (we were using it in dlthub)
- there are very long module and function docstrings. those should be split and converted into comments. right now they are too sloppy. you can ask claude to apply rules from `@docstrings.md` or or try `/review-vocabulary` (which is based on Boeing simple simple English) and see what happens

PR #4348, opened 2026-08-12
fix: raise ValueError coercing a non-finite decimal to bigint

eeshsaxena:
`coerce_value("bigint", "decimal", ...)` leaks `decimal.InvalidOperation` on a non-finite value instead of the `ValueError` the coercion layer expects:

```python
from decimal import Decimal
from dlt.common.data_types import coerce_value

coerce_value("bigint", "decimal", Decimal("Infinity"))
# decimal.InvalidOperation: [<class 'decimal.InvalidOperation'>]
```

`_numeric_to_bigint` checks for a fractional part with `value % 1`, but for `Decimal("Infinity")` / `Decimal("-Infinity")` (and `Wei`) that modulo raises `InvalidOperation` rather than returning a remainder. `coerce_value` only catches `OverflowError`, so it propagates uncaught and takes down normalization instead of being handled as a failed coercion (which would fall back to a variant column).

The text coercers already convert `InvalidOperation` into `ValueError` (`_text_to_decimal`, `_text_to_wei`); this makes `_numeric_to_bigint` consistent. `Decimal("NaN")` already raised `ValueError` and still does. Added regression cases to `test_coerce_type_to_bigint`.

### DIFF
--- a/dlt/common/data_types/type_helpers.py
+++ b/dlt/common/data_types/type_helpers.py
@@ -166,7 +166,14 @@ def _float_to_wei(value: float) -> Wei:
 
 
 def _numeric_to_bigint(value: Any) -> int:
-    if value % 1 != 0:
+    try:
+        if value % 1 != 0:
+            raise ValueError(value)
+    except InvalidOperation:
+        # A non-finite Decimal/Wei (Infinity, -Infinity) has no integer form,
+        # and `value % 1` raises InvalidOperation rather than returning a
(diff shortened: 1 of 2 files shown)


zilto (approved):
The change makes sense. Though, I'm curious where you encounter these infinity values.

zilto:
These changes could be expanded to include `Decimal("NaN")`

PR #4362, opened 2026-08-16
Feat/3269 snowflake workload identity provider

### LINKED ISSUE #3269: Add workload identity parameters to Snowflake destination
### Feature description

Now that Snowflake supports Workload Identity Federation, it would be great if the `workload_identity_provider` parameter is supported in `SnowflakeCredentials` since now it will be ignored and Snowflake complains about it missing.

The workaround for now is using `query={"workload_identity_provider": "<provider>"}`.

### Are you a dlt user?

None

### Use case

We want to not rely on secrets when authenticating to Snowflake.

### Proposed solution

Add `workload_identity_provider` as a supported parameter in `SnowflakeCredentials`, just like `private_key`.  It looks like it should also be added to the `SnowflakeCredentials.__query_params__` list. I'd be happy to do it if this is the right approach.

### Related issues

_No response_

richacode007-byte:
Description
Adds support for Snowflake Workload Identity Federation by accepting workload_identity_provider as a first-class credentials field (and connection-string query param). It is forwarded to the Snowflake connector via to_connector_params(), so users can set authenticator=WORKLOAD_IDENTITY with a provider such as AWS, AZURE, GCP, or OIDC.

Includes a unit test covering URL resolve → connector params, and a short WIF note in the Snowflake destination docs.

Related Issues
Fixes #3269
Additional Context
Requires snowflake-connector-python>=3.17.0 for WIF (documented; dependency not bumped in this PR).
Local check: tests/load/snowflake/test_snowflake_configuration.py — 28 passed.

### DIFF
--- a/dlt/destinations/impl/snowflake/configuration.py
+++ b/dlt/destinations/impl/snowflake/configuration.py
@@ -31,6 +31,7 @@ class SnowflakeCredentialsWithoutDefaults(ConnectionStringCredentials):
     role: Optional[str] = None
     authenticator: Optional[str] = None
     token: Optional[str] = None
+    workload_identity_provider: Optional[str] = None
     private_key: Optional[TSecretStrValue] = None
     private_key_path: Optional[str] = None
     private_key_passphrase: Optional[TSecretStrValue] = None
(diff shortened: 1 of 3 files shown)


rudolfix (approved):
LGTM!

PR #4367, opened 2026-08-17
fix(incremental): keep last_value forward-only when it is exactly 0

### LINKED ISSUE #4363: incremental `lag` rewinds the cursor by one window when `last_value` is exactly 0
### dlt version

1.30.0 (and devel as well, a4243c994b9fade227d23e4bb18dab70037507cf)

### Describe the problem

The forward-only guard for `lag` is a **truthiness test**, so it is skipped when the persisted
`last_value` is a falsy value, for example, exactly `0`:

https://github.com/dlt-hub/dlt/blob/a4243c994b9fade227d23e4bb18dab70037507cf/dlt/extract/incremental/__init__.py#L688-L689

| `last_value_func` | after run 1 | after run 2 |
|---|---|---|
| `max` | `0` | `-5` |
| `min` | `0` | `5` |
| `max` | `10` | `10` |
| `min` | `10` | `10` |

Note: Numeric cursors only. Epoch-0 datetimes are truthy, and are unaffected (arguably the issue)

### Expected behavior

`0` ought to

- behave like `10`, with the cursor stays put.
- match behaviour with epoch-0 datetimes.

### Steps to reproduce

Run this:

```python
import dlt

def run_case(first_cursor: int) -> int:
    """run 1 leaves last_value at `first_cursor`; run 2 yields only a row below the lagged bound"""
    is_second_run = False

    @dlt.resource(name="events", primary_key="id", write_disposition="append")
    def events(_=dlt.sources.incremental("counter", lag=5, last_value_func=max)):
        yield {"id": 2, "counter": first_cursor - 10} if is_second_run else {"id": 1, "counter": first_cursor}

    p = dlt.pipeline(pipeline_name=f"lag_zero_{first_cursor}", destination="duckdb", dev_mode=True)
    p.run(events())
    is_second_run = True
    p.run(events())
    return p.state["sources"][p.default_schema_name]["resources"]["events"]["incremental"]["counter"]["last_value"]

control, subject = run_case(10), run_case(0)
print(f"cursor 10 -> {control!r}   (control)")
print(f"cursor  0 -> {subject!r}")

assert control == 10, f"control regressed to {control}"
assert subject == 0, f"cursor regressed to {subject} instead of staying at 0"
```

And you get:

```
cursor 10 -> 10   (control)
cursor  0 -> -5
AssertionError: cursor regressed to -5 instead of staying at 0
```

### Operating system

macOS

### Runtime environment

Local

### Python version

3.14

### dlt data source

N/A

### dlt destination

DuckDB

### Other deployment details

N/A. It local. 

### Additional information

`apply_lag_with_suppression` tests the same value with `is None`:

https://github.com/dlt-hub/dlt/blob/a4243c994b9fade227d23e4bb18dab70037507cf/dlt/extract/incremental/lag.py#L110

so it adds to the "codebase internal self-inconsistency" argument.

chjnett:
Fixes dlt-hub/dlt#4363

## Summary
The forward-only guard for `lag` used a truthiness test on the persisted `last_value`:

```python
if self.lag and (cached_last_value := cached_state.get("last_value")):
```

When the cursor is exactly `0` (falsy), the guard was skipped, so `lag` rewound the cursor by one window on the next run (`0` → `-5` for `max` with `lag=5`). Numeric cursors at epoch-0 were the affected case; the issue reporter noted epoch-0 datetimes are truthy and unaffected.

Changed the guard to an identity check (`is not None`), so `0` behaves like any other value: the cursor stays put.

## Test
Added `test_incremental_lag_forward_only_with_zero_last_value`, parametrized over first cursors `10` (control) and `0` (subject) — both now keep `last_value == first_cursor` across two runs. Verified locally against the reporter's reproducer: before the fix `0` regressed to `-5`, after it stays `0`.

### DIFF
--- a/dlt/extract/incremental/__init__.py
+++ b/dlt/extract/incremental/__init__.py
@@ -686,7 +686,7 @@ class Incremental(
 
         cached_state = self._cached_state
         # ensure last_value maintains forward-only progression when lag is applied
-        if self.lag and (cached_last_value := cached_state.get("last_value")):
+        if self.lag and (cached_last_value := cached_state.get("last_value")) is not None:
             transformer.last_value = self.last_value_func(
                 (transformer.last_value, cached_last_value)
             )
--- a/tests/extract/test_incremental.py
+++ b/tests/extract/test_incremental.py
@@ -3705,6 +3705,39 @@ def test_incremental_lag_int_with_initial_values(lag: float, last_value_func) ->
         assert result == expected_results[int(lag)]
 
 
+@pytest.mark.parametrize("first_cursor", [10, 0])
+def test_incremental_lag_forward_only_with_zero_last_value(first_cursor: int) -> None:
+    """
+    The forward-only guard for `lag` must not use a truthiness test on the
+    persisted `last_value`: when the cursor is exactly `0` (falsy), the guard
+    was skipped and `lag` rewound the cursor by one window on the next run.
+
+    Regression for #4363.
+    """
+
+    pipeline = dlt.pipeline(
+        pipeline_name="p" + uniq_id(),
+        destination=dlt.destinations.duckdb(credentials=duckdb.connect(":memory:")),
+    )
+    is_second_run = False
+
+    @dlt.resource(name="events", primary_key="id", write_disposition="append")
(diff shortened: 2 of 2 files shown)


zilto (approved):
The fix is sound. Thank you.

PR #4375, opened 2026-08-19
fix(config): resolve GCP credentials from a file path, aggregate union parse errors

### LINKED ISSUE #3328: dlt tries OAuth in GCP Instead of Service Account Credentials when given Service Account Credentials via credential file

nanjeshramesh:
### Description

`GcpServiceAccountCredentials`/`GcpOAuthCredentials.parse_native_representation()` called
`json.loads()` directly on the given string, so pointing `credentials` at a file path (the
common `GOOGLE_APPLICATION_CREDENTIALS` convention) failed to parse. Since GCS/GDrive type
credentials as `Union[GcpServiceAccountCredentials, GcpOAuthCredentials]`, the resolver
silently fell through to try OAuth next, which also failed - and only OAuth's confusing
error surfaced to the user, even though the value was a valid service account file all along.

Two changes:
- `gcp_credentials.py`: read the file's contents first when the given string is an existing
  file path, for both ServiceAccount and OAuth credentials.
- `resolve.py`: when every spec in a `Union` fails to parse, report why each one failed
  instead of only the last one tried. This is generic to all Union-typed credentials in dlt,
  not just GCP, so it improves error messages beyond this one bug.

### Related Issues

- Fixes #3328

### Additional Context

Added tests for both changes (file-path resolution for both credential types, and the
aggregated union error message). Full `tests/common/configuration/` suite passes locally
(249 tests), `ruff check`/`ruff format --check` clean.

### DIFF
--- a/dlt/common/configuration/resolve.py
+++ b/dlt/common/configuration/resolve.py
@@ -308,6 +308,7 @@ def _resolve_config_fields(
             if not current_value:
                 if len(specs_in_union) > 1:
                     is_optional = is_optional_type(hint)
+                    invalid_native_value_exceptions: List[Tuple[Type[Any], InvalidNativeValue]] = []
                     for idx, alt_spec in enumerate(specs_in_union):
                         # return first resolved config from an union
                         try:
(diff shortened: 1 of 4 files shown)


nanjeshramesh:
Friendly bump on this one. It lets GCP credentials be given as a file path and makes the union parse errors report every attempted type instead of only the last one. Happy to adjust anything or split it up if that makes review easier.

nanjeshramesh:
@zilto following up on this one when you have a moment. It lets GCP credentials be given as a file path (the usual `GOOGLE_APPLICATION_CREDENTIALS` convention), and makes the union parse errors report every attempted type instead of only the last one. CI has not run on it, so it may need a maintainer to approve the workflow. Happy to rebase or split it up if that makes review easier.

PR #4385, opened 2026-08-21
fix(filesystem): drop the hard orjson import in the filesystem destination

tushardev-365:
### Description

`filesystem.py` imports `orjson` directly, only to catch `orjson.JSONDecodeError` in `get_storage_versions`. That one import causes two distinct problems.

**It breaks the filesystem destination entirely on PyPy and emscripten.** The dependency markers in `pyproject.toml` deliberately exclude orjson there:

```
"orjson>=3.6.7,<4,... ; platform_python_implementation != 'PyPy' and sys_platform != 'emscripten'",
```

so on those platforms `import orjson` at the top of the module raises `ModuleNotFoundError` the moment the filesystem (and with it delta/iceberg) destination is imported. The pluggable backend in `dlt/common/json` exists precisely to avoid this, and this file was the only direct orjson import in the package outside that backend and one test file.

**It catches the wrong error with the simplejson backend.** `json` in this module is dlt's pluggable json, and `DLT_USE_JSON=simplejson` is a documented switch. simplejson's `JSONDecodeError` is not a subclass of orjson's:

```python
>>> issubclass(simplejson.JSONDecodeError, orjson.JSONDecodeError)
False
>>> issubclass(simplejson.JSONDecodeError, ValueError), issubclass(orjson.JSONDecodeError, ValueError)
(True, True)
```

so with that backend a corrupt `init` file bypassed the curated `ValueError` naming the corrupt file path and leaked a raw decode error with no context instead.

Catching `(KeyError, ValueError)` covers both backends, since both decode errors subclass `ValueError`. The `import orjson` line is removed.

The three invalid-content cases in `test_get_storage_version_invalid` now pass `match="Invalid content"`, so a leaked backend error (which would not carry that message) fails the test rather than satisfying a bare `pytest.raises(ValueError)`.

### Related Issues

None filed; happy to open one if you prefer tracking it that way.

### Additional Context

I ran the exception-class checks above against installed simplejson and orjson; I did not run the filesystem test suite locally since it needs the load-test harness. The changed module compiles and the test edit is a `match=` tightening of three existing assertions.


rudolfix:
@tushardev-365 this is definitely a bug. I pushed additional test to prevent that in the futyre

rudolfix (approved):
LGTM

tushardev-365:
Thank you!

.github/workflows/lint.yml
```diff
@@ -40,6 +40,16 @@ jobs:
...
 
+      # emscripten drops the marker-excluded deps (orjson), so this catches a hard
+      # import of one of them that plain CPython lint cannot see
+      - name: Setup Node
+        uses: actions/setup-node@v5
+        with:
+          node-version: '22'
+
+      - name: Check dlt imports under emscripten
+        run: make lint-emscripten
```
zilto, line 51: @rudolfix was it intended to include those checks? the `lint.yml` is a bottleneck in our GHA graph. Now having `node` there is a somewhat significant slow down. The make command `lint-emscripten` is very much a test and not a linting operation.

I recommend moving it to a dedicate branch of the graph and call it a test

PR #4386, opened 2026-08-23
chore: move flake8 config from tox.ini to pyproject.toml

### LINKED ISSUE #3346: chore: move all tooling to `pyproject.toml`

nanjeshramesh:
### Description

mypy and pytest settings were already moved into pyproject.toml at some point. This finishes
that up by moving the last piece, the flake8 config in tox.ini.

Added Flake8-pyproject as a dev dependency so flake8 reads its config from pyproject.toml,
moved the settings from tox.ini into a new [tool.flake8] section, and deleted tox.ini.

One catch: flake8-tidy-imports parses banned-modules with a plain string.split("\n") call
internally, so that option has to stay a TOML string instead of a list like everything else.
Found this by testing before deleting tox.ini, not by assuming the conversion would just work.

I checked this doesn't change any actual linting behavior by running flake8 on the full dlt/,
tools/, and tests/ trees with the old config and the new one and diffing the output (identical
both times), and by running it against a small file with a banned import, a banned typing
symbol, and a print statement, both inside and outside tests/, to make sure per-file-ignores
still exempts the right things.

### Related Issues

- Closes #3346

### Additional Context

Ran the full lint-core suite locally, no changes in output.


nanjeshramesh:
@zilto could you take a look, and approve the workflow run if needed? The three CI jobs on this PR were skipped, so nothing has actually validated it yet. It moves the flake8 config from `tox.ini` into `pyproject.toml` (mypy and pytest already live there) and adds flake8-pyproject so flake8 still reads it. Happy to adjust anything.

PR #4407, opened 2026-08-27
docs: unify readme examples around a single spotify dataset

elviskahoro:
## Why
The README's examples each used a different toy dataset (Pokémon, chess players, event guests, Luma), so a reader following the doc top-to-bottom had to re-orient at every section instead of building on one mental model. Unifying every snippet around a single Spotify tracks/playlists dataset lets the examples compose — the same `tracks`/`playlist_tracks` resource reappears in the schema, transform, and Ibis sections — so the README reads as one continuous example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,24 +59,27 @@ from dlt.sources.rest_api import rest_api_source
 
 # 1. Describe the API declaratively
 source = rest_api_source({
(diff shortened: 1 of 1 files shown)


zilto (approved):
LGTM

PR #4417, opened 2026-08-30
(feat) background agent launcher

rudolfix:
Adds a job kind whose body is an agent loop. Three terms carry it:

- an **agent definition**, an `AGENT.md` or a decorated function: the system prompt, the inputs and output, the tools, skills and rules the agent uses, and the access it needs (pydantic-ai calls this an `AgentSpec`);
- an **agent job**, `run.agent(...)`: the definition plus the runtime settings that say how it operates: model, token limit, trigger, instructions, loop, identity (what pydantic-ai and claude-agent-sdk call an Agent, and what the web UI lists as one);
- an **agent run**, one execution of the job: inputs in (a job run id, say), output and trace out, with settings a caller may override for that run alone.

Agent jobs run on the same launcher machinery as every other dlt job: locally with `dlthub local run`, remotely by the runner, on a pluggable loop (pydantic-ai or claude-agent-sdk). Part 1 is the user guide, on the job-inspector example. Part 2 is how it works.

# Part 1. User guide

## 1. Write an agent definition

An agent definition is a system prompt plus a declaration of what the agent takes, returns and may touch. It is an `AGENT.md` or a decorated Python function; both make the same agent job.

### 1a. As an `AGENT.md`

Installed with a toolkit (`dlthub ai toolkit install dlthub-platform`) it lives in `.claude/dlthub/agents/<name>/AGENT.md` (`.cursor/...`, `.agents/...` for the other hosts), a folder Claude Code does not scan for subagents, so dlt agents and native subagents never mix. Or keep it anywhere in the workspace as a folder holding an `AGENT.md` and refer to it by path. The frontmatter declares, the markdown body is the system prompt, and only the body is required: a file with no frontmatter is a working agent definition named after its folder.

```yaml
---
name: job-inspector
description: Inspects a failed job run and reports a diagnosis. Read-only.
tools: [telemetry]                          # dlthub MCP feature groups, see §8
skills: [dlthub-platform:debug-deployment]  # loaded natively or inlined, see §9
rules:  [dlthub-platform:job-resources]     # always inlined into the prompt

access:                                     # what the job may touch, see §7
  local:   [read]                           # read | write | execute | network
  data:    [read]                           # read | write, governed by the profile
  context: [read]                           # telemetry, runs, job definitions

inputs:
  type: object
  properties:
    failed_run_id:
      type: string
      description: run id of the job that failed
      entity_type: job-run                  # this input names a workspace entity
    failed_job_ref:
      type: string
      description: job ref of the failed job
      entity_type: job
  required: {}

output:
  type: object
  properties:
    classification: { enum: [config, credentials, code, upstream_data, unknown] }
    confidence:     { enum: [high, medium, low] }
    evidence:       { type: array, items: { type: string } }
  required: [classification, confidence]

defaults:                                   # the runtime may override all of these
  trigger: [job.fail:*]
  model: sonnet
  limits: { max_turns: 30, max_tokens: 1000000 }
  loop_run_args: { retries: 1 }
---
You are a job inspector for a dltHub Platform workspace. You run unattended, seconds
after a job failed. Explain the failure; do not repair it.

Investigate job_ref '{{ failed_job_ref }}' from trigger `{{ run_context.trigger }}`
with failed run id '{{ failed_run_id }}'. Any of the three may be empty.

1. With a run id, read the run and its logs.
2. Without one, resolve it from the job ref or the trigger.
3. With neither, you have nothing to inspect: return `status: aborted` naming what was
   missing.
```

**Inputs.** `inputs` is a JSON Schema, and every property in it is a job configuration key: `-c failed_run_id=...` on the command line, `jobs.<section>.<job>.failed_run_id` in `config.toml` or the environment, or a run argument the trigger carries. The body refers to inputs as `{{ name }}`, and to the run as `{{ run_context.trigger }}` or `{{ run_context.run_id }}`; placeholders are rendered before the first turn. A required input with no value fails the run like any missing job argument. An optional one nobody supplied renders blank, is logged, and is listed in the trace as unresolved; an input the body never mentions is a warning at manifest time. There is no `inputs.prompt`: the task belongs in the body.

**Entities.** An input with `entity_type` names a workspace entity: `job-run`, `job`, `pipeline`, `dataset` or `workspace`. The value the agent receives is the bare id: a run id, a job ref, a pipeline name. Declaring it buys two things. The agent run reports the entity in its result's `object` list, so the run shows up on that entity's page. And the first entity-typed input lands in the manifest as `expose.object_input`, which is how a UI offers the agent job from a failed run's row and knows which key to pass the run id under (§6).

**Outputs.** `output` is optional. `status` and `summary` are always part of it, with descriptions dlt writes so the model knows when to report which outcome; `output` adds the definition's own fields. Descriptions matter: the model reads them when it fills a field. `status` is `succeeded` or `failed` as the prompt defines them, and `aborted` when the task could not be done at all; `aborted` ends the run with an exception that carries `summary`.

### 1b. As a function

No file, no toolkit. The docstring is the system prompt, the parameters are the inputs, the return type is the output; together they are the agent definition, and the decorator arguments are the agent job's settings. The body drives the loop it finds in `run_context["ai_loop"]`.

```python
from typing import Annotated, List, Literal

import dlt
from dlt.common.typing import NotRequired
from dlt.hub import run


class CrashReport(run.TAgentOutput):
    """`status` and `summary` come from the base."""
    classification: Annotated[
        Literal["config", "credentials", "code", "upstream_data", "unknown"],
        run.Doc("What kind of failure it was"),
    ]
    confidence: Literal["high", "medium", "low"]
    evidence: NotRequired[List[str]]


@run.agent(
    access={"local": ["read"], "data": ["read"], "context": ["read"]},
    tools=["telemetry"],
    skills=["dlthub-platform:debug-deployment"],
    trigger="job.fail:*",
    limits={"max_turns": 30},
)
async def crash_inspector(
    failed_run_id: Annotated[str, run.Entity("job-run")] = dlt.config.value,
    depth: int = 2,
    run_context: run.TJobRunContext = None,
) -> CrashReport:
    """Inspects a failed job run and reports a diagnosis.

    You run unattended, seconds after a job failed. Explain the failure; do not repair it.
    Investigate run '{{ failed_run_id }}' from `{{ run_context.trigger }}`, going
    {{ depth }} runs back. Return `aborted` when the inputs give you nothing to inspect.
    """
    loop = run_context["ai_loop"]
    report = await loop.run(inputs={"failed_run_id": failed_run_id, "depth": depth})
    if loop.trace["total_tokens"] > 800_000:
        report["summary"] += "\n\n_Investigation was expensive; consider narrowing the trigger._"
    return report
```

| in Python | in the agent definition |
|---|---|
| function name | `name`, the agent job's name, and with `<module>:` in front the agent definition reference when no `AGENT.md` is behind it |
| docstring | system prompt, placeholders included; its first line is the `description` |
| parameters | `inputs`, and so the job's configuration: `-c failed_run_id=...` fills them, typed |
| `Annotated[str, run.Entity("job-run")]` | an entity-typed input |
| `dlt.config.value` default | a required input |
| `run_context` parameter | the implicit input the launcher passes; in neither `inputs` nor `config_keys` |
| return type deriving from `run.TAgentOutput` | `output`; `run.Doc(...)` on a field is its description |
| `access=`, `tools=`, `skills=`, `rules=` | the matching `AGENT.md` fields of the definition |
| `model=`, `limits=`, `loop_run_args=`, `instructions=`, `trigger=`, `loop=` | the agent job's settings, `defaults` in an `AGENT.md` |

The schemas come from pydantic, `TypeAdapter(func).json_schema()` for the signature and the return type for the output, so `Optional`, `Literal`, `List`, nested models and `NotRequired` behave as they do everywhere else. The return type derives from `TAgentOutput`, the payload, not from the result envelope, so the schema the model sees never contains a launcher-owned key.

The function may be `def` or `async def`. Returning the loop's output as-is is the common case; the body above exists to show that the job owns the run: it can inspect `loop.trace`, run the loop twice, or not run it at all. A function that only drives an installed agent definition is a one-liner, with `agent=` as the base that the decorator arguments and the function override:

```python
@run.agent(agent="dlthub-platform:job-inspector", loop="claude-agent-sdk")
async def inspect(run_context: run.TJobRunContext = None) -> run.TAgentOutput:
    return await run_context["ai_loop"].run()
```

## 2. Declare the agent job

The agent job is the agent definition plus the settings that say how it operates: model, limits, trigger, instructions, loop, and later the identity that grants its access. It is what pydantic-ai and claude-agent-sdk call an Agent, and what the web UI lists as one.

```python
from dlt.hub import run

inspector = run.agent(
    "dlthub-platform:job-inspector",
    trigger="job.fail:tag:ingest",
    model="opus",
    limits={"max_turns": 20},
    instructions="focus on the loader step",
)
```

The job is named after the agent definition (`job-inspector` becomes `job_inspector`) in the declaring module's section. Instead of a `"<toolkit>:<agent>"` reference you can pass a workspace-relative path to a folder holding an `AGENT.md`, or a `TAgentSpec` dict declared inline. Also accepted: `loop=`, `loop_run_args=`, `verbosity=`, `inputs_validator=`, `outputs_validator=`, `name=`, `section=`, plus the usual `trigger`, `execute`, `expose`, `require`, `spec`.

`job.fail:` and `job.success:` take a job ref or any selector dlt already has, the same grammar `--select` and the runtime's trigger API use:

```
job.fail:tag:ingest          every job tagged `ingest`
job.fail:batch:              every batch job
job.fail:jobs.mod.*          every job in section `mod`
job.fail:*                   every job
```

A selector expands at manifest time to one concrete `job.fail:jobs.<section>.<name>` per matching job, because the runtime looks a completion trigger up by exact string. It never expands onto the job that declares it, and interactive jobs are never targets. A job event never becomes a job's `default_trigger`: run by hand, an inspector sees `manual:` and aborts, as its body says it should.

## 3. Start an agent run

Every trigger of the agent job produces an agent run: inputs in, output and trace out. Settings may be overridden for that run alone, the way `Agent.run()` takes overrides in the frameworks:

```bash
dlthub local run job_inspector -c failed_run_id=89826ee6-... -c agent.instructions="explain, do not fix"
```

Locally and on the runner the entry point is the same, `dlt._workspace.deployment.launchers.agent`, and it delegates to the ordinary job launcher: config env vars, profile activation, interval injection, signal interception and result delivery are those of any function job.

- **`agent.instructions`** is the user prompt, the run's first message. The system prompt says who the agent is; instructions say what to do this time. Without any, the loop sends a bare "Go ahead" and the system prompt speaks alone. It resolves like `model`: decorator argument, then configuration, then `loop.run(instructions=...)` for a direct call.
- **`agent.verbosity`** is how much of the transcript goes to stdout: `0` the turns, tool names and results; `1` (default) plus the agent's thoughts and the tool arguments; `2` plus the full rendered system prompt, nothing cut. The user sets it on the agent job; configuration overrides it per run.
- **`agent.model`, `agent.max_turns`, `agent.max_tokens`** raise or lower the job's settings for one run.
- Precedence for every setting: loop default, then `AGENT.md` defaults, then the decorator argument, then configuration, so a runtime value always wins.

Everything resolves through normal dlt config (env, toml, `-c`):

| what | key |
|---|---|
| declared inputs | `jobs.<section>.<name>.<input>`, so `-c failed_run_id=...` |
| loop, instructions, model, limits, `loop_run_args`, verbosity | `jobs.<section>.<name>.agent.*`, so `-c agent.verbosity=2` |
| the model endpoint | `agent.model` / `agent.api_key` / `agent.api_url` / `agent.api_version`, each with a `runtime_*` twin the runtime supplies; one whole set is used, never a mix |
| trace export | `agent.trace_url` / `agent.trace_key` |

An agent job's inputs are its job configuration: they take the type from the schema (`-c depth=3` arrives as `int`), and a required input with no value fails the run like any missing job argument.

## 4. Read the result

The agent run prints its transcript to stdout as it goes, in color whether or not a terminal is attached (`NO_COLOR` turns it off): what the model thinks, says and calls, what each tool handed back, and a finish line with the tools, skills and MCP tools it used. When it ends, the launcher prints and delivers the job result:

```jsonc
{
  "type": "job.background_agent.dlthub-platform:job-inspector",
  "engine_version": 1,
  "job_ref": "jobs.__deployment__.job_inspector",
  "status": "succeeded",
  "summary": "## Root cause: platform runner token conflict ...",
  "result": { "status": "succeeded", "summary": "...", "classification": "transient", "confidence": "high", "evidence": [...] },
  "object": [
    { "type": "job-run", "id": "job-run/89826ee6-ed5d-478e-8ad1-415251f225e0" },
    { "type": "job", "id": "job/jobs.maintenance.compact_delta_tables" }
  ],
  "trace": { "loop_type": "pydantic-ai", "model": "anthropic:claude-sonnet-5", "turn_count": 3, "total_tokens": 13215, ... }
}
```

- **`status`** and **`summary`** are lifted out of the agent output to the top, so a list never opens `result`. `succeeded` and `failed` mean what the prompt says they mean; `aborted` means the task could not be done at all, and raises `JobAbortedException` with `summary` as its text, after the result is delivered.
- **`result`** is the agent output, exactly as the definition's `output` schema declares it.
- **`object`** is the entities the run acted on.
- **`trace`** is what the loop ran with and did: agent and file, loop type, model, limits, `loop_run_args`, the user turn, the resolved inputs, declared `access` with the `local_tools` the loop wired, MCP features, native and inlined skills, unresolved placeholders, turn count and token counts, per-turn tool calls, `tools_used` / `skills_used` / `mcp_tools_used`, and `cost_usd` / `stop_reason` where the loop reports them.

# Part 2. Mechanisms and architecture

## 5. One envelope for every job, shaped like Activity Streams 2.0

Any job can declare a structured result; it names only the payload shape:

```python
run.result({"rows": 10}, type="etl_summary")
```

The launcher completes the envelope when the run ends: `type`, `job_ref`, `object`, and for agents `status`, `summary` and `trace`, then delivers it to the telemetry beacon as payload type `job_result`, synchronously and before an abort is raised, so the trace is on the wire even when the process exits non-zero. The body carries no run id; the beacon knows the run from the token in the DSN, as it does for pipeline traces.

- **`type`** is `job.<category>.<name>`, two discriminators in one string, and it links a result to its shape. The category (`background_agent`, `pipeline`, `batch`, ...) names the envelope: `background_agent` means `status`, `summary` and `trace` are present. The name names the payload schema of `result`. For a plain job it is whatever `run.result(..., type=)` said. For an agent job it is the **agent definition reference**, because the definition's `output` schema is the payload schema: `dlthub-platform:job-inspector` for a toolkit agent, the workspace path for a definition referenced by path, or `<module>:<function>` (`agents:crash_inspector`) for a decorated function with no `AGENT.md` behind it. Every job that runs the same definition produces the same `type`, so one template renders all their results; the schema itself is `output` on the job definition the result's `job_ref` points at. The job never writes the prefix; the launcher adds it, because only the launcher knows the category. The first two segments are closed vocabularies with no dots, so the name may carry dots, colons and slashes.
- **`object`** is derived the same way for every job kind, from its entity-typed inputs and outputs (§6).
- **`result`** is the payload. **`trace`** is agent-specific.

The envelope follows the Activity Streams 2.0 model the dltHub audit log is being designed on: an actor performs a verb on an object, optionally toward a target, and the activity carries a `result` and a `summary`.

| Activity Streams 2.0 | job result |
|---|---|
| `actor` | the agent job, `job_ref`, acting through one run, the beacon token; its kind is the agent definition reference, the name segment of `type` |
| `type`, the verb | implicit and constant: a job ran (`run:job`, `end:run` in the audit vocabulary). The envelope's own `type` is not the verb but the schema discriminator above |
| `object`, `target` | `object`, the entities the run acted on |
| `result` | `result`, and `status` |
| `summary` | `summary` |
| `id`, deduplicated | deduplicated on `job_ref` within the run |

An agent run therefore projects onto the audit log as `run:job` by the agent job, with the inspected run as its object, and an entity's activity view can list job results next to audit events without a second model.

## 6. Entities: from an input to `object` and `expose.object_input`

An input is an entity reference when its schema says so: `entity_type: job-run` in `AGENT.md`, or `Annotated[str, Entity("job-run")]` on a parameter. Known types are `job-run`, `job`, `pipeline`, `dataset`, `workspace`; they are URI path segments, hence the hyphen. The value carries only the unique id; dlt composes `{type}/{id}` when it reports: `job-run/89826ee6-...`, `dataset/duckdb_prod/github_events`, `pipeline/github_actions`, `job/jobs.agents.job_inspector`. At the end of a run, `object` is every entity-typed input the run received, overwritten by an output property of the same name, so an agent that resolves a run id it was not given still reports the run it inspected. The loops hand the model the same schema with `entity_type` moved into `$comment`: strict validators such as the Claude CLI's reject keywords they do not know, and `$comment` is the standard slot a model still reads.

The manifest generator takes the first entity-typed input and writes it to `expose`:

```yaml
expose:
  category: background_agent
  object_input:
    entity_type: job-run
    input: jobs.__deployment__.job_inspector.failed_run_id
```

That is all a UI needs to start an agent run for an entity. On a list of job runs, a failed run's row offers every agent job whose `object_input.entity_type` is `job-run`; choosing one starts a run with the run id under the key `object_input.input` names, plus an optional prompt:

```json
{
  "script_id_or_ref": "jobs.__deployment__.job_inspector",
  "trigger": "manual:",
  "config": {
    "jobs.__deployment__.job_inspector.failed_run_id": "89826ee6-ed5d-478e-8ad1-415251f225e0",
    "jobs.__deployment__.job_inspector.agent.instructions": "explain, do not fix"
  }
}
```

The launcher turns `config` into job configuration under the job's section, the inputs resolve as if typed on the command line, and the result comes back with that run in `object`, closing the loop: offered from the entity, reported on the entity. `object_input` names the primary entity only; `inputs.properties[*].entity_type` lists every entity the job can be pointed at.

## 7. Access: what a job may touch

`access` is a job-level manifest field (`TJobDefinition.access`), not an agent-only one. Agents are the first job kind to declare it and the only producer for now; the field lives where the next job kind can fill it.

| axis | verbs | meaning |
|---|---|---|
| `local` | `read`, `write`, `execute`, `network` | what may be done on the machine the workspace lives on |
| `data` | `read`, `write` | workspace data, governed by the dlt profile |
| `context` | `read` | the context graph: telemetry, runs, job definitions |

`all` is shorthand for every verb on an axis, and a single verb may be written without the list (`data: read`). The type carries a wider `context` ladder (`write`, `execute`, `deploy`) than any runtime serves, so a manifest declaring one is refused at generation with a message saying it is typed but unserved.

**Only `local` depends on the loop.** `data` and `context` are served by the workspace MCP server, started with the declared access, so nothing can be granted beyond it. A loop that lacks a tool for a verb wires less, and the run proceeds narrowed. What it did wire is in the trace as `local_tools` and in the result view, each tool under its verb: `Grep (read), Bash (execute), WebSearch (network)`.

**`access.local` maps to one standard tool set**, named as the Claude Code harness names its tools, so the same declaration means the same thing on both loops:

| verb | tools |
|---|---|
| `read` | `Read`, `Glob`, `Grep` |
| `write` | `Write`, `Edit` |
| `execute` | `Bash` (`PowerShell` on Windows), `RunPython` |
| `network` | `WebFetch`, `WebSearch` |

The tools themselves are `LocalTools` in `deployment/agent/loops/tools.py`: plain methods confined to the workspace root and the system temp folder, tested on their own. The pydantic-ai loop wraps them as function tools, and `network` maps to the provider's own web capabilities. Both loops name the workspace and the temp folder in the system prompt so the model knows where scratch files go; the claude loop also adds the temp folder to the harness's directories and attaches what it has beyond the standard names to the name it extends (`Read` brings `NotebookRead`, `Edit` brings `MultiEdit` and `NotebookEdit`, `Bash` brings `BashOutput` and `KillShell`). The shell is one per platform, as Claude Code offers it: `Bash` everywhere but Windows, `PowerShell` there, run without a profile and non-interactively; on both loops the model is told which one it has, and tool output is UTF-8 whatever the console code page. Paths the tools return are posix and relative to the workspace root on every platform. No verb, no tool: an agent that declares no `access` gets no file tool and no shell, and its MCP server serves only the tools that require nothing, the toolkit catalogue.

Nothing here is an OS sandbox: `execute` runs in the same process tree and virtualenv as the job, so what contains it is the runner the job already runs in. Two things are enforced regardless of the declared access:

- **credential files are never readable by a file tool**: `*secrets.toml`, `.env`, `.env.*`, on both loops (deny rules on the harness, a refusal inside dlt's own tools);
- **`execute_sql_query` runs one read-only statement**, whole-tree checked against mutating statements and host functions after [GHSA-xxqf-4m3r-qx6c](https://github.com/dlt-hub/dlt/security/advisories/GHSA-xxqf-4m3r-qx6c).

## 8. MCP tools: how they are injected

`tools:` names feature groups to request from the dlthub MCP server: `telemetry`, `pipeline`, `jobs`, `logs`, whatever the installed plugins contribute through `plug_mcp`. The loop turns that into a server the agent talks to over stdio:

```
dlthub ai mcp run --stdio --access data:read,context:read --no-default-features --features telemetry
```

Three parts, each doing one job:

1. **`--features`**: exactly the groups the agent asked for. `--no-default-features` keeps the interactive defaults (workspace, pipeline, toolkit, secrets, context) out: an agent that asks for one group does not silently get the rest.
2. **`--access`**: the agent definition's `access`, serialized as `axis:verb` pairs. Server-side it is parsed back and every tool is checked against it before `add_tool`, so a tool the grant does not cover does not exist in the schema the model sees. Of the 17 tools dlt serves, `data:read` leaves 10, `local:write` leaves 3, and a grant of nothing, spelled `local:,data:,context:` (an axis with no verb grants nothing on it), leaves the 2 that need nothing (the toolkit catalogue). An empty `--access` is refused rather than read as anything. A person at a terminal running `dlthub ai mcp run` without `--access` is granted everything explicitly: the command hands the full grant to the server, which always filters against the grant it holds.
3. **the environment**: the workspace directory and `WORKSPACE__PROFILE` the launcher exported, so the server reads the same profile and config as the job.

Tools declare what they require next to what they return, and dlt reads it; pydantic drops metadata it does not recognise, so nothing of it reaches the tool schema:

```python
def list_pipelines() -> Annotated[List[str], RequiresAccess(data=["read"])]: ...
def get_workspace_info() -> Annotated[Dict[str, Any], RequiresAccess(local=["read"])]: ...
def export_schema(...) -> Annotated[str, RequiresAccess(data=["read"], local=["write"])]: ...
```

Silence is not a grant: a tool that declares nothing is assumed to need everything, so it reaches no narrowed caller until someone has decided what it may touch. `RequiresAccess()` is how a tool says it needs nothing.

Wiring differs per loop, the surface does not:

- **pydantic-ai** attaches the server as an `MCPToolset` over `StdioTransport`, id `dlt-workspace-mcp`, booted when the agent enters its context. Tool errors, MCP and local alike, follow `loop_run_args.retries`: at 0 (the default) every error reaches the model as a failed call and the run goes on, as on the claude loop; above 0 pydantic-ai asks the model to correct the call that many times per tool and ends the run when the budget is spent.
- **claude-agent-sdk** passes it as `mcp_servers` with `strict_mcp_config`, and allows `mcp__dlt-workspace-mcp__*`. The project's own `.mcp.json` is never loaded; the loop always spawns its own server.

`tools: []` (or no `tools`) means no server is started at all.

## 9. Loops

`pydantic-ai` (default) and `claude-agent-sdk` ship; `loop="claude-agent-sdk"` on the decorator, overridden by `agent.loop` in config. A third-party loop registers through the `plug_agent_loop` plugin hook and needs no changes here. Each loop maps to a dependency group (`agent-loop-pydantic-ai`, `agent-loop-claude-agent-sdk`) which the job stamps into `require.dependency_groups` and dlt writes into the requirements manifest, so the runner installs only the loop that job needs. claude-agent-sdk ships a bundled Claude Code CLI binary; nothing else is installed.

**What each loop does with the agent definition.** `access` becomes the tool surface (§7), `tools` the MCP server (§8), the rendered body the framework's system prompt, `instructions` the user turn, `output` structured output with `entity_type` moved into `$comment`. Rules are inlined into the system prompt on both loops. Skills follow the loop: pydantic-ai inlines their text; claude-agent-sdk lists the declared skills by name and loads one when the agent invokes it, the way Claude Code does natively. For that it reads the project settings, adds the `Skill` tool to the base tool set only when skills are declared, and keeps everything else in the project out of the agent's context: `.claude/rules/*.md` are excluded through the `--settings` layer (merged with any `settings` in `loop_run_args`) because in a dlthub workspace they name every installed skill, while `CLAUDE.md` and `CLAUDE.local.md` load as in any Claude Code session. The pydantic loop attaches the prompt as a function rather than through `AgentSpec.instructions`, because that field is re-rendered as a Handlebars template when `deps_schema` is set and would drop the `{{ }}` dlt has already resolved.

**Limits are dlt's.** `limits.max_tokens` is counted by dlt itself after each turn, so it means the same on every loop; pydantic-ai's native limit and claude-agent-sdk's lack of one no longer matter, and a run that passes it stops with `AgentTokenLimitExceeded`. `limits.max_turns` is handed to the framework.

**Transcript.** Every loop emits the same events (start, system prompt, prompt, turn, thinks, says, tool call, tool result, finish) and one renderer prints them, with the verbosity levels of §3. On pydantic-ai a tool the loop did not build itself is counted as the MCP server's. Every run ends with the tools, skills and MCP tools it used, on the console and as an INFO line in the log.

## 10. Code map

- **`run.agent` and `AgentJobFactory`** (`deployment/decorators.py`): the agent job. When the job definition is built it resolves the agent definition (referenced, given inline, or read off the function), emits the manifest `agent` block, the job-level `access`, `inputs`, `output` and `expose.object_input`, and synthesizes the job's config spec from the declared inputs.
- **`deployment/agent/reflection.py`**: signature, return type and docstring to `TAgentSpec`, the agent definition, merged over the one the decorator referenced. Pydantic renders both JSON Schemas.
- **`deployment/agent/manifest.py`**: reads `AGENT.md`, resolves `<toolkit>:<agent>` refs through the toolkit index installed in the workspace, renders `{{ }}` placeholders against the run's inputs, inlines rules and skills a loop cannot load natively.
- **`_workspace/access.py`**: the access vocabulary shared by agents, tools and the MCP server: `TWorkspaceAccess`, `RequiresAccess`, and the parse/format helpers.
- **`deployment/agent/loop.py`**: the `AgentLoop` contract: `init(agent_spec)` takes the agent definition, `run(instructions, inputs, model, limits, run_args)` is one agent run: it renders the system prompt against the inputs, builds the framework's Agent and runs it. Settings resolve loop default < `AGENT.md` defaults < decorator argument < configuration.
- **`deployment/agent/loops/{pydantic_ai,claude_sdk}.py`**: one adapter each; **`loops/tools.py`**: `LocalTools` and the MCP server command.
- **`deployment/launchers/agent.py`**: a thin call into `launchers/job.py` with an `invoke` that builds and drives the loop, so all job-launcher setup is shared by construction rather than copied.
- **`deployment/job_result.py`**: `run.result`, the per-run result container, `result_type` / `parse_result_type`, and beacon delivery; **`deployment/entity.py`**: `THubEntity` construction and the `object` merge.
- **`deployment/_run_views.py`**: the transcript and result renderer.
- **Packaging**: the toolkit index under `.dlt/.toolkits` ships with the deployment package so `<toolkit>:<agent>` refs resolve on the runner; the agent definition's files travel as ordinary workspace files.

## 11. Required runtime changes

The runtime imports dlt's `TJobDefinition` into pydantic models with `extra="forbid"` and freezes it into nine generated mirrors (three OpenAPI specs, three attrs clients including `dlthub-client`, two TypeScript schemas). Every item below therefore needs a dlt pin bump plus regeneration before any agent job can deploy.

1. **New job-definition fields.** `engine_version` (the manifest engine the definition is written for, always the manifest's own; `migrate_job_definition` stamps it, so a definition stored on its own can be migrated later), `agent` (the `TAgentDefinition` block: engine_version, `agent_file`, name, description, tools/skills/rules, instructions, model), `access` (`TWorkspaceAccess`, without `toolkits`), `inputs` / `output` (JSON Schema, on every job kind, not just agents), `expose.object_input` (`TJobObjectInput`: `entity_type` and the full config key of the first entity-typed input), and `background_agent` in `TExposeSpec.category`. All are optional; nothing in the runtime reads them yet, so this is a schema move, not a behaviour change. They ride manifest engine 2 (the #4033 bump) and need no migration of their own. Every workspace's manifest hash moves once, because a job that takes arguments now carries `inputs`. The frozen category enum blocks deploying any agent job today; `tests/hub/test_manifest_contract.py` fails until it lands.
2. **Run requests take `config`.** `CreateRunRequest` gains a `config` map of full config keys (`jobs.<section>.<job>.<key>`), carried to the vault and passed as `cli_config` to `build_runtime_entry_point`, stripped of the `<job_ref>.` prefix and nested by the remaining dots. This is what §6 and a "run with prompt" dialog need; the launcher side already works.
3. **Telemetry beacon must accept `job_result`.** Payloads are `PUT` as payload type `job_result` with the envelope of §5. The beacon's payload type enum has `trace` and `state` only and its dedup key falls back to `transaction_id`, so today it answers `ignored` and drops every job result, agent traces included. Add the type, dedup on `job_ref` within the run, store it, and expose `status`, `summary`, `object` on the run response and `result`, `trace` on a detail call.
4. **Loop dependency groups and the MCP server.** The runner installs the groups a job names in `require.dependency_groups`, which it already does by flattening them into the run's requirements. Both loops reach the platform through the workspace MCP server, so the job environment needs the `dlthub` console script and the `hub` extra; the loops spawn `dlthub ai mcp run --stdio` themselves.
5. **Agent credentials, including the API version.** The runtime supplies the model endpoint as `agent.runtime_api_key` / `agent.runtime_api_url` / `agent.runtime_api_version` and `agent.runtime_model` in the job's config section (or workspace-wide `AGENT__RUNTIME_*`). The model id belongs to the endpoint that serves it, so the four are one set: setting any one of `agent.model` / `agent.api_key` / `agent.api_url` / `agent.api_version` makes the run use the user's four and ignore all four runtime values, and the loop logs at INFO which endpoint it picked. Both sets stay readable side by side. dlt maps each to the argument the named provider uses (`azure_endpoint` and `api_version` for Azure, `api_base` for LiteLLM, `api_host` for xAI, `base_url` elsewhere) and refuses an `api_url` on a provider with a fixed endpoint rather than silently calling the wrong host.
6. **Profile.** dlt does not infer a profile from the agent. It exports `WORKSPACE__PROFILE` for the job and the tools read it; the runtime decides the profile from the declared `access.data` (`write` for prod, `read` for access) or from what the user pinned on the job.
7. **`default_trigger` recompute.** The runtime recomputes it with dlt's own `compute_default_trigger`, so the job-event exclusion arrives with the pin bump; jobs already stored keep their column until redeployed. Null is already handled everywhere it is read: an explicitly named job with no default is fired as `manual:<ref>`.
8. **No change needed, verified:** the run repackager passes the deployment tarball through unchanged, so the shipped `.dlt/.toolkits` index and the agent's files reach the runner as-is.

### DIFF
--- a/.claude/skills/review-vocabulary/SKILL.md
+++ b/.claude/skills/review-vocabulary/SKILL.md
@@ -130,6 +130,9 @@ Groups defined so far:
 | [G3 — Transformations and materialization](#g3--transformations-and-materialization) | relations, transformations, model jobs, eager and lazy paths |
 | [G4 — Identifiers and SQL generation](#g4--identifiers-and-sql-generation) | naming conventions, case-folding, query binding |
 | [G5 — Configuration and credentials](#g5--configuration-and-credentials) | configs, credentials, secrets |
+| [G6 — Jobs, triggers and the deployment manifest](#g6--jobs-triggers-and-the-deployment-manifest) | jobs, job runs, launchers, triggers, selectors |
+| [G7 — Agents, loops and prompts](#g7--agents-loops-and-prompts) | agent definitions, agent jobs, agent runs, loops, system prompts, user turns, traces |
+| [G8 — Workspace access and tools](#g8--workspace-access-and-tools) | the `access` declaration, axes, verbs, tools, feature groups |
 
 Two rules apply across every group:
 
@@ -160,9 +163,11 @@ Two rules apply across every group:
 
 **Rulings**
 
-- **`access` is a VERB.** Do not introduce noun uses — no "data access", no "access is one-way".
-  Write "the engine accesses the data", "only one direction accesses the data". Pre-existing noun
-  uses stay.
+- **`access` is a verb for the act, and a noun for what a job may touch.** Keep the verb for the
+  act of reading data: "the engine accesses the data", not "data access is one-way". The noun is
+  legal only in the sense G8 defines — the `access` declaration, and the grant that answers it.
+  dlt's docs already write "grant access", "read access" and "denied access" 40 times over; that
+  usage was always compliant and stays.
 - **`reach` is not always `access`.** "`SET SESSION` would not reach the cloned sessions" means
   *propagate to*, not *read data from*. A literal swap changes the meaning — restructure
   (Rule 9.1).
@@ -260,10 +265,157 @@ Two rules apply across every group:
 
 ---
 
+### G6 — Jobs, triggers and the deployment manifest
+
+**Included**
+
+| Concept | Write |
+|---|---|
+| the unit a workspace deploys and runs | **job** |
+| one execution of it | **job run** |
+| its identity | **job ref** |
+| its entry in the manifest | **job definition** |
+| the dlt module that starts a job | **launcher** |
+| the machine the platform gives the job | **runner** |
+| the dltHub platform itself | **runtime** |
+| the string that starts a job | **trigger** |
+| the pattern that expands into triggers | **selector** |
+| the trigger a manual run stands in for | **default trigger** |
+| the file that describes every job | **deployment manifest** |
+
+**Excluded**
+
+| Never | Because |
+|---|---|
+| task, workload, script (for a dlt job) | one name — **job**; the runtime's `Script` model keeps its own name |
+| job execution, invocation (for one run) | **job run**, the noun the manifest and the beacon use |
+| primary trigger | **default trigger**, the name of the manifest field |
+| manifest (unqualified, for `AGENT.md`) | **deployment manifest** is the only manifest; see G7 |
+
+**Rulings**
+
+- **`run` is a verb; `job run` is the noun.** "dlt runs the job" and "the job run failed". The
+  command `dlthub local run` is a technical name (Rule 8.6) and stays.
+- **launcher, runner and runtime are three things.** The launcher is dlt code. The runner is a
+  machine. The runtime is the platform. Never swap them, and never write "the runtime launches".
+- **`task` is legal in two places only:** the work an agent must do (G7), and an Airflow task when
+  the sentence says "Airflow task". It is never a dlt job.
+
+---
+
+### G7 — Agents, loops and prompts
+
+Three nouns carry the feature, and they nest: a definition is written once, a job adds how it
+operates, a run is one execution.
+
+**Included**
+
+| Concept | Write |
+|---|---|
+| what an `AGENT.md` or a decorated function declares: system prompt, inputs, output, tools, skills, rules, access | **agent definition** (`TAgentSpec` in code, `TAgentDefinition` for the block the manifest carries; pydantic-ai's `AgentSpec`) |
+| the file it is read from | **`AGENT.md`**, and **agent file** for the path |
+| the name a definition is referred to by | **agent definition reference**: `<toolkit>:<agent>`, a workspace path, or `<module>:<function>` for a function with no `AGENT.md` behind it |
+| the definition plus the runtime settings that say how it operates: model, limits, trigger, instructions, loop, identity | **agent job** (`run.agent(...)`; what pydantic-ai and claude-agent-sdk call an Agent, and what the web UI lists as one) |
+| one execution of an agent job: inputs in, output and trace out, settings overridable for that run | **agent run** (`Agent.run()` in the frameworks) |
+| what the model returns: `status`, `summary` and the declared fields | **agent output** |
+| the envelope the launcher delivers | **job result** (G6 owns it; an agent job's is `TAgentJobResult`) |
+| the binding to one agent framework | **loop**, or **agent loop** |
+| the text the model gets as its role and task | **system prompt** |
+| the first message of the run | **user turn** |
+| what a person tells this run to do | **instructions** |
+| one model request | **turn** |
+| the record of what the loop ran and did | **agent trace** |
+| `{{ name }}` in a body | **placeholder** |
+
+**Excluded**
+
+| Never | Because |
+|---|---|
+| agent spec (in prose) | **agent definition**; `TAgentSpec` keeps its name as a type |
+| agent reference, agent ref (in prose) | **agent definition reference**; the field `agent_ref` keeps its name |
+| agent (bare) where the sentence needs one of the three | say which: **agent definition**, **agent job** or **agent run** |
+| harness (for the Claude Code CLI) | the **dltHub AI Harness** owns that word; write **the Claude Code CLI** |
+| adapter (for a loop) | dlt's adapters are `bigquery_adapter` and friends; write **loop** |
+| prompt (bare) | say which one: **system prompt** or **user turn** |
+| spec (bare, for an agent) | bare `spec` is the configspec; write **agent definition** |
+| trace (bare, for an agent) | bare `trace` is the pipeline trace; write **agent trace** |
+| manifest (for `AGENT.md`) | **agent file**; the manifest is the deployment manifest (G6) |
+| framework (as a loop's name) | name it: **pydantic-ai**, **claude-agent-sdk** |
+
+**Rulings**
+
+- **Bare `agent` is the actor.** "The agent reads the logs", "the agent returns `aborted`": the
+  model acting during a run. That stays. The moment a sentence is about the file, the `run.agent`
+  product or one execution, it names the definition, the job or the run.
+- **`Agent` (capitalised) is the agent job.** That is how the web UI labels it and how pydantic-ai
+  and claude-agent-sdk name the same thing. Legal in UI copy and when talking about a framework;
+  in dlt prose write **agent job**.
+- **An agent run is a job run.** Use **agent run** when the sentence is about what the agent
+  received and produced; **job run** when it is about scheduling, status or logs, as for any job.
+- **`instructions` means two opposite things across the boundary.** In dlt it is the user turn. In
+  pydantic-ai, `AgentSpec.instructions` is the system prompt. Qualify every mention of the
+  framework field: "pydantic-ai's `instructions` field (its system prompt)".
+- **`turn` is one model request.** The **user turn** is the first message. A turn counter counts
+  requests. Do not let one word carry both without the qualifier.
+- **`loop` is the agent loop.** In a file that also has Python loops, write **agent loop** once,
+  then `loop`.
+- **Model names and aliases are technical nouns** (Rule 1.8): `sonnet`, `opus`, `claude-sonnet-5`.
+
+---
+
+### G8 — Workspace access and tools
+
+**Included**
+
+| Concept | Write |
+|---|---|
+| the declaration of what a job may touch | **access** (the `access` block, `TWorkspaceAccess`) |
+| one of its four keys | **access axis** |
+| a value on an axis | **verb** |
+| what the runtime or a loop does with the declaration | **grant**, **deny** (verbs) |
+| what it did grant | **granted access** |
+| what it did not grant | **denied access** |
+| what a loop supplies beyond the declaration | **over-granted** |
+| what a tool needs before it is served | **required access** |
+| what a model can call | **tool** |
+| the MCP grouping a manifest requests | **feature group** |
+
+**Excluded**
+
+| Never | Because |
+|---|---|
+| permission, entitlement, scope (for the declaration) | **access**, the word dlt's docs already use; `permission` is legal only when naming the SDK's `permission_mode` |
+| narrowed, narrowing | say the fact: **with less access** |
+| grant (as a noun) | **the declared access**; `granted` as an adjective is legal (Rule 3.3) |
+| capability (for a dlt tool) | **tool**; legal only for pydantic-ai `capabilities` |
+| ceiling, floor (as prose metaphor) | say what it does: "access is the most a loop can wire" |
+| honor, honour, honored, unhonoured (any spelling) | access is **granted** or **denied**, never honored |
+
+**Rulings**
+
+- **Grant and deny are the verbs.** A job declares access; the runtime and the loop grant what they
+  can; the rest is denied. dlt's docs already write "grant access" 13 times and "denied access"
+  twice, so this is the house pairing, not a new one. A loop that supplies more than the
+  declaration **over-grants**.
+- **The declaration is a request, not a claim.** A manifest `access` block says what the job wants.
+  Nothing in it is granted until the runtime says so. Say "the job declares", never "the job has".
+- **`access` is the noun, and no synonym joins it.** `permissions`, `scopes`, `capabilities` and
+  `entitlements` each name this concept somewhere in the industry; dlt named it `access` before this
+  feature existed — the built-in `access` profile is coarse-grained data access, and `profile_for()`
+  derives that profile from `access.data`. One concept, one word (Rule 1.11).
+- **`tools:` in an `AGENT.md` holds feature groups, not tools.** Write "feature groups" whenever the
+  sentence is about that field, or a reader counts 4 tools and gets 19.
+---
+
 ### Legal technical nouns — never replace, any group (Rules 1.5, 1.8)
 
 attach, attach alias, attach info, attach statement, catalog, config, data location, dataset,
 destination, duckdb, iceberg, materialization, model job, pipeline, relation, scanner, vended.
+
+access, access axis, agent, agent definition, agent definition reference, agent file, agent job,
+agent output, agent run, agent trace, default trigger, feature group, instructions, job, job definition, job ref, job result,
+job run, launcher, loop, placeholder, runner, runtime, selector, skill, system prompt, tool, toolkit,
+trigger, turn, user turn, verb.
 
 ## Rules this codebase breaks most
 
--- a/.github/workflows/test_destinations_remote.yml
+++ b/.github/workflows/test_destinations_remote.yml
@@ -285,7 +285,7 @@ jobs:
           PYTEST_ARGS: ${{ matrix.rerun_args || '' }}
 
   run_vault_providers:
-    name: vault providers
+    name: vault providers and agents
     env:
       # Prevent implicit resyncs by `uv run`; this is ignored by explicit `uv sync` commands.
       UV_NO_SYNC: "1"
@@ -318,11 +318,16 @@ jobs:
           enable-cache: ${{ github.event_name != 'pull_request_target' }}
 
       - name: Install dependencies
-        # google-api-python-client for google secrets, botocore (s3 extra) for aws secrets manager
-        run: uv sync --group providers --extra s3
+        # google-api-python-client for google secrets, botocore (s3 extra) for aws secrets manager,
+        # the workspace and both agent loops for the agent runs
+        run: uv sync --group providers --extra s3 --group workspace-deps --extra cli --group agent --group agent-claude
 
       - name: create secrets.toml
         run: pwd && echo "$DLT_SECRETS_TOML" > tests/.dlt/secrets.toml
 
       - name: Run vault provider tests
         run: uv run pytest tests/helpers/providers
+
+      - name: Run agents on real loops and models
+        # skipped unless the secrets carry `jobs.agent.api_key`
+        run: make test-workspace-agents
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ has-uv:
 	uv --version
 
 dev: has-uv ## Prepares development environment
-	uv sync --all-extras --no-extra hub --group workspace-deps --group dev --group providers --group pipeline --group sources --group sentry-sdk --group ibis --group adbc --group dashboard-tests
+	uv sync --all-extras --no-extra hub --group workspace-deps --group dev --group providers --group pipeline --group sources --group sentry-sdk --group ibis --group adbc --group dashboard-tests --group agent --group agent-claude
 
 dev-airflow: has-uv ## Prepares development environment with airflow support
 	uv sync --all-extras --no-extra hub --group workspace-deps --group providers --group pipeline --group sources --group sentry-sdk --group ibis --group airflow
@@ -278,12 +278,15 @@ test-pipeline-arrow:
 # ----------------------------------------------------------------------
 
 install-workspace:
-	uv sync $(UV_SYNC_ARGS) --group workspace-deps --extra cli --group streamlit
+	uv sync $(UV_SYNC_ARGS) --group workspace-deps --extra cli --group streamlit --group agent --group agent-claude
 
-TEST_WORKSPACE_PATHS = tests/workspace
+TEST_WORKSPACE_PATHS = tests/workspace --ignore tests/workspace/deployment/test_agent_e2e.py
 
 test-workspace:
 	$(call RUN_XDIST_SAFE_SPLIT,$(TEST_WORKSPACE_PATHS))
+
+test-workspace-agents: ## Runs agents on real loops and models (needs jobs.agent.api_key in tests/.dlt/secrets.toml)
+	$(PYTEST_BASE) tests/workspace/deployment/test_agent_e2e.py
 
 # ----------------------------------------------------------------------
 # CI: hub minimal (no ibis)
--- a/dlt/_workspace/_plugins.py
+++ b/dlt/_workspace/_plugins.py
@@ -132,7 +132,7 @@ def plug_mcp_context(features: Set[str]) -> Optional[McpFeatures]:
 
 
 @_plugins.hookimpl(specname="plug_agent_loop")
-def plug_agent_loop_pydantic_ai(loop_type: str) -> Optional[Type[Any]]:
+def plug_agent_loop_pydantic_ai(loop_type: str) -> Optional[Type[_plugins.SupportsAgentLoop]]:
     """Contribute the Pydantic AI agent loop."""
     # the plugin module loads with the run context, so the deployment package stays out of it
     from dlt._workspace.deployment.launchers import LOOP_PYDANTIC_AI
@@ -145,7 +145,7 @@ def plug_agent_loop_pydantic_ai(loop_type: str) -> Optional[Type[Any]]:
 
 
 @_plugins.hookimpl(specname="plug_agent_loop")
-def plug_agent_loop_claude_sdk(loop_type: str) -> Optional[Type[Any]]:
+def plug_agent_loop_claude_sdk(loop_type: str) -> Optional[Type[_plugins.SupportsAgentLoop]]:
     """Contribute the Claude Agent SDK loop."""
     from dlt._workspace.deployment.launchers import LOOP_CLAUDE_AGENT_SDK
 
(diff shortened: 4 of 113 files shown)


sh-rp:
I had a look at how the runtime and the agents in dlt / dlthub interact and below is my first round of feedback. 
There will probably be more :)

## Credentialing: inject env vars, don't namespace under `agent`

Platform access needs no new config surface. The CLI already resolves credentials off `runtime_config` (dlt's `RuntimeConfiguration`, section `runtime`) and env is already a config provider, so the runner just injects `RUNTIME__API_KEY`, `RUNTIME__WORKSPACE_ID`, `RUNTIME__ORGANIZATION_ID` and `RUNTIME__API_BASE_URL` into the job environment. Every job should be able to get an identity, not just agents. And the above will simply work with the future SDK.

## Tools: SDK wrappers vs MCP

The prototype ([runtime#1565](https://github.com/dlt-hub/runtime/pull/1565)) used pydantic-ai tools wrapping an SDK, with a plan to keep that SDK always current with the deployed runtime API. MCP is good too, but it moves the freshness problem rather than removing it — the tools come from whichever `dlthub` is installed in the job. Either way it's API stability that fixes it, which is under way.

## Why is an agent job an `expose.interface`?

No direct objection, but it doesn't read like a job with a particular interface. `gui`, `rest_api` and `mcp` all say *how you reach a running job* — the web filters the interactive grid on it. An agent job is headless, so `background_agent` becomes a value meaning "not reachable". `expose.category` is the field that says what a job *is*. What's the reasoning for interface? Maybe have a job type "agent" next to "interactive" and "batch"? I am unsure.

## `TAgentJobOutput` — a job output, with a notify signal?

The mechanism already looks generic — `run.output(...)` is offered to any job and the beacon payload type is `job_output` — so perhaps only the name is agent-shaped, with the agent envelope as the specialization. Could it also carry a notify signal by default, a boolean or a severity level? It would land in the events stream for users to subscribe to, which covers "notify me if something is wrong" and a deterministic job raising a flag alike. I'm not sure yet which shape covers enough cases. In any case every job should use it and by default should conform to the base shape.

## Reasoning in the logs, on by default?

I'd assumed reasoning was part of the trace, but it isn't — `TAgentTurn` holds tool names and token counts only. `log_activity` already emits thinks / says / tool calls at INFO, which is most of what's wanted, but `log_level` defaults to `WARNING`, so a run says nothing unless someone raises it, and `log_excerpt` caps each line at 800 chars. Could reasoning and tool calls be on by default, interleaved in their real order? The prototype just printed everything, and being able to follow an agent while it works turned out to be worth a lot — both for watching a live run and for auditing why it concluded what it did. Maybe reasoning output to logs should be a setting on the jobdef?

## Is the trace part of v1?

If reasoning and tool calls are in the logs and the conclusion is in the output, the trace may not be needed yet.

sh-rp:
Correction to the above, the env var name is RUNTIME__AUTH_TOKEN, since we will not be using an api key here.

burnash:
Opened #4478 against this branch: `dlthub ai toolkit info` now lists a toolkit's agents next to skills.

PR #4420, opened 2026-09-01
docs: upgrade docs build tooling

zilto:
This PR aims to streamline how building docs works. Now, CI is much faster and checking things locally is trivial, enabling agents to edit docs.

Sorry for the massive diff and messy `git` history. I recommend starting with pulling the branch locally
```shell
git pull
git switch docs/upgrade-tooling
```
and looking at the "final code" to get an idea of how it works.

### Tooling entrypoint consolidation
- Replace the `docs/` Python package and sprawl of Python tools by standalone scripts with inline dependencies under `docs/tools`
- Move the docs start and build process from `package.json` to `docs/Makefile`
- Use `prek` to run all checks in parallel in local dev environment and CI. This facilitates the work of agents for iterating over docs quickly
- Create `make` commands for common checks (e.g., type check snippets)
- Added `mdsmith` for markdown linting, formatting and more (formatting will be done in a follow-up PR to make the diff more manageable)
- Simplify the GHA workflows and leverage caching more

### Snippet processing
- snippet formatting, linting, type checking and testing is done with `pytest-examples`
- custom directive in the code fence allow to opt-in / opt-out (`execute`, `notype`, `nolint`)
- Removed unused/orphaned snippets
- Using inline variable and type definitions should improve user experience, LLM training data, and agents using our docs as input

### Examples / Cookbooks
- Remove dynamically transforming `.py` file into `.md`. The code from markdown snippet is tested directly. This also allows to define the sidebar determnistically
- Temporarily disable examples / cookbooks. They should be checked in CI via a separate path from the main "check docs" because they required secrets, API keys, and make network calls. This has been a frequent cause of failures in the past. Using `pytest-examples` and `prek` it will be easier to "only run examples that changed"
- Remove the examples archive (those have been untested for a while)

## Future work
- enable more rules in `mdsmith` for markdown linting and formatting
- remove `notype` directive from snippets and fix them
- re-enable the cookbook examples that require live services in a separate CI branch (only run on edits)
- enable grammar checker hook via `harper-cli`. This will require making many manual fixes to current docs

### Optional / ideas
- look at opportunities to simplify the JS/TS tooling 
- centralize the content of `docs/Makefile` to the root `Makefile`. This would require some renaming, but could reduce sprawl
- Add more docstring code snippets example in the main `dlt` code and use `pytest-examples` to run them
- Do we need separate `docs_build` and `docs_check` workflows? Now that they are much faster, it probably makes sense to centralize. Having a valid docs build with invalid content doesn't seem too useful


Travior:
https://github.com/dlt-hub/dlt/blob/c8247c08f9a2c46a2e0465a78c3ef6173e472340/docs/website/tools/update_versions.js#L160-L162

This passes right now because we fetch clean `master` but this PR renames the make target to `process-docs`.

(also "normal" comment because I can't add untouched files to a review 🙄)

docs/test_snippets.py
```diff
@@ -0,0 +1,146 @@
...
+) -> None:
+    """Run docs snippets and standalone examples.
+
+    Snippets (`snippets` marker) and examples (`examples` marker) share the same
+    execution logic and can be selected independently, e.g. `pytest -m examples`.
+
+    If shared_state is True, share the state between individual snippets
+    of the same docs page.
+    """
+    page_path = WEBSITE_DOCS_DIR / page_relative_path
```
Travior, line 128: is this necessary? i think if I properly pass the relative path from invocation `page_relative_path` should be the correct path right? If so we could also rename the param to `page_path`?
This mean `--pages` is currently broken

docs/Makefile
```diff
@@ -35,36 +54,42 @@ lint-notebooks: ## Lints the notebooks in the education folder
...
-	uv run preprocess-docs --watch
+# Order matters: preprocess-markers wipes and recreates docs_processed, so
+# generators that write into it must run afterwards.
+process-docs: ## Build website/docs_processed from sources (markers + API reference)
+	$(MAKE) preprocess-markers
+	$(MAKE) generate-api-ref
 
-test-preprocess-molabs: ## Tests functions used to build Molabs
-	uv run pytest docs_tools/education/tests
+test-tools: ## Tests docs tools
```
Travior, line 73: are we running this somewhere? It's broken for me right now:
```
============================================================================ ERRORS =============================================================================
______________________________________________________ ERROR collecting tools/test_preprocess_to_molab.py _______________________________________________________
ImportError while importing test module '/Users/benjamin/reviews/dlt-hub/dlt/pr-4420/docs/tools/test_preprocess_to_molab.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../../../.local/share/uv/python/cpython-3.12.11-macos-aarch64-none/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tools/test_preprocess_to_molab.py:2: in <module>
    from docs_tools.education.preprocess_to_molab import (
E   ModuleNotFoundError: No module named 'docs_tools'
======================================================================= 1 error in 0.93s ========================================================================
```

docs/Makefile
```diff
@@ -6,26 +6,45 @@ help: ## Shows this help message
...
+
+cli-check: ## Checks CLI reference docs are up to date (CI)
+	uv --quiet run --script tools/generate_cli_docs.py website/docs/reference/command-line-interface.md --compare
+
 # Generated example test files (docs/examples/**/test_*.py) are gitignored and
-# created by prepare-examples-tests, so exclude them from mypy
+# created by prepare-examples-tests, so exclude them from mypy.
+# generate_api_reference.py declares its deps inline (PEP 723) and runs in its
+# own env, so they are not resolvable here
 lint-core: ## Lints the docs tooling, non-embedded snippets, examples and notebooks
```
Travior, line 43: same as `test-tools`

Travior:
I think we also need to update the source code links in the cookbook examples

Travior:
https://github.com/dlt-hub/dlt/blob/c8247c08f9a2c46a2e0465a78c3ef6173e472340/.claude/rules/documentation.md?plain=1#L23-L32

The Claude rule is also out of date now

Travior (changes requested):

docs/Makefile
zilto: I included this in `.pre-commit-config.yaml` to run with other hooks locally and in CI

docs/Makefile
zilto: fixed the import path

docs/test_snippets.py
zilto: You're right. I will remove


### Context

In previous iterations, `prek` was responsible for passing `.md` file paths. This allowed to only process snippets that were recently updated. Though, this created conflicts with `pytest-xdist` and was not worth the complexity.

zilto (commented):
## Summary of changes since last review
I needed to rebase from `devel` to solve merge conflicts. This breaks "see changes since last review" and rewrites history. You should reviewing from commit `62089fe`

### Changes
- brought back files for the `chess` dbt project. This was used by `tests/` so I moved the files there
- remove the hardcoded links in the `examples/` entries. The `.md` page now include the full source code.
- revert `process-docs` back to `preprocess-docs` to avoid issues with `update_versions.js` script
- enable snippet testing for `examples/` that only run locally without external service or secrets
- remove unused custom `pytest --pages` flag
- fix logic bug that was skipping snippets in `test_snippets.py::test_snippets_run`
- update Claude Code rules

docs/test_snippets.py
```diff
@@ -0,0 +1,142 @@
...
+    the dbt runner snippet expects `profiles.yml` next to `dbt.md`).
+    """
+    secret_dir = str(WEBSITE_DOCS_DIR / ".dlt")
+    config_dir = str(page_dir / ".dlt")
+
+    def _initial_providers(self: Any) -> list[Any]:
+        return [
+            EnvironProvider(),
+            SecretsTomlProvider(settings_dir=secret_dir),
+            ConfigTomlProvider(settings_dir=config_dir),
```
Travior: i think we resolve the wrong path if we switch the working directory below

docs/tools/preprocess_to_molab.py
Travior, line 1: we need to update the path traversal here to account for the move of the file (e.g. remove one .parent)

https://github.com/dlt-hub/dlt/blob/942c3861a04a447d212785c8db9940728c0ee97c/docs/tools/preprocess_to_molab.py#L8

Travior (changes requested):
Found two more things. The main question we should probably answer is what quality level we want from the cookbooks/examples. A couple of them were broken before and are now marked as `noexecute`, I’d suspect that this will probably stay until the end of time 😅
If that’s fine then this looks quite good to me, if not then we need to revisit a lot of the examples

docs/tools/preprocess_to_molab.py
zilto: good catch. The `make` command was successfully checking the existing `marimo` notebook but was failing to recreate them from the `.ipynb`

docs/test_snippets.py
zilto: fixed!

zilto:
(I had to rebase and force push to resolve merge conflicts again)

>The main question we should probably answer is what quality level we want from the cookbooks/examples. A couple of them were broken before and are now marked as noexecute, I’d suspect that this will probably stay until the end of time 😅 If that’s fine then this looks quite good to me, if not then we need to revisit a lot of the examples

## Answer
I would move forward with the current PR and fix examples in a follow-up. This PR focuses on the docs build tooling (rather than content) and dropping the tests for examples from the "main development loop" is intentional. In their current form, examples can't easily run locally (requires secrets, live services), are leaky (recurrent OSS maintenance cost) and slow (~8min on `devel` to test the same examples on every commit despite not being changed for 2+ years)

(note that examples are still linted, formatted, and type-checked. This will catch many dlt-related regressions)

### Details
- Out of the existing examples, only a few of them were actually tested. Only the following examples are being changed to "untested": `chess_production.md`,`custom_destination_bigquery.md`,`custom_naming.md`,`google_sheets.md`,`nested_data.md`,`pdf_to_weaviate.md`,`postgres_to_postgres.md`.
    - Some could easily be adapted to not require live services. For example, `nested_data.md` requires a live MongoDB instance just to show normalization process. Testing a custom MongoDB source seems completely out-of-scope for `dlt-hub/dlt` (it even diverges from `dlt-hub/verified-sources` source)
- Most examples are low traffic and the maintenance cost is high
- We will want to curate the examples / cookbook and many are expected to change or be deprecated
- We agreed that the examples / cookbooks are worth keeping (can be found by search, part of LLM corpus) even if less thoroughly tested
- If the examples are useful tests for `dlt`, they should be moved to `tests/`

zilto:
The `docs_check` was failing because of what seems to be a bug in `marimo check`. It would move the comments `#` to the wrong line. I edited the `ipynb` content to avoid facing this issue

edit: there could be an overlap with the `black` version used. The main error was the formatting of a tuple assignment, which seems to have changed between `black` `<=23` and `>=24`

docs/website/docs/examples/custom_config_provider.md
```diff
@@ -17,8 +18,9 @@
 * Instantiate custom provider (CustomLoaderDocProvider) from the loader
 * Register provider instance to be queried
 
-"""
+### Full source code
 
+```py
```
Travior: I think this can be `py execute`

```suggestion
```py execute
```

Travior (dismissed):
Direction around the cookbook examples makes sense, found one more snippet that was previously linted and passes locally. Otherwise LGTM

docs/website/docs/examples/custom_config_provider.md
zilto: Fixed!

Travior (approved):

PR #4435, opened 2026-09-08
fix(rest-client): fall back to other stop conditions on null total with default total_path

### LINKED ISSUE #4429: REST API: OffsetPaginator raises on null total even when stop_after_empty_page is set
### dlt version

1.30.0

### Describe the problem

`RangePaginator.update_state` raises when `total_path` resolves to `None`, even when `stop_after_empty_page` is `True` and would have ended the walk correctly.

`total_path` defaults to `"total"` on `OffsetPaginator` and `PageNumberPaginator`, and `stop_after_empty_page` defaults to `True`. So for an API whose response has no `total`, or has `"total": null`, the paginator raises on the first non-empty page unless you know to pass `total_path=None`.

`"total": null` is the API saying it does not know the total, which is different from a response that left the field out by mistake. We hit this on a vendor that returns a number for `total` on some endpoints and `null` on others, with the same envelope on both.

The part that costs time is that `stop_after_empty_page=True` is both the default and a complete stop condition, but it does not prevent the raise. The two settings look like they overlap and they do not.

### Expected behavior

If `total_path` gives no usable value but another stop condition is set (`stop_after_empty_page`, `maximum_value` or `has_more_path`), use that and log a warning instead of raising.

If raising is preferred, the error should mention `total_path=None` as the fix. That is not obvious from the current message.

### Steps to reproduce

```py


class Response:
    @staticmethod
    def json():
        return {"data": [{"id": 1}], "total": None}


paginator = OffsetPaginator(limit=100, stop_after_empty_page=True)
paginator.update_state(Response(), data=[{"id": 1}])
```

```
ValueError: Total `items` not found in the response in `OffsetPaginator`.
Expected a response with a `total` key, got `{'data': [{'id': 1}], 'total': None}`.
```

Passing `total_path=None` works as expected.

### Operating system

macOS

### Runtime environment

Local

### Python version

3.12

### dlt data source

REST API

### dlt destination

_No response_

### Other deployment details

_No response_

### Additional information

_No response_

bunnysayzz:
Fixes #4429.

A `null` (or absent) total with the default `total_path` no longer raises: following the design agreed in the issue, the default is now a sentinel instead of the literal `"total"`, so `RangePaginator` can tell an inherited default from an explicitly passed path.

- inherited default + missing total: warn once, ignore the total, and let the other stop conditions (`stop_after_empty_page`, `maximum_value`, `has_more_path`) end the walk;
- explicit `total_path` + missing total: still raises on page one (a mistyped path on an API that clamps page numbers must stay loud), and the error now points at `total_path=None` as the escape hatch;
- constructing a range paginator with no usable stop condition at all fails fast in `__init__`.

The sentinel is normalized to `"total"` in `__init__` and never stored, so paginators stay trivially deep-copyable (`client.paginate` deep-copies the paginator).

Tests: new cases in `tests/sources/helpers/rest_client/test_paginators.py` for inherited-null, inherited-missing, explicit-null, explicit-missing, empty-page stop, and fail-fast construction, plus `PageNumberPaginator` coverage. Three existing tests that asserted the old raise for default-constructed paginators now pass `total_path` explicitly. Full file green (109 passed); `black`, `flake8`, and `mypy` clean on both touched files.

### DIFF
--- a/dlt/sources/helpers/rest_client/paginators.py
+++ b/dlt/sources/helpers/rest_client/paginators.py
@@ -6,7 +6,18 @@ from urllib.parse import urljoin, urlparse
 from requests import Request, Response
 
 from dlt.common import jsonpath
+from dlt.common import logger
 from dlt.common.utils import str2bool
+
+
+_DEFAULT_TOTAL_PATH: Any = object()
+"""Sentinel marking the default `total_path` of range paginators.
+
+Lets `RangePaginator` tell an inherited default apart from an explicitly
+passed path: a response without a total only raises in the latter case.
+The sentinel is normalized to `"total"` in `__init__` and never stored,
+so paginators stay trivially copyable.
+"""
 
 
 class BasePaginator(ABC):
@@ -133,8 +144,14 @@ class RangePaginator(BasePaginator):
                 Defaults to None.
         """
         super().__init__()
+        # An inherited default is not a reliable stop condition: the response
+        # may simply not carry a total. Only an explicitly passed `total_path`
+        # counts as one for the validation below.
+        self._total_path_explicit = total_path is not None and total_path is not _DEFAULT_TOTAL_PATH
+        if total_path is _DEFAULT_TOTAL_PATH:
+            total_path = "total"
         if (
-            total_path is None
+            (total_path is None or not self._total_path_explicit)
             and maximum_value is None
             and has_more_path is None
             and not stop_after_empty_page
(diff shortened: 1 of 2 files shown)


dlt/sources/helpers/rest_client/paginators.py
```diff
@@ -548,7 +587,7 @@ def __init__(
             )
 
         if (
-            total_path is None
+            (total_path is None or total_path is _DEFAULT_TOTAL_PATH)
```
lucasmfraser, line 590: This also makes `OffsetPaginator(stop_after_empty_page=False)` raise at
construction, where today it constructs fine and stops on the total at the
default key. Just checking that was intentional alongside the null-total fix?
Same applies to PageNumberPaginator above.

lucasmfraser (commented):
Nice work, looks good overall to me with the sentinel and one-time warning. Just added a minor comment.

dlt/sources/helpers/rest_client/paginators.py
bunnysayzz: Good spot, yes, intentional. With the sentinel design the inherited default cannot count as a stop condition, otherwise a response with no total and no other stop condition would walk forever once the runtime fallback clears the path. So OffsetPaginator(stop_after_empty_page=False) with nothing else now fails fast at construction. The migration is one word: pass total_path='total' explicitly if the API returns the total under the default key. I added that hint to all three constructor errors plus tests pinning both the raise and the explicit-path construction. 112 paginator tests green in this push.

PR #4436, opened 2026-09-08
docs(dataset): align where/filter docstring params with the signature

simpleqt:
`Relation.where()` and `Relation.filter()` (both in `dlt/dataset/relation.py`) document a `column_name (str)` parameter — the actual parameter is `column_or_expr: SqlglotExprOrStr`. The two Args entries now match. Small self-contained docs correction (exempt from the issue requirement per the contributing guide).

### DIFF
--- a/dlt/dataset/relation.py
+++ b/dlt/dataset/relation.py
@@ -918,7 +918,7 @@ class Relation(WithSqlClient):
         This is identical to `Relation.filter()`.
 
         Args:
-            column_name (str): The column to filter on.
+            column_or_expr (str | sqlglot.Expression): The column or SQL expression to filter on.
             operator (TFilterOperation): The operator to use. Available operations are: eq, ne, gt, lt, gte, lte, in, not_in
             value (Any): The value to filter on.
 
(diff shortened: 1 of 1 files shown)


dlt/dataset/relation.py
```diff
@@ -918,7 +918,7 @@ def where(
         This is identical to `Relation.filter()`.
 
         Args:
-            column_name (str): The column to filter on.
+            column_or_expr (str | sqlglot.Expression): The column or SQL expression to filter on.
```
anuunchin, line 921: Let's do `column_or_expr (SqlglotExprOrStr):..`

dlt/dataset/relation.py
```diff
@@ -998,7 +998,7 @@ def filter(  # noqa: A003
         This is identical to `Relation.where()`.
 
         Args:
-            column_name (str): The column to filter on.
+            column_or_expr (str | sqlglot.Expression): The column or SQL expression to filter on.
```
anuunchin, line 1001: same here

anuunchin (changes requested):

simpleqt:
Done — both docstring entries now use the `SqlglotExprOrStr` alias (`dae36e5`), matching the signatures on `where`/`filter`. Thanks for the pointer.

PR #4444, opened 2026-09-09
fix(providers): serialize vault cache resolution

### LINKED ISSUE #4443: VaultDocProvider can report existing secrets as missing during concurrent resolution
### dlt version

The affected code is present on `devel` at commit `a1c530114cc347496d1f00f38891475a047b6d05`.

### Describe the problem

`VaultDocProvider` can return `None` for an existing secret when two threads resolve the same key concurrently.

In [`_update_from_vault`](https://github.com/dlt-hub/dlt/blob/a1c530114cc347496d1f00f38891475a047b6d05/dlt/common/configuration/providers/vault.py), the lookup is marked complete before the fetched value is stored in the document cache:

```python
secret = self._look_vault(full_key, hint)
self._vault_lookups[full_key] = pendulum.now()
if secret is not None:
    self.set_fragment(key, secret, pipeline_name, *sections)
```

There is no synchronization protecting this sequence. If thread A pauses after setting `_vault_lookups[full_key]` but before publishing the value, thread B skips the lookup because the key is already in `_vault_lookups`. It then reads the still-empty document and returns `None`.

This affects the shared vault-provider implementation, independently of any particular secrets backend. A caller resolving a required secret can consequently receive a missing-configuration error even though the secret exists.

### Expected behavior

Both concurrent callers should receive the existing value. A lookup in progress should not be treated as a completed lookup with no value.

### Steps to reproduce

The following standalone example uses an in-memory vault, dummy data, and no network requests. Events deliberately expose the cache-publication window:

```python
from concurrent.futures import ThreadPoolExecutor
from threading import Event

from dlt.common.configuration.providers.vault import VaultDocProvider
from dlt.common.typing import TSecretValue

entered, release = Event(), Event()


class DemoVault(VaultDocProvider):
    @property
    def name(self):
        return "demo"

    def _list_vault(self):
        return {"password"}

    def _look_vault(self, full_key, hint):
        return "dummy-secret"

    def set_fragment(self, key, value, pipeline_name, *sections):
        entered.set()
        if not release.wait(5):
            raise TimeoutError("test synchronization")
        super().set_fragment(key, value, pipeline_name, *sections)


provider = DemoVault(
    only_secrets=True,
    only_toml_fragments=False,
    list_secrets=True,
)

with ThreadPoolExecutor(max_workers=2) as pool:
    first = pool.submit(provider.get_value, "password", TSecretValue, None)
    try:
        assert entered.wait(5)
        second = pool.submit(provider.get_value, "password", TSecretValue, None)
        value, _ = second.result(timeout=2)
        print("Concurrent lookup incorrectly reports missing:", value is None)
    finally:
        release.set()
    value, _ = first.result(timeout=2)
    print("Original lookup succeeds:", value == "dummy-secret")
```

Output demonstrating the bug:

```text
Concurrent lookup incorrectly reports missing: True
Original lookup succeeds: True
```

The overridden `set_fragment` only pauses before delegating to the real implementation; it does not change the cached value or simulate a backend failure.

### Additional information

The lookup and document/cache publication need to be synchronized, including fragment loading. Simply moving the cache marker may not address all concurrent document updates or duplicate requests.

A regression test should release the first thread independently so a correctly blocked second reader can complete. It should verify that both readers obtain the value, while preserving existing behavior for genuinely missing optional keys and allowing recovery after lookup exceptions.

sdebruyn:
Concurrent vault resolutions can report an existing secret as missing or return a general fragment's value before its more specific override is loaded. Serialize resolution and lookup-cache clearing per provider, and mark a lookup complete only after its fragment has been stored. Retrieval, listing, and publication exceptions remain visible and allow a later call to retry; completed missing lookups remain cached.

Add in-memory regression tests for concurrent single values, global and pipeline documents, fragment precedence, cache clearing, missing keys, and failure recovery. Update the vault documentation to describe synchronization, caching, and backend-controlled retries. No backend-specific dependency is required.

Validation: all 27 vault-provider tests pass; the configuration suite has 301 passed and 11 skipped. Running the new cases against the unmodified upstream implementation reproduces 11 failures. Ruff, Black, and mypy for the changed files with silent import following pass. The broader mypy invocation reports three errors in unchanged runtime/storage files.

Fixes dlt-hub/dlt#4443

### DIFF
--- a/dlt/common/configuration/providers/vault.py
+++ b/dlt/common/configuration/providers/vault.py
@@ -2,6 +2,7 @@ import abc
 import contextlib
 import re
 import string
+from threading import Lock
 from typing import Any, Dict, Optional, Set, Tuple
 
 from dlt.common import logger
@@ -64,6 +65,7 @@ class VaultDocProvider(BaseDocProvider):
         self.only_secrets = only_secrets
         self.only_toml_fragments = only_toml_fragments
         self.list_secrets = list_secrets
+        self._vault_lock = Lock()
         self._vault_lookups: Dict[str, Any] = {}
         self._available_keys: Optional[Set[str]] = None
         if list_secrets and (only_toml_fragments or only_secrets):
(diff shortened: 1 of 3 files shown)


dlt/common/configuration/providers/vault.py
```diff
@@ -64,6 +65,7 @@ def __init__(
         self.only_secrets = only_secrets
         self.only_toml_fragments = only_toml_fragments
         self.list_secrets = list_secrets
+        self._vault_lock = Lock()
```
schrodervictor: We should use `RLock` here instead, to prevent dead locks caused by the same thread trying to re-acquire this lock (could happen in case of subclasses that use the vault mechanism itself).

docs/website/docs/general-usage/credentials/vaults.md
```diff
@@ -38,7 +38,11 @@ Fragments are TOML documents and are merged into the in-memory configuration in
 
 ### Caching
 
-Every successful or failed lookup (including “not found”) is cached for the lifetime of the process. Changes in the vault will not be picked up until the process restarts.
+Retrieved values and completed lookups (including “not found”) are cached by each provider instance. Changes in the vault are normally picked up when the process restarts. `clear_lookup_cache()` clears the lookup markers; it does not clear already loaded values or the list of available keys.
+
+Concurrent `get_value()` calls on the same provider are serialized, including vault requests and fragment merging. A caller waits for an ongoing resolution before reading the shared document, so it cannot mistake a pending lookup for a missing secret or read a partially merged fragment.
+
+Exceptions raised while listing, retrieving, or storing a fragment are propagated and do not mark that lookup as completed. A later call can try again. The shared vault provider does not add automatic retries on top of the backend SDK's retry policy. A backend response treated as “not found” (including permission errors that a provider handles this way) remains cached.
```
schrodervictor: Please trim this part to reduce verbosity and prevent documentation drift. The docs are supposed to give instructions to the technical user, but should not go too deep into implementation details. The source code remains as the authoritative source of truth when it comes to this level of detail.

schrodervictor (changes requested):
Thank you for your contribution @sdebruyn! The approach looks good to cover the situation described. We provided some feedback below to hammer out some final details. Please take a look when you can.

PR #4450, opened 2026-09-15
docs: enabled `mdsmith` formatting rules

zilto:
## Summary
Enable automatted and consistent formatting of markdown files.

PR #4420 introduced automated docs checks. This PR enables the majority of formatting rules for `mdsmith` and fixes pre-existing issues.

## Changes
For each commit on this PR, a rule or rule group was enabled and offending cases were fixed. The majority of changes are format-only and don't affect content (e.g., whitespaces, table, lists, admonitions). 

Some other rules required light content changes, such as:
- avoid large cells in tables required shortening some cells
- consistent heading hierarchies and avoiding "bold as heading" required rewriting some titles

## Decisions
There are two groups that I didn't enable: `line` and `prose`

- `line` has a single rule that sets the max length of a line. This can be useful, but it would trigger a huge diff and require "reflowing" all the text. During previous discussions, this check was divisive
- `prose` has useful checks such as "conciseness" and "readability" scoring. Enabling those should be discussed first  and they would require editing the content (rather than just the form)


docs/website/docs/general-usage/credentials/setup.md
```diff
@@ -3,8 +3,7 @@ title: Overview
 description: Learn where configs are stored and how to write them
 keywords: [credentials, secrets.toml, secrets, config, configuration, environment variables, provider]
 ---
-
-## Key features
+# Key features
```
burnash: i'm not sure if mdsmith converted this to h1, but this page already has a frontmatter title. Docusaurus does not make a title from the frontmatter when the page has an H1. It shows the H1 instead. 

This page now shows Key features as a title and "overview" in the menu:

Image

some number of pages also now specify h1. could be because of first-line-heading rule.

docs/Makefile
```diff
@@ -88,3 +88,9 @@ snippets-typecheck:
 
 snippets-test:
 	uv run pytest test_snippets.py::test_run_snippets
+
+md-check:  ## Check mardown file formatting errors in website/docs
```
burnash:
```suggestion
md-check:  ## Check markdown file formatting errors in website/docs
```

burnash (changes requested):

docs/website/docs/general-usage/credentials/setup.md
zilto: good find! So `mdsmith` doesn't autofix these issues. I had applied to script:

1. if there's no level 1 heading, set an heading 1 matching the frontmatter
2. if this is an `index.md`, set the heading 1 as `Overview`

It seems that `2.` was a bad heuristic because we have some flat `/docs/page_name/index.md`. 

OTOH: I know heading 1 is now redundant with frontmatter, but not all markdown renderers render the frontmatter. Also, having an explicit heading 1 allows us to check for valid heading hierarchies (e.g., we don't jump from heading 1 to heading 3)

burnash (approved):

burnash:
This is really good, thank you @zilto

PR #4465, opened 2026-09-19
Add clickhouse upsert

prevostc:
### Description

Add the `upsert` merge strategy on the ClickHouse destination.

ClickHouse `delete-insert` deletes matching keys from the live table, then inserts. That hole is visible to concurrent readers: ClickHouse does not run those two steps in a transaction. We do not want merge to mutate the live table that way.

This PR adds support for `upsert`, but rebuilds a copy instead to make the change happend-only from the pov of clickhouse engine:

1. create a new table with the destination structure
2. insert the load
3. append live rows whose primary keys are not in that load
4. `EXCHANGE TABLES`
5. drop the previous table

Keys omitted from the load are kept. The live table is unchanged until the swap.

Default `merge` (`delete-insert`) is unchanged. The same copy-then-swap path could later be used for `delete-insert` so that strategy also never deletes from the live table.

Tests cover generated SQL for both strategies. Docs list ClickHouse under `upsert`.

### Related Issues

- Related to #1628 (ClickHouse was excluded because it has no `MERGE INTO`)

### Additional Context

Requires an `Atomic`, `Replicated`, or `Shared` database engine, same as `staging-optimized` replace.

```py
write_disposition={"disposition": "merge", "strategy": "upsert"}
```

`primary_key` is required. `hard_delete` is supported. Nested tables: children of replaced parents are dropped; children of kept parents stay.

SQL generation and live ClickHouse destination tests pass locally, including `upsert` merge cases.


prevostc:
very slow for tables with large history, will use ReplacingMergeTree instead

PR #4469, opened 2026-09-21
(fix) Add missing typing for the last_trace property

VioletM:
fixes: Issue similar to https://github.com/dlt-hub/dlt/issues/4321 , but for `last_trace`

[pipeline.last_trace](https://github.com/dlt-hub/dlt/blob/devel/dlt/pipeline/pipeline.py#L1099) reports return type as PipelineTrace but can potentially [be None](https://github.com/dlt-hub/dlt/blob/devel/dlt/pipeline/trace.py#L349-L362).

Code snippet
```
import dlt

pipeline = dlt.pipeline()

# stuff, potentially not leading to a trace ...

if pipeline.last_trace:
    # do something with the last traceand run ty with check for redundant-condition:
```

type-check
```
2 ▲ a/div ❱ uv run ty check trace_ty_issue_example.py
warning[redundant-condition]: A 8-element tuple is always truthy
 --> trace_ty_issue_example.py:7:4
  |
7 | if pipeline.last_trace:
  |    ^^^^^^^^^^^^^^^^^^^ Inferred type is `PipelineTrace`

Found 1 diagnostic
112 ▼ a/div ❱ uv run dlt --version
```
dlt 1.30.0


zilto (approved):
LGTM

